PR #2679, opened 2022-02-08
General: Add settings for CleanUpFarm and disable the plugin by default

iLLiCiTiT:
## Brief description
It could be potentially dangerous to have CleanUpFarm plugin enabled by default https://github.com/pypeclub/OpenPype/pull/2390#issuecomment-1032811166.

## Changes
- added settings for `CleanUpFarm` plugin and disable the plugin by default


antirotor (approved):
👍🏻 disabled state is better as default

Issue #2620, opened 2022-02-01
fps rounding issue in validate_maya_units.py


m-u-r-p-h-y:
we did some math magic to round fps correctly because of the specific way ftrack was storing fps

this was changed apparently and now our rounding is causing troubles in Maya

```
 # openpype.hosts.maya.api.lib : Setting scene FPS to: '29.97fps'
.
// pyblish.ValidateMayaUnits : Units (time): 29.98 FPS // 
.
AssertionError: Scene must be 29.97 FPS(now is 29.98) //
```

ordinary python rounding should fix the problem
`fps = round(context.data['fps'], 2)`

[cuID:OP-2549]

antirotor:
Closed by #4424

PR #2680, opened 2022-02-08
Maya configurable unit validator

antirotor:
## Feature

this is adding configurable options for Maya Units validator.

image

Most of the options should be set later in different place - by something like scene defaults settings.

🗒️ There is also issue #2620 with fps decimal rounding that needs to be investigated between supported maya versions and ftrack.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/validate_maya_units.py
+++ b/openpype/hosts/maya/plugins/publish/validate_maya_units.py
@@ -30,8 +30,8 @@ class ValidateMayaUnits(pyblish.api.ContextPlugin):
     def process(self, context):
 
         # Collected units
-        linearunits = context.data('linearUnits')
-        angularunits = context.data('angularUnits')
+        linearunits = context.data.get('linearUnits')
+        angularunits = context.data.get('angularUnits')
         # TODO(antirotor): This is hack as for framerates having multiple
         # decimal places. FTrack is ceiling decimal values on
         # fps to two decimal places but Maya 2019+ is reporting those fps
@@ -39,7 +39,7 @@ class ValidateMayaUnits(pyblish.api.ContextPlugin):
         # rounding, we have to round those numbers coming from Maya.
         # NOTE: this must be revisited yet again as it seems that Ftrack is
         # now flooring the value?
-        fps = float_round(context.data['fps'], 2, ceil)
+        fps = float_round(context.data.get('fps'), 2, ceil)
 
         asset_fps = lib.get_asset()["data"]["fps"]
 
--- a/openpype/hosts/maya/plugins/publish/validate_maya_units.py
+++ b/openpype/hosts/maya/plugins/publish/validate_maya_units.py
@@ -37,6 +37,8 @@ class ValidateMayaUnits(pyblish.api.ContextPlugin):
         # fps to two decimal places but Maya 2019+ is reporting those fps
         # with much higher resolution. As we currently cannot fix Ftrack
         # rounding, we have to round those numbers coming from Maya.
+        # NOTE: this must be revisited yet again as it seems that Ftrack is
+        # now flooring the value?
         fps = float_round(context.data['fps'], 2, ceil)
 
         asset_fps = lib.get_asset()["data"]["fps"]
@@ -53,7 +55,8 @@ class ValidateMayaUnits(pyblish.api.ContextPlugin):
             and linearunits
             and linearunits != self.linear_units
         ):
-            self.log.error("Scene linear units must be centimeters")
+            self.log.error("Scene linear units must be {}".format(
+                self.linear_units))
             valid = False
 
         if (
@@ -61,7 +64,8 @@ class ValidateMayaUnits(pyblish.api.ContextPlugin):
             and angularunits
             and angularunits != self.angular_units
         ):
-            self.log.error("Scene angular units must be degrees")
+            self.log.error("Scene angular units must be {}".format(
+                self.angular_units))
             valid = False
 
         if self.validate_fps and fps and fps != asset_fps:
--- a/openpype/hosts/maya/plugins/publish/validate_maya_units.py
+++ b/openpype/hosts/maya/plugins/publish/validate_maya_units.py
@@ -19,6 +19,14 @@ class ValidateMayaUnits(pyblish.api.ContextPlugin):
     hosts = ['maya']
     actions = [openpype.api.RepairContextAction]
 
+    validate_linear_units = True
+    linear_units = "cm"
+
+    validate_angular_units = True
+    angular_units = "deg"
+
+    validate_fps = True
+
     def process(self, context):
 
         # Collected units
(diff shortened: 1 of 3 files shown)


openpype/hosts/maya/plugins/publish/validate_maya_units.py
```diff
@@ -37,26 +45,44 @@ def process(self, context):
...
+        ):
+            self.log.error("Scene linear units must be centimeters")
+            valid = False
+
+        if (
+            self.validate_angular_units
+            and angularunits
+            and angularunits != self.angular_units
+        ):
+            self.log.error("Scene angular units must be degrees")
```
BigRoy, line 64: Error message is incorrect - should be formatted with `self.angular_units`

openpype/hosts/maya/plugins/publish/validate_maya_units.py
```diff
@@ -37,26 +45,44 @@ def process(self, context):
...
-                                                        "must be degrees")
-        assert fps and fps == asset_fps, "Scene must be {} FPS"\
-                                         "(now is {})".format(asset_fps, fps)
+        # Check if units are correct
+        if (
+            self.validate_linear_units
+            and linearunits
+            and linearunits != self.linear_units
+        ):
+            self.log.error("Scene linear units must be centimeters")
```
BigRoy, line 56: Error message is incorrect - should be formatted with `self.linear_units`

openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_publish.json
```diff
@@ -129,6 +129,62 @@
...
+                          "type": "enum",
+                          "multiselection": false,
+                          "defaults": "cm",
+                          "enum_items": [
+                            {"mm": "millimeter"},
+                            {"cm": "centimeter"},
+                            {"m":  "meter"},
+                            {"in":  "inch"},
+                            {"ft": "foot"},
+                            {"yd": "yard"}
```
BigRoy, line 161: Not saying we REALLY need them all but it does seem like [the Maya command has more possible settings for linear units](https://download.autodesk.com/us/maya/2011help/Commands/currentUnit.html#flaglinear).

> Valid strings are:
> [mm | millimeter | cm | centimeter | m | meter | km | kilometer | in | inch | ft | foot | yd | yard | mi | mile]

Which means we are missing **(km) kilometer** and **(mi) mile**.
_(Not that one should ever mess with Maya's default units. But hey, some people like to be daredevils!)_

BigRoy (changes requested):
Logged messages are still wrong and we might be missing some linear unit options.

Note that also linear units and angular units are not set by default when a New Scene is created in [`on_new`](https://github.com/Colorbleed/colorbleed-config/blob/acre/colorbleed/maya/__init__.py#L186-L190) callback. It might be worth considering whether we'd like that too. I can imagine having your units ready from the get go is even better!

openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_publish.json
antirotor: strange thing as Maya (at least 2020) is not exposing them from UI:
image

openpype/hosts/maya/plugins/publish/validate_maya_units.py
antirotor: sure, that's why it is still draft. Putting kids to bed isn't compatible with finishing pull requests :) But when they are old enough, I'll let them do it :)

antirotor:
>  I can imagine having your units ready from the get go is even better!

True but I believe this is task for another PR. This was meant to quickly allow to publish something with non-standard stuff. Setting some pre-configured default on new scene is another beast and should probably include more thoughts and features :)

openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_publish.json
BigRoy: Even they weren't confident enough. I guess then we should ignore the other options. I'll mark this resolved.

BigRoy:
Should we refactor the usage of `context.data()` too? It would then be `context.data.get()` It just feels off against most of the rest of the code base. There is some more ancient `context.data()` and `instance.data()` in OpenPype which seems to appear in about five plug-ins in total.

Especially because the implementation was once for a ["temporary transition" in Pyblish according to the docstring](https://github.com/pyblish/pyblish-base/blob/deaaa9e8e157937cb0d47f0f45c80e28b2b89db7/pyblish/plugin.py#L667-L689) and the overriding logic for the key "name" seems extra confusing if you don't know about it..

BigRoy:
@antirotor I just tried setting to `kilometer` here in Maya 2022. It does set it. 

```python
from maya import cmds

cmds.currentUnit(linear="kilometer")
print(cmds.currentUnit(query=True))
# km
```

The preferences/settings in Maya does not show it, instead it shows "millimeter". I mean even at `0.01` I basically can't see the grid. The camera far clipping plane is `0.01` :') 
afbeelding

Whenever you start to try navigating in the viewport you'll be flying warp speed however.
I must admit - I wonder if this is EVER useful and whether OpenPype should expose it.

I'm a bit worried about the can of worms we'll open if someone ever sets the OP setting to kilometer and the vague reports we might be getting - especially with Maya not showing it by default in the UI.

antirotor:
I actually worked few times in kilometers but it was no joy. I would still leave it there as it is just validator and I see no reason to limit validating it, but of course when we do the scene settings I vote for omitting those not that much used units.

m-u-r-p-h-y (approved):
the validator is now configurable

tried to make it optional and published asset with validator turned off succesfully

openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_publish.json
```diff
@@ -129,6 +129,69 @@
...
+                          "label": "Linear units",
+                          "type": "enum",
+                          "multiselection": false,
+                          "defaults": "cm",
+                          "enum_items": [
+                            {"mm": "millimeter"},
+                            {"cm": "centimeter"},
+                            {"m":  "meter"},
+                            {"km":  "kilometer"},
+                            {"in":  "inch"},
```
BigRoy, line 165: Why do some of these entries have two spaces before the value? Were these entered manually? Or does the settings tool do that?

openpype/settings/defaults/project_settings/maya.json
```diff
@@ -188,18 +188,27 @@
...
             "validate_mesh": false,
             "validate_collision": true
         },
+        "ValidateCycleError": {
+            "enabled": true,
+            "optional": false,
+            "families": [
+                "rig"
+            ]
+        },
```
BigRoy, line 211: I've seen this before where some keys got shuffled on "save settings". Should we enable `sort_keys=True` on the JSON dump if it currently isn't yet? Or would that produce issues with dictionary values where users might expect the same "ordering" of how they set up the dictionary when they get back to it?

It's not a big issue, but it does stand out to me that sometimes the order just shifts around a bit.

openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_publish.json
antirotor: thats just probably my PyCharm forcing his idea of padding of json files.

openpype/settings/defaults/project_settings/maya.json
iLLiCiTiT: Note: Unrelated to this PR + there will always be some issues because some are modyfing defaults manually.

PR #2682, opened 2022-02-09
Ftrack: Hierarchical attributes are queried properly

iLLiCiTiT:
## Brief description
It was discovered that `ContextCustomAttributeValue` does return default value of custom attribute if entity does not have set value explicitly which is an issue for querying of hierarchical attribute values.

## Changes
- 'query_custom_attributes' has ability to return only explicitly set values or all values with defaults
    - added new argument `only_set_values` to define if `CustomAttributeValue` or `ContextCustomAttributeValue` table is used to get values
- almost all places that query custom attributes are using 'query_custom_attributes'

### Prereq
- Create new project with some structure
- Find one hierarchical custom attribute (e.g. `fps`) and set default value to something

### Sync to avalon - action
1. Set different then defaut value on project
2. Run sync to avalon
3. Check attribute values on assets (e.g. in project manager) they should be same as project value

### Sync to avalon - event
1. Run event server
2. Turn on auto sync on the project
3. Change `fps` on project
4. Check if the value is propagated in avalon and on all assets that should inherit the value

### Push hierarchical values
- this pushes changes from hierarchical custom attribute to non-hierarchical (and back for event)
1. Find/Create custom attribute that has both hierarchical and non-hierarchical version (e.g. `frameStart` on `Shot`)
2. Run ftrack event server
3. Change value on one of them
4. The value should be change on the other too
- same way it works for action but the value is propagated only from hierarchical to non-hierarchical

### DIFF
--- a/openpype/modules/default_modules/ftrack/event_handlers_server/action_push_frame_values_to_task.py
+++ b/openpype/modules/default_modules/ftrack/event_handlers_server/action_push_frame_values_to_task.py
@@ -2,7 +2,10 @@ import sys
 import json
 import collections
 import ftrack_api
-from openpype_modules.ftrack.lib import ServerAction
+from openpype_modules.ftrack.lib import (
+    ServerAction,
+    query_custom_attributes
+)
 
 
 class PushHierValuesToNonHier(ServerAction):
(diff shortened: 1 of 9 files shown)


mkolar:
Task linked: [OP-2610 Ftrack hiearchical attributes sync uses default values](https://app.clickup.com/t/6658547/OP-2610)

m-u-r-p-h-y (approved):
attributes are correctly propagated to child entities

image

PR #2691, opened 2022-02-10
Nuke: prerender with review knob

jakubjezek001:
## Brief description
Adding re-viewable option to prerender instance.

## Description
This will allow to publish reviewable components to Ftrack for example.

## Testing notes:
1. make sure the settings `project_settings/nuke/create/CreateWritePrerender/reviewable` is set to ON
2. make sure there are no project overrides on `project_settings/ftrack/publish/CollectFtrackFamily/profiles/6/families/2` and `prerender` family is available there
3. also make sure `project_settings/ftrack/publish/IntegrateFtrackInstance/family_mapping` is not having project overides and `prerender` key is available in mappings.
4. open your nuke testing workfile and create a prerender. 
5. review toggle should be set on and set for example some frame range constrain in rendering attribures on the prerender node.
6. publish and notice new version in ftrack with the prerender family

### DIFF
--- a/openpype/hosts/nuke/plugins/create/create_write_prerender.py
+++ b/openpype/hosts/nuke/plugins/create/create_write_prerender.py
@@ -98,13 +98,13 @@ class CreateWritePrerender(plugin.OpenPypeCreator):
                                    "/{subset}.{frame}.{ext}")})
 
         self.log.info("write_data: {}".format(write_data))
-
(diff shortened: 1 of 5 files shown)


mkolar:
Task linked: [OP-2588 Nuke prerender family with reviewable option](https://app.clickup.com/t/6658547/OP-2588)

PR #2698, opened 2022-02-11
Tweaks to Maya vrscene loader fixes PR

BigRoy:
Tweaks made as explained in [this comment](https://github.com/pypeclub/OpenPype/pull/2633#issuecomment-1028943913)

> - I did notice that it's possible to "delete" the node in the outliner. We might want to lock the "shape" node so that you cannot delete it. 

You cannot delete it now.

> - And also that the extra `_GRP` above the V-Ray Scene transform is a bit redundant because the VRayScene will never exist of multiple nodes, right? So potentially that could be removed.

Removed the extra _GRP

> - The group/root did not get colorized even though the code is there. That's because **vrayscene_layer** was not exposed in the Maya Loader Subset Outliner Colors settings. I've added those and matched the default color to `vrayproxy` color.

Fixed color settings

> - Since the Mesh was parented first to the Transform and then the V-Ray Scene Maya shows the mesh icon on the transform when collapsed in the hierarchy. I've re-ordered them so that the V-Ray Scene icon is shown. That will be much clearer for artists that they are not dealing with a 'regular mesh'.

Reordered dag order so that now Transform shows the VRayScene icon instead in outliner.


antirotor (approved):
Code looks ok, I'll need to check it somewhere with valid vray license.

Issue #2697, opened 2022-02-11
Settings: Allow searching (ctrl+F) functionality to browse settings


BigRoy:
### Feature

I noticed the Settings UI has a Breadcrumbs address bar that autocompletes to valid setting names. Which made me think it shouldn't be impossible to actually implement a 'search' within the settings as well!

I imagine the user clicking `CTRL + F` in the Settings UI or a little search icon next the breadcrumb bar. When doing so a little Search UI shows up where you can e.g. search `Shad` and it lists all entries that are related to "Shader"

```
project_settings/maya/publish/ValidateShaderName
project_settings/maya/publish/ValidateShadingEngine
```

This can be a rudimentary search just based on path, but potentially also on Label and Value. Once the basic functionality for the search is in of course it'll be more trivial to look into improving what you can query on.

I've needed this so many times, just to figure out where something was again. New users could do CTRL + F, type "Tools" and see they are in `Anatomy > Attributes > Tools`. In the little UI they could click on it and it'd set the path in the breadcrumbs address bar, and voila. The user is directly there.

#### Additional notes

Likely a little search UI like this is much more trivial than actually having a real-time search bar that filters the widgets inside the UI since the UI is relatively heavy on the amount of widgets.

I imagine this also reduces the need for #2686 because you'd already be able to identify specific things much faster. User goes CTRL + F, types "Maya" , presses down key to click through the list of search matches and clicks enter. The user is now directly at the right element in settings.

[cuID:25cq45q]

BigRoy:
Here's a functioning prototype. :)


https://user-images.githubusercontent.com/2439881/153628914-eced1388-cb14-4352-8639-5cfc667c8ab9.mp4

iLLiCiTiT:
> Here's a functioning prototype. :)

Where is the PR to merge?

PR #2699, opened 2022-02-11
Settings: Implementation for search in settings

BigRoy:
## Brief description

A prototype for #2697 

Press CTRL+F to open the search UI. Note that it is just a prototype. Might crash, have ugly code - just to show that it _could_ work! (Probably CMD+F on Mac)

Current state:

https://user-images.githubusercontent.com/2439881/153628914-eced1388-cb14-4352-8639-5cfc667c8ab9.mp4


openpype/tools/settings/settings/search.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def get_entity_children(entity):
+
+    if hasattr(entity, "values"):
+        return entity.values()
+    return []
+
+
+class RecursiveSortFilterProxyModel(QtCore.QSortFilterProxyModel):
```
iLLiCiTiT, line 16: `RecursiveSortFilterProxyModel` can be imported from `openpype.tools.utils`.

BigRoy: That's correct. But - this is a changed one. This filters to ALL columns and actually uses Qt 5.10+ feature to enable recursive filtering that Qt allows. Since I assumed the Settings tool would never be used in older Qt (e.g. in hosts with older Python, etc.)

In essence this class name should just be updated to match what it does instead - but I couldn't think of a good name. Maybe `FilterAllColumnsProxyModel`?

openpype/tools/settings/settings/categories.py
```diff
@@ -379,7 +381,12 @@ def set_category_path(self, category, path):
         """Change path of widget based on category full path."""
         pass
 
+    def change_path(self, path):
+        """Change path and go to widget."""
+        self.breadcrumbs_widget.change_path(path)
```
iLLiCiTiT, line 386:
```suggestion
        self.breadcrumbs_bar.change_path(path)
```

BigRoy:
Thanks for the changes. Works for me! ✅

iLLiCiTiT (approved):
Approved for me.

BigRoy:
@iLLiCiTiT should we maybe make the search case insensitive? :) Currently `JPEG or `jpeg` doesn't find `Jpeg`?

iLLiCiTiT:
> Currently JPEG or jpeg doesn't find Jpeg for example.

Oh, sh*t, Yes!

iLLiCiTiT:
Created PR https://github.com/pypeclub/OpenPype/pull/2810

PR #2708, opened 2022-02-12
Publish UI: Show states instead of processed plugin

iLLiCiTiT:
## Brief description
Do not show processing plugin but current publish state.

## Changes
- publish footer does not show just processed plugin but current publishing state
- removed fade out effect so the state is visible all the time

image


BigRoy:
Nice!

Some notes:
- If a Collector fails it currently ends with a label "Collected" which in essence is 'ok' but would it be better that if there were any errors at the end of collecting that it says: "Collected, with errors" or something?
- When validating ends due to failures the label remains "Validating..." instead it should say something like "Validation failed." or "Stopped, due to errors" or "Validated, with errors" (or whatever is short and reads like 'some things were not valid')
- If publishing finished (validations passed and publishing continued) but an Integrator failed then it should read: [Published, with errors.](https://github.com/pyblish/pyblish-qml/blob/103fef81938a8e4102dda831898d80c0ac2ba8de/pyblish_qml/control.py#L979) - preferably without the triple dots because that makes it look like it's still actively doing something.
- Typo: Published message currently reads "Pulished"


Here's a quick snippet to test that case with a failing Integrator - that label at the end should read "Published, with errors"
```python
import pyblish.api
import os

# Cleanup so no plug-ins are found from OpenPype
os.environ["PYBLISHPLUGINPATH"] = ""
pyblish.api.deregister_all_hosts()
pyblish.api.deregister_all_paths()
pyblish.api.deregister_all_targets()
pyblish.api.deregister_all_plugins()
pyblish.api.deregister_all_discovery_filters()
pyblish.api.deregister_all_callbacks()

# Register our own
class Collector(pyblish.api.ContextPlugin):
    order = pyblish.api.CollectorOrder
    
    def process(self, context):
        instance = context.create_instance("test")
        instance.data["family"] = "test"
        
class Integrator(pyblish.api.InstancePlugin):
    order = pyblish.api.IntegratorOrder
    families = ["test"]
    
    def process(self, instance):
        raise RuntimeError("Invalid.")
        
pyblish.api.register_plugin(Collector)
pyblish.api.register_plugin(Integrator)
```

openpype/tools/pyblish_pype/control.py
```diff
@@ -293,6 +297,7 @@ def load_plugins(self):
     def on_published(self):
         if self.is_running:
             self.is_running = False
+        self._current_state = "Pulished"
```
BigRoy, line 300: Typo: "Pulished" -> "Published"

64qam (approved):

BigRoy (changes requested):
Just marking this as "request changes" with regard to my earlier comments.

BigRoy (approved):

mkolar:
@iLLiCiTiT could you please resolve the conflict and merge?

PR #2711, opened 2022-02-14
Settings: Studio Project anatomy is queried using right keys

iLLiCiTiT:
## Brief description
Querying of studio project anatomy is not using right key to get data document.

## Changes
- fix keys which are used to get studio anatomy settings

## Testing notes:
1. Change and save values in studio default project anatomy
2. Restart process and look if the values are loaded


m-u-r-p-h-y (approved):
default project fps setting changed, saved, OP reloaded and fps is shown correctly now

PR #2713, opened 2022-02-14
Fusion: Moved implementation into OpenPype

iLLiCiTiT:
## Brief description
Moved implementation from avalon-core into openpype repository.

## Changes
- moved the code
- fixed imports to match new code locations
- moved prelaunch hook code from api to prelaunch hook
- removed unused scripts from avalon

## Testing notes:
1. Fusion should work as before (not sure if it worked?)

Related to PR https://github.com/pypeclub/avalon-core/pull/422


mkolar:
Task linked: [OP-2542 fusion](https://app.clickup.com/t/6658547/OP-2542)

BigRoy:
Thanks. At first glance the code changes look solid. I'll test this as soon as possible. What do we want to do with the hound errors?

BigRoy:
I've given this a go. And it works fine - has the same 'problems' that were there before the move so I think it's best to address them separately. I'll work towards a PR regarding that sometime soon. But in the basis the issues would be:

1. Loaders don't have "render" family set, and thus won't work for e.g. on Maya render outputs.
2. Fusion 17 isn't in the applications list currently.
3. The logic in the prelaunch hook checks whether `os.path.normpath({envvar})` is a directory, but if the env var was not set an empty string would be passed to `os.path.normpath` which returns the current working directory. As such the logic is not correctly checking whether the variable was set at all.
4. Fusion needs to be manually set to use Python 3 by default instead of Python 2. Otherwise the menu script will not work.
5. Python 3.6 will need to be installed locally by default.
5. The Python 3.6 locally installed will need to have PySide2 (or another Qt) installed for the tools to work.

I'll set up a quick PR to fix 1 and 2 as they seemed the more pressing issues.
I'm also working on a bigger refactor of the Fusion implementation on the side already so I'm thinking to try and address the rest separate from this PR since this is mostly just for the move.

BigRoy:
Updating a loader fails - fixed in my PR with [this commit](https://github.com/pypeclub/OpenPype/pull/2716/commits/4f9699fa000ceaba0679e5491e26419cac65d8f4).
```
  File "S:\openpype\OpenPype\openpype\tools\sceneinventory\view.py", line 275, in <lambda>
    lambda: self._show_version_dialog(items))
  File "S:\openpype\OpenPype\openpype\tools\sceneinventory\view.py", line 726, in _show_version_dialog
    api.update(item, version)
  File "S:\openpype\OpenPype\repos\avalon-core\avalon\pipeline.py", line 1563, in update
    return loader.update(container, new_representation)
  File "<string>", line 208, in update
  File "<string>", line 253, in _get_first_image
NotADirectoryError: [WinError 267] The directory name is invalid: \path\to\file.1001.exr'
```

iLLiCiTiT:
> I'll set up a quick PR to fix 1 and 2 as they seemed the more pressing issues.

3. Changed
4. Changed by removing `use_python_2` key from schemas
5. + 6. are nice features but not goal of this PR
Issue with templates seems to be more important than 5. and 6. but also not related to this PR.

jakubjezek001 (approved):
works suprisingly well ;)

BigRoy (approved):

PR #2729, opened 2022-02-15
General: Fix loading of unused chars in xml format

iLLiCiTiT:
## Brief description
Class `ElementTree` in xml parser don't know how to handle all escaped values which cause parse error.

## Description
Not sure what is proper fix. Propably would be to modify xml parser which is more complicated or define all possible espace values (e.g. from this [source](https://www.obkb.com/dcljr/charstxt.html)). It currently breaks loading of data for some exr.

## Changes
- replace `&` in some unused xml ampresand characters with `&amp;` so `ElementTree` can parse it

### DIFF
--- a/openpype/lib/transcoding.py
+++ b/openpype/lib/transcoding.py
@@ -197,6 +197,8 @@ def parse_oiio_xml_output(xml_string, logger=None):
     # Fix values with ampresand (lazy fix)
     # - oiiotool exports invalid xml which ElementTree can't handle
     #   e.g. "&#01;"
+    # WARNING: this will affect even valid character entities. If you need those values correctly, this must
+    # take care of valid character ranges. See https://github.com/pypeclub/OpenPype/pull/2729
     matches = XML_CHAR_REF_REGEX_HEX.findall(xml_string)
     for match in matches:
         new_value = match.replace("&", "&amp;")
--- a/openpype/lib/transcoding.py
+++ b/openpype/lib/transcoding.py
@@ -31,37 +31,9 @@ INT_TAGS = {
     "subimages",
 }
 
-XML_UNUSED_CHARS = {
-    "&#00;",
-    "&#01;",
-    "&#02;",
-    "&#03;",
-    "&#04;",
-    "&#05;",
-    "&#06;",
-    "&#07;",
-    "&#08;",
-    "&#11;",
-    "&#12;",
-    "&#14;",
-    "&#15;",
-    "&#16;",
-    "&#17;",
-    "&#18;",
-    "&#19;",
-    "&#20;",
-    "&#21;",
-    "&#22;",
-    "&#23;",
-    "&#24;",
-    "&#25;",
-    "&#26;",
-    "&#27;",
-    "&#28;",
-    "&#29;",
-    "&#30;",
-    "&#31;"
-}
+CHAR_REF_REGEX_DECIMAL = re.compile(r"&#[0-9]+;")
+CHAR_REF_REGEX_HEX = re.compile(r"&#x[0-9a-zA-Z]+;")
+
 # Regex to parse array attributes
 ARRAY_TYPE_REGEX = re.compile(r"^(int|float|string)\[\d+\]$")
 
@@ -226,10 +198,13 @@ def parse_oiio_xml_output(xml_string, logger=None):
     # Fix values with ampresand (lazy fix)
     # - ElementTree can't handle all escaped values with ampresand
     #   e.g. "&#01;"
-    for unused_char in XML_UNUSED_CHARS:
-        if unused_char in xml_string:
-            new_char = unused_char.replace("&", "&amp;")
-            xml_string = xml_string.replace(unused_char, new_char)
+    matches = (
+        set(CHAR_REF_REGEX_HEX.findall(xml_string))
+        | set(CHAR_REF_REGEX_DECIMAL.findall(xml_string))
+    )
+    for match in matches:
+        new_value = match.replace("&", "&amp;")
+        xml_string = xml_string.replace(match, new_value)
 
     if logger is None:
         logger = logging.getLogger("OIIO-xml-parse")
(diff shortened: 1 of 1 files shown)


BigRoy:
@iLLiCiTiT does this resolve itself when [parsing from a `unicode` string](https://stackoverflow.com/a/22256466/1838864)? Or see [this other topic about it](https://stackoverflow.com/a/62892284/1838864).

iLLiCiTiT:
Encoding is not an issue in this case. The issue is that one attribute has value with escaped xml value but `ElementTree ` can't handle that.

Example
```
<attrib name="tech_details_color_space" type="string">&#02;</attrib>
```
The `&#02;` should be hexadecimal string `0x02` but the `&` raises ParseError.

antirotor:
the only predefined character sequences in xml are:

| Name | Character | Unicode code point (decimal) | Standard | Name |
|---|---|---|---|---|
|quot|"|U+0022 (34)|XML 1.0|[quotation mark](https://en.wikipedia.org/wiki/Quotation_mark)|
|amp|&|U+0026 (38)|XML 1.0|[ampersand](https://en.wikipedia.org/wiki/Ampersand)|
|apos|'|U+0022 (39)|XML 1.0|[apostrophe](https://en.wikipedia.org/wiki/Apostrophe) (1.0: apostrophe-quote)|
|lt|<|U+003C (60)|XML 1.0|[less-than sign](https://en.wikipedia.org/wiki/Less-than_sign)|
|qt|>|U+003E (3E)|XML 1.0|[greater-than sign](https://en.wikipedia.org/wiki/Greater-than_sign)|

everything else is illegal except character and entity references:

```
'&#' [0-9]+ ';' | '&#x' [0-9a-fA-F]+ ';'
```
but:

**Well-formedness constraint: Legal Character**

[Characters referred to using character references must match the production for ]()[Char](https://www.w3.org/TR/2008/REC-xml-20081126/#NT-Char).

And that is defined as:

```
Char	   ::=   	#x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]	/* any Unicode character, excluding the surrogate blocks, FFFE, and FFFF. */
```
so `&#02;` is really invalid character in xml

antirotor (changes requested):
I am afraid this is not enough, see my comment. This must be solved with character ranges.

iLLiCiTiT:
> This must be solved with character ranges.

To be honest I don't know how. Right now this breaks extract review because xml export from oiiotool put `&#02;` into some irrelevant node which is not used. If you know how, do it.

BigRoy:
Maybe it makes more sense to rely on `BeautifulSoup` or `lxml` since they appear to have some possible solutions to this?

openpype/lib/transcoding.py
```diff
@@ -31,37 +31,9 @@
...
-    "&#24;",
-    "&#25;",
-    "&#26;",
-    "&#27;",
-    "&#28;",
-    "&#29;",
-    "&#30;",
-    "&#31;"
-}
+CHAR_REF_REGEX_DECIMAL = re.compile(r"&#[0-9]+;")
```
antirotor, line 34: you can use one regex to rule them all:
```&#x?(?P<char>[0-9a-fA-F]+);`

File: openpype/lib/transcoding.py [hunk 2]
```diff
@@ -226,10 +198,13 @@ def parse_oiio_xml_output(xml_string, logger=None):
     # Fix values with ampresand (lazy fix)
     # - ElementTree can't handle all escaped values with ampresand
```
antirotor, line 199: I think you are to harsh to ElementTree - it just wants valid xml

openpype/lib/transcoding.py
iLLiCiTiT: I wanted to add the ranges validations...

antirotor:
https://www.tjohearn.com/2018/01/24/safe-ampersand-parsing-in-xml-files/

> Only these are valid ranges? `#x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]`

this is in [xml specs](https://www.w3.org/TR/2008/REC-xml-20081126/#dt-character).

openpype/lib/transcoding.py
antirotor: sure, just that one will pick all character references no matter if they are hex or decimal and you can do the range check on `m.group("char")` then. You just need to make sure it is cast properly to number.

iLLiCiTiT: But I've modified it like you suggested here. I don't want be dependent on oiiotool "correctness" of xml.

iLLiCiTiT:
Modified to use regex which checks if xml from oiiotool contain valid values and replace ampresand of invalid values. The loaded value matches string value from xml text. These values are metadata that are not needed for us so I would not care about their real value until they're needed.

```
<attrib name="tech_details_color_space" type="string">&#02;</attrib>
```
Is loaded as node with `&#02;` text value.

openpype/lib/transcoding.py
iLLiCiTiT: > they are building the xml as a string

I though that this is the issue, hoped not.

openpype/lib/transcoding.py
```diff
@@ -191,6 +194,14 @@ def parse_oiio_xml_output(xml_string, logger=None):
     if not xml_string:
         return output
 
+    # Fix values with ampresand (lazy fix)
+    # - oiiotool exports invalid xml which ElementTree can't handle
+    #   e.g. "&#01;"
```
antirotor, line 199: I would just add explicit warning about what it does and point it to this discussions.

```suggestion
    #   e.g. "&#01;"
    # WARNING: this will affect even valid character entities. If you need those values correctly, this must
    # take care of valid character ranges. See https://github.com/pypeclub/OpenPype/pull/2729
```

antirotor (changes requested):
I would just add there comment what is a result of this - this will affect all character entities, even the valid one. This fix is quick and dirty, but I can imagine that someone will try in the future get something from these information and he'll hit it?

antirotor (approved):

PR #2731, opened 2022-02-15
Loader UI: Multiple asset selection and underline colors fixed

iLLiCiTiT:
## Brief description
Loader errors when it is clicked on a group with merge group in it and underline colors in assets widget are not cleared.

## Description
Code does not expect that group can have children item that is another group but expects it has version document which cause not updating of version info widget on click. Also underline color are not cleared at all righ now.

## Changes
- implemented missing part of clearing of underline colors in assets widget
- looping through group item expect that can have another group as children
- it is possible to show loaders on group items

## Testing notes:
1. Click on group with merged subsets in loader (multi asset selection with same subsets in same group is required)
2. Nothing should crash and should be possible to run loaders on the group
3. Underline colors should be cleared when clicked on different subset or subsets are deselected

Resolves https://github.com/pypeclub/OpenPype/issues/2728

### DIFF
--- a/openpype/tools/libraryloader/app.py
+++ b/openpype/tools/libraryloader/app.py
@@ -396,9 +396,7 @@ class LibraryLoaderWindow(QtWidgets.QDialog):
             self._versionschanged()
             return
 
-        selected_subsets = self._subsets_widget.selected_subsets(
-            _merged=True, _other=False
-        )
+        selected_subsets = self._subsets_widget.get_selected_merge_items()
 
         asset_colors = {}
         asset_ids = []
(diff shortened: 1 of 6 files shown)


jakubjezek001 (changes requested):
The problem with the error is still happening and now it is even worse. The error is popping up multiply. Before when I had closed the error and RMB clicked again to the subset group the context menu had finally open and I could load but now the error is poping up all the time. 

Here is how it looks 
image

iLLiCiTiT:
Should be fixed now.

jakubjezek001 (approved):
works well

PR #2733, opened 2022-02-16
Maya: set Deadline job/batch name to original source workfile name instead of published workfile

BigRoy:
## Description

When submitting from Maya to deadline with Work file publish enabled Deadline will render using the published work file. The deadline batch and job name were set using that filename.

The problem however was that a user could have a work file name `shot_v001_{subversion}` like `shot_v001_testfile` as current workfile. On submitting to deadline the Deadline job would however read as batch name and job name `shot_v001` which made it harder for artists to trace back their render to their work file and track e.g. some test subversions they submitted.

This change should always force the usage of the original work file name for Job + Batch name.

## Testing notes:

1. test a deadline submission from workfile with subversion string with Use Published enabled + workfile publishing enabled as well
2. test same settings, but from workfile without subversion string.


antirotor (approved):
Works, merging now. Thanks!

Issue #2626, opened 2022-02-02
Work Files: Preserve subversion part of filename on incrementing


BigRoy:
**Is your feature request related to a problem? Please describe.**

Some Artists are reporting to me that they felt it was off that OpenPype stripped off the 'subversion' suffix part of the filename when incrementing current work file when e.g. submitting the renders to Deadline.

Should the subversion part of the filename persist?

The same happens in the workfiles tool. My current file has a "subversion" part of the string, when clicking "Save" the subversion is empty and if I'd like to preserve the value I need to retype it manually. Should it by default maybe also set the current subverison string in there?

**Describe the solution you'd like**

Preserve as much of the user's file when incrementing workfile as possible.

**Describe alternatives you've considered**

This came up earlier too - where an artist was trying to keep two branches simultaneously. E.g. a special scene with a `_smoke` suffix that is similar to the original lighting scene but optimized and dedicated to rendering only a very small subset with a potentially different renderer too. For that case I've recommened making another `lighting_smoke` task if they want to keep two of those branches simultaneously. Is that what you would recommend too?

[cuID:OP-2560]

BigRoy:
Actually the incrementing of the file for deadline submission doesn't use the same logic as the Work Files tool. Instead that uses [`openpype.lib.path_tools.version_up`](https://github.com/pypeclub/OpenPype/blob/4ce12ffc7c2bf0a406f6cd1bb558e06b27e283f2/openpype/lib/path_tools.py#L22-L74). So we could also only patch that - but somehow I feel both cases are relevant to this issue even though they can be fixed separately.

antirotor:
Connected to this is also request to have dropdown menu in workfile tool with all subversion string so you could select the one that's already there without typing it manually every time you want to save. This is probably more issue for them because it doesn't preserve the subversion.

BigRoy:
> Connected to this is also request to have dropdown menu in workfile tool with all subversion string so you could select the one that's already there without typing it manually every time you want to save. This is probably more issue for them because it doesn't preserve the subversion.

Ah, that would be nice too! :)

Issue #2719, opened 2022-02-14
Workfiles: add dropdown to select existing subversion string


antirotor:
## Feature

in **Workfiles**, you can specify subversion string when saving file. This is arbitrary string. It would help to be able to select from existing in dropdown menu then writing every time from the scratch.

![image](https://user-images.githubusercontent.com/33513211/153887810-815b1a0c-32e8-45ef-9cd1-abb9fb28c93e.png)

[cuID:25xn3bh]

tokejepsen:
Or maybe preserve it from previous versions; https://github.com/pypeclub/OpenPype/issues/2626

PR #2734, opened 2022-02-16
Work Files: Preserve subversion comment of current filename by default

BigRoy:
## Brief description

This is an implementation for #2626 where clicking on Save in the Work Files tool the subversion field will be prefilled with what is assumed to be the current 'comment' in the filename. 

It uses regex so is always prone to error I suppose. Howevr this case should be a pretty accurate detection against the filename template because all other "data" of the current context is formatted like also used for making a new save version- then only the {comment} and {version} are parsed against the current filename.

If the regex cannot detect any comment or current filename can't be matched against the template then nothing changes. Comment field will remain empty.

## Testing notes:
1. try and break the logic
2. test against some unique filename templates used by OP teams

### DIFF
--- a/openpype/tools/workfiles/app.py
+++ b/openpype/tools/workfiles/app.py
@@ -1,5 +1,6 @@
 import sys
 import os
+import re
 import copy
 import getpass
 import shutil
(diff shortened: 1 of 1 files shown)


BigRoy:
With https://github.com/pypeclub/OpenPype/pull/2734/commits/18f8741224d06e6aee41c924a9ad7796810072cd this now also implements #2719 

Admittedly the code might benefit from some cleanup - moving widgets to a `widgets.py` or alike. But it's functional at least. :)

openpype/tools/workfiles/app.py
```diff
@@ -37,6 +38,182 @@
...
+        if not self.fname_regex:
+            return
+
+        fname = os.path.basename(filepath)
+        match = self.fname_regex.match(fname)
+        if match:
+            return match.group(1)
+
+
+class SubversionLineEdit(QtWidgets.QWidget):
```
BigRoy, line 151: This is basically similar to the "subset" line edit as in the Creator tool.
Wasn't sure how to implement this simple widget with less code than this.

openpype/tools/workfiles/app.py
```diff
@@ -37,6 +38,182 @@
...
+        menu = self._menu
+        button = self._button
+
+        state = any(values)
+        button.setEnabled(state)
+        if state is False:
+            return
+
+        # Include an empty string
+        values = [""] + sorted(values)
```
BigRoy, line 201: Empty string is just there so you can also quickly clear the field when checking the dropdown menu.

antirotor:
It doesn't get this:
image

I thought there should be `test` in subversion textfield?

BigRoy:
It should indeed!

What is your project's file template?

BigRoy:
Ok. Found it. It's a Py2 <-> Py3 difference in how regex escapes.

In Python 2 a lot more characters are actually being escaped with `re.escape()`. I've fixed the logic.
Can you try again @antirotor?

antirotor (approved):
That did the trick! Working fine.

Issue #2743, opened 2022-02-16
Maya: work file not saved before publish


m-u-r-p-h-y:
When publishing in Maya without saving the workfile first, there is an issue with copying the wrong workfile to publish.

We should **always** save the workfile before publish.

Is there any reason for this filtering in 
OpenPype\openpype\hosts\maya\plugins\publish\save_scene.py ?

[cuID:OP-2668]

BigRoy:
> Is there any reason for this filtering in OpenPype\openpype\hosts\maya\plugins\publish\save_scene.py ?

I think this might have been speed constraints in our older pipeline - saving could just take long. [It looks like a close to copy of our studio config](https://github.com/Colorbleed/colorbleed-config/blob/acre/colorbleed/plugins/maya/publish/save_scene.py#L14). And we just loved living on the edge.

There shouldn't be any reason not to in general, but the use case you're describing would already work by adding [`"workfile"` family to this SaveScene plug-in](https://github.com/pypeclub/OpenPype/blob/926eb5b95ece6feb27957c04633bdfb3785e9c7a/openpype/hosts/maya/plugins/publish/save_scene.py#L12)

Also, we could add checking for modifications as in [our plugin](https://github.com/Colorbleed/colorbleed-config/blob/acre/colorbleed/plugins/maya/publish/save_scene.py#L14) - that at least skips saving if the user saved just before publishing already.

BigRoy:
> (but we should not do it twice before and after publish)

I've set up PR #2744 but that excludes this particular point. It is an optimization but also prone to error.

Technically we could **store** some data during Save Scene like `context.data["sceneSavedDuringPublish"] = True` and then instead of saving a new version in [IncrementCurrentFile](https://github.com/pypeclub/OpenPype/blob/926eb5b95ece6feb27957c04633bdfb3785e9c7a/openpype/hosts/maya/plugins/publish/increment_current_file_deadline.py#L33-L40) we could e.g. do:

```python
# psuedocode
    # Rename current file so that Maya knows the new filepath
    cmds.file(rename=new_filepath)
    
    if context.data.get("sceneSavedDuringPublish", False):
        # Copy file without saving to optimize
        shutil.copyfile(current_filepath, new_filepath)
    else:
        # Save with the correct file type associated with extension
        if new_filepath.endswith(".ma"):
            fileType = "mayaAscii"
        elif new_filepath.endswith(".mb"):
            fileType = "mayaBinary"
    
[...]

PR #2744, opened 2022-02-16
Maya: Save current scene on workfile publish

BigRoy:
## Brief description

Implementation for #2743 

## Testing notes:
1. make a workfile save it, change something drastic (**DONT SAVE!**), publish your file
2. ensure the published workfile contains the drastic change

### DIFF
--- a/openpype/hosts/maya/plugins/publish/save_scene.py
+++ b/openpype/hosts/maya/plugins/publish/save_scene.py
@@ -17,5 +17,11 @@ class SaveCurrentScene(pyblish.api.ContextPlugin):
         current = cmds.file(query=True, sceneName=True)
         assert context.data['currentFile'] == current
 
+        # If file has no modifications, skip forcing a file save
+        if not cmds.file(query=True, modified=True):
+            self.log.debug("Skipping file save as there "
+                           "are no modifications..")
+            return
+
         self.log.info("Saving current file..")
         cmds.file(save=True, force=True)
--- a/openpype/hosts/maya/plugins/publish/save_scene.py
+++ b/openpype/hosts/maya/plugins/publish/save_scene.py
@@ -9,7 +9,7 @@ class SaveCurrentScene(pyblish.api.ContextPlugin):
     label = "Save current file"
     order = pyblish.api.IntegratorOrder - 0.49
     hosts = ["maya"]
-    families = ["renderlayer"]
+    families = ["renderlayer", "workfile"]
 
     def process(self, context):
         import maya.cmds as cmds


m-u-r-p-h-y (changes requested):
It works as expected.

The question is, if the timing in the que is right, saving the file after running the extractors, alembics etc.

Shouldn't we make sure we save the scene before any possible changes are done down the publishing line?

BigRoy:
>The question is, if the timing in the que is right, saving the file after running the extractors, alembics etc.
> Shouldn't we make sure we save the scene before any possible changes are done down the publishing line?

Extractors should (or **must**?) be build to be atomic operations that leave the current scene unaltered, otherwise the User's scene would break anyway since they don't reopen the scene. Technically you are correct that it might be safer from a technical standpoint to save before we're trying to do anything that might interfere with the current scene - but honestly, if you'd need to rely on that you will also always need to reopen the scene after a publish.

We could move the Save before the Extraction - but if indeed the local scene is broken then yes, the user usually continues there work with a broken scene too. I feel like we're currently at a stage it's safe to assume things are left unaltered - if not, that's a bug.

m-u-r-p-h-y (approved):
I agree it is more important to push this through to make sure we have some data integrity in place and we can discuss the philosophical matter later :)

thanks for the quick fix!

antirotor:
Now, would extraction stop on the first failed extractor?

BigRoy:
> Now, would extraction stop on the first failed extractor?

No. It should continue by default - but a User _can_ stop it in-between and with slow extractors tends to do so once an error has been detected.

Pyblish can be set up to fail/stop directly on ANY errors whatsoever whenever you want - you can register a custom test. I actuall recommended and discussed to do so recently [here](https://github.com/pypeclub/OpenPype/pull/2390#issuecomment-1032989829)

> Because if not, this should really be first thing to do

Let me push the move to `ExtractorOrder` in this PR.

> And in that regard I might have done something terrible in another PR - [I've created plugin to clean all mess extractors would produce during their work](https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/maya/plugins/publish/clean_nodes.py)

Yes. That definitely shouldn't be the way to do that. That should be wrapped up in a context manager, and could use for example the [`delete_after`](https://github.com/pypeclub/OpenPype/blob/926eb5b95ece6feb27957c04633bdfb3785e9c7a/openpype/hosts/maya/api/lib.py#L944-L949) context manager to ensure nodes are deleted after they have been extracted.

BigRoy:
For what it's worth here are more **Save Scene** logic across the hosts - most are at beginning of Extractor:

- [After Effects - extract_save_scene](https://github.com/pypeclub/OpenPype/blob/b432613e726770d4f21360f8db65ff3936af8429/openpype/hosts/aftereffects/plugins/publish/extract_save_scene.py#L8)
- [Maya - save_scene](https://github.com/pypeclub/OpenPype/blob/c4c7266529844ad97df4de45acdfbe406290f33f/openpype/hosts/maya/plugins/publish/save_scene.py#L10)
- [Houdini - save_scene](https://github.com/pypeclub/OpenPype/blob/7b2932c9a78aa0e829a6bd06b148eb29a6f31e7d/openpype/hosts/houdini/plugins/publish/save_scene.py#L9)
- [Houdini - save_scene_deadline](https://github.com/pypeclub/OpenPype/blob/7b2932c9a78aa0e829a6bd06b148eb29a6f31e7d/openpype/hosts/houdini/plugins/publish/save_scene_deadline.py#L8)
- [Harmony](https://github.com/pypeclub/OpenPype/blob/5a3ce847485753ebaaa13482d14fdd41b7372d71/openpype/hosts/harmony/plugins/publish/extract_save_scene.py#L9)
- [Photoshop](https://github.com/pypeclub/OpenPype/blob/c1d6eaa5f948f823eea373b723a405176c27278b/openpype/hosts/photoshop/plugins/publish/extract_save_scene.py#L8)

antirotor:
> That should be wrapped up in a context manager

True but that would work only inside of one plugin. You see, I need to duplicate some meshes, combine them, reparent - and then I want to use FBX extractor to extract it. Now I am thinking that the only reasonable way to do it is actually pull code out of the FBX extractor to library, do my stuff with context manager and reuse the code in the same plugin.

antirotor:
> We might want to unify and even opt to do a Global plug-in across multiple hosts using the host's workio.save() to save the scene? Could be worth a separate issue?

That is interesting idea definitely worth exploring in separate issue. I am not sure now if we have it implemented across hosts but I think that is it so,

BigRoy:
> True but that would work only inside of one plugin. You see, I need to duplicate some meshes, combine them, reparent - and then I want to use FBX extractor to extract it. Now I am thinking that the only reasonable way to do it is actually pull code out of the FBX extractor to library, do my stuff with context manager and reuse the code in the same plugin.

You could also use a Context manager stack where you only append your extra context manager that does that if the family is the specific unreal family, like `contextlib.nested` (or Py2: `contextlib.ExitStack`)

Example:
```python
# pseudocode
stack = [a, b, c]
if family == "unreal":
    stack.append(d)

with contextlib.nested(*stack):
    export_fbx()
```

PR #2746, opened 2022-02-16
Flame: adding validator source clip

jakubjezek001:
## Brief description
Checking if source clip is present in reels.

## Description
Due to Flame API limits we have to have available source clips for timeline segments in Reels as they are then used for rendering rather then render all segments in timeline.


mkolar:
Task linked: [OP-2650 flame: validator instance data has flameSourceClip](https://app.clickup.com/t/6658547/OP-2650)

iLLiCiTiT (approved):

PR #2747, opened 2022-02-16
Houdini: Move Houdini Save Current File to beginning of ExtractorOrder

BigRoy:
## Brief description

In the spirit of this PR https://github.com/pypeclub/OpenPype/pull/2744 with [this commit](https://github.com/pypeclub/OpenPype/pull/2744/commits/c4c7266529844ad97df4de45acdfbe406290f33f) moving the order for the Save Scene plug-ins to beginning of ExtractorOrder this makes Houdini align with that too.

As [discussed in the comments this seemed like the more sensible moment to save the current workfile](https://github.com/pypeclub/OpenPype/pull/2744#issuecomment-1041974051).

### Additional note

In Houdini there seemed to be two plug-ins involved with saving - one operating at a different target for different families.
We could technically add both in a single file and do something like:

```python
class SaveCurrentScene(pyblish.api.ContextPlugin):
    """Save current scene"""

    label = "Save current file"
    order = pyblish.api.ExtractorOrder - 0.49
    hosts = ["houdini"]
    targets = ["deadline"]

    def process(self, context):
        ...

# Inherit from SaveCurrentScene to always save for all families for target `deadline`
class SaveCurrentSceneDeadline(SaveCurrentScene):
    hosts = ["houdini"]
    targets = ["deadline"]
```

That way the logic could be less duplicated and the Save plug-ins close together.


antirotor:
Shouldn't we add more families there, or maybe use it for all of them?

BigRoy:
> Shouldn't we add more families there, or maybe use it for all of them?

Quickly checked with our Houdini artist - he had no issue with scene being saved on all families by default. We can always tone that down if we notice issues with that approach.

BigRoy:
> or maybe use it for all of them?

Implemented with https://github.com/pypeclub/OpenPype/pull/2747/commits/7de42ee923dabba0c534774d6367fc44df67ecf6

antirotor (approved):

PR #2748, opened 2022-02-17
Aftereffects 2022 and Deadline

kalisp:
## Brief description
Version of AE wasn't propagated correctly into DL.

## Description
This adds possibility to query AE to get version that should be pushed to Deadline. With this version proper executable of AE is selected.

Introduced `Multiprocess` field in `project_settings/deadline/publish/AfterEffectsSubmitDeadline`   to enable/disable multiprocess rendering on DL. It looks that AE 2022 (at least current version) has some issue (throws error in DL about wrong `-mp` argument).

In case of multiple versions of AE working on same project (not recommended) its possible to use environment variable `MULTIPROCESS` set to 'No' in System Settings to override project setting for specific variant of AE.

`MULTIPROCESS` env var is not required, only 2022 has it set in defaults. 

## Additional info
Also contains fix for `get_audio_url` issue, kept in same PR to not increase extension version unnecessarily.

Requires reinstall of extension!

## Testing notes:
1. start with this step
2. follow this step


mkolar:
Task linked: [OP-2551 Aftereffects 2022 and Deadline](https://app.clickup.com/t/6658547/OP-2551)

PR #2749, opened 2022-02-17
Maya - Easy turntable

antirotor:
## Feature

Easy turnable creator for Maya

🛠️ WIP


mkolar:
Task linked: [OP-1431 Easy Turntable](https://app.clickup.com/t/6658547/OP-1431)

openpype/hosts/maya/plugins/create/create_turntable.py
```diff
@@ -0,0 +1,76 @@
...
+            # reference template file
+            cmds.file(path,
+                    reference=True,
+                    groupReference=True,
+                    namespace=namespace,
+                    groupName="_GRP",
+            )
+
+            # # get animation time and set key frames
+            cmds.setKeyframe(instance, t=1001, v=0.0, at="ry", itt="linear", ott="linear")
```
antirotor, line 47: This won't work as start frame can be different then 1001 - it should be pulled from current maya timeline (that should be set to asset data)

openpype/hosts/maya/plugins/create/create_turntable.py
```diff
@@ -0,0 +1,76 @@
...
+                        asset=asset,
+                        options= {"useSelection":True},
+                        data={
+                            "frameStartHandle": 1001,
+                            "frameEndHandle": int(frames)+1001,
+                        }
+                        )
+
+            cmds.setAttr("defaultRenderGlobals.startFrame", 1001)
+            cmds.setAttr("defaultRenderGlobals.endFrame", int(frames)+1001)
```
BigRoy, line 76: It feels weird to me to have this kind of logic in the Creator.

- What if after creation you want to update the start or end frame?
- What if after creation the model you're turntabling changed in size?
- Why does this need to mess with my render globals? Don't touch my render settings :)

I'd say, move this logic elsewhere, e.g. an Extractor or alike, where it temporarily creates this during the context. Or am I overthinking this?

antirotor:
A agree. I would:

- change how this handles frame range
- change where it does it - but be aware that as it is using render family and render family by default render from published workfile, you would need to have those changes saved in  that workfile, so they can be temporary.
-  `turntable.mb` is pulling with itself dependencies that will stop Deadline for rendering until they are satisfied
-  During creation it is failing for me with:

image

mkolar:
This doesn't have high priority on the internal backlog now, and we feel it will need to be reworked from the ground up. Closing it down as it's not really hitting the mark at the moment.

PR #2754, opened 2022-02-17
Harmony: Rendering in Deadline didn't work in other machines than submitter

kalisp:
## Brief description
DL used `c:/Users/SUBMITTER` folder to render files, which didn't work on different machine than submitter.

## Description
Because work folder in Harmony is local, it must be replaced with folder where workfile will be finally published. In some cases naming convention of expected files didn't match correct value, which resulted in not updating the path.
This explicitly replaces dir name of expected file with staging folder where renders will be created on a DL machine.

## Additional info
technical information.


mkolar:
Task linked: [OP-2669 Harmony use_published for rendering on DL not working](https://app.clickup.com/t/6658547/OP-2669)

64qam (approved):

PR #2755, opened 2022-02-18
Add root keys and project keys to create starting folder

ClementHector:
While keeping the previous operation of create_starting_folder, the project[name], project[code] and root[work..] keys can be used in project folder structure, for example:
```
{
     "{root[work]}": {
         "project[code]": {
             "template": {}
         }
}
```

### DIFF
--- a/openpype/lib/path_tools.py
+++ b/openpype/lib/path_tools.py
@@ -6,10 +6,11 @@ import logging
 import six
 
 from openpype.settings import get_project_settings
-from openpype.settings.lib import get_site_local_overrides
 
 from .anatomy import Anatomy
 from .profiles_filtering import filter_profiles
+
+import avalon.api
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


iLLiCiTiT:
The same settings are also used to create default project hiearchy in ftrack. This change would need to separate the settings as with this change it can be used only for disk folders creation. @mkolar ?

ClementHector:
I don't know how it works with ftrack but it is always possible to make the two types of dictionary coexist:
  - for ftack:
```
"__project_root__": {
     "assets[ftrack.Library]": {
       "characters[ftrack]": {},
       "locations[ftrack]": {}
     },
     "shots[ftrack.Sequence]": {
       "editorial[ftrack.Folder]": {}
     }
}
```
  - for OpenPype:
```
{
      "{root[work]}": {
          "project[code]": {
              "template": {}
          }
}
```
Otherwise I can also add the features that are missing in OP to create the folders of each asset, but maybe all the ftrack entities are not found in OpenPype

iLLiCiTiT:
> I don't know how it works with ftrack but it is always possible to make the two types of dictionary coexist:

It seems that this feature with folder structure started to do things that were not planned to do and it looks to me that it is getting over it's capabality.
Originaly it was added to:
1. Create default project structure in ftrack (or other project manager) - let's say "default asset documents" in terms of avalon
2. Create default folders on disk that can't be created automatically on application start but it is good to have it for production
It was not meant to precreate workdirs of known hosts. The question would be "Why?".

From my opinion I agree that these two functions should have split settings but there were anti-arguments, and if the functions won't precreate workdirs of known hosts and will use default `work` template to create the default folder structure the anti-arguments are still valid. But this PR is breaking that rules and we have to first split the settings to be able merge this PR. That is something we have to discuss with @mkolar .

### Why is the feature getting over it's capabality?
> Otherwise I can also add the features that are missing in OP to create the folders of each asset,

Workfiles can have multiple templates based on host name and task type, `work` root and `work` template are just a defaults but you can do much more. It is just not possible to precreate workdirs properly for known hosts because you don't know contexts in which they should be created.

For example unreal has separated workfiles template and creating workdir of unreal workdirs for each asset is ridiculous and creating workfirs for tasks which are meant only for unreal is confusing. Also there are productions having more work templates because are using different roots for different hosts.

What I want to say is that for simple scenarios it is doable and the question "Why?" is not in the place but I have no idea how would you handle complex scenarios in a reasonable way, thus the question "Why to create the folders of each asset?" is raised.

mkolar:
I agree that this is stretching the target of this functionality quite a bit and has potentially of introducing confusion as to what it should be used for. @ClementHector did you have a particular use case that this was solving?

ClementHector:
In the case where several root exist in my project. Currently the hierarchy under `__project_root__` is built for all project roots.

mkolar:
Right. Let's merge this then. @ClementHector could you do me a favour please and adjust the default settings for this to look like this? It is really supposed to be just a simple example, so let's get rid of the ftrack dependency in there, that should only really be in ftrack settings anyways. 
```
{
    "__project_root__": {
        "prod": {},
        "resources": {
            "footage": {
                "plates": {},
                "offline": {}
            },
            "audio": {},
            "art_dept": {}
        },
        "editorial": {},
        "assets": {
            "characters": {},
            "locations": {}
        },
        "shots": {
        }
    }
}
```

ClementHector:
It's done as you requested ;)

mkolar (approved):

mkolar:
Thank you @ClementHector Merging

PR #2757, opened 2022-02-18
Loader UI: Fix right click in representation widget

iLLiCiTiT:
## Brief description
[This PR](https://github.com/pypeclub/OpenPype/pull/2731) broke right click of representation widget. It also used ItemRole which would not work with older python + PySide.

## Description
Don't use `ItemRole` in loader since it won't pass right object in older Python + PySide combination because proxy model will convert it to known python oject which cause lost of methods/attributes.

## Changes
- use item id in representations widget

## Testing notes:
Site sync module must be enabled
1. Open loader in host
2. Right click on representation in representation widget

### DIFF
--- a/openpype/tools/loader/model.py
+++ b/openpype/tools/loader/model.py
@@ -1068,6 +1068,7 @@ class RepresentationModel(TreeModel, BaseRepresentationModel):
         self._icons = lib.get_repre_icons()
         self._icons["repre"] = qtawesome.icon("fa.file-o",
                                               color=style.colors.default)
+        self._items_by_id = {}
 
     def set_version_ids(self, version_ids):
         self.version_ids = version_ids
@@ -1075,6 +1076,9 @@ class RepresentationModel(TreeModel, BaseRepresentationModel):
 
     def data(self, index, role):
         item = index.internalPointer()
+
+        if role == ITEM_ID_ROLE:
+            return item["id"]
 
         if role == self.IdRole:
             return item.get("_id")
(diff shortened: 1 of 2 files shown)


BigRoy:
> Don't use ItemRole in loader since it won't pass right object in older Python + PySide combination because proxy model will convert it to known python oject which cause lost of methods/attributes.

Where could I detect specifically this problem? Since the problem also seems to exist in e.g. Maya 2020 I'd assume it's not 'older PySide' unless you're referring to older PySide2 with Python 2?

mkolar (approved):

mkolar:
Works for me now. @BigRoy the problem is visible in loader. When you right click on a representation, you don't get any context menu with loaders at all without the fix

BigRoy:
> When you right click on a representation, you don't get any context menu with loaders at all without the fix

I understand the bug - but I'm not sure the way it's fixed or at least what's described here to be the issue to be correct. I've not seen before that Qt would convert the type of the returned object and thus it breaking.

This sounds more like an issue with the Proxy Model than of Qt itself. I've seen that happen with proxy models if they weren't implemented correctly to pass in the right data.

> because proxy model will convert it to known python oject which cause lost of methods/attributes.

Could you point to me which proxy model is involved breaking this scenario? :)

kalisp (approved):

64qam (approved):

iLLiCiTiT:
> This sounds more like an issue with the Proxy Model than of Qt itself.

It is partially true but there's not much we can do about it. It happens in dcc which have older Qt binding or with small tweaks in c++. Nuke had this issue I think up to v 10, it seems that flame has similar issue.

Accessing data through roles is not meant to return complex objects but `QVariant` (Documentation: "acts like a union for the most common Qt data types"). It seems that older versions are trying to convert the value into python common data types which for `Item` class is `dict` thus [this happens](https://github.com/pypeclub/OpenPype/issues/2728). I've tried overridde proxy model methods to avoid this but it fixed only few cases and at the end it just add complexity into parts of code where it should not matter just to be able get value which is not subset of `QVariant` types...

I don't think this fix is ideal but accessing `Item` object is not either.

PR #2759, opened 2022-02-18
Maya: Fix `unique_namespace` when in an namespace that is empty

BigRoy:
## Brief description

This is a moved implementation of https://github.com/pypeclub/avalon-core/pull/412

## Testing notes:

1. See the tests in https://github.com/pypeclub/avalon-core/pull/412 and run those 
2. Load some assets, make sure it works and namespaces come in fine

### DIFF
--- a/openpype/hosts/maya/api/lib.py
+++ b/openpype/hosts/maya/api/lib.py
@@ -206,21 +206,51 @@ def unique_namespace(namespace, format="%02d", prefix="", suffix=""):
         format (str, optional): Formatting of the given iteration number
         suffix (str, optional): Only consider namespaces with this suffix.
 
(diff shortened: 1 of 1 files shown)


m-u-r-p-h-y (approved):
works.

the linked testing process needs to be modified not to use avalon

```
from maya import cmds
import openpype.hosts.maya.api.lib

cmds.namespace(add="foo")
cmds.namespace(set="foo")
openpype.hosts.maya.api.lib.unique_namespace("namespace") 
```

BigRoy:
Thanks @m-u-r-p-h-y - even better would be [the full set of tests in the comment below it](https://github.com/pypeclub/avalon-core/pull/412#issuecomment-1012682541). 🥇

PR #2760, opened 2022-02-18
Update docusaurus to latest version

mkolar:
## Brief description
Update docusaurus to latest beta and fix build


mkolar:
Seems, that the CI build is now running as it should.

PR #2765, opened 2022-02-18
Avoid renaming udim indexes

gabormarinov:
## Brief description
Texture batch publishing renamed texture udim indexes like they were image sequence indexes. This fixes that, leaving udim indexes untouched.


kalisp (approved):

kalisp:
Thanks for the fix.

PR #2766, opened 2022-02-19
General: Extract template formatting from anatomy

iLLiCiTiT:
## Brief description
Extracted general logic of string formatting out of anatomy to separated lib part so it can be used outside of project anatomy.

## Description
With extraction the template formatting logic was a little bit modified. It should be easier to use it elsewhere and add some more logic to formatting logic. Template is separated into chunks of strings to format, optional parts and other parts of template.

## Additional info
This PR may affect some changes that are changed in different forks (e.g. [this PR](https://github.com/pypeclub/OpenPype/pull/2218)). @ClementHector would it be OK for you?

## Changes
- separated string template formatting into `openpype/lib/path_templates.py`
- templates are split into string chunks with separated logic
    - first are separated formatting string
    - second are optional parts that may contain formatting strings or other optional parts
        - this is possible `<task[name]<_comment>>` now
- main class is `StringTemplate` which has `format` and `format_strict`
    - `format_strict` will format the template and will raise an `TemplateUnsolved` if template can't be filled
    - `format` method returns `TemplateResult` but won't crash if there are missing keys or invalid types
- Anatomy templates are using the new logic

## Testing notes:
1. Anatomy should work as before
2. It should be possible to use `StringTemplate` which has `format` and `format_strict`


ClementHector:
I haven't taken the time to go through with this change (other production priorities have come) so you can modify, I'll come back to it as soon as possible.
Path Templating is a critical library, I wish it was independent from openpype and what is its testing lot to avoid bugs. What do you think?

iLLiCiTiT:
> Path Templating is a critical library, I wish it was independent from openpype and what is its testing lot to avoid bugs. What do you think?

I'm not against but I did this change in my free time and I probably won't have much in next few days.

> I'll come back to it as soon as possible.

It's not priority at this time. There are still places where we should but can't use it right now.

ClementHector:
And think what do you think of the integration of [this libirary](https://github.com/4degrees/lucidity) instead of anatomy, which we could maintain outside of OP?
I have a python2/3 fork with optional fields and unit test.

iLLiCiTiT:
> And think what do you think of the integration of [this librairy](https://github.com/4degrees/lucidity) instead of anatomy

Lucidity just don't fit to most of usecases we need. Only possible advantage that lucidity could have is attempt to recreate data from template and formatted result, but it can't work with templates and data used with OpenPype to be safely used in production.

Side note: Anatomy is not just templates but all base attributes related to project.

mkolar:
I agree with @iLLiCiTiT about lucidity. We considered it at the beginning, but it's nowhere near to what we actually need. What it does, it does well, but it's philosophy is not quite the same.

kalisp:
Currently failing in ExtractBurnin
Screenshot 2022-02-25 093926

iLLiCiTiT:
> Currently failing in ExtractBurnin

Should be fixed now.

kalisp (approved):
Issue fixed, seems ok.

PR #2767, opened 2022-02-20
Houdini: Fix open last workfile

### LINKED ISSUE #2714: Houdini: open last workfile doesn't work
## Bug

Auto-opening of last workfile for Houdini doesn't work as it does in Maya for example.

[cuID:OP-2631]

BigRoy:
## Brief description

This fixes #2714 

## Testing notes:
1. test whether last workfile is opened when enabled in admin settings

✅ I tested this in Houdin 19.0.455 - works fine, the file argument is added on launch and the file is opened.

### DIFF
--- a/openpype/hooks/pre_add_last_workfile_arg.py
+++ b/openpype/hooks/pre_add_last_workfile_arg.py
@@ -17,6 +17,7 @@ class AddLastWorkfileToLaunchArgs(PreLaunchHook):
         "nuke",
         "nukex",
         "hiero",
+        "houdini",
         "nukestudio",
         "blender",
         "photoshop",


BigRoy:
Found two bugs:

1. When opening Houdini with last workfile then it's possible for`api.install()` is not triggered and the startup scripts `123.py` or `houdinicore.py` don't seem to run at all. This only occurs if the scene itself produces error on open - e.g. a missing plug-in. This could actually turn into a crucial issue. 

I tried whether that could be worked around by passing the `waitforui` argument before the HIP file to the executable. [See Houdini command line documentation](https://www.sidefx.com/docs/houdini/ref/commandline.html) - but faced the same issue. 

We could also looking into maybe passing the install script before the filename as argument? But render farm should still rely on `123.py`/`houdinicore.py`.
Another approach might to instead of passing the last file to pass a script that basically loads the last file.

Here's a test file that requires Arnold so you can try to reproduce the issue:
- [test_asset1_debug_v008.zip](https://github.com/pypeclub/OpenPype/files/8105066/test_asset1_debug_v008.zip)

---

2. The Skip Opening last file feature doesn't seem to work - whenever set to "skip" it still adds the filename as argument and thus opens the scene? I'm seeing the red little dot on the application just fine and right click context menu shows the "

If how this got added for Houdini is the way to do it then the bug must be elsewhere. Maybe @kalisp knows where to best investigate the root of this?

BigRoy:
Actually [`123.py` not triggering when launching Houdini with a file so this might be the correct behavior as per the documentation](https://www.sidefx.com/docs/houdini/hom/locations.html#startup)

> Houdini runs this script when it is started **_without_** a scene (.hip) file. Houdini will only run the first 123.py script it finds in the path. This is useful for customizing the empty scene, for example if you want to start every scene with a default lighting rig.

BigRoy:
I've pushed a fix/workaround for Houdini not triggering the launch script when launching with last workfile. It's a bit verbose and explicit, but it works fine on my end. I still feel it's off that it doesn't work otherwise.

Also feel it's weird that I had to explicitly open the file (last argument) in the launch script because just `houdini.exe script.py file.hip` didn't actually open the file. Which seems odd but that's almost the [exact example used in the documentation.](https://www.sidefx.com/docs/houdini/ref/commandline.html)

> `houdini -foreground myscript.cmd waitforui setupshelf.py my.hip`

The waitforui flag made no difference as to whether it'd automatically load the hip file. Any Houdini artists that have more experience with opening a file on Houdini launch whilst also triggering startup scripts?

openpype/hosts/houdini/startup/scripts/openpype_launch.py
```diff
@@ -0,0 +1,40 @@
...
+        # If the last argument is a Houdini file open it directly
+        workfile_path = args[-1].replace("\\", "/")
+        print("Opening workfile on launch: {}".format(workfile_path))
+
+        # We don't use `workio.open_file` because we want to explicitly ignore
+        # load warnings. Otherwise Houdini will fail to start if a scene load
+        # produces e.g. errors on missing plug-ins
+        hou.hipFile.load(workfile_path,
+                         suppress_save_prompt=True,
+                         ignore_load_warnings=True)
```
BigRoy, line 37: This does have the downside that warnings aren't even logged to the output window. So it's not perfect.

BigRoy:
After some more investigating it seems that `123.py` and `houdinicore.py` are intended to run a script on Houdini launch to "customize" the empty scene it starts with. As such we basically abused them all this time. 

Instead to "always run" on startup Houdini uses `pythonrc.py` files. I've replaced the logic and now adding a workfile argument became as trivial as other hosts.

As per [houdini documentation](https://www.sidefx.com/docs/houdini/hom/locations.html#startup).

iLLiCiTiT (approved):
Works in Houdini 18.5, Python 3, Windows. @antirotor do you have Houdini with Python 2?

antirotor (approved):
It works with opening last workfile (tested on H18.5 with python2 and H19) I can confirm that skipping it doesn't work. I would merge this and create new issue #2788.

PR #2752, opened 2022-02-17
Fusion: migrate colorbleed fork code

jakubjezek001:
## Brief description
Integrating Colorbleed's work and adding `Set Framerange` to menu items.

## Description
Original PR description here https://github.com/BigRoy/OpenPype/pull/2
- added `Set Framerange` to menu items to set framerange to comp from asset document data.

PR #2769, opened 2022-02-20
Fusion: Tweak Fusion integration

BigRoy:
## Brief description

This is a replacement PR for #2752 with the code coming from our studio's `colorbleed` branch which had tweaks done to the Fusion integration.

## Description

This is a first pass to expose the tweak we've been doing on our end to discuss the changes.

1. Instead of copying scripts into the user's local Fusion folder this instead relies on using Fusion masterprefs - which is way to override settings, through which we can customize the 'path map' locations for scripts and thus load them directly from OpenPype. **--> Note: users that had launched the older integration would still have those original scripts in their scripts folders!**
2. The OpenPype "menu" is now accessible as top-level _OpenPype_ menu as opposed to hidden in Fusion's Comp scripts.
3. At the top of the OpenPype menu (the Qt floating menu) in Fusion it now shows the current context (where Avalon thinks it's operating in the current session) with regards to the current asset - since the Python scripts in Fusion run in a completely standalone process it can be non-trivial to find this out without a visual clue.

- [x] Global In/Out will now be considered as handles for your publish for all areas outside of the Current Time Range.
- [x] Added some tweaks to support Fusion 18+ (note that Fusion 18 supports multiple Py3 versions, in particular Python 3.3-3.10)
- [x] Added default settings for a Fusion 18 application entry
- [x] Added OCIO config setting which will set OCIO env var on Fusion launch if the setting is enabled (or maybe I should remove the enabled checkbox and allow to just skip if an empty OCIO config path was set in settings?
- [x] Fixed loader issue where it wouldn't align frames correctly if there were start handles.
- [x] Allow to minimze the OpenPype Menu window.
- [x] Fix Set Resolution in OpenPype menu
- [x] Fix Set Frame Range in OpenPype menu

## Additional Context

- [Related "Fusion Thread" in OpenPype Discord](https://discord.com/channels/517362899170230292/936207795039121488/943502368476123146)
- [Brief discussion about some pipeline implementation details in _Davinci Resolve Plug-in Developers_ discord](https://discord.com/channels/793508729785155594/793508730276282439/939563124246315028)

## Testing notes:

1. go through testing Fusion from the ground up - as if a new user


BigRoy:
This is currently a code merge of a few things which I haven't tested yet. I'm expecting to test by Tuesday the earliest.

antirotor:
So, could this be tweaked a little bit to provide PySide2 and OTIO for Resolve? Just asking how much is it bound to Fusion logic.

BigRoy:
> So, could this be tweaked a little bit to provide PySide2 and OTIO for Resolve

This actually still doesn't provide PySide2 and OTIO out of the box - or at least PySide2 it doesn't. I'm not sure what'd be needed on OTIO end since I hadn't tested that at all.

The master prefs might be fusion specific - I'm not sure about that. Also, with regards to the top level menu in a `.fu` file - I'm not sure how well that transfers to Resolve unfortunately.

Can't provide much information here.

BigRoy:
Tweaked some more things for this PR and tested it in Fusion. Seemed to work fine.

**Aside of these issues:**

- Save Prompt opens behind Work Files tool -> fixed with #2784 
- OpenPype logs are errors in the Fusion console -> separate issue #2797 
- Set Resolution in OpenPype menu does nothing - to be implemented
    - should set: `Comp settings > Frame Format -> Resolution`
- There's no button to set the "Current FPS"? And there's no "new scene" callback to enforce it for the current comp. So should there maybe be a `Set FPS` button like Set Resolution and Set Frame Range until callbacks are in place?
    - should set: `Comp Settings > Frame Format > Frame Rate`

---

Not a direct issue, but settings we might want to set for the comp by default too:
- Comp Settings > General > Frames Start From > Sequence (instead of start from one)

At the same time since the Loaders already do so for published files this might actually be "forcing too much" by default? Nonetheless it makes me think once we have new scene callback/default settings implemented we might want to just expose a user to apply custom prefs overrides in OP admin settings.

---

@jakubjezek001 any chance you could also take a look at the current state and let me know what feels off to you?

jakubjezek001 (changes requested):
Hey @BigRoy for the update! I was testing and at the moment found following problems:
- when some work has been done in one shot context and then in workfile want to open other shot context, a dialogue window (in red rectangle) is showing under the workfile window

image

-  I have to change `FUSION16_PYTHON36_HOME` environment key in prelaunch hook as it is not supported by app setting yet. Reversing it to `PYTHON36` was fixing it for me for the test. https://github.com/pypeclub/OpenPype/blob/237ce7938dc04d163308cf34960433eac3132609/openpype/settings/defaults/system_settings/applications.json#L698

- publishing created render node was working well 
- loading rendered image sequence or mov works but having an issue with correctly setting frameranges

BigRoy:
> when some work has been done in one shot context and then in workfile want to open other shot context, a dialogue window (in red rectangle) is showing under the workfile window

Correct - [as mentioned](https://github.com/pypeclub/OpenPype/pull/2769#issuecomment-1048273383) that's fixed in #2784

> I have to change FUSION16_PYTHON36_HOME environment key in prelaunch hook as it is not supported by app setting yet. Reversing it to PYTHON36 was fixing it for me for the test. 

I set that in the Application environment - just like how it was set before for PYTHON36.

fusion_env

> loading rendered image sequence or mov works but having an issue with correctly setting frameranges

What's the issue? :)

mkolar:
Task linked: [OP-2667 Fusion: set framerange from menu button](https://app.clickup.com/t/6658547/OP-2667)

openpype/hosts/fusion/deploy/fusion_shared.prefs
```diff
@@ -0,0 +1,19 @@
+{
+Locked = true,
+Global = {
+ Paths = {
+  Map = {
+   ["OpenPype:"] = "$(OPENPYPE_FUSION)/deploy",
+   ["Reactor:"] = "$(REACTOR)",
```
jakubjezek001, line 7: Could not find it. Is there set somewhere in app environments?

BigRoy: Ah, REACTOR is only there because it's used by almost everyone working with Fusion in a more than average capacity. The unfortunate side effect of masterprefs is that it "locks" the settings to those values. So a user/studio couldn't set custom path maps for their studio unless they were to change the `.prefs` file.

`REACTOR` would be an environment variable set by the studio as to where they have that installed (if they use it). If they don't it's mostly disregarded by fusion because the pathmapping doesn't resolve to anything.

You'd set it in Admin Settings, like `FUSION16_PYTHON36_HOME` as in my screenshot [here](https://github.com/pypeclub/OpenPype/pull/2769#issuecomment-1049985998).

jakubjezek001:
> What's the issue? :)

So I was testing it one more time and found out there is problem with the set_framerange. It is defining render range exluding handles and if publishing of the render node is done during those render range defined, it will render only range without handles. 

Once it is loaded back to workfile, it is set to the first workfile frame and it seems to be wrong. I wonder if the render range could be ignored during publishing. The handles identified by those limits shown in image are cool and we are used to work with those in Nuke. User is interactivly shown what is in edit and what is handle. 

image

It would be good to tel rendering plugin that it should render full range.

BigRoy:
@jakubjezek001 Thanks for being so verbose about the issue - but somehow my mind is having a hard time grasping exactly what issue you're having with the handles. Any chance you could do a screen recording of the issue?

BigRoy:
I'm thinking of also exposing `FUSION_PROFILE_DIR` as a setting to the OpenPype fusion integration so that it by default also uses a custom settings location for Fusion. So that whatever gets "set" as settings inside OpenPype never influences what is set outside of OpenPype.

What would be a good default location for OpenPype to expose Host specific settings/files? Of course we could expose the default value in settings. Would `%LOCALAPPDATA%/pypeclub/hosts/fusion` be a decent default?

Also, the most [recent Fusion release 17.4.6 also adds a `/execute` argument](https://forum.blackmagicdesign.com/viewtopic.php?f=22&t=157536) that you can pass along with Fusion launch to directly run a lua script on Fusion start. Here's an example of that:
```
fusion /execute "print(fusion:GetCurrentComp():GetAttrs().COMPS_FileName)" "P:\path\to\file.comp"
```
So I expect we could also use that to force a script to run at startup, e.g.
```
fusion /execute "fusion:RunScript('/path/to/python/script.py3')" "P:\path\to\file.comp"
```
However it does not run early enough to set up Fusion Path mapping (which is needed to customize where Fusion looks for Scripts, Menus, Callbacks, etc.) and requires to restart Fusion at least once before it's picked up - something we don't want. And thus we'll still need to rely on using Fusion Master Prefs to override until Black Magic Design exposes alternatives in future versions.

BigRoy:
> What would be a good default location for OpenPype to expose Host specific settings/files? Of course we could expose the default value in settings. Would %LOCALAPPDATA%/pypeclub/hosts/fusion be a decent default?

Anyone any pointers on this? :)

mkolar:
>Anyone any pointers on this? :)

`%LOCALAPPDATA%/pypeclub/hosts/fusion` is totally reasonable. I'm, however, unable to comment on much of the rest as I'm really not versed in fusion in general.

jakubjezek001:
> > What would be a good default location for OpenPype to expose Host specific settings/files? Of course we could expose the default value in settings. Would %LOCALAPPDATA%/pypeclub/hosts/fusion be a decent default?
> 
> Anyone any pointers on this? :)

To be honest I am not very happy with the `pypeclub` part of the path. I would rather prefare`%LOCALAPPDATA%/OpenPype/hosts/fusion`. BTW, I do realize we are using `pypeclub` in the locals but I disagree with it. Pypeclub is the business name and not the package name.

jakubjezek001:
Hey @BigRoy how are you doing with this one? Can I help somehow to finalize this one? I believe it might be just a little bit from the grand finale, right?

mkolar:
To be honest I'd be happy to merge this, considering the testing going on in colourbleed. We just need to sort out the conflicts.

BigRoy:
I've merged with latest develop and fixed the merge conflicts.
I have also tested some first tweaks with regards to [this comment](https://github.com/pypeclub/OpenPype/pull/2769#issuecomment-1050018849) and making sure start + end frames are set as expected. But I'll still need to test+fix those changes - hopefully this week!

openpype/hosts/fusion/api/lib.py
```diff
@@ -18,12 +18,14 @@
     legacy_io,
 )
 from .pipeline import get_current_comp, comp_lock_and_undo_chunk
-
+from openpype.api import (
```
iLLiCiTiT, line 21: This should be changed to `from openpype.pipeline.context_tools import get_current_project_asset`

openpype/hosts/fusion/api/menu.py
```diff
@@ -1,26 +1,21 @@
-import os
 import sys
 
 from Qt import QtWidgets, QtCore
 
-from openpype import style
+from avalon import api
```
iLLiCiTiT, line 5: Wow. This is such old PR?

BigRoy: Definitely is - I'll test it soon and clean up all the bits :D

BigRoy:
- Fixed refactoring for merge with latest develop
- Added some tweaks to support Fusion 18+ (note that Fusion 18 supports multiple Py3 versions, in particular Python 3.3-3.10
- Added OCIO config setting which will set `OCIO` env var on Fusion launch if the setting is enabled (or maybe I should remove the `enabled` checkbox and allow to just skip if an empty OCIO config path was set in settings?
- Fixed loader issue where it wouldn't align frames correctly if there were start handles.

openpype/settings/entities/schemas/projects_schema/schemas/schema_anatomy_imageio.json
```diff
@@ -427,6 +427,35 @@
                 }
             ]
         },
+
+        {
```
iLLiCiTiT, line 431: Please move this fro manatomy to fusion settings (regarding issue https://github.com/pypeclub/OpenPype/issues/3865 ).

jakubjezek001: suggestion: `project_anatomy/imageio/fusion` > `project_settings/fusion/imageio`

BigRoy: It's now moved to `project_settings/fusion/imageio/ocio/configFilePath`:

afbeelding

I still need to update the Hook..

Does the hook have access to `self.data["project_settings"]` as I quickly refactored in https://github.com/pypeclub/OpenPype/pull/2769/commits/7aa905898e6726e910c05be8db6514e5e656b8c1 ?

openpype/hosts/fusion/hooks/pre_fusion_setup.py
```diff
@@ -1,8 +1,6 @@
 import os
-import shutil
-
-import openpype.hosts.fusion
 from openpype.lib import PreLaunchHook, ApplicationLaunchFailed
+from openpype.hosts.fusion import HOST_DIR
```
jakubjezek001, line 3:
```suggestion
from openpype.hosts.fusion import FUSION_HOST_DIR
```

openpype/hosts/fusion/hooks/pre_fusion_setup.py
```diff
@@ -14,101 +12,40 @@ class FusionPrelaunch(PreLaunchHook):
...
+        # TODO: Set this for EITHER Fu16-17 OR Fu18+, don't do both
+        # Fusion 18+ does not look in FUSION16_PYTHON36_HOME anymore
+        # but instead uses FUSION_PYTHON3_HOME and requires the Python to
+        # be available on PATH to work. So let's enforce that for now.
+        self.launch_context.env["FUSION_PYTHON3_HOME"] = py36_dir
+        self.launch_context.env["PATH"] += ";" + py36_dir
+
+        # Add our Fusion Master Prefs which is the only way to customize
+        # Fusion to define where it can read custom scripts and tools from
+        self.log.info(f"Setting OPENPYPE_FUSION: {HOST_DIR}")
```
jakubjezek001, line 45:
```suggestion
        self.log.info(f"Setting OPENPYPE_FUSION: {FUSION_HOST_DIR}")
```

openpype/hosts/fusion/hooks/pre_fusion_setup.py
```diff
@@ -14,101 +12,40 @@ class FusionPrelaunch(PreLaunchHook):
...
+        self.launch_context.env["OPENPYPE_FUSION"] = HOST_DIR
```
jakubjezek001, line 46:
```suggestion
        self.launch_context.env["OPENPYPE_FUSION"] = FUSION_HOST_DIR
```

openpype/hosts/fusion/hooks/pre_fusion_setup.py
```diff
@@ -14,101 +12,40 @@ class FusionPrelaunch(PreLaunchHook):
...
+
+        pref_var = "FUSION16_MasterPrefs"   # used by both Fu16 and Fu17
+        prefs = os.path.join(HOST_DIR, "deploy", "fusion_shared.prefs")
```
jakubjezek001, line 49:
```suggestion
        prefs = os.path.join(FUSION_HOST_DIR, "deploy", "fusion_shared.prefs")
```

openpype/hosts/fusion/hooks/pre_fusion_setup.py
```diff
@@ -14,101 +12,40 @@ class FusionPrelaunch(PreLaunchHook):
...
-        if not py36_dir:
-            raise ApplicationLaunchFailed(
-                "Required environment variable \"PYTHON36\" is not set."
-                "\n\nFusion implementation requires to have"
-                " installed Python 3.6"
-            )
-
-        py36_dir = os.path.normpath(py36_dir)
-        if not os.path.isdir(py36_dir):
+        py36_var = "FUSION16_PYTHON36_HOME"
```
jakubjezek001, line 15: shouldn't this be renamed to something like "FUSION_PYTHON_HOME"?

jakubjezek001 (changes requested):
I have managed to test most of it and it worked. I am suggesting following changes before we merge it:
1. change python home variable name
2. process code suggestions from comments

openpype/hosts/fusion/hooks/pre_fusion_setup.py
BigRoy: I agree - I think `FUSION_PYTHON3_HOME` would be the most sensible - it would even match the Fu18+ name for it too.

BigRoy:
@jakubjezek001 The default environment for Fusion is currently set to this:
```json
{
    "FUSION_UTILITY_SCRIPTS_SOURCE_DIR": [],
    "FUSION_UTILITY_SCRIPTS_DIR": {
        "windows": "{PROGRAMDATA}/Blackmagic Design/Fusion/Scripts/Comp",
        "darwin": "/Library/Application Support/Blackmagic Design/Fusion/Scripts/Comp",
        "linux": "/opt/Fusion/Scripts/Comp"
    },
    "PYTHON36": {
        "windows": "{LOCALAPPDATA}/Programs/Python/Python36",
        "darwin": "~/Library/Python/3.6/bin",
        "linux": "/opt/Python/3.6/bin"
    },
    "PYTHONPATH": [
        "{PYTHON36}/Lib/site-packages",
        "{VIRTUAL_ENV}/Lib/site-packages",
        "{PYTHONPATH}"
    ],
    "PATH": [
        "{PYTHON36}",
        "{PYTHON36}/Scripts",
        "{PATH}"
    ],
    "OPENPYPE_LOG_NO_COLORS": "Yes"
}
```

---

But I feel like that should instead be something like this at its core:
```json
{
    "FUSION_PYTHON3_HOME": {
        "windows": "{LOCALAPPDATA}/Programs/Python/Python36;",
        "darwin": "~/Library/Python/3.6/bin",
        "linux": "/opt/Python/3.6/bin"
    }
}
```
Thoughts?

Or would you rather prefer like this:
```json
{
    "PYTHON36": {
        "windows": "{LOCALAPPDATA}/Programs/Python/Python36;",
        "darwin": "~/Library/Python/3.6/bin",
        "linux": "/opt/Python/3.6/bin"
    },
    "FUSION_PYTHON3_HOME": "{PYTHON36}",
}
```

jakubjezek001:
Also I believe `FUSION_PYTHON3_HOME` should be also part of `openpype/settings/defaults/system_settings/applications.json` It is missing at the moment.

BigRoy:
> Also I believe FUSION_PYTHON3_HOME should be part of openpype/settings/defaults/system_settings/applications.json It is missing at the moment.

Yup, that's what I was referring too for the default setting. Will tweak the PR.

BigRoy:
Have tweaked the addon to set `OPENPYPE_LOG_NO_COLORS`, tweaked the default fusion environment in applications defaults to match what is needed for current implementation to work, fixed the "Set Resolution" button in the OpenPype menu to do anything.

Currently there's no logic anywhere which sets the Comp's default aspect ratio, fps, etc. I feel like that would still be good to add.
There are no real working callbacks for Fusion that I've been able to get to work consistently which would also be able to quickly run against the python code because it'd each run a new python process as far as I am aware. But, that might be best as a separate issue to solve.

BigRoy:
@jakubjezek001 Any chance you're able to look into this PR quickly? :) I'm working on another Fusion integration tweak which might be quite big on top of this PR and I'd love to PR it separate.

It would have the OpenPype menu context try and auto-match with the currently active comp so that if you were to have multiple comps open at the same time from different assets that the UI would be able to change along with when you click into another comp. It'd be based on a timer (running e.g. every 2.5 second - since it runs in a separate process it doesn't trigger any slowdowns in Fusion and runs in 0.0009 seconds or anyway).

jakubjezek001 (approved):
Works perfect! Thanks you ;)

jakubjezek001:
> Actually thinking about that now that could technically be dangerous during publishing if some of the logic there would rely on the context to remain the same during the publishing instead of the user potentially clicking into another comp as he starts the the publish. Maybe instead it should highlight the asset name at the top of the menu with a different color to notify it's not matching - and maybe add a button to "refresh" or "sync" it again. Sounds like a good reason to keep that a separate PR already.

I wonder if the context is collected only once at start of publishing and then it is not required anymore. Even Integrate is taking Context data from collected, right? @iLLiCiTiT

iLLiCiTiT:
>  wonder if the context is collected only once at start of publishing and then it is not required anymore. Even Integrate is taking Context data from collected, right? @iLLiCiTiT

It is, but there is a lot of plugins that are still using `legacy_io` or `os.environ` directly and it's hard to identify if we can "just switch" them because I'm not sure when context is filled during farm publishing.

BigRoy:
> Definetely not issue for this PR :)

Yes - different issue, different PR.

@iLLiCiTiT any chance this PR could get your approval for a merge?

Issue #2740, opened 2022-02-16
Houdini: Reset frame range and resolution


antirotor:
## Feature

Implement reset framerange and resolution the same way as it is in Maya,

[cuID:263dfgu]

BigRoy:
> **Reset resolution**

As far as I know Houdini doesn't have a 'current resolution' except for a resolution _per camera_. There is no default camera, and I'm a bit wary about just changing resolutions on ALL cameras in the file. Thoughts?

> **Reset framerange**

Implemented with #2770

antirotor:
Agree, changing resolution here seems to be out of the scope for Houdini.

BigRoy:
> Agree, changing resolution here seems to be out of the scope for Houdini.

Thanks - because of that I've linked #2770 directly to this issue.

PR #2770, opened 2022-02-20
Houdini: Implement Reset Frame Range

BigRoy:
## Brief description

Implement **Reset Frame Range** for Houdini as per issue #2740 

- Adds menu entry that allows user to reset at any time
- On Houdini open or new scene it will now default to asset frame range.

### Note

It seems that whenever I do new scene the frame range is set correctly, but the start frame just doesn't want to go to the right start frame. It stays at 1 in the playbar. On Houdini launch however it seems fine, also with clicking the entry in the menu bar it also seems fine.

## Testing notes:
1. launch houdini, check whether frame range is correct
2. click CTRL+N (new scene), check whether frame range is correct
2. set your frame range incorrectly, click _reset frame range_ in OpenPype menu, check whether frame range is correct

### DIFF
--- a/openpype/hosts/houdini/api/lib.py
+++ b/openpype/hosts/houdini/api/lib.py
@@ -542,3 +542,37 @@ def maintained_selection():
         if previous_selection:
             for node in previous_selection:
                 node.setSelected(on=True)
+
+
+def reset_framerange():
+    """Set frame range to current asset"""
+
+    asset_name = api.Session["AVALON_ASSET"]
+    asset = io.find_one({"name": asset_name, "type": "asset"})
+
+    frame_start = asset["data"].get("frameStart")
+    frame_end = asset["data"].get("frameEnd")
+    # Backwards compatibility
+    if frame_start is None or frame_end is None:
+        frame_start = asset["data"].get("edit_in")
+        frame_end = asset["data"].get("edit_out")
+
+    if frame_start is None or frame_end is None:
+        log.warning("No edit information found for %s" % asset_name)
+        return
+
+    handles = asset["data"].get("handles") or 0
+    handle_start = asset["data"].get("handleStart")
+    if handle_start is None:
+        handle_start = handles
+
+    handle_end = asset["data"].get("handleEnd")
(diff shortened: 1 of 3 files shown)


antirotor (commented):
It resets the frame range perfectly, but i doesn't work on new scenes and on launch... And I can't find it in the code?

BigRoy:
I think I forgot to push the last commit. Let me do that in a few hours when behind a PC.

BigRoy:
@antirotor pushed the missing commits just now. 👍

openpype/hosts/houdini/api/lib.py
```diff
@@ -542,3 +542,37 @@ def maintained_selection():
...
+    handle_start = asset["data"].get("handleStart")
+    if handle_start is None:
+        handle_start = handles
+
+    handle_end = asset["data"].get("handleEnd")
+    if handle_end is None:
+        handle_end = handles
+
+    frame_start -= int(handle_start)
+    frame_end += int(handle_end)
```
BigRoy, line 574: This logic is basically a straight copy from the implementation in Maya. :)

BigRoy:
I noticed the `on_new` event also got triggered when opening files. That is now fixed with https://github.com/pypeclub/OpenPype/pull/2770/commits/823164e6c49732fa96d6960f886b66d2e1d688fb. 

That implementation could also be moved directly to [here](https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/houdini/api/pipeline.py#L107-L108) so that we wouldn't even emit the Avalon signal. Thoughts?

antirotor (approved):
Now it works perfectly.

| so that we wouldn't even emit the Avalon signal. Thoughts?

any reason why we shouldn't emit it?

BigRoy:
> any reason why we shouldn't emit it?

Mostly that if anyone was to use the Avalon callback in their own config to rely on the "new scene" callback that they might not know that for Houdini it also emits (just before) loading a file. Anyway, it's fine as is. 👍 

Read to merge.

PR #2775, opened 2022-02-21
Slack: Added regex for filtering on subset names

kalisp:
## Brief description
Subset name filtering could be used to limit notification to Slack only for specific AOVs

## Testing notes:
1. configure subset name in `project_settings/slack/publish/CollectSlackFamilies/profiles`
2. run publish


mkolar:
Task linked: [OP-2642 Configure which reviewables are posted on Slack](https://app.clickup.com/t/6658547/OP-2642)

PR #2784, opened 2022-02-22
Work Files: Fix parenting of save prompt QMessageBox

BigRoy:
## Brief description

This makes the save prompt messagebox an actual child to the parent Qt UI element. 

#### Resolves issue

This fixes the save prompt appearing behind the workfiles UI in Fusion.

#### Original "crash" case

Setting the windowFlags without the original messagebox.windowFlags() was the culprit as to why the messagebox previously wouldn't show when parented. Likely because then it's missing the Dialog window flag and thus would try to embed itself into the parent UI, which you then cannot exec()

## Additional Context

(cherry picked from commit 290e2b601d0e20f4aaba356d4f053bf733de406b)

## Testing notes:
1. Test whether save prompt (when current scene has modifications and you are **opening** a file through Work Files) works as expected
2. Please test Blender integration, since original code mentioned "crashes" in Blender. ⚠️ 

I tested in Fusion 17 and Maya 2020 - those worked fine. ✅

### DIFF
--- a/openpype/tools/workfiles/app.py
+++ b/openpype/tools/workfiles/app.py
@@ -544,10 +544,6 @@ class FilesWidget(QtWidgets.QWidget):
         # file on a refresh of the files model.
         self.auto_select_latest_modified = True
 
-        # Avoid crash in Blender and store the message box
-        # (setting parent doesn't work as it hides the message box)
-        self._messagebox = None
-
         files_view = FilesView(self)
 
         # Create the Files model
@@ -726,9 +722,9 @@ class FilesWidget(QtWidgets.QWidget):
         self.file_opened.emit()
 
     def save_changes_prompt(self):
-        self._messagebox = messagebox = QtWidgets.QMessageBox()
-
-        messagebox.setWindowFlags(QtCore.Qt.FramelessWindowHint)
+        messagebox = QtWidgets.QMessageBox(parent=self)
+        messagebox.setWindowFlags(messagebox.windowFlags() |
+                                  QtCore.Qt.FramelessWindowHint)
         messagebox.setIcon(messagebox.Warning)
         messagebox.setWindowTitle("Unsaved Changes!")
         messagebox.setText(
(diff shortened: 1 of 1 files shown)


BigRoy:
@iLLiCiTiT 

afbeelding

It's a trend. :)

iLLiCiTiT (approved):
The crash in blender is "occasional" so it is hard to test. I would recommend to keep store it in the attribute, as it won't affect your change.

EDITED:
I don't think it is still an issue but blender is already so broken so why to break it even more? :)

iLLiCiTiT (changes requested):
I accidentally approved.

BigRoy:
@iLLiCiTiT since the original comment mentioned this:

> (setting parent doesn't work as it hides the message box)

I felt like the issue originally in blender was purely due to the parent relationship not existing in the first place and thus the messagebox getting garbage collected. (Due to how Blender would process events even if you'd explicitly call `exec()` I think).

Should we maybe just try and see if this now holds up in Blender as is? Maybe with just doing some preliminary testing - because I do recall the original crash was somewhat reproducable.

The original implementation of storing the messagebox came from [here](https://github.com/getavalon/core/pull/465) which also describes that. It also came up [here](https://github.com/getavalon/core/issues/361#issuecomment-518662113) as:

> A reference to pop-up windows (like the 'Save changes?' in Work Files) is needed, else they are garbage collected and immediately disappear. I now use self.message for example instead of just message. So this one is fixed in my branch.

Which should now also be solved with the parenting.

Other issues related to crashes in blender all seemed to be [related to the Window instead of the Message Box](https://github.com/getavalon/core/pull/466#issuecomment-585802564).

iLLiCiTiT:
I can confirm that it works on windows (even without `messagebox.windowFlags()`) but it would be good to try it on linux where both Blender and Qt are using totally different code for crutial parts.

BigRoy:
> For me it is worth it to store the dialog into attribute just to make sure it is not garbage collected. It affects nothing.

I'd rather avoid keeping around legacy code that we might not need. Anyway, reverted that with this commit https://github.com/pypeclub/OpenPype/pull/2784/commits/bb105209e8bc34a63e2e04bc8de42a603f3368d4 

Let me know if you still want to change something. :)

Issue #2786, opened 2022-02-22
Pyblish Pype: Preserve typed comment on reset


BigRoy:
**Is your feature request related to a problem? Please describe.**

What happens often is the user typing a comment for the publish in the comment box and starts to validate/publish. If an error occurred he needs to fix, reset and try again.

However, currently resetting Pyblish Pype will remove the current comment.

**Describe the solution you'd like**

Preserve the user typed comment on reset, similar to how Pyblish QML does it.

BigRoy:
Implemented with #2793

PR #2793, opened 2022-02-22
Pyblish Pype: Preserve comment on reset

BigRoy:
## Brief description

Draft implementation for #2786 where the typed comment is preserved on reset.

Of course this has the side effect that if you were to keep the pyblish window open after publishing, go into another asset, then reset that your old comment would still be there. Not sure if that'd be an issue for anyone? If so, we might want to just "clear" the comment after a full publish. (e.g. Publish finished without errors)?

## Additional info

Not sure if this is the cleanest/best implementation. Let me know if you feel there's something off with this approach.

## Testing notes:

1. start pyblish, type a comment, reset. Your comment should be preserved
2. ensure the actual pyblish plug-ins are picking up the preserved comment on the publish

### DIFF
--- a/openpype/tools/pyblish_pype/app.py
+++ b/openpype/tools/pyblish_pype/app.py
@@ -90,9 +90,8 @@ def show(parent=None):
         install_fonts()
         install_translator(app)
 
-        ctrl = control.Controller()
-
         if self._window is None:
+            ctrl = control.Controller()
             self._window = window.Window(ctrl, parent)
             self._window.destroyed.connect(on_destroyed)
 
--- a/openpype/tools/pyblish_pype/control.py
+++ b/openpype/tools/pyblish_pype/control.py
@@ -229,7 +229,13 @@ class Controller(QtCore.QObject):
         self.log.debug("Resetting pyblish context object")
 
         comment = None
-        if self.context is not None and self.context.data.get("comment"):
+        if (
+            self.context is not None and
+            self.context.data.get("comment") and
+            # We only preserve the user typed comment if we are *not*
+            # resetting from a successful publish without errors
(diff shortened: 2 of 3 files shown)


iLLiCiTiT (approved):
Fixed placeholder issue and seems to do what is described. @mkolar can you approve the feature?

BigRoy:
Thanks @iLLiCiTiT 

@mkolar be extre aware of my note:
> Of course this has the side effect that if you were to keep the pyblish window open after publishing, go into another asset, then reset that your old comment would still be there. Not sure if that'd be an issue for anyone? If so, we might want to just "clear" the comment after a full publish. (e.g. Publish finished without errors)?

mkolar (changes requested):
> If so, we might want to just "clear" the comment after a full publish. (e.g. Publish finished without errors)?

I agree with this. In faster project this might very well produce some weird situations, but clearing it after success, si a very good solution I think.

BigRoy:
> I agree with this. In faster project this might very well produce some weird situations, but clearing it after success, si a very good solution I think.

@mkolar Implemented

openpype/tools/pyblish_pype/app.py
```diff
@@ -90,9 +90,8 @@ def show(parent=None):
         install_fonts()
         install_translator(app)
 
-        ctrl = control.Controller()
-
         if self._window is None:
+            ctrl = control.Controller()
```
BigRoy, line 94: This technically wasn't crucial to this PR but I noticed the controller would only be used if there was no original window so I moved it here solely because I was confused at first thinking the controller was always "cleared" even if the window persisted.

PR #2796, opened 2022-02-22
Fix `OPENPYPE_LOG_NO_COLORS` being True actually resulting in `use_colors` being False

BigRoy:
## Brief description

If I'm not mistaken the logic for `OPENPYPE_LOG_NO_COLORS` was incorrectly setting `use_colors` to True when `OPENPYPE_LOG_NO_COLORS` was set to True.

I believe this would fix that - and thus would avoid this printed output when the value is enabled for hosts that don't support `blessed`:
> Module `blessed` failed on import or terminal creation. Pype terminal won't use colors.

## Testing notes:
1. Make sure I'm not misreading the logic. 😄 
2. Test it for hosts that have `OPENPYPE_LOG_NO_COLORS` set.

### DIFF
--- a/openpype/lib/terminal.py
+++ b/openpype/lib/terminal.py
@@ -49,8 +49,8 @@ class Terminal:
         """
 
         from openpype.lib import env_value_to_bool
-        use_colors = env_value_to_bool(
-            "OPENPYPE_LOG_NO_COLORS", default=Terminal.use_colors
+        use_colors = not env_value_to_bool(
+            "OPENPYPE_LOG_NO_COLORS", default=not Terminal.use_colors
         )
         if not use_colors:
             Terminal.use_colors = use_colors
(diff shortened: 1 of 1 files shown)


openpype/lib/terminal.py
```diff
@@ -49,8 +49,8 @@ def _initialize():
         """
 
         from openpype.lib import env_value_to_bool
-        use_colors = env_value_to_bool(
-            "OPENPYPE_LOG_NO_COLORS", default=Terminal.use_colors
+        use_colors = not env_value_to_bool(
+            "OPENPYPE_LOG_NO_COLORS", default=not Terminal.use_colors
         )
         if not use_colors:
             Terminal.use_colors = use_colors
```
iLLiCiTiT, line 56: It is right but I had to read it 3 times to understand the logic. I suggest for readibility this change
```suggestion
        log_no_colors = env_value_to_bool(
            "OPENPYPE_LOG_NO_COLORS", default=None
        )
        if log_no_colors is not None:
            Terminal.use_colors = not log_no_colors
        
        if not Terminal.use_colors:
```

iLLiCiTiT (approved):

PR #2800, opened 2022-02-23
Ftrack: Unset task ids from asset versions before tasks are removed

iLLiCiTiT:
## Brief description
Ftrack delete action still crashes on `CircularDependencyError`. Seems that changes from [this PR](https://github.com/pypeclub/OpenPype/pull/1579) are not enough.

## Description
It seems that the circular dependcy happens because is trying to delete asset versions that have relation to removed task.

## Changes
- unset `task_id` from AssetVersions that are related to tasks which are going to be removed

## Test notes
1. Try run ftrack Delete Asset/Subset action in ftrack on entities with tasks that have asset versions assigned

## Note
Not sure if this will fix the issue. The error happens without more details, this is just a guess fix.


mkolar:
Task linked: [OP-2737 Ftrack circular dependency error](https://app.clickup.com/t/6658547/OP-2737)

64qam (approved):

Issue #969, opened 2021-02-03
Manager: Update all to latest button


tokejepsen:
**Is your feature request related to a problem? Please describe.**
When updating many containers in the manager, it takes time and clicks to select them > right-click > update to latest.

**Describe the solution you'd like**
Add a button at the top to update all outdated containers to latest version.

![Capture](https://user-images.githubusercontent.com/1860085/106766528-df41e380-6631-11eb-8e72-7ffc172607a9.PNG)

[cuID:PYPE-1209]

mkolar:
I agree. Small thing but certainly a pleasant impact on users

BigRoy:
For the time being I tend to recommend to:
- enable filter to outdated (top right), 
- select all
- right click > update to latest

PR #2805, opened 2022-02-24
Manager: Update all to latest button

BigRoy:
## Brief description

A draft implementation for #969 to add a "Update all outdated" button to the Manager / Scene Inventory.

update_all_to_latest

## Description

Clicking the button will update only the items that are currently shown as outdated. It will **not** query the database to see if since the last refresh of the manager there were any new updates. This just made most sense to me - it'll only update what I currently can see is outdated.

## Additional info

- We might want to add a confirmation box "Are you sure you want to update all?" because updating tons of assets can be slow, might not be undoable and users might accidentally click the button next to the refresh button.
- It will always update **all outdated** even when the user had a Search entry at the top bar or if he had selected any items in the view. User selection/search is completely ignored and all entries in the model are considered. It **does** however filter to Cherry-Picked hierarchies (right click, cherry pick) because in that case the source model is already 'filtered'.
- Would be nice if the button would disable `setEnabled(False)` if no outdated entries currently exist in the model. However the model isn't currenty optimized to easily tell us whether there is any outdated entry. We could cache that.
- The logic to update from the model is a bit scattered and copied around. We might want to simplify the logic and separate it some more into re-usable functions.
- The logic to update is currently on the Window but might be better suited on the SceneInventoryView instead - so that the Window just calls `view.update_all_to_latest()`
- Also fixed some code typos.

⚠️ See notes above ☝️  - I'll keep this up for review but I wouldn't say it's polished and mergable yet.

## Testing notes:

1. Test whether updating all works ✅ 
2. There's "groups in groups" functionality in the manager which I haven't seen used much. Make sure it also works for those cases. ❓ 
3. The manager can also list containers that have invalid representation ids - make sure those don't break the update all feature. Easiest way is to cripple a container and change the `id` value. I recall seeing "invalid entries" but as I tested just now it seemed to just cripple the full model and show nothing instead. (a separate bug?) ❓


mkolar:
@iLLiCiTiT could you please check this out. Functionally this is good with me

openpype/tools/sceneinventory/window.py
```diff
@@ -158,6 +171,42 @@ def _on_outdated_state_change(self):
             self._outdated_only_checkbox.isChecked()
         )
 
+    def _on_update_all(self):
```
iLLiCiTiT, line 174: For now I would move this into view. For example the `_show_version_error_dialog` is not availabe here.

BigRoy: Moved with https://github.com/pypeclub/OpenPype/pull/2805/commits/882a17b04a164c833de3bdbca06a1af5505eaf78

openpype/tools/sceneinventory/view.py
```diff
@@ -796,3 +796,40 @@ def _show_version_error_dialog(self, version, items):
         ).format(version_str)
         dialog.setText(msg)
         dialog.exec_()
+
+    def update_all(self):
+        """Update all items that are currently 'outdated' in the view"""
+        # Get the source model through the proxy model
+        model = self.model().sourceModel()
```
BigRoy, line 803: I'm personally not too big of a fan to hardcode this way to get to the source model instead of the proxy. This is solely to preserve the behavior as in the original comment of this PR. E.g. search does not influence the update all, etc.

iLLiCiTiT: Me too, but both ways are not ideal, this way is at least update logic at one place.

All tools will need to be re-done in a way that their core logic won't be in Qt classes at all so I would not care about perfect solution now.

iLLiCiTiT (approved):

PR #2806, opened 2022-02-24
Flame Babypublisher optimalization

jakubjezek001:
## Brief description
Addressing client's feedback

## Description
- renamed to Babypublisher as it is already internally used a lot
- taking shot name form Shot name segment attribute
- ignoring hidden and empty tracks and hidden segments
- debugging ftrack multiple created tasks on one asset entity (was crashing because of duplicity)


mkolar:
Task linked: [OP-2735 Flame: babyPublisher shot name form sequence shot name attribute](https://app.clickup.com/t/6658547/OP-2735)

iLLiCiTiT (approved):
I can't test it but the code doesn't look to break anything which was already there :)

openpype/hosts/flame/startup/openpype_babypublisher/openpype_babypublisher.py
```diff
@@ -16,10 +16,11 @@ def flame_panel_executor(selection):
     if "panel_app" in sys.modules.keys():
         print("panel_app module is already loaded")
         del sys.modules["panel_app"]
+        import panel_app
+        reload(panel_app)  # noqa
```
BigRoy, line 20: Python 3 has no `reload`.

Latest flame versions use Python 3.7 according to [this](https://knowledge.autodesk.com/support/flame-products/learn-explore/caas/CloudHelp/cloudhelp/2022/ENU/Flame-WhatsNew/files/GUID-112C2B67-4D0E-473B-9C14-A7B7BA5EDE80-htm.html).

BigRoy:
> renamed to Babypublisher as it is already internally used a lot

Somehow I feel **Babypublisher** should really need some explaining. Should at least a docstring somewhere describe why it's named like that? To me it's not an obvious choice. Unless we're actually publishing babies here. 👶

iLLiCiTiT:
> Should at least a docstring somewhere describe why it's named like that?

I agree @jakubjezek001 

Babypublishing is focused for one specific client usage who needed a really quick solution. It's not something that could be considered as proper flame implementation (at least from my point of view).

PR #2810, opened 2022-02-25
Settings UI: Search case sensitivity

iLLiCiTiT:
## Brief description
Search in settings is not case sensitive.

## Description
Case sensitivity is not welcommed as searching of `jpeg` won't find `JPEG`.

## Changes
- regex is case insensitive
- fixed filter text change timer to be singleshot
- fixed issue with missing entity on reset which can happen if schema is invalid or there was an issue with creation

## Testing notes:
1. Open settings UI
2. Open search dialog (CTRL + f)
3. Find something with wrong case sensitivity

### DIFF
--- a/openpype/tools/settings/settings/search_dialog.py
+++ b/openpype/tools/settings/settings/search_dialog.py
@@ -30,7 +30,7 @@ class RecursiveSortFilterProxyModel(QtCore.QSortFilterProxyModel):
         regex = self.filterRegExp()
         if not regex.isEmpty() and regex.isValid():
             pattern = regex.pattern()
-            compiled_regex = re.compile(pattern)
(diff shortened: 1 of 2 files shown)


openpype/tools/settings/settings/search_dialog.py
```diff
@@ -75,6 +75,7 @@ def __init__(self, parent):
 
         filter_changed_timer = QtCore.QTimer()
         filter_changed_timer.setInterval(200)
+        filter_changed_timer.setSingleShot(True)
```
BigRoy, line 78: Ah, good catch. :)

BigRoy (approved):
Works for me.

64qam (approved):

PR #2811, opened 2022-02-25
Nuke: adding Reformat to baking mov plugin 

jakubjezek001:
## Brief description
Reformat node is definable in settings

## Description
- Reformat node is disabled by default and needs to be activated. 
- Added subset name filtering on baking profiles
- reformat knobconfig with defined types
- extract review condition to keep input resolution if `reformated` tag in representation

## Testing notes:
1. Start nuke in your testing workfile, make sure it is more than fullHD resolution
2. Enable `project_settings/nuke/publish/ExtractReviewDataMov/outputs/baking/reformat_node_add`
3. make sure your Extract review preset is not having reformat to fullHD resolution (project_settings/global/publish/ExtractReview/profiles/0/outputs/h264/width or height is set to 0)
4. publish render write
5. notice that extracted review mp4 is fullHD resolution rather then the original workfile resolution


mkolar:
Task linked: [OP-2739 Nuke - Add Reformat node for baked Mov](https://app.clickup.com/t/6658547/OP-2739)

64qam:
It is working well. The only thing is that the node graph is a bit messy after publishing 
image

64qam (approved):
Looks good now

PR #2812, opened 2022-02-25
global: letter box calculated on output as last process

jakubjezek001:
## Brief description
Extract review Letter box enhancements

## Description
Clients were complaning the process of calculation for letterbox was too complex and not very transparent. We had simplified it to the point that the letterbox is calculated from output resolution after an overscan crop and reformat. 

## Additional info
In case where no overscan and reformat is defined as zeros, so source resolution is applied, then letter box is applied and calculated to source resolution.


mkolar:
Task linked: [OP-2758 Letter box Ratio](https://app.clickup.com/t/6658547/OP-2758)

openpype/plugins/publish/extract_review.py
```diff
@@ -996,70 +993,119 @@ def get_letterbox_filters(
...
         line_color_alpha = float(l_alpha) / 255
-
-        if input_res_ratio == output_res_ratio:
-            ratio /= pixel_aspect
-        elif input_res_ratio < output_res_ratio:
-            ratio /= scale_factor_by_width
-        else:
-            ratio /= scale_factor_by_height
-
+        height_letterbox = int(output_height - (output_width * (1 / ratio)))
```
iLLiCiTiT, line 996: I think you broke how pillar boxes are calculated with this change. The ratio recalculation is here due to state of letterbox or pillarbox.

openpype/plugins/publish/extract_review.py
```diff
@@ -996,70 +993,119 @@ def get_letterbox_filters(
...
-                    ratio, line_thickness, line_color_hex, line_color_alpha
+                    widht=output_width,
+                    height=output_height,
+                    ratio=ratio,
+                    l_thick=line_thickness,
+                    l_color=line_color_hex,
+                    l_alpha=line_color_alpha
                 )
-                output.extend([top_line, bottom_line])
+                if height_letterbox > 0:
```
iLLiCiTiT, line 1053: All these conditions should not be here or should be at the beginning because this whole function can be skipped when it's true.

jakubjezek001: eh.. yeah.. good catch ;)

iLLiCiTiT:
It seems that the `letterbox` and `pillarbox` states are not used at all and their existence doesn't make much sense. I would suggest remove it and change logic of lines draw based on ration input vs. output.

Also the letterboxes are drawed before rescale which also does not make much sense as rescale can pad the output.

jakubjezek001:
> It seems that the `letterbox` and `pillarbox` states are not used at all and their existence doesn't make much sense. I would suggest remove it and change logic of lines draw based on ration input vs. output.
> 
> Also the letterboxes are drawed before rescale which also does not make much sense as rescale can pad the output.

I had removed the enumerator and all is done by dynamic switch: if preset ratio is smaller then image retio > create pillar box; if otherwise do letter box.

openpype/plugins/publish/extract_review.py
```diff
@@ -1319,6 +1345,22 @@ def rescaling_filters(self, temp_data, output_def, new_repre):
                 "setsar=1"
             ])
 
+        # letter_box
+        if letter_box_enabled:
+            filters.extend([
```
iLLiCiTiT, line 1350: Output should be already scaled up so this filter is not needed.

openpype/plugins/publish/extract_review.py
```diff
@@ -1319,6 +1345,22 @@ def rescaling_filters(self, temp_data, output_def, new_repre):
...
+                "scale={}x{}:flags=lanczos".format(
+                    output_width, output_height
+                ),
+                "setsar=1"
+            ])
```
iLLiCiTiT, line 1355: Output should be already scaled up so this filter is not needed.
```suggestion
```

openpype/plugins/publish/extract_review.py
```diff
@@ -997,68 +993,117 @@ def get_letterbox_filters(
...
+        pillar = output_ratio > ratio
+        need_mask = format(output_ratio, ".3f") != format(ratio, ".3f")
 
-        if state == "letterbox":
+        if need_mask and not pillar:
             if fill_color_alpha > 0:
                 top_box = (
-                    "drawbox=0:0:iw:round((ih-(iw*(1/{})))/2):t=fill:c={}@{}"
-                ).format(ratio, fill_color_hex, fill_color_alpha)
+                    "drawbox=0:0:{widht}:round("
```
iLLiCiTiT, line 1004: All have typo `widht` -> `width`.

iLLiCiTiT:
Submodule `avalon-unreal-integration` was added back. Please remove it.

iLLiCiTiT (approved):
Did some smaller tweaks with pillar boxes and it works.

PR #2814, opened 2022-02-26
Modules: Remove 'default_modules' folder logic

BigRoy:
## Brief description

This removes the old logic used to load modules from folder `default_modules`. With the recent move of all modules that logic has been replaced and the default folder does not exist thus this didn't do anything by default.

It also avoids the warning on startup:
```
*** WRN: >>> { ModulesLoader }: [  Could not find path when loading OpenPype modules "\openpype\modules\default_modules"  ] 
```

## Testing notes:
1. Test whether everything for modules still works.

### DIFF
--- a/openpype/modules/base.py
+++ b/openpype/modules/base.py
@@ -143,13 +143,6 @@ class _LoadCache:
     modules_loaded = False
 
 
-def get_default_modules_dir():
-    """Path to default OpenPype modules."""
-    current_dir = os.path.abspath(os.path.dirname(__file__))
-
-    return os.path.join(current_dir, "default_modules")
-
-
 def get_dynamic_modules_dirs():
     """Possible paths to OpenPype Addons of Modules.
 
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
I would not remove this logic as few features in development are still using the ability of that folder. If the warning is only reason of this PR then there are better solutions for that.

BigRoy:
Thanks. Not sure what modules in development you are talking about. An open PR? Shotgun/Kitsu modules?

Anyway, I see now I never ended up posting my edited comment with the warning that if any studio relied on putting things into the default modules folder they'd also be affected.

How about instead of removing the logic to log a deprecation warning if any modules are found in that folder?

iLLiCiTiT:
Deprecation warning for `default_modules` is good. This means to do the existing path check inside `get_default_modules_dir` and do not return invalid path.

I was also thinking about adding 1 other default path `~/modules/addons` to have ability develop addons faster. As starting point it is good to have some kind of folder inside openpype that is loaded without changing settings even if is not existing in git.

BigRoy:
Closing this in favor of recent changes on how modules are loaded since this code is outdated.

PR #2815, opened 2022-02-27
General: Inclusive handles

iLLiCiTiT:
## Brief description
Added ability to have inclusive handles per project.

## Description
Inclusive handles changes what `frameStart` and `frameEnd` stored on entities represents. With inclusive handles they represent full range. Even when handles should not be used we need to know their value to be able know frame range.

| | Frame Start | Frame End | Handle Start | Handle End | Full range | Range without handles |
| --- | --- | --- | --- | --- | --- | --- |
| Extend handles | 1001 | 1100 | 10 | 10 | 991-1110 | 1001-1100 |
| Inclusive handles | 1001 | 1100 | 10 | 10 | 1001-1100 | 1011-1090 |

It requires a lot of changes when data about frames are loaded from asset and version because they need recalculations.

To keep the code simpler it was created a unified frame info which is used when frame data are loaded from asset or version document. Based on anatomy attribute `handlesExtend` is recalculating frame ranges. This allows to keep current code unchanded in most of cases as inclusive handles are converted into extending.

I hope I've changed all publishing and loading code where frame start and frame end are used when reading from asset and version document.

New Anatomy key `handlesExtend` was also added to ftrack custom attributes.

## TODOs
Editorial publishing was not modified. I couldn't find out how handles are handles handled in that case. I know that editorial can create hierarchy (asset documents) where it should be stored the way project expect it so I'll need @jakubjezek001 help.

It is not properly validated if the frame range is valid.

## Notes
Changing the `handlesExtend` during production would probably break already existing version documents.

## Testing notes
1. Change `handlesExtend` to `False` on a project.
    - since that moment all frames should be recalculated
2. Check if frames are calculated in right way.

If `handlesExtend` is set to `True` all should work as before.


mkolar:
Task linked: [OP-2585 Frame range with included handles](https://app.clickup.com/t/6658547/OP-2585)

openpype/lib/avalon_context.py
```diff
@@ -261,6 +262,200 @@ def get_hierarchy(asset_name=None):
...
+
+    @property
+    def frame_range_without_handles(self):
+        """Frame range without handles."""
+        return (self.frame_end - self.frame_start) + 1
+
+    @property
+    def frame_range(self):
+        """Frame range with handles."""
+        return (self.handle_frame_end - self.handle_frame_start) + 1
```
BigRoy, line 363: To me these aren't frame ranges but `durations` (or "number of frames"). A frame range to me consists of two values, like a start and end.

The loader currently shows these as **Duration** (excluding the handles):
afbeelding

openpype/plugins/publish/collect_avalon_entities.py
```diff
@@ -59,43 +64,22 @@ def process(self, context):
...
-        context.data["frameStartHandle"] = frame_start_h
-        context.data["frameEndHandle"] = frame_end_h
+        anatomy = context.data["anatomy"]
+
+        frame_info = get_frame_info(asset_entity, anatomy)
+        if frame_info is None:
+            frame_info = UnifiedFrameInfo(1, 2, 0, 0)
+            self.log.warning(
+                "Missing frame information on asset. Using default 1-2 frames."
+            )
```
BigRoy, line 74: I see that the old logic does the same - but why would it default to 1-2 instead of just a single frame 1-1?

openpype/plugins/publish/integrate_new.py
```diff
@@ -923,15 +923,53 @@ def create_version_data(self, context, instance):
...
+            optional_values["frameStart"],
+            optional_values["frameEnd"],
+            optional_values.get("handleStart"),
+            optional_values.get("handleEnd"),
+            anatomy
+        )
+        optional_values["frameStart"] = frame_info.real_frame_start
+        optional_values["frameEnd"] = frame_info.real_frame_end
+        optional_values["handleStart"] = frame_info.handle_start
+        optional_values["handleEnd"] = frame_info.handle_end
```
BigRoy, line 970: I think this function or just above these lines it's best to have a comment in the code as to why this is necessary. Something like:
```# We always store the frame start and frame end into the backend with handles excluded (extend) so we need to convert based on how the project is set.
```

Also, I'd change the logic so that only during Collect we retrieve the `UnifiedFrameInfo` and pass that along with the instance as `instance.data["frameInfo"]` or alike and then we don't need to duplicate the logic and build it again here. Or why would we want this code here too?

openpype/lib/avalon_context.py
```diff
@@ -261,6 +262,200 @@ def get_hierarchy(asset_name=None):
...
+        self._handles_extend = True
+        self.change_handles_extend(handles_extend)
+
+        self._log = None
+
+    @property
+    def log(self):
+        if self._log is None:
+            self._log = PypeLogger.get_logger(self.__class__.__name__)
+        return self._log
```
BigRoy, line 319: Log seems unused on this class. Remove it?

openpype/lib/avalon_context.py
```diff
@@ -261,6 +262,200 @@ def get_hierarchy(asset_name=None):
     return "/".join(hierarchy_items)
 
 
+class UnifiedFrameInfo:
```
BigRoy, line 265: This class really needs some tests. :)

openpype/plugins/publish/integrate_new.py
iLLiCiTiT: Because these are data that were collected on instance and may not be the same as are on asset document.

openpype/plugins/publish/collect_avalon_entities.py
iLLiCiTiT: To be honest, I don't know.

openpype/lib/avalon_context.py
iLLiCiTiT: Agree. Other attributes are clear?

openpype/plugins/publish/integrate_new.py
BigRoy: > Because these are data that were collected on instance and may not be the same as are on asset document.

I understand why we are doing the conversion at the end. I am just saying that we'd already construct the frame info during collect and use that one here too

openpype/lib/avalon_context.py
BigRoy: I had to re-read the class multiple times and I am still not sure I fully grasp the variables. So at first sight. No.

Let me read up again coming week when I have some time to let this sink in and see if I can come up with better input.

openpype/plugins/publish/integrate_new.py
iLLiCiTiT: > I am just saying that we'd already construct the frame info during collect and use that one here too

Not exactly. Frame info on instance can be collected from workfile/scene and can be different than frame info from asset.

openpype/hosts/maya/plugins/load/actions.py
```diff
@@ -29,18 +30,27 @@ def load(self, context, name, namespace, data):
...
+            ))
             return
 
-        cmds.playbackOptions(minTime=start,
-                             maxTime=end,
-                             animationStartTime=start,
-                             animationEndTime=end)
+        handle_start = version_data.get("handleStart")
+        handle_end = version_data.get("handleStart")
+        frame_info = get_frame_info(
```
BigRoy, line 45: The version data in the database/backend will always be inclusive. No? The database need to either choose one of the two formats or additionally should publish along whether it's inclusive or not. But `get_frame_info` applies the current settings to it and thus results in different frame ranges to be set. However once the project settings change that should NOT influence already published content.

Again. The handles being inclusive or not is solely a visual thing:
1. On create instances.
2. In the tools that display frame info.

openpype/plugins/publish/integrate_new.py
iLLiCiTiT: Modified it and version document always use extending handles.

openpype/hosts/maya/plugins/load/actions.py
antirotor: > And created instance don't have frame info in metadata, as far as I know.

They actually have:
image

BigRoy: That ☝️ - and I'd still prefer changing the admin setting doesn't influence my existing scenes so I'd recommend adding a "how to interpret handles" attribute onto the instances.

iLLiCiTiT: And where they come from? It's taken from Maya timeline?

antirotor: > And where they come from? It's taken from Maya timeline?

yes, these all taken from timeline set in Maya that is initialized from asset data (but can be changed during the work ofc)

antirotor:
yeah, I vote for additive/subtractive handles

openpype/hosts/maya/plugins/load/actions.py
iLLiCiTiT: So they're always stored in scene as additive handles, no matter what is set on project.

BigRoy: So even for the studios using the subtractive handles the settings on instances in the scene are considered additive?

I am not sure I understand. Where then does it actually influence the experience for those users?

iLLiCiTiT: I would say that only place where it is affected are asset documents (project management).

I had in mind that all other places are defined by dcc itself (e.g. maya timeline probably can't show subtractive handles, or?) or are "metadata" driven so "User"/"Artist" should not see and care about it.

But as I can see I am wrong because these are "metadata" but can be modified by an artist. I wonder how he know that he should set them (e.g. `handleStart` or `handleEnd`)?

The question is if this should be affected by project settings too? If yes then are there other places where frame ranges are defined in metadata which are meant to be changed by an artist?

jakubjezek001:
From my perspective it would be better ti have anatomy attribute name 'handlesInclusive' and during publishing of versions to add the attribute to version data. So if the attribute is not available - it is exclusive, and we keep backward compatibility even this is changing during a ongoing production. 

Regarding editorial, only ideal plugin to dealnwith this is the global 'CollectHierarchy' and that is what you have done, so :+1:

iLLiCiTiT:
> From my perspective it would be better ti have anatomy attribute name 'handlesInclusive' and during publishing of versions to add the attribute to version data

This would mean to change code of all loaders using those data. I'm not against, but I'm not capable. That was the reason why additive/subtractive handles are handled only on asset documents in this PR.

mkolar:
After some extensive discussion internally, we're pausing dev on this. We believe it is an unnecessary complication at the moment, until we get a more solid usecase.

PR #2819, opened 2022-02-28
Ftrack: Job killer with missing user

iLLiCiTiT:
## Brief description
Job killer action can crash if job's username was removed from ftrack.

## Description
Added fallback for cases when job does not have valid user id and made job killer faster using explicit queries.

## Testing notes:
1. Try to kill jobs without valid user (not sure how to do that without removing ftrack user)


mkolar:
Task linked: [OP-2514 Ftrack: Job killer kill jobs without user](https://app.clickup.com/t/6658547/OP-2514)

64qam (approved):

PR #2820, opened 2022-03-01
Settings UI: Fix "Apply from" action

iLLiCiTiT:
## Brief description
Apply from actions were using lambda callbacks created inside loop but it looks like lambda will keep using variable values from the scope in which was created so the last project name was always used.

## Changes
- create custom class for action callbacks `_Callback` which stores function to call and arguments to be used
- replace `lambda` with `_Callback` object

## Testing notes:
1. Open project settings UI
2. Set some settings in one project and make sure they are different then all other projects
    - don't do that on last project in list
3. Run Apply from -> "the project from step 2." action in other project and validate that values are the same

### DIFF
--- a/openpype/tools/settings/settings/base.py
+++ b/openpype/tools/settings/settings/base.py
@@ -12,6 +12,22 @@ from .lib import create_deffered_value_change_timer
 from .constants import DEFAULT_PROJECT_LABEL
 
 
+class _Callback:
+    """Callback wrapper which stores it's args and kwargs.
+
+    Using lambda has few issues if local variables are passed to called
+    functions in loop it may change the value of the variable in already
+    stored callback.
+    """
+    def __init__(self, func, *args, **kwargs):
+        self._func = func
+        self._args = args
+        self._kwargs = kwargs
+
+    def __call__(self):
+        self._func(*self._args, **self._kwargs)
+
+
 class BaseWidget(QtWidgets.QWidget):
     allow_actions = True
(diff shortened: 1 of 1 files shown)


mkolar:
Task linked: [OP-2753 Project Settings Apply values from ...](https://app.clickup.com/t/6658547/OP-2753)

64qam (approved):

openpype/tools/settings/settings/base.py
```diff
@@ -325,7 +341,11 @@ def _apply_values_from_project_action(self, menu, actions_mapping):
...
             action = QtWidgets.QAction(project_name)
             submenu.addAction(action)
-            actions_mapping[action] = lambda: self._apply_values_from_project(
+            # Use custom callback object instead of lambda
+            #   - project_name value is changed each value so all actions will
+            #       use the same source project
+            actions_mapping[action] = _Callback(
+                self._apply_values_from_project,
                 project_name
             )
```
BigRoy, line 350: Shouldn't this just use `functools.partial(self._apply_values_from_project, project_name)` instead of its own Callback wrapper class?

iLLiCiTiT: Yep... I forgot about `functools`.

iLLiCiTiT: Changed in https://github.com/pypeclub/OpenPype/pull/2825

PR #2822, opened 2022-03-01
General: Custom function for find executable

iLLiCiTiT:
## Brief description
We're using `find_executable` from distutils but it seems that they'll get deprecated soon and may not work as expected also some older hosts have issues with using distutils too.

## Description
Added custom `find_executable` which should return full path to an executable file.

## Changes
- implemented `find_executable` in `openpype/lib`
- use the function in places where it is needed
- `maketx` in extract look does not add `".exe"`

## Testing notes:
- try run applications
- oiio and ffmpeg functions returning path should work and return full path to file

### DIFF
--- a/openpype/lib/vendor_bin_utils.py
+++ b/openpype/lib/vendor_bin_utils.py
@@ -58,7 +58,7 @@ def find_executable(executable):
     paths = path_str.split(os.pathsep)
     for path in paths:
         for variant in variants:
-            filepath = os.path.abspath(os.path.join(path, executable))
+            filepath = os.path.abspath(os.path.join(path, variant))
             if os.path.isfile(filepath):
                 return filepath
     return None
--- a/openpype/lib/vendor_bin_utils.py
+++ b/openpype/lib/vendor_bin_utils.py
@@ -7,6 +7,25 @@ import subprocess
 log = logging.getLogger("Vendor utils")
 
 
+def is_file_executable(filepath):
+    """Filepath lead to executable file.
+
+    Args:
+        filepath(str): Full path to file.
+    """
+    if not filepath:
+        return False
+
+    if os.path.isfile(filepath):
+        if os.access(filepath, os.X_OK):
+            return True
+
+        log.info(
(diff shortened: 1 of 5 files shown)


openpype/lib/vendor_bin_utils.py
```diff
@@ -3,9 +3,65 @@
...
+    if not path_str:
+        return None
+
+    paths = path_str.split(os.pathsep)
+    for path in paths:
+        for variant in variants:
+            filepath = os.path.abspath(os.path.join(path, executable))
+            if os.path.isfile(filepath):
+                return filepath
+    return None
```
BigRoy, line 67: Out of curiosity. What is this logic based on? It looks like `distutils.find_executable` but does not [look like recent code](https://github.com/python/cpython/blob/main/Lib/distutils/spawn.py#L95-L129)?

iLLiCiTiT: Source was distutils, but slightly modified. In distutils `find_executable` always expect `".exe"` on windows, this adds other options from `PATHEXT` and `.sh` for linux and mac.

BigRoy: For reference sake this is what [`which`](https://github.com/getavalon/core/blob/master/avalon/lib.py#L103-L124) does in Avalon core which looks somewhat similar

iLLiCiTiT: Handle less cases. When passed `executable` is existing path it should just return it. If the passed `executable` is path but don't have extension then this function will just add extension and return it. Then will start to look into PATH for possible variants.
Adding the `os.access(filepath, os.X_OK)` seems like a good idea.

antirotor:
Just warning about `os.access(filepath, os.X_OK)` as it might fail on some storage configurations - we've seen this already

kalisp (approved):
Tested in Windows on AE and Nuke, locally and on Deadline, works fine.

Should be probably tested on different OSes.

iLLiCiTiT:
> Should be probably tested on different OSes.

@64qam could you find a time to test this on linux?

PR #2826, opened 2022-03-02
Maya: Deformer node ids validation plugin

2-REC:
## Brief description
In Maya, deformers can create copies of shape nodes without keeping the cbId from the original nodes, breaking subsequent references.

## Description
In Maya, when a deformer is added to a mesh, a new shape node is created, copy of the original shape node.
The "deformed" shape node will hold the result of the deformation applied to the original shape node.
The new node doesn't get the cbId of the original node.

Depending on the deformer, the new node will be used as the original node or as the deformed node. Additionally, if applying more deformers, more nodes can be created.
In many cases, the deformed node will not have a cbId and will get a newly generated one, unmatching the original node cbId.

This causes problems downstream in the pipeline, as any reference to the original node's cbId will not match with the deformed shape node's cbId.

## Additional info
Typically, this issue will occur when assigning a published look to a rig.

To reproduce the problem, the following steps can be executed:
* Create and publish a model.
* Import the published model in a new file ("Load" as a reference), create a look for that model and publish it.
* Import the published model in a new file ("Load" as a reference), create a rig for that model (with a skin deformer) and publish it.
* In a new scene, import the published rig, and using the "Look Assigner" tool, assign the published look to the rig.

As a result, the look is not assigned to the mesh, and the following message is displayed in the "History" panel:
```
# Warning: None of the items can be added to the set # 
```

The issue is that the process is trying to assign the shaders only to the original shape nodes (referred to by their cbId), which are marked as "intermediate objects" (because of the deformer applied to them).
Shaders cannot be assigned to these nodes, and no other nodes match the cbIds, hence the shaders aren't assigned to anything.

## Solution

To fix this issue, and to allow downstream tasks to refer to the deformed nodes as the original nodes without distinction, the solution is to get the cbId from the original shape nodes and "transfer" them to the new deformed shape nodes.

In general, the original shape node can be retrieved from the deformed shape node's history.

In some cases however, it is not possible to retrieve the original shape node this way, as another copy of the original node has been made and is now used as the original for the deformer.

For example, consider the following node graph.
deformer_graph

The shape node "MeshShapeDeformedOrig" is used as the original node for the deformer, but its relation with the original "MeshShape" has been lost.

In this case, it is not possible to get the original shape node from the deformed shape node's history (only the new node which is a copy of the original node).

The nodes do however share the same transform node parent, and are related in that way.

Following is a suggested solution:
The original shape node can be found as a sibling of the deformed shape node.
As mentioned earlier, the original node should have the attribute "intermediate object" set.
Additionally, the node should also have a cbId assigned to it, as opposed to the new node.

A node satisfying these conditions should be the original shape node.
The cbId of this node can then be transferred to the deformed node.


2-REC:
> I believe you might even get by with only adding `rig` family to [`validate_node_ids_deformed_shapes`](https://github.com/pypeclub/OpenPype/blob/7017410173647da9a122e0659fda2c7af5593c80/openpype/hosts/maya/plugins/publish/validate_node_ids_deformed_shapes.py).
Thanks for the quick feedback!

I thought that as well, but the issue is that the original shape node might be in the deformed shape node's history anymore. I am not sure how this happened (I didn't do the rig), but the deformed shape node's history refers to a new "shapeOrig" node, and "lost" the connection with the original node.
This is why in my case, looking at the history didn't work.
All the names share the same parent though, and the relation can be retrieved this way.

Maybe a mix between the 2 solutions would be ideal (in case the history isn't enough, look at the siblings).

2-REC:
I saw your other comments after posting, so had to delete my previous comment ;)

As you suggest, a mix between the 2 solutions would probably be ideal (in case the history isn't enough, look at the siblings).
A flag "history_only" seems perfect.


As for the referenced meshes, yes, it seems to happen in that case (I think it might be related to foster parents nodes when trying to modify the referenced nodes), though I feel like I had cases where the problem also occurred with imported meshes.
I will try again and confirm (or not).

2-REC:
> Another note: It's also technically possible to have multiple mesh shapes under a single transform, which will 'invalidate' with this current logic. I have that seen being used very rarely but I have seen it. It's something to keep in mind at least.

Indeed.
A rare case, but possible...
Not sure how to solve this case, maybe by name similarities?

BigRoy:
> I thought that as well, but the issue is that the original shape node might be in the deformed shape node's history anymore. I am not sure how this happened (I didn't do the rig), but the deformed shape node's history refers to a new "shapeOrig" node, and "lost" the connection with the original node.
> This is why in my case, looking at the history didn't work.

This can happen when the rigger did for example a delete history in-between. (which I believe would disconnect) But Maya being Maya there's likely a lot more reasons that could happen.

> Indeed.
> A rare case, but possible...
> Not sure how to solve this case, maybe by name similarities?

I'd say for now we might want to add this Validator in Admin Settings so that it can be enabled/disabled. Then if someone ever has the case where they need to have multiple shapes under a transform like that, they can disable it for themselves. I wouldn't say it's directly problematic but it's what makes this a potentially difficult case to "always solve correctly".

2-REC:
OK, so I can refactor the lib function 'get_id_from_history' as you suggested, and use it in this plugin with the added parameter 'history_only' to False.
If it seems ok for you, I will make the changes now and commit asap.

About the admin settings, I could have a look, but it might be a bit complicated as I haven't touched that (yet).

BigRoy:
> OK, so I can refactor the lib function 'get_id_from_history' as you suggested, and use it in this plugin with the added parameter 'history_only' to False.
> If it seems ok for you, I will make the changes now and commit asap.

I think you could remove your Validator and apply your tweak in this [Validate Rig Out Set Node Ids](https://github.com/pypeclub/OpenPype/blob/7017410173647da9a122e0659fda2c7af5593c80/openpype/hosts/maya/plugins/publish/validate_rig_out_set_node_ids.py) so that it does `get_id_from_siblings(history_only=self.allow_history_only)`. Then at the top of the class below `actions = []` add `self.allow_history_only = False`.

Then we can expose that particular True/False settings in Admin Settings - the plug-in would need to be added [in this .json file](https://github.com/pypeclub/OpenPype/blob/develop/openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_publish.json). Likely we'll want to add it inside the [Rig validators](https://github.com/pypeclub/OpenPype/blob/develop/openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_publish.json#L497-L500) entry where you'd want to expose the `allow_history_only` **key** similar to how it's done [here for another validator](https://github.com/pypeclub/OpenPype/blob/develop/openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_publish.json#L224-L247). Let me know if you can figure it out like that or need more input.

Once you have it added in the settings all you'd need to do is set the default settings values. You can do that by running `tools/run_settings.ps1`, then Save with "Modify Defaults" enabled on the Project tab (since this is for project settings). That will regenerate the relevant defaults .json file.

2-REC:
OK, perfect.
Thank you for all the details.

I should be able to do it, I think I have enough info.
I won't have time to finish today, but will definitely have it done tomorrow.

openpype/hosts/maya/api/lib.py
```diff
@@ -1791,6 +1797,22 @@ def _get_parent(node):
...
+        # Get siblings of same type
+        similar_nodes = cmds.listRelatives(parent,
+                                           type=node_type,
+                                           fullPath=True)
+        # Exclude itself
+        similar_nodes = [x for x in similar_nodes if x != node]
+
+        for similar_node in similar_nodes:
+            # Check if "intermediate object"
+            if cmds.getAttr(similar_node + ".io"):
```
BigRoy, line 1810: Can we use the full attribute name `.intermediateObject` here? For readability.

openpype/hosts/maya/api/lib.py
```diff
@@ -1751,18 +1751,24 @@ def remove_other_uv_sets(mesh):
...
-def get_id_from_history(node):
+def get_id_from_sibling(node, history_only=True):
     """Return first node id in the history chain that matches this node.
 
     The nodes in history must be of the exact same node type and must be
     parented under the same parent.
 
+    If no matching node is found in history, the siblings of the node
+    are checked. Additionally to having the same parent, the sibling must
+    be marked as 'intermediate object'.
```
BigRoy, line 1762: The description here and also in the *Args* description down below this feels incorrect. It **always** checks the siblings but only takes those that are also in the history. Right?

openpype/hosts/maya/plugins/publish/validate_rig_out_set_node_ids.py
```diff
@@ -51,10 +52,12 @@ def get_invalid(cls, instance):
                          noIntermediate=True)
 
         for shape in shapes:
-            history_id = lib.get_id_from_history(shape)
-            if history_id:
+            sibling_id = \
+                lib.get_id_from_sibling(shape,
+                                        history_only=cls.allow_history_only)
```
BigRoy, line 57: Should this maybe change in code style to the following?
```
sibling_id = lib.get_id_from_sibling(
    shape,
    history_only=cls.allow_history_only
)
```
I don't recall seeing much of the code style in OP code base that you're using now. (I might be wrong)

BigRoy:
Also don't forget to add the default settings and make sure to commit them:

> Once you have it added in the settings all you'd need to do is set the default settings values. You can do that by running tools/run_settings.ps1, then Save with "Modify Defaults" enabled on the Project tab (since this is for project settings). That will regenerate the relevant defaults .json file.

They currently appear to be missing.

openpype/hosts/maya/api/lib.py
2-REC: I see what you mean, and you are right for the "Args" description.

However, for the main description, I think it is correct:
```
If no matching node is found in history, the siblings of the node are checked.
...
```
The interesting nodes in the history are indeed also siblings, but in the case none is found, the process will look in the siblings regardless of the history.

The way you say it is correct when considering the end result: The siblings are always checked, and we only takes those also in the history (if found). (else we look at all the siblings)
And I think the way I describe it is closer to the process itself: If not found in history (from the siblings), the process looks at (all) the siblings.

But I see your point, and I should probably rewrite it.
(I have never been good at documentation/comments)

openpype/hosts/maya/plugins/publish/validate_rig_out_set_node_ids.py
BigRoy: Actually - what you're showing there would be even better. But I believe it was moved with `\` due to the exceeding the PEP08 line length. In that case the code style I showed in my comment I think is what is preferred here.

openpype/hosts/maya/api/lib.py
```diff
@@ -1791,6 +1797,22 @@ def _get_parent(node):
         if _id:
             return _id
 
+    if not history_only:
+        # Get siblings of same type
+        similar_nodes = cmds.listRelatives(parent,
+                                           type=node_type,
```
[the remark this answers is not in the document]
2-REC: Yes, this seems indeed like an issue.
Thank you for the detailed explanation!

So I will add the "ls" command with "exactType" parameter after calling "listRelatives" (and after excluding the node itself).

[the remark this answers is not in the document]
BigRoy: Yes, make sure to include `long=True` because afterwards you're excluding itself which needs to match exactly in name. :)

openpype/hosts/maya/plugins/publish/validate_rig_out_set_node_ids.py
```diff
@@ -24,6 +24,7 @@ class ValidateRigOutSetNodeIds(pyblish.api.InstancePlugin):
         openpype.hosts.maya.api.action.SelectInvalidAction,
         openpype.api.RepairAction
     ]
+    allow_history_only = False
```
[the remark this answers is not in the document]
2-REC: Good question.

Preserve backwards compatibility is essential, though in this case it is somehow a "bug fix", which can now be ignored if desired (for specific cases when have more than 1 shape).

So I feel like the default should be set to ```True```, even if it can break backwards compatibility...

But of course I will not make the decision :)

mkolar:
>So I feel like the default should be set to True, even if it can break backwards compatibility...

I actually agree that in this case, it's more a bugfix, than a pure change. If someone already has a scene prepped, it should be validated against this in the new publish, to prevent potential issues.

openpype/hosts/maya/api/lib.py
```diff
@@ -1791,6 +1799,24 @@ def _get_parent(node):
...
+
+        # Exclude itself
+        similar_nodes = [x for x in similar_nodes if x != node]
+
+        for similar_node in similar_nodes:
+            # Check if "intermediate object"
+            if cmds.getAttr(similar_node + ".intermediateObject"):
+                _id = get_id(similar_node)
+                if _id:
+                    return _id
```
BigRoy, line 1817: Should we log a warning if multiple intermediate shapes are found that have different ids with the info that we're taking only one of them? Just so we are aware of those conflicts? The other logic with the "history" took the first one (which is the closest one in history) which makes sense to prioritize. But for siblings I'm not so sure how we could correctly decide which one should be more likely to be correct?

openpype/hosts/maya/plugins/publish/validate_rig_out_set_node_ids.py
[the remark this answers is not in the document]
BigRoy: > So I feel like the default should be set to True, even if it can break backwards compatibility...

Actually - `False` - right? Because when False the siblings without history are also considered? So the current implementation has the correct default. 👍

openpype/hosts/maya/api/lib.py
2-REC: Indeed, so I think this is probably the best solution:
Set the default value to ```False``` ("allow_history_only"), with a warning in the particular case of more than 1 matching sibling.
I will make the changes (add the warning) if you all agree with that.

mkolar:
Just a note that this has now been tested on a studio side with one of our clients as well and works as expected

openpype/hosts/maya/api/lib.py
```diff
@@ -1971,6 +1979,35 @@ def _get_parent(node):
...
+                    # Check if already found an id
+                    if first_id:
+                        log.warning(("Found more than 1 matching intermediate"
+                                     " shape for '{}'. Using id of first"
+                                     " found: '{}'".format(node, found_node)))
+                        break
+                    first_id = _id
+                    found_node = similar_node
+
+        return first_id
```
BigRoy, line 2008: This logic has the downside to also log a warning if both shapes found had the same id - which would still be 'safe'.
We could do `if first_id != _id:` or alike.

Or here's an alternative approach:

```python
# pseudocode
from collections import OrderedDict
...
        
        # Get all unique ids from siblings in order since
        # we consistently take the first one found
        sibling_ids = OrderedDict()
        for similar_node in similar_nodes:
            # Check if "intermediate object"
            if not cmds.getAttr(similar_node + ".intermediateObject"):
                continue
                
            _id = get_id(similar_node)
            if not _id:
                continue
                
            if _id in sibling_ids:
                sibling_ids[_id] = [similar_node]
            else:
                sibling_ids[_id].append(similar_node)
                    
        if sibling_ids:
            # Log a warning if we've found multiple unique ids
            first_id, found_nodes = next(iter(sibling_ids.items()))
            if len(sibling_ids) > 1:
                log.warning(("Found more than 1 intermediate shape with unique"
                             " id: for '{}'. Using id of first"
                             " found: '{}'".format(node, found_nodes[0])))
                break
               
            return first_id
```

2-REC: Ah yes, I didn't know the case with same IDs was valid/possible.
In that case, your alternative approach seems good, I will make the change.

I will check that ok, and update the changes.

BigRoy: > (and the user needs to check by themself anyway, as the tool will not fix it).

Actually, the tool does fix it if you right click -> repair. Anyway, agreed. It's something left to the user to debug.

BigRoy:
@2-REC Great work. Testing this now!

- Can you quickly also [fix up the logged error message](https://github.com/pypeclub/OpenPype/pull/2826/files#diff-47c4492dbbf79fbde2f55b6277e7ef35289fb11a3b7ea5ed8933e3db9c392bfeR36-R37)? It's not an issue introduced by you. But as I started testing I noticed it's off. Something like this?
```python
            raise RuntimeError("Nodes found with mismatching "
                               "IDs: {0}".format(invalid))
```

Other than that I can confirm it does take the id from the sibling if no history connection is present. ✅ 

---

As a TD my preference would be that it'd log clearly where it's taking the ID from but I think most artists would likely say it confuses them more so I'm fine with current behavior. Let's leave it as is for now.

mkolar (approved):

mkolar:
Very nice. Thank you. @BigRoy happy on your end? we can included in 3.9.1 coming out today with a swarm of other fixes

BigRoy (approved):
Great work @2-REC - sorry for the so much back and forth on this. Really appreciate you sticking through it.

PR #2837, opened 2022-03-03
Maya: add loaded containers to published instance

### LINKED ISSUE #2830: Maya: add option to attach loaded containers to specific families
## Feature

For `layout`, `assembly`, `setdress` and possibly `mayaScene` add to Maya Scene Extractor ability to attach `AVALON_CONTAINERS` directly to published data so it will pull those along.

antirotor:
## Enhancement

Adding ability to include loaded containers to published instance.

This is mainly useful for Layouts and Setdress families. When you create them from loaded models for example, this will pull their containers along published Layout so OpenPype will retain information about them when Layout is loaded into the new scene.

This can be set in the Settings (`project_settings/maya/publish/ExtractMayaSceneRaw/add_for_families`):

image

### How to test:

In the new scene load bunch of models with OpenPype, create Layout and publish it. When you load this layout, you should see all content in the layout in Manager.

Close #2830


openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py
```diff
@@ -57,10 +58,22 @@ def process(self, instance):
         else:
             members = instance[:]
 
+        loaded_containers = None
+        if {f.lower() for f in self.add_for_families}.intersection(
```
BigRoy, line 62: Why should this be non case sensitive in the check even though families in Pyblish plugins are case sensitive, right?

Also, are we missing `self.add_for_families` attribute being explicitly set in the code? (Or am I just missing it in the GitHub view somehow?)

Instead of adding these in the extractor - shouldn't these originally be "collected" along instead?

openpype/hosts/maya/plugins/load/load_reference.py
```diff
@@ -158,3 +165,17 @@ def _post_process_rig(self, name, namespace, context, options):
...
+    def _organize_containers(nodes):
+        # type: (list) -> None
+        for node in nodes:
+            id_attr = "{}.id".format(node)
+            if not cmds.attributeQuery("id", node=node, exists=True):
+                print("-" * 80)
+                print("skipping {}".format(node))
+                continue
+            if cmds.getAttr(id_attr) == AVALON_CONTAINER_ID:
+                print("=" * 80)
```
BigRoy, line 179: Should this printing be logging instead? Seems quite verbose?

BigRoy:
What's the reason this setting is configurable? When would you NOT want this for e.g. for layout, etc.?

openpype/hosts/maya/plugins/load/load_reference.py
antirotor: leftover debug prints that shouldn't be there, removed.

openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py
antirotor: `add_for_families` is [added there directly from Settings](https://github.com/pypeclub/OpenPype/blob/1ccc2ad45822f3b9c5b2db29e3000ba158bb5c2b/openpype/lib/plugin_tools.py#L267-L354), no need to manually specify.

With case sensitivity - I am not fan of case sensitive family names as I belive it just adds to chaos - there is no reason why `layout` should be different from `Layout` even more when you allow user to set family name by typing it somewhere in the Settings.

antirotor:
> What's the reason this setting is configurable? When would you NOT want this for e.g. for layout, etc.?

Reason is that it allows you the decision between old behaviour and new one, even if the old one was somewhat limited. It should stay there imho for a while as deprecated feature problably (and I should state it somewhere that it is so).

openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py
BigRoy: I'd avoid the case-insensitivity here. Pyblish plug-in families are case-sensitive. This logic is tied very close to those exact `families`. There are more OpenPype settings exposed that related to Pyblish families and those are also case sensitive.

It's less code/logic to maintain, less cases to test and also more explicit. We're only making things more complicated by allowing the other cases. Plus I'd personally rather not see code with `LaYouT` and have it considered valid too.

To me plug-in families are constants. If we want to avoid human error in the settings then it should be an enum with a list to specify from. User might just as well mistype with a different character as recently was found with aftereffects being mistyped.

> add_for_families is [added there directly from Settings](https://github.com/pypeclub/OpenPype/blob/1ccc2ad45822f3b9c5b2db29e3000ba158bb5c2b/openpype/lib/plugin_tools.py#L267-L354), no need to manually specify.

A code linter will likely not detect that? And most plug-ins that I've seen that specify custom settings also define the attribute on the class for readability. We should find a consistency between those two - I'd prefer the explicit method for readability.

I'd even go as far as not allowing Settings to set attributes on the plug-ins that do not exist yet to enforce this explicit code style.

BigRoy:
> Reason is that it allows you the decision between old behaviour and new one, even if the old one was somewhat limited. It should stay there imho for a while as deprecated feature problably (and I should state it somewhere that it is so).

Somehow I feel this should and could be simplified but I'm having a hard time thinking of a clean solution now. Adding a setting that we already know would be considered "deprecated" behavior just doesn't feel right and we're only introducing more technical/code debt that way, no?

mkolar:
>Somehow I feel this should and could be simplified but I'm having a hard time thinking of a clean solution now

I agree and we should probably log an issue or discussion so it doesn't get covered by a dust of time, but for now, this works well enough for production use-cases we have.

openpype/hosts/maya/plugins/load/load_reference.py
```diff
@@ -125,6 +127,11 @@ def process_reference(self, context, name, namespace, options):
 
             return new_nodes
 
+    def load(self, context, name=None, namespace=None, options=None):
+        super(ReferenceLoader, self).load(context, name, namespace, options)
+        # clean containers if present to AVALON_CONTAINERS
+        self._organize_containers(self[:])
```
BigRoy, line 133: Another question - what's the reason we want this? Isn't it much nicer if they actually remain within the container of what loaded it? You'd get these containers in the containers

E.g. I think this is much better:
```
AVALON_CONTAINERS > layoutMain > modelMain
```
Than:
```
AVALON_CONTAINERS > layoutMain
AVALON_CONTAINERS > modelMain
```

openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py
BigRoy: @mkolar Any preference on this from your end? ☝️

openpype/hosts/maya/plugins/load/load_reference.py
BigRoy: > Without any organizing my loaded containers loads to the scene root and because they are not hidden in the Outliner, they start to clutter it

Ah, of course. With the change where only reference nodes are added into the avalon container then of course other scene nodes aren't added in there either. I guess that's a good thing.

Then one more question. What's the reason this organizing isn't added directly to the `openpype.hosts.maya.api.plugin.ReferenceLoader` class but only on the inherited one?

mkolar:
Task linked: [OP-2772 Avalon containers to be parented to instance by default](https://app.clickup.com/t/6658547/OP-2772)

mkolar (approved):

openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py
BigRoy: @mkolar I saw you approved this PR - could you check in regarding this question above?

openpype/hosts/maya/plugins/load/load_reference.py
BigRoy: > Then one more question. What's the reason this organizing isn't added directly to the `openpype.hosts.maya.api.plugin.ReferenceLoader` class but only on the inherited one? I'd recommend moving it there.

Note that this is also still an open question. It'd reduce the amount of lines of code.

antirotor: I've moved the code to the parent.

openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py
antirotor: and yes, after giving it more thought than it deserves :) I agree with your point.  But I believe with this we really need to implement some central way of managing families - we are currently unable to present (for example in Settings) all families we have, so in many places you just need to write it down manually and that is source of errors.

mkolar (approved):

mkolar:
Good with me. let's merge it, it's needed in production

openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py
```diff
@@ -57,10 +58,21 @@ def process(self, instance):
         else:
             members = instance[:]
 
+        loaded_containers = None
+        if set(self.add_for_families).intersection(
+                set(instance.data.get("families")),
+                set(instance.data.get("family").lower())):
```
BigRoy, line 64: 1) `set(instance.data.get("family").lower())` this is incorrect? Family is not a list and thus this would turn into e.g. `set("model".lower())` which turns into `set("model")` which becomes `{"m", "o", "d", "e", "l"}` :)
2) `.lower()` can be removed?

openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py
```diff
@@ -57,10 +58,21 @@ def process(self, instance):
...
+            loaded_containers = self._add_loaded_containers(members)
+
+        selection = members
+        if loaded_containers:
+            self.log.info(loaded_containers)
+            selection += loaded_containers
```
BigRoy, line 70: This if statement can be moved into the if statement above.

```python
# psuedocode
families = set([instance.data.get("family", "")] + instance.data.get("families", []))
if families.intersection(set(self.add_for_families)):
    member_containers = self._get_loaded_containers(members)  # note that I also refactored this method to a more logical name
    members += member_containers
                set(instance.data.get("families")),
                set(instance.data.get("family").lower()))
```

openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py
```diff
@@ -83,3 +95,33 @@ def process(self, instance):
...
+
+            id_attr = "{}.id".format(obj_set)
+            if cmds.getAttr(id_attr) != AVALON_CONTAINER_ID:
+                continue
+
+            set_content = set(cmds.sets(obj_set, query=True))
+            if set_content.intersection(refs_to_include):
+                loaded_containers.append(obj_set)
+
+        return loaded_containers
```
BigRoy, line 127: I think this logic would fail to detect e.g. loaded `gpuCache` containers, `yetiCache` containers, etc. and others that aren't referenced. So this logic would not be correct?

```python
members_with_refs= set(members) + set(refs_to_include)
...

if set_content.intersection(members_with_refs)
```

Likely you'd need something like that?

BigRoy (changes requested):

BigRoy:
> Good with me. let's merge it, it's needed in production

Ouch. Please see my notes above. :)

openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py
BigRoy: Also, are you sure you require the `family` to also be present in `families` which is what the current logic is doing because `a.intersection(b, c)` returns a set with items that is present in all 3 sets.

I'd rather detect whether it's at least present in either as in [my pseudocode mockup](https://github.com/pypeclub/OpenPype/pull/2837#discussion_r824517669)

PR #2839, opened 2022-03-04
General: Move change context functions

iLLiCiTiT:
## Brief description
Context change related functions moved from avalon into openpype.

## Description
These functions recalculate session changes and change of workdir used primarily in workfiles tool.

## Changes
- moved functions into `openpype/lib/avalon_context.py` - at this moment there is not a better place
- changed how fusion is getting wordir path
- use functions in workfiles tool

## Testing notes:
1. Workfiles tool should work as before
    - changes context on save
    - calculate proper workdir and uses right template
2. Fusion scripts handling workdir should work - not sure if it did work before?

### DIFF
--- a/openpype/hosts/fusion/scripts/fusion_switch_shot.py
+++ b/openpype/hosts/fusion/scripts/fusion_switch_shot.py
@@ -5,11 +5,12 @@ import logging
 
 # Pipeline imports
 import avalon.api
-from avalon import io, pipeline
+from avalon import io
 
 from openpype.lib import version_up
 from openpype.hosts.fusion import api
 from openpype.hosts.fusion.api import lib
+from openpype.lib.avalon_context import get_workdir_from_session
 
 log = logging.getLogger("Update Slap Comp")
 
@@ -46,12 +47,8 @@ def _format_version_folder(folder):
 
 def _get_work_folder(session):
     """Convenience function to get the work folder path of the current asset"""
-
     # Get new filename, create path based on asset and work template
-    template_work = self._project["config"]["template"]["work"]
-    work_path = pipeline._format_work_template(template_work, session)
-
-    return os.path.normpath(work_path)
+    return get_workdir_from_session(session)
 
 
 def _get_fusion_instance():
--- a/openpype/hosts/fusion/scripts/fusion_switch_shot.py
+++ b/openpype/hosts/fusion/scripts/fusion_switch_shot.py
@@ -45,12 +45,6 @@ def _format_version_folder(folder):
     return version_folder
 
 
-def _get_work_folder(session):
-    """Convenience function to get the work folder path of the current asset"""
-    # Get new filename, create path based on asset and work template
-    return get_workdir_from_session(session)
-
-
 def _get_fusion_instance():
     fusion = getattr(sys.modules["__main__"], "fusion", None)
     if fusion is None:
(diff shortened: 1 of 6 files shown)


mkolar:
Task linked: [OP-2844 Move change context functions](https://app.clickup.com/t/6658547/OP-2844)

openpype/hosts/fusion/scripts/fusion_switch_shot.py
```diff
@@ -46,12 +47,8 @@ def _format_version_folder(folder):
 
 def _get_work_folder(session):
     """Convenience function to get the work folder path of the current asset"""
-
     # Get new filename, create path based on asset and work template
-    template_work = self._project["config"]["template"]["work"]
-    work_path = pipeline._format_work_template(template_work, session)
-
-    return os.path.normpath(work_path)
+    return get_workdir_from_session(session)
```
BigRoy, line 51: Should we remove the `_get_work_folder` function and refactor other code to just use `get_workdir_from_session` directly?
Especially since it's now just a single function call. I'd argue the same for the removal of [`_get_context_directory` in `switch_ui.py`](https://github.com/pypeclub/OpenPype/pull/2839/files#diff-202acaf33ac1ec4e4421ad5a56783a07d1dc51ff2da3ee3d62aaf8cca8c399f1R161)

openpype/lib/avalon_context.py
```diff
@@ -644,6 +644,161 @@ def get_workdir(
     )
 
 
+def template_data_from_session(session):
+    """ Return dictionary with template from session keys.
+
+    Args:
+        session (dict, Optional): The Session to use. If not provided use the
+            currently active global Session.
```
BigRoy, line 652: The docstring is either incorrect or the argument should be `session=None`?

BigRoy:
Separate from this PR we should look into what's up with [this code](https://github.com/pypeclub/OpenPype/blob/59c5c464ccd3960dbf2e5260024d42c2c58436eb/openpype/tools/assetcreator/app.py#L177-L184).

That appears extremely connected to Maya only - and it seems to be part of a feature that I personally was unaware of. Is that used? Should that be removed/cleaned?

openpype/hosts/fusion/scripts/fusion_switch_shot.py
iLLiCiTiT: Done. I actually wasn't sure if the script is even used?

iLLiCiTiT:
> That appears extremely connected to Maya only - and it seems to be part of a feature that I personally was unaware of. Is that used? Should that be removed/cleaned?

It is extremely connected to maya and ftrack. I think it wasn't used for 3 years. Should be removed, but not in this PR.

openpype/hosts/fusion/scripts/fusion_switch_shot.py
BigRoy: Yes, it's old - probably unused in OpenPype - and it's been a while since we have used it too. We could consider it code debt for a future separate PR and discuss a more thorough cleanup.

mkolar:
>Should that be removed/cleaned?

The whole tools should be removed and eventually re-written, this was an urgent production requirement a few years back when OP was used only in a handful of places. I'd leave it for a separate PR though

mkolar (approved):

PR #2841, opened 2022-03-04
New: Validation exceptions

iLLiCiTiT:
## Brief description
Use custom validation exception in validation plugins.

## Description
In new publisher validators should raise `PublishValidationError` and to make live easier we prepared `PublishXmlValidationError` which will load xml file with information for the plugin so the plugin does not have to contain all descriptions of error. Is backwards compatible.

## Changes
- added `detail` to PublishValidationError` exception
- imlemented `PublishXmlValidationError` which looks for xml files next to plugin location based on same filename as plugin's filename
- changed exceptions in hosts:
    - aftereffects
    - harmony
    - houdini
    - standalone publisher
    - tvpaint
- removed few validation plugins

## Testing notes:
1. Trigger publishing in one of affected hosts and cause that valitors will crash
2. Current publisher should work as did
3. New publisher should show more description in UI


kalisp (approved):
Tested in PS, AE, Harmony, SP, seems fine.

PR #2848, opened 2022-03-07
General: Add vendors from avalon

iLLiCiTiT:
## Brief description
Add vendors from avalon into openpype dependencies.

## Description
Some modules already are part of openpype dependencies but the rest have to be added into our system.

## Changes
- replaced `filelink` with new function `create_hard_link`
- added `qargparse` into `openpype/vendor`
- added `qtawesome` into poetry
    - has dependency `qtpy`
- `jsonschema` in poetry was lowered from `3.2.0` to `2.6.0`
- added `functools32` into python 2 vendor files

## Testing notes
- run `create_env` script
- change avalon-core branch to https://github.com/pypeclub/avalon-core/pull/426 (just to be sure)
- OpenPype should load in all hosts
- Try open tools using `qargparse` (loaders)
- Try tools using `qtawesome` inside hosts
    - especially in hosts with older python and Qt
    - `flame` @jakubjezek001 please try ui tools there
    - `maya2018` @antirotor if possible?

Depends on https://github.com/pypeclub/avalon-core/pull/426

### DIFF
--- a/openpype/lib/path_tools.py
+++ b/openpype/lib/path_tools.py
@@ -4,6 +4,7 @@ import abc
 import json
 import logging
 import six
+import platform
 
 from openpype.settings import get_project_settings
 


jakubjezek001 (approved):
tested on macos Flame and it was ok

openpype/lib/vendor_bin_utils.py
```diff
@@ -86,6 +86,41 @@ def find_executable(executable):
...
+
+        res = CreateHardLink(dst_path, src_path, None)
+        if res == 0:
+            raise ctypes.WinError()
+    # Raises not implemented error if gets here
+    raise NotImplementedError(
+        "Implementation of hardlink for current environment is missing."
+    )
+
+
```
BigRoy, line 123: Why is this in `vendor_bin_utils`? That seems confusing?

openpype/lib/vendor_bin_utils.py
```diff
@@ -86,6 +86,41 @@ def find_executable(executable):
     return None
 
 
+def create_hard_link(src_path, dst_path):
```
BigRoy, line 89: Should this function be called `create_hardlink`? Since the name of the actual feature is "hardlink" without a space, right?

openpype/lib/vendor_bin_utils.py
iLLiCiTiT: Moved to `path_tools.py`.

PR #2851, opened 2022-03-08
WebPublisher: Fix wrong number of frames for video file

kalisp:
## Brief description
Wrong value 0-1 was hardcoded, now `ffprobe` is used to try get frame count from a file

## Description
In the case of any weird video format, fallback to values stored in DB (Ftrack) is used.

## Testing notes:
1. Publish any video file as a 'render' family via WebPublisher
2. Compare number of frames in the file and in published record (shoul be visible in Loader or when Loaded to Nuke)


mkolar:
Task linked: [OP-2860 Nuke loader - loading mp4 from web publisher incorrect duration](https://app.clickup.com/t/6658547/OP-2860)

PR #2838, opened 2022-03-03
AfterEffects: Implemented New Publisher

kalisp:
## Brief description
Implementation of New Publisher in AfterEffects

## Description
Implementation should allow use of old and New publisher at same time

PR contains https://github.com/pypeclub/OpenPype/pull/2841 , after that is merged, list of modified files should be much smaller.

Known issue:
- old publisher shows all 3 creators, will be fixed with https://github.com/pypeclub/OpenPype/pull/2854

## Testing notes:
- Enable New Publisher in Local Settings
- Test full create > publish process on empty workfile
- Test publishing in old workfile with created instances in NP and via old publish

PR #2856, opened 2022-03-09
AfterEffects: Allow configuration of default variant via Settings

kalisp:
## Brief description
Variant name for subset name creation was hardcoded. Now it is possible to configure it via Settings.

## Description
Implemented as a part of move to New Publisher.
Multiple variants possible.

## Additional info
Depends on https://github.com/pypeclub/OpenPype/pull/2838. Should be tested by merging to #2838 as that branch might contain new updates/fixes for general behavior in NP.

## Testing notes:
1. Configure required value in `Settings:project_settings/aftereffects/create/RenderCreator`
2. Open render creator in NP, check value in the middle bottom (just above gray colored final subset name)


BigRoy:
@kalisp any reason you are not making this PR to #2838 directly?

kalisp:
@BigRoy its separate issue in our tracker, just red tape reasons

64qam (approved):

64qam (approved):

mkolar:
> its separate issue in our tracker, just red tape reasons

Let's pull off that red tape. It is possible to manually link PR to clickup task for these situations, which should make it easier then.

Issue #2573, opened 2022-01-21
Assembly collection broken


mkolar:
Assembly family in maya is not functional at all on publishing end, because of the changes to how containers store data inside their maya sets. Previously we stored all the nodes from the reference, however now we only keep the reference node itself. The collector is trying to find transforms in the set instead of traversing one level deeper to reference content

BigRoy:
> Previously we stored all the nodes from the reference, however now we only keep the reference node itself. The collector is trying to find transforms in the set instead of traversing one level deeper to reference content

- The `get_container_members` in [this PR](https://github.com/pypeclub/OpenPype/pull/2508) would solve that. :)

LiborBatek:
Did a test with latest ```develop``` (3.15.3 Nigthly.2) and got this error msg...

![image](https://user-images.githubusercontent.com/112623825/226372996-3064586f-4ad2-4d0a-a802-af0cebad343e.png)

PR #2862, opened 2022-03-09
Bugfix/op 2472 maya assembly collection broken

Allan-I:
Assembly collection fix for issue #2573 

Passing matrix transforms to the instance.data by recognizing the children of the root transform as well and grabbing the transforms of the children to be passed.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/collect_assembly.py
+++ b/openpype/hosts/maya/plugins/publish/collect_assembly.py
@@ -48,6 +48,9 @@ class CollectAssembly(pyblish.api.InstancePlugin):
 
             # Retrieve the hierarchy
             parent = cmds.listRelatives(root, parent=True, fullPath=True)[0]
+
+            # Retrieves children of root
+            children = cmds.listRelatives(parent, children=True, fullPath=True)[0]
             hierarchy_nodes.append(parent)
 
             # Temporary warning for GPU cache which are not supported yet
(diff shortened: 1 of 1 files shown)


Allan-I:
My initial thinking is that we need to pass children of the root nodes to the hierarchy as well as the `matrix_data` through `get_container_members ` but I'm not entirely sure of the step afterwards, since `matrix_data` is getting its matrix through `get_matrix_data`, so storing the children would require putting them in a dictionary, maybe?

mkolar:
There are actually more problems with this it turns out. 

1. we've accidentally broken the collection also in an unrelated plugins (openpype\plugins\publish\collect_resources_path.py) by relying on `hierarchy` inside instance.data. Unfortunately that is a global plugin and in editorial it holds something else thatn in assembly. So firstly that key should be renamed in the assembly collector to something else. 

Nevertheless, due to other changes in how we construct namespaces of loaded subsets I have a feeling that we broken the assembly publisher in general, because even after I bypass the hierarchy problem. The alembic groups that are published don't match the .json so it cannot be loaded into a scene correctly anyways. That's because we strip namespaces during the extraction I suppose. 

I'm wondering considering this was discovered only recently, it mean this is not being used in production at all, and might be actually worth finally doing it properly with USD structure. that we were discussing internally for a while and @antirotor is working on a rudimentary resolver, that would solve a lot of these issues. @BigRoy are you guys using assemblies at all. The idea was good, but we found the implementation was never quite enough for production.

BigRoy:
> are you guys using assemblies at all.

No. Kill it with fire 🔥 if you can!

mkolar:
Agreed across the board. We'll soon open up wide conversation about USD implementation instead.

PR #2866, opened 2022-03-10
Nuke: gizmo precollect fix

jrsndl:
## Brief description
Precollect checks Nuke groups for having "write" in families. Publishing Group as OP Gizmo fails.

## Testing notes:
1. create group of nodes in Nuke
2. Create OP Gizmo from selected group
3. Publish


jakubjezek001 (approved):
works well

PR #2874, opened 2022-03-11
Fixes for attaching loaded containers (#2837)

antirotor:
## Fixes for #2837

This is adding additional fixes for #2837

### DIFF
--- a/openpype/hosts/maya/api/plugin.py
+++ b/openpype/hosts/maya/api/plugin.py
@@ -247,6 +247,11 @@ class ReferenceLoader(Loader):
 
             self.log.warning("Ignoring file read error:\n%s", exc)
 
+        shapes = cmds.ls(content, shapes=True, long=True)
+        new_nodes = (list(set(content) - set(shapes)))
+
+        self._organize_containers(new_nodes, container["objectName"])
+
         # Reapply alembic settings.
         if representation["name"] == "abc" and alembic_data:
             alembic_nodes = cmds.ls(
--- a/openpype/hosts/maya/api/plugin.py
+++ b/openpype/hosts/maya/api/plugin.py
@@ -178,7 +178,7 @@ class ReferenceLoader(Loader):
                 loader=self.__class__.__name__
             )
             loaded_containers.append(container)
-            self._organize_containers([ref_node], container)
+            self._organize_containers(nodes, container)
             c += 1
             namespace = None
 
@@ -318,6 +318,7 @@ class ReferenceLoader(Loader):
     @staticmethod
     def _organize_containers(nodes, container):
         # type: (list, str) -> None
+        """Put containers in loaded data to correct hierarchy."""
         for node in nodes:
             id_attr = "{}.id".format(node)
             if not cmds.attributeQuery("id", node=node, exists=True):
--- a/openpype/hosts/maya/api/plugin.py
+++ b/openpype/hosts/maya/api/plugin.py
@@ -247,10 +247,7 @@ class ReferenceLoader(Loader):
 
             self.log.warning("Ignoring file read error:\n%s", exc)
 
-        shapes = cmds.ls(content, shapes=True, long=True)
-        new_nodes = (list(set(content) - set(shapes)))
-
-        self._organize_containers(new_nodes, container["objectName"])
+        self._organize_containers(content, container["objectName"])
 
         # Reapply alembic settings.
         if representation["name"] == "abc" and alembic_data:
@@ -289,7 +286,6 @@ class ReferenceLoader(Loader):
                 to remove from scene.
 
         """
-
         from maya import cmds
 
         node = container["objectName"]
--- a/openpype/hosts/maya/plugins/load/load_reference.py
+++ b/openpype/hosts/maya/plugins/load/load_reference.py
@@ -121,17 +121,11 @@ class ReferenceLoader(openpype.hosts.maya.api.plugin.ReferenceLoader):
             if family == "rig":
                 self._post_process_rig(name, namespace, context, options)
             else:
-
                 if "translate" in options:
                     cmds.setAttr(group_name + ".t", *options["translate"])
 
+            print(new_nodes)
             return new_nodes
-
-    def load(self, context, name=None, namespace=None, options=None):
-        container = super(ReferenceLoader, self).load(
-            context, name, namespace, options)
-        # clean containers if present to AVALON_CONTAINERS
-        self._organize_containers(self[:], container[0])
 
     def switch(self, container, representation):
         self.update(container, representation)
--- a/openpype/hosts/maya/plugins/load/load_reference.py
+++ b/openpype/hosts/maya/plugins/load/load_reference.py
@@ -123,8 +123,6 @@ class ReferenceLoader(openpype.hosts.maya.api.plugin.ReferenceLoader):
             else:
                 if "translate" in options:
                     cmds.setAttr(group_name + ".t", *options["translate"])
-
-            print(new_nodes)
             return new_nodes
 
     def switch(self, container, representation):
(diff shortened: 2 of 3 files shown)


openpype/hosts/maya/plugins/load/load_reference.py
```diff
@@ -121,18 +121,12 @@ def process_reference(self, context, name, namespace, options):
             if family == "rig":
                 self._post_process_rig(name, namespace, context, options)
             else:
-
                 if "translate" in options:
                     cmds.setAttr(group_name + ".t", *options["translate"])
 
+            print(new_nodes)
```
BigRoy, line 127: Debug can be removed

openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py
```diff
@@ -58,16 +58,12 @@ def process(self, instance):
...
-                set(instance.data.get("family").lower())):
-            loaded_containers = self._add_loaded_containers(members)
-
         selection = members
-        if loaded_containers:
-            self.log.info(loaded_containers)
-            selection += loaded_containers
+        if set(self.add_for_families).intersection(
+                set(instance.data.get("families"))) or \
+                instance.data.get("family") in self.add_for_families:
```
BigRoy, line 64: Are we sure `family` is just a fallback and that `families` always includes what is in `family`?

With cleaning up "technical debt" before OP v4 I'd say removing `family` altogether and solely keeping `families` would be high on the list. :)

openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py
```diff
@@ -58,16 +58,12 @@ def process(self, instance):
...
+            selection += self._add_loaded_containers(members)
+            self.log.info(selection)
```
BigRoy, line 66: I'd recommend refraining from logging all nodes. It tends to be a very large list, thus still very hard to debug even if it's there and it slows down current interfaces as well with that many nodes. Unlike before where it was just the loaded containers that were logged.

openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py
```diff
@@ -58,16 +58,12 @@ def process(self, instance):
...
-            loaded_containers = self._add_loaded_containers(members)
-
         selection = members
-        if loaded_containers:
-            self.log.info(loaded_containers)
-            selection += loaded_containers
+        if set(self.add_for_families).intersection(
+                set(instance.data.get("families"))) or \
+                instance.data.get("family") in self.add_for_families:
+            selection += self._add_loaded_containers(members)
```
BigRoy, line 65: I'd recommend refactoring the method to `_get_loaded_containers` because it doesn't add anything.

openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py
mkolar: I agree. But it needs to be done a different PR, otherwise it's untrackable. That being said we'll then have to rely on the first item in `families` to determine the main family of the subset. Also not ideal

openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py
mkolar: Agreed

openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py
BigRoy: >  That being said we'll then have to rely on the first item in families to determine the main family of the subset. Also not ideal

Agreed - it's less explicit. I think it'd be good to identify exactly why we need to be able to identify a "primary family" as well.

Worst case we could have a `primaryFamily` data key that MUST also be present in `families`. (e.g. have a global validator so we make sure it's always the case if it exists) So that wherever check for family related things we can always rely on solely `families` and we just use `primaryFamily` to know what the primary is. We could also use the old `family` key for that - but might be harder fix the behavior for and it's less explicit.

BigRoy: > Are we sure family is just a fallback and that families always includes what is in family?

Nevermind - I read the logic wrong. It actually does check both.

openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py
```diff
@@ -58,16 +58,12 @@ def process(self, instance):
...
-                set(instance.data.get("families")),
-                set(instance.data.get("family").lower())):
-            loaded_containers = self._add_loaded_containers(members)
-
         selection = members
-        if loaded_containers:
-            self.log.info(loaded_containers)
-            selection += loaded_containers
+        if set(self.add_for_families).intersection(
+                set(instance.data.get("families"))) or \
```
BigRoy, line 63: Be aware that `families` _might_ not exist and you can't get make a set of `None`.

openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py
antirotor: I think we need to handle families in more robust way in v4 anyway and that should be another discussion.

openpype/hosts/maya/api/plugin.py
```diff
@@ -247,6 +247,11 @@ def update(self, container, representation):
 
             self.log.warning("Ignoring file read error:\n%s", exc)
 
+        shapes = cmds.ls(content, shapes=True, long=True)
+        new_nodes = (list(set(content) - set(shapes)))
```
BigRoy, line 251: I started to type the comment below - but I noticed you are trying to mimic the behavior that `process_reference` does for the `ReferenceLoader`. I believe there original shape filtering was once done to avoid shapes being added to the containers. Which has since been gone completely anyway - yay!

Anyway, just leaving this comment here as a side note - ignore it for now.

---

I'd recommend to not filter out the shapes. It's redundant.

I believe I can also technically parent (instance) a referenced shape into a transform that I want to currently export. `cmds.parent(referenced_shape, my_transform, add=True, shape=True)` with something like this. So technically I could be extracting a hierarchy that only has referenced shape. We'd still want to collect that reference too.

```my_transform > referenced_shape```

What's the reason you'd want to get these `new_nodes` like this?
You should be able to pass `content` directly to `self._organize_containers`. No?

File: openpype/hosts/maya/plugins/publish/extract_maya_scene_raw.py [hunk 7]
```diff
@@ -97,15 +92,15 @@ def process(self, instance):
...
+    def _get_loaded_containers(members):
         # type: (list) -> list
         refs_to_include = [
             cmds.referenceQuery(ref, referenceNode=True)
             for ref in members
             if cmds.referenceQuery(ref, isNodeReferenced=True)
         ]
 
-        refs_to_include = set(refs_to_include)
+        members_with_refs = set(refs_to_include).union(members)
```
BigRoy, line 103: Should we directly turn this into a set like this?
```python
        refs_to_include = {
            cmds.referenceQuery(ref, referenceNode=True)
            for ref in members
            if cmds.referenceQuery(ref, isNodeReferenced=True)
        }

        members_with_refs = refs_to_include.union(members)
```

Also I feel the variable name in the loop is confusing - it's not referring to a `reference` but to `node` that might be referenced or not.
```python
        refs_to_include = {
            cmds.referenceQuery(node, referenceNode=True)
            for nodein members
            if cmds.referenceQuery(node, isNodeReferenced=True)
        }

        members_with_refs = refs_to_include.union(members)
```
This makes more sense to me.

openpype/hosts/maya/api/plugin.py
antirotor: agree, I've just copy pasted the code without thinking much about it.

mkolar (approved):

mkolar:
This can make it's way to 3.9.1 as it seems to cover good amount of cases now. Happy to keep adjusting as we go along and do a bit of production battle testing with this.

PR #2875, opened 2022-03-11
AfterEffects: Fix rendering for single frame in DL

kalisp:
## Brief description
Single video file ('Asset_renderCompositionMain_v001.avi`) or single frame sequence (`Asset_renderCompositionMain_v001.0000.png`) caused failure when running publishing job in Deadline

## Testing notes:
0. Easiest to test it is to merge this to `feature/OP-2765_AE-to-new-publisher` branch
1. Publish single video (mov or avi - set in Composition Settings in AE) to farm
2. Publish single frame (png) to farm
3. Publish ordinary multi frame sequence to farm 
4. Nothing should fail in `Publish` step of rendering process in DL


openpype/lib/delivery.py
```diff
@@ -13,18 +13,30 @@ def collect_frames(files):
...
+    if len(files) == 1:
+        real_file_name = list(files)[0]
+
     sources_and_frames = {}
     if collections:
         for collection in collections:
             src_head = collection.head
             src_tail = collection.tail
 
+            if src_head.endswith("_v"):
```
BigRoy, line 32: This seems rather hardcoded. Anatomy does allow productions to override the filename structure so that the version string could be elsewhere in the filename?

It could be `v001_asset01_subset.exr`? - The number picked up would still not be the frame number?

openpype/lib/delivery.py
```diff
@@ -71,14 +83,15 @@ def path_from_representation(representation, anatomy):
...
 
     if os.path.exists(dst_path):
         return
     try:
-        create_hard_link(
+        filelink.create(
             src_path,
-            dst_path
+            dst_path,
+            filelink.HARDLINK
```
BigRoy, line 94: This is reverting to old code. This is wrong.

openpype/modules/deadline/plugins/publish/submit_aftereffects_deadline.py
```diff
@@ -102,24 +103,18 @@ def get_job_info(self):
 
     def get_plugin_info(self):
         deadline_plugin_info = DeadlinePluginInfo()
-        context = self._instance.context
-        script_path = context.data["currentFile"]
 
         render_path = self._instance.data["expectedFiles"][0]
 
-        if len(self._instance.data["expectedFiles"]) > 1:
```
BigRoy, line 110: Why stop relying on expected files. That seems wrong?

openpype/lib/delivery.py
BigRoy: All your created tests show the frame numbers are separated with a `.` like `file.0001.exr`. If we ALWAYS know that's the case, we can provide a more explicit pattern to `clique.assemble` which would be amazing.

But we'd need to ensure we never produce content like `file_0001.exr`.

openpype/modules/deadline/plugins/publish/submit_aftereffects_deadline.py
kalisp: Because it was replacing numbers with "#" only in cases where there were more expected files, not when there was only single one (but still with '0000' in its name). I think better option is to rely on presence of frame substring in a name.

openpype/lib/delivery.py
kalisp: You are right, I tried to create more generic, though safe, pattern. Updated tests.
I think that this function is used (or at least I am using it) in case that I don't know anything about format of files (in my case files are produced locally by AE render). 
I am still depending that version substring starts with 'v', following with any number of numeric characters.

openpype/lib/delivery.py
```diff
@@ -4,27 +4,41 @@
...
         Returns:
             (dict): {'/asset/subset_v001.0001.png': '0001', ....}
     """
     collections, remainder = clique.assemble(files, minimum_items=1)
 
+    real_file_name = None
     sources_and_frames = {}
+    if len(files) == 1:
+        real_file_name = list(files)[0]
+        sources_and_frames[real_file_name] = None
```
BigRoy, line 31: A single file could still be a frame range based render - we'd still like to collect the actual frame for that file. No?

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -601,6 +601,7 @@ def _get_representations(self, instance, exp_files):
                 "files": os.path.basename(remainder),
                 "stagingDir": os.path.dirname(remainder),
             }
+            representations.append(rep)
```
BigRoy, line 604: Interesting - was this just a bug that it was missing before? Seems like something that should've be found earlier :D

openpype/lib/delivery.py
```diff
@@ -4,27 +4,41 @@
...
+
     if collections:
         for collection in collections:
             src_head = collection.head
             src_tail = collection.tail
 
+            # version recognized as a collection
+            if re.match(".*([a-zA-Z0-9]%[0-9]+d).*", collection.format()):
```
BigRoy, line 39: Wouldn't you be better of taking just the info from the collection itself instead of trying to parse the formatted string (which I'd consider just to be consistent for user reading purposes instead of parsing purposes). I wouldn't expect the clique api to try and keep the formatting api super consistent across versions since it's just there to format a default readable format? (unless you explicitly pass it the string to format?)

Looking at your logic you're saying that if there's directly any alphabetic character or number before the parsed frame number) then you consider it to be a parsed version?

Somehow this logic doesn't feel intuitive to me. What exactly are we trying to solve?

64qam (approved):

openpype/lib/delivery.py
```diff
@@ -4,27 +4,41 @@
...
 def collect_frames(files):
     """
         Returns dict of source path and its frame, if from sequence
 
-        Uses clique as most precise solution
+        Uses clique as most precise solution, used when anatomy template that
+        created files is not known.
+
+        Depends that version substring starts with 'v' with any number of
+        numeric characters after.
```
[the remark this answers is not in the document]
kalisp: Removed

openpype/lib/delivery.py
```diff
@@ -14,31 +13,24 @@ def collect_frames(files):
         Uses clique as most precise solution, used when anatomy template that
         created files is not known.
 
-        Depends that version substring starts with 'v' with any number of
-        numeric characters after.
+        Assumption is that frames are separated by '.', negative frames are not
+        allowed.
```
BigRoy, line 17: I'm wondering. Should we also implement a test showing that negative frames do not work?
Like have a test that shows that e.g. `frame.-001.png` does not work? Similar to how we have a test that shows that `frame_001.png` does not parse to a frame number?

tests/README.md
```diff
@@ -21,3 +21,27 @@ Specific location could be provided to this command as an argument, either as ab
 (eg. `python ${OPENPYPE_ROOT}/start.py start.py runtests ../tests/integration`) will trigger only tests in `integration` folder.
 
 See `${OPENPYPE_ROOT}/cli.py:runtests` for other arguments.
+
+Run in IDE:
+-----------
+If you would prefer to run/debug single file dirrectly in IDE of your choice, you might encounter issues with imports.
```
BigRoy, line 27: Typo `dirrectly` -> `directly`

tests/README.md
```diff
@@ -21,3 +21,27 @@ Specific location could be provided to this command as an argument, either as ab
...
+It would manifest like `KeyError: 'OPENPYPE_DATABASE_NAME'`. That means you are importing module that depends on OP to be running, eg. all expected variables are set.
+
+In some cases your tests might be so localized, that you don't care about all env vars to be set properly.
+In that case you might add this dummy configuration BEFORE any imports in your test file
```
BigRoy, line 31: I like this being documented - but I'd love it even more if we'd be able to streamline running the tests even more so that we wouldn't even need to pass in the mockup data manually. Running tests should be made as easy as possible.

openpype/lib/delivery.py
kalisp: Cleaned up, used `clique.PATTERNS["frames"]` which handles it more nicely

openpype/lib/delivery.py
kalisp: It is already there `test_collect_frames_single_sequence_negative`

tests/README.md
kalisp: I hope this will be sorted out after this refactoring and clean up that is in progress. I find this PITA too.

kalisp:
After discussion on Discord implementation changed to follow `clique.PATTERNS["frames"]` , eg. frames are separated by '.' as it is most common use case.
If there will be any issue, we might need to create some configurable patterns in Settings.

PR #2879, opened 2022-03-11
Nuke: ExtractReviewSlate can handle more codes and profiles

iLLiCiTiT:
## Brief description
Plugin `ExtractReviewSlate` don't know how to reuse data from ffprobe to recreate almost same output as needed.

## Description
But we have a base of that logic in burnins so it is a good idea to move it to transcoding functions and share it across code. Also some functions from vendor lib can be moved to transcoding too.

## Changes
- function `ffprobe_streams` is split into 2 function `get_ffprobe_data` and `get_ffprobe_streams` and is moved to `transcoding`
    - changed related code to use this function
- moved function related to "recreate as close as possible" stream using ffmpeg from burnins into `transcoding`
    - changed related code to use it
- renamed function `get_fps` into `convert_ffprobe_fps_value` and moved to `transcoding`

## Testing notes:
1. Burnins should work as they did
2. ExtractReviewSlate should be able to create more outputs (dnxhd for example)

### DIFF
--- a/openpype/hosts/standalonepublisher/plugins/publish/extract_trim_video_audio.py
+++ b/openpype/hosts/standalonepublisher/plugins/publish/extract_trim_video_audio.py
@@ -2,6 +2,9 @@ import os
 import pyblish.api
 import openpype.api
 
+from openpype.lib import (
+    get_ffmpeg_tool_path,
+)
 from pprint import pformat
 
 
@@ -27,7 +30,7 @@ class ExtractTrimVideoAudio(openpype.api.Extractor):
             instance.data["representations"] = list()
 
         # get ffmpet path
-        ffmpeg_path = openpype.lib.get_ffmpeg_tool_path("ffmpeg")
+        ffmpeg_path = get_ffmpeg_tool_path("ffmpeg")
 
         # get staging dir
         staging_dir = self.staging_dir(instance)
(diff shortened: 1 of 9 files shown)


maxpareschi:
@iLLiCiTiT @jakubjezek001

Hi all, the dnhxr profiles introduces ambiguity in some cases when dealing with standard dnxhd.
Real problem is that ffprobe does get the bitrate, but it's inconsistent with the bitrate profiles that standard dnxhd REQUIRES to encode correctly.
These lines fix the problem albeit it's not the cleanest solution:

TO BE ADDED AFTER LINE 169 in extract_review_slate.py
```

            if "dnxhd" in codec_args:
                for arg in repre["outputDef"]["ffmpeg_args"].get("output", []):
                    if arg.startswith("-b:v") or arg.startswith("-vb"):
                        output_args.append(arg)
                        self.log.debug("DNxHD found, adding bitrate flag found in the ExtractReview Settings: `{}`".format(arg))

```

iLLiCiTiT:
> Since the review dnx gets encoded using user input parameters defined in the Extract Review plugin

It is not 100% guaranteed that it comes from ExtractReview.

> we could just copy the correct bitrate from there, no need to guess.

That's what is actually done there. The `source_ffmpeg_cmd = repre.get("ffmpeg_cmd")` is full command from extract review. We can't expect that someone filled output ffmpeg arguments as expected. You can have in settings one string there `"-pix_fmt yuv422p -c:v dnxhd -b:v 145M"` and will work.
If the guess means that bitrate from input is used that is "backup solution".

maxpareschi:
that's cool right now the branch is working as expected! much better than my hack.
Thanks

jakubjezek001 (approved):
Dnxhd is already working, but I found there was a wrong order for the output resolution definition in case reformed video material is coming in from upstream nuke bake processes. I had committed it here as it is related.

PR #2886, opened 2022-03-14
General: Move loader logic from avalon to openpype

iLLiCiTiT:
## Brief description
Move load workflow and related functions from avalon-core into OpenPype.

## Description
Moved bases of Loader plugin and load functions and also change names of functions and classes to identify their job. Load logic also consist of few database query functions that are loading related and at this moment don't have a place where they could be moved so they're part of load workflow. Added functions to register/deregister plugins/paths and to discover them for loading.

### Renamed variables
- `Loader` class renamed to `LoaderPlugin`
- `SubsetLoader` class renamed to `SubsetLoaderPlugin`
- `load` function renamed to `load_representation`
- `remove` function renamed to `remove_container`
- `update` function renamed to `update_container`
- `switch` function renamed to `switch_container`

## Note
This is not final structure of load workflow. The final structure rely on other structures and functions that are not yet moved and at the same time they can't be moved without this PR.

## Testing notes:
1. Run host
2. Show loader tool
3. Check if loader are available
4. Try to trigger load (at least one Loader)

### Tested hosts
- [x] Nuke
- [x] Flame
- [x] Maya
- [x] Houdini
- [x] Blender
- [x] Photoshop
- [x] AfterEffects
- [x] TVPaint
- [x] Harmony
- [x] Standalone Publisher

Related to PR https://github.com/pypeclub/avalon-core/pull/431


BigRoy:
> - load function renamed to load_representation
> - remove function renamed to remove_container
> - update function renamed to update_container
> - switch function renamed to switch_container

I'm not a fan of the new names. I understand that with full OpenPype potentially exposing a much larger API that their might be the urge to refactor the straight `load`, `remove`, `update` and `switch` but the new functions names to me feel less explicit because it doesn't sound like they belong together. All these functions are related to loading containers and managing their loaded state and having `load_representation` be the counterpart to `remove_container` just feels off to me. Can we simplify or streamline this more?

I could get used to the new names of course - but at first sight this seems like something that could also confuse newcomers.

BigRoy:
> Loader class renamed to LoaderPlugin
> SubsetLoader class renamed to SubsetLoaderPlugin

Pretty much same question as my previous comment - is this absolutely necessary? Sometimes less is more. Anyway, it's fine with me - it just strikes me as odd that it wasn't done with e.g. `LegacyCreator`? Similarly we don't do it on e.g. all the Loaders that are implemented for the hosts - they are called `PointcacheLoader`, `CameraLoader`, etc.

I feel like less is more and actually is more explicit too in this case.

openpype/pipeline/load/plugins.py
```diff
@@ -0,0 +1,130 @@
...
+def deregister_loader_plugins_path(path):
+    import avalon.api
+
+    avalon.api.deregister_plugin_path(LoaderPlugin, path)
+
+
+def register_loader_plugins_path(path):
+    import avalon.api
+
+    return avalon.api.register_plugin_path(LoaderPlugin, path)
```
BigRoy, line 125: Is the goal to strip of `avalon` completely? If so, this doesn't feel like the best strategy. Why still register with Avalon in the e/d

Also, having all these `register_loader_plugin`, etc. separate functions that all do basically the same thing - register a plug-in into a single simple plug-in system. Why do we need all these separate functions everywhere? I know it's more explicit but it introduces so much more code overhead to maintain the logic. OpenPype already is a massive beast - and we're now merging in Avalon **and** at the same time growing Avalon's code into something much bigger too. That feels off.

I feel like that at worst we should merge Avalon code and slim down where we can, no? Unless we're now adding actual functionality that we absolutely require? Or are we just adding convenience function but are we keeping the same minimal system in-place and are we refactoring these utility functions to do `openpype.api.register_plugin_path(LoaderPlugin, path)` in the end?

mkolar:
> All these functions are related to loading containers and managing their loaded state and having `load_representation` be the counterpart to `remove_container` just feels off to me. Can we simplify or streamline this more?

I agree with this one. the function should be `load_container` to keep consistency. 

> I could get used to the new names of course - but at first sight, this seems like something that could also confuse newcomers.

I guess that will be a matter of perspective. Exactly due to the fact that OpenPype is a large codebase, with overlaps into many other APIs and hosts, we'd like to be super explicit on any core functions to be able to keep track of everything. Searching for `LoaderPlugin` in the codebase gives you relevant results. Searching for `Loader` give you a whole lot of everything, especially because a GUI also called Loader :).

But in general all of the work in the first phase is to remove the avalon repository dependency. Once that is done, the real cleanup and refactoring can start, but at the moment, ambiguity in names, needs to be lowered to make that process simpler to keep track of.

openpype/pipeline/load/plugins.py
mkolar: > merging in Avalon and at the same time growing 

Purely merging now. All of these functions exist in avalon repo, hence we are keeping them in the first pass. Once that's done we can start trimming and separating into smaller repos is the ultimate step to allow for leaner codebase and better testability. 

Removing the code from avalon and changing it at the same time feels like a road to hell for us though, so we're just being careful here.

openpype/hosts/houdini/api/pipeline.py
```diff
@@ -53,7 +53,7 @@ def install():
     pyblish.api.register_host("hpython")
 
     pyblish.api.register_plugin_path(PUBLISH_PATH)
-    register_loader_plugin_path(LOAD_PATH)
+    register_loader_plugins_path(LOAD_PATH)
     avalon.api.register_plugin_path(LegacyCreator, CREATE_PATH)
```
BigRoy, line 57: Are we sure this was originally a typo? :) The surrounding functions line 55 and 57 also don't have the plural on `plugin`?

iLLiCiTiT: > The surrounding functions line 55 and 57 also don't have the plural on plugin?

I would say that it is because it is registering a path for a plugin class, but `register_loader_plugins_path` is registering path for loader plugins. Both variants are acceptable for me.

openpype/pipeline/load/plugins.py
iLLiCiTiT: > Why do we need all these separate functions everywhere?

Well with this functions you can change it to anything at one place.

> I feel like that at worst we should merge Avalon code and slim down where we can, no? 

Not in single PR. It is a massive amount of changes at one time. It would not be doable and reviewable. It's affecting almost each code in OpenPype and it would also require a refactor of OpenPype because everything is affecting everything.

openpype/hosts/houdini/api/pipeline.py
BigRoy: > I would say that it is because it is registering a path for a plugin class, but register_loader_plugins_path is registering path for loader plugins. Both variants are acceptable for me.

Aren't they both just registering a path from where plug-ins get loaded? I'd stick to the naming conventions that already exist if the function intend to do the same? In this case we might be somewhat biased because the other 'naming conventions' come from Pyblish and Avalon API - but still I wouldn't see why we wouldn't match for consistency.

jakubjezek001 (approved):
Flame works

mkolar (approved):

mkolar:
@iLLiCiTiT ready to merge

PR #2889, opened 2022-03-14
General: Reduce style usage to OpenPype repository

iLLiCiTiT:
## Brief description
There are still few places where `avalon.style` is used.

## Description
Colors from `avalon.style.colors` are used at few places primarily in tools and some plugins. Maya look assigner is using avalon styles fully. Colors must be defined in openpype style and plugins should not use the colors from there (at least not now).

## Changes
- plugins using `avalon.style.*` have now color hardcoded
- UI in action "Delete old version" is using openpype styles
- openpype style has 4 new functions
    - `get_default_tools_icon_color` - tool icons colors (refresh, set current asset, etc.)
    - `get_default_entity_icon_color` - color for entity icons (asset, task, container, etc.)
    - `get_disabled_entity_icon_color` - color of disabled entity (used for no task item, or missing workfile)
    - `get_deprecated_entity_font_color` - color of deprecated entity (deprecated asset)
- added function to get project icon and it's color
- maya look assigner is using openpype style and was a little bit modified to propagate styles
- changed default color of project, assets, tasks, subsets and containers (scene inventory)

## Testing notes:
1. Colors in UIs should be a little bit modified
    - affected project, assets, tasks, subsets, containers and plugins (loader icons)

Related to PR https://github.com/pypeclub/avalon-core/pull/430


mkolar (approved):

openpype/style/__init__.py
```diff
@@ -7,11 +7,28 @@
...
-_STYLESHEET_CACHE = None
-_FONT_IDS = None
 
-current_dir = os.path.dirname(os.path.abspath(__file__))
+# Default colors
+# - default color used in tool icons
+_TOOLS_ICON_COLOR = "#ffffff"
+# - entities icon color - use 'get_default_asset_icon_color'
+_DEFAULT_ENTITY_ICON_COLOR = "#fb9c15"
+# - disabled entitie
```
BigRoy, line 18: Typo `entitie` -> `entity`

openpype/style/__init__.py
```diff
@@ -7,11 +7,28 @@
...
+_DISABLED_ENTITY_ICON_ICON_COLOR = "#808080"
```
BigRoy, line 19: What's the reason we still need these globals if they are read from `openpype/style/data.json` by default? Shouldn't we choose one of the two instead of both?

It's added [here](https://github.com/pypeclub/OpenPype/pull/2889/files#diff-5b574387dd928faaa5b0ac039d1849c29f5cb781162b8b3a48024f83252bbb83R61-R63).

Would `data.json` ever not exist? And why wouldn't it?

openpype/tools/mayalookassigner/app.py
```diff
@@ -147,11 +155,14 @@ def remove_connection(self):
 
     def showEvent(self, event):
         self.setup_connections()
-        super(App, self).showEvent(event)
+        super(MayaLookAssignerWindow, self).showEvent(event)
+        if self._first_show:
+            self._first_show = False
+            self.setStyleSheet(style.load_stylesheet())
```
BigRoy, line 161: Why do we need this additional set stylesheet call? Seems something is off? It's already being set [here](https://github.com/pypeclub/OpenPype/pull/2889/files#diff-24a0bf91d45dc87a044e19664ed3f6271bf0544dd6b300e650095a8d20bbe9e0R281) no?

openpype/tools/utils/lib.py
```diff
@@ -126,18 +122,42 @@ def get_qta_icon_by_name_and_color(icon_name, icon_color):
...
+        if icon:
+            return icon
+
+    return get_qta_icon_by_name_and_color(
+        "fa.map", get_default_entity_icon_color()
+    )
+
+
+def get_asset_icon_name(asset_doc, has_children=True):
+    if asset_doc:
```
BigRoy, line 138: Why does this need an `if asset_doc`? To me it'd feel very weird for the function to succeed if I don't pass it a valid document? Shouldn't that just error?

openpype/tools/utils/lib.py
```diff
@@ -149,7 +169,9 @@ def get_task_icon():
 
     Icon should be defined by task type which is stored on project.
     """
-    return get_qta_icon_by_name_and_color("fa.male", style.colors.default)
+    return get_qta_icon_by_name_and_color(
+        "fa.male", get_default_entity_icon_color()
+    )
```
BigRoy, line 174: This is hardcoded to a single icon, unlike what the docstring describes. Avalon allowed to have a different icon per task name (which was basically it's type). Did we break that logic in OpenPype somewhere? Shouldn't this return an icon fitting with that particular task type?

openpype/tools/mayalookassigner/app.py
iLLiCiTiT: The `show` function is not used in Maya. Removed the setStyleSheet there.

PR #2892, opened 2022-03-15
Flame: support for comment with xml attribute overrides

jakubjezek001:
## Brief description
Supporting timeline segment comments with xml overrides.

## Description
Client was requesting multiple resolution with custom values to be used during exporting in publish. This could only by supported by some sort of input values and segment's comment is the best way to do it as it is supporting multiple selected segments' input. The supported syntax is described lower.

## Additional info
We are supporting following attributes: *width*, *height*, *pixelAspect*, *resizeType*, *resizeFilter* and additionally  also *resolution* (with multiple input string >> 1920x1080x1.5). User can add to the segment comment following text: `resolution: 1920x1080x1.5; resizeType: fill` and it will be resolved to: 
- width: 1920
- height: 1080
- pixelAspect: 1.50
- resizeType: fill

This will be dynamically added to xml preset override before the the sequence will be processed. It will be also set on *resolutionWidth*, *resolutionHeight*, *pixelRatio* on a shot/asset data.

# dependencies
this PR is based on https://github.com/pypeclub/OpenPype/pull/2869


openpype/hosts/flame/plugins/publish/collect_timeline_instances.py
```diff
@@ -126,6 +148,84 @@ def process(self, context):
...
+                "width": width,
+                "height": height,
+                "pixelRatio": pixel_ratio
+            })
+
+    def _split_comments(self, comment_string):
+        # first split comment by comma
+        split_comments = []
+        if "," in comment_string:
+            split_comments.extend(iter(comment_string.split(",")))
```
iLLiCiTiT, line 221: Why are you adding the `iter` everywhere? `split_comments.extend(comment_string.split(","))` does the job.

openpype/hosts/flame/plugins/publish/collect_timeline_instances.py
```diff
@@ -126,6 +148,94 @@ def process(self, context):
...
+        key, value = split.split(":")
+
+        for a_name, a_type in self.xml_preset_attrs_from_comments.items():
+            # exclude all not related attributes
+            if a_name.lower() not in key.lower():
+                continue
+
+            # get pattern defined by type
+            pattern = TXT_PATERN
+            if "number" in a_type or "float" in a_type:
```
iLLiCiTiT, line 191: Don't use for string comparison `"number" in "not-number"` if you're expecting specific value.
```suggestion
            if a_type in ("number" , "float"):
```

iLLiCiTiT (approved):

PR #2896, opened 2022-03-16
Documentation: New publisher develop docs

iLLiCiTiT:
## Brief description
Documentation about development for new publisher.

## Changes
- few minor changes in creator plugins
    - update list contains namedtuple `UpdateData`
    - `identifier` attribute is not abstract but returns `family` by default
    - `collect_instances` and `update_instances` have docstrings
    - some getter methods in creator plugins have default return value an attribute
        - `get_instance_attr_defs` returns `instance_attr_defs`
        - `get_pre_create_attr_defs` returns `pre_create_attr_defs`
        - `get_default_variant` returns `default_variant`
- added new category `"Host integrations"` with subcategory `"Publishing"` into develop documentation
- tried to describ how creation and publishing works together in new publisher

## Testing notes:
1. Run documentation website
2. Read and criticize


kalisp:
I am not a fan incorporating full examples of classes (like `CreateRender` and auto), I would prefer just links to github, as sooner than later implementation and documentation will split and become inconsistent. 
(I like smaller examples though.)

mkolar:
>I would prefer just links to github, as sooner than later implementation and documentation will split and become inconsistent.

That's actually a good point. We could just make a repo with public examples for all documentation purposes

kalisp:
And I would add some pictures :).

kalisp (approved):
I think its decent enough.

PR #2898, opened 2022-03-16
Refactor Integrate Asset

BigRoy:
## Brief description

Draft attempt at refactoring/rewriting some logic of the Integrator. My first step was to try and figure out the pieces that are currently putting it together to really figure out all logic that's there and clean up what really was redundant.

Note: The current state it not tested and 100% sure won't work as is. It's a WIP to initiate first discussion - and I'd be happy to revamp a lot more. But I hope to find some time tomorrow to improve readability of the code further so it's easier to understand what we are actually altering. A lot of it might also be "commenting" specific areas of the code.

### Related topics

- [Define Representation as data class](https://github.com/pypeclub/OpenPype/pull/1572)
- [Some questions on Discord from my end as I started work on this](https://discord.com/channels/517362899170230292/517382145552154634/953635260019654667)


BigRoy:
### Notable changes

Wanted to mark these changes specifically since they were obvious changes that I felt could break things. So naming them here to keep an eye out. But the refactor is so massive that it's likely a ton more.

---

Removed Slate exception:

```
            # exception for slate workflow
            if index_frame_start and "slate" in instance.data["families"]:
                index_frame_start -= 1
```

This "fix" should be moved to outside of the integrator itself.

---

Removed this logic for 'single file' processing for representations.
It will now always use `name` of the representation
```
            template_data["representation"] = repre['ext']
```

I believe this behavior actually was a bug because other areas of the code believed it was actually using `name` instead?

openpype/plugins/publish/integrate_new.py
```diff
@@ -705,40 +676,58 @@ def copy_file(self, src, dst):
...
+
+        for _family in (instance.data.get("families") or []):
+            if _family not in families:
+                families.append(_family)
+
+        return families
+
+    def register_subset(self, instance):
+        # todo: rely less on self.prepare_anatomy to create this value
+        asset = instance.data.get("assetEntity") # <- from prepare_anatomy :(
```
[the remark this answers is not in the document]
BigRoy: I'm actually quite amazed that after all of this refactoring on this very early draft this is really the only nitpick the hound has this time. 🥇

openpype/plugins/publish/integrate_new.py
```diff
@@ -304,441 +565,245 @@ def register(self, instance):
...
+            # frame indices of the published collection. It will start from
+            # that `frameStart` index instead. Thus if that frame start
+            # differs from the collection we want to shift the destination
+            # frame indices from the source collection.
+            destination_indexes = list(src_collection.indexes)
+            destination_padding = len(get_first_frame_padded(src_collection))
+            if repre.get("frameStart") is not None:
+                index_frame_start = int(repre.get("frameStart"))
 
                 # TODO use frame padding from right template group
```
BigRoy, line 654: @mkolar do you happen to know what the "right template group" would be? :) This is an old to do comment and I'm wondering if I could easily resolve this as I go.

BigRoy: Hmm - I can see the entries there - and I also see the Templates in Anatomy however it's not clear how using a custom template name I'd overwrite the frame padding value instead?

openpype/plugins/publish/integrate_new.py
```diff
@@ -304,441 +565,245 @@ def register(self, instance):
...
+        data = repre.get("data") or {}
+        data.update({'path': repre["published_path"], 'template': template})
+        representation = {
+            "_id": repre_id,
+            "schema": "openpype:representation-2.0",
+            "type": "representation",
+            "parent": version["_id"],
+            "name": repre['name'],
+            "data": data,
+            "dependencies": instance.data.get("dependencies", "").split(),
```
BigRoy, line 768: Can someone confirm for me that this is unused/deprecated?

I believe it originated from the RigLoader in Maya which stored ['dependencies'](https://github.com/pypeclub/OpenPype/blob/0f2ca2e0432655562739813fd4114f06870e7184/openpype/hosts/maya/plugins/load/load_reference.py#L159) on load. Oddly enough it's storing just the representation id of the actual rig that was loaded (which seems odd) - the value also didn't get updated on updates of the rigs.

It seems extra weird that the value on the instance wasn't just a list to begin with but expecting a concatenated string?

Anyway, with 'input dependencies' the source rig representation for a published alembic would already be retrievable and thus also more reliable than this old unused logic. @mkolar  Correct?

BigRoy:
The current refactored state as of https://github.com/pypeclub/OpenPype/pull/2898/commits/56bcd8cec35f201ead80a18c09f6c070b76209c1 now can actually integrate e.g. a simple Maya Look (with a texture) and a Maya Pointcache. So it's somewhat functional as before.

Likely Hero Versions are broken. Likely Reviews as well. Likely site sync functionality isn't completely functional either.

Also haven't tested what happens if a publish actually fails and whether the file transaction rolls back correctly. But separate "tests" can be implemented to define our expected behavior explicitly and adapt the FileTransaction class to behave as we'd want.

Nonetheless would love input at this stage @mkolar @iLLiCiTiT @m-u-r-p-h-y as to what I should continue to do from here.
I've mostly focused on breaking the code down into more understandable 'parts'.

I've noticed however that if representations fails that the Subset and Version are registered nonetheless. Would be extra nice if we could somehow even merge those transactions together so that we can write all of those even closer together in the code.

openpype/plugins/publish/integrate_new.py
```diff
@@ -108,46 +273,44 @@ class IntegrateAssetNew(pyblish.api.InstancePlugin):
     exclude_families = ["clip"]
     db_representation_context_keys = [
         "project", "asset", "task", "subset", "version", "representation",
-        "family", "hierarchy", "task", "username"
+        "family", "hierarchy", "task", "username", "frame", "udim"
```
iLLiCiTiT, line 276: These keys should be always added when are available in formatting data but `frame` and `udim` should be added only if filled template used them.
```suggestion
        "family", "hierarchy", "task", "username"
```

openpype/plugins/publish/integrate_new.py
```diff
@@ -31,6 +27,175 @@
...
+    def _create_folder_for_file(self, path):
+        dirname = os.path.dirname(path)
+        try:
+            os.makedirs(dirname)
+        except OSError as e:
+            if e.errno == errno.EEXIST:
+                pass
+            else:
+                self.log.critical("An unexpected error occurred.")
+                six.reraise(*sys.exc_info())
```
iLLiCiTiT, line 196: I would simplify this, all the try->except does is reraise exception.
```suggestion
        if not os.path.exists(dirname):
            os.makedirs(dirname)
```

BigRoy: I believe the original behavior came from the fact that _technically_ the folder could have been created in-between the call to `os.makedirs` and `os.path.exists` and thus the `try..except` is the only way to correctly capture the folder didn't exist at the time of the call to `os.makedirs`. I know it's an extreme edge case, but it's considered safer as it currently is.

If we were python 3.2+ (which got further fixes in 3.4.1+) everywhere we could rely on `os.makedirs(dirname, exist_ok=True)` but unfortunately we aren't.

openpype/plugins/publish/integrate_new.py
BigRoy: Isn't that exactly what the code does? It stores only the values that are used in the template data around lines 284-290. No?

iLLiCiTiT: So they don't have to be defined here which may cause accidental adding of them into context where it has nothing to do?

BigRoy: > So they don't have to be defined here which may cause accidental adding of them into context where it has nothing to do?

They aren't added into the context if they aren't used in the template formatting. That's how I'm reading the code before + after. But let me look at this again fresh tomorrow - I might be missing something obvious.

BigRoy:
With the current state being "barebones functional and highly untested" with "more than likely broken in major areas" I do feel it's a good moment to digest what the integrator currently does (and why - if I can) and note some of the changes made (and why).

### What did the Integrator do that I have removed or have moved into its own

- Separated registering "subset", "version", "representation" logic more into isolation where possible.
- It previously prepared anatomy data which I mostly removed with [this commit](https://github.com/pypeclub/OpenPype/pull/2898/commits/177e244bd80bf0b1d472948cba45f40dfecd672e) due to it being collected by [CollectAnatomyContextData](https://github.com/pypeclub/OpenPype/blob/4c0309263c8ff300d337bbecc528e531e6b7a9b9/openpype/plugins/publish/collect_anatomy_context_data.py) and [CollectAnatomyInstanceData](https://github.com/pypeclub/OpenPype/blob/4c0309263c8ff300d337bbecc528e531e6b7a9b9/openpype/plugins/publish/collect_anatomy_instance_data.py). It should not be the responsibility of the integrator.
    - It still [prepares some anatomy data](https://github.com/pypeclub/OpenPype/blob/47259f8ef7b177892c76a8dbfde6d147cf664d39/openpype/plugins/publish/integrate_new.py#L311-L318) - it only still adds `intent` value - which [should be moved](https://github.com/pypeclub/OpenPype/blob/47259f8ef7b177892c76a8dbfde6d147cf664d39/openpype/plugins/publish/integrate_new.py#L293-L295).
- I've moved the "site sync" functionality in such a way that it's more encapsulated and we can discuss how to best clean that up. The logic now comes down to defining the `sites` once (because they got cached originally anyway)
    - The available "sites" logic is [here](https://github.com/pypeclub/OpenPype/blob/47259f8ef7b177892c76a8dbfde6d147cf664d39/openpype/plugins/publish/integrate_new.py#L988-L1086) and only used by a single call to `compute_resource_sync_sites` [here](https://github.com/pypeclub/OpenPype/blob/47259f8ef7b177892c76a8dbfde6d147cf664d39/openpype/plugins/publish/integrate_new.py#L422) - other than that it's quite "contained".
    - Then those sites are used to [prepare the file info](https://github.com/pypeclub/OpenPype/blob/47259f8ef7b177892c76a8dbfde6d147cf664d39/openpype/plugins/publish/integrate_new.py#L965-L986) which is the data Site Sync uses.

### What does the Integrator currently do?

It runs **per instance** and for each instance it does this:

1. Register the documents in the database
    1. [Register the subset](https://github.com/pypeclub/OpenPype/blob/47259f8ef7b177892c76a8dbfde6d147cf664d39/openpype/plugins/publish/integrate_new.py#L369)
    2. [Register the version](https://github.com/pypeclub/OpenPype/blob/47259f8ef7b177892c76a8dbfde6d147cf664d39/openpype/plugins/publish/integrate_new.py#L371)
    3. [Register the representation](https://github.com/pypeclub/OpenPype/blob/47259f8ef7b177892c76a8dbfde6d147cf664d39/openpype/plugins/publish/integrate_new.py#L379-L448) - still quite messy
2. Prepare data for template formatting to define publish file destinations
    1. _To be moved, see comment in code:_ [Quite some template data preparation for subset group](https://github.com/pypeclub/OpenPype/blob/47259f8ef7b177892c76a8dbfde6d147cf664d39/openpype/plugins/publish/integrate_new.py#L830-L857)
    2. [Handle the template formatting to published files](https://github.com/pypeclub/OpenPype/blob/47259f8ef7b177892c76a8dbfde6d147cf664d39/openpype/plugins/publish/integrate_new.py#L528-L741) - still quite complex
3. Integrate the files safely

#### Why still the complexity?

The complexity comes mostly from the fact that:
- It tries to "rollback" if it fails along the way but also needs to apply changes before it can do some other changes - making it non-trivial.
    - The file transferring happens in-between registering subset/versions and the representations - making it still a bit hard to follow when what is happening. Especially because some data for the representation currently requires the files to be in destination location (like the file info/hash for site sync).

- It can update data on an existing subset, like [set a new `subsetGroup` and add new "families" into the subset](https://github.com/pypeclub/OpenPype/blob/47259f8ef7b177892c76a8dbfde6d147cf664d39/openpype/plugins/publish/integrate_new.py#L768-L791)

- It can overwrite/update existing Versions. 
    - For which during _register version_ it'll [archive existing representations](https://github.com/pypeclub/OpenPype/blob/47259f8ef7b177892c76a8dbfde6d147cf664d39/openpype/plugins/publish/integrate_new.py#L505-L514)
    - More complexity due to the [instance also allowing to `append`](https://github.com/pypeclub/OpenPype/blob/47259f8ef7b177892c76a8dbfde6d147cf664d39/openpype/plugins/publish/integrate_new.py#L482-L484) where the [original representations are kept _unless_ when they are overwritten by new ones with the same name](https://github.com/pypeclub/OpenPype/blob/47259f8ef7b177892c76a8dbfde6d147cf664d39/openpype/plugins/publish/integrate_new.py#L501-L504).
    - Due to allowing to "overwrite" or "update" existing versions this means Integration also needs to safely handle potentially overwriting existing files with rollback. Now handled by [FileTranscation class](https://github.com/pypeclub/OpenPype/blob/47259f8ef7b177892c76a8dbfde6d147cf664d39/openpype/plugins/publish/integrate_new.py#L43-L198).

- Confusion between `instance.data["representation"]` and actual representation documents for the database as they aren't a one to one match yet the it's hard to make the distinction exactly in the code as noted in [this todo](https://github.com/pypeclub/OpenPype/blob/47259f8ef7b177892c76a8dbfde6d147cf664d39/openpype/plugins/publish/integrate_new.py#L708-L710)

- [Quite a lot of lines of code](https://github.com/pypeclub/OpenPype/blob/47259f8ef7b177892c76a8dbfde6d147cf664d39/openpype/plugins/publish/integrate_new.py#L988-L1086) related to computing the available "sites" (Site Sync functionality)

---

Is the Hound failing to run or how come I'm not making mistakes? I'm a bit worried the dog is sleeping and bites me later. 🔥

BigRoy:
Few questions!

1. Is it correct that the "archived_representation" type during the Integrator only got created 'during integration' and then directly deleted. Originally it [would archive representations](https://github.com/pypeclub/OpenPype/blob/develop/openpype/plugins/publish/integrate_new.py#L280-L300) and then before registering the new representations [delete all archived ones](https://github.com/pypeclub/OpenPype/blob/develop/openpype/plugins/publish/integrate_new.py#L633-L638) [(`existing_repres` defined here)](https://github.com/pypeclub/OpenPype/blob/develop/openpype/plugins/publish/integrate_new.py#L311-L314). So if instead I'd turn the registering into a single Bulk Write I wouldn't need the intermediate "archived_representation". Correct?

2. Can I just move the ["intent" anatomy data](https://github.com/pypeclub/OpenPype/blob/develop/openpype/plugins/publish/integrate_new.py#L318-L323) to [CollectAnatomyContextData](https://github.com/pypeclub/OpenPype/blob/644b2f639c8abfeff634adf990b3b9cbdfb1187b/openpype/plugins/publish/collect_anatomy_context_data.py)? Intent is defined at the beginning, no? Since [it's set directly in Pyblish Pype UI](https://github.com/pypeclub/OpenPype/blob/644b2f639c8abfeff634adf990b3b9cbdfb1187b/openpype/tools/pyblish_pype/window.py#L536-L539). Or would there be a reason not to?

3. Could we move [`_get_subset_group`](https://github.com/pypeclub/OpenPype/blob/develop/openpype/plugins/publish/integrate_new.py#L814-L871) into a global Collector of its own to just store `instance.data["subsetGroup"]`?

4. Is it correct that some publishes might not have a rootless path stored due to [`instance.data["source"]` being taken directly if it exists](https://github.com/pypeclub/OpenPype/blob/develop/openpype/plugins/publish/integrate_new.py#L909-L915)? Looking over the code base e.g. After Effects sets `instance.data["source"]` and [VRayScene collector in Maya](https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/maya/plugins/publish/collect_vrayscene.py#L125) and [Maya Collect Render](https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/maya/plugins/publish/collect_render.py#L336).
    - This is likely due to some publishers setting to `"source"` to [`"webpublisher"`](https://github.com/pypeclub/OpenPype/blob/50c0580feff4142db32f183452d17a801d2dcf13/openpype/hosts/webpublisher/plugins/publish/collect_published_files.py#L91) or ["standalone_publisher"](https://github.com/pypeclub/OpenPype/blob/6eaf7017eb66d85ca0089a84dbd63ebd874cf9f1/openpype/hosts/standalonepublisher/plugins/publish/collect_texture.py#L311)
    - However there are other areas where another value is expected in `source` like expecting a valid path [here in After Effects](https://github.com/pypeclub/OpenPype/blob/467208309b96095a63ddbf297a02ed145653d354/openpype/hosts/aftereffects/plugins/publish/validate_scene_settings.py#L136) and [here](https://github.com/pypeclub/OpenPype/blob/8432e94615a75216e7172cf46f1370fff806bf13/openpype/plugins/publish/integrate_resources_path.py#L28)

5. Glancing over the original integrator code **it looks to me as if site sync would not work for published Look textures** - as those files not part of a representation do not get their file info sorted for site sync. For representation files they were originally collected [here](https://github.com/pypeclub/OpenPype/blob/develop/openpype/plugins/publish/integrate_new.py#L617-L619) but I wasn't entirely sure how files directly in `instance.data["transfers"]` would be processed too. I'm pretty sure it's broken in the current refactored state but I'm also doubting whether it worked in the original. Anyone can confirm that published textures in a Maya look publish in Site Sync used to work as expected? Likely I'm misreading the original code somewhere [due to how it overrides the global `instance.data["transfers"]`](https://github.com/pypeclub/OpenPype/blob/644b2f639c8abfeff634adf990b3b9cbdfb1187b/openpype/plugins/publish/integrate_new.py#L358-L361). E.g. maybe originally both representations would have the "resources" files added uniquely as site sync entries? If it does work I'd love to know whether this is the case that the entries are duplicated.

6. Shouldn't version data `fps` rely on `instance.data["fps"]` over `context.data["fps"]`? [Originally context data would override instance data. Is that as expected](https://github.com/pypeclub/OpenPype/blob/644b2f639c8abfeff634adf990b3b9cbdfb1187b/openpype/plugins/publish/integrate_new.py#L925-L927)?

openpype/plugins/publish/integrate_new.py
```diff
@@ -299,469 +514,266 @@ def register(self, instance):
...
+            assert not _remainder, "This is a bug"
+            assert len(dst_collections) == 1, "This is a bug"
+            dst_collection = dst_collections[0]
+
+            # Update the destination indexes and padding
+            dst_collection.indexes.clear()
+            dst_collection.indexes.update(set(destination_indexes))
+            dst_collection.padding = destination_padding
+            assert len(src_collection.indexes) == \
+                   len(dst_collection.indexes), "This is a bug"
```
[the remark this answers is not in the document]
BigRoy: YAY! The hound works - and I was expecting is to bark about this line! 🐕‍🦺 ❤️

BigRoy:
Did some quick tests of the current code.

Tested Published + Loading:
- Maya Pointcache ✅ 
- Maya Model ✅ 
- Maya Look with a texture ✅ 
- Maya Pointcache with a file already existing in the destination location ✅
- Houdini Pointcache ✅ 
- Houdini VDB sequence ✅  <- this is the only "sequence" type I have tested.

I have not tested:
- Site Sync
- Hero Versions
- Reviews
- Other hosts
- Farm rendering
- etc.

antirotor:
So far so good. It did try publishing in Blender, Hero versions and farm rendering. It failed on the last one because of missing `published_path` - or I should say, In **IntegrateFtrackInstance** failed there, expecting it. Old iintegrator provided it. Now I am not sure what this variable does so I am just leaving it here as a note until I have time to dig in deeper.

openpype/plugins/publish/integrate_new.py
```diff
@@ -1,38 +1,88 @@
...
+    family = instance.data.get("family")
+    families = []
+    if family:
+        families.append(family)
+
+    for _family in (instance.data.get("families") or []):
+        if _family not in families:
+            families.append(_family)
+
+    return families
```
BigRoy, line 67: Anyone know of a more optimal way to write this code? :)

Basically it's just adding `[family] + families` and ensuring they are unique whilst preserving the order?
I was thinking of actually doing that:
```python
data = instance.data
families = [data.get("family")] + data.get("families", [])
families = [family for family in families if family]  # keep only actual values
return unique(families)
```
Where `unique` would be keep unique and preserve order e.g. like [this](https://stackoverflow.com/a/480227/1838864)
But I felt I was overcomplicating things. Nonetheless I feel the code can be simplified. Ideas?

openpype/plugins/publish/integrate_new.py
```diff
@@ -1,38 +1,88 @@
...
+
+def get_first_frame_padded(collection):
+    """Return first frame as padded number from `clique.Collection`"""
+    start_frame = next(iter(collection.indexes))
+    return get_frame_padded(start_frame, padding=collection.padding)
+
+
+def bulk_write(writes):
+    """Convenience function to bulk write into active project database"""
+    return io._database[io.Session["AVALON_PROJECT"]].bulk_write(writes)
```
BigRoy, line 83: This `bulk_write` should actually be implemented in `avalon.io` (or `openpype.io` when it's moved). Yes?

openpype/plugins/publish/integrate_new.py
```diff
@@ -110,801 +156,494 @@ class IntegrateAssetNew(pyblish.api.InstancePlugin):
...
+            # Update existing subset data with new data and set in database.
+            # We also change the found subset in-place so we don't need to
+            # re-query the subset afterwards
+            subset["data"].update(data)
+            bulk_writes.append(UpdateOne(
+                {"type": "subset", "_id": subset["_id"]},
+                {"$set": {
+                    "data": subset["data"]
+                }}
+            ))
```
BigRoy, line 390: Like before we currently update only the `data` of the existing Subset instead of replacing the document. Would it however be fine to just Replace the original document? Would there ever be any information on the existing Subset entity/document in the database that is outside of `data` which we want to preserve from the original subset?

If we could just replace the original document that would allow to simplify the code some more.

openpype/plugins/publish/integrate_new.py
```diff
@@ -913,21 +652,19 @@ def create_version_data(self, context, instance):
...
+            "families": get_instance_families(instance),
             "time": context.data["time"],
             "author": context.data["user"],
             "source": source,
             "comment": context.data.get("comment"),
             "machine": context.data.get("machine"),
-            "fps": context.data.get(
-                "fps", instance.data.get("fps")
-            )
+            "fps": instance.data.get("fps", context.data.get("fps"))
```
BigRoy, line 664: Having `instance.data["fps"]` override `context.data["fps"]` if it exists made more sense to me. Is that correct thinking?

openpype/plugins/publish/integrate_new.py
```diff
@@ -979,142 +741,104 @@ def get_rootless_path(self, anatomy, path):
...
-                                                  already_attached_sites,
-                                                  rec)
-
-            log.debug("final sites:: {}".format(rec["sites"]))
-
-        return rec
-
-    def _get_sites(self, sync_project_presets):
+class SiteSync(object):
+    """Logic for Site Sync Module functionality"""
```
BigRoy, line 785: The logic defined in this temporary SiteSync class should actually be coming from the Site Sync module. @kalisp has it on its todo list to look into our options regarding exposing that as API on the Site Sync module.

But for now at least this keeps it separated from the Integrator plug-in.

BigRoy (commented):
Added some comments with questions.

openpype/plugins/publish/integrate_new.py
```diff
@@ -110,801 +156,494 @@ class IntegrateAssetNew(pyblish.api.InstancePlugin):
...
-    def _get_subset_group(self, instance):
-        """Look into subset group profiles set by settings.
+                # TODO use frame padding from right template group
+                render_template = anatomy.templates["render"]
+                frame_start_padding = int(
+                    render_template.get(
+                        "frame_padding",
+                        render_template.get("padding")
+                    )
+                )
```
BigRoy, line 511: Similar to before the refactor the Render Template padding is only used when the `frameStart` value is defined in the representation data in the instance. Is this to be expected? Or would we expect the Render Template Frame Padding to **always** be used for sequences even when `frameStart` is not set?

openpype/plugins/publish/integrate_new.py
```diff
@@ -110,801 +156,494 @@ class IntegrateAssetNew(pyblish.api.InstancePlugin):
...
+        if not prepared_representations:
+            # Even though we check `instance.data["representations"]` earlier
+            # this could still happen if all representations were tagged with
+            # "delete" and thus are skipped for integration
+            raise RuntimeError("No representations prepared to publish.")
+
+        # Each instance can also have pre-defined transfers not explicitly
+        # part of a representation - like texture resources used by a
+        # .ma representation. Those destination paths are pre-defined, etc.
+        # todo: should we move or simplify this logic?
```
mkolar, line 264: We've had a discussion about this internally and agreed that we should actually be adding all resources into `['fils']` list to ensure that they are correctly handled by site sync, tracked properly in the db, plus is would actually remove the need for `transfers` and `hardlinks` lists.

It would, however, require that is `representation['files']` wouldn't be only a simple list, but a list of dictionaries, to allow specifying extra information for each file. Especially considering that currently we assume everything in `files` has the same stagingDir, which is not always true`. With dictionaries we could define `file_name`, `mode`(hardlink or copy) and similar when need be.

openpype/plugins/publish/integrate_new.py
```diff
@@ -110,801 +156,494 @@ class IntegrateAssetNew(pyblish.api.InstancePlugin):
...
+        for src, dst in instance.data.get("transfers", []):
+            file_transactions.add(src, dst, mode=FileTransaction.MODE_COPY)
+        for src, dst in instance.data.get("hardlinks", []):
+            file_transactions.add(src, dst, mode=FileTransaction.MODE_HARDLINK)
+
+        # Bulk write to the database
+        # todo: Can we move this even to after the file transfers?
```
mkolar, line 271: I believe we could

openpype/plugins/publish/integrate_new.py
```diff
@@ -110,801 +156,494 @@ class IntegrateAssetNew(pyblish.api.InstancePlugin):
...
+                        "parent": version["_id"],
+                        "name": {"$in": list(delete_names)}
+                    }
+                ))
+
+        # Write representations to the database
+        bulk_write(representation_writes)
+
+        # Backwards compatibility
+        # todo: can we avoid the need to store this?
```
mkolar, line 337: To be able to remove it, we'd need to handle deleting representations that were tagged for deletion (hence skipped here) directly in the integrator.

Another option might be re-using the `delete` flag logic from the top of this file in the following integrators that should only operate on representations that were actually already placed in the db

openpype/plugins/publish/integrate_new.py
mkolar: Hmm. I think we planned to allow defining custom padding in each template category, but eventually removed it to reduce complexity. Looks like it would be safe to remove this TODO

openpype/plugins/publish/integrate_new.py
```diff
@@ -110,801 +156,494 @@ class IntegrateAssetNew(pyblish.api.InstancePlugin):
...
+        # Use previous representation's id if there is a name match
+        existing = existing_repres_by_name.get(repre["name"].lower())
+        if existing:
+            repre_id = existing["_id"]
+        else:
+            repre_id = ObjectId()
+
+        # Backwards compatibility:
+        # Store first transferred destination as published path data
+        # todo: can we remove this?
```
mkolar, line 597: Following integrators need to know the final destination of each representation without having to re-calculate it. So unless we find a better way, I don't think we can remove it.

openpype/plugins/publish/integrate_new.py
```diff
@@ -110,801 +156,494 @@ class IntegrateAssetNew(pyblish.api.InstancePlugin):
...
+            "parent": version["_id"],
+            "name": repre['name'],
+            "data": data,
+            "dependencies": instance.data.get("dependencies", "").split(),
+
+            # Imprint shortcut to context for performance reasons.
+            "context": repre_context
+        }
+
+        # todo: simplify/streamline which additional data makes its way into
```
mkolar, line 619: This is directly relate to work on this https://github.com/pypeclub/OpenPype/pull/1572 which unfortunately got stalled, but we considered it a prerequisite before moving onto a major integration refactor. It is mean to solve this exact issue and make representation a lot more defined.

openpype/plugins/publish/integrate_new.py
```diff
@@ -110,801 +156,494 @@ class IntegrateAssetNew(pyblish.api.InstancePlugin):
...
+        if is_sequence_representation and repre.get("frameStart") is not None:
+            representation['context']['frame'] = template_data["frame"]
+
+        return {
+            "representation": representation,
+            "anatomy_data": template_data,
+            "transfers": transfers,
+            # todo: avoid the need for 'published_files' used by Integrate Hero
+            # backwards compatibility
+            "published_files": [transfer[1] for transfer in transfers]
```
mkolar, line 633: Maybe if we refactor `files` on representation to be a dictionary as mentioned above, we could just add each file final destination in there, per file, that would also create a direct one to one mapping between source and dest files for any further plugins processing

openpype/plugins/publish/integrate_new.py
mkolar: I agree

openpype/plugins/publish/integrate_new.py
mkolar: Agreed this keep it contained and we can move it in another PR

mkolar (changes requested):
I'd say that it's crucial to turn this into a new integrator plugin (rename the original to `integrate_old` and leave this one as `integrate_new` maybe?)

The reason is that we can then start transferring families one by one always testing it against the new one, while making sure we're not braking anything else. Last refactor was done that way and was quite a good way of keeping track. 

Other it's unmergable due to testing complexity of everything everywhere.

openpype/plugins/publish/integrate_new.py
BigRoy: > It is mean to solve this exact issue and make representation a lot more defined.

Super. I'll look into it!

openpype/plugins/publish/integrate_new.py
BigRoy: > Following integrators need to know the final destination of each representation without having to re-calculate it. So unless we find a better way, I don't think we can remove it.

We're already storing the 'transfers' or `files` data in the actual representation entities. We might be better of using either of those data then duplicating the data and also storing `published_files` separately?

openpype/plugins/publish/integrate_new.py
BigRoy: It would mean however that if two machines would publish at the same time there's a higher chance they publish into the same version without warning and both write a unique version id with the same version number. Especially if files copying takes long that chance increases. It's not entirely safe. Unless we'd maybe do another check JUST before writing to the database and after the file transfer. We could e.g. do that only if more time passed than "safety amount".

openpype/plugins/publish/integrate_new.py
BigRoy: Ah - the `repre` delete flag is not just a "delete publish instance" during this publishing. But delete the actual existing representation in the database? Doesn't sound like something the Integrator should do but would love to know the particular use cases!

openpype/plugins/publish/integrate_new.py
BigRoy: To be clear. 

1. Are we saying to add all files to the `instance.data["representations"]` or the actual representation entity in the database?
2. We wouldn't want database representation entity `.ma` to also have the Alembic file in `files` of representation `.abc`, correct? So we're just talking about duplicating the `files` entries for "resources" like textures in Maya Lookdev? So in essence:

```
# random example
representation .ma
    files = ["representationfile.ma", "texture1.exr", "texture2.exr"]

representation .abc
    files = ["representationfile.abc", "texture1.exr", "texture2.exr"]
```

openpype/plugins/publish/integrate_new.py
mkolar: Damn. you're totally right and that's actually why it's where it is, to reserve the version. Not sure there any benefit of moving it later really apart from visual clarity in the code (io operation clustered together). probably not worth the hassle

openpype/plugins/publish/integrate_new.py
mkolar: no no `delete` flag tells the publishing that this particular representation (not instance!) should not be integrated and the files will be deleted in the last publishing step. It doesn't delete anything in the db. The usecase is when studio is technically rendering image sequence from their DCC, but are only integrating Prores video for editorial that was generated from them automatically. 

Heavily used in 2d animation workflows with remote people to prevent integrating of massive file sequences, that give no benefit over 4444 prores files and only take more time to transfer and more data. 

The sequence representation is used during publishing to generate prores and then marked for deletion. Hence it is skipped in integrator and a cleanup pluing will remove it from the disk.

openpype/plugins/publish/collect_subset_group.py
```diff
@@ -0,0 +1,94 @@
...
+
+from openpype.lib.profiles_filtering import filter_profiles
+from openpype.lib import (
+    prepare_template_data,
+    StringTemplate,
+    TemplateUnsolved
+)
+
+
+class CollectSubsetGroup(pyblish.api.InstancePlugin):
```
iLLiCiTiT, line 20: You can't collect subset groups during collection because there are instances that yet don't have right family.

BigRoy: Could you identify for me which ones those are?  Because if that's the case - then whatever logic is involved in those areas should actually move to Collectors! :)

openpype/plugins/publish/integrate_new.py
```diff
@@ -110,801 +156,494 @@ class IntegrateAssetNew(pyblish.api.InstancePlugin):
     exclude_families = ["clip"]
     db_representation_context_keys = [
         "project", "asset", "task", "subset", "version", "representation",
-        "family", "hierarchy", "task", "username"
+        "family", "hierarchy", "task", "username", "frame"
```
iLLiCiTiT, line 159: The `frame` really should not be here.

BigRoy: Looking at the code changes the behavior should be the same as before. `frame` is only added if a valid value is present for frame and if used for the sequence - this is because a single file pops out the frame data anyway.

Or what am I missing? Or why must this not be here?

openpype/plugins/publish/integrate_new.py
```diff
@@ -110,801 +156,494 @@ class IntegrateAssetNew(pyblish.api.InstancePlugin):
...
+                # todo: add support for hardlink transfers
+                file_transactions.add(src, dst)
+
+            prepared_representations.append(prepared)
+
+        if not prepared_representations:
+            # Even though we check `instance.data["representations"]` earlier
+            # this could still happen if all representations were tagged with
+            # "delete" and thus are skipped for integration
+            raise RuntimeError("No representations prepared to publish.")
```
iLLiCiTiT, line 259: But that may be intended and is not reason to raise exception.

openpype/plugins/publish/integrate_new.py
iLLiCiTiT: You can use `anatomy.templates[template_name]["frame_padding"]`.

openpype/plugins/publish/integrate_new.py
iLLiCiTiT: We have to know which files were published for the representation not for instance.

openpype/plugins/publish/integrate_new.py
iLLiCiTiT: It should be added only if is used, not if is available.

openpype/plugins/publish/collect_subset_group.py
iLLiCiTiT: TVPaint rendering, Nuke rendering and some standalone publisher extractors.

openpype/plugins/publish/integrate_new.py
iLLiCiTiT: These keys should be always added to representation context no matter what was used.

openpype/plugins/publish/integrate_new.py
BigRoy: Thanks, good point!

I don't think we would want versions integrated in the database without **any** representations or files. Or what use case would having no representations to integrate have whatsoever? Would be good to have a use case so I could add it to the comment when that might be needed and could be allowed - that way it remains clear in the code why something was designed a certain way.

Or do you mean that instead or erroring it would suffice to "skip the integrator" completely for that instance and just log a warning that says something like "No representations to publish. Integration skipped..."?

openpype/plugins/publish/integrate_new.py
BigRoy: > It should be added only if is used, not if is available.

Could you explain by what it should be used? Is that used for destination template? Or for something else?

Currently they are only added when like this like before. This means these keys are added if in template data - not necessarily if used by the template. So you are correct - that sounds wrong if they only should be added if used. Thanks!
```
        for key in self.db_representation_context_keys:
            value = template_data.get(key)
            if not value:
                continue
            repre_context[key] = template_data[key]
```
_Which is the same logic as before [here](https://github.com/pypeclub/OpenPype/blob/develop/openpype/plugins/publish/integrate_new.py#L562-L566)._

The `frame` key would never be added if not used. It's always used in the Sequence/Collection, but never for a single file... and in that case it's popped. But I think you might be referring to that `frame` data would come [from the "used_values" here](https://github.com/pypeclub/OpenPype/blob/develop/openpype/plugins/publish/integrate_new.py#L441-L442). The interesting bit being, that if we only want to store if used - why do we need to be explicit for the others with the `db_representation_context_keys`? Do we always **require** `hierarchy`, `task`, etc even when NOT used?

openpype/plugins/publish/collect_subset_group.py
BigRoy: Thanks, that's good information. Here's what I found quickly:

- So in TV Paint we're changing the family "renderpass" and "renderlayer" to `render` in the Extractor.
    - Which also breaks the logic that `family` is the primary family and should also be in `families`.
    - For these cases - what's holding us back from adding `renderpass` or `renderlayer` to `instance.data["families"]` at the very beginning and set `instance.data["family"] = "render"` there too?

- [Nuke in Extract Render Local](https://github.com/pypeclub/OpenPype/blob/06a8f5014023ec1c792263fc66eb82c15159f908/openpype/hosts/nuke/plugins/publish/extract_render_local.py#L94-L105)
    - Same point, why not address this in a collector? It seems like we're breaking Pyblish workflow here.

- [Standalone Publisher changing family + families in Extractor](https://github.com/pypeclub/OpenPype/blob/8432e94615a75216e7172cf46f1370fff806bf13/openpype/hosts/standalonepublisher/plugins/publish/extract_bg_for_compositing.py#L174-L183)
- [Standalone Publisher creating new instances in Extractor](https://github.com/pypeclub/OpenPype/blob/8432e94615a75216e7172cf46f1370fff806bf13/openpype/hosts/standalonepublisher/plugins/publish/extract_images_from_psd.py#L90-L130)
    - In both these cases the PSD data should have been retrieved in a Collector itself, no?

These seem like "issues" on their own and go against the nature of Pyblish and how Validators should process the data, etc.

Worst case scenario we could work around this by moving `CollectSubsetGroup` to `IntegratorOrder - 0.499`. However to me this seems we're looking at code debt that we'll need to solve anyway and cleanup for these Extractors instead.

openpype/plugins/publish/integrate_new.py
BigRoy: Looking at what @mkolar writes [here](https://github.com/pypeclub/OpenPype/pull/2898#discussion_r840424267) deleted representations should be for versions that would still be writing e.g. a Prores file and integrating that instead. Which means that if NO representations are detected it's still the unexpected result and thus an error is valid. If e.g. the prores wasn't marked for delete then this would pass fine and things work as expected.

openpype/plugins/publish/integrate_new.py
iLLiCiTiT: It `template_data` may contain `"frame"` but template may not have `{frame}` in it.

BigRoy: > The rest of data are to be able get context from representation itself when something goes wrong (e.g. see in what context it was published against in what context is now).

Check. Totally agree - unused data should be ignored.
So the other keys from `db_representation_context_keys` sounds like are absolutely required instead of optional? 

Or are they "optional" for that we only add them if they were set we put them in the context even if unused by template? That is the current and old behavior. That does still allow the cases that we can't rely on those values being present.

---

Anyway, sounds like we could get by with this:

```
# Always add to representation context even if not used by template
self.representation_context_keys = [
    "project", "asset", "task", "subset", "version", "representation",
    "family", "hierarchy", "username"
]

# Only add to representation context if used by template for
# destination path formatting
self.representation_template_used_context_keys = ["frame", "udim"]
```

---

After all this time only now I see that `"task"` appears twice in `db_representation_context_keys` (also before the refactor) which seems redundant - will clean that up too.

openpype/plugins/publish/integrate_new.py
BigRoy: > We have to know which files were published for the representation not for instance.

That would still be the case, because those `files` list would also only have the files related to that particular representation. Albeit they would ALSO contain the `instance.data["transfers"]` since those files are also relevant to that representation. As shown in [this example](https://github.com/pypeclub/OpenPype/pull/2898#discussion_r840022014).

(Note that currently it does note store `instance.data["transfers"]` as files after the refactor at all and thus Site Sync would not work for those.)

openpype/plugins/publish/integrate_new.py
BigRoy: Resolving this discussion as it continued [here](https://github.com/pypeclub/OpenPype/pull/2898#discussion_r840439032).

openpype/plugins/publish/integrate_new.py
BigRoy: Removed `"frame"` from `db_representation_context_keys` with https://github.com/pypeclub/OpenPype/pull/2898/commits/3ec9684239b7afc326cad7e184a7c6ed4e7a6058

The `repre_context` always contains all used values - so that would still include "frame" only if it's used in destination template. Is this sufficient now?

openpype/plugins/publish/integrate_new.py
BigRoy: Not tested yet, but implemented with https://github.com/pypeclub/OpenPype/pull/2898/commits/3e095bc7554a24ef13282ccfd87e0327eb3b8745

openpype/plugins/publish/integrate_new.py
BigRoy: Updated the comment to clarify why we write version in advance. Only reason I had to move it later is to ensure file transfer went OK and try and implement the database writing into a single transaction that can be rolled back easily. But I'll consider it ok for now. 👍

openpype/plugins/publish/integrate_new.py
BigRoy: > We've had a discussion about this internally and agreed that we should actually be adding all resources into ['fils'] list to ensure that they are correctly handled by site sync, tracked properly in the db, plus is would actually remove the need for transfers and hardlinks lists.

@mkolar Currently I have implemented this on the database side (so the Integrator does actually integrate into `representation["files"]` in [the way I described in my previous comment](https://github.com/pypeclub/OpenPype/pull/2898#discussion_r840022014). Added with https://github.com/pypeclub/OpenPype/pull/2898/commits/89376a97e4ef85069a3afdfd5e3115b33bd27284

In essence each representation contains:
- Its own files, e.g. representation `ma` has the `.ma` and representation `abc` has the `.abc` file.
- The version's "resource" files, like textures. These are any files not explicitly assigned to a representation and considered required files for all representations. (What is currently stored in `instance.data["transfers"]` and `instance.data["hardlinks"]`)

To do is still to change how files are actually added as transfers, etc. in the extractors, etc. But currently at least the final integrator logic should now do what you expected it to do for Site Sync, etc. Yes?

openpype/plugins/publish/integrate_new.py
```diff
@@ -113,816 +159,538 @@ class IntegrateAssetNew(pyblish.api.InstancePlugin):
...
+            transfers = [(src, dst)]
+
+        # todo: Are we sure the assumption each representation
+        #       ends up in the same folder is valid?
+        if not instance.data.get("publishDir"):
+            instance.data["publishDir"] = (
+                anatomy_filled
+                [template_name]
+                ["folder"]
+            )
```
BigRoy, line 618: Can anyone confirm the question in this todo? This code is new after a recent pull of latest `develop` with merge conflicts.

openpype/plugins/publish/integrate_new.py
```diff
@@ -113,816 +159,538 @@ class IntegrateAssetNew(pyblish.api.InstancePlugin):
...
+        }.items():
+            # Allow to take value from representation
+            # if not found also consider instance.data
+            if key in repre:
+                value = repre[key]
+            elif key in instance.data:
+                value = instance.data[key]
+            else:
+                continue
+            template_data[anatomy_key] = value
```
BigRoy, line 499: After a recent merge conflict with latest develop which introduced `originalBasename` I refactored this code to first consider representation, if not found then consider `instance.data` values (which is where `originalBasename` was taking it from in latest develop).

This change should also allow to pick up `fps` from `instance.data` too which I believe was also in the old integrator code.

BigRoy: This does change the logic that a null value or zero value can still be set in `template_data`. Previously values were check with `if value:` - now they aren't. Thoughts on this?

openpype/plugins/publish/integrate_new.py
mkolar: @kalisp what behaviour will this trigger with respect to site sync? If I have a file in two representations, I assume both will try to sync the file independently correct?

openpype/plugins/publish/integrate_new.py
mkolar: I'd say that instead of error, we should be skipping integration in case of no representations present. From developer and TD standpoint you want to have the exception to make sure nothing is missed magically, however, this is mostly encounter by users (that's if the situation happens at all of course) in which case any error that they cannot affect and doesn't really break anything just invokes a small panic every time.

openpype/plugins/publish/integrate_new.py
mkolar: >Subset entity/document in the database that is outside of data which we want to preserve from the original subset?

We should do all we can to preserve the ID in the mongo. It could be already referenced elsewhere. Even if not directly by our tool, then possibly by a studio custom tool relying on its persistence.

openpype/plugins/publish/integrate_new.py
mkolar: Can you ever get a sequence without `frameStart`?

openpype/plugins/publish/integrate_new.py
mkolar: `published_path` is not aboutthe file itself, but rather being able to tell later on in the process where this publish in general ended up being integrated into. There is no other place where to get that information. And it needs to be the "main" file of the repre. If we're looking just at the `representation['files']` list, that you can't be sure which one of those files is representative of the publish location really, considering some might be a folder deeper in `resources` subfolder. 

This is another piece that could be solved by more robust `files` data on the repre though. As mentioned here https://github.com/pypeclub/OpenPype/pull/2898/files/a7a908d1348381ab0c4df9c29861d7c02be635cb#r839325287

To give a simple pseudoexample:

```
representation["files"] = [
	{
	source: "my_original_filename.abc", 							# filled before the integrator
	transfer_type: "copy", 										# filled before the integrator
	destination: "path/to/final/integrated/directory/new_filename.abc", 	# filled by integrator
	resource_type: "primary", 									# filled before the integrator
	},
	{
	source: "additional_texture.exr",
	transfer_type: "hardlink",
	destination: "path/to/final/integrated/directory/resources/new_filename.exr",
	resource_type: "auxiliary",
	}
]
```

That could cover all use cases of `transfers`, `hardlinks`, `published_path` and would also possibly match the files data model in the DB exactly and could be inserted (with exter keys like `created_dt`) directly into the db. It would also allow distinguishing between the `main` representation files and `resources` even from the db record, rather than just a simple list of files.

kalisp: What purpose have `transfer_type` and `resource_type`? 
I guess "gimme all hardlinks", "gimme all primary files", is there anything else we would need for it to be useful? (Just brainstorming here.)

mkolar: Well that `transfer_type` would be needed for the publishing process itself so the integrator know whether to copy or hardlink. Not really important for the DB, however, useful to query how much stuff we copy vs hardlink when looking into the project data in general. 

`resource_type` could be used by loaders to determine how to approach the various files. For instance look loader now simply expects that resources are in the `resources` subfolder in the publish (hardcoded), but with this info the loader could actually know what is the main shader file, vs what are the textures needed for it.

openpype/plugins/publish/integrate_new.py
mkolar: I believe it should be the case. Don't think there is a way to define where else representation might go.

openpype/plugins/publish/integrate_new.py
BigRoy: I'll log a warning and return early. We'll need to make sure other plug-ins afterwards don't do odd stuff then of course.
Or would you rather have it log just an info message?

antirotor:
adding to this conversation - related to his https://github.com/pypeclub/OpenPype/pull/2898/files/a7a908d1348381ab0c4df9c29861d7c02be635cb#r844801266

Imagine publishing of texture pack within one subset - you have your files that are not sequences and for some reason, they are correctly named from the beginning (and you have a validator to check it). This needs robust `files` as well because current state is that we expects that if `files` are list, they are sequence.

BigRoy:
> Imagine publishing of texture pack within one subset - you have your files that are not sequences and for some reason, they are correctly named from the beginning (and you have a validator to check it). This needs robust files as well because current state is that we expects that if files are list, they are sequence

This is exactly what explicit `transfers` did originally. They would describe a straight file copy from A to B. 

Nonetheless in essence I feel it's a matter of defining what types of A->B integrations we'd expect and continuously see if we could by with less or a simpler feature set to manage the same results. I think so far I've seen:

1. **Single file renamed to publish template** (matched by profile to set what template to use)
2. **Sequence renamed to publish template** (matched by profile to set what template to use)
    - Potentially shifted start frame
    - Might be UDIM sequence too - so there are cases to _never shift_ the files.
3. **Single file without rename into publish folder**
4. **_Sequence without rename into publish folder_*** -> _which is basically a list of single files instead of a single collection, without rename. So this is basically the same as 3)._
5. **Straight transfer to explicit custom destination path** (custom; for everything out of the ordinary where we want to be have full control over destination) -> currently used by e.g. maya lookdev to publish its textures without rename and move them into a `resources/` subfolder.

Then some questions remain: 
- Would we ever allow to write outside of the Publish folder. E.g. does 5) allow control to even break _that_ rule?
- Would the file for 3) be considered the published representation? Or is that still a 'resource' file alongside a representation file that follows naming template?

Anyway, it's an interesting problem where we should continuously define actual use cases why we absolutely require e.g. to not rename a file in a certain case, etc. aside of just being _requested_. 

It might actually point to a different issue altogether, e.g. if we are able to validate the actual filenames anyway like you're saying I assume they are following a template? Wouldn't that just mean those files are actually like 1)/2) but should have their templates set correctly? Of course it might not be that trivial, but it's always good to consider _why_ we must do it a certain way and also keep track of those decisions/goals for a feature somewhere.

openpype/plugins/publish/collect_subset_group.py
iLLiCiTiT: I think you're right in all of these. Only remaining issue I can think of is editorial publishing where asset documents may not exists at the collection time because are created during extraction. I don't know how it's handled now, but I'm 99% sure it can't be changed.

> Worst case scenario we could work around this by moving CollectSubsetGroup to IntegratorOrder - 0.499.

If it must be moved out of integrator then it's best solution for current state.

> Standalone Publisher creating new instances in Extractor

Oh my god! ...and it was me who did it?!!!

mkolar:
Just noting that this will need a merge of develop, that is removing avalon-core and some magic to make it compliant with that PR. Unfortunately the last step interfered with this.

BigRoy:
> Just noting that this will need a merge of develop, that is removing avalon-core and some magic to make it compliant with that PR. Unfortunately the last step interfered with this.

@mkolar Should be fixed.

antirotor:
> Please let me know what you need on this PR from my end to make this testable by OP team.

I'll try to test it this week per battle plan. I'll let you know what happened :)

BigRoy:
Looking forward to hearing about the test runs!

antirotor:
So far so good!

I've tested:
---
**Maya 2022**

- [X] Animation
- [ ] Ass StandIn
- [ ] Assembly
- [ ] Camera
- [ ] Camera Rig
- [X] Layout
- [X] Look
- [X] Maya Scene
- [X] Model
- [ ] Multiverse USD
- [ ] Multiverse USD Composition
- [ ] Multiverse USD Override
- [X] Point Cache
- [ ] Redshift Proxy
- [X] Render
- [ ] Render Setup Preset
- [X] Review
- [X] Rig
- [ ] Set Dress
- [X] Unreal Skeletal Mesh
- [X] Unreal Static Mesh
- [ ] VRay Proxy
- [ ] VRay Scene
- [ ] Xgen Interactive
- [ ] Yeti Cache
- [ ] Yeti Rig

---
**Blender 3.0**

- [ ] Action
- [ ] Animation
- [ ] Camera
- [X] Layout
- [ ] Model
- [ ] Point Cache
- [ ] Rig

---
**Houdini 19**

- [X] Camera
- [ ] Arnold ASS
- [ ] Composite (Image Sequence)
- [ ] Houdini Digital Asset (Hda)
- [ ] Point Cache
- [ ] Redshift ROP
- [ ] USD
- [ ] USD Render
- [ ] VDB Cache

---

BigRoy:
> So far so good!

Nice! We should actually add the following to testing too:

_Preferably each should be tested for publishes with a single file, multiple files, sequences and publishes with resources like Maya lookdev_

- [ ] Hero versions
- [ ] Site Sync
- [ ] A failing publish (e.g. Failing Extractor - does the Integrator fail nicely?)
- [ ] Farm Publishing (e.g. Maya Renders with Deadline)
- [ ] A publish that writes into an existing version (e.g. Maya Render re-publish with different frame range -> customize frame range in Deadline)
- [ ] A Python 2 host

antirotor:
Would you be so kind to split it in separate integrator so we can slowly add hosts there for better testing? Because I have a feeling that for Maya, this can be used right away (for example).

openpype/plugins/publish/collect_anatomy_context_data.py
```diff
@@ -92,5 +92,13 @@ def process(self, context):
...
             })
 
+        # todo: some code actually expects the dict itself and others doesn't
+        #       question: what should it be?
+        intent = context.data.get("intent")
+        if intent and isinstance(intent, dict):
+            intent = intent.get("value")
+        if intent:
+            context_data["intent"] = intent
+
```
iLLiCiTiT, line 102: Intent should be a dictionary with "value" and "label", to be able tell if you want use value or label of the intent in templates.
```suggestion
```

openpype/plugins/publish/integrate_new.py
iLLiCiTiT: Just had to handle that situation and moved filtering of representations as first step before anything happens in `IntegrateAssetNew` in [PR](https://github.com/pypeclub/OpenPype/pull/3398).

mkolar:
@BigRoy could you please separate the integrator into another file? that way we could actually merge it and start moving host by host to it. I've tested quite a bit in maya and the chances are if it works there it's good in other places, However, to be able to deply it into production I think we should expose the family lists of both the existing and the re-worked integrator to the settings temporarily. That way if something goes wrong with the new one, we can easily reassign a family to the old one and unblock production situation.

BigRoy:
Sorry, I've been so busy with day to day production that this keeps slipping through. Beginning of new day now, let's start! :) On it.

openpype/plugins/publish/collect_anatomy_context_data.py
BigRoy: Removed with recent commit. However, this logic with the if statement still feels like we should somehow correct other logic elsewhere so that we know it's always a `dict` to begin with.

openpype/plugins/publish/integrate_new.py
BigRoy: Great. That should resolve it - both old vs. new integrator now solve it by raising an error if I'm not mistaken.

BigRoy:
@mkolar Could you check if I missed anything crucial?

mkolar:
Thx. gonna try now

mkolar:
This should probably still be addressed https://github.com/pypeclub/OpenPype/pull/2898#discussion_r855003720

PR #2902, opened 2022-03-17
Fixing gap detection in extract review

jakubjezek001:
## Brief description
Gap was detect even there was none

### DIFF
--- a/openpype/plugins/publish/extract_review.py
+++ b/openpype/plugins/publish/extract_review.py
@@ -747,10 +747,14 @@ class ExtractReview(pyblish.api.InstancePlugin):
         collections = clique.assemble(files)[0]
         assert len(collections) == 1, "Multiple collections found."
         col = collections[0]
-        # do nothing if sequence is complete
-        if list(col.indexes)[0] == start_frame and \
-                list(col.indexes)[-1] == end_frame and \
-                col.is_contiguous():
+
+        # do nothing if no gap is found in input range
+        not_gap = True
+        for fr in range(start_frame, end_frame + 1):
+            if fr not in col.indexes:
+                not_gap = False
+
+        if not_gap:
             return []
 
         holes = col.holes()


iLLiCiTiT (approved):

openpype/plugins/publish/extract_review.py
```diff
@@ -747,10 +747,14 @@ def fill_sequence_gaps(self, files, staging_dir, start_frame, end_frame):
...
-        # do nothing if sequence is complete
-        if list(col.indexes)[0] == start_frame and \
-                list(col.indexes)[-1] == end_frame and \
-                col.is_contiguous():
+
+        # do nothing if no gap is found in input range
+        not_gap = True
+        for fr in range(start_frame, end_frame + 1):
+            if fr not in col.indexes:
+                not_gap = False
```
BigRoy, line 755: Shouldn't this just be:
```python
# Detect whether all expected frames are present
no_gaps = all(fr in col.indexes for range(start_frame, end_frame + 1))
```

Also. A little below in the code, this monster:
```
        # generate ideal sequence
        complete_col = clique.assemble(
            [("{}{:0" + str(col.padding) + "d}{}").format(
                col.head, f, col.tail
            ) for f in range(start_frame, end_frame)]
        )[0][0]  # type: clique.Collection
```
Could also just be:
```
frames = range(start_frame, end_frame + 1)
complete_col = clique.Collection(col.head, 
                                 col.tail, 
                                 col.padding, frames=set(frames))
```
I feel that's much more readable and explicit?

We could even do just the check `complete_col == col`?

BigRoy:
We could also use the [concat demuxer](https://trac.ffmpeg.org/wiki/Slideshow#Concatdemuxer) potentially to not even require to duplicate the frames on disk? Or even pass them in through [pipe](https://trac.ffmpeg.org/wiki/Slideshow#pipe)?

openpype/plugins/publish/extract_review.py
BigRoy: This seems like the exact technical debt that we should be cleaning @mkolar, no? Making sure the code is easy to parse and understand what it does. Like replacing the monstrosity of that clique assemble line ;)

jakubjezek001: @iLLiCiTiT so how would you write it yourself? Just to see the alternative and not only pointing and mistakes ;)

jakubjezek001: I am happy to change this one, i agree that the readibility is not the best one..

iLLiCiTiT: Not sure if this PR is best place to discuss?

This is probably the best variant:
```
if f_subsets:
    if any(re.search(filter_pattern, subset) for filter_pattern in f_subsets):
        continue
```

jakubjezek001: here is the correct PR https://github.com/pypeclub/OpenPype/pull/2903/files

PR #2903, opened 2022-03-17
Nuke: improving readability

jakubjezek001:
## Brief description
Code was too complex to read

## Testing notes:
1. configure `project_settings/nuke/publish/ExtractReviewDataMov/outputs` to have multiple presets with use of filters
2. open your testing nuke script
3. publish render write node

### DIFF
--- a/openpype/hosts/nuke/plugins/publish/extract_review_data_mov.py
+++ b/openpype/hosts/nuke/plugins/publish/extract_review_data_mov.py
@@ -51,39 +51,18 @@ class ExtractReviewDataMov(openpype.api.Extractor):
                 f_subsets = o_data["filter"]["sebsets"]
 
                 # test if family found in context
-                test_families = any([
-                    # first if exact family set is matching
-                    # make sure only interesetion of list is correct
-                    bool(set(families).intersection(f_families)),
-                    # and if famiies are set at all
-                    # if not then return True because we want this preset
-                    # to be active if nothig is set
-                    bool(not f_families)
-                ])
+                # using intersection to make sure all defined
+                # families are present in combinantion
+                test_families = not f_families or any(
+                    set(families).intersection(f_families))
 
                 # test task types from filter
-                test_task_types = any([
-                    # check if actual task type is defined in task types
-                    # set in preset's filter
-                    bool(task_type in f_task_types),
-                    # and if taskTypes are defined in preset filter
-                    # if not then return True, because we want this filter
-                    # to be active if no taskType is set
-                    bool(not f_task_types)
-                ])
+                test_task_types = not f_task_types or any(
+                    task_type in f_task_types)
 
                 # test subsets from filter
-                test_subsets = any([
-                    # check if any of subset filter inputs
-                    # converted to regex patern is not found in subset
-                    # we keep strict case sensitivity
-                    bool(next((
-                        s for s in f_subsets
-                        if re.search(re.compile(s), subset)
-                    ), None)),
-                    # but if no subsets were set then make this acuntable too
-                    bool(not f_subsets)
-                ])
+                test_subsets = not f_subsets or any(
+                    re.search(s, subset) for s in f_subsets)
 
                 # we need all filters to be positive for this
                 # preset to be activated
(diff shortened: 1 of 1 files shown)


BigRoy:
I agree with @iLLiCiTiT [this comment](https://github.com/pypeclub/OpenPype/pull/2902#discussion_r829244845) is likely the best format. Which makes it something like this:

```python
# please consider example as pseudocode - untested
...

                # process only allowed families
                if f_families:
                    if set(families).intersection(f_families):
                        continue

                # process only allowed task types
                if f_task_types:
                    if not task_type in f_task_types:
                        continue
                        
                # process only subsets matching patterns
                if f_subsets:
                    if any(re.search(pattern, subset) 
                           for pattern in f_subsets)):
                       continue
```

This also means the lines below it where you check all can be removed:
```python
                # we need all filters to be positive for this
                # preset to be activated
                test_all = all([
                    test_families,
                    test_task_types,
                    test_subsets
                ])

                # if it is not positive then skip this preset
                if not test_all:
                    continue
```

jakubjezek001:
> I agree with @iLLiCiTiT [this comment](https://github.com/pypeclub/OpenPype/pull/2902#discussion_r829244845) is likely the best format. Which makes it something like this:
> 
> ```python
> # please consider example as pseudocode - untested
> ...
> 
>                 # process only allowed families
>                 if f_families:
>                     if set(families).intersection(f_families):
>                         continue
> 
>                 # process only allowed task types
>                 if f_task_types:
>                     if task_type in f_task_types:
>                         continue
>                         
>                 # process only subsets matching patterns
>                 if f_subsets:
>                     if any(re.search(pattern, subset) 
>                            for pattern in f_subsets)):
>                        continue
> ```
> 
> I believe above logic says that if it matches any of the filters then they are continued and excluded which might be the opposite of what you'd need? :D
> 
> This also means the lines below it where you check all can be removed:
> 
> ```python
>                 # we need all filters to be positive for this
>                 # preset to be activated
>                 test_all = all([
>                     test_families,
>                     test_task_types,
>                     test_subsets
>                 ])
> 
>                 # if it is not positive then skip this preset
>                 if not test_all:
>                     continue
> ```

Brilliant! Thank you @BigRoy . Code looks and reads much better this way ;)

openpype/hosts/nuke/plugins/publish/extract_review_data_mov.py
```diff
@@ -51,50 +51,20 @@ def process(self, instance):
...
-                    # first if exact family set is matching
-                    # make sure only interesetion of list is correct
-                    bool(set(families).intersection(f_families)),
-                    # and if famiies are set at all
-                    # if not then return True because we want this preset
-                    # to be active if nothig is set
-                    bool(not f_families)
-                ])
+                # using intersection to make sure all defined
+                # families are present in combinantion
```
BigRoy, line 55: Typo: `combinantion` -> `combination`

openpype/hosts/nuke/plugins/publish/extract_review_data_mov.py
```diff
@@ -58,7 +72,7 @@ def process(self, instance):
                     continue
 
                 # test task types from filter
-                if f_task_types and not any(
+                if f_task_types and not bool(
                         task_type in f_task_types):
```
BigRoy, line 76: There's no need to turn this into a `bool` explicitly.

```suggestion
                if f_task_types and task_type not in f_task_types:
```

openpype/hosts/nuke/plugins/publish/extract_review_data_mov.py
```diff
@@ -50,51 +52,33 @@ def process(self, instance):
...
-                    bool(set(families).intersection(f_families)),
-                    # and if famiies are set at all
-                    # if not then return True because we want this preset
-                    # to be active if nothig is set
-                    bool(not f_families)
-                ])
+                # using intersection to make sure all defined
+                # families are present in combination
+                if f_families and not any(
+                        set(families).intersection(f_families)):
```
BigRoy, line 71: No need to check `any` because if the set intersection is empty it will already be used as `False`.
Thus you should be able to just do:
```
               if f_families and not set(families).intersection(f_families):
```

Even better if you change `families` into a set outside of the for loop at the top:

```
families = set(instance.data["families"])
```

And then here just do:
```
               if f_families and not families.intersection(f_families):
```

jakubjezek001: nice, thanks

iLLiCiTiT (approved):

PR #2904, opened 2022-03-17
General: Fix use of Anatomy roots

iLLiCiTiT:
## Brief description
Get representation path used `roots_obj` from Anatomy instead of `roots`.

## Changes
- use `roots` attribute from Anatomy object instead of `roots_obj` when different project roots should be used

## Testing notes:
1. Try copy path from loader launched from tray


BigRoy (approved):
Looks good to me.

PR #2907, opened 2022-03-17
General: Change how OPENPYPE_DEBUG value is handled

iLLiCiTiT:
## Brief description
Env variable `OPENPYPE_DEBUG` must be set to 2 or higher number to enable debugging which is unexpected for most of cases.

## Changes
- `OPENPYPE_DEBUG` can be set to any number higher than `0` to start emit debug log
- all changes of `OPENPYPE_DEBUG` setting the value has been changed to set it to `0`

## Testing notes:
1. Run any command with env variable `OPENPYPE_DEBUG` set to `"1"`

### DIFF
--- a/openpype/cli.py
+++ b/openpype/cli.py
@@ -101,7 +101,7 @@ def eventserver(debug,
     on linux and window service).
     """
     if debug:
-        os.environ['OPENPYPE_DEBUG'] = "3"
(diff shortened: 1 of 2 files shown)


BigRoy (approved):
LGTM

mkolar (approved):

PR #2911, opened 2022-03-18
Nuke: Add no-audio Tag

jrsndl:
Allow skipping audio for reviews.

## Brief description
Some reviews can't have audio. This tag allows to flag reviews to be mute.

## Description
Ffmpeg can fail adding audio to files that cant' have any, like OpAtom MXFs

## Testing notes:
1. Add no-audio flag in extractReview profile
2. Publish shot with audio available (from Nuke)
3. Check that review is mute

### DIFF
--- a/openpype/settings/entities/schemas/projects_schema/schemas/schema_representation_tags.json
+++ b/openpype/settings/entities/schemas/projects_schema/schemas/schema_representation_tags.json
@@ -24,6 +24,9 @@
         },
         {
             "sequence": "Output as image sequence"
+        },
+        {
+            "no-audio": "Do not add audio"
         }
     ]
 }


jakubjezek001 (commented):
I wonder if this will solve the issue. By adding tag to newly created representation which is having audio included, we are telling publish plugins which will follow downstream to ignore audio. 

In case you want to exclude audio processing on particular output profile, it has to be added as bool attribute.

mkolar:
I think this does solve the issue. The `extractReview` code is already counting of the possibility of this tag being present in the output profile, but we're not actually able to use it because it's not present in the schema. This just makes it available.

jakubjezek001 (approved):

PR #2912, opened 2022-03-18
General: Move remaining plugins from avalon

iLLiCiTiT:
## Brief description
Move remained plugins/actions from avalon into OpenPype.

## Description
Moved last plugins and actions from avalon repository `ThumbnailResolver `, `InventoryAction` and `Action`. Action `Action` was renamed to `LauncherAction`.

## Changes
- added `InventoryAction` and `LauncherAction` into `openpype/pipeline/actions.py`
- added `ThumbnailResolver` into `openpype/pipeline/thumbnails.py`
- removed registering of inventory path in hosts without inventory actions
- changed imports of moved code

## Testing notes:
1. launcher actions should work in Launcher UI
2. thumbnails should be visible if are published
3. actions in Scene Inventory tools should be visible

Related to https://github.com/pypeclub/avalon-core/pull/433

### DIFF
--- a/openpype/pipeline/__init__.py
+++ b/openpype/pipeline/__init__.py
@@ -41,6 +41,22 @@ from .publish import (
     OpenPypePyblishPluginMixin
 )
 
+from .actions import (
+    LauncherAction,
+
+    InventoryAction,
+
+    discover_launcher_actions,
+    register_launcher_action,
+    register_launcher_action_path,
+
+    discover_inventory_actions,
+    register_inventory_action,
+    register_inventory_action_path,
+    deregister_inventory_action,
+    deregister_inventory_action_path,
+)
+
 
 __all__ = (
     "attribute_definitions",
@@ -82,5 +98,19 @@ __all__ = (
     "PublishValidationError",
     "PublishXmlValidationError",
     "KnownPublishError",
-    "OpenPypePyblishPluginMixin"
+    "OpenPypePyblishPluginMixin",
+
+    # --- Plugins ---
+    "LauncherAction",
+    "InventoryAction",
+
+    "discover_launcher_actions",
+    "register_launcher_action",
+    "register_launcher_action_path",
+
+    "discover_inventory_actions",
+    "register_inventory_action",
+    "register_inventory_action_path",
+    "deregister_inventory_action",
+    "deregister_inventory_action_path",
 )
--- a/openpype/tools/launcher/actions.py
+++ b/openpype/tools/launcher/actions.py
@@ -1,6 +1,7 @@
 import os
 
-from avalon import api
+from Qt import QtWidgets, QtGui
+
 from openpype import PLUGINS_DIR
 from openpype import style
 from openpype.api import Logger, resources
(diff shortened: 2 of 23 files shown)


openpype/pipeline/__init__.py
```diff
@@ -82,5 +98,19 @@
     "PublishValidationError",
     "PublishXmlValidationError",
     "KnownPublishError",
-    "OpenPypePyblishPluginMixin"
+    "OpenPypePyblishPluginMixin",
+
+    # --- Plugins ---
```
BigRoy, line 103: Might this be better labeled actions? Because above in load there are also "plugins"?

```suggestion
    # --- Actions ---
```

openpype/tools/sceneinventory/view.py
```diff
@@ -5,13 +5,14 @@
...
 
-from avalon import io, api
+from avalon import io
 
 from openpype import style
 from openpype.pipeline import (
     HeroVersionType,
     update_container,
     remove_container,
+    discover_inventory_actions,
```
BigRoy, line 15: Command at the end is redundant - it should work as is but might be cleaner without?
```suggestion
    discover_inventory_actions
```

openpype/tools/launcher/actions.py
```diff
@@ -74,7 +81,7 @@ class ApplicationAction(api.Action):
     @property
     def log(self):
         if self._log is None:
-            self._log = Logger().get_logger(self.__class__.__name__)
+            self._log = Logger.get_logger(self.__class__.__name__)
```
BigRoy, line 84: Good catch - both should work but it's definitely easier to read and more explicit that we're using (in this case) a classmethod. A quick search over the code base does show that `Logger().get_logger(`  is used in many more places. Would it be worth it to remap all of them for consistency (potentially separate PR of course)

iLLiCiTiT: Yes, but I would wait after refactor. We will move code from place to place during the refactor.

jakubjezek001 (changes requested):
I had the same issue here as it was #2918 

```
  File "C:\CODE\__PYPE\OpenPype\openpype\tools\launcher\window.py", line 412, in run_action
    action().process(filtered_session)
  File "C:\CODE\__PYPE\OpenPype\openpype\tools\launcher\actions.py", line 115, in process
    **self.data
  File "C:\CODE\__PYPE\OpenPype\openpype\lib\applications.py", line 365, in launch
    return self.manager.launch(self.full_name, *args, **kwargs)
  File "C:\CODE\__PYPE\OpenPype\openpype\lib\applications.py", line 504, in launch
    return context.launch()
  File "C:\CODE\__PYPE\OpenPype\openpype\lib\applications.py", line 1140, in launch
    prelaunch_hook.execute()
  File "C:\CODE\__PYPE\OpenPype\openpype\hooks\pre_global_host_data.py", line 45, in execute
    prepare_context_environments(temp_data)
  File "C:\CODE\__PYPE\OpenPype\openpype\lib\applications.py", line 1528, in prepare_context_environments
    _prepare_last_workfile(data, workdir)
  File "C:\CODE\__PYPE\OpenPype\openpype\lib\applications.py", line 1612, in _prepare_last_workfile
    last_workfile_path = avalon.api.last_workfile(
```

jakubjezek001 (approved):
tests were all passing

Issue #2393, opened 2021-12-13
Maya: Invalid review flag on rendered AOVs 🏴


antirotor:
## Bug

AOVs sent to farm are marked for review even when they shouldn't be - the don't match regex specified in Settings. This is causing "technical" data to have created thumbnails and review videos. That works by itself until ffmpeg hits data pass with values it doesn't undrestand, like ID passes where there is no rgba channel.

### How to reproduce

In `project_settings/deadline/publish` set **Reviewable subsets filter** for Maya to regex `^(?:(?!_).)*$`. This regex matches `Main.0001.exr` but not `Main_foo.0001.exr`, or matches `Second.0001.exr` but not `Second_foo_boo.0001.exr`. This should skip all sequences with AOVs but not beauty pass, as beauty is handled by regex `.+(?:\.|_)([Bb]eauty)(?:\.|_).*` set before and has files like `Main.0001.exr` or `Main_beauty.0001.exr` and AOVs are rendered as `Main_P.0001.exr` for example. Publish to farm and observer as thumbails and reviews are generated.

📓 **This is only for non-multilayered exr**

### Technical detail

Problem is probably somewhere here:
https://github.com/pypeclub/OpenPype/blob/9df8cd1b582c0e14f4dc40dc2bbd1d1cbf0fe2ba/openpype/modules/default_modules/deadline/plugins/publish/submit_publish_job.py#L441-L455

[cuID:1xb4fta]

antirotor:
Could be problem with filtering by file name vs actual AOV name from the list.

antirotor:
This must also solve generation of reviews for Redshift where there is no `_beauty` attached.

PR #2915, opened 2022-03-18
Maya: invalid review flag on rendered AOVs

Allan-I:
Fixes issue #2393.  

The commit fix issue in `submit_publish_job.py ` and `deadline.json `to properly compare the AOV pattern we're grabbing from the settings against the resulting AOV strings appended by publishing a render (in Maya).  

The regex within `deadline.json` was incorrect and wasn't matching `aov_filter `.


mkolar:
@Allan-I please fix the hound comments. Otherwise it works as expected now.

openpype/settings/defaults/project_settings/deadline.json
```diff
@@ -95,6 +68,33 @@
...
+            "deadline_department": "",
+            "deadline_pool": "",
+            "deadline_group": "",
+            "deadline_chunk_size": 1,
+            "deadline_priority": 50,
+            "publishing_script": "",
+            "skip_integration_repre_list": [],
+            "aov_filter": {
+                "maya": [
+                    ".*(?:[\\._-])*([Bb]eauty)(?:[\\.|_])*.*"
```
BigRoy, line 83: The extra * after both of the non capture groups `(?:[\\._-])` & `(?:[\\.|_])` sounds off. We allow zero, one or more occurrences of it? If so, why do we even try and match it if it can also not be there at all?
`(?:` marks a non capture group. So the group it captures can be ignored.
Then after the non capture group we allow anything any amount of times `.*`?

If we don't care about `(?:[\\.|_])*` this particular match. Couldn't we just make this?
```
.*([Bb]eauty).*
```

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -448,7 +448,11 @@ def _create_instances_for_aov(self, instance_data, exp_files):
             preview = False
             if app in self.aov_filter.keys():
                 for aov_pattern in self.aov_filter[app]:
-                    if re.match(aov_pattern, aov):
+                    # Matching against the AOV pattern in the render files
+                    # In order to match the AOV name, we must compare against the render filename string
+                    # We are grabbing the render filename string from the collection that we have grabbed from expected files (exp_files)
+                    render_file_name = os.path.basename(col[0])
```
BigRoy, line 454: According to [these lines](https://github.com/pypeclub/OpenPype/blob/509f3289418a297ac17e3a0de93a987390cd8370/openpype/modules/deadline/plugins/publish/submit_publish_job.py#L403-L408) can `col` also be a single "remainder" which is not a collection but a single filepath? If so, doing `col[0]` would just return the first character of that string.

You can examine a similar "problem" being worked around [here](https://github.com/pypeclub/OpenPype/blob/509f3289418a297ac17e3a0de93a987390cd8370/openpype/modules/deadline/plugins/publish/submit_publish_job.py#L428-L431).

openpype/settings/defaults/project_settings/deadline.json
mkolar: True that simpler regex works too, from a quick test. Keep in mind also that this is technically an example default, that often needs adjusting in the studio, but I agree, let's keep it simpler.

openpype/modules/deadline/plugins/publish/submit_publish_job.py
mkolar: Damn. You're right, that should be caught.

Allan-I:
Thanks guys will make these changes!

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -448,7 +448,16 @@ def _create_instances_for_aov(self, instance_data, exp_files):
             preview = False
             if app in self.aov_filter.keys():
                 for aov_pattern in self.aov_filter[app]:
-                    if re.match(aov_pattern, aov):
+                    # Matching against the AOV pattern in the render files
+                    # In order to match the AOV name
+                    # we must compare against the render filename string 
+                    # We are grabbing the render filename string
+                    # from the collection that we have grabbed from exp_files 
+                    if isinstance(col, list):
```
BigRoy, line 456: The other fallback checks compare again `(list, tuple)`. Not sure if that is either too wide or this one is to narrow.
Anyway, we don't use the remainder any differently than the collection - but I do see now that the remainder branch will bug out anyway, because I believe it wouldn't have a `.head` attribute which is used [here](https://github.com/pypeclub/OpenPype/blob/80e08d43ce3617d57280a04ae426c0d42a595709/openpype/modules/deadline/plugins/publish/submit_publish_job.py#L422). Or am I mistaken?

@mkolar what's the best way to approach this? I feel the best way would be trying to work around needing all these if/else statements everywhere. Most of the code would just need a list of files looking at what it does and it's trivial to turn both cases into a valid `files` variable. E.g. `files = list(col[0])` and `files = [rem]`.

I know this PR is technically not about solving this - but the code can be greatly simplified if some time is spent fixing these issues too.

Allan-I: Hey thanks man.  I opted for `list `only as I'm supposing the incoming object is a `list ` only (as that's what's being grabbed higher in the file, if I'm not mistaken).  But I'm not entirely sure to be honest whether I should be checking for `tuple `too, or not.

Also, I think there's some work being done by @antirotor on the data incoming and on a new version of this, maybe, but I have no idea as to the state nor am I 100% sure that that's even in the works.

antirotor: The check for tuple is not necessary as col will never be a tuple (it is casted to list or it should be just a string in case of single file. The reason why we don't use even single file as a list is because we still have no way in representation to differentiate between single file sequence and single file. I agree that this code needs refactoring but in connection to other issues we have down the line.

antirotor: But I think we introduced `is_sequence` tag to representation later on, so it might be obsolete. Now I am not sure. Anyway, I would leave it to different PR

openpype/pipeline/farm/patterning.py
```diff
@@ -0,0 +1,16 @@
...
+    In order to match the AOV name
+    we must compare against the render filename string
+    that we are grabbing the render filename string
+    from the collection that we have grabbed from exp_files.
+    """
+    if app in self.aov_filter.keys():
+        for aov_pattern in self.aov_filter[app]:
+            if re.match(aov_pattern, render_file_name):
+                preview = True
+                return preview
```
BigRoy, line 16: I don't really understand why we're currently putting this into its own file, especially because it still uses `self` as if it's on the class - it's not self-contained enoug with its arguments to belong on its own. Why are we making this so much more 'complicated'?

BigRoy:
Why empty file `flagging.py` added with https://github.com/pypeclub/OpenPype/pull/2915/commits/c6cfdfbd3aac60dd8469048354b423f39edd4e9a? Seems unused and redundant.

openpype/pipeline/farm/patterning.py
Allan-I: Thanks Roy, the idea is that we're going about refactoring some stuff to reorganize the code a bit, it's being used in a couple of places in that file and it will be reused in some other submitters as well iirc the discussion that took place. 

`self` being passed might present an issue later on but it's not necessarily breaking oop principles, for now imho.

Allan-I:
> Why is empty file `flagging.py` added with [c6cfdfb](https://github.com/pypeclub/OpenPype/commit/c6cfdfbd3aac60dd8469048354b423f39edd4e9a)? Seems unused and redundant.

Thanks Roy, 

> Why is empty file `flagging.py` added with [c6cfdfb](https://github.com/pypeclub/OpenPype/commit/c6cfdfbd3aac60dd8469048354b423f39edd4e9a)? Seems unused and redundant.

Thank you I don't know how I missed this.

openpype/pipeline/farm/patterning.py
BigRoy: If we're moving it to a separate file we should clean this up and make it generic enough to be usable - otherwise it's just a separate confusing part that we'll be left unused. I feel like we should only move it out if we're also refactoring other uses (e.g. not in the same plug-in).

Currently that'd maybe be `collect_render` for After Effects? (And that doesn't store it as `self.aov_filter` so the logic doesn't hold?

In essence this function does this:
```python
def match_patterns(patterns, value):
    if not patterns:
        return False
    return any(re.match(pattern, value) for pattern in patterns)
```

Could likely even be optimized to:
```python
def match_patterns(patterns, value):
    return re.match("|".join(patterns), value)
```

With the usage in the other file being:
```python
aov_filter_patterns = self.aov_filter.get(app, [])
match_patterns(aov_filter_patterns, render_file_name)
```

I don't see why we are currently in dire need to refactor this logic and warrant having it out in its own file, especially with how non-generic the code currently is.

Allan-I: Ah, I see and understand your point.  For the logic, yes indeed it's doing the exact same thing, and even though that's more pythonic, it seems a little difficult to read for me

As for the refactoring, I think the guys have a better understanding of what's happening with the refactoring

BigRoy: I'm totally for code re-use and refactoring to avoid code duplication. My point mostly is that the current state of this refactor doesn't make my sense and I'd hold of merging until cleaned.

Allan-I: Thank you, I'm waiting for the guys to give it a looksie, I do lack the thorough understanding of the codebase to kind of segregate things out further, but from what I understood from the discussion the exact bit is to be reused in an identical way in a couple other places (checking if in context) and so on, so that's why I approached it like such, otherwise I really want to take out the contextual aspect of it

BigRoy: > "I think it can be used even outside the deadline - in royal render module or in muster (if resurrected)"

If that's the case - then lets set it up like that. 

> And in that case signature of externalized function def match_aov_pattern(app, aov_patterns, render_file_name) might be more useful, as it would provide you more reusability.

Exactly this - and only if the function body has some complexity to it that requires consistency in multiple areas. If it's just a matter of regex matching I'm not sure separating now is the best time, but if we expect more complexity over time. Sure.

However if just a preparating I'd opt for adding it as a separate function within the Plugin file until it is actually used elsewhere. Unless we expect to directly create another PR to actually use that lib function. But that'd just be a personal preference just so we don't create clutter where we don't need it yet.

Allan-I: Hey guys, thanks a lot for the notes.  We could potentially skip `aov_filter_patterns = self.aov_filter.get(app, [])` through directly using `aov_patterns = self.aov_filter.keys()` outside the function and `aov_pattern in aov_patterns[app]` within it as such:

```
for aov_pattern in aov_patterns[app]:
        if not aov_patterns:
            return False
        else:
            return any(re.match(aov_pattern, render_file_name) for aov_pattern in aov_patterns)   ```


What do you think?

And additionally testing within the for loop for it even existing through `if aov_pattern` would perhaps be a good idea too

Any input would be helpful!

Comment (Allan-I) [reply openpype/pipeline/farm/patterning.py:16]: Ah, I see!  Thanks man I appreciate it!

File: openpype/pipeline/farm/patterning.py [hunk 5]
```diff
@@ -20,4 +20,6 @@ def match_aov_pattern(app, aov_patterns, render_file_name):
     """
     aov_pattern = aov_patterns.get(app, [])
     if aov_pattern:
-        return any(re.match(aov_pattern, render_file_name))    
+        if re.match(aov_pattern, render_file_name):
+            preview = True
+            return preview
```
BigRoy, line 25: This would basically be:
```suggestion
        return re.match(aov_pattern, render_file_name):
```

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -446,12 +447,15 @@ def _create_instances_for_aov(self, instance_data, exp_files):
...
-                    if re.match(aov_pattern, aov):
-                        preview = True
-                        break
 
+            if isinstance(col, list):
+                render_file_name = os.path.basename(col[0])
+            else:
+                render_file_name = os.path.basename(col)
+            aov_patterns = self.aov_filter.keys()
+            preview = match_aov_pattern(app, aov_patterns, render_file_name)
```
BigRoy, line 456: I don't think we should be passing just the keys to the function. This seems like a bug?

openpype/pipeline/farm/patterning.py
Allan-I: Hey man thank you, the problem is if it doesn't match it would return a `None` so I thought perhaps being a bit more explicit would be helpful.

openpype/modules/deadline/plugins/publish/submit_publish_job.py
Allan-I: Thank you it might be, will take a look in a bit

Allan-I: Yep you are right, definite bug I'm looking into it

openpype/pipeline/farm/patterning.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+import re
+
+
+def match_aov_pattern(app, aov_patterns, render_file_name):
```
iLLiCiTiT, line 5: Please change `app` to `host_name`.

openpype/pipeline/farm/patterning.py
```diff
@@ -0,0 +1,27 @@
...
+    Returns:
+        bool: Review state for rendered file (render_file_name).
+    """
+    aov_pattern = aov_patterns.get(app, [])
+    if aov_pattern:
+        if re.match(aov_pattern, render_file_name):
+            preview = True
+            return preview
+        else:
+            return False
```
iLLiCiTiT, line 27: I think that @BigRoy  wanted to say that you don't need those conditions at all because the can be replaced completely with the result of `match`. There is also unhandled situation when the pattern is not set.
```suggestion
    if not aov_pattern:
        return False
    return re.match(aov_pattern, render_file_name)
```

Allan-I: Thank you @iLLiCiTiT, I'll fix the unhandled situation for an unset pattern, but for `re.match` my understanding is it returns `None` if it doesn't match.  I might be misunderstanding the docs here: https://docs.python.org/3/library/re.html#re.match 

Please let me know so I can fix this if I'm wrong.

openpype/pipeline/farm/patterning.py
Allan-I: Thank you, just fixed it!

openpype/pipeline/farm/patterning.py
iLLiCiTiT: You're right. It can still be simplified (Edited the suggestion). Question is how to handle "empty regex string". Does it match in that case, or it should return `False` like in the suggestion?

Allan-I: Thank you.  Based on my understanding it should return `False ` so it sets the review status (and that was the intended behavior at the beginning).  If a studio (or artist) wants to, they'd need to specify the regex, imho I think assuming intended behavior needing to be explicit would always be a good idea.

Allan-I: I wonder if it's worth it to use  `re.fullmatch` (https://docs.python.org/3/library/re.html#re.fullmatch) so as to simplify mainly adding to the benefit of being explicit about it, like such:

```
if re.fullmatch(aov_pattern, render_file_name):
    return True
else:
    return False

```

openpype/pipeline/farm/patterning.py
```diff
@@ -0,0 +1,24 @@
...
+    """Matching against a `AOV` pattern in the render files.
+
+    In order to match the AOV name we must compare
+    against the render filename string that we are
+    grabbing the render filename string  from the collection
+    that we have grabbed from `exp_files`.
+
+    Args:
+        app (str): Host name.
+        aov_patterns (list): List of AOV patterns from AOV filters.
```
antirotor, line 15: aov_pattern isn't list, its dictionary:
```suggestion
        aov_patterns (dict):  AOV patterns from AOV filters.
```

openpype/pipeline/farm/patterning.py
```diff
@@ -0,0 +1,24 @@
...
+        render_file_name (str): Incoming file name to match against.
+
+    Returns:
+        bool: Review state for rendered file (render_file_name).
+    """
+    aov_pattern = aov_patterns.get(host_name, [])
+    if not aov_pattern:
+        return False
+    return re.match(aov_pattern, render_file_name) is not None
```
antirotor, line 24: Its crashing for me. I think it is because `re.match` accepts only string as a pattern, not list.

Now - `aov_patterns` are set by default like this:
```python
{'maya': ['.*([Bb]eauty).*'], 'nuke': ['.*'], 'aftereffects': ['.*'], 'celaction': ['.*'], 'harmony': ['.*']}
```
so in case of maya and my test, the return line becomes:

```python
return re.match(['.*([Bb]eauty).*'], "Second_N.1001.exr") is not None
```
and this will crash with:

```
  File "E:\projects\openpype\3x\.venv\lib\site-packages\pyblish\plugin.py", line 522, in __explicit_process
    runner(*args)
  File "E:\projects\openpype\3x\openpype\modules\deadline\plugins\publish\submit_publish_job.py", line 850, in process
  File "E:\projects\openpype\3x\openpype\modules\deadline\plugins\publish\submit_publish_job.py", line 460, in _create_instances_for_aov
  File "E:\projects\openpype\3x\openpype\pipeline\farm\patterning.py", line 24, in match_aov_pattern
    return re.match(aov_pattern, render_file_name) is not None
  File "D:\gfx\Autodesk\Maya2022\Python37\lib\re.py", line 173, in match
    return _compile(pattern, flags).match(string)
  File "D:\gfx\Autodesk\Maya2022\Python37\lib\re.py", line 276, in _compile
    return _cache[type(pattern), pattern, flags]
TypeError: unhashable type: 'list'
```
Untested suggestion:
```suggestion
    return any(re.match(p, render_file_name) for p in aov_pattern)
```

antirotor (changes requested):

antirotor (approved):
It works for me, but another set of eyes would be better.

m-u-r-p-h-y (approved):
did some render tests and it seems it is working as expected.

Allan-I:
You got it!

PR #2410, opened 2021-12-15
Unreal: Rendering implementation

simonebarbieri:
This PR adds support for rendering in Unreal.

PR #2917, opened 2022-03-18
Unreal: Render Publishing

simonebarbieri:
## Brief description
Implementation of publishing for renders in Unreal

## Additional info
Requires #2410


openpype/hosts/unreal/plugins/publish/validate_sequence_frames.py
```diff
@@ -0,0 +1,45 @@
...
+        for repr in representations:
+            frames = []
+            for x in repr.get("files"):
+                # Get frame number. The last one contains the file extension,
+                # while the one before that is the frame number.
+                # `lstrip` removes any leading zeros. `or "0"` is to tackle
+                # the case where the frame number is "00".
+                frame = int(str(x).split('.')[-2])
+                frames.append(frame)
+            frames.sort()
```
BigRoy, line 35: For consistency with other areas of OpenPype code base it might make more sense to rely on `clique` for sequences. No?

In this case (where you require the frame number between two dots) you can use exactly [this logic](https://github.com/pypeclub/OpenPype/pull/2875/files#diff-dfd3bf8586623379142bdb58490bf540cca132c24f8fdfd8104095fc66698797R24-R26):

Something like this:
```python
patterns = [clique.PATTERNS["frames"]]
collections, remainder = clique.assemble(files, minimum_items=1,
                                         patterns=patterns)
assert not remainder, "Must not have remainder"
assert len(collections) == 1, "Must detect single collection"
collection = collections[0]
frames = list(collection.indexes)
```

simonebarbieri: Done!

openpype/hosts/unreal/plugins/publish/collect_remove_marked.py
```diff
@@ -0,0 +1,24 @@
...
+
+        self.log.debug(context)
+        # make ftrack publishable
+        instances_to_remove = []
+        for instance in context:
+            if instance.data.get('remove'):
+                instances_to_remove.append(instance)
+
+        for instance in instances_to_remove:
+            context.remove(instance)
```
BigRoy, line 24: This logic is a duplicate of other hosts.

In [StandalonePublisher](https://github.com/pypeclub/OpenPype/blob/7017410173647da9a122e0659fda2c7af5593c80/openpype/hosts/standalonepublisher/plugins/publish/collect_remove_marked.py) and [Maya](https://github.com/pypeclub/OpenPype/blob/97cd765363961d8a2e756c17cf0a91aba95f3c95/openpype/hosts/maya/plugins/publish/collect_remove_marked.py). Would there be a way to unify this behavior?

E.g. make a single plug-in in global and add `hosts = ["maya", "standalonepublisher", "unreal"]` filter or alike?

openpype/hosts/unreal/plugins/publish/collect_render_instances.py
```diff
@@ -0,0 +1,103 @@
...
+                    new_data["frameStart"] = s.get('frame_range')[0]
+                    new_data["frameEnd"] = s.get('frame_range')[1]
+                    new_data["sequence"] = seq.get_path_name()
+                    new_data["master_sequence"] = data["master_sequence"]
+                    new_data["master_level"] = data["master_level"]
+
+                    self.log.debug(f"new instance data: {new_data}")
+
+                    project_dir = unreal.Paths.project_dir()
+                    render_dir = (f"{project_dir}/Saved/MovieRenders/"
```
mkolar, line 81: For the first production testing this is good enough, but we'll need too move this to settings or to an exposed area inside of UE so it can be tweaked.

openpype/hosts/unreal/plugins/publish/collect_remove_marked.py
BigRoy: For what it's worth. I think the logic in [StandalonePublished CollectRemoveMarked](https://github.com/pypeclub/OpenPype/blob/7017410173647da9a122e0659fda2c7af5593c80/openpype/hosts/standalonepublisher/plugins/publish/collect_remove_marked.py#L19-L21) isn't even safe because it's removing instances whilst iterating the same list?

More info as to why [that is not safe](https://stackoverflow.com/a/6260097/1838864).

openpype/hosts/unreal/api/rendering.py
```diff
@@ -0,0 +1,125 @@
...
+    del executor
+    del queue
+
+
+def _job_finish_callback(job, success):
+    # You can make any edits you want to the editor world here, and the world
+    # will be duplicated when the next render happens. Make sure you undo your
+    # edits in OnQueueFinishedCallback if you don't want to leak state changes
+    # into the editor world.
+    unreal.log("Individual job completed.")
```
BigRoy, line 25: Somehow I feel like we can avoid these globals to exist. E.g. we could define both these functions within the `start_rendering` function?

mkolar (approved):

PR #2920, opened 2022-03-18
General: Configurable add review collector

kalisp:
## Brief description
Adds 'review' family to instance 

## Description
Configuration based on profiles on:
- host name
- task type
- task name
- subset name

Screenshot 2022-03-18 173224

## Additional info
**Still WIP - will be discussed if advanced filtering is useful.**
Hosts should be updated to use this in separate PR, when necessary.

Added mainly for hosts that have functionality to render locally (AE, Harmony). Currently only place to add review to instance based on subset name is in Deadline `submit_publish_job`.

In the future I think that only this collector should be adding review family to simplify review situation, if possible.

## Testing notes:
1. Configure in `Settings:project_settings/global/publish/CollectReviewFamily`
2. Publish in configured hosts

### DIFF
--- a/openpype/plugins/publish/collect_review_family.py
+++ b/openpype/plugins/publish/collect_review_family.py
@@ -1,0 +1,107 @@
+"""
+Requires:
+    none
+
+Provides:
+    instance     -> families ([])
+"""
+import pyblish.api
+import avalon.api
+
+from openpype.lib.plugin_tools import filter_profiles
+
+
+class CollectReviewFamily(pyblish.api.InstancePlugin):
+    """
+        Adds explicitly 'review' to families to trigger ExtractReview
+
+        Uses selection by combination of hosts/families/tasks/subset names via
+        profiles resolution.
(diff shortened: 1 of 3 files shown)


BigRoy:
> In the future I think that only this collector should be adding review family to simplify review situation, if possible.

Unifying across all hosts - yes please!

kalisp:
Currently a bit on hold as this naive approach won't work in a lot of cases. Mainly in cases when reviewable instances are created during extraction process (which is pretty often unfortunately).
This collector would be fine for cases that you know what reviewable instances there will be during collection (I think it should work for AE, PS, Harmony), but it will not sort out problem of having logic in multiple places.
This need much more brainstorming.

mkolar:
@kalisp could we close this and move it's collected information to an issue?

kalisp:
Closing for now, issue created https://github.com/pypeclub/OpenPype/issues/3524 instead.

PR #2922, opened 2022-03-18
Flame: centos related debugging

jakubjezek001:
## Brief description
Flame was behaving differently on centos

## Description
- Only collect selected segments once and pass it to context data
- head and tails could sometimes be string so condition is setting it to 0
- wiretap color policy is not file path but wiretap node path

### DIFF
--- a/openpype/hosts/flame/plugins/publish/collect_timeline_instances.py
+++ b/openpype/hosts/flame/plugins/publish/collect_timeline_instances.py
@@ -34,119 +34,124 @@ class CollectTimelineInstances(pyblish.api.ContextPlugin):
     def process(self, context):
         project = context.data["flameProject"]
         sequence = context.data["flameSequence"]
+        selected_segments = context.data["flameSelectedSegments"]
+        self.log.debug("__ selected_segments: {}".format(selected_segments))
+
         self.otio_timeline = context.data["otioTimeline"]
         self.clips_in_reels = opfapi.get_clips_in_reels(project)
         self.fps = context.data["fps"]
 
         # process all sellected
-        with opfapi.maintained_segment_selection(sequence) as segments:
-            for segment in segments:
-                comment_attributes = self._get_comment_attributes(segment)
-                self.log.debug("_ comment_attributes: {}".format(
-                    pformat(comment_attributes)))
+        for segment in selected_segments:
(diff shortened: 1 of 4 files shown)


openpype/hosts/flame/plugins/publish/collect_timeline_instances.py
```diff
@@ -34,119 +34,124 @@ class CollectTimelineInstances(pyblish.api.ContextPlugin):
...
+            family = marker_data["family"]
+            families = [str(f) for f in marker_data["families"]]
+            families.insert(0, str(family))
+
+            # form label
+            label = asset
+            if asset != clip_name:
+                label += " ({})".format(clip_name)
+            label += " {}".format(subset)
+            label += " {}".format("[" + ", ".join(families) + "]")
```
iLLiCiTiT, line 102:
```suggestion
            label += " {} [{}]".format(subset, ", ".join(families))
```

openpype/hosts/flame/plugins/publish/collect_timeline_instances.py
```diff
@@ -34,119 +34,124 @@ class CollectTimelineInstances(pyblish.api.ContextPlugin):
...
+            ))
+            self.log.debug("_ comment_attributes: {}".format(
+                pformat(comment_attributes)))
+
+            clip_data = opfapi.get_segment_attributes(segment)
+            clip_name = clip_data["segment_name"]
+            self.log.debug("clip_name: {}".format(clip_name))
+
+            # get openpype tag data
+            marker_data = opfapi.get_segment_data_marker(segment)
```
iLLiCiTiT, line 58: I would say that these conditions should be first step of the loop before getting comments and clip data.
```
for segment in selected_segments:
    marker_data = opfapi.get_segment_data_marker(segment)
    self.log.debug("__ marker_data: {}".format(
        pformat(marker_data)))

    if not marker_data:
        continue

    if marker_data.get("id") != "pyblish.avalon.instance":
        continue

    comment_attributes = self._get_comment_attributes(segment)
    ...
```

jakubjezek001: Ah yeah, you are right.

iLLiCiTiT (approved):

PR #2925, opened 2022-03-21
Workfiles: Open published workfiles

iLLiCiTiT:
## Brief description
It is handy to have ability to open published workfile through Workfiles tool in some cases.

## Description
We've agreed that we don't want to allow direct access to published workfile but want add ability to open it. To avoid opening of published workfile directly we decided to create a copy of workfile on user's machine and then open it. This won't close workfiles tool so it is possible to save it right away to a selected context.

Because this way there are created files in app data there should be ability to remove them without user's direct interaction. That is achieved using simple metadata cache mechanism that will store each opened workfile into appdata and store metadata with it's expiration time (1 hour right now). Cache is validated on initialization of workfiles tool which may not be enough so it is also possible to clear cache manually using button in workfiles tool. Next to clear temp is information about used disk storage.

Button label in "publish mode" has label `View` (any better suggestions?). When any published workfile is opened it is copied to cache folder, opened from there using host implementation and `Published` checkbox is unchecked.

### How are selected workfiles?
All subsets with all versions and all representations are loaded for selected asset (it may take some time to load). First filtering is by extensions in representation context which must match one of workfile extensions of host. When task is selected it is also checked for task name in the context. All of them are added to view.

### Screenshot
image

## Changes
- workfiles tool uses `QStandardItemModel` for filling data
- widgets in workfiles tool were separated into more files (it was hard to work in single file)
- widget `FilesWidget` has 2 views and models
    - one is for workarea files which works same way as did before
    - second is for published files which loads paths of representations matching current context

## Testing notes:
1. Open workfiles tool in any host
2. Check `Published` checkbox
3. Open any published workfile
4. `Published` checkbox should be unchecked
5. Now you can store opened file into any context
6. The cached file in appsdata should be removed over time on it's own

Modified implementation of PR https://github.com/pypeclub/OpenPype/pull/2442/files


64qam (approved):

openpype/tools/workfiles/lib.py
```diff
@@ -0,0 +1,214 @@
...
+    minute_in_seconds = 60
+    hour_in_seconds = 60 * minute_in_seconds
+    day_in_seconds = 24 * hour_in_seconds
+
+    def __init__(self):
+        root_dir = appdirs.user_data_dir(
+            "published_workfiles_temp", "openpype"
+        )
+        if not os.path.exists(root_dir):
+            os.makedirs(root_dir)
```
BigRoy, line 57: Why not use an actual temp directory?

openpype/tools/workfiles/lib.py
```diff
@@ -0,0 +1,214 @@
...
+            json.dump({"pid": os.getpid()}, stream)
+
+        try:
+            data = self._get_data()
+            yield data
+            with open(self._metadata_path, "w") as stream:
+                json.dump(data, stream)
+
+        finally:
+            os.remove(self._lock_path)
```
BigRoy, line 127: What's with the lock file. Why do we need to go through this locking mechanism? Would be great to clarify in a docstring or alike. It feels redundant when glancing over the code quickly.

openpype/tools/workfiles/window.py
```diff
@@ -0,0 +1,393 @@
...
+
+from .files_widget import FilesWidget
+from .lib import TempPublishFiles, file_size_to_string
+
+
+class SidePanelWidget(QtWidgets.QWidget):
+    save_clicked = QtCore.Signal()
+    published_workfile_message = (
+        "<b>INFO</b>: Published workfiles you'll opened will be stored in"
+        " temp directory on your machine. Current temp size: <b>{}</b>."
```
BigRoy, line 24: Message has typos `you'll opened` is invalid grammar.

```suggestion
        "<b>INFO</b>: Opened published workfiles will be stored in"
        " temp directory on your machine. Current temp size: <b>{}</b>."
```

Also, UX wise it somehow feels odd how it's in that panel there. Can't think of a better placement directly but just wanted to raise that point.

openpype/tools/utils/delegates.py
```diff
@@ -287,9 +287,6 @@ class PrettyTimeDelegate(QtWidgets.QStyledItemDelegate):
...
     def displayText(self, value, locale):
-
-        if value is None:
-            # Ignore None value
-            return
-
-        return pretty_timestamp(value)
+        if value is not None:
+            return pretty_timestamp(value)
+        return None
```
BigRoy, line 292: With the restructured if statement logic technically this `return None` is redundant too.

openpype/tools/workfiles/lib.py
```diff
@@ -0,0 +1,214 @@
...
+    @property
+    def size(self):
+        if os.path.exists(self.filepath):
+            s = os.stat(self.filepath)
+            return s.st_size
+        return 0
+
+
+class TempPublishFiles(object):
+    """Directory where """
```
BigRoy, line 47: Incomplete docstring

openpype/tools/workfiles/lib.py
```diff
@@ -0,0 +1,214 @@
...
+import time
+import json
+import logging
+import contextlib
+
+import appdirs
+
+
+class TempPublishFilesItem(object):
+    """Object representing on subfolder in app temp files.
```
BigRoy, line 13: Seems like a typo? Not sure what is being referred to with "Object representing on subfolder"?

openpype/tools/workfiles/lib.py
iLLiCiTiT: > Is that just so we could 'ease' "checking" a published workfile version without doing anything with it?

Yes

> why not directly allow "resaving" it to a new version

That'll probably be next PR. In some cases you don't want to save it to workfiles but just look into it.

openpype/tools/workfiles/lib.py
iLLiCiTiT: Making sure that the metadata file is locked for whole process of adding data. Will fill docs.

openpype/tools/workfiles/lib.py
```diff
@@ -0,0 +1,272 @@
...
+import time
+import json
+import logging
+import contextlib
+
+import appdirs
+
+
+class TempPublishFilesItem(object):
+    """Object representing copied workfile in app temp folfer.
```
BigRoy, line 13:
```suggestion
    """Object representing copied workfile in app temp folder.
```

openpype/tools/workfiles/lib.py
```diff
@@ -0,0 +1,272 @@
...
+    Directory is located in appdirs on the machine. Folder contains file
+    with metadata about stored files. Each item in metadata has id, filename
+    and expiration time. When expiration time is higher then current time the
+    item is removed from metadata and it's files are deleted. Files of items
+    are stored in subfolder named by item's id.
+
+    Metadata file can be in theory opened and modified by multiple processes,
+    threads at one time. For those cases is created simple lock file which
+    is created before modification begins and is removed when modification
+    ends. Existince of the file means that it should not be modified by
```
BigRoy, line 58:
```suggestion
    ends. Existence of the file means that it should not be modified by
```

openpype/tools/workfiles/lib.py
iLLiCiTiT: > Also, what happens if you have a different published workfile with the same name that you are opening (e.g. from another project if e.g. project name isn't in the published filename?)

Each opened publish file will create new item with new id so even if you would open same file twice it would create 2 items and store it to a different place. The metadata just keep track of when they were created and when they should be removed.

> And instead opt for simple yet effective - copy the file once and just open it directly. Leave it up to the user to do whatever they'd like at that moment of time.

We want avoid to give option to open the published workfile directly but want give option to just open it and not save in any workdir (at least from what I understood). Give option to save it is in plan, but in different PR.

BigRoy: I understand we'd want to avoid opening the published file for safety reasons.

I'm just worried that the "temp file" logic in place now is worth the amount of code involved to make this work. It means we'll need to maintain the functionality and I wonder whether it provides enough "benefits" over just allowing a user to copy a published workfile back into the workfiles as a new version.

Aside of potentially easier "debugging" of the published workfile I'm personally doubting whether this warrants being this much extra code to maintain. It feels like feature creep. Whereas just saving as another workfile potentially would need only a few extra lines of codes. It could even be a `Loader` in that case! (But I can understand why having it visually in Work Files might be nice).

PR #2927, opened 2022-03-22
General: Removed silo references and terminal splash

iLLiCiTiT:
## Brief description
Silo is not used in OpenPype for a long time and still there are references to it in the code. Also there is terminal splash which is not used anymore and can be removed.

## Description
All references to silo were removed in the code with few deprecated comments related to it. Removed terminal splash was used at first versions of OpenPype 3 but is not used because of huge amount of lines added to stdout. Also reduced Loader tool logic related to roots which is now simplified so library loader logic does not need `LibrarySubsetWidget` anymore.

## Changes
- removed silo references from code
- removed `terminal_splash.py` and `splash.txt` from `openpype/lib`
- Loader tool is not using `registered_root` but `Anatomy.roots` based on current project in context

## Testing notes:
1. Both loader and library loader should work when copying source of version
2. Everything else should work as expected (all removed code was not used)


antirotor:
I will mourn the removal of animated splash even if it wasn't used. R.I.P. ✝️

PR #2933, opened 2022-03-23
Photoshop: New Publisher

kalisp:
## Brief description
Implemented New Publisher for Photoshop

## Description
Should be backward compatible.

## Testing notes:
1. Try end to end publish from empty workfile
2. Try to use New publisher on old workfile with already existing instances


64qam (approved):

mkolar:
This is only waiting for user documentation for the new publisher

ClementHector:
I think there is regression in the code in the path template formatting, the key {subset} is formatted with the value of {family}.
For example .../publish/{family}/{subset}/{@version} gives me .../publish/workfile/workfile/v004

PR #2935, opened 2022-03-23
General: Move plugins register and discover

iLLiCiTiT:
## Brief description
Move plugins register and discover logic from avalon-core repository.

## Description
Discovery and register of plugins was moved to `openpype/pipeline/plugin_discover.py` and is used at all places where register and discover plugins is used. Added register and discover function into create plugins.

## Changes
- implemented `PluginDiscoverContext` - it's object can be used to register and discover plugins
    - added object of `PluginDiscoverContext` for global access
    - discover method returns `DiscoverResult` which holds all discovered, crashed, ignored, duplicated and abstract plugins
- added register/deregister and discover functions into create plugins
- discover of legacy creators and loader will trigger `set_plugin_attributes_from_settings`
    - this was added to `discover_loader_plugins` and `discover_legacy_creator_plugins`
- function `classes_from_module` does not check inheritance by class names

## Testing notes:
- pre/post launch hooks must work as did
    - affected by change of dynamic import functions
- all plugins (create, load, inventory, launcher action, thumbnail) must be discovered properly

Related to PR https://github.com/pypeclub/avalon-core/pull/436


iLLiCiTiT:
> I feel we're once again increasing complexity where we don't need it. Making it much harder to maintain the code over time and have anyone go through the code and grasp what exactly is going o

Current avalon's approach is based on singleton access. It is NOT possible to have registered multiple paths for different contexts which would be now possible. It's not used that way yet but until it's possible it will never be.

> To me personally it makes just as much sense if python_module_tools and plugins_discover would be in a single file in openpype/pipeline/plugins.py. 

Content of `python_module_tools` is usabled (and used) at many different places. It is more like a python module adding some functionality that can be used elsewhere. Not just in discovery of pipeline plugins.

Also I've kept the `plugins` in function names to keep backwards compatibility but the register and discovery is not just for plugins but also for actions and thumbnail resolvers.

> Additionally the new classes (e.g. GlobalDiscover acting as a singleton) provide no docstrings or documentation and feel less python than just storing the registered paths directly.

Will add a docstring. All it does is that it gives ability to use it the same way as was in avalon.

> Also having discover not just return the plug-ins but a "DiscoverResult" just so we can print a report and list "duplicate" plug-ins feels like redundant complexity.

Because it's not used yet doesn't mean it will not be used. Many many times we wanted to show in UI some report that something went wrong and give artist ability to send it to us. That few steps away so current usage just print it out. I can modify the logic that it will not return the result object but plugins by default and will return the result only when requested.

> I'm additionally worried that I might need to alter any scripts/tools I have in place that actually run discover().

We want to add ability to have a context wrapper around the actions and plugins and add them some initialization to add more abilities we're lacking right now. Again not used that way now but it must be prepared to be able do it that way so the PRs are reviewable.
One example is current applying of settings on loaders and creators. That's something what should be done during initialization and both of them should be usable to load/create multiple times using the same object but it is not possible right now.

> Having things be "simple yet effective" is much better, no? We should only make things more complex if we absolutely need to. A pipeline is complex enough by itself.

It's getting more and more complicated to have it "simple". We're hacking a lot because it is done that way and we will never stop hack it until we make small steps heading to make it usable without hacks. It looks simple when you see just those functions but it is much more complicated when we're bending it to make it work as we need. Also there is so much we can't do because of that...

For example easily switch Qt UI for any other UI (detached from the host process).

BigRoy:
> It's not used that way yet but until it's possible it will never be.

I do feel that might be the wrong approach. We should implement the features that we need - not expose functionality that we might never be used. It's much more likely that 'errors' will go by unnoticed.

E.g. likely it won't work with the current implementation either since it still relies on having the global singleton everywhere? In essence that part doesn't behave much different than before?

> Content of python_module_tools is usabled (and used) at many different places. It is more like a python module adding some functionality that can be used elsewhere. Not just in discovery of pipeline plugins.

Agreed that it's relatively generic - but I have the same approach as the above. We might want to only break out the things that we know will be used. Having it separated or exposing it as some sort of "API" or 'use it wherever you see fit' means we should be much less inclined to ever changing the logic/results of the functions in the future. Since we don't know what users/teams/studios might have been using it for elsewhere.

Currently it isn't actually used anywhere else. I'm fine with having it separated. It just makes it appear bigger than I felt it needs to be.

> Because it's not used yet doesn't mean it will not be used. Many many times we wanted to show in UI some report that something went wrong and give artist ability to send it to us. That is few steps away so current usage just print it out. I can modify the logic that it will not return the result object but plugins by default and will return the result only when requested.

I'd much prefer actually to have that feature added - so we know it actually works the way we intend it to work. Additionally we should implement tests that the feature keeps working as we ever intended it to be.

---

In short, it'd help a lot if the decision making process about making these things more complicated is much clearer. Currently I can't accurately and actively feedback the code as to decide whether it does what it should which means I could only say "this errors".

Anyway, thanks for the thorough input.

openpype/pipeline/plugin_discover.py
```diff
@@ -0,0 +1,264 @@
...
+        """Return all currently registered plug-in paths"""
+        # Prohibit editing in-place
+        duplicate = {
+            superclass: paths[:]
+            for superclass, paths in self._registered_plugin_paths.items()
+        }
+        return duplicate
+
+    def deregister_plugin(self, superclass, plugin):
+        """Oppsite of `register_plugin()`"""
```
BigRoy, line 218:
```suggestion
        """Opposite of `register_plugin()`"""
```

openpype/pipeline/plugin_discover.py
```diff
@@ -0,0 +1,264 @@
...
+            self._registered_plugin_paths[superclass].append(path)
+
+    def registered_plugin_paths(self):
+        """Return all currently registered plug-in paths"""
+        # Prohibit editing in-place
+        duplicate = {
+            superclass: paths[:]
+            for superclass, paths in self._registered_plugin_paths.items()
+        }
+        return duplicate
```
BigRoy, line 215: Why is this called `duplicate`? Seems odd for the function name?

Or is it called `duplicate` solely due to it being a copy of the original data (with `paths` also being copied?)

openpype/pipeline/plugin_discover.py
```diff
@@ -0,0 +1,264 @@
...
+        """
+
+        if superclass not in self._registered_plugins:
+            self._registered_plugins[superclass] = list()
+
+        if cls not in self._registered_plugins[superclass]:
+            self._registered_plugins[superclass].append(cls)
+
+    def register_plugin_path(self, superclass, path):
+        """Register a directory of one or more plug-ins
```
BigRoy, line 193: This currently reads to me as if it could be for one or more plug-in types.
```suggestion
        """Register a directory containing plug-ins of type `superclass`
```

openpype/pipeline/plugin_discover.py
```diff
@@ -0,0 +1,264 @@
...
+                    lines.append(10 * "*")
+                    lines.extend(traceback.format_exception(*exc_info_args))
+                    lines.append(10 * "*")
+
+        return "\n".join(lines)
+
+    def print_report(self, only_errors=True, exc_info=True):
+        report = self.get_report(only_errors, exc_info)
+        if report:
+            print(report)
```
BigRoy, line 86: Shouldn't the printed report either be logged instead? Or put behind a verbosity flag?

openpype/pipeline/plugin_discover.py
```diff
@@ -0,0 +1,264 @@
...
+    classes_from_module,
+)
+
+
+class DiscoverResult:
+    """Hold result of publish plugins discovery.
+
+    Stores discovered plugins duplicated plugins and file paths which
+    crashed on execution of file.
+    """
```
BigRoy, line 16: We might want to be explicit about mentioning that this holds the discovered result of a single superclass type.

Also I feel like we don't need to mention "Hold result of.." but can just say: "Result of Plug-ins discovery of a single `superclass` type."

openpype/lib/python_module_tools.py
```diff
@@ -129,20 +130,14 @@ def classes_from_module(superclass, module):
...
-        if not inspect.isclass(obj):
-            continue
-
-        # These are subclassed from nothing, not even `object`
-        if not len(obj.__bases__) > 0:
+        if not inspect.isclass(obj) or obj is superclass:
             continue
 
         # Use string comparison rather than `issubclass`
         # in order to support reloading of this module.
```
BigRoy, line 137: The code below has changed ignoring this comment. So either this comment is incorrect and the original 'issue' is resolved and this comment can removed? Or we reintroduced an old issue?

iLLiCiTiT:
> I do feel that might be the wrong approach. We should implement the features that we need - not expose functionality that we might never be used. It's much more likely that 'errors' will go by unnoticed.

It is "a piece of logic" which contains some functions and attributes that do some logic. All I did is to put them in class so it is visible what are the logic boundries and what functions and attributes that logic has. Which might never be used as I described but it is at least visible what is affecting what.

> I'd much prefer actually to have that feature added - so we know it actually works the way we intend it to work.

We just can't do everything in one run. We have limited time to do features. When we have one simple feature to do but the feature requires to rewrite the core function used everywhere it will never happen because it will take too much time and too much risk. Preparing it slowly is the only way I can see we could do it. At this moment we have ability to prepare it as much close to what we need as possible.

> Currently it isn't actually used anywhere else. I'm fine with having it separated. It just makes it appear bigger than I felt it needs to be.

The functions were there before this PR. It is used in openpype modules, application and getting of openpype version.

One of refactor steps will be to remove all pipeline specific code from `openpype/lib` to `openpype/pipeline`. Main reason is that most of them are needed only for pipeline specific cases but not for application as it is. OpenPype is not just a pipeline anymore but an application that contain pipeline. When we need just handle application version or connection to server, we should not touch the pipeline part at all because it has high ability of application failure.

> Currently I can't accurately and actively feedback the code as to decide whether it does what it should which means I could only say "this errors".

Currently the code does exactly the same like did in avalon. Only the core logic was wrapped into class and some hacks were removed from places where were not needed.

openpype/pipeline/plugin_discover.py
iLLiCiTiT: I've copied it from new publisher. Will change it.

openpype/pipeline/plugin_discover.py
iLLiCiTiT: Copied from avalon...

openpype/lib/python_module_tools.py
iLLiCiTiT: Now you scared me... I've read the comment 5 times and still don't get how I meant it?

BigRoy: Basically the comment states that it is _not_ using `issubclass` for a reason but you have changed the lines below it to use `issubclass`. :) So now the comment makes no sense with the code.

Basically you deviated from exactly what the comment was trying to explain not to do. Either remove the comment and "disallow it still working when reloading that module" OR we preserve the old behavior.

iLLiCiTiT: I've look in history and it was added with `LaunchHooks` but why would using class name instead of issubclass checks help to reload module? Nonsense.

openpype/pipeline/plugin_discover.py
BigRoy: By the way. Should this class maybe be implemented similar to how RenderProducts are implemented?  Would that make it "simpler" to maintain and/or more explicit that it's just a data storage?

openpype/lib/python_module_tools.py
BigRoy: Say you have a file `a` defining class `Plugin`. You have other files, e.g. file `b`, importing it and inheriting from `Plugin`.

Once you reload file `a` then the plugin defined in `b` isn't a subclass (until you'd afterwards also reload `b`) anymore due to it inheriting from the original class before the reload. However by name they'd still match.

With how discover is implemented and thus plugins always reloading this might not seem like a direct issue.. but you would still hit the issue if e.g. plugin file `b` used an intermediate class type, like the Maya ReferenceLoader or anything from a file that doesn't get as dynamically loaded then the issue would be pretty apparent.

I believe however the issue would only occur if you'd reload file `a`. Or to be more descriptive, if you would reload the files from where the Plugin superclasses are loaded.

openpype/__init__.py
```diff
@@ -27,60 +23,17 @@
...
-    ]
-
-    set_plugin_attributes_from_settings(filtered_plugins, superclass)
-
-    return filtered_plugins
-
-
-@import_wrapper
 def install():
     """Install Pype to Avalon."""
```
antirotor, line 27:
```suggestion
    """Install OpenPype to Avalon."""
```

antirotor (changes requested):
It is failing in Houdini 19.0.498:

```
  File "E:\projects\openpype\3x\openpype\tools\loader\app.py", line 307, in _assetschanged
    self._thumbnail_widget.set_thumbnail(asset_ids)
  File "E:\projects\openpype\3x\openpype\tools\loader\widgets.py", line 871, in set_thumbnail
    thumbnail_ent, "thumbnail", self.dbcon
  File "E:\projects\openpype\3x\openpype\pipeline\thumbnail.py", line 17, in get_thumbnail_binary
    resolvers = discover_thumbnail_resolvers()
  File "E:\projects\openpype\3x\openpype\pipeline\thumbnail.py", line 134, in discover_thumbnail_resolvers
    return discover(ThumbnailResolver).plugins
AttributeError: 'list' object has no attribute 'plugins'
```

antirotor:
same problem in Blender 3.0

```
  File "E:\projects\openpype\3x\openpype\tools\loader\app.py", line 307, in _assetschanged
    self._thumbnail_widget.set_thumbnail(asset_ids)
  File "E:\projects\openpype\3x\openpype\tools\loader\widgets.py", line 870, in set_thumbnail
    thumbnail_bin = get_thumbnail_binary(
  File "E:\projects\openpype\3x\openpype\pipeline\thumbnail.py", line 17, in get_thumbnail_binary
    resolvers = discover_thumbnail_resolvers()
  File "E:\projects\openpype\3x\openpype\pipeline\thumbnail.py", line 134, in discover_thumbnail_resolvers
    return discover(ThumbnailResolver).plugins
AttributeError: 'list' object has no attribute 'plugins'
```

iLLiCiTiT:
> It is failing in Houdini 19.0.498:

Should be fixed now

antirotor (approved):
works now, thanks!

PR #2937, opened 2022-03-24
Workfiles tool: Save as published workfiles

iLLiCiTiT:
## Brief description
Give option to save as published workfiles without storing it in temp folder.

## Description
Removed `View` and added 2 new buttons into published workfiles context.
1. `Copy & Open` which will save published workfile to it's workarea
2. `Choose different context` which will allow to change context of workarea where the file will be stored
When button n2 is pressed artist can change context of asset and task and confirm it with `Copy & Open` or cancel the operation with `Cancel`. During this time he can see other workfiles in workarea but the view is disabled.

## Changes
- removed `View` button from workfiles tool
- added `Copy & Open` and `Choose different context` buttons into published workfiles
- Save as dialog does not allow to change extension when published workfile is opened
- the `Published` checkbox is hidden if saving is disabled

### When **Published** is checked
image
### When **Choose different context** is clicked
image

## Testing notes:
1. Open workfiles tool in host
2. Check published
3. Try both `Copy & Open` and `Choose different context` buttons and play with it


64qam (approved):

BigRoy:
The Save As and Save As (to context) feels a bit odd.
What options do we have to just have a single "Save As" button - where it'd default to current context and potentially allow to save elsewhere?

Preferably the labels would be "Open" (with tooltip that says 'Open temporary copy'?) since "View" almost sounds like it's going to open the file in an image viewer - haha.

And have just a single "Save As" button that would always allow to decide where to save. Does it pop-up the "Version Dialog" like regular "Save" for non-workfiles? Could that pop-up dialog have a "Pick Context" button or "Change Destination"? Also because the "Select Context" functionality isn't instantly obvious as to what the user then needs to do.

iLLiCiTiT:
We should also disable to see publishing workfiles when `save` is set to `False`.

mkolar (changes requested):
This came from a bit of misunderstanding from a client request. 

Overall, the functionality is good, but we need to work on the UX and labels here.

1. `View` is misleading and if we want to keep the option of just opening it up temporarily, then it should be explicit `Open Temporarily`. However, I believe this function should actually be removed because it has a huge potential for loss of work. Artist will do it, work on the file, keep pressing ctrl+s, but it won't actually be stored in the project. The backlash in larger teams that we see could be massive. 

2. `Save As` should be more like `Copy & Open`
4. `Save As (to context)` would be better written as `Choose different context`, followed by `Copy & Open` again.

BigRoy:
> View is misleading and if we want to keep the option of just opening it up temporarily, then it should be explicit Open Temporarily. However, I believe this function should actually be removed because it has a huge potential for loss of work. Artist will do it, work on the file, keep pressing ctrl+s, but it won't actually be stored in the project. The backlash in larger teams that we see could be massive.

Sounds like removing the "open temporary file" might the best bet and would reduce quite some code length. :)

mkolar:
>And have just a single "Save As" button that would always allow to decide where to save. Does it pop-up the "Version Dialog" like regular "Save" for non-workfiles? Could that pop-up dialog have a "Pick Context" button or "Change Destination"? Also because the "Select Context" functionality isn't instantly obvious as to what the user then needs to do.

I agree that would be a better option, but we need to make a new widget for that which combines context picker and the file name widget. We'll first try with a guide text over the file list that will tell the use `choose context on the left`.

We'll see what users say and change accordingly if it's not received well... **`tested on people`**

mkolar (approved):

BigRoy:
Some UX notes about this implementation for "Choose different context:"

When clicking `Choose different context` the "Cancel" button also gets disabled when not having a Task selected which feels odd. I think cancel should always be enabled. What do you think?

Extra bonus if pressing "escape" button could also cancel. :)

---

Also, I think the "please choose context on the left" message should also state 'what file we're coping from where to where'

```
Choose context on the left to save:

X

to

[no selection]
```

And once selected:

```
Choose context on the left to save:

X

to

Y
```

PR #2938, opened 2022-03-24
Settings: Save default settings sorted

iLLiCiTiT:
## Brief description
Use `sort_keys` when dumpings default settings values.

## Description
Using sorted keys prevent to randomly change order of settings values. Settings were resaved.

## Testing notes:
1. Settings should be loaded fine.

### DIFF
--- a/openpype/modules/base.py
+++ b/openpype/modules/base.py
@@ -1346,7 +1346,7 @@ class JsonFilesSettingsDef(ModuleSettingsDef):
             os.makedirs(dirpath)
 
         with open(path, "w") as file_stream:
-            json.dump(data, file_stream, indent=4)
+            json.dump(data, file_stream, sort_keys=True, indent=4)
 
     def save_system_defaults(self, data):
         """Save default system settings values.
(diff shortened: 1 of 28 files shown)


BigRoy:
This would at least re-order the versions of applications so that latest versions are always listed at the bottom, correct? Is that preferred? And are there any other areas where the ordering might actually influence behavior differently than before? 

It'd also restructure e.g. the JSON data which users can manually "type" in e.g. Maya Scripts Menu Defintion `project_settings/maya/scriptsmenu/definition` - would we maybe want to exclude auto sorting those text based entries so what the user types doesn't get adjusted?

The same would happen to any "environments" the user typed for applications? For example. I tend to somewhat sort the environment variables keys based on what they are related to, for example the entries in `applications/maya/environment` - would those now also auto-sort?

iLLiCiTiT:
> would we maybe want to exclude auto sorting those text based entries so what the user types doesn't get adjusted?

Probably yes, but in that case it should not be done this way... Closing

PR #2939, opened 2022-03-24
General: Python specific vendor paths on env injection

iLLiCiTiT:
## Brief description
Move adding of python version specific vendor paths from prelaunch hook to base logic of application launch.

## Description
The python version specific vendor modules are not available on farm because in that case is not used prelaunch hook which would add them.

## Changes
- removed prelaunch hook adding python 2 vendors into PYTHONPATH
- added function that adds python version specific paths into PYTHONPATH so environments injection on farm can get them

## Testing notes:
1. run older maya publishing on deadline
2. should not crash due to missing python modules
3. publishing job should finish too

Resolves https://github.com/pypeclub/OpenPype/issues/2940


antirotor (approved):
It is working for me as far as I was able to test. Thanks.

PR #2942, opened 2022-03-24
TVPaint: Extractor to convert PNG into EXR

iLLiCiTiT:
## Brief description
TVPaint extractor is creating PNG sequences but it may be better to store them as exr files.

## Description
Since the TVPaint extraction is based on compositing in python we can't extract EXRs directly from TVPaint. Extractor is using oiiotool to convert pngs into exr. Plugin is disabled by default. Has ability to choose output exr compression and if should replace input pngs or keep them. If pngs are replaced then their files are removed during cleanup.

Source [PR from Pype 2](https://github.com/pypeclub/OpenPype/pull/2048).

## Notes
Right now there are hardcoded color conversions `--colorconvert sRGB linear`. Not sure if we should do it or if we should add some kind of settings for this? If yes, what kind of settings?

## Testing notes:
1. Enable the plugin in project settings
2. Publish from TVPaint

### DIFF
--- a/openpype/hosts/tvpaint/plugins/publish/extract_convert_to_exr.py
+++ b/openpype/hosts/tvpaint/plugins/publish/extract_convert_to_exr.py
@@ -22,10 +22,12 @@ class ExtractConvertToEXR(pyblish.api.InstancePlugin):
     hosts = ["tvpaint"]
     families = ["render"]
 
-    active = False
+    enabled = False
 
+    # Replace source PNG files or just add
     replace_pngs = True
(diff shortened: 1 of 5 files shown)


tokejepsen:
> Right now there are hardcoded color conversions --colorconvert sRGB linear. Not sure if we should do it or if we should add some kind of settings for this? If yes, what kind of settings?

Think hardcoded is fine. Not a lot of options in TVP for this and most production dont deal this colorspace in TVP.

64qam (approved):

PR #2950, opened 2022-03-24
General: Smaller fixes and typos

iLLiCiTiT:
## Brief description
Fix of few smaller issues.

## Description
Fixed import of attribute definitions in tray publisher plugin. Fixed typo of attribute `_first_show`. Log information about not find icon and also log that passed icon name is not exact match of used icon name (prefix from qtawesome was added). Changed `Name:` label to `Variant:` in creator dialog window.

## Testing notes:
- Tray publisher should show workfiles creator
- log viewer should not crash
- icons should be visible in tools
- creator dialog in new publisher should show `Variant:` instead of `Name:`

### DIFF
--- a/openpype/hosts/traypublisher/plugins/publish/validate_workfile.py
+++ b/openpype/hosts/traypublisher/plugins/publish/validate_workfile.py
@@ -6,7 +6,7 @@ from openpype.pipeline import PublishValidationError
 class ValidateWorkfilePath(pyblish.api.InstancePlugin):
     """Validate existence of workfile instance existence."""
 
-    label = "Collect Workfile"
+    label = "Validate Workfile"
     order = pyblish.api.ValidatorOrder - 0.49
     families = ["workfile"]
     hosts = ["traypublisher"]
--- a/openpype/tools/utils/lib.py
+++ b/openpype/tools/utils/lib.py
@@ -126,7 +126,7 @@ def get_qta_icon_by_name_and_color(icon_name, icon_color):
         log.info("Didn't find icon \"{}\"".format(icon_name))
 
     elif used_variant != icon_name:
-        log.info("Icon \"{}\" was not found \"{}\" is used instead".format(
+        log.debug("Icon \"{}\" was not found \"{}\" is used instead".format(
             icon_name, used_variant
         ))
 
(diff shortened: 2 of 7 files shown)


openpype/tools/utils/lib.py
```diff
@@ -111,13 +113,23 @@ def get_qta_icon_by_name_and_color(icon_name, icon_color):
...
         except Exception:
             pass
 
+    if used_variant is None:
+        log.info("Didn't find icon \"{}\"".format(icon_name))
+
+    elif used_variant != icon_name:
+        log.info("Icon \"{}\" was not found \"{}\" is used instead".format(
+            icon_name, used_variant
+        ))
```
BigRoy, line 131: Wouldn't this happen quite often? Anywhere where not explicitly `fa.` is prepended to the icon name for example?

E.g. [here](https://github.com/pypeclub/OpenPype/blob/57feaee44b8bff1eb7156b25bad70c978b715fec/openpype/tools/utils/lib.py#L155-L156)?

Should this be `log.debug`? And still be prepared for quite some logging?

iLLiCiTiT: It is called once for each icon.

openpype/hosts/traypublisher/plugins/publish/validate_workfile.py
```diff
@@ -6,7 +6,7 @@
 class ValidateWorkfilePath(pyblish.api.InstancePlugin):
     """Validate existence of workfile instance existence."""
 
-    label = "Collect Workfile"
+    label = "Validate Workfile"
     order = pyblish.api.ValidatorOrder - 0.49
     families = ["workfile"]
     hosts = ["traypublisher"]
```
BigRoy, line 12: Noting this here even though it's somewhat unrelated to this PR.

This particular plug-in checks `sourceFilepath` in data. A quick query over the code base shows that ONLY this validator and collector use that particular `instance.data`. It doesn't appear to be used for any output, etc.

Could we somehow make to this more generic so that there's no need for this additional data key for this single use case but we could do e.g. a validator over all `instance.data[representation]` and its files? (even though most representations get added during Extractors after Validator).

Validating our input content exists seems like something we (need to do or are doing) elsewhere too so I imagine we might want to unify where possible?

iLLiCiTiT: The whole logic of traypublisher will be more generic, but not in this PR.

kalisp (approved):

PR #2952, opened 2022-03-25
Maya - added transparency into review creator

kalisp:
## Brief description
Transparency configuration was missing from `CreateReview`

## Testing notes:
1. Create camera in Maya scene and Create Review family
2. check created review group, there should be new item to select transparency
3. publish and check that selected transparency was applied

### DIFF
--- a/openpype/hosts/maya/plugins/create/create_review.py
+++ b/openpype/hosts/maya/plugins/create/create_review.py
@@ -15,6 +15,14 @@ class CreateReview(plugin.Creator):
     keepImages = False
     isolate = False
     imagePlane = True
+    transparency = [
+        "preset",
+        "simple",
+        "object sorting",
+        "weighted average",
+        "depth peeling",
+        "alpha cut"
+    ]
 
     def __init__(self, *args, **kwargs):
         super(CreateReview, self).__init__(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


kalisp:
Should `CreateReview` class attributes added to Settings? (`keepImages`, `isolate`, 'imagePlane`, `transparency`)

antirotor:
> Should `CreateReview` class attributes added to Settings? (`keepImages`, `isolate`, `imagePlane`, `transparency`)

I don't think so, those are set usually per scene. Except maybe transparency. Any thought about that @tokejepsen ?

tokejepsen:
> I don't think so, those are set usually per scene. Except maybe transparency. Any thought about that @tokejepsen ?

Yeah, we set it in the scene usually, since its dependent on the project/shot/task. I guess eventually if you could have project/shot/task settings, then we might configure it on a larger set of tasks, but for now setting it in the scene works well.

antirotor (approved):
works for me!

m-u-r-p-h-y:
another example of arbitrary attributes cluttering the code and the UI . . .

tokejepsen:
> another example of arbitrary attributes cluttering the code and the UI . . .

What was the solution to this?

PR #2956, opened 2022-03-28
Settings: Conditional dictionary avoid invalid logs

iLLiCiTiT:
## Brief description
Conditional dictionary which has children with same key but different types does not cause warning logs when loading settings anymore.

## Description
Added option to not log invalid value type on loading of settings which is by default enabled but (currently only) dict-conditional may disable it if is updating values of children which may have different type which is not importatnt.

## Changes
- added argument to `update_default_values`, `update_studio_values` and `update_project_values` which will define if invalid value type is logged out
- when this argument is passed and value has invalid type it is not logged out
- conditional dictionary changes the enabled logging based on enum value from the values

## Testing notes:
1. Run settings UI
2. Check logs in console if there are any warnings

### DIFF
--- a/openpype/settings/entities/base_entity.py
+++ b/openpype/settings/entities/base_entity.py
@@ -173,6 +173,10 @@ class BaseItemEntity(BaseEntity):
         # Entity has set `_project_override_value` (is not NOT_SET)
         self.had_project_override = False
 
+        self._default_log_invalid_types = True
+        self._studio_log_invalid_types = True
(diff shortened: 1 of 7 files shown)


kalisp (approved):
Removed `project_settings/nuke/publish/ExtractReviewDataMov/outputs/baking/reformat_node_config/0/value Got invalid value type for default values. Expected types: (<class 'bool'>,) | Got Type: <class 'str'> | Value: "to format" `. 
Save and restart works.

Issue #2913, opened 2022-03-18
Nuke: Review with Slate doesn't have source timecode


jrsndl:
**Describe the bug**
Reviews with Nuke slate do not have source timecode

**To Reproduce**
1. Setup mov review
2. Publish a version from Nuke
3. The review mov has a timecode
4. Change the review to have a slate
5. Publish a version from Nuke with Nuke slate
6. The review mov has a slate, but no timecode

**Expected behavior**
For one frame slate at the start, the source timecode should be present, with value of start_frame timecode - 1
For frame range 1001-1010 where frame 1001 has stc 01:00:00:00 @ 25 fps, slate at frame 1000 should have stc 00:59:59:24

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Host: Nuke 12.1v3

jrsndl:
Yeah. I will do a PR with attempt to fix it.

PR #2962, opened 2022-03-28
Bugfix/fix slate timecode and rendering range

maxpareschi:
## Brief description
Timecode was not correctly added when slate was present, nor was correctly offset when skip handle frames option was selected in the ExtractReview tags. Review duration was off for first time renders with slate but ok without.

## Description
Timecodes in review movs (from Nuke) were not correctly updated to cater for extra slate frame on head.
Timecode should shift -1 frame from starting timecode of extracted representation:
This means that each representation (and its frame length, especially in the case of "Skip handle frames" in outupt videos) should always be:
- rendered (the data mov) with the correct timecode at head and the correct length if slate is present
- checked against new lengths in case of trimming
- have a -1 frame shift in timecode in case of slate being present
- updated with new timecodes per representation before concat/merge

This needed additions in _OpenPype/lib/editorial.py_ with function to deal with timecode and frames which are now available in `openpype.lib` package, plus Logic fixes in the slate extraction, extract review and integrate new.

This should fix problems mentioned in this issue:
[Nuke: Review with Slate doesn't have source timecode #2913 ](https://github.com/pypeclub/OpenPype/issues/2913)
Still needs silent audio in slate as per another PR though

## Testing notes:
1. Render out any asset from nuke
2. create ExtractReview profiles with slate
3. publish and check timecode in a variety of scenarios


mkolar:
Even though this seems correct in principle, I think we'll have to keep it stale for a bit longer, merge #2898 and then revisit this, so we don't get too mixed up in it.

PR #2966, opened 2022-03-29
SiteSync: Download all workfile inputs

kalisp:
## Brief description
Each workfile collects loaded containers and stores them in DB. This feature allows when artists downloads workfile for local work to all these reference files be marked for downloading also.

## Additional info
Now it seems that referenced item is only one step away, no 'referencing of referenced', function was implemented to actually support this. Probably not needed for now.

## Testing notes:
1. In non local setup (`active_site` == studio) publish workfile that contains loaded item
2. Turn on local setup (`active_site` == local site) in Local Setting
3. Try to download previously published workfile via Loader
4. Local site skeleton should be added in DB not only to workfile, but for all loaded containers (this should be visible in Site Queue list)

### DIFF
--- a/openpype/lib/avalon_context.py
+++ b/openpype/lib/avalon_context.py
@@ -1971,3 +1971,123 @@ def get_last_workfile(
         return os.path.normpath(os.path.join(workdir, filename))
 
     return filename
+
+
+@with_avalon
+def get_linked_ids_for_representations(project, repre_ids, dbcon=None,
+                                       link_type=None, max_depth=0):
+    """Returns list of linked ids of particular type (if provided).
+
+    Goes from representations to version, back to representations
+    Args:
+        project (str)
+        repre_ids (list) or (ObjectId)
+        dbcon (avalon.mongodb.AvalonMongoDB, optional): Avalon Mongo connection
+            with Session.
+        link_type (str): ['reference', '..]
+        max_depth (int): limit how many levels of recursion
+    Returns:
+        (list) of ObjectId - linked representations
+    """
+    if not dbcon:
+        log.debug("Using `avalon.io` for query.")
+        dbcon = avalon.io
+        # Make sure is installed
+        dbcon.install()
+
+    if dbcon.Session["AVALON_PROJECT"] != project:
+        dbcon.Session["AVALON_PROJECT"] = project
+
+    if not isinstance(repre_ids, list):
+        repre_ids = [repre_ids]
+
+    versions = avalon.io.find(
+        {
+            "_id": {"$in": repre_ids},
+            "type": "representation"
+        },
+        projection={"parent": True}
+    )
+    version_ids = [version["parent"] for version in versions]
+
+    graph_lookup = {
+        "from": project,
+        "startWith": "$data.inputLinks.id",
+        "connectFromField": "data.inputLinks.id",
+        "connectToField": "_id",
+        "as": "outputs_recursive",
+        "depthField": "depth"
+    }
+    if max_depth != 0:
+        # We offset by -1 since 0 basically means no recursion
+        # but the recursion only happens after the initial lookup
+        # for outputs.
+        graph_lookup["maxDepth"] = max_depth - 1
+
+    match = {
+        "_id": {"$in": version_ids},
+        "type": "version"
+    }
+
+    pipeline_ = [
+        # Match
+        {"$match": match},
+        # Recursive graph lookup for inputs
+        {"$graphLookup": graph_lookup}
+    ]
+
+    result = dbcon.aggregate(pipeline_)
+    referenced_version_ids = _process_referenced_pipeline_result(result,
+                                                                 link_type)
+
+    representations = avalon.io.find(
+        {
+            "parent": {"$in": list(referenced_version_ids)},
+            "type": "representation"
+        },
+        projection={"_id": True}
+    )
+    ref_ids = {representation["_id"] for representation in representations}
+    return list(ref_ids)
+
+
+def _process_referenced_pipeline_result(result, link_type):
+    """Filters result from pipeline for particular link_type.
+
+    Pipeline cannot use link_type directly in a query.
+    Returns:
+        (list)
+    """
+    referenced_version_ids = set()
+    correctly_linked_ids = set()
+    for item in result:
+        correctly_linked_ids = _filter_input_links(item["data"]["inputLinks"],
+                                                   link_type,
+                                                   correctly_linked_ids)
+
+        # outputs_recursive in random order, sort by _id
+        outputs_recursive = sorted(item.get("outputs_recursive", []),
+                                   key=lambda d: d["_id"])
+        # go from oldest to newest
+        # only older _id can reference another newer _id
+        for output in outputs_recursive[::-1]:
+            if output["_id"] not in correctly_linked_ids:  # leaf
+                continue
+
+            correctly_linked_ids = _filter_input_links(
+                output["data"].get("inputLinks", []),
+                link_type,
+                correctly_linked_ids)
+
+            referenced_version_ids.add(output["_id"])
+
+    return referenced_version_ids
+
+
+def _filter_input_links(input_links, link_type, correctly_linked_ids):
+    for input_link in input_links:
+        if not link_type or input_link["type"] == link_type:
+            correctly_linked_ids.add(input_link.get("id") or
+                                     input_link.get("_id"))  # legacy
+
+    return correctly_linked_ids
--- a/openpype/modules/sync_server/sync_server_module.py
+++ b/openpype/modules/sync_server/sync_server_module.py
@@ -23,7 +23,7 @@ from openpype.settings.lib import (
 from .providers.local_drive import LocalDriveHandler
 from .providers import lib
 
-from .utils import time_function, SyncStatus
+from .utils import time_function, SyncStatus, SiteAlreadyPresentError
 
 
 log = PypeLogger().get_logger("SyncServer")
@@ -129,7 +129,7 @@ class SyncServerModule(OpenPypeModule, ITrayModule):
 
     """ Start of Public API """
     def add_site(self, collection, representation_id, site_name=None,
-                 force=False):
+                 force=False, force_only_broken=False):
         """
             Adds new site to representation to be synced.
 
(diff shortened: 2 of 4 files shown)


openpype/plugins/load/add_site.py
```diff
@@ -12,21 +22,61 @@ class AddSyncSite(load.LoaderPlugin):
...
+
+        add_ids = set(add_ids)
+        self.log.info("Add to repre_ids {}".format(add_ids))
+        is_main = True
+        for add_repre_id in add_ids:
+            self.add_site_to_representation(project_name,
+                                            add_repre_id,
+                                            data["site_name"],
+                                            is_main)
+            is_main = False
```
BigRoy, line 57: Are you setting it only on the first entry because you're expecting the first entry to be the initial `repre_id`? If so, then your `is_main` logic is wrong because you're turning it into a `set` to ensure unique entries which will mangle up the order.

Unlike `dict` even in Python 3 sets are unsorted. See also [here](https://stackoverflow.com/a/45589593/1838864).

Here's an alternative. **Pseudocode!**
```suggestion
        repre_id = data["_id"]
        self.add_site_to_representation(project_name,
                                        repre_id,
                                        data["site_name"],
                                        is_main=True)

        if family == "workfile":
            links = get_linked_ids_for_representations(project_name,
                                                       add_ids,
                                                       link_type="reference")
            for link_repre_id in links:
                self.add_site_to_representation(project_name,
                                                link_repre_id,
                                                data["site_name"],
                                                is_main=False)
```

openpype/lib/avalon_context.py
```diff
@@ -1971,3 +1971,123 @@ def get_last_workfile(
...
+                                                                 link_type)
+
+    representations = avalon.io.find(
+        {
+            "parent": {"$in": list(referenced_version_ids)},
+            "type": "representation"
+        },
+        projection={"_id": True}
+    )
+    ref_ids = {representation["_id"] for representation in representations}
```
BigRoy, line 2050: I believe this could also be:
```suggestion
    ref_ids = io.distinct("_id", 
        filter={
            "parent": {"$in": list(referenced_version_ids)},
            "type": "representation"
        }
    )
```
_Untested, so please consider as **pseudocode**_

openpype/lib/avalon_context.py
```diff
@@ -1971,3 +1971,123 @@ def get_last_workfile(
...
+    for item in result:
+        correctly_linked_ids = _filter_input_links(item["data"]["inputLinks"],
+                                                   link_type,
+                                                   correctly_linked_ids)
+
+        # outputs_recursive in random order, sort by _id
+        outputs_recursive = sorted(item.get("outputs_recursive", []),
+                                   key=lambda d: d["_id"])
+        # go from oldest to newest
+        # only older _id can reference another newer _id
```
BigRoy, line 2072: Are we sure about this? As I refactored the Integrator I'm pretty sure existing versions and representations can be "replaced" or overwritten with new content and thus technically an earlier generated `id` might have newer content than another id.

Also, are we sure [the IDs are sortable like that as if they are timestamps?](https://stackoverflow.com/a/12100023/1838864)

BigRoy (commented):
Lovely!
Some quick comments as I glanced over the current code.

openpype/lib/avalon_context.py
mkolar: >Also, are we sure [the IDs are sortable like that as if they are timestamps?](https://stackoverflow.com/a/12100023/1838864)

I'm pretty sure that mongoIDs are timestamps, however, this will not be the case in the v4 probably, so we should keep it in mind here

kalisp: I switched that 'depth' that comes from pipeline, that should be deterministic. (Not solving v4 though.)

openpype/lib/avalon_context.py
```diff
@@ -1971,3 +1971,123 @@ def get_last_workfile(
...
+        # Make sure is installed
+        dbcon.install()
+
+    if dbcon.Session["AVALON_PROJECT"] != project:
+        dbcon.Session["AVALON_PROJECT"] = project
+
+    if not isinstance(repre_ids, list):
+        repre_ids = [repre_ids]
+
+    versions = avalon.io.find(
```
iLLiCiTiT, line 2004:
```suggestion
    versions = dbcon.find(
```

openpype/lib/avalon_context.py
```diff
@@ -1971,3 +1971,123 @@ def get_last_workfile(
...
+        {"$match": match},
+        # Recursive graph lookup for inputs
+        {"$graphLookup": graph_lookup}
+    ]
+
+    result = dbcon.aggregate(pipeline_)
+    referenced_version_ids = _process_referenced_pipeline_result(result,
+                                                                 link_type)
+
+    ref_ids = avalon.io.distinct(
```
iLLiCiTiT, line 2043:
```suggestion
    ref_ids = dbcon.distinct(
```

openpype/lib/avalon_context.py
```diff
@@ -1971,3 +1971,123 @@ def get_last_workfile(
         return os.path.normpath(os.path.join(workdir, filename))
 
     return filename
+
+
+@with_avalon
+def get_linked_ids_for_representations(project, repre_ids, dbcon=None,
+                                       link_type=None, max_depth=0):
```
iLLiCiTiT, line 1978: Just more specific arg name
```suggestion
def get_linked_ids_for_representations(project_name, repre_ids, dbcon=None,
                                       link_type=None, max_depth=0):
```

openpype/lib/avalon_context.py
```diff
@@ -1971,3 +1971,123 @@ def get_last_workfile(
...
+        repre_ids = [repre_ids]
+
+    versions = avalon.io.find(
+        {
+            "_id": {"$in": repre_ids},
+            "type": "representation"
+        },
+        projection={"parent": True}
+    )
+    version_ids = [version["parent"] for version in versions]
```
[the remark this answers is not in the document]
BigRoy: Actually, instead of doing that. We could also change the query itself to:
```python
version_id = `io.distinct("parent", {
        "_id": {"$in": repre_ids},
        "type": "representation"
    })
```
Especially if we're just projecting to `parent` only, don't use any other data and only want to know the unique parent ids.


At first I thought it felt wrong storing `version["parent"]` into a variable named `version_ids` is confusing. That doesn't sound like version ids but subset ids. No? But then I noticed the query above. It's not actually querying `versions` but `representations`. So this would actually be `[repre["parent"] for repre in representations]` etc.

Anyway, the `distinct` example solves it either way. :)

openpype/lib/avalon_context.py
```diff
@@ -1971,3 +1971,118 @@ def get_last_workfile(
...
+        (list) of ObjectId - linked representations
+    """
+    if not dbcon:
+        log.debug("Using `avalon.io` for query.")
+        dbcon = avalon.io
+        # Make sure is installed
+        dbcon.install()
+
+    if dbcon.Session["AVALON_PROJECT"] != project_name:
+        dbcon.Session["AVALON_PROJECT"] = project_name
```
BigRoy, line 1999: If we're forcing it to be `project_name` anyway, why not just make it:
```suggestion
    dbcon.Session["AVALON_PROJECT"] = project_name
```

Or do we expect a high cost involved with updating the Session?

m-u-r-p-h-y (approved):
tested with animation file referencing model abc

all files synced

openpype/plugins/load/add_site.py
```diff
@@ -12,21 +22,58 @@ class AddSyncSite(load.LoaderPlugin):
...
+                                   site_name, is_main):
+        """Adds new site to representation_id, resets if exists
+
+        Args:
+            project_name (str)
+            representation_id (ObjectId):
+            site_name (str)
+            is_main (bool): true for really downloaded, false for references,
+                force redownload main file always, for references only if
+                broken
```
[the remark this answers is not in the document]
kalisp: Changed.

openpype/plugins/load/add_site.py
```diff
@@ -12,21 +22,60 @@ class AddSyncSite(load.LoaderPlugin):
...
+
+        Args:
+            project_name (str)
+            representation_id (ObjectId):
+            site_name (str)
+            is_main (bool): true for main representation, false for referenced
+                loaded repres. Drives if site state should be reset.
+                (it should be for main, not for referenced as they might be
+                shared from multiple workfiles). In necessary cases, referenced
+                repres should be reset (re-downloaded) manually.
```
BigRoy, line 70: Thanks - it is more readable. But since I don't have too much knowledge about site sync and this area of the code base I must admit that it's still rather hard to understand exactly what is happening.

In essence for a particular project's representation and a given site name this would make sure all files are set up to sync at the given site.

When `force` is set (for `is_main=True`) then the files will be redownloaded even if they already existed locally?
When `force_only_broken` is set, then what happens? Does it only redownload if the file locally has an invalid checksum, or what is considered "broken"?

openpype/plugins/load/add_site.py
```diff
@@ -12,21 +22,60 @@ class AddSyncSite(load.LoaderPlugin):
...
+        """
+        try:
+            self.sync_server.add_site(project_name, representation_id,
+                                      site_name,
+                                      force=is_main,
+                                      force_only_broken=not is_main)
```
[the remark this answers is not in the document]
kalisp: Yeah, its a bit cumbersome to have there two similar arguments. I cleaned up them a bit.
Developer should only use 'force' argument and catch exception if necessary.

openpype/modules/sync_server/sync_server_module.py
```diff
@@ -1418,14 +1423,16 @@ def reset_site_on_representation(self, collection, representation_id,
             pause (bool or None): if True - pause, False - unpause
             force (bool): hard reset - currently only for add_site
 
-        Returns:
-            throws ValueError
+        Throws:
```
iLLiCiTiT, line 1426:
```suggestion
        Raises:
```

openpype/plugins/load/add_site.py
```diff
@@ -1,9 +1,19 @@
 from openpype.modules import ModulesManager
 from openpype.pipeline import load
+from openpype.lib.avalon_context import get_linked_ids_for_representations
+from openpype.modules.sync_server.utils import SiteAlreadyPresentError
```
iLLiCiTiT, line 4: Not change for this PR but this plugin should be moved to sync server module.

m-u-r-p-h-y (approved):
Tested again after code refactoring. 

improved "Validate files on active site" in Sync Queue to check for nonexistent files and fix database record

Syncing works as expected.

Will continue in improving the user experience with SiteSync.

PR #2967, opened 2022-03-29
Hero versions: Use custom templates

iLLiCiTiT:
## Brief description
Add ability to use more than default `hero` template for hero version publishing.

## Description
Added ability to define different template key from anatomy templates, similar to how it's used in `IntegrateAssetNew` using profile filters based on host, task name, task type and family.

## Changes
- added template key profile into Integrate Hero version setttings
- modified logic of hero version integrator to use template key based on settings
- removed unused color template from settings

## Testing notes:
1. Create new template in project anatomy
    - with `folder`, `file` and `path` that match hero template keys (without `{version}`)
    - different than default `hero` template values!
2. Use the template key as template key in hero version integrator settings (`project_settings/global/publish/IntegrateHeroVersion/template_name_profiles`)
3. Enable hero version publishing and publish something that will be published as hero version
4. Check if right template was used

Resolves https://github.com/pypeclub/OpenPype/issues/2865#event-6324535346


kalisp (approved):
Works

antirotor (approved):

PR #2968, opened 2022-03-29
Maya: Fix getting non-active model panel.

kalisp:
## Brief description
When capturing multiple cameras with image planes that have file sequences playing, only the active (first) camera will play through the file sequence.

## Additional info
Reimplementation of https://github.com/pypeclub/OpenPype/pull/1751 for OP

## Testing Notes:
1. Prepare and image sequence with frame numbers.
2. Load image sequence into Maya as image plane on two cameras (dont use any default cameras).
3. Setup two reviews for each camera.
4. Publish.
5. Validate both reviews have the image sequence playing correctly.


BigRoy:
This should be done through a context manager instead to ensure that if any errors occur in between the Model Panel is reverted back to its original state.

Also, I'd love to have a simple reproducable "test" for this issue since I recall having solved this before but simpler.

---

Also, would it be sufficient to use `capture` with `off_screen=False`? Since that'd actually playblast to an independent model panel anyway and it'd save you quite a lot of logic in this area?

mkolar:
>Also, would it be sufficient to use capture with off_screen=False? Since that'd actually playblast to an independent model panel anyway and it'd save you quite a lot of logic in this area?

I'd say this is certainly worth a shot

tokejepsen:
> Also, would it be sufficient to use capture with off_screen=False? Since that'd actually playblast to an independent model panel anyway and it'd save you quite a lot of logic in this area?

Would this not limit the resolution? Dont think you can output 1920x1080 on a monitor set to 1920x1080 unless you render off screen?

BigRoy:
@tokejepsen good point actually. I recall that being the case in the early Maya days but I am not actually sure that is still an issue Today. Would be good to double check yes.

kalisp:
There is a bug in a publish process limiting exporting 2 reviews from same workfile, until it is resolved, this couldn't be tested properly.

tokejepsen:
> There is a bug in a publish process limiting exporting 2 reviews from same workfile, until it is resolved, this couldn't be tested properly.

Got a link to the issue?

kalisp:
It should be https://github.com/pypeclub/OpenPype/issues/3034

antirotor:
We should move forward with this or close it. Since #3034 is closed, this can be tested if it is still the problem (and I believe that it is)

antirotor:
What is status of this @kalisp? Is it ready for review or still in draft.

BigRoy:
I'd say this might need updating after this PR by @tokejepsen to instead use the `panel` data which the tweaked plugin there collects: https://github.com/ynput/OpenPype/pull/4549

LiborBatek (changes requested):
Followed the testing steps and having 2 camera reviews with 2 imagePlanes gives me errors when trying to publish em. Failing on Burnins too?

image

image

image

BigRoy:
> Followed the testing steps and having 2 camera reviews with 2 imagePlanes gives me errors when trying to publish em. Failing on Burnins too?

Looking at the errors those acutally don't seem to be specifically related to this PR as far as I can see.

tokejepsen:
@LiborBatek could you check your settings for `project_settings/global/publish/ExtractBurnin/profiles/0/burnins`. Looks like you have `comment` to burn into the review but the `Collect Instance Comment` cannot find any comments.
Unrelated to this PR but should get you to a state where you can test this properly.

LiborBatek:
Ok, thx for idea so I have reverted to defaults in terms of Burnins but it didnt help :(

here are my settings:
image

and again the error I ve got:

image

The ImagePlane has connected simple Jpeg sequence which has more frames than my output frame range.

tokejepsen:
@LiborBatek the issue coming from this line; https://github.com/ynput/OpenPype/blob/develop/openpype/plugins/publish/extract_burnin.py#L489 which means the `comment` on the instance is not collected. Could you check the plugin `Collect Instance Comment` for any issues?

tokejepsen:
@LiborBatek are you experiencing this issue when testing for other PRs?
Maybe try with just the review instance in the scene, to see if its related to this PR.

LiborBatek:
It also happens when creating fresh new Camera and ```Review``` Instance in the scene. Also even when no ImagePlane is used...

running this PR branch...

LiborBatek:
It seems that it happens in latest ```develop``` branch but just in that particular shot...

LiborBatek (approved):
All works as expected. I have published 2 reviews from diff cameras and both having animated ImagePlane so all works ok.

image

image

image

LiborBatek:
So to summarize that all the issues were caused by instances created before when tested ```New Publisher``` in that particular shot...now all fine working!

@BigRoy thx for clarifying it

openpype/hosts/maya/plugins/publish/extract_playblast.py
```diff
@@ -112,6 +112,15 @@ def process(self, instance):
...
 
+        # Image planes do not update the file sequence unless the active panel
+        # is viewing through the camera.
+        panel_camera = cmds.modelPanel(
+            instance.data["panel"], query=True, camera=True
+        )
+        cmds.modelPanel(
+            instance.data["panel"], edit=True, camera=preset["camera"]
+        )
+
```
BigRoy, line 123: Does this also make sense to be done through a contextmanager?

```python
import contextlib
from maya import cmds

@contextlib.contextmanager
def panel_camera(panel, camera):
    original_camera = cmds.modelPanel(panel, query=True, camera=True)
    try:
        cmds.modePalen(panel, edit=True, camera=camera)
        yield
    finally:
        cmds.modePalen(panel, edit=True, camera=original_camera)


with panel_camera(instance.data["panel"], preset["camera"]):
    # playblasting here
```

Additionally - should any of this be done for thumbnails as well? Or is it just that they don't update over time?

tokejepsen: Yup will have a look later.

tokejepsen: Done

openpype/hosts/maya/plugins/publish/extract_playblast.py
```diff
@@ -25,6 +36,31 @@ class ExtractPlayblast(publish.Extractor):
     optional = True
     capture_preset = {}
 
+    def _capture(self, preset, override_viewport_options, instance):
+        filename = preset.get("filename", "%TEMP%")
```
BigRoy, line 40: Filename isn't used within this method and can just remain in the original process method. Then this method can just return only `path`

openpype/hosts/maya/plugins/publish/extract_playblast.py
```diff
@@ -138,28 +174,25 @@ def process(self, instance):
...
+                panel_camera(instance.data["panel"], preset["camera"])
+            ):
+                filename, path = self._capture(
+                    preset, override_viewport_options, instance
+                )
+        else:
+            with contextlib.ExitStack() as stack:
+                stack.enter_context(lib.maintained_time())
+                stack.enter_context(
+                    panel_camera(instance.data["panel"], preset["camera"])
```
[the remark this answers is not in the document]
tokejepsen:
```suggestion
        # Need to ensure Python 2 compatibility.
        # TODO: Remove once dropping Python 2.
        if getattr(contextlib, "nested", None):
            # Python 3 compatibility.
            with contextlib.nested(
                lib.maintained_time(),
                panel_camera(instance.data["panel"], preset["camera"])
            ):
                filename, path = self._capture(
                    preset, override_viewport_options, instance
                )
        else:
            # Python 2 compatibility.
            with contextlib.ExitStack() as stack:
                stack.enter_context(lib.maintained_time())
                stack.enter_context(
                    panel_camera(instance.data["panel"], preset["camera"])
```

openpype/hosts/maya/plugins/publish/extract_playblast.py
```diff
@@ -25,6 +36,31 @@ class ExtractPlayblast(publish.Extractor):
     optional = True
     capture_preset = {}
 
+    def _capture(self, preset, override_viewport_options, instance):
+        filename = preset.get("filename", "%TEMP%")
+
+        # Force viewer to False in call to capture because we have our own
+        # viewer opening call to allow a signal to trigger between
+        # playblast and viewer
+        preset['viewer'] = False
```
BigRoy, line 45: I feel like we could also move this outside of the function. There's actually no reason why this was originally in the maintained time thing.

openpype/hosts/maya/plugins/publish/extract_playblast.py
tokejepsen: Yup nice.

openpype/hosts/maya/plugins/publish/extract_playblast.py
tokejepsen: Yeah, reducing it to just the debug and capture now.

openpype/hosts/maya/plugins/publish/extract_playblast.py
```diff
@@ -25,6 +36,15 @@ class ExtractPlayblast(publish.Extractor):
...
     capture_preset = {}
 
+    def _capture(self, preset, override_viewport_options, instance):
+        self.log.info(
+            "Using preset:\n{}".format(
+                json.dumps(preset, sort_keys=True, indent=4)
+            )
+        )
+
+        return capture.capture(log=self.log, **preset)
```
BigRoy, line 46: This doesn't need arguments `override_viewport_options` and `instance`

openpype/hosts/maya/plugins/publish/extract_playblast.py
```diff
@@ -138,28 +158,41 @@ def process(self, instance):
...
+                stack.enter_context(lib.maintained_time())
+                stack.enter_context(
+                    panel_camera(instance.data["panel"], preset["camera"])
                 )
-            )
 
-            path = capture.capture(log=self.log, **preset)
+                path = self._capture(
+                    preset, override_viewport_options, instance
+                )
```
BigRoy, line 195: Looking at the code the returned `path` isn't used other than debug logging down below. Should we move that debug log into the capture method and return nothing.

It does feel a bit weird that we're using `filename` instead of `path` that it returned from capture? No idea if that returns anything usable.

But currently `path` variable seems quite redundant to pass around.

openpype/hosts/maya/plugins/publish/extract_playblast.py
tokejepsen: Good point.

openpype/hosts/maya/plugins/publish/extract_playblast.py
tokejepsen: Yeah, lets get rid of that `path` variable.

BigRoy (approved):
Did not do a test run myself, but code looks good to me. Maybe @LiborBatek should give it another go with recent changes?

LiborBatek:
@BigRoy did another test pass this time in ```maya 2022``` and all works fine...

image

tokejepsen (approved):
Tested in Maya 2020 succesfully.

antirotor (approved):

LiborBatek:
@antirotor as you wanted to re run test again...it all works as expected. no problems found.
image

PR #2971, opened 2022-03-29
Maya looks: support for native Redshift texture format

### LINKED ISSUE #2599: Maya looks: support for native Redshift texture format
## Feature

As we support OpenImageIO `.tx` files for Arnold, we should also add support for Redshift native texture support to avoid double conversions from `.tx` to `.rstexbin`.

I suggest using similar mechanism we use for conversion of `.tx` files - calling oiio `maketx`, but using `redshiftTextureProcessor` instead. This must be checked for existence and should reside in `$REDSHIFT_COREDATAPATH/bin` path (defined by redshift plugin in Maya).

Allan-I:
## Brief description
Add support for native Redshift textures handling. Closes #2599

## Description
Uses Redshift's Texture Processor executable to convert textures being used in renders to the Redshift ".rstexbin" format.

## Additional info
Essentially emulating what `maketx` does by passing a number of arguments to a `subprocess` and returning a status code to indicate whether the operation was successful or not.


openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -42,6 +43,58 @@ def find_paths_by_hash(texture_hash):
...
+        "bin",
+        "redshiftTextureProcessor"
+        )
+        
+    texture_processor_path = find_executable(redshift_bin_path)
+    
+    cmd = [
+        texture_processor_path,
+        escape_space(source),
+
```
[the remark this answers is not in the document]
iLLiCiTiT: He's right

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -42,6 +43,58 @@ def find_paths_by_hash(texture_hash):
...
+    Args:
+        source (str): Path to source file.
+        *args: Additional arguments for `redshiftTextureProcessor`.
+
+    Returns:
+        str: Output of `redshiftTextureProcessor` command.
+
+    """
+    if "REDSHIFT_COREDATAPATH" not in os.environ:
```
iLLiCiTiT, line 61: Note: I think we should validate this environment during validation, only when is needed if it's possible to know.

openpype/hosts/maya/plugins/publish/extract_look.py
Allan-I: Damn it skipped a line, fixed now. Thank you!

openpype/hosts/maya/plugins/publish/extract_look.py
Allan-I: Ah, I see! Thank you, I'll take a look about moving that

iLLiCiTiT: That's something I don't know. Maybe it's not needed or impossible to validate before this plugin, can't tell.

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -42,6 +43,58 @@ def find_paths_by_hash(texture_hash):
...
+
+    cmd = [
+        texture_processor_path,
+        escape_space(source),
+        
+    ]
+
+    cmd.extend(args)
+
+    cmd = " ".join(cmd)
```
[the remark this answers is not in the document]
Allan-I: Thank you!  This is helpful, makes me think whether we should consider refactoring the `maketx` function as well to simplify a little bit.

Will hop on to something else in the meantime.  I appreciate your time!

openpype/hosts/maya/plugins/publish/extract_look.py
antirotor: This is more complex as this is set only if studio is using and has configured Redshift. So I think the logic should be - add option `make redshift texture` on the same place as we have with make tx (look instance) and then probably have validator that will run on look family and will validate this env var.

openpype/lib/vendor_bin_utils.py
```diff
@@ -120,6 +120,26 @@ def get_oiio_tools_path(tool="oiiotool"):
...
+    Returns:
+        str: Full path to redshift texture processor executable.
+    """
+    redshift_tool_path = os.path.join(
+        os.environ["REDSHIFT_COREDATAPATH"],
+        "bin",
+        tool_name
+    )
+
+    return find_executable(redshift_tool_path)
```
BigRoy, line 140: Not a huge issue - but this does feel off here since unlike the rest of the methods here this Redshift binary isn't actually vendorized? I do like that there's unison in the function itself however - nonetheless it's not perfect.

Allan-I: Thank you @BigRoy, you bring up a good point.  I wonder where it would be good to place the function then @antirotor @iLLiCiTiT would love your take on where I should be placing this

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -450,7 +515,7 @@ def resource_destination(self, instance, filepath, do_maketx):
             resources_dir, basename + ext
         )
 
-    def _process_texture(self, filepath, do_maketx, staging, linearize, force):
+    def _process_texture(self, filepath, do_rstex, do_maketx, staging, linearize, force):
```
antirotor, line 518: some as comment above

```suggestion
    def _process_texture(self, filepath, processors, staging, linearize, force):
```

antirotor: and you should iterate over processors on file path there

openpype/lib/vendor_bin_utils.py
antirotor: I would probably define separate ENV var for the tool that should be defined in the tool itself to use it.

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -1,12 +1,14 @@
...
 """Maya look extractor."""
+from abc import ABC, abstractmethod
 import os
 import sys
 import json
 import tempfile
 import platform
 import contextlib
 import subprocess
+from openpype.lib.vendor_bin_utils import get_redshift_tool
```
antirotor, line 11: I think we shouldn't put it in `vendor_bin_utils` because we are not vendoring it. We should leave the code for tool detection in extractor because I think it is not currently used anywhere else

antirotor (changes requested):
Just added few comments. I would merge latest develop here. Don't bother with Renderman, we'll handle it in separate issue

openpype/hosts/maya/plugins/publish/extract_look.py
Allan-I: Hey thank you, moved it back to the same file!

openpype/hosts/maya/plugins/publish/extract_look.py
Allan-I: Done!

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -482,34 +566,9 @@ def _process_texture(self, filepath, do_maketx, staging, linearize, force):
...
-                "-sattrib",
-                "sourceHash",
-                escape_space(texture_hash),
-                colorconvert,
-            )
-
-            return converted, COPY, texture_hash
-
+        for processor in processors:
+            processor().process(filepath)
```
BigRoy, line 570: Should we make the `process` method a `staticmethod` or `classmethod` if it doesn't actually store anything on the Processor instance itself nor takes any init arguments?

BigRoy: Also shouldn't we be using the returned filepath?
For example:
```processed_path = processor.process(filepath)```

File: openpype/hosts/maya/plugins/publish/extract_look.py [hunk 8]
```diff
@@ -482,34 +566,9 @@ def _process_texture(self, filepath, do_maketx, staging, linearize, force):
...
+            self.log.info("Generating .rstexbin file for %s .." % filepath)
```
BigRoy, line 571: This logged message is incorrect - since it's not only about `.rstexbin` files anymore - correct?

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -22,6 +24,28 @@
...
+    """
+    redshift_os_path = os.environ["REDSHIFT_COREDATAPATH"]
+
+    redshift_tool_path = os.path.join(
+        redshift_os_path,
+        "bin",
+        tool_name
+    )
+
+    return find_executable(redshift_tool_path)
```
BigRoy, line 46: Should this maybe be a staticmethod on the `MakeRSTexBin` class. @iLLiCiTiT thoughts?

openpype/hosts/maya/plugins/publish/extract_look.py
Allan-I: Fixed!

openpype/hosts/maya/plugins/publish/extract_look.py
Allan-I: You are indeed correct, fixed!

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -482,35 +566,10 @@ def _process_texture(self, filepath, do_maketx, staging, linearize, force):
...
-                colorconvert,
-            )
-
-            return converted, COPY, texture_hash
-
-        return filepath, COPY, texture_hash
+        for processor in processors:
+            processed_path = processor().process(filepath)
+            self.log.info("Generating texture file for %s .." % filepath)
+        return processed_path, COPY, texture_hash
```
BigRoy, line 572: This will return only the last processed_path instead of all processed files.
Also, if no processors were active then `processed_path` would never be defined.

Allan-I: Ah, you are right.  Then I suppose when need to be using `processed_path` at the end after it's been processed.

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -43,64 +67,123 @@ def find_paths_by_hash(texture_hash):
     return legacy_io.distinct(key, {"type": "version"})
 
 
-def maketx(source, destination, *args):
-    """Make `.tx` using `maketx` with some default settings.
+class TextureProcessor(object):
+    __metaclass__ = ABCMeta
```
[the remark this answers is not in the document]
Allan-I: Why's maya screaming at the python3 syntax then

m-u-r-p-h-y (changes requested):
```
Instance: 
lookMain
Message: 
can't concat str to bytes
Line: 
569
Traceback: 
    runner(*args)
  File "D:\REPO\OpenPype\openpype\hosts\maya\plugins\publish\extract_look.py", line 340, in process
  File "D:\REPO\OpenPype\openpype\hosts\maya\plugins\publish\extract_look.py", line 486, in process_resources
  File "D:\REPO\OpenPype\openpype\hosts\maya\plugins\publish\extract_look.py", line 569, in resource_destination
TypeError: can't concat str to bytes
```

image

Allan-I:
Thank you Murphy, can you please recreate your environment through `create_env` in tools and try again?

m-u-r-p-h-y:
run the command
image

but the publish ended with exactly the same error :(

Allan-I:
Thank you Murphy, investigating now

openpype/lib/vendor_bin_utils.py
m-u-r-p-h-y: > I would probably define separate ENV var for the tool that should be defined in the tool itself to use it.

is it necessary? why can't we use the mandatory  `{REDSHIFT_COREDATAPATH}/bin/redshiftTextureProcessor.exe ` (and the appropriate lnx and mac variant)

openpype/hosts/maya/plugins/publish/extract_look.py
iLLiCiTiT: I agree, it's used only there.

m-u-r-p-h-y (approved):
look with textures extracted and published successfully

image

BigRoy:
> [look with textures extracted and published successfully](https://github.com/pypeclub/OpenPype/pull/2971#pullrequestreview-1097251132)

Shouldn't that produce `.rstexbin` files with this PR instead of `.tx` files?

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -520,38 +628,45 @@ def _process_texture(self, filepath, do_maketx, staging, linearize, force):
...
+                        return processed_path, COPY, texture_hash
+                    else:
+                        self.log.info("maketx has returned nothing")
+                elif processor is MakeRSTexBin:
+                    processed_path = processor().process(filepath)
+                    self.log.info("Generating texture file for %s .." % filepath) # noqa
+                    if processed_path:
+                        return processed_path, COPY, texture_hash
+                    else:
+                        self.log.info("redshift texture converter has returned nothing") # noqa
```
BigRoy, line 667: Isn't it odd that we define something like a `Processor` base class but then end up needing to pass different types of input variables for the conversion - shouldn't those at least have the same method signatures so that when adding an extra type of processor it's just a means of implementing that base class?

m-u-r-p-h-y:
> > [look with textures extracted and published successfully](https://github.com/pypeclub/OpenPype/pull/2971#pullrequestreview-1097251132)
> 
> @m-u-r-p-h-y - Shouldn't that produce `.rstexbin` files with this PR instead of `.tx` files?
> 
> Did you enable to `rstex` attribute on the look instance?

Ohh you are completely right. I had old look instance in my scene. 

Redid the test and it is correctly producing rstexbin files.

my question is, do we really need to specify the format in the instance? Isn't it clear that we should do

arnold -> tx
redshift -> rxtexbin


this way it opens a space for human error as I already demonstrated . . . 

image

BigRoy:
> my question is, do we really need to specify the format in the instance? Isn't it clear that we should do
> arnold -> tx
> redshift -> rxtexbin

Actually I'd say yes. I believe Redshift can actually render both `.tx` and `.rstexbin` files but they serve a different purpose. (I might be wrong however!) I recall having done a test on our infrastructure and it was usually faster to render without `.rstexbin` files and use regular textures or `.tx` files. But then again, I might be misremembering.

Check the **[Render Time Considerations](https://docs.redshift3d.com/display/RSDOCS/Texture+Processor+Tool#TextureProcessorTool-RenderTimeConsiderations)**

antirotor:
I agree that we need to specify format, but we'll need to add in the future validators to deal with unsupported features enabled on  existing render instance - for example arnold/vray/renderman won't understand .rstexbin, renderman will not understand .tx/.rstexbin but will read .tex and so on.

m-u-r-p-h-y:
> I agree that we need to specify format, but we'll need to add in the future validators to deal with unsupported features enabled on existing render instance - for example arnold/vray/renderman won't understand .rstexbin, renderman will not understand .tx/.rstexbin but will read .tex and so on.

the toggle in render instance could only state

- [ ] convert textures to native format

and we should know, depending on renderer which format to use . . .

Allan-I:
Would that be ideal?  To have that toggle on the instance?

m-u-r-p-h-y:
> Would that be ideal? To have that toggle on the instance?

we have this currently  
image

and it is not ideal definitely

openpype/hosts/maya/plugins/publish/validate_look_contents.py
```diff
@@ -112,3 +114,23 @@ def validate_files(cls, instance):
                 invalid.append(node)
 
         return invalid
+
+    @classmethod
+    def validate_renderer(cls, instance):
+
+        renderer = cmds.getAttr(
+                'defaultRenderGlobals.currentRenderer').lower()
```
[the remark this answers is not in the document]
Allan-I: Okay hound I'll fix it buddy, and get you a treat, too

m-u-r-p-h-y (approved):
tested successfully

Maya2023 and RS 5.3.09 

image

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -514,36 +629,45 @@ def _process_texture(self, filepath, do_maketx, staging, linearize, force):
...
-                filepath,
-                converted,
-                additional_args,
-                self.log
-            )
-
-            return converted, COPY, texture_hash
-
-        return filepath, COPY, texture_hash
+        config_path = get_ocio_config_path("nuke-default")
```
iLLiCiTiT, line 632: I have to say few notes regarding subprocess arguments. If it is possible (and here it is possible) arguments should be passed to subprocess functions as list of strings so you don't have to call anything like `escape_space`.

Also I agree with @BigRoy about those explicit arguments for `TextureProcessor`. If one has different arguments then other then what is the reason of having them in abstract class... They should not be prepared for specific processor before they're passed to their processing method and all of the methods should expect the same arguments.

- `colorconvert` variable should be split to `src_colorspace` and `dst_colorspace` which are passed to process method
- the same for `color_config` which is just path to config file where colorspaces can be found
- the same for `texture_hash`
- the same for `staging_dir` (handling `converted` varible probably should not happen if MakeTX is not executed?)
In that case eash processor's `process` method arguments should expect these arguments:
```
def process(
    self,
    src_filepath,
    staging_dir,
    color_config_path,
    src_colorspace,
    dst_colorspace
):
    pass
```


So `process` method of `MakeTX` can be changed to:
```
    def process(
        self,
        src_filepath,
        staging_dir,
        color_config_path,
        src_colorspace,
        dst_colorspace
    ):
        """Make `.tx` using `maketx` with some default settings.
        The settings are based on default as used in Arnold's
        txManager in the scene.
        This function requires the `maketx` executable to be
        on the `PATH`.
        Args:
            ...fill
        Returns:
            str: Output filepath/s.
        """
        from openpype.lib import get_oiio_tools_path

        maketx_path = get_oiio_tools_path("maketx")

        if not os.path.exists(maketx_path):
            print(
                "OIIO tool not found in {}".format(maketx_path))
            raise AssertionError("OIIO tool not found")

        dirpath, filename = os.path.split(src_filepath)
        fname, ext = os.path.splitext(filename)
        dst_filepath = os.path.join(staging_dir, "resources", fname + ".tx")
        if not os.path.exists(os.path.dirname(dst_filepath)):
            os.makedirs(os.path.dirname(dst_filepath))

        cmd = [
            maketx_path,
            "-v",  # verbose
            "-u",  # update mode
            # unpremultiply before conversion (recommended when alpha present)
            "--unpremult",
            "--checknan",
            # use oiio-optimized settings for tile-size, planarconfig, metadata
            "--oiio",
            "--filter", "lanczos3",
            src_filepath,
        ]

        additional_args = []
        if src_colorspace and dst_colorspace and color_config_path:
            additional_args.extend([
                "--colorconfig", color_config_path,
                "--colorconvert", src_colorspace, dst_colorspace
            ])
        cmd.extend(additional_args)
        cmd.extend(["-o", dst_filepath])

        cmd = " ".join(cmd)

        kwargs = dict(args=cmd, stderr=subprocess.STDOUT)
        if hasattr(subprocess, "CREATE_NO_WINDOW"):
            kwargs["creationflags"] = subprocess.CREATE_NO_WINDOW

        subprocess.call(**kwargs)

        return dst_filepath
```

antirotor (changes requested):

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -68,59 +95,129 @@ def find_paths_by_hash(texture_hash):
...
+        if sys.platform == "win32":
+            kwargs["creationflags"] = CREATE_NO_WINDOW
+        try:
+            subprocess.check_output(**kwargs)
+        except subprocess.CalledProcessError as exc:
+            print(exc)
+            import traceback
+
+            traceback.print_exc()
+            raise
```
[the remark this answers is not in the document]
mkolar: We should not be raising `PublishValidationError` in extractors. That is too late and should be standard error exception. validation error is reserved for things that the artist can affect and fix and should be caught in the validation step.

PR #2972, opened 2022-03-29
Add toggle-all button to Loaders' family filter

icyvapor:
**Note:  Please squash all commits before merging!**

## Description
One minor user-friendly issue we noticed is, when there are many different assets published under a given asset, user has to rely on the filter to pick the family of asset  they want to load. 

Currently this requires them to right click, "Disable all", then pick the target family. Got complaint from artists, so I added an toggle button to toggle them all. 

Modified both the embedded loader window and the library loader. 

image

Note: 

Got a few commits from syncing up my fork and some changes I reverted. Could you make sure to **squash** before merging in this PR?  Thanks a lot.


BigRoy:
I love exposing better ways to quickly change the family filtering but I'm not necessarily a fan of this button there. It's a bit of visual clutter which I feel is more clutter than the amount of usage the feature might warrant or get in the end.

We have a feature implemented in Maya for selection masks where control clicking an entry would Disable all others and Isolate the single entry you clicked. It's something you "need to know" is possible because UX wise it's not a familiar default but it's been super pleasant to everyone in our team over the years. I can imagine similar behavior here would be lovely flavour. It's really nice if you know it and we can put "Tip!" info in the documentation about it. Control clicking an entry if it is already isolated (only selected entry) re-enables all entries.

I'm not a huge fan of the button because it takes so much screen space for just "toggling". I'd much rather have right click menu added features like: "Isolate selected" (if there's a current active selection) and a "Toggle All" button in the context menu.

I wonder if there's other better UX approaches aside of needing the actual button there all the time.

Thoughts?

icyvapor:
This button is an actual request by someone in the studio. 
My screenshot above only showed a sandbox project, but for actual production, the subset window on the right quickly fills up with each department publishing multiple subsets per family. And we have been creating our own custom families. So from both observation and my own experience testing, every time we open the loader, the workflow is almost always: 1) right-click 2) disable all 3) click on the specific family 4) find the specific asset on the right. 

So I think a button that does 1) and 2) by one click will be easier and faster than the context menu way.  Ctrl-selection is a nice tip for sure, though I doubt artists mainly using Wacom tablets will be super happy with reaching for Ctrl on keyboard. 

I'm totally fine with this staying as our local customization. Just putting this out here in case it's useful.

BigRoy:
> the workflow is almost always: 1) right-click 2) disable all 3) click on the specific family 4) find the specific asset on the right.

Note that you can manage the default visible families per host in the Admin Settings. So you could technically disable almost all. (I wouldn't disable all by default since that'd definitely be confusing!) But as an example we have workfiles disabled by default and some others - since they tend to not be used 99% of the cases.

In Fusion e.g. we also don't have a lot of Maya related publishes visible since there isn't a sensible usage/loader for it by default.

> So I think a button that does 1) and 2) by one click will be easier and faster than the context menu way. Ctrl-selection is a nice tip for sure, though I doubt artists mainly using Wacom tablets will be super happy with reaching for Ctrl on keyboard

Totally agree that if you need to do this _that_ often I can imagine it might be nice. I'm mostly in a visual disagreement haha - and would love to know other UX approaches that might solve the same thing and maybe even in a better (or more generic way).

tokejepsen:
Here is an alternative to the button;

Capture

This is from the Deadline Monitor.

iLLiCiTiT:
The families are now filtered by selected asset so you may not see any families at the moment when you click the button.

I would like to know the use case. Because it looks like the "Toggle all" is actually "disable family filtering". In that case I like @tokejepsen (or deadline's?) idea or I would say that the view is big enough to add button (with some suitable icon) to it's right bottom corner which would disable families filter completely without changing checkbox values.

image

BigRoy:
> But yea, a pre-configured family filtering may take wind out of this button

I guess you could definitely roll out both now and see where you land.

icyvapor:
> @icyvapor have you tried the family configurations? It feels to me that this button doesn't add that much value to that system. Let us know please if this is still requested by your artists even with a proper config. We can then revisit it.
> 
> Otherwise, I'd close the PR. I'll wait for your input though.

Arrhg, I didn't get around to look into that, there are many (bigger, and sometimes not OP related) fishes to fry. 
I deployed this internally, and artists over here have been using the button. 
I'll close this, later I'll play with the family config to possibly deprecate this button.

Thanks!

PR #2975, opened 2022-03-30
General: Tools with host filters

iLLiCiTiT:
## Brief description
Tools can be filtered for specific host or application.

## Description
Right now are all tools environments applied to all applications which may break then and it is not possible to have different versions for different versions of application. It was added a hosts and application variants filter for each tool variant. Both filters must be matched. Missing filters mean that the tool will be used at all cases. The filtering using host will probably be used in most of cases but for more granular filtering was also added a application variant filter.

## Additional info
This change require to change how tools are stored so a "hasckish" backwards compatible conversion in settings was added when studio settings overrides are loaded so old environment values are not lost. But changes are still showed in settings.

Application variants are not "live" updated so there may be confusion on user's side. Added/renamed/removed application variant needs save and refresh to update combobox of applications.

## Changes
- changed settings schemas to contain host and app variants filter
- EnvironmentTool uses the host and app filtering which is checked using `is_valid_for_app`

## Testing notes:
1. Set filters of tool 1 for **app one** of tool 2 for **app two** and remove filters for tool 3
2. Run one of application and check if environments were added as expected

### DIFF
--- a/openpype/lib/applications.py
+++ b/openpype/lib/applications.py
@@ -571,13 +571,8 @@ class EnvironmentTool:
         # Backwards compatibility 3.9.1 - 3.9.2
         # - 'variant_data' contained only environments but contain also host
         #   and application variant filters
-        host_names = []
-        app_variants = []
-        if "host_names" in variant_data:
-            host_names = variant_data["host_names"]
-
-        if "app_variants" in variant_data:
-            app_variants = variant_data["app_variants"]
+        host_names = variant_data.get("host_names", [])
+        app_variants = variant_data.get("app_variants", [])
 
         if "environment" in variant_data:
             environment = variant_data["environment"]
--- a/openpype/lib/applications.py
+++ b/openpype/lib/applications.py
@@ -211,6 +211,7 @@ class ApplicationGroup:
         data (dict): Group defying data loaded from settings.
         manager (ApplicationManager): Manager that created the group.
     """
+
     def __init__(self, name, data, manager):
         self.name = name
         self.manager = manager
@@ -374,6 +375,7 @@ class ApplicationManager:
             will always use these values. Gives ability to create manager
             using different settings.
     """
+
     def __init__(self, system_settings=None):
         self.log = PypeLogger.get_logger(self.__class__.__name__)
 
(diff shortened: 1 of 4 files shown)


openpype/settings/lib.py
```diff
@@ -265,11 +265,43 @@ def save_project_anatomy(project_name, anatomy_data):
...
+
+            for key in set(variants.keys()):
+                if key in METADATA_KEYS:
+                    continue
+
+                variant_value = variants[key]
+                if "environment" not in variant_value:
+                    variants[key] = {
+                        "environment": variant_value
+                    }
```
BigRoy, line 292: Should this be more explicit and come with a version check? If `settings version < 3.9.2` do this?

openpype/lib/applications.py
```diff
@@ -560,15 +562,35 @@ class EnvironmentTool:
...
+        # Backwards compatibility 3.9.1 - 3.9.2
+        # - 'variant_data' contained only environments but contain also host
+        #   and application variant filters
+        host_names = []
+        app_variants = []
+        if "host_names" in variant_data:
+            host_names = variant_data["host_names"]
+
+        if "app_variants" in variant_data:
+            app_variants = variant_data["app_variants"]
```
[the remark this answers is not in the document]
BigRoy: Additionally with `_system_settings_backwards_compatible_conversion` implemented. Would `environments` ever NOT be there? It seems to already resolve this case and the above logic I suggest here should be fine?

BigRoy (commented):
I love to see this incoming! Nice work.

openpype/settings/lib.py
iLLiCiTiT: > Should this be more explicit and come with a version check?

It would be nice but don't know how to 100% achieve it. Since the zip files are created with "some version" which may have custom name with custom changes. It would be random until 3.9.2 realease.

But I think it would be possible to check if the version is lower than 3.9.2, so it's more clear?

BigRoy: Ah of course - studios could do whatever with the version numbers. It's a bit odd that this compatible conversion logic would also potentially trigger for newer versions in the code base. Would have loved if we could've cleaned that up somehow.

antirotor (commented):
Question: What should be the behavior if you select only **Host** but not **Applications**, shouldn't it apply to all?

iLLiCiTiT:
> What should be the behavior if you select only Host but not Applications, shouldn't it apply to all? Currently it seems it doesn't set it at all in this case.

If you select just a host then it should apply to all applications having the host.

antirotor:
> > What should be the behavior if you select only Host but not Applications, shouldn't it apply to all? Currently it seems it doesn't set it at all in this case.
> 
> If you select just a host then it should apply to all applications having the host.

well it doesn't work for me

image

`MTOA_VERSION` should be set but it isn't:

image

antirotor:
ahh, but it isn't assigned to project, nevermind... stupid me.

antirotor (approved):
works! thanks

PR #2983, opened 2022-03-30
SiteSync: Added compute_resource_sync_sites to sync_server_module

kalisp:
## Brief description
Part or refactoring of `integrate_new`, should separate Site Sync logic out.

## Description
Exposed `compute_resource_sync_sites` as a public method of `sync_server_module`. Integrate process should call only this 
method with project name argument.
It will return prepared dictionary of configured sites.


BigRoy:
This seems like a close copy from the Integrator code so I assume this will work. I will however hold of on testing/feedbacking until more detailed review was done on #2898 so the logic really is aligned as much as possible.

Thanks @kalisp - appreciate the quick effort into this!

tests/unit/openpype/modules/sync_server/test_module_api.py
```diff
@@ -0,0 +1,46 @@
...
+    def setup_sync_server_module(self, dbcon):
+        """Get sync_server_module from ModulesManager"""
+        from openpype.modules import ModulesManager
+
+        manager = ModulesManager()
+        sync_server = manager.modules_by_name["sync_server"]
+        yield sync_server
+
+    def test_get_alt_site_pairs(self, setup_sync_server_module):
+        conf_sites = {'SFTP': {"alternative_sites": ["studio"]},
```
BigRoy, line 36: Consistency in quotes
```suggestion
        conf_sites = {"SFTP": {"alternative_sites": ["studio"]},
```

tests/unit/openpype/modules/sync_server/test_module_api.py
```diff
@@ -0,0 +1,46 @@
...
+                      "studio2": {"alternative_sites": ["studio"]}}
+
+        ret = setup_sync_server_module._get_alt_site_pairs(conf_sites)
+        expected = {"SFTP": ["studio", "studio2"],
+                    "studio": ["SFTP", "studio2"],
+                    "studio2": ["studio", "SFTP"]}
+        assert ret == expected, "Not matching result"
```
BigRoy, line 43: I feel like just one level of alternate sites that matches doesn't accurately represent whether we tackle 'cycles' well enough and whether the recursion works across a depth of one.

Should we write a test for case like this:
```
        conf_sites = {"A": {"alternative_sites": ["C"]},
                      "B": {"alternative_sites": ["C"]},
                      "C": {"alternative_sites": ["D"]},
                      "D": {"alternative_sites": ["A"]},
                      "F": {"alternative_sites": ["A"]},
                      }
```
This tackles A -> B -> C -> D

Also - your test only passes if the order of the output is actually like you have it defined here - however. Is that order really relevant for the pass to test? Like, do we absolutely require them to be ordered a certain way in the lists? I just wasn't sure if that had any value - if it doesn't have value we might want to use `set` for the alternatives?

openpype/modules/sync_server/sync_server_module.py
```diff
@@ -184,6 +183,151 @@ def remove_site(self, collection, representation_id, site_name,
...
+        """Returns dict of site and its alternative sites.
+
+        If `site` has alternative site, it means that alt_site has 'site' as
+        alternative site
+        Args:
+            conf_sites (dict)
+        Returns:
+            (dict): {'site': [alternative sites]...}
+        """
+        alt_site_pairs = defaultdict(list)
```
BigRoy, line 303: This could be `defaultdict(set)` see https://github.com/pypeclub/OpenPype/pull/2898/commits/1a03bbe48a37cc62918152985488a0bd99d43473

BigRoy:
Should I already refactor my Refactor Integrator PR to include this? That way there's less duplicated work going on with maintaining functionality. Especially because this PR comes with tests I'd rather start using this asap.

tests/unit/openpype/modules/sync_server/test_module_api.py
kalisp: Good idea, added more complex unit test.

BigRoy (approved):
Nice work. Changes look solid to me!

mkolar:
>Should I already refactor my Refactor Integrator PR to include this? That way there's less duplicated work going on with maintaining functionality. Especially because this PR comes with tests I'd rather start using this asap.

Definitelly. Merging now, you go for it.

PR #2986, opened 2022-03-30
Maya: Don't do hardlinks on windows for look publishing

iLLiCiTiT:
## Brief description
Hardlink on windows cause that the hardlinked file gets locked at  the moment event the exact path is not used.

## Description
Disabled hardlinks on windows platform during look publishing. We should consider where else this may affect the publishing. The proper fix should be considered in future that we would create "new hardlink" for hardlinked files and remove the temp hardlink but the main issue is with hero publishing where whole folder is renamed to temp.

## Testing notes:
1. Publish look
2. Integration should not crash because cannot remove temporary file

### DIFF
--- a/openpype/hosts/maya/plugins/publish/extract_look.py
+++ b/openpype/hosts/maya/plugins/publish/extract_look.py
@@ -337,6 +337,9 @@ class ExtractLook(openpype.api.Extractor):
         hashes = {}
         # Temporary fix to NOT create hardlinks on windows machines
         if platform.system().lower() == "windows":
+            self.log.info(
+                "Forcing copy instead of hardlink due to issues on Windows..."
+            )
             force_copy = True
         else:
             force_copy = instance.data.get("forceCopy", False)


openpype/hosts/maya/plugins/create/create_look.py
```diff
@@ -22,4 +22,6 @@ def __init__(self, *args, **kwargs):
         self.data["maketx"] = self.make_tx
 
         # Enable users to force a copy.
-        self.data["forceCopy"] = True
+        # - on Windows is "forceCopy" always changed to `True` because of
+        #   windows implementation of hardlinks
```
BigRoy, line 26: If the fix is temporary and we log it to the end user (see below in the code) we might be fine with not having this comment in the Creator code. I'm fine with either.

BigRoy (commented):
Would've been nice if instead we would have been able to solve the issue but I guess it's an important hotfix.

I actually believe the #2898 will work around the issue on Windows because the issue was solely with renaming the hardlinked files from `.tmp` to non `.tmp` and the new integrator does that slightly differently so just might not run into the same issue. :) It only renames backup files.

antirotor (approved):

PR #2988, opened 2022-03-31
Publishing textures for Unreal

antirotor:
## Feature

This adds simple family to publish texture with [Unreal naming convention](https://github.com/Allar/ue5-style-guide).

### Usage

Launch **Standalone Publisher**, selecte `simple Unreal Texture` family and asset and drag file named like `T_Foo_D.png`(where `Foo` is asset name). It should publish normal and hero version. Hero version should keep the same name. Validation will check if the file name is starting with asset name `T_[ASSET].*`.

### Setup

It defines two additional templates `simpleUnrealTextureHero` adn `simpleUnrealTexture`:

image

Their use must be configured in **Template name profiles**:
`project_settings/global/publish/IntegrateAssetNew/template_name_profiles`
image

Added here to **Integrate Hero Version**:
`project_settings/global/publish/IntegrateHeroVersion`
image

And then select the template:
`project_settings/global/publish/IntegrateHeroVersion/template_name_profiles`
image

---

| 📥 includes: #2967 |
|---|


m-u-r-p-h-y (approved):
tested with standalone Publish

the texture was copied to the correct publish folder and versioned accordingly . .

PR #2990, opened 2022-03-31
Fix: Disable setuptools auto discovery

antirotor:
## Problem

With updated [setuptools](https://github.com/pypa/setuptools) new behaviour emerged that is not backwards compatible,
introducing non-backwards compatible breaking changes.

Setting packages to an empty list will disable autodiscovery and should permit build to run as before.

Described here pypa/setuptools#3197

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -188,5 +188,5 @@ setup(
         }
     },
     executables=executables,
-    package_dir=[]
+    packages=[]
 )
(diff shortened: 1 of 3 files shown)


m-u-r-p-h-y (approved):
works for me.

remove .potery
remove .venv

.\tools\create_env.ps1
.\tools\build.ps1

success

`*** All done in 56 secs. You will find OpenPype and build log in '.\build' directory.`

mkolar (approved):

PR #2991, opened 2022-03-31
PS: fix renaming subset incorrectly in PS

kalisp:
## Brief description
Validator Repair method used layer name as a subset name.

## Testing notes:
1. create multiple layers with broken names (layer name containing `/`, or `;` ...)
2. create for all of them instances (in single Creator, 'Use Selection')
3. use `Repair` action when `Validate Naming` fails
4. compare published file names format with unbroken layer names

### DIFF
--- a/openpype/settings/defaults/project_settings/photoshop.json
+++ b/openpype/settings/defaults/project_settings/photoshop.json
@@ -18,7 +18,7 @@
             "active": true
         },
         "ValidateNaming": {
-            "invalid_chars": "[ \\\\/+\\*\\?\\(\\)\\[\\]\\{\\}:,]",
+            "invalid_chars": "[ \\\\/+\\*\\?\\(\\)\\[\\]\\{\\}:,;]",
             "replace_char": "_"
         },
         "ExtractImage": {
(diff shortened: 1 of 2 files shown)


openpype/settings/defaults/project_settings/photoshop.json
```diff
@@ -18,7 +18,7 @@
             "active": true
         },
         "ValidateNaming": {
-            "invalid_chars": "[ \\\\/+\\*\\?\\(\\)\\[\\]\\{\\}:,]",
+            "invalid_chars": "[ \\\\/+\\*\\?\\(\\)\\[\\]\\{\\}:,;]",
```
iLLiCiTiT, line 21: Should these be in settings? If it's related to subset names then we should not allow any characters that are not in `SUBSET_NAME_ALLOWED_SYMBOLS` (`openpype.pipeline.create`).

64qam (approved):

openpype/settings/defaults/project_settings/photoshop.json
kalisp: SUBSET_NAME_ALLOWED_SYMBOLS is allowing only ascii characters, we should also support utf-8 characters (as they might appear in layer names).
But thats a question for different PR.

iLLiCiTiT: Created issue https://github.com/pypeclub/OpenPype/issues/2993

PR #2992, opened 2022-03-31
nuke: bypass baking

jakubjezek001:
## Brief description
Added support for review without baking

## Description
Since someone can set workfow rendering to png with sRGB so screenspace already baked into image sequence, they can skip baking because only purpose of it is to bake screen space into reviewable files. 

## Testing notes:
1. go to settings on your project and disable `project_settings/nuke/publish/ExtractReviewDataMov`
2. go to settings on your project and enable `project_settings/nuke/publish/ExtractReviewData`
3. go to settings `project_anatomy/imageio/nuke/nodes/requiredNodes` and set `CreateWriteRender` to be png `file_type`, remove [`compression`, `cropped`], set `datatype` to "16 bit" and `colorspace` to "sRGB" or "Output - sRGB" (if you are in ACES)
4. Render locally, on farm, from rendered files
5. all should look the same and produce reviewable version on Ftrack

### DIFF
--- a/openpype/hosts/nuke/plugins/publish/extract_review_data.py
+++ b/openpype/hosts/nuke/plugins/publish/extract_review_data.py
@@ -25,7 +25,7 @@ class ExtractReviewData(openpype.api.Extractor):
             instance.data["families"].remove("review")
 
         for repre in representations:
-            if ext not in repre["ext"]:
+            if ext != repre["ext"]:
                 continue
 
             if not repre.get("tags"):
--- a/openpype/hosts/nuke/plugins/publish/extract_review_data.py
+++ b/openpype/hosts/nuke/plugins/publish/extract_review_data.py
@@ -21,9 +21,15 @@ class ExtractReviewData(openpype.api.Extractor):
 
         representations = instance.data.get("representations", [])
 
-        if "render.farm" in instance.data["families"]:
+        # review can be removed since `ProcessSubmittedJobOnFarm` will create
+        # reviable representation if needed
+        if (
+            "render.farm" in instance.data["families"]
+            and "review" in instance.data["families"]
+        ):
             instance.data["families"].remove("review")
 
+        # iterate representations and add `review` tag
         for repre in representations:
             if ext != repre["ext"]:
                 continue
--- a/openpype/hosts/nuke/plugins/publish/extract_review_data.py
+++ b/openpype/hosts/nuke/plugins/publish/extract_review_data.py
@@ -1,0 +1,41 @@
+import os
+import pyblish.api
+import openpype
+from pprint import pformat
+
+
+class ExtractReviewData(openpype.api.Extractor):
+    """Extracts review tag into available representation
+    """
+
+    order = pyblish.api.ExtractorOrder + 0.01
+    # order = pyblish.api.CollectorOrder + 0.499
+    label = "Extract Review Data"
+
+    families = ["review"]
+    hosts = ["nuke"]
+
+    def process(self, instance):
+        fpath = instance.data["path"]
+        ext = os.path.splitext(fpath)[-1][1:]
+
+        representations = instance.data.get("representations", [])
+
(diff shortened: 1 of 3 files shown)


openpype/hosts/nuke/plugins/publish/extract_review_data.py
```diff
@@ -0,0 +1,41 @@
+import os
+import pyblish.api
+import openpype
+from pprint import pformat
+
+
+class ExtractReviewData(openpype.api.Extractor):
```
iLLiCiTiT, line 7: Why it's extractor and not collector?

jakubjezek001: hoo, good catch :D - that hurts..

openpype/hosts/nuke/plugins/publish/extract_review_data.py
```diff
@@ -0,0 +1,41 @@
...
+        fpath = instance.data["path"]
+        ext = os.path.splitext(fpath)[-1][1:]
+
+        representations = instance.data.get("representations", [])
+
+        if "render.farm" in instance.data["families"]:
+            instance.data["families"].remove("review")
+
+        for repre in representations:
+            if ext not in repre["ext"]:
```
iLLiCiTiT, line 28:
```suggestion
            if ext != repre["ext"]:
```

openpype/hosts/nuke/plugins/publish/extract_review_data.py
```diff
@@ -0,0 +1,41 @@
...
+    families = ["review"]
+    hosts = ["nuke"]
+
+    def process(self, instance):
+        fpath = instance.data["path"]
+        ext = os.path.splitext(fpath)[-1][1:]
+
+        representations = instance.data.get("representations", [])
+
+        if "render.farm" in instance.data["families"]:
```
iLLiCiTiT, line 24: Won't this cause that farm publishing will not produce review?

jrsndl (commented):
I tested on ACES 1.2 project, works as described.
Prores422HQ and mp4 is color matched to PNG very well.
Can't check farm publish right now, sorry


Changing CreateWriteRender to png didn't change the already existing exr Write to png. Assuming that is by design.

64qam:
I am not sure if I test it correctly, but the `bake` file is still there (both local and farm render)

openpype/hosts/nuke/plugins/publish/extract_review_data.py
jakubjezek001: this is done by `ProcessSubmittedJobOnFarm` so no need to have `review` in families. If you keep it there it will be switching `ExtractReview` during publishing on local processing, which is not desirable.

jakubjezek001:
> I am not sure if I test it correctly, but the `bake` file is still there (both local and farm render)

this could be due older bake files, if you didnt remove all before you tested it. Could you confirm please?

openpype/hosts/nuke/plugins/publish/extract_review_data.py
iLLiCiTiT: Then maybe add comment here which explaint why it's removed and also would be good to check if `"review"` is in families before removing it?

jakubjezek001: good point, thanks ;)

64qam:
All older files have been deleted and all files have been recreated with bake.mov. The same for brand new task.

jakubjezek001:
> All older files have been deleted and all files have been recreated with bake.mov. The same for brand new task.

did that happened even you had followed step 1. also at project overrides?

64qam:
Yes, first of all in global project settings I disabled `project_settings/nuke/publish/ExtractReviewDataMov` save it, enable `project_settings/nuke/publish/ExtractReviewDataMov`, save it, then did the same procedure on my test project.

64qam (approved):

PR #2995, opened 2022-04-01
Deadline: priority configurable in Maya jobs

kalisp:
## Brief description
Priority is pulled from Settings when instance is being created and it is propagated into Deadline.

## Description
Added separate `tile_priority` to differentiate regular render and tile jobs.

maya_dl_priority

## Testing notes:
1. modify `project_settings/deadline/publish/MayaSubmitDeadline` priorities
2. create instances of 'render' family in Maya
3. priorities should be visible when clicked on instance in a list of user configurable attributes
4. submit to Deadline, monitor priorities (accessible in `Job Priorities` (double click on job name), section `General`)


openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
```diff
@@ -272,37 +276,12 @@ def process(self, instance):
...
-        )
-
-        self.group = (
-            context.data["project_settings"].get(
-                "deadline", {}).get(
-                "publish", {}).get(
-                "MayaSubmitDeadline", {}).get(
-                "group", "none")
-        )
+        self.limit_groups = self.limit
```
BigRoy, line 284: Would be nicer if somehow we could refactor the rest of the code to just use the class variables directly instead of re-assigning them. I agree the `self.limit` name isn't very descriptive however.

If we'd just use the existing settings info we could however remove these three lines completely.

Should we refactor that too to cleanup some technical debt there? It might mean to also tweak the settings names however?

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
```diff
@@ -254,7 +254,11 @@ class MayaSubmitDeadline(pyblish.api.InstancePlugin):
     use_published = True
     tile_assembler_plugin = "OpenPypeTileAssembler"
     asset_dependencies = False
+    priority = 50
+    tile_priority = 50
     limit_groups = []
+    jobInfo = None
+    pluginInfo = None
```
BigRoy, line 261: Looking at the previous code these should defaulto to `{}`?

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -937,6 +931,11 @@ def process(self, instance):
                 self.deadline_url = instance.data.get("deadlineUrl")
             assert self.deadline_url, "Requires Deadline Webservice URL"
 
+            if self.deadline_priority:
+                render_job["Props"]["Pri"] = self.deadline_priority
+            else:
+                render_job["Props"]["Pri"] = instance.data.get("priority")
+
```
BigRoy, line 938: I know this matches older code but this does seem odd.

Instead of changing the Priority of the job that we are submitting we're changing the properties data of the already submitted job to update how we're setting the priority for the Publish Job.

It feels like we should be moving this to [around line 245](https://github.com/pypeclub/OpenPype/blob/700184cfdc2f3dbabb3b49bbc1d9464fd3939a25/openpype/modules/deadline/plugins/publish/submit_publish_job.py#L245) and actually just set:
```python
            if self.deadline_priority:
                priority = self.deadline_priority
            else:
                priority = instance.data.get("priority")
```

And pass that to `priority` on [line 257](https://github.com/pypeclub/OpenPype/blob/700184cfdc2f3dbabb3b49bbc1d9464fd3939a25/openpype/modules/deadline/plugins/publish/submit_publish_job.py#L257).

openpype/settings/defaults/project_settings/deadline.json
```diff
@@ -95,6 +70,33 @@
...
             "department": "",
             "multiprocess": true
+        },
+        "ProcessSubmittedJobOnFarm": {
+            "enabled": true,
+            "deadline_department": "",
+            "deadline_pool": "",
+            "deadline_group": "",
+            "deadline_chunk_size": 1,
+            "deadline_priority": 40,
```
BigRoy, line 80: This default value has been changed from `50` to `40`. Is that intended?

openpype/settings/entities/schemas/projects_schema/schema_project_deadline.json
```diff
@@ -117,6 +117,18 @@
                             "key": "asset_dependencies",
                             "label": "Use Asset dependencies"
                         },
+                        {
+                            "type": "number",
+                            "key": "priority",
+                            "label": "Priority",
+                            "default": 50
```
BigRoy, line 124: Does having default in the schema make any sense if at all times the defaults are loaded anyway from the default settings? I don't recall ever seeing `default` set on objects in the schema.

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
BigRoy: Nah it's fine - just wondered that if we were doing cleanup anyway whether this would also be the moment to clean this up in settings now so that names match with the code better, etc.

Would be good to keep track of this and resolve it for v4 however. Especially since that's a completely new database we can do whatever we want it feels like.

kalisp: I don't think that how Settings are stored/parsed will be much different in v4. 
But I am all up for refactoring and clean up, but we don't have any tests for deadline publishing yet.

BigRoy: > I don't think that how Settings are stored/parsed will be much different in v4.

The only reason not to rename now is that if a studio had a custom value set now that they'd lose their setting and had to re-apply their setting in the new setting attribute. With v4 I imagine with the new database that studios might need to set their settings anyway, OR we'll provide a 'convert v3 -> v4' script in which we can refactor names anyway? Not as nice, but it'd somewhat be a one time thing.

openpype/settings/defaults/project_settings/deadline.json
kalisp: Nope, fixed

openpype/settings/entities/schemas/projects_schema/schema_project_deadline.json
kalisp: Probably doesn't make sense, fixed.

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
kalisp: Fixed

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -235,6 +235,8 @@ def _submit_deadline_post_job(self, instance, job, instances):
             if mongo_url:
                 environment["OPENPYPE_MONGO"] = mongo_url
 
+        priority = self.deadline_priority or instance.data.get("priority")
```
BigRoy, line 238: This could potentially result in a `None` value - however I don't believe you can submit to Deadline with None value as Priority? If None is not a valid value for the submission we might need to do `instance.data.get("priority", 50)`.

Or if instead we believe the priority instance data is a fallback that should always exist I think we should rely on it with `instance.data["priority"]`?

kalisp: Used first variant, as it seems safer.

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
antirotor: I am wondering if we could utilize something like this:

https://github.com/pypeclub/OpenPype/blob/develop/openpype/modules/deadline/abstract_submit_deadline.py

in that file is list of all options deadline job can have. Maybe create Settings widget (entity) to easily add from that list with human readable description so it is unified across the hosts...? I'll create new issue about that to keep track.

BigRoy: Oh yes baby. That sounds lovely.

m-u-r-p-h-y (approved):
The new priority is correctly propagated down the line

from 
project_settings/deadline/publish/MayaSubmitDeadline/priority

to
Maya render instance

to 
Deadline submission


image

PR #2996, opened 2022-04-01
Nuke: removing redundant Ftrack asset when farm publishing

jakubjezek001:
## Brief description
No need to publish sequence with review if baked profiles are used. 

## Description
This was happenning due wild card regex match but since we are adding `bakingNukeScripts` atribute to instance data, it is quite clear we don't need to review also .exr image sequence.

## Additional info
We had also not very nice way of Ftrack asset naming in case multiple reviewable representations were processed. I had hopefully preserved backward compatibility for single reviewable, but it needs to be tested. The feature was made optional and can be activated in settings `project_settings/ftrack/publish/IntegrateFtrackInstance/nicer_asset_name`

## Testing notes:
Ideal test scenarios are following:
- activate `project_settings/ftrack/publish/IntegrateFtrackInstance/nicer_asset_name` on your project
- nuke single baking preset - ftrack component asset name should be in old way (subset name)
- nuke multiple baking presets - asset name should be in new way (each component {subset}_{repre.name}) 
image

# Todo:
- I have noticed we are also generating correct thumbnails for each reviewable, but those are not used anywhere. Anyway `IntegrateFttrackInstances` is not supporting multiple Thumbnails so this could be also improved. 
image
Problem is that each of the representations (noLut, withLut) is having differently looking image and this will be great to display as thumbnail.

### DIFF
--- a/openpype/modules/deadline/plugins/publish/submit_publish_job.py
+++ b/openpype/modules/deadline/plugins/publish/submit_publish_job.py
@@ -528,11 +528,8 @@ class ProcessSubmittedJobOnFarm(pyblish.api.InstancePlugin):
             # preview video rendering
             for app in self.aov_filter.keys():
                 if os.environ.get("AVALON_APP", "") == app:
-                    # no need to add review if baking in nuke present
-                    if (
-                        app == "nuke"
-                        and instance.get("bakingNukeScripts")
-                    ):
+                    # no need to add review if `hasReviewableRepresentations`
+                    if instance.get("hasReviewableRepresentations"):
                         break
 
                     # iteratre all aov filters
--- a/openpype/modules/deadline/plugins/publish/submit_publish_job.py
+++ b/openpype/modules/deadline/plugins/publish/submit_publish_job.py
@@ -509,8 +509,8 @@ class ProcessSubmittedJobOnFarm(pyblish.api.InstancePlugin):
         most cases, but if not - we create representation from each of them.
 
         Arguments:
-            instance (pyblish.plugin.Instance): instance for which we are
-                                                setting representations
+            instance (dict): instance data for which we are
+                             setting representations
             exp_files (list): list of expected files
 
         Returns:
@@ -528,6 +528,14 @@ class ProcessSubmittedJobOnFarm(pyblish.api.InstancePlugin):
             # preview video rendering
             for app in self.aov_filter.keys():
                 if os.environ.get("AVALON_APP", "") == app:
+                    # no need to add review if baking in nuke present
+                    if (
+                        app == "nuke"
+                        and instance.get("bakingNukeScripts")
+                    ):
+                        break
+
+                    # iteratre all aov filters
                     for aov in self.aov_filter[app]:
                         if re.match(
                             aov,
--- a/openpype/modules/ftrack/plugins/publish/integrate_ftrack_instances.py
+++ b/openpype/modules/ftrack/plugins/publish/integrate_ftrack_instances.py
@@ -169,7 +169,8 @@ class IntegrateFtrackInstance(pyblish.api.InstancePlugin):
         # Change asset name of each new component for review
         is_first_review_repre = True
         not_first_components = []
-        extended_asset_name = False
+        extended_asset_name = ""
+        multiple_reviewable = len(review_representations) > 1
         for repre in review_representations:
             # Create copy of base comp item and append it
             review_item = copy.deepcopy(base_component_item)
@@ -178,8 +179,7 @@ class IntegrateFtrackInstance(pyblish.api.InstancePlugin):
             # expand name to better label componenst
             if (
                 not self.keep_first_subset_name_for_review
-                and is_first_review_repre
-                and len(review_representations) > 1
+                and multiple_reviewable
             ):
                 asset_name = review_item["asset_data"]["name"]
                 # define new extended name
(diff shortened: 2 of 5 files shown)


openpype/modules/ftrack/plugins/publish/integrate_ftrack_instances.py
```diff
@@ -168,7 +168,31 @@ def process(self, instance):
...
         is_first_review_repre = True
         not_first_components = []
+        extended_asset_name = False
         for repre in review_representations:
+            # Create copy of base comp item and append it
+            review_item = copy.deepcopy(base_component_item)
+
+            # condition for multiple reviewable representations
+            # expand name to better label componenst
+            if is_first_review_repre and len(review_representations) > 1:
```
iLLiCiTiT, line 178: But we don't want to add representation name for first review. That's the whole logic of `is_first_review_repre` attribute.

64qam (approved):

openpype/modules/ftrack/plugins/publish/integrate_ftrack_instances.py
```diff
@@ -168,7 +168,31 @@ def process(self, instance):
...
+                review_item["asset_data"]["name"] = extended_asset_name
+                # and rename all already created components
+                for _ci in component_list:
+                    _ci["asset_data"]["name"] = extended_asset_name
+
+                # and rename all already created src components
+                for _sci in src_components_to_add:
+                    _sci["asset_data"]["name"] = extended_asset_name
+
+                first_thumbnail_component[
```
iLLiCiTiT, line 193: This will also crash if thumbnail is not available.

openpype/modules/ftrack/plugins/publish/integrate_ftrack_instances.py
jakubjezek001: well it is the point of view. I would like to have those differentiated as it makes sense to me. From your logic it doesn't make sense. Now who is right to decide what is right, right?

iLLiCiTiT: I didn't, @mkolar did.

openpype/modules/ftrack/plugins/publish/integrate_ftrack_instances.py
```diff
@@ -168,7 +168,33 @@ def process(self, instance):
...
+                # and rename all already created components
+                for _ci in component_list:
+                    _ci["asset_data"]["name"] = extended_asset_name
+
+                # and rename all already created src components
+                for _sci in src_components_to_add:
+                    _sci["asset_data"]["name"] = extended_asset_name
+
+                # rename also first thumbnail component if any
+                if first_thumbnail_component is not None:
```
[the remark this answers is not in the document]
jakubjezek001: from the line 161 you can see that it is not

openpype/modules/ftrack/plugins/publish/integrate_ftrack_instances.py
jakubjezek001: I can offer condition with Nuke host only.

mkolar: Unfortunately, this adds a new use case to this functionality, but the use case is totally relevant. Limiting it to one Nuke is not very systematic and will explode in our faces at some point I'm afraid. 

The reason why we have the first review as the main one is to keep some consistency between multiple versions and because these extra ftrack version don't map very well to our database structure. Ftrack doesn't allow for multiple reviewable components (our representations), so we have to hack it and create "related" versions. 

Seeing this new usecase, I'd say it's reasonable to have it explicit in all cases, but I'm worried of the implication it might have if someone does this for a single shot only. Suddenly one shot has a different name format than the rest...

I think that @tokejepsen is using this in production as well. What do you think?

mkolar: :The issue is that you'll loose version progress of the subset on ftrack if you'll add representation name and there is none that has only subset name

I assume you mean, in case where v03 was published with 2 reviewable representations, but v04, suddenly only has 1 reviewable repre?

iLLiCiTiT: Yes, in that case you'll have for example:
- version 3 reviews on `renderCompositingMain_mov` and `renderCompositingMain_mov_lut`
- version 4 review on `renderCompositingMain`

And related version would be empty (or with gaps in versions)
image

jakubjezek001: I will remove my change if you feel it is important to remain it consistently broken. But it is rather step back then forward.

jakubjezek001: Another option would be to let users to control this behaviour from Settings. Would taht be option for you?

openpype/modules/ftrack/plugins/publish/integrate_ftrack_instances.py
```diff
@@ -249,6 +281,14 @@ def process(self, instance):
                 continue
             # Create copy of base comp item and append it
             other_item = copy.deepcopy(base_component_item)
+
+            # add extended name if any
+            if (
+                    self.nicer_asset_name is not False
```
iLLiCiTiT, line 287: I would change the settings key to `keep_first_subset_for_review`.

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -528,6 +528,14 @@ def _get_representations(self, instance, exp_files):
             # preview video rendering
             for app in self.aov_filter.keys():
                 if os.environ.get("AVALON_APP", "") == app:
+                    # no need to add review if baking in nuke present
+                    if (
+                        app == "nuke"
```
iLLiCiTiT, line 533: I would check here for more general key on instance data like `"addPreviewOnSubmit"` which would be set in Nuke's submitter or other plugin, to avoid host specific code in general plugin.

iLLiCiTiT (approved):

PR #2998, opened 2022-04-01
Fix - remove doubled dot in workfile created from template

kalisp:
## Brief description
Newly created workfile contained 2 dots before extension

### DIFF
--- a/openpype/lib/avalon_context.py
+++ b/openpype/lib/avalon_context.py
@@ -1965,6 +1965,7 @@ def get_last_workfile(
         data.pop("comment", None)
         if not data.get("ext"):
             data["ext"] = extensions[0]
+        data["ext"] = data["ext"].replace('.', '')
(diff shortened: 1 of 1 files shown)


64qam (approved):

openpype/lib/avalon_context.py
```diff
@@ -1965,6 +1965,7 @@ def get_last_workfile(
         data.pop("comment", None)
         if not data.get("ext"):
             data["ext"] = extensions[0]
+        data["ext"] = data["ext"].replace('.', '')
```
BigRoy, line 1968: This sounds like a "hack" to fix another issue. The fill data shouldn't have a `.` to begin with.
This logic now breaks usages of `ext` where the actual extension also contains a dot, like `.ass.gz`.

If we'd want to ensure `ext` never starts with a dot, fine with me. We could raise an error so that whenever it happens we can fix the root of the problem where fill data got the wrong information. Right?

PR #3000, opened 2022-04-01
TVPaint: Render scene family

iLLiCiTiT:
## Brief description
Added `renderScene` family which is created during publishing and will create render of whole scene for all visible layers.

## Description
The instance is collected dynamically during publishing and it's result are all visible layers in the scene published as PNG sequences. The plugin is by default disabled. Since the result is almost the same as render layer, it is possible to define layer name (or variant) for subset name template to fill, render pass is filled with `"beauty"`. Instance is disabled if there are other `renderPass` or `renderLayer` instances collected from the scene, the reason is that it can be used as lazy publishing of files from remote artists.

## Testing notes:
1. Open TVPaint without any created instances
2. Run publishing
3. Render scene instance should be created and enabled
4. Publishing should result with image sequence of the file

5. Open TVPaint with created instances
6. Run publishing
7. The render scene instance should be disabled

### DIFF
--- a/openpype/hosts/tvpaint/plugins/publish/collect_instances.py
+++ b/openpype/hosts/tvpaint/plugins/publish/collect_instances.py
@@ -20,21 +20,35 @@ class CollectInstances(pyblish.api.ContextPlugin):
             json.dumps(workfile_instances, indent=4)
         ))
 
+        filtered_instance_data = []
+        # Check if there is any created instance
+        any_created_instance = False
         # Backwards compatibility for workfiles that already have review
         #   instance in metadata.
         review_instance_exist = False
         for instance_data in workfile_instances:
-            if instance_data["family"] == "review":
+            family = instance_data["family"]
+            if family == "review":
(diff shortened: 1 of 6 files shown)


64qam:
Collect Anatomy Instance Data says:
```
Traceback: 
  File "C:\Program Files (x86)\OpenPype\dependencies\pyblish\plugin.py", line 522, in __explicit_process
    runner(*args)
  File "C:\Users\pypeclub\AppData\Local\pypeclub\openpype\openpype-v3.9.2-theline220401\openpype\plugins\publish\collect_anatomy_instance_data.py", line 46, in process
  File "C:\Users\pypeclub\AppData\Local\pypeclub\openpype\openpype-v3.9.2-theline220401\openpype\plugins\publish\collect_anatomy_instance_data.py", line 60, in fill_missing_asset_docs
KeyError: 'asset'
```

64qam (approved):

PR #3002, opened 2022-04-01
Nuke: fixing unicode type detection in effect loaders

jakubjezek001:
## Brief description
Fixing unicode type detection in effect loaders

## Testing notes:
1. open testing nuke workfile in nuke12
2. Load Effect family subset to scene either as Input process or as nodes

### DIFF
--- a/openpype/hosts/nuke/plugins/load/load_effects.py
+++ b/openpype/hosts/nuke/plugins/load/load_effects.py
@@ -1,6 +1,7 @@
 import json
 from collections import OrderedDict
 import nuke
+import six
 
 from avalon import io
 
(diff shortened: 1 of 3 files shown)


BigRoy:
Should this also patch [this one](https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/nuke/plugins/load/load_gizmo_ip.py#L246)?

jakubjezek001:
> Should this also patch [this one](https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/nuke/plugins/load/load_gizmo_ip.py#L246)?
> 
> I think that might've been missed in PR https://github.com/pypeclub/OpenPype/pull/2948

yes you are right, I will add gizmo loader to this pr. Thanks

64qam (approved):

PR #3004, opened 2022-04-01
Hosts: Remove path existence checks in 'add_implementation_envs'

iLLiCiTiT:
## Brief description
Function `add_implementation_envs` should not remove not existing paths because they may be not yet resolved environment value.

## Description
Resolving of environment variable values is finished after collection of all environments. Last step is that host implementation may modify environments value in `add_implementation_envs` function. But in some cases are also verifying existence of paths that may still have unfilled values of other environments. This should not happen at all, it is better to keep the invalid value then  remove it completely which cause confusion.

## Changes
- host's implementation of `add_implementation_envs` does not remove not existing paths from env values

## Testing notes:
1. Hosts should still work
2. It should be possible to use env keys in variables that are modified in host implementation

Resolves https://github.com/pypeclub/OpenPype/issues/3003

### DIFF
--- a/openpype/hosts/blender/__init__.py
+++ b/openpype/hosts/blender/__init__.py
@@ -29,12 +29,12 @@ def add_implementation_envs(env, _app):
         env.get("OPENPYPE_BLENDER_USER_SCRIPTS") or ""
     )
     for path in openpype_blender_user_scripts.split(os.pathsep):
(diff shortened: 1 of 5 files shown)


m-u-r-p-h-y (approved):
tested on Houdini.

successfully passing compound environments to the host

image

PR #3006, opened 2022-04-01
Fix support for Renderman in Maya

antirotor:
## Fix

This is restoring basic support for Renderman in Maya and OpenPype. It is still very basic and has no support for more advanced features like light groups, and so on.

### Testing

Install Renderman for Maya, create simple scene, create render instance, submit to farm.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/validate_rendersettings.py
+++ b/openpype/hosts/maya/plugins/publish/validate_rendersettings.py
@@ -117,13 +117,14 @@ class ValidateRenderSettings(pyblish.api.InstancePlugin):
             cls.log.error("Animation needs to be enabled. Use the same "
                           "frame for start and end to render single frame")
 
-        if renderer != "renderman":
-            if not prefix.lower().startswith(required_prefix):
-                invalid = True
-                cls.log.error(
-                    "Wrong image prefix [ {} ] - doesn't start with: '{}'".format(
-                        prefix, required_prefix)
-                )
+        if renderer != "renderman" and not prefix.lower().startswith(
+                required_prefix):
+            invalid = True
+            cls.log.error(
+                ("Wrong image prefix [ {} ] "
+                 " - doesn't start with: '{}'").format(
+                    prefix, required_prefix)
+            )
 
         if not re.search(cls.R_LAYER_TOKEN, prefix):
             invalid = True
--- a/openpype/hosts/maya/plugins/publish/validate_rendersettings.py
+++ b/openpype/hosts/maya/plugins/publish/validate_rendersettings.py
@@ -130,9 +130,10 @@ class ValidateRenderSettings(pyblish.api.InstancePlugin):
 
         if not prefix.lower().startswith(required_prefix):
             invalid = True
-            cls.log.error("Wrong image prefix [ {} ] - "
-                          "doesn't start with: '{}'".format(
-                prefix, required_prefix))
+            cls.log.error(
+                "Wrong image prefix [ {} ] - doesn't start with: '{}'".format(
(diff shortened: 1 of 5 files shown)


m-u-r-p-h-y (approved):
Publish and submit to deadline works ok.

just to note there is another issue with rendering but it seems more deadline related
Cannot find procedure "rmanCreateAllGlobalsNodes"

BigRoy:
> Cannot find procedure "rmanCreateAllGlobalsNodes"

That sounds more like the deadline machine can't find the Renderman plug-in. Did you happen to have Renderman installed only locally and submitted the render to a different machine than your own?

m-u-r-p-h-y:
It is not an issue of not loading the plugin, but as you said the function injected by Deadline to startup mel script was most probably deprecated by Pixar.

MayaBatch.py ->
` 'if (!`pluginInfo -q -l RenderMan_for_Maya`) {loadPlugin RenderMan_for_Maya;} rmanCreateAllGlobalsNodes();string $rl=""; string $rp=""; string $spoolmode=""; int $chunksize=100; int $rib=' + ('1' if self.Renderer == "rendermanexport" else '0' ) + '; int $ui=0; string $renderer=""; string $globals="";'
`

I already submitted the issue to Thinkbox

openpype/settings/defaults/system_settings/tools.json
```diff
@@ -52,10 +52,39 @@
...
+                "__dynamic_keys_labels__": {
+                    "24-3-maya": "24.3 RFM"
+                }
+            }
+        },
         "__dynamic_keys_labels__": {
             "mtoa": "Autodesk Arnold",
             "vray": "Chaos Group Vray",
-            "yeti": "Pergrine Labs Yeti"
+            "yeti": "Pergrine Labs Yeti",
```
BigRoy, line 86: Not related to this PR but it's Per**e**grine Labs instead of Pergrine:
```suggestion
            "yeti": "Peregrine Labs Yeti",
```

openpype/hosts/maya/plugins/publish/validate_rendersettings.py
```diff
@@ -121,7 +121,8 @@ def get_invalid(cls, instance):
             cls.log.error("Animation needs to be enabled. Use the same "
                           "frame for start and end to render single frame")
 
-        if not prefix.lower().startswith("maya/<scene>"):
+        if not prefix.lower().startswith("maya/<scene>") and \
+                renderer != "renderman":
```
BigRoy, line 125: Shouldn't we still consider it invalid for Renderman too if it doesn't write into `maya/scenes`?

Should it be something like this?

**Consider this untested pseudocode**
```suggestion
        required_prefix = "maya/<scene>"
        if renderer == "renderman":
            # Renderman requires to start with <ws> to write into the workspace
            required_prefix = "<ws>/" + required_prefix

        if not prefix.lower().startswith(required_prefix):
```

BigRoy: Also looking at the [Renderman Documentation](https://rmanwiki.pixar.com/display/RFM22/Workspace) shouldn't this prefix actually also contain [the render folder from the actual workspace](https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/maya/resources/workspace.mel#L11)?

Something like:

`<ws>/renders/maya/<scene>`?

antirotor:
This should now render and publish properly in renderman 24.3

m-u-r-p-h-y (approved):
Tested again after code refactoring.

Maya2022, RenderMan for Maya 24.3

Successfully submitted to Deadline and rendered

image

PR #3008, opened 2022-04-02
Resolve environment variable in google drive credential path 

ClementHector:
## Brief description
Allow resolving a path with an environment variable in gdrive credentials.
For example:
 if OpenPype is installed in `path/to/OP` and setting `project_settings/global/sync_server/sites/gdrive/credentials_url` is `{OPENPYPE_ROOT}/path/to/cred_file.json` will give `/path/to/OP/path/to/cred_file.json`

### DIFF
--- a/openpype/modules/sync_server/providers/gdrive.py
+++ b/openpype/modules/sync_server/providers/gdrive.py
@@ -3,7 +3,7 @@ import os.path
 import time
 import sys
 import six
-import platform
+import acre
 
 from openpype.api import Logger
 from openpype.api import get_system_settings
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
I don't think acre merge should be used for this. I understand that part of the logic is same but acre is meant for environment variables and merges all same values together which may be dangerous in this case (calling the `acre.merge`).

Function which would only parse the dictionary as is and tries to fill env values in, is the way == skip the merge. Something like this:
```
def prepare_platform_values(data, env=None):
    if env is None:
        env = os.environ
    env_copy = env.copy()
    parsed_data = acre.parse(data)
    for key in tuple(parsed_data.keys()):
        value = parsed_data[key]
        try:
            parsed_data[key] = value.format(**env_copy)
        except KeyError, ValueError:
            pass
   return parsed_data
```

BigRoy:
@iLLiCiTiT isn't that just doing `acre.prepare` and `acre.compute`? :)

ClementHector:
> @iLLiCiTiT isn't that just doing `acre.prepare` and `acre.compute`? :)

I ain't find `acre.prepare`

ClementHector:
> I don't think acre merge should be used for this. I understand that part of the logic is same but acre is meant for environment variables and merges all same values together which may be dangerous in this case (calling the `acre.merge`).

I didn't think it was a problem because the envrionment is copy and not reinject at the end. Using accre made sense to me, I will do a simple format.

iLLiCiTiT:
> I ain't find acre.prepare

That's because I wanted to write `acre.parse` :)

> Using accre made sense to me, I will do a simple format.

There is possibility that you'll have environment with name `EXAMPLE` and will have key stored under `EXAMPLE` and they'll be merged using `os.path.sep` which is not expected. In other words the merge does more than should.

kalisp (approved):
Works.

PR #3011, opened 2022-04-05
General: default workfile subset name for workfile

kalisp:
## Brief description
Added default template for subset workfile to '{family}{Task}'. (eg. for TestTask subset would be `workfileTestTask`)

## Description
Standard method `get_subset_name` might not be used for workfiles on all places (it should), but it seems when it will be used, 
it should translate into workfileTaskname format.

### DIFF
--- a/openpype/settings/defaults/project_settings/global.json
+++ b/openpype/settings/defaults/project_settings/global.json
@@ -245,6 +245,15 @@
                 },
                 {
                     "families": [
+                        "workfile"
+                    ],
+                    "hosts": [],
+                    "task_types": [],
+                    "tasks": [],
+                    "template": "{family}{Task}"
+                },
+                {
+                    "families": [
                         "render"
                     ],
                     "hosts": [],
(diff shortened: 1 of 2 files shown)


openpype/settings/defaults/project_settings/global.json
```diff
@@ -243,6 +243,15 @@
...
                     "template": "{family}{variant}"
                 },
+                {
+                    "families": [
+                        "workfile"
+                    ],
+                    "hosts": [],
+                    "task_types": [],
+                    "tasks": [],
+                    "template": "{family}{Task}"
```
mkolar, line 253: shouldn't this rather be `task[name]`?

kalisp: I was thinking about it, but I found that we are still using only {Task} mostly. I think it would be good ideas to create issue to standardize.
(And figure out how to make it backward compatible, as {Task} will be prominent in customer's Settings, I am guessing.)

64qam (approved):

PR #3015, opened 2022-04-05
Deadline: Use more suitable name for sequence review logic

iLLiCiTiT:
## Brief description
Changed key `"hasReviewableRepresentations"` -> `"useSequenceForReview"` to make it more clear for future logic.

## Description
The logic for forcing to not create reviewable from sequence was reversed by checking for `"useSequenceForReview"`. The value must be explicitly set to `False` to not add review tag to the representation. Logic result did not change.

## Additional info
This change was made based on discussion. Result of discussion is that we should probably modify logic of expected files in submit publish job which is limited a lot and can't handle few cases properly.

## Testing notes:
1. Farm publishing of baking nuke scripts should work as did before this PR

### DIFF
--- a/openpype/modules/deadline/plugins/publish/submit_publish_job.py
+++ b/openpype/modules/deadline/plugins/publish/submit_publish_job.py
@@ -524,26 +524,31 @@ class ProcessSubmittedJobOnFarm(pyblish.api.InstancePlugin):
         for collection in collections:
             ext = collection.tail.lstrip(".")
             preview = False
-            # if filtered aov name is found in filename, toggle it for
-            # preview video rendering
-            for app in self.aov_filter.keys():
-                if os.environ.get("AVALON_APP", "") == app:
-                    # no need to add review if `hasReviewableRepresentations`
-                    if instance.get("hasReviewableRepresentations"):
-                        break
+            # TODO 'useSequenceForReview' is temporary solution which does
+            #   not work for 100% of cases. We must be able to tell what
+            #   expected files contains more explicitly and from what
+            #   should be review made.
+            # - "review" tag is never added when is set to 'False'
+            use_sequence_for_review = instance.get(
+                "useSequenceForReview", True
+            )
+            if use_sequence_for_review:
+                # if filtered aov name is found in filename, toggle it for
+                # preview video rendering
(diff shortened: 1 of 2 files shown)


openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -524,26 +524,31 @@ def _get_representations(self, instance, exp_files):
...
+                            if re.match(
+                                aov,
+                                list(collection)[0]
+                            ):
+                                preview = True
+                                break
+
+                # toggle preview on if multipart is on
+                if instance.get("multipartExr", False):
+                    preview = True
```
BigRoy, line 551: Would this logic be more readable when moved to `use_collection_for_review` method?

BigRoy: Here's what the logic could be like:

_consider as pseudocode_
```python
def use_collection_for_review(self, collection, instance):
    
    # TODO 'useSequenceForReview' is temporary solution which does
    #   not work for 100% of cases. We must be able to tell what
    #   expected files contains more explicitly and from what
    #   should be review made.
    # - "review" tag is never added when is set to 'False'
    if not instance.get("useSequenceForReview", True):
        return False
        
    # toggle preview on if multipart is on
    if instance.get("multipartExr", False):
        return True
    
    # if filtered aov name is found in filename, toggle it for
    # preview video rendering
    app = os.environ.get("AVALON_APP", "")
    aov_filters = self.aov_filter.get(app, [])
    for aov_filter in aov_filters:
        first_filename = list(collection)[0]
        if re.match(aov_filter, first_filename):
            return True
```

iLLiCiTiT: The aov part have already existing PR which (I think) eradicate the remaining code into 2 lines and the `useSequenceForReview` is not proper way how to achieve the result so it's not needed. And I would say that in this case it's better that we'll see it everytime we look into the code :)

BigRoy: If it's being solved already elsewhere then ignore my comment. Thanks!

antirotor (approved):

PR #3016, opened 2022-04-05
Console Interpreter: Changed how console splitter size are reused on show

iLLiCiTiT:
## Brief description
Use stored splitter sizes as relative ratio rather then absolute size.

## Description
Restored sizes are now set explicitly on initialization but the size of window is not yet known so the bottom part is squeezed and when sizes would be set as absolute size of first show it may cause issues when machine has more displays with different resolution. The best solution seems to be se restore size of window before first show and change splitter sizes after showing up.

## Testing notes:
1. Open tray > Admin > Console
2. Change sizes of window and splitter parts
3. Restart tray
4. Open tray > Admin > Console again
5. Size of window and splitter parts should be the same, or more close then were before this PR

### DIFF
--- a/openpype/modules/python_console_interpreter/window/widgets.py
+++ b/openpype/modules/python_console_interpreter/window/widgets.py
@@ -648,24 +648,8 @@ class PythonInterpreterWidget(QtWidgets.QWidget):
 
         # Skip if number of size items does not match to splitter
         splitters_count = len(self._widgets_splitter.sizes())
-        if len(splitter_size_ratio) != splitters_count:
-            return
-
-        # Don't use absolute sizes but ratio of last stored sizes
-        ratio_sum = sum(splitter_size_ratio)
-        sizes = []
-        max_size = self._widgets_splitter.height()
-        cur_size = 0
-        ratio = max_size / ratio_sum
-        for size in splitter_size_ratio:
-            item_size = int(ratio * size)
-            cur_size += item_size
-            if cur_size > max_size:
-                item_size -= cur_size - max_size
-            if not item_size:
-                item_size = 1
-            sizes.append(item_size)
-        self._widgets_splitter.setSizes(sizes)
+        if len(splitter_size_ratio) == splitters_count:
+            self._widgets_splitter.setSizes(splitter_size_ratio)
 
     def closeEvent(self, event):
         self.save_registry()
--- a/openpype/modules/python_console_interpreter/window/widgets.py
+++ b/openpype/modules/python_console_interpreter/window/widgets.py
@@ -389,7 +389,8 @@ class PythonInterpreterWidget(QtWidgets.QWidget):
 
         self._append_lines([openpype_art])
 
-        self.setStyleSheet(load_stylesheet())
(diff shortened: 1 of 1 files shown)


openpype/modules/python_console_interpreter/window/widgets.py
```diff
@@ -627,8 +628,45 @@ def add_tab(self, tab_name, index=None):
...
+        if self._first_show:
+            self._first_show = False
+            self._on_first_show()
+
         self._output_widget.scroll_to_bottom()
 
+    def _on_first_show(self):
+        # Change stylesheet
+        self.setStyleSheet(load_stylesheet())
+        # Check if splitter size raio is set
```
BigRoy, line 641:
```suggestion
        # Check if splitter size ratio is set
```

openpype/modules/python_console_interpreter/window/widgets.py
```diff
@@ -627,8 +628,45 @@ def add_tab(self, tab_name, index=None):
...
+        # - first store value to local variable and then unset it
+        splitter_size_ratio = self._splitter_size_ratio
+        self._splitter_size_ratio = None
```
BigRoy, line 644: It does seem weird that we're storing this value to `self` just so it's used this one time on first show and only then. I understand the `init_from_registry()` takes all settings from registry but nonetheless this feels odd.

openpype/modules/python_console_interpreter/window/widgets.py
```diff
@@ -627,8 +628,45 @@ def add_tab(self, tab_name, index=None):
...
+        max_size = self._widgets_splitter.height()
+        cur_size = 0
+        ratio = max_size / ratio_sum
+        for size in splitter_size_ratio:
+            item_size = int(ratio * size)
+            cur_size += item_size
+            if cur_size > max_size:
+                item_size -= cur_size - max_size
+            if not item_size:
+                item_size = 1
```
BigRoy, line 666: Are you sure this is necessary?

I believe that `QSplitter.setSizes()` is always relative sizes right?

The documentation for [`QSplitter.setSizes`](https://doc.qt.io/qt-5/qsplitter.html#setSizes) states:
> any additional/missing space is distributed amongst the widgets according to the relative weight of the sizes.

Which does bring me to "why did it misbehave in the first place"? Maybe that was only down to being triggered before the UI was shown?

iLLiCiTiT: > Which does bring me to "why did it misbehave in the first place"? Maybe that was only down to being triggered before the UI was shown?

Tried and you're right, changed.

openpype/modules/python_console_interpreter/window/widgets.py
iLLiCiTiT: > It does seem weird that we're storing this value to self just so it's used this one time on first show and only then.

I agree but at the same time loading the data from registry twice is weirder. Since splitter sizes must be set after show and window resize must happen before show, it's just decision.

64qam (approved):

PR #3022, opened 2022-04-06
General: Add example addons to ignored

iLLiCiTiT:
## Brief description
Example addons folder is not auto-loaded because is not in ignored filenames variable.

## Description
Added `example_addons` and `default_modules` folders into ignored folders. Also added condition to ignore all openpype module folders that don't have `__init__.py` files in. Side change is removement of imports in ftrack action `action_create_folders.py` which were accidentally added back during merge resolving.

## Testing notes:
1. OpenPype modules should be loaded.
2. Example addons must not be added automatically unless their path is explicitly added in settings.
3. Folders of OpenPype modules containing only pycache files should be skipped.

### DIFF
--- a/openpype/modules/base.py
+++ b/openpype/modules/base.py
@@ -38,6 +38,7 @@ IGNORED_DEFAULT_FILENAMES = (
     "base.py",
     "interfaces.py",
     "example_addons",
+    "default_modules",
 )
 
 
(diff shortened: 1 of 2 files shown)


64qam (approved):

openpype/modules/base.py
```diff
@@ -341,7 +352,16 @@ def _load_modules():
...
+                # Check existence of init fil
+                init_path = os.path.join(fullpath, "__init__.py")
+                if not os.path.exists(init_path):
+                    log.debug((
+                        "Module directory does not contan __init__.py file {}"
+                    ).format(fullpath))
+                    continue
+
+            elif ext not in (".py", ):
                 continue
```
BigRoy, line 365: Why is this logic duplicated from above? Wouldn't these cases already be 'continued' upon?

PR #3029, opened 2022-04-07
General: Fix validate asset docs plug-in filename and class name

BigRoy:
## Brief description
Fix **Validate asset docs** plug-in filename and class name which was invalid [as reported on Discord](https://discord.com/channels/517362899170230292/517382145552154634/961312392711667773).

Plus some minor typos/grammar changes.

This should also avoid a class name conflict with [Validate Containers](https://github.com/pypeclub/OpenPype/blob/2207d3a279a7974d187a43d404861cd68f06c05c/openpype/plugins/publish/validate_containers.py#L17).

## Testing notes:
1. Check whether the plug-in shows up and runs as expected.

### DIFF
--- a/openpype/plugins/publish/validate_asset_docs.py
+++ b/openpype/plugins/publish/validate_asset_docs.py
@@ -3,7 +3,7 @@ from openpype.pipeline import PublishValidationError
 
 
 class ValidateAssetDocs(pyblish.api.InstancePlugin):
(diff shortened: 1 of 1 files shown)


kalisp (approved):
Works, I am just questioning practicality of this validator. If instance doesn't have `assetEntity`, it will probably explode much sooner in some collector. (This is why no one notice it wasn't showing up.)

PR #3036, opened 2022-04-07
Ftrack: Custom attributes can be filled in slate values

iLLiCiTiT:
## Brief description
Extended settings for slate values and added ability to get custom data from ftrack into some template fill data.

## Description
Slate extraction in nuke now have templates where is possible to fill values for Netflix's Nuke slate node. Right now were added 3 keys: `f_submission_note`, `f_submitting_for` and `f_vfx_scope_of_work` (but they can be extended in future). Each input in settings can be disabled, in that case plugin won't change the value of the key at all. Default settings match previous default values.

Ftrack has new collector where can be collected custom attribute values into custom data which can be used in slate or in burnins. Collector has settings where can be specified custom attributes that are collected. Hierarchical attributes are not handled properly.

## Additional info
Added new key `customData` into `instance.data`. There already was key `custom_burnin_data` but it's name is specific for burnins (kept backwards compatibility).

## Testing notes:
1. Create some custom attribute in ftrack (text for `Shot` entity type)
2. Choose one entity with the custom attribute and fill it's value
3. Enable `project_settings/nuke/publish/ExtractSlateFrame/f_vfx_scope_of_work` in settings
4. Fill `f_vfx_scope_of_work` template with `{custom[ftrack][<your custom attribute key]}`
5. Publish something to chosen entity from Nuke which would add the slate
6. The same template value can be used in ExtractBurnin plugin


64qam:
The custom attribute is not filling in to the slate

64qam (approved):

PR #3038, opened 2022-04-08
Maya:  disable rendering of all lights render setup deadline

### LINKED ISSUE #2241: Maya: Disable rendering of all lights - Render Setup / Deadline
## Feature

Normally Maya will render all enabled lights in scene. If - for some reason - you want to selectively render lights based on render setup settings, you can do it by manually setting this option in render setup or using `MAYA_RENDER_SETUP_INCLUDE_ALL_LIGHTS` set to `0`.

However Deadline will always override this environment variable unless job includes `RenderSetupIncludeLights` set to zero in **PluginInfo** part of job definition.

Having this set per scene is useful so I propose to add it to render instance settings.

Related #2215 

[cuID:1rw2mw7]

Allan-I:
## Brief description
Fixes #2241 by adding flag "RenderSetupIncludeLights" to project settings.

## Description
Disables rendering all lights in the scene to conform with render layers.

## Additional info
Appends "RenderSetupIncludeLights" to Deadline payload template and appending the value assigned in the project settings through `instance.data`


m-u-r-p-h-y (changes requested):
did some initial testing and it seems the new settings are not propagated in Maya
none of the settings is changing the Render setup settings

image

Allan-I:
> did some initial testing and it seems the new settings are not propagated in Maya none of the settings is changing the Render setup settings
> 
> image

Thank you, @m-u-r-p-h-y! 

The project setting in its current form overrides the Maya option (not taking it into account nor does it modify that value in the local file it applies the change to), it merely applies it to the Deadline payload, will adjust it shortly to reflect the change within the Maya file as well as some naming changes.

However, one usability decision needs to be made about whether we want one of the settings to have control over the other, meaning do we need the project settings to be changing the Maya setting or vice versa?  Or both ways?  I'm not sure.

I'd appreciate any usability design choice you could provide for this.

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -266,6 +266,7 @@ def _submit_deadline_post_job(self, instance, job, instances):
                 "Version": self.plugin_pype_version,
                 "Arguments": " ".join(args),
                 "SingleFrameOnly": "True",
+                "RenderSetupIncludeLights": instance.data.get("RenderSetupIncludeLights")  # noqa
```
BigRoy, line 269: This key has no functionality on the submit publish job plug-in I think. This should be in Submit Maya Deadline, right? I feel like Deadline wouldn't get this value currently.

openpype/hosts/maya/plugins/create/create_render.py
```diff
@@ -199,7 +204,13 @@ def process(self):
...
                 renderer = 'renderman'
 
-            self._set_default_renderer_settings(renderer)
+        if self.instance.data.get("RenderSetupIncludeLights"):
+            from maya.mel import eval
+            eval('optionVar -iv "renderSetup_includeAllLights" 1')
+        if not self.instance.data.get("RenderSetupIncludeLights"):
+            from maya.mel import eval
+            eval('optionVar -iv "renderSetup_includeAllLights" 0')
+        self._set_default_renderer_settings(renderer)
```
BigRoy, line 213: Be aware that this is **not** a scene-wide setting but a local machine preference. Another user opening the same scene will not have this set correctly.

If we're trying to set Include All Lights on a project wide level I'd really urge and recommend to use the environment variable `MAYA_RENDER_SETUP_INCLUDE_ALL_LIGHTS` instead.

We've had it before and it's super confusing if one opens a scene on another machine and it renders differently due to this option var. So I'd recommend always using the `MAYA_RENDER_SETUP_INCLUDE_ALL_LIGHTS` environment variable to force a preferred setting in the project.

openpype/modules/deadline/plugins/publish/submit_publish_job.py
Allan-I: Thank you @BigRoy I've done a couple of test renders seems like it goes through!

openpype/hosts/maya/plugins/create/create_render.py
Allan-I: Hi! @BigRoy yes thank you, I'm coming from your thread on chaos's website discussing this, made the adjustment because this would make the option local, I'm figuring out what best way to deal with this

Allan-I: @antirotor what do you think?

openpype/modules/deadline/plugins/publish/submit_publish_job.py
BigRoy: > I've done a couple of test renders seems like it goes through!

I have a feeling that if it works that it's not necessarily due to this line of code but maybe due to it existing in the [payload skeleton in Submit Maya Deadline](https://github.com/pypeclub/OpenPype/blob/6dcd0ae47ed1baa2b5e11c97ce560948374bb499/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py#L65) here.

Could you describe how you're detecting whether the value went through?

Are you able to see it on the Maya Render Job in deadline in **Modify Job Properties..** in `Submission Params` in `Plugin Info Parameters` as `RenderSetupIncludeLights=False`? And when the setting is toggled in Maya does it show `RenderSetupIncludeLights=True`?

Allan-I: I'll elaborate, before adding the functionality to toggle the option on and off (which is actually the broken bit here), the flag was being propagated to the payload, yes, and it was going through into the submission params, but now that I'm trying to have it toggle the setting it's erroring out for me, which I'll remove from that commit i made and see if it works again.

openpype/hosts/maya/plugins/create/create_render.py
m-u-r-p-h-y: There are two points which need to be done.

1. Properly set Deadline submission parameters to pass the correct settings to render job to ensure consistency in rendering, based on the way we decide below
2. Finding the way to control the behavior as Maya stores this setting locally, which can lead to different behavior on different machines.
- the easiest way is to set the ENV variable in Maya Application settings, but this is defining the behavior system-wide
- I'm not sure if there is any mechanism to add per project ENV variable overrides to Maya in OP and/or how difficult it would be to add it. I would personally prefer to do it in this generic way, instead of creating buttons to set every single thing in OP settings.

Example:
image


This way we'd be giving the user freedom to set this and potentially other things on a project level.

openpype/modules/deadline/plugins/publish/submit_publish_job.py
Allan-I: Hi Roy, you were right, for some reason I was remembering that this was working in that commit and I don't really understand why, it should indeed be in `Submit Maya Deadline` as part of the payload skeleton.  Now I can see the flag in there in the payload, testing for a layer.

Allan-I:
I just confirmed it works for turning on and off (the flag, I mean).  Created a couple of renders, one with flag on and off, and indeed got one light when the flag was off and two lights when it was on.

  
We just have to align on what we need to be doing for this in terms of toggling.  The function that toggles that option in Maya render setup returns a `None` type, so that would be useless in terms of it toggling the setting off itself, we could potentially add our own button in there that toggles that off I think.

openpype/hosts/maya/plugins/create/create_render.py
```diff
@@ -108,6 +111,8 @@ def __init__(self, *args, **kwargs):
         self._project_settings = get_project_settings(
             legacy_io.Session["AVALON_PROJECT"])
 
+
+
```
BigRoy, line 115: Isn't this too many empty lines? Why isn't the hound reacting?
```suggestion
```

openpype/hosts/maya/plugins/create/create_render.py
```diff
@@ -108,6 +111,8 @@ def __init__(self, *args, **kwargs):
...
         # project_settings/maya/create/CreateRender/aov_separator
```
BigRoy, line 116: Isn't this too many empty lines? Why isn't the hound reacting?
```suggestion
            legacy_io.Session["AVALON_PROJECT"])

        # project_settings/maya/create/CreateRender/aov_separator
```

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
```diff
@@ -29,7 +29,8 @@
 
 import clique
 import requests
-
+from openpype.api import (
+    get_current_project_settings)
```
BigRoy, line 33: Doesn't most other code in OP format it like this?
```suggestion
from openpype.api import (
    get_current_project_settings
)
```

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
```diff
@@ -500,7 +502,7 @@ def process(self, instance):
 
         self.payload_skeleton["JobInfo"]["Comment"] = comment
         self.payload_skeleton["PluginInfo"]["RenderLayer"] = renderlayer
-
+        self.payload_skeleton["PluginInfo"]["RenderSetupIncludeLights"] = get_current_project_settings()["maya"]["create"]["CreateRender"]["disable_all_lights"] # noqa
```
BigRoy, line 505: Shouldn't this use `context.data["project_settings"]` like [here](https://github.com/pypeclub/OpenPype/blob/8b8269a1324045faf134aafd6d654bb0c4fe9a0d/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py#L332)?

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -270,6 +270,7 @@ def _submit_deadline_post_job(self, instance, job, instances):
                 "Version": self.plugin_pype_version,
                 "Arguments": " ".join(args),
                 "SingleFrameOnly": "True",
+                "RenderSetupIncludeLights": None  # noqa
```
BigRoy, line 273: I still think this line shouldn't be here at all. `RenderSetupIncludeLights` isn't a Plugin setting for the Deadline OpenPype Plugin.
```suggestion
```

openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_create.json
```diff
@@ -63,6 +63,11 @@
                     "type": "text",
                     "key": "default_render_image_folder",
                     "label": "Default render image folder"
+                },
+                {
+                    "type": "boolean",
+                    "key": "disable_all_lights",
+                    "label": "Disable Rendering of All Lights"
```
BigRoy, line 70: I feel like this needs a clarification that this is the case for "Render Setup Layers"

With [this comment taken into consideration](https://github.com/pypeclub/OpenPype/pull/3038/files#r861015724) I feel like this could be labeled or documented: "Include all lights in Render Setup Layers by default" so it's very close to what Maya shows in the [menu entry as visible in screenshot in this comment](https://github.com/pypeclub/OpenPype/pull/3038#pullrequestreview-954451610).

openpype/hosts/maya/plugins/create/create_render.py
```diff
@@ -72,6 +72,9 @@ class CreateRender(plugin.Creator):
     # renderSetup instance
     _rs = None
 
+    # Disable "render all lights" flag for the scene.
+    disable_all_lights = True
```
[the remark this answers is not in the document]
Allan-I: Thank you Roy, indeed you are right and if you look in the history I did fix that but I had to do a hard reset because of some issues, was going to fix that but I went for a break, fixing it in a minute.

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
Allan-I: For some reason my linter panicked when I did that, fixing it.

openpype/hosts/maya/plugins/create/create_render.py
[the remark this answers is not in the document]
Allan-I: No worries!  I appreciate your time very much, I am indeed forgetful and this helps a ton!

openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_create.json
Allan-I: Done!

openpype/modules/deadline/plugins/publish/submit_publish_job.py
Allan-I: Fixed!

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
Allan-I: I kinda like the explicit one better that just plops it down from that explicit function tbh.

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
Allan-I: Done!

openpype/hosts/maya/plugins/create/create_render.py
Allan-I: Fixed!

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
mkolar: We should keep to the settings that have already been collected, rather than calling for them again, that's exactly why we collect them to the context. Keep in mind that all of these are db calls.

Allan-I: Oh my god you're absolutely right, fixing that now!

Allan-I: Done!

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
```diff
@@ -503,7 +503,13 @@ def process(self, instance):
...
         self.payload_skeleton["JobInfo"]["Comment"] = comment
         self.payload_skeleton["PluginInfo"]["RenderLayer"] = renderlayer
-        self.payload_skeleton["PluginInfo"]["RenderSetupIncludeLights"] = get_current_project_settings()["maya"]["create"]["CreateRender"]["enable_all_lights"] # noqa
+        self.payload_skeleton["PluginInfo"]["RenderSetupIncludeLights"] = (
+            context.data["project_settings"].get(
+                "maya", {}).get(
+                "create", {}).get(
+                "CreateRender", {}).get(
+                "enable_all_lights", {})
+        )
```
BigRoy, line 509: A bit off topic.

Anyone else feels like we should have a library function that allows to quickly to basically do this nested retrieval and fall back on a default value in a more readable way:

```python
get(settings, "maya.create.CreateRender.enable_all_lights", default=None)
get(settings, "maya", "create", "CreateRender", "enable_all_lights", default=None)
```

mkolar: >Anyone else feels like we should have a library function that allows to quickly to basically do this nested retrieval and fall back on a default value in a more readable way:

Topic for another PR, but in general I kinda agree.

antirotor (changes requested):
I am not sure it is working correctly:

this is set in Maya:
image

this is result in Deadline:
image
 is it possible we are injecting the value only from the settings and ignoring whats on instance?

Allan-I:
> I am not sure it is working correctly:
> 
> this is set in Maya: image
> 
> this is result in Deadline: image is it possible we are injecting the value only from the settings and ignoring whats on instance?

Hi Ondrej, you are absolutely right that's exactly what's being done right now, it's only propagating the values form the settings and not taking into account the instance value.  How would you like that to be approached?

antirotor:
It still doens't work as I expected. Even with turned off "render setup include lights" on the instance, I still get `RenderSetupIncludeLight=True` in Deadline job.

image

@m-u-r-p-h-y can you confirm or is my setup somehow broken (because it might)

openpype/hosts/maya/plugins/publish/validate_rendersettings.py
```diff
@@ -239,6 +239,15 @@ def get_invalid(cls, instance):
...
 
+        settings_lights_flag = instance.context.data["project_settings"].get(
+            "maya", {}).get(
+            "create", {}).get(
+            "CreateRender", {}).get(
+            "enable_all_lights", {})
+
+        instance_lights_flag = instance.data.get("RenderSetupIncludeLights")
+        if settings_lights_flag != instance_lights_flag:
+            cls.log.warning("Warning: render settings differ from instance")
```
antirotor, line 250: I thinks this message should also explain why.

Allan-I: This should work now, was missing variable collection from the collector

antirotor:
btw even if it doesn't work for me I still think the message mentioned there should be fixed. This said, I am passing it to @m-u-r-p-h-y  at least until I'll fix a new project for testing.

openpype/hosts/maya/plugins/publish/validate_rendersettings.py
```diff
@@ -243,6 +243,15 @@ def get_invalid(cls, instance):
...
 
+        settings_lights_flag = instance.context.data["project_settings"].get(
+            "maya", {}).get(
+            "create", {}).get(
+            "CreateRender", {}).get(
+            "enable_all_lights", {})
+
+        instance_lights_flag = instance.data.get("RenderSetupIncludeLights")
+        if settings_lights_flag != instance_lights_flag:
+            cls.log.warning("Instance flag is set to {0} and Settings flag is set to {1}".format(instance_lights_flag, settings_lights_flag)) # noqa
```
antirotor, line 254:
```suggestion
            cls.log.warning('Instance flag for "Render Setup Include Lights" is set to {0} and Settings flag is set to {1}'.format(instance_lights_flag, settings_lights_flag)) # noqa
```

Allan-I: Thank you Ondrej, committing now!

m-u-r-p-h-y (changes requested):
same issue here 
image

openpype/hosts/maya/plugins/publish/validate_rendersettings.py
```diff
@@ -243,6 +243,15 @@ def get_invalid(cls, instance):
...
 
+        settings_lights_flag = instance.context.data["project_settings"].get(
+            "maya", {}).get(
+            "create", {}).get(
+            "CreateRender", {}).get(
+            "enable_all_lights", {})
+
+        instance_lights_flag = instance.data.get("RenderSetupIncludeLights")
+        if settings_lights_flag != instance_lights_flag:
+            cls.log.warning('Instance flag for "Render Setup Include Lights" is set to {0} and Settings flag is set to {1}'.format(instance_lights_flag, settings_lights_flag)) # noqa
```
BigRoy, line 254: Somehow I feel it would be clearer if this would explicitly mention "Project Settings" instead of "Settings"?

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
```diff
@@ -504,7 +504,13 @@ def process(self, instance):
...
         self.payload_skeleton["JobInfo"]["Comment"] = comment
         self.payload_skeleton["PluginInfo"]["RenderLayer"] = renderlayer
-
+        self.payload_skeleton["PluginInfo"]["RenderSetupIncludeLights"] = (
+            context.data["project_settings"].get(
+                "maya", {}).get(
+                "create", {}).get(
+                "CreateRender", {}).get(
+                "enable_all_lights", {})
+        )
```
BigRoy, line 513: This is the issue that others are reporting where the value in Deadline is incorrect - it'll always use the project settings value instead of the value on the instance. 

Instead it should likely be `instance.data.get("RenderSetupIncludeLights")` like [here](https://github.com/pypeclub/OpenPype/pull/3038/files#diff-b331d4b66a42158b9abaf95dab978dcc14b362e492941dc40c40e3ce82e9aa31R252).

Note that this instance data key does seem to deviate from most others I have seen being stored as pyblish instance data - I don't recall them starting with upper case characters before - I'd have expected `renderSetupIncludeLights`?

BigRoy (changes requested):

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
antirotor: agree, all keys should start with lower letter.

Allan-I: Thanks guys, will address both points right now.

BigRoy:
Did this get deleted by accident? Or was this removed in favor of something else?

Allan-I:
Way too many merge conflicts to waste time on, I'm creating another branch and PR

PR #3042, opened 2022-04-08
LibraryLoader: Use current project for asset query in families filter

iLLiCiTiT:
## Brief description
Family filter now query asset from current project instead of project selected in library loader.

## Description
Library loader queried asset document which defines family filters from the project defined in library loader instead of from current project which caused missing asset document bug.

## Testing notes:
1. You have to have at least one library project
2. Open host in other project
3. Open Library loader in the host
4. Select a library project
5. Tool should not crash due to missing asset document

### Source traceback
```
#   File ".\openpype\openpype-v3.9.1\openpype\tools\libraryloader\app.py", line 278, in on_project_change
#     self._refresh_assets()
#   File ".\openpype\openpype-v3.9.1\openpype\tools\libraryloader\app.py", line 343, in _refresh_assets
#     self._families_filter_view.refresh()
#   File ".\openpype\openpype-v3.9.1\openpype\tools\loader\widgets.py", line 1047, in refresh
#     self._family_model.refresh()
#   File ".\openpype\openpype-v3.9.1\openpype\tools\loader\widgets.py", line 943, in refresh
#     family_config = self.family_config_cache.family_config(family)
#   File ".\openpype\openpype-v3.9.1\openpype\tools\utils\lib.py", line 354, in family_config
#     self._refresh()
#   File ".\openpype\openpype-v3.9.1\openpype\tools\utils\lib.py", line 417, in _refresh
#     tasks_info = asset_doc.get("data", {}).get("tasks") or {}
# AttributeError: 'NoneType' object has no attribute 'get'
```

### DIFF
--- a/openpype/tools/utils/lib.py
+++ b/openpype/tools/utils/lib.py
@@ -409,6 +409,7 @@ class FamilyConfigCache:
         project_name = os.environ.get("AVALON_PROJECT")
         asset_name = os.environ.get("AVALON_ASSET")
         task_name = os.environ.get("AVALON_TASK")
+        host_name = os.environ.get("AVALON_APP")
         if not all((project_name, asset_name, task_name)):
             return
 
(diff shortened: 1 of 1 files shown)


BigRoy:
This doesn't sound like the right way to do this? 

For the Library Loader it shouldn't at all rely on `os.environ` because that is the current local project, but not the library project?

Instead it should be getting them from `dbcon.Session`?

iLLiCiTiT:
> This doesn't sound like the right way to do this?

The filtering works based on your current context, not on selected context in UI, so it is actually the right way :)

BigRoy:
> The filtering works based on your current context, not on selected context in UI, so it is actually the right way :)

Well, yes and no. (I might need more coffee so please correct me again if I'm wrong!)

Say I'm browsing project "Library X" and I'm working in "Project Y".
I'm checking out library project "Library X" asset "AssetX".

It will then try to find that "AssetX" name in "Project Y". And likely won't find an asset with a name match... and thus won't be able to retrieve an asset type either. So it's still wrong?

---

It sounds like what you want is to use the Profiles settings from Project Y but sample the data from Library X so your icons and filters match your current project and not from the library project? Which might get us in all some kind of edge case but I can see why you might want that. In that way if Project X would never require "renders" you can e.g. always filter them by default too in the Library Loader. That makes sense.

kalisp (approved):
Fixes the exception.

PR #3045, opened 2022-04-08
General: Local overrides for environment variables

iLLiCiTiT:
## Brief description
Give ability to define white list of environment variables that can be changed per machine.

## Description
In `system_settings/general` was added new key `local_env_white_list` where is possible to define environment variable keys that can be changed per machine. Values of the environment variable keys can be set in Local settings UI. Values are applied on application launch and only if launch environments contain the keys.

## Additional info
These environment variable changes are not applied on OpenPype launch but only on application launch.

### Questions
1. Not sure about the placement of the env white list in the `system_settings/general/local_env_white_list`, it could be maybe added under applications as it's related to them?

## Documentation
TODO modify documentation based on answer of question 1

## Testing notes:
1. Add a env key into `system_settings/general/local_env_white_list` (e.g. `"TEST"`)
2. Choose one application (ideally DCC with script editor) and and env settings add the env key with some value (e.g. `"TEST": "OriginValue"`)
3. Open local settings and set value of the key to somethig different (e.g. `"MyValue"`)
4. Launch the application and check the value

For more complex test also add tool and use the environment key in other key as value.
5. Choose one tool and add there different key using the key from step 1 (e.g.: "SUBKEY_TEST": "This is {TEST}"`)
6. Make sure the tool is added to asset on which you're testing
7. Launch the application and check value of the key from step 5 (`"SUBKEY_TEST"` should have `"MyValue"`)

When key is removed from `system_settings/general/local_env_white_list` it should not be overriden even if is still available in local settings.

### Screenshots
image
image
image

### DIFF
--- a/openpype/lib/applications.py
+++ b/openpype/lib/applications.py
@@ -1399,8 +1399,27 @@ def prepare_app_environments(data, env_group=None, implementation_envs=True):
 
     app = data["app"]
     log = data["log"]
+    source_env = data["env"].copy()
 
-    _add_python_version_paths(app, data["env"], log)
+    _add_python_version_paths(app, source_env, log)
+
+    # Use environments from local settings
+    filtered_local_envs = {}
+    system_settings = data["system_settings"]
+    whitelist_envs = system_settings["general"].get("local_env_white_list")
+    if whitelist_envs:
+        local_settings = get_local_settings()
+        local_envs = local_settings.get("environments") or {}
+        filtered_local_envs = {
+            key: value
+            for key, value in local_envs.items()
+            if key in whitelist_envs
+        }
+
+    # Apply local environment variables for already existing values
+    for key, value in filtered_local_envs.items():
+        if key in source_env:
+            source_env[key] = value
 
     # `added_env_keys` has debug purpose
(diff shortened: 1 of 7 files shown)


mkolar:
Will this require new binaries?

iLLiCiTiT:
> Will this require new binaries?

Nope

antirotor (approved):
It works for me perfectly. I would leave the whitelist where it is as you can use it even for global variables, not only application specific - for example overriding some variable storing your license server or whatever.

iLLiCiTiT:
> I would leave the whitelist where it is as you can use it even for global variables

That's the issue. You can't use it for general variables now.

mkolar:
> That's the issue. You can't use it for general variables now.

Yeah we've had that conversation in the office a few time now. Not sure where is the reasonable line to draw here. I was also suggesting ability to override any variable, albeit, that can be totally deadly to OP

iLLiCiTiT:
I've added ability to change general environments too, only condition is that the key must be available in settings of general environments (Still not require new binaries).

pberto:
@iLLiCiTiT good one.

PR #3046, opened 2022-04-08
Houdini: New Publisher

### LINKED ISSUE #2960: Convert Houdini to new publisher
## Feature

We should convert Houdini creators to new publisher infrastructure before more features are added to ease transition. This involves creation of host based creator class that will be used as super class for rest of the creators.

antirotor:
## Enhancement

New publisher implementation for Houdini. All creators were changed to new system of publishing.

### Details

Instance metadata is stored on output nodes in **Extra** folder the same way as before, Context metadata are stored on hidden node **OpenPypeContext**.

Collected instances are available under `houdini_cached_subsets` for all creators in collected shared data under their respective creator identifiers, legacy subsets are cached in the same place under `houdini_cached_legacy_subsets` under their respective families. Convertor can convert existing legacy subsets to the new one by adding creator identifiers mapped to families, but artist still needs to set task on converted instances in the UI.

### Testing

All creators should work as expected and publishing should produce expected outputs. Legacy subsets in old scene should be converted on demand.

Close #2960

> **Note**
> This PR includes #4020, #3995 and #3972


openpype/hosts/houdini/api/lib.py
```diff
@@ -397,8 +401,20 @@ def read(node):
...
-    return {parameter.name(): parameter.eval() for
-            parameter in node.spareParms()}
+    data = {}
+    for parameter in node.spareParms():
+        value = parameter.eval()
+        # test if value is json encoded dict
+        if isinstance(value, six.string_types) and \
+                len(value) > 0 and value.startswith("JSON:::"):
+            try:
+                value = json.loads(value.lstrip("JSON:::"))
```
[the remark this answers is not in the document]
BigRoy:
```suggestion
                value = json.loads(value[len("JSON:::"):]))
```
Should likely be this?

openpype/hosts/houdini/api/lib.py
```diff
@@ -460,3 +476,60 @@ def reset_framerange():
     hou.playbar.setFrameRange(frame_start, frame_end)
     hou.playbar.setPlaybackRange(frame_start, frame_end)
     hou.setFrame(frame_start)
+
+
+def load_creator_code_to_asset(
+        otl_file_path, node_type_name, source_file_path):
```
BigRoy, line 482: Can we please add a docstring to describe what this function does?

openpype/hosts/houdini/api/lib.py
```diff
@@ -460,3 +476,60 @@ def reset_framerange():
...
+
+
+def get_main_window():
+    """Acquire Houdini's main window"""
+    if self._parent is None:
+        self._parent = hou.ui.mainQtWindow()
+    return self._parent
+
+
+def parm_to_template(key, value):
```
BigRoy, line 506: Function name feels a bit confusing - it's not actually turning a `hou.Parm` into a template. It's take a parameter name and a default value and for that define a corresponding parm template.

Options/Ideas:
- `parm_template_from_value(name, value)`
- `auto_parm_template(name, value)`

openpype/hosts/houdini/api/lib.py
```diff
@@ -397,8 +401,20 @@ def read(node):
...
     """
     # `spareParms` returns a tuple of hou.Parm objects
-    return {parameter.name(): parameter.eval() for
-            parameter in node.spareParms()}
+    data = {}
+    for parameter in node.spareParms():
+        value = parameter.eval()
+        # test if value is json encoded dict
+        if isinstance(value, six.string_types) and \
+                len(value) > 0 and value.startswith("JSON:::"):
```
BigRoy, line 409:
```suggestion
        if isinstance(value, six.string_types) and value.startswith("JSON:::"):
```
No need to check length, right?

Also shouldn't the `"JSON:::"` prefix be hardcoded elsewhere? Feels a bit arbitrary having this hardcoded here?

openpype/hosts/houdini/api/pipeline.py
```diff
@@ -28,78 +30,113 @@
...
-        on_file_event_callback
-    )
+    def update_context_data(self, data, changes):
+        root_node = hou.node("/")
+        lib.imprint(root_node, data)
+
+    def get_context_data(self):
+        from pprint import pformat
+
+        self.log.debug(f"----" + pformat(lib.read(hou.node("/"))))
```
BigRoy, line 138:
```suggestion
```
Remove debugging at some point

openpype/hosts/houdini/plugins/create/create_pointcache.py
```diff
@@ -1,48 +1,45 @@
...
+            "use_sop_path": True,
+            "build_from_path": True,
+            "path_attrib": "path",
             "prim_to_detail_pattern": "cbId",
-            "format": 2,  # Set format to Ogawa
-            "facesets": 0,  # No face sets (by default exclude them)
-            "filename": "$HIP/pyblish/%s.abc" % self.name,
+            "format": 2,
+            "facesets": 0,
+            "filename": "$HIP/pyblish/{}.abc".format(self.identifier)
```
BigRoy, line 32: This shouldn't be the identifier but the unique subset's name so that two Pointcache instances don't write to the same `pointcache.abc` file, right?

openpype/hosts/houdini/api/lib.py
```diff
@@ -293,46 +293,50 @@ def imprint(node, data):
...
     Returns:
         None
 
     """
+    if not data:
+        return
 
-    parm_group = node.parmTemplateGroup()
+    current_parameters = node.spareParms()
+    current_keys = [p.name() for p in current_parameters]
```
BigRoy, line 308:
```suggestion
    current_parms = {parm.name(): parm for parm in node.spareParms()]
```

openpype/hosts/houdini/api/lib.py
```diff
@@ -293,46 +293,50 @@ def imprint(node, data):
...
     node.setParmTemplateGroup(parm_group)
 
+    if update_keys:
+        parms = node.parmTuplesInFolder(("Extra",))
+        for parm in parms:
+            for key, value in update_keys:
+                if parm.name() == key:
+                    node.replaceSpareParmTuple(
+                        parm.name(), parm_to_template(key, value))
+
```
BigRoy, line 339: Why can't we just use the existing collected parms by name from above but instead need to look into `node.parmTuplesInFolder(("Extra",))` here?

BigRoy (commented):
Some early notes

openpype/hosts/houdini/api/lib.py
antirotor: this was from the early attempts. Idea was to have creators as nodes and use Houdini to execute creator python code on them. It was unnecessary complicated and ... useless :) so this is now removed.

openpype/hosts/houdini/api/lib.py
antirotor: yeah, I've optimized all of it a little bit. Still wondering wheter update should be imprint or if it should be moved to separate function.

antirotor:
This now works on single creator - **Create PointCache**. Publishing fails because more work has to be done on collectors, but it can serve as a base to rewrite other existing creators.

✏️ **Note**: It won't correctly display creator attributes in Python 2 variant of Houdini (#3782)

BigRoy:
Wanted to quickly test this PR. Getting this error when "resetting" in the UI after creating a pointcache instance I get this error:
```python
  File "S:\openpype\OpenPype\openpype\tools\publisher\window.py", line 368, in _on_reset_clicked
    self.controller.reset()
  File "S:\openpype\OpenPype\openpype\tools\publisher\control.py", line 578, in reset
    self.save_changes()
  File "S:\openpype\OpenPype\openpype\tools\publisher\control.py", line 717, in save_changes
    self.create_context.save_changes()
  File "S:\openpype\OpenPype\openpype\pipeline\create\context.py", line 1121, in save_changes
    self._save_context_changes()
  File "S:\openpype\OpenPype\openpype\pipeline\create\context.py", line 1129, in _save_context_changes
    self.host.update_context_data(data, changes)
  File "S:\openpype\OpenPype\openpype\hosts\houdini\api\pipeline.py", line 137, in update_context_data
    lib.imprint(root_node, data)
  File "S:\openpype\OpenPype\openpype\hosts\houdini\api\lib.py", line 345, in imprint
    node.setParmTemplateGroup(parm_group)
  File "C:/PROGRA~1/SIDEEF~1/HOUDIN~1.303/houdini/python3.7libs\hou.py", line 15725, in setParmTemplateGroup
    return _hou.Node_setParmTemplateGroup(self, parm_template_group, rename_conflicting_parms)
hou.OperationFailed: The attempted operation failed.
Spare parameters are not alowed on the root node.
```
Re-opening the UI resolved it and then I was able to reset as expected.

antirotor:
> Spare parameters are not alowed on the root node

That at least answers question about storing context data on the root node. We need to find better place.

antirotor:
also I am wondering if this #3674 applies here too. I don't think so, because in new publisher all those parameters must have a value, but I need to think about it some more.

BigRoy:
> That at least answers question about storing context data on the root node. We need to find better place

Potentially [Houdini Variables](https://www.sidefx.com/docs/houdini/ref/windows/alias.html#variables). Those are per scene, get saved along and clear out when opening new scene. So it's persistent data with the scene. 

- Edit > Aliases and Variables _(Alt + Shift + V)_

Like in this image:
IMG_3225

However, as you can see it is quite accessible to the user and variables become even accessible in Houdini nodes/wrangles, etc. So might not be what we want?

---

Otherwise this trick might do it to store it on the scene root.
```python
hou.node(“/”).setUserData(“foo”, “bar”) 
```
[Source](https://www.sidefx.com/forum/topic/42767/?page=1#post-192166)

Instead of visible parameters store as user data that would just be hidden user data I believe.

openpype/hosts/houdini/api/pipeline.py
```diff
@@ -35,71 +40,138 @@
...
+
+    def has_unsaved_changes(self):
+        return hou.hipFile.hasUnsavedChanges()
+
+    def get_workfile_extensions(self):
+        return [".hip", ".hiplc", ".hipnc"]
+
+    def save_workfile(self, dst_path=None):
+        # Force forwards slashes to avoid segfault
+        filepath = dst_path.replace("\\", "/")
```
BigRoy, line 94: `dst_path=None` default wouldn't work because you can't perform `.replace` on None type.

BigRoy: Also feels weird that the argument is `dst_path` but `open_workfile` uses `filepath` as argument.

Should this maybe be `dest_filepath` or just `filepath`?

openpype/hosts/houdini/api/plugin.py
```diff
@@ -84,3 +90,109 @@ def process(self):
                 OpenPypeCreatorError,
                 OpenPypeCreatorError("Creator error: {}".format(er)),
                 sys.exc_info()[2])
+
+
+@six.add_metaclass(ABCMeta)
```
iLLiCiTiT, line 95: There is no need to add `ABCMeta` it is already in `NewCreator`.

openpype/hosts/houdini/api/pipeline.py
```diff
@@ -371,3 +428,35 @@ def main_take(no_update=True):
             instance_node.bypass(not new_value)
     except hou.PermissionError as exc:
         log.warning("%s - %s", instance_node.path(), exc)
+
+
+def list_instances(creator_id=None):
```
iLLiCiTiT, line 433: It's used only in `HoudiniCreator` I would move it there (with `remove_instance`).

openpype/hosts/houdini/api/plugin.py
```diff
@@ -84,3 +90,109 @@ def process(self):
...
+                self.family,
+                subset_name,
+                instance_data,
+                self)
+            self._add_instance_to_context(instance)
+            imprint(instance_node, instance.data_to_store())
+            return instance
+
+        except hou.Error as er:
+            six.reraise(
```
iLLiCiTiT, line 149: This is done to keep traceback? Clever...

openpype/hosts/houdini/hooks/set_operators_path.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+from openpype.lib import PreLaunchHook
+import os
+
+
+class SetOperatorsPath(PreLaunchHook):
```
iLLiCiTiT, line 6: Shouldn't this happen in `add_implementation_envs` inside addon?

BigRoy:
I have two questions.

First:
- Does this PR mean that any existing scenes with old creators would now not work anymore since the creator identifiers have changed? That seems like quite a breaking change, no?

Second:
afbeelding

It feels to me that **members** and **instance_id** don't seem like data that should persist in the scene. These values are "collected on the fly" in the Pyblish window and for e.g. the Camera instances aren't actually the data that should be user-settable as an attribute like that.

---

For what it's worth, in houdini if we were ever to set members to node paths - why wouldn't the attribute be an Operator List parameter type? Should be much cleaner, no?

iLLiCiTiT:
> t feels to me that members and instance_id don't seem like data that should persist in the scene. These values are "collected on the fly" in the Publisher window and for e.g. the Camera instances aren't actually the data that should be user-settable as an attribute like that.

Members could be created on the fly but `instance_id` is not. Instance id must persist on refresh or in any other process.

BigRoy:
> Members could be created on the fly but instance_id is not.

Wouldn't the unique identifier in the scene be the node path in Houdini instead of that identifier? (It's just as unique?)
Especially because it seems the `OpenPypeContext` or global data gets no `unique_identifier`. What's the reason we need this unique identifier to be that string attribute? Why could the Creator be able to return its own unique identifier for an instance, which could then be any string value?

BigRoy:
By the way, noticed a bug. If I change **Validate Containers** on the global Options then it persists that setting on the `/obj/OpenPypeContext` object. However, if I then reset the setting to its default value the `publish_attributes` will remain to be set to False. It seems that whenever `publish_attributes` was ever generated at least once on `/obj/OpenPypeContext` then changing the settings afterwards will not persist.

Also, the _"Instance with id {} is already added to context."_ is logged so many times that I almost feel it's not worth being a warning log. With three instances in my scene this is the log on resetting the Publisher:

```
Instance with id 8acc2afe-9f8b-496c-b617-94180ebc183d is already added to context.
Instance with id 3b8282d9-17dc-45b1-be96-5d2d95298a2e is already added to context.
Instance with id 1be068f3-4a3a-46f3-8e40-3eda511dd6a5 is already added to context.
Instance with id 8acc2afe-9f8b-496c-b617-94180ebc183d is already added to context.
Instance with id 3b8282d9-17dc-45b1-be96-5d2d95298a2e is already added to context.
Instance with id 1be068f3-4a3a-46f3-8e40-3eda511dd6a5 is already added to context.
Instance with id 8acc2afe-9f8b-496c-b617-94180ebc183d is already added to context.
Instance with id 3b8282d9-17dc-45b1-be96-5d2d95298a2e is already added to context.
Instance with id 1be068f3-4a3a-46f3-8e40-3eda511dd6a5 is already added to context.
Instance with id 8acc2afe-9f8b-496c-b617-94180ebc183d is already added to context.
Instance with id 3b8282d9-17dc-45b1-be96-5d2d95298a2e is already added to context.
Instance with id 1be068f3-4a3a-46f3-8e40-3eda511dd6a5 is already added to context.
Instance with id 8acc2afe-9f8b-496c-b617-94180ebc183d is already added to context.
Instance with id 3b8282d9-17dc-45b1-be96-5d2d95298a2e is already added to context.
Instance with id 1be068f3-4a3a-46f3-8e40-3eda511dd6a5 is already added to context.
Instance with id 8acc2afe-9f8b-496c-b617-94180ebc183d is already added to context.
Instance with id 3b8282d9-17dc-45b1-be96-5d2d95298a2e is already added to context.
Instance with id 1be068f3-4a3a-46f3-8e40-3eda511dd6a5 is already added to context.
Instance with id 8acc2afe-9f8b-496c-b617-94180ebc183d is already added to context.
Instance with id 3b8282d9-17dc-45b1-be96-5d2d95298a2e is already added to context.
Instance with id 1be068f3-4a3a-46f3-8e40-3eda511dd6a5 is already added to context.
Instance with id 8acc2afe-9f8b-496c-b617-94180ebc183d is already added to context.
Instance with id 3b8282d9-17dc-45b1-be96-5d2d95298a2e is already added to context.
Instance with id 1be068f3-4a3a-46f3-8e40-3eda511dd6a5 is already added to context.
Instance with id 8acc2afe-9f8b-496c-b617-94180ebc183d is already added to context.
Instance with id 3b8282d9-17dc-45b1-be96-5d2d95298a2e is already added to context.
Instance with id 1be068f3-4a3a-46f3-8e40-3eda511dd6a5 is already added to context.
```
Which seems an awfully lot - and a lot of duplicates too?

iLLiCiTiT:
> Does this PR mean that any existing scenes with old creators would now not work anymore since the creator identifiers have changed? That seems like quite a breaking change, no?

If it does not then it can be changed. Collection phase of creators gives option to use backwards compatibile approach.

> Wouldn't the unique identifier in the scene be the node path in Houdini instead of that identifier? (It's just as unique?)
Especially because it seems the OpenPypeContext or global data gets no unique_identifier. What's the reason we need this unique identifier to be that string attribute? Why could the Creator be able to return its own unique identifier for an instance, which could then be any string value?

There must be some unique identifier which is not host specific so you can point to it without knowing the host (or creator's) logic underhood. Also it helps to work with instances in publisher tool on change or refresh. If you want to do something with instance you don't have to know how it's unique in terms of creator but just know it's id and no matter when, you can have access to it event after restart or in different process.

> Which seems an awfully lot - and a lot of duplicates too?

That's an easy fix.

antirotor:
> I have two questions.
I am currently dealing with both these questions :) There should definitely be backwards compatibility with already created instances.

`members` were just transient attribute I want to get rid of as it should be determined on-the-fly by creators. It's just that it was easier to make current plugins work until I'll figure better way. For most of the current creators memeber == instance node itself so they can add that directly (without being it exposed as visible attribute to the artist). For others it should be collected by its collectors.

openpype/hosts/houdini/api/pipeline.py
```diff
@@ -35,71 +40,139 @@
...
+
+        sys.path.append(hou_pythonpath)
+
+        # Set asset settings for the empty scene directly after launch of
+        # Houdini so it initializes into the correct scene FPS,
+        # Frame Range, etc.
+        # TODO: make sure this doesn't trigger when
+        #       opening with last workfile.
+        _set_context_settings()
+        shelves.generate_shelves()
```
BigRoy, line 84: Somehow I can't start Houdini with this line present. As soon as I remove this `shelves.generate_shelves` then it launches fine. With this line in the codebase Houdini just appears to hang during launch on the splash screen with the console only saying:
```
Installing OpenPype ...
Setting scene FPS to 25
```
And I have no idea why this happening.
Tested with Houdini 19.5.303 Py 3.7 on Windows.

Seems odd no?

iLLiCiTiT: There is strange logic, from my point of view. It crashes with `FileNotFoundError` instead of warning log - I don't think shelves are so crutial to crash. Maybe that's the issue.

There is in default settings empty shelf item which could cause triggering the error
image

BigRoy: It didn't crash. It just didn't launch. Note that I was using Houdini core!

openpype/hosts/houdini/plugins/create/create_vbd_cache.py
```diff
@@ -1,38 +1,36 @@
...
+"""Creator plugin for creating VDB Caches."""
 from openpype.hosts.houdini.api import plugin
+from openpype.pipeline import CreatedInstance
 
 
-class CreateVDBCache(plugin.Creator):
+class CreateVDBCache(plugin.HoudiniCreator):
     """OpenVDB from Geometry ROP"""
-
+    identifier = "io.openpype.creators.houdini.vdbcache"
```
iLLiCiTiT, line 9: Why `io`?

openpype/hosts/houdini/api/plugin.py
```diff
@@ -84,3 +90,109 @@ def process(self):
...
+        parent_node = hou.node(parent)
+        instance_node = parent_node.createNode(
+            node_type, node_name=node_name)
+        instance_node.moveToGoodPosition()
+        return instance_node
+
+    def create(self, subset_name, instance_data, pre_create_data):
+        try:
+            if pre_create_data.get("use_selection"):
+                self.selected_nodes = hou.selectedNodes()
```
iLLiCiTiT, line 124:
```suggestion
            if pre_create_data.get("use_selection"):
                self.selected_nodes = hou.selectedNodes()
            else:
                self.selected_nodes = []
```

openpype/hosts/houdini/plugins/create/create_composite.py
```diff
@@ -1,44 +1,36 @@
...
     icon = "gears"
 
-    def __init__(self, *args, **kwargs):
-        super(CreateCompositeSequence, self).__init__(*args, **kwargs)
+    def create(self, subset_name, instance_data, pre_create_data):
+        import hou  # noqa
 
-        # Remove the active, we are checking the bypass flag of the nodes
-        self.data.pop("active", None)
+        instance_data.pop("active", None)
```
iLLiCiTiT, line 18: Is there a reason to expect `"active"` key in instance data?

openpype/hosts/houdini/api/plugin.py
jakubjezek001: Also, shouldn't there be error raised - since `use selection` is ON and no selection is active?

openpype/hosts/houdini/api/plugin.py
```diff
@@ -84,3 +90,109 @@ def process(self):
...
+    def update_instances(self, update_list):
+        for created_inst, _changes in update_list:
+            instance_node = hou.node(created_inst.get("instance_node"))
+            current_data = read(instance_node)
+
+            imprint(
+                instance_node,
+                {
+                    key: value[1] for key, value in _changes.items()
+                    if current_data.get(key) != value[1]
```
jakubjezek001, line 185:
```suggestion
                    if current_data.get(key) != value[0]
```

jakubjezek001: I was implementing at Nuke and found out index was working this way rather then to `1`

openpype/hosts/houdini/plugins/create/create_hda.py
```diff
@@ -1,28 +1,22 @@
...
     label = "Houdini Digital Asset (Hda)"
     family = "hda"
     icon = "gears"
     maintain_selection = False
 
-    def __init__(self, *args, **kwargs):
-        super(CreateHDA, self).__init__(*args, **kwargs)
-        self.data.pop("active", None)
-
     def _check_existing(self, subset_name):
```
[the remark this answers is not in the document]
BigRoy: Yes, looks like old creator code.

openpype/hosts/houdini/api/plugin.py
BigRoy: @jakubjezek001 That should be wrong. This should actually read:

```python
                {
                    key: new_value for key, (old_value, new_value) in _changes.items()
                    if current_data.get(key) != new_value
                }
```

But might be more readable by not doing a dict comprehension, something like:
```python
updates = {}
for key, (old_value, new_value) in _changes:
    if key not in current_data or new_value != current_data[key]:
        updates[key] = new_value
        
imprint(instance_nodes, updates, update=True)
```

openpype/hosts/houdini/plugins/create/create_composite.py
antirotor: That is part of the old creator stuff. Active was reflecting the state of the Houdini node but that can be handled better now.

openpype/hosts/houdini/plugins/create/create_hda.py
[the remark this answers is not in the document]
antirotor: this is one family that wasn't converted yet as it needs different approach than other creators

openpype/hosts/houdini/api/plugin.py
```diff
@@ -84,3 +90,116 @@ def process(self):
...
+        """
+        for name in parameters:
+            try:
+                parm = node.parm(name)
+                parm.lock(True)
+            except AttributeError:
+                self.log.debug("missing lock pattern {}".format(name))
+
+    def collect_instances(self):
+        cached_instances = self.collection_shared_data.get(
```
iLLiCiTiT, line 170: It was meant more like this:
```
if "houdini_cached_instances" not in self.collection_shared_data:
    instances_by_identifier = collections.defaultdict(list)
    for instance in all_instances:
        identifier = instance.get("creator_identifier")
        if identifier:
            instances_by_identifier[identifier].append(instance)

    self.collection_shared_data["houdini_cached_instances"] = (
        instances_by_identifier
    )

instances = (
    self.collection_shared_data
    ["houdini_cached_instances"]
    .get(self.identifier)
) or []
for instance in instances:
    created_instance = CreatedInstance.from_existing(
        read(instance), self
    )
    self._add_instance_to_context(created_instance)
```
So the `list_instances` is called only once.

antirotor: isn't it doing exactly the same? In pseudo code:

```
get cached instances for given creator
if nothing instances:
    query the scene
     fill the cache
```

iLLiCiTiT: With current implementation you can skip the caching completelly. Each creator will call `list_instances` and store only instances for him -> that does not need caching at all.

antirotor: now I am little lost. So how do I get the list of instances for `collect_instances()`? Is it done automatically now?

openpype/hosts/houdini/api/plugin.py
antirotor: I think it is more readable to be explicit - that by simple look at the code you know you shouldn't use that class directly.

openpype/hosts/houdini/api/plugin.py
iLLiCiTiT: The shared data are meant as "Shared data" across multiple creators for `collect_instances`. If each creator will cache data only for the creator it is pointless.

The goal is for example to collect all nodes in scene with `id` which is `pyblish.avalon.instance` -> store these to shared data so other creators don't have to collect the nodes again (instead of going through 1.000.000 nodes every time go through them only once and store relevant nodes for other creators).

m-u-r-p-h-y:
I have some trouble testing the new publisher

```
  File "D:\REPO\OpenPype\openpype\tools\experimental_tools\dialog.py", line 171, in _on_btn_trigger
    tool.execute()
  File "D:\REPO\OpenPype\openpype\tools\experimental_tools\tools_def.py", line 61, in execute
    self.callback(*args, **kwargs)
  File "D:\REPO\OpenPype\openpype\tools\experimental_tools\tools_def.py", line 169, in _show_publisher
    self._publisher_tool = publisher.PublisherWindow(
AttributeError: module 'openpype.tools.publisher' has no attribute 'PublisherWindow'
```

In old publisher it complains about `No registered families`

image


Houdini 19.5.403 Aprentice

antirotor:
> I have some trouble testing the new publisher as it is not opening:
> 
> ```
> Traceback (most recent call last):
>   File "D:\REPO\OpenPype\openpype\tools\experimental_tools\dialog.py", line 171, in _on_btn_trigger
>     tool.execute()
>   File "D:\REPO\OpenPype\openpype\tools\experimental_tools\tools_def.py", line 61, in execute
>     self.callback(*args, **kwargs)
>   File "D:\REPO\OpenPype\openpype\tools\experimental_tools\tools_def.py", line 169, in _show_publisher
>     self._publisher_tool = publisher.PublisherWindow(
> AttributeError: module 'openpype.tools.publisher' has no attribute 'PublisherWindow'
> ```

This was fixed yesterday in develop. I've merged it back in here so it should work.
 
> In old publisher it complains about `No registered families`

This is right, there are no families for the old publisher. Old menu is there only because there might be someone with old creators still.

mkolar:
I'm able to publish some basic families but not a whole lot of success overall. Even though I'm not sure if those are issues with me publishing wrong stuff, or problem with the new publisher. 

nevertheless, when I delete an instance from the publisher, it's corresponding data is left in the scene, shouldn't we be cleaning up after ourselves a bit?

mkolar (approved):

PR #3050, opened 2022-04-11
General: Extract Review handle invalid characters for ffmpeg

iLLiCiTiT:
## Brief description
Extract review prepare input exrs with invalid characters for ffmpeg.

## Description
FFmpeg can't read exr input if it's metadata attributes contain invalid characters, from that moment ffmpeg can't read any information about input (like width, height, ...) and will crash on using it as source. Added check for invalid characters (similar to string length check) which can cause that attribute is removed before ffmpeg processing. Right now the invalid chars contain only `"` but will be probably extended over time. Another issue was when data from input could not be read by ffmpeg the temporary staging dir was not set back to previous directory so integration failed because could not find source exrs.

## Testing notes:
1. Publish something that will produce exrs with `"` in attributes (e.g. cryptomatte adds json string)
2. Extract review should not skip processing of the file and if do skip it then should not crash on integration because of missing file

### DIFF
--- a/openpype/lib/transcoding.py
+++ b/openpype/lib/transcoding.py
@@ -422,7 +422,12 @@ def convert_for_ffmpeg(
         compression = "none"
 
     # Prepare subprocess arguments
-    oiio_cmd = [get_oiio_tools_path()]
+    oiio_cmd = [
+        get_oiio_tools_path(),
+
+        # Don't add any additional attributes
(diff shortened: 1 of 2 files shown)


m-u-r-p-h-y (approved):
manually recreated the flow of conversions and I can confirm ffmpeg is processing the intermediate oiio exrs without a problem now.

image

PR #3051, opened 2022-04-12
Deadline: reworked pools assignment

kalisp:
## Brief description
Previous solution was nonconsistent, expected values for pools in multiple places in Setting.

## Description
New collector and validator was introduced. Collector adds pool names from Setting (merged and removed separate pool settings in hosts sections) if not already present on a instance (as Maya has it, potentially others with New Publisher).

Validator checks if name of pools on the instances matches to current values on DL (does a query of DL).

## Additional info
Refactored `_requests_get` and `requests_post` wrappers into `openpype.lib.connections` - name up for commenting/changing.

## Testing notes:
1. configure `project_settings/deadline/publish/CollectDeadlinePools`
1.a - insert non existent values
1.b - insert values matching situation on DL
3. publish on DL and check pools

### DIFF
--- a/openpype/modules/deadline/deadline_module.py
+++ b/openpype/modules/deadline/deadline_module.py
@@ -3,8 +3,7 @@ import requests
 import six
 import sys
 
-
-from openpype.lib import requests_get
+from openpype.lib import requests_get, PypeLogger
 from openpype.modules import OpenPypeModule
 from openpype_interfaces import IPluginPaths
 
@@ -59,9 +58,7 @@ class DeadlineModule(OpenPypeModule, IPluginPaths):
 
         """
         if not log:
-            from openpype.lib import PypeLogger
-
-            log = PypeLogger().get_logger(__name__)
+            log = PypeLogger.get_logger(__name__)
 
         argument = "{}/api/pools?NamesOnly=true".format(webservice)
         try:
--- a/openpype/modules/deadline/deadline_module.py
+++ b/openpype/modules/deadline/deadline_module.py
@@ -1,6 +1,18 @@
 import os
+import requests
+import six
+import sys
+
+
+from openpype.lib import requests_get
 from openpype.modules import OpenPypeModule
 from openpype_interfaces import IPluginPaths
+
+
+class DeadlineWebserviceError(Exception):
+    """
+    Exception to throw when connection to Deadline server fails.
+    """
 
 
 class DeadlineModule(OpenPypeModule, IPluginPaths):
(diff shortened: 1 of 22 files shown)


m-u-r-p-h-y:
- Maya render instance (and all render instances in other hosts) should check the settings and prefill the pools and groups from the settings
- Pools and Groups should not be a drop-down, but text fields as we can combine multiple groups to fine-tune specific scene requirements. Ex. I need to submit my heavy scene to group Maya, Ram256 and GPU to render on machines with Maya installed, with more than 256 GB RAM and GPU card)




I would suggest to add pools and group to Nuke write instance (the same as we have in Maya)

image

m-u-r-p-h-y:
there is also an issue in collect render layers (I'm not sure if it is connected to this PR though)

```
// pyblish.CollectMayaRender : collecting layer: rs_Main // 
    runner(*args)
  File "D:\REPO\OpenPype\openpype\hosts\maya\plugins\publish\collect_render.py", line 391, in process
KeyError: 'SecondaryPool'
    runner(*args)
  File "<string>", line 391, in process
KeyError: 'SecondaryPool' // 
b'Processing Context Label'
```

openpype/modules/deadline/deadline_module.py
```diff
@@ -32,3 +44,37 @@ def get_plugin_paths(self):
...
+            webservice (str): Server url.
+            log (Logger)
+        Returns:
+            list: Pools.
+        Throws:
+            RuntimeError: If deadline webservice is unreachable.
+
+        """
+        if not log:
+            from openpype.lib import PypeLogger
```
iLLiCiTiT, line 62: Move the import to the top of file

openpype/modules/deadline/deadline_module.py
```diff
@@ -32,3 +44,37 @@ def get_plugin_paths(self):
...
+
+            log = PypeLogger().get_logger(__name__)
```
iLLiCiTiT, line 64: Initialization of `PypeLogger` is not needed.
```suggestion
            log = PypeLogger.get_logger(__name__)
```

kalisp:
`collect_render` issue was resolved.

m-u-r-p-h-y (approved):
project pool is correctly propagated to render instance 

image

BigRoy:
I'd expect users might want to set the Pool per project and per application, so that e.g. Maya runs in a different pool than Nuke or Harmony. Thoughts?

m-u-r-p-h-y:
You are correct. I've already created issue #3057  to do it outside of this PR.

kalisp:
@BigRoy I thought that pools are per project and groups would be per host.
But maybe we should add there profile for host, task name?

PR #3053, opened 2022-04-12
TVPaint: Added init file for worker to triggers missing sound file dialog

iLLiCiTiT:
## Brief description
TVPaint worker will open TVPaint init file which has missing sound file so all missing sound files can be ignored at the start of worker.

## Description
This dialog can be triggered randomly at any time on worker which is "not nice" as someone has to connect to worker machine and check "Ingore missing dependency" and hit "ok". Because of that the dialog can be shown at the start of worker so for rest of running process are missing sound files ignored.

## Testing notes:
1. Run TVPaint worker
2. Dialog showing missing sound file should show at the start


tokejepsen:
We have the issue on occasion as well when rendering on Deadline.

Have you found any validations we could maybe check before sending it off to render?

Also am I right in thinking that the following is the workflow for getting around this?

- open a temp TVP file that has missing sounds dependencies
- force close the TVP file
- now all TVP files opened after this will ignore the missing sound dependencies

iLLiCiTiT:
> Have you found any validations we could maybe check before sending it off to render?

Only way that could work is to remove all audio references or in theory could work to read the file as binary, find loaded audio, remove it and store the binary content back :) (not tested).

> Also am I right in thinking that the following is the workflow for getting around this?

Yes, that's the goal.

64qam (approved):

PR #3055, opened 2022-04-12
Update collect_render.py

tokejepsen:
Better error message for non selected render camera edge case.

Capture

Previously error message would read `no file names were generated, this is bug`, which does not help fix the issue.


tokejepsen:
Did wonder whether this should be a validator or something, but its a bit of a "chicken and egg" situation i guess.

BigRoy:
⚠️ 

This PR introduced a **BUG** where a renderlayer with a single camera and no camera tokens would never be allowed to get collected. That's because the camera only gets appended if the render product had a camera token in the file prefix.

PR #3060, opened 2022-04-13
Nuke: Add aov matching even for remainder and prerender

kalisp:
## Brief description
Previously filtering used only for collections and 'render' family.

## Description
This should allow more control when to add 'review' tag to representation.

## Testing notes:
1. Create `prerender` write node in Nuke
2. configure `project_settings/deadline/publish/ProcessSubmittedJobOnFarm/aov_filter`
2a. configure that pattern matches created subset name
2b. configure that pattern doesn't match created subset name
4. publish
5. check Ftrack

### DIFF
--- a/openpype/lib/profiles_filtering.py
+++ b/openpype/lib/profiles_filtering.py
@@ -44,12 +44,6 @@ def _profile_exclusion(matching_profiles, logger):
     Returns:
         dict: Most matching profile.
     """
-
-    logger.info(
-        "Search for first most matching profile in match order:"
-        " Host name -> Task name -> Family."
-    )
-
     if not matching_profiles:
         return None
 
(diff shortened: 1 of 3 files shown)


BigRoy:
I'd actually recommend doing `clique.assemble` with an explicit frames pattern and `minimum_items=1`. It'd allow less types of naming conventions to pass for collections but I feel like they'd still match all that we wanted to support with OpenPype. The new integrator is actually already using a [refactored way of `clique.assemble`](https://github.com/pypeclub/OpenPype/blob/7fa224a022e6892b8fd9a1759082741ac52aad3d/openpype/plugins/publish/integrate_new.py#L20-L52). It's more explicit and avoids errors. It however does only allow filenames rendered as `filename_whatever.{frame}.{ext}` to be assembled as a collection and won't allow e.g. `filename_v{frame}.{ext}` or `filename{frame}.{ext}`.

Without this rendering a single frame (that still has a frame number) would currently not be assembled as a sequence.

64qam (approved):

kalisp:
Lets test it with minimum_items=1 in next PR, this one seems to be working and it is needed to get release out.

PR #3063, opened 2022-04-14
Flame: extract presets distribution

jakubjezek001:
## Brief description
Added filtering option to extract presets

## Description
Some presets should be activated only if path to clip or segment is containing particular regex expression or string value. Setting presets can be now deactivated - this will add benefits on project overriding. Settings preset ui had been redesigned for better readability.  The logic for adding parsed attributes from segment's comment had been switched.

## Additional info
- removing dependency on reel clips, because individual clips are now imported on demand
- added option to settings for deactivating `IntegrateBatchGroup` plugin


openpype/hosts/flame/plugins/publish/extract_subset_resources.py
```diff
@@ -131,161 +165,157 @@ def process(self, instance):
...
+            preset_orig_xml_path = str(os.path.join(
+                preset_dir, preset_file
+            ))
+
+            preset_path = opfapi.modify_preset_file(
+                preset_orig_xml_path, staging_dir, modify_xml_data)
+
+            # define kwargs based on preset type
+            if "thumbnail" in unique_name:
+                export_kwargs["thumb_frame_number"] = in_mark + (
```
iLLiCiTiT, line 239: This will create float number, is it ok?

openpype/hosts/flame/plugins/publish/extract_subset_resources.py
```diff
@@ -131,161 +165,157 @@ def process(self, instance):
...
+                export_dir_path, files, extension)
+
+            # fix representation in case of nested folders
+            if n_stage_dir:
+                representation_data["stagingDir"] = n_stage_dir
+                files = n_files
+
+            # add files to represetation but add
+            # imagesequence as list
+            if (
```
iLLiCiTiT, line 285: I think don't we support to have multiple files in `representation["files"]` (like 2 and more .move files).
So this could be simplified to:
```
if len(files) == 1:
    representationdata["files"] = files[0]
else:
    representationdata["files"] = files
```

openpype/hosts/flame/plugins/publish/extract_subset_resources.py
```diff
@@ -131,161 +165,157 @@ def process(self, instance):
...
+            # create preset path
+            preset_orig_xml_path = str(os.path.join(
+                preset_dir, preset_file
+            ))
+
+            preset_path = opfapi.modify_preset_file(
+                preset_orig_xml_path, staging_dir, modify_xml_data)
+
+            # define kwargs based on preset type
+            if "thumbnail" in unique_name:
```
iLLiCiTiT, line 238: Should the `unique_name` be explicitly `"thumbnail"` or any name with `"thumbnail"` in name?
Simplified question: Should this happen if `unique_name` is `"thumbnail_source"`?

iLLiCiTiT (approved):
Added few comments, please read them. But it looks ok (Can't test).

openpype/hosts/flame/plugins/publish/extract_subset_resources.py
jakubjezek001: Yeah there could be a case where `thumbnail_source` can happen - for example, when we support multiple thumbnails in the future.

PR #3070, opened 2022-04-19
General: Add global log verbose arguments

iLLiCiTiT:
## Brief description
Added ability to turn on/off mongo logging and handle global arguments for verbose and debug.

## Description
Logging to mongo can be tuend on/off in system settings, change is propagated after restart. Added `--verbose` argument expecting verbose level int [0-50] or string matching python log level enum, value is stored in env variable `OPENPYPE_LOG_LEVEL`. With verbose argument seems `--debug` argument deprecated but there are usages for it so it was also added global flag argument `--debug` which affects `OPENPYPE_DEBUG` value.

## Additional info
This is partially fix and partially enhancement. Current usage of `--debug` argument is not propagated in current process because is handled too late.

## Testing notes:
- change logging to mongo in settings and check if is propagated
- use all new global arguments
    - verbose
    - debug

Resolves https://github.com/pypeclub/OpenPype/issues/2961

### DIFF
--- a/openpype/cli.py
+++ b/openpype/cli.py
@@ -20,6 +20,10 @@ from .pype_commands import PypeCommands
                     "to list staging versions."))
 @click.option("--validate-version", expose_value=False,
               help="validate given version integrity")
+@click.option("--debug", is_flag=True, expose_value=False,
+              help=("Enable debug"))
+@click.option("--verbose", expose_value=False,
+              help=("Change OpenPype log level (debug - critical or 0-50)"))
 def main(ctx):
     """Pype is main command serving as entry point to pipeline system.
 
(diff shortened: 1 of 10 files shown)


kalisp:
I didn't expect line with 'DEBUG' when running `openpype_console --verbose info`.
Screenshot 2022-04-21 113811

And I think it should be added to documentation, probably here https://openpype.io/docs/admin_use

kalisp:
Actually I wouldn't expect DEBUG lines in console of external python process (for Photoshop for example) with no OPENPYPE_DEBUG.
Screenshot 2022-04-21 115143

Not sure if it is because of this PR or it was there before.

iLLiCiTiT:
> I didn't expect line with 'DEBUG' when running openpype_console --verbose info.

The ftrack debug is actually print, changed.

> And I think it should be added to documentation, probably here https://openpype.io/docs/admin_use

Good point, will do.

> Actually I wouldn't expect DEBUG lines in console of external python process (for Photoshop for example) with no OPENPYPE_DEBUG.

I'm changing log level of openpype logger without affecting other python loggers. I'm not sure if we want to set all python loggers to debug...

kalisp (approved):

PR #2643, opened 2022-02-03
Maya: Refactor/Cleanup Create Render, Render Settigns and Render Products

BigRoy:
## Brief description

As #2640 PR just opened I thought it'd be good to also open this as a draft PR. Some of the things in this PR somewhat align with changes implemented in that PR.

What this PR does:
- Don't force separator from OP settings into the filepath, but actually retrieve it from the render settigns in the current scene. This would fix #2636
- Work towards `lib_renderproducts` detecting the `aov_separator` so we could add a validator for it.
- Cleanup/simplify CreateRender logic - by removing some duplicated code, using available library functions, moving logic closer together, etc.
- Creates functionality for "RenderSettings" separately from Create Render - because I'd like to also be able to 'set the render defaults' without requiring to create a render instance. We tend to have multiple renderers used on a project in a scene, and each of those layers would have to get the defaults applied correctly. I'd personally opt for having it removed from `CreateRender` and pushed somewhere else. E.g. add it to the top menu for artists `OpenPype > Set Default Render settings for current renderer` and also add a Validator that actually ensures the output is valid for the current project.

@antirotor - it could be worth investigating how we could either join the effort, or have one step back and push the other step first with a higher priority. The issues we're facing with rendering currently are getting in the way of our current production so would love to tackle this somewhat soon and actually fix and validate our production cases. :)

PR #3071, opened 2022-04-19
Maya: render creator has configurable options 

Allan-I:
## Brief description
Adds configurable render options in the settings.

## Description
Adds schema specified render options to be triggered through the settings instead of hard-coded render options in the render creator plugins.

## Additional info
The intent is to have the configurable options grabbed through a function that's within the render creator class, and that will append the settings by being triggered within the render creator function that sets the renderer settings.


antirotor:
So the plan is:

- [ ] take code from #2643 as a base
- [ ] move setting schemas that define render settings from creator to its own category called `Render Settings` under Maya
- [ ] add menu item to OpenPype menu in Maya called `Set Render Settings` that will take whatever is set in the Settings and apply it
- [ ] add checkbox to `Create Render` in the Settings called `Apply Render Settings on creation` that if checked will apply render settings when Render instance is created
- [ ] connect all of that to render settings validator and its repair action, so repair action would do the same thing as `Set Render Settings` in OpenPype menu
- [ ] make validate render settings optional (or add ability to configure it so in the Settings)

openpype/hosts/maya/api/lib_renderproducts.py
```diff
@@ -80,6 +80,13 @@
 }
 
 
+def has_tokens(string, tokens):
+    """Return whether any of tokens is in input string (case-insensitive)"""
+    pattern = "({})".format("|".join(re.escape(token) for token in tokens))
```
BigRoy, line 85: I believe this doesn't need to be grouped and the pattern can just be:
```suggestion
    pattern = "|".join(re.escape(token) for token in tokens)
```
```python
pattern = "|".join(re.escape(token) for token in tokens)
```

BigRoy: Also, is this function unused? :)

Allan-I: Hello Roy, thanks very much, I'm not entirely sure about this to be honest, will check it out later and let you know!

Allan-I: Ah, seems like it's unused, will check with the guys perhaps it's of use later on.

PR #3074, opened 2022-04-20
Fix resolution sync

ClementHector:
Check if FPS is not "" and if resolution contain "x" before update data.


BigRoy:
Is this PR against the wrong branch? :) Seems like a lot of changes for just a resolution/fps sync change?

ClementHector:
Sorry I forgot to change branch!

PR #3078, opened 2022-04-20
Nuke: added suspend_publish knob

kalisp:
## Brief description
There was no control on Deadline of publishing. This PR adds artist controlable knob on Write node.

## Description
suspend_publish_Nuke

If checked, DL publish job gets created in suspended state. By default is unchecked, eg current state.

Also refactored code of all 3 Creators of write nodes, as they contained almost same code.

## Testing notes:
1. Create Render Write node (farm rendered)
2. Check new knob in Deadline tab of a write node
3. Publish and check `Publish - ....` job in DL monitor, it should be gray, eg suspended
4. Test all `Write` creators

### DIFF
--- a/openpype/hosts/nuke/plugins/create/create_write_still.py
+++ b/openpype/hosts/nuke/plugins/create/create_write_still.py
@@ -1,10 +1,10 @@
 import nuke
 
+from openpype.hosts.nuke.api import plugin
 from openpype.hosts.nuke.api.lib import create_write_node
-from openpype.hosts.nuke.plugins.create import create_write_render
 
 
-class CreateWriteStill(create_write_render.CreateWriteRender):
+class CreateWriteStill(plugin.AbstractWriteRender):
     # change this to template preset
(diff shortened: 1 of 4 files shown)


64qam (approved):

openpype/hosts/nuke/plugins/create/create_write_still.py
```diff
@@ -1,12 +1,10 @@
-from collections import OrderedDict
-
 import nuke
 
-from openpype.hosts.nuke.api import plugin
 from openpype.hosts.nuke.api.lib import create_write_node
+from openpype.hosts.nuke.plugins.create import create_write_render
```
iLLiCiTiT, line 4: I don't think we should import from other plugins at all, it adds import dependency between plugins so you (or others) may not know that changes in `CreateWriteRender` will affect other plugins. If you want extract common logic then it should be inside `openpype.hosts.nuke.api.<something>`.

PR #3086, opened 2022-04-22
Transcoding: Use operational patter to recognize OP Atom mxf format

iLLiCiTiT:
## Brief description
Use operational pattern ul to find out if input has OP Atom format instead of looking to product name.

## Description
We used product name to find out if input uses OP Atom format but product name is arbitrary key which does not define what format input has (e.g. Nuke adds there foundry specific name).  Operational pattern ul on the other side 
was designed to recognize different variants of mxf  e.g. [OP1a format](https://www.loc.gov/preservation/digital/formats/fdd/fdd000266.shtml) has `"060e2b34.04010101.0d010201.01010900"` and OP Atom has `"060e2b34.04010102.0d010201.10030000"`.

## Testing notes:
1. Create MXF OP Atom in Extract Review
2. Make sure Extract burnins will process it (this is step where modified code is executed)
3. Output after publishing should be OP Atom


jakubjezek001 (commented):
works well

jakubjezek001 (approved):

PR #3087, opened 2022-04-22
General: Remove repos related logic

iLLiCiTiT:
## Brief description
Remove logic related to repos directory in OpenPype which is removed in [PR](https://github.com/pypeclub/OpenPype/pull/3066).

## Description
Repos directory related logic was removed and few places based on repos directory path was modified. Repos directory is not handled at all from this moment. Zip files copied from version repository are removed.

## Additional info
PR is extending [this PR](https://github.com/pypeclub/OpenPype/pull/3066). This change may affect how OpenPype works so it was separated to this PR to see changes better.

## Testing notes:
- Test if OpenPype from code works
- Zip can be created from code
- Build can be created from code
- Build can load zip from openpype repository
    - zip copied to Local App data should be removed when extracted

### DIFF
--- a/igniter/bootstrap_repos.py
+++ b/igniter/bootstrap_repos.py
@@ -668,9 +668,9 @@ class BootstrapRepos:
         self._progress_callback = progress_callback
 
         if getattr(sys, "frozen", False):
-            self.live_repo_dir = Path(sys.executable).parent / "repos"
+            self.live_repo_dir = Path(sys.executable).parent
         else:
-            self.live_repo_dir = Path(Path(__file__).parent / ".." / "repos")
+            self.live_repo_dir = Path(Path(__file__).parent / "..")
 
     @staticmethod
     def get_version_path_from_list(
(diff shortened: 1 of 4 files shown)


igniter/bootstrap_repos.py
```diff
@@ -668,9 +668,9 @@ def empty_progress(x: int):
         self._progress_callback = progress_callback
 
         if getattr(sys, "frozen", False):
-            self.live_repo_dir = Path(sys.executable).parent / "repos"
+            self.live_repo_dir = Path(sys.executable).parent
```
antirotor, line 671: I think we should remove the use of `self.live_repo_dir` as it has no effect now. That probably involves slight refactor of `create_version_from_live_code()`

antirotor (changes requested):
I think we should clean it a little bit more

igniter/bootstrap_repos.py
iLLiCiTiT: Changed it to use `OpenPypeVersion.get_installed_version()` which seems to be easiest way how to get all needed information.

You can do changes into branch if you'd like to...

PR #3090, opened 2022-04-22
Ftrack: Update Create Folders action

iLLiCiTiT:
## Brief description
Create folders action use task as dictionary. (Variant for 3.9.x)

## Description
Changed how data for template collecting are prepared. Action used task as string instead of dictionary. Also made faster queries to get full hierarchy data.

## Testing notes:
1. Run the action
2. Check that work and publish folders are created


iLLiCiTiT:
Closed. Created from wrong branch.

PR #3093, opened 2022-04-22
StandalonePublisher: removed Extract Background plugins

kalisp:
## Brief description
`Extract Background Layout` and `Extract Background Comp` support was missing from 2.x

## Description
Necessary plugins were moved to separate AddOn as this functionality is customer specific.
Build itself need to contain `dataclasses` as library used in an addon requires them and vendorizying it doesn't work.


kalisp:
Converted to Addon.

64qam (changes requested):
It is not working for multiple files and `psd-tools` is not installed from addons

64qam (approved):

PR #3100, opened 2022-04-25
Maya: Implement `iter_visible_nodes_in_range` for extracting Alembics

### LINKED ISSUE #3085: Maya: animation publish option to not export hidden geometries to alembic
Alembic exporter now exports both visible and hidden geometries if they are in the export set. There should be an option to export only visible geometries.

 Use case: animating a complex character rig which represents multiple character variations in one rig, like a Fighter with multiple armor variations. If rig set to variation "A" only geometries visible in variation "A" should be exported, geometries visible in variations "B", "C" etc. not.

There should be an option attribute on the export set, to exclude hidden geometries from alembic export. 

There's a flag in Maya alembic exporter called "Renderable Only", but unfortunately it only  exclude a geometry from exporting if the visibility attribute is set to 0 value. If there's a connection to the visibility attribute, then the geometry will be exported regardless attribute's calculated value.

Fortunately in Colorbleed's Avalon configuration @BigRoy implemented a logic which before export detects which dag nodes are visible within the export frame range and only export those.  Related logic here, it is optimized for large hierarchies and heavy scenes https://github.com/Colorbleed/colorbleed-config/blob/acre/colorbleed/plugins/maya/publish/extract_alembic.py#L90

Related ` get_visible_in_frame_range` function: https://github.com/Colorbleed/colorbleed-config/blob/76f537eea041f8fed135f2b3334d70e1cf22b528/colorbleed/maya/lib.py#L2414

The default value of the option attribute on the export set could be false (export hidden geometries, works just like now) as it wouldn't cause any performance hit by running ` get_visible_in_frame_range` if you don't need this feature.

[cuID:OP-3115]

BigRoy:
## Brief description

This fixes #3085 by implementing a per frame visibility check for all dag nodes in the pointcache instance and only returning those visible nodes for exporting.

## Description

The `visibleOnly` attribute already existed on Animation and Pointcache instances in Maya in the Creators but was unused previously as far as I'm aware.

## Additional info

Here's a TEST for just the `iter_visible_nodes_in_range` function. You can also try and publish these particular nodes in a pointcache to confirm the functionality in that regard:

```python
from maya import cmds
from openpype.hosts.maya.api.lib import iter_visible_nodes_in_range


def cube(name):
    return cmds.ls(cmds.polyCube(name=name, constructionHistory=False), long=True)[0]


nodes = {
    name: cube(name) for name in [
        "on",
        "off",
        "animated_always_on",
        "animated_always_off",
        "animated_on_between_10_20",
        "connected_always_on",
        "connected_always_off",
        "connected_on_between_10_20",
        "expression_on",
        "expression_off",
        "expression_on_between_10_20"
    ]
}

cmds.setAttr(nodes["on"] + ".visibility", True)
cmds.setAttr(nodes["off"] + ".visibility", False)

# Animated
cmds.setKeyframe(nodes["animated_always_on"], attribute="visibility", time=1.0, value=True)
cmds.setKeyframe(nodes["animated_always_off"], attribute="visibility", time=1.0, value=False)
cmds.setKeyframe(nodes["animated_on_between_10_20"], attribute="visibility", time=0.0, value=False, itt="flat", ott="step")
cmds.setKeyframe(nodes["animated_on_between_10_20"], attribute="visibility", time=10.0, value=True, itt="flat", ott="step")
cmds.setKeyframe(nodes["animated_on_between_10_20"], attribute="visibility", time=20.0, value=False, itt="flat", ott="step")

# Connected
for src, dest in {
    "animated_always_on": "connected_always_on",
    "animated_always_off": "connected_always_off",
    "animated_on_between_10_20": "connected_on_between_10_20"
}.items():
    cmds.connectAttr(nodes[src] + ".visibility", 
                     nodes[dest] + ".visibility", 
                     force=True)
    
# Expressions
cmds.expression(name="visibilityExpressions", s= """
{0[expression_on]}.visibility = 1;
{0[expression_off]}.visibility = 0;
{0[expression_on_between_10_20]}.visibility = (time1.outTime >= 10) && (time1.outTime <= 20);
""".format(nodes))

# Expected test results
ALWAYS_ON = {
    nodes["on"],
    nodes["animated_always_on"],
    nodes["connected_always_on"],
    nodes["expression_on"],
}

ON_IN_10_20 = {
    nodes["animated_on_between_10_20"],
    nodes["connected_on_between_10_20"],
    nodes["expression_on_between_10_20"]
} | ALWAYS_ON

# Test `iter_visible_nodes_in_range`
node_names = list(nodes.values())
assert set(iter_visible_nodes_in_range(node_names, start=1, end=9)) == ALWAYS_ON
assert set(iter_visible_nodes_in_range(node_names, start=1, end=25)) == ON_IN_10_20
assert set(iter_visible_nodes_in_range(node_names, start=15, end=15)) == ON_IN_10_20
assert set(iter_visible_nodes_in_range(node_names, start=21, end=30)) == ALWAYS_ON
```

## Testing notes:
1. Test whether visibility is correctly detected in a wide range of scenarios. (e.g. Parenting relationships are not tested in the above example test case but can easily be tested against as well.)


m-u-r-p-h-y (changes requested):
there is some issue  with sets containing hidden geo

```
    runner(*args)
  File "D:\REPO\OpenPype\openpype\hosts\maya\plugins\publish\extract_pointcache.py", line 93, in process
  File "D:\REPO\OpenPype\openpype\hosts\maya\api\lib.py", line 1241, in extract_alembic
    cmds.AbcExport(j=job_str, verbose=verbose)
  File "D:\REPO\OpenPype\openpype\hosts\maya\plugins\publish\extract_pointcache.py", line 2, in AbcExport
RuntimeError: -selection specified but nothing is actively selected.

    runner(*args)
  File "<string>", line 93, in process
  File "D:\REPO\OpenPype\openpype\hosts\maya\api\lib.py", line 1241, in extract_alembic
    cmds.AbcExport(j=job_str, verbose=verbose)
  File "<string>", line 2, in AbcExport
RuntimeError: -selection specified but nothing is actively selected.
```

image

BigRoy:
You're trying to export a pointcache with NO geometry whatsoever due to it always being hidden and thus being excluded. :) I was actually expecting that to happen!

Thus this should get validated for both scenarios, however this visibility check is _relatively_ slow for large scenes so it could potentially mean the Validator would be slow. Plus, we'd have the plug-in likely "cache" the visibility states so the Extractor doesn't have to do it again but that also means a Validator is technically "collecting" information. Moving that to a Collector means that technically the collecting stage can be slow.

OR We'd just log a warning in the Extractor and do nothing instead? Thoughts?

Anyway, thanks - will look into this edge case.

mkolar:
>OR We'd just log a warning in the Extractor and do nothing instead? Thoughts?

I think, this might be the right approach in this situation. Probably more intuitive to export nothing if everything is hidden rather than crashing.

mkolar:
@BigRoy just a ping that if we do the last tweak mention in you comment, this will be good to merge

BigRoy:
> just a ping that if we do the last tweak mention in you comment, this will be good to merge

@mkolar I've just pushed the change to the extractors so that the message is logged and the extractor continues. However, the integrator will currently still error on any instances that do **not** produce any files whatsoever.

See:
afbeelding

mkolar:
>How would you like me to proceed?

That's somewhat tricky. We should either figure this out in the validator, or this particular instance should never get to integration stage. No files, nothing to integrate. 

It could get really confusing for the artist though, so I think validation should flag this, if we're able to catch it. If not, then the extractor, should figure out it produced no files, hence not create a representation and the integrator can then gracefully skip

BigRoy:
I've changed the function to an iterator to allow a much faster validation for 99% of the cases where at least a single node is detected to be visible. It's good to be aware that the validation allows any dag node to be visible. That also includes e.g. a group node which I felt makes sense to be the least restrictive and it also wouldn't fail on the extraction (which is what we were trying to solve anyway.)

Should be good to test now @mkolar

openpype/hosts/maya/plugins/publish/validate_visible_only.py
```diff
@@ -0,0 +1,51 @@
...
+            start, end = self.get_frame_range(instance)
+            raise RuntimeError("No visible nodes found in "
+                               "frame range {}-{}.".format(start, end))
+
+    @classmethod
+    def get_invalid(cls, instance):
+
+        if instance.data["family"] == "animation":
+            # Special behavior to use the nodes in out_SET
+            nodes = instance.data["out_hierarchy"]
```
BigRoy, line 38: ### Warning

It's good to be aware that this logic depends on [THIS Collector plug-in](https://github.com/pypeclub/OpenPype/blob/3bb6d8aea083220ee570b61b9c741e03aa7524c9/openpype/hosts/maya/plugins/publish/collect_animation.py#L56-L57).

The bug/issue is that if the instance in Maya is set to `active=False` on the objectSet/instance directly that that Collector does not run.

There are two solutions to that:
- Have the instance collector NOT collect the active state directly but instead collect as e.g. `__active`. Then have a very later Collector (`CollectorOrder + 0.499`) which sets `instance.data["active"] = instance.data,pop("__active", True)` - so that up to that point all collectors do run as expected.
- Use/implement something like Post collectors like `PYBLISH_QML_POST_COLLECT` see: https://github.com/pyblish/pyblish-qml/pull/356 - that way we can have these collectors that should always run for an active instance in the post collection step after the user toggled states. Upside to this is that collecting will be MUCH faster since many collectors could move to the post collector order.

iLLiCiTiT:
I have stupid question. What does it return? I miss one word in the function name: "Get visible ... in frame range".

BigRoy:
> I have stupid question. What does it return? I miss one word in the function name: "Get visible ... in frame range".

It's currently been renamed to `iter_visible_in_frame_range` - not sure if that's more descriptive for you.
But basically it yields the nodes from the input nodes that are visible at least once during the frame range.
Would you like me to refactor it to: `iter_visible_nodes_in_frame_range`?

iLLiCiTiT:
> Would you like me to refactor it to: iter_visible_nodes_in_frame_range?

I would say it's more important then it's from frame range :) (Maybe `iter_visible_nodes_in_range` would make it shorter)

m-u-r-p-h-y (approved):
Testing with my old setup

validation for visibility works great

image

and publish as well

image

BigRoy:
Can we merge this?

mkolar (approved):

mkolar:
Merging. Thank you.

Issue #3054, opened 2022-04-12
TrayPublisher: Simple families from settings


iLLiCiTiT:
## Description
Right now there is workfile creator in tray publisher but the same logic could be generalized for simple cases and extended by defying what file extensions are allowed to publish to what family. This simple solution could not have much specific logic (like batch publishing) but would work for most of simple cases we would need now.

## Suggested solution
Have settings with list where each item define **family** and **allowed extensions** (maybe allow sequences?) as required values and optional **label** and **icon**.  This could be enough for start and would cover 50% of current standalone publisher.

### Note
File definition and it's widgets are not perfect at this moment so it requires their enhancement too (missing sequence option).

[cuID:26b71vb]

PR #3105, opened 2022-04-26
Tray publisher: Simple families from settings

iLLiCiTiT:
## Brief description
Added simplified version of TrayPublisher publishing.

## Description
It is possible to "add" creators into TrayPublisher using settings. These creators must define family, allowed extensions and if supports sequences and optionally can enter identifier, icon, label, descritption and if "review" family should be added. Single "creation" allows only one file item: "single file" or "single sequence of files" AKA frame sequence. Removed workfile creator and added `workfile` family to default settings. Previous 'workfile' publishing plugins were generalized. Few publisher UI enhancements/fixes.

## Additional info
Implemeneted default `TrayPublishCreator ` which has implemented all basic methods related to tray publisher and `SettingsCreator` used for simple variant of tray publisher. Slightly modified handling of `FileDef` and it's UI. Result of `FileDef` is `FileDefItem` which can be converted to and from dictionary. Holds more information about files in it, for example if represents sequence and which frames of sequence so creator can know that user added files as sequence or as single files if needed.

FileDef UI does not show text input but always shows view. View not allow to drop files with unsupported extensions. Sequences can be split into multiple files (cannot be joined back). Delete button is not on each item but at bottom of view.

## TODO / Known issues
- ~~Multiselection of FileDef UI is not handled~~
- FileDef UI does not show allowed extensions (not 100% sure where to put it?)
- Remove button can be over items
- Add more generic creators

## Testing notes:
**Try workfile publishing**
**Add new creator**
1. Add new creator in `project_settings/traypublisher/simple_creators`
- for example `"image"` family with `".png"` and `".jpg"` extension that allows sequences
2. Try publish something using new creator


kalisp:
`filepath` produces weird result for workfiles with spaces:
Screenshot 2022-04-27 125617

mkolar:
> `filepath` produces weird result for workfiles with spaces)

spaces......ouch.

iLLiCiTiT:
That's just has `pprint` resolves too long strings but it is ok.

mkolar (changes requested):
`enable review` checkbox has no effect, I'm assuming it's because we don't have plugins that would support the data it adds.

Secondly these two don't show up anywhere. 
image

mkolar:
resolves #3054

iLLiCiTiT:
> enable review checkbox has no effect, I'm assuming it's because we don't have plugins that would support the data it adds.

Added `traypublisher` to ExtractReview and ExtractBurnin plugins.

> Secondly these two don't show up anywhere.

Fixed

PR #3109, opened 2022-04-28
TVPaint: Match renderlayer key with other hosts

iLLiCiTiT:
## Brief description
Render layer key in TVPaint implementation is without underscode (`render_layer` -> `renderlayer`).

## Description
Add 'renderlayer' to instance.data for TVPaint render instances. Render passes had `render_layer` which was changed to `renderlayer` and layers pass their `variant` as value. This is to match Maya specifically. There is a secondary reason to do so, to match subset group name filtering. Also usage of `render_pass` was modified to use `renderpass` to keep consistency.

## Additional info
For some time there must be bost ways how are keys stored. Release variant of [PR](https://github.com/pypeclub/OpenPype/pull/3110).

## Testing notes:
- New files
1. Open new TVpaint file
2. Create some render layers/passes
3. Publish
4. All should work and if subset group profile is matching, it should be grouped

- Old files
1. Open TVPaint file with instances created before this PR
2. publish
3. All should work and subset groups should match too

### DIFF
--- a/openpype/hosts/tvpaint/plugins/create/create_render_pass.py
+++ b/openpype/hosts/tvpaint/plugins/create/create_render_pass.py
@@ -36,7 +36,7 @@ class CreateRenderPass(plugin.Creator):
 
         # TODO remove - Backwards compatibility for old subset name templates
         # - added 2022/04/28
-        dynamic_data["renderpass"] = dynamic_data["render_pass"]
+        dynamic_data["render_pass"] = dynamic_data["renderpass"]
 
         return dynamic_data
 
--- a/openpype/hosts/tvpaint/plugins/create/create_render_pass.py
+++ b/openpype/hosts/tvpaint/plugins/create/create_render_pass.py
@@ -20,7 +20,9 @@ class CreateRenderPass(plugin.Creator):
     icon = "cube"
     defaults = ["Main"]
 
-    dynamic_subset_keys = ["render_pass", "render_layer"]
+    dynamic_subset_keys = [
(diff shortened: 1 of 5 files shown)


openpype/hosts/tvpaint/plugins/create/create_render_pass.py
```diff
@@ -29,9 +31,13 @@ def get_dynamic_data(
         dynamic_data = super(CreateRenderPass, cls).get_dynamic_data(
             variant, task_name, asset_id, project_name, host_name
         )
-        dynamic_data["render_pass"] = variant
+        dynamic_data["renderpass"] = variant
         dynamic_data["family"] = "render"
 
+        # TODO remove - Backwards compatibility for old subset name templates
+        # - added 2022/04/28
+        dynamic_data["renderpass"] = dynamic_data["render_pass"]
```
BigRoy, line 39: Won't this always overwrite the `variant` value set above?

iLLiCiTiT: This is called in `get_subset_name`, before creation.

BigRoy: It's more that a few lines above `renderpass` is set to `variant` and then here we're directly setting it to `dynamic_data["render_pass"]`. So I have the feeling that either of the two lines is redundant.

iLLiCiTiT: That's because I swapped keys :pensive: .

64qam (approved):

PR #3111, opened 2022-04-28
General: Python 3 compatibility in queries

iLLiCiTiT:
## Brief description
Get last versions in queries works for Py3 hosts.

## Description
There is an issue that dictionary in Py3 returns `dict_keys` when `.keys()` is called which causes crash in mongod queries, so we're making sure that result of `keys()` is converted to list.

## Testing notes:
1. Run build workfile in Python 3 host

Older release variant of [PR](https://github.com/pypeclub/OpenPype/pull/3112).


BigRoy:
Isn't `dict_keys` a Py3 thing instead of Py2?

PR #3117, opened 2022-04-28
Nuke: render instance with subset name filtered overrides

jakubjezek001:
## Brief description
Adding ability to override Anatomy imagio required nodes with subset filtering.

## Description
Some studios were requesting overriding per subset name regex filtering. This will allow to have different colorspace on subset `renderCompMain` then on `renderCompMask` for example. Overrides and also adding missing knob values if they are not found in the original Required Plugin nodes. 

## Additional info
- if override knob is having the same name but different value, value is overriten
- if override knob is not having its equivalent in original knobs, it is added
- if override knob is having matching equivalent in original knobs and having empty value, it is removed from original knobs.

## Testing notes:
1. Start testing workfile in nuke
2. go to Studio settings `project_anatomy/imageio/nuke/nodes/overrideNodes` at your project
3. configure new override similary to the image, and save settings
image
4. in nuke go to openpype/create/Create Write Render, set Subset to `Boho` and Create
5. Select the node and CTRL+Enter to see content of render group. Double click at write node and see properties bin. It should look similar to following image, if your colorspace was originaly defined as ACES
image
6. now try to create again but Subset name as `Main` and repeate step 5 and look to `project_anatomy/imageio/nuke/nodes/requiredNodes/0/knobs` and compare. It shold be set exactly as it is in settings.

### DIFF
--- a/openpype/hosts/nuke/api/lib.py
+++ b/openpype/hosts/nuke/api/lib.py
@@ -560,7 +560,13 @@ def get_created_node_imageio_setting(**kwarg):
                         "_ overriding knob: `{}` > `{}`".format(
                             knob, oknob
                         ))
-                    knob["value"] = oknob["value"]
+                    if not oknob["value"]:
+                        # remove original knob if no value found in oknob
+                        imageio_node["knobs"].remove(knob)
+                    else:
+                        # override knob value with oknob's
+                        knob["value"] = oknob["value"]
+
                 # add missing knobs into imageio_node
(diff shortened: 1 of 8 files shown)


64qam (approved):

jrsndl:
Works nicely for string knob overrides. It ignores boolean settings, and fails on integer settings.
It expects first character to be uppercase in the settings, to match to the Subset in the Instance Creator.

PR #3122, opened 2022-04-29
Support for Unreal 5

### LINKED ISSUE #3101: Support for Unreal 5
## Feature

Unreal 5 has few things in different places we need to handle - rename main executable from `UE4Editor` to `UnrealEditor`, `UnrealBuildTool` is under its own directory now, and so on.

When this fixed unreal will open without the project (probably project metadata) and icon in toolbar is missing. Python paths to openpype are not set.

[cuID:OP-3130]

antirotor:
## Fixing support for Unreal 5

This is adding support for both UE4 and UE5, dealing with the need for two versions of integration plugin and adding `UE_PYTHONPATH` to defaults for UE5.

There is still one caveat and that is that we need to install pyside2 to Unreal and we can't do that when Unreal is installed to privileged location (like Program Files on Windows) where you need admin rights to write.

### To test it:
Install both UE4 and UE5 versions, enable them on project, make sure you have `{"UE_PYTHONPATH": "{PYTHONPATH}"}` in environment for UE5 and try to create projects in both version.

✏️ Note: You'll need Visual Studio if you don't have already built integration plugins.

Close #3101


antirotor:
We are now missing support for ue 4.2x (we will probably need to retain two versions of integration plugin). And rendering is not working #3207

m-u-r-p-h-y (approved):
basic functionality (OP integration works)

image

not sure what other things to test at this stage

openpype/hosts/unreal/__init__.py
```diff
@@ -1,15 +1,19 @@
...
 import openpype.hosts
+from openpype.lib.applications import Application
 
 
-def add_implementation_envs(env, _app):
+def add_implementation_envs(env: dict, _app: Application) -> None:
     """Modify environments to contain all required for implementation."""
     # Set OPENPYPE_UNREAL_PLUGIN required for Unreal implementation
+
+    ue_plugin = "UE_5.0" if _app.name[:1] == "5" else "UE_4.7"
```
iLLiCiTiT, line 10: It is already merged but I would recommend to add separated input for Unreal where you would tell which version of unreal it is... Using a name which is defined by admins is not best approach. If they'd want use name `testing`, you'd be lost.

PR #3123, opened 2022-05-02
Nuke: Expose write attributes to settings

jakubjezek001:
## Brief description
Created node knobs have settings for overrides

## Description
Now we are able to define knob overrides for example for `render` or `deadlinePriority`.

## Additional info
In settings it is important to know knob type to be able to define it correctly - finally no guessing on this.

## Testing notes:
1. open your favorite testing workfile in nuke
2. hover above created node's knob to find out exact name of knob 
image
3. add some knob override to `project_settings/nuke/create/CreateWriteRender/knobs` to your project override and save settings.
4. Create Render node and see that the default value was set.

### DIFF
--- a/openpype/hosts/nuke/api/lib.py
+++ b/openpype/hosts/nuke/api/lib.py
@@ -858,6 +858,7 @@ def create_write_node(name, data, input=None, prenodes=None,
     Return:
         node (obj): group node with avalon data as Knobs
     '''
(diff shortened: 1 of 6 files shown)


tokejepsen (changes requested):
Just that Hound error to resolve. Rest looks good to me.

64qam (approved):

PR #3128, opened 2022-05-02
Maya: fix image prefix warning in validator

### LINKED ISSUE #3127: Maya Render Settings validator prefix confusion
## Issue

There is a little issue with Render Settings validator that if `Merge AOV` is selected,  This validator will complain about superfluous `<RenderPass>` token. But If that token is correctly removed, Render Settings validator will issue warning about image prefix being different than the default one.

[cuID:296qnmz]

antirotor:
## Fix

This is fixing issue when `Merge AOV` option forces render settings validator to complain about superfluous `<RenderPass>` token. But if removed, it issues warning about image prefix not being the one recommended.

## Test

Create scene with render instance, in Render Settings check `Merge AOVs` in Arnold settings. Leave prefix as it is. Hit Publish -> Validate. Render Settings Validator should fail on unnecessary `<RenderPass>` token. Remove it from the prefix, validate again and it should pass.

## Note

This is quick fix but I think Render Settings validator still needs big overhaul to deal with different renderer settings more gracefully. Ideally providing validator for every renderer.

Close #3127


openpype/hosts/maya/plugins/publish/validate_rendersettings.py
```diff
@@ -92,6 +92,7 @@ def process(self, instance):
     def get_invalid(cls, instance):
 
         invalid = False
+        multipart = False
```
BigRoy, line 95: Looking at all the changes this addition seems redundant.
```suggestion
```

openpype/hosts/maya/plugins/publish/validate_rendersettings.py
```diff
@@ -211,14 +213,20 @@ def get_invalid(cls, instance):
...
                                   "You can't use '<renderpass>' token "
                                   "with merge AOVs turned on".format(prefix))
+                default_prefix = re.sub(
+                    cls.R_AOV_TOKEN, "", default_prefix)
+                # remove aov token from prefix to pass validation
+                # first resolve it and then remove if dangling
+                default_prefix = default_prefix.replace(
+                    "{aov_separator}", instance.data.get("aovSeparator", "_"))
+                default_prefix = default_prefix.rstrip(
+                    instance.data.get("aovSeparator", "_"))
```
BigRoy, line 223: Why first resolve it and then strip the right hand side aov separator tokens, that seems redundant and maybe even not as explicit as one might want. If it's stripping it anyway shouldn't it just replace the explicit `{aov_separator}` placeholder with an empty string?

```suggestion
                # remove aov token separator
                default_prefix = default_prefix.replace("{aov_separator}", "")
```

Additionally, since the image prefix is hardcoded in the validator anyway (doesn't seem to allow to be overridden by settings) this "fix" could even be made more explicit by explicitly replacing the end of the default image prefix alltogether?

We could even get by with just a single `re.sub(re.escape("{aov_separator}<RenderPass>") + "$", "", default_prefix)`? Or even as simple as `prefix = prefix[:-len("("{aov_separator}<RenderPass>")]` to just strip off the hardcoded end or `prefix = prefix.split("{aov_separator}")[0]`?

m-u-r-p-h-y (approved):
Validators are passing correctly with Merge AOV option turned on

image

openpype/hosts/maya/plugins/publish/validate_rendersettings.py
mkolar: @antirotor just pinging you here in case you didn't notice

PR #3131, opened 2022-05-02
General: Cleanup some Loader docstrings

BigRoy:
## Brief description

Purely cosmetics (except for one!) change to fix some loader docstrings to be more accurate.

## Description

I believe these are also used as tooltip or status tip for the Loaders in the loader context menu so artists might actually see this information too.

## Additional info

It might still be worth looking into doing a more thorough cleanup and unifying of how Loaders are described across all hosts.

### DIFF
--- a/openpype/hosts/houdini/plugins/load/load_image.py
+++ b/openpype/hosts/houdini/plugins/load/load_image.py
@@ -42,9 +42,9 @@ def get_image_avalon_container():
 
 
 class ImageLoader(load.LoaderPlugin):
-    """Specific loader of Alembic for the avalon.animation family"""
+    """Load images into COP2"""
 
-    families = ["colorbleed.imagesequence"]
+    families = ["imagesequence"]
     label = "Load Image (COP2)"
     representations = ["*"]
     order = -10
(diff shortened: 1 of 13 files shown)


openpype/hosts/houdini/plugins/load/load_image.py
```diff
@@ -42,9 +42,9 @@ def get_image_avalon_container():
 
 
 class ImageLoader(load.LoaderPlugin):
-    """Specific loader of Alembic for the avalon.animation family"""
+    """Load images into COP2"""
 
-    families = ["colorbleed.imagesequence"]
+    families = ["imagesequence"]
```
BigRoy, line 47: Note: This is not just a cosmetic change. However the `colorbleed.imagesequence` family does not exist in the OpenPype code base and thus this didn't work previously anyway.

It should however be part of the Houdini cleanup and improvements to actually fix these up and test the behaviors as I believe OpenPype actually doesn't use the `imagesequence` family either but instead uses `render` as family. Nonetheless there are still quite a few plug-ins that are set to operate on `imagesequence` instead.

mkolar (approved):
Thank you this is all looking good

openpype/hosts/houdini/plugins/load/load_image.py
mkolar: Very true.

PR #3132, opened 2022-05-02
Cosmetics: Minor typo fix in ModulesLoader

BigRoy:
## Brief description

Fixes minor typo in logged message, `contan` -> `contain`.

## Description

Before:
```
  - { ModulesLoader }: [  Module directory does not contan __init__.py file /path/to/module
```

After:
```
  - { ModulesLoader }: [  Module directory does not contain __init__.py file /path/to/module
```

### DIFF
--- a/openpype/modules/base.py
+++ b/openpype/modules/base.py
@@ -353,7 +353,7 @@ def _load_modules():
             basename, ext = os.path.splitext(filename)
 
             if os.path.isdir(fullpath):
-                # Check existence of init fil
+                # Check existence of init file
                 init_path = os.path.join(fullpath, "__init__.py")
                 if not os.path.exists(init_path):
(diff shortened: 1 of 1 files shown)


BigRoy:
I also feel like the logic to load default modules could be unified with the logic that is loading the files/directories from `get_module_dirs()`. What's the reason we need the logic almost duplicated there?

The non-default modules also do a double check whether `__init__.py` exists in the folders too, seems redundant too?

iLLiCiTiT (approved):

iLLiCiTiT:
> The non-default modules also do a double check whether __init__.py exists in the folders too, seems redundant too?

Created PR https://github.com/pypeclub/OpenPype/pull/3137.

PR #3133, opened 2022-05-02
Terminal: Tweak coloring of TrayModuleManager logging enabled states

BigRoy:
## Brief description

This tweaks the regex for colorizing in the terminal to change how the TrayModulesManager logs the enabled states' colors.

## Description

**Before:**

traymodulemanager_logs

**After:**

traymodulemanager_changed

## Additional info

I'm not too knowledgeable about this area of the code base so please double check the change.

## Testing notes:

1. Check whether logging messages are colored as intended and messages come out ok.


antirotor (approved):
Thanks, finally it is fixed :D

PR #3136, opened 2022-05-03
General: Missing version on headless mode crash properly

iLLiCiTiT:
## Brief description
Added missing variable in start.py.

## Description
Usage of missing OP version in headless mode did not crash properly. It crashed because of missing 'openpype_versions' variable. Moved logic of validate versions and print versions from `boot` into separated functions.

## Testing notes:
1. Start of OpenPype should work as did
2. When version which is not available should be used, it should crash properly in headless mode
    - run e.g. `openpype_console --headless --use-version="4.4.4"` or `start.py --headless --use-version="4.4.4"`

## Traceback
```
Requested version "3.11.0" was not found.
  File "C:\Users\JakubTrllo\Desktop\Prace\openpype3_2\start.py", line 1050, in boot
    version_path = _bootstrap_from_code(use_version, use_staging)
  File "C:\Users\JakubTrllo\Desktop\Prace\openpype3_2\start.py", line 830, in _bootstrap_from_code
    use_version
igniter.tools.OpenPypeVersionNotFound: Requested version "3.11.0" was not found.

During handling of the above exception, another exception occurred:

  File "C:\Users\JakubTrllo\Desktop\Prace\openpype3_2\start.py", line 1175, in <module>
    boot()
  File "C:\Users\JakubTrllo\Desktop\Prace\openpype3_2\start.py", line 1056, in boot
    list_versions(openpype_versions, local_version)
UnboundLocalError: local variable 'openpype_versions' referenced before assignment
```

### DIFF
--- a/start.py
+++ b/start.py
@@ -1029,6 +1029,9 @@ def boot():
             message = str(exc)
             _print(message)
             if os.environ.get("OPENPYPE_HEADLESS_MODE") == "1":
+                openpype_versions = bootstrap.find_openpype(
+                    include_zips=True, staging=use_staging
+                )
                 list_versions(openpype_versions, local_version)
             else:
                 igniter.show_message_dialog("Version not found", message)
@@ -1053,6 +1056,9 @@ def boot():
             message = str(exc)
             _print(message)
             if os.environ.get("OPENPYPE_HEADLESS_MODE") == "1":
+                openpype_versions = bootstrap.find_openpype(
+                    include_zips=True, staging=use_staging
+                )
                 list_versions(openpype_versions, local_version)
             else:
                 igniter.show_message_dialog("Version not found", message)
--- a/start.py
+++ b/start.py
@@ -931,11 +931,11 @@ def _boot_print_versions(use_staging, local_version, openpype_root):
     else:
         _print("--- This will list only staging versions detected.")
         _print("    To see other version, omit --use-staging argument.")
-    _openpype_root = OPENPYPE_ROOT
+
     openpype_versions = bootstrap.find_openpype(include_zips=True,
                                                 staging=use_staging)
     if getattr(sys, 'frozen', False):
-        local_version = bootstrap.get_version(Path(_openpype_root))
+        local_version = bootstrap.get_version(Path(openpype_root))
     else:
         local_version = OpenPypeVersion.get_installed_version_str()
 
--- a/start.py
+++ b/start.py
@@ -897,6 +897,51 @@ def _bootstrap_from_code(use_version, use_staging):
     return version_path
 
 
+def _boot_validate_versions(use_version, local_version):
+    _print(f">>> Validating version [ {use_version} ]")
+    openpype_versions = bootstrap.find_openpype(include_zips=True,
+                                                staging=True)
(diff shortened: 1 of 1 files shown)


start.py
```diff
@@ -897,6 +897,51 @@ def _bootstrap_from_code(use_version, use_staging):
...
+        _print("--- This will list only non-staging versions detected.")
+        _print("    To see staging versions, use --use-staging argument.")
+    else:
+        _print("--- This will list only staging versions detected.")
+        _print("    To see other version, omit --use-staging argument.")
+    _openpype_root = OPENPYPE_ROOT
+    openpype_versions = bootstrap.find_openpype(include_zips=True,
+                                                staging=use_staging)
+    if getattr(sys, 'frozen', False):
+        local_version = bootstrap.get_version(Path(_openpype_root))
```
BigRoy, line 938: I think this needs to be changed so that `openpype_root` is actually used from input arguments.
```suggestion
    openpype_versions = bootstrap.find_openpype(include_zips=True,
                                                staging=use_staging)
    if getattr(sys, 'frozen', False):
        local_version = bootstrap.get_version(Path(openpype_root))
```

start.py
```diff
@@ -1029,6 +1037,9 @@ def boot():
             message = str(exc)
             _print(message)
             if os.environ.get("OPENPYPE_HEADLESS_MODE") == "1":
+                openpype_versions = bootstrap.find_openpype(
+                    include_zips=True, staging=use_staging
+                )
                 list_versions(openpype_versions, local_version)
             else:
                 igniter.show_message_dialog("Version not found", message)
```
BigRoy, line 1045: I'm thinking that since this logic is duplicated just below, should we maybe encapsulate that in its own function, like:

```python
def _debug_version_not_found(exc):
    message = str(exc)
    _print(message)
    if os.environ.get("OPENPYPE_HEADLESS_MODE") == "1":
        openpype_versions = bootstrap.find_openpype(include_zips=True, 
                                                    staging=use_staging)
        list_versions(openpype_versions, local_version)
    else:
        igniter.show_message_dialog("Version not found", message)
```

We could even define this within the boot() function just before this part in the code - just so it's clear that it's re-used?

start.py
```diff
@@ -897,6 +897,51 @@ def _bootstrap_from_code(use_version, use_staging):
...
+    result = bootstrap.validate_openpype_version(
+        bootstrap.get_version_path_from_list(
+            use_version, openpype_versions))
+
+    _print("{}{}".format(
+        ">>> " if result[0] else "!!! ",
+        bootstrap.validate_openpype_version(
+            bootstrap.get_version_path_from_list(
+                use_version, openpype_versions)
+        )[1])
```
BigRoy, line 923: Isn't this the same?
```suggestion
    _print("{}{}".format(
        ">>> " if result[0] else "!!! ", result[1]
        )
```

start.py
BigRoy: Ah, you actually JUST did so. :)

iLLiCiTiT: Oh yes, I blindly copied the code and it crashed :)

BigRoy (approved):
Looks good to me.

PR #3140, opened 2022-05-03
Houdini: Add loader for alembic through Alembic Archive node

BigRoy:
## Brief description

Adds a Houdini loader for Alembic files through the `alembicarchive` node.

_Requested by one of our artists_.

## Description

The original Alembic loader would load alembic geometry directly into SOPs with an `alembic` node. The downside to that is that not all nodes/hierarchies can be accurately represented in SOPs since it's only geometry data. Thus empty transforms, or locators or parents in the hierarchies can't be correctly represented in SOPs only.

This Loader instead does allow to rebuild the full Alembic Hierarchy as a Houdini Hierarchy in the Network Editors.

## Additional Info

- Logic is loosely based on the Camera Loader which already was using the `alembicarchive` node.
- Any local changes made INSIDE the Alembic Archive in the Houdini are lost upon updating since the contents are "Rebuild".

## Testing notes:
1. Test whether loading works as expected
2. Test whether updating works as expected.

### DIFF
--- a/openpype/hosts/houdini/plugins/load/load_alembic_archive.py
+++ b/openpype/hosts/houdini/plugins/load/load_alembic_archive.py
@@ -1,0 +1,75 @@
+import os
+from openpype.pipeline import (
+    load,
+    get_representation_path,
+)
+from openpype.hosts.houdini.api import pipeline
+
+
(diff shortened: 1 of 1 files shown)


antirotor (approved):
Thanks, its working!

PR #3143, opened 2022-05-04
Nuke: settings for reformat node in CreateWriteRender node

jakubjezek001:
## Brief description
Reformat node can be altered in creator settings.

## Description
We are supporting reformat knob configurations at creator plugin settings. Since we needed to refectory the way settings are working with knob configurations we are now able to do per subset overrides. Pre nodes were added to all creators plugins in settings so the studio can add some other nodes before a write node with defined knobs - also with ability to format templated values (this needs to be enhanced for custom key inputs in future PR)

## Additional info
We needed to unify the way nuke knobs are defined in settings, so all knob configurations are now using only one template schema. This will allow for imageio driven node overrides with subset filter. 

## Known issue
PR is backward compatible, because most of legacy code had been renamed as `_legacy` and `AbstractWriteRender` plugin class is having `is_legacy` function for legacy detection. But in case someone will re-save project settings, anatomy imageio will be broken. 

The way convert old settings to new is to open the project anatomy settings in older version, screengrab settings in "project_anatomy/imageio/nuke/nodes/requiredNodes" particular Knobs are what we are interested in. Then open settings in this PR and see that original values had disappear (schemas had changed). Now you need to add manually those screengrabed values.

### General behaviour
1. Before you check out this PR save the anatomy info (look to `Known issue`)
2. Checkout to this PR and add the saved knob settings
3. Save the settings nad open Nuke in some project's task
4. Create Write render/prerender/still 
5. all should look as you would expect from settings

### ImageIO Plugin node overrides
1. add some node override to `project_anatomy/imageio/nuke/nodes/overrideNodes`
2. example here: 
image
3. should result in 
image


openpype/hosts/nuke/api/lib.py
```diff
@@ -374,23 +375,18 @@ def add_write_node(name, **kwarg):
     Returns:
         node (obj): nuke write node
     """
+    file_path = kwarg["path"]
+    knobs = kwarg["knobs"]
```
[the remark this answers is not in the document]
jakubjezek001: yeah, that makes sense. thanks ;)

jrsndl:
Works as described. I tested Numbers, Booleans and Hexa.

jrsndl (approved):

openpype/hosts/nuke/plugins/create/create_write_still.py
```diff
@@ -2,6 +2,10 @@
 
 from openpype.hosts.nuke.api import plugin
 from openpype.hosts.nuke.api.lib import create_write_node
+from openpype.api import (
+    Logger
+)
+log = Logger.get_logger(__name__)
```
iLLiCiTiT, line 8: Variable `log` is not used.

openpype/settings/entities/schemas/projects_schema/schemas/template_nuke_knob_inputs.json
```diff
@@ -0,0 +1,258 @@
...
+                    {
+                        "key": "hex",
+                        "label": "Hexadecimal (0x)",
+                        "children": [
+                            {
+                                "type": "text",
+                                "key": "name",
+                                "label": "Name"
+                            },
+                            {
```
iLLiCiTiT, line 76: Wouldn't be easier to use `"color"` instead of `"text"`?

jakubjezek001: good point, thanks ;)

PR #3144, opened 2022-05-05
Deadline: fix the output directory

antirotor:
## Fix

This is fixing bug introduced with renderman support changing output directory in deadline, Because renderman has different prefix for file and folder.

This is quick fix, but I still vote for optimizing the code to have as little as possible of hardcoded renderer specific code.

### DIFF
--- a/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
+++ b/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
@@ -266,7 +266,6 @@ def get_renderer_variables(renderlayer, root):
 
     filename_prefix = cmds.getAttr(prefix_attr)
     return {"ext": extension,
-            "renderer": renderer,
             "filename_prefix": filename_prefix,
             "padding": padding,
             "filename_0": filename_0}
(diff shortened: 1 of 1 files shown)


openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
```diff
@@ -440,7 +441,8 @@ def process(self, instance):
 
         output_filename_0 = filename_0
 
-        dirname = os.path.dirname(output_filename_0)
+        if render_variables["renderer"] == "renderman":
+            dirname = os.path.dirname(output_filename_0)
```
BigRoy, line 445: Should we add a comment as to why this workaround is needed for renderman?

BigRoy: Also might  be good to be aware that elsewhere in the code `renderer` is taken from `instance.data` instead of re-read from the layer: https://github.com/pypeclub/OpenPype/blob/3744b21266258fdd69814807a4152e268dca3d27/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py#L731

As you said, cleanup of this code is needed. :)

BigRoy (approved):
Works on our end with Maya 2022 + Redshift + Deadline. Renders come through, files in correct location and publish works.

Previously rendering in Deadline did work, but files were in wrong location and thus publish jobs failed.

Thanks!

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
antirotor: you are right, I've added those changes

m-u-r-p-h-y (approved):
Submitted RenderMan job from Maya to Deadline and everything seems working. Rendered and published successfully 

image

PR #3147, opened 2022-05-05
Hooks: Tweak logging grammar

BigRoy:
## Brief description

This tweaks the Hooks logging grammar to (I think) be more valid plus it also adds the actual Hook's name that is being skipped due to not being valid for a context.

## Testing notes:
1. Launch an app from tray and note the logs about the Hooks

### DIFF
--- a/openpype/lib/applications.py
+++ b/openpype/lib/applications.py
@@ -1044,8 +1044,8 @@ class ApplicationLaunchContext:
                     hook = klass(self)
                     if not hook.is_valid:
                         self.log.debug(
-                            "Hook is not valid for current "
-                            "launch context: {}".format(klass.__name__)
+                            "Skipped hook invalid for current launch context: "
+                            "{}".format(klass.__name__)
                         )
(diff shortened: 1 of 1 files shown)


openpype/lib/applications.py
```diff
@@ -1044,7 +1044,8 @@ def discover_launch_hooks(self, force=False):
                     hook = klass(self)
                     if not hook.is_valid:
                         self.log.debug(
-                            "Hook is not valid for current launch context."
+                            "Hook is not valid for current "
+                            "launch context: {}".format(str(hook))
```
iLLiCiTiT, line 1048: I would change the `str(hook)` to `klass.__name__` to avoid long names in logs:
- e.g. `<pre_copy_template_workfile.CopyTemplateWorkfile object at 0x000001D52BD48388>`

BigRoy: Agreed. Will also change the lines below to match that.

Issue #3098, opened 2022-04-25
Houdini bgeo publishing


antirotor:
## Feature

Add bgeo publishing/loading to Houdini. Support for `bgeo` and `bgeo.sc`.

[cuID:OP-3129]

BigRoy:
Has work on this started yet? If not, I'd be happy to set up the barebones functionality for this.

antirotor:
Nope, nothing is done yet.

antirotor:
FYI I've just added for one client very simple loader of bgeo to houdini. Nothing fancy. Publishing from standalone publisher has one limitation - it doesn't support `bgeo.sc` adn `bgeo.gz` but only `bgeosc` and `bgeogz` due to the limitations we have of handling file extentions.

BigRoy:
Due to other priorities and production needs on our end lately I don't feel me as assignee is the best candidate for this on short-term - sorry about that.

PR #3153, opened 2022-05-06
Compressed bgeo publishing in SAP and Houdini loader

antirotor:
## Enhancement

This PR is adding support for standalone publisher to publish `.bgeosc` and `.bgeogz` (compressed bgeo sequences/files). It is also adding basic loader for Houdini for `bgeo*`

Related to #3098


m-u-r-p-h-y (approved):
- bgeosc sequence created and published via Standalone publisher
- the published sequence is available in Loader 

image

PR #3162, opened 2022-05-10
add silent audio to slate

jrsndl:
## Brief description
Allows Nuke slate with audio

## Description
Slate is concatenated with review using ffmpeg concat. Concatenation requires same number, type and layout of streams in concatenated files. Adding silence to the slate before concatenation allows to keep the audio in the review.

## Additional info

- Some review files do not support audio. Make sure you use Do not add audio tag to avoid encoding fail.
- It seems that some file formats produce small audio offsets. That might be a limitation of concat.

## Testing notes:
1. make sure you have an audio available
2. setup a review with Add Slate Frame tag
3. Publish from Nuke, with slate
4. Check that review file has slate and audio

### DIFF
--- a/openpype/plugins/publish/extract_review_slate.py
+++ b/openpype/plugins/publish/extract_review_slate.py
@@ -79,7 +79,7 @@ class ExtractReviewSlate(openpype.api.Extractor):
             )
             # Get video metadata
             for stream in streams:
-                input_timecode = None
+                input_timecode = ""
                 input_width = None
                 input_height = None
                 input_frame_rate = None
@@ -89,18 +89,20 @@ class ExtractReviewSlate(openpype.api.Extractor):
                         tags = stream.get("tags") or {}
                         input_timecode = tags.get("timecode") or ""
                         if "width" in stream and "height" in stream:
-                            input_width = int(stream.get("width"))
-                            input_height = int(stream.get("height"))
+                            input_width = stream.get("width")
+                            input_height = stream.get("height")
                         if "r_frame_rate" in stream:
                             # get frame rate in a form of
                             # x/y, like 24000/1001 for 23.976
-                            input_frame_rate = str(stream.get("r_frame_rate"))
+                            input_frame_rate = stream.get("r_frame_rate")
                         if (
-                            input_timecode
-                            and input_width
+                            input_width
                             and input_height
                             and input_frame_rate
                         ):
+                            input_width = int(input_width)
+                            input_height = int(input_height)
+                            input_frame_rate = str(input_frame_rate)
                             break
             # Raise exception of any stream didn't define input resolution
             if input_width is None:
@@ -131,7 +133,10 @@ class ExtractReviewSlate(openpype.api.Extractor):
                         and audio_channel_layout
                         and audio_codec
                     ):
-                        input_audio = True
+                        input_audio = str(input_audio)
+                        audio_sample_rate = str(audio_sample_rate)
+                        audio_channel_layout = str(audio_channel_layout)
+                        audio_codec = str(audio_codec)
                         break
             # Get duration of one frame in micro seconds
             one_frame_duration = "40000us"
@@ -207,7 +212,7 @@ class ExtractReviewSlate(openpype.api.Extractor):
             input_args.extend(["-r {}".format(input_frame_rate)])
             input_args.extend(["-frames:v 1"])
             # add timecode from source to the slate, substract one frame
-            offset_timecode = "00:00:00:00"
+            offset_timecode = ""
             if input_timecode:
                 offset_timecode = str(input_timecode)
                 offset_timecode = self._tc_offset(
@@ -311,7 +316,8 @@ class ExtractReviewSlate(openpype.api.Extractor):
             # Create slate with silent audio track
             if input_audio:
                 # silent slate output path
-                slate_silent_path = slate_path.replace(".png", "Silent" + ext)
+                slate_silent_path = "_silent".join(
+                    os.path.splitext(slate_v_path))
                 _remove_at_end.append(slate_silent_path)
 
                 slate_silent_args = [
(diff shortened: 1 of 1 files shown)


openpype/plugins/publish/extract_review_slate.py
```diff
@@ -314,6 +308,41 @@ def process(self, instance):
...
+                    slate_silent_path
+                ]
+                slate_silent_subprocess_cmd = " ".join(slate_silent_args)
+                # run slate generation subprocess
+                self.log.debug(
+                    "Silent Slate Executing: {}".format(
+                        slate_silent_subprocess_cmd))
+                openpype.api.run_subprocess(
+                    slate_silent_subprocess_cmd, shell=True, logger=self.log
+                )
```
iLLiCiTiT, line 341:
```suggestion
                openpype.api.run_subprocess(
                    slate_silent_args, logger=self.log
                )
```

openpype/plugins/publish/extract_review_slate.py
```diff
@@ -314,6 +308,41 @@ def process(self, instance):
...
+                    "anullsrc=r={}:cl={}:d={}".format(
+                        audio_sample_rate,
+                        audio_channel_layout,
+                        one_frame_duration
+                    ),
+                    "-c:v", "copy",
+                    "-c:a", audio_codec,
+                    "-map", "0:v",
+                    "-map", "1:a",
+                    "-shortest",
```
iLLiCiTiT, line 330: There are missing format arguments.

openpype/plugins/publish/extract_review_slate.py
```diff
@@ -314,6 +308,41 @@ def process(self, instance):
                 slate_subprocess_cmd, shell=True, logger=self.log
             )
 
+            # Create slate with silent audio track
+            if input_audio:
```
iLLiCiTiT, line 312: How this second "silent" output is integrated? Because it's created but unused...

openpype/plugins/publish/extract_review_slate.py
iLLiCiTiT: Maybe irrelevat comment if this is true https://github.com/pypeclub/OpenPype/pull/3162/files#r869999097

openpype/plugins/publish/extract_review_slate.py
```diff
@@ -122,10 +122,14 @@ def process(self, instance):
                     if stream["channel_layout"]:
                         audio_channel_layout = str(
                             stream.get("channel_layout"))
+                    if stream["codec_name"]:
+                        audio_codec = str(
+                            stream.get("codec_name"))
```
iLLiCiTiT, line 127: Value is used for check if there is a codec. But if `None` is converted to string it will be `"None"` which will pass the condition even if there is not information about it.
```suggestion
                        audio_codec = stream.get("codec_name")
```
The same should be changed for `channel_layout` `sample_rate` and `channels`.

openpype/plugins/publish/extract_review_slate.py
```diff
@@ -314,6 +308,41 @@ def process(self, instance):
                 slate_subprocess_cmd, shell=True, logger=self.log
             )
 
+            # Create slate with silent audio track
+            if input_audio:
+                # silent slate output path
+                slate_silent_path = slate_path.replace(".png", "Silent" + ext)
```
iLLiCiTiT, line 314: Longer but more safe proof to reuse `slate_v_path`.
```suggestion
                basename, ext = os.path.splitext(slate_v_path)
                slate_silent_path = "{}_silent{}".format(basename, ext)
```

Or singleline option
```
slate_silent_path = "_silent".join(os.path.splitext(slate_v_path))
```

openpype/plugins/publish/extract_review_slate.py
```diff
@@ -314,6 +308,41 @@ def process(self, instance):
...
+                # run slate generation subprocess
+                self.log.debug(
+                    "Silent Slate Executing: {}".format(
+                        slate_silent_subprocess_cmd))
+                openpype.api.run_subprocess(
+                    slate_silent_subprocess_cmd, shell=True, logger=self.log
+                )
+
+                # replace slate with silent slate for concat
+                slate_v_path = slate_silent_path
```
iLLiCiTiT, line 344: The source `slave_v_path` should be probably removed?

jrsndl: Not sure if I follow. `slate_v_path`  is used for encoding the slate video. If audio is present in the main review, `slate_v_path` is used as a source for `slate_silent_path`. In the following code, concat needs to use either `slate_v_path` or `slate_silent_path` depending on if audio is present or not. Therefore I overwritten the `slate_v_path` by `slate_silent_path` in the end of `if input_audio` block.

iLLiCiTiT: No, you change the filepath to `{dir}/{basename}Silent{ext}` (https://github.com/pypeclub/OpenPype/pull/3162/files#r870019552).

openpype/plugins/publish/extract_review_slate.py
jrsndl: will do

openpype/plugins/publish/extract_review_slate.py
jrsndl: will do

openpype/plugins/publish/extract_review_slate.py
jrsndl: will do

openpype/plugins/publish/extract_review_slate.py
jrsndl: Feeling more dumb than normal today. IMHO it is cleaned together with other in-betweens when not needed:
https://github.com/pypeclub/OpenPype/blob/b0c9e8a29d4c16d3276989fee37c8d8762d67809/openpype/plugins/publish/extract_review_slate.py#L414

'cause `slate_v_path` is appended to _remove_at_end
https://github.com/pypeclub/OpenPype/blob/b0c9e8a29d4c16d3276989fee37c8d8762d67809/openpype/plugins/publish/extract_review_slate.py#L294

I just tested, and found no leftovers after publish

openpype/plugins/publish/extract_review_slate.py
```diff
@@ -89,18 +89,20 @@ def process(self, instance):
                         tags = stream.get("tags") or {}
                         input_timecode = tags.get("timecode") or ""
                         if "width" in stream and "height" in stream:
-                            input_width = int(stream.get("width"))
-                            input_height = int(stream.get("height"))
+                            input_width = stream.get("width")
+                            input_height = stream.get("height")
```
iLLiCiTiT, line 93: This is actually only place where the `int` conversion should be done here (to check if it's not `0`).

jakubjezek001 (approved):
tested and works fine

PR #3164, opened 2022-05-11
Fix for attaching render to subset

### LINKED ISSUE #2753: Maya: Attach Rende to Look
## Problem

Attaching render to look instance doesn't work.

antirotor:
## Fix

This is fixing issue when renders couldn't be attached as a review to published subsets.

### Old behaviour:

- Create look
- Create Render
- Drag Look instance under render layer set in Render instance
- Set your render settings as needed
- Publish

Look will be published just fine, job will be submited to farm and when complete, render will be published as render. It will be shown on ftrack as render asset type and you won't find it in Maya Loader.

This is fixing it, render should attach correctly and look subset should be found along others in Loader as usual.

Close #2753


antirotor:
> Still grouping under renderLightingMain

should be fixed now

m-u-r-p-h-y (approved):
The published look is listed correctly in Loader now . . .

image

PR #3166, opened 2022-05-11
General: Oiio conversion for ffmpeg checks for invalid characters

iLLiCiTiT:
## Brief description
Fix check for invalid characters in oiio conversion.

## Description
Check for invalid characters never happened because of invalid condition before.

## Testing notes:
1. Create renders with `"` in some metadata attribute but shorted then 8196 characters
2. Make it go through ExtractReview plugin
3. It should not crash on ffmpeg conversion


m-u-r-p-h-y (approved):
Submitted job from Maya2022 using Arnold to render cryptomatte exrs

Publish and extract review works as expected, all mp4 created.

image

PR #3169, opened 2022-05-12
General/Hiero: better clip duration calculation

jakubjezek001:
## Brief description
Improving clip duration calculation

## Description
Frame duration should not be excluding one frame as it is disturbing other hosts. Better to fix it in global then hack particular dcc (flame)

## Testing notes:
1. open your testing hiero script
2. publish a clip
3. all should look as expected


jrsndl:
Not quite sure how to properly check the duration. I compared published OTIO from Hiero to old one, and found the durations the same and correct.

mkolar (approved):

mkolar:
This is certainly a better approach

PR #3171, opened 2022-05-12
Enhancement/op 2598 blender dynamic workflow using linked lib

kaamaurice:
# PR Moved
- It was to messy with all the commits to rebase properly so please follow this PR now : #3463

### Brief description
- Use collections and overrides for the instances.

### Description
- Change Blender plugins to work with a collection instead of an empty to store custom properties of the avalon instance.
use overrides.
https://github.com/pypeclub/OpenPype/discussions/2598

- Based on the work of [jojoba106](https://github.com/jojoba106)
https://github.com/pypeclub/OpenPype/pull/2771


openpype/hosts/blender/plugins/publish/extract_blend.py
```diff
@@ -23,24 +24,39 @@ def process(self, instance):
...
 
+        plugin.deselect_all()
+
         data_blocks = set()
 
         for obj in instance:
             data_blocks.add(obj)
+            # Skip and non Object type like Collections.
+            if not isinstance(obj, bpy.types.Object):
+                continue
```
Tilix4, line 35: If continuing, the `add` shouldn't be processed.
```suggestion
            # Skip and non Object type like Collections.
            if not isinstance(obj, bpy.types.Object):
                continue
            data_blocks.add(obj)
```

kaamaurice: Actually I want to add all instance data_blocks but juste skip for non object type to prevent invalid attribute error with obj.type because Collection hasn't this property. I use a "not condition" with continue to spare increments. I change this with simple condition.

openpype/hosts/blender/api/pipeline.py
```diff
@@ -372,8 +371,11 @@ def ls() -> Iterator:
     called containers.
     """
 
-    for container in lib.lsattr("id", AVALON_CONTAINER_ID):
-        yield parse_container(container)
+    collections = lib.lsattr("id", AVALON_CONTAINER_ID)
+    scene_collections = list(bpy.context.scene.collection.children_recursive)
```
Tilix4, line 375: I'm not sure casting it to `list` is necessary. It seems to work without.

openpype/hosts/blender/api/plugin.py
```diff
@@ -88,6 +97,102 @@ def create_blender_context(active: Optional[bpy.types.Object] = None,
...
+        if color_tag:
+            container.color_tag = color_tag
+        bpy.context.scene.collection.children.link(container)
+        return container
+
+
+def remove_container(container, content_only=False):
+    """
+    Remove the container with all this objects and child collections.
+
```
Tilix4, line 118: Please document this fuction's arguments

openpype/hosts/blender/api/plugin.py
```diff
@@ -88,6 +97,102 @@ def create_blender_context(active: Optional[bpy.types.Object] = None,
...
+            objects_to_remove.add(container)
+
+    # Remove objects
+    for obj in objects_to_remove:
+        # Append object data if exists.
+        if obj.data:
+            data_to_remove.add(obj.data)
+        obj.name = f"{obj.name}.removed"
+        bpy.data.objects.remove(obj)
+    # Remove objects
```
Tilix4, line 156:
```suggestion
    # Remove collections
```

openpype/hosts/blender/api/plugin.py
```diff
@@ -281,6 +543,152 @@ def _load(self,
...
+        will not be removed, only unlinked. Normally this should not be the
+        case though.
+
+        Warning:
+            No nested collections are supported at the moment!
+        """
+        object_name = container["objectName"]
+        asset_group = (
+            bpy.data.collections.get(object_name) or
+            bpy.data.objects.get(object_name)
```
Tilix4, line 616: Is it now possible an object will be used as an asset_group?

openpype/hosts/blender/plugins/publish/extract_blend.py
```diff
@@ -23,24 +50,28 @@ def process(self, instance):
         # Perform extraction
         self.log.info("Performing extraction..")
 
+        plugin.deselect_all()
+
         data_blocks = set()
+        objects = set()
 
         for obj in instance:
```
Tilix4, line 58: Could you please add a comment here? I don't get explicitly the purpose of this loop...

openpype/hosts/blender/api/pipeline.py
kaamaurice: Yes, you 're right, I had in memory this :
```python
"collection_name" in bpy.context.scene.collection.children
```
Is valid but,
```python
collection in bpy.context.scene.collection.children
```
while raise :
`bpy_prop_collection.__contains__: expected a string or a tuple of strings`

but `children_recursive` return list indeed

openpype/hosts/blender/plugins/publish/extract_blend.py
kaamaurice: done

openpype/hosts/blender/api/plugin.py
kaamaurice: rectified

openpype/hosts/blender/api/plugin.py
kaamaurice: done

openpype/hosts/blender/api/plugin.py
kaamaurice: Only with model representation loaded in layout or setdress (actually it is a very early behavior, we need to discuss this feat)

Tilix4: With the current system, you containerize always with a `Collection` no? Then I don't see how the case you describe may happen.

openpype/hosts/blender/plugins/load/load_model.py
```diff
@@ -30,78 +21,81 @@ class BlendModelLoader(plugin.AssetLoader):
...
-            if obj.type == 'MESH':
-                for material_slot in list(obj.material_slots):
-                    bpy.data.materials.remove(material_slot.material)
-                bpy.data.meshes.remove(obj.data)
-            elif obj.type == 'EMPTY':
-                objects.extend(obj.children)
-                bpy.data.objects.remove(obj)
-
-    def _process(self, libpath, asset_group, group_name):
+    def _load_blend(self, libpath, asset_group):
```
Tilix4, line 28: Can't all these `_load_blend` functions be DRYer? There is a lot of duplicate code accross all `load_.py` and `plugin.py` files.

PR #3173, opened 2022-05-12
Traypublisher: basic implementation of transcoding workflow

kalisp:
## Additional info
Temporary placement of Creator for as Addons currently cannot introduce its own Creators until https://github.com/pypeclub/OpenPype/issues/3165

This PR is only for testing, shouldn't be merged.


kalisp:
Shouldn't be merged at all.

PR #3178, opened 2022-05-13
Feature/multiverse

pberto:
## Brief description
Finalized Multiverse integration.

## Description
- updating labels for the various tools as per discussed
- first version of the documentation on website
- added jcube to op contributors on main page of website
- Adding support for choosing .usd/.usda/.usdz.
- fixed empty composition due to relative paths
- formatting / cosmetics

## Documentation

This PR includes also documentation.

## Notes

We will add a Multiverse Look creator at the next PR.


openpype/hosts/maya/plugins/publish/extract_multiverse_usd.py
```diff
@@ -129,13 +131,19 @@ def parse_overrides(self, instance, options):
 
         return options
 
+    def get_file_format(self, instance):
```
iLLiCiTiT, line 134: You call `get_file_format` but it does not return a thing. Instead it is changing attribute of object so if instance does have empty fileFormats it will use `scene_type` of previous instance.

Also not sure what the code should do? If it would be rewritten to simplified version:
```
instance_file_format = ["usd", "usda"]
file_formats = ["usd", "usda", "usdz"]
some_range = range(len(file_formats)) # that's the same as [0, 1, 2]

# This will never be true as the list with file formats won't be in iter of itegers
if instance_file_format in some_range:
    # You change attribut on the class instead of returning it
    # + it would crash if it would get here because it's using
    #    a list to get item from list (TypeError)
    self.scene_type = file_formats[instance_file_format]
```

It kinda looks like you wanted to use `for` instead of `if`? But that's anti-pythonic way of looping through list. Also only last item would be used in that case it would be better to use:
```
# Backwards compatibility for instances without fileFormat
file_format = instance.data.get("fileFormat")
# Check if is set
if file_format:
    # return last item of the list
    return file_format[-1]
# Return default scene type
return self.default_file_format
```

BigRoy: Why would be `instance.data["fileFormat"]` be a list or tuple in the first place - that sounds confusing?

Why can't it be this?
```python
file_format = instance.data.get("fileFormat",
                                self.default_file_format)
```

Or even a hardcoded backwards compatible default:
```python
file_format = instance.data.get("fileFormat", 
                                # Backwards compatibility
                                "usd")
```

With this simplification it doesn't need its own `get_file_format` method. Nonetheless, like mentioned before.. please return the value instead of storing it on the class if you do end up using a defined method for this.

openpype/hosts/maya/plugins/publish/extract_multiverse_usd.py
```diff
@@ -199,10 +201,10 @@ def process(self, instance):
             instance.data["representations"] = []
 
         representation = {
-            'name': 'usd',
-            'ext': 'usd',
+            'name': self.scene_type,
```
BigRoy, line 204: @iLLiCiTiT would you feel it's wrong if `name` would always be `usd` no matter the extension? That way updating in the manager can be done even if file format changed in-between. Thoughts?

BigRoy:
> We will add a Multiverse Look creator at the next PR.

It might be confusing that the current documentation does already mention Multiverse Looks (e.g. in screenshots). Do you have an ETA on that next PR?

openpype/hosts/maya/plugins/publish/collect_multiverse_look.py
```diff
@@ -189,6 +192,22 @@ def get_file_node_files(node):
...
+def get_mipmap(fname):
+    for colour_space in COLOUR_SPACES:
+        for mipmap_ext in MIPMAP_EXTENSIONS:
+            mipmap_fname = '.'.join([fname, colour_space, mipmap_ext])
+            if os.path.exists(mipmap_fname):
+                return mipmap_fname
+    return None
+
+
+def is_mipamp(fname):
```
BigRoy, line 204: typo
```suggestion
def is_mipmap(fname):
```

dmo-j-cube: Thanks!! Fixed it!!

openpype/hosts/maya/plugins/publish/extract_multiverse_usd.py
dmo-j-cube: Sorry for the delay. For simplicity, I did your suggestion of using that `get` method.

pberto:
@mkolar please note I refactored a bit the Maya docs as the page was too long. I personally think this new organization is easier to read for the user and is also faster to find content.

mkolar:
>@mkolar please note I refactored a bit the Maya docs as the page was too long. I personally think this new organization is easier to read for the user and is also faster to find content.

Thank you having a look

m-u-r-p-h-y:
The link in the documentation is not valid

image

BigRoy:
> but please first change the family names to be prefixed with mv

Are we sure? The data is just regular `usd` data that is generated in the end. Nothing Multiverse about that - except for the published multiverse looks that is.

To load the multiverse USDs into Houdini I'd then have to allow them to load `mvUsd` data instead of regular USD data. This will spread much further than we'd probably like?

I admit the Loaders themselves (so that loaded content can be managed correctly in the future) are `MultiverseUsdLoader`, etc. but they already are. 👍

mkolar:
MV structures the USD data in certain way that makes sense from it's perspective and even though they are just native USD files, it's kinda like ditching `model` family and instead we start using `alembic` instead. Yes it's universal, but it removes all structure.

mkolar:
Houdini implementation at the moment is a good example of such mixup. We have `usdCamera` which makes no sense. It should be `Camera` family that has `USD` representation, albeit generic USD scene family will be immensely useful as fallback, the same way as `mayaScene` is. The question is, what to call is `USD`, `usdScene`, usdStage`? And it doesn't really fit what `usdComposition` and `usdOverrides` are as they have very specific structure that could be validated

BigRoy:
Thanks for the follow up - I understand. 👍 Definitely worth a good brainstorm on other potential approaches for this across the whole "family". (Pun intended.)

mkolar (changes requested):
please check comment in https://github.com/pypeclub/OpenPype/pull/3178#issuecomment-1147591757

mkolar (approved):
Thank you guys. This is great and I think we can squeeze it into **3.11.0**

PR #3179, opened 2022-05-13
General: Creator plugins from addons can be registered

iLLiCiTiT:
## Brief description
Addons can register creator plugin paths and Creators can have host name filter.

## Description
Creator plugins have host name filtering so e.g. tray publisher creators won't appear in other hosts. Addon can implement new method which return creator plugin paths for specific host name. Default implementation is using `get_plugin_paths` for backwards compatibility. Installation of openpype plugins will also register creators from modules. Creator plugin has also ability to define host name filter.

## Additional information
We wanted to have filtering ability of creator plugins on both addon and creator side so creator plugin has `host_name` attribute which can define on which host is working. This is for backwards compatibility not abstract. But it is a question if it should be required? In most of cases it is only matter of addon implementation where this filter is "required" (It probably won't happen that host on it's own will register plugins for different host). Also at this time are new creators used only in 3 hosts so it could be easy to change it.

## Testing notes:
1. Create addon and add it to addons path
2. Make sure the addon inherit from `IPluginPaths`
3. Create some creator plugin for TrayPublisher inside the addon
4. Make sure addon implement `get_creator_plugin_paths` which returns path to the creator dir
5. Run tray publisher
6. The creator should be visible there

Resolves https://github.com/pypeclub/OpenPype/issues/3165

### DIFF
--- a/openpype/hosts/traypublisher/api/plugin.py
+++ b/openpype/hosts/traypublisher/api/plugin.py
@@ -14,6 +14,7 @@ from .pipeline import (
 
 class TrayPublishCreator(Creator):
     create_allow_context_change = True
+    host_name = "traypublisher"
 
     def collect_instances(self):
         for instance_data in list_instances():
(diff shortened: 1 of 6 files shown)


kalisp (approved):
Working for creator in add on.

PR #3180, opened 2022-05-13
General: OpenPype modules publish plugins are registered in host

iLLiCiTiT:
## Brief description
Module's publish plugin paths are registered during installation of OpenPype plugins.

## Description
Changed how module plugin paths are registered. Instead of using environment variables are plugin paths registered during installation of OpenPype plugins which is more suitable and possible dynamic.

## Additional info
This approach gives more dynamic approach of registering publish plugins (e.g. does not require restart of tray when module change state). Can be extended in future to install plugin paths per project and more dynamically, which is not as easy now because modules' plugin paths are there from moment when openpype starts.

PR is based on https://github.com/pypeclub/OpenPype/pull/3179 where creator plugins are installed same way. I wanted to separate the PRs as this change should be fully tested.

## Testing notes:
1. All modules should register their publish plugins in all hosts
2. The same must work for farm publishing, standalone publishing and webpublisher

### DIFF
--- a/openpype/pipeline/context_tools.py
+++ b/openpype/pipeline/context_tools.py
@@ -119,7 +119,7 @@ def install_openpype_plugins(project_name=None, host_name=None):
     modules_manager = ModulesManager()
     publish_plugin_dirs = modules_manager.collect_plugin_paths()["publish"]
     for path in publish_plugin_dirs:
-        pyblish.api.register_plugin_path(PUBLISH_PATH)
+        pyblish.api.register_plugin_path(path)
 
     if host_name is None:
         host_name = os.environ.get("AVALON_APP")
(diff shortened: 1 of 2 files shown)


openpype/pipeline/create/context.py
```diff
@@ -861,6 +865,17 @@ def reset_plugins(self, discover_publish_plugins=True):
...
+
+            # Filter by host name
+            if (
+                creator_class.host_name
+                and creator_class.host_name != self.host_name
+            ):
+                self.log.info((
+                    "Creator's host name is not supported for current host {}"
+                ).format(creator_class.host_name, self.host_name))
+                continue
```
BigRoy, line 877: For consistency, shouldn't we support similar host name filtering like Pyblish does? Like allowing `hosts = ["maya", "houdini"]` and maybe have default value of `hosts = ["*"]`?

iLLiCiTiT: I don't think you can have same creator plugins for 2 different hosts....

BigRoy: Ah, of course that might make a lot less sense.

kalisp (approved):
Tested in AE,PS, Nuke, Standalone publisher

PR #3181, opened 2022-05-13
Maya: added clean_import option to Import loader

kalisp:
## Brief description
By selecting this option all occurrences of cbId of imported nodes will be removed.

## Description
Usable for sharing assets between projects.

## Testing notes:
1. Load model via `Import` loader - set `Clean import` option (rectangle after Import label)
2.`cbId` in `Extra Attributes` should be removed (also check log for 'Removing ..' in Script Editor)

### DIFF
--- a/openpype/hosts/maya/plugins/load/actions.py
+++ b/openpype/hosts/maya/plugins/load/actions.py
@@ -1,7 +1,7 @@
 """A module containing generic loader actions that will display in the Loader.
 
 """
-
+import qargparse
 from openpype.pipeline import load
 from openpype.hosts.maya.api.lib import (
     maintained_selection,
@@ -98,12 +98,23 @@ class ImportMayaLoader(load.LoaderPlugin):
     icon = "arrow-circle-down"
     color = "#775555"
 
+    options = [
+        qargparse.Boolean(
+            "clean_import",
+            label="Clean import",
+            default=False,
+            help="Should all occurences of cbId be purged?"
+        )
+    ]
+
     def load(self, context, name=None, namespace=None, data=None):
         import maya.cmds as cmds
 
         choice = self.display_warning()
         if choice is False:
             return
+
+        clean_import = data.get("clean_import", False)
 
         asset = context['asset']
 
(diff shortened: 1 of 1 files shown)


m-u-r-p-h-y (changes requested):
The main purpose of this option is to simplify the reuse of assets (from other projects, assets, etc.)

We have to make sure the attributes are removed from all nodes, especially sets, curves (from rigging), etc.

It seems the attribute is removed from the geometry shape only.



image

openpype/hosts/maya/plugins/load/actions.py
```diff
@@ -114,13 +125,21 @@ def load(self, context, name=None, namespace=None, data=None):
...
+                              preserveReferences=True,
+                              namespace=namespace,
+                              returnNewNodes=True,
+                              groupReference=True,
+                              groupName="{}:{}".format(namespace, name))
+
+            if clean_import:
+                shapes = cmds.ls(nodes, shapes=True, long=True)
+                for shape in shapes:
+                    meshes = cmds.ls('{}.cbId'.format(shape))
```
BigRoy, line 139: This should be `cmds.attributeQuery("cbId", node=shape, exists=True)` to be more explicit.
I'd also disregard the filtering to shapes and just check on all nodes because transforms of shapes also get a cbId.
```python
remove_attributes = ["cbId"]
for node in nodes:
    for attr in remove_attributes:
        if cmds.attributeQuery(attr, node=node, exists=True):
            cmds.deleteAttr("{}.{}".format(attr))

BigRoy:
> Usable for sharing assets.

Is it? Because if saving the scene in OpenPype again it will auto-generate new ids. So it's not really easy to use for sharing with e.g. another team outside of the pipeline? Or did you mean sharing assets internally between projects?

m-u-r-p-h-y (approved):
Imported couple of different assets (models, rigs)

and all cbId attributes and all entities were removed  successfully
- shapes
- transform nodes
- sets
- rigs

tested with M2023

PR #3198, opened 2022-05-17
Maya: Fix Yeti errors on Create, Publish and Load

BigRoy:
## Brief description

I was unable to create a Yeti rig publish and load it correctly so I started working on fixing the implementation.

**This is still a draft WIP - not fully tested.**

Current status:

-  [x] Creators work
-  [x] Extractors work
-  [x] Loaders work

### Questions

Currently the Yeti Rig publish will also try to create a single frame Yeti Cache alongside the rig publish - which as far as I know isn't required for a Yeti Rig to work. Would it make sense to remove that logic and if one would ever want both a "rig" and a "cache" from that same file they would just create two publish instances, one `yetiRig` family and one `yetiCache` family.

---

The way it's currently broken it seems like this has hardly ever been tested - or maybe has seen numerous changes in other areas of OpenPype but hasn't been tested in quite some time and thus has been unused too.

Would it be worth maybe refactoring the family name `yeticache` to `yetiCache` for readability?

## Additional info

I'm currently testing with Maya 2022 + Yeti 4.1.4

## Testing notes:
1. Create a Yeti Rig and Yeti Cache
2. Publish it.
3. Load it.

Preferably also test cases that uses textures/files in the Yeti Graph since those should be published along similar to Lookdev publishes.

### DIFF
--- a/openpype/hosts/maya/plugins/create/create_yeti_cache.py
+++ b/openpype/hosts/maya/plugins/create/create_yeti_cache.py
@@ -22,7 +22,8 @@ class CreateYetiCache(plugin.Creator):
         # Add animation data without step and handles
         anim_data = lib.collect_animation_data()
         anim_data.pop("step")
-        anim_data.pop("handles")
+        anim_data.pop("handleStart")
+        anim_data.pop("handleEnd")
         self.data.update(anim_data)
 
         # Add samples
--- a/openpype/hosts/maya/plugins/load/load_yeti_cache.py
+++ b/openpype/hosts/maya/plugins/load/load_yeti_cache.py
@@ -1,15 +1,13 @@
 import os
 import json
 import re
-import glob
 from collections import defaultdict
-from pprint import pprint
 
+import clique
 from maya import cmds
 
 from openpype.api import get_project_settings
 from openpype.pipeline import (
-    legacy_io,
     load,
     get_representation_path
 )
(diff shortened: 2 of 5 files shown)


BigRoy:
This appears to be more broken than I anticipated - might take some more time than a quick fix now. Will look into it later this week.

BigRoy:
The biggest issue I'm facing now is this:

The Yeti implementation is intended to allow publishing of Yeti Rigs and Yeti Caches which exist of _potentially_ multiple `pgYetiMaya` nodes with each their own `.fur` sequence export in the cache.

These **should** then all be published along to the destination path:
```
{subset}.rigsettings       # (only for Yeti Rig family)
{subset}.fursettings       # fur cache settings (like display/render width, etc.)
body.1000.fur
body.1001.fur
head.1000.fur
head.1001.fur
hands.1000.fur
hands.1001.fur
```

However due to how the files are added to the representation: `representation["files"] = generated_files` this means that the Integrator will see it as a sequence and rename it to the destination path/template based on the subset name.

What would be the way to go about publishing something like this?

BigRoy:
Should be working now by using explicit "transfers" for the sequences which are not being renamed at all. 👍 Not super pretty and nice - but it works.

antirotor:
> Should be working now by using explicit "transfers" for the sequences which are not being renamed at all. 👍 Not super pretty and nice - but it works.

I am afraid there's currently probably no other way around it.

antirotor:
So is it ready for review?

mkolar:
BTW, it's exactly as you said @BigRoy this has been used and supported in production in version 2 the last I remember. Since then a lot has changed and the funding for yeti dried up on the side of the studio that originally needed it. Unfortunate indeed.

BigRoy:
No problem - we've just used Yeti in a very simple project here within OpenPype just fine. I believe it was exactly this PR that worked for us. :)

iLLiCiTiT (approved):
We cannot test it and code looks ok, so we decided that we'll trust it. Since it's affecting only yeti code which is now 100% broken we have nothing to loose.

PR #3199, opened 2022-05-17
Loader UI: Speed issues of loader with sync server

iLLiCiTiT:
## Brief description
A little bit enhanced loader UI speec when sync server is enabled.

## Description
Cache sync server related information for some time to avoid refresh of sync server on each selection change.

## Additional info
Loader was refreshing sync server module, it's setting for all projects on each change of context in loader UI which is also happening separatelly for subset widget and representation widget.

This is a "quick" change, full fix would require to modify some logic in sync server and partially rewrite how loader UI logic works.

## Testing notes:
Loader didn't change but should be faster when sync server is enabled.

### DIFF
--- a/openpype/tools/loader/model.py
+++ b/openpype/tools/loader/model.py
@@ -1,6 +1,7 @@
 import copy
 import re
 import math
+import time
 from uuid import uuid4
 
 from Qt import QtCore, QtGui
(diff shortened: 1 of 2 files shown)


openpype/tools/loader/model.py
```diff
@@ -47,28 +56,53 @@ def reset_sync_server(self, project_name=None):
...
+                active_provider = 'studio'
+
+            remote_site = sync_server.get_remote_site(project_name)
+            remote_provider = sync_server.get_provider_for_site(
+                project_name, remote_site)
+            if remote_site == 'studio':  # for studio use explicit icon
+                remote_provider = 'studio'
 
-                repre_icons = lib.get_repre_icons()
+            repre_icons = lib.get_repre_icons()
```
BigRoy, line 105: Would you want to cache this too?

iLLiCiTiT: I don't know. I just made quick fix so it's usable in production...

BigRoy:
> This is a "quick" change, full fix would require to modify some logic in sync server and partially rewrite how loader UI logic works.

If we are merging this lets keep an issue open somewhere to remind us of doing that cleanup.

64qam (approved):

kalisp (approved):

PR #3200, opened 2022-05-17
Loader UI: Speed issues of loader with sync server

iLLiCiTiT:
## Brief description
A little bit enhanced loader UI speec when sync server is enabled.

## Description
Cache sync server related information for some time to avoid refresh of sync server on each selection change.

## Additional info
Loader was refreshing sync server module, it's setting for all projects on each change of context in loader UI which is also happening separatelly for subset widget and representation widget.

This is a "quick" change, full fix would require to modify some logic in sync server and partially rewrite how loader UI logic works.

## Testing notes:
Loader didn't change but should be faster when sync server is enabled.


64qam (approved):
It is approx. two times faster now

PR #3214, opened 2022-05-19
Deadline: instance data overwrite fix

antirotor:
## Bug description

submission of publishing job to deadline had a serious bug for some quite time: It is creating new instances based on stuff that is being rendered. For this it is using some skeleton data that are part of every newly created instance.

When the instance was being created, it `copy()` the skeleton data but that is only shallow copy. Anything that was modifying data on newly created instance edited them in skeleton data too, influencing all other newly created instances.

### Test case:

```python
from copy import copy


skeleton = {
   "name": "A",
   "families": []
}

new_instance_names = ["B", "C", "D"]

new_instances = []
for new_name in new_instance_names:
    new_instance = copy(skeleton)
    new_instance["name"] = new_name
    if new_name == "C":
        new_instance["families"].append("FOO")
    new_instances.append(new_instance)

for instance in new_instances:
    print(f"{instance['name']}")
    print(f"{instance['families']}")
    print("-" * 10)
```

This will result in:
```
B
['FOO']
----------
C
['FOO']
----------
D
['FOO']
----------
```
All instances have `FOO` even if we expect ony **C** instance to have it.

This PR is solving it by changing `copy()` to `deepcopy()`

### DIFF
--- a/openpype/modules/deadline/plugins/publish/submit_publish_job.py
+++ b/openpype/modules/deadline/plugins/publish/submit_publish_job.py
@@ -466,7 +466,7 @@ class ProcessSubmittedJobOnFarm(pyblish.api.InstancePlugin):
             if instance_data.get("multipartExr"):
                 preview = True
 
-            new_instance = copy(instance_data)
+            new_instance = deepcopy(instance_data)
             new_instance["subset"] = subset_name
             new_instance["subsetGroup"] = group_name
             if preview:


BigRoy (approved):
This is definitely much safer!

m-u-r-p-h-y (approved):
B
[]
----------
C
['FOO']
----------
D
[]
----------

PR #3216, opened 2022-05-19
Project manager: Sped up project load

iLLiCiTiT:
## Brief description
Speed of project loading was enhanced by ignoring signals during initial project item creation.

## Description
Blocking of signals during that time cause that filtering and repaint of view is not triggered which makes everything much smoother. Also refresh button forces to refresh current project.

## Testing notes:
1. Open project manager
2. Select project
3. Load of project entities should be 3x faster

### DIFF
--- a/openpype/tools/project_manager/project_manager/model.py
+++ b/openpype/tools/project_manager/project_manager/model.py
@@ -264,6 +264,8 @@ class HierarchyModel(QtCore.QAbstractItemModel):
         if not project_doc:
             return
 
+        self.blockSignals(True)
+
         # Create project item
         project_item = ProjectItem(project_doc)
         self.add_item(project_item)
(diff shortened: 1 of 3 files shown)


64qam (approved):

iLLiCiTiT:
> The only side effect after this PR is that now it doesn't show the "Working..." overlay anymore and thus it feels like it's hanging more than before.

Project manager never had that overlay?

BigRoy:
This PR has introduced a bug where the **Tools** column doesn't show the correct data visually until when double clicked and editable.

### After

What I'm seeing on refresh (note the empty line compared to the next screenshot):
afbeelding

When double clicking the tools for a specific asset
afbeelding

### Before

afbeelding

---

Note also how the table view doesn't show all the editable columns. I have a hunch that maybe delegates aren't being registered correctly?

PR #3218, opened 2022-05-20
Project Manager: Fix persistent editors on project change

iLLiCiTiT:
## Brief description
Open persistent editors on project change/refresh in Project manager. Issue caused by [PR](https://github.com/pypeclub/OpenPype/pull/3216).

## Description
Insert signals are ignored on initial items fill on project refresh which cause that persistent editors are not opened. Opening of persistent editors is "faked" after project refresh. This change make it a little bit slower.

## Testing notes:
1. Open project manager
2. Select a project
3. All editors should be visible on all items and their values should match
4. The same should work on project change or on save

### DIFF
--- a/openpype/tools/project_manager/project_manager/model.py
+++ b/openpype/tools/project_manager/project_manager/model.py
@@ -264,8 +264,6 @@ class HierarchyModel(QtCore.QAbstractItemModel):
         if not project_doc:
             return
 
-        self.blockSignals(True)
-
         # Create project item
         project_item = ProjectItem(project_doc)
         self.add_item(project_item)
@@ -378,8 +376,6 @@ class HierarchyModel(QtCore.QAbstractItemModel):
                 task_items.append(task_item)
 
             self.add_items(task_items, asset_item)
-
-        self.blockSignals(False)
 
         # Emit that project was successfully changed
         self.project_changed.emit()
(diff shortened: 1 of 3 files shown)


BigRoy:
It seems Persistent Editors are so slow for relatively bigger (50 assets+shots with each 3 tasks; that's not even huge) project that I'd urge to look for another solution. It's literally a difference between instant feedback and waiting 20 seconds.

If it's solely for being able to edit multiple entries at the same time you can actually still edit multiple values of selected rows by double right clicking on one of the cells. If we play a bit with the click events of the View I have the feeling we can actually get the same "single click to edit" or "left double click to edit" as it was before whilst preserving the selection.

If it's solely for the user to be more aware that the fields are editable I think we might get away with stylesheets or Column delegates that can tweak the visuals for those columns when not being edited currently.

iLLiCiTiT:
I agree it's slowing it down but compared to removing block signals it's 2x faster.

BigRoy:
@iLLiCiTiT can you confirm on your end that clicking the Refresh Button top left doesn't actually Refresh the currently open project's view? Is that a new bug or was that always the case?

iLLiCiTiT:
> can you confirm on your end that clicking the Refresh Button top left doesn't actually Refresh the currently open project's view? Is that a new bug or was that always the case?

That was fixed with https://github.com/pypeclub/OpenPype/pull/3216/files#diff-bb26071d6f912a16a2e840d34e68d7c493900b4b2ebb99ab9789c27f4293a5d2R194

64qam (approved):

PR #3219, opened 2022-05-20
Hiero: debugging frame range and other

jakubjezek001:
## Brief description
Frame range is correctly defined now.

## Description
Framerange was not correctly collected in otio exporter. Hiero  is for not known reason having one frame longer duration at timeline and it was influencing calculation of handles and other aspects of timing. Loading of published clips were one (to two) frames longer and startHandle was also not correctly defined. Removed callback on timeline clip change - because it was slowing down work with bigger timelines. 

## Additional info
- also refactory of get_track_items because validation was not working properly
- small code improvements because we want nicer code
- added new function to api `get_timeline_selection`, for better timeline selected clips return from any context

## Known issue:
- when video clip is published it is creating longer but image sequences correct framerange

## Testing notes:
1. open your testing hrox file with some timeline clips
2. select couple clips which are tagged as publishable instances and publish
3. after publish go to loader and make sure the subset is having correct framerange and length of handles
4. load it to timeline as original timing and to actual timeline. Make sure framerange and handles are set correctly and there is no frame offset when you use diff tool on pixels with original clilp.
5. repeat 3 - 4 on retimed clips and Timewarped too.4
6. all should be correct.

# Linked PR
https://github.com/pypeclub/OpenPype/pull/3222 is 3.10 release alternative


64qam:
When creating an instance it crashes with the error:
```Traceback (most recent call last):
  File "C:\Users\qam\pypeclub\OpenPype\openpype\tools\creator\window.py", line 435, in _on_create
    data={"variant": variant}
  File "C:\Users\qam\pypeclub\OpenPype\openpype\pipeline\create\legacy_create.py", line 147, in legacy_create
    plugin = Creator(name, asset, options, data)
  File "C:\Users\qam\pypeclub\OpenPype\openpype\hosts\hiero\api\plugin.py", line 617, in __init__
    selection=timeline_selection
  File "C:\Users\qam\pypeclub\OpenPype\openpype\hosts\hiero\api\lib.py", line 177, in get_track_items
    with contextlib.suppress(AttributeError):
AttributeError: 'module' object has no attribute 'suppress' ```
```
The same when publishing already created instances: 
```
Message: 
'module' object has no attribute 'suppress'
Line: 
177
Traceback: 
  File "C:\Users\qam\pypeclub\OpenPype\.venv\lib\site-packages\pyblish\plugin.py", line 522, in __explicit_process
    runner(*args)
  File "C:\Users\qam\pypeclub\OpenPype\openpype\hosts\hiero\plugins\publish\precollect_instances.py", line 26, in process
  File "C:\Users\qam\pypeclub\OpenPype\openpype\hosts\hiero\api\lib.py", line 177, in get_track_items
    with contextlib.suppress(AttributeError):
AttributeError: 'module' object has no attribute 'suppress'```

64qam (approved):

PR #3225, opened 2022-05-20
Deadline: publishing of animation and pointcache on a farm

kalisp:
## Brief description
Some publishes (or extractors) might be too long to run in on a artist machine, so it should be offloaded to a farm.

## Description
Farm runs a host plugin which triggers new python script. This script does pyblish publish process end-to-end.

Animation and pointcache instances have now new 'Farm' field which if set to True, instance will be sent to DL, collected, validated, extracted and integrated there.

## Additional info
THIS PR IS OPEN FOR DISCUSSION

Current implementation is per instance, eg. only some instances could be published on the farm from a single publish. This matches more future move to New Publisher.

For each remote publishable instance, separate full publish job is triggered, which might not be optimal. There should be a discussion of separating "extract" part (separate jobs) and final publish job (in similar fashion as currently renders are implemented.)

`MayaSubmitRemotePublishDeadline` is currently created, most of logic should be yanked out into Abstract class to allow similar implementation in a different host (eg. Houdini).

## Testing notes:
0. Deploy `openpype/modules/deadline/repository/custom/plugins/GlobalJobPreLoad.py` to `DL_repo/custom/plugins`
1. Create instance of Pointcache family
2. Set its Farm field to True
3. Publish
4. Check DL and eventually published folder


tokejepsen (approved):
Code looks good to me, but can't unfortunately test myself.

openpype/modules/deadline/plugins/publish/submit_maya_remote_publish_deadline.py
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        # Include critical environment variables with submission + api.Session
+        keys = [
+            "FTRACK_API_USER",
+            "FTRACK_API_KEY",
+            "FTRACK_SERVER"
+        ]
+        environment = dict({key: os.environ[key] for key in keys
+                            if key in os.environ}, **legacy_io.Session)
```
BigRoy, line 106: Doesn't this already insert the full `legacy_io.Session` into `environment`?

openpype/modules/deadline/plugins/publish/submit_maya_remote_publish_deadline.py
```diff
@@ -0,0 +1,134 @@
...
+
+        # TODO replace legacy_io with context.data ?
+        environment["AVALON_PROJECT"] = legacy_io.Session["AVALON_PROJECT"]
+        environment["AVALON_ASSET"] = legacy_io.Session["AVALON_ASSET"]
+        environment["AVALON_TASK"] = legacy_io.Session["AVALON_TASK"]
+        environment["AVALON_APP_NAME"] = os.environ.get("AVALON_APP_NAME")
+        environment["OPENPYPE_LOG_NO_COLORS"] = "1"
```
BigRoy, line 113: Is this up to the submission to supply? Wouldn't Deadline global job or other areas affecting how Deadline launches things set this value when needed?

openpype/hosts/maya/plugins/publish/extract_animation.py
```diff
@@ -23,6 +23,14 @@ class ExtractAnimation(openpype.api.Extractor):
...
 
     def process(self, instance):
+        if instance.data.get("farm"):
+            path = os.path.join(
+                os.path.dirname(instance.context.data["currentFile"]),
+                "cache",
+                instance.data["name"] + ".abc"
+            )
+            instance.data["expectedFiles"] = [os.path.normpath(path)]
+            return
```
BigRoy, line 33: Some questions:
1. What's this needed for? Why do we need to set `expectedFiles`?
2. The path here wouldn't be correct, right? The farm would still extract to a temporary staging dir, no? Or did I miss the code that influences staging dir on the farm?

openpype/plugins/publish/integrate_new.py
```diff
@@ -139,6 +139,10 @@ def process(self, instance):
                     ef, instance.data["family"], instance.data["families"]))
                 return
 
+        # instance should be published on a farm
+        if instance.data["farm"]:
```
antirotor, line 143: shouldn't this be checked more carefully? What if instance doesn't have `farm` (as it happened to me right now)

antirotor (changes requested):
having integrate error (see comment) below

openpype/modules/deadline/plugins/publish/submit_maya_remote_publish_deadline.py
kalisp: I think DL uses these variables to call `extractenvironments` command which then provides environment for DL to set into rendering/publishing script environment.

openpype/hosts/maya/plugins/publish/extract_animation.py
kalisp: Yeah, you are right, removed.

openpype/hosts/maya/api/pipeline.py
```diff
@@ -71,8 +71,15 @@ def install():
     if lib.IS_HEADLESS:
         log.info(("Running in headless mode, skipping Maya "
                  "save/open/new callback installation.."))
+
+        # Register default "local" target
+        print("Registering pyblish target: remote")
```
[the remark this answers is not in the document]
kalisp: Created explicit HEADLESS_PUBLISH env var to differentiate it more precisely.

openpype/modules/deadline/plugins/publish/submit_maya_remote_publish_deadline.py
kalisp: I don't think anyone knows why is this in all DL submit jobs, so it might be cleaned up a bit, but I would keep it to another PR, as it would mean to retest all publishes to DL for all the implemented hosts.

antirotor:
Getting this error on Deadline job:
```
ScriptFilename={OPENPYPE_REPOS_ROOT}/openpype/scripts/remote_publish.py
```

Seems  `OPENPYPE_REPOS_ROOT` isn't resolved (even when it is set in env)

antirotor (approved):
Tested it, published, loaded... all works. Thanks!

PR #3230, opened 2022-05-23
Project manager: Speed up but without persistent editors

iLLiCiTiT:
## Brief description
Do not create persistent editors but create then on demand.

## Description
Creation of persistent editors is making the UI initialization slow but at the same time it's easy to see what can be modified. This change cause that editors are not visible so it's unclear what can be modified but it's really fast on initialization. All other functionality should stay the same. Also the columns are not resized to editor's size hint because are not known at the moment.

### Screenshots
image
image

## Testing notes:
1. Run project manager
2. Select project
3. Loading of project should be fast
4. Change any value on items


iLLiCiTiT:
Other possible solution, in theory: Persistent editors would be "dynamic" -> Create persistent editor on "first show" of row. So the slowdown would not be propagated at once but in smaller chunks and items that were not showed yet won't affect performance of UI.

mkolar (approved):
Good enough I'd say, considering this tool will probably be deprecated in v4 and this solves the immediate problem of the speed

PR #3236, opened 2022-05-23
Maya: Fix support for multiple resolutions

antirotor:
## Bug

Resolution was taken from default, so overridden values were ignored. This is changing how width, height and pixel aspect are resolved.

### How to test

Create two render layers, override on one resolution. Render / review should respect that.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/collect_render.py
+++ b/openpype/hosts/maya/plugins/publish/collect_render.py
@@ -339,9 +339,15 @@ class CollectMayaRender(pyblish.api.ContextPlugin):
                 "source": filepath,
                 "expectedFiles": full_exp_files,
                 "publishRenderMetadataFolder": common_publish_meta_path,
-                "resolutionWidth": cmds.getAttr("defaultResolution.width"),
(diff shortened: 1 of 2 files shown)


antirotor:
One question for @iLLiCiTiT: Even when it renders and publish in multiple resolutions now, the resolution on ftrack is still ... probably taken fron an asset? Because it doesn't reflect real resolution of renders.

iLLiCiTiT:
> Even when it renders and publish in multiple resolutions now, the resolution on ftrack is still ... probably taken fron an asset?

I can look tomorrow, but I'm 99% sure that we use resolution from settings (in ExtractReview) and if are set to `0` then we use resolution from input.

m-u-r-p-h-y (approved):
Published a Maya 2022 scene with 2 render layers with resolution overrides

sequences rendered correctly with proper resolution.
image

review media was generated at the same resolution though, but it is for another discussion on how this should be tackled.

- should the review media follow the source resolution?
- should the review media follow the  default project/asset resolution (as it is now)
- should we have overrides for review?
- etc.

tokejepsen:
> should the review media follow the source resolution?

+1. Think this is the most intuitive option.

PR #3247, opened 2022-05-25
Nuke: Change default icon path in settings

ClementHector:
## Brief description
I added an icon in resource for nuke studio and modified the default paths for nukex and nukestudio.

## Additional info
It's the only icon nuke studio that looks like the others nuke, if you find better I can replace it :)


m-u-r-p-h-y (changes requested):
would you mind taking a closer look at all the Nuke icons once again? Each of them is different levels of black, different resolution and different graphical style

image

it is not looking good

image


thank you!

m-u-r-p-h-y:
I was looking at the nuke installation folder, and found all the icons, but only in 48px resolution  :-(

plugins\icons\NukeStudioApp48.png
plugins\icons\NukeXApp48.png

Nuke is the only one with higher-res ones
plugins\icons\NukeApp256.png
plugins\icons\NukeApp128.png
plugins\icons\NukeApp64.png
plugins\icons\NukeApp48.png


would it be enough?

ClementHector:
I added the icons you provided :)

m-u-r-p-h-y (approved):
looks good and consistent now :)

image

thanks Clément

PR #3252, opened 2022-05-26
Unreal: Added plugin support for both UE4 and UE5

simonebarbieri:
UE4 and UE5 use different modules, so we need two different plugin integrations.


antirotor:
I've already merged the same thing there, but just with slightly different logic - the name of parent folders containing  integration are based on minimum compatible version so we are safe if something like another integration plugin will be needed fo UE_5.8 for example (and I hope it will not come to this :))

PR #3255, opened 2022-05-26
Unreal: Fix Camera Loading if Layout is missing

simonebarbieri:
## Brief description
Fixes a problem when loading a camera asset when the layout is missing.

## Description
When loading a camera, the loader checks for the master level. If no layout has been loaded, the master level is missing. Now when loading the camera, if there's no master level, it is created similarly to the master level sequence.


antirotor (approved):
works! thanks.

PR #3264, opened 2022-05-30
Maya: reference loaders could store placeholder in referenced url

kalisp:
## Brief description
All reference loaders in Maya now could instead of real published folder use environment variable placholder. This would allow to use this workfile in different locations (remote artist computer, studio etc.)

## Description
`OPENPYPE_PROJECT_ROOT_WORK` is calculated for values in Anatomy or overridden values from Local Setting.
This placeholder is physically stored in workfile, each value is resolved when workfile is opened. 
This allows remote artists to reference assets in a workfile and publish it. Same workfile will then work everywhere (studio, different artist) if they have Anatomy or Local Setting configured to existing folder on their machine.

This limits opening workfiles without Openpype, this could be remedied by adding used environment variable to artist environment. (By adding to theirs mel script??)

## Additional info
Flag to trigger this behavior is hardcoded to True for testing. Placement of this toggle in Settings **need to be decided**.
Easiest option is to add it as a plugin variable, this would mean add/modify this behavior on each plugin separately. Usage of
environment variable could be also widened to not only Maya. It could be used in any host which stores full path of referenced files in its workfile metadata. In that case, more general location to enable/disable this behavior should be chosen.

## Testing notes:
1. Open Maya
2. Use Reference Loader to pull in any asset
3. Save workfile
4. Check content of workfile, ${} placeholder should be used instead of real physical root.

### DIFF
--- a/openpype/hosts/maya/plugins/load/load_reference.py
+++ b/openpype/hosts/maya/plugins/load/load_reference.py
@@ -8,7 +8,6 @@ from openpype.pipeline import (
     legacy_create,
 )
 import openpype.hosts.maya.api.plugin
-from openpype.api import Anatomy
 from openpype.hosts.maya.api.lib import maintained_selection
 
 
@@ -52,8 +51,8 @@ class ReferenceLoader(openpype.hosts.maya.api.plugin.ReferenceLoader):
 
         with maintained_selection():
             cmds.loadPlugin("AbcImport.mll", quiet=True)
-            anatomy = Anatomy(context["project"]["code"])
-            file_url = anatomy.replace_root_with_env_key(self.fname, '${{{}}}')
+            file_url = self.prepare_root_value(self.fname,
+                                               context["project"]["code"])
             nodes = cmds.file(file_url,
                               namespace=namespace,
                               sharedReferenceFile=False,
--- a/openpype/hosts/maya/plugins/load/load_reference.py
+++ b/openpype/hosts/maya/plugins/load/load_reference.py
@@ -8,6 +8,7 @@ from openpype.pipeline import (
     legacy_create,
 )
 import openpype.hosts.maya.api.plugin
+from openpype.api import Anatomy
 from openpype.hosts.maya.api.lib import maintained_selection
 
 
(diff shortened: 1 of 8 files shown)


openpype/hosts/maya/plugins/load/load_reference.py
```diff
@@ -51,7 +51,9 @@ def process_reference(self, context, name, namespace, options):
 
         with maintained_selection():
             cmds.loadPlugin("AbcImport.mll", quiet=True)
-            nodes = cmds.file(self.fname,
+            file_url = self.prepare_root_value(self.fname,
+                                               context["project"]["code"])
+            nodes = cmds.file(file_url,
```
BigRoy, line 56: Somehow it looks to me like this particular change - if applicable to **all** references should actually happen in the `ReferenceLoader` class in `load` and potentially just do `self.fname = self.prepare_root_value(self.fname)`.

Just so the logic is encapsulated to that area only?

We should be able to make that change ONLY in the `ReferenceLoader` class and have that work for all implementations already, right?

kalisp: Imho no, each inheriting class of `ReferenceLoader` could calculate file url differently (or at least some existing loaders do). Plus we sometimes need it in update logic.

openpype/settings/entities/schemas/projects_schema/schema_project_maya.json
```diff
@@ -22,6 +22,12 @@
             "label": "Maya Directory Mapping",
             "is_group": true,
             "children": [
+                {
+                    "type": "boolean",
+                    "key": "use_env_var_as_root",
+                    "label": "Use env var placeholder in referenced url",
```
mkolar, line 28:
```suggestion
                    "label": "Use env var placeholder in referenced paths",
```

openpype/settings/entities/schemas/projects_schema/schema_project_maya.json
```diff
@@ -22,6 +22,12 @@
...
+                    "docstring": "Use ${} placeholder instead of physical value of root when storing into workfile metadata."
```
mkolar, line 29:
```suggestion
                    "docstring": "Use ${} placeholder instead of absolute value of a root in referenced filepaths."
```

m-u-r-p-h-y (approved):
works as expected.

image

will the path be resolved correctly when published to render farm?

kalisp:
It is working on a farm too.

PR #3267, opened 2022-05-31
Unreal: add support for skeletalMesh and staticMesh to loaders

antirotor:
## Enhancement

This small PR is just adding support for `skeletalMesh` and `staticMesh` to Unreal loaders.

### Testing

Just publish skeletal mesh and static mesh in Maya and load it in Unreal


m-u-r-p-h-y (approved):
Skeletal and Static Meshes are available now in Unreal


image

PR #3268, opened 2022-05-31
General: Updated windows oiio tool

iLLiCiTiT:
## Brief description
Changed url to new version (2.3.10) of OpenImageIO tools.

## Testing notes:
1. Fetch 3rd party libs
2. Run any publishing which is using oiio or try run the oiio manually


64qam:
Hash needs to be updated as well

PR #3269, opened 2022-06-01
Maya: better handling of legacy review subsets names

kalisp:
## Brief description
Multiple reviews from a Maya workfile are blocked by legacy subset names without variant. These names could be used in later process so we cannot replace them.

## Description
Legacy subset name (`{task}Review}`) is queried for presence in DB for published asset. If at least one subset found, subset name is kept (which limits possibility of multiple reviews), in other case proper from Creator (which could be configured by 
Validator to check uniqueness of subset names was added to catch multiple reviews with legacy name.

## Testing notes:
Test A
0. Before pulling this PR create camera and Review instance in Maya and publish
1. Create at least 2 cameras in Maya
2. create for them 'Review' instances
3. Publish
4. It should fail on Validation

Test B
0. Use different asset then in Test A
1. Create at least 2 cameras in Maya
2. create for them 'Review' instances
3. 3. Publish
4. It **shouldnt** fail on Validation

### DIFF
--- a/openpype/hosts/maya/plugins/publish/collect_review.py
+++ b/openpype/hosts/maya/plugins/publish/collect_review.py
@@ -77,15 +77,14 @@ class CollectReview(pyblish.api.InstancePlugin):
                 instance.data['remove'] = True
                 self.log.debug('isntance data {}'.format(instance.data))
         else:
-            if self.legacy:
-                instance.data['subset'] = task + 'Review'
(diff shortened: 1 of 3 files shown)


m-u-r-p-h-y (approved):
multiple reviews work as expected

image


now we should have a look at the broken arbitrary attributes of the review instance

image

PR #3274, opened 2022-06-02
Maya look: skip empty file attributes

antirotor:
## Fix

This is fixing look collector so it will skip empty filename attributes. Useful mainly on nodes that defines file attribute but can accept also input from other nodes.

### Testing

Create look with shader network, connecting **PxrTexture** to **PxrBump** for example in renderman (leaving filename attribute in **PxrBump** empty. Publishing and loading of the look should be successfull

### DIFF
--- a/openpype/hosts/maya/plugins/publish/collect_look.py
+++ b/openpype/hosts/maya/plugins/publish/collect_look.py
@@ -601,8 +601,10 @@ class CollectLook(pyblish.api.InstancePlugin):
                                                      source,
                                                      computed_source))
 
-            if not source:
-                self.log.info("source is empty, skipping...")
+            # renderman allows nodes to have filename attribute empty while
+            # you can have another incoming connection from different node.
+            if not source and cmds.nodeType(node).lower().startswith("pxr"):
+                self.log.info("Renderman: source is empty, skipping...")
                 continue
             # We replace backslashes with forward slashes because V-Ray
             # can't handle the UDIM files with the backslashes in the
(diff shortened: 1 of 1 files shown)


openpype/hosts/maya/plugins/publish/collect_look.py
```diff
@@ -601,6 +601,9 @@ def collect_resources(self, node):
                                                      source,
                                                      computed_source))
 
+            if not source:
+                self.log.info("source is empty, skipping...")
+                continue
```
BigRoy, line 606: Would it be worth adding a comment in code to state why we'd sometimes want to skip collecting on a file node.

Since technically with this change it also now allows a user to publish a look using a `file` node that doesn't have a texture set without that being validated against, yes?

tokejepsen: Doesnt Arnold throw an error if this happens? If so we should defo validate.

antirotor: I seems this is allowed only for Renderman as it seems. I can add logic just for that but I hate to have so much renderer specific codes....

mkolar: I'm not quite sure if it should be our responsibility to check if a file node is empty or not. I would just allow it across the board and leave it to the artist. It's way too wide of a problem to potentially deal with it for all renderers and their usecases.

antirotor: I've limited it to renderman nodes, I believe this is the safest and easiest approch to it

m-u-r-p-h-y (approved):
image

publish is passing through with an empty file path in render man node

PR #3280, opened 2022-06-03
SiteSync: speedup of starting process

kalisp:
## Brief description
Enabled Site Sync increased loading time of Tray and manipulations in Loader quite significantly on slow (Mongo on Cloud) connection.

## Description
Most expensive operation is to get all Site Sync enabled projects as its pulling project settings for all projects. This was moved into thread, so it is not blocking start. (For specific customer with cloud Mongo stationed far away `get_project_settings` takes about 3 seconds :/)
Currently this check is being done on every loop (to start syncing as soon as project is enabled), if this would be problematic, resolving this only once Tray session would need to be implemented.

## Testing notes:
1. Everything should work as before with Site Sync, just slightly faster on cloud MongoDB


mkolar (approved):
hard to tell how much difference it makes, but I can attest that at the very least everything seems to work for me.

64qam (changes requested):
It works on our cloud mongo. However, when using client cloud DB it is almost impossible to work with. Everything is extremely laggy.

64qam (approved):

PR #3282, opened 2022-06-03
Webpublisher: return only active projects in ProjectsEndpoint

kalisp:
## Brief description
Previously even archived projects were shown in Webpublisher front end

3.9.x version of https://github.com/pypeclub/OpenPype/pull/3281
## Testing notes:
1. Check FE if only active projects are visible


kalisp:
Wrong target branch

PR #3288, opened 2022-06-06
General: Define query functions

iLLiCiTiT:
## Brief description
Define basic query functions to replace direct mongo queries for project specific entities.

## Description
We need to define functions that will cover all direct mongo queries in code and use only functions in a way that functions logic can be changed to work with any other database system, or at least as much as possible. This is first step of the change by defying basic query functions which are used (for now) only in tools. `AvalonMongoDB` objects and `legacy_io` have new functions `active_project` and `current_project` (difference is that `current_project` does not require installation).

## Additional information
At this stage all tools should work as they did. Content of `openpype.client` must be self contained so there is no place to use 
any other context then context passed to functions which are called there.

Some functions can be split into more, some functions can be merged into one, some are missing completely because there was not a use case for them in tools.

## Future goals
1. Conver remaining parts of codebase to use these functions.
2. Move MongoDB connection handling from `openpype.lib` to `openpype.client` (with backwards compatibility).
3. Find out how to handle "creation" and "update" with some abstraction which is not MongoDB based.

## Testing notes:
1. Tools should work as they did


mkolar:
I'll just add that this is part of the continuous work towards a properly defined V4 API on the client side as well as server. To be able to make the move, we first need to unify what's going on with the DB call across the codebase.

mkolar (changes requested):
When I select asset in tray loader 
image

Launcher doesn't show anything inside a project


Selecting asset and task in standalone publisher 
image


Can't test further, because without launcher, i can't launch hosts :D

iLLiCiTiT:
Should work now

mkolar (changes requested):
- [x] representations and dependencies widgets in loader are empty.
- [x] Right click in the inventory doesn't work 

Digging up some more.

iLLiCiTiT:
> representations and dependencies widgets in loader are empty.

I think that representations were not visible even before this PR, can you confirm?

mkolar:
In develop I can see the representations just fine. 
image

antirotor (approved):
As far as I've tested it works for me.

kalisp (approved):
Tested in AE/PS.
Tested Loader and Site Sync.
So far it works fine.

PR #3293, opened 2022-06-07
Remove obsolete maya ass exporter

### LINKED ISSUE #1244: Export maya ass job
## Description
- script `~/pype/export_maya_ass_job.py` does not work in Pype 3

mkolar:
Considering this script hasn't worked for over a year and no-one noticed, I think it's safe to assume it's obsolete and can be removed. Resolve #1244


antirotor:
Then it should be removed also from the parts of the code leading to it. Probably only here:

https://github.com/pypeclub/OpenPype/blob/develop/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py

and here:

https://github.com/pypeclub/OpenPype/blob/develop/openpype/modules/deadline/plugins/publish/submit_publish_job.py

mkolar:
I guess a good question would be, should we rather fix it? It seem that these parts are never really used, so it feels like wasted effort.

PR #3297, opened 2022-06-08
Maya: point cache publish handles Maya instances

icyvapor:
This is to fix pointCache extractor, so Maya instances can be correctly published.

Our environment team reported issue where, multiple *Maya instances* sharing the same underlying mesh could not be published correctly as pointCache. Changes here fixed the pointCache extractor so it behaves correctly.
 
## Additional info
Currently the pointCache published from a selection set of multiple *Maya instances*, loaded back in, would only have one copy of the shared geo mesh, parented to one of instance transform node, then the rest of the instance transform nodes don't  have child mesh. Maya's native abc export tool can cache out Maya instances correctly. 

After some testing, seems the `selection` flag would correct this behavior, without any effects on regular abc export. 

PLEASE SQUASH BEFORE MERGING. Thanks!

## Testing notes:
1. In Maya scene, create instances using "Duplicate - special"
2. Create a 'PointCache'  OP instance including those instances
3. Publish it
4. OP Load the newly published PointCache back in


BigRoy:
Thank you very much. I can reproduce the issue on my end and I do suspect this change does "workaround" the problem that we have in OpenPype for the Alembic export.

I tested quickly with just Maya's `AbcExport` command to export instances with the `-selection` flag set and that does work, so it's not an issue with the Alembic export command itself. This lead me to suspect the issue is actually in the `nodes` being selected. I believe the root of the issue actually originates [here](https://github.com/pypeclub/OpenPype/blob/8d0bb493e86f252064b7b7535862c10d0c8addd2/openpype/hosts/maya/plugins/publish/collect_instances.py#L89-L95).

Whenever you use `maya.,cmds.listRelatives` with the `allDescendents` flags it will skip the instanced paths and returns only the first one as the [Maya documentation states](https://download.autodesk.com/global/docs/maya2012/en_us/CommandsPython/listRelatives.html#flagallDescendents):

> Returns all the children, grand-children etc. of this dag node. **If a descendent is instanced, it will appear only once on the list returned.** Note that it lists grand-children before children.

So to really solve the root of this issue we'd probably need to get all the children [like this](https://github.com/Colorbleed/colorbleed-config/blob/76f537eea041f8fed135f2b3334d70e1cf22b528/colorbleed/plugins/maya/publish/collect_instances.py#L27-L64) and it was part of a [fix we did on our previous pipeline](https://github.com/Colorbleed/colorbleed-config/blob/acre/colorbleed/plugins/maya/publish/collect_instances.py#L140-L142) codebase to indeed solve working with Instances all around the Maya codebase - hence why initially seeing this pop up had me wondering since I knew it worked for us in the past. Without fixing it there the issue could also be present in e.g. exporting to other formats.

I'd personally prefer reverting this change and resolving it in that area. @mkolar thoughts?

mkolar:
>I'd personally prefer reverting this change and resolving it in that area. @mkolar thoughts?

I agree with @BigRoy. We should probably just gut out the solution from colorbleed config.

icyvapor:
@BigRoy  thanks for looking into this! 
Yes, I wasn't exactly proud of the one-liner hack, totally agree we should fix the root cause (pun not-intended ..)

That said, I find it a bit strange for the formulated abcexport command to rely on the `selection` flag. As I understand, OP (avalon framework) works by publishing what's in the selection set, not requiring any selection.  The native `abcexport` has the `-root` flag for specifying the top nodes of what to be cached, so I feel it's a bit roundabout to still requiring them being selected. 

I'll move Roy's logic of children discovery over, and let this extractor use that with '-root'.

BigRoy:
The reason originally to use the selection flag was two reasons:
- Match code of other plugins like Maya ASCII exports, etc. Most Maya exporters tended to act like an "Export Selection". It didn't match for the end user really.
- We originally also used the export functionality in a way that it also explicitly excluded some children nodes in the hierarchy. As far as I am aware that's only possible with the selection flag because there is no argument (I think) to exclude specific nodes. I am not sure whether anything in OpenPype relies on being able to exclude children. Since it allows that feature which otherwise would not be possible I don't see an instant reason to change it around (unless it's breaking something).

BigRoy (approved):
Thanks! This works for me in Maya 2022 and fixes the issue.

mkolar:
thank you guys. merging

BigRoy:
It's good to note that this code influences the behavior for `collect_instances` on how it collects nodes that are IN objectSets directly in the instance objectSet.

E.g.
```
yetiRigMain                         (yetiRig publish instance; objectSet)
    - input_SET                     (objectSet)
        - input_member_GEO          (transform)
```

Previously `input_member_GEO` would have been collected as a child node because `cmds.listRelatives(members, allDescendents=True)` would actually also include the "children" of the `input_SET`. I'd argue that the old behavior is actually unexpected and wrong.

This thus could influence other areas of the code unexpectedly - for example this does influence data that yetiRig validators try to validate, like the [**Validate Yeti Rig Input Shapes In Instance**](https://github.com/pypeclub/OpenPype/blob/2074e65d2d033c0874dae919870a478c0c4493c9/openpype/hosts/maya/plugins/publish/validate_yeti_rig_input_in_instance.py). The old collector would allow the validator to pass (I believe) by accident.

PR #3299, opened 2022-06-08
Documentation: Add app key to template documentation

ClementHector:
## Brief description
I added the app key which was missing in the documentation on the list of existing keys in the templates.


mkolar (approved):
Thank you Clement

PR #3300, opened 2022-06-08
Maya: Fix swaped width and height in reviews

iLLiCiTiT:
## Brief description
Reviews from Maya had swaped width and height collected from scene.

## Testing notes:
1. Publish from maya with some review in scene


m-u-r-p-h-y (approved):
fixed . . .

Issue #2724, opened 2022-02-14
Implement Package family


antirotor:
## Feature

We need to create family **Package** that would hold just information about various subset grouping them together. This would be useful then for bundling things together for exports, filtering in UI, and so on. It should be just simple list of subset versions that could be created directly from tools without publishing, but versioned.

antirotor:
This shouldn't just depend on the artist. Produciton could put together things from different assets, for exports, and so on. Put together stuff like geo, looks, clothes, grooms, ...

tokejepsen:
Kinda similar to the projection workflow we have experimented with; https://github.com/pypeclub/OpenPype/pull/1110

BigRoy:
> Produciton could put together things from different assets, for exports, and so on. Put together stuff like geo, looks, clothes, grooms, ...

But who or where defines what goes together? That's mostly what I'm asking. Just to figure out the approach to implementation here.

antirotor:
> But who or where defines what goes together? That's mostly what I'm asking. Just to figure out the approach to implementation here.

we thought about actually using Loader (or something very similar to it). Maybe even action of ftrack. So you wouldn't need to go through publishing process

mkolar:
I've been doing a some brainstorming on this front a multiple question have arisen. To avoid ambiguity I will use the term `package` in completely generic way, but would like to define two distinct approaches that need to be supported that I will name `assembly` and `bundle` for not, just to have some differentiation. Names are not important at this point. 

- usecase one is where we have multiple subsets relate to each other in a very simplistic way without any special relationships. Example would be 
	- a set of renderlayers (not know how they should be connected)
	-  characters that always tend to appear together in the shot
	- a collection of trees that we grouped together to make it easier to load them into a scene when building a forest, etc.
All of these just need information that they belong together and I will call them `bundles` for now. 

- usecase two is multiple subsets that not only belong together, but should be connected (assembled) in a particular way. 
	- A look and groom applied to an alembic cache. Can give you a render ready assembly of a character
	- A proxy look applied to a rig can give you animation ready assembly of a character
	- Cutlery and plates placed on a table and published together gives you an assembly that you can use for set dressing.
All of these need extra information about how things are connected, what their transforms might be and so on. I will call these `assemblies` for now. 

My feeling is that these can and should be two separate families. 
[...]

PR #3304, opened 2022-06-08
Maya: Allow more data to be published along camera 🎥 

antirotor:
## Enhancement

This PR is releasing restrictions in camera extractors to allow more data to be published along with the camera. There is still validator that validates camera content and by default allows only camera to be passed down the line, but it can be set to allow more data. Usefull for proxy geometry / locators / pointcloud to be published with the camera.

### How to test it

in the Settings, there are new options for **Validate Camera Content** - `project_settings/maya/publish/ValidateCameraContents`. It can be disabled, it can be set as optional and also you can disable validating presence of shapes - in another words, disabling this option will allow more then cameras to pass this validator.

image

* Create camera instance and put for example box and camera under it
* Publish it with `Validate presence of shapes` **ON**
* Validator should trigger

---

* Disable `Validate presence of shapes` and try again
* Validator should pass and publish should be successful
* Loading camera as abc and ma should pull in camera and whatever geometry attached.

---

❔ not sure about the `Validate presence of shapes` as it is technically incorrect

### DIFF
--- a/openpype/hosts/maya/plugins/publish/extract_camera_alembic.py
+++ b/openpype/hosts/maya/plugins/publish/extract_camera_alembic.py
@@ -34,7 +34,6 @@ class ExtractCameraAlembic(openpype.api.Extractor):
                           dag=True, type="camera")
 
         # validate required settings
-        assert len(cameras) == 1, "Not a single camera found in extraction"
         assert isinstance(step, float), "Step must be a float value"
         camera = cameras[0]
 
@@ -44,8 +43,12 @@ class ExtractCameraAlembic(openpype.api.Extractor):
         path = os.path.join(dir_path, filename)
 
         # Perform alembic extraction
+        member_shapes = cmds.ls(
+            members, leaf=True, shapes=True, long=True, dag=True)
         with lib.maintained_selection():
-            cmds.select(camera, replace=True, noExpand=True)
+            cmds.select(
+                member_shapes,
+                replace=True, noExpand=True)
 
             # Enforce forward slashes for AbcExport because we're
             # embedding it into a job string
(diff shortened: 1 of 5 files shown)


BigRoy:
1. Also - if we're allowing other arbitrary data in there doesn't this just make it a `pointcache` publish? What's the reason to not choose a `pointcache` family whenever needing additional geometry data with your cameras?
2. Even when allowing other shapes would we still want to validate that only a single camera is present?

mkolar:
>If we're allowing other arbitrary data in there doesn't this just make it a pointcache publish? What's the reason to not choose a pointcache family whenever needing additional geometry data with your cameras?

The main reason is correct filtering and categorization in the loader. Having everything as pointcaches would become very confusing fairly quickly.

>Even when allowing other shapes would we still want to validate that only a single camera is present?

I think it does still make sense. This is about adding tracking or proxy geometry to the single camera. But it should still be recognizably a single camera.

BigRoy:
> I think it does still make sense. This is about adding tracking or proxy geometry to the single camera. But it should still be recognizably a single camera.

I think this might also just be a +1 for implementing subset packages like #2724 which would actually solve this too?

antirotor:
> > I think it does still make sense. This is about adding tracking or proxy geometry to the single camera. But it should still be recognizably a single camera.
> 
> I think this might also just be a +1 for implementing subset packages like #2724 which would actually solve this too?

I believe it solves it only partially. I can imagine integration with tracking software that is producing some geometry bound to specific camera - and it puts it to one single file (fbx or whatever). Without a full integration, you would drop such file into standalone publisher, select `Camera` and publish it.

It all depends how you feel about data being tightly bound to camera - proxy geo can live on its own, but some partial track, or pointcloud can be total nonsense without specific camera. So yes, in some cases Packages/Bundles would solve it, but I still believe that for particular workflows the camera and additional data are unseparable.

antirotor:
Single camera is still validated by validator, I think extractors shouldn't care.

mkolar (approved):
I've tested all combination I could think of and all is working as expected

PR #3305, opened 2022-06-08
General: Vendorized modules for Python 2 and update poetry lock

iLLiCiTiT:
## Brief description
Added some python modules to Python 2 vendor, fix incompatible dependencies and updated poetry lock.

## Description
Current dependencies are incompatible which cause that creation of env and build crashes. One reason is that gazu `0.8` has set `requests` version explicitly and is not compatible with our requirements. That has changed in `0.8.28` and updated which added new dependencies. Requests and some of it's dependencies now have to be vendorized for Python 2 hosts and poetry lock was updated.

## Additional info
Python module `charset_normalizer` have to be faked for python 2 requests to raise `ImportError` (it would raise `SyntaxError` which is not handled). Didn't test all possible clashes.

## Testing notes:
1. Run create env tool to update python dependencies
2. Test both Python 2 and Python 3 hosts
    - in a way that they use modified dependencies (e.g. deadline, ftrack, gazu, ...)


kalisp (approved):
Build is fixed, tested in PS/AE/Nuke, publishing works (even to Ftrack).

mkolar (approved):

PR #3306, opened 2022-06-08
Nuke: fixing farm publishing if review is disabled

jakubjezek001:
## Brief description
Nuke is now publishing on farm without review

## Description

## Dependency
this PR also includes https://github.com/pypeclub/OpenPype/pull/3303

### DIFF
--- a/openpype/hosts/maya/api/lib.py
+++ b/openpype/hosts/maya/api/lib.py
@@ -1737,8 +1737,11 @@ def apply_shaders(relationships, shadernodes, nodes):
             log.warning("No nodes found for shading engine "
                         "'{0}'".format(id_shading_engines[0]))
             continue
+        try:
+            cmds.sets(filtered_nodes, forceElement=id_shading_engines[0])
+        except RuntimeError as rte:
+            log.error("Error during shader assignment: {}".format(rte))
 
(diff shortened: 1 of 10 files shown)


64qam (changes requested):
It doesn't create an Ftrack review (e.g. mp4), but it still makes an empty link on Ftrack. 
image

iLLiCiTiT:
There is missing `"representations"` key on instances if they go to farm which cause crashes. Would be possible to add the key in https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/nuke/plugins/publish/precollect_instances.py#L139 for each instance?

iLLiCiTiT:
Also can we remove adding "instances" key to context here:
https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/nuke/plugins/publish/precollect_instances.py#L161
https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/nuke/plugins/publish/precollect_workfile.py#L104

It's not used anywhere...

64qam (changes requested):
When `review` is disabled, it still creates a new empty Ftrack version, but with exr component in it. 
image

64qam (approved):

PR #3310, opened 2022-06-09
Ftrack: Action to easily create daily review session

iLLiCiTiT:
## Brief description
Added server action to create daily review session with ability to auto-create dailies each day.

## Description
Action creates review session with name based on template defined in settings. If there is review session with same name then rest of logic is skipped. If there is not review session with same name and then is review session from "previous day" which does not contain any reviewables then that is just renamed otherwise new review session object is created. Action does daily cycles during which checks if project has enabled "daily creation" of review sessions and create them automatically for all enabled so the action does not have to be executed manually.

## Additional info
If template does not contain day formatting key it probably won't be "daily" so it can be used for e.g. weeklies too. Default template is `{yy}{mm}{dd}` (2 digits year, 2 digits month and 2 digits day).

## Testing notes:
1. Run ftrack event server
2. Run the action on any project
3. Enable cycle creation on a project and wait until next day

### DIFF
--- a/openpype/modules/ftrack/event_handlers_server/action_create_review_session.py
+++ b/openpype/modules/ftrack/event_handlers_server/action_create_review_session.py
@@ -37,7 +37,9 @@ class CreateDailyReviewSessionServerAction(ServerAction):
     default_template = "{yy}{mm}{dd}"
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


jakubjezek001:
here is what I can see in Project actions  
image

iLLiCiTiT:
It's server action, not local.

jakubjezek001 (approved):
works as expected

PR #2202, opened 2021-11-03
Maya: Build workfile by template

ClementHector:
Following discussion #2013 , here is a first working draft for maya, before going any further, your opinion would be welcome!

## How does it work:

1. Build your template:
    - Open Maya
    - Add place holder with `Create Place Holder` button
image
   - Fill in the extra attribute on Place Holder
image
    - Arrange your maya scene the way you want
    - Save your scene
2. Create profile:
    - Go to Project Settings
    - Add a profile
image    - 
3. Build your first workfile:
    - Open maya
    - Build your workfile
image

## Future development and idea

- [x]  Check operation with builder type linked asset 
- [x] Use asset, hierarchy for filtering import
- [x] Keep the information of the places holders in maya in order to be able to check if everything is up to date, useful if the list of linked assets has changed.
- [x] Extract class PlaceHolder from TemplateLoader
- [x] Better error feedback
- [x] Write documentation

PR #3315, opened 2022-06-10
Maya: Build workfile by template

ClementHector:
## Brief description
Following discussion #2013 and resumption of work from ellipsanime #2202

## Templated Build Workfile

Building a workfile using a template designed by users. Helping to assert homogeneous subsets hierarchy and imports. Template stored as file easy to define, change and customize for production needs.

 **1. Make a template**

Make your template. Add families and everything needed for your tasks. Here is an example template for the modeling task using a placeholder to import a gauge.

maya-workfile-outliner

If needed, you can add placeholders when the template needs to load some assets. **OpenPype > Template Builder > Create Placeholder**

maya-create_placeholder

- **Configure placeholders**

Fill in the necessary fields (the optional fields are regex filters)

maya-placeholder_new

    - Builder type: Wether the the placeholder should load current asset representations or linked assets representations

    - Representation: Representation that will be loaded (ex: ma, abc, png, etc...)

    - Family: Family of the representation to load (main, look, image, etc ...)

    - Loader: Placeholder loader name that will be used to load corresponding representations

    - Order: Priority for current placeholder loader (priority is lowest first, highet last)

- **Save your template**

 **2. Configure Template**

- **Go to Studio settings > Project > Your DCC > Templated Build Settings**
- Add a profile for your task and enter path to your template

template_build_workfile

**3. Build your workfile**

- Open maya

- Build your workfile

maya-build_workfile_from_template


tokejepsen:
Exciting stuff!

Should we have a task name regex for the profile, so we can be flexible with tasks of the same type but different names?

ClementHector:
> Should we have a task name regex for the profile, so we can be flexible with tasks of the same type but different names?

Do we have to put a taks list like for the orignal build ?

tokejepsen:
> > Should we have a task name regex for the profile, so we can be flexible with tasks of the same type but different names?
> 
> Do we have to put a taks list like for the orignal build ?

Have only skimmed this PR, so ignore my thought if it requires more work.

ClementHector:
Don't worry, i will do it!

openpype/lib/build_template.py
```diff
@@ -0,0 +1,61 @@
+from openpype.pipeline import registered_host
```
iLLiCiTiT, line 1: Can we move this to `openpype.pipeline`? We're in process of reducing pipeline related functions from `openpype.lib` to avoid cross imports.

openpype/lib/avalon_context.py
```diff
@@ -922,6 +923,118 @@ def save_workfile_data_to_doc(workfile_doc, data, dbcon=None):
...
+            }
+
+        output[asset_id]["subsets"][subset_id]["version"]["repres"].append(
+            repre
+        )
+
+    return output
+
+
+@with_pipeline_io
```
iLLiCiTiT, line 1023: This function does not need the decorator.
```suggestion
```

openpype/lib/build_template.py
iLLiCiTiT: This looks like a package of workfile template build. So maybe a folder with this file, `abstract_template_loader.py` and build_template_exceptions.py`?

ClementHector: Should I put the file `build_template.py`, `abstract_template_loader.py` and `build_template_exceptions.py`
in `openpype/lib/build_workfile` or `openpype/build_workfile`?

ClementHector: @iLLiCiTiT Can you tell me which folder please.

m-u-r-p-h-y (changes requested):
This is a big feature and a massive code change. I tested just the basic functionality

- creating templates
- modifying project settings
- the initial building of a scene
- -  
this is working good so far- 



during further testing, I encountered a couple of issues

when  running Update Workfile from a template (after the initial build of the scene with no asset loaded)

`# Error: ValueError: file D:\REPO\OpenPype\openpype\hosts\maya\api\template_loader.py line 83: No object matches name: AVALON_CONTAINERS # `



another issue is with Update Placeholder.
when you run it, it will open a window. If you close it without doing any change, it will remove custom attributes from the placeholder

image

openpype/lib/build_template.py
iLLiCiTiT: I'm sorry for late response, just `./openpype/pipeline`. Keep updated changes in `avalon_context.py` most of those functions were changed or removed over time.

m-u-r-p-h-y (approved):
the two small things were fixed and it seems quite stable now.
we should let it go to production for a real-world test :)


thanks!

mkolar (approved):

iLLiCiTiT (dismissed):
Please move "context specific" functionality from `openpype.lib` to `openpype.pipeline`. Function `collect_last_version_repres` changed code over time so please update the code from latest develop. And `get_loaders_by_name` should be moved into `openpype.pipeline.load.utils.py`.

PR #3316, opened 2022-06-10
updated poetry installation source

maxpareschi:
## Brief description
updated poetry installation source

## Description
So we can get rid of the create_env message at the start

## Additional info
both .ps1 and .sh updated 

## Testing notes:
.\tools\create_env.ps1


antirotor (approved):
I was planning to do it for very long time but couldn't get to it. Thank you!

PR #3318, opened 2022-06-10
Maya: add pointcache family to gpu cache loader

### LINKED ISSUE #3313: Maya: pointcache loader doesn't support "load as GPU cache"
## Bug

Loading **pointcache** family should have similar options a loading alembics from **model** family. Currently it is missing `load as GPU cache` option.

[cuID:2f3jtkb]

antirotor:
## Enhancement

This simple PR is adding `pointcache` family to GPU loader so alembic published as pointcaches can be loaded with it.

### Testing

Try to load pointcache alembic as GPU cache in Maya

Close #3313

### DIFF
--- a/openpype/hosts/maya/plugins/load/load_gpucache.py
+++ b/openpype/hosts/maya/plugins/load/load_gpucache.py
@@ -10,7 +10,7 @@ from openpype.api import get_project_settings
 class GpuCacheLoader(load.LoaderPlugin):
     """Load Alembic as gpuCache"""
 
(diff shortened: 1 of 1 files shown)


BigRoy:
@antirotor While you're at it... also add the `animation` family like [here](https://github.com/BigRoy/OpenPype/blob/colorbleed/openpype/hosts/maya/plugins/load/load_gpucache.py#L13). Should've made a PR when I did that, sorry!

mkolar (approved):

Issue #3271, opened 2022-06-01
Settings serializable to file


antirotor:
## Feature request

It would be very useful to be able to serialize parts of Settings to file, to store it as backup or reference or to be able to reapply it on other projects (for example review profiles).

It should work exactly the same as copy & paste feature works now, but using files.

PR #3323, opened 2022-06-10
Settings: Settings can be extracted from UI

iLLiCiTiT:
## Brief description
Added option to extract current values into json file.

## Description
Extraction takes current values and extract them to a json file. It takes all values no matter if are overriden or not. Values are stored under keys with double underscores to be able differentiate what is value and what are metadata. Also added option to paste values from clipboard that are copied from different source or when whole file is copied.

## Additional information
Didn't add option to "open json file" because it would be hard to find out what user want's to do with it. This is not meant as backup of Project settings as it does not save only overrides but all values.

### Extract values to file
1. Open settings UI
2. Right click on any item in settings and hit `Extract to file`
3. Select output file and hit Save
4. Check output

### Using extracted file
1. Open settings UI
2. Copy file in file explorer
3. Right click in settings where would you like to paste values
4. `Paste` action should be visible and if the saved file contains path there should be also `Paste to same place`
5. Try them

### Paste value as text
1. Open settings UI
2. Copy json content as text
```
# For example 'system_settings/modules/avalon'
{
    "AVALON_TIMEOUT": 666,
    "AVALON_THUMBNAIL_ROOT": {
        "windows": "D:/evil/directory",
        "darwin": "",
        "linux": ""
    }
}
```
3. Try to paste it into settings UI

Resolves https://github.com/pypeclub/OpenPype/issues/3271

### DIFF
--- a/openpype/tools/settings/settings/base.py
+++ b/openpype/tools/settings/settings/base.py
@@ -12,27 +12,70 @@ from openpype.tools.settings import CHILD_OFFSET
 
 from .widgets import ExpandingWidget
 from .lib import create_deffered_value_change_timer
-from .constants import DEFAULT_PROJECT_LABEL
+from .constants import (
+    DEFAULT_PROJECT_LABEL,
+    SETTINGS_PATH_KEY,
+    ROOT_KEY,
+    VALUE_KEY,
+    SAVE_TIME_KEY,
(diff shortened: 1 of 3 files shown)


antirotor:
Related to #3271

PR #3325, opened 2022-06-10
Project Manager: Directly make task editable when clicking on the add Task button

BigRoy:
(cherry picked from commit 96ca15b565b952d4e17c2010e8846ca18272d681)

## Brief description

This makes the "Task" button create a directly editable task for name/type like when doing right click -> Add Task.
Otherwise it makes a task for which it's unclear what to do next.

## Description

Before:
afbeelding

After:
afbeelding

## Additional info

~I know this is not the way to fix it because it calls a private method. However, I'm opening this to start the discussion and approach a better solution. Maybe make that method not private?~ Made public with https://github.com/pypeclub/OpenPype/pull/3325/commits/cbc6fb8b330abc88c0e8c64c57a966cd09fd8077

## Testing notes:
1. Open Project Manager
2. Select an asset
3. Click on the "Task" button in the top left
4. The task should now directly be clearly editable and have a default type.

### DIFF
--- a/openpype/tools/project_manager/project_manager/window.py
+++ b/openpype/tools/project_manager/project_manager/window.py
@@ -245,7 +245,6 @@ class ProjectManagerWindow(QtWidgets.QWidget):
         self.hierarchy_view.add_asset()
 
     def _on_add_task(self):
-        # Colorbleed edit: force the task to directly be in edit mode
         self.hierarchy_view.add_task_and_edit()
 
     def _on_create_folders(self):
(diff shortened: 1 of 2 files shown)


openpype/tools/project_manager/project_manager/window.py
```diff
@@ -245,7 +245,8 @@ def _on_add_asset(self):
         self.hierarchy_view.add_asset()
 
     def _on_add_task(self):
-        self.hierarchy_view.add_task()
+        # Colorbleed edit: force the task to directly be in edit mode
+        self.hierarchy_view._add_task_and_edit()
```
iLLiCiTiT, line 249: Please change `_add_task_and_edit` to `add_task_and_edit`. Otherwise approved.

BigRoy: Should be resolved.

iLLiCiTiT (approved):

PR #3326, opened 2022-06-10
Maya: Speed up validate mesh shader connections for large scenes

BigRoy:
(cherry picked from commit 46752511e500a54ba9335a325ca09224ef600227)

## Brief description

This optimizes the Validate Mesh Shader connection for large scenes - that is scenes with many objects that e.g. all have the lambert1 assigned in a model scene.

## Description

Instead of querying the members of e.g. the lambert1 shading engine per object it caches the members during the process to speed up the query. This can make it go down from many seconds on a large scene to near instant.

## Testing notes:
1. Publish some models from large and small scenes - confirm it passes fine as before and is faster.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/validate_mesh_shader_connections.py
+++ b/openpype/hosts/maya/plugins/publish/validate_mesh_shader_connections.py
@@ -12,28 +12,41 @@ def pairs(iterable):
         yield i, y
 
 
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
@antirotor please check this PR?

antirotor (approved):

PR #3328, opened 2022-06-11
Nuke: Load full model hierarchy by default

gabormarinov:
## Brief description
Fix Nuke model loader to select all geo nodes in the hierarchy for loading by default

## Description
The current implementation only selects top level geo nodes form the hierarchy, this fixes it by always selecting all geo nodes

## Additional info
Nuke pops up the Hierarchy window during Alembic loading to show which geo nodes are selected for loading. To avoid this popup, you have to set the following in preferences:

```python
preferencesNode = nuke.toNode('preferences')
preferencesNode.knob('DfltAbcAlwaysCreateAllInOne').setValue(True)
```

We put this into Nuke's `init.py` in our pipeline, we didn't want to hardcode it in here.  Without it loading/version switching especially on multiple geo nodes is really annoying.  Maybe there's a more streamlined solution to avoid this window?

### DIFF
--- a/openpype/hosts/nuke/plugins/load/load_model.py
+++ b/openpype/hosts/nuke/plugins/load/load_model.py
@@ -60,6 +60,10 @@ class AlembicModelLoader(load.LoaderPlugin):
                 inpanel=False
             )
             model_node.forceValidate()
+
+            # workaround to load all geo nodes, not just top level ones
+            model_node.knob('scene_view').setAllItems(model_node.knob('scene_view').getAllItems(), True)
+
             model_node["frame_rate"].setValue(float(fps))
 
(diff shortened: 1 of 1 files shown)


tokejepsen:
Can you use the scene API?

```
# Ensure all items are imported and selected.
        scene_view = node['scene_view']
        scene_view.setImportedItems(scene_view.getAllItems())
        scene_view.setSelectedItems(scene_view.getAllItems())
```

gabormarinov:
It works flawlessly, thanks!

mkolar (approved):

mkolar:
Thank you @gabormarinov

PR #3333, opened 2022-06-13
Webpublisher: Use client query functions

iLLiCiTiT:
## Brief description
Webpublish routes are using query functions from [PR](https://github.com/pypeclub/OpenPype/pull/3288) instead of direct use of mongo queries.

## Description
Webpublish routes are using query functions from `openpype.client`. Webpublish routes do not inherit from private class of `_RestApiEndpoint` from avalon module instead has it's own implementation which was split into 2 because of new function calls. Did some logical renaming (e.g. `OpenPypeRestApiResource` -> `WebpublishRestApiResource`).

## Testing notes:
1. Run blender
2. Frame start/end should be set to asset's resolution
3. Publish layout

PR waits for merge of [PR](https://github.com/pypeclub/OpenPype/pull/3288)


kalisp (approved):
Webpublisher still works with this one

PR #3335, opened 2022-06-14
Enhancement/ftrack restrict folder creation for shot and asset build only

maxpareschi:
## Brief description
Option to skip folders without tasks when creating publish and work folders ahead of time
Option to create the local folder structure directly from Ftrack (only in the work root, where it makes sense)

## Description
Current behaviour was creating publish and work folders in every ftrack folder indiscriminately, and also replicated the whole structure in all the root paths when triggered from the standalone project manager.
Made the first step optional using a checkbox on Creation dialog, and also added the option (still through checkbox) to create the complete project directory structure taken from the global project settings directly from Ftrack.

## Additional info
There was a need to override the logic in lib/path_tools.py (to restrict the creation to the work root only).
I exposed a new named arg that has the default value reflecting the old behaviour for compatibility (and while the old behaviour seems like a bug to me it can actually be correct in other places so..)  

no docs

## Testing notes:
1. go to ftrack
2. get to the project
3. actions/create folders
4. set the checkbox as needed (we check all of three to get the desidered output)
5. enjoy the folders if you need them!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,23 +1,39 @@
 # Changelog
 
-## [3.11.0-nightly.1](https://github.com/pypeclub/OpenPype/tree/HEAD)
+## [3.11.0-nightly.2](https://github.com/pypeclub/OpenPype/tree/HEAD)
 
 [Full Changelog](https://github.com/pypeclub/OpenPype/compare/3.10.0...HEAD)
 
 ### 📖 Documentation
 
+- Documentation: Add app key to template documentation [\#3299](https://github.com/pypeclub/OpenPype/pull/3299)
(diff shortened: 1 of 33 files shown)


iLLiCiTiT:
Please make PRs clean, without code from another PRs if they are independent. This branch is created from branch in [this PR](https://github.com/pypeclub/OpenPype/pull/3320) but is not related to it.

maxpareschi:
> Please make PRs clean, without code from another PRs if they are independent. This branch is created from branch in [this PR](https://github.com/pypeclub/OpenPype/pull/3320) but is not related to it.

Sorry, my bad. I've tried to rebase but cannot really see it working, maybe it's best if we just close this and I'll apply the changes to a new branch?

PR #3338, opened 2022-06-14
Enhancement/ftrack-restrict-folder-creation-for-shot-and-asset-build-only-(rebranched)

maxpareschi:
## Brief description
Option to skip folders without tasks when creating publish and work folders ahead of time
Option to create the local folder structure directly from Ftrack (only in the work root, where it makes sense)

## Description
Current behaviour was creating publish and work folders in every ftrack folder indiscriminately, and also replicated the whole structure in all the root paths when triggered from the standalone project manager.
Made the first step optional using a checkbox on Creation dialog, and also added the option (still through checkbox) to create the complete project directory structure taken from the global project settings directly from Ftrack.

## Additional info
REBRANCHED FROM DEVELOP TO FIX BRANCH INDEPENDENCE ISSUE!
There was a need to override the logic in lib/path_tools.py (to restrict the creation to the work root only).
I exposed a new named arg that has the default value reflecting the old behaviour for compatibility (and while the old behaviour seems like a bug to me it can actually be correct in other places so..)

## Documentation (add "type: documentation" label)
no docs

Testing notes:
1. go to ftrack
2. get to the project
3. actions/create folders
4. set the checkbox as needed (we check all of three to get the desidered output)
5. enjoy the folders if you need them!

### DIFF
--- a/openpype/modules/ftrack/event_handlers_user/action_create_folders.py
+++ b/openpype/modules/ftrack/event_handlers_user/action_create_folders.py
@@ -75,10 +75,10 @@ class CreateFolders(BaseAction):
             {
                 "type": "label",
                 "value": "Create <b>Local Work</b> folders from OpenPype \
-                    template"
+                    Project Folders Template"
             },
             {
-                "name": "from_local_template",
+                "name": "from_local",
                 "type": "boolean",
                 "value": False
             },
@@ -103,11 +103,11 @@ class CreateFolders(BaseAction):
         with_interface = event["data"]["values"]["with_interface"]
         with_childrens = True
         restrict_folders = False
-        from_local_template = False
+        from_local = False
         if with_interface:
             with_childrens = event["data"]["values"]["children_included"]
             restrict_folders = event["data"]["values"]["restrict_folders"]
-            from_local_template = event["data"]["values"]["from_local_template"]
+            from_local = event["data"]["values"]["from_local"]
 
         filtered_entities = []
         for entity in entities:
@@ -230,15 +230,18 @@ class CreateFolders(BaseAction):
 
         self.log.info("Creating local template folders...")
 
-        if from_local_template:
+        if from_local:
             try:
                 # Get paths based on presets
                 basic_paths = get_project_basic_paths(project_name)
                 if not basic_paths:
                     pass
                 # Invoking OpenPype API to create the project folders
-                create_project_folders(basic_paths, project_name,
-                                    root_paths=[work_root])
+                create_project_folders(
+                    basic_paths,
+                    project_name,
+                    root_paths=[work_root]
+                )
                 self.log.info("successfully created local template folders.")
             except Exception as exc:
                 self.log.warning(
--- a/openpype/modules/ftrack/event_handlers_user/action_create_folders.py
+++ b/openpype/modules/ftrack/event_handlers_user/action_create_folders.py
@@ -1,7 +1,11 @@
 import os
 import collections
 import copy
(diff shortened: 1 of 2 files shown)


openpype/modules/ftrack/event_handlers_user/action_create_folders.py
```diff
@@ -51,13 +55,33 @@ def interface(self, session, entities, event):
...
                 "type": "label",
-                "value": "With all chilren entities"
+                "value": "Create with all children entities"
             },
             {
                 "name": "children_included",
                 "type": "boolean",
                 "value": False
             },
+            {
```
iLLiCiTiT, line 65: This is misleading label. Something like "Create folders only for entities with tasks" would be more suitable.

openpype/modules/ftrack/event_handlers_user/action_create_folders.py
```diff
@@ -196,6 +228,27 @@ def launch(self, session, entities, event):
             if not os.path.exists(path):
                 os.makedirs(path)
 
+        self.log.info("Creating local template folders...")
+
+        if from_local:
```
iLLiCiTiT, line 233: Not sure if this belongs here. The logic is already in different action (`CreateProjectFolders`).

openpype/modules/ftrack/event_handlers_user/action_create_folders.py
```diff
@@ -196,6 +228,27 @@ def launch(self, session, entities, event):
...
+                basic_paths = get_project_basic_paths(project_name)
+                if not basic_paths:
+                    pass
+                # Invoking OpenPype API to create the project folders
+                create_project_folders(
+                    basic_paths,
+                    project_name,
+                    root_paths=[work_root]
+                )
+                self.log.info("successfully created local template folders.")
```
iLLiCiTiT, line 245: Moved log and fix condition.
```suggestion
        if from_local:
            self.log.info("Creating local template folders...")
            try:
                # Get paths based on presets
                basic_paths = get_project_basic_paths(project_name)
                if basic_paths:
                    # Create the project folders
                    create_project_folders(
                        basic_paths,
                        project_name,
                        root_paths=[work_root]
                    )
                    self.log.info("successfully created local template folders.")
```

openpype/modules/ftrack/event_handlers_user/action_create_folders.py
mkolar: I can't quite tell exactly what this does either actually

maxpareschi: Well the children included is somethintg that was already there, i don't know why you would need the check, maybe it was some feature request of a studio?

maxpareschi: Aside of the fact that if you uncheck the children included the script actually does nothing.

mkolar: The question is about `Create publish and work folders`

`Children included` should generate folder for all children as well as selected entity. so If you generate for `sequence`. With the boolean on, you will get all the shots as well, with it Off, you will only get the sequence.

mkolar: If the intention is that it will skip creating publish and work folders for entities that are other than `shots` and `asset builds` than that is way too specific to your workflow. We have users that have entities named `poster`, `book`, `library` and so on and they want these folders generated there. We can't make such assumption

PR #3349, opened 2022-06-15
Docs: Clarify about staging version naming.

icyvapor:
Make it a bit clearer that the staging version needs to be put into the version.py file.

### DIFF
--- a/website/docs/admin_distribute.md
+++ b/website/docs/admin_distribute.md
@@ -54,9 +54,10 @@ The default locations are:
 ### Staging vs. Production
 You can have version of OpenPype with experimental features you want to try somewhere but you
 don't want to disrupt your production. You can tag version as **staging** simply by appending `+staging`
-to its name.
+to its name in `<Project_Root>/openpype/version.py`.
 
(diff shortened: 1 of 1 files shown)


website/docs/admin_distribute.md
```diff
@@ -54,9 +54,10 @@ The default locations are:
 ### Staging vs. Production
 You can have version of OpenPype with experimental features you want to try somewhere but you
 don't want to disrupt your production. You can tag version as **staging** simply by appending `+staging`
-to its name.
+to its name in `<Project_Root>/openpype/version.py`.
 
-So if you have OpenPype version like `OpenPype-v3.0.0.zip` just name it `OpenPype-v3.0.0+staging.zip`.
+So if you have OpenPype version like `3.0.0` just name it `3.0.0+staging`, thus generating the zip file named
+`OpenPype-v3.0.0+staging.zip`.
```
mkolar, line 60: This should actually be amended a bit differently. You don't need this naming gymnastics anymore, because you can be explicit as to what is considered staging in the System settings. No need for changing of names at all. @antirotor what do you think?

PR #3350, opened 2022-06-16
Feature/multiverse

dmo-j-cube:
## Brief description
Sensible defaults & Project Settings UI for Multiverse USD. Exposed ExtractAlembic family settings in the UI

## Description
Adding more sensible defaults to Multiverse USD Asset Creator and to expose the Multiverse Creator Plugins to the Studio Projects Settings UI. Minor update to the documentation is also included.

For ExtractAlembic, the families which it responds to can be configured in the Project Settings UI - by default, nothing changes since all 3 existing families are set as defaults, but, for example, `model` can be removed so the Alembic publish can be skipped for a studio.


antirotor (approved):
Thanks, looks good.

PR #3353, opened 2022-06-16
Maya: Implementation of JSON layout for Unreal workflow

simonebarbieri:
## Brief description
Implementation for the extraction of JSON layout in Maya for Unreal workflow.

## Description
This implementation is similar to the current workflow in Blender. It uses the same JSON format.


antirotor (commented):
I've just did a quick test, but it seems it doesn't translate the loaded objects in the layout correctly. This is my very stupid maya layout:

image

and this happens in UE5:

image

(basically everything is loaded to 0,0,0)

antirotor (approved):
It works for me fine!

PR #3354, opened 2022-06-16
Ftrack: Removed requirement of pypeclub role from default settings

iLLiCiTiT:
## Brief description
Ftrack settings don't have `pypeclub` role for custom attributes creation in defaults.

## Descritpion
Create/Update of custom attributes does not require `pypeclub` role by default. Changed label of action to `Create/Update Custom Attributes`.

## Testing notes:
1. Remove `pypeclub` role from ftrack (if is there)
2. Run action to Create/Update Avalon attributes which should not fail if the role is not in ftrack


BigRoy:
Haven't tested - but code looks good to me.

64qam (approved):

PR #3356, opened 2022-06-16
deadline: fixing misidentification of revieables

jakubjezek001:
## Brief description
fixing wrong `review` attribute distribution

## Testing notes:
1. start your host supporting farm publishing
2. set your farm instance and set it to `review` **on** and publish
2.1. it should be reviewable on ftrack
3. set your farm instance and set it to `review` **off** and publish
3.1. it should not be reveiwable on ftrack


64qam (approved):

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -754,10 +755,6 @@ def process(self, instance):
                 "family": "prerender",
                 "families": []})
```
iLLiCiTiT, line 756: Review is removed here for prerender, is it ok?

PR #3357, opened 2022-06-16
Hosts: More options for in-host callbacks

iLLiCiTiT:
## Brief description
Modules are acknowledged on host installation that host was installed so they can register their specific callbacks. Workfiles tool also trigger events on workfile open. Moved start timer callback on task change to timers manager.

## Description
Modules have method `on_host_install` which is called when `install_host` is called, these modules are kept in memory (they're not refreshed for rest of process) that is because modules would be garbage collected and their callbacks would not work. Workfiles tool is triggering `workfile.open.before` and `workfile.open.after`. All events triggered from workfiles tool also contain context information (host name, project name, asset name and task name) so it's not just blind information. Task changed topic has human readable keys. Logic related to start timer on task change was moved to timers manager module.

## Additional information
Reason of this feature is to be able register event callbacks which are module specific and to be able catch workfiles opening when workfiles tool is used.

## Testing notes:
Change related to timers manager:
1. Make that you have some module with time management enabled (e.g. ftrack module)
2. Run application on a task (timer should start for the task)
3. Change context to different task using workfiles tool inside host
4. Check if timer for the new task was started

New feature adding ability to catch host installation:
1. Implement in any module method `on_host_install` where you can do some logic which you want.

### DIFF
--- a/openpype/pipeline/context_tools.py
+++ b/openpype/pipeline/context_tools.py
@@ -33,6 +33,9 @@ from . import (
 _is_installed = False
 _registered_root = {"_": ""}
 _registered_host = {"_": None}
+# Keep modules manager (and it's modules) in memory
+# - that gives option to register modules' callbacks
+_modules_manager = None
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


mkolar (approved):

openpype/pipeline/context_tools.py
```diff
@@ -70,10 +88,12 @@ def install_host(host):
             avalon host-interface.
     """
     global _is_installed
+    global _modules_manager
```
BigRoy, line 91: Line is not needed since `_get_modules_manager()` is called which already sets/reads the global variable.

iLLiCiTiT: yep, fixed

PR #3370, opened 2022-06-20
Maya: Handle excluding `model` family from frame range validator.

### LINKED ISSUE #2562: Maya: Validate Frame Range
In case of nested sets under reviewMain  Validate Frame Range function is picking up wrong set to query frameStart

```
plugin is: <class 'D:\REPO\OpenPype\openpype\hosts\maya\plugins\publish\validate_frame_range.py.ValidateFrameRange'>
Preparing <class 'openpype.action.RepairAction'>..
Action prepared.
// pyblish.pyblish.plugin.Action : Finding failed instances.. // 
    runner(*args)
  File "D:\REPO\OpenPype\openpype\action.py", line 65, in process
    plugin.repair(instance)
  File "D:\REPO\OpenPype\openpype\hosts\maya\plugins\publish\validate_frame_range.py", line 89, in repair
RuntimeError: setAttr: No object matches name: modelMain.frameStart

    runner(*args)
  File "D:\REPO\OpenPype\openpype\action.py", line 65, in process
    plugin.repair(instance)
  File "<string>", line 89, in repair
RuntimeError: setAttr: No object matches name: modelMain.frameStart
```

Allan-I:
## Brief description
This resolves #2562 by excluding the `model` and some other families from having their `frameStart` and `frameEnd` attributes validated (which they wouldn't have because of their being geometry).

## Description
The `validate_frame_range.py` currently validates the incorrect subset for published Review instances when there's a nested `model` subset in it.  This fix bypasses the validator checks when it detects the families are of `model` instance, and some others.

## Testing notes:
1. Create model and then create OP `model` instance.
2. Create camera and then create OP `review` instance.
3. Append model instance subset into review subset.
4. Publish.
5. If frame range is set incorrectly, it should give an error and the validator should be able to repair that error by conforming the review instance frame range to the one set in the mongo document.


BigRoy:
This doesn't seem like a 'clean fix' and feels rather arbitrary? How come the `model` family is problematic? Wouldn't this e.g. also happen with `mayaScene`, `layout`, `look`, `cameraRig`, `rig` families? Or is it solely due to [`CollectModel` which forces a single frame extraction](https://github.com/pypeclub/OpenPype/blob/8432e94615a75216e7172cf46f1370fff806bf13/openpype/hosts/maya/plugins/publish/collect_model.py#L23-L26)?

Sounds like something that points to another potential deeper issue?

mkolar:
There are a few problems in play here. Most would be solved by implementing the new publisher in  Maya, so we wouldn't have to hack the whole review system, however, that is just constantly slipping in priorities :(.

@bigroy is right and this is not just an issue with model, but anything that get's nested under a review set inside maya. Right now, it really poses two questions. 

1. Should model instance inherit the review framerange, or keep it's own? 
2. Does it make sense to validate frame ranges whatsoever for publishes that are not shot based?

For 1. I'd say that the review framerange should be propagated to the model (or other nested instance) so that it can be properly validated, if required. 

as for 2. we should probably just expose the families attribute to settings, so the studio can choose not to validate frameranges on a set of families for themselves.

BigRoy:
> For 1. I'd say that the review framerange should be propagated to the model (or other nested instance) so that it can be properly validated, if required.

This sounds unwanted since it opens up for motion/multiple time samples to make it's way into the model `.abc` in the Extract Alembic.

> Does it make sense to validate frame ranges whatsoever for publishes that are not shot based?

Good question. I personally think it's not necessarily only a Shot validation. Maybe at best the Project Manager could also have a state where start/end frame is NOT SET like a boolean stating "use frame range".

mkolar:
>Good question. I personally think it's not necessarily only a Shot validation. Maybe at best the Project Manager could also have a state where start/end frame is NOT SET like a boolean stating "use frame range".

True, however, I'm not sure it's worth doing it in v3 data model now.  In v4 there is differentiation between asset types and they can technically have empty values on attributes, which will make this a lot easier. 

>This sounds unwanted since it opens up for motion/multiple time samples to make it's way into the model .abc in the Extract Alembic.

Also a very good point. But in that case the only option we have now to make it at least functional is to skip validation on certain families which is exactly what this PR attempts. I just think those families should be configurable

openpype/hosts/maya/plugins/publish/validate_frame_range.py
```diff
@@ -27,6 +27,7 @@ class ValidateFrameRange(pyblish.api.InstancePlugin):
                 "yeticache"]
     optional = True
     actions = [openpype.api.RepairAction]
+    exclude_families = ["model"]
```
mkolar, line 30: If we add this to settings, we'll solve the production situation right now and can focus on doing it properly as Roy suggests in v4

BigRoy:
> Also a very good point. But in that case the only option we have now to make it at least functional is to skip validation on certain families which is exactly what this PR attempts. I just think those families should be configurable

Or could we reverse that logic? Only check `review` instance family frame range if NOT attached to another instance or if another of the families is set to validate with this validator  (e.g. `animation`)?

antirotor:
Hmm...  to add to it there seems to be probem with Repair Action for render instances (as it is running against non-existent instances in Maya - they are created on-the-fly by collector).

Maybe repair action should run only on those instances that are "physically" present in Maya and have frame range attributes.

mkolar:
>Or could we reverse that logic? Only check review instance family frame range if NOT attached to another instance or if another of the families is set to validate with this validator (e.g. animation)?

This is a bit limited to pyblish capabilities right now it's checking in `families`, whereas, here it would be beneficial to only go by `family`, I still think that simply excluding `model`, `rig` and maybe a few others that are not time based, would be the best option right now.

openpype/settings/defaults/project_settings/maya.json
```diff
@@ -204,7 +204,8 @@
         "ValidateFrameRange": {
             "enabled": true,
             "optional": true,
-            "active": true
+            "active": true,
+            "exclude_families": []
```
antirotor, line 208: I would move default families from `validate_frame_range.py#30` to here.
```suggestion
            "exclude_families": ["model", "rig", "staticMesh"]
```

m-u-r-p-h-y (changes requested):
The Frame Range Validator always passes no matter what numbers are entered in Review Instance

image

mkolar:
>The Frame Range Validator always passes no matter what numbers are entered in Review Instance

that's the whole point of the PR. to not validate on models ;)

m-u-r-p-h-y (approved):
Ohh. I was reading the original Description and was expecting something else.

In this case it works as expected :)

mkolar (approved):

PR #3371, opened 2022-06-20
Sync Queue: Added far future value for null values for dates

kalisp:
## Brief description
Null values were sorted as last, this keeps queued items together with last synched.

## Testing notes:
1. experiment with sorting by `Active site` (or `Remote site`)
2. queued items (without any value in column) should be sorted close to last updated

### DIFF
--- a/openpype/modules/sync_server/tray/models.py
+++ b/openpype/modules/sync_server/tray/models.py
@@ -1,6 +1,7 @@
 import os
 import attr
 from bson.objectid import ObjectId
+import datetime
 
 from Qt import QtCore
 from Qt.QtCore import Qt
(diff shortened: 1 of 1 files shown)


mkolar (approved):

openpype/modules/sync_server/tray/models.py
```diff
@@ -413,6 +414,23 @@ def get_index(self, id):
...
+        """Converts 'date_value' to string.
+
+        Value of date_value might contain date in the future, used for nicely
+        sort queued items next to last downloaded.
+        """
+        try:
+            converted_date = None
+            # ignore date in the future - for sorting only
+            if date_value and date_value < current_date:
+                converted_date = date_value.strftime("%Y%m%dT%H%M%SZ")
```
BigRoy, line 427: Could this be dangerous for different time zones or a machine that is say an hour behind with its hardware clock?

What sorting are we trying to solve, the one in Qt only? Or in the mongodb queries itself?
I feel like we could solve this simpler/better for either of those.

mkolar: I don't think this is a problem. Especially considering the re-write of the site-sync as a OPv4 addon anyways. Not worth the time now.

PR #3379, opened 2022-06-21
Resolve: Use client query functions

iLLiCiTiT:
## Brief description
Modified Resolve host to use query functions from [PR](https://github.com/pypeclub/OpenPype/pull/3288).

## Description
Resolve is using query functions from `openpype.client` to get versions. Added `"plate"` family into load sequence loader.

## Testing notes:
1. Loader plugin `Load clip` should be able to load and update


jakubjezek001 (approved):
testing was working for me

PR #3381, opened 2022-06-21
Maya: vray device aspect ratio fix

m-u-r-p-h-y:
## Description
Vray has a different attribute name for device aspect ratio from other renderers.

 (.aspectRatio instead of .deviceAspectRatio)

## Testing notes:

- open/create a maya scene with vray renderer set
- run OpenPype -> Set Resolution

resolution should be set without an error message about nonexistent vray attribute

### DIFF
--- a/openpype/hosts/maya/api/lib.py
+++ b/openpype/hosts/maya/api/lib.py
@@ -2126,9 +2126,11 @@ def set_scene_resolution(width, height, pixelAspect):
 
     control_node = "defaultResolution"
     current_renderer = cmds.getAttr("defaultRenderGlobals.currentRenderer")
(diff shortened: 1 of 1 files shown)


BigRoy:
Wouldn't this now break it for all non v-ray settings? :)

m-u-r-p-h-y:
I thought it was part of a code dealing with vray settings only, if not, I will ask @antirotor to make necessary changes. Thanks for checking @BigRoy

antirotor (approved):
It works! (I hope :))

m-u-r-p-h-y:
I've tested it and it works correctly for all renderers (vray, maya hw, prman, mtoa)

PR #3385, opened 2022-06-21
Maya: Use client query functions

iLLiCiTiT:
## Brief description
Modified Maya host to use query functions from [PR](https://github.com/pypeclub/OpenPype/pull/3288). Also small tweaks in code related to using the functions which in some cases can make the process faster.

## Description
Use query functions from `openpype.client`. Some functions were modified to be able match to functions abilities or were enhanced by reducing queries. Most of places used during publishing are using `assetEntity` instead of querying it again by name.

## Testing notes:
- [x] Generate uuids action should work (on validators)
- [x] Setdress workflow
- [x] Look assignment workflow
- [x] Load look
- [x] Load audio
- [ ] Load image plane

### DIFF
--- a/openpype/hosts/maya/api/action.py
+++ b/openpype/hosts/maya/api/action.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import
 
 import pyblish.api
 
+from openpype.client import get_asset_by_name
 from openpype.pipeline import legacy_io
 from openpype.api import get_errored_instances_from_context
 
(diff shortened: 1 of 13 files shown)


antirotor:
Only loading image plane doesn't work but that might be related to #3067

```
  File "C:\Users\annat\Documents\pype\3.0\openpype\tools\loader\widgets.py", line 1556, in _load_representations_by_loader
    options=options
  File "C:\Users\annat\Documents\pype\3.0\openpype\pipeline\load\utils.py", line 285, in load_with_repre_context
  File "C:\Users\annat\Documents\pype\3.0\openpype\hosts\maya\plugins\load\load_image_plane.py", line 184, in load
    image_plane_shape.listConnections(type="expression")[0]
IndexError: list index out of range
```

iLLiCiTiT:
> Only loading image plane doesn't work but that might be related to https://github.com/pypeclub/OpenPype/pull/3067

So it's not working even in develop? Can I do somthing about it?

antirotor (approved):
Appart of image plane loader that was broken before everything seems to work.

PR #3386, opened 2022-06-21
Maya: Camera extra data - additional fix for #3304

antirotor:
## Fix

This is adding forgotten commits to already merged PR #3304. It solves issues of parented geometry inside camera instance.

### Testing

Try to publish camera with  parented geometry. Validator should trigger.


m-u-r-p-h-y (approved):
validate on:
image

validate off:
image

openpype/hosts/maya/plugins/publish/extract_camera_alembic.py
```diff
@@ -61,10 +61,11 @@ def process(self, instance):
...
             if bake_to_worldspace:
                 job_str += ' -worldSpace'
-                for member in member_shapes:
-                    self.log.info(f"processing {member}")
+
+                for member in members:
                     transform = cmds.listRelatives(
-                        member, parent=True, fullPath=True)[0]
+                        member, parent=True, fullPath=True)
+                    transform = transform[0] if transform else member
```
BigRoy, line 68: Not entirely sure but this is something to be aware of. Solely looking at this code if a camera instance previously contained a parent group of the camera instead of the camera itself the old logic would still select the 'camera' for export.

I suspect this will now change the behavior for older camera instances that originally would not include the parent hierarchy.

```
group1/group2/camera
```

Would previously export as
```
camera
```

I suspect after this change the export is now:
```
group1/group2/camera
```

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
```diff
@@ -172,18 +172,24 @@ def process(self, instance):
...
-                            unlock(plug)
-                            cmds.setAttr(plug, value)
+                            shapes=True, long=True), attrs.items()):
+                        # the above call still pull in shapes that are not
+                        # cameras, so we filter them out here
+                        if cmds.nodeType(cam) != "camera":
+                            continue
+                        plug = "{0}.{1}".format(cam, attr)
+                        unlock(plug)
+                        cmds.setAttr(plug, value)
```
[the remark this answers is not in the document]
antirotor: that check there was because I've overlooked the `shape` flag and was wondering why the hell I still get other shapes than camera. Apart from that, the code is maybe harder to read but is getting rid of one loop and re-setting of `attrs` that is constant in each iteration.

openpype/hosts/maya/plugins/publish/extract_camera_alembic.py
```diff
@@ -61,10 +61,25 @@ def process(self, instance):
...
+                camera_root = cmds.listRelatives(
+                    camera, parent=True, fullPath=True)[0]
+                job_str += ' -root {0}'.format(camera_root)
+
+                for member in members:
+                    member_content = cmds.listRelatives(
+                        member, ad=True, fullPath=True) or []
+                    # skip hierarchy if it contains only camera
+                    # `member_content` will contain camera + its parents
+                    if camera in member_content \
```
BigRoy, line 76: Please do not do a lookup per member whether the camera is in member_content while it is a set. This could be very slow, right? Should this be a `set`?

```python
and len(member_content) == len(camera.split("|")) - 2```

What's that trying to solve? This looks like an odd comparison to me.

File: openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py [hunk 4]
```diff
@@ -172,18 +172,19 @@ def process(self, instance):
...
-                                 "backgroundColorB": 0.0,
-                                 "overscan": 1.0}
-                        for attr, value in attrs.items():
-                            plug = "{0}.{1}".format(cam, attr)
-                            unlock(plug)
-                            cmds.setAttr(plug, value)
+                            long=True), attrs.items()):
+                        plug = "{0}.{1}".format(cam, attr)
+                        unlock(plug)
+                        cmds.setAttr(plug, value)
```
BigRoy, line 187: I know it's functionally the same - but do we feel `itertools.product` is really that much more readable then the nested for loop? I'm fine with either but somehow I have a harder time reading this new code.

openpype/hosts/maya/plugins/publish/extract_camera_alembic.py
antirotor: What this is trying to achieve is to allow situations where you have camera inside some hierarchy and  at the same time handle cases where you have hierarchy with camera AND something else.

```
alice_GRP
    |---> bob_GRP
    |        \---> camera1_CAM
    \---> box_GEO
```

in that case you would expect alembic with:

```
camera1_CAM
alice_GRP
    \---> box_GEO
```

but that will fail because `alice_GRP` and `camera1_CAM` have parent relationship.

So what it does is that it gets all descendants of the member and check if the count of them is exactly the same as the path to the camera (minus 2 for the camera transform and shape).

Camera is `|alice_GRP|bob_GRP|camera1_CAM|camera1_CAMShape` and all descendants for member `|alice_GRP` are `["|alice_GRP", "|alice_GRP|bob_GRP"]`

if  count of `member_content` is higher, it means there are more things there than camera.  I think there should be warning logged before `continue`.

> This could be very slow, right?

actually Maya is pretty fast for that one. I've just ran some benchmarks and for 1000 calls I get something aroung  ~0.02 sec

BigRoy: Why wouldn't we just check whether this:
```python
descendents = cmds.listRelatives(member, allDescendents=True, fullPath=True) or []
shapes = cmds.ls(descendents, shapes=True, noIntermediate=True, long=True)
cameras = cmds.ls(shapes, type="camera", long=True)
if set(shapes) - set(cameras):
    # There are other shapes than just cameras present
    ...
```

I have a feeling we're overthinking or over-engineering this concept somehow. The fact that this starts behaving differently in a certain edge also feels less consistent than we'd like for a publish. No?

antirotor: damn you are right. Why is it the sets always eludes me even if the are the easiest solution.

mkolar (approved):

mkolar (approved):

PR #3395, opened 2022-06-22
Houdini: Use client query functions

iLLiCiTiT:
## Brief description
Modified Houdini host to use query functions from [PR](https://github.com/pypeclub/OpenPype/pull/3288).

## Description
Houdini is using query functions from `openpype.client` to get full hierarchy. Modified avalon uri handler in vendorized module to use Anatomy.

## Testing notes
- [x] Reset frame range and fps should work
- [x] Create HDA should work
- [x] USD publish should work (collect, validate, extract)
- [ ] SelectAssetDialog in USD should work

### DIFF
--- a/openpype/hosts/houdini/api/lib.py
+++ b/openpype/hosts/houdini/api/lib.py
@@ -4,6 +4,7 @@ from contextlib import contextmanager
 
 import six
 
+from openpype.client import get_asset_by_name
 from openpype.api import get_asset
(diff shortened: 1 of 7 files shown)


mkolar (approved):
`SelectAssetDialog` is not actually used anywhere, but the rest works as expected. 

@BigRoy do you know what it was for originally?

PR #3408, opened 2022-06-24
Houdini: fix loading and updating vbd/bgeo sequences

antirotor:
## Fix

This is fixing loading and updating vdb and bgeo sequences in Houdini. There was problem with sequence token replacement and some syntax errors.

### Testing

Try to load vbd and bgeo sequence, try to switch it to older version and back. Observe file name on the node - it should point to proper sequence and should have frame token in it (`$F4`)

### DIFF
--- a/openpype/hosts/houdini/plugins/load/load_bgeo.py
+++ b/openpype/hosts/houdini/plugins/load/load_bgeo.py
@@ -70,7 +70,6 @@ class BgeoLoader(load.LoaderPlugin):
         # The path is either a single file or sequence in a folder.
         if not is_sequence:
             filename = path
-            print("single")
         else:
             filename = re.sub(r"(.*)\.(\d+)\.(bgeo.*)", "\\1.$F4.\\3", path)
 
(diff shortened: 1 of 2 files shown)


openpype/hosts/houdini/plugins/load/load_vdb.py
```diff
@@ -31,6 +31,7 @@ def load(self, context, name=None, namespace=None, data=None):
 
         # Create a new geo node
         container = obj.createNode("geo", node_name=node_name)
+        is_sequence = bool(context["representation"]["context"].get("frame"))
```
BigRoy, line 34: Would this be a false negative if the first frame of the sequence is zero (`0`)?

openpype/hosts/houdini/plugins/load/load_bgeo.py
```diff
@@ -94,9 +93,10 @@ def update(self, container, representation):
 
         # Update the file path
         file_path = get_representation_path(representation)
-        file_path = self.format_path(file_path)
+        is_sequence = bool(representation["context"].get("frame"))
+        file_path = self.format_path(file_path, is_sequence)
```
BigRoy, line 97: How about changing `format_path()` arguments to `(file_path, representation)`. The `is_sequence = bool(representation["context"].get("frame")` statement would then just move into the method.

Thoughts?

openpype/hosts/houdini/plugins/load/load_vdb.py
antirotor: It won't as the `frame` is string. But it is place to revisit in v4  (along with so many others)

openpype/hosts/houdini/plugins/load/load_bgeo.py
antirotor: sure, it's much more cleaner.

m-u-r-p-h-y (approved):
Switching versions works as expected, paths are generated properly with correct frame token

image

PR #3411, opened 2022-06-24
Attribute Defs UI: Files widget show what is allowed to drop in

iLLiCiTiT:
## Brief description
Show label of what is allowed to drop into files widget. It should not matter if extension is with dot or not.

## Description
Show relevant label to user what he can drop into files widget. Show if user can add single or multiple folders, files or file sequences.

## Testing notes:
Easiest is possible using Tray publisher tool
1. Define creators with some extensions in simple cases
2. Extensions should be showed in files widget

It is also possible to add custom creator with any variant of files attribute definition.

### DIFF
--- a/openpype/widgets/attribute_defs/files_widget.py
+++ b/openpype/widgets/attribute_defs/files_widget.py
@@ -26,26 +26,122 @@ IS_SEQUENCE_ROLE = QtCore.Qt.UserRole + 7
 EXT_ROLE = QtCore.Qt.UserRole + 8
 
 
+class SupportLabel(QtWidgets.QLabel):
+    pass
+
+
 class DropEmpty(QtWidgets.QWidget):
-    _drop_enabled_text = "Drag & Drop\n(drop files here)"
+    _empty_extensions = "Any file"
 
-    def __init__(self, parent):
+    def __init__(self, single_item, allow_sequences, parent):
         super(DropEmpty, self).__init__(parent)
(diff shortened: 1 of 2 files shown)


mkolar (changes requested):
Can we please add a bit of gap here? 
image

openpype/widgets/attribute_defs/files_widget.py
```diff
@@ -47,6 +47,7 @@ def __init__(self, single_item, allow_sequences, parent):
         layout.addWidget(
             drop_label_widget, 0, alignment=QtCore.Qt.AlignCenter
         )
+        layout.addSpacing(10)
```
mkolar, line 50: `layout.addSpacing(30)`

openpype/widgets/attribute_defs/files_widget.py
```diff
@@ -47,6 +47,7 @@ def __init__(self, single_item, allow_sequences, parent):
         layout.addWidget(
```
mkolar, line 47: `layout.addSpacing(20)`

mkolar (changes requested):

openpype/widgets/attribute_defs/files_widget.py
mkolar: Sorry this is support to be at the line above.

mkolar (approved):

PR #3413, opened 2022-06-26
Nuke: option to bake viewer ip before slate

maxpareschi:
## Brief description
Introduce an optional workflow to bake viewer input process before the slate. 

## Description
This pr introduces an optional workflow that is intended to not change the colorspace of the slate by applying the viewer input process baking before the actual slate.
We use this since our slate layouts itself after the input resolution, and our Input process has also the ability to repo and reformat.
Also, it does not make sense to us to apply a lut on the slate, especially if CDL + LUT are very invasive.
The code also compensate for viewer process baking, where if Rec.709 is used (for instance) to bake the slate the source gts converted form Rec.709 to linear before the slate to have correct colors in the slate itself (in case of thumbnails)

## Additional info
Setting for this live under Project Settings > Nuke > Publish Plugins > ExtractSlateFrame
there's a new boolean switch called "Bake Viewer IP berfore Slate"
Also, we found the problem with nuke.nodeCopy and nodePaste, which is that the system clipboard when OP is driving nuke is sometimes not available. Fortunately the same functionality can be used by specifying a text file, which we save in the instance staging directory.

no docs

## Testing notes:
1. enable this under Project Settings > Nuke > Publish Plugins > ExtractSlateFrame > Bake Viewer IP berfore Slate
2. You should see the viewer input process node copied before the slate and the slate connected after it when publishing

### DIFF
--- a/openpype/hosts/nuke/plugins/publish/extract_slate_frame.py
+++ b/openpype/hosts/nuke/plugins/publish/extract_slate_frame.py
@@ -49,12 +49,12 @@ class ExtractSlateFrame(openpype.api.Extractor):
 
     def render_slate(self, instance):
         node_subset_name = instance.data.get("name", None)
-        node = instance[0] # group node
+        node = instance[0]  # group node
         # get the slate node
         slate_node = instance.data.get("slateNode")
         # get the last comp node before slate
         last_comp_node = slate_node.input(0)
- 
+
         self.log.info("Creating staging dir...")
 
         if "representations" not in instance.data:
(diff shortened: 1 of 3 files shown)


openpype/hosts/nuke/plugins/publish/extract_slate_frame.py
```diff
@@ -98,18 +108,65 @@ def render_slate(self, instance):
         if "#" in fhead:
             fhead = fhead.replace("#", "")[:-1]
 
-        previous_node = node
+        # Create a text file in staging dir to use as a
```
iLLiCiTiT, line 111: Few comments regarding this. There is no reason why to store clipboard path to instance.data. Also adding temp file to staging dir is not a good idea. Then it seems that the clipboard path is used for duplication of nodes, if it's so then there is a function `duplicate_node` in `openpype.hosts.nuke.api.lib`. If it's not enough then function `_duplicate_node_temp` can be made as public and used here. Which does actually similar thing.

maxpareschi: unfortunately i've seen that the code changes we did are built on 3.11.1, whereas develop changed a lot of the stuff in extract_slate_frame. There was no function like that at that time.
I need to test this on develop to actually see if this just works as we would like and propose the fixes there.

jakubjezek001:
This issue was solved in https://github.com/pypeclub/OpenPype/pull/3245. It is baking the thumbnail slate before the slate node input. It is ofcourse depending on options comming from `ExctractReviewDataMov`. 

Thank you very much @maxpareschi for your approach, we had already in development the multistream workflow and slate processing was the next step.

PR #3427, opened 2022-06-28
Nuke: Slate frame is integrated

iLLiCiTiT:
## Brief description
Slate frame is missing in representation files. It is primarilly caused by https://github.com/pypeclub/OpenPype/pull/3407 but the same issue could happen even before.

## Description
Add slate filename to representation files in Extract Slate.

## Addiotional information
I've used input with different frame range then write node frame range which caused that slate was not rendered at all (file was rendered but without slate content). Maybe we should force duplicated netflix gizmo to use frame we define as slate frame (e.g. prepend FrameRange/Retime node)?

## Testing notes:
1. Create render in Nuke
2. Add Netflix before the write node
3. Hit publish
+ Farm publishing (I didn't test it)

### DIFF
--- a/openpype/hosts/nuke/plugins/publish/extract_slate_frame.py
+++ b/openpype/hosts/nuke/plugins/publish/extract_slate_frame.py
@@ -4,6 +4,7 @@ import nuke
 import copy
 
 import pyblish.api
+import six
 
 import openpype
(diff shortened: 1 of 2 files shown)


jakubjezek001 (approved):
Tested on Farm publish withou out rendered frames and with and all seems to be ok.

PR #3437, opened 2022-06-29
Maya: Collect instances preserve handles even if frameStart + frameEnd matches context

BigRoy:
## Brief description

This removes the fallback to context `handleStart` and `handleEnd` if instance `frameStart` and `frameEnd` matches context.

## Description

Now it will **always** inherit `handleStart` and `handleEnd` from the context (current working asset's handles in project manager) **unless** the instance explicitly sets values for the handles.

This means that for a `pointcache` instance it will never match the context because the [pointcache instance has its own `handleStart` and `handleEnd` attribute](https://github.com/pypeclub/OpenPype/blob/3bb6d8aea083220ee570b61b9c741e03aa7524c9/openpype/hosts/maya/api/lib.py#L370-L371) on the publish instance. _(This is true for all time based instances in Maya that use the `collect_animation_data` which I think is all time based instances in Maya!)_ We might be off improving that more so that by default on creating an instance it takes the handles and its values from the current working context.

If we really want to keep some of the old behavior we could add an extra attribute on publish instances that toggles the use of the `handleStart`/`handleEnd` from the instance, like `useCustomHandles` or `handlesFromInstance` or something like which can e.g. have a default value in settings so studios can decide whether it's more up to the artists or not.

## Additional info

Without this change the artist could never set custom start/end handles if the start and end frame would match that of the asset in the project manager.

## Testing notes:
1. Check whether publish frames for instances now work as expected.


mkolar:
I remember discussing this at quite some length with a client at the time when this was implemented. The general issue at hand was, how to differentiate between the cases, where the range should be validated against the DB, so we don't produce BS data, vs allowing the artist to bypass it somehow reasonably per instance. 

The intended workflow was that if the artist is customizing the frame range, it is best to simply set different start and end, whether we have handles or not, if they want to override.

Truth be told, the only reason this is there is the fact that the publishing GUI is not very friendly towards choosing to validate one instance, but skip another. This would be trivial to change and make it proper in the new published GUI, unfortunately, we're failing to find the time to start converting maya families to it.

mkolar:
After quite some time in production, I agree that it's probably more intuitive to allow the override anytime, but then the validation of frame range is completely useless

mkolar:
Just a note that we haven't forgotten about this at all, just considering if it can nastily influence anything in production.

BigRoy:
Should this get another "test" some time soon for consideration?

BigRoy:
Just pinging once more here to check on what's the best way to get this in.

BigRoy:
Before merging this PR note that Collect Render should actually be fixed too since it does [the same thing here](https://github.com/ddesmond/OpenPype/blob/9d8d3df8b8f5c9546e05e18b3caf9640a44ffd84/openpype/hosts/maya/plugins/publish/collect_render.py#L224-L239)

LiborBatek (approved):
Had been tested and it keeps the custom frame handles for ```pointcache``` family intact and works as expected. Other subset matches frame ranges with handles from DB values.

image

and DB values
image

Works ok.

antirotor (approved):

Issue #3415, opened 2022-06-27
Support for multiple build versions


antirotor:
## Feature:

Add support for multiple build versions on the same machine. It would allow easier testing and rollback (no need to re-installing and zip version shuffling)

### Steps:

- change Inno Setup to allow multiple versions intalled on Windows
- add zips to <major>.<minor> subfolder to user data

Build 3.12.x should look for its zip updates in `%LOCALAPPDATA%\pypeclub\openpype\3.12\*`

BigRoy:
This would be extra nice because it might allow to install/update to a new version without requiring to have all dependencies unlinked of a previous version (e.g. having all mayabatch closed, etc.)

PR #3444, opened 2022-06-30
Windows installer: Clean old files and add version subfolder

antirotor:
## Enhancement

This PR is addressing issue where old file were left when OP version was reinstalled to the same folder.  This has one important effect - this will currently delete all files there without any remorse (deleting any user stuff that might be there).

I've also added version subfolder to the path so OpenPype will install to `C:\Program Files (x86)\OpenPype\3.12.1-nightly.1`
This is in preparation for implementing #3415


antirotor:
Damn, wrong branch...

PR #3446, opened 2022-06-30
General: Use query functions in load utils

iLLiCiTiT:
## Brief description
Load utils are using query functions to get documents from project.

## Description
Load utils are using query functions from `openpype.client`. Function `_get_project_connection` in client requires passed project.  Function `get_representations_parents` always returns parents for all passed representations but they are set to `None` by default (tuple expanding will work all the time).

## Testing notes:
1. Load plugins should be visible
2. Run load plugin in Library loader
3. Run load plugin Loader
4. Run subset load plugin in any tool (we have probably only delete old versions?)


kalisp (approved):
Tested in Maya, with Site Sync, seems to work.

PR #3456, opened 2022-06-30
Support for mutliple installed versions

### LINKED ISSUE #3415: Support for multiple build versions

### LINKED ISSUE #3416: Deadline: support for version specific job
## Feature

OpenPype had an ability to set OP version per Deadline job and that requested version was used to process it. It was limited because of "one build version per machine/user" limitation, but once this (#3415) will be solved, jobs on farm could be locked to specific OpenPype version (or range of them). This could be influenced by the Settings where you could set "always use the one set for studio" or use per-job one.

This would geratly ease testing of new features on the farm.

[cuID:]

### LINKED ISSUE #3439: Deadline: add information about used OP version to Job
## Enhancement

Mainly for debugging purposes it would be nice to have information about the OpenPype version used to submit job on the farm. Currently this information is lost and even with #3416 it would be nice to have it there (because it might be configured to "use studio version".

We can store it as Job Extra Info fields.

antirotor:
## Feature

This PR is adding support for multiple installed versions (builds) on the same system. It will allow easy rollback and testing on farms and without disrupting existing running sessions.

On Deadline, it is now passing OpenPype version as `OPENPYPE_VERSION` to the job. If this environment variable is found on the job, Deadline will try to find compatible version build to use for launching that version of OpenPype. So if job was submitted from OpenPype version `3.12.2` Deadline will try to find any build that matches `3.12` and will use the latest available. Deadline will try to find it in directories specified in `OpenPypeInstallationDirs` configuration option.

It relies on functionality implemented here #3445.

## Deployment Notes

Since #3445 you can have multiple versions of OpenPype installed on Windows. This was possible even before with Linux, but it was missing logic that will load only versions from the update zips/repository compatible with the currently running build.

So with this PR, running build will "see" only versions compatible with it and will use only them. For Deadline, where this process is reversed - you know what version was used to create the submission and you need to find correct build to launch it, you need to put builds under some common directory do OpenPype can detect them. On Windows, this is by default `C:\Program Files (x86)\OpenPype` but it can be any directory following logic that its direct subdirectories should contain separate build available. So with Deadline, you need to point it to the directory with following structure: `/some/op/dir/3.12.2-nightly.2`, `/some/op/dir/3.13.0` and so on.

## Testing

For proper testing, you need to create few "mock" build versions of OpenPype. It should be enough to change `openpype/version.py` and running `tools/build.ps1|sh` and after each run store produced version somewhere.

### Visibility of compatible versions
OpenPype should by default see only versions compatible with the running build. So issuing command like:

```powershell
.\.poetry\bin\poetry.exe run python start.py --list-versions
```
or
```powershell
& "C:\Program Files (x86)\OpenPype\3.12.2-nightly.3\openpype_console.exe" --list-versions
```
should show only versions compatible with that particular build.

### Deadline
When testing, don't forget to update Deadline plugins in Deadline repository.

You need to configure OpenPype Deadline plugin:

image

Add path to **OpenPype Installation Directories**. This is path-per-line, first found is used.

Jobs submitted to Deadline should have set `OPENPYPE_VERSION` to correct one. In job log, you should find messages about version resolution. 

### Implementation steps:

- [X] change Inno Setup to allow multiple versions intalled on Windows #3445 
- [X] implement compatibility checks to functions detecting OP versions
- [x] change unzipping and tools to look into version folders in user data folder
- [x] mark as deprecated looking for versions in common folder
- [x] fix unit tests
- [x] implement version selector in the Deadline global event / OP plugin[^draft]

[^draft]: This would go over all specified path to OpenPype in Deadline and try to find requested version among them.

🧑‍🏭 **TODO:** find out how to deal with multiple installations on mac.

Resolves #3415
Resolves #3416
Resolves #3439


m-u-r-p-h-y (dismissed):
sry wrong PR

antirotor:
Currently there is one limitation and that is that Deadline cannot work with the sources. It will find the version correctly, but then it is looking for `openpype_console` executable that is present only in the build. Question is how to handle this in platform independent manner, because running OP from sources basically involves running Poetry like this `./.poetry/bin/poetry run python start.py`

iLLiCiTiT:
It's hard to review when there are changes from develop.

m-u-r-p-h-y (changes requested):
created a build and put it in the defined folder



```
Error: Global job preload script 'C:\ProgramData\Thinkbox\Deadline10\workers\WS01\plugins\62e251cea96635187cc30a45\GlobalJobPreLoad.py': TypeError : can only concatenate str (not "list") to str (FranticX.Scripting.PythonNetException)
  File "C:\ProgramData\Thinkbox\Deadline10\workers\WS01\plugins\62e251cea96635187cc30a45\GlobalJobPreLoad.py", line 290, in __main__
    inject_openpype_environment(deadlinePlugin)
  File "C:\ProgramData\Thinkbox\Deadline10\workers\WS01\plugins\62e251cea96635187cc30a45\GlobalJobPreLoad.py", line 105, in inject_openpype_environment
    "The path to the render executable can be configured " + 
```

image

antirotor:
Closing in favour of pure 3.13.x based branch to ease out testing.

PR #3457, opened 2022-07-01
General: Use query functions in rest api calls

iLLiCiTiT:
## Brief description
Rest api calls in avalon module use query functions instead of `AvalonMongoDB`.

## Additional info
Not sure if are the rest points used anywhere? Maybe we could removed them.

## Testing notes:
It should be possible to receive project documents using rest api points. Try access `http://localhost:8079/avalon/projects` when tray is running.

### DIFF
--- a/openpype/modules/avalon_apps/rest_api.py
+++ b/openpype/modules/avalon_apps/rest_api.py
@@ -5,7 +5,12 @@ from bson.objectid import ObjectId
 
 from aiohttp.web_response import Response
 
-from openpype.pipeline import AvalonMongoDB
+from openpype.client import (
+    get_projects,
+    get_project,
+    get_assets,
+    get_asset_by_name,
+)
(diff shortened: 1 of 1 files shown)


64qam (approved):

openpype/modules/avalon_apps/rest_api.py
```diff
@@ -14,19 +19,13 @@ def __init__(self, resource):
...
-        output = []
-        for project_name in self.dbcon.database.collection_names():
-            project_doc = self.dbcon.database[project_name].find_one({
-                "type": "project"
-            })
-            output.append(project_doc)
+        output = [
+            project_doc
+            for project_doc in get_projects()
+        ]
```
BigRoy, line 28: Is there a reason this isn't just `output = list(get_projects())`?

Issue #2801, opened 2022-02-23
Delete Old Versions relies on Ftrack session


BigRoy:
**Describe the bug**

The [Delete Old Versions library loader action tries to access ftrack](https://github.com/pypeclub/OpenPype/blob/a4b440d080a8a392aa26bf9e333baec20f735c9b/openpype/plugins/load/delete_old_versions.py#L375-L405) even when the Ftrack module is disabled.

**To Reproduce**
1. Disable Ftrack + set no valid ftrack urls
2. Run the delete old versions action
3. You will get some errors

**Expected behavior**

- Skip any Ftrack logic if module is disabled.

**Additional context**

Note that the logic is the same in the [Ftrack Delete Old Versions action](https://github.com/pypeclub/OpenPype/blob/a4b440d080a8a392aa26bf9e333baec20f735c9b/openpype/modules/default_modules/ftrack/event_handlers_user/action_delete_old_versions.py) and there's a lot of duplicated code. If possible it might be nice to unify the logic where possible so there's less duplicated code to maintain. 

I'm assuming the same might be true for parts of `delivery` action in Ftrack and the one in the Library Loader even though that logic at first sight looks much less alike.

kaamaurice:
Hi !

Agreed ! I have issues with `delete_old_versions` in Blender 3.1 and Ftrack is disable and I don't want to use it.

I will propose a bugfix soon to skip Ftrack logic if module is disabled.

** To Reproduce **
- Load an asset in Blender 3.1 with the OpenPYPE loader tool.

## Traceback
```>>> [ *** FAILED to load 1 files
- /Volumes/D11-TRAVAIL/DEV/OpenPype/openpype/plugins/load/delete_old_versions.py
**********
Traceback (most recent call last):

  File "/Volumes/D11-TRAVAIL/DEV/OpenPype/openpype/lib/python_module_tools.py", line 92, in modules_from_path
    module = import_filepath(full_path, mod_name)

  File "/Volumes/D11-TRAVAIL/DEV/OpenPype/openpype/lib/python_module_tools.py", line 37, in import_filepath
    module_loader.exec_module(module)

  File "<frozen importlib._bootstrap_external>", line 883, in exec_module

  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed

  File "/Volumes/D11-TRAVAIL/DEV/OpenPype/openpype/plugins/load/delete_old_versions.py", line 7, in <module>
    import ftrack_api

[...]

PR #3462, opened 2022-07-01
General: Delete old versions is safer when ftrack is disabled

kaamaurice:
# Brief description
The [Delete Old Versions](https://github.com/pypeclub/OpenPype/blob/a4b440d080a8a392aa26bf9e333baec20f735c9b/openpype/plugins/load/delete_old_versions.py#L375-L405) should use openpype.settings to get if **ftrack** module is enable and proceed ftrack logic only if needed.

## Description
see issue : Delete Old Versions relies on Ftrack session - #2801

## Testing:
1. Disable Ftrack in the OpenPYPE Studio settings
2. Use Loader tool to action **Delete Old Versions**

Resolves: https://github.com/pypeclub/OpenPype/issues/2801

### DIFF
--- a/openpype/plugins/load/delete_old_versions.py
+++ b/openpype/plugins/load/delete_old_versions.py
@@ -4,13 +4,14 @@ import uuid
 
 import clique
 from pymongo import UpdateOne
-import ftrack_api
+import importlib
 import qargparse
 from Qt import QtWidgets, QtCore
 
 from openpype import style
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
Good catch. There are few minor changes regarding this (made [PR](https://github.com/kaamaurice/OpenPype/pull/6)).

iLLiCiTiT (approved):
Works as expected

PR #3469, opened 2022-07-05
Migrate basic families to the new Tray Publisher

mkolar:
This move all the families that don't require a custom collector from standalone publisher to the new Tray publisher GUI. This is a first step in getting rid of the old standalone publisher.

Testing should be as easy as running the tray publisher from the OpenPype tray menu and trying to publish some families that are available by default. 

This is really just a change in the default settings, to have some reasonable defaults.

### DIFF
--- a/openpype/settings/defaults/project_settings/traypublisher.json
+++ b/openpype/settings/defaults/project_settings/traypublisher.json
@@ -42,7 +42,7 @@
                 "Sculpt"
             ],
             "description": "Clean models",
-            "detailed_description": "Models should only contain geometry data, without any extras like cameras, locators or bones. It should be ready to be loaded into other scenes as is.\n",
+            "detailed_description": "Models should only contain geometry data, without any extras like cameras, locators or bones.\n\nKeep in mind that models published from tray publisher are not validated for correctness. ",
             "allow_sequences": false,
             "extensions": [
                 ".ma",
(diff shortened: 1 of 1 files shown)


jakubjezek001 (dismissed):
Tested `render` family only with exr sequence and mov file. EXR was not able to finish the publishing at all. It was freezed on integrate new plugin at coppying temp files process. 

Mov file was showing review toggle on but it was not adding it into instance data so no reviewable was generated.

iLLiCiTiT:
Not sure how much of these issues are related to this PR :)

> Mov file was showing review toggle on but it was not adding it into instance data so no reviewable was generated.

This looks like a bug of the collector plugin.

> Tested render family only with exr sequence and mov file. EXR was not able to finish the publishing at all. It was freezed on integrate new plugin at coppying temp files process.

This is strange, but again not related to this PR.

kalisp (approved):
Tested these:
Screenshot 2022-07-12 113846

Seems to work fine.

mkolar:
> Tested `render` family only with exr sequence and mov file. EXR was not able to finish the publishing at all. It was freezed on integrate new plugin at coppying temp files process.
> 
> Mov file was showing review toggle on but it was not adding it into instance data so no reviewable was generated.

Was that publishing both at the same time as two representations or two individual publishes? I managed to publish just fine.

PR #3470, opened 2022-07-05
Flame: solved problem with multi-selected loading

jakubjezek001:
## Brief description
Loading with multiselect now works as it suppose.

## Description
But which was blocking to load clips was resolved by adding optionable template keys into `clip_name_template` of each loader plugin and settings.


iLLiCiTiT (approved):
Can't test but code almost didn't change.

PR #3474, opened 2022-07-07
General: thumbnail extractor fix

jakubjezek001:
## Brief description
The extractor is bypassed to ffmpeg if oiiotool is failing

## Description
Oiiotool was failing on some version of CentOS that seems to not handle oiiotool correctly, so as a fallback we now use ffmpeg when oiiotool fails to do its conversion.

# Dependency
- also containing: https://github.com/pypeclub/OpenPype/pull/3455


Allan-I (approved):
Thumbnail extractor code looks good to me!

PR #3475, opened 2022-07-08
General: Better arguments order in creator init

iLLiCiTiT:
## Brief description
Change order of creator init arguments to have most used project settings as first to less important arguments as last.

## Description
In most of cases when creator want to override initialization it's because of using project settings thus it's better to have it as first argument, then system settings and then what is probably not used in custom init just to be passed to super init.

## Additional info
Breaks backwards compatibility. So this should be done now before new publisher is used in all hosts.

Order was change from most used to less used arguments.
```
# Argument in init
# /// Before \\\
def __init__(self, create_context, system_settings, project_settings, headless=False):

# /// Now \\\
def __init__(self, project_settings, system_settings, create_context, headless=False):

# Demonstration when project settings are used
# /// Before \\\
def __init__(self, create_context, system_settings, project_settings, *args, **kwargs):
    super().__init__(create_context, system_settings, project_settings, *args, **kwargs)

# /// Now \\\
def __init__(self, project_settings, *args, **kwargs):
    super().__init__(project_settings, *args, **kwargs)
```

### DIFF
--- a/openpype/hosts/aftereffects/plugins/create/create_render.py
+++ b/openpype/hosts/aftereffects/plugins/create/create_render.py
@@ -17,11 +17,8 @@ class RenderCreator(Creator):
 
     create_allow_context_change = True
 
-    def __init__(
-        self, create_context, system_settings, project_settings, headless=False
-    ):
-        super(RenderCreator, self).__init__(create_context, system_settings,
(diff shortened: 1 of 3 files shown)


kalisp (approved):
NP Create works in PS, AE

PR #3484, opened 2022-07-11
NewPublisher: Added ability to use label of instance

iLLiCiTiT:
## Brief description
Instance can have defined label in it's data which is shown in instances view.

## Description
Previously instance used only subset name of instance which makes it complicated when instances are created in bulk for multiple different contexts so each has same name. If label is not available then subset name is used as fallback.

## Additional info
To use label just pass `"label"` to instance data of `CreatedInstance`.

## Testing notes:
1. Create instance which has `"label"` key in data
2. Label should be visible in create UI

### DIFF
--- a/openpype/pipeline/create/context.py
+++ b/openpype/pipeline/create/context.py
@@ -497,7 +497,7 @@ class CreatedInstance:
         return self._data["subset"]
 
     @property
-    def instance_label(self):
+    def label(self):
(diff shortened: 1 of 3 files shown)


jakubjezek001 (approved):

jakubjezek001:
resolve https://github.com/pypeclub/OpenPype/issues/3482

PR #3494, opened 2022-07-12
TrayPublisher: Added more options for grouping of instances

iLLiCiTiT:
## Brief description
Grouping of instances has much more abilities. Instances can have their own group defined in data and creator plugin can define it's group label. This is just UI purpose change.

## Description
Group label is defined by this order (if value is empty then next value is used):
- key `"group"` in created instance data
- attribute `group_label` on creator plugin
- attribute `label` on creator plugin
- attribute `identifier` on creator plugin

## Additional info
Identifier must be available on creator plugin so it is end point. Default implementation returns creator's `family` which can be considered as last source.

## Testing notes:
Creators group label
1. Define `group_label` on creator plugin and create it's instances
2. They should be grouped under that label

Instance group label
1. Create an instance with some string in `"group"` key in it's data
2. Instance should be grouped under that group in UI

Resolves https://github.com/pypeclub/OpenPype/issues/3490

### DIFF
--- a/openpype/pipeline/create/context.py
+++ b/openpype/pipeline/create/context.py
@@ -505,7 +505,7 @@ class CreatedInstance:
 
     @property
     def group_label(self):
-        label = self._data.get("group_label")
+        label = self._data.get("group")
         if label:
             return label
         return self.creator.get_group_label()
--- a/openpype/pipeline/create/context.py
+++ b/openpype/pipeline/create/context.py
@@ -504,6 +504,13 @@ class CreatedInstance:
         return label
 
     @property
+    def group_label(self):
+        label = self._data.get("group_label")
+        if label:
+            return label
+        return self.creator.get_group_label()
+
+    @property
     def creator_identifier(self):
         return self.creator.identifier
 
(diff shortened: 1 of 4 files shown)


jakubjezek001 (changes requested):
I have tested two scenarios described in the PR description:
- added `group` key to instance data > didnt work
- added `group_label` key to instance data > this worked

openpype/pipeline/create/context.py
```diff
@@ -503,6 +503,13 @@ def label(self):
             label = self.subset_name
         return label
 
+    @property
+    def group_label(self):
+        label = self._data.get("group_label")
```
iLLiCiTiT, line 508: Fix key
```suggestion
        label = self._data.get("group")
```

iLLiCiTiT:
> I have tested two scenarios described in the PR description:
> 
> * added `group` key to instance data > didnt work
> * added `group_label` key to instance data > this worked

Copy->Paste issue

jakubjezek001 (approved):

PR #3496, opened 2022-07-12
Kitsu: Use query function from client

iLLiCiTiT:
## Brief description
Use query functions instead of database queries in kitsu/zou sync.

## Description
Goal is to replace all mongo specific queries in code with functions that can use different source of data. At this moment we're replacing just receiving of data. In some cases the logic was slightly modified as direct queries by custom subkeys in data are not allowed at this moment. Places where it could make potential issues (even before this PR) were commented.

## Additional information
Made sure that `dbcon` won't change type from `AvalonMongoDB` to `pymongo.Collection` during sync which could cause some issues especially with these changes.

## Testing notes:
1. Kitsu integration should work as did
2. Sync from and to OP should work as expected


openpype/modules/kitsu/utils/sync_service.py
```diff
@@ -335,14 +343,25 @@ def _new_task(self, data):
...
+            project_name,
+            asset_names=[parent_name],
+            fields=["_id", "data.zou.id", "data.tasks"]
         )
+        asset_doc = None
+        for _asset_doc in asset_docs:
+            doc_zou_id = _asset_doc.get("data", {}).get("zou", {}).get("id")
+            if doc_zou_id == parent_zou_id:
+                asset_doc = _asset_doc
+                break
```
Tilix4, line 364: Don't you have an API function `get_single_asset` like function? It'd be better than having to filter for every module no?

Or use your `get_asset_by_name`? Doing `get_asset_by_name(task['entity']['name'])`

openpype/modules/kitsu/utils/update_op_with_zou.py
```diff
@@ -33,16 +39,29 @@ def create_op_asset(gazu_entity: dict) -> dict:
     }
 
 
+def get_kitsu_project_name(project_id: str):
```
Tilix4, line 42:
```suggestion
def get_kitsu_project_name(project_id: str)->str:
```

openpype/modules/kitsu/utils/update_op_with_zou.py
```diff
@@ -33,16 +39,29 @@ def create_op_asset(gazu_entity: dict) -> dict:
...
+    """Get project name based on project id in kitsu.
+
+    Args:
+        project_id (str): Id of project in Kitsu.
```
Tilix4, line 46:
```suggestion
        project_id (str): UUID of project in Kitsu.
```

openpype/modules/kitsu/utils/update_op_with_zou.py
Tilix4: I think this should belong into an `utils.py` file as it's shared into both `update_` python files.

openpype/modules/kitsu/utils/update_op_with_zou.py
```diff
@@ -33,16 +39,29 @@ def create_op_asset(gazu_entity: dict) -> dict:
...
+
+    Returns:
+        str: Project name which has project in Kitsu.
```
Tilix4, line 49:
```suggestion
        str: Name of Kitsu project.
```

openpype/modules/kitsu/utils/update_op_with_zou.py
```diff
@@ -162,14 +179,32 @@ def update_op_assets(
...
         if visual_parent_doc_id is None:
             # Find root folder doc
-            root_folder_doc = dbcon.find_one(
-                {
-                    "type": "asset",
-                    "name": entity_parent_folders[-1],
-                    "data.root_of": substitute_item_type,
-                },
-                ["_id"],
+            root_folder_docs = get_assets(
```
Tilix4, line 182: As the function is `get_assets`, `root_folder_docs` is not explicit enough, if you want to mention that they are `docs`, use `root_folder_assets_docs`.

openpype/modules/kitsu/utils/update_op_with_zou.py
```diff
@@ -162,14 +179,32 @@ def update_op_assets(
...
+                project_name,
+                asset_name=[entity_parent_folders[-1]],
+                fields=["_id", "data.root_of"]
             )
+            # NOTE: Not sure why it's checking for entity type?
```
Tilix4, line 187: If I remember correctly, it is for sorting by `Assets` or `Shots`

Tilix4 (changes requested):

openpype/modules/kitsu/utils/sync_service.py
iLLiCiTiT: I'm ok with that. I wasn't sure if the check of `zou.id` should stay?

openpype/modules/kitsu/utils/sync_service.py
```diff
@@ -119,17 +124,16 @@ def _update_project(self, data):
 
         # Write into DB
         if update_project:
-            self.dbcon = self.dbcon.database[project_name]
+            self.dbcon.Session["AVALON_PROJECT"] = project_name
             self.dbcon.bulk_write([update_project])
 
     def _delete_project(self, data):
         """Delete project."""
-        project_doc = self.dbcon.find_one(
```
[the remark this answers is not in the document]
iLLiCiTiT: You actually have to query from kitsu in all cases because you have to get the project name from there. Without project name you can't query project document by zou id because you don't know in which collection it should be found.

openpype/modules/kitsu/utils/update_op_with_zou.py
iLLiCiTiT: But if you would have `Bob` in `Shots` and `Assets` OpenPype would not be able to work properly because it would be duplicated asset name.

Long store short: If there is and asset with same name in OpenPype there are 2 possible:
1. There is duplicated name in project which can't happen and should skip it's creation
2. It is the same entity but was moved from `Shots` to `Assets` (for example) so should update the type
In both cases it should not create new asset document with same name.

openpype/modules/kitsu/utils/update_op_with_zou.py
iLLiCiTiT: I just followed naming with `root_folder_doc`.

openpype/modules/kitsu/utils/sync_service.py
Tilix4: It is not supposed to change, and the name is supposed to be unique in both DB (zou and OP), then we can drop this UUID check.

Tilix4 (changes requested):

openpype/modules/kitsu/utils/update_op_with_zou.py
iLLiCiTiT: >  I don't understand why you talk about same name issue here.

Because this code here can cause the duplicated name issue.

The logic here is:
Query asset with name <NAME> and type <TYPE> if was not found then create the asset.

But, what if the asset with the <NAME> already exists but does not have the <TYPE>? In that case you will create second asset document with name <NAME> and from that moment OpenPype won't work correctly.

iLLiCiTiT: Similar issue to this https://github.com/pypeclub/OpenPype/pull/3496#discussion_r919721371

Maybe just skip the check of `root_of`, query the asset byt name and it would do the same thing.

openpype/modules/kitsu/utils/sync_service.py
iLLiCiTiT: Done

Tilix4 (approved):

PR #3497, opened 2022-07-12
Nuke: load clip with options from settings

jakubjezek001:
## Brief description
Loading of options can be set in settings.

## Description
In case production needs to set loading options for all retimed plates to ignore retime settings, this can be now done from settings.

## Testing notes:
1. open nuke workfile
2. make sure you are having some retimed plate to load - ideally more that one version.
3. Load first older version with option set to add_retime to False
4. Update to latest version
5. It should load first read node without retime node connected to it, and then during update it should just update the read node without retime nodes.

### DIFF
--- a/openpype/hosts/nuke/plugins/load/load_clip.py
+++ b/openpype/hosts/nuke/plugins/load/load_clip.py
@@ -54,25 +54,27 @@ class LoadClip(plugin.NukeLoader):
     script_start = int(nuke.root()["first_frame"].value())
 
     # option gui
-    defaults = {
+    options_defaults = {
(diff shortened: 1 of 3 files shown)


jrsndl:
Using Load works as advertised. Changing load clip options in load tool or project settings correctly adds or skips adding the TimeWarp node. I tested with Nuke 12.2, speedup and slowdown, one or two layers.

Not sure if connected to this PR, but using Manage to switch version of the plate(s) to latest produces warning "node LoadClip_exr already exists", and TimeWarp nodes are not replaced, but added.
image

jrsndl (approved):

PR #3498, opened 2022-07-12
NewPublisher: Keep plugins with mismatch target in report

iLLiCiTiT:
## Brief description
Add plugins that don't have matching targets to publish report so we have available all loaded plugins.

## Description
Filtered plugins are marked as skipped so they are visible in the report view. The main reason is confusion that plugin is not visible in view and is not in crashed plugin paths.

## Testing notes:
1. Run validation in new publisher
2. Go to report view
3. Unhide skipped plugins
4. You should see plugins that were skipped because if mismatched targets

### DIFF
--- a/openpype/pipeline/create/context.py
+++ b/openpype/pipeline/create/context.py
@@ -709,6 +709,7 @@ class CreateContext:
         self.manual_creators = {}
 
         self.publish_discover_result = None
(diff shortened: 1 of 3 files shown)


kalisp (approved):
Looks good.

jakubjezek001 (approved):

PR #3499, opened 2022-07-12
General: Event system

iLLiCiTiT:
## Brief description
It is possible to create independent event systems with own callback and topics.

## Description
This is small change related to current event system which works only in global scope but in some cases it would be helpful to have independent event system from global scope. For example in tools which are not based on Qt signals which is not usefull all cases especially when their controller should be Qt independent.

## Additional information
For example new publisher already has it's own implementation of independent event system but it's not nice to read or maintain as most of the logic is related to handling references to callbacks. This would make the controller much cleaner.

## Note
Event system is not used anywhere because this change should be properly tested before it's used so we don't break current usages.

## Testing notes:
Event callbacks should work as they did.
- for example `taskChanged` from workfiles tool should be captured in hosts (e.g. Maya or Nuke) and frame start/end should be updated to new context


jakubjezek001 (approved):
Tested in Nuke by context change in workfiles and Ftrack tracking had switched, the nuke context too.

PR #3501, opened 2022-07-12
NewPublisher: Groups work and enum multivalue

iLLiCiTiT:
## Brief description
Instances refresh use right source for groups and enum attribute can handle multivalue.

## Description
Groups on update of instance were still using old implementation. And enum multivalue didn't work because combobox did not have line edit available.

## Testing notes:
Change view few times after instance creation.
- it should not crash and should use right groups

Select multiple instances which have defined enumerator attribute.
- it should not crash and `< Multivalue >` should be showed as label if instances have different values


jakubjezek001 (changes requested):
Follow the screenshot gif please. It seems to have a bug still.
traypublisherv03

PR #3502, opened 2022-07-13
Add pack and unpack convenience scripts

mkolar:
These scripts provide easier way to export a full project with database and all of it's data and subsequently load it back.

Courtesy of @jakubjezek001

The most common use is for testing purposes. This makes it possible to export a current state of projects, do testing publishes and then return to the previous state. The result is a zip file containing everything needed for the project, so it is not suitable for packing full project at production scale. 

## Additional info
The pack project currently doesn't include project settings

## Testing notes:
packing
1. run `pack_project.ps1` with a project name as an argument. 
2. you should be left with a zip file next to the project root folder

unpacking
1. run `unpack_project.ps1 path/to/project.zip`
2. check if the project has been unpacked and the original project folder renamed for a backup


iLLiCiTiT (approved):
Scripts are working

PR #3513, opened 2022-07-14
TrayPublisher: Simple creation enhancements and fixes

iLLiCiTiT:
## Brief description
Simple creators can handle multiple path items and can (really) create reviews from representations.

## Description
Simple creators can have "multiple item" in filepath which will cause that there is more then one representaion on the instance. Review collector was removed because there also must be a way how to mark a representation for review which is tricky with only publish plugin attributes thus review marking was moved to creator and it's related collector. For simple creators is last representation used for instance source. Simple creators also have second input for review representation.

## Additional info
Moved order of collecting of cleanup keys earlier (it has none dependency and zero effect). Removed collect review family collector from tray publisher. Validation of source filepaths will crash when source filepath is set with empty list. Instance data containt data from asset which are collected in collect instance anatomy data (`fps`, `frameStart`, `frameEnd`, `handleStart`, `hanleEnd`) which are required for most of plugins dealing with rendering (e.g. extract review). Reset of publish view will first save changes before refresh (refresh will not cause loose of changed attributes). Last representation path is used as source for instance. Small change to be able zoom text in report view. FileDef can have defined custom extensions label.

## Testing notes:
1. Open tray publisher
2. Create something reviewable using simple creators and mark it to be review
3. Hit publish and review should be created
- if filepaths are empty validator should tell you
- publishing should crash if **only** review path is entered in simple creation
- review is not created if there is no file in reviewable input


kalisp (approved):
Solves my issues.
Two separate boxes for representations vs review representations are bit weird, but I understand why they are there.

PR #3514, opened 2022-07-14
Maya: Renderman display output fix

antirotor:
## Problem

Imagine you have two displays in a scene - `beauty` and let say `mattes`. In the `beauty`, there are channels `Z`, `Env`, `id` (and of course `Ci` and `a`) and in the `mattes` there are `foo`, `bar`, `baz`:

```
beauty
   |- Z
   |- Env
   \- id
mattes
   |- foo
   |- bar
   \- baz
```

publishing **without** prior IPR render will result correctly in two displays (and two expected sequences) for example `Layer_beauty.####.exr` and `Layer_mattes.####.exr`. 

But when you do the IPR before publishing, Renderman will breakout existing displays (`beauty` and `mattes`) into `beauty_Z`, `beauty_Env`, `beauty__id` and similar with mattes - `mattes_foo`, `mattes_bar`, `mattes_baz`.

## Solution

is to force rfm to output only displays for render destination (not all - viewport/it/driver). This PR is doing exactly this, plus adding proper `multipart` flag on formats supporting it.

## Testing

Create test scene and add few displays with some channels. Start IPR render, close it and then publish the render job to farm. After rendering, it should publish ok.

This code will show you all expected displays (display == output file sequence):

```python
from rfm2.api.displays import get_displays

displays = get_displays(override_dst="render")["displays"]
print("v" * 80)
for name, display in displays.items():
    print("-" * 80)
    print(f"display name: {name}")
    channels = ", ".join(display['params']['displayChannels']['value'])
    print(f"channels: {channels}")
print("-" * 80)
print("^" * 80)
```

### DIFF
--- a/openpype/hosts/maya/api/lib_renderproducts.py
+++ b/openpype/hosts/maya/api/lib_renderproducts.py
@@ -1087,7 +1087,7 @@ class RenderProductsRenderman(ARenderProducts):
             "d_tiff": "tif"
         }
 
-        displays = get_displays()["displays"]
+        displays = get_displays(override_dst="render")["displays"]
         for name, display in displays.items():
             enabled = display["params"]["enable"]["value"]
             if not enabled:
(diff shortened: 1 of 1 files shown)


m-u-r-p-h-y (approved):
Tested with M2022 and RFM 24.3 and RFM 24.4

output aftrer running and closing IPR

```
vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
--------------------------------------------------------------------------------
display name: beauty
channels: Ci, a1, P
--------------------------------------------------------------------------------
display name: _id
channels: id
--------------------------------------------------------------------------------
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

PR #3517, opened 2022-07-15
Traypublisher: added thumbnail extract 

kalisp:
## Brief description
Used global plugin for thumbnail extraction even in Traypublisher

## Additional info
Added check if thumbnail representation is not present, created by different plugin by any chance.
ExtractThumbnailSP should be removed when SP is removed, no need to copy and have 2 plugins.

## Testing notes:
1. try publish render with TrayPublisher
2. it should create thumbnail in publish folder


iLLiCiTiT:
I would just reorder the words in method name to `_already_has_thumbnail`.

iLLiCiTiT (approved):
Small comment otherwise approved

PR #3518, opened 2022-07-15
TrayPublisher: copied ValidateFrameRange from SP

kalisp:
## Brief description
Validate frame range was missing from TrayPublisher

## Additional info
SP will be removed in the future, copied this to TrayPublisher to keep functionality.

## Testing notes:
1. publish render sequence with non matching frame count against the DB
2. Validator should fail
3. check that validator could be disabled


openpype/hosts/traypublisher/plugins/publish/validate_frame_ranges.py
```diff
@@ -0,0 +1,72 @@
...
+    def process(self, instance):
+        # Skip the instance if is not active by data on the instance
+        if not self.is_active(instance.data):
+            return
+
+        if any(re.search(pattern, instance.data["task"])
+               for pattern in self.skip_timelines_check):
+            self.log.info("Skipping for {} task".format(instance.data["task"]))
+
+        asset_data = lib.get_asset(instance.data["asset"])["data"]
```
iLLiCiTiT, line 38:
```suggestion
        asset_doc = instance.data["assetEntity"]
        asset_data = asset_doc["data"]
```

openpype/hosts/traypublisher/plugins/publish/validate_frame_ranges.py
```diff
@@ -0,0 +1,72 @@
...
+            self.log.info("No representations, skipping.")
+            return
+
+        ext = repre[0]['ext'].replace(".", '')
+
+        if not ext or ext.lower() not in self.check_extensions:
+            self.log.warning("Cannot check for extension {}".format(ext))
+            return
+
+        files = instance.data.get("representations", [None])[0]["files"]
```
iLLiCiTiT, line 56: This is strange code, it deserve at least some refactor
```suggestion
        repres = instance.data.get("representations")
        if not repres:
            self.log.info("No representations, skipping.")
            return
            
        first_repre = repres[0]
        ext = first_repre['ext'].replace(".", '')

        if not ext or ext.lower() not in self.check_extensions:
            self.log.warning("Cannot check for extension {}".format(ext))
            return

        files = first_repre["files"]
```

openpype/hosts/traypublisher/plugins/publish/validate_frame_ranges.py
```diff
@@ -0,0 +1,72 @@
...
+    check_extensions = ["exr", "dpx", "jpg", "jpeg", "png", "tiff", "tga",
+                        "gif", "svg"]
+    skip_timelines_check = []  # skip for specific task names (regex)
+
+    def process(self, instance):
+        # Skip the instance if is not active by data on the instance
+        if not self.is_active(instance.data):
+            return
+
+        if any(re.search(pattern, instance.data["task"])
```
iLLiCiTiT, line 34: This logs out that it's skipping but actually is not.

openpype/hosts/traypublisher/plugins/publish/validate_frame_ranges.py
```diff
@@ -0,0 +1,72 @@
...
+            return
+
+        files = instance.data.get("representations", [None])[0]["files"]
+        if isinstance(files, str):
+            files = [files]
+        frames = len(files)
+
+        msg = "Frame duration from DB:'{}' ". format(int(duration)) +\
+              " doesn't match number of files:'{}'".format(frames) +\
+              " Please change frame range for Asset or limit no. of files"
```
iLLiCiTiT, line 63:
```suggestion
        msg = (
            "Frame duration from DB:'{}'  doesn't match number of files:'{}'"
            " Please change frame range for Asset or limit no. of files"
        ). format(int(duration), frames)
```

openpype/hosts/traypublisher/plugins/publish/validate_frame_ranges.py
kalisp: Added condition to check only if `skip_timelines_check` is configured.

64qam (approved):

PR #3521, opened 2022-07-15
Nuke: double slate

jrsndl:
## Brief description
This changes a way to merge slate with review. Now it reencodes, so all apps can read the result with slate correctly.
Also, reviews with audio and slate now have audio again.

I tested on op-atom MXF, DNxHD mov, Prores mov, H264 in mp4. With audio (except MXF) and with two ExtractReviewDataMov Output Definitions. Local publish, clean working folder.

## Testing notes:
1. Make a slated output with and with no audio as MXF and H264 in mp4
2. Compare lengths of the sequence output with MFX and mp4 in Resolve
3. See that Resolve reads all frames correctly


openpype/plugins/publish/extract_review_slate.py
```diff
@@ -328,19 +324,25 @@ def process(self, instance):
                 copy_args = (
                     "-metadata",
                     "-metadata:s:v:0",
+                    "-codec:v",
+                    "-pixfmt",
```
iLLiCiTiT, line 328: Args for `codec` and `pixfmt` should be filled same way as here https://github.com/pypeclub/OpenPype/pull/3521/files#diff-1064edbd703275fe1bdff4855b452f266ceef997f708712a16a9a21740653a68R186

jrsndl: is the link working?

iLLiCiTiT: > is the link working?

Seems like it's not. There is `format_args` and `codec_args` calculated from source (approximately line 186). That must be used as source for codec and pixfmt instead.

They actually must be filled otherwise ffmpeg will use default values for the codec and format.

jakubjezek001 (approved):
I could not replicate the problem you had described, but this PR is working well althought this needs to beet tested as much as possible!

PR #3530, opened 2022-07-18
Refactor Integrate Asset

mkolar:
This is a followup PR from https://github.com/pypeclub/OpenPype/pull/2898 which was merged to an intermediate branch and we've added a few commits. 

To be able to merge this, we've moved `hosts` and `families` filters for both the old and the new integrator to their respective GUI settings. The names of classes don't help here, but we can't easily rename the original class as we need to keep production backwards compatibility with existing project settings overrides.

The settings are here
`project_settings/global/publish/IntegrateAssetNew` (the original integrator)
`project_settings/global/publish/IntegrateAsset` (the refactored integrator)

This PR makes all of the hosts and families default to the refactored integrator. The plan is to merge do develop and look out for any issues that appear while testing any other PRs or production deployments. As soon as an issue appears, that particular combination of host and family should be moved to the legacy integrator in the settings. That action will revert it to the original and won't affect production, however, it will also automatically start generating a list of problematic combinations that need to be addressed in further PRs on the path to eradicate the old and ugly integrator plugin. 

Please make sure that before you deploy develop, or next minor release into production, someone has the rights to quickly change system settings on these two plugins in case of any production issues.

_________________________________________________________________________________

## Instructions for fixing issues cause by this PR

If something breaks during integration, you can hotfix it by editing the exclusion filters in settings `project_settings/global/publish/IntegrateAsset`. 

You can exclude any combination of `host + family` and at from that moment the legacy integrator will be used instead. This should allow you to fix any problems cause on the spot without even restarting your DCC. If you need to do this at any point, please make sure to report the `host + family` combination to us, so it can be looked at. 

This change will make it into the next patch release, considering it can be reverted selectively and doesn't need a new build.


openpype/plugins/publish/integrate.py
```diff
@@ -0,0 +1,835 @@
...
+            ).format(instance))
+
+        else:
+            self.log.debug(
+                "Establishing staging directory "
+                "@ {0}".format(instance_stagingdir)
+            )
+
+        # Ensure at least one representation is set up for registering.
+        repres = instance.data.get("representations")
```
iLLiCiTiT, line 219: Check for representations and their filtering should be first part. Missing representations should just skip the instance instead of raise an error.

openpype/plugins/publish/integrate.py
```diff
@@ -0,0 +1,835 @@
...
+
+    # Representation context keys that should always be written to
+    # the database even if not used by the destination template
+    db_representation_context_keys = [
+        "project", "asset", "task", "subset", "version", "representation",
+        "family", "hierarchy", "username"
+    ]
+
+    # Attributes set by settings
+    template_name_profiles = None
```
iLLiCiTiT, line 177: If we'll change place where publish templates are defined the we should move it from publish plugins to separated part so it's not needed to know plugin name to receive it when it's needed elsewhere.

openpype/settings/entities/schemas/projects_schema/schemas/schema_global_publish.json
```diff
@@ -39,6 +39,61 @@
                 }
             ]
         },
+        {
+            "type": "dict",
+            "collapsible": true,
+            "key": "CollectSubsetGroup",
+            "label": "Collect Subset Group",
```
iLLiCiTiT, line 46:
```suggestion
            "key": "IntegrateSubsetGroup",
            "label": "Integrate Subset Group",
```

iLLiCiTiT:
## Modified code:
- New integrator has same host and families filters as legacy integrator (all of them)
- New integrator marks instances for legacy integrator to skip them and avoid double integration
- Make sure that order of legacy integrator is after new integrator
- New integrator ha settings where is possible to "skip" host and families combinations to easily fix production when something breaks. If instance is skipped using this settings it will be integrated using legacy integrator.
- Settings for publish templates are taken from legacy integrator
    - they will be moved elsewhere out of plugin settings in future PR (this is to avoid break compatibility of settings twice) 
- New integrator does not crash if all representations are marked for deletion and representations filtering happens before registering
- `CollectSubsetGroup` was renamed to `IntegrateSubsetGroup` and was changed it's order to happen in integration
+ few minor changes

### Explanations (To avoid confusion)
- New integrator: Plugin `IntegrateAsset` in `./openpype/plugins/publish/integrate.py`
- Legacy integrator: Plugin `IntegrateAssetNew` in `./openpype/plugins/publish/integrate_legacy.py`

kalisp:
It worked on basic published in:
- AE
- PS
- Nuke

iLLiCiTiT (approved):

mkolar:
Just fixed the exclusion condition. Ready to merge

PR #3531, opened 2022-07-18
General: Mongo core connection moved to client

iLLiCiTiT:
## Brief description
Moved core of Mongo connection `OpenPypeMongoConnection` from `openpype.lib` into `openpype.client`. This change makes `openpype.client` independent from openpype lib functions and can be considered as self contained.

## Description
Changed parts of code using `OpenPypeMongoConnection` to use new source and marked lib functions as deprecated. That is unfortunatelly not possible for classes located there too.

## Testing notes:
- [x] Logging to mongo should work (can be validated using LogViewer in tray)
- [x] Webpublisher should work
- [x] Maya shader publishing can be possible (not sure if is used?)
- [x] OpenPype MongoURL can be changed in local settings UI
- [x] Remote (automated) publishing can get and update jobs to do
- [x] Ftrack event server (service) can process ftrack events
- [x] Integration to slack should work
- [x] Project Settings can be loaded and saved

### DIFF
--- a/openpype/client/__init__.py
+++ b/openpype/client/__init__.py
@@ -1,3 +1,7 @@
+from .mongo import (
+    OpenPypeMongoConnection,
+)
+
 from .entities import (
     get_projects,
     get_project,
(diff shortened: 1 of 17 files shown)


kalisp (approved):
Tested webpublishing, automatic testing, works.

PR #3540, opened 2022-07-20
Maya: add additional validators to Settings

antirotor:
## Enhancement

With this PR all Maya validators that are more or less safe to disable/enable are added to the Settings.

It is difficult to test, but the state of the validators should be the same as before. They show up in the Settings, where they can be configured (enabled/optional/active)


m-u-r-p-h-y:
Just going through the settings and found some inconsistency in the way the validators are marked as enabled/disabled

image

what is the difference in making one active with the active toggle or enabling/disabling it completely?

m-u-r-p-h-y:
I can see it now. Enabled but not active ;)

image

m-u-r-p-h-y (approved):
I was playing with validators a bit.

enabling, disabling, and running the validators and did not experience any trouble.

seems good to go . . .

PR #3555, opened 2022-07-22
Nuke: publish existing frames with slate with correct range

jakubjezek001:
## Brief description
Range is correct when published slate from existing frames

## Description
The miscalculation was removed.

## Testing notes:
1. open nuke workfile with slate workflow
2. publish from existing frames
3. notice that frame range is correct and not frame shorter

### DIFF
--- a/openpype/hosts/nuke/api/plugin.py
+++ b/openpype/hosts/nuke/api/plugin.py
@@ -181,8 +181,6 @@ class ExporterReview(object):
             # get first and last frame
             self.first_frame = min(self.collection.indexes)
             self.last_frame = max(self.collection.indexes)
-            if "slate" in self.instance.data["families"]:
-                self.first_frame += 1
         else:
             self.fname = os.path.basename(self.path_in)
             self.fhead = os.path.splitext(self.fname)[0] + "."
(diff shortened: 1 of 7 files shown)


jrsndl (changes requested):
The "farm" and "using existing frames" options with slate works with this PR.

Using local publish produces double slate in review files (EXRs are OK). Local publish works with current develop, so I assume fixed farm publish broke local publish.

jakubjezek001:
> The "farm" and "using existing frames" options with slate works with this PR.
> 
> Using local publish produces double slate in review files (EXRs are OK). Local publish works with current develop, so I assume fixed farm publish broke local publish.

Thanks for catching this! Please try it again now.

jrsndl (changes requested):
Local publish makes correct H264 and MXF, but there is one  frame offset in published EXR files. Here the shot has range 997-1006, and slate at 996. The published exr sequence ands up being 997-1007.
image

jrsndl (approved):
When I exclude Integrate Asset for Nuke host and render + write Families, publishing works fine for local and farm publish, with clean or "dirty" work folder.

I tested h264 and MXF.

jakubjezek001:
@jrsndl please try it one more time. I found some flaws in Farm and Local publishing.

jrsndl (approved):
Renders have correct range now. I've tested EXR, MXF and H264 outputs with slate, on farm and local renders, with clean and dirty work folder.
I also made a prerender, with review. All good.

I've tested together with OP-3642_Nuke-Local-vs-Farm-Publish-family-inconsistency.
image

PR #3556, opened 2022-07-22
General: Remove hosts filter on integrator plugins

iLLiCiTiT:
## Brief description
Removed hosts filters from integrator plugins.

## Description
Integrators accidentaly kept hosts filters that should not be there.

## Testing notes:
1. Integration should work in all hosts


antirotor (approved):
👍🏻

PR #3559, opened 2022-07-25
NewPublisher: Python 2 compatible html escape

iLLiCiTiT:
## Brief description
Escape of html syntax symbols is implemented in openpype instead of using Python 3 module.

## Description
We've used Python 3 module `html` for escaping of html symbols which is not available in Python 2 hosts. Thus we needed to add that option for them too. The basic escaping is easy to implement and does not need special python module for that so a function was added.

## Testing notes:
1. Open NewPublisher in host with Python 2
2. UI should show their labels and not crash

Resolves https://github.com/pypeclub/OpenPype/issues/3558

### DIFF
--- a/openpype/tools/publisher/publish_report_viewer/model.py
+++ b/openpype/tools/publisher/publish_report_viewer/model.py
@@ -1,9 +1,9 @@
 import uuid
-import html
 from Qt import QtCore, QtGui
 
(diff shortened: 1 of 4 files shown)


kalisp (approved):
Solves the issue in Maya

PR #3561, opened 2022-07-25
General: Global thumbnail extractor is ready for more cases

iLLiCiTiT:
## Brief description
It seems that only adding `traypublisher` to global thumbnail extractor is not enough. There are few issues related to the plugin which was not prepared to be used elsewhere then on farm.

## Description
Added or modified preflight checks. First check is if instance have representations and skip when not. Instance check for review and already existing thumbnail were moved earlier. Thumbnail representation has staging in temp rather then reuse source staging, that is because if input is `.jpg` it is just replaced so source is "replaced" (happend to my jpg sequence - not fun in production). Limited creation of thumbnail into 2 conditions and the creation of thumbnail is tried on all representations that can be used for that (if first crashes it will try create it from next). Added few logs.

## Notes
- We could maybe remove the cryptomate check? The plugin won't crash if conversion fails.
- We should probably remove the check for `instance.data.get("review", True)`. It's "deprecated" way how to tag instance for review for more then 2 years.
    - At the same time we should probably check if passed instance have `review` family in families rather. This may have massive impact on farm publishing so we should do it as totally separated PR.
- We should probably add more families to the plugin e.g. `image` family won't be processed by the plugin.

## Testing notes:
1. Publish `render` family through tray publisher
2. Should be integrated and thumbnail should be created

1. Publish something on farm
2. Thumbnails should be created as before

### DIFF
--- a/openpype/plugins/publish/extract_thumbnail.py
+++ b/openpype/plugins/publish/extract_thumbnail.py
@@ -41,9 +41,14 @@ class ExtractThumbnail(pyblish.api.InstancePlugin):
             "Processing instance with subset name {}".format(subset_name)
         )
 
-        # Skip if instance does not have review
+        # Skip if instance have 'review' key in data set to 'False'
         if not self._is_review_instance(instance):
             self.log.info("Skipping - no review set on instance.")
+            return
+
+        # Check if already has thumbnail created
+        if self._already_has_thumbnail(instance_repres):
+            self.log.info("Thumbnail representation already present.")
             return
 
         # skip crypto passes.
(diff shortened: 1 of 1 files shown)


m-u-r-p-h-y (changes requested):
there is an error during render publish

```
ERROR:pyblish.plugin:Traceback (most recent call last):
    runner(*args)
  File "D:\REPO\OpenPype\openpype\plugins\publish\extract_thumbnail.py", line 75, in process
AttributeError: 'str' object has no attribute 'formap'
    runner(*args)
  File "<string>", line 75, in process
AttributeError: 'str' object has no attribute 'formap'
```


image

m-u-r-p-h-y (approved):
- publishing from tray works as expected
- publishing from DCC to farm works as expected as well

PR #3562, opened 2022-07-25
Maya: Fix animated attributes (ie. overscan) on loaded cameras breaking review publishing.

### LINKED ISSUE #2807: Maya: Camera Overscan Attribute
**Describe the bug**
When animating camera attributes and publishing a camera alembic, then referencing this back into Maya, it'll connect all camera attributes to the alembic node. Regardless of whether the attribute were animated in the first place.

**To Reproduce**
1. Animated camera scale attribute.
2. Publish camera.
3. Load camera.
4. Publish review with the loaded camera.
5. Error on Playblast extraction because overscan attribute is connected to the alembic node.

**Expected behavior**
The overscan attribute is rarely (is ever) animated, so a solution would be to disconnect the overscan attribute when loading the camera.
Also might be useful to have a validator to check whether the attribute is connected or even locked, because else it'll just error the extraction.

Allan-I:
## Brief description
Fixes #2807 by checking for locked attributes and skipping them.

## Description
Issue #2807 stems from published cameras being unable to distinguish animated attributes on them when they're published as alembics, so when they're loaded in again, the attributes are automatically connected to the alembic nodes.  This solves that issue by testing if the attribute is locked and passes changing those attributes if they're connected.

## Additional info
It's mainly handled in `capture.py`, with the soft fail wrapper by @kalisp included in case Maya acts up. 

## Testing notes:
1. In Maya, animate attribute on a created camera shape.
2. Create camera instance.
3. Publish camera instance.
4. Load published camera in as alembic.
5. Create review instance from published camera.

Should run smoothly as opposed to breaking during the extract playblast stage.


m-u-r-p-h-y (approved):
working seamlessly

Maya2022.4

- animated focal length
- published and used for review
- everything went smoothly, playblast was created and published to ftrack

image

PR #3567, opened 2022-07-25
Nuke: Validation refactory to new publisher

jakubjezek001:
## Brief description
Validators are compatible with New publisher

## Description
- Validate Backdrop
- Validate Gizmo
- Validate Asset Name
- Validate Node's Knobs
- Validate Render Output Resolution
- Validate Proxy Mode
- Validate Rendered Frames
- Validate Script Attributes


openpype/hosts/nuke/plugins/publish/validate_script_attributes.py
```diff
@@ -0,0 +1,141 @@
...
+    actions = [RepairScriptAttributes]
+
+    def process(self, instance):
+        root = nuke.root()
+        knob_data = get_avalon_knob_data(root)
+        project_name = legacy_io.active_project()
+        asset = get_asset_by_name(
+            project_name,
+            instance.context.data["asset"]
+        )
```
iLLiCiTiT, line 46:
```suggestion
        asset = instance.data["assetEntity"]
```

iLLiCiTiT, line 46:
```suggestion
        asset = instance.data["assetEntity"]
```

openpype/hosts/nuke/api/lib.py
```diff
@@ -2224,15 +2208,14 @@ def get_write_node_template_attr(node):
     avalon_knob_data = read_avalon_data(node)
     # get template data
     nuke_imageio_writes = get_imageio_node_setting(
-        node_class=avalon_knob_data["family"],
+        node_class=avalon_knob_data["families"],
```
iLLiCiTiT, line 2211: Why is this `"families"`?

iLLiCiTiT (changes requested):
Few smaller change request.

openpype/hosts/nuke/api/lib.py
jakubjezek001: it is actually used later as families, so why not to call it families from begginging.. There are 3 different subsets with main family sharing one `Write` family in families. It is for simplification of some publishing plugins and also for creator settings.
image

jrsndl:
I tested in Nuke 12.2v3

- [x] Validate Asset Name: tested on render and gizmo. Repair on render worked, on gizmo the repair removed the gizmo from the script

- [x] Validate Gizmo: Correctly warns about more than one output

- [x] Validate Proxy Mode: Correctly detects proxy mode, repair works

- [x] Validate Script Attributes: Correctly detects and repairs fps, resolution, 

- [x] Validate Render Output Resolution: Correctly detects wrong reformat inside render group. Requires workfile to be published

- [x] Validate Backdrop: Correctly detects "empty" backdrop

- [ ] Validate Rendered Frames:
Seems to fail every time. With or without nuke slate in the script. Maybe the condition in ValidateRenderedFrames should just compare lengths? It looks like publish can ignore if artist renders more frames then necessary.
```
if (
    collected_frames_len != frame_length
):
```
Only available for "use existing frames" option.

- [ ] Validate Nuke Write Node: It fails on integer values:
```
Write node's knobs values are not correct!
Knob 'dw_compression_level' > Correct: `60` > Wrong: `60.0`
```

- [ ] Validate Node's Knobs: It didn't work for me with settings below. Validator didn't detect the review was off.
```
{
    "render": {
        "review": true
    }
}
```

jrsndl (dismissed):
Validate Rendered Frames Seems to fail every time.
Validate Nuke Write Node fails on integer values
Validate Node's Knobs: didn't detect the review was off.
```
{
    "render": {
        "review": true
    }
}
```

jakubjezek001:
> ```
> {
>     "render": {
>         "review": true
>     }
> }
> ```
Nice catch.. fixed now @jrsndl

iLLiCiTiT:
Already removed OcioConfig submodule is added back to repository.

PR #3569, opened 2022-07-26
Maya: fix Review image plane attribute 

### LINKED ISSUE #3272: Maya: Review Image plane attribute
**Describe the bug**

animation review instance attributes are not working while I was testing the new feature with multiple instances in #3269

**To Reproduce**
two review instances 
one with image plane on
other with image plane off

The resulting playblasts will have image plane on in both cases

image

[cuID:2eqzt4u]

moonyuet:
## Brief description
Fixing the bug regarding to OP-3356_Maya-Review-Image-plane-attribute

## Description
The main issue of this bug is review image plane still showing up as thumbnail in ftrack even when its related attribute (Image Plane) in Maya is turned off. 

## Solution
Adjust the code in _extract_thumbnail.py_ to tell the system not to create the thumbnail image in the project path of the ftrack

resolves #3272


mkolar (changes requested):
Hi Kayla. 

I think you might have misunderstood the issue at hand. The problem is that thumbnails always have imageplane visible in them, regardless whether it's turned on or off in on the maya set options. 

You are skipping thumbnail creation altogether when ImagePlane option is off, rather than hiding the image plane and still creating the thumbnail. Albeit I will admit that not generating a thumbnail doesn't result in generating a wrong one :) #technicallythetruth

To fix it, you just need to replicate this behaviour of the extract_playblast, which already deals with it correctly .https://github.com/pypeclub/OpenPype/blob/cfc11bb84027635850670f6c99ae67a89faf4b92/openpype/hosts/maya/plugins/publish/extract_playblast.py#L111-L116

mkolar (approved):
Perfect. Works as it should Thank you

PR #3571, opened 2022-07-26
General: Lib cleanup

iLLiCiTiT:
## Brief description
Small cleanup of code in `openpype.lib` which should not be moved now but deserve attention. Some unused code was removed and raising few questions related to deprecated deprecations.

## Description
Proposal of some `./openpype/lib/` functions cleanup. Removed git progress which is unused and contains only private class. Renamed `./openype/lib/config.py` to `./openpype/lib/dateutils.py` which we wanted to do since OpenPype 3 was released - kept `config` with deprecation warnings. Removed `timeit` function from log lib - unused function, at wrong place.

Remaining changes are related to custom logger logic which has strange overusage of old name `Pype`. Removed `Pype` from most of class names to make it less specific. Most important question is what to do with `Logger` and `PypeLogger`.

## Logger question
Right now in develop `./openpype/api.py` is `Logger` commented as deprecated which logically means that `PypeLogger` should be used instead. But it's `OpenPype` for a long time not `Pype`, so the question is how to handle this.
I see 4 options:
1. mark `PypeLogger` as deprecated and use `Logger` instead
2. mark `Logger` as deprecated and use `PypeLogger` instead
3. come up with some new name and throw even more confusion into the world
4. keep both and don't mark any as deprecated

I vote for 1.) but that is my personal opinion and would like to hear some other inputs, options and opinions. "Mark as deprecated" meants that the class will stay for some time in code base but when is used a deprecation warning log is thrown and will be removed after 2 or more releases.

## Testing notes:
1. Everything should work as did


mkolar (approved):
I vote for option 1. Other than that the seems good to me.

PR #3572, opened 2022-07-26
Enable write color sets on animation publish automatically

moonyuet:
## Brief description
This is the request from the Line. They want the "write color sets" to be on by default on the animation publish
See the ticket here: https://app.clickup.com/t/6658547/OP-3371
## Solution
Set` self.data["writeColorSets"] = True` in _create_animation.py_ in ./maya/plugins/publish folder

### DIFF
--- a/openpype/hosts/maya/plugins/create/create_animation.py
+++ b/openpype/hosts/maya/plugins/create/create_animation.py
@@ -22,7 +22,7 @@ class CreateAnimation(plugin.Creator):
             self.data[key] = value
 
         # Write vertex colors with the geometry.
-        self.data["writeColorSets"] = False
+        self.data["writeColorSets"] = True
         self.data["writeFaceSets"] = False
 
         # Include only renderable visible shapes.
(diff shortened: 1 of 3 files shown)


m-u-r-p-h-y:
I'm not sure if this should be on by default, as this will affect many studios out there. Should be configurable and let them turn it on in their environment . . . and keep it off by default to keep some continuity in settings.

moonyuet:
yes I agree. I leave the solution into the clickup first. If it is possible, just send them a script to replace the original one or just ask them to set self.data["writeColorSets"]  = True by their own.

mkolar (changes requested):
This definitely needs to be added to the settings, with the default staying as it was

moonyuet:
I have added switching on/off checkbox in Create Animation slot in the Openpype setting 
image

m-u-r-p-h-y (commented):
shouldn't we approach Point Cache instance the same way as well?

It would be more consistent imo.

image

moonyuet:
i updated the new one with the correct branch in https://github.com/pypeclub/OpenPype/pull/3582

PR #3576, opened 2022-07-26
Nuke: render family integration consistency 

jakubjezek001:
## Brief description
Families are integrating to the same family from any render target.

## Description
Any chosen render target should now produce the same result in published folder.

## Additional info
Integrator is using `anatomyData` keys for filling up templates and choosing what template to use from Anatomy. So family has to be correct.

## Testing notes:
1. Nuke render local, farm, existing frames from render and prerender node
2. published versions should be the same family

closes https://github.com/pypeclub/OpenPype/issues/3568


jrsndl:
Farm publish doesn't make an anatomy {output} for thumbnail, local publish does.
image

jrsndl (approved):
Family consistency is fixed.
I tested farm and local publish, with clean and "dirty" work folder.

PR #3582, opened 2022-07-27
Enable write color sets on animation publish automatically

moonyuet:
## Brief description
Add the switching on/off option for write color sets in Create Cache in Openpype settings
image

## Description
Add the switching on/off option for write color sets in Create Cache in Openpype settings, which overrides the write color sets attributes in maya when creating point cache instance for the geometry. 
It is implemented for being consistent regarding to the same implementation of switching on/off options in Create Animation for maya.

### DIFF
--- a/openpype/hosts/maya/plugins/create/create_animation.py
+++ b/openpype/hosts/maya/plugins/create/create_animation.py
@@ -22,7 +22,7 @@ class CreateAnimation(plugin.Creator):
             self.data[key] = value
 
         # Write vertex colors with the geometry.
-        self.data["writeColorSets"] = False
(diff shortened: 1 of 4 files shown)


m-u-r-p-h-y (approved):
It works as expected. Animation and Point Cache instances are created with proper values set from settings.

I would suggest unifying the naming in settings though. 

Cache -> Point Cache

image

mkolar:
@moonyuet I agree with Murphy. all good, just please fix the label and the hound comments.

moonyuet:
image
Get the name changed!

m-u-r-p-h-y (approved):
Nice! Ready to go . . .

PR #3586, opened 2022-07-27
Fix general settings environment variables resolution

### LINKED ISSUE #3585: Environment variables config bug on Windows
**Describe the bug**
On Windows, you can't put an environment variable in the PYTHONPATH set in OpenPype settings, unlike on Linux.

**To Reproduce**
OpenPype crashes at the start if you have something like this in the settings:
```
{
    "STUDIO_SOFTWARE": "/prod/softprod/apps",
    "PYTHONPATH": [
        "{PYTHONPATH}",
        "{STUDIO_SOFTWARE}"
    ]
}
```
You need to have something like that in order to work on Windows:
```
{
    "STUDIO_SOFTWARE": "/prod/softprod/apps",
    "PYTHONPATH": [
        "{PYTHONPATH}",
        "/prod/softprod/apps"
    ]
}
```

**Screenshots**
This is the error we got:
image

**Expected behavior**
The first pattern should work on Windows as it works on Linux.

 - OS: Windows

[cuID:2nbq137]

antirotor:
# Fix

Fix how variables are resolved in general settings. This will now allow to define new variables and resolve them in one pass.

For example:

```
{
    "FOO": "something",
    "BAR": ["{FOO}", "else"]
}
```
will now resolve correctly. This was crashing OpenPype on start.

Resolves #3585


antirotor:
It is coming from different base, I'll re-create it from the right one

PR #3590, opened 2022-07-28
Bugfix: Add OCIO as submodule to prepare for handling `maketx` color space conversion.

### LINKED ISSUE #3584: maketx built with OCIO support requires config
## Bug

`maketx` tool as a part of the OpenImageIO frameworks needs OCIO config when OIIO is build with OCIO support. It doesn't know even the simplest color profiles. This makes look workflow broken in Maya if OCIO isn't enabled.

## Suggested quick fix:

- Add [OpenColorIO-Configs](https://github.com/imageworks/OpenColorIO-Configs) to `/vendor` as git submodule
- Add argument to `maketx` in Extract Look to add `nuke-default` as its OCIO config

## More robust solution:

Implement proper color management in OpenPype :)

from `maketx` help
```
Color Management Options (OpenColorIO enabled)
    --colorconfig FILENAME   Explicitly specify an OCIO configuration file
    --colorconvert IN OUT    Apply a color space conversion to the image. If the output color space is not the same
                               bit depth as input color space, it is your responsibility to set the data format to
                               the proper bit depth using the -d option.  (choices: NONE)
    --unpremult              Unpremultiply before color conversion, then premultiply after the color conversion.
                               You'll probably want to use this flag if your image contains an alpha channel.
```

[cuID:OP-3657]

Allan-I:
## Brief description
`maketx` executable requires an `OCIO` config file path appended when using the `--colorconfig` flag.  This resolves #3584 and prepares for preventing `maketx` from crashing when publishing OpenPype `Look` instances.

## Description
Adding OCIO as a submodule and adding functionality to grab the OCIO config file path based on a profile folder parameter as the function argument.

Testing: 
1. Create some geo and OP model instance.
2. Publish it.
3. load model instance in.
4. Create shader, apply a test base color texture (that's in sRGB).
5. publish it. 

The publishing should produce a `.tx` file that's optimized for `MtoA`. (Make sure create .tx is enabled in the look creator)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,3 +5,6 @@
 [submodule "tools/modules/powershell/PSWriteColor"]
     path = tools/modules/powershell/PSWriteColor
     url = https://github.com/EvotecIT/PSWriteColor.git
+[submodule "vendor/configs/OpenColorIO-Configs"]
+	path = vendor/configs/OpenColorIO-Configs
+	url = https://github.com/imageworks/OpenColorIO-Configs
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,6 +5,3 @@
 [submodule "tools/modules/powershell/PSWriteColor"]
     path = tools/modules/powershell/PSWriteColor
     url = https://github.com/EvotecIT/PSWriteColor.git
-[submodule "vendor/configs/OpenColorIO-Configs"]
-	path = vendor/configs/OpenColorIO-Configs
-	url = https://github.com/imageworks/OpenColorIO-Configs
--- a/openpype/hosts/maya/plugins/publish/extract_look.py
+++ b/openpype/hosts/maya/plugins/publish/extract_look.py
@@ -514,6 +514,9 @@ class ExtractLook(openpype.api.Extractor):
                 colorconvert = "--colorconvert sRGB linear"
             else:
                 colorconvert = ""
+
+            config_path = get_ocio_config_path("nuke-default")
+            color_config = "--colorconfig {0}".format(config_path)
             # Ensure folder exists
             if not os.path.exists(os.path.dirname(converted)):
                 os.makedirs(os.path.dirname(converted))
@@ -523,10 +526,11 @@ class ExtractLook(openpype.api.Extractor):
                 filepath,
                 converted,
                 # Include `source-hash` as string metadata
-                "-sattrib",
+                "--sattrib",
                 "sourceHash",
                 escape_space(texture_hash),
                 colorconvert,
+                color_config
             )
 
             return converted, COPY, texture_hash
--- a/openpype/hosts/maya/plugins/publish/extract_look.py
+++ b/openpype/hosts/maya/plugins/publish/extract_look.py
@@ -27,6 +27,28 @@ def escape_space(path):
     return '"{}"'.format(path) if " " in path else path
 
 
+def get_ocio_config_path(profile_folder):
+    """Path to OpenPype vendorized OCIO.
+
+    Vendorized OCIO config file path is grabbed from the specific path
+    hierarchy specified below.
+
+    "{OPENPYPE_ROOT}/vendor/OpenColorIO-Configs/{profile_folder}/config.ocio"
+    Args:
+        profile_folder (str): Name of folder to grab config file from.
+
+    Returns:
+        str: Path to vendorized config file.
+    """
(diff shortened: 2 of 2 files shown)


.gitmodules
```diff
@@ -5,3 +5,6 @@
 [submodule "tools/modules/powershell/PSWriteColor"]
     path = tools/modules/powershell/PSWriteColor
     url = https://github.com/EvotecIT/PSWriteColor.git
+[submodule "vendor/configs/OpenColorIO-Configs"]
```
jakubjezek001, line 8: @antirotor shouldn't we do this rather with poetry?

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -27,6 +27,29 @@ def escape_space(path):
     return '"{}"'.format(path) if " " in path else path
 
 
+def get_ocio_config_path(profile_folder):
```
jakubjezek001, line 30: @iLLiCiTiT I wonder if this should not be somewhere in global scope?

jakubjezek001 (changes requested):

.gitmodules
jakubjezek001: @Allan-I look to `pyproject.toml` to line 34 

> acre = { git = "https://github.com/pypeclub/acre.git" }

openpype/hosts/maya/plugins/publish/extract_look.py
jakubjezek001: I would create new `openpype.lib` module `colorspace.py` and add it there.

m-u-r-p-h-y (approved):
tested per instructions

publishing went smoothly, with tx file generated 

```
// pyblish.ExtractLook : tx: converting sRGB -> linear // 
// pyblish.ExtractLook : Generating .tx file for D:\PROJECTS\murphy_22101\assets\props\testProp2\work\lookdev\macbeth_srgb.tif .. // 
// pyblish.ExtractLook : file will be copied C:\Users\murph\AppData\Local\Temp\pyblish_tmp_ozlxpjrw\resources\macbeth_srgb.tx -> D:\PROJECTS\murphy_22101\assets\props\testProp2\publish\look\lookMain\v001\resources\macbeth_srgb.tx // 
// 
```

.gitmodules
Allan-I: Thank you very much Jeza, I'm on it!

openpype/hosts/maya/plugins/publish/extract_look.py
iLLiCiTiT: We've discussed this on discord and conclusion was that it is right now used only here and is used for very specific usecase (fix maketx bug). Handling ocio configs properly would require much more logic and having it in `openpype.lib` would maybe cause illusion of us handling the colorspace and using them which is not true.

.gitmodules
Allan-I: Due to the package being in a format that poetry has problems with, and due to it giving me issues as I was grabbing it I've opted to keep including it as a git submodule unless further discussion arises, right now using poetry (according to what I understand from discussions online) might present future issues for that repository.

openpype/hosts/maya/plugins/publish/extract_look.py
Allan-I: I think we could just move it elsewhere if we need to, soon as we decide to tackle the color management.

jakubjezek001 (approved):

.gitmodules
jakubjezek001: I am only very afraid to do it as submodule as it is quite difficult to remove it lately..

Allan-I: I absolutely agree with you on this also, it was a nightmare trying to remove it in the initial testing to see how it goes.  Also, it's quite a heavy repo actually.

BigRoy:
Just want to mention here that this PR adds a 2.3 GB file size to OpenPype builds and installs as [I mentioned on discord](https://discord.com/channels/517362899170230292/563751989075378201/1006889129755295794). The code currently only uses the 2.3 MB `nuke-default` folder so it might be the easiest quick fix for now to only include that somehow.

antirotor:
we've decided to have our own forked and curated set that should be used instead, not just for that particular issue but also for proper support of color management.

PR #3591, opened 2022-07-28
Photoshop: implemented {layer} placeholder in subset template

kalisp:
## Brief description
Previously when layer name is added to subset name (in case of multiple instances created in one go), group (layer) name
was concatenated (capitalized). Customer wants better control.

## Description
{layer} placeholder (and all its variants {Layer}, {LAYER}) could be added to `project_settings/global/tools/creator/subset_name_profiles` to control how layer name should look like.

## Additional info
{layer} means keep layer name as it was, not lowercased!
Be careful with delimiters as '_{layer}' as for single subset it would leave trailing underline.

layer_subset

## Testing notes:
1. Configure Setting accordingly to attached screenshot (`{Variant}{layer}` would be actually preferred)
2. create instances from multiple selected layers (in legacy and New Publisher)


64qam (approved):

openpype/hosts/photoshop/plugins/create/create_image.py
```diff
@@ -143,3 +152,9 @@ def _handle_legacy(self, instance_data):
     def _clean_highlights(self, stub, item):
         return item.replace(stub.PUBLISH_ICON, '').replace(stub.LOADED_ICON,
                                                            '')
+    @classmethod
+    def get_dynamic_data(
+        cls, variant, task_name, asset_id, project_name, host_name
+    ):
+        """Called by UI, empty value for layer must be provided."""
+        return {"layer": ""}
```
iLLiCiTiT, line 160: Maybe simple variant would be to return formattings placeholder and fill it in create?
```
def get_dynamic_data(cls, *args, **kwargs):
    return prepare_template_data({"layer": "{layer}"})
```

And in create:
```
group_fill = prepare_template_data({"layer": group.name})
subset_name = subset_name.format(**group_fill)
```

You don't need to duplicate code with `get_subset_template` and `get_subset_name_for_multiple` and artist will see `{layer}` insubset name before create.

kalisp: This wouldn't work if layer name is not used in subset name (which might not be all the time).

iLLiCiTiT: It would just not fill it. In that case the subset name would stay unchanged which is not an issue.
```
layer_name = "something"
subset_name = "renderMain"
layer_fill = prepare_template_data({"layer": layer_name})
# This would just won't do anything
subset_name = subset_name.format(**layer_fill)
print(subset_name)
```

PR #3594, opened 2022-07-28
General: Add context to pyblish context

iLLiCiTiT:
## Brief description
Store project name, asset name and task name to pyblish context so it's not needed to use `legacy_io` during publishing to get it. Goal is to reduce being dependent on `legacy_io.Session` and `os.environ` and their specific keys.

## Description
Added global collector (-0.5 order) which adds to `context.data` for publishing `"projectName"`, `"asset"` and `"task"` from `legacy_io` if they're not filled. Changed few usages of `AVALON_PROJECT` in other plugins to use already available data from publish context.

## Additional information
Not sure when context is collected on farm? It looks like it happens in `CollectRenderedFiles` with -0.2 order which is too late and the plugin should probably be split into 2 (or more) plugins or move it earlier with all dependent plugins to it.

Strange thing is that `CollectRenderedFiles` can expected multiple paths in `"OPENPYPE_PUBLISH_DATA"` environment variable but it probably should not be allowed and strictly require single path or validate contexts of content.

## Testing notes:
1. Render publishing should work
2. Standard publishing should work


kalisp (approved):
It works ( in AE - local or farm, PS, Nuke)

PR #3596, opened 2022-07-29
General: Use query functions in general code

iLLiCiTiT:
## Brief description
Use query functions in general code of openpype.

## Description
Replace direct mongo calls with query functions in general places or forgotten places.

## Additional information
There are 2 duplicated fusion scripts and both seems to be unused.

## Testing notes:
1. Applications should launch
2. Thumbnails should be visible in loader (as they were)
3. Tray publisher can work


kalisp (approved):
Seems to work.
TrayPublisher works only with https://github.com/pypeclub/OpenPype/pull/3600

PR #3601, opened 2022-07-29
General: Naive implementation of document create, update, delete

iLLiCiTiT:
## Brief description
Simple variant of creating, updating and deleting of entities can be done without direct access to mongo documents.

## Description
Created basic operations for create, update and delete. All of them require project name and entity type, update and delete require entity id, creation requires full document and update requires changes of document.

There is almost none validation of documents even not on creation. That can be achieved by moving `schema.py` out of `openpype.pipeline` to top level. We should also consider that json schemas would be at the same folder.

## Additional info
There are missing docstrings and I didni't fully validated functionality, yet. Class and function/method names are not final. Any comments welcomed. There is no way to "replace or insert" operation. Not sure if we should allow that option/s.

## Possible enhancements
- schema validation of documents to create
- maybe also validate if created document is not duplicate of already existing document (by entity specific queries) before commit

## Testing notes:
At this moment are operations used only in new integrator.
1. Run publishing
2. Run publishing which will integrate to already existing version
3. Farm publishing which appends representations

### DIFF
--- a/openpype/client/operations.py
+++ b/openpype/client/operations.py
@@ -1,0 +1,249 @@
+import uuid
+import copy
+from abc import ABCMeta, abstractmethod
+
+import six
+from bson.objectid import ObjectId
+from pymongo import DeleteOne, InsertOne, UpdateOne
+
+from .mongo import get_project_connection
+
+REMOVED_VALUE = object()
+
+
+@six.add_metaclass(ABCMeta)
+class AbstractOperation(object):
+    """Base operation class."""
+
+    def __init__(self, entity_type):
+        self._entity_type = entity_type
+        self._id = uuid.uuid4()
+
+    @property
+    def id(self):
+        return self._id
+
+    @property
+    def entity_type(self):
+        return self._entity_type
+
+    @abstractmethod
+    def to_mongo_operation(self):
+        pass
+
+
+class CreateOperation(AbstractOperation):
+    def __init__(self, project_name, entity_type, data):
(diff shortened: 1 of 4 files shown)


openpype/client/operations.py
```diff
@@ -0,0 +1,587 @@
...
+    return {
+        "_id": _create_or_convert_to_mongo_id(entity_id),
+        "name": project_name,
+        "type": CURRENT_PROJECT_SCHEMA,
+        "entity_data": data,
+        "config": config
+    }
+
+
+def new_asset_document(
```
kalisp, line 63: Not particularly fond of this signature, `parents` mostly. How would I get those? Wouldn't 'parent_id' be enough for method to handle it for me?

kalisp (approved):
Tested in AE, PS, Nuke, compared new representation, version, subset documents with old ones, visually looked ok.

openpype/client/operations.py
iLLiCiTiT: > Wouldn't 'parent_id' be enough for method to handle it for me?

I agree with this statement but we don't have an easy way or a server which would care about it. We have to care about it manually on a client side. It would require to query all the parents by visualParent ids. First of all that is waste of time (a ton of queries if a full project is created) and secondly they may not exists yet in database (if full project is created at once). At all places where an asset is created, we know it's parents.

It can be changed or improved in future, this is just wrapping what we already have. In v4 it will be calculated on server.

Issue #3534, opened 2022-07-19
Maya: Look Data Contents fails with custom attribute on group


m-u-r-p-h-y:
**Describe the bug**
When a group which is a member of look instance contains custom attributes `Look Data Contents` validator will fail

![image](https://user-images.githubusercontent.com/352795/179717992-94ba6669-c027-44f9-aa2e-e512491ca308.png)

![image](https://user-images.githubusercontent.com/352795/179718315-9a1b6898-3cb4-496e-bf3c-32449fde3b99.png)

**Screenshots**
![image](https://user-images.githubusercontent.com/352795/179717902-9c362364-928d-4bb1-886c-93fb33bf489b.png)

**Additional context**
It seems this attribute is coming from FBX import

[cuID:OP-3607]

m-u-r-p-h-y:
I can replicate the fail on latest develop.

just select the group and add the custom attribute with MEL:

`addAttr -s true -ci true -k true -sn "MaxHandle" -ln "MaxHandle" -smn 0 -smx 0;`

![image](https://user-images.githubusercontent.com/352795/179764338-d9d21729-f12b-4b17-8995-727d6b4b34ad.png)

m-u-r-p-h-y:
resolved in #3607

PR #3607, opened 2022-08-02
Maya look data contents fails with custom attribute on group

moonyuet:
## Brief description
Solving the failure of validating look data contents when custom attributes are set up on group

## Solution
Set up the condition to tell the relative validating scripts to exclude the nodes with custom attributes from the invalid list, when the system is checking if the attributes are on nodes with IDs.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/collect_look.py
+++ b/openpype/hosts/maya/plugins/publish/collect_look.py
@@ -551,11 +551,9 @@ class CollectLook(pyblish.api.InstancePlugin):
                 if cmds.getAttr(attribute, type=True) == "message":
                     continue
                 node_attributes[attr] = cmds.getAttr(attribute)
-                
             # Only include if there are any properties we care about
             if not node_attributes:
                 continue
-            
             attributes.append({"name": node,
                                "uuid": lib.get_id(node),
                                "attributes": node_attributes})
--- a/openpype/hosts/maya/plugins/publish/collect_look.py
+++ b/openpype/hosts/maya/plugins/publish/collect_look.py
@@ -551,7 +551,11 @@ class CollectLook(pyblish.api.InstancePlugin):
                 if cmds.getAttr(attribute, type=True) == "message":
                     continue
                 node_attributes[attr] = cmds.getAttr(attribute)
-
+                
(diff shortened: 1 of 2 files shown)


BigRoy:
This seems like an odd fix. 

Shouldn't all entries in `lookdata["attributes"]` have attribute changes in `attr_changes["attributes"]` to begin with?

moonyuet:
Hi Roy, I have checked the function of collect_attributes_changed from _collect_look.py_ particularly for this issue. I saw this below:
`            attributes.append({"name": node,
                               "uuid": lib.get_id(node),
                               "attributes": node_attributes})
`
If I understand correctly, the for-loop condition in _validate_look_data_contents.py_ shows that `attr_changes["name"]`, `attr_changes["uuid"]` and `attr_changes["attributes"]` are the elements iterating through `lookdata["attributes"]`.

BigRoy:
Ah yes, the root of the issue does appear to be there! I originally assumed cases without actual values would be ignored as [this code tries to do](https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/maya/plugins/publish/collect_look.py#L542-L544). However, the [code just below skips some additional cases of attributes](https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/maya/plugins/publish/collect_look.py#L546-L553) with the `continue` statement.

As such, it's probably better to also exclude that node completely when `not node_attributes`.

It might thus make most sense to change the collecting to ignore the cases correctly [around line 554](https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/maya/plugins/publish/collect_look.py#L554):
```python
            # Only include if there are any properties we care about
            if not node_attributes:
                continue
```

What do you think?

moonyuet:
I agree. I think it'd be logically concise to implement your suggested conditions to entirely exclude the node when it isn't with `node_attributes`. 
Will add your suggested code later in `collect_look.py`

m-u-r-p-h-y:
linking with issue #3534

m-u-r-p-h-y (approved):
now it passes the validation with custom attributes on the group

image


so the issue is resolved, but not sure if all doubts were resolved as well . .

PR #3610, opened 2022-08-03
AfterEffects: refactored integrate doesnt work formulti frame publishes

kalisp:
## Brief description
Frame pattern used in new integrate is not catching frames pattern in AE

## Description
Locally rendered files are renamed into frame pattern that integrator plugin understands.

## Additional info
~~(Added Kuba and Ondrej as reviewers to only highlight this.)
Added `render.farm` for AE into Settings to skip integrating. New publisher should have 'render' as a primary family (and 'render.farm` in families), which would trigger integrating (legacy integrator has 'render.farm' in `exclude_families` , this functionality is currently missing in refactored integrator).~~

~~I think we should keep `skip_host_families` not as a temporary solution to broken publishes for new integrator, but use it instead of 'exclude_families' which might be always useful.~~

## Testing notes:
1. Publish locally in AE
2. Publish on farm in AE

### DIFF
--- a/openpype/plugins/publish/integrate.py
+++ b/openpype/plugins/publish/integrate.py
@@ -23,41 +23,6 @@ from openpype.pipeline.publish import KnownPublishError
 log = logging.getLogger(__name__)
 
 
-def assemble(files):
-    """Convenience `clique.assemble` wrapper for files of a single collection.
-
-    Unlike `clique.assemble` this wrapper does not allow more than a single
-    Collection nor any remainder files. Errors will be raised when not only
-    a single collection is assembled.
-
-    Returns:
-        clique.Collection: A single sequence Collection
-
-    Raises:
-        ValueError: Error is raised when files do not result in a single
-                    collected Collection.
-
-    """
-    # todo: move this to lib?
-    # Get the sequence as a collection. The files must be of a single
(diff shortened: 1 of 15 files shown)


openpype/plugins/publish/integrate.py
```diff
@@ -167,7 +167,7 @@ class IntegrateAsset(pyblish.api.InstancePlugin):
     skip_host_families = []
 
     def process(self, instance):
-        if self._temp_skip_instance_by_settings(instance):
+        if self.skip_instance_by_settings(instance):
```
iLLiCiTiT, line 170: why the rename?

iLLiCiTiT:
> I think we should keep skip_host_families not as a temporary solution to broken publishes for new integrator, but use it instead of 'exclude_families' which might be always useful.

That was removed on purpose and was replaced by having `instance.data["farm"] = True` (you've added it...). If it should be added back then it should not be in settings but handled on purpose. It was removed to reduce responsibility of the integrator.

iLLiCiTiT:
With this PR https://github.com/pypeclub/OpenPype/pull/3611 this PR is not needed

64qam (approved):

kalisp:
This PR is still useful for new testing class and for farm publishes.

PR #3611, opened 2022-08-03
Integrator: Don't force to have dot before frame

iLLiCiTiT:
## Brief description
Creation of collection in new integrator does not require to have frame starting with dot.

## Description
It is not forced to use determined frame pattern in clique assemble. It seems to be added to support single file sequence which is not supported across multiple plugins so it's pointless and breaks a lot of current extractors output (or publish templates). Output sequence is created using anatomy format for each frame/udim.

## Testing notes:
It was discovered in TVPaint but it is also issue for AfterEffects.

Resolves https://github.com/pypeclub/OpenPype/issues/3608


kalisp (approved):
Works for AE.

PR #3614, opened 2022-08-03
Core: translated validate_containers.py into New publisher style

kalisp:
## Brief description
`Validate Containers` could now be used in New Publisher as a optional plugin.
First sentence is brief description.

## Additional info
AE could be already using NP.

## Testing notes:
1. Try to publish in AE with New Publisher in the scene with obsolete loaded containers.

### DIFF
--- a/openpype/plugins/publish/help/validate_containers.xml
+++ b/openpype/plugins/publish/help/validate_containers.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<root>
+<error id="main">
+<title>Not up-to-date assets</title>
+<description>
+## Obsolete containers found
+
+Scene contains one or more obsolete loaded containers, eg. items loaded into scene by Loader.
+
+### How to repair?
+
+Use 'Scene Inventory' and update all highlighted old container to latest OR
+    refresh Publish and switch 'Validate Containers' toggle on 'Options' tab.
+
+    WARNING: Skipping this validator will result in publishing (and probably rendering) old version of loaded assets.
+</description>
+<detail>
+### __Detailed Info__ (optional)
+
+This validator protects you from rendering obsolete content, someone modified some referenced asset in this scene, eg.
+    by skipping this you would ignore changes to that asset.
+</detail>
+</error>
+</root>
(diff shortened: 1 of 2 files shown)


64qam (approved):

openpype/plugins/publish/help/validate_containers.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<root>
+<error id="main">
+<title>Not up-to-date assets</title>
+<description>
+## Obsolete containers found
+
+Scene contains one or more obsolete loaded containers, eg. items loaded into scene by Loader.
```
BigRoy, line 8: I wonder whether "obsolete" is the right word to use here? It sounds like 'don't use the subset anymore please' instead of being "outdated" or "old". Thoughts?

openpype/plugins/publish/help/validate_containers.xml
```diff
@@ -0,0 +1,24 @@
...
+Use 'Scene Inventory' and update all highlighted old container to latest OR
+    refresh Publish and switch 'Validate Containers' toggle on 'Options' tab.
+
+    WARNING: Skipping this validator will result in publishing (and probably rendering) old version of loaded assets.
+</description>
+<detail>
+### __Detailed Info__ (optional)
+
+This validator protects you from rendering obsolete content, someone modified some referenced asset in this scene, eg.
+    by skipping this you would ignore changes to that asset.
```
BigRoy, line 21: > someone modified some referenced asset in this scene, eg.
>    by skipping this you would ignore changes to that asset.

I don't understand that part of the detailed info. It sounds like "modifications to the referenced asset IN THIS SCENE." but instead it should be about a newer version of the subset being available.

So basically:

- This validates whether you're working with the latest versions of published content loaded into your scene. This protects you from using outdated versions of an asset.

Or something along those lines.

openpype/plugins/publish/validate_containers.py
```diff
@@ -24,5 +30,9 @@ class ValidateContainers(pyblish.api.ContextPlugin):
     actions = [ShowInventory]
 
     def process(self, context):
+        if not self.is_active(context.data):
+            return
```
BigRoy, line 34: Shouldn't this be Pyblish's responsibility? Or why do we need to do this manually?

PR #3615, opened 2022-08-03
Ftrack: Comment template can contain optional keys

iLLiCiTiT:
## Brief description
Reduced empty notes and descriptions during publishing when integrating to ftrack, intent can be nullable and `StringTemplate` support to keep less/greater symbols if don't contain any formatting `<...>` (`<br/>` is kept unchanged).

## Description
The main issue is that templates in most of cases consist from `{intent}: {comment}` which cause that result is `{-: }`. One reason is that intent can't be empty at this moment so it's filled with other symbol (e.g. `"-"`) and we're unable to know which part should be skipped if value is missing.

Intent can have nullable option so studio can tell that empty intent is possible and we can handle it in some way during publishing.

`StringTemplate` support to keep less/greater symbols if they do not contain any formatting. That is important for ftrack integrate note where html tags are used. If anybody would enter to template `"<<b>{asset}</b>><><br/>"` result after formatting would be `"<b>Bob</b><><br/>"` with `asset = "Bob"` and `"<><br/>"` if `asset` is not available. This is not a perfect solution for cases when someone would like to keep the symbols even between formatting but for current issue it's enough.

Integrate description skip integration if `comment` and `intent` are empty. Also template support optional keys. Integrate note can't skip if `comment` and `intent` are not available because are not only keys that can be used to fill the template but also support optional keys (was reason why `StringTemplate` needed modification).

## Additional info
This does not handle all cases because we can use optional keys only "one way". So for combination of `comment` and `intent` when template looks like `"{intent}: {comment}"` there are 2 possibilities how to enter optional keys `"<{intent}: >{comment}"` or `"{intent}<: {comment}>"` but we can't support both at a single time. I think that it would be possible with some "magical expression logic" (First letters make MEL, coinsidence?) in template `"<{intent}>{%: ?intent && comment%}<{comment}>"`.

## Testing notes:
1. Integrate description should skip if intent and comment are empty
2. It should be possible to use optional keys in both integrate description and integrate note

### DIFF
--- a/openpype/lib/path_templates.py
+++ b/openpype/lib/path_templates.py
@@ -211,15 +211,28 @@ class StringTemplate(object):
                 if counted_symb > -1:
                     parts = tmp_parts.pop(counted_symb)
                     counted_symb -= 1
+                    # If part contains only single string keep value
+                    #   unchanged
                     if parts:
                         # Remove optional start char
                         parts.pop(0)
-                        if counted_symb < 0:
-                            out_parts = new_parts
-                        else:
-                            out_parts = tmp_parts[counted_symb]
-                        # Store temp parts
-                        out_parts.append(OptionalPart(parts))
+
+                    if not parts:
+                        value = "<>"
+                    elif (
+                        len(parts) == 1
+                        and isinstance(parts[0], six.string_types)
+                    ):
+                        value = "<{}>".format(parts[0])
+                    else:
+                        value = OptionalPart(parts)
+
+                    if counted_symb < 0:
+                        out_parts = new_parts
+                    else:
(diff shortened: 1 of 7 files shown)


kalisp:
Could we add mention of optional keys to documentation (probably to https://openpype.io/docs/admin_settings_project_anatomy/)? I found it only for legacy documentation for v2 and was under impression it is `()` and not `<>`.

64qam (approved):

iLLiCiTiT:
> Could we add mention of optional keys to documentation

Made [PR](https://github.com/pypeclub/OpenPype/pull/3618).

PR #3617, opened 2022-08-04
Maya: Fix types of default settings

iLLiCiTiT:
## Brief description
Default settings for maya RenderSettings use list instead of dictionary which is defined in settings schemas.

## Description
Resaved default settings using settings UI which reorganized formatting of few lines in default settings and fixed wrong types.

## Testing notes:
I hope nothing should change.


antirotor (approved):
Looks good, but beware there are changes in `traypublisher.json` too, not related to Maya. Just formality though.

Allan-I (approved):
Looks good to me!

PR #3620, opened 2022-08-04
General: Extract review aspect ratio scale is calculated by ffmpeg

iLLiCiTiT:
## Brief description
Ratio scaling calculations happens in ffmpeg instead of our python code.

## Description
It happened that `1080` after python float calculations becomes `1081` which cause a lot of troubles. The main reason why we did that was to keep aspect ratio but ffmpeg scale has `force_original_aspect_ratio` option which keep ration by decreasin or increasing (in our case decreasing) and calculate padding difference in ffmpeg too.

## Additional info
This is huge change and should be tested properly. Changes affect cases when scale is needed and when input and output have different aspect ratio (scale is needed).

## Testing notes:
Force extract review to output a video which has different aspect ratio (visible with an eye) then converted input. The output should have right resolution and should be padded the right way.

### DIFF
--- a/openpype/plugins/publish/extract_review.py
+++ b/openpype/plugins/publish/extract_review.py
@@ -1390,9 +1390,11 @@ class ExtractReview(pyblish.api.InstancePlugin):
             self.log.debug("height_half_pad: `{}`".format(height_half_pad))
 
             filters.extend([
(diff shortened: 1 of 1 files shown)


antirotor (approved):
Tested in Maya and it works.

PR #3623, opened 2022-08-05
General: Plugin settings handled by plugins

iLLiCiTiT:
## Brief description
Loader and creator settings are handled by plugins.

## Description
`LoaderPlugin` and `LegacyCreator` have new classmethod `apply_settings` which by default contains logic from `openpype.lib.set_plugin_attributes_from_settings`. 

Function `filter_pyblish_plugins` which apply settings for pyblish plugins was moved to `openpype.pipeline.publish.lib`. For pyblish plugins was added ability to implement classmethod `def apply_settings(cls, project_settings, system_settings)` too. Can customize how settings are applied on a plugin. This is optional. If method is not implemented the "automated" approach is used as before - this option was added to be able handle some cases when path location of plugin does not exactly match settings path. Functions in lib were marked as deprecated.

## Testing notes:
1. Settings on loader plugins are applied
2. Settings on creator plugins are applied
3. Settings on publish plugins are applied
E.g. in Maya, Nuke, ...

### DIFF
--- a/openpype/lib/plugin_tools.py
+++ b/openpype/lib/plugin_tools.py
@@ -57,7 +57,7 @@ def deprecated(new_destination):
                 stacklevel=4
             )
             return decorated_func(*args, **kwargs)
-        return wrapper-
+        return wrapper
 
     if func is None:
         return _decorator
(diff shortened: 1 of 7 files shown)


64qam (approved):
Tested in Nuke, AE, PS and all looked ok

PR #3624, opened 2022-08-05
Nuke: support environment variables in gizmo menu

friquette:
## Brief description
A little update to accept environment variables in the command part of the gizmo menu.

## Description
We can now use environment variables in the python command part of the gizmo menu definition

image

### DIFF
--- a/openpype/hosts/nuke/api/gizmo_menu.py
+++ b/openpype/hosts/nuke/api/gizmo_menu.py
@@ -56,7 +56,7 @@ class GizmoMenu():
                 if item_type == ("python" or "file"):
                     parent.addCommand(
                         item["title"],
-                        command=str(item["command"]),
+                        command=str(item["command"].format(**os.environ)),
                         icon=item.get("icon"),
                         shortcut=item.get("hotkey")
                     )


BigRoy:
Note: This could be a potential security hazard if one was to set the environment value to e.g. close the brackets in the actual command and inject their own code to execute there too.

This will also generate errors on running the software when the environment variable is set even before running the command itself due to the environment itself being set.

What's holding you from changing the actual python script to be more explicit and have the code be:
```python
import os
nuke.nodePaste(os.environ["QUAD_PLUGIN_PATH"] + "/host/nuke/scripts/flockOfBirds.nk")
```
And if multi-line is not supported maybe doing: `import os; nuke.nodePaste(os.environ["QUAD_PLUGIN_PATH"] + "/host/nuke/scripts/flockOfBirds.nk")`?

friquette:
> Note: This could be a potential security hazard if one was to set the environment value to e.g. close the brackets in the actual command and inject their own code to execute there too.
> 
> This will also generate errors on running the software when the environment variable is NOT set even before running the command itself due to the environment itself being set.
> 
> What's holding you from changing the actual python script to be more explicit and have the code be:
> 
> ```python
> import os
> nuke.nodePaste(os.environ["QUAD_PLUGIN_PATH"] + "/host/nuke/scripts/flockOfBirds.nk")
> ```
> 
> And if multi-line is not supported maybe doing: `import os; nuke.nodePaste(os.environ["QUAD_PLUGIN_PATH"] + "/host/nuke/scripts/flockOfBirds.nk")`?
> 
> This leaves less potential errors up to the OpenPype implementation and allows anyone to continue to do with their scripts what they'd like.
> 
> Especially because just formatting the string plain and simple will also need extra 'escaping' if the Python command itself already had formatting too. Say python snippet was: `print("{0}_{1}".format("foo", "bar"))` I suspect that'll now also break with this code change.
> 
> As such I'd recommend against this implementation change.

There's some points here I didn't think of. My change seems to add more problems than solutions, thank you for your comment !

iLLiCiTiT:
Closing the PR based on the conversation above.

Issue #2509, opened 2022-01-10
Collect Input Links specifically for each published subsets


BigRoy:
**Is your feature request related to a problem? Please describe.**

We want to know exactly which loaded versions were used in the generation of a specific published version of a subset.

OpenPype currently does have some basic logic that [stores all loaded container versions] but [it only so when integrating a workfile](https://github.com/pypeclub/OpenPype/blob/f03472c017bed16d660a5f0f273533a35c94a9e8/openpype/plugins/publish/integrate_inputlinks.py). In a way that makes sense since it collects everything that was loaded in the scene, and not anything that was likely used in the generation of the specific subset. And that's exactly what we'd like to know instead.

We want to know which characters are in renderlayer 'Character' and which backgrounds are in renderlayer 'Environment' and exclude any content that isn't actually in that particular renderlayer even thought they might have been loaded into the work file (for being in the other renderlayer). 

**Describe the solution you'd like**

In colorbleed-config we previously had it implemented like this:

For example like this:
- https://github.com/Colorbleed/colorbleed-config/blob/acre/colorbleed/plugins/maya/publish/collect_inputs.py
- https://github.com/Colorbleed/colorbleed-config/blob/acre/colorbleed/plugins/houdini/publish/collect_inputs.py
- https://github.com/Colorbleed/colorbleed-config/blob/acre/colorbleed/plugins/fusion/publish/collect_inputs.py
And upon integrating:
- https://github.com/Colorbleed/colorbleed-config/blob/76f537eea041f8fed135f2b3334d70e1cf22b528/colorbleed/lib.py#L853-L864

We had tools in place to debug input dependencies and have used small scripts to sync up loaded versions of different geo based on used versions in their input history. Often used for FX or Comp to sync specific layers / assets.

One of the tools was in the repo - but I'm not sure that's currently the latest state but the code shows the overall idea. This tool allows to [show inputs/outputs of a subset](https://github.com/Colorbleed/colorbleed-config/blob/76f537eea041f8fed135f2b3334d70e1cf22b528/colorbleed/plugins/global/load/show_representation_inputs.py) using the [Representation Input/Output tool.](https://github.com/BigRoy/core/blob/acre/avalon/tools/representationinput.py)

Our implementation was functional for Maya, Houdini and Fusion.

**Additional context**

Related to:
- https://github.com/pypeclub/OpenPype/issues/2092 Issue
- https://github.com/pypeclub/OpenPype/pull/2192 PR

[...]

Issue #2092, opened 2021-09-30
Typed asset version dependencies


davidlatwe:
## Goal

As an artist/lead/developer, I would like to know the dependencies of a version (workfile/published output/etc.) so that I can make educated decisions without having to open or inspect the file/version in question.

## Use cases

* Knowing which versions were used in an output during a review.

* Ability to make better decisions in data transfer and cleanup tools.

## Details

Asset version dependencies should be typed (best configurable) to be able to represent different dependencies.

Example:

* Generative - No "active" dependency, the dependant was generated from the provider. Such as an image rendered from a Nuke script.

* Reference - The dependant actively uses the provider. Such as a rig referenced in a Maya scene or an image input into a Nuke script.

* Transitive: A indirect dependency, the dependent was generated from a provider that itself has a dependency on "the" provider. Such as the dependency from a rig to a model without references (where the direct dependency is "model">"workfile">"rig"),

* Member: The provider belongs to a "package" as a member.

Through typed dependencies, better decisions can be made. For ex. when transferring files only "reference" dependencies are actually needed etc..

### brainstorm in Miro

https://miro.com/app/board/o9J_lAB4oQ0=/

## Tasks
* [ ] publish incoming and outgoing links
* [ ] simple gui for displaying asset links
* [ ] simple gui for editing asset links

---
**Internal ticket: PYPE-1425**

[cuID:PYPE-1955]

davidlatwe:
For the data that need to be published along with, I propose (after a short discuss with @mkolar) the following data model which will be written/updated in **version** type document's `data.link`.

```
# say in a pointcache version document's 'data.link'
{
    "involved": [
        ObjectId("layoutB"),
        ObjectId("renderS"),
    ],
    "inputs": [
        {id: ObjectId("rigB"), by: publish, type: generative}
        {id: ObjectId("animation-workfile"), by: publish, type: reference}
    ],
}
```

The `involved` field is for recording which product is involved, e.g. modelA-v001 may involved with renderX-v005.

The `inputs` fields is for links. From the Miro board, there are 4 types of links : generative, reference, transitive, member. For generative and reference type of links, they should be logged in document of course. But the other two, they seems indirect and can be calculated on the fly.

The `outputs` field is missing in above model, since they can be computed from other's `inputs`.

mkolar:
[...]

PR #3629, opened 2022-08-08
General: Collect (generative) input links

BigRoy:
## Brief description

This is a (currently **DRAFT**) pull request to implement specific input links used in the generation of a specific publish.

## Description

This basically implements the following code mentioned in [this comment](https://github.com/pypeclub/OpenPype/issues/2092#issuecomment-1113324596):

> https://github.com/Colorbleed/colorbleed-config/blob/acre/colorbleed/plugins/maya/publish/collect_inputs.py
> https://github.com/Colorbleed/colorbleed-config/blob/acre/colorbleed/plugins/houdini/publish/collect_inputs.py
> https://github.com/Colorbleed/colorbleed-config/blob/acre/colorbleed/plugins/fusion/publish/collect_inputs.py

Note that the [Houdini code actually already exists in the OpenPype code base](https://github.com/pypeclub/OpenPype/blob/e93558f961c32615b36177f0e2a212ed59f22e24/openpype/hosts/houdini/plugins/publish/collect_inputs.py) and thus is **NOT** in this PR.

## TODO

- [x] Collectors for Maya, Fusion, Houdini 
- [ ] Make sure the data link type is correct, either `generative` or `reference` depending on the publish type
- [x] Actually integrate the data into the database

To mark the input data link type correctly we'd probably need to allow the Extractors define that relationship. For example a `mayaScene` extractor would mark the input links as `reference` (published data will still needs those references) whereas e.g. an alembic cache would mark its inputs as just `generative` (references are just used to generate output that works standalone) 

## Additional info

This relates to:
- #2509 
- #2092 
- [OpenPype Asset Dependencies Miro Board](https://miro.com/app/board/o9J_lAB4oQ0=/)
- [OpenPype Discord Input Links generation thread](https://discord.com/channels/517362899170230292/1006207855503876156)

## Testing notes:
1. Open any of Maya, Fusion, Houdini
2. Load something with the loader
3. Create/Publish something with the loaded content used as input.
4. (Optional) Also test with loading something and NOT using that as input. This content should _not_ be collected for the instance.
5. Publish.
6. Check the input links / dependencies.

### DIFF
--- a/openpype/hosts/fusion/plugins/publish/collect_inputs.py
+++ b/openpype/hosts/fusion/plugins/publish/collect_inputs.py
@@ -1,0 +1,112 @@
+import pyblish.api
+
+from openpype.pipeline import registered_host
+
+
+def collect_input_containers(tools):
+    """Collect containers that contain any of the node in `nodes`.
+
+    This will return any loaded Avalon container that contains at least one of
+    the nodes. As such, the Avalon container is an input for it. Or in short,
+    there are member nodes of that container.
+
+    Returns:
+        list: Input avalon containers
+
+    """
+
+    # Lookup by node ids
+    lookup = frozenset([tool.Name for tool in tools])
+
+    containers = []
+    host = registered_host()
+    for container in host.ls():
+
+        name = container["_tool"].Name
+
+        # We currently assume no "groups" as containers but just single tools
+        # like a single "Loader" operator. As such we just check whether the
+        # Loader is part of the processing queue.
+        if name in lookup:
+            containers.append(container)
+
+    return containers
+
+
+def iter_upstream(tool):
+    """Yields all upstream inputs for the current tool.
+
+    Yields:
+        tool: The input tools.
+
+    """
+
+    def get_connected_input_tools(tool):
+        """Helper function that returns connected input tools for a tool."""
+        inputs = []
+
+        # Filter only to actual types that will have sensible upstream
+        # connections. So we ignore just "Number" inputs as they can be
+        # many to iterate, slowing things down quite a bit - and in practice
+        # they don't have upstream connections.
+        VALID_INPUT_TYPES = ['Image', 'Particles', 'Mask', 'DataType3D']
+        for type_ in VALID_INPUT_TYPES:
+            for input_ in tool.GetInputList(type_).values():
+                output = input_.GetConnectedOutput()
+                if output:
+                    input_tool = output.GetTool()
+                    inputs.append(input_tool)
+
+        return inputs
+
+    # Initialize process queue with the node's inputs itself
+    queue = get_connected_input_tools(tool)
+
+    # We keep track of which node names we have processed so far, to ensure we
+    # don't process the same hierarchy again. We are not pushing the tool
+    # itself into the set as that doesn't correctly recognize the same tool.
+    # Since tool names are unique in a comp in Fusion we rely on that.
+    collected = set(tool.Name for tool in queue)
+
+    # Traverse upstream references for all nodes and yield them as we
+    # process the queue.
+    while queue:
+        upstream_tool = queue.pop()
+        yield upstream_tool
+
+        # Find upstream tools that are not collected yet.
+        upstream_inputs = get_connected_input_tools(upstream_tool)
+        upstream_inputs = [t for t in upstream_inputs if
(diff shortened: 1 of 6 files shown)


BigRoy:
Current example of "Workfile" links provided by @kalisp:

openpype_input_links_workfiles_example_by_kalisp

Most input links for publishes and containers "used in the generation of that publish" will likely be **generative** by default but some (like a `mayaScene`) will still be dependent on the input references/used files and thus will be of link type **reference** as described on the [Miro board](https://miro.com/app/board/o9J_lAB4oQ0=/)

afbeelding

I was thinking of having the Extractor or a family setting whether the input links are Generative or Reference, with Generative being the default due to being most common?

BigRoy:
Current code should work as expected. 

Only thing to note still is that [ALL these links will be `generative` due to how `inputVersions` are linked in the Integrate Input Links integrator](https://github.com/pypeclub/OpenPype/blob/develop/openpype/plugins/publish/integrate_inputlinks.py#L77-L84). Technically for a `mayaScene` publish this would be "incorrect" since the input versions would actually be `reference` links because the publishes maya scene would still need the original 'references' to exist, etc.

BigRoy:
We're using this in production now!

However I've noticed that input versions don't make their way into published renderlayers from e.g Maya through Deadline. That's now fixed with https://github.com/pypeclub/OpenPype/pull/3629/commits/eab14fc5e9204126554e8be5898010e0db0398ca however that'll include the input versions with each AOV. For many input ids and many AOV this could quickly bloat the `.json` file quite a bit. Another option would be instead to store it outside of the instance skeleton data and just store it globally in the metadata json and then we assume each AOV will share the same input versions? (However I'm not sure how true that logic is for potential other hosts/submissions supporting metadata.json files)

Also, I'm mapping the `inputVersions` objectIds to `str` so the `json.dump` works nicely. Instead of doing that we could also use a custom JSON encoder which forces all objectIds to become str anyway. But I felt explicitly for this value made more sense.

antirotor:
Thanks, this looks great! Can't wait to test it. One note about the AOV/input links - I was thinking hard about the uses in different DCCs and I think that this applies. In one deadline submission - ie. connected to one json file, we have output from the same data, after effects composition, harmony scene or maya render layer. There might be of course exceptions if you go to big enough detail - like should AOV of lightgroup have input link to referenced light rig even if it doesn't include them, and so on.

BigRoy:
> There might be of course exceptions if you go to big enough detail - like should AOV of lightgroup have input link to referenced light rig even if it doesn't include them, and so on.

I love you. That's a perfect use case where an AOV could differ.

BigRoy:
Would be great if this could see some testing soon!

antirotor (approved):
I've tested it, not in depth but so far it works and it doesn't break publishing so I think it is safe to merge. I've found gaps in workfile links but those are no concern of this PR

BigRoy:
>  I've found gaps in workfile links.

Could you elaborate?

antirotor:
> Could you elaborate?

it's actually difficult question. If you publish model and with it workfile, should that published workfile have outgoing link to the model? That's what happened to me - that published workfiles are not linking to published stuff.

mkolar:
> > Could you elaborate?
> 
> it's actually difficult question. If you publish model and with it workfile, should that published workfile have outgoing link to the model? That's what happened to me - that published workfiles are not linking to published stuff.

Ideally that should be the case, removing as many workfile dependencies as possible of course. But I'd personally merge this and continue that conversation in an improvement issue. This in itself will be a massive improvement right away.

mkolar (approved):

BigRoy:
> it's actually difficult question. If you publish model and with it workfile, should that published workfile have outgoing link to the model? That's what happened to me - that published workfiles are not linking to published stuff.

Wait a second - for me it does that! 

afbeelding

You are saying a workfile is not showing its outputs?
Are you seeing any outputs whatsoever?

BigRoy:
@antirotor Could you confirm [my last comment](https://github.com/pypeclub/OpenPype/pull/3629#issuecomment-1243870103)?

Would love to resolve and get this PR merged.

mkolar:
We can merge this. It didn't work for Ondrej, but it did for me. I reckon we can iron it out over time if we find the cause.

PR #3630, opened 2022-08-08
Editorial: Mix audio use side file for ffmpeg filters

iLLiCiTiT:
## Brief description
Mix of audio using ffmpeg command should not cause error when ffmeg can't process too long command.

## Description
Modified command preparations for audio mix to be able store filters into a file and tell ffmpeg where to look for filters to process. Also was modified how a time for `dropout_transition` is defined from hardcoded `1000` calculate it using longest input audio in miliseconds.

## Testing notes:
1. Publish a timeline with at least 30 items containing audio (probably can be one very long audio item?)
2. It should not crash because of "too long command" in ffmpeg (or any other error in the `Extract OTIO Audio Tracks` plugin)

### DIFF
--- a/openpype/plugins/publish/extract_otio_audio_tracks.py
+++ b/openpype/plugins/publish/extract_otio_audio_tracks.py
@@ -57,15 +57,7 @@ class ExtractOtioAudioTracks(pyblish.api.ContextPlugin):
         audio_inputs.insert(0, empty)
 
         # create cmd
-        cmd = path_to_subprocess_arg(self.ffmpeg_path) + " "
-        cmd += self.create_cmd(audio_inputs)
-        cmd += path_to_subprocess_arg(audio_temp_fpath)
-
-        # run subprocess
-        self.log.debug("Executing: {}".format(cmd))
-        openpype.api.run_subprocess(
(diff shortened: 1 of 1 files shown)


openpype/plugins/publish/extract_otio_audio_tracks.py
```diff
@@ -245,46 +237,80 @@ def create_empty(self, inputs):
...
-        for index, input in enumerate(inputs):
-            input_format = input.copy()
-            input_format.update({"i": index})
-            input_format["mediaPath"] = path_to_subprocess_arg(
-                input_format["mediaPath"]
+        input_args = []
+        filters = []
+        tag_names = []
+        for index, audio_input in enumerate(audio_inputs):
+            input_args.extend([
```
jakubjezek001, line 261: I wonder if this could not be also added into the sidecar file. The similar issue was someone solving here https://stackoverflow.com/questions/64407313/is-it-possible-to-provide-input-files-list-to-ffmpeg-in-a-text-file-instead-of-o

jakubjezek001 (approved):
tests were successful. Tested on 382 individual audio clips (192 L, 192 R)

PR #3631, opened 2022-08-08
Maya: FBX support for update in reference loader

simonebarbieri:
## Brief description
Fixes a problem with Maya for which it is impossible to set hero version for Unreal Static Meshes.

## Description
It happens because Unreal Static Meshes are FBXs, and the FBX format was unsupported in the update function of the Reference Loader.

## Testing notes:
1. Publish a model as Unreal Static Mesh with hero version enabled for the `staticMesh` family.
2. Load the model as hero version *or* change to hero version for a loaded Static Mesh.

### DIFF
--- a/openpype/hosts/maya/api/plugin.py
+++ b/openpype/hosts/maya/api/plugin.py
@@ -236,10 +236,16 @@ class ReferenceLoader(Loader):
                                            representation["context"]
                                                          ["project"]
                                                          ["name"])
+
+            params = {
+                "loadReference": reference_node,
+                "returnNewNodes": True
+            }
+            if file_type != "fbx":
+                params["type"] = file_type
+
             content = cmds.file(path,
-                                loadReference=reference_node,
-                                type=file_type,
-                                returnNewNodes=True)
+                                **params)
         except RuntimeError as exc:
             # When changing a reference to a file that has load errors the
             # command will raise an error even if the file is still loaded
--- a/openpype/hosts/maya/api/plugin.py
+++ b/openpype/hosts/maya/api/plugin.py
@@ -209,7 +209,7 @@ class ReferenceLoader(Loader):
             "ma": "mayaAscii",
             "mb": "mayaBinary",
             "abc": "Alembic",
-            "fbx": "fbx"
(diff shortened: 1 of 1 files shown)


BigRoy:
Could you describe why this works? Somehow you CAN update a referenced fbx when you don't pass the type again when reloading it. That seems... odd?

simonebarbieri:
> Could you describe why this works? Somehow you CAN update a referenced fbx when you don't pass the type again when reloading it. That seems... odd?
> 
> Isn't the fix instead to just have `fbx` added as file type on top?

I tried that, but when calling `cmds.file` with `type = 'fbx'`, it kept giving me `Reference file not found`. The only way to make it work (that I found) was not giving it the type parameter.

BigRoy:
@simonebarbieri Could you try with the fbx type set to uppercase as `"fbx": "FBX"`

I suspect that'll work. When manually replacing the FBX path in the reference editor then maya runs this in MEL:
**Succeeds:**
```
file -loadReference "cube1RN" -type "FBX" -options "v=0;" "C:/Users/Roy/Desktop/sphere1.fbx";
```
Which works. But if I replace it manually to lowercase it also fails like your case.
**Fails:**
```
file -loadReference "cube1RN" -type "fbx" -options "v=0;" "C:/Users/Roy/Desktop/sphere1.fbx";
```

simonebarbieri:
@BigRoy OMG it worked 😅 Thanks a lot, I'll fix it now!

openpype/hosts/maya/api/plugin.py
```diff
@@ -234,11 +235,17 @@ def update(self, container, representation):
             path = self.prepare_root_value(path,
                                            representation["context"]
                                                          ["project"]
-                                                         ["code"])
+                                                         ["name"])
```
BigRoy, line 238: Also, this change got in there by accident, or?

simonebarbieri: That was actually a bug! `prepare_root_value` takes the project name, not the code!

iLLiCiTiT (approved):
I agree it looks ok.

PR #3632, opened 2022-08-08
Nuke: color settings for render write node is working now

jakubjezek001:
## Brief description
Getting correct settings for write node was not working.

## Additional info
Key name in avalon data was defined as `family` but in reality in settings and in publishing plugins we are using `families`.

## Testing notes:
1. open nuke workfile and create render write node.
2. open settings and go to `project_anatomy/imageio/nuke/nodes/requiredNodes/0/knobs/6/value` and set to different colorspace setting.
3. Save settings and then go back to nuke and in openpype menu chose `Apply all`
4. colorspace on write node should change (this can only be seen inside the write node group)


jrsndl (approved):
Works as described. Tested In Nuke 12.2v3

PR #3633, opened 2022-08-09
Refactor `get_output_link_versions` to query id instead of input

BigRoy:
## Brief description

This fixes `get_output_link_versions` query.

Now it search for `data.inputLinks.id` instead of `data.inputLinks.input`

## Description

With this change the Outputs in the Loader are now also correctly shown:

afbeelding

## Additional info

Note: The old logic is likely due to how the ids were stored previously. So we might want to add backwards compatibility similar to [how the SimpleLinkView does for inputs here](https://github.com/pypeclub/OpenPype/blob/a04840eadfc7e0be3d69e7a0c2dd13bd9109393b/openpype/tools/assetlinks/widgets.py#L64-L68).

Or if that "old style" is **ancient legacy** we can also remove the backwards compatibility there and only consider entries with `id` data present for inputs.

---

This will be a useful fix once #3629 is merged as well. It'll allow you to trace easily from a specific version in what publishes it's used down the line.

### DIFF
--- a/openpype/client/entities.py
+++ b/openpype/client/entities.py
@@ -819,7 +819,7 @@ def get_output_link_versions(project_name, version_id, fields=None):
     # Does make sense to look for hero versions?
     query_filter = {
         "type": "version",
-        "data.inputLinks.input": version_id
+        "data.inputLinks.id": version_id
     }
     return conn.find(query_filter, _prepare_fields(fields))
 


iLLiCiTiT:
> Or if that "old style" is ancient legacy we can also remove the backwards compatibility there and only consider entries with id data present for inputs.

Good question. I think it is "ancient legacy", maybe from avalon? But we actually have one more legacy key, it was also `data.inputLinks._id`. Not sure for how long and if it's still relevant.

For full backwards compatibility it would be (not nice):
```
query_filter = {
    "type": "version",
    "$or": [
        {"data.inputLinks.id": version_id},
        {"data.inputLinks.input": version_id},
        {"data.inputLinks._id": version_id},
    ]
}
```

BigRoy:
> For full backwards compatibility it would be (not nice):

Ouch. Let me know what the full verdict is on how we'd want to support it and we go for it. I'd say the ancient one we can maybe ignore if it currently already isn't supported anyway?

iLLiCiTiT (approved):
Fixes current issue. We'll discuss possible needs of backwards compatibility but it looks like all of backwards compatible keys are not used or did not contain all needed data.

PR #3638, opened 2022-08-09
General: Hero version representations have full context

iLLiCiTiT:
## Brief description
Add all keys from anatomy data to representation context even if it's already there.

## Description
Integrate hero versions skipped reusing of anatomy data values in repre context if the key was already there but e.g. for `project` or `task` the key can be there but not all subkeys were used. Was found out because project name was filled but project code was missing in representation context.

## Testing notes:
1. Publish hero version which does not use project code in templates
2. Check DB if representations contain project code in `context`


kalisp (approved):
Screenshot 2022-08-12 130722

Issue #3233, opened 2022-05-23
Maya: Importing Pymel in Maya 2020.4 is failing


m-u-r-p-h-y:
**Describe the bug**

`import pymel.core as pm`

is not working in Maya2020 launched from OP

`# Error: NameError: file C:\Program Files\Autodesk\Maya2020\Python\lib\site-packages\pymel\core\__init__.py line 77: name 'windows' is not defined # `

**To Reproduce**

see the step above

**Expected behavior**

vanilla Maya is importing pymel with no issue

Maya 2020, Cut Number 202011110415
or
Maya2020.4

[cuID:OP-3287]

m-u-r-p-h-y:
Another user reported the same issue on MacOS

![image](https://user-images.githubusercontent.com/352795/170291035-04ace98b-a5f7-422a-8777-3bb1debf9977.png)

m-u-r-p-h-y:
I still have trouble consistently replicating the issue. It is very random and it seems it is dependent on loading the other plugins like arnold etc.

antirotor:
My research shows that this is directly connected to Maya 2020 and MtoA 5.1.3. It might affect different versions of MtoA and Maya simply because pymel is very sensitive to node definitions and it just happens that MtoA 5.1.3 is defining `aiSky` node in pymel incompatible way.

There is probably solution to in from OpenPype point of view - monkeypatch that pymel function that is checking for node superclass definitions using assert, but by my opinion it is not worth the effort as this doesn't happen in Maya 2022 and MtoA 5.1.3 or even in Maya 2023.

2-REC:
Having the same issue (when trying to load a component via the OpenPype loader) in Maya 2020.2 with MtoA 5.1.3.1.

The problem occurs when importing Pymel if MtoA is loaded.

So as a "solution", the following can bypass the issue:
* unload MtoA
* import Pymel module
* load MtoA

Once the Pymel module has been imported, following imports will not have the problem, even if MtoA is loaded.

It is of course not a real solution, but if somehow stuck with these 2 versions, it might be better than nothing.

Issue #2353, opened 2021-12-01
Maya: Missing PyMEL issue


antirotor:
## Problem

Some of publishing plugins are not working in Maya 2022 because of missing pymel module:

```
Skipped: "%s" (%s) collect_ass No module named 'pymel'
Skipped: "%s" (%s) collect_renderable_camera No module named 'pymel'
Skipped: "%s" (%s) collect_review No module named 'pymel'
Skipped: "%s" (%s) extract_playblast No module named 'pymel'
Skipped: "%s" (%s) extract_thumbnail No module named 'pymel'
Skipped: "%s" (%s) validate_attributes No module named 'pymel'
Skipped: "%s" (%s) validate_mesh_arnold_attributes No module named 'pymel'
Skipped: "%s" (%s) validate_mesh_overlapping_uvs No module named 'pymel'
Skipped: "%s" (%s) validate_no_namespace No module named 'pymel'
Skipped: "%s" (%s) validate_render_image_rule No module named 'pymel'
Skipped: "%s" (%s) validate_rig_output_ids No module named 'pymel'
```

PyMel is can be missing from Maya (can be installed separately) and we should account for that.

It can be installed like `mayapy -m pip install pymel` or `mayapy -m pip install --user pymel`

We should use Validator to find out if we have pymel available and instruct user to install it, or we should do it automatically (tricky, because it will either neeed Administrator privilegues as it will modify maya install location, or it will install per user)

[cuID:OP-2084]

BigRoy:
Without Pymel the user is also unable to e.g. load a published model - resulting in an error like this:
![maya_load_model_no_pymel](https://user-images.githubusercontent.com/2439881/146188533-21bef1ed-d40b-4bde-9d91-bc874c1a8900.png)

BigRoy:
> We should use Validator to find out if we have pymel available and instruct user to install it, or we should do it automatically (tricky, because it will either neeed Administrator privilegues as it will modify maya install location, or it will install per user)

That wouldn't be the case if just provide the `pymel` package as a dependency to OpenPype however. Nevertheless, the easier fix might be to not reply on Pymel altogether.

ClementHector:
We can verify / solve this problem in several places:
  - In the launcher, check with `mayapy pip seach mayapy`
  - When starting maya, check with `import pymel`
  - In pyblish, check with a pyblish plugin and a resolver

ClementHector:
[...]

PR #3643, opened 2022-08-10
Maya: Refactor moved usage of CreateRender settings

### LINKED ISSUE #3642: Maya: Validate Images File Rule (Workspace) always validates "None"
**Describe the bug**

Validator doesn't read the right value from settings:

afbeelding

This is incorrect:
```python
    @staticmethod
    def get_default_render_image_folder(instance):
        return instance.context.data.get('project_settings')\
            .get('maya') \
            .get('create') \
            .get('CreateRender') \
            .get('default_render_image_folder')
```

And should instead be:
```python
    @staticmethod
    def get_default_render_image_folder(instance):
        return instance.context.data.get('project_settings')\
            .get('maya') \
            .get('RenderSettings')
            .get('default_render_image_folder')
```

Settings are now [here](https://github.com/pypeclub/OpenPype/blob/e0c5c5a664093a29cb5aa2d5792e687b7be06e08/openpype/settings/defaults/project_settings/maya.json#L34-L63).

### Issue might also be with `aov_separator`

I suspect the same issue would be true for the [`aov_separator` in `lib_rendersettings.py`](https://github.com/pypeclub/OpenPype/blob/e0c5c5a664093a29cb5aa2d5792e687b7be06e08/openpype/hosts/maya/api/lib_rendersettings.py#L59-L68) and thus will now always hit a key error with recent settings?

I believe this might be broken since this PR https://github.com/pypeclub/OpenPype/pull/3097

[cuID:]

BigRoy:
## Brief description

This is a draft untested PR to fix #3642 

## Description

Please test publishing from Maya with up-to-date saved settings.

## Testing 

1. Ensure render images file rule works as expected.
2. Ensure aov separator works as expected

### DIFF
--- a/openpype/hosts/maya/api/lib_rendersettings.py
+++ b/openpype/hosts/maya/api/lib_rendersettings.py
@@ -60,8 +60,7 @@ class RenderSettings(object):
         try:
             aov_separator = self._aov_chars[(
                 self._project_settings["maya"]
-                                      ["create"]
-                                      ["CreateRender"]
+                                      ["RenderSettings"]
                                       ["aov_separator"]
             )]
         except KeyError:
--- a/openpype/hosts/maya/plugins/publish/collect_render.py
+++ b/openpype/hosts/maya/plugins/publish/collect_render.py
@@ -203,8 +203,7 @@ class CollectMayaRender(pyblish.api.ContextPlugin):
             aov_dict = {}
             default_render_file = context.data.get('project_settings')\
                 .get('maya')\
-                .get('create')\
-                .get('CreateRender')\
+                .get('RenderSettings')\
                 .get('default_render_image_folder') or ""
             # replace relative paths with absolute. Render products are
             # returned as list of dictionaries.
--- a/openpype/hosts/maya/plugins/publish/validate_render_image_rule.py
+++ b/openpype/hosts/maya/plugins/publish/validate_render_image_rule.py
@@ -41,6 +41,5 @@ class ValidateRenderImageRule(pyblish.api.InstancePlugin):
     def get_default_render_image_folder(instance):
         return instance.context.data.get('project_settings')\
             .get('maya') \
-            .get('create') \
-            .get('CreateRender') \
+            .get('RenderSettings') \
             .get('default_render_image_folder')
--- a/openpype/hosts/maya/plugins/publish/validate_render_image_rule.py
+++ b/openpype/hosts/maya/plugins/publish/validate_render_image_rule.py
@@ -33,9 +33,13 @@ class ValidateRenderImageRule(pyblish.api.InstancePlugin):
 
     @classmethod
     def repair(cls, instance):
-        default = cls.get_default_render_image_folder(instance)
-        cmds.workspace(fileRule=("images", default))
-        cmds.workspace(saveWorkspace=True)
+
+        required_images_rule = cls.get_default_render_image_folder(instance)
+        current_images_rule = cmds.workspace(fileRuleEntry="images")
+
+        if current_images_rule != required_images_rule:
+            cmds.workspace(fileRule=("images", required_images_rule))
+            cmds.workspace(saveWorkspace=True)
 
     @staticmethod
     def get_default_render_image_folder(instance):
--- a/openpype/hosts/maya/plugins/publish/validate_render_image_rule.py
+++ b/openpype/hosts/maya/plugins/publish/validate_render_image_rule.py
@@ -1,13 +1,7 @@
-import maya.mel as mel
-import pymel.core as pm
+from maya import cmds
 
 import pyblish.api
 import openpype.api
-
-
-def get_file_rule(rule):
-    """Workaround for a bug in python with cmds.workspace"""
-    return mel.eval('workspace -query -fileRuleEntry "{}"'.format(rule))
 
 
 class ValidateRenderImageRule(pyblish.api.InstancePlugin):
@@ -28,7 +22,7 @@ class ValidateRenderImageRule(pyblish.api.InstancePlugin):
     def process(self, instance):
 
         required_images_rule = self.get_default_render_image_folder(instance)
-        current_images_rule = get_file_rule("images")
+        current_images_rule = cmds.workspace(fileRuleEntry="images")
 
         assert current_images_rule == required_images_rule, (
             "Invalid workspace `images` file rule value: '{}'. "
@@ -40,8 +34,8 @@ class ValidateRenderImageRule(pyblish.api.InstancePlugin):
     @classmethod
     def repair(cls, instance):
         default = cls.get_default_render_image_folder(instance)
-        pm.workspace.fileRules["images"] = default
-        pm.system.Workspace.save()
+        cmds.workspace(fileRule=("images", default))
+        cmds.workspace(saveWorkspace=True)
 
     @staticmethod
     def get_default_render_image_folder(instance):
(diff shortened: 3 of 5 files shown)


openpype/hosts/maya/api/lib_rendersettings.py
```diff
@@ -60,8 +60,7 @@ def set_default_renderer_settings(self, renderer=None):
         try:
             aov_separator = self._aov_chars[(
                 self._project_settings["maya"]
-                                      ["create"]
-                                      ["CreateRender"]
+                                      ["RenderSettings"]
                                       ["aov_separator"]
             )]
```
BigRoy, line 65: Both before and after I feel the `except KeyError` code isn't the most "explicit" we have now. We are currently unable to identify whether the key error is due to the `self._aov_chars` dictionary not having the right key or whether the Project Settings didn't have the stored setting key.

I feel like both are different errors and potentially should be handled differently. No?

BigRoy:
Only thing still missing to be able to submit renders from Maya again is the actual populating of [`full_exp_files` and thus `expectedFiles` remaining empty currently.] It's [missing this line that was there originally](https://github.com/pypeclub/OpenPype/blob/CI/3.10.1-nightly.3/openpype/hosts/maya/plugins/publish/collect_render.py#L272). 

[This commit would fix it](https://github.com/BigRoy/OpenPype/commit/baaa98bed0e8aee85bbe6bb5b4255c0522f0a5a7).

Wasn't sure whether it made sense to include it in this PR.

m-u-r-p-h-y (commented):
shouldn't we put some default value here?

image

m-u-r-p-h-y (changes requested):
still not passing through

```
    runner(*args)
  File "D:\REPO\OpenPype\openpype\hosts\maya\plugins\publish\validate_render_image_rule.py", line 30, in process
AssertionError: Workspace's `images` file rule must be set to: renders/
    runner(*args)
  File "<string>", line 30, in process
AssertionError: Workspace's `images` file rule must be set to: renders/ // 
```

image

BigRoy:
> still not passing through

I believe you should set it to just `renders` without the slash?

> shouldn't we put some default value here?

I'd just put whatever the [default `workspace.mel` sets that comes with OP, which is indeed `renders`](https://github.com/pypeclub/OpenPype/blob/9d2c35aa56bea90211b6469a0055d19e51606e4e/openpype/hosts/maya/resources/workspace.mel#L3)

m-u-r-p-h-y:
Yes, it is the slash so we should change the wording in the validator message

BigRoy:
> Yes, it is the slash so we should change the wording in the validator message

Technically this would also work `renders/lighting` or whatever subfolders you'd like.
What wording would describe it better for you?

Might just be best to strip any starting/trailing slashes when checking against it?

Also I'm not sure whether the `workspace.mel` actually allows storing the value with a slash at the end - if it does then technically it could also explicitly check the value directly with the trailing slash.

m-u-r-p-h-y:
as a "user" I was strictly following the validators messages to use the trailing slash:

Validates "images" file rule is set to "renders/"
and 
Workspace's `images` file rule must be set to: renders/

image

BigRoy:
@m-u-r-p-h-y How about like this? (Just configured it to `foobar` to test)

afbeelding

BigRoy:
Changed default value and changed the validator's wording.

> Might just be best to strip any starting/trailing slashes when checking against it to avoid user error anyway?

Would you still like me to update the validator for this?

BigRoy:
Another question. Is there any reason why this:
```python
def get_file_rule(rule):
    """Workaround for a bug in python with cmds.workspace"""
    return mel.eval('workspace -query -fileRuleEntry "{}"'.format(rule))

value = get_file_rule("images")
```

Isn't just this:
```python
value = cmds.workspace(fileRuleEntry="images")
```

---

And is there a reason why we're setting the workspace rule with Pymel using:
```python
        pm.workspace.fileRules["images"] = default
        pm.system.Workspace.save()
```

Instead of doing this:
```python
cmds.workspace(fileRule=("images", value))
cmds.workspace(saveWorkspace=True)
```

Especially since we might want to use as little Pymel as possible also with things like #2353 and #3233

m-u-r-p-h-y:
> > Might just be best to strip any starting/trailing slashes when checking against it to avoid user error anyway?
> 
> Would you still like me to update the validator for this?

I'd just treat it the standard way we do it in other places where we deal with paths. No special treatment.

m-u-r-p-h-y (commented):
render images file rule works as expected now

I have some troubles with aov separator:
(trying to break it on purpose)


set the aov separator to - (dash)
created couple of aovs

this is the default Redshift aov settings:
`<BeautyPath>/<BeautyFile>.<RenderPass>`

but the error I got from Render Settings validator

```
// Error: pyblish.ValidateRenderSettings : AOV (BeautyAux) image prefix is not set correctly <BeautyPath>/<BeautyFile>.<RenderPass> != <BeautyPath>/<BeautyFile>_<RenderPass> // 
// Error: pyblish.ValidateRenderSettings : AOV (Cryptomatte) image prefix is not set correctly <BeautyPath>/<BeautyFile>.<RenderPass> != <BeautyPath>/<BeautyFile>_<RenderPass> // 
// Error: pyblish.ValidateRenderSettings : AOV file format is not the same as the one set globally 1 != 6 // 
// Error: pyblish.ValidateRenderSettings : AOV (Z) image prefix is not set correctly <BeautyPath>/<BeautyFile>.<RenderPass> != <BeautyPath>/<BeautyFile>_<RenderPass> // 
    runner(*args)
  File "D:\REPO\OpenPype\openpype\hosts\maya\plugins\publish\validate_rendersettings.py", line 91, in process
AssertionError: Invalid render settings found for 'Main'!
    runner(*args)
  File "<string>", line 91, in process
AssertionError: Invalid render settings found for 'Main'! // 
```


is using underscore

also, the file format error is not very clear, as the user is not aware of 1 and 6 and what format it actually represents.

BigRoy:
Nice! :) 

---

```
// Error: pyblish.ValidateRenderSettings : AOV file format is not the same as the one set globally 1 != 6 // 
```

This is a known bug: https://github.com/pypeclub/OpenPype/issues/3410

---

```
// Error: pyblish.ValidateRenderSettings : AOV (Z) image prefix is not set correctly <BeautyPath>/<BeautyFile>.<RenderPass> != <BeautyPath>/<BeautyFile>_<RenderPass> // 
```

The others show that it's not being validated against the correct AOV separator.

That bug comes from the fact that it validates against `instance.data["aovSeparator"]` which is the aov separator actually being collected by the layer/AOV as collected in CollectRender and not the aov separator that the settings define.

As such, instead the validator should just be doing `instance.data["aovSeparator"] != settings_aov_separator` however that would technically be incorrect since Redshift allows a separator per AOV to be set instead of per layer and thus we should be validating that PER aov and not per layer.

As such this identifies an issue that might not be a simple fix - should we separate into an issue of its own?

antirotor:
>  should we separate into an issue of its own?

Yeah, it is not only redshift that allows separator to be defined per AOV (or rather the whole prefix, but  that is question if we even want to support). So I vote for another issue as that one will be harder to solve. I thought we have one already for this but I cannot find it.

m-u-r-p-h-y:
should this be merged than to get `images file rule` fixed?

BigRoy:
I'd say so. This is a decent hotfix now for the issue at hand except for maybe this missing commit:

> [This commit would fix it](https://github.com/BigRoy/OpenPype/commit/baaa98bed0e8aee85bbe6bb5b4255c0522f0a5a7).
> - [source comment](https://github.com/pypeclub/OpenPype/pull/3643#issuecomment-1210825259)

m-u-r-p-h-y (approved):
approved for ` images file rule`

antirotor (approved):

PR #3644, opened 2022-08-10
General: Extract Review can scale with pixel aspect ratio

iLLiCiTiT:
## Brief description
This [PR](https://github.com/pypeclub/OpenPype/pull/3620) removed ability to scale source using pixel aspect ratio which is fixed here.

## Description

## Additional info
technical information.

## Testing notes:
1. Make sure `project_settings/global/publish/ExtractReview/profiles/0/outputs/h264/scale_pixel_aspect` is set to true
2. Publish something with pixel aspect ratio different then 1
- from what I know it's possible from nuke or editorial
- e.g. anamorphic image (pixel aspect is 2 so 100x100 should have resolution 200x100)
3. Check output which should be scaled as expected

Resolves https://github.com/pypeclub/OpenPype/issues/3641

### DIFF
--- a/openpype/plugins/publish/extract_review.py
+++ b/openpype/plugins/publish/extract_review.py
@@ -1210,7 +1210,6 @@ class ExtractReview(pyblish.api.InstancePlugin):
 
         # Get instance data
         pixel_aspect = temp_data["pixel_aspect"]
-
         if reformat_in_baking:
             self.log.debug((
                 "Using resolution from input. It is already "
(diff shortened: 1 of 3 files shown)


jrsndl (approved):
Works as described. I tested in Nuke 12.2v3, on 2:1 anamorphic, with local publish. 

With Scale pixel aspect off, the burnin pixel aspect doesn't match the pixel aspect of the image. I take it as a current limitation.
image

PR #3646, opened 2022-08-10
Backend: Implementation of addon deployment logic

kalisp:
## Brief description
Should handle updating addon zips provided by v4 server.

## Description
Goal is for OP client to learn which addons (and their version) are activated on the server, download and unzip client side code. 

## Additional info
Implements basic logic to process returned json with information about addons (actual get from v4 server is not included yet).
Establishes json for addon information (from payload received from `localhost:5000/api/addons?details=1`)
```
"versions": {
          "1.0.0": {
               "clientPyproject": {
                    "tool": {
                         "poetry": {
                          "dependencies": {
                           "nxtools": "^1.6",
                           "orjson": "^3.6.7",
                           "typer": "^0.4.1",
                           "email-validator": "^1.1.3",
                           "python": "^3.10",
                           "fastapi": "^0.73.0"
                          }
                     }
                }
               },
               "hasSettings": True,
               "clientSourceInfo": [
                   {
                       "type": "http",
                       "url": "https://drive.google.com/file/d/1TcuV8c2OV8CcbPeWi7lxOdqWsEqQNPYy/view?usp=sharing"  # noqa
                   },
                   {
                       "type": "filesystem",
                       "path": {
                           "windows": ["P:/sources/some_file.zip",
                                       "W:/sources/some_file.zip"],  # noqa
                           "linux": ["/mnt/srv/sources/some_file.zip"],
                           "darwin": ["/Volumes/srv/sources/some_file.zip"]
                       }
                   }
               ],
               "frontendScopes": {
                    "project": {
                     "sidebar": "hierarchy"
                    }
               }
          }
     },
     "title": "Slack addon",
     "name": "openpype_slack",
     "productionVersion": "1.0.0",
      "hash": "4be25eb6215e91e5894d3c5475aeb1e379d081d3f5b43b4ee15b0891cf5f5658"   TODO - missing from v4 payload - should be `sha256` of zipped file to check that download was correct
       "description": "Lorem Ipsum ...",
       "license": "Apache-2.0",  TODO - missing from v4 payload
       "authors": "Petr Kalis <petr.kalis.orbi.tools>"  TODO - missing from v4 payload
}
```

`Description, license, authors` could be used in the future for some Addon store or pulling info to some customer internal CRM, where same server endpoint could be reused. For current deployment logic it is not used.

Two types of source urls are implemented:
- OS - mapped drive to machine
- HTTP - downloading from HTTP server, handles GDrive sharable links too

Compares provided sha256 hashes with downloaded .zip file.
Unzips addon zip file into destination folder (zip file is expected without "inner" addon folder, just bare code) and removes zip.

It is missing another important part of preparation of dependencies for addons. That might be a separate PR.


antirotor:
you have there md5 implemented instead of sha256 but I guess that's just placeholder. I was wondering about the metadata that is coming from it. Actually if there is a reason to extend it with more information, to create for example browsable repository, display more information about the addon or so. What about sending along something like stuff defined in [PEP-621](https://peps.python.org/pep-0621/) - then these could be read from pyproject.toml by server for example... `name`, `version`, `description`, `license`, `authors`. And for example `post-install-script` that could be run after the addon is unzipped to handle some checks, cleanups? This could be ofc done by properly designed hook api...

kalisp:
Implemented basic http get to mock server returning testing `get_addons_info` payload.

distribution/addon_distribution.py
```diff
@@ -0,0 +1,201 @@
...
+import logging
+import requests
+
+from distribution.file_handler import RemoteFileHandler
+
+
+class UrlType(Enum):
+    HTTP = "http"
+    GIT = "git"
+    OS = "os"
```
iLLiCiTiT, line 14: I would call it `filesystem`

antirotor: or "local"? since you can't encode easily UNC paths in `file:` schema?

iLLiCiTiT: I don't think there will be used schema for the type. In that case `addon_url` is just the path (which may be UNC) without `file:` at the start.

distribution/addon_distribution.py
```diff
@@ -0,0 +1,201 @@
...
+    # TODO temp
+    # addon_info = AddonInfo(
+    #     **{"name": "openpype_slack",
+    #        "version": "1.0.0",
+    #        "addon_url": "c:/projects/openpype_slack_1.0.0.zip",
+    #        "type": UrlType.FILESYSTEM,
+    #        "hash": "4f6b8568eb9dd6f510fd7c4dcb676788"})  # noqa
+    #
+    # http_addon = AddonInfo(
+    #     **{"name": "openpype_slack",
```
iLLiCiTiT, line 129: Just thinking about this. I would say that single item is representing single addon version and the item should have something like `"sources"` where are the source items with `"type"` and type related data. I would expect that `type` is only required key and the type handler would define remaining required keys. At the end you don't want download each source but check if the addon is available locally and if not then try sources one by one with source type specific handlers.

Also the data are what server will return so it will contain only json compatible types.

Example:
```javascript
{
    "name": "openpype_slack",
    "version": "1.0.0",
    // Hash is probably same for all sources?
    "hash": "4f6b8568eb9dd6f510fd7c4dcb676788",
    "sources": [
        {
            "type": "http",
            "url": "https://some/path"
        },
        {
            "type": "filesystem",
            "path": {
                "windows": ["P:/sources/some_file.zip", "W:/sources/some_file.zip"],
                "linux": ["/mnt/srv/sources/some_file.zip"],
                "darwin": ["/Volumes/srv/sources/some_file.zip"]
            }
        },
        {
            "type": "filesystem",
            "path": "P:/sources/some_file.zip"
        }
    ]
}
```

kalisp: Updated format

iLLiCiTiT:
Now it's question if we're able to make server send us these information?

I would add some readme to the folder that it is v4 related code.

Few notes related to client folder structure. They don't have to be handled in this PR.
1. I think there should be some parent for `distribution`. Import of `distribution` can be dangerous in python world. With combination that we will probably need more common code in future. I would suggest to put `distribution` into folder named like `openpype_common` (or else) so we would do imports like `from openpype_common import distribution`.
- Good / Bad ???

2. I'm nost sure about this because v4 client is not yet fully defined but we may need even one more additional parent folder so the "imported folder" is not in the root of build. That is because we may want add the path explicitly to `sys.path` on start which we may not be able to do if it will be in root.
```
# Structure (combined with 1.)
├ <build root>
│  ├ igniter
│  │ └ ...
│  ├ common (this is path which will be added to sys.path)
│  │  └ openpype_common
│  │    ├ distribution
│  │    └ ...
│  ├ start.py
│  └ ...
└ ...
```
- Good / Bad ???

kalisp:
Added readme.

distribution/addon_distribution.py
```diff
@@ -0,0 +1,244 @@
...
+    EXISTS = "exists"
+    UPDATED = "updated"
+    FAILED = "failed"
+
+
+@attr.s
+class MultiPlatformPath(object):
+    windows = attr.ib(default=None)
+    linux = attr.ib(default=None)
+    darwin = attr.ib(default=None)
```
antirotor, line 28: I am wondering wheter this should use `pathlib.Path` type directly

distribution/addon_distribution.py
```diff
@@ -0,0 +1,244 @@
...
+
+        Args:
+            addon_path (str): local path to addon zip file
+            addon_hash (str): sha256 hash of zip file
+        Raises:
+            ValueError if hashes doesn't match
+        """
+        if not os.path.exists(addon_path):
+            raise ValueError(f"{addon_path} doesn't exist.")
+        if addon_hash != RemoteFileHandler.calculate_md5(addon_path):
```
antirotor, line 90: shouldn't this match whats in docstring? There is **sha256**

distribution/addon_distribution.py
```diff
@@ -0,0 +1,244 @@
...
+    downloader_factory.register_format(UrlType.FILESYSTEM, OSAddonDownloader)
+    downloader_factory.register_format(UrlType.HTTP, HTTPAddonDownloader)
+
+    test_endpoint = "https://34e99f0f-f987-4715-95e6-d2d88caa7586.mock.pstmn.io/get_addons_info"  # noqa
+    if os.environ.get("OPENPYPE_SERVER"):  # TODO or from keychain
+        server_endpoint = os.environ.get("OPENPYPE_SERVER") + "get_addons_info"
+    else:
+        server_endpoint = test_endpoint
+
+    check_addons(server_endpoint, addon_folder, downloader_factory)
```
antirotor, line 244: are you sure about including these hardcoded values? I understand that it is still all WIP, but...?

antirotor, line 244: are you sure about including these hardcoded values? I understand that it is still all WIP, but...?

distribution/file_handler.py
```diff
@@ -21,7 +21,7 @@ class RemoteFileHandler:
                                'tar.gz', 'tar.xz', 'tar.bz2']
 
     @staticmethod
-    def calculate_md5(fpath, chunk_size):
+    def calculate_md5(fpath, chunk_size=10000):
```
antirotor, line 24: I vote for sha256 instead of md5 even when it is used just for file integrity check and not security

tests/lib/testing_classes.py
```diff
@@ -10,7 +10,7 @@
 import platform
 
 from tests.lib.db_handler import DBHandler
-from tests.lib.file_handler import RemoteFileHandler
+from distribution.file_handler import RemoteFileHandler
```
antirotor, line 13: I agree with @iLLiCiTiT about the module name and the question is if this shouldn't live in openpype module?

kalisp: I don't think it good idea, I think importing it from openpype might be problematic.

distribution/addon_distribution.py
kalisp: What would be benefit? Does it handle something like `source["path"].get(platform.system().lower())` ?

distribution/file_handler.py
kalisp: Ok, changed.

kalisp:
Introduced deeper folder structure

common/openpype_common/distribution/addon_distribution.py
```diff
@@ -0,0 +1,238 @@
...
+
+@attr.s
+class MultiPlatformPath(object):
+    windows = attr.ib(default=None)
+    linux = attr.ib(default=None)
+    darwin = attr.ib(default=None)
+
+
+@attr.s
+class AddonSource(object):
```
iLLiCiTiT, line 33: Comment which was not written, only discussed:
I would not make this strict and define required attributes. From my point of view only require attribute of addon source should be `"type"`. For each type should be dedicated type handler (what `AddonDownloader` is right now) which validate rest of the attributes.

I'm 99.99% convinced each type handler will require it's attributes with possible different types - and they won't be the same for all sources. Even now there is `url` and `path`. By adding other source type handlers the attribute list would grows and will be unmaintainable with each new source handler. Any possible future attribute change would cause clashes and custom implementations would be a lottery.

common/openpype_common/distribution/addon_distribution.py
```diff
@@ -0,0 +1,238 @@
...
+    addons_info = get_addons_info(server_endpoint)
+    result = update_addon_state(addons_info,
+                                addon_folder,
+                                downloaders)
+    if UpdateState.FAILED.value in result.values():
+        raise RuntimeError(f"Unable to update some addons {result}")
+
+
+def cli(*args):
+    raise NotImplemented
```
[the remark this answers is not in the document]
iLLiCiTiT: He's right...

common/openpype_common/distribution/addon_distribution.py
kalisp: Updated AddonSource, created 2 classes inheriting from it.

common/openpype_common/distribution/addon_distribution.py
```diff
@@ -0,0 +1,267 @@
+import os
+from enum import Enum
+from abc import abstractmethod
+import attr
+import logging
+import requests
+import platform
+import shutil
+
+from common.openpype_common.distribution.file_handler import RemoteFileHandler
```
iLLiCiTiT, line 10: You can use relative import here
```suggestion
from .file_handler import RemoteFileHandler
```

Or this (which would break current tests):
```
from openpype_common.distribution.file_handler import RemoteFileHandler
```

iLLiCiTiT (approved):
One [last comment](https://github.com/pypeclub/OpenPype/pull/3646/files#r972725145) regarging import otherwise approved.

common/openpype_common/distribution/addon_distribution.py
kalisp: Ok.

kalisp:
Merging it as working prototype.

Next steps will be done after OP client will be prepared, to hook this up, polish it up and test.

PR #3647, opened 2022-08-11
Maya: Hosts as modules

iLLiCiTiT:
## Brief description
Hosts are OpenPype modules so they have more abilities and more defined structure.

## Description
First step to be able define hosts as OpenPype modules which can make them self contained parts. Added interface `IHostModule` which marks a module as host module so it can handle host related logic before a DCC or process, where host is integrated, is running. Host module must have set which host it represents with it name (`host_name`). Can define it's workfile extensions out of integration so it's available before DCC is running. As module it can handle some logic more dynamically without predetermined paths. For example can define path to launch hooks using `get_launch_hook_paths` or handle launch environments implementing `add_implementation_envs` as module method.

Changed how paths to prelaunch hooks are received. An object of application which is launched is passed as an argument to `get_launch_hook_paths` so it is possible to filter out if it make sense to add certain launch hook paths for passed application. Module does not have to inherit from `ILaunchHookPaths` to be able define launch hooks.

To demonstrate the functionality a maya implementation is now used as module.

## Testing notes:
1. Launch Maya
2. Last workfile should be used on launch (if exists and the feature is enabled)
3. Maya pre launch hooks should work (workspace mel is created)
4. OpenPype is integrated in launched maya

### DIFF
--- a/openpype/host/host.py
+++ b/openpype/host/host.py
@@ -19,8 +19,15 @@ class MissingMethodsError(ValueError):
         joined_missing = ", ".join(
             ['"{}"'.format(item) for item in missing_methods]
         )
+        if isinstance(host, HostBase):
+            host_name = host.name
+        else:
+            try:
+                host_name = host.__file__.replace("\\", "/").split("/")[-3]
+            except Exception:
+                host_name = str(host)
         message = (
-            "Host \"{}\" miss methods {}".format(host.name, joined_missing)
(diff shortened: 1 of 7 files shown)


m-u-r-p-h-y (approved):
- maya launched with latest workfile loaded
- workspace.mel created
- openPype menu loaded

mkolar (approved):

Issue #3124, opened 2022-05-02
Loader: Show which subsets are already loaded in the scene


BigRoy:
**Is your feature request related to a problem? Please describe.**

Artists requested being able to highlight or see which subsets in the Loader are already loaded into the current scene.

**Describe the solution you'd like**

Likely an extra column that might show whether it's loaded with a potential number of loaded entries as value so that the user could sort by that column would already allow the user to be able to find unloaded content without too much effort.

It would be nice to make it a custom delegate that could also show e.g. an orange warning icon when the loaded content is actually loading an older version than the latest version and maybe a green check mark if it's currently using the latest loaded version.

Bonus points if hovering over it would show a tooltip or alike that could detail similar information as the Scene Inventory would show about the loaded content, like the container names, used loaders, etc. to find out whether e.g. the Alembic or another representation was loaded.

Having this visual in the Loader allows to also easily detect any entries that might have gotten a new publish under a new subset name or alike, e.g. an FX derivative of a (part of a) character.

**Describe alternatives you've considered**

Artists would currently open the Scene Inventory / Manager next to it to check against currently loaded content but visually it's not a 1-to-1 correspondence to the Loader list of subsets so it's relatively time consuming to compare against it - plus large scenes with many loaded assets get extra difficult in this way. Being able to easily find or filter already content in the Loader would solve that.

[cuID:OP-3167]

mkolar:
That's a very good idea actually. It does muddle the waters a bit between loader and inventory, but it woudl certainly be useful

BigRoy:
This has been implemented with #3649 but doesn't tell you anything about whether it's latest version or not, but I believe it's proven useful as it is.

I'll close this issue until anyone has additional requests for the feature.

PR #3649, opened 2022-08-11
Loader: Show whether a subset is loaded into the current scene

BigRoy:
## Brief description

This is a prototype for #3124 

## Description

This implements an extra column in the Loader that shows whether _any container of that **subset** is loaded into the current workfile_.

afbeelding

## Additional info

Due to `host.ls()` call the database queries to get the subsets for the loaded containers this of course will be slower than without this functionality. Please stress test in heavy scene with heavy databases over slow connections. :) 

~This currently uses `get_representations_parents` because I was lazy. That's far from the optimal query to get to the unique subset ids of the scene containers, see the `# todo` code.~

Note that the loaded containers only refreshes at most in every three seconds so if `yes`/`no` is not accurate, then wait a few seconds and press the refresh icon top left.

## Testing notes:

1. open a host
2. load some content (or open existing scene)
3. refresh the loader page.


openpype/tools/loader/model.py
```diff
@@ -614,6 +668,37 @@ def _fill_subset_items(
...
+            data.update(
+                self._get_last_repre_info(repre_info_by_version_id,
+                                          last_version["_id"]))
+
+            item = Item()
+            item.update(data)
+            self.add_child(item, parent_item)
+
+            index = self.index(item.row(), 0, parent_index)
+            self.set_version(index, last_version)
```
BigRoy, line 700: This was solely moved to a function inside the method to avoid the code duplication just a bit further down.

openpype/tools/loader/model.py
```diff
@@ -526,6 +560,17 @@ def refresh(self):
             self.doc_fetched.emit()
             return
 
+        # Collect scene container representations to compare loaded state
+        # This runs in the main thread because it involves the host DCC
+        if self._host:
+            time_since_refresh = time.time() - self._host_loaded_refresh_time
+            print(time_since_refresh)
```
BigRoy, line 567: Forgot to remove this print statement. Will push commit tomorrow.

antirotor (dismissed):
There is now one little conflict, but with it works nice.

BigRoy:
Thanks - will look into this tomorrow.

I was also thinking of turning the column into a delegate and storing `True/False` (boolean) instead of a "yes", "no" string in the model to clean things up. That way also the delegate could take care of the colorizing instead of the model data.

BigRoy:
Fixed merge conflicts + stored values as bool instead of "yes/no" string in model + added delegate which shows it as "yes" or "no" with some nice colors to it.

Now looks like this:
openpype_loader_show_containers_in_scene

BigRoy:
Should be ready to merge - so please do one more test-run.

openpype/tools/loader/model.py
```diff
@@ -533,6 +564,16 @@ def refresh(self):
             self.doc_fetched.emit()
             return
 
+        # Collect scene container representations to compare loaded state
+        # This runs in the main thread because it involves the host DCC
+        if self._host:
+            time_since_refresh = time.time() - self._host_loaded_refresh_time
+            if time_since_refresh > self._host_loaded_refresh_timeout:
+                repre_ids = {con.get("representation")
+                             for con in self._host.ls()}
```
iLLiCiTiT, line 573:
```suggestion
                if isinstance(self._host, ILoadHost):
                    containers = self._host.get_containers()
                else:
                    containers = self._host.ls()
                    
                repre_ids = {con.get("representation")
                             for con in self._host.ls()}
```

openpype/tools/utils/delegates.py
```diff
@@ -291,3 +291,30 @@ class PrettyTimeDelegate(QtWidgets.QStyledItemDelegate):
     def displayText(self, value, locale):
         if value is not None:
             return pretty_timestamp(value)
+
+
+class LoadedInSceneDelegate(QtWidgets.QStyledItemDelegate):
```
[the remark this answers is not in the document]
BigRoy: No - pretty tightly coupled currently. I intended originally to generalize it so it was a delegate that would allow a mapping of value to str and color, e.g. but felt that there was no need currently to make it more complex than needed.

Like this? https://github.com/pypeclub/OpenPype/pull/3649/commits/d5cb828edc611f9ed9f98b61ae5ae3591d797640

mkolar (commented):
I just ran it and it's working as expected, however the positioning of the column kind of breaks with site sync turned on. 

image

mkolar:
All good now, even though I'd expand the width of `in scene` a tiny bit at least. Other than that, happy to merge

BigRoy:
> All good now, even though I'd expand the width of in scene a tiny bit at least. 

Your wish is my command. [Done](https://github.com/pypeclub/OpenPype/pull/3649/commits/c3588e2f9d337109aea3c2630fb12079b617d81d).

mkolar (approved):

PR #3650, opened 2022-08-11
General: Optimized OCIO configs

antirotor:
## Description

OCIO configs vendorized as git submodule had a huge size, adding almost 3Gb to OpenPype.  This is changing it to use reduced [OCIO configs](pypeclub/OpenColorIO-Configs) managed by Poetry.

### Testing

This should build correctly and that darn `maketx` color space issue in Maya should be still fixed.

🔧 **This will go to 3.14 release**


Allan-I (commented):
Code looks good to me, will test it first thing tomorrow!

antirotor:
So Github has bandwidth limit on Git with LFS, so this is not option. We need to reimplement this to serve it as a zip file the same way as we are now doing with ffmpeg/oiio.

poetry.lock
```diff
@@ -147,25 +147,25 @@ python-versions = ">=3.5"
...
 
 [[package]]
 name = "attrs"
-version = "21.4.0"
+version = "22.1.0"
 description = "Classes Without Boilerplate"
 category = "main"
 optional = false
-python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
+python-versions = ">=3.5"
```
iLLiCiTiT, line 162: I think we're still using `attrs` in Python 2 DCCs. We should maybe add it to python 2 vendor.

Allan-I:
What do I need to do to test this out?  Just normally pull in and publish a look?

iLLiCiTiT:
You should run `~/openpype/tools/create_env` script to update third party libs.

Allan-I:
Thank you Kuba, this currently fails because the path evaluates to `C:\Users\Dumbledoor\Documents\Code\OpenPype\vendor\bin\ocioconfigOpenColorIOConfigs\nuke-default\config.ocio` which doesn't seem to exist.

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -40,11 +40,13 @@ def get_ocio_config_path(profile_folder):
...
         str: Path to vendorized config file.
     """
+
     return os.path.join(
         os.environ["OPENPYPE_ROOT"],
         "vendor",
-        "configs",
-        "OpenColorIO-Configs",
+        "bin",
+        "ocioconfig"
```
Allan-I, line 48: "ocioconfig"`

Allan-I, line 48: This is missing a comma I believe
from `"ocioconfig"` to `"ocioconfig",`

Allan-I (changes requested):

openpype/hosts/maya/plugins/publish/extract_look.py
iLLiCiTiT:
```suggestion
        "ocioconfig",
```

Allan-I (approved):
Tested and works!

PR #3652, opened 2022-08-11
Houdini: Shelf Manager

friquette:
## Brief description
Adding a shelf manager in houdini.

## Description
The shelf manager must make it possible to quickly implement shelves and tools in one or more menus without having to create them manually.

## Testing notes:
1. Open OpenPype Project settings
2. Go to project_settings/houdini/shelves/
3. Fill the settings
houdini_shelves

If you fill the Shelf Set Path, you don't have to fill the other settings.


openpype/hosts/houdini/api/shelves.py
```diff
@@ -0,0 +1,188 @@
...
+
+    # load configuration of houdini shelves
+    project_settings = get_project_settings(os.getenv("AVALON_PROJECT"))
+    shelves_set_config = project_settings["houdini"]["shelves"]
+
+    if not shelves_set_config:
+        log.warning(
+            "SHELF WARNGING: No custom shelves found in project settings."
+        )
+        return
```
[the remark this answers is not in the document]
iLLiCiTiT: You may have set them to empty list, but in that case it should be lower log level (`info` max).

openpype/settings/defaults/project_settings/houdini.json
```diff
@@ -1,4 +1,27 @@
...
+            "shelf_set_name": "OpenPype Shelves",
+            "shelf_set_source_path": {
+                "windows": "",
+                "darwin": "",
+                "linux": "/path/to/your/shelf_set_file"
+            },
+            "shelf_definition": [
+                {
+                    "shelf_name": "OpenPype Shelf",
+                    "tools_list": [
```
iLLiCiTiT, line 13: Invalid default values in settings are irrelevant.

openpype/hosts/houdini/api/lib.py
```diff
@@ -460,3 +460,35 @@ def reset_framerange():
...
+
+    new_tool = hou.shelves.newTool(
+        file_path='',
+        name='new_tool',
+        label='New Tool',
+        script='',
+        language=hou.scriptLanguage.Python,
+        icon='',
+        help='This is a new tool'
+    )
```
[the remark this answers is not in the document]
friquette: I'm sorry I forgot to remove this function, this is no longer needed.

openpype/hosts/houdini/api/shelves.py
```diff
@@ -0,0 +1,188 @@
...
+                    "SHELF WARNGING: No name found in shelf definition."
+                )
+                return
+
+            shelf = get_or_create_shelf(shelf_name)
+
+            for tool_definition in shelf_definition.get('tools_list'):
+                # We verify that the name and script attibutes of the tool
+                # are set
+                mandatory_attributes = ['name', 'script']
```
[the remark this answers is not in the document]
friquette: What is the point to define this as a set here?

openpype/hosts/houdini/api/shelves.py
```diff
@@ -0,0 +1,188 @@
...
+                if not all(
+                    [v for k, v in tool_definition.items() if
+                        k in mandatory_attributes]
+                ):
```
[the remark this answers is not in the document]
friquette: If the tool definition exists, so are the name and script keys. I added a condition to check if the tool definition exists before this loop.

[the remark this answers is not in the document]
BigRoy: I'd still recommend the code change to just iterate the `mandatory_attributes` since you wouldn't care about other attributes anyway.

openpype/hosts/houdini/api/shelves.py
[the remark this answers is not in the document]
BigRoy: You have logic that does a "is value in list" check which is `On` complexity to run, by using a `set` instead that should be an `O1` complexity lookup and faster. (The list in this case was two values so speed difference might be negatable, but usually  you'd do lookups for values using a `set` because it uses the hash to check instead of iterating over the list until it finds the exact values).

E.g. [more info here](https://towardsdatascience.com/faster-lookups-in-python-1d7503e9cd38).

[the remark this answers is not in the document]
friquette: Ok thank you, this is good to know !

openpype/hosts/houdini/api/pipeline.py
```diff
@@ -309,6 +310,7 @@ def _set_context_settings():
         fps
         resolution
         renderer
+        shelves
```
BigRoy, line 314:
```suggestion
        renderer

```

The `_set_context_settings` doesn't seem to do anything with shelves so adding it to its docstring seems incorrect?

openpype/hosts/houdini/api/shelves.py
```diff
@@ -0,0 +1,196 @@
...
+    shelf_name = shelf_label.replace(' ', '_').lower()
+    new_shelf = hou.shelves.newShelf(
+        name=shelf_name,
+        label=shelf_label
+    )
+    return new_shelf
+
+
+def get_or_create_tool(tool_definition, shelf):
+    """This function verifies if the tool exsist and update it. If not, creates
```
BigRoy, line 156:
```suggestion
    """This function verifies if the tool exists and updates it. If not, creates
```

openpype/hosts/houdini/api/shelves.py
```diff
@@ -0,0 +1,196 @@
...
+    a new one.
+
+    Arguments:
+        tool_definition {dict} -- Dict with label, script, icon and help
+        shelf {hou.Shelf} -- The parent shelf of the tool
```
BigRoy, line 161:
```suggestion
        tool_definition (dict): Dict with label, script, icon and help
        shelf (hou.Shelf): The parent shelf of the tool
```

Shouldn't the type be inside `()` instead of `{}`?
And shouldn't it be followed by `: ` instead ` -- `

openpype/hosts/houdini/api/shelves.py
```diff
@@ -0,0 +1,196 @@
...
+    ]
+
+    if existing_tool:
+        tool_definition.pop('name', None)
+        tool_definition.pop('label', None)
+        existing_tool[0].setData(**tool_definition)
+        return existing_tool[0]
+
+    tool_name = tool_label.replace(' ', '_').lower()
+    log.warning(tool_definition)
```
BigRoy, line 180: Seems odd to always log a warning in this case?

BigRoy (commented):
Again, nice work. Thanks for the tweaks - here's some follow up notes now. I still feel [my original comment's question is still open for discussion too](https://github.com/pypeclub/OpenPype/pull/3652#pullrequestreview-1072345452).

openpype/hosts/houdini/api/shelves.py
friquette: You are right, I used autogenerated docstrings but I changed to match OpenPype's

antirotor:
If I may add to the discussion - this is probably similar thing (in concept) as the one we already have for Maya/Nuke and that is scriptsmenu support. I agree this is on the verge of OpenPype responsibility, but from my experience it is good to have some UI entry point to expose stuff that might or might not be OpenPype related - that is custom per studio but at the same time bound to some OpenPype functionality so there is some reason to put it and configure it together.

friquette:
> Again, nice work. Thanks for the tweaks - here's some follow up notes now. I still feel [my original comment's question is still open for discussion too](https://github.com/pypeclub/OpenPype/pull/3652#pullrequestreview-1072345452).

As for your original comment, I think this feature has its place here, like the maya script menu or the nuke gizmo menu. But @iLLiCiTiT will surely have a better opinion on this.

iLLiCiTiT:
> But @iLLiCiTiT will surely have a better opinion on this.

I'm glad you've mentioned me, but I'm not the one who should decide. From my point of view it make sense to have prepared settings to give studio option to easily add custom scripts, like in nuke or maya. But at the same time I'm not artist with experiences so I can't tell if it make sense in houdini (and this way) because I've never worked with Houdini (or Maya). So I'll mark @mkolar and @antirotor  to tell their opinions :)

openpype/hosts/houdini/api/shelves.py
```diff
@@ -0,0 +1,195 @@
...
+    shelf_name = shelf_label.replace(' ', '_').lower()
+    new_shelf = hou.shelves.newShelf(
+        name=shelf_name,
+        label=shelf_label
+    )
+    return new_shelf
+
+
+def get_or_create_tool(tool_definition, shelf):
+    """This function verifies if the tool exsists and updates it. If not, creates
```
BigRoy, line 156:
```suggestion
    """This function verifies if the tool exists and updates it. If not, creates
```

openpype/hosts/houdini/api/shelves.py
```diff
@@ -0,0 +1,195 @@
...
+    current_os = platform.system().lower()
+
+    # load configuration of houdini shelves
+    project_settings = get_project_settings(os.getenv("AVALON_PROJECT"))
+    shelves_set_config = project_settings["houdini"]["shelves"]
+
+    if not shelves_set_config:
+        log.info(
+            "SHELF INFO: No custom shelves found in project settings."
+        )
```
BigRoy, line 28: Should this just be a debug message?
Also I steel feel like doing `log.info` and also including "INFO" in the message is kind of saying the same things twice.

I'd expect something like this:
```suggestion
        log.debug("SHELF: No custom shelves found in project settings.")
```

And if the logger at the top would be:
```python
log = logging.getLogger("openpype.hosts.houdini.shelves")
```
(Or anything that at least hints that this logger relates to the shelves)...
Then I'd say the logged message could even be:

```suggestion
        log.debug("No custom shelves found in project settings.")
```

It's the same with the other logged messages that include INFO, WARNING or ERROR in the message itself - shouldn't the logging level (`log.info`, `log.warning`, `log.error`) describe that itself?

mkolar:
I agree with @antirotor. It's on the verge, but we have a precedence with maya shelfs, and scriptsmenu. OpenPype always aimed to come with batteries included and this feels like an extra little battery that an make TDs life easier, so I'm happy to include it.

BigRoy:
Thanks for the follow up everyone - on my questions - and thanks @friquette for the contribution. Solid work!

openpype/hosts/houdini/api/shelves.py
```diff
@@ -0,0 +1,200 @@
...
+            "No custom shelves found in project settings."
+        )
+        return
+
+    for shelf_set_config in shelves_set_config:
+        shelf_set_filepath = shelf_set_config.get('shelf_set_source_path')
+
+        if shelf_set_filepath[current_os]:
+            if not os.path.isfile(shelf_set_filepath[current_os]):
+                raise FileNotFoundError(
```
antirotor, line 36: `FileNotFoundError` is not global in Python 2, so it fails with `NameError: global name 'FileNotFoundError' is not defined` in Houdini 18.5 with py2.

Might be patched with this added somewhere at the top of the file:

```python
import six
if six.PY2: FileNotFoundError = IOError
```

antirotor (changes requested):
Apart that python 2 compatibility thing it's great work!

antirotor (approved):
Thank you for making the changes, merging now.

PR #3654, opened 2022-08-12
Maya: Redshift Skip aov file format check for Cryptomatte

### LINKED ISSUE #3410: Maya: Redshift Validate Render Settings for Cryptomatte reports invalid file format
**Describe the bug**

The code [here](https://github.com/pypeclub/OpenPype/blob/2eab1574eebac97cfd943b2510288f97c7eac857/openpype/hosts/maya/plugins/publish/validate_rendersettings.py#L190-L195) is to validate the extensions/file formats of the AOVs are consistent with the global render file format set.

The Redshift Cryptomatte AOV uses a custom `Cryptomatte (EXR)` file format and thus will always differ from e.g. the global `OpenEXR` file format.

The "repair" method does allow the user to repair and pass the validator. However, Redshift will still enforce the Cryptomatte (EXR) and after repair will actually spew out Maya errors whenever you select the Cryptomatte AOV with the Attribute Editor visible - this is due to that particular AOV not havingt he `OpenEXR` file format entry available.

**To Reproduce**
1. Create a render scene with Redshift
2. Add Cryptomatte AOV
3. Check the errors, perform a repair
4. See the issue described above.

**Expected behavior**

Should just allow the Cryptomatte (EXR) file format to be set for cryptomatte AOVs. 

**Screenshots**

 - OS: Windows
 - Host: Maya 2022 py3 Redshift 3.5.03

[cuID:2hz0tnq]

BigRoy:
## Brief description

This fixes #3410 

## Description

The file format check for AOVs is skipped for Cryptomatte AOVs because Redshift always write it as "Cryptomatte (EXR)" format.

## Testing notes:
1. Set up scene with Redshift + Cryptomatte AOV and others
2. Ensure validation works as expected for file format (extension)

### DIFF
--- a/openpype/hosts/maya/plugins/publish/validate_rendersettings.py
+++ b/openpype/hosts/maya/plugins/publish/validate_rendersettings.py
@@ -108,8 +108,9 @@ class ValidateRenderSettings(pyblish.api.InstancePlugin):
 
         # Get the node attributes for current renderer
         attrs = lib.RENDER_ATTRS.get(renderer, lib.RENDER_ATTRS['default'])
+        # Prefix attribute can return None when a value was never set
         prefix = lib.get_attr_in_layer(cls.ImagePrefixes[renderer],
-                                       layer=layer)
+                                       layer=layer) or ""
         padding = lib.get_attr_in_layer("{node}.{padding}".format(**attrs),
                                         layer=layer)
 
--- a/openpype/hosts/maya/plugins/publish/validate_rendersettings.py
+++ b/openpype/hosts/maya/plugins/publish/validate_rendersettings.py
@@ -201,7 +201,7 @@ class ValidateRenderSettings(pyblish.api.InstancePlugin):
                     labels = get_redshift_image_format_labels()
                     cls.log.error(
                         "AOV file format {} does not match global file format "
-                        "{}".format(labels[default_ext], labels[aov_ext])
+                        "{}".format(labels[aov_ext], labels[default_ext])
                     )
                     invalid = True
 
--- a/openpype/hosts/maya/plugins/publish/validate_rendersettings.py
+++ b/openpype/hosts/maya/plugins/publish/validate_rendersettings.py
@@ -337,8 +337,7 @@ class ValidateRenderSettings(pyblish.api.InstancePlugin):
                 cmds.optionMenuGrp("vrayRenderElementSeparator",
                                    v=instance.data.get("aovSeparator", "_"))
                 cmds.setAttr(
-                    "{}.fileNameRenderElementSeparator".format(
-                        node),
+                    "{}.fileNameRenderElementSeparator".format(node),
                     instance.data.get("aovSeparator", "_"),
                     type="string"
                 )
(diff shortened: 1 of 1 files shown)


openpype/hosts/maya/plugins/publish/validate_rendersettings.py
```diff
@@ -180,14 +180,17 @@ def get_invalid(cls, instance):
                         redshift_AOV_prefix
                     ))
                     invalid = True
-                # get aov format
-                aov_ext = cmds.getAttr(
-                    "{}.fileFormat".format(aov), asString=True)
 
-                default_ext = cmds.getAttr(
-                    "redshiftOptions.imageFormat", asString=True)
```
BigRoy, line 188: I've removed the `asString` argument since the returned values were returned as integers anyway.

BigRoy:
It was mentioned [here](https://github.com/pypeclub/OpenPype/pull/3643#pullrequestreview-1069610148) that the file format error isn't very clear

> **Murphy**: the file format error is not very clear, as the user is not aware of 1 and 6 and what format it actually represents.

We could turn the integer into redshift image format labels (which is what the Attribute Editor shows) or the extensions with this:

```python
from maya import mel

def redshift_get_image_format_ext(index, as_labels=False):
    if as_labels:
        key = "g_redshiftImageFormatLabels"
    else:
        key = "g_redshiftImageFormatExtensions"
    labels = mel.eval("${0}=${0}".format(key))
    return labels[index]
```
These get defined (on Maya launch?) in Redshift's `Plugins/Maya/Common/scripts/redshiftCreateImageFormats.mel`.

antirotor (changes requested):
> Let me know if I should add it in this PR.

I think we should add it even as static map mapping values to strings and showing it to the artist. Yes, I am traumatized by windows errors like `Error 0x0004a2c12 has occured`...

BigRoy:
Improved the logged messages to show the readable labels for the image formats:
afbeelding

---

Additionally squashed some other bugs:
- `cmds.getAttr("defaultRenderGlobals.imageFilePrefix") would return `None` if no value has ever been set, so I added `or ""` to ensure a string value.
- Repair for `defaultRenderGlobals.animation` wasn't implemented - it now is.

antirotor (approved):
I think it fine now, thanks for that!. Before merging, I would like @m-u-r-p-h-y  to test it too as my current setup is pretty much broken on Redshift side.

m-u-r-p-h-y:
where is the global file format taken from?

image

BigRoy:
This validation checks the AOV file format of the **AOV** where the "global setting" is the setting on the Layer itself.
So basically you have:

- OpenPype Settings: Output Image Format - _unrelated to this error actually_
- Redshift Render Settings > Common > Image Format - this is the currently set 'global' file format for the layer.
- Redshift Render Settings > Advanced > AOV - each AOV also has a file format of its own. That is the AOV file format.

m-u-r-p-h-y:
From my point of view, the wording is actually opposite.

OP settings is set to TIFF
Render Globals is set to TIFF
Beauty AOV is set OpenExr
Cryptomatte is set to OpenExr

so:
AOV file format OpenExr is not matching Global settings TIFF (OP + Render settings)

BigRoy:
You were right - the formatting was swapped. Fixed.

m-u-r-p-h-y (approved):
the check is correctly skipping cryptomatte file format check with Redshift renderer

image

PR #3656, opened 2022-08-12
General: Fix finding of last version

antirotor:
## Bug

There was a bug that affected finding of last OP version. This resulted in invalid behaviour mainly in headless mode, where it tries to reinstall already installed version.

### DIFF
--- a/igniter/bootstrap_repos.py
+++ b/igniter/bootstrap_repos.py
@@ -411,16 +411,7 @@ class OpenPypeVersion(semver.VersionInfo):
 
         # DEPRECATED: backwards compatible way to look for versions in root
         dir_to_search = Path(user_data_dir("openpype", "pypeclub"))
-        versions = OpenPypeVersion.get_versions_from_directory(
-            dir_to_search, compatible_with=compatible_with
-        )
-        if compatible_with:
-            dir_to_search = Path(
-                user_data_dir("openpype", "pypeclub")) / f"{compatible_with.major}.{compatible_with.minor}"  # noqa
-            versions += OpenPypeVersion.get_versions_from_directory(
-                dir_to_search, compatible_with=compatible_with
-            )
-
+        versions = OpenPypeVersion.get_versions_from_directory(dir_to_search)
 
         filtered_versions = []
         for version in versions:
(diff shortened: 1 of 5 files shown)


BigRoy:
This would require deploying new executables/installers, yes? :)

Were you able to reproduce the original issue of it trying to do a force install in headless mode? And did this avoid it on your end? Just wondering how "safe" it is for me to just throw this PR into production to see if it'd would fix the issue or whether this might do more harm.

antirotor:
This still needs little refactor to move the compatibility check out...

antirotor:
> This would require deploying new executables/installers, yes? :)

yes, this will need new build


> Were you able to reproduce the original issue of it trying to do a force install in headless mode? And did this avoid it on your end? Just wondering how "safe" it is for me to just throw this PR into production to see if it'd would fix the issue or whether this might do more harm.

I haven't reproduced it, but find similar issue with headless mode that lead me to this. The re-install logic is sound I think, problem is that it can't find the latest version correctly so it is trying to reinstall it.

But please wait with deployment until the compatibility check is refactored.

BigRoy:
> But please wait with deployment until the compatibility check is refactored.

If you could ping me when that's the case, please do so! :)

igniter/bootstrap_repos.py
```diff
@@ -515,17 +503,27 @@ def get_versions_from_directory(
...
+                          item.name.split(".")[
+                              1] == installed_version.minor  # noqa: E051
+                if item.is_dir() and ver_dir:
+                    _versions = OpenPypeVersion.get_versions_from_directory(
+                        item)
+                    if _versions:
+                        openpype_versions.append(_versions)
+            except IndexError:
+                pass
+            # if file exists, strip extension, in case of dir don't.
```
BigRoy, line 526: Wouldn't the `item.name` match exactly with `f"{installed_version.major}.{installed_version.minor}"`?
So wouldn't this just be `if item.name == f"{installed_version.major}.{installed_version.minor}"` or alike?

```
major_minor = f"{installed_version.major}.{installed_version.minor}"
if item.is_dir() and item.name == major_minor:
    _versions = OpenPypeVersion.get_versions_from_directory(item)
    openpype_versions.extend(_versions)
```

And shouldn't we also "continue" if we've already identified it as a major minor subfolder, so add
```
    continue
```

igniter/bootstrap_repos.py
```diff
@@ -629,11 +615,7 @@ def get_latest_version(
 
         all_versions.sort()
         latest_version: OpenPypeVersion
```
iLLiCiTiT, line 617:
```suggestion
```

BigRoy:
We have deployed this in production now. (Since this requires the build/install to be updated too I've actually batch removed the old install and then batch deployed the new to make sure it's using this newer build logic.)

So far so good in Deadline. If anything pops up with errors I'll report back.

antirotor:
> Added UI message if installed version is not compatible with zip and is not launched in headless mode.

It works.

PR #3658, opened 2022-08-12
Scene Inventory: Add subsetGroup column

BigRoy:
## Brief description

This adds a `subsetGroup` column to the Scene Inventory / Manager.

## Description

We tend to group publishes in the Loader (shortcut: `ctrl + g`) to sometimes label things as "extras" or even group them to be "deprecated".
By showing the column in the Manager this is now also much easier to trace to that information.

afbeelding

## Additional info

Code might be a bit messy - so please double check. 

The "icon" for the group is the default group icon. Originally in Avalon groups could be configured on the project to allow custom icons per groups, etc. but that's currently not supported in OpenPype as far as I know. Thus I had chosen for now to just hardcode the icon.

This would be extra useful as soon as better UX for filtering is added to tools - then we could also filter by subset group, etc.

## Testing notes:
1. Load some content with and without subset groups.
2. Check the scene inventory

### DIFF
--- a/openpype/tools/sceneinventory/model.py
+++ b/openpype/tools/sceneinventory/model.py
@@ -34,7 +34,8 @@ from .lib import (
 class InventoryModel(TreeModel):
     """The model for the inventory"""
 
-    Columns = ["Name", "version", "count", "family", "loader", "objectName"]
+    Columns = ["Name", "version", "count", "family",
+               "subsetGroup", "loader", "objectName"]
 
     OUTDATED_COLOR = QtGui.QColor(235, 30, 30)
     CHILD_OUTDATED_COLOR = QtGui.QColor(200, 160, 30)
@@ -157,8 +158,13 @@ class InventoryModel(TreeModel):
                 # Family icon
                 return item.get("familyIcon", None)
 
+            column_name = self.Columns[index.column()]
+
+            if column_name == "subsetGroup" and item.get("subsetGroup"):
+                return qtawesome.icon("fa.object-group",
+                                      color=get_default_entity_icon_color())
+
             if item.get("isGroupNode"):
-                column_name = self.Columns[index.column()]
                 if column_name == "active_site":
                     provider = item.get("active_site_provider")
                     return self._site_icons.get(provider)
@@ -423,6 +429,7 @@ class InventoryModel(TreeModel):
             group_node["familyIcon"] = family_icon
             group_node["count"] = len(group_items)
             group_node["isGroupNode"] = True
+            group_node["subsetGroup"] = subset["data"].get("subsetGroup")
 
             if self.sync_enabled:
                 progress = get_progress_for_repre(
--- a/openpype/tools/sceneinventory/model.py
+++ b/openpype/tools/sceneinventory/model.py
@@ -35,7 +35,7 @@ class InventoryModel(TreeModel):
     """The model for the inventory"""
 
     Columns = ["Name", "version", "count", "family",
-               "subsetGroup", "loader", "objectName"]
+               "group", "loader", "objectName"]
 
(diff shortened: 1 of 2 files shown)


openpype/tools/sceneinventory/window.py
```diff
@@ -88,7 +88,8 @@ def __init__(self, parent=None):
         view.setColumnWidth(1, 55)   # version
         view.setColumnWidth(2, 55)   # count
         view.setColumnWidth(3, 150)  # family
-        view.setColumnWidth(4, 100)  # namespace
```
BigRoy, line 91: FYI: Namespace column didn't exist - this was actually setting the `loader` column. I felt the default width for the loader column was slightly too small and most loader names got cut off. So I've set the loader column width to 150.

iLLiCiTiT (approved):
Code looks ok and works. Can someone check it from UX perspective ( @mkolar @antirotor )?

mkolar (approved):
I would personally just name it `Group` rather than `subsetGroup`. 

In general It's a good idea, however it's becoming painfully obvious that we'll need to implement hiding and unhiding of columns. For example most of clients on support don't actually use this grouping that much and for those it's suddenly and empty column that's there all the time and is never really populated. Hence needless confusion. 

That's more food for though though and I'm happy with merging this.

openpype/tools/sceneinventory/model.py
```diff
@@ -34,7 +34,8 @@
 class InventoryModel(TreeModel):
     """The model for the inventory"""
 
-    Columns = ["Name", "version", "count", "family", "loader", "objectName"]
+    Columns = ["Name", "version", "count", "family",
+               "subsetGroup", "loader", "objectName"]
```
mkolar, line 38:
```suggestion
               "group", "loader", "objectName"]
```

openpype/tools/sceneinventory/model.py
```diff
@@ -157,8 +158,13 @@ def data(self, index, role):
                 # Family icon
                 return item.get("familyIcon", None)
 
+            column_name = self.Columns[index.column()]
+
+            if column_name == "subsetGroup" and item.get("subsetGroup"):
```
mkolar, line 163:
```suggestion
            if column_name == "group" and item.get("group"):
```

openpype/tools/sceneinventory/model.py
```diff
@@ -423,6 +429,7 @@ def add_items(self, items, parent=None):
             group_node["familyIcon"] = family_icon
             group_node["count"] = len(group_items)
             group_node["isGroupNode"] = True
+            group_node["subsetGroup"] = subset["data"].get("subsetGroup")
```
mkolar, line 432:
```suggestion
            group_node["group"] = subset["data"].get("subsetGroup")
```

mkolar (changes requested):
Let's please tweak the name of the column, then we can merge

BigRoy:
@mkolar How about solely updating the header column's name but keeping it the same in the code? :) Somehow I feel `subsetGroup` keeps it linked to the original data attribute name and describes what it is exactly. Or do you feel `group` is clear enough and one should now that these relate to the subset instead of the version or representation?

openpype/tools/sceneinventory/window.py
```diff
@@ -88,7 +88,8 @@ def __init__(self, parent=None):
...
+        view.setColumnWidth(4, 120)  # subsetGroup
```
BigRoy, line 91:
```suggestion
        view.setColumnWidth(4, 120)  # group
```

iLLiCiTiT:
@mkolar are you ok with current state?

antirotor (approved):

mkolar (approved):

PR #3659, opened 2022-08-12
Scene Inventory: Maya add actions to select from or to scene

BigRoy:
## Brief description

Ever wanted to quickly identify some containers that are currently loaded into the scene? Now you can!

## Description

- **Select in scene**:  Select nodes in the scene from selected containers in scene inventory
- **Select from scene**: Select containers in scene inventory from the current scene selection

containers_maya_select

## Additional info

This now works because https://github.com/pypeclub/OpenPype/pull/3657 is merged.

Note: that this will _not_ select or consider the top group because it is not returned by `get_container_members` (that is e.g. the non-referenced top group where references get added into by default)

I added quite a high order to the actions so they'll always appear near the bottom.

## Testing notes:

1. Run maya
2. Load some content
3. Try the right click actions in the Scene Inventory


openpype/hosts/maya/plugins/inventory/select_containers.py
```diff
@@ -0,0 +1,46 @@
...
+    color = "#888888"
+    order = 100
+
+    def process(self, containers):
+
+        selection = set(cmds.ls(selection=True, long=True, objectsOnly=True))
+        host = registered_host()
+
+        to_select = []
+        for container in host.ls():
```
iLLiCiTiT, line 38:
```suggestion
        for container in host.get_containers():
```

BigRoy: Thanks, works!

tokejepsen:
Maybe it's better to have some more elaborate labels? Just from the labels I'm getting confused as to what they do.
"Select nodes in scene" and "Select containers in scene"

BigRoy:
> Maybe it's better to have some more elaborate labels? Just from the labels I'm getting confused as to what they do.
"Select nodes in scene" and "Select containers in scene"

That's a good remark - but I just couldn't come up with the right label. I think "Select containers in scene" sounds wrong since it's not selecting inside maya but it instead would be selecting the container entries in the Scene Inventory UI.

The best description I could still think of was this:

> **Select in scene:** Select nodes in the scene from selected containers in scene inventory
> **Select from scene:** Select containers in scene inventory from the current scene selection

I think this would be somewhat clearer - but it remains unclear 'based on what' it'd decide to select.

1. "Select nodes in scene" 
2. "Select containers in inventory"

I think 1) is somewhat clear (also in UX because you're right clicking on the containers) but 2) just remains odd UX without a better label.
Other ideas:

- "Match scene selection" (match it where?)
- "Match container selection by scene selection" (too long and still not 100% clear)

Anyone has better ideas?

iLLiCiTiT:
First shot ideas `Selection From scene` or `Mark Scene selection`.

mkolar:
I would keep the first one as `Select in scene`. Many maya users are not primarily familiar with nodes in general. Like animators, quite often. `Select in scene` seems more than clear to me. 

The other one is trickier..... 1. it should be a button in the gui and not and action on a selected container. That's I believe would clear up the confusion and if it was just next to the buttons on the top, we could have it as `Highlight by scene selection`

BigRoy:
> The other one is trickier..... It should be a button in the gui and not and action on a selected container. That's I believe would clear up the confusion and if it was just next to the buttons on the top, we could have it as Highlight by scene selection

It's trickier due to this being a host specific implementation and we might not want to expose it for all hosts? (For some it might even be irrelevant, etc.)

How about also allowing right click actions to be added that are not specific to the selection (and thus also work in an empty space in the Scene Inventory) and appear outside of the `Actions` menu? :)

For this first draft we could also just make them:

1. Select in scene 
2. Highlight by scene selection

Then we can look into more streamlining separate from this PR?

BigRoy:
Let me know if there's a verdict and I'll see if I can update the actions to match along with the feedback.

iLLiCiTiT:
> It's trickier due to this being a host specific implementation and we might not want to expose it for all hosts? (For some it might even be irrelevant, etc.)

This is unfortunatelly true and is partially connected to https://github.com/pypeclub/OpenPype/issues/3467 which could make it available. For now I agree with
```
For this first draft we could also just make them:

    1. Select in scene
    2. Highlight by scene selection
```

antirotor (approved):
Works even in totally broken scene 🤣

PR #3661, opened 2022-08-14
Maya: Disable rendering of all lights for render instances submitted through Deadline.

### LINKED ISSUE #2241: Maya: Disable rendering of all lights - Render Setup / Deadline

Allan-I:
## Brief description
Fixes #2241 by adding flag "RenderSetupIncludeLights" to deadline job information.

## Description
Disables rendering all lights in the scene to conform with render layers, such that all the lights in the scene aren't automatically rendered

## Additional info
Previously, jobs submitted through deadline would render all the lights because of the lack of the `RenderSetupIncludeLights` flag.  This PR fixes the issue by adding that flag to the project settings to be a global variable, that can be overridden through the created render instance in the `Extra Attributes` tab in the attribute editor in Maya.  

The old PR (found here: https://github.com/pypeclub/OpenPype/pull/3038) was discarded in favor of this one.

## Testing notes:
1. Create multiple lights in the scene.
2. Create geometry.
3. Create render instance selecting only one light and the geometry.
4. Create another render layer selecting another light and the geometry through the render setup panel in Maya (Windows->Rendering Editors->Render Setup).
5. Make sure the layers each contain collection that includes the geometry and the intended light, remove any wildcards (*) that exist.
6. Publish.

The jobs submitted should have the flag disabled as `False` in the `Job Properties`.


m-u-r-p-h-y (commented):
Can you pls merge latest develop to fix the previous bugs preventing the testing of this PR?

thx!

image

Allan-I:
> 

You got it, Murphy, done!

openpype/hosts/maya/plugins/create/create_render.py
```diff
@@ -71,7 +71,7 @@ class CreateRender(plugin.Creator):
     label = "Render"
     family = "rendering"
     icon = "eye"
-
+    enable_all_lights = True
```
BigRoy, line 74: Is this attribute unused?

m-u-r-p-h-y (approved):
Followed the testing instructions and submitted a scene with two render layers and
Deadline job attribute is correctly set to False in Job Properties

`RenderSetupIncludeLights=False`

image

openpype/hosts/maya/plugins/create/create_render.py
BigRoy:
```suggestion
```
I suspect this attribute is unused and can be removed.

openpype/hosts/maya/plugins/create/create_render.py
```diff
@@ -220,6 +220,12 @@ def _create_render_settings(self):
         self.data["tilesY"] = 2
         self.data["convertToScanline"] = False
         self.data["useReferencedAovs"] = False
+        self.data["renderSetupIncludeLights"] = (
+            self._project_settings.get(
+                "maya", {}).get(
+                "RenderSettings", {}).get(
+                "enable_all_lights", {})
```
BigRoy, line 227:
```suggestion
                "enable_all_lights", False)
```

This should not default to a dict if `enable_all_lights` isn't found because a `dict` isn't settable on the created node and also is the wrong type?

openpype/hosts/maya/plugins/publish/validate_rendersettings.py
```diff
@@ -242,6 +242,14 @@ def get_invalid(cls, instance):
             instance.context.data["project_settings"]["maya"]["publish"]["ValidateRenderSettings"].get(  # noqa: E501
                 "{}_render_attributes".format(renderer)) or []
         )
+        settings_lights_flag = instance.context.data["project_settings"].get(
+            "maya", {}).get(
+            "RenderSettings", {}).get(
+            "enable_all_lights", {})
```
BigRoy, line 248: Is there a reason why even the `enable_all_lights` get should default to `{}` even though the value should be a boolean. I'd expect something like this:
```suggestion
            "enable_all_lights", False)
```

Or just:
```suggestion
            "enable_all_lights")
```

openpype/hosts/maya/plugins/create/create_render.py
Allan-I: Ah, good catch, thank you Roy

Allan-I:
Thank you Roy!

BigRoy (approved):
Solid work! (Haven't tested the code, but code changes look good to me)

m-u-r-p-h-y:
do I need to redo the test?

Allan-I:
> do I need to redo the test?

I don't think so!

PR #3463, opened 2022-07-01
Blender: Dynamic workflow using linked libraries and overrides

kaamaurice:
### Brief description
- Use collections and overrides for the instances.

### Description
- Change Blender plugins to work with a collection instead of an empty to store custom properties of the Avalon instance.

- Discussion : https://github.com/pypeclub/OpenPype/discussions/2598

- Based on the work of [jojoba106](https://github.com/jojoba106)
https://github.com/pypeclub/OpenPype/pull/2771

- Squashed and rebase from : https://github.com/pypeclub/OpenPype/pull/3171

Resolve #2890

PR #3662, opened 2022-08-15
Blender: validators code correction with settings and defaults

kaamaurice:
## Brief description
Blender Validators for publishing check with more conditions and are setupable with Project Settings.

## Description
- Validators check instance members for correct object type.
- Settings and defaults for OP project settings added to control validators for publishing.
- category, order, logger messages rectified.

### DIFF
--- a/openpype/hosts/blender/plugins/publish/validate_camera_zero_keyframe.py
+++ b/openpype/hosts/blender/plugins/publish/validate_camera_zero_keyframe.py
@@ -1,9 +1,11 @@
 from typing import List
 
 import mathutils
+import bpy
 
 import pyblish.api
-import openpype.hosts.blender.api.action
+from openpype.api import ValidateContentsOrder
+from openpype.hosts.blender.api.action import SelectInvalidAction
 
 
 class ValidateCameraZeroKeyframe(pyblish.api.InstancePlugin):
@@ -14,21 +16,21 @@ class ValidateCameraZeroKeyframe(pyblish.api.InstancePlugin):
     in Unreal and Blender.
     """
 
-    order = openpype.api.ValidateContentsOrder
+    order = ValidateContentsOrder
     hosts = ["blender"]
     families = ["camera"]
     category = "geometry"
     version = (0, 1, 0)
     label = "Zero Keyframe"
-    actions = [openpype.hosts.blender.api.action.SelectInvalidAction]
+    actions = [SelectInvalidAction]
 
     _identity = mathutils.Matrix()
 
-    @classmethod
-    def get_invalid(cls, instance) -> List:
+    @staticmethod
+    def get_invalid(instance) -> List:
         invalid = []
-        for obj in [obj for obj in instance]:
-            if obj.type == "CAMERA":
+        for obj in set(instance):
+            if isinstance(obj, bpy.types.Object) and obj.type == "CAMERA":
                 if obj.animation_data and obj.animation_data.action:
                     action = obj.animation_data.action
                     frames_set = set()
@@ -45,4 +47,5 @@ class ValidateCameraZeroKeyframe(pyblish.api.InstancePlugin):
         invalid = self.get_invalid(instance)
         if invalid:
             raise RuntimeError(
-                f"Object found in instance is not in Object Mode: {invalid}")
+                f"Camera must have a keyframe at frame 0: {invalid}"
+            )
--- a/openpype/hosts/blender/plugins/publish/validate_camera_zero_keyframe.py
+++ b/openpype/hosts/blender/plugins/publish/validate_camera_zero_keyframe.py
@@ -1,11 +1,10 @@
 from typing import List
 
-import mathutils
 import bpy
 
 import pyblish.api
-from openpype.api import ValidateContentsOrder
-from openpype.hosts.blender.api.action import SelectInvalidAction
+import openpype.api
+import openpype.hosts.blender.api.action
 
 
 class ValidateCameraZeroKeyframe(pyblish.api.InstancePlugin):
@@ -16,18 +15,17 @@ class ValidateCameraZeroKeyframe(pyblish.api.InstancePlugin):
     in Unreal and Blender.
     """
 
-    order = ValidateContentsOrder
+    order = openpype.api.ValidateContentsOrder
     hosts = ["blender"]
     families = ["camera"]
-    category = "geometry"
     version = (0, 1, 0)
     label = "Zero Keyframe"
-    actions = [SelectInvalidAction]
+    actions = [openpype.hosts.blender.api.action.SelectInvalidAction]
 
     @staticmethod
     def get_invalid(instance) -> List:
         invalid = []
-        for obj in set(instance):
+        for obj in instance:
             if isinstance(obj, bpy.types.Object) and obj.type == "CAMERA":
                 if obj.animation_data and obj.animation_data.action:
                     action = obj.animation_data.action
--- a/openpype/hosts/blender/plugins/publish/validate_camera_zero_keyframe.py
+++ b/openpype/hosts/blender/plugins/publish/validate_camera_zero_keyframe.py
@@ -24,8 +24,6 @@ class ValidateCameraZeroKeyframe(pyblish.api.InstancePlugin):
     label = "Zero Keyframe"
     actions = [SelectInvalidAction]
 
-    _identity = mathutils.Matrix()
-
     @staticmethod
     def get_invalid(instance) -> List:
         invalid = []
--- a/openpype/hosts/blender/plugins/publish/validate_mesh_has_uv.py
+++ b/openpype/hosts/blender/plugins/publish/validate_mesh_has_uv.py
@@ -13,7 +13,7 @@ class ValidateMeshHasUvs(pyblish.api.InstancePlugin):
     order = ValidateContentsOrder
     hosts = ["blender"]
     families = ["model"]
-    category = "geometry"
+    category = "uv"
     label = "Mesh Has UV's"
     actions = [SelectInvalidAction]
     optional = True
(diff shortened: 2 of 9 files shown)


openpype/hosts/blender/plugins/publish/validate_camera_zero_keyframe.py
```diff
@@ -1,9 +1,11 @@
 from typing import List
 
 import mathutils
+import bpy
 
 import pyblish.api
-import openpype.hosts.blender.api.action
+from openpype.api import ValidateContentsOrder
+from openpype.hosts.blender.api.action import SelectInvalidAction
```
BigRoy, line 8: I'm not sure this code style change is the best move. Especially since the majority of the OpenPype code base doesn't seem to import these like these for the other hosts, etc.

openpype/hosts/blender/plugins/publish/validate_mesh_has_uv.py
```diff
@@ -3,18 +3,19 @@
...
 
 class ValidateMeshHasUvs(pyblish.api.InstancePlugin):
     """Validate that the current mesh has UV's."""
 
-    order = pyblish.api.ValidatorOrder
+    order = ValidateContentsOrder
     hosts = ["blender"]
     families = ["model"]
-    category = "geometry"
+    category = "uv"
```
BigRoy, line 16: This remark isn't necessarily about this change - it's more a general comment regarding the `category` attribute.

I believe `category` is actually an unused attribute across the tooling. Unless the new publishers use this? I'm wondering whether we should just clean up all these `category` definitions.

@mkolar do you know more?

openpype/hosts/blender/plugins/publish/validate_mesh_has_uv.py
```diff
@@ -33,20 +34,21 @@ def has_uvs(obj: bpy.types.Object) -> bool:
     @classmethod
     def get_invalid(cls, instance) -> List:
         invalid = []
-        # TODO (jasper): only check objects in the collection that will be published?
-        for obj in [
-            obj for obj in instance]:
+        for obj in set(instance):
```
BigRoy, line 37: When would the `instance` not have unique entries? I'd say maybe we should solve that instead of making use of `set` here? I suspect this could be slower for many instances with many nodes?

openpype/hosts/blender/plugins/publish/validate_mesh_has_uv.py
```diff
@@ -33,20 +34,21 @@ def has_uvs(obj: bpy.types.Object) -> bool:
...
             try:
-                if obj.type == 'MESH':
-                    # Make sure we are in object mode.
-                    bpy.ops.object.mode_set(mode='OBJECT')
+                if isinstance(obj, bpy.types.Object) and obj.type == 'MESH':
+                    if obj.mode != 'OBJECT':
+                        # Make sure we are in object mode.
+                        bpy.ops.object.mode_set(mode='OBJECT')
```
BigRoy, line 42: Somehow I feel this check should be outside the for loop since it looks like we're setting the object mode globally? Couldn't we query that globally as well? (This isn't necessarily a comment about the code change - it was there originally too)

openpype/hosts/blender/plugins/publish/validate_mesh_no_negative_scale.py
```diff
@@ -3,29 +3,27 @@
...
-            obj for obj in bpy.data.objects if obj.type == 'MESH'
-        ]:
-            if any(v < 0 for v in obj.scale):
-                invalid.append(obj)
-
-        return invalid
+        for obj in set(instance):
+            if isinstance(obj, bpy.types.Object) and obj.type == 'MESH':
+                if any(v < 0 for v in obj.scale):
+                    invalid.append(obj)
```
BigRoy, line 26: Resolves the original #todo. Much better - nice!

openpype/hosts/blender/plugins/publish/validate_no_colons_in_name.py
```diff
@@ -12,20 +15,21 @@ class ValidateNoColonsInName(pyblish.api.InstancePlugin):
...
         invalid = []
-        for obj in [obj for obj in instance]:
+        for obj in set(instance):
             if ':' in obj.name:
                 invalid.append(obj)
-            if obj.type == 'ARMATURE':
+            if isinstance(obj, bpy.types.Object) and obj.type == 'ARMATURE':
                 for bone in obj.data.bones:
                     if ':' in bone.name:
                         invalid.append(obj)
```
BigRoy, line 35: I have no knowledge about how the blender logic works for OpenPype. But in Maya terminology the "history" of the instance (like its bones for a rig) are included into the `instance` as nodes/objects. Wouldn't that make more sense here too? So that we wouldn't have to traverse into the data armatures to begin with and collect its history / armature bones in a collector? (Again, not a comment about the code change - sorry!)

openpype/hosts/blender/plugins/publish/validate_object_mode.py
```diff
@@ -10,22 +12,18 @@ class ValidateObjectIsInObjectMode(pyblish.api.InstancePlugin):
...
-                if obj.type == 'MESH' or obj.type == 'ARMATURE':
-                    # Check if the object is in object mode.
-                    if not obj.mode == 'OBJECT':
-                        invalid.append(obj)
-            except Exception:
-                continue
+        for obj in set(instance):
+            if isinstance(obj, bpy.types.Object):
+                if not obj.mode == 'OBJECT':
+                    invalid.append(obj)
```
BigRoy, line 26: ❤️ Nice cleanup compared to the original code.

I suspect that now we might be checking more objects than before. So technically this might give different results in production?

BigRoy (commented):
Nice work. Sorry for commenting some things that aren't directly related to your code changes. I've commented some things that stood out to me. (Note that I don't use Blender so please take them with a grain of salt if I'm saying stupid things!)

openpype/hosts/blender/plugins/publish/validate_no_colons_in_name.py
kaamaurice: Hi ! Blender has very atypical way to construct rigs. Bones are not objects but rather sub-elements of an armature type object. But this could be possible to append `bpy.type.Bone` in `collect_instances.py` as members now that we check members classes as `bpy.types.Object` in all Validators how needed it.

openpype/hosts/blender/plugins/publish/validate_mesh_has_uv.py
kaamaurice: Yes It will be absolutely true with #3463
Maybe not totaly true with this : [collect_instances.py : line 99](https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/blender/plugins/publish/collect_instances.py#L99)
But we can simplify the loops without `set()`

openpype/hosts/blender/plugins/publish/validate_camera_zero_keyframe.py
kaamaurice: Ok, agree, thanks for the reviews :)

openpype/hosts/blender/plugins/publish/validate_mesh_has_uv.py
BigRoy: I'd say update the Collector to make to do `list(set(members))` once so that all validators, etc. benefit from it. But I'll leave it up to you whether that's sensible in the context of blender.

openpype/hosts/blender/plugins/publish/validate_no_colons_in_name.py
BigRoy: I see - if they aren't objects like that I can see how this might not be the most sensible way to embed it as history. Thanks.

kaamaurice: Yes actually Extractors could raise an `AttributeError` if we add `bpy.types.Bone`: [extract_blend.py#L31](https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/blender/plugins/publish/extract_blend.py#L31)

openpype/hosts/blender/plugins/publish/validate_mesh_has_uv.py
mkolar: To be honest I personally completely forgot this was there, and I agree that it should be investigated and tackled across the board. It doesn't hurt here, but doesn't really add anything either.

openpype/hosts/blender/plugins/publish/validate_object_mode.py
kaamaurice: bpy.types.Object type could be `‘MESH’, ‘CURVE’, ‘SURFACE’, ‘META’, ‘FONT’, ‘CURVES’, ‘POINTCLOUD’, ‘VOLUME’, ‘GPENCIL’, ‘ARMATURE’, ‘LATTICE’, ‘EMPTY’, ‘LIGHT’, ‘LIGHT_PROBE’, ‘CAMERA’, ‘SPEAKER’`
All of then could and should be in mode `'OBJECT'` and most of them have only `'OBJECT'` mode so results should be good for all cases.

openpype/hosts/blender/plugins/publish/validate_mesh_has_uv.py
kaamaurice: Yes absolutely and I'm not happy with this ops and I think it should stay as it to be consistent with the Object mode Validator.

openpype/hosts/blender/plugins/publish/validate_mesh_has_uv.py
kaamaurice: Ok, I leave the code as it is for `category = "geometry"` attributes in validators if there are consistent, but removed it for inconsistent as in camera validators.

simonebarbieri (approved):

PR #3666, opened 2022-08-15
Deadline: Global job pre load is not Pype 2 compatible

iLLiCiTiT:
## Brief description
OpenPype deadline global plugin is not Pype 2 compatible anymore.

## Description
Because of pype 2 compatibility the deadline prelaunch hook is always triggered which cause issues with jobs that are not for OpenPype because in that case Pype 2 logic starts.

Side change is to make the plugin Python 2 compatible.

## Testing notes:
1. Copy the plugin to deadline repository
2. Try submit a job without openpype
3. It should not crash

Resolves https://github.com/pypeclub/OpenPype/issues/3664


BigRoy (approved):
This looks similar we had done to our deployments - so looks good to me.

_Haven't tested this particular PR_

m-u-r-p-h-y (approved):
image

submission of none OP job works as expected

PR #3667, opened 2022-08-15
TrayPublisher: Removed from experimental tools

iLLiCiTiT:
## Brief description
Tray publisher is added to tray menu by default and artist don't have to enable the experimental option.

## Description
Tray Publisher is now in ready to use state and should work at least the same way as standalone publisher did. We should collect as much input from artists as possible thus the tool is removed from experimental tools and is added as available feature in tray.

## Additional information
Changed label to `Tray Publish` so it is possible to differentiate in standalone publisher and tray publisher.

## Testing notes:
1. Tray publisher is visible in tray menu without enabling it in local settings
2. Tray publisher is not visible in local settings


kalisp (approved):
Screenshot 2022-08-17 130031

Tool is opening up.

mkolar:
This is confusing. If we take it out of experimental, we need to at the same time hide the old one and make that optional. Two publishers at the same time is a road to hell I'm afraid. Do we now have 1-1 functionality match?

iLLiCiTiT:
> Do we now have 1-1 functionality match?

There is not 100% but at the same time we should add an easy access to use tray publisher so somebody in production can test what we have there now.

We can add standalone publisher and tray publisher into module settings so they can be turned enabled/disabled. Also we can change label of standalone publisher action.

m-u-r-p-h-y:
It is not clear what should happen next.

current `Publish` to be renamed and hidden? (Legacy Publish and create settings to put it back on, or?)

current `Tray Publish`  renamed to Publish (or keep Tray Publish? or?)

mkolar:
I would remove the old publisher and keep the new one just called `Publisher` (to keep in liner with the `loader` and `launcher` rather than `load` and `launch`). I agree that we should keep the option to turn the old publisher back on somewhere in the settings. In which case the old should be called `Publisher (legacy)`

iLLiCiTiT:
Standalone publisher is disabled by default (can be enabled) and has label `Publisher (legacy)`. Tray publisher has label `Publisher`.

64qam (approved):

PR #3670, opened 2022-08-15
Blender: Publisher collect workfile representation

kaamaurice:
## Brief description
Like `collect_workfile` behavior from Maya host, Blender should ba able to publish workfiles.

## Description
- I have updated `collect_current_file.py` with the code from maya plugin `collect_workfile` to be able to publish workfile representation.
- Action are available to save file if needed.

### DIFF
--- a/openpype/hosts/blender/plugins/publish/collect_current_file.py
+++ b/openpype/hosts/blender/plugins/publish/collect_current_file.py
@@ -13,11 +13,7 @@ class SaveWorkfiledAction(pyblish.api.Action):
     icon = "save"
 
     def process(self, context, plugin):
-        current_file = workio.current_file()
-        if current_file:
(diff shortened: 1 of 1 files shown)


openpype/hosts/blender/plugins/publish/collect_current_file.py
```diff
@@ -1,19 +1,82 @@
...
+        has_unsaved_changes = workio.has_unsaved_changes()
+
+        context.data["currentFile"] = current_file
 
-        assert current_file != '', "Current file is empty. " \
+        assert current_file, (
+            "Current file is empty. Save the file before continuing."
+        )
+
+        assert not has_unsaved_changes, (
```
iLLiCiTiT, line 42: I would say that validations should happen in validator plugin.

PR #3671, opened 2022-08-15
Ftrack: More logs related to auto sync value change

iLLiCiTiT:
## Brief description
More information in logs when autosync is turned on/off.

## Description
Log information than auto sync was turned off and added infromation about username who triggered it.

## Testing notes:
1. Start ftrack event server
2. Change auto sync on a project
3. A log describing who turned on/off auto sync on which project should appear in process of ftrack event server


kalisp (approved):
Screenshot 2022-08-19 113230

Works (with --debug).

kalisp:
Actually wouldn't be beneficial to have there timestamp? But no other logs have timestamps, so idk.

iLLiCiTiT:
> Actually wouldn't be beneficial to have there timestamp? But no other logs have timestamps, so idk.

This is mostly checked in log viewer tool which has timestamps available.

PR #3672, opened 2022-08-16
Loader: Fix representation check per version logic

BigRoy:
## Brief description

Fix logic that reports whether a representation was found for the versions or not.

## Description

Noticed an odd if statement which looked like a bug:
```python
if version_id not in version_ids:
    version_ids.remove(version_id)
```

Which in the end resulted in all loaders to always log this:

> Subsets "subset" don't have representation 'representation'

So ended up investigating and refactoring the code.

## Testing notes:
1. Loaders should work as expected and report when representation isn't found.

### DIFF
--- a/openpype/tools/loader/widgets.py
+++ b/openpype/tools/loader/widgets.py
@@ -584,9 +584,9 @@ class SubsetWidget(QtWidgets.QWidget):
             for repre_doc in repre_docs:
                 repre_ids.append(repre_doc["_id"])
 
+                # keep only version ids without representation with that name
                 version_id = repre_doc["parent"]
-                if version_id not in version_ids:
-                    version_ids.remove(version_id)
+                version_ids.remove(version_id)
 
             for version_id in version_ids:
                 joined_subset_names = ", ".join([
--- a/openpype/tools/loader/widgets.py
+++ b/openpype/tools/loader/widgets.py
@@ -586,7 +586,7 @@ class SubsetWidget(QtWidgets.QWidget):
 
                 # keep only version ids without representation with that name
                 version_id = repre_doc["parent"]
-                version_ids.remove(version_id)
+                version_ids.discard(version_id)
 
             if version_ids:
                 # report versions that didn't have valid representation
--- a/openpype/tools/loader/widgets.py
+++ b/openpype/tools/loader/widgets.py
@@ -567,12 +567,12 @@ class SubsetWidget(QtWidgets.QWidget):
 
             # Trigger
             project_name = self.dbcon.active_project()
-            subset_names_by_version_id = collections.defaultdict(set)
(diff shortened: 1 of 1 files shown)


openpype/tools/loader/widgets.py
```diff
@@ -567,12 +567,12 @@ def on_context_menu(self, point):
 
             # Trigger
             project_name = self.dbcon.active_project()
-            subset_names_by_version_id = collections.defaultdict(set)
+            subset_name_by_version_id = dict()
             for item in items:
                 version_id = item["version_document"]["_id"]
-                subset_names_by_version_id[version_id].add(item["subset"])
+                subset_name_by_version_id[version_id] = item["subset"]
```
BigRoy, line 573: A version id would never have multiple subsets - so storing it as a set of subsets per version id didn't make much sense.

openpype/tools/loader/widgets.py
```diff
@@ -584,14 +584,15 @@ def on_context_menu(self, point):
...
+                version_ids.remove(version_id)
 
-            for version_id in version_ids:
+            if version_ids:
+                # report versions that didn't have valid representation
                 joined_subset_names = ", ".join([
-                    '"{}"'.format(subset)
-                    for subset in subset_names_by_version_id[version_id]
+                    '"{}"'.format(subset_name_by_version_id[version_id])
+                    for version_id in version_ids
```
BigRoy, line 595: Now this prints just reports just a single `Subsets x, y, z don't have representation 'name'` whereas previously it echoed it per subset name.

iLLiCiTiT (approved):
Looks valid.

EDITED: Found one issue

openpype/tools/loader/widgets.py
```diff
@@ -584,14 +584,15 @@ def on_context_menu(self, point):
             for repre_doc in repre_docs:
                 repre_ids.append(repre_doc["_id"])
 
+                # keep only version ids without representation with that name
                 version_id = repre_doc["parent"]
-                if version_id not in version_ids:
-                    version_ids.remove(version_id)
+                version_ids.remove(version_id)
```
iLLiCiTiT, line 589: I think it was meant as this because if the version id is not in the set it will crash.
```suggestion
                if version_id in version_ids:
                    version_ids.remove(version_id)
```

BigRoy: Ah yes, you are right. Should we instead do [`set.discard()`](https://docs.python.org/3/library/stdtypes.html#frozenset.discard)? :) 

```python
a = set()
a.discard("b")
print(a)
# set()
a = set(["b"])
a.discard("b")
print(a)
# set()
```

iLLiCiTiT: Much better.

PR #3673, opened 2022-08-16
General: Use client projects getter

iLLiCiTiT:
## Brief description
Use `get_projects` from `openpype.client` instead of `projects` method from `AvalonMongoDB`.

## Description
Use `get_projects` function in code to query projects.

## Additional information
I didn't notice this method during query getter refactor.

## Testing notes:
Inactive projects should be hiddedn as expected
- [x] Settings should load projects
- [x] Sync server should load projects and sync properly
- [x] Library loader should show right projects (active library projects in hosts and all active projects from tray)
- [ ] Kitsu (zou) sync should work
- [x] Launcher is showing active projects
- [x] Project manager is showing active projects
- [x] Standalone publisher is showing active projects

### DIFF
--- a/openpype/modules/kitsu/utils/update_zou_with_op.py
+++ b/openpype/modules/kitsu/utils/update_zou_with_op.py
@@ -6,7 +6,11 @@ from typing import List
 import gazu
 from pymongo import UpdateOne
 
-from openpype.client import get_project, get_assets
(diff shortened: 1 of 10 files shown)


kalisp (approved):
Checked and marked most of the points in description.

PR #3674, opened 2022-08-16
Maya: `containerise` dont skip empty values

BigRoy:
## Brief description

Maya host `containerise` was the only implementation that skipped imprinting values if the value was empty/None/zero (`if not value`). Instead of ever allowing to do so this now forces the required attributes to actually be generated on containerise.

Since all values should be string anyway I've also simplifies the logic that adds and sets the attributes.

## Additional info

I was investigating an issue where some loaded containers were missing `loader` and `representation` attributes and I'm unsure how that happened. However, this part of the code could have potentially done so without warnings/errors if `loader` or `representation` values were ever considered null.

## Testing notes:

1. Load containers into Maya
2. Everything should load as usual and be visible in the Scene Inventory, etc.

### DIFF
--- a/openpype/hosts/maya/api/pipeline.py
+++ b/openpype/hosts/maya/api/pipeline.py
@@ -344,7 +344,7 @@ def containerise(name,
         ("id", AVALON_CONTAINER_ID),
         ("name", name),
         ("namespace", namespace),
-        ("loader", str(loader)),
+        ("loader", loader),
         ("representation", context["representation"]["_id"]),
     ]
 
--- a/openpype/hosts/maya/api/pipeline.py
+++ b/openpype/hosts/maya/api/pipeline.py
@@ -350,7 +350,7 @@ def containerise(name,
 
     for key, value in data:
         cmds.addAttr(container, longName=key, dataType="string")
-        cmds.setAttr(container + "." + key, value, type="string")
+        cmds.setAttr(container + "." + key, str(value), type="string")
 
     main_container = cmds.ls(AVALON_CONTAINERS, type="objectSet")
     if not main_container:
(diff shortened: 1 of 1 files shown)


openpype/hosts/maya/api/pipeline.py
```diff
@@ -349,16 +349,8 @@ def containerise(name,
...
-
-        if isinstance(value, (int, float)):
-            cmds.addAttr(container, longName=key, attributeType="short")
-            cmds.setAttr(container + "." + key, value)
-
-        else:
-            cmds.addAttr(container, longName=key, dataType="string")
-            cmds.setAttr(container + "." + key, value, type="string")
+        cmds.addAttr(container, longName=key, dataType="string")
+        cmds.setAttr(container + "." + key, value, type="string")
```
iLLiCiTiT, line 353: Not sure if Maya convert values automatically but I think that representation id is `ObjectId`.
```suggestion
        cmds.setAttr(container + "." + key, str(value), type="string")
```

BigRoy: Maya does not automatically convert all types no - however, this would change any `None` value into "None". If we're fine with that I'm happy with refactoring like this.

The ObjectId being set hasn't been an issue so far somehow. This works fine:
```python
from maya import cmds
from bson.objectid import ObjectId
id = ObjectId()
plug = "blabla_CON.representation"
cmds.setAttr(plug, id, type="string")
```

This does error:
```python
cmds.setAttr(plug, None, type="string")
```

iLLiCiTiT: Shouldn't be the values validated before are really stored into the scene? Because this way you'll end up with container that has half of data anyway.

BigRoy: I feel like they should be - yes. :) Haven't seen direct issues in the past either. I'll refactor this to use `str()` like you describe because many other hosts seem to do something similar.

antirotor (approved):
Tested and it works...

PR #3675, opened 2022-08-16
Ftrack: Set task status on task creation in integrate hierarchy

iLLiCiTiT:
## Brief description
Have ability to set task status on creation during integrate hierarchy (editorial publishing).

## Description
Added profiles using task name and task type to be able tell what status should be set on new created task during integrate hierarchy.

## Testing notes:
1. Add a profile which define some status name for specific task type (`project_settings/ftrack/publish/IntegrateHierarchyToFtrack/create_task_status_profiles`)
2. Publish editorial which creates new tasks
3. Check if statuses of new tasks are right


jakubjezek001 (approved):
works ;)

PR #3677, opened 2022-08-16
Ftrack: Store ftrack entities on hierarchy integration to instances

iLLiCiTiT:
## Brief description
Ftrack integrate hierarchy is storing the created entities to instances with created task entities.

## Description
Created entities during integrate hierarchy are stored to instances that should have stored those entities. Key `"ftrackEntity"` was already set (more then once) but `"ftrackTask"` was not.

Removed settings in ftrack server event handlers for `first_version_status (Set status on first created version)` which is not used anywhere.

## Additional information
This is minor change which probably won't have much affect as it is handled only during editorial publishing which probably won't create asset versions at the time.

## Testing notes:
1. Publish editorial which create new entities in ftrack
2. Rest of publishing should not crash
3. If there are reviewables for new created entities they should be also added to related tasks


jakubjezek001 (approved):
tested with traypublisher on editorial and worked as expected

PR #3678, opened 2022-08-16
Webpublisher:change create flatten image into tri state

kalisp:
## Brief description
`CollectColorCodedInstances` allows publish only flattened subset of 'image' family.

## Description
Previously it was possible to set in Settings to create flatten subset of 'image' family from published subsets. But separate layers marked by publishable colors were published also.
This PR brings more granularity, now it is possible to:
- create flattened 'image' subset of all publishable layers + create 'image' subset per publishable layer
- create ONLY flattened 'image' subset of all publishable layers
- not create flattened 'image' subset at all, only separate subsets per marked layer.

## Testing notes:
1. set `project_settings/photoshop/publish/CollectColorCodedInstances/flatten_subset_template`
2. 
Screenshot 2022-08-16 170951

3. mark some top layers in PS to be published
4. check that only flatten subset was produced

### DIFF
--- a/openpype/hosts/photoshop/plugins/publish/collect_color_coded_instances.py
+++ b/openpype/hosts/photoshop/plugins/publish/collect_color_coded_instances.py
@@ -32,7 +32,7 @@ class CollectColorCodedInstances(pyblish.api.ContextPlugin):
     # TODO check if could be set globally, probably doesn't make sense when
     # flattened template cannot
     subset_template_name = ""
-    create_flatten_image = False
+    create_flatten_image = "no"
     # probably not possible to configure this globally
     flatten_subset_template = ""
 
@@ -98,13 +98,16 @@ class CollectColorCodedInstances(pyblish.api.ContextPlugin):
                     "Subset {} already created, skipping.".format(subset))
                 continue
 
-            instance = self._create_instance(context, layer, resolved_family,
-                                             asset_name, subset, task_name)
+            if self.create_flatten_image != "only":
+                instance = self._create_instance(context, layer,
+                                                 resolved_family,
+                                                 asset_name, subset, task_name)
+                created_instances.append(instance)
+
             existing_subset_names.append(subset)
             publishable_layers.append(layer)
-            created_instances.append(instance)
 
-        if self.create_flatten_image and publishable_layers:
+        if self.create_flatten_image != "no" and publishable_layers:
             self.log.debug("create_flatten_image")
             if not self.flatten_subset_template:
                 self.log.warning("No template for flatten image")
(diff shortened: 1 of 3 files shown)


openpype/hosts/photoshop/plugins/publish/collect_color_coded_instances.py
```diff
@@ -32,7 +32,7 @@ class CollectColorCodedInstances(pyblish.api.ContextPlugin):
     # TODO check if could be set globally, probably doesn't make sense when
     # flattened template cannot
     subset_template_name = ""
-    create_flatten_image = False
+    create_flatten_image = "no"
```
iLLiCiTiT, line 35: I would add some explanation about possible values and what is the difference especially what is difference in `yes` and `only`. Maybe I would change the value to tell what only e.g. `flatten_only`.

kalisp: Added more docstrings.
In setting it is actually as enum

iLLiCiTiT: > In setting it is actually as enum

I know but it doesn't meant the value can't tell what it means.

As I'm reading the docstrings wouldn't be better to have "Flatten only (flatten_only)" and "Flatten with image (flatten_with_image)". Because what you have now as flatten only more suits value yes, but current yes does more then flatten.

kalisp: Ok, makes sense. Updated.

openpype/hosts/photoshop/plugins/publish/collect_color_coded_instances.py
```diff
@@ -9,14 +9,22 @@
...
+    coding. Top level layers (group) must be marked by specific color to be
+    published as an instance of 'image' family.
 
     Can add group for all publishable layers to allow creation of flattened
     image. (Cannot contain special background layer as it cannot be grouped!)
 
+    Based on value `create_flatten_image` from Settings:
+    - "yes": create flattened 'image' subset of all publishable layers + create
+        'image' subset per publishable layer
+    - "only": create ONLY flattened 'image' subset of all publishable layers
```
iLLiCiTiT, line 24:
```suggestion
    - "flatten_with_images": create flattened 'image' subset of all publishable layers + create
        'image' subset per publishable layer
    - "flatten_only": create ONLY flattened 'image' subset of all publishable layers
```

iLLiCiTiT (approved):
Code looks ok, but can't test it without Photoshop.

64qam (approved):

Issue #3396, opened 2022-06-23
Maya: Collect Render slows down publishing a lot


BigRoy:
**Describe the bug**

Collect Render is really slow for large scenes with many layers. The only reason why it's slow is due to these logging lines:
- https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/maya/plugins/publish/collect_render.py#L217
- https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/maya/plugins/publish/collect_render.py#L298
- https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/maya/plugins/publish/collect_render.py#L413

That changes a scene that takes 15+ seconds to collect back down to <1 second to collect when they are removed.

The issue also is that with that much logged data somehow the Pyblish Pype UI gets relatively slow. The UI remains repsonsive just fine when not processing (except when looking into the logs of that plug-in) but it gets really really sluggish during processing when needing to log more. For example, with the logs still enabled validation also becomes very slow for my test scene - but with those logs just removed in `collect_render` then validation also becomes close to instant.

It's a huge speed difference.

I've quickly tested moving these into an if branch so it only runs if `os.environ.get("OPENPYPE_DEBUG") == "1"` and that also worked.

**To Reproduce**
1. Create 10 Render Setup layers (so there's quite a few layers to process)
2. Set render frame range to have 200+ frames, e.g. `1001-1300` (just so expected files list is large to log)
3. Try to publish/submit scene. It'll be very slow.

**Expected behavior**

Should be instant.

**Desktop (please complete the following information):**
 - OS: Windows
 - Host: Maya 2022

BigRoy:
##### Note

I should add that I had done a similar optimization for the verbose logging in `CollectAnatomyContextData`, `CollectAnatomyInstanceData` and `CollectInstances` to not log the full json dumped data unless `OPENPYPE_DEBUG == 1`. That's also why collecting on my end might have been under a second after this change - if you try and reproduce the speed up you migth still see slow-downs due to those collectors also being very slow.

iLLiCiTiT:
I agree that logging full instance or context data is overkill especially when it's in each plugin.

PR #3679, opened 2022-08-16
Global: making collect audio plugin global

jakubjezek001:
## Brief description
Collect audio plugin could be used by other hosts.

## Description
Reviewable videos can be created with included audio tracks if any audio subset is available. Subset name with audio variant can be defined in settings `project_settings/global/publish/CollectAudio/audio_subset_name`

## Testing notes:
1. standalone publish any image sequence or video on shot with available audio subset.
2. reviewable should have audio track


openpype/plugins/publish/collect_audio.py
```diff
@@ -0,0 +1,76 @@
...
+            if not repre_docs:
+                self.log.warning(
+                    "Version document does not contain any representations"
+                )
+            else:
+                repre_doc = repre_docs[0]
+
+        # Add audio to instance if representation was found
+        if repre_doc:
+            instance.data["audio"] = [{
```
iLLiCiTiT, line 71: What if is `"audio"` already filled?

jakubjezek001: good catch, yeah it is fixed now ;)

openpype/plugins/publish/collect_audio.py
```diff
@@ -0,0 +1,77 @@
...
+    legacy_io,
+    get_representation_path,
+)
+
+
+class CollectAudio(pyblish.api.InstancePlugin):
+    """ Collecting available audio subset to instance
+
+    """
+    label = "Colect Audio"
```
BigRoy, line 18:
```suggestion
    label = "Collect Audio"
```

openpype/plugins/publish/collect_audio.py
```diff
@@ -0,0 +1,77 @@
...
+            if not repre_docs:
+                self.log.warning(
+                    "Version document does not contain any representations"
+                )
+            else:
+                repre_doc = repre_docs[0]
+
+        # Add audio to instance if representation was found
+        if repre_doc:
+            if not instance.data.get("audio"):
```
iLLiCiTiT, line 71: This should probably the first condition of the plugin instead of last.

openpype/plugins/publish/collect_audio.py
```diff
@@ -0,0 +1,83 @@
...
+from openpype.pipeline import (
+    legacy_io,
+    get_representation_path,
+)
+
+
+class CollectAudio(pyblish.api.InstancePlugin):
+    """ Collecting available audio subset to instance
+
+    """
```
BigRoy, line 17: Can we improve this docstring? To me it doesn't really explain what this does?

Should this be more elaborate like:

```python
"""Collect asset's last published audio.

The audio subset name searched for is defined in:
    project settings > Collect Audio
"""
```

openpype/plugins/publish/collect_audio.py
```diff
@@ -0,0 +1,83 @@
...
+
+        # Add audio to instance if representation was found
+        if repre_doc:
+            instance.data["audio"] = [{
+                "offset": 0,
+                "filename": get_representation_path(repre_doc)
+            }]
+            self.log.info("Audio Data added to instance ...")
+
+        self.log.debug("instance.data: {}".format(pformat(instance.data)))
```
BigRoy, line 83: I'd always refrain from just logging all `instance.data` - if this is a necessity then I feel there's a UX problem with the Pyblish workflows in general. The instance data can be very big and logging it (and formatting it!) can be very slow see #3396 
```suggestion
```

openpype/plugins/publish/collect_audio.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def process(self, instance):
+        if instance.data.get("audio"):
+            self.log.info(
+                "Skipping Audio collecion. It is already collected"
+            )
+            return
+
+        # Add audio to instance if exists.
+        self.log.info('Collecting Audio Data ...')
```
BigRoy, line 49: I feel this would be more useful if it would state what audio subset name it would look for.
```suggestion
        self.log.info("Searching for audio subset '{subset}' in asset '{asset}'".format(asset=instance.data["asset"], self.audio_subset_name))
```
☝️ Note: Line length of course is too long in suggested example. Consider psuedocode

openpype/plugins/publish/collect_audio.py
```diff
@@ -0,0 +1,83 @@
...
+            # Try to find it's representation (Expected there is only one)
+            repre_docs = list(get_representations(
+                project_name, version_ids=[last_version_doc["_id"]]
+            ))
+            if not repre_docs:
+                self.log.warning(
+                    "Version document does not contain any representations"
+                )
+            else:
+                repre_doc = repre_docs[0]
```
BigRoy, line 73: In most publishes many instances will share the same asset. These database queries can be slow for many instances too. To optimize it would be nice to "cache" the queries so we'd do much fewer database queries.

```python
cache = context.data.get("__cache_asset_audio", {})
if asset in cache:
    return cache[asset]

# Else query as usual and store in cache
```

References where this is also done:
- https://github.com/pypeclub/OpenPype/blob/4a4bb22f60353bb9cfcdb64c516b2b145cb7c966/openpype/hosts/maya/plugins/publish/collect_inputs.py#L116-L126

The alternative option would be to turn this in a ContextPlugin so it runs once and collects for all instances manually

openpype/plugins/publish/collect_audio.py
```diff
@@ -0,0 +1,83 @@
...
+    legacy_io,
+    get_representation_path,
+)
+
+
+class CollectAudio(pyblish.api.InstancePlugin):
+    """ Collecting available audio subset to instance
+
+    """
+    label = "Collect Audio"
```
BigRoy, line 18: Would this be more descriptive as `Collect Asset Audio` or `Collect Audio from Asset`?

openpype/plugins/publish/collect_audio.py
jakubjezek001: good point, thanks

iLLiCiTiT:
Already removed OcioConfig submodule is added back to repository.

openpype/plugins/publish/collect_audio.py
```diff
@@ -0,0 +1,103 @@
...
+        # Add audio to instance if representation was found
+        if repre_doc:
+            instance.data["audio"] = [{
+                "offset": 0,
+                "filename": get_representation_path(repre_doc)
+            }]
+            self.log.info("Audio Data added to instance ...")
+
+    def _get_repre_doc(self, instance):
+        cache = instance.context.data.get("__cache_asset_audio", {})
```
iLLiCiTiT, line 69:
```suggestion
        cache = instance.context.data.get("__cache_asset_audio")
        if cache is None:
            cache = {}
            instance.context.data["__cache_asset_audio"] = cache
```

openpype/plugins/publish/collect_audio.py
```diff
@@ -0,0 +1,103 @@
...
+            if not repre_docs:
+                self.log.warning(
+                    "Version document does not contain any representations"
+                )
+            else:
+                repre_doc = repre_docs[0]
+
+        # update cache
+        cache[asset_name] = repre_doc
+        instance.context.data["__cache_asset_audio"].update(cache)
```
iLLiCiTiT, line 101:
```suggestion
```

openpype/settings/defaults/project_settings/global.json
```diff
@@ -3,6 +3,10 @@
         "CollectAnatomyInstanceData": {
             "follow_workfile_version": false
         },
+        "CollectAudio": {
+            "enabled": true,
```
iLLiCiTiT, line 7: I would turn this plugin off by default as most of studio don't use audio and current state of plugin could do a lot of queries to mongo.

jrsndl:
I tested on EXR sequence, available audio was added to the reviewable.

iLLiCiTiT (approved):
Looks ok for me.

PR #3683, opened 2022-08-17
General: version string in subset name break file sequence collection in review

moonyuet:
## Brief description
When there is a version string in subset name, like foo_bar_v004  and that subset name is used for review, clique  has trouble collecting playblasted sequences for it because it tries to parse version string as a frame number, resulting in multiple collections.

## Solutions
Defines the pattern of the filename in `clique.assemble()`
Instead of 
```
        collections, remainder = clique.assemble(collected_files,
                                                 minimum_items=1)
```
Uses
```
        pattern_frame = [r'\.(?P<index>(?P<padding>0*)\d+)\.\D+\d?$']
        collections, remainder = clique.assemble(collected_files,
                                                 minimum_items=1,
                                                 patterns=pattern_frame)
```

### DIFF
--- a/openpype/hosts/maya/plugins/publish/extract_playblast.py
+++ b/openpype/hosts/maya/plugins/publish/extract_playblast.py
@@ -138,10 +138,10 @@ class ExtractPlayblast(openpype.api.Extractor):
         self.log.debug("playblast path  {}".format(path))
 
         collected_files = os.listdir(stagingdir)
-        pattern_frame = [r'\.(?P<index>(?P<padding>0*)\d+)\.\D+\d?$']
+        patterns = [clique.PATTERNS["frames"]]
         collections, remainder = clique.assemble(collected_files,
                                                  minimum_items=1,
-                                                 patterns=pattern_frame)
+                                                 patterns=patterns)
 
         self.log.debug("filename {}".format(filename))
         frame_collection = None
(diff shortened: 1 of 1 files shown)


BigRoy:
This pattern would again force it to only recognize file patterns with a `.` in the filename.
Also, I believe you should also be able to just use:
```python
patterns = [clique.PATTERNS["frames"]]
collections, remainder = clique.assemble(frames,
                                         minimum_items=1,
                                         patterns=patterns)
```

That way you don't need to manage the regex pattern yourself and can use one of the ones coming with `clique`.

moonyuet:
> This pattern would again force it to only recognize file patterns with a `.` in the filename. (Which might not be a problem for Extract playblast) but would potentially be for others as came up in this PR #3611
> 
> Also, I believe you should also be able to just use:
> 
> ```python
> patterns = [clique.PATTERNS["frames"]]
> collections, remainder = clique.assemble(frames,
>                                          minimum_items=1,
>                                          patterns=patterns)
> ```
> 
> That way you don't need to manage the regex pattern yourself and can use one of the ones coming with `clique`.
> 
> Can you confirm that before this PR that clique fails when trying to `assemble` just a single frame but succeeds with multiple frames. I recall clique doing something like that.

moonyuet:
> This pattern would again force it to only recognize file patterns with a `.` in the filename. (Which might not be a problem for Extract playblast) but would potentially be for others as came up in this PR #3611
> 
> Also, I believe you should also be able to just use:
> 
> ```python
> patterns = [clique.PATTERNS["frames"]]
> collections, remainder = clique.assemble(frames,
>                                          minimum_items=1,
>                                          patterns=patterns)
> ```
> 
> That way you don't need to manage the regex pattern yourself and can use one of the ones coming with `clique`.
> 
> Can you confirm that before this PR that clique fails when trying to `assemble` just a single frame but succeeds with multiple frames. I recall clique doing something like that.

I agree. Thanks for telling me about this.

antirotor (approved):

PR #3685, opened 2022-08-17
Ftrack: Addiotional component metadata

iLLiCiTiT:
## Brief description
Added ability to add additional metadata into components.

## Description
Additional metadata are not stored under `ftr_meta` (required for play and cinesync) but as keys next to it. It is possible to set which keys are added to component in settings. If value for a key is not available component it is skipped. There are available keys `openpype_version`, `frame_start`, `frame_end`, `duration`, `width`, `height`, `fps` and `codec`.

### Screenshots
image

## Testing notes:
1. Add additional metadata keys in `project_settings/ftrack/publish/IntegrateFtrackInstance/additional_metadata_keys` (e.g. `openpype_version`)
2. Publish something which creates components in ftrack
3. Check it's metadata


m-u-r-p-h-y (approved):
tested with a couple of different publishes and it works as expected:

Maya - playblast
image

Maya - lookdev
image

PR #3686, opened 2022-08-17
General: Benevolent context label collector

iLLiCiTiT:
## Brief description
Context label collector is more benevolent to available data.

## Description
Collector does not require `currentFile` to be filled and use just host name in that case. If context data contains `contextLabel` the plugin just set it as label of context and skip rest of logic, this was added to have ability define custom context label and skip the plugin logic. Result is that standalone publisher and tray publisher don't have to be explicitly named in the plugin.

Pyblish pype does not force to have `"label"` in `instance.data` and `context.data`.

## Testing notes:
1. Context label should be collected as did

### DIFF
--- a/openpype/plugins/publish/collect_context_label.py
+++ b/openpype/plugins/publish/collect_context_label.py
@@ -1,5 +1,6 @@
 """
-Requires:
+Optional:
+    context     -> hostName (str)
     context     -> currentFile (str)
 Provides:
     context     -> label (str)
@@ -16,16 +17,16 @@ class CollectContextLabel(pyblish.api.ContextPlugin):
     label = "Context Label"
 
     def process(self, context):
+        host_name = context.data.get("hostName")
+        if not host_name:
+            host_name = pyblish.api.registered_hosts()[-1]
+        # Use host name as base for label
+        label = host_name.title()
 
-        # Get last registered host
-        host = pyblish.api.registered_hosts()[-1]
-
-        # Get scene name from "currentFile"
-        path = context.data.get("currentFile") or "<Unsaved>"
-        base = os.path.basename(path)
+        # Get scene name from "currentFile" and use basename as ending of label
+        path = context.data.get("currentFile")
+        if path:
+            label += " - {}".format(os.path.basename(path))
 
         # Set label
-        label = "{host} - {scene}".format(host=host.title(), scene=base)
-        if host == "standalonepublisher":
-            label = host.title()
         context.data["label"] = label
--- a/openpype/plugins/publish/collect_context_label.py
+++ b/openpype/plugins/publish/collect_context_label.py
@@ -17,6 +17,12 @@ class CollectContextLabel(pyblish.api.ContextPlugin):
     label = "Context Label"
 
     def process(self, context):
+        # Add ability to use custom context label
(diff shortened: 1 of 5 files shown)


openpype/plugins/publish/collect_context_label.py
```diff
@@ -16,16 +17,22 @@ class CollectContextLabel(pyblish.api.ContextPlugin):
...
-        host = pyblish.api.registered_hosts()[-1]
-
-        # Get scene name from "currentFile"
-        path = context.data.get("currentFile") or "<Unsaved>"
-        base = os.path.basename(path)
+        # Add ability to use custom context label
+        context_label = context.data.get("contextLabel")
+        if context_label:
+            context.data["label"] = context_label
+            return
```
BigRoy, line 24: If this is the case - why not instead have it do `if context.data.get("label")` and ignore this plug-in if a label was already set for the context. Or is `context.data["label"]` set usually?

openpype/plugins/publish/collect_context_label.py
```diff
@@ -16,16 +17,22 @@ class CollectContextLabel(pyblish.api.ContextPlugin):
...
+
+        host_name = context.data.get("hostName")
+        if not host_name:
+            host_name = pyblish.api.registered_hosts()[-1]
```
BigRoy, line 28: Would this still work with the new class based hosts implementations?

iLLiCiTiT: This is a backup solution when `hostName` is not collected. I think this will never be used in UI.

openpype/plugins/publish/collect_context_label.py
iLLiCiTiT: I think that `context.data["label"]` is set to `Context` by default.

iLLiCiTiT: Did some research and found out it's used that way in pyblish pype.

iLLiCiTiT: Changed

openpype/plugins/publish/collect_context_label.py
iLLiCiTiT: To answer question. The `"hostName"` on context data has some logic how it can be filled and one possible approach can be to use registered host.

openpype/tools/pyblish_pype/model.py
```diff
@@ -616,9 +611,19 @@ def type(self):
 
     def data(self, role=QtCore.Qt.DisplayRole):
         if role == QtCore.Qt.DisplayRole:
+            label = None
             if settings.UseLabel:
-                return self.instance.data["label"]
-            return self.instance.data["name"]
+                label = (
+                    self.instance.data.get("label")
+                    or getattr(self.instance, "label", None)
```
BigRoy, line 618: Has `instance.label` ever been I thing? I noticed that logic is also on the PluginModel and plugins do have `label` attribute but have never seen it on Pyblish instances.

iLLiCiTiT: I'm actually not sure? I maybe added it in past (like 3-4 years ago) and was scared to remove later. If it's not a thing I can remove it.

64qam (approved):

PR #3690, opened 2022-08-18
Maya: write color sets in  create_model

moonyuet:
## Brief description
Adding the on/off option of write color sets attributes in _create_rig.py_

Adding swtiching on/off clickbox to override the boolean option of write_color_sets_arttirbute in _create_model.py_ and  _create_rig.py_

image
image

### DIFF
--- a/openpype/hosts/maya/plugins/create/create_rig.py
+++ b/openpype/hosts/maya/plugins/create/create_rig.py
@@ -13,12 +13,15 @@ class CreateRig(plugin.Creator):
     label = "Rig"
     family = "rig"
     icon = "wheelchair"
+    write_color_sets = False
+    def __init__(self, *args, **kwargs):
+        super(CreateRig, self).__init__(*args, **kwargs)
+        self.data["writeColorSets"] = self.write_color_sets
 
     def process(self):
 
         with lib.undo_chunk():
             instance = super(CreateRig, self).process()
-
             self.log.info("Creating Rig instance set up ...")
             controls = cmds.sets(name="controls_SET", empty=True)
             pointcache = cmds.sets(name="out_SET", empty=True)
--- a/openpype/hosts/maya/plugins/create/create_rig.py
+++ b/openpype/hosts/maya/plugins/create/create_rig.py
@@ -18,6 +18,7 @@ class CreateRig(plugin.Creator):
     def __init__(self, *args, **kwargs):
         super(CreateRig, self).__init__(*args, **kwargs)
         self.data["writeColorSets"] = self.write_color_sets
+        self.data["writeFaceSets"] = False
 
     def process(self):
 
--- a/openpype/hosts/maya/plugins/create/create_rig.py
+++ b/openpype/hosts/maya/plugins/create/create_rig.py
@@ -14,11 +14,12 @@ class CreateRig(plugin.Creator):
     family = "rig"
     icon = "wheelchair"
     write_color_sets = False
+    write_face_sets = False
 
     def __init__(self, *args, **kwargs):
         super(CreateRig, self).__init__(*args, **kwargs)
         self.data["writeColorSets"] = self.write_color_sets
-        self.data["writeFaceSets"] = False
+        self.data["writeFaceSets"] = self.write_face_sets
 
     def process(self):
 
--- a/openpype/hosts/maya/plugins/create/create_rig.py
+++ b/openpype/hosts/maya/plugins/create/create_rig.py
@@ -14,6 +14,7 @@ class CreateRig(plugin.Creator):
     family = "rig"
     icon = "wheelchair"
     write_color_sets = False
+
     def __init__(self, *args, **kwargs):
         super(CreateRig, self).__init__(*args, **kwargs)
         self.data["writeColorSets"] = self.write_color_sets
--- a/openpype/hosts/maya/plugins/create/create_rig.py
+++ b/openpype/hosts/maya/plugins/create/create_rig.py
@@ -13,18 +13,12 @@ class CreateRig(plugin.Creator):
     label = "Rig"
     family = "rig"
     icon = "wheelchair"
-    write_color_sets = False
-    write_face_sets = False
-
-    def __init__(self, *args, **kwargs):
-        super(CreateRig, self).__init__(*args, **kwargs)
-        self.data["writeColorSets"] = self.write_color_sets
-        self.data["writeFaceSets"] = self.write_face_sets
 
     def process(self):
 
         with lib.undo_chunk():
             instance = super(CreateRig, self).process()
+
             self.log.info("Creating Rig instance set up ...")
             controls = cmds.sets(name="controls_SET", empty=True)
             pointcache = cmds.sets(name="out_SET", empty=True)
--- a/openpype/settings/defaults/project_settings/maya.json
+++ b/openpype/settings/defaults/project_settings/maya.json
@@ -187,8 +187,6 @@
         },
         "CreateRig": {
             "enabled": true,
-            "write_color_sets": false,
-            "write_face_sets": false,
             "defaults": [
                 "Main",
                 "Sim",
--- a/openpype/settings/defaults/project_settings/maya.json
+++ b/openpype/settings/defaults/project_settings/maya.json
@@ -102,6 +102,7 @@
         "CreateAnimation": {
             "enabled": true,
             "write_color_sets": false,
+            "write_face_sets": false,
             "defaults": [
                 "Main"
             ]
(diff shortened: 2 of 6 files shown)


openpype/hosts/maya/plugins/create/create_rig.py
```diff
@@ -13,12 +13,16 @@ class CreateRig(plugin.Creator):
     label = "Rig"
     family = "rig"
     icon = "wheelchair"
+    write_color_sets = False
+
+    def __init__(self, *args, **kwargs):
+        super(CreateRig, self).__init__(*args, **kwargs)
+        self.data["writeColorSets"] = self.write_color_sets
```
BigRoy, line 20: This shouldn't make much sense since the rig comes out as maya scene and does not strip any color set data. Animation instances however would extract as Alembic.

openpype/hosts/maya/plugins/publish/extract_playblast.py
```diff
@@ -138,8 +138,10 @@ def process(self, instance):
         self.log.debug("playblast path  {}".format(path))
 
         collected_files = os.listdir(stagingdir)
+        pattern_frame = [r'\.(?P<index>(?P<padding>0*)\d+)\.\D+\d?$']
         collections, remainder = clique.assemble(collected_files,
-                                                 minimum_items=1)
+                                                 minimum_items=1,
+                                                 patterns=pattern_frame)
```
BigRoy, line 144: Seems unrelated to the PR description.

moonyuet: my bad, this is related to another ticket.

m-u-r-p-h-y: > Seems unrelated to the PR description.

coming from #3683 ?

moonyuet: yes, it is

m-u-r-p-h-y:
I would add Face Sets options as well to make it complete

image

moonyuet:
> I would add Face Sets options as well to make it complete
> 
> image

added into the _create_rig.py_

openpype/hosts/maya/plugins/create/create_rig.py
m-u-r-p-h-y: we are trying to unify colors sets (and other sets) flow and even though currently rig creator can only output Maya scene we plan to add fbx representation (replacing the skeletal mesh) so it will make sense then.

It is not influencing anything currently but we thought it would be good to have it there for future use and consistency.

BigRoy: > It is not influencing anything currently but we thought it would be good to have it there for future use and consistency

I see what you're going for - but it'd just confuse the artists into thinking it would do something useful.
Also, most usual "rigs" won't be suitable to do an "fbx rig" export for - and then also automate `animation` instance creation for in maya when loading them again. I'd say `fbxRig` is a family of its own.

m-u-r-p-h-y: so we have to decide if we remove the component sets from Rig Creator @antirotor

m-u-r-p-h-y:
> added into the _create_rig.py_

can you also add the Face sets to the corresponding setting?

openpype/hosts/maya/plugins/create/create_rig.py
m-u-r-p-h-y: After internal discussion, we decided to keep the sets for Rig Creator there. We understand it is not very clean from the code perspective as these options are actually modifying the animation output, not the rig output itself, but as it does not have its own creator we gave preference to keeping all attributes settings in one place in preferences and not moving it outside just for the rig. 
We think it will be more consistent from the user's point of view.

m-u-r-p-h-y (approved):
the settings are correctly propagated to their respective creators.

note that the settings for Rig Creator are not modifying rig publish, but output from rig animation, which is handled automatically without its own creator. 

image

BigRoy:
> note that the settings for Rig Creator are not modifying rig publish, but output from rig animation, which is handled automatically without its own creator.

Where is the code/logic for that? I don't recall seeing any Rig loader logic that sets anything like that for the to be created animation instance? Looking at the code the animation instance values have nothing to do with the setting that were set on the rig instance when being published. (That data is not even in the database or on disk, right?)

Or what obvious thing am I missing?

antirotor:
> Or what obvious thing am I missing?

No, but it is somewhat convoluted - animation creator is called when rig is loaded into the scene:

https://github.com/pypeclub/OpenPype/blob/377df371db52412f9485272da413448ef647107c/openpype/hosts/maya/plugins/load/load_reference.py#L155-L167

but since this creator is never called by user (and shouldn't be even listed in Creator dialog - but we can't hide it now), it's logical to move settings for it somewhere. Not saying it is ideal, but the options are:

- have the settings in animation creator: that is logical from the point of code, but not very user intuitive
- have that option on rig loader - but again, it actually influence subsequent publishing of animation from the loaded rig
- have it on the rig creator as we have in this PR
- other option? :)

BigRoy:
The point I was trying to make was that the values you set on the Rig instance [here](https://github.com/pypeclub/OpenPype/pull/3690/files#diff-28c70455cabe6808db98be267a240ada7af79a80a9beb2c69a131e8dcb393feaR21-R22) don't seem to influence the values on the Animation instance at all. Those values are not published along with the Rig in such a way that it seems that the Rig loader set anything related on the animation in any way.

As such the values on CreateRig and the rig instance itself have no use. They do nothing, they don't propagate to the animation instance when the rig is loaded. Or where is that logic?

---

> have the settings in animation creator: that is logical from the point of code, but not very user intuitive
> have that option on rig loader - but again, it actually influence subsequent publishing of animation from the loaded rig
> have it on the rig creator as we have in this PR
> other option? :)

I'd say that IF the rig instance can set default values for the animation instance that could be quite useful. So that an advanced rigger could say "for this rig please publish animation with write color sets enabled by default" - I'd love to see that feature with "visibleOnly" to be settable per rig instance too. But the logic involved with trickling down decisions like that are much more complicated (the code isn't there yet as far as I know).

What happens if:
- Rig instance was already loaded and an animation family was created, now the Rig version gets updated which has different values for the "write color data" published along. What do we do with the existing animation instance in the scene? Do we preserve any local adjustments a user might have made or do we force revert it?
- If the Rigger allows to set a preferred value does it really make sense to allow the animator to still change the value? Like, how "strict" are these? 

---

TL;DR

Having the settings on the animation/pointcache instance is worthwhile - but I'm not seeing how the logic added now for rig instances is doing anything useful.

openpype/hosts/maya/plugins/create/create_rig.py
```diff
@@ -13,12 +13,18 @@ class CreateRig(plugin.Creator):
     label = "Rig"
     family = "rig"
     icon = "wheelchair"
+    write_color_sets = False
+    write_face_sets = False
+
+    def __init__(self, *args, **kwargs):
+        super(CreateRig, self).__init__(*args, **kwargs)
+        self.data["writeColorSets"] = self.write_color_sets
+        self.data["writeFaceSets"] = self.write_face_sets
```
mkolar, line 22: @BigRoy roy is correct. This code has absolutely no bearing on any result anywhere. When we load the rig, an animation instance is created automatically, but that uses the settings from animation creator, rather than rig. So this introduces two attributes that are just extra and will create confusion. 

It would of course be possible to have these parsed into the animation instance that is created, so that rigger could choose what values should be used down the line as starting point, but I'm not sure that it's actually worth the hassle at the moment, considering that simply adding the attributes to the animation creator solves the immediate issue for the upcoming production. 

that being said, the confusion between `animation` family, `animation` creator and how and if they are affected from the rig, is almighty and should be addressed. I'm not convinced it is the responsibility of this PR though, so I reckon let's just remove the obsolete code from the rig creator here and separately have a look at how to clean up the rig -> animation workflow to make it a bit more clear.

mkolar (changes requested):

m-u-r-p-h-y:
we should unblock this one as some studios are waiting for create_model at least

@BigRoy is right that the added attribute on create_rig is not doing anything so @moonyuet can you remove it pls so we can merge this pls?

openpype/settings/defaults/project_settings/maya.json
```diff
@@ -33,7 +33,7 @@
     },
     "RenderSettings": {
         "apply_render_settings": true,
-        "default_render_image_folder": "",
+        "default_render_image_folder": "renders",
```
mkolar, line 36: this seems unrelated to this PR, other than that I'm happy to merge

mkolar (changes requested):
All code seems good to me, let's just please remove unrelated settings change as per comment

antirotor (approved):

PR #3695, opened 2022-08-19
RoyalRender: handle host name that is not set

antirotor:
## Description

Not setting `AVALON_APP_NAME` to proper configured host name resulted in crashes in  `extract_review` and `integrate_subset_gorups` that relied on `app` resp. `hostName` to be set in context.

This is not directly related to RoyalRender, only that RoyalRender publisher doesn't set this as it is oblivious about host name to be used.

### DIFF
--- a/openpype/plugins/publish/integrate_subset_group.py
+++ b/openpype/plugins/publish/integrate_subset_group.py
@@ -93,6 +93,6 @@ class IntegrateSubsetGroup(pyblish.api.InstancePlugin):
         return {
             "families": anatomy_data["family"],
             "tasks": task.get("name"),
-            "hosts": anatomy_data.get("app"),
+            "hosts": instance.context.data["hostName"],
             "task_types": task.get("type")
         }
--- a/openpype/plugins/publish/integrate_subset_group.py
+++ b/openpype/plugins/publish/integrate_subset_group.py
@@ -93,6 +93,6 @@ class IntegrateSubsetGroup(pyblish.api.InstancePlugin):
         return {
             "families": anatomy_data["family"],
             "tasks": task.get("name"),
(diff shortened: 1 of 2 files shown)


iLLiCiTiT:
This mean that all filters based on host name won't work. Extract review won't produce what is expected for the host, integrator may not use right publish templates etc. I would recommend to set `AVALON_APP` to host name which is source of the publish sources.

Note: Using `AVALON_APP_NAME` was just a different approach how to get the host name and was requirement for royal render in past.

openpype/plugins/publish/integrate_subset_group.py
```diff
@@ -93,6 +93,6 @@ def get_profile_filter_criteria(self, instance):
         return {
             "families": anatomy_data["family"],
             "tasks": task.get("name"),
-            "hosts": anatomy_data["app"],
+            "hosts": anatomy_data.get("app"),
```
iLLiCiTiT, line 96: Well if we change it then use the right key :)
```suggestion
            "hosts": instance.context.data["hostName"],
```

antirotor:
> I would recommend to set `AVALON_APP` to host name which is source of the publish sources.

This is not unfortunately possible. Source application can be determined, but with completely different names than what we have configured in application manager so you would need to maintain arbitrary mapping of RoyalRender app names and ours. And you would actually need to have this application configured in OpenPype even if there is no other integration for it at all.

antirotor:
> This mean that all filters based on host name won't work

for this particular case I don't care. I've always considered this as poor mans alternative to proper integration. There are tons of things missing there, not only these functions.

iLLiCiTiT (approved):

PR #3697, opened 2022-08-19
Houdini: Publishing workfiles

antirotor:
## Feature

This is adding ability to publish **Houdini HIP files** as workfiles the same way as is currently done in Maya. It adds Validator that checks if paths inside specific nodes doesn't contain unresolved `$HIP`/`$JOB` variables (because one this workfile is copied to published area, those links would be broken). Validator has repair action to resolve those to absolute paths.

Validator is configurable to add more environments variable that should trigger it (albeit repair action will only resolve `$HIP` and `$JOB`)


m-u-r-p-h-y (approved):
workfile publishing works as expected.

published workfile visible in workfile loader.

image

iLLiCiTiT (changes requested):
There is again added submodule `OpenColorIO-Configs`

EDITED: Fixed

PR #3700, opened 2022-08-22
Settings: Lock settings UI session

iLLiCiTiT:
## Brief description
Handle opened sessions of settings UI and last saved settings, to avoid unwanted loss of settings values when multiple people are modifying setting at the same time.

## Description
First who opens settings will create lock in mongo (OP version specific lock). From that moment anyone who will open settings UI will see message saying that other session of settings is opened with information about username and machine where it's happening. The dialog give options to open settings in view only mode or take control over the settings. In view mode is save disabled and when control is taken the lock is changed to the user who did it. From that moment the previous locked user can't save with only option to reset settings, is shown with dialog and information about person who took the control.

Also was added one more information to saved overrides which is information about last save. That is also used before save to validate if last saved settings are the same as at the start of session. This is for cases when settings are modified out of settings UI. Lock does not work for project anatomy which is versionless and it's changes are not always propagated using settings handlers.

## Testing notes:
1. Open settings with `~openpype/tools/run_settings` script
2. Open other session of settings (script or from tray)
3. The second session should show a dialog
From this point you can "play" with possible scenarios and try to save or refresh settings.

### DIFF
--- a/openpype/settings/handlers.py
+++ b/openpype/settings/handlers.py
@@ -453,13 +453,12 @@ class CacheValues:
             return {}
         return copy.deepcopy(self.data)
 
-    def update_data(self, data, version=None):
+    def update_data(self, data, version):
         self.data = data
         self.creation_time = datetime.datetime.now()
-        if version is not None:
-            self.version = version
+        self.version = version
 
-    def update_from_document(self, document, version=None):
+    def update_from_document(self, document, version):
         data = {}
         if document:
             if "data" in document:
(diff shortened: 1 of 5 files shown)


m-u-r-p-h-y (approved):
works as expected.


I would suggest changing the window title to `Open Pype Settings [View Only]` and/or some other visual clue (color change)

image

BigRoy:
Might also be worth stripping the milliseconds from that time stamp for readability?

iLLiCiTiT:
> I would suggest changing the window title to Open Pype Settings [View Only] and/or some other visual clue (color change) in case of going to read only mode.

Changed. Not sure it window title change is propagated on all OS properly (when Qt is used).

> Might also be worth stripping the milliseconds from that time stamp for readability?

Used `pretty_date` from version delegate.

64qam (approved):

PR #3701, opened 2022-08-22
Unreal: Define unreal as module and use host class

iLLiCiTiT:
## Brief description
Unreal is OpenPype module and it's implementation registered as object inherited from `HostBase`.

## Description
Unreal in `openpype.hosts` works as openpype module and unreal integration is registered using `UnrealHost` object instead of `openpype.hosts.unreal.api` module.

## Testing notes:
1. Unreal should initialize project as usually
2. Unreal should launch successfully
3. Integration works as did (with publishing)

### DIFF
--- a/openpype/hosts/unreal/api/pipeline.py
+++ b/openpype/hosts/unreal/api/pipeline.py
@@ -14,6 +14,7 @@ from openpype.pipeline import (
 )
 from openpype.tools.utils import host_tools
 import openpype.hosts.unreal
+from openpypr.host import HostBase, ILoadHost
 
 import unreal  # noqa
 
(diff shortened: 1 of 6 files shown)


openpype/hosts/unreal/api/pipeline.py
```diff
@@ -14,6 +14,7 @@
 )
 from openpype.tools.utils import host_tools
 import openpype.hosts.unreal
+from openpypr.host import HostBase, ILoadHost
```
simonebarbieri, line 17: There's a typo here that does not allow openpype to be loaded correctly. Fixed that, everything seems to be working fine!

simonebarbieri (changes requested):

simonebarbieri (approved):

PR #3709, opened 2022-08-23
Maya: Use project name instead of project code

iLLiCiTiT:
## Brief description
Use project name for root preparation instead of project code.

## Description
Using project code can cause that anatomy won't be created for right project so root values will be invalid.

## Testing notes:
1. Loading in Maya should work


BigRoy:
Testing notes:

- This would only be relevant when testing with project settings enabled for `maya.maya-dirmap.use_env_var_as_root`.

kalisp (approved):
Reference Loaders work, reopening and resolving works too.

PR #3710, opened 2022-08-23
Loader: Subset groups using client operations

iLLiCiTiT:
## Brief description
Use client opeartions to change subset group on subset documents.

## Testing notes:
1. Open any host on project where you have something published
2. Open loader tool in the host
3. Select subset/s
4. Hit Ctrl + G
5. Change group of subset
6. Group change should be propagated

### DIFF
--- a/openpype/tools/loader/model.py
+++ b/openpype/tools/loader/model.py
@@ -272,11 +272,13 @@ class SubsetsModel(TreeModel, BaseRepresentationModel):
 
                 # update availability on active site when version changes
                 if self.sync_server.enabled and version_doc:
-                    repre_info = self.sync_server.get_repre_info_for_versions(
-                        project_name,
-                        [version_doc["_id"]],
-                        self.active_site,
-                        self.remote_site
+                    repre_info = list(
+                        self.sync_server.get_repre_info_for_versions(
+                            project_name,
+                            [version_doc["_id"]],
+                            self.active_site,
+                            self.remote_site
+                        )
                     )
                     if repre_info:
                         version_doc["data"].update(
--- a/openpype/tools/loader/model.py
+++ b/openpype/tools/loader/model.py
@@ -272,7 +272,7 @@ class SubsetsModel(TreeModel, BaseRepresentationModel):
 
                 # update availability on active site when version changes
                 if self.sync_server.enabled and version_doc:
-                    repre_info = list(
+                    repres_info = list(
                         self.sync_server.get_repre_info_for_versions(
                             project_name,
                             [version_doc["_id"]],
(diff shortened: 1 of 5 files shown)


openpype/tools/loader/widgets.py
```diff
@@ -614,26 +615,30 @@ def on_context_menu(self, point):
...
 
     def group_subsets(self, name, asset_ids, items):
-        field = "data.subsetGroup"
+        subset_ids = {
+            item["_id"]
+            for item in items
+            if item.get("_id")
+        }
+        if not subset_ids:
+            return
```
BigRoy, line 624: Does this also work when multi-selecting (multiple assets) in the loader left hand side and then trying to group on the subsets? I feel like the old code might have been that way to support that case - but I'm not sure?

iLLiCiTiT: Good point, testing.

iLLiCiTiT: Yes it work as expected, and found other unrelated bug :)

64qam (approved):

openpype/tools/loader/model.py
```diff
@@ -272,11 +272,13 @@ def setData(self, index, value, role=QtCore.Qt.EditRole):
...
-                        self.active_site,
-                        self.remote_site
+                    repre_info = list(
+                        self.sync_server.get_repre_info_for_versions(
+                            project_name,
+                            [version_doc["_id"]],
+                            self.active_site,
+                            self.remote_site
+                        )
                     )
```
BigRoy, line 282: Wouldn't `next` make more sense here since you always just want the first entry?

openpype/tools/loader/model.py
```diff
@@ -272,15 +272,17 @@ def setData(self, index, value, role=QtCore.Qt.EditRole):
...
-                        project_name,
-                        [version_doc["_id"]],
-                        self.active_site,
-                        self.remote_site
+                    repres_info = list(
+                        self.sync_server.get_repre_info_for_versions(
+                            project_name,
+                            [version_doc["_id"]],
+                            self.active_site,
+                            self.remote_site
```
BigRoy, line 280: Also, does this logic fail to update when switching from a version to a hero version (where the hero version is not the same version number as the one that was previously selected?)

openpype/tools/loader/model.py
iLLiCiTiT: To be honest I don't like using `next` for such long expressions especially when it would make it even longer.
```
repre_info = next(
    self.sync_server.get_repre_info_for_versions(
        project_name,
        [version_doc["_id"]],
        self.active_site,
        self.remote_site
    ),
    None
)
```

openpype/tools/loader/model.py
iLLiCiTiT: It's based on version id so ti should be fine.

BigRoy: Except that it only triggers in the else statement just above. Right?

iLLiCiTiT: Hmm you're right. But that's out of bounds of this PR.

iLLiCiTiT: Created issue https://github.com/pypeclub/OpenPype/issues/3711

PR #3712, opened 2022-08-23
General: Added helper getters to modules manager

iLLiCiTiT:
## Brief description
Added helper methods to ModulesManager to access modules by name.

## Description
Added dictionary like access to module by their names and special method to get module by name which also checks if is module enabled.

## Testing notes:
New methods on created object of `ModulesManager` in different cases (e.g. in Tray > Admin > Console).

```
from openpype.modules import ModulesManager

manager = ModulesManager()
print(manager["project_manager"])
print(manager.get("project_manager"))
print(manager.get_enabled_module("project_manager"))
```

### DIFF
--- a/openpype/modules/base.py
+++ b/openpype/modules/base.py
@@ -562,6 +562,40 @@ class ModulesManager:
         self.initialize_modules()
         self.connect_modules()
 
+    def __getitem__(self, module_name):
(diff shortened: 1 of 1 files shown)


kalisp (approved):
Both new ones work as expected.

PR #3713, opened 2022-08-23
General: Move constants from lib to client

iLLiCiTiT:
## Brief description
Moved entity specific constants to client code or reuse those which are already there.

## Description
These constants are related to creation of new entities. Some of them were already available there. 

## Additional info
Hard to tell to which degree these constants are "constants" and should be defined at one place? They're connected to schema validations but schema validation is in pipeline part. I would say that logic related to validation of project entities should be moved to client side and rest should stay in pipeline. I would like to avoid changing it and discussing it in this PR as this is refactor PR.

Variable `PROJECT_NAME_ALLOWED_SYMBOLS` is not used in `avalon_context.py` but is imported from there in `__init__.py`. I would like to keep it until the backwards compatibility is removed.

## Testing notes:
Technically logic didn't change.
- project manager can create and update assets
- ftrack sync works

### DIFF
--- a/openpype/client/operations.py
+++ b/openpype/client/operations.py
@@ -1,3 +1,4 @@
+import re
 import uuid
 import copy
 import collections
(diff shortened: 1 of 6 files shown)


kalisp (approved):
Project manager works, sync to avalon too.

PR #3714, opened 2022-08-23
General: Move creators plugin getter

iLLiCiTiT:
## Brief description
Moved `get_creator_by_name` function into `openpype.pipeline.create.creator_plugins` and renamed to `get_legacy_creator_by_name`.

## Testing notes:
Creation during loading in Blender and Maya should work as before.

### DIFF
--- a/openpype/hosts/blender/plugins/load/load_layout_blend.py
+++ b/openpype/hosts/blender/plugins/load/load_layout_blend.py
@@ -6,12 +6,12 @@ from typing import Dict, List, Optional
 
 import bpy
 
-from openpype import lib
 from openpype.pipeline import (
     legacy_create,
     get_representation_path,
     AVALON_CONTAINER_ID,
 )
+from openpype.pipeline.create import get_legacy_creator_by_name
 from openpype.hosts.blender.api import plugin
 from openpype.hosts.blender.api.pipeline import (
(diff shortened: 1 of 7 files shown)


BigRoy:
Could you show me how I could get the non legacy creators by name? (Or basically what the new equivalent is of this functionality)

iLLiCiTiT:
> Could you show me how I could get the non legacy creators by name? (Or basically what the new equivalent is of this functionality)

Creator plugins are wrapped into CreateContext. Creators can be accessed by their identifier from there. So there is not direct equivalent of this functionality.

Programmatically creation is not "ideal" now but ready to be added. The development focus was and still is elsewhere.

```


context = CreateContext(registered_host())
render_creator = context.creators["render"]
...
# Do some stuff
...
context.save_changes()
```

BigRoy:
> Creator plugins are wrapped into CreateContext. Creators can be accessed by their identifier from there. So there is not direct equivalent of this functionality.
> Programmatically creation is not "ideal" now but ready to be added. The development focus was and still is elsewhere.

Thanks - that helps. That still looks painful, yes. Moving to legacy seem fine - we'll just need to make sure not to mark it deprecated until the new system is at the same ease of use.

simonebarbieri (approved):

PR #3716, opened 2022-08-24
Settings: Fix studio default anatomy save

iLLiCiTiT:
## Brief description
It is possible to save changes in studio default anatomy.

## Description
Add information about last save into studio default project anatomy. Right now it's crashing because the last save information is missing on cache.

## Additional information
The last save information is not used from cache but it is probably good idea to store that information too.

## Testing notes:
1. Open settings UI
2. Go to Project tab
3. Keep default project and change anything in Anatomy category
4. Hit save
5. Save should work fine


m-u-r-p-h-y (approved):
settings are saved correctly, without any error

PR #3717, opened 2022-08-24
Hiero: Define hiero as module

iLLiCiTiT:
## Brief description
Hiero is defined as OpenPype module and fix protobuf issue.

## Description
Added `HieroModule` for hiero host which cares about handling hiero host related code. Also was added `protobuf 3.20.1` into hiero vendor. That's because we override hiero's inner site-packages with ours where we have `protobuf 4.21.5` which breaks hiero on start (issue for Hiero/Nukestudio 13.*).

## Additional information
The protobuf issue should be maybe fixed by changing version in pyproject.toml but that would require new build. Question is if this is bad solution or not.

## Testing notes:
1. Hiero launch should work
2. Last workfile should be opened if is available and feature is enabled
3. Hiero and NukeStudio 13.0 and higher should not crash

Resolves https://github.com/pypeclub/OpenPype/issues/3715


64qam (approved):

m-u-r-p-h-y (approved):
tested with Nuke 13.2.v3, Hiero 13.2.v3

PR #3722, opened 2022-08-24
git: update gitignore

jakubjezek001:
## Brief description
Adding developer tools ignoring

## Description
all developer tools in /tools folder starting with prefix `dev_` will be ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -107,3 +107,6 @@ website/.docusaurus
 mypy.ini
 
 tools/run_eventserver.*
+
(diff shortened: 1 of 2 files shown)


mkolar:
Is this something that you're using locally? I have no problem with it at all, just wondering where it's coming from

antirotor (changes requested):
I have no problem with it too, but I would mention it somewhere in README.md or documentation?

antirotor (approved):

PR #3723, opened 2022-08-24
Genetal: Move subset name functionality

iLLiCiTiT:
## Brief description
Moved functions related to subset names into `openpype.pipeline.create`.

## Description
Functions `get_subset_name` and `get_subset_name_with_asset_doc` in `openpype.lib` are marked as deprecated. Both are calling new function in `openpype.pipeline.create.subset_name` called `get_subset_name` which is copy of `get_subset_name_with_asset_doc`.

## Additional info
Because `get_subset_name` is not used it was not added to new location at all and `get_subset_name_with_asset_doc` was renamed to `get_subset_name`. New function can expect project settings to make calculations faster.

I moved the logic under "create" category because it is used for instance creation. But it's also used in publish plugins when instances are "auto-created" (which is still technically creation?).

I've named the file `subset_name.py` because I didn't know where elso to put it. I think it's more lib function for creation because at the end subset name is defined by creators. Let comments if you agree with having `subset_name.py` or use different file. At this moment is content not directly imported so it would be easy to change.

## Testing notes:
1. Creation of instances should work in any host and subset name should be filled correctly.


64qam (approved):
Tested in Nuke, PS, AE

BigRoy:
_I had to change that PR title because - oh boy - I read that the wrong way._

PR #3724, opened 2022-08-24
Nuke: Define nuke as module

iLLiCiTiT:
## Brief description
Nuke is defined as OpenPype module and fix protobuf issue.

## Description
Added `NukeModule` for nuke host which cares about handling nuke host related code. Also was added `protobuf 3.20.1` into nuke vendor. That's because we override nuke's inner site-packages with ours where we have `protobuf 4.21.5` which may break nuke on start (issue for Nuke/NukeX 13.*).

## Testing notes:
1. Nuke/NukeX launch should work
2. Last workfile should be opened if is available and feature is enabled
3. Nuke/NukeX 13.* and higher should not crash on launch (it was crashing on 13.2)


m-u-r-p-h-y (approved):
Tested with 13.2.v3 and just to make sure older 12.2v9

PR #3725, opened 2022-08-24
General: Replace PypeLogger with Logger

iLLiCiTiT:
## Brief description
Code cleanup related to OpenPype logger.

## Description
Class `PypeLogger` should not be used anymore, is marked as deprecated and logs out warning message when `get_logger` is called. It's usage was replaced with `Logger`. Also was removed calling of the class which is not needed. In some cases was logger object moved into objects from global scope.

Most changes that could cause issues happened in sync server.

## Testing notes:
1. Logging should work as expected
2. SyncServer sync should work

Resolves https://github.com/pypeclub/OpenPype/issues/3581

### DIFF
--- a/openpype/client/entities.py
+++ b/openpype/client/entities.py
@@ -1455,7 +1455,7 @@ def get_workfile_info(
 """
 ## Custom data storage:
 - Settings - OP settings overrides and local settings
-- Logging - logs from PypeLogger
+- Logging - logs from Logger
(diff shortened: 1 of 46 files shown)


kalisp (approved):
Site Sync works, Adobe, Maya seems to work.
Was clicking around, couldn't find an issue.
Didn't test Deadline publishing though!

PR #3729, opened 2022-08-24
Blender: Define blender as module

iLLiCiTiT:
## Brief description
Blender host is OpenPype module.

## Description
Implemented `BlenderModule` for blender host which gives public information about the host.

## Testing notes:
- Blender can be launched
- Integration works
    - PySide2 is auto install when missing
- Other addons are registered too
- Last workfile is launched if available and feature is enabled

### DIFF
--- a/openpype/hosts/blender/__init__.py
+++ b/openpype/hosts/blender/__init__.py
@@ -1,52 +1,6 @@
-import os
+from .module import BlenderModule
 
 
-def add_implementation_envs(env, _app):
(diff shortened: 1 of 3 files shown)


64qam (changes requested):
OpenPype is not integrated into the Blender menu bar.

iLLiCiTiT:
> OpenPype is not integrated into the Blender menu bar.

I guess that's because I've named `blender` as `nuke` :D

64qam (approved):

PR #3732, opened 2022-08-25
Flame: Defined flame as addon

iLLiCiTiT:
## Brief description
Flame host is addon.

## Description
Implemented `FlameAddon` for flame host which gives public information about the host.

## Testing notes:
- Flame can be launched and integration works (as before)


jakubjezek001 (approved):
works as usual

PR #3735, opened 2022-08-25
Houdini: Define houdini as addon

iLLiCiTiT:
## Brief description
Houdini host is addon.

## Description
Implemented `HoudiniAddon` for houdini host which gives public information about the host. Fixed adding vendor to sys.path. Modified how current directory is set for launched houdini.

## Testing notes:
- Houdini can be launched and integration works (as before)


antirotor (approved):
Tested in 19.5.303 and it works!

PR #3739, opened 2022-08-25
Kitsu|Change: Drop 'entities root' setting.

### LINKED ISSUE #3738: Kitsu module: Drop entities root settings
### Discussed in https://github.com/pypeclub/OpenPype/discussions/3545

<div type='discussions-op-text'>

<sup>Originally posted by **Tilix4** July 20, 2022</sup>
In settings, you can customize the entities root folders for `assets` and `shots`:
image

This allows you to create a tree of folders where all assets or shots will be put under. The issue is, I don't see any use case where we will need more than `Assets/` or `Shots/` to put all the related entities.
Moreover, that's how things are classified in Kitsu: 
image

Therefore, I propose to drop this settings feature to keep a simple `Assets` and `Shots`, not customizable.
image
</div>

[cuID:OP-3831]

Tilix4:
## Brief description
Removed "entities root" setting to match with Kitsu's native behaviour without adding overhead in code, neither for admin.

Closes #3738

### DIFF
--- a/openpype/modules/kitsu/utils/update_op_with_zou.py
+++ b/openpype/modules/kitsu/utils/update_op_with_zou.py
@@ -166,50 +166,21 @@ def update_op_assets(
 
         # Substitute item type for general classification (assets or shots)
         if item_type in ["Asset", "AssetType"]:
-            substitute_item_type = "assets"
+            entity_root_asset_name = "Assets"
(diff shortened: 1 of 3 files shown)


iLLiCiTiT (approved):
We don't have capacity to test this PR but changes make sense and are reasonable.

PR #3741, opened 2022-08-25
General: Logger tweaks

iLLiCiTiT:
## Brief description
Fix logger issue in timers manager and modify logs in sync server.

## Description
Timers manager rest api codebase is using wront attribute name during initialization for logger. Sync server provides have their own loggers so they don't have to use loggers in global scope.

## Testing notes:
1. TimersManager can register webserver callbacks
    - can start/stop timer from host (e.g. on task change in workfiles tool)
2. Sync server provides should not crash and still can log


kalisp (approved):
Tested Site Sync, Timers, works.

PR #3742, opened 2022-08-25
General: Custom workfile template use env keys to fill path

maxpareschi:
## Brief description
Allow env vars to be parsed in workfile template path resolving

## Description
The current workfile template path resolving matches either an absolute path or just the anatomy context.
This extends the anatomy context used by the parser to also contain all the os.environ vars
using dict.update (so there's no conflict with keys)
Priority would be in environ then, which makes this flexible.

## Testing notes:
1. put an env var in the workfiles template path
2. seems to work!

### DIFF
--- a/openpype/pipeline/workfile/path_resolving.py
+++ b/openpype/pipeline/workfile/path_resolving.py
@@ -408,6 +408,9 @@ def get_custom_workfile_template(
     # add root dict
     anatomy_context_data["root"] = anatomy.roots
 
+    # extend anatomy context with os.environ
+    anatomy_context_data.update(os.environ)
+
     # get task type for the task in context
     current_task_type = anatomy_context_data["task"]["type"]
 
--- a/openpype/pipeline/workfile/path_resolving.py
+++ b/openpype/pipeline/workfile/path_resolving.py
@@ -409,7 +409,7 @@ def get_custom_workfile_template(
     anatomy_context_data["root"] = anatomy.roots
 
     # extend anatomy context with os.environ
-    full_context_data = os.environ
+    full_context_data = os.environ.copy()
     full_context_data.update(anatomy_context_data)
 
     # get task type for the task in context
--- a/openpype/pipeline/workfile/path_resolving.py
+++ b/openpype/pipeline/workfile/path_resolving.py
@@ -409,10 +409,11 @@ def get_custom_workfile_template(
     anatomy_context_data["root"] = anatomy.roots
 
     # extend anatomy context with os.environ
-    anatomy_context_data.update(os.environ)
+    full_context_data = os.environ
+    full_context_data.update(anatomy_context_data)
 
     # get task type for the task in context
-    current_task_type = anatomy_context_data["task"]["type"]
+    current_task_type = full_context_data["task"]["type"]
 
     # get path from matching profile
     matching_item = filter_profiles(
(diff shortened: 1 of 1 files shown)


openpype/pipeline/workfile/path_resolving.py
```diff
@@ -408,6 +408,9 @@ def get_custom_workfile_template(
     # add root dict
     anatomy_context_data["root"] = anatomy.roots
 
+    # extend anatomy context with os.environ
+    anatomy_context_data.update(os.environ)
```
BigRoy, line 412: At the very least I'd reverse the logic so that if any anatomy context data would never be overwritten if one were to match a key in `os.environ` by accident. Just to foolproof it?

maxpareschi: done with that, i just got os.environ and updated it with anatomy afterwards. check it out!

openpype/pipeline/workfile/path_resolving.py
```diff
@@ -408,8 +408,12 @@ def get_custom_workfile_template(
     # add root dict
     anatomy_context_data["root"] = anatomy.roots
 
+    # extend anatomy context with os.environ
+    full_context_data = os.environ
+    full_context_data.update(anatomy_context_data)
```
BigRoy, line 413: Ouch - this will actually update your local `os.environ` data. You'll need to work with a copy of the dict instead!

```suggestion
    full_context_data = os.environ.copy()
    full_context_data.update(anatomy_context_data)
```
OR:
```suggestion
full_context_data = dict(os.environ, **anatomy_context_data)
```

maxpareschi: good catch, i like the copy approach. merged in.

openpype/pipeline/workfile/path_resolving.py
```diff
@@ -408,8 +408,12 @@ def get_custom_workfile_template(
     # add root dict
     anatomy_context_data["root"] = anatomy.roots
 
+    # extend anatomy context with os.environ
+    full_context_data = os.environ.copy()
+    full_context_data.update(anatomy_context_data)
+
```
BigRoy, line 414: I hate to be the one to provide more comments after you've made fixes - sorry, I tend to glimpse over PRs like this too quickly on a first read. But looking at the code we can actually move this logic way further down, even into the if statement for when a matching item is found. That way the environment is only ever copied on a match and the logic is closer to where it's used.

It would turn the logic into something like this:
```python
    # get task type for the task in context
    current_task_type = anatomy_context_data["task"]["type"]

    # get path from matching profile
    matching_item = filter_profiles(
        template_profiles,
        {"task_types": current_task_type}
    )
    # when path is available try to format it in case
    # there are some anatomy template strings
    if matching_item:
        # extend anatomy context with os.environ to allow
        # formatting environment keys in template path
        full_context_data = os.environ.copy()
        full_context_data.update(anatomy_context_data)
    
        template = matching_item["path"][platform.system().lower()]
        return StringTemplate.format_strict_template(
            template, full_context_data
        ).normalized()
```

maxpareschi: one fix coming right up!

BigRoy:
Thanks - solid work, nice compact change like this too. I wonder what OP crew will say about the functionality and whether this patch is in the right place and doing "what's expected" for them as well.

iLLiCiTiT (approved):
Looks good

PR #3756, opened 2022-08-30
FIx: version resolution from string

antirotor:
## Fix

Fix version detection in string. This fixed error could break running `create_zip` command in case the version of OP was "pure" - `3.15.1` (instead of `3.15.1-nightly.1`).


antirotor:
This should be probably based on develop branch

PR #3757, opened 2022-08-30
Fix version resolution

antirotor:
## Fix

This is fixing edge case where if version of OP is "pure" - ie. `3.14.1`  then `OpenPypeVersion.version_in_str()` will fail (it will return None). This mainly affects `create_zip` script and probably finding versions.

This is change in Igniter, so in theory it should go to new minor release but because it has low impact and is useful for subsequent patch releases, let's put it to develop.


64qam (approved):

tools/create_zip.ps1
```diff
@@ -96,6 +96,7 @@ Write-Color -Text ">>> ", "Cleaning cache files ... " -Color Green, Gray -NoNewl
 Get-ChildItem $openpype_root -Filter "__pycache__" -Force -Recurse|  Where-Object {( $_.FullName -inotmatch '\\build\\' ) -and ( $_.FullName -inotmatch '\\.venv' )} | Remove-Item -Force -Recurse
 Get-ChildItem $openpype_root -Filter "*.pyc" -Force -Recurse | Where-Object {( $_.FullName -inotmatch '\\build\\' ) -and ( $_.FullName -inotmatch '\\.venv' )} | Remove-Item -Force
 Get-ChildItem $openpype_root -Filter "*.pyo" -Force -Recurse | Where-Object {( $_.FullName -inotmatch '\\build\\' ) -and ( $_.FullName -inotmatch '\\.venv' )} | Remove-Item -Force
+Write-Color -Text "OK" -Color Green
```
BigRoy, line 99: Is this really related to this PR?

antirotor: it's not, but it was missing new line there and I've noticed it when fixing it. Nothing crutial and I was really lazy to create separate PR jus for this.

PR #3759, opened 2022-08-30
Maya: Refactor submit deadline to use AbstractSubmitDeadline

BigRoy:
## Brief description

Refactor submit deadline to use AbstractSubmitDeadline.

## Description

Current state is functional in that it's able to do a straight maya submission (and I was able to submit a Tile Render submission, but haven't tested whether the tile assembly came out as needed yet).

I'll leave as "Draft" for the time being as I expect to still perform some fixes on my end that I can probably test easily soon.

## Additional info

Yes, this is a massive PR and likely breaks on many things - but I wanted to start raising discussions on approach. I can try to separate into standalone smaller PRs that first cleanup the old submitter and then refactor - but there will always be a tipping point where it needs to be refactored to the new logic and we'll still need to end up testing the new functionality. Thus I feel it might be worth just cleaning the monster ones and test thoroughly the different cases.

Especially because likely many cases (Tile Rendering, V-Ray submissions, Arnold submissions) were previously untested in their latest state too and might have been broken for some time.

## Testing notes:

1. Open Maya, start submitting things.
2. Open other hosts that use the AbstractDeadlineSubmitter as base - and make sure they still work too.


openpype/modules/deadline/abstract_submit_deadline.py
```diff
@@ -66,6 +67,58 @@ def requests_get(*args, **kwargs):
...
+        self[index] = other
+        return self
+
+    def __setitem__(self, key, value):
+        if not isinstance(key, int):
+            raise TypeError("Key must be an integer: {}".format(key))
+
+        if key < 0:
+            raise ValueError("Negative index can't be set: {}".format(key))
+        dict.__setitem__(self, key, value)
```
BigRoy, line 119: Likely a better implementation can be build for this, but I wanted to make it easy to set specific indices whilst also allowing the ability to easily append values like was done before with the old setters/getters.
```python
# Append value
job_info.EnvironmentKeyValue += "key=value"
# Query index (must exist)
job_info.EnvironmentKeyValue[0]
# Get index
job_info.EnvironmentKeyValue.get(0)
# Set by index
job_info.EnvironmentKeyValue[0] = "hello"
```

Setting/querying specific indices might not be super relevant for `EnvironmentKeyValue` but it is instead for `ExtraExtraInfoKeyValue`.

I'd have loved if the implementation would also allow to "uniqify" added keys (e.g. for `EnvironmentKeyValue` or `AssetDependency`) but that isn't super crucial.

m-u-r-p-h-y (approved):

openpype/hosts/maya/plugins/publish/collect_render.py
```diff
@@ -293,6 +293,7 @@ def process(self, context):
                 "source": filepath,
                 "expectedFiles": full_exp_files,
                 "publishRenderMetadataFolder": common_publish_meta_path,
+                "renderProducts": layer_render_products,
```
BigRoy, line 296: Note that this is **not** JSON serializable and thus the logging below had to be removed to avoid an error. Is this a problem?

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
```diff
@@ -1121,7 +640,7 @@ def smart_replace(string, key_values):
 
         return result
 
-    def _patch_workfile(self, file, patches):
+    def _patch_workfile(self):
         # type: (str, dict) -> [str, None]
```
antirotor, line 644: type hint doesn't match now

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
```diff
@@ -863,65 +467,38 @@ def _get_maya_payload(self, data):
...
-                int(self._instance.data["frameEndHandle"])),
-
-            "Plugin": self._instance.data.get(
-                "mayaRenderPlugin", "MayaPype"),
-            "FramesPerTask": self._instance.data.get("framesPerTask", 1)
-        }
 
-        plugin_info_ext = {
-            # Renderer
+        plugin_info = {
```
antirotor, line 494: just note that we should probably move all those to the similar class like `MayaPluginInfo`

antirotor:
so far it looks very good, I would like to go over it and remove form it ass and vrscene exporting jobs because I don't think they currently work and it will clean up the code.

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
BigRoy: These are actually PluginInfo for different plug-ins across the file. So this is one for the Tile Assembler, others I believe are for VRay plugin and Arnold plugin. But yes, it'd be more consistent if each would get their own plugin info accordingly.

BigRoy:
> so far it looks very good, I would like to go over it and remove form it ass and vrscene exporting jobs because I don't think they currently work and it will clean up the code.

I'm fine with removing it. However, might be more work to add them again later instead of fixing now? Thoughts?

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
BigRoy: I've now refactored them to their own PluginInfo data classes - I wonder if that's more readable now. It feels a lot cleaner like this.

openpype/modules/deadline/abstract_submit_deadline.py
```diff
@@ -218,146 +309,38 @@ class DeadlineJobInfo(object):
...
-            out["OutputDirectory{}".format(index)] = v
-        return out
-
-    @OutputDirectory.setter
-    def OutputDirectory(self, val):  # noqa: N802
-        self._outputDirectory.append(val)
+    OutputFilename = attr.ib(factory=partial(DeadlineIndexedVar,
+                                             "OutputFilename"))
+    OutputFilenameTile = attr.ib(factory=partial(DeadlineIndexedVar,
+                                                 "OutputFilename{}Tile"))
```
BigRoy, line 336: Anyone know of a better more pythonic way to initialize these with `attr`? I feel like I'm missing something obvious.

openpype/modules/deadline/abstract_submit_deadline.py
```diff
@@ -381,23 +364,35 @@ def serialize(self):
...
+            self.EnvironmentKeyValue,
+            self.ExtraInfo,
+            self.ExtraInfoKeyValue,
+            self.TaskExtraInfoName,
+            self.OutputFilename,
+            self.OutputFilenameTile,
+            self.OutputDirectory,
+            self.AssetDependency
+        ]:
+            serialized.update(attribute.serialize())
```
BigRoy, line 387: Does anyone know if it's possible to implement a method on a class so `attr.as_dict` knows how to serialize it? E.g. a `__json__` method or alike?

BigRoy: I guess the right way would be to pass [`value_serializer` to `attr.todict](https://www.attrs.org/en/stable/extending.html?highlight=serialize#customize-value-serialization-in-asdict)

kalisp (approved):
Tested and works in AE.

openpype/modules/deadline/abstract_submit_deadline.py
iLLiCiTiT: @kalisp or @antirotor do you know?

m-u-r-p-h-y:
I have tested Maya submits

- redshift
- arnold
- render man

submitting itself is passing through correctly. 

I'm just curious why the Deadline integrator is falling back to use workfiles, instead of published ones.
image

I double checked my project settings and it set to defaults to use published one . . .

image

BigRoy:
> I'm just curious why the Deadline integrator is falling back to use workfiles, instead of published ones.

Thank you very much - that was a bug! Fixed with https://github.com/pypeclub/OpenPype/pull/3759/commits/49dff63f08207eea0218cf37e4824795d08e3895

m-u-r-p-h-y (approved):
submitting to Deadline works and using published workfile correctly now.  good job!


image

PR #3768, opened 2022-08-31
General: Move create project folders to pipeline

iLLiCiTiT:
## Brief description
Moved project folder creation functions into `openpype.pipeline`.

## Description
Moved workdir extra folders creation to workfile api and default project structure into new file `project_structure.py`. Their logic was a little bit tweaked. Lib functions were kept and marked as deprecated with new functions locations (if there is any).

## Additional info
Functionality of project folders creation is based on pipeline logic (e.g. Anatomy) so moving it from general lib is logical step. If you have any better idea for the filename (instead of `project_structure.py`) please write it to comments.

## Testing notes:
- [ ] Extra folders in workdir should be created if `project_settings/global/tools/Workfiles/extra_folders` has defined any
- [ ] Default project structure can be created from ftrack and from project manager


iLLiCiTiT:
> Crreate Starting Folders in Project Manager throws this:

Should be fixed now...

PR #3770, opened 2022-08-31
General: Move queries of asset and representation links

iLLiCiTiT:
## Brief description
Moved function to get and extract linked ids and entities into client.

## Description
Moved functions to get asset and representation links from `openpype.lib` to `openpype.client`.

## Testing notes:
1. Adding workfile to sync through loader should add to sync also related representations
2. Get linked assets should work (not sure if we're able to set them?)


kalisp (approved):

openpype/client/entities.py
```diff
@@ -32,17 +32,37 @@ def _prepare_fields(fields, required_fields=None):
...
+def convert_ids(in_ids):
+    """Helper function for conversion of ids from string to ObjectId.
+
+    Args:
+        in_ids (Iterable[Union[str, ObjectId, Any]]): List of entity ids that
+            should be converted to right type for queries.
+
+    Returns:
+        List[ObjectId]: Converted ids to ObjectId.
+    """
```
BigRoy, line 60: I feel like this docstring should mention that the input ids gets uniqified?

BigRoy:
Based on these functions it's a bit unclear to me how to get output links? E.g. find all links for which representation or version is used.

openpype/client/entities.py
```diff
@@ -32,17 +32,37 @@ def _prepare_fields(fields, required_fields=None):
...
+        in_id (Union[str, ObjectId, Any]): Entity id that should be converted
+            to right type for queries.
+
+    Returns:
+        Union[ObjectId, Any]: Converted ids to ObjectId or in type.
+    """
+
     if isinstance(in_id, six.string_types):
         return ObjectId(in_id)
     return in_id
```
BigRoy, line 48: Shouldn't this also ensure the returned type is actually `ObjectId`.
If I'd pass an `int` or `float` to the function it would come out exactly the same as an `int` or `float`.

Also the `in_id` type hint thus seems to broad with `Any` in there. No?

iLLiCiTiT:
> Based on these functions it's a bit unclear to me how to get output links? E.g. find all links for which representation or version is used.
> 
> Didn't we have that functionality somewhere? I recall seeing that.

I agree it's unclear and it's confusing but goal of this PR is to move functions from `openpype.lib` to different place, with smaller tweaks like more arguments options to be called with or small modification of docstrings. If it's unclear now, it was probably even before this PR :)

BigRoy:
> If it's unclear now, it was probably even before this PR :)

I just recall having seen a function for that before. I think I may have been confused with this [`get_output_link_versions`](https://github.com/pypeclub/OpenPype/blob/377df371db52412f9485272da413448ef647107c/openpype/client/entities.py#L790)?

PR #3777, opened 2022-09-01
Fix - changed format of version string in pyproject.toml

kalisp:
## Brief description
Since Poetry 1.2 version in .toml must follow PEP-440, in our case there must not be dash sign. Plus sign is valid though.

## Additional info
Version in pyproject.toml is not used anywhere yet (only maybe when uploading wheel to pip, which is not really used).


64qam (approved):

tools/ci_tools.py
```diff
@@ -101,7 +101,7 @@ def bump_file_versions(version):
 
     # bump pyproject.toml
     filename = "pyproject.toml"
-    regex = "version = \"(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-((0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*))?\" # OpenPype"
+    regex = "version = \"(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(\+((0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*))?\" # OpenPype"
```
antirotor, line 104: just a note that in semver, you can have `+` for the build number. This regex won't match `
image

antirotor (approved):
It fixes build problems even if it is leaving proper PEP-440 resolution open, so I am for merging it.

PR #3792, opened 2022-09-06
Hiero: retimed clip publishing is working

jakubjezek001:
## Brief description
Retimed clips for lest then 100% is working now.

## Description
Publishing is working as expected.

## Additional info
Precollect instances was falsely detecting timeline range for otio clip data.

## Testing notes:
1. open Hiero with testing project
2. create clip with retime for about 80% speed
3. Create publishable clip from the 2.
4. publish and see that Precollect intances will not fail


jrsndl (approved):
Works as advertised. Tested in Hiero 12.2v3, with speeds 12%, 80%, 85%, 191%

PR #3806, opened 2022-09-07
Enhancement/Ftrack ignore OP templating str on folder create

maxpareschi:
## Brief description
Make Ftrack ignore OP templating strings (eg. {root[work]}) on project creation

## Description
Ftrack does not like names with {}, but recognizes them as folders on folder creation and messes up everything. This re.match ignores OP templating stings on folder creation.
This avoids a bug in ftrack concerning the {}, and also circumvents the recursive creation of folders in multiple roots (which seems to me like an OP bug)

Is this really the correct way though? maybe templating would be nice to have as a feature. Otherwise we would probably need to modify the folder creation logic to work only on a single root when using "__project_root__" as a key?

## Testing notes:
[Discord Messages](https://discord.com/channels/517362899170230292/563751989075378201/1017104897293951106)


maxpareschi:
is no one interested in this?

ddesmond:
I have the same issue with doubling the folder structure when Ftrack syncs.
Has this been tested? i still see double folder when i change those 3 lines.
Can anyone help about this issue or this is not an issue but requesries different folder structure settings in OP?

thanks!

maxpareschi:
You must use a single root in the project structure, like “root[work]”
Otherwise it will do all roots

PR #3809, opened 2022-09-08
General: Better pixmap scaling

iLLiCiTiT:
## Brief description
Enabled high pixmap scaling in tray application.

## Description
Enabling these flags enhance pixmap look in UIs a lot.

### Screenshot
image

## Testing notes:
1. Compare look of UI images in this PR and in develop

### DIFF
--- a/openpype/tools/tray/pype_tray.py
+++ b/openpype/tools/tray/pype_tray.py
@@ -9,11 +9,11 @@ import platform
 from Qt import QtCore, QtGui, QtWidgets
 
 import openpype.version
-from openpype.api import (
-    resources,
-    get_system_settings
+from openpype import resources, style
+from openpype.lib import (
+    get_openpype_execute_args,
+    Logger,
 )
-from openpype.lib import get_openpype_execute_args, Logger
 from openpype.lib.openpype_version import (
     op_version_control_available,
     get_expected_version,
@@ -25,8 +25,8 @@ from openpype.lib.openpype_version import (
     get_openpype_version,
 )
 from openpype.modules import TrayModulesManager
-from openpype import style
 from openpype.settings import (
+    get_system_settings,
     SystemSettings,
     ProjectSettings,
     DefaultsNotDefined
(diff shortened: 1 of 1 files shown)


openpype/tools/tray/pype_tray.py
```diff
@@ -778,6 +778,14 @@ def main():
     if not app:
         app = QtWidgets.QApplication([])
 
+    for attr_name in (
+        "AA_EnableHighDpiScaling",
+        "AA_UseHighDpiPixmaps"
+    ):
+        attr = getattr(QtCore.Qt, attr_name, None)
+        if attr is not None:
+            app.setAttribute(attr)
```
BigRoy, line 787: If this is only for Pype Tray - don't we know what Qt version would be set up in the environment for that and whether it should be available?
Also, should this log a debug message/warning if an attribute does not exist so we can identify changes of it in the future?

iLLiCiTiT: > If this is only for Pype Tray - don't we know what Qt version would be set up in the environment for that and whether it should be available?
Also, should this log a debug message/warning if an attribute does not exist so we can identify changes of it in the future?

I've added it as "secure" check for MacOS or linux distributions where it may not be available, depending on installed Qt build. It's a "fancy" feature, definetely should not be breaking. Logs are added, but I can assure you that no one will notice a log message, but everyone will notice pixelated images :)

BigRoy: Good point - nice work!

openpype/tools/tray/pype_tray.py
```diff
@@ -9,11 +9,11 @@
 from Qt import QtCore, QtGui, QtWidgets
 
 import openpype.version
-from openpype.api import (
-    resources,
```
BigRoy, line 13: Nice catch!

BigRoy (approved):
Haven't tested the code - but code changes look good to me.

PR #3810, opened 2022-09-08
General: lock task workfiles when they are working on

moonyuet:
## Brief description
Adding lock file when user opens the files to work on while removing it when user finished the work and exit the maya scene. 
image
image
image


openpype/hosts/maya/api/pipeline.py
```diff
@@ -426,6 +468,49 @@ def on_before_save():
     return lib.validate_fps()
 
 
+def after_file_open():
```
BigRoy, line 471: Function name is unclear - best to rename to `check_lock_on_current_file`? Additionally it looks like this function isn't used on "after open" but "on save"?

openpype/pipeline/workfile/lock_workfile.py
```diff
@@ -0,0 +1,67 @@
...
+    if not process_id:
+        process_id = str(uuid4())
+        os.environ["OPENPYPE_PROCESS_ID"] = process_id
+    info = get_workstation_info()
+    info["process_id"] = process_id
+    with open(lock_filepath, "w") as stream:
+        json.dump(info, stream)
+
+
+def get_username(filepath):
```
iLLiCiTiT, line 55: One more request, change name of this function to `get_user_from_lock` please.

openpype/hosts/maya/api/pipeline.py
```diff
@@ -426,6 +468,49 @@ def on_before_save():
...
+        cmds.columnLayout(adjustableColumn=True)
+        comment = " %s is working the same workfile!" % username
+        cmds.text(comment, align='center')
+        cmds.text(vis=False)
+        cmds.rowColumnLayout(numberOfColumns=3,
+                             columnWidth=[(1, 300), (2, 100)],
+                             columnSpacing=[(2, 10)])
+        cmds.separator(vis=False)
+        quit_command = "cmds.quit(force=True);cmds.deleteUI('%s')" % reminder
+        cmds.button(label='Ok', command=quit_command)
```
BigRoy, line 495: Better maybe to not quit Maya but instead "close" the file.

Also feel like this UI logic is better implemented in Qt to stick with OpenPype styling and also potentially introduce this logic in another host. I feel like this most of this logic should actually be in `openpype.pipeline.lock_workfile` too.

openpype/hosts/maya/api/pipeline.py
```diff
@@ -229,6 +265,12 @@ def _before_scene_save(return_code, client_data):
     )
 
 
+def _remove_workfile_lock():
+    filepath = current_file()
+    if filepath:
+        remove_workfile_lock(filepath)
```
BigRoy, line 271:
```suggestion
def remove_current_workfile_lock():
    """Remove workfile lock on current file"""
    filepath = current_file()
    if filepath:
        remove_workfile_lock(filepath)
```
Does this make more sense?

And feels like this could actually be exposed in the `openpype.pipeline.lock_workfile` too so that potential other hosts who would like to implement this could do so - especially because it just uses the workio API `current_file` function.

openpype/pipeline/workfile/lock_workfile.py
```diff
@@ -0,0 +1,67 @@
...
+    lock_filepath = _get_lock_file(filepath)
+    process_id = os.environ["OPENPYPE_PROCESS_ID"]
+    data = _read_lock_file(lock_filepath)
+    return data["process_id"] == process_id
+
+
+def delete_workfile_lock(filepath):
+    lock_filepath = _get_lock_file(filepath)
+    if not os.path.exists(lock_filepath):
+        return
```
BigRoy, line 37: No need to check this since `is_workfile_locked_for_current_process` already checks that too. Right?

openpype/pipeline/workfile/lock_workfile.py
```diff
@@ -0,0 +1,67 @@
...
+    if not is_workfile_locked():
+        return False
+
+    lock_filepath = _get_lock_file(filepath)
+    process_id = os.environ["OPENPYPE_PROCESS_ID"]
+    data = _read_lock_file(lock_filepath)
+    return data["process_id"] == process_id
+
+
+def delete_workfile_lock(filepath):
```
BigRoy, line 34: Should this function allow a `force` argument to even force it if not a file of the current user? Or considered too dangerous?

What happens if you force quit (kill process) maya, wouldn't the `.lock` file remain?

openpype/pipeline/workfile/lock_workfile.py
```diff
@@ -0,0 +1,67 @@
...
+    info = get_workstation_info()
+    info["process_id"] = process_id
+    with open(lock_filepath, "w") as stream:
+        json.dump(info, stream)
+
+
+def get_username(filepath):
+    lock_filepath = _get_lock_file(filepath)
+    with open(lock_filepath, "r") as stream:
+        data = json.load(stream)
```
BigRoy, line 58:
```suggestion
    lock_filepath = _get_lock_file(filepath)
    data = _read_lock_file(lock_filepath)
```
Shouldn't it be this?

openpype/pipeline/workfile/lock_workfile.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def _read_lock_file(lock_filepath):
+    with open(lock_filepath, "r") as stream:
+        data = json.load(stream)
+    return data
+
+
+def _get_lock_file(filepath):
+    return filepath + ".lock"
```
BigRoy, line 14: What happens if potentially an application's workfile does have a `.lock` file of its own. Should we maybe add a safeguard in `_read_lock_file` with some error logging if the file wasn't readable or laid out as OP expected?

openpype/hosts/maya/api/pipeline.py
```diff
@@ -426,6 +468,49 @@ def on_before_save():
...
+    """Delete the lock file after user quitting the Maya Scene"""
+    log.info("Closing Maya...")
+    # delete the lock file
+    filepath = current_file()
+    remove_workfile_lock(filepath)
+
+
+def before_file_open():
+    """check lock file when the file changed"""
+    log.info("check lock file when file changed...")
```
BigRoy, line 509: I feel like this logging could be clearer.

```suggestion
    log.info("Removing lock on current file before scene open...")
```

Or more generic which could be in the lock workfile api instead:
```suggestion
    log.debug("Removing lock on current file...")
```

I actually wonder whether maybe this should be debug logging in the `openpype.pipeline.lock_workfile` logic instead?

openpype/hosts/maya/api/pipeline.py
```diff
@@ -426,6 +468,49 @@ def on_before_save():
...
+                             columnSpacing=[(2, 10)])
+        cmds.separator(vis=False)
+        quit_command = "cmds.quit(force=True);cmds.deleteUI('%s')" % reminder
+        cmds.button(label='Ok', command=quit_command)
+        cmds.showWindow(reminder)
+
+
+def on_before_close():
+    """Delete the lock file after user quitting the Maya Scene"""
+    log.info("Closing Maya...")
```
BigRoy, line 501: Same as logging comment below - this can be more descriptive. Best if this explains what it is about to do right?

Also, I really wonder how reliable the "maya exiting callback" will be for this feature. Maya crashes a lot. We need to make sure that won't get in the way for users if maya force quits, breaks, crashes and you try to re-open the same scene.

BigRoy (commented):
Some comments on the logic.

Also, wanted to add. Can we please add logic to enable/disable this in OpenPype settings. I'm a bit wary of the complications this will bring and we'll likely want to disable this feature. Maya crashes way too much for this to feel like something we can rely on.

openpype/pipeline/workfile/lock_workfile.py
moonyuet: I didn't use this function in other script. I guess it wont be problem if it gets removed.

openpype/hosts/maya/api/pipeline.py
moonyuet: It is implemented in "on_open", which binds to the kAfterOpen callback

openpype/pipeline/workfile/lock_workfile.py
iLLiCiTiT: This is to delete the lock (not matter who created it). This should give ability to "unlock" it e.g. if maya crashes.

openpype/pipeline/workfile/lock_workfile.py
iLLiCiTiT: It should also check if the file exists and return None if not?

openpype/pipeline/workfile/lock_workfile.py
iLLiCiTiT: Function `remove_workfile_lock` is to remove the lock for current process and `delete_workfile_lock` should be used to delete the lock file no matter how is opened.

moonyuet: I tested it with` cmds.quit(force =True)` in maya scene with Openpype, the scripts of removing/deleting lock still works even if  it is forced to be closed.

openpype/hosts/maya/api/pipeline.py
iLLiCiTiT: That was actually my point that callback to `kAfterOpen` should have `on_open` callback rather then `check_lock_on_current_file`.

openpype/hosts/maya/api/pipeline.py
iLLiCiTiT: > Better maybe to not quit Maya but instead "close" the file? (E.g. new scene?)

True

> Also feel like this UI logic is better implemented in Qt to stick with OpenPype styling and also potentially introduce this logic in another host. I feel like this most of this logic should actually be in openpype.pipeline.lock_workfile too.

I agree but I'm not sure if @moonyuet should do it, and it should be done in this PR. We do this because of single request and spending so much time because of single request without knowing if it actually be used is waste of work time for us. But the lock should be in settings and disabled by default.

openpype/pipeline/workfile/lock_workfile.py
iLLiCiTiT: Good point. My suggestion is `.oplock` but any other suggestions are welcomed :)

openpype/hosts/maya/api/pipeline.py
iLLiCiTiT: > especially because it just uses the workio API current_file function

I don't think this is a good idea. Functions in `lock_workfile` are so simple on purpose. Keep in mind that not each host handle workfiles the same way.

openpype/pipeline/workfile/lock_workfile.py
iLLiCiTiT: I had call with @moonyuet and we did changes that should add enable/disable with settings and ability of ignoring the lock if is existing (so you can take control even if lock is there).

Like I've wrote this is "feature propose" so there are some dirty hacks. We'll do modifications in this PR only if it cause slowdown, it will cause issues or the feature won't work.

m-u-r-p-h-y (commented):
It works, the lock file is created and warning is displayed.

I would suggest improving the UI of the warning message a little.

instead of
image

something like

image

moonyuet:
> It works, the lock file is created and warning is displayed.
> 
> I would suggest improving the UI of the warning message a little.
> 
> instead of image
> 
> something like
> 
> image

image
I have added the workfile dialog done by @iLLiCiTiT in maya and workfile tools

BigRoy:
Looking at the design of this dialog this could have also been a [`QMessageBox.warning`](https://doc.qt.io/qt-6/qmessagebox.html#warning)? The screenshot still doesn't clarify to me what "cancel" would do

Psuedocode example for potential QMessageBox:
```python
from Qt import QtWidgets
message = QtWidgets.QMessageBox
message.warning(
    parent, 
    title="File is already open by other user", 
    text="{user} on {machine} is working on the same workfile",
    buttons = message.Close | message.Ignore,
    defaultButton = message.Close
)
```

Thoughts?

---

Also, the pop-up shows BEFORE the file is opened, correct? So that 'cancel' avoids even trying to open the file to begin with. In that case "Cancel" makes sense.

openpype/hosts/maya/api/pipeline.py
```diff
@@ -426,6 +478,47 @@ def on_before_save():
...
+
+    # add lockfile dialog
+    from Qt import QtWidgets
+    from openpype.tools.workfiles.lock_dialog import WorkfileLockDialog
+
+    top_level_widgets = {w.objectName(): w for w in
+                         QtWidgets.QApplication.topLevelWidgets()}
+    parent = top_level_widgets.get("MayaWindow", None)
+    workfile_dialog = WorkfileLockDialog(filepath, parent=parent)
+    workfile_dialog.show()
```
BigRoy, line 503: Technically the lock could have been removed between "is_workfile_locked" and initializing the WorkfileLockDialog.
Would it make sense more to instead do it like this:

```python
lockfile_data = get_lockfile_data(filepath)
if not lockfile_data:
    # File is not locked, create lock, etc.
    return
 
dialog = WorkfileLockDialog(lock_data, parent=parent)
dialog.show()
```
Otherwise one could run into the race condition where the WorkfileLockDialog would be unable to get the data and then raise an error doing so?

openpype/tools/workfiles/lock_dialog.py
```diff
@@ -0,0 +1,55 @@
...
+
+        self.setStyleSheet(load_stylesheet())
+
+    def _on_ignore_click(self):
+        # Result is '1'
+        self.accept()
+
+    def _on_cancel_click(self):
+        # Result is '0'
+        self.reject()
```
BigRoy, line 55: When doing things like this where you are connecting signals to just call a single function, you can also just directly connect that function. Like so:

```suggestion
        cancel_btn.clicked.connect(self.reject)
        ignore_btn.clicked.connect(self.accept)

    def showEvent(self, event):
        super(WorkfileLockDialog, self).showEvent(event)
        self.setStyleSheet(load_stylesheet())
```

openpype/hosts/maya/api/pipeline.py
iLLiCiTiT, line 503: This import should be at the top:
```
from openpype.tools.workfiles.lock_dialog import WorkfileLockDialog
```

The dialog can be executed (wait for user response) and handle result.
```suggestion
    if is_workfile_locked(filepath):
        # Add ability to ignore lock
        workfile_dialog = WorkfileLockDialog(filepath, parent=parent)
        if not workfile_dialog.exec_():
            return
    create_workfile_lock(filepath)
```

openpype/tools/workfiles/files_widget.py
```diff
@@ -452,8 +459,19 @@ def _get_event_context_data(self):
...
+            return False
+        if not is_workfile_locked(filepath):
+            return False
+        return not is_workfile_locked_for_current_process(filepath)
+
     def open_file(self, filepath):
         host = self.host
+        if self._is_workfile_locked(filepath):
+            # add lockfile dialog
+            WorkfileLockDialog(filepath, parent=self)
```
iLLiCiTiT, line 473:
```suggestion
            dialog = WorkfileLockDialog(filepath, parent=self)
            # Skip if user cancel the action
            if not dialog.exec_():
                return
```

openpype/tools/workfiles/lock_dialog.py
iLLiCiTiT: This is actually my code. I preffer to have callback for a button click even if the callback just call other method. But in this case It's not a big deal so it can be commited.

m-u-r-p-h-y (commented):
I was testing the workflow and have a couple of notes:

- what will happen to the lock when I up-version the workfile? (it seems that lock file is removed from the old workfile, but new one is not created)

- when I was opening the workfile using `Work file ...`  tool in Maya the lock dialogue is displayed twice (answer the first and it is displayed again)

image

openpype/hosts/maya/api/pipeline.py
```diff
@@ -426,6 +479,46 @@ def on_before_save():
...
+    # add the lock file when opening the file
+    filepath = current_file()
+
+    if is_workfile_locked(filepath):
+        # add lockfile dialog
+        from Qt import QtWidgets
+        top_level_widgets = {w.objectName(): w for w in
+                             QtWidgets.QApplication.topLevelWidgets()}
+        parent = top_level_widgets.get("MayaWindow", None)
+        workfile_dialog = WorkfileLockDialog(filepath, parent=parent)
```
iLLiCiTiT, line 498: I think it will work even without the parent. If it does not then please use `get_main_window` from `.lib` to get it.
```suggestion
        workfile_dialog = WorkfileLockDialog(filepath)
```

With parent
```
        parent = get_main_window()
        workfile_dialog = WorkfileLockDialog(filepath, parent=parent)
```

iLLiCiTiT:
> Looking at the design of this dialog this could have also been a [QMessageBox.warning](https://doc.qt.io/qt-6/qmessagebox.html#warning)?
> The pop-up shows BEFORE the file is opened, correct? So that 'cancel' avoids even trying to open the file to begin with. In that case "Cancel" makes sense and then the pop-up still doesn't clarify to me what "cancel" would do to the scene I just opened. If it's before however, it's fine as is.

You would have to implement the same logic in each host if you would like to use it. If the content of the lock would change or we would like to change the message we would have to change it in each individual implementation. `WorkfileLockDialog` is to avoid it.

But we can change the labels.

BigRoy:
> You would have to implement the same logic in each host if you would like to use it. If the content of the lock would change or we would like to change the message we would have to change it in each individual implementation. WorkfileLockDialog is to avoid it.

Sure. I was just thinking of reducing the code complexity by adding the "warn_user" dialog in the workfile_lock library using `QMessageBox` functionality. So like now it could run something from there - the library would just be using `QMessageBox` to have less lines of code to maintain.

iLLiCiTiT:
> Sure. I was just thinking of reducing the code complexity by adding the "warn_user" dialog in the workfile_lock library using QMessageBox functionality.

We don't mix UI and logic. Qt (if possible) should not be imported anywhere in `openpype.pipeline`.

iLLiCiTiT:
Not sure if this is Maya 2020 specific but `current_file` returns `None` in callbacks 😢 
```
  File "C:\Users\JakubTrllo\Desktop\Prace\openpype3_3\openpype\lib\events.py", line 160, in process_event
    callback()
  File "C:\Users\JakubTrllo\Desktop\Prace\openpype3_3\openpype\hosts\maya\api\pipeline.py", line 590, in on_open
    check_lock_on_current_file()
  File "C:\Users\JakubTrllo\Desktop\Prace\openpype3_3\openpype\hosts\maya\api\pipeline.py", line 492, in check_lock_on_current_file
    if is_workfile_locked(filepath):
  File "C:\Users\JakubTrllo\Desktop\Prace\openpype3_3\openpype\pipeline\workfile\lock_workfile.py", line 23, in is_workfile_locked
    lock_filepath = _get_lock_file(filepath)
  File "C:\Users\JakubTrllo\Desktop\Prace\openpype3_3\openpype\pipeline\workfile\lock_workfile.py", line 19, in _get_lock_file
    return filepath + ".oplock"
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str' # 
```

openpype/hosts/maya/api/pipeline.py
```diff
@@ -229,6 +266,22 @@ def _before_scene_save(return_code, client_data):
     )
 
 
+def _remove_workfile_lock():
+    """Remove workfile lock on current file"""
+    if not handle_workfile_locks():
+        return
+    filepath = current_file()
```
iLLiCiTiT, line 273:
```suggestion
    filepath = current_file()
    log.info("Removing lock on current file {}...".format(filepath))
```

openpype/hosts/maya/api/pipeline.py
```diff
@@ -426,6 +479,46 @@ def on_before_save():
...
+    log.info("Closing Maya...")
+    # delete the lock file
+    filepath = current_file()
+    if handle_workfile_locks():
+        remove_workfile_lock(filepath)
+
+
+def before_file_open():
+    """check lock file when the file changed"""
+    log.info("Removing lock on current file before scene open...")
```
iLLiCiTiT, line 517:
```suggestion
```

moonyuet:
> I was testing the workflow and have a couple of notes:
> 
> * what will happen to the lock when I up-version the workfile? (it seems that lock file is removed from the old workfile, but new one is not created)
> * when I was opening the workfile using `Work file ...`  tool in Maya the lock dialogue is displayed twice (answer the first and it is displayed again)
> 
> image

the first issue can be resolved by creating lockfile function in aftersave callback for both workfile and maya.(the callbacks have been added in the latest commit)
The reason to cause the second issue contributes to having the lock_dialog in both workfile and maya callback so that's why you see the dialog pop up twice. In the latest commit, I didn't execute the dialog in the workfile at all , but I am not sure it is a good case as it is supposed that the lock function is also implemented in other dcc software.

m-u-r-p-h-y (changes requested):
it is not working now on file open (from Launcher nor Work File tool in Maya)

the lock file is only created when up versioning in Work File tool

m-u-r-p-h-y:
> it is not working now on file open (from Launcher nor Work File tool in Maya)
> 
> the lock file is only created when up versioning in Work File tool



I was wrong. It works, I just did not have the correct settings. With correct setting set
image
 the lock file is created correctly and changed accordingly when I switch versions or upversion


It means there is still a problem, that the lock files is created with settings OFF and doing upversion in Work File tool . . . (will double check)

m-u-r-p-h-y (approved):
I tested different scenarios with settings on/off

upversioning and reopening using Work File tool and it works as expected

I'm pretty sure it will open a flood of ideas on how to improve it, but for the first implementation is quite solid.

good job!

Issue #3812, opened 2022-09-08
Maya: "Look" instance inside "Render" Instance causes Collecting errors during Pyblish of the Look/Render


LiborBatek:
I wanted publish look instance and also render turntable of the look via "Render" instance. So I did following...

Creating Look instance and Render instance using Creator in maya scene and then afterwards putting "Look" instance inside "Render" instance in Outliner (see Hierarchy below)  

-RenderingMain
              -LookMain

...caused following collecting errors:

![image](https://user-images.githubusercontent.com/112623825/189121719-b1bb015d-1fa9-4ed8-843c-466aade846e0.png)

```
Message: 
'NoneType' object has no attribute 'group'
Line: 
115
Traceback: 
Traceback (most recent call last):
  File "D:\REPO\OpenPype\.venv\lib\site-packages\pyblish\plugin.py", line 522, in __explicit_process
    runner(*args)
  File "D:\REPO\OpenPype\openpype\hosts\maya\plugins\publish\collect_render.py", line 115, in process
AttributeError: 'NoneType' object has no attribute 'group'

```

[cuID:OP-3941]

BigRoy:
Good catch!

To fix your particular case - that `lookMain` should go into the actual layer you'd like to attach it to. So for example into that `Main` set inside.

Nonetheless - this shows that the logic and error logging can be improved for the end user.

Likely it's best to log a clearer message `self.log.warning("Ignoring invalid member of 'renderingMain': {}".format(layer))` or alike.
And potentially highlight to the user that it should maybe be moved into the layer-specific sets inside of it.

BigRoy:
@LiborBatek was this issue resolved with #3816 and did that improve the logging message enough to be clear for the user?

LiborBatek:
I did try both cases putting ```Look``` instance into ```RenderMain``` and next to it and gives no warning at all :) just publishes ok in both cases.

But I think it was resolved to work in both cases.

PR #3816, opened 2022-09-08
Maya: Warn correctly about nodes in render instance with unexpected names

BigRoy:
## Brief description

This is a fix for #3812 

## Description

This fixes the reported error and the node added incorrectly into the set is now logged with a warning. However, what the user originally tried in #3812 might still be unclear on how to solve correctly.

They should have done this: 🟢 
openpype_valid_attach_to_rendering

Instead of this: 🔴 
openpype_invalid_attach_to_rendering

Note how the `lookMain` that was intended to be attached to the layer is not a member of `_renderingMain:Main` (which represents the Main layer) and instead `lookMain` is a member of the `renderingMain` instance directly.

We might want improve the logging or workflow to ensure the User is more aware of what they did wrong, because this is what is logged as warning now:

afbeelding

In the current stage it DOES allow the user to continue and publish - but it won't attach the render to `lookMain`.

## Testing notes:
1. Add `lookMain` into `renderingMain`
2. Should now log warning instead of error with even more confusing error to end user - now at least the name is logged, etc.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/collect_render.py
+++ b/openpype/hosts/maya/plugins/publish/collect_render.py
@@ -102,21 +102,23 @@ class CollectMayaRender(pyblish.api.ContextPlugin):
         }
 
         for layer in collected_render_layers:
-            try:
-                if layer.startswith("LAYER_"):
-                    # this is support for legacy mode where render layers
-                    # started with `LAYER_` prefix.
-                    expected_layer_name = re.search(
-                        r"^LAYER_(.*)", layer).group(1)
-                else:
-                    # new way is to prefix render layer name with instance
-                    # namespace.
-                    expected_layer_name = re.search(
-                        r"^.+:(.*)", layer).group(1)
-            except IndexError:
+            if layer.startswith("LAYER_"):
+                # this is support for legacy mode where render layers
+                # started with `LAYER_` prefix.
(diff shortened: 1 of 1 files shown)


BigRoy:
I have also made the `Processing "{layer}` log a bit more detailed so it's clearer what is happening if it fails after, e.g. like this:

afbeelding

(I made it fail on purpose by adding in a set that was also in that namespace, but did not relate to a layer)

---

Side-note: The same potential bug/error as #3812 describes I believe is also in the [V-Ray Scene Collector's logic since it follows the same logic as this originally did](https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/maya/plugins/publish/collect_vrayscene.py#L48-L54).

LiborBatek (approved):
When doing this:
image

Yep, its working as it should. Gives warning yet publish without errors. 
image

So approved!

m-u-r-p-h-y:
just wondering, what will be the logic if there is more than 1 render layer and the look instance will be placed outside of them?

BigRoy:
> just wondering, what will be the logic if there is more than 1 render layer and the look instance will be placed outside of them?

It'd still ignore the look instance. It would only attached if added into the ones per layer I think.

mkolar (approved):

PR #3820, opened 2022-09-09
Maya+Ftrack: Change typo in family name `mayaascii` -> `mayaAscii`

BigRoy:
## Brief description

Fix typo in legacy family name `mayaascii` -> `mayaAscii`.

## Description

Not sure how relevant this tweak is because I think this is the legacy family name for `mayaScene` family. Correct?

[Mentioned on discord](https://discord.com/channels/517362899170230292/517382145552154634/1017751095734452264)

## Testing notes:

1. Not sure where to test this best. Publish a Maya scene and check with Ftrack?

### DIFF
--- a/openpype/modules/ftrack/plugins/publish/integrate_ftrack_instances.py
+++ b/openpype/modules/ftrack/plugins/publish/integrate_ftrack_instances.py
@@ -74,11 +74,14 @@ class IntegrateFtrackInstance(pyblish.api.InstancePlugin):
         version_number = int(instance_version)
 
         family = instance.data["family"]
-        family_low = family.lower()
 
+        # Perform case-insensitive family mapping
+        family_low = family.lower()
         asset_type = instance.data.get("ftrackFamily")
-        if not asset_type and family_low in self.family_mapping:
-            asset_type = self.family_mapping[family_low]
+        if not asset_type:
+            for map_family, map_value in self.family_mapping.items():
+                if map_family.lower() == family_low:
+                    asset_type = map_value
 
         if not asset_type:
             asset_type = "upload"
--- a/openpype/modules/ftrack/plugins/publish/integrate_ftrack_instances.py
+++ b/openpype/modules/ftrack/plugins/publish/integrate_ftrack_instances.py
@@ -89,15 +89,6 @@ class IntegrateFtrackInstance(pyblish.api.InstancePlugin):
         self.log.debug(
             "Family: {}\nMapping: {}".format(family_low, self.family_mapping)
         )
-
-        # Ignore this instance if neither "ftrackFamily" or a family mapping is
-        # found.
-        if not asset_type:
-            self.log.info((
-                "Family \"{}\" does not match any asset type mapping"
-            ).format(family))
-            return
-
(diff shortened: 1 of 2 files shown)


iLLiCiTiT:
This would actually break it. Current logic expect that the family is lowered in settings. I suggest these changes to replace lines 79-81 in `https://integrate_ftrack_instances.py/`.

```
# I would maybe replace this with 'instance.data.get("ftrackAssetType")'
# - it's not used (as far as I know) and 'ftrackAssetType' makes more sense
asset_type = instance.data.get("ftrackFamily")
if not asset_type:
    for key, value in self.family_mapping.items():
        if key.lower() == family_lower:
            asset_type = value
            break
```

BigRoy:
Thanks!

That means that instead currently [`mayaScene`](https://github.com/pypeclub/OpenPype/blob/d0665c3928ecff4176f6e22b676384f09485c173/openpype/settings/defaults/project_settings/ftrack.json#L476) and [`unrealStaticMesh`](https://github.com/pypeclub/OpenPype/blob/d0665c3928ecff4176f6e22b676384f09485c173/openpype/settings/defaults/project_settings/ftrack.json#L484) are broken in settings?

iLLiCiTiT:
> That means that instead currently [mayaScene](https://github.com/pypeclub/OpenPype/blob/d0665c3928ecff4176f6e22b676384f09485c173/openpype/settings/defaults/project_settings/ftrack.json#L476) and [unrealStaticMesh](https://github.com/pypeclub/OpenPype/blob/d0665c3928ecff4176f6e22b676384f09485c173/openpype/settings/defaults/project_settings/ftrack.json#L484) are broken in settings?

Yeah. Looks like it is...

BigRoy:
@iLLiCiTiT Still believe [this](https://github.com/pypeclub/OpenPype/pull/3820#issuecomment-1243392485) is the recommended fix?

iLLiCiTiT:
> @iLLiCiTiT Still believe https://github.com/pypeclub/OpenPype/pull/3820#issuecomment-1243392485 is the recommended fix?

Yes I think so. If you don't want to be blamed I can create PR to your branch :D

BigRoy:
No worries - I break things all the time.

openpype/modules/ftrack/plugins/publish/integrate_ftrack_instances.py
```diff
@@ -74,27 +74,22 @@ def process(self, instance):
...
             "Family: {}\nMapping: {}".format(family_low, self.family_mapping)
         )
-
-        # Ignore this instance if neither "ftrackFamily" or a family mapping is
-        # found.
-        if not asset_type:
-            self.log.info((
-                "Family \"{}\" does not match any asset type mapping"
-            ).format(family))
-            return
```
BigRoy, line 96: Removed this logic since it would never happen since above it's forced as `asset_type = "upload"` if it wasn't set.

BigRoy:
Ok - should be refactored to match key without case-sensitivity correctly.

iLLiCiTiT (approved):

PR #3821, opened 2022-09-09
Maya: Render settings validation attribute check tweak logging

BigRoy:
## Brief description

If settings was previously set up with an attribute name that was NOT `{node_type}.{attr}` like e.g. `foobar` without a dot the logic would've made it into `foobar` as node_name and would have tried to getAttr `node.` which seems odd. Now the logic will warn the user about the invalid settings.

I also believe the previous `isinstance` list check would have never proceeded to detect no nodes found since `cmds.ls()` always returns a list, even when no matching nodes detected.

It's mostly logging related cosmetics.

## Testing notes:

1. Submit renders for different renderers with different attribute value check implemented in settings.
2. Test with invalid attr names - you should be warned.. but publish should not error hard.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/validate_rendersettings.py
+++ b/openpype/hosts/maya/plugins/publish/validate_rendersettings.py
@@ -269,7 +269,8 @@ class ValidateRenderSettings(pyblish.api.InstancePlugin):
             nodes = cmds.ls(type=node_type)
 
             if not nodes:
-                cls.log.info("No nodes of type '{}' found.".format(node_type))
+                cls.log.warning(
+                    "No nodes of type '{}' found.".format(node_type))
                 continue
 
             for node in nodes:
--- a/openpype/hosts/maya/plugins/publish/validate_rendersettings.py
+++ b/openpype/hosts/maya/plugins/publish/validate_rendersettings.py
@@ -257,14 +257,18 @@ class ValidateRenderSettings(pyblish.api.InstancePlugin):
         # go through definitions and test if such node.attribute exists.
         # if so, compare its value from the one required.
         for attr, value in OrderedDict(validation_settings).items():
-            # first get node of that type
             cls.log.debug("{}: {}".format(attr, value))
-            node_type = attr.split(".")[0]
-            attribute_name = ".".join(attr.split(".")[1:])
+            if "." not in attr:
+                cls.log.warning("Skipping invalid attribute defined in "
+                                "validation settings: '{}'".format(attr))
+
+            node_type, attribute_name = attr.split(".", 1)
+
+            # first get node of that type
             nodes = cmds.ls(type=node_type)
 
-            if not isinstance(nodes, list):
-                cls.log.warning("No nodes of '{}' found.".format(node_type))
+            if not nodes:
+                cls.log.info("No nodes of type '{}' found.".format(node_type))
                 continue
 
             for node in nodes:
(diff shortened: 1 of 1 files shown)


m-u-r-p-h-y:
We are not sure we understand exactly the testing procedure here. But I encountered an error while testing render settings

image

```
[['enableProgressiveRender', '1'], ['motion_blur_enable', '0']]
# Error: No object matches name: enableProgressiveRender
#   File "D:\REPO\OpenPype\openpype\hosts\maya\api\menu.py", line 131, in <lambda>
#     command=lambda *args: lib_rendersettings.RenderSettings().set_default_renderer_settings()    # noqa
#   File "D:\REPO\OpenPype\openpype\hosts\maya\api\lib_rendersettings.py", line 88, in set_default_renderer_settings
#     self._set_arnold_settings(width, height)
#   File "D:\REPO\OpenPype\openpype\hosts\maya\api\lib_rendersettings.py", line 141, in _set_arnold_settings
#     self._additional_attribs_setter(additional_options)
#   File "D:\REPO\OpenPype\openpype\hosts\maya\api\lib_rendersettings.py", line 236, in _additional_attribs_setter
#     if (cmds.getAttr(str(attribute), type=True)) == "long":
# ValueError: No object matches name: enableProgressiveRender # 
```

(as a side note, could it be possible to make the input key text box wider as the attributes name can get quite long and it makes it unreadable in the settings. or should I create a separate issue for it?)

BigRoy:
Thank you! I'm not in the office Today so sorry for the short reply but this looks like I didn't fix it correctly yet.

BigRoy:
Ha, that's funny @m-u-r-p-h-y - I have taken another look.

I think that's actually an other area of the code base, that's the Render Settings settings. This PR relates to the Validate Render Settings logic.

This PR is about:
```
project_settings/maya/publish/ValidateRenderSettings/arnold_render_attributes
project_settings/maya/publish/ValidateRenderSettings/vray_render_attributes
project_settings/maya/publish/ValidateRenderSettings/redshift_render_attributes
project_settings/maya/publish/ValidateRenderSettings/renderman_render_attributes
```
_These take a `{node_type}.{attribute_name}` and a value_


You are referring to an issue with:
```
project_settings/maya/RenderSettings/arnold_renderer/additional_options
project_settings/maya/RenderSettings/vray_renderer/additional_options
project_settings/maya/RenderSettings/redshift_renderer/additional_options
```
_These take a `{node_name}.{attribute_name}` and a value_

So that label in the documentation for the additional options in Render Settings I believe is wrong. 

afbeelding

It shouldn't be just the attribute's name but it should include the node name as well.
So instead of:
> Add additional options - put attribute and value, like AASamples

It would be:
> Add additional options - put attribute and value, like defaultArnoldSettings.AASamples

I also feel that the render settings logic should log a warning if the `{node_name}.{attribute_name}` isn't found instead of an actual error stopping the setting of the render settings.

I'd be happy to rework that logic too - but I'm not sure if it makes sense to add into this PR.

LiborBatek:
I have tested several attributes like
defaultArnoldRenderOptions.AASamples and it works (OP>Set Render Settings - sets it accordingly) also I can confirm that this
formulation should be used as hint:

"Add additional options - put attribute name and value, like defaultArnoldRenderOptions.AASamples 4" - I can see putting even the example value as very useful to user to understand it

LiborBatek (approved):
Working ok!
Small note: the hint message in Admin/Project section should be adjusted-see my comment.

BigRoy:
> Small note: the hint message in Admin/Project section should be adjusted-see my comment.

Should be fixed now with https://github.com/pypeclub/OpenPype/pull/3821/commits/484e3175c9b0ebb694e565a628e0c991ace0495b

afbeelding

LiborBatek:
I can confirm that the "hint msg" in OP Project settings is updated and fine...

mkolar (approved):

PR #3825, opened 2022-09-09
Flame: OpenPype submenu to batch and media manager

jakubjezek001:
## Brief description
Invoking OpenPype submenu from more UI contexts

## Description
We can invoke OpenPype submenu with other actions not only from timeline editor but also from Batch and Media manager panels.

## Testing notes:
1. start Flame from Openpype
2. Go to Batch tab and RMB click to node graph > you should see OpenPype menu at bottom
3. Go to Media manager panel and RMB click any item in panel > you should see OpenPype menu at bottom


iLLiCiTiT (approved):
Approved

PR #3828, opened 2022-09-09
General: Fix Pattern access in client code

iLLiCiTiT:
## Brief description
Client code does not crash on python distributions where is not available `re.Pattern`.

## Description
Replace `re.Pattern` with custom variable `PatternType` by getting type of compiled string.


antirotor (approved):
looks fine - as fine as there hacks can look.

PR #3833, opened 2022-09-13
Publisher: Add new publisher to host tools

iLLiCiTiT:
## Brief description
Added new publisher to host tools. Code copied from https://github.com/pypeclub/OpenPype/pull/3046 which is in draft and the code is already needed for other PRs.

## Description
Added new publisher to host tools to be able easily open the tool from hosts.

## Testing notes:
1. Calling `show_publisher` in host tools should work

```
from openpype.tools.utils.host_tools import show_publisher

show_publisher()
```

### DIFF
--- a/openpype/tools/utils/host_tools.py
+++ b/openpype/tools/utils/host_tools.py
@@ -32,6 +32,7 @@ class HostToolsHelper:
         self._workfiles_tool = None
         self._loader_tool = None
         self._creator_tool = None
+        self._publisher_tool = None
         self._subset_manager_tool = None
         self._scene_inventory_tool = None
         self._library_loader_tool = None
@@ -205,6 +206,7 @@ class HostToolsHelper:
         pyblish_show = self._discover_pyblish_gui()
         return pyblish_show(parent)
 
+
     def _discover_pyblish_gui(self):
         """Return the most desirable of the currently registered GUIs"""
         # Prefer last registered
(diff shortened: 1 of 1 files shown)


openpype/tools/utils/host_tools.py
```diff
@@ -269,6 +271,30 @@ def show_experimental_tools_dialog(self, parent=None):
             dialog.activateWindow()
             dialog.showNormal()
 
+    def get_publisher_tool(self, parent):
+        """Create, cache and return scene inventory tool window."""
+        if self._scene_inventory_tool is None:
```
BigRoy, line 276:
```suggestion
        if self._publisher_tool is None:
```

iLLiCiTiT: You're faster then Sonic

antirotor (approved):
It worked in the original PR so I think it should work here too :)

openpype/tools/utils/host_tools.py
```diff
@@ -205,6 +206,7 @@ def show_publish(self, parent=None):
         pyblish_show = self._discover_pyblish_gui()
         return pyblish_show(parent)
 
+
```
BigRoy, line 209: Shouldn't the hound complain about these?
```suggestion

```

openpype/tools/utils/host_tools.py
```diff
@@ -269,6 +271,31 @@ def show_experimental_tools_dialog(self, parent=None):
             dialog.activateWindow()
             dialog.showNormal()
 
+    def get_publisher_tool(self, parent):
+        """Create, cache and return scene inventory tool window."""
```
BigRoy, line 275:
```suggestion
        """Create, cache and return publisher window."""
```

PR #3834, opened 2022-09-13
Ftrack: Url validation does not require ftrackapp

iLLiCiTiT:
## Brief description
Self deployed ftrack server does not have to end with `.ftrackapp.com` which is now possible to handle in OpenPype.

## Description
The validation does 2 validations of server connection, one with `ftrackapp.com` and one without. This way it is backwards compatible but works when url does not end with `.ftrackapp.com`. Function `check_ftrack_url` was moved to module level.

## Additional info
To get ftrack url it is required to connect to ftrack server which can slow down tray launch. That is because environment variable `FTRACK_SERVER` is set on launch. We should maybe not set the environment variable on start but get it from module when needed - has high potential to break ftrack code (created [issue](https://github.com/pypeclub/OpenPype/issues/3835)).

## Testing notes:
1. Set ftrack url to self deployed server without `ftrackapp.com` at the end
2. Start OpenPype
3. Ftrack login should work
4. Publishing to ftrack should work

### DIFF
--- a/openpype/modules/ftrack/ftrack_module.py
+++ b/openpype/modules/ftrack/ftrack_module.py
@@ -6,7 +6,7 @@ import platform
 import click
 
 from openpype.modules import OpenPypeModule
-from openpype_interfaces import (
+from openpype.modules.interfaces import (
     ITrayModule,
     IPluginPaths,
     ISettingsChangeListener
--- a/openpype/modules/ftrack/ftrack_module.py
+++ b/openpype/modules/ftrack/ftrack_module.py
@@ -520,8 +520,9 @@ def check_ftrack_url(url, log_errors=True, logger=None):
 
     if ftrack_url:
         logger.debug("Ftrack server \"{}\" is accessible.".format(ftrack_url))
+
     elif log_errors:
-        logger.error("Entered Ftrack URL \"{}\" is not accessible!".format(url))
+        logger.error("Ftrack server \"{}\" is not accessible!".format(url))
 
     return ftrack_url
 
--- a/openpype/modules/ftrack/ftrack_module.py
+++ b/openpype/modules/ftrack/ftrack_module.py
@@ -12,8 +12,10 @@ from openpype_interfaces import (
     ISettingsChangeListener
 )
 from openpype.settings import SaveWarningExc
+from openpype.lib import Logger
 
 FTRACK_MODULE_DIR = os.path.dirname(os.path.abspath(__file__))
+_PLACEHOLDER = object()
 
 
 class FtrackModule(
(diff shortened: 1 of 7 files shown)


openpype/modules/ftrack/ftrack_module.py
```diff
@@ -479,6 +482,50 @@ def cli(self, click_group):
...
+        result = requests.get(url, allow_redirects=False)
+    except requests.exceptions.RequestException:
+        return False
+
+    if (result.status_code != 200 or "FTRACK_VERSION" not in result.headers):
+        return False
+    return True
+
+
+def check_ftrack_url(url, log_errors=True, logger=None):
```
BigRoy, line 498: I think `resolve_ftrack_url` is a better name maybe?

openpype/modules/ftrack/ftrack_module.py
```diff
@@ -479,6 +482,50 @@ def cli(self, click_group):
...
+        return None
+
+    if not url.startswith("http"):
+        url = "https://" + url
+
+    ftrack_url = None
+    if not url.endswith("ftrackapp.com"):
+        ftrackapp_url = url + ".ftrackapp.com"
+        if _check_ftrack_url(ftrackapp_url):
+            ftrack_url = ftrackapp_url
```
BigRoy, line 516: I wonder why we shouldn't just enforce the user to put in full the url instead. So if they use a `ftrackapp.com` deployment why not have that included in the URL? Seems a bit magical almost?

I know it added it before already so it would somewhat break backwards compatibility? But maybe we should at least deprecate the automatic resolving?

openpype/modules/ftrack/lib/avalon_sync.py
```diff
@@ -19,11 +19,8 @@
     CURRENT_PROJECT_SCHEMA,
     CURRENT_PROJECT_CONFIG_SCHEMA,
 )
-from openpype.api import (
-    Logger,
-    get_anatomy_settings
-)
-from openpype.lib import ApplicationManager
+from openpype.settings import get_anatomy_settings
+from openpype.lib import ApplicationManager, Logger
```
BigRoy, line 23: Is this an intended refactor? (Just because it seems out of place in this PR and might be an oversight)

openpype/modules/ftrack/ftrack_module.py
```diff
@@ -479,6 +482,50 @@ def cli(self, click_group):
...
+
+    if not ftrack_url and _check_ftrack_url(url):
+        ftrack_url = url
+
+    if ftrack_url:
+        logger.debug("Ftrack server \"{}\" is accessible.".format(ftrack_url))
+    elif log_errors:
+        logger.error("Entered Ftrack URL \"{}\" is not accesible!".format(url))
```
BigRoy, line 524:
```suggestion
        logger.error("Entered Ftrack URL \"{}\" is not accessible!".format(url))
```

Or even better and align more with the other message?
```suggestion
        logger.error("Ftrack server \"{}\" is not accessible!".format(url))
```

openpype/modules/ftrack/ftrack_module.py
```diff
@@ -6,14 +6,16 @@
...
+from openpype.modules.interfaces import (
     ITrayModule,
     IPluginPaths,
     ISettingsChangeListener
 )
 from openpype.settings import SaveWarningExc
+from openpype.lib import Logger
 
 FTRACK_MODULE_DIR = os.path.dirname(os.path.abspath(__file__))
+_PLACEHOLDER = object()
```
BigRoy, line 18: Maybe more descriptive as `_URL_NOT_SET = object()`?

openpype/modules/ftrack/ftrack_module.py
BigRoy: Also `log_errors` argument is unused?

64qam (approved):

openpype/modules/ftrack/lib/avalon_sync.py
iLLiCiTiT: Not related to PR but intended.

openpype/modules/ftrack/ftrack_module.py
iLLiCiTiT: Not an issue or task for this PR.

openpype/modules/ftrack/ftrack_module.py
iLLiCiTiT: Done

PR #3839, opened 2022-09-13
Houdini: Implement review family with opengl node

### LINKED ISSUE #2720: Houdini: add support for reviews
## Feature

Add support for reviews in Houdini. More description in PR implementing it - #3839

[cuID:OP-2635]

BigRoy:
## Brief description

Implements a first pass for Reviews publishing in Houdini. Resolves #2720 

## Description

Uses the `opengl` ROP node to produce PNG images.

## Additional info

Required:

- [x] Working with Extract Review
- [x] Working with Extract burnins
- [x] Might need refactoring after #3046 
- [x] Support (animated) focal length burnin like maya does since #4531 
- [x] Expose some crucial default settings?
- [x] Maya has a "delete images" option. 
- [x] ~Might need a validator for the file extension - it's assuming it remains set to `.png` currently.~ Different file formats can now be set - it is not validated however.
- [x] Selected nodes should be filtered to 'cameras' and '/obj' level paths to go into `camera` and `scenepath` parm.
- [x] Add documentation! :)

Skipped for this PR (should be separate issues/PRs):

- [ ] Allow resolution of 0, 0 to represent "take resolution from asset data"?
- [ ] Add **Attach To** functionality (also [see discussion](https://github.com/pypeclub/OpenPype/discussions/3859))
- [ ] Default settings in OpenPype settings for specific Display Options on the ROP node, etc. (Probably not needed for minimum viable product of the feature)
- [ ] The Houdini logic for "Collect Frames" currently requires a period (`.`) before the frame numbers - something to keep an eye out for. (Probably best as separate issue/PR)

## Testing notes:

1. Launch Houdini
2. Create review instance
3. Set the "object path" and "camera" to something correct (needs to be done manually for now)


antirotor:
> Might need refactoring after https://github.com/pypeclub/OpenPype/pull/3046

As I am looking at the code, only minimal.

What are your thoughts about dealing with the resolution? In Maya, it is rather complex thing - getting resolution from asset but allowing you to override it through OP Settings and even Maya viewport settings.

BigRoy:
> What are your thoughts about dealing with the resolution? In Maya, it is rather complex thing - getting resolution from asset but allowing you to override it through OP Settings and even Maya viewport settings.

@antirotor With [this commit](https://github.com/pypeclub/OpenPype/pull/3839/commits/79812118215b11280dac110f5ab2736591e6c820) I've made it so that the resolution by default is set on the OpenGL rop and does not just inherit it from the camera it's rendering so it's more explicit.

Other than that I was thinking of potentially having it similar to maya. The integer fields that allow to override resolution on the OpenGL node actually allow negative values (go figure!). So I was thinking to do similar to maya that if -1 was found as value that it'd fall back to use asset resolution. So basically match the Maya behavior.

Would you prefer different?

BigRoy:
About a feature like "attachTo" in maya. What would integrators and publish plug-ins need for me to be able to do the same in Houdini?

antirotor:
> About a feature like "attachTo" in maya. What would integrators and publish plug-ins need with regards to `instance.data` (or `context.data`?) for me to be able to do the same in Houdini?

With the old publisher it takes just a subset name since there is no way to attach reviews to already published subsets. With the new one, we could maybe do something similar, but there are now ids there that are more robust and in the future we could extend it - find if the connected node to review node is publish instance or loaded instance and either add review to existing version or add to the new one.

Another question is how to deal with 'isolate' feature to create review just for the connected data or for the whole scene. But that's maybe something for the future PRs.

BigRoy:
> With the old publisher it takes just a subset name since there is no way to attach reviews to already published subsets. With the new one, we could maybe do something similar, but there are now ids there that are more robust and in the future we could extend it - find if the connected node to review node is publish instance or loaded instance and either add review to existing version or add to the new one.


For now:
`instance.data["attachTo"] = "other_subset_name"`
Check, thanks!

Not sure how attach to technically works down the line - but somehow I'd have expected to also be able to attach a review to multiple subsets. But probably since I haven't used the feature yet I might just be overthinking things.

> Another question is how to deal with 'isolate' feature to create review just for the connected data or for the whole scene. But that's maybe something for the future PRs.

OpenGL can specify what "objects" to render - so isolating certain parts of your scene is totally doable. You can also explicitly exclude certain objects, etc. Would leave that up to the artist to set up for now if they'd like to tweak it. That would be ok?

BigRoy:
I've made this work as barebones as it was but now with the new publisher.

I tried implementing this on the creator:
```python
    def get_instance_attr_defs(self):

        return [
            BoolDef("override_resolution",
                    label="Override resolution",
                    tooltip="When disabled the resolution set on the camera "
                            "is used instead.",
                    default=True),
            NumberDef("resx",
                      label="Resolution Width",
                      default=1280,
                      minimum=2,
                      decimals=0),
            NumberDef("resy",
                      label="Resolution Height",
                      default=720,
                      minimum=2,
                      decimals=0),
            NumberDef("aspect",
                      label="Aspect Ratio",
                      default=1.0,
                      minimum=0.0001,
                      decimals=3)
        ]
```

But as I've mentioned on discord a few times it just doesn't seem like the Houdini integration works in that regard. I started getting errors on refresh/reset and gave up again.

Additionally I would've liked it the HoudiniCreator plug-in allowed me to "intercept" certain attribute definitions on how they'd be read and imprinted into the scene easily. We might want to add a dedicated method to allow interception certain variables too. So that e.g. these attributes shown here could be actually stored and set using the dedicated attributes that the Houdini node itself already has for it. That would make much more sense.

@antirotor Any idea how we could go about making this production-ready and workable?

BigRoy:
@fabiaserra You mentioned you were interested in generating "reviews" from Houdini. This was a prototype that I believe worked. I've pulled in latest develop branch so I'd say give it a test run for some feedback.

@antirotor I'm not sure how valid [these notes addressed to you still are](https://github.com/ynput/OpenPype/pull/3839#issuecomment-1405157718) but if there's also someone on OP team able to give this a test run that'd be welcome!

fabiaserra:
> @fabiaserra You mentioned you were interested in generating "reviews" from Houdini. This was a prototype that I believe worked. I've pulled in latest develop branch so I'd say give it a test run for some feedback.
> 
> @antirotor I'm not sure how valid [these notes addressed to you still are](https://github.com/ynput/OpenPype/pull/3839#issuecomment-1405157718) but if there's also someone on OP team able to give this a test run that'd be welcome!

Amazing, thank you @BigRoy !

openpype/hosts/houdini/plugins/create/create_review.py
```diff
@@ -0,0 +1,74 @@
...
+    # Default settings for the ROP
+    # todo: expose in OpenPype settings?
+    override_resolution = True
+    width = 1280
+    height = 720
+    aspect = 1.0
+
+    def create(self, subset_name, instance_data, pre_create_data):
+
+        import hou
```
fabiaserra, line 25: why not a global import?

openpype/hosts/houdini/plugins/publish/extract_opengl.py
```diff
@@ -0,0 +1,55 @@
...
+    hosts = ["houdini"]
+    optional = True
+
+    def process(self, instance):
+
+        ropnode = hou.node(instance.data["instance_node"])
+
+        # Get the filename from the filename parameter
+        # `.evalParm(parameter)` will make sure all tokens are resolved
+        output = ropnode.evalParm("picture")
```
fabiaserra, line 24: I guess it's quite unlikely this could change to another rop but would it be better to use the util function?
```
output = get_output_parameter(ropnode).eval()`
```

openpype/hosts/houdini/plugins/publish/extract_opengl.py
```diff
@@ -0,0 +1,55 @@
...
+
+        # We run the render
+        self.log.info("Extracting '%s' to '%s'" % (file_name, staging_dir))
+        render_rop(ropnode)
+
+        # Unfortunately user interrupting the extraction does not raise an
+        # error and thus still continues to the integrator. To capture that
+        # we make sure all files exist
+        files = instance.data["frames"]
+        missing = [fname for fname in files
```
fabiaserra, line 37: maybe raise the error as soon as you find if a file doesn't exist?

fabiaserra (commented):
I will give it a test soon but first pass looks good to me!

openpype/hosts/houdini/api/lib.py
```diff
@@ -127,6 +127,8 @@ def get_output_parameter(node):
         return node.parm("filename")
     elif node_type == "comp":
         return node.parm("copoutput")
+    elif node_type == "opengl":
```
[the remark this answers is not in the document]
BigRoy: The Arnold if branch has another nested if branch in it because it can either render or export ass files and then uses a different filename parameter. So this might not be as trivial for some to be a dict mapping.

openpype/hosts/houdini/plugins/create/create_review.py
BigRoy: This actually happens across the code base that some imports happen in the pyblish plugins. I am not entirely sure why we would for `hou` so yes we can make this global.

I do wonder also why it's done in most others too. I vaguely recall that in very early versions of Pyblish there was a bug that some of the imported libs would already get garbage collected in the plugin files before process ran or something, but that was a long time ago that I had seen anything like that.

openpype/hosts/houdini/plugins/create/create_review.py
```diff
@@ -0,0 +1,74 @@
...
+            subset_name,
+            instance_data,
+            pre_create_data)  # type: CreatedInstance
+
+        instance_node = hou.node(instance.get("instance_node"))
+
+        frame_range = hou.playbar.frameRange()
+
+        parms = {
+            "picture": '$HIP/pyblish/`chs("subset")`/`chs("subset")`.$F4.png',
```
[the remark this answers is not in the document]
BigRoy: Only in Houdini extractors, they all default to using a pyblish folder in the work area. That could be refactored in a separate PR I'd say if there is anything more sensible to design. :)

openpype/hosts/houdini/plugins/publish/extract_opengl.py
BigRoy: The benefit of doing it like this is that we are able to report to the user all missing files so they can investigate. Otherwise they'd see the single file reported only.

openpype/hosts/houdini/plugins/create/create_review.py
antirotor: with those host specific modules the reason is mostly that it make discovery of the plugins slow so we move it usually to the place where it is directly used.

openpype/hosts/houdini/plugins/create/create_review.py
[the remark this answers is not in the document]
antirotor: It might be set in the Settings the same way you can change render directory in Maya probably.

openpype/hosts/houdini/plugins/publish/extract_opengl.py
fabiaserra: yeah makes sense

openpype/hosts/houdini/plugins/create/create_review.py
fabiaserra: awesome, good to know, thank you!

openpype/hosts/houdini/plugins/create/create_review.py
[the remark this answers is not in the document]
BigRoy: Yes, could be nice for a separate PR to refactor it across all the creators.

BigRoy:
> We might solve it by slightly changing what imprint accepts and use ch() to link "native" parameters to ours. That would be probably the easiest way to implement it.

It would be for both imprint and loading the data.

BigRoy:
So I've updated it again and done some test runs with fixes. I've also added focal length burnin support like @tokejepsen dit with #4531 for Maya.

This PR should put it in a first rough usable state. However do still take a look at the [**Nice-to-haves** list on the initial post](https://github.com/ynput/OpenPype/pull/3839#issue-1371695117) and let me know if there's anything there that should be crucial to getting this PR merged.

@antirotor @fabiaserra it would be gold if any of you could test this out.

BigRoy:
The last three commits accidentally implemented the fix of PR #4739 as well - it shouldn't be a problem, but just so you know. My software apparently didn't switch branches quickly enough while I was already committing the fixes. 🐌

BigRoy:
@tokejepsen What do you think? Should I update the [`maya_burnin` default preset](https://github.com/tokejepsen/pype/blob/b7f65898869774f147e3ffcede90f1d8fed0a8e7/openpype/settings/defaults/project_settings/global.json#L253-L275) to also operate on Houdini host and call it the `focallength_burnin` instead since this PR now also supports the animated focal length?

moonyuet:
Works as expected after cleaning up of the enum in the creator's settings. The updated PR can export image sequences with different image formats. Also it has a similar option of deleting images as if maya.
Basically the user can choose whether they want to keep image sequences or not. By default, the sequences are removed after publishing and only the video is in the related publish folder.
image

Not sure if it is a good idea(it's just personal opinion) - I think the selected node option should mainly allow user selecting the camera for the playblast, as it seems the camera is mandatory for opengl, while the scene path is always left as `/obj/` by default , When you hit render, it will yell at you if there is no existing camera while it won't yell if the scene path is left as  `/obj/` , there should be always validation for the scene path to make sure the user won't keep it empty or set a wrong path.

BigRoy:
> Not sure if it is a good idea(it's just personal opinion) - I think the selected node option should mainly allow user selecting the camera for the playblast, as it seems the camera is mandatory for opengl, while the scene path is always left as /obj/ by default , When you hit render, it will yell at you if there is no existing camera while it won't yell if the scene path is left as /obj/ , there should be always validation for the scene path to make sure the user won't keep it empty or set a wrong path.

Yes, agree that workflow could be better.

Our houdini artist actually mentioned he'd prefer the default scenepath to always be `/obj` and then **Candidate Objects** which defaults to `*` in houdini to be cleared (so it doesn't show anything by default) but then in Force Objects it'd add your selection (whatever in your selection is not a camera) so that what you had selected will always be rendered even if at some point in time the user turns off the display flag.

That'd make the rop node way more explicit and not dependent on display flags of your scene at any point in time. Thoughts?

moonyuet:
> > Not sure if it is a good idea(it's just personal opinion) - I think the selected node option should mainly allow user selecting the camera for the playblast, as it seems the camera is mandatory for opengl, while the scene path is always left as /obj/ by default , When you hit render, it will yell at you if there is no existing camera while it won't yell if the scene path is left as /obj/ , there should be always validation for the scene path to make sure the user won't keep it empty or set a wrong path.
> 
> Yes, agree that workflow could be better.
> 
> Our houdini artist actually mentioned he'd prefer the default scenepath to always be `/obj` and then **Candidate Objects** which defaults to `*` in houdini to be cleared (so it doesn't show anything by default) but then in Force Objects it'd add your selection (whatever in your selection is not a camera) so that what you had selected will always be rendered even if at some point in time the user turns off the display flag.
> 
> That'd make the rop node way more explicit and not dependent on display flags of your scene at any point in time. Thoughts?
Yes, I agree that we should make sure the rop node is independent on display flags.

openpype/hosts/houdini/plugins/create/create_review.py
```diff
@@ -0,0 +1,106 @@
...
+                "tres": override_resolution,
+                "res1": pre_create_data.get("resx"),
+                "res2": pre_create_data.get("resy"),
+                "aspect": pre_create_data.get("aspect"),
+            })
+
+        if self.selected_nodes:
+            # todo: allow only object paths?
+            node_paths = " ".join(node.path() for node in self.selected_nodes)
+            parms.update({"scenepath": node_paths})
```
moonyuet, line 62: This is the code I have in mind after the discussion. Not the best one but it works at some point.
```suggestion
        if self.selected_nodes:
            for node in self.selected_nodes:
                path = node.path()
                if node.type().name() == "cam":
                    parms.update({"camera": path})
                else:
                    node_paths = " ".join(path)
                    parms.update({"forceobjects": node_paths})

**Comment by moonyuet** [comment]
Test it just now. Nice, the camera params of the opengl nodes would be set to the path of the selected camera, while the force objects params would be set to the paths of the selected nodes. 
 I just checked again on the nice-to-have in this PR, and find out you want to use 0, 0 to represent resolution taken from asset data...And somehow the validator does not allow this happened. I double check the opengl node in Houdini and find out it actually works if the user playblast in 0, 0 resolution.  Maybe it is good for testing if the playblast works before the full review playblast as I know some scene can still be huge even there is a cache. 
I am not quite sure about the image format validator though.

**Comment by BigRoy** [comment]
> And somehow the validator does not allow this happened. I double check the opengl node in Houdini and find out it actually works if the user playblast in 0, 0 resolution. Maybe it is good for testing if the playblast works before the full review playblast as I know some scene can still be huge even there is a cache.

I can't really see a production case where you'd want to actually end up playblasting a 0x0 resolution. I think with the validator for now is fine. Just need to add documentation I think.

**Comment by BigRoy** [comment]
@antirotor this should be ready. Could you take a look?
@tokejepsen any chance you could take a peek too and maybe check the documentation?

File: website/docs/pype2/admin_presets_plugins.md [hunk 7]
```diff
@@ -294,17 +294,17 @@ If source representation has suffix **"h264"** and burnin suffix is **"client"**
...
+  | resolution_height | Resolution height.                                                                                                                                                                                                                                      |
+  | fps | Fps of an output.                                                                                                                                                                                                                                       |
+  | timecode | Timecode by frame start and fps.                                                                                                                                                                                                                        |
+  | focalLength | **Only available in Maya and Houdini**<br /><br />Camera focal length per frame. Use syntax `{focalLength:.2f}` for decimal truncating. Eg. `35.234985` with `{focalLength:.2f}` would produce `35.23`, whereas `{focalLength:.0f}` would produce `35`. |
```
BigRoy, line 307: @tokejepsen I noticed only now that this documentation that you added here in this PR https://github.com/ynput/OpenPype/pull/4531 is actually part of the `pype2` documentation. 🤣

tokejepsen (changes requested):

BigRoy:
> Yeah, good point. Should be moved to here; https://openpype.io/docs/project_settings/settings_project_global#publish-plugins

@tokejepsen I'll be moving that in a separate PR. There is no **Extract Burnin** documentation for the OP v3 it seems but only pype v2. Adding that here would make this PR more bloated than needed and I expect that the documentation might get some more feedback going and I'd rather get this PR in separately from that. Does that mean you could approve this PR?

BigRoy:
This PR should be usable and pretty much ready to merge. ❤️

fabiaserra (approved):

tokejepsen (approved):
The docs and code looks good to me.

One thing you might need to have a look out for, cause it has pop up on @kalisp tests, is because the burnin extractor is used on reviews as well as renders, the renders might error with missing focal length.

kalisp:
I created issue [[focalLenght](https://github.com/ynput/OpenPype/issues/4789)](https://github.com/ynput/OpenPype/issues/4789) for Toke's comment.

Looking into code, I don't think it should be applicable for Houdini, imho

BigRoy:
This should be ready to merge @antirotor

antirotor (approved):
Works fine! The only problem (not related to this PR and probably partially solved by #4700) is that the ROP node is not handling gamma correction very nice - I had to manually set gamma to 2.2 when rendering to pngs otherwise the resulting video would be dark. This would be handled by OCIO better, so I suppose it isn't of much concern.

BigRoy:
> Works fine! The only problem (not related to this PR and probably partially solved by https://github.com/ynput/OpenPype/pull/4700) is that the ROP node is not handling gamma correction very nice - I had to manually set gamma to 2.2 when rendering to pngs otherwise the resulting video would be dark. This would be handled by OCIO better, so I suppose it isn't of much concern

Thanks, that does sounds like something we can track in a separate dedicated issue to resolve. 👍

PR #3840, opened 2022-09-13
Houdini: Increment current file on workfile publish

### LINKED ISSUE #2687: Houdini: Increment workfile version on publish
## Request

Houdini is currently missing feature to increment workfile version on publish as for example Maya has.

[cuID:OP-2614]

BigRoy:
## Brief description

This fixes #2687

## Description

Matches logic more to Maya where it also increments current file with the workfile family.
Previous plug-ins had `targets` "local" and "deadline" which I believe aren't even used in OpenPype - so have been non-functional for that time.

## Testing notes:

1. Publish with Houdini
2. Ensure scene incrementing behavior is as expected


mkolar (approved):
Well spotted and does what one would expect. Thank you

Issue #3404, opened 2022-06-24
Maya: Yeti Validate Rig Input


m-u-r-p-h-y:
**Describe the bug**
There is a problem with the validation of Yeti Rig Input meshes. 

```
b'Processing Subset Name'
b'Processing Subset Name'
b'Processing Resources Unique'
// pyblish.ValidateResources : No resources to validate..
b'Processing Resources Unique'
// pyblish.ValidateResources : No resources to validate..
b'Processing Yeti Rig Input Shapes In Instance'
// Error: pyblish.plugin : Traceback (most recent call last):
  File "C:\Program Files (x86)\OpenPype\dependencies\pyblish\plugin.py", line 592, in __implicit_process
    provider.invoke(runner)
  File "C:\Program Files (x86)\OpenPype\dependencies\pyblish\plugin.py", line 113, in invoke
    return func(**inject)
  File "C:\Program Files (x86)\OpenPype\openpype\hosts\maya\plugins\publish\validate_yeti_rig_input_in_instance.py", line 21, in process
RuntimeError: Yeti Rig has invalid input meshes
Traceback (most recent call last):
  File "C:\Program Files (x86)\OpenPype\dependencies\pyblish\plugin.py", line 592, in __implicit_process
    provider.invoke(runner)
  File "C:\Program Files (x86)\OpenPype\dependencies\pyblish\plugin.py", line 113, in invoke
    return func(**inject)
  File "<string>", line 21, in process
RuntimeError: Yeti Rig has invalid input meshes // 
b'Processing Maya Units'
// pyblish.ValidateMayaUnits : Units (linear): cm
// pyblish.ValidateMayaUnits : Units (angular): deg
// pyblish.ValidateMayaUnits : Units (time): 25.0 FPS
```

**Screenshots**
![image](https://user-images.githubusercontent.com/352795/175478650-619cb8a1-f3a4-459e-8bee-8391bc3e3c5f.png)

Maya 2022.3
OP 3.11.1

This particular scene is working in V2 without problems and failing in V3. As the code of Yeti implementation did not change it seems we have a problem somewhere else.

@BigRoy (thanks a lot!) is tracing the issue down to PR #3297

[cuID:OP-3454]

BigRoy:
@m-u-r-p-h-y any thoughts on the above?

LiborBatek:
@BigRoy  Hi! went through your explanation and have a question about dependencies (incomming connections to some entities).

[...]

PR #3841, opened 2022-09-13
Maya + Yeti: Fix yeti cache load using wrong name in edge case

BigRoy:
## Brief description

We hit this issue when the code was trying to create a group node that by pure luck had a node name that matched exactly with a name that was also in the namespace - thus maya would auto-rename the created group node.

As such `group_name != group_node` and the `.setOutlinerColor` failed to do its job correctly.

## Description

This fixes it by actually using the node name returned by maya.

## Testing notes:

Might be hard to test the specific edge case since I'm not entirely sure how we got the exact name match. But just testing a Yeti publish + load to begin with should be a good test.

1. Load a Yeti Cache.

### DIFF
--- a/openpype/hosts/maya/plugins/load/load_yeti_cache.py
+++ b/openpype/hosts/maya/plugins/load/load_yeti_cache.py
@@ -73,8 +73,8 @@ class YetiCacheLoader(load.LoaderPlugin):
 
         c = colors.get(family)
         if c is not None:
-            cmds.setAttr(group_name + ".useOutlinerColor", 1)
-            cmds.setAttr(group_name + ".outlinerColor",
+            cmds.setAttr(group_node + ".useOutlinerColor", 1)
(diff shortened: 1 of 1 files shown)


LiborBatek:
Hi Roy,
will test it out ASAP. We got finally free license for Yeti from Peregrine so it just depends when it will be running @m-u-r-p-h-y

LiborBatek:
Tried out to test it however I have no clue how to get conditions like that :) ...cant get my head around this problem resp how to create such conditions.

However when tested at least yeti publish/load Im encountering the same issue as in here [#3404 ](https://github.com/pypeclub/OpenPype/issues/3404)
So I guess it need to be sorted first.

LiborBatek (commented):
I am not able to prepare such conditions for testing...sorry. Anyone have idea?!

antirotor (approved):
`group_node` is better to use then previous `group_name`, thanks

PR #3843, opened 2022-09-14
General: Simple script for getting license information about used packages

antirotor:
## Enhancement

This is adding simple script that will crawl over python packages installed by Poetry and printing them out along with the version information and license type in Markdown table. This could be then used to update [documentation](https://openpype.io/docs/dev_requirements/#python-modules-we-use-and-their-licenses).

### Usage:

```sh
./.poetry/bin/poetry run python ./tools/get_python_packages_info.py
```


jakubjezek001 (approved):
fantastic!

mkolar (approved):

PR #3846, opened 2022-09-14
Publisher: Instances can be marked as stored

iLLiCiTiT:
## Brief description
Give ability to mark instance object as stored to clear changes.

## Description
New created instance have all attributes marked as changed (because there were none data before) but in most of cases are also stored to scene during creation so they should be marked as stored. Implemented method `mark_stored` to "clear changes" and from that moment instance object behave like has no changes.

Method `mark_as_stored` must be called explicitly in creator. I didn't find a good place how to automate it with being sure that the instance was actually stored. Traypublisher does it.

## Testing notes:
Testing requires some code changes. Easiest is probably using prints in tray publisher:
1. In `~/openpype/hosts/traypublisher/api/pipeline.py` add print of `update_list` variable in `update_instances`
2. Open TrayPublisher
3. Create an instance
4. Hit refresh at the bottom
5. Check console if there is printed output from the print
    - in this branch there should be no print
    - in develop there should be some print with a list of changes

Resolves https://github.com/pypeclub/OpenPype/issues/3772

### DIFF
--- a/openpype/hosts/traypublisher/api/plugin.py
+++ b/openpype/hosts/traypublisher/api/plugin.py
@@ -104,6 +104,8 @@ class TrayPublishCreator(Creator):
 
         # Host implementation of storing metadata about instance
         HostContext.add_instance(new_instance.data_to_store())
+        new_instance.mark_stored()
+
         # Add instance to current context
         self._add_instance_to_context(new_instance)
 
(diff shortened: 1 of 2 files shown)


antirotor (approved):
Run simple tests in publisher refactor for Houdini and it seems to be working well

PR #3852, opened 2022-09-14
Maya: Extract Playblast fix textures + labelize viewport show settings

BigRoy:
## Brief description

Adds a `displayTextures` setting that actually does what you expect it to do - toggle display of textures.
The old `textures` setting is preserved, but now labelized that it is actually about Texture Placements.

I've also labelized all settings to match the Show menu in Maya to avoid confusion.

### FIXES

- This **fixes** "textures" by adding the new but correct `displayTextures` key-value for settings to toggle Display Textures.
- This **fixes** the toggle for HUD / Heads Up Display from #3850 
- This **labelizes** all Viewport Show Options and Camera Options to match more with Maya

## Description

<details>
  <summary>Click to expand and see screenshots</summary>

New settings in OpenPype:
afbeelding

Display Textures is added higher:
afbeelding

Reference: Maya **Show** menu in viewport:
afbeelding

</details>

## Testing notes:

1. Playblast away with nice settings
    - Tweak settings for Extract Review > Viewport Options > Show
    - Publish playblasts with reviews in Maya
2. Test the "displayTextures" option: `project_settings/maya/publish/ExtractPlayblast/capture_preset/Viewport Options/displayTextures`
3. Test the "HUD" option: `project_settings/maya/publish/ExtractPlayblast/capture_preset/Viewport Options/headsUpDisplay`

### DIFF
--- a/openpype/hosts/maya/api/lib.py
+++ b/openpype/hosts/maya/api/lib.py
@@ -2483,7 +2483,7 @@ def load_capture_preset(data=None):
     # DISPLAY OPTIONS
     id = 'Display Options'
     disp_options = {}
-    for key in preset['Display Options']:
+    for key in preset[id]:
         if key.startswith('background'):
             disp_options[key] = preset['Display Options'][key]
             if len(disp_options[key]) == 4:
--- a/openpype/settings/defaults/project_settings/maya.json
+++ b/openpype/settings/defaults/project_settings/maya.json
@@ -678,9 +678,6 @@
                     "isolate_view": true,
                     "off_screen": true
                 },
-                "PanZoom": {
-                    "pan_zoom": true
-                },
                 "Renderer": {
                     "rendererName": "vp2Renderer"
                 },
--- a/openpype/settings/defaults/project_settings/maya.json
+++ b/openpype/settings/defaults/project_settings/maya.json
@@ -686,9 +686,7 @@
                 },
                 "Resolution": {
                     "width": 1920,
-                    "height": 1080,
-                    "percent": 1.0,
-                    "mode": "Custom"
+                    "height": 1080
                 },
                 "Viewport Options": {
                     "override_viewport_options": true,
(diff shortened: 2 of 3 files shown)


BigRoy:
Ok. Cleaning up some more of these settings - I have now also labelized the Camera Settings.

Screenshot:

afbeelding


---

## Question

Does anyone know what this setting is supposed to do:
```
project_settings/maya/publish/ExtractPlayblast/capture_preset/Resolution/mode
```

It defaults to text value:
```
Custom
```

It seems unused?

There is also a `percent` setting:
```
project_settings/maya/publish/ExtractPlayblast/capture_preset/Resolution/percent
```

Of which the value also seems unused in the code base for Extract Playblast and Extract Thumbnail.
Both seem unused in [`load_capture_preset`](https://github.com/pypeclub/OpenPype/blob/ad8a7c86e4b655014e6dc776c813e9966cb9e1f3/openpype/hosts/maya/api/lib.py#L2461).

Remove?

antirotor:
> There is also a `percent` setting

Maybe I am wrong but I think this was added because of feature never implemented to specify overscan as percent value. If it is not used in the code I would remove it.

m-u-r-p-h-y:
is this one doing something btw?
image

BigRoy:
[> to scale up/down the output resolution with the coefficient. (no need to calculate and input width and scale)](https://github.com/pypeclub/OpenPype/pull/3852#issuecomment-1246910209)

I thought that was the case but couldn't find any logic whatsoever related to actually doing that percent scaling. I assume it never worked. If you can confirm that it'd be gold.

[> is this one doing something btw?](https://github.com/pypeclub/OpenPype/pull/3852#issuecomment-1246911466)

`PanZoom` and `pan_zoom` strings only ever occur in the OP settings, nowhere else in code. Also not in the capture lib.

So also seems UNUSED. Will remove for now. If we'd like that feature implemented, best to create an issue and produce that separate from this PR.

m-u-r-p-h-y (approved):
`Viewport Options / Show ` options work as expected, including hardware texturing

image

mkolar (approved):

PR #3853, opened 2022-09-14
Settings + Houdini: Fix missing default settings value for houdini shelf definition

BigRoy:
## Brief description

This #3652 PR introduced a Houdini Shelf Manager - however the `shelf_definition` was missing in default values for the settings and thus OP would complain about it.

## Description

Saved default settings for Houdini.

## Testing notes:

1. Houdini Shelf manager?


BigRoy:
Fixed with https://github.com/pypeclub/OpenPype/pull/3870 instead

PR #3854, opened 2022-09-14
Photoshop: synchronize image version with workfile

kalisp:
## Brief description
Adds configurable Setting to control if publishable instances ('image', 'review' family) should take version from workfile.

## Testing notes:
1. configure `project_settings/photoshop/publish/CollectVersion`
Screenshot 2022-09-16 143114

3. publish only workfile (disable image and review instances) - that way gap in their versions will be created
4. publish from Photoshop (manually and from WP)
5. version of image (and review) should match to workfile


openpype/settings/entities/schemas/projects_schema/schema_project_photoshop.json
```diff
@@ -131,6 +131,28 @@
...
+                        {
+                            "type": "boolean",
+                            "key": "optional",
+                            "label": "Optional"
+                        },
+                        {
+                            "type": "label",
+                            "label": "Synchronize version for image and review instances by workfile version."
+                        },
+                        {
```
iLLiCiTiT, line 149: I would replace `sync_workfile_version` with `enabled`. It doesn't make sense to show the plugin if it's functionality is disabled.

iLLiCiTiT: Just realized it's collector... You can't have optional collector (maybe in new publisher?).

BigRoy:
How does this differ from the global plug-in behavior defined with the setting `project_settings/global/publish/CollectAnatomyInstanceData/follow_workfile_version`? (and maybe also `project_settings/global/publish/CollectSceneVersion`?)

That should make it so that publish versions would match your workfile version.

iLLiCiTiT:
> That should make it so that publish versions would match your workfile version.

That part "prefills" anatomy data but it doesn't change `instance.data["version"]` (and it should not). This is optional feature based on workfile version. Not all hosts have workfiles and can sync instances...

BigRoy:
> That part "prefills" anatomy data but it doesn't change instance.data["version"] (and it should not). This is optional feature based on workfile version. Not all hosts have workfiles and can sync instances...

I'm not entirely sure I understand. The `sync_workfile_version` setting I mentioned seems to do what I described it does for Maya and Houdini. 

Description-wise this feature tries to do exactly that? Your reply doesn't explain to me why Photoshop in this case needs unique functionality in that regard.

It sounds like what you're looking to do is be able to regulate what hosts or publishes would need to sync publishes with the workfile version.

In that case, should maybe the original sync workfile version start using profiles to be able to enable/disable against Apps, Families, Tasks, etc. for if one would like to enable just for a subset of hosts?

kalisp:
So you are saying enhance `CollectAnatomyInstanceData/follow_workfile_version` with profiles:
- Host
- Task
- Family

and remove current implementation of `sync_workfile_version` (currently only implemented for Maya, Nuke, Harmony).
That way we would get this behavior (force version from **workfile** name to all published artifacts, eg. published workfile, image, review whatever) for all hosts for cheap, with reasonable granularity.

I second that.

Things I don't particularly like:
- configuring this in `CollectAnatomyInstanceData` seems kinda hidden 
- do we want allow artist to disable this for particular publish?

(Just note to myself - there are use cases - Webpublisher/Tray publisher - where we shouldn't depend on version in uploaded workfile, as freelance artist might not know correct version, might not follow naming convention etc.)

iLLiCiTiT:
If we'll somehow manage to make it a global plugin it should not happen in this PR. This PR is to fix current issue on client site, conversion to global plugin has a lot of caviads that would stop currently needed fix.

I would just change `sync_workfile_version` to `enabled` and it's ready to merge from my point of view.

kalisp:
I would rather stick with `sync_workfile_version` to tie this to existing implementations that when they will be removed, this won't get skipped accidentally.

BigRoy:
That was indeed what I was saying. But what I meant to say more exactly is... **it already is global, right?**. We've enabled that single setting and it has synced published versions to work files for both Maya + Houdini on our end.

I'm seeing no logic that is Houdini or Maya specific that implements this - except for the logic that makes it persist and come across with maya renders -> deadline -> publish job 

> Things I don't particularly like:
>     configuring this in CollectAnatomyInstanceData seems kinda hidden
>     do we want allow artist to disable this for particular publish?

Agreed, don't like that either.

iLLiCiTiT:
> I would rather stick with sync_workfile_version to tie this to existing implementations that when they will be removed, this won't get skipped accidentally.

But they have it that way because it's "side feature" of a collector plugin. Here it is the only purpose of the plugin.

BigRoy:
> But they have it that way because it's "side feature" of a collector plugin. Here it is the only purpose of the plugin

I like that. Can we have this functionality then replace the other one. So that this would also run for Houdini, Maya, etc.?

kalisp:
> But they have it that way because it's "side feature" of a collector plugin. Here it is the only purpose of the plugin.

Ok, renamed.

iLLiCiTiT (approved):

64qam:
The image version should be the same as the published workfile version. Now it is working for Web publisher (the published worfiles have the same version as published image files), but not for PS. When the image is published from Photoshop, it results in the version of the working (unpublished) workfile. E.g., after publishing from PS, there are working workfile v079, published workfile v010, and published image v079. But when publishing from WP, there are published workfile v011 and published image v011.

kalisp:
This is expected behavior confirmed on internal discussion.

PR #3856, opened 2022-09-15
Tray Publisher: skip plugin if otioTimeline is missing

antirotor:
## Fix

This is quick fix for publishing workfile with tray publisher, skipping **Extract OTIO File** if data is missing on instance.

It should be fixed more thoroughly but I hadn't time to dig deeper as how this OTIO stuff works with workfilles and this is needed by client.


iLLiCiTiT:
Created issue related to proper fix https://github.com/pypeclub/OpenPype/issues/3861

iLLiCiTiT (approved):
Approved as quick fix

PR #3860, opened 2022-09-15
Flame: make migratable projects after creation

jakubjezek001:
## Brief description
Flame projects are migratable after they are created.

## Description
It was reported that when project is created by `user01` at `machine01`, Flame is creating its own unmanaged project data which are locked to 722 and therefore they are not migratable to other machines. This is workaround. The process will open immediately all those files after the project is crated.

### DIFF
--- a/openpype/hosts/flame/hooks/pre_flame_setup.py
+++ b/openpype/hosts/flame/hooks/pre_flame_setup.py
@@ -129,7 +129,7 @@ class FlamePrelaunch(PreLaunchHook):
                     for name in set(dirs) | set(files):
                         path = os.path.join(root, name)
                         st = os.stat(path)
-                        if oct(st.st_mode) != self.permissions:
+                        if oct(st.st_mode) != self.permisisons:
                             os.chmod(path, self.permisisons)
 
                 except OSError as exc:
--- a/openpype/hosts/flame/hooks/pre_flame_setup.py
+++ b/openpype/hosts/flame/hooks/pre_flame_setup.py
@@ -126,12 +126,14 @@ class FlamePrelaunch(PreLaunchHook):
         for dirtm in dirs_to_modify:
             for root, dirs, files in os.walk(dirtm):
                 try:
-                    for d in dirs:
-                        os.chmod(os.path.join(root, d), self.permisisons)
-                    for f in files:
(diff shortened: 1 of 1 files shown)


openpype/hosts/flame/hooks/pre_flame_setup.py
```diff
@@ -99,8 +101,39 @@ def execute(self):
...
+
+        Reported issue: when project is created locally on one machine,
+        it is impossible to migrate it to other machine. Autodesk Flame
+        is crating some unmanagable files which needs to be opened to 0o777.
+
+        Args:
+            project_name (str): project name
+            volume_name (str): studio volume
+        """
+        dirs_to_modify = [
```
iLLiCiTiT, line 120: Not sure if flame can be opened on windows?
```suggestion

        # Skip if 'chmod' is not available
        if not hasattr(os, "chmod"):
            return

        dirs_to_modify = [
```

openpype/hosts/flame/hooks/pre_flame_setup.py
```diff
@@ -99,8 +101,39 @@ def execute(self):
...
+
+        for dirtm in dirs_to_modify:
+            for root, dirs, files in os.walk(dirtm):
+                try:
+                    for d in dirs:
+                        os.chmod(os.path.join(root, d), self.permisisons)
+                    for f in files:
+                        os.chmod(os.path.join(root, f), self.permisisons)
+                except OSError as _E:
+                    self.log.warning("Not able to open files: {}".format(_E))
```
iLLiCiTiT, line 134: It's a little bit longer but I would skip all dirs or files which already have expected permissions.
```suggestion
                try:
                    for name in set(dirs) | set(files):
                        path = os.path.join(root, name)
                        st = os.stat(path)
                        if oct(st.st_mode) != self.permissions:
                            os.chmod(path, self.permisisons)

                except OSError as exc:
                    self.log.warning("Not able to open files: {}".format(exc))
```

openpype/hosts/flame/hooks/pre_flame_setup.py
jakubjezek001: No it cannot. It is only Linux and Macos..

iLLiCiTiT (approved):
Approved

PR #3862, opened 2022-09-15
Publisher: Increase size of main window

iLLiCiTiT:
## Brief description
Main window of publisher is slighly bigger.

## Description
The publisher window can be bigger then is now to add more space for attributes and create dialog is smaller then the window and is clearly visible it is not the same window - it blended before.

## Testing notes:
1. Open Publisher in a host which has creators (e.g. tray publisher)


m-u-r-p-h-y (approved):
Before
image


After
image



Approved!

PR #3864, opened 2022-09-15
Github issues adding `running version` section

jakubjezek001:
Adding running version


BigRoy:
Yes, good idea. I think Running Version could be renamed to "OpenPype version"?

jakubjezek001:
> Yes, good idea. I think Running Version could be renamed to "OpenPype version"?

Let's keep the `openpype` out of it. Just in case we need to rename it in future..

BigRoy (approved):

PR #3867, opened 2022-09-16
TrayPublisher: added text filter on project name to Tray Publisher

kalisp:
## Brief description
This quick text filter should help studios with a large amount of projects 

## Additional info
Screenshot 2022-09-16 114118

## Testing notes:
1. type some substring of project name


openpype/tools/traypublisher/window.py
```diff
@@ -59,10 +61,17 @@ def __init__(self, publisher_window):
         btns_layout.addWidget(cancel_btn, 0)
         btns_layout.addWidget(confirm_btn, 0)
 
+        txt_filter = QtWidgets.QLineEdit()
```
iLLiCiTiT, line 64: Use `PlaceholderLineEdit` from `openpype.tools.utils` to draw text color properly and please pass `content_widget` as parent.

openpype/tools/traypublisher/window.py
```diff
@@ -79,11 +88,15 @@ def __init__(self, publisher_window):
         projects_view.doubleClicked.connect(self._on_double_click)
         confirm_btn.clicked.connect(self._on_confirm_click)
         cancel_btn.clicked.connect(self._on_cancel_click)
+        txt_filter.textChanged.connect(
+            lambda: projects_proxy.setFilterRegularExpression(
```
iLLiCiTiT, line 92: I would prefer speciam method as callback. Lambdas has few issues on QObject destroy.

kalisp:
Updated according to comments.

openpype/tools/utils/models.py
```diff
@@ -356,10 +356,12 @@ def lessThan(self, left_index, right_index):
 
     def filterAcceptsRow(self, source_row, source_parent):
         index = self.sourceModel().index(source_row, 0, source_parent)
+        string_pattern = self.filterRegularExpression().pattern()
         if self._filter_enabled:
             result = self._custom_index_filter(index)
             if result is not None:
-                return result
+                project_name = index.data(PROJECT_NAME_ROLE)
+                return string_pattern in project_name
```
iLLiCiTiT, line 364: Default project has name set to `None`
```suggestion
        if result is not None:
                project_name = index.data(PROJECT_NAME_ROLE)
                if project_name is None:
                    return result
                return string_pattern.lower() in project_name.lower()
```

iLLiCiTiT:
Not related to this PR but can you add sorting to `showEvent` after project refresh?
```
    def showEvent(self, event):
        self._projects_model.refresh()
        self._projects_proxy.sort(0)
```

iLLiCiTiT:
Also would be good to change case sensitivity of projects sorting in `openpype/tools/utils/models.py`
```
class ProjectSortFilterProxy(QtCore.QSortFilterProxyModel):
    def __init__(self, *args, **kwargs):
        super(ProjectSortFilterProxy, self).__init__(*args, **kwargs)
        self.setSortCaseSensitivity(QtCore.Qt.CaseInsensitive)
        self._filter_enabled = True
```

m-u-r-p-h-y (approved):
searching works as expected, 

it is case sensitive

approved . . .

openpype/tools/utils/models.py
kalisp: I tested it, it never crashed.

iLLiCiTiT: Please do it anyway. It's not TrayPublisher's proxy model but general proxy model in utils...

kalisp:
Added safer resolving.
Added sorting after refresh.
Changed case sensitivity.

openpype/tools/utils/models.py
```diff
@@ -335,6 +335,9 @@ class ProjectSortFilterProxy(QtCore.QSortFilterProxyModel):
     def __init__(self, *args, **kwargs):
         super(ProjectSortFilterProxy, self).__init__(*args, **kwargs)
         self._filter_enabled = True
+        # Disable case sensitivity
+        self.setSortCaseSensitivity(QtCore.Qt.CaseInsensitive)
+        self._filter_enabled = True
```
iLLiCiTiT, line 340: Duplicated attribute
```suggestion
```

iLLiCiTiT (approved):
Approved

PR #3873, opened 2022-09-16
Maya: Remove hardcoded requirement for maya/ start for image file prefix

### LINKED ISSUE #3872: Maya: Remove hardcoded "maya" subfolder for renders

The Maya render implementation currently forces the `maya/` folder in the filename prefix to avoid clashes with renders of potential other DCCs. 

I have two issues with this:
- This is now hardcoded but should be customizable.
  - We use a custom work dir template that includes `{app}` and thus our renders would never clash with other DCCs - making this 'hardcoded' entry very annoying. 
- It doesn't actually ensure maya's renders go into that subfolder. (e.g. interactive renders from maya go into `/renders/tmp/maya` instead of `/renders/maya/tmp`

Do not hardcode this as a requirement.

And instead of forcing it into the file prefix, it would be better to tweak Maya's workspace to enforce the `maya` subfolder for those who would like that.
It would be as trivial as setting the default "images" rule of the workspace to `renders/maya` instead of `renders`. This would also mak the `tmp/` renders go into that subfolder (and also even those that render with file prefix not set - by any odd chance)

**Additional context**

For setting `project_anatomy/templates/work/folder` we use:
```
{root[work]}/{project[name]}/{hierarchy}/{asset}/work/{task[name]}/{app}
```
Instead of:
```
{root[work]}/{project[name]}/{hierarchy}/{asset}/work/{task[name]}
```

[Mentioned on OpenPype discord](https://discord.com/channels/517362899170230292/563751989075378201/1020287872974209024)

[cuID:OP-4073]

BigRoy:
## Brief description

This PR is to resolve #3872 

This is just a code prototype - I haven't tested this. If someone could, that'd be very appreciated!

## Description

This will remove the `maya/` requirement from the file prefix and by default moves that behavior into the `workspace.mel` by adding it into the "images" file rule defining where Maya puts its renders.

Moving it there allowed less logic to be implemented, and also allowing the customization as described in #3872. To NOT render into a `maya/` subfolder one would only need to set the "images" file rule to `renders` instead of `renders/maya`

## Testing notes:

1. Launch Maya
2. Submit renders
3. Ensure renders end up in similar position as before - that is `renders/maya/{scene}/{renderlayer}_{renderpass}` or alike.


openpype/hosts/maya/plugins/publish/validate_rendersettings.py
```diff
@@ -116,7 +116,7 @@ def get_invalid(cls, instance):
         prefix = prefix.replace(
             "{aov_separator}", instance.data.get("aovSeparator", "_"))
 
-        required_prefix = "maya/<scene>"
+        required_prefix = "<scene>"
```
BigRoy, line 119: Isn't it odd that we're still hardcoding the beginning part of the file prefix as a not customizable requirement. I'd say we can remove this completely and rely on the project settings for image prefixes per renderer.

Thoughts?

BigRoy: Solid. Let me know if I should remove that `required_prefix` validation in this PR.

BigRoy: Required prefix gets removed in this PR: https://github.com/pypeclub/OpenPype/pull/3880

LiborBatek:
Seems like it gives error when validating the scene before publish of render to deadline

image

when tried to get rid of "maya/" to just : "<Scene>/<RenderLayer>/<RenderLayer>_<RenderPass>" also doesnt work and gives different error:
image

BigRoy:
Had another look - so this is what likely is the case:

The creation of a workspace.mel is only done in the code if the workspace.mel does not exist yet - otherwise it's skipped. As such, your workspace 'images' file rule is likely still set to the old setting of the previous workspace.mel definition.

However, a "repair" of the `ValidateRenderImageRule` plug-in with label **Images File Rule (Workspace)** should work. That should solve it for those scenes.

For new maya workspaces it should be correct by default.

I've merged with latest develop and removed a 'required prefix' which should allow users to keep the old settings as override on a project if they'd like.

LiborBatek:
Not sure it its the case because when I ve deleted the workspace.mel (backed up) new one was created at the launch but they were completely same...

however it works normally now...validating, Publish and renders sitting were they should. 

image

Should I test other render prefix situation? or just the above one?
Let me know if not I will Review it as ok.
Thx

BigRoy:
> Should I test other render render prefixes situation? or just the above one?
> Let me know
> if not I will Review it as ok.

Please check within a workspace that was previously created before this PR - we'd want as little issues and interuption to existing users as possible. We don't want to break their projects in an annoying way. So at the very least that IF the change of settings requires a repair during validation then that message should be clear to the user AND the repair should work as expected.

Maybe it's also good - just to be sure to check a version with and without AOVs - and maybe with and without merge AOVs. (Just so we have spread a relatively wide testing net)

LiborBatek:
When testing with this OP Project Settings:
image

Seems ok too
image

Also tested with AOVs on and MergeAOV on/off.  All renders being in correct locations.

LiborBatek (approved):
Seems working ok. Being tested with both cases of ```workspace.mel``` without/with matching  image file prefixes ```renders``` and ```renders/maya``` and with/without AOVs non merged/merged files.
Render image locations were correct for both and all the validating/publishing ran smoothly.

PR #3874, opened 2022-09-16
Remove lockfile during publish

moonyuet:
## Brief description
During publish, the lockfile isn't removed currently for the maya scene when incrementing and saving the scene.\

### Solution
Adding the remove_lockfile function in `save_scene.py`

### DIFF
--- a/openpype/hosts/maya/plugins/publish/save_scene.py
+++ b/openpype/hosts/maya/plugins/publish/save_scene.py
@@ -1,5 +1,9 @@
 import pyblish.api
-
+from openpype.pipeline.workfile.lock_workfile import(
+    is_workfile_lock_enabled,
+    remove_workfile_lock
+)
+from openpype.pipeline import legacy_io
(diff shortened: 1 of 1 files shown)


openpype/hosts/maya/plugins/publish/save_scene.py
```diff
@@ -23,5 +27,9 @@ def process(self, context):
                            "are no modifications..")
             return
 
+        active_project = legacy_io.active_project()
+        # remove lockfile before saving
+        if is_workfile_lock_enabled("maya", active_project):
```
iLLiCiTiT, line 32:
```suggestion
        project_name = context.data["projectName"]
        project_settings = context.data["project_settings"]
        # remove lockfile before saving
        if is_workfile_lock_enabled("maya", project_name, project_settings):
```

iLLiCiTiT (approved):
Seems reasonable

PR #3875, opened 2022-09-16
OP-1233/ Import Reference during Publish

moonyuet:
## Brief description
The references are imported during publish if you set the attribute of "import reference" in look instance to True. So if the users want to import references, they can enable the function. While they dont want, they can just leave it by default(which doesn't import any reference). 
image

### DIFF
--- a/openpype/hosts/maya/plugins/publish/extract_look.py
+++ b/openpype/hosts/maya/plugins/publish/extract_look.py
@@ -255,6 +255,14 @@ class ExtractLook(publish.Extractor):
         hashes = results["fileHashes"]
         remap = results["attrRemap"]
 
+        # Import Reference if the option is enabled
+        ref_import = instance.data.get("importReference", True)
+        if ref_import:
+            reference_node = cmds.ls(type="reference")
+            for r in reference_node:
+                rFile = cmds.referenceQuery(r, f=True)
+                cmds.file(rFile, importReference=True)
+
         # Extract in correct render layer
         layer = instance.data.get("renderlayer", "defaultRenderLayer")
         with lib.renderlayer(layer):
--- a/openpype/hosts/maya/plugins/create/create_look.py
+++ b/openpype/hosts/maya/plugins/create/create_look.py
@@ -21,6 +21,9 @@ class CreateLook(plugin.Creator):
         # Whether to automatically convert the textures to .tx upon publish.
         self.data["maketx"] = self.make_tx
 
+        # Enable users to import reference
+        self.data["importReference"] = False
+
         # Enable users to force a copy.
         # - on Windows is "forceCopy" always changed to `True` because of
         #   windows implementation of hardlinks


openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -255,6 +255,14 @@ def process(self, instance):
         hashes = results["fileHashes"]
         remap = results["attrRemap"]
 
+        # Import Reference if the option is enabled
+        ref_import = instance.data.get("importReference", True)
+        if ref_import:
+            reference_node = cmds.ls(type="reference")
+            for r in reference_node:
+                rFile = cmds.referenceQuery(r, f=True)
+                cmds.file(rFile, importReference=True)
```
BigRoy, line 264: Sorry. Why would I want to import into my workfile during publish? And why on look publishing. This seems odd.

The [resulting published look already does NOT preserve references due to this argument](https://github.com/pypeclub/OpenPype/blob/8ae92351a7e3e1758cca4a62e4170749ed93a914/openpype/hosts/maya/plugins/publish/extract_look.py#L284) and thus results in a file without references as if they were imported.

m-u-r-p-h-y: I think this want meant for rendering not look publish, 

We should have the possibility to import all references for published render workfile which is used for rendering. (it  should not affect the workfile itself)

- to prevent render farm issues with referenced files
- to help with cloud rendering and dealing with dependencies

openpype/hosts/maya/plugins/publish/extract_import_reference.py
```diff
@@ -0,0 +1,124 @@
...
+from openpype.settings import get_project_settings
+from openpype.hosts.maya.api import lib
+
+
+def _get_project_setting():
+    project_name = legacy_io.active_project()
+    project_setting = get_project_settings(project_name)
+    maya_enabled = (
+        project_setting["maya"]["publish"]["ImportReference"]["enabled"]
+    )
```
BigRoy, line 17: Checking this should be redundant since if the setting is disabled the plug-in shouldn't show to begin with. Right? It should automatically apply that setting if I'm not mistaken.

openpype/hosts/maya/plugins/publish/extract_import_reference.py
```diff
@@ -0,0 +1,124 @@
...
+            # use extension mapping for first family found
+            for family in self.families:
+                try:
+                    self.scene_type = ext_mapping[family]
+                    self.log.info(
+                        "Using {} as scene type".format(self.scene_type))
+                    break
+                except KeyError:
+                    # no preset found
+                    pass
```
BigRoy, line 59: `self.scene_type` might be undefinded if no ext mapping is defined or a keyerror is hit. There currently appears to be no fallback scene type.

openpype/hosts/maya/plugins/publish/extract_import_reference.py
```diff
@@ -0,0 +1,124 @@
...
+        tmp_name = instance.name + self.tmp_format
+        m_ref_fname = "{0}.{1}".format(tmp_name, self.scene_type)
+
+        m_ref_path = os.path.join(dir_path, m_ref_fname)
+
+        self.log.info("Performing extraction..")
+        current = cmds.file(query=True, sceneName=True)
+        cmds.file(save=True, force=True)
+
+        self.log.info("Performing extraction..")
```
BigRoy, line 76: Logging the same message twice - probably wrong?

openpype/hosts/maya/plugins/publish/extract_import_reference.py
```diff
@@ -0,0 +1,124 @@
...
+
+        # create temp scene with imported
+        # reference for rendering
+        reference_node = cmds.ls(type="reference")
+        for r in reference_node:
+            rFile = cmds.referenceQuery(r, f=True)
+            if r == "sharedReferenceNode":
+                cmds.file(rFile, removeReference=True, referenceNode=r)
+            cmds.file(rFile, importReference=True)
```
BigRoy, line 85: Avoid the usage of the uppercase variable `rFile` but use lowercase e.g. `ref_file`

openpype/hosts/maya/plugins/publish/extract_import_reference.py
```diff
@@ -0,0 +1,124 @@
...
+
+        instance.data["files"].append(m_ref_path)
+
+        if instance.data.get("representations") is None:
+            instance.data["representations"] = []
+
+        ref_representation = {
+            "name": self.scene_type,
+            "ext": self.scene_type,
+            "files": os.path.basename(m_ref_fname),
```
BigRoy, line 116: Shouldn't `files` be a list? (Not entirely sure)

openpype/hosts/maya/plugins/publish/extract_import_reference.py
```diff
@@ -0,0 +1,124 @@
...
+        The temp scene with imported reference is
+        published for rendering if this extractor is activated
+
+    """
+
+    label = "Import Reference"
+    order = pyblish.api.ExtractorOrder - 0.48
+    hosts = ["maya"]
+    families = ["renderlayer", "workfile"]
+    active = _get_project_setting()
```
BigRoy, line 40: Setting the active state like this probably won't work because it'd get overridden by OpenPype settings for plug-in discovery due to the `project_settings/maya/publish/ImportReference/enabled` setting existence.

I suspect this is wrong.

Also, it feels off that we're toggling this based on `MayaSubmitDeadline/use_published` setting but it also applies to `workfile` family - which means that it'd ALWAYS trigger for every workfile, even if not used for rendering.

We're mixing logic here.

openpype/hosts/maya/plugins/publish/extract_import_reference.py
```diff
@@ -0,0 +1,124 @@
...
+            rFile = cmds.referenceQuery(r, f=True)
+            if r == "sharedReferenceNode":
+                cmds.file(rFile, removeReference=True, referenceNode=r)
+            cmds.file(rFile, importReference=True)
+
+        if current.endswith(self.scene_type):
+            current_path = os.path.dirname(current)
+            tmp_path_name = os.path.join(current_path, tmp_name)
+            cmds.file(rename=tmp_path_name)
+            cmds.file(save=True, force=True)
```
BigRoy, line 91: If the scene type matches - why do we need to save it as temporary file?

openpype/hosts/maya/plugins/publish/extract_import_reference.py
moonyuet: I will set the enable option of import reference into the submit maya deadline instead of creating two enabled options

openpype/hosts/maya/plugins/publish/extract_import_reference.py
moonyuet: the temp file is storing the scene with the imported reference and it is used for submitting to the deadline while the actual workfile scene isn't. I think we still need to figure out the way to clean up temp scene after the render finished.

openpype/hosts/maya/plugins/publish/extract_import_reference.py
moonyuet: I am still thinking about which families is the best suited for this case, as it is asking the system to publish the temporary scene as a published workfile, while saving a temporary scene as a "current" workfile to use for rendering if the published workfile option is turned off. Perhaps we can introduce `optional = True ` in this extractor so that the user can choose to turn on/off in the publisher

openpype/hosts/maya/plugins/publish/extract_import_reference.py
antirotor: I think it can be both, but I am not sure either :)

antirotor:
I've noticed this during the integration step:
image

That means the results are two files - `OPC_s010_workfileModeling_v098.ma` and `OPC_s010_workfileModeling_v098.mb` - that seems to be a bit confusing.

antirotor:
And the job in deadline still points to the original scene:
image

antirotor:
So what about doing it like this:

```python
import sys
import os
from openpype.lib import source_hash, run_subprocess

ref_scene_name = "some_scene_file.ma"
script = f"""
import maya.standalone

maya.standalone.initialize()
reference_node = cmds.ls(type='reference')
for ref in reference_node:
    ref_file = cmds.referenceQuery(r, f=True)
    if ref == 'sharedReferenceNode':
        cmds.file(ref_file, removeReference=True, referenceNode=ref)
    else:
        cmds.file(ref_file, importReference=True)

cmds.file(rename='{ref_scene_name}')
cmds.file(save=True, force=True)
"""
mayapy_exe = os.path.join(os.getenv("MAYA_LOCATION"), "bin", "mayapy")
if sys.platform == "windows":
    mayapy_exe = mayapy_exe + ".exe"

subprocess_args = [
    mayapy_exe,
    "-c",
   script.replace("\n", ";")
]


try:
    out = run_subprocess(subprocess_args)
except Exception:
    self.log.error("Maketx converion failed", exc_info=True)
    raise
```

Note that this is just code fragment written from top of my head and needs to be adjusted ofc and integreted...

openpype/hosts/maya/plugins/publish/extract_import_reference.py
```diff
@@ -0,0 +1,135 @@
...
+        subprocess_args = [
+            mayapy_exe,
+            "-c",
+        script.replace("\n", ";")
+        ]
+        try:
+            out = run_subprocess(subprocess_args)
+        except Exception:
+            self.log.error("Import reference failed", exc_info=True)
+            raise
```
antirotor, line 101: Please consider this - I've tested it and it works (taking a lot of time executing in on bigger scenes

```suggestion
        # This generates script for mayapy to take care of reference
        # importing outside current session. It is passing current scene
        # name and destination scene name.
        script = ("""
# -*- coding: utf-8 -*-
'''Script to import references to given scene.'''
import maya.standalone

maya.standalone.initialize()

# scene names filled by caller
current_name = "{current_name}"
ref_scene_name = "{ref_scene_name}"

print(">>> Opening {{}} ...".format(current_name))
cmds.file(current_name, open=True, force=True)
reference_node = cmds.ls(type='reference')
print(">>> Processing references")
for ref in reference_node:
    ref_file = cmds.referenceQuery(ref, f=True)
    print("--- {{}}".format(ref))
    print("--> {{}}".format(ref_file))
    if ref == 'sharedReferenceNode':
        cmds.file(ref_file, removeReference=True, referenceNode=ref)
    else:
        cmds.file(ref_file, importReference=True)

print(">>> Saving scene as {{}}".format(ref_scene_name))
try:
    cmds.file(rename=ref_scene_name)
except SyntaxError:
    cmds.file(rename=ref_scene_name)
    cmds.file(save=True, force=True)
print("*** Done")
        """).format(current_name=current_name, ref_scene_name=ref_scene_name)
        mayapy_exe = os.path.join(os.getenv("MAYA_LOCATION"), "bin", "mayapy")
        if sys.platform == "windows":
            mayapy_exe += ".exe"
            mayapy_exe = os.path.normpath(mayapy_exe)

        # can't use TemporaryNamedFile as that can't be opened in another
        # process until handles are closed by context manager.
        with tempfile.TemporaryDirectory() as tmp_dir_name:
            tmp_file_name = os.path.join(tmp_dir_name, "import_ref.py")
            tmp = open(tmp_file_name, "w+t")

            subprocess_args = [
                mayapy_exe,
                tmp_file_name
            ]
            self.log.info("Using temp file: {}".format(tmp.name))
            try:
                tmp.write(script)
                tmp.close()
                run_subprocess(subprocess_args)
            except Exception:
                self.log.error("Import reference failed", exc_info=True)
                raise
```

antirotor:
btw is this PR ready for review? I suppose so. Switching from draft, but please correct me if I am wrong.

openpype/hosts/maya/plugins/publish/extract_import_reference.py
```diff
@@ -0,0 +1,160 @@
...
+from openpype.hosts.maya.api import lib
+
+
+def _import_reference():
+    project_name = legacy_io.active_project()
+    project_setting = get_project_settings(project_name)
+    import_reference = (
+        project_setting["deadline"]["publish"]["MayaSubmitDeadline"]["import_reference"] # noqa
+    )
+    return import_reference
```
antirotor, line 21: I am not sure if this is good idea to do it like this. First: it won't work in new publisher, second is that it will be overwritten by the plugin settings itself when (if) they are created. For example if TD would liek to control `optional` attribute.

I suggest implementing class method `apply_settings(project_setting, system_settings)`.

Should be enough to add:

```python

@classmethod
def apply_settings(cls, project_setting, system_settings):
    cls.active = project_setting["deadline"]["publish"]["MayaSubmitDeadline"]["import_reference"] # noqa
```
but later on, when other settings are defined it must handle all of them here.

antirotor (changes requested):
Otherwise it looks good

moonyuet:
> I've just tested it in the complex scene and it fails on the nested references (references inside references) with the error:
> 
> ```
> The specified reference file cannot be imported because its parent file is not the top-level scene file. 
> ```
> 
> Good way to replicate it is to publish a set dress from Maya with referenced models and then load that set dress as a reference into the render scene. To make it usable, we'll need to be able to flatten all the references.
> 
> **Extractor error:** image
> 
> **Offending reference:** image

Yes, I agree. For the updated PR I take reference from setdress publisher and rewrite the code to import reference while reunning it in maya standalone. It should import references of reference and also import reference.
image

antirotor (changes requested):
So one additional case we need to catch:

```
  File "C:\Users\annat\AppData\Local\Temp\tmp78o_0ryn\import_ref.py", line 15, in <module> 
    cmds.file(ref, importReference=True) 
RuntimeError: The specified reference file cannot be imported because it is unloaded. 
```

Guess we need to skip all unloaded references.

m-u-r-p-h-y:
> Guess we need to skip all unloaded references.

good catch!

openpype/hosts/maya/plugins/publish/extract_import_reference.py
```diff
@@ -0,0 +1,156 @@
...
+maya.standalone.initialize()
+# scene names filled by caller
+current_name = "{current_name}"
+ref_scene_name = "{ref_scene_name}"
+print(">>> Opening {{}} ...".format(current_name))
+cmds.file(current_name, open=True, force=True)
+print(">>> Processing references")
+all_reference = cmds.file(q=True, reference=True) or []
+for ref in all_reference:
+    if cmds.referenceQuery(ref, f=True, il=True):
```
antirotor, line 85: It seems maya is bitch:

```
RuntimeError: These flags are mutually exclusive (except for 'ea/editAttrs' and 'en/editNodes' which may be used together): 'f/filename', 'rfn/referenceNode', 'inr/isNodeReferenced', 'n/nodes', 'es/editStrings', 'en/editNodes', and 'ea/editAttrs'. 
```

We need to drop the `f` flag here.

```suggestion
    if cmds.referenceQuery(ref, il=True):
```

antirotor (changes requested):
It doesn't work for file references (like alembic), but with simple fix it should be ok.

Also it seems that submission doesn't use file with imported references:

image

it should point to `cg_street_workfileSetdress_v006_imported.ma` that is correctly integrated.

openpype/hosts/maya/plugins/publish/extract_import_reference.py
```diff
@@ -0,0 +1,156 @@
...
+                tmp_file_name
+            ]
+            self.log.info("Using temp file: {}".format(tmp.name))
+            try:
+                tmp.write(script)
+                tmp.close()
+                run_subprocess(subprocess_args)
+            except Exception:
+                self.log.error("Import reference failed", exc_info=True)
+                raise
```
BigRoy, line 121: Why are we opening the file with `+` mode? Why is that required?

I feel this would make for more readable code (note that this doesn't capture the script file writing in try..except like before). What do you think?
```suggestion
            tmp_script_path = os.path.join(tmp_dir_name, "import_ref.py")
            self.log.info("Using script file: {}".format(tmp_script_path))
            with open(tmp_script_path, "wt") as tmp:
                tmp.write(script)
                
            try:
                run_subprocess([mayapy_exe, tmp_script_path])
            except Exception:
                self.log.error("Import reference failed", exc_info=True)
                raise
```

openpype/modules/deadline/abstract_submit_deadline.py
```diff
@@ -516,15 +517,17 @@ def from_published_scene(self, replace_in_path=True):
...
         if workfile_instance is None:
             return
 
         # determine published path from Anatomy.
         template_data = workfile_instance.data.get("anatomyData")
-        rep = workfile_instance.data.get("representations")[0]
+        if self.import_reference:
+            rep = workfile_instance.data.get("representations")[1]
+        else:
+            rep = workfile_instance.data.get("representations")[0]
```
BigRoy, line 530: This seems oddly hardcoded - looking at the logic [here](https://github.com/ynput/OpenPype/pull/3875/files#diff-2ccaafdd3a3dd7b8cf293860eb9bf6551427f5b38cdb63fe37a8a2101a1d5674R141-R142) I'm pretty sure we are not _that_ sure the representations will be second one?

moonyuet: > When the extractor publishing the workfile, it will publish two workfiles, one is the scene with reference, the other is the scene with imported references(with the name e.g. {scene_name}_imported.ma). I set up this condition to take the second scene in order to match the submitted scene with imported reference. If not doing this, the rendering in deadline would error out.

BigRoy: I understand what the if statement is trying to do. I'm merely saying that purely by looking at the code it seems quite an implicit expectation that the first will be the non-imported and the second will be the imported version. Somehow I'd have expected we should somehow be able to do a more explicit check like:
```python
representations = workfile_instance.data["representations"]
rep = next(rep for rep in representations if rep["outputName"] == "imported")
```
Consider the above pseudocode. :) 

Additionally it seems odd we're doing `workfile_instance.data.get` since we're explicitly trying to index it with `[0]` and `[1]` making the assumption that the key will actually exist. Thus I'd recommend to at least change it to:
```suggestion
        if self.import_reference:
            rep = workfile_instance.data["representations"][1]
        else:
            rep = workfile_instance.data["representations"][0]
```

antirotor:
Argh... We still need to tweak something. There is a logic that will switch the scene to the published one here:

https://github.com/ynput/OpenPype/blob/5dfd92d303f1ab39820de1978e6de6b936f7c275/openpype/modules/deadline/abstract_submit_deadline.py#L498-L586

I think it will replace the scene with imported references for the published one.

PR #3877, opened 2022-09-16
Publisher: Nice checkbox visible in Python 2

iLLiCiTiT:
## Brief description
NiceCheckbox widget should work in Python 2 environment.

## Description
NiceCheckbox is expecting auto-conversions to float which does not happen in Python 2 so it's width is `0` and the checker does not move. By converting numbers to float at all places where it's expected should be the issue resolved.

## Additional information
I don't have any Python 2 application available so let me know if something does not work.

## Testing notes:
1. Open Publisher in DCC with Python 2 (e.g. Houdini 18.5)
2. Checkboxes should be visible and should behave right way on changing check state

Resolves https://github.com/pypeclub/OpenPype/issues/3782

### DIFF
--- a/openpype/widgets/nice_checkbox.py
+++ b/openpype/widgets/nice_checkbox.py
@@ -111,14 +111,14 @@ class NiceCheckbox(QtWidgets.QFrame):
         return QtCore.QSize(width, height)
 
     def get_width_hint_by_height(self, height):
-        return (
-            height / self._base_size.height()
-        ) * self._base_size.width()
+        return int((
+            float(height) / self._base_size.height()
+        ) * self._base_size.width())
 
     def get_height_hint_by_width(self, width):
-        return (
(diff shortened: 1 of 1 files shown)


BigRoy:
Would this also have been resolved if at the top of `nice_checkpox.py` we would have done this `from __future__ import division`?
It'll make the Py2 division behavior similar to Py3.

iLLiCiTiT:
> Might be a nicer way to fix this particular one since it will keep things clean for Py3-only in the future?

To be honest I don't like futures module. Once it's imported it does "some magic" and everyone is from that moment scared to remove it. Even if it's not Python 2 related. Convert to right types is maybe less nice for Python 3 but more explicit.

antirotor (approved):
rendering nice in Houdini 18, thanks.

PR #3878, opened 2022-09-17
Maya: Render Settings remove double setting of additional attributes for Arnold

BigRoy:
## Brief description

I'm noticing a lot of "duplicated" code and logic that contradicts itself in separate areas regarding Maya + Render Settings. Specifically:
- Sometimes checking the project settings and sometimes using a hardcoded default
- `lib_rendersettings` is totally decoupled from the validating of the rendersettings. This should be unified where possible.

The issue is relatively widespread so I'm not sure how to best clean this without going into a big refactor of logic. As such, I'll try to make tiny PRs for each individual problem where I can first.

## Description

Remove double setting of additional attributes for Arnold renderer.

## Testing notes:

1. Create render settings for arnold (with the "set render settings" enabled in settings)

### DIFF
--- a/openpype/hosts/maya/api/lib_rendersettings.py
+++ b/openpype/hosts/maya/api/lib_rendersettings.py
@@ -133,20 +133,7 @@ class RenderSettings(object):
 
         cmds.setAttr(
             "defaultArnoldDriver.mergeAOVs", multi_exr)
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
Works normally. 
L.

PR #3879, opened 2022-09-17
Maya: RenderSettings set default image format for V-Ray+Redshift to exr

BigRoy:
## Brief description

 Set default image format for V-Ray+Redshift to `exr` instead of `png` and `iff` in OP settings.

## Testing notes:

1. Test setting default rendersettings with V-Ray and Redshfit.

### DIFF
--- a/openpype/settings/defaults/project_settings/maya.json
+++ b/openpype/settings/defaults/project_settings/maya.json
@@ -49,7 +49,7 @@
         "vray_renderer": {
             "image_prefix": "maya/<scene>/<Layer>/<Layer>",
             "engine": "1",
-            "image_format": "png",
+            "image_format": "exr",
             "aov_list": [],
             "additional_options": []
         },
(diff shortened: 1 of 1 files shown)


Allan-I:
Off the top of my head I don't remember precisely but I think Redshift required a list index to be passed as opposed to a file format string.  But I could be wrong

BigRoy:
> Off the top of my head I don't remember precisely but I think Redshift required a list index to be passed as opposed to a file format string. But I could be wrong

That's good to know, but that logic hasn't changed. [It's still doing that conversion from settings extension name to index](https://github.com/pypeclub/OpenPype/blob/a2fbb472b4f85ab9ac996d73b34dc8efcfd0b4c8/openpype/hosts/maya/api/lib_rendersettings.py#L162-L163).

This only touches the default setting.

Allan-I:
> > Off the top of my head I don't remember precisely but I think Redshift required a list index to be passed as opposed to a file format string. But I could be wrong
> 
> That's good to know, but that logic hasn't changed. [It's still doing that conversion from settings extension name to index](https://github.com/pypeclub/OpenPype/blob/a2fbb472b4f85ab9ac996d73b34dc8efcfd0b4c8/openpype/hosts/maya/api/lib_rendersettings.py#L162-L163).
> 
> This only touches the default setting.

Oh damn yeah I'd forgotten writing that, yes sir that works just fine

Allan-I (approved):
LGTM!

PR #3884, opened 2022-09-19
changed default image prefix to comply with multilayer exr Maya Arnold render settings

m-u-r-p-h-y:
our defaults for Arnold renderer have conflict of Multilayer exr and `<RenderPass>` image token

image
 

image

### DIFF
--- a/openpype/settings/defaults/project_settings/maya.json
+++ b/openpype/settings/defaults/project_settings/maya.json
@@ -39,7 +39,7 @@
         "aov_separator": "underscore",
         "reset_current_frame": false,
         "arnold_renderer": {
-            "image_prefix": "maya/<Scene>/<RenderLayer>/<RenderLayer>",
+            "image_prefix": "maya/<Scene>/<RenderLayer>/<RenderLayer>{aov_separator}<RenderPass>",
             "image_format": "exr",
             "multilayer_exr": true,
             "tiled": true,
--- a/openpype/settings/defaults/project_settings/maya.json
+++ b/openpype/settings/defaults/project_settings/maya.json
@@ -39,7 +39,7 @@
         "aov_separator": "underscore",
         "reset_current_frame": false,
         "arnold_renderer": {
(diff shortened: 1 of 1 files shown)


openpype/settings/defaults/project_settings/maya.json
```diff
@@ -39,7 +39,7 @@
         "aov_separator": "underscore",
         "reset_current_frame": false,
         "arnold_renderer": {
-            "image_prefix": "maya/<Scene>/<RenderLayer>/<RenderLayer>_<RenderPass>",
+            "image_prefix": "maya/<Scene>/<RenderLayer>/<RenderLayer>",
```
BigRoy, line 42:
```suggestion
            "image_prefix": "maya/<Scene>/<RenderLayer>/<RenderLayer>{aov_separator}<RenderPass>",
```

I believe if the setting is like this it'll actually work.

It'll split off everything after the `{aov_separator}` when merge AOVs is enabled and thus should have no separate renderpasses.

m-u-r-p-h-y:
it is actually not working with the

BigRoy:
Actually, changing that value in settings wouldn't make it pass that validator either. The used prefix is currently hardcoded, see [here](https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/maya/plugins/publish/validate_rendersettings.py#L120).

Anyway, I'd be happy to look into this but feel merging the other render setting tweaks first would likely ease the process. That is: 
#3880 and https://github.com/pypeclub/OpenPype/pull/3873

LiborBatek:
Hi guys, seems like it hasnt been updated yet? Tried to test it out but behaves still same...

@m-u-r-p-h-y  did u already fix the preset or?

BigRoy:
As @m-u-r-p-h-y mentioned - the commit I suggested doesn't actually work. Let me look into it with this PR: https://github.com/pypeclub/OpenPype/pull/3880#issuecomment-1251931201

BigRoy:
Should have a fix for this now implemented in #3880

BigRoy:
I suspect we can close this in favor of #3880

PR #3887, opened 2022-09-19
Fix kitsu openpype sync without project resolution

ClementHector:
## Brief description
In the case where the resolution of the project has been removed by a soft user that the resolution does not contain x, the update of openpype by zou crashes.

## Description
In case the character 'x' is not in the resolution of kitsu, this information is not updated.

### DIFF
--- a/openpype/modules/kitsu/utils/update_op_with_zou.py
+++ b/openpype/modules/kitsu/utils/update_op_with_zou.py
@@ -18,7 +18,7 @@ from openpype.client import (
     create_project,
 )
 from openpype.pipeline import AvalonMongoDB
-from openpype.settings import get_project_settings, get_anatomy_settings
+from openpype.settings import get_project_settings
 from openpype.modules.kitsu.utils.credentials import validate_credentials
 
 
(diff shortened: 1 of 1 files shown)


openpype/modules/kitsu/utils/update_op_with_zou.py
```diff
@@ -268,12 +267,15 @@ def write_project_to_op(project: dict, dbcon: AvalonMongoDB) -> UpdateOne:
...
-            "resolutionWidth": int(project["resolution"].split("x")[0]),
-            "resolutionHeight": int(project["resolution"].split("x")[1]),
             "zou_id": project["id"],
         }
     )
 
+    proj_res = project["resolution"]
+    if "x" in proj_res:
+        project_data['resolutionWidth'] = int(proj_res.split("x")[0])
+        project_data['resolutionHeight'] = int(proj_res.split("x")[1])
```
BigRoy, line 277: Would it make sense to at least a log a warning or debug message that resolution data is incorrect?

iLLiCiTiT: I agree and maybe set resolution to some default (`0`, `0`?).

ClementHector: - Should I add a logger instead of all print?
- I would prefer not to set the resolution to zero because in my case the resolution was deleted by mistake, it would be problematic in the software to have a resolution set to 0x0 during artist work.

mkolar: Log would be good here, and I agree that setting to 0 is not a good way here. It would introduce more mistakes potentially

openpype/modules/kitsu/utils/update_op_with_zou.py
```diff
@@ -268,12 +270,18 @@ def write_project_to_op(project: dict, dbcon: AvalonMongoDB) -> UpdateOne:
...
-            "resolutionWidth": int(project["resolution"].split("x")[0]),
-            "resolutionHeight": int(project["resolution"].split("x")[1]),
             "zou_id": project["id"],
         }
     )
 
+    match_res = re.match(r"(\d+)x(\d+)", project["resolution"])
+    if match_res:
+        project_data['resolutionWidth'] = match_res.group(1)
+        project_data['resolutionHeight'] = match_res.group(2)
```
BigRoy, line 280:
```suggestion
        project_data['resolutionWidth'] = int(match_res.group(1))
        project_data['resolutionHeight'] = int(match_res.group(2))
```
Previously these were integers.

BigRoy (approved):
**Haven't tested the code** - but code changes look good to me.

mkolar (approved):
Agreed looking good

PR #3888, opened 2022-09-20
Publisher: Files Drag n Drop cleanup

iLLiCiTiT:
## Brief description
Cleanup code of files widgets used in publisher.

## Description
Handle moving around of items in files widget. Stack widgets in files widget to handle their size hints. Proper showing of empty widget on moving items around. Hide Drag'n'Drop label if user can't drop any.

## Additional info
Files widget changed a look when have no items in. The view is painted underneath so the beackground is dark grey.

### Screenshot
image

## Testing notes:
1. Open TrayPublisher where files widget is available
2. Try to play with it to validate if reported issues that should be resolved do not happen.

Resolves https://github.com/pypeclub/OpenPype/issues/3837
Resolves https://github.com/pypeclub/OpenPype/issues/3836

### DIFF
--- a/openpype/widgets/attribute_defs/files_widget.py
+++ b/openpype/widgets/attribute_defs/files_widget.py
@@ -138,11 +138,13 @@ class DropEmpty(QtWidgets.QWidget):
             allowed_items = [item + "s" for item in allowed_items]
 
         if not allowed_items:
+            self._drop_label_widget.setVisible(False)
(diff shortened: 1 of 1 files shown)


antirotor:
It fixed those issues but there is till some confusion here:

tray_publisher_items

I would expect the file dragged to the upper box to be added there, not to replace existing one?

antirotor (approved):

PR #3889, opened 2022-09-20
Fix - hardlink issue in integrate Hero

kalisp:
## Brief description
Disk must be NTFS format or it will throw "[WinError 1", which matches to EINVAL. Still raise different errors.

## Description
Occured on disk formatted by `exFAT`.

## Testing notes:
1. publish Hero version


LiborBatek (approved):
Tested it out and worked ok!!

iLLiCiTiT:
We should maybe normalize destination path (as first step) to fix possible issues with double slashes.

LiborBatek:
Regarding those doubled slashes, small note it seems it doesnt caused any issues in the end. At least from the maya host user perspective....

PR #3892, opened 2022-09-21
Fusion: New Publisher

BigRoy:
## Brief description

This converts the old publishing system to the new one. It implements Fusion as a new host addon.

## Description

- Create button removed in OpenPype menu in favor of the new Publisher
- Draft refactor validations to raise `PublishValidationError`
- Implement Creator for New Publisher
- Implement Fusion as Host addon

## Additional info

I have close to no knowledge of the new Publisher so there's probably a lot broken bits and pieces. Admittedly the API for Creator's and the UX for the new publisher can still use some work before doing this full code refactor and switching to the new publisher.

## Testing notes:

1. Launch Fusion
2. Create a saver instance
3. Publish


openpype/hosts/fusion/plugins/create/create_workfile.py
```diff
@@ -0,0 +1,129 @@
...
+        if not comp:
+            self.log.error("Unable to find current comp")
+            return
+
+        # TODO: Is this really necessary?
+        # Force kill any existing "workfile" instances
+        for instance in self.create_context.instances:
+            if instance.family == self.family:
+                self.log.debug(f"Removing instance: {instance}")
+                self._remove_instance_from_context(instance)
```
BigRoy, line 83: I'm still not entirely sure if the AutoCreator has to implement both a `create()` and a `collect_instances()`. It feels like both are trying to do the same thing - which is automatically give me the instance from the scene.

This confusing mixture in responsibility makes it so that I still don't understand what should be in `collect_instances` and what should be in `create` and why?

In my logic here I'm force killing it - in other Workfile creators or the documentation example it tries to update it in-place. Why would we care to do one over the other?

openpype/hosts/fusion/plugins/create/create_workfile.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def get_icon(self):
+        return qtawesome.icon("fa.file-o", color="white")
+
+    def _imprint(self, comp, data):
+
+        # TODO: Should this keys persist or not? I'd prefer not
+        # Do not persist the current context for the Workfile
+        for key in ["variant", "subset", "asset", "task"]:
+            data.pop(key, None)
```
BigRoy, line 122: The After Effects workfile creator "overrides" asset, task and subset with the current context if it differs from what's stored in the scene. If that's the case - why would we store it to begin with?

PR #3897, opened 2022-09-22
Publisher: Enhancement proposals

iLLiCiTiT:
## Brief description
Proposing possible enhancements related to new publisher.

## Description
Instance object has `lifetime_data` where can creator add objects from scene or other objects not related to instance data, during publishing can be accessed using `lifetimeData` on instance. Renamed `INewPublisher` to `IHostPublish` -> match other interfaces and get rid of "new publisher" (`INewPublisher` kept for backwards compatibility, but we could remove it now?). Creators have method `apply_settings` to simplify initialization.

## Additional information
The `lifetime_data` were requested from Fusion implementation where node name is not enough to be able identify the node. I would say it's reasonable enough to provide this option too.

### DIFF
--- a/openpype/host/__init__.py
+++ b/openpype/host/__init__.py
@@ -5,6 +5,7 @@ from .host import (
 from .interfaces import (
     IWorkfileHost,
     ILoadHost,
+    IPublishHost,
     INewPublisher,
 )
 
@@ -16,6 +17,7 @@ __all__ = (
 
     "IWorkfileHost",
     "ILoadHost",
+    "IPublishHost",
     "INewPublisher",
 
     "HostDirmap",
--- a/openpype/host/interfaces.py
+++ b/openpype/host/interfaces.py
@@ -282,7 +282,7 @@ class IWorkfileHost:
         return self.workfile_has_unsaved_changes()
 
 
-class INewPublisher:
+class IPublishHost:
     """Functions related to new creation system in new publisher.
 
     New publisher is not storing information only about each created instance
@@ -306,7 +306,7 @@ class INewPublisher:
                 workflow.
         """
 
-        if isinstance(host, INewPublisher):
+        if isinstance(host, IPublishHost):
             return []
 
         required = [
@@ -330,7 +330,7 @@ class INewPublisher:
             MissingMethodsError: If there are missing methods on host
                 implementation.
         """
-        missing = INewPublisher.get_missing_publish_methods(host)
+        missing = IPublishHost.get_missing_publish_methods(host)
         if missing:
             raise MissingMethodsError(host, missing)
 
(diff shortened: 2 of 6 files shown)


BigRoy:
> Creators have method apply_settings to simplify initialization.

Could you explain the idea behind this?

> Instance object has lifetime_data where can creator add objects from scene or other objects not related to instance data, during publishing can be accessed using lifetimeData on instance.

I'm not too sure about the name `lifetimeData`. It feels like it still needs a lot of explanation as to what to do with it. Maybe we should name it something that sounds like the opposite of "persistent data" (basically something that doesn't save with the scene - does `lifetimeData` do that sufficiently?)
Maybe `transientData`?

I also feel that storing it like that on the actual pyblish instance is a bit off since makes it feel less tightly coupled somehow even though it's crucial data to get back to the node that generated the instance. Couldn't we maybe store it into pyblish instance like:
```python
instance.data = {}
instance.data.update(CreatedInstance.transientData)
instance.data.update(copy.deepcopy(CreatedInstance.data))
```

So that e.g. `instance.data["tool"]` would be possible.

iLLiCiTiT:
> Could you explain the idea behind this?

If you want apply settings you don't have to override `__init__` method and hit potential future changes, but just override `apply_settings` method where you don't have to call `super`.

> I'm not too sure about the name lifetimeData. 

I'm ok with transient data.

> I also feel that storing it like that on the actual pyblish instance is a bit off since makes it feel less tightly coupled somehow even though it's crucial data to get back to the node that generated the instance. Couldn't we maybe store it into pyblish instance like:

So you have to explain that you must not use keys that are same as in instance data because you would loose the alues and those keys would be automatically propagated to pyblish instance data. I would say it's better to have it under specific key so you know you can safely access it from that key without any worries.

jakubjezek001:
I wonder if there is really not a way you could get node object by node name attribute. Something like:

```
node_name = "NodeName01"
compList = fu.GetCompList()
for index, node in sorted(compList.items()):
  	if node.GetAttrs()["TOOLS_Name"] == node_name:
		return node
```

BigRoy:
@jakubjezek001 I actually implemented that currently - see [here](https://github.com/pypeclub/OpenPype/pull/3892/files#diff-366db2d06eed704cb122f69f5c605fac2ff7a51c2fc93a6632bfc86a4ac15799R115-R124).

However, it introduces the problem with the user having multiple comps open at the same time and it's likely that they might activate a different comp tab as they are running the Publish UI. It'd be hard to know which of the open comps that particular tool belonged to - whereas if we'd have the transient data we know exactly because we'd get the comp from the tool instead, using `tool.Comp()` or alike.

BigRoy:
This PR seems functional with regards to `transient_data`. This is what it allowed me to do as a change in Fusion logic, [example commit](https://github.com/BigRoy/OpenPype/commit/bdfe2414583480c06f0ee35113b0deea3fced1c6). Pretty good!

PR #3900, opened 2022-09-23
General: Fix links query on hero version

iLLiCiTiT:
## Brief description
Fixed function to find links which crashed on hero versions.

## Description
Don't lookup for hero versions in `get_linked_representation_id` because they don't have links. Made safer access to `"data"` key on found items to avoid possible crashes on items without links.

## Testing notes:
1. Run Add to Site action on hero version
2. It should not crash on finding linked representations

### DIFF
--- a/openpype/client/entity_links.py
+++ b/openpype/client/entity_links.py
@@ -132,7 +132,9 @@ def get_linked_representation_id(
 
     match = {
         "_id": version_id,
-        "type": {"$in": ["version", "hero_version"]}
+        # Links are not stored to hero versions at this moment so filter
+        #   is limited to just versions
+        "type": "version"
     }
 
     graph_lookup = {
--- a/openpype/client/entity_links.py
+++ b/openpype/client/entity_links.py
@@ -189,7 +189,7 @@ def _process_referenced_pipeline_result(result, link_type):
     referenced_version_ids = set()
     correctly_linked_ids = set()
     for item in result:
-        input_links = item["data"].get("inputLinks")
+        input_links = item.get("data", {}).get("inputLinks")
         if not input_links:
             continue
 
@@ -205,7 +205,7 @@ def _process_referenced_pipeline_result(result, link_type):
             continue
 
         for output in sorted(outputs_recursive, key=lambda o: o["depth"]):
-            output_links = output["data"].get("inputLinks")
+            output_links = output.get("data", {}).get("inputLinks")
             if not output_links:
                 continue
 


kalisp:
Right mouse click on representation of Hero version in bottom right in Loader throws this:
Screenshot 2022-09-23 205406

kalisp (changes requested):
See last comment.

openpype/client/entity_links.py
```diff
@@ -127,12 +128,20 @@ def get_linked_representation_id(
...
+
     if max_depth is None:
         max_depth = 0
 
     match = {
         "_id": version_id,
-        "type": {"$in": ["version", "hero_version"]}
+        # Links are not stored to hero versions at this moment so filter
+        #   is limited to just versions
+        "type": "version"
```
BigRoy, line 144: Actually - are we sure? The input link collection I believe is retrieved from containers in the scene - those I suspect would have the representation id of the hero version. As such by doing this "fix" we'd never be able to list those inputs even though they are stored?

iLLiCiTiT: Links are added only to [versioned version documents](https://github.com/pypeclub/OpenPype/blob/develop/openpype/plugins/publish/integrate_inputlinks.py#L45). and [hero version document](https://github.com/pypeclub/OpenPype/blob/develop/openpype/plugins/publish/integrate_hero_version.py#L195) don't have `"data"` at all.

But because hero versions are based on versioned documents we can reuse links from there.

BigRoy: Yes. Getting inputs FROM the hero version is then fixed. But any hero version IN the inputs is not? That hero version's inputs are not collected? (But might not be relevant for the UI we have no since it only shows direct inputs instead of a further depth.)

That does raise the question too that when we collect the inputs on publish would we want to resolve the hero version to the exact version it actually was at the time or do we want to just keep the link to the hero version? If we'd 'bake' it down to the exact version on collecting inputs the issue I mentioned would be gone plus you get the benefit of actually being able to see what version was used at the time in input link. Maybe we could tag it extra as "hero" so that the does know it originally was a hero link but at the time of publishing that was referring to v038.

@mkolar thoughts?

Tilix4:
Hi, is there anything to do for helping this fix to land?

kalisp (approved):

openpype/client/entity_links.py
kalisp: Should we spin this up as a new issue?

PR #3901, opened 2022-09-23
Flame: loading multilayer exr to batch/reel is working

jakubjezek001:
## Brief description
Loading multilayer EXR is now available.

## Description
We are now able to load multilayer exr into Batch, so users can access individual layers for their compositing workflows. 

## Known limits
In case where on a loaded openclip file next version of exr will change set of layers or rename some, we are only supporiting last available set of layers and all previous versions at the openclip which are not having cross matched versions will be hidden. Solution for this would be generate black frame placeholders for those which will be missing, but it will have also its disadvantages. Lets say during several versions will be set of layers changed multiply, and the column of versions will be very long and messy, because there is no way to find out, that layer stack is empty black frame. 

## example of batch exr loader
image


openpype/hosts/flame/api/plugin.py
```diff
@@ -783,52 +800,81 @@ def _update_open_clip(self):
...
-            out_feeds.append(tmp_xml_feed)
-
-            self.log.info(
-                "Appending new feed: {}".format(
-                    self.feed_version_name))
-            feed_added = True
-
-        if feed_added:
+        # loop tmp tracks
+        updated_any = []
```
iLLiCiTiT, line 804: This should be boolean

openpype/hosts/flame/api/plugin.py
```diff
@@ -783,52 +800,81 @@ def _update_open_clip(self):
...
+                self._clear_handler(tmp_xml_feed)
+
+                # update fps from MediaInfoFile class
+                if self.fps:
+                    tmp_feed_fps_obj = tmp_xml_feed.find(
+                        "startTimecode/rate")
+                    tmp_feed_fps_obj.text = str(self.fps)
+
+                # update start_frame from MediaInfoFile class
+                if self.start_frame:
```
iLLiCiTiT, line 840:
```suggestion
                if self.start_frame is not None:
```

openpype/hosts/flame/api/plugin.py
```diff
@@ -783,52 +800,81 @@ def _update_open_clip(self):
...
+                new_path_obj = tmp_xml_feed.find(
+                    "spans/span/path")
+                new_path = new_path_obj.text
+
+                # check if feed path already exists in track's feeds
+                if (
+                    out_track_element is not None
+                    and self._feed_exists(out_track_element, new_path)
+                ):
+                    continue
```
BigRoy, line 827: Wouldn't this also skip updating any potential changes to FPS, Start Frame, End Frame, etc.?

jakubjezek001: Good catch, thanks!

jakubjezek001: eh, actually there is no need for continuing if the condition is not met. Problem is that if the same path is already in some previous versions, there is no need for updating it to new version with the same representation data. I thing this contition is working as it should.

iLLiCiTiT (approved):

PR #3906, opened 2022-09-23
Add version to shortcut name

antirotor:
## Enhancement

This little fix is adding verison string to the shortcut name created on Windows when installed to differentiate between multiple installed versions.

### DIFF
--- a/inno_setup.iss
+++ b/inno_setup.iss
@@ -48,8 +48,8 @@ Source: "build\{#build}\*"; DestDir: "{app}"; Flags: ignoreversion recursesubdir
 ; NOTE: Don't use "Flags: ignoreversion" on any shared system files
 
 [Icons]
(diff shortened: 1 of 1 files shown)


iLLiCiTiT (approved):

BigRoy:
What does this version name look like? Is it just the `major.minor` version? or the full identifier including patch and nightly suffix? Since you'd only ever be installing separate `major.minor` versions right?

antirotor:
> What does this version name look like? Is it just the `major.minor` version? or the full identifier including patch and nightly suffix? Since you'd only ever be installing separate `major.minor` versions right?

`$AppVer` is full version. Not sure if I can get easily from Inno Setup just `major.minor`... Maybe I could fix it with Scripted Constant :) Time to refresh my Pascal skills?

PR #3907, opened 2022-09-23
General: Remove deprecated functions from lib

iLLiCiTiT:
## Brief description
Removed functions marked as deprecated from `openpype.lib`.

## Description
These functions were marked as deprecated for a long time and should be safe to remove them.

## Testing notes:
1. Everything should work as before


jakubjezek001 (approved):
seems to be working as it should

PR #3908, opened 2022-09-23
Photoshop: create single frame image in Ftrack as review

kalisp:
## Brief description
Previously PS was creating .mov even for single frame review. This PR allows to create and upload single image instead.

## Description

## Additional info
Still in progress as Play button under Versions metadata is finicky, Play button on thumbnail in version list seems to work fine.

## Testing notes:
Screenshot 2022-09-20 165806

1. Double check jpg options for Photoshop ExtractReview in Setting
2. Publish workfile that has single or no `image` instance.
3. Check that `publish/review` folder contains high res review
4. Check in Ftrack by clicking on Thumbnail in Versions list that high res image pops up

### DIFF
--- a/openpype/modules/ftrack/plugins/publish/integrate_ftrack_instances.py
+++ b/openpype/modules/ftrack/plugins/publish/integrate_ftrack_instances.py
@@ -150,38 +150,38 @@ class IntegrateFtrackInstance(pyblish.api.InstancePlugin):
         # TODO what if there is multiple thumbnails?
         first_thumbnail_component = None
         first_thumbnail_component_repre = None
-        for repre in thumbnail_representations:
-            if review_representations and not has_movie_review:
-                break
-            repre_path = self._get_repre_path(instance, repre, False)
-            if not repre_path:
-                self.log.warning(
-                    "Published path is not set and source was removed."
+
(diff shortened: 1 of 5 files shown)


openpype/modules/ftrack/plugins/publish/integrate_ftrack_instances.py
```diff
@@ -147,6 +151,8 @@ def process(self, instance):
         first_thumbnail_component = None
         first_thumbnail_component_repre = None
         for repre in thumbnail_representations:
+            if review_representations and not has_movie_review:
+                break
```
BigRoy, line 155: Both these variables don't seem to be related to the `repre` we are currently iterating and thus it feels off that we're checking these variables inside the for loop instead of prior to it? (It looks confusing because it reads as if related to the current repre being iterated)

kalisp:
There is still discrepancy between clicking on thumbnail in Version list and Side panel 'Play' https://discord.com/channels/@me/928645495822368799/1026857661368651856 which is being solved with Ftrack Support.

Customer is fine going ahead with this for now, so we can go ahead and test it

64qam (approved):

Issue #3865, opened 2022-09-15
Move ImageIO settings from anatomy to hosts


iLLiCiTiT:
## Description
ImageIO settings are changing a lot which breaks project anatomy because is not versioned (as other settings). Also each host need a different approach how image io is defined thus it would be unmaintainable to keep all hosts in anatomy and foolish to try do it "universaly".

## Suggested solution
This is painfull change which will require to keep anatomy schemas same as were for at least one release to not loose any values when testing new version of OpenPype. This change require to move the values on update from anatomy to project settings (it is not possible to have "empty" settings).
So we should create new schemas in host's project settings which are exactly the same as the one in anatomy.

### On update
For each project should be values copy pasted from project anatomy to project settings

BigRoy:
Was this resolved with #3959 already @jakubjezek001 @iLLiCiTiT ?

iLLiCiTiT:
Yes

PR #3909, opened 2022-09-23
Settings: Move imageio from project anatomy to project settings

BigRoy:
## Brief description

This implements #3865 by moving `imageio` settings per host from project anatomy to project settings.

Before:
```
project_anatomy/imageio/{app}
```

After:
```
project_settings/{app}/imageio
```

## Description

Currently untested - no backwards compatible or fallbacks implemented. Basically your old project anatomy customized settings will be lost. 🔥 - So as #3865 we'll need to look into how we'll want to transfer those settings over _automagically_ 🪄 .

## Additional info

Note that each `imageio` group has now set `is_group` to True as such changing one setting in the imageio for a host will now override that full imageio block of settings for the project. It's up for discussion whether we want that - if we want to change around we'll need to add some labels to some of the settings (required for `is_group` to be False) but that should be fine.

It could be that some documentation might also need to be updated along. 
- Even though admittedly the old [Color Managements docs are very slim](https://openpype.io/docs/admin_settings_project_anatomy#colour-management-and-formats) - it's just the header - ha! 
- But potentially a comment on and an update to description [of this Nuke Color Management tutorial video](https://openpype.io/docs/artist_hosts_nuke_tut/#nuke-color-management) would be worthwhile too to make users aware the settings have moved over in the future. 
- Hiero documentation will need to be updated around [apply colorspace project](https://openpype.io/docs/artist_hosts_hiero/#apply-colorspace-project)

#### Copying ImageIO settings from Anatomy to Project Settings

Any potential backwards compatible "updating" would only need to basically 'copy the settings values' from `project_anatomy/imageio/{app}` on top of `project_settings/{app}/imageio` to transfer the exact settings. So in essence it is:

```python
for app in anatomy["imageio"]:
    value = app.value
    project_settings[app.key]["imageio"].set(value)
    project_settings.save()
```

---

I'm not awaiting the moment this PR gets 'assigned' to me and I get all the blame in the future. 🗡️ 

## Testing notes:

1. 🔥 🚒 Get your fire extinguishers ready
2. Test color management related features across the hosts

- [x] Flame 
- [x] (and Flare?)
- [ ] Maya
- [x] Nuke
- [x] Hiero
- [x] (Did Resolve have any? I couldn't find it. If it didn't, let's check this box and continue our ignorance.)


schema/config-2.0.json
```diff
@@ -23,9 +23,6 @@
         "roots": {
             "type": "object"
         },
-        "imageio": {
-            "type": "object"
-        },
```
BigRoy, line 28: Can we just remove this? Are old projects ever checked against this schema and invalidated if it finds an entry that isn't in this config?

It's not a required key so that helps, but still.

openpype/settings/lib.py
```diff
@@ -316,22 +316,6 @@ def _system_settings_backwards_compatible_conversion(studio_overrides):
...
-            return
-        value = value[key]
-
-    for item in value:
-        for node in item.get("knobs") or []:
-            if "type" in node:
-                break
-            node["type"] = "__legacy__"
-
-
```
BigRoy, line 334: Had no use with this move of the settings as far as I know - _so burned it down to the floor._

iLLiCiTiT:
I would keep schemas for `openpype/settings/defaults/project_anatomy/imageio.json` so you can copy->paste values on update without loosing them on update.

BigRoy:
> I would keep schemas for `openpype/settings/defaults/project_anatomy/imageio.json` so you can copy->paste values on update without loosing them on update.

Since we are matching the settings 1-to-1 wouldn't it be much nicer if it'd automatically transfer the settings from Project Anatomy somehow? I feel from a User perspective that'd be much nicer than still having to "fix" each project manually (even though sure, you could copy/paste the values - but similarly we could offer a "click here and it's fixed button" by copying the data project anatomy in database to the settings?

I'm mostly worried about having both setting visible and the user misunderstanding that those in Project Anatomy essentially do nothing. We could put in a big deprecation label in the UI, but still. If the intend is for production to notice the least the settings _should_ be copied over thus it sounds like we should just make that happen.

iLLiCiTiT:
> I'm mostly worried about having both setting visible and the user misunderstanding that those in Project Anatomy essentially do nothing. We could put in a big deprecation label in the UI, but still. If the intend is for production to notice the least the settings should be copied over thus it sounds like we should just make that happen.

The problem is that you will loose all those data from anatomy on first save of project setting. So if you're "testing" new version in production you will actually break all projects on project settings save.

BigRoy:
**Question:**

Should we maybe swap these around so the settings are 
- `project_settings/imageio/{app}`

instead of

- `project_settings/{app}/imageio`?


The benefit I'd see from that is that we can just make one MASTER config setting that sets `OCIO` env var as a hook based on the config path set in `project_settings/imageio/config_path`. Potentially still allowing apps to have an explicit `project_settings/imageio/app/override_config_path` like setting?

Or is it very likely that the different applications should use different OCIO configs in a project.

What do we feel is the most sensible 'project setting'? @mkolar @maxpareschi

iLLiCiTiT:
That is ultimate goal but we can't do it at this moment. Nuke imageio has totally different settings then Maya or Flame or Hiero. We need this PR to move the settings from anatomy to project setting so we can possibly change their behavior from version to version with versioned settings which is not possible if they're in anatomy.

BigRoy:
Great - then I suspect this PR should be ok to start that transition.

openpype/hosts/nuke/api/lib.py
```diff
@@ -563,7 +563,7 @@ def get_node_path(path, padding=4):
 
 
 def get_nuke_imageio_settings():
-    return get_anatomy_settings(Context.project_name)["imageio"]["nuke"]
+    return get_project_settings(Context.project_name)["nuke"]["imageio"]
```
jakubjezek001, line 566: I believe there should be some backward compatibility check. Some clients might have Anatomy ImageIO set from 3.9 and they didn't change `__legacy__` type yet. We would need to go trough all projects and do this manually (impossible). 
For those projects it doesnt even make sense to copy paste it from Anatomy to `./nuke/imageio`.

I will do the changes in separate PR. Please could you return the `___legacy__` check (https://github.com/pypeclub/OpenPype/pull/3909#discussion_r979023257)

openpype/settings/lib.py
jakubjezek001: https://github.com/pypeclub/OpenPype/pull/3909#discussion_r992079184

jakubjezek001 (approved):
good for merge.. tested Nuke, Hiero. Resolve is not done anyway. Flame should be fine, too.

jakubjezek001:
> good for merge.. tested Nuke, Hiero. Resolve is not done anyway. Flame should be fine, too.

huh.. just realized that the `__legacy__` (https://github.com/pypeclub/OpenPype/pull/3909#discussion_r979023257) needs to be returned before the merge..

iLLiCiTiT (changes requested):
We need explicit backwards compatibility in project settings. @jakubjezek001 can you do required changes in new branch (sourcing from this) and create new PR?

jakubjezek001:
This PR is closing in favor of https://github.com/pypeclub/OpenPype/pull/3959

PR #3910, opened 2022-09-24
Publisher: Fix PublishIconButton drawing disabled icon with the color specified

BigRoy:
## Brief description

Fix PublishIconButton drawing disabled icon with the color specified.

## Description

Previously the color to draw in the `paint_image` logic was ignored when button was disabled because default color was applied to the disabled state. As such the disabled image always had a default "gray" state which QPushButtons generate.

Before:
afbeelding
(Only "reset" button is actually enabled here!)

After:
afbeelding

## Additional info

Note that the disabled color is a bit hardcoded like this and randomly chosen. I picked the direct color from `openpype/style/data.json` for `font-disabled` which was `#99A3B2` but that just didn't show well enough.

## Testing notes:

1. Check in different platforms, Python and Qt versions

### DIFF
--- a/openpype/style/data.json
+++ b/openpype/style/data.json
@@ -20,7 +20,7 @@
     "color": {
         "font": "#D3D8DE",
         "font-hover": "#F0F2F5",
-        "font-disabled": "#99A3B2",
+        "font-disabled": "#5b6779",
         "font-view-selection": "#ffffff",
         "font-view-hover": "#F0F2F5",
 
--- a/openpype/tools/publisher/widgets/widgets.py
+++ b/openpype/tools/publisher/widgets/widgets.py
@@ -125,28 +125,19 @@ class PublishIconBtn(IconButton):
     def __init__(self, pixmap_path, *args, **kwargs):
         super(PublishIconBtn, self).__init__(*args, **kwargs)
 
-        loaded_image = QtGui.QImage(pixmap_path)
+        icon = self.generate_icon(pixmap_path,
+                                  enabled_color=QtCore.Qt.white,
+                                  disabled_color=QtGui.QColor("#5b6779"))
+        self.setIcon(icon)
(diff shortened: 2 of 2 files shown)


iLLiCiTiT:
I like the color we can add it into styles data to `font-disabled`.

You can then use something like:
```
from openpype.style import get_objected_colors

...
        colors = get_objected_colors()
        icon = self.generate_icon(
            pixmap_path,
            enabled_color=colors["font"].get_qcolor(),
            disabled_color=colors["font-disabled"].get_qcolor())
```

mkolar (approved):
agreed,  the colour is nice. Lets' do it. Approving visualy. @iLLiCiTiT will approve and merge once the code is good

BigRoy:
I've made the changes.

@iLLiCiTiT @mkolar  I'm a bit worried that each init of a button in the GUI would then turn into a full read of the colors `data.json`. Especially because there's likely many buttons to be initialized. Should we maybe cache the `get_objected_colors` result? `openpype.style` does appear to have a cache but it doesn't look like the `data.json` result is actually cached too. So any call to `get_objected_colors` would result in the full `.json` read and conversion of colors. Should we cache that too?

iLLiCiTiT:
I'm not worried about it much. It takes at about 0.0009s. But if you're worried you can add it to `_Cache` :)

BigRoy:
> I'm not worried about it much. It takes at about 0.0009s. But if you're worried you can add it to _Cache :)

If you run it locally from your specific SSD or NVME disk maybe - yes. But if the build might be on a network or alike those numbers could differ greatly. I'm thinking now that this might actually also influence any "nice checkbox" or other buttons in the Settings UI? :) It could be reading the `data.json` for each button you initialize and thus be making 1000s of 0.0009s calls?

PR #3912, opened 2022-09-26
Maya: Add additional playblast options.

Allan-I:
## Brief description
Adds more options for playblasts to preferences/settings.

## Description
Adds the following as options in generating playblasts, matching viewport settings. 

- Use default material
- Wireframe on shaded
- X-ray
- X-ray Joints
- X-ray active component 

## Testing notes:
1. Change studio settings to select one of the available settings.
2. Create review camera.
3. Create review.
4. Publish.

Note:  Seems that some commits I'd had on this have been lost, going to investigate why that's the case.


openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_capture.json
```diff
@@ -348,6 +348,31 @@
...
+                        {
+                            "type": "boolean",
+                            "key": "jointXray",
+                            "label": "Toggle Joint X-Ray Mode"
+                        },
+                        {
+                            "type": "boolean",
+                            "key": "backfaceCulling",
+                            "label": "Enable Back Face Culling"
+                        },
```
BigRoy, line 375: Could we move this to above the label **Show** and then add a label just above it that says **Shading**?

That way it matches the Maya Viewport's menu name for Shading.
Then I'd actually labelize the settings the same way as Maya does too:
afbeelding

So that would be like this refactor, but then moved higher.
```suggestion
                        {
                            "type": "label",
                            "label": "<b>Shading</b>"
                        }, 
                        {
                            "type": "boolean",
                            "key": "udm",
                            "label": "Use Default Material"
                        },
                        {
                            "type": "boolean",
                            "key": "wos",
                            "label": "Wireframe On Shaded"
                        },
                        {
                            "type": "boolean",
                            "key": "xray",
                            "label": "X-Ray"
                        },
                        {
                            "type": "boolean",
                            "key": "jointXray",
                            "label": "X-Ray Joints"
                        },
                        {
                            "type": "boolean",
                            "key": "backfaceCulling",
                            "label": "Backface Culling"
                        },
```

Allan-I: You got it!

openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_capture.json
```diff
@@ -348,6 +348,35 @@
...
+                        }, 
+                        {
+                            "type": "boolean",
+                            "key": "udm",
+                            "label": "Use Default Material"
+                        },
+                        {
+                            "type": "boolean",
+                            "key": "wos",
+                            "label": "Wireframe On Shaded"
```
BigRoy, line 363:
```suggestion
                        {
                            "type": "boolean",
                            "key": "useDefaultMaterial",
                            "label": "Use Default Material"
                        },
                        {
                            "type": "boolean",
                            "key": "wireframeOnShaded",
                            "label": "Wireframe On Shaded"
```

BigRoy:
Changes look good to me but haven't tested the functionality.

Note that I think that when you save defaults with the Settings UI that it might re-order the [openpype/settings/defaults/project_settings/maya.json](https://github.com/pypeclub/OpenPype/pull/3912/files#diff-c6017fd21a1f9bf3c79271e6fe7f506dbaf5b5468724e9649e4ba15524146c86) file and thus I'd say to regenerate them now if they do - instead of us manually editing the file here. Since dicts are ordered in Py3 I suspect with the re-ordering of the attributes that likely these defaults would be ordered in the newly ordered way too.

By updating those defaults now it'd avoid having that re-ordering happen in a future PR.

openpype/settings/defaults/project_anatomy/attributes.json
```diff
@@ -19,8 +19,7 @@
         "blender/2-91",
         "harmony/20",
         "photoshop/2021",
-        "aftereffects/2021",
-        "unreal/4-26"
```
BigRoy, line 23: Seems unrelated to this PR?

openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_capture.json
```diff
@@ -221,6 +221,31 @@
                         {
                             "type": "splitter"
                         },
+                        {
+                            "type": "boolean",
+                            "key": "useDefaultMaterial",
+                            "label": "Use Default Material"
+                        },
```
BigRoy, line 228: I feel keeping the "Display" label above this will still make it feel more like the Maya viewport's menu and will help the user navigate the settings. Thoughts?

openpype/settings/defaults/project_anatomy/attributes.json
Allan-I: Ah, that's extra, will remove that, thanks man!

openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_capture.json
Allan-I: I agree!

antirotor:
@LiborBatek @m-u-r-p-h-y can you please check the state of this?

moonyuet:
checked the Openpype setting, the related attributes are there.
image
and also checked the playblast settings by applying the review instance. 
I only find out that the geometry has gone when the option of "Use Default Material" is on. It is supposed that the default material is applied on the mesh and display the mesh in a lambert material.
image

I tested with maya 2023

moonyuet (approved):

tokejepsen:
@Allan-I is this ready to review? It'll need some conflict resolving for a merge.

LiborBatek:
@tokejepsen  @Allan-I  is not working at YNPUT anymore so I guess you have to dig into it... you wont get any answer I guess

PR #3913, opened 2022-09-26
Style: Cache `get_objected_colors` function

BigRoy:
## Brief description

The `get_objected_colors` function is called very often - e.g. over 1500 times when opening the Settings UI - and did a full read of the style's color `data.json` and converted it to the color objects with an average of 0.0009s per call on my machine.

## Description

See original discussion in the comments [here](https://github.com/pypeclub/OpenPype/pull/3910#issuecomment-1257727856) where I've elaborated on the usage of the function in the Settings UI.

## Additional info

Do we want to return a deep copy to ensure any usage of the data doesn't influence the cache? - It does seem that with a deep copy the time taken for the function returns almost to the same as without the caching. (However that is only if the data.json is read from my local SSD of course)

## Testing notes:

1. Test whether colors are still correct in the UIs.

### DIFF
--- a/openpype/style/__init__.py
+++ b/openpype/style/__init__.py
@@ -1,4 +1,5 @@
 import os
+import copy
 import json
 import collections
 import six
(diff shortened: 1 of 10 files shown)


iLLiCiTiT:
We should maybe cache result of `get_colors_data` too (which actually read the file).

BigRoy:
> We should maybe cache result of get_colors_data too (which actually read the file).

Usage in code to that function is very limited it seems! :) As in, only the function in `openpype.style` actually use it. So it was at most ever called five times. Anyway, added a commit.

Should we still ensure the functions return deep copies? It might lose the speed benefits then however.

iLLiCiTiT (approved):

PR #3915, opened 2022-09-26
Publisher: Add instance to context only if is not added yet

iLLiCiTiT:
## Brief description
Add instance to context only if is not added yet so there are no warnings on helper functions.


BigRoy:
I'm not sure whether this hides the root of the problem - why are the same instances potentially being added anyway I wonder?

iLLiCiTiT:
> I'm not sure whether this hides the root of the problem - why are the same instances potentially being added anyway I wonder?

I just re-read the comment and you're right. I though it's because someone called `_add_instance_to_context` multiple times but that's not the case.

Closing as this is not solution...

PR #3919, opened 2022-09-26
Houdini: Tweak Shelves set manager to ignore invalids

BigRoy:
## Brief description

I was unable to start **Houdini Core 19.5.303 Py 3.7** on Windows due to the Shelf Set Manager implemented with #3652 

It would just hang on Houdini splash screen without continuing. What causes it exactly I'm not sure - but I had no shelves set or defined but the shelves still tried to create and or set shelves which apparently broken my Houdini start.

With the changes implemented here the default is now correctly fully empty to avoid warnings, it'll now also not try `get_or_create_shelf_set` if the shelf definition for that shelf is empty (like previous default value in settings).

Also instead of returning it'll now continue on entries that are invalid in settings.

### Additional Info

I'd still like to figure out what exactly stops my Houdini from continuing past the splash screen - still investigating now.

[Originally mentioned in comments of this other PR](https://github.com/pypeclub/OpenPype/pull/3046#discussion_r979923594)

## Testing notes:
1. Start Houdini
2. Set up shelves in OpenPype settings for Houdini - ensure they work as intended.

### DIFF
--- a/openpype/hosts/houdini/api/shelves.py
+++ b/openpype/hosts/houdini/api/shelves.py
@@ -1,7 +1,6 @@
 import os
 import logging
 import platform
-import six
 
 from openpype.settings import get_project_settings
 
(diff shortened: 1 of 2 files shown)


iLLiCiTiT (approved):
Changes look good for me. @antirotor can you test it?

antirotor (approved):
Seems to work. I had troubles replicating the original issue but the code looks sound.

PR #3923, opened 2022-09-27
Maya: Fix Scene Inventory possibly starting off-screen due to maya preferences

### LINKED ISSUE #3921: Reset window position

It is happening quite often recently that opening an OP window (Scene Inventory in Maya for example) is not working for the user.

There is only a message in Script Editor `Using existing QApplication..`

image

This "error" means the window is already opened, but it is hidden for the user outside of the visible area. (happening a lot with dual monitor setups and remoting to that machine)

We could implement some hot key to reset the window position back to the default location (and print the hotkey to the console with the `Using existing QApplication..` message

[cuID:OP-4154]

BigRoy:
## Brief description

Remove "saveWindowPref" property.

This should fix #3921 

## Description

Now Maya will not restore/remember the previous Scene Inventory position.
The UI will now behave similar to other hosts.

## Testing notes:
1. Start Scene Inventory - move it off-screen off to the side.
2. Close it down, close maya, save maya preferences, etc.
3. Restart Maya. 
 
Previously before this PR Maya would have reopened Scene Inventory in position where it was last time, now it doesn't.

### DIFF
--- a/openpype/tools/sceneinventory/window.py
+++ b/openpype/tools/sceneinventory/window.py
@@ -40,8 +40,6 @@ class SceneInventoryWindow(QtWidgets.QDialog):
         project_name = os.getenv("AVALON_PROJECT") or "<Project not set>"
         self.setWindowTitle("Scene Inventory 1.0 - {}".format(project_name))
         self.setObjectName("SceneInventory")
-        # Maya only property
-        self.setProperty("saveWindowPref", True)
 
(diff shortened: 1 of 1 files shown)


m-u-r-p-h-y (approved):
confirmed that previously the position of the Inventory window was stored and remembered.

Now it is always opened in default, central position. Thx!

PR #3925, opened 2022-09-28
Maya: Remove unused 'openpype.api' imports in plugins

iLLiCiTiT:
## Brief description
Almost every Maya validator plugin has import of `openpype.api` which is not used.

## Description
Removed `openpype.api` imports from Maya validarors. They could be there because of hidden imports (found out 1 where I had to add different import).

## Testing notes:
1. All changed plugins should be discovered
2. All changed plugins should work

### DIFF
--- a/openpype/hosts/maya/plugins/publish/validate_animation_content.py
+++ b/openpype/hosts/maya/plugins/publish/validate_animation_content.py
@@ -1,5 +1,4 @@
 import pyblish.api
-import openpype.api
 import openpype.hosts.maya.api.action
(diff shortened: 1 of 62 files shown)


BigRoy (approved):
Nice. At first glance this looks good to me.

LiborBatek:
I have done publish of the model, look, rig and also loaded all into some scene and no any errors found.
Seems ok to me.

LiborBatek (approved):
Works ok. Not any errors appeared. Tested on mode, look, rig families...

PR #3928, opened 2022-09-28
Fusion: Implement callbacks to Fusion's event system thread

BigRoy:
## Brief description

Implements "save", "open" and "new" OpenPype events based on Fusion notifications of certain events.

## Description

Note that the Fusion events may not be 100% reliable. It could still be that a 'different action' might need to be listened to - but these events implemented now seemed reliable enough.

## Additional info

This adds a background QThread with the OpenPype menu which will listen to the Fusion events for comp open, save and new comp. This has the downside that IF you were to ever open the OpenPype menu twice (or more times) that the Events will be heard multiple time and each running instance of the menu would then trigger the callbacks to happen.

As such - a next step likely would be to figure out a way for the menu to just get raised instead of re-requested to open if it was already opened for the current fusion session so that only ever a single OpenPype session/menu would be open for that Fusion session. This is difficult with Fusion because the scripts run in a separate background process each, disconnected from the Fusion app process itself.

Likely we'll need to keep a port open on the process that will highlight to fusion that the current process ID already has a menu open, or alike. However, at the same time - multiple Fusions might be open which do each want their own menu. So we'll still need to be able to identify per Fusion session too.

## Testing notes:

1. Start Fusion, Start OpenPype menu.
2. You should get outdated containers reports on Comp open (if they have any outdated containers) and on Save it should also report if your current comp doesn't match the asset FPS, etc.

### DIFF
--- a/openpype/hosts/fusion/api/pipeline.py
+++ b/openpype/hosts/fusion/api/pipeline.py
@@ -6,10 +6,12 @@ import sys
 import logging
 
 import pyblish.api
+from Qt import QtCore
 
 from openpype.lib import (
     Logger,
-    register_event_callback
+    register_event_callback,
+    emit_event
 )
 from openpype.pipeline import (
     register_loader_plugin_path,
@@ -86,10 +88,10 @@ def install():
         "instanceToggled", on_pyblish_instance_toggled
     )
 
-    # Fusion integration currently does not attach to direct callbacks of
-    # the application. So we use workfile callbacks to allow similar behavior
-    # on save and open
-    register_event_callback("workfile.open.after", on_after_open)
+    # Register events
+    register_event_callback("open", on_after_open)
+    register_event_callback("save", on_save)
+    register_event_callback("new", on_new)
 
 
 def uninstall():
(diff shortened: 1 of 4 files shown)


openpype/hosts/fusion/api/pipeline.py
```diff
@@ -39,12 +42,13 @@
...
+class FusionLogHandler(logging.Handler):
+    # Keep a reference to fusion's Print function (Remote Object)
+    _print = getattr(sys.modules["__main__"], "fusion").Print
+
     def emit(self, record):
         entry = self.format(record)
-        comp = get_current_comp()
-        if comp:
-            comp.Print(entry)
+        self._print(entry)
```
BigRoy, line 51: This is mostly an optimization.

openpype/hosts/fusion/api/pulse.py
```diff
@@ -19,9 +19,12 @@ def run(self):
...
                 return
-            try:
-                app.Test()
-            except Exception:
+
+            # We don't need to call Test because PyRemoteObject of the app
+            # will actually fail to even resolve the Test function if it has
+            # gone down. So we can actually already just check by confirming
+            # the method is still getting resolved. (Optimization)
+            if app.Test is None:
```
BigRoy, line 27: As comment explains this will mostly result in less delays (since calling `Test()` actually called against the main fusion UI).

openpype/hosts/fusion/api/pipeline.py
BigRoy: **Warning:** I've just found that this logic does not work in Fusion 17.2 (released around June 2021) but does work in Fusion 17.4 since the `fusion` instance does not have a `Print` method in the older Fusion.

We might need to introduce a fallback to support older fusion versions.

PR #3930, opened 2022-09-29
Extract Review code refactor

BigRoy:
## Brief description

Trying to reduce complexity of Extract Review plug-in

- Re-use profile filtering from lib
- Remove "combination families" additional filtering which supposedly was from OP v2
- Simplify 'formatting' for filling gaps
- Use `legacy_io.Session` over `os.environ`

## Testing notes:

1. Extract some reviews


mkolar:
> I'll leave it as this for now since further cleanup/refactor might be warranted only after a check of current changes to avoid making way too big of a PR.
> 
> Pretty decent score: afbeelding

Should we make this into a new game? The best add/reduce ratio of a PR gets a cookie.

openpype/plugins/publish/extract_review.py
```diff
@@ -777,73 +786,53 @@ def fill_sequence_gaps(self, files, staging_dir, start_frame, end_frame):
...
-            for frame, file in new_files.items():
-                self.log.info(
-                    "Filling gap {} with {}".format(frame, file))
+        # Both collection frame indexes and the missing indexes are sorted
+        # so we can keep track of the nearest existing frame with an iterator
+        existing_indices_iter = iter(col.indexes)
+        prev_idx = next(existing_indices_iter, None)
+        next_idx = next(existing_indices_iter, prev_idx)
+        if prev_idx is None:
+            raise RuntimeError("No indices in input collection. This is a bug")
```
iLLiCiTiT, line 809: Please us `KnownPublishError` from `openpype.pipeline.publish`.

BigRoy: It actually isn't a known error. This should technically NEVER happen. Since it wouldn't have an assembled collection to begin with earlier in the code if there would've been no "indices". This really would be an odd code error/bug if it would occur.

iLLiCiTiT: If you raise error before other possible error then it's known error. If it should never happen then the check is redundant and should be removed...

Note: I would simplify it to this:
```
        first_frame = col.indexes[0]
        prev_frame = None
        hole_frame_to_nearest = {}
        # First prepare which hole is filled with frame
        for frame in range(int(frame_start), int(frame_end) + 1):
            if frame in col.indexes:
                prev_frame = frame
            elif prev_frame is None:
                hole_frame_to_nearest[frame] = first_frame
            else:
                hole_frame_to_nearest[frame] = prev_frame
        
        # Calculate paths
        holes_to_nearest = {}
        for hole_frame, src_frame in hole_frame_to_nearest.items():
            src_fpath = os.path.join(staging_dir, col_format % src_frame)
            hole_fpath = os.path.join(staging_dir, col_format % hole_frame)
            holes_to_nearest[hole_fpath] = nearest_fpath
```

mkolar: I agree that this is indeed a known error case or there'sno point in raising it. By definition if you are catching it and raising it, you know that is can happen and should use the `KnownPublishError` so we can make any know errors less scary for the user.

openpype/plugins/publish/extract_review.py
```diff
@@ -384,10 +393,10 @@ def _render_output_definitions(
     def input_is_sequence(self, repre):
         """Deduce from representation data if input is sequence."""
         # TODO GLOBAL ISSUE - Find better way how to find out if input
-        # is sequence. Issues( in theory):
-        # - there may be multiple files ant not be sequence
-        # - remainders are not checked at all
-        # - there can be more than one collection
+        #  is sequence. Issues (in theory):
```
iLLiCiTiT, line 396:
```suggestion
        # is sequence. Issues (in theory):
```

iLLiCiTiT, line 396: Why these have 2 spaces?

openpype/plugins/publish/extract_review.py
mkolar: @BigRoy could we address this? we'll hopefully get to some proper testing on more complicated data soon.

iLLiCiTiT:
Merged 'develop', resolved few conflicts, modified few lines to match develop changes. Removed task name filtering from ExtractReview profiles -> it is not available in settings, and was not for a long time. Replaced `AssertionError` and `ValueError` with `KnownPublishError`. Modified hole frame mapping to more readable snipped (commit https://github.com/ynput/OpenPype/pull/3930/commits/5b7e5bff53128d509dcc9e21c07a540ed1da73c3 ).

iLLiCiTiT (approved):
Looks good. @BigRoy let me know if you agree/disagree with my changes.

BigRoy:
> I would say it was actually never used, it was added for edge cases of one client. Finding nearest frame is unnecessary complex and not sure how much useful? So I vote for option 2 (previous frame). You're right it could simplified even more with min, can you commit the change?

I have cleaned it up with https://github.com/ynput/OpenPype/pull/3930/commits/d13b8208f5618952362171241605d4cdd1094c04 but kept the "previous" frame logic instead of nearest frame.

I guess someone should give the PR a few test runs?

kalisp (approved):
Seems to provide same results in AE, Nuke, Maya, PS (according to tests).

PR #3934, opened 2022-10-04
General: Direct settings imports

iLLiCiTiT:
## Brief description
Use direct import of settings functions.

## Testing notes:
Nothing should change or crash
Hosts:
- [x] Nuke
- [ ] Flame
- [ ] Maya
- [ ] Resolve
- [x] TrayPublisher
- [ ] TVPaint
- [ ] Unreal

Addons:
- [ ] Ftrack
- [ ] Kitsu
- [ ] Shotgrid

Tools:
- [x] Creator
- [x] Pyblish Pype
- [x] Local Settings

### DIFF
--- a/openpype/tools/pyblish_pype/control.py
+++ b/openpype/tools/pyblish_pype/control.py
@@ -22,7 +22,7 @@ import pyblish.version
 from . import util
 from .constants import InstanceStates
 
-from openpype.settings import get_project_settings
+from openpype.settings import get_current_project_settings
 
 
 class IterationBreak(Exception):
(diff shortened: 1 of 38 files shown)


openpype/tools/pyblish_pype/control.py
```diff
@@ -22,7 +22,7 @@
 from . import util
 from .constants import InstanceStates
 
-from openpype.api import get_project_settings
+from openpype.settings import get_project_settings
```
BigRoy, line 25: Could be replaced with `from openpype.settings import get_current_project_settings` looking at the code down below. No?

iLLiCiTiT: `get_current_project_settings` is not the best function available in settings api. But it is probably better to use it if env is already used in code.

kalisp (approved):
Clicked around Tray Publisher, Local Settings, Creator

PR #3936, opened 2022-10-04
Publisher: Create dialog is part of main window

iLLiCiTiT:
## Brief description
Reduced dialogs in publisher. Create content is now part of publisher main window. Window is split into more pages using tabs.

## Description
Didn't change any logic or workflow just reorganized UI elements. Publisher's main window now has 4 tabs with different content.
- Create is first page with context selection, creators selection, their attributes and list of created instanced.
- Publish (which is actually attributes part) is to see all created instances, their context and attributes
- Report page with validation errors or brief info for artist
- Details - publish report details (for developers or TDs)

Controller callbacks are based on EventSystem with topics instead of handling them on it's own. Creator details widget was removed and replaced with help dialog. At this moment can only show help for creators but more can be added in future. Publish progress overlay can be shrunked.

Changed how controller is accessing current context information. Is using host's method instead of environment variables or `legacy_io`. Controller events are not sending objects via events but only relevant data.

## Additional information
There is a lot of TODOs but this is already a massive change that should be reviewed before next work.

### Screenshot
image

## Testing notes:
Publisher UI should work as before (shouldn't have any "new bugs"). Easiest to test is to use Tray Publisher.


jakubjezek001:
What is this button for? It seems to me it does nothing. Do we need it? 
image

jakubjezek001:
I wonder if we could find a way to make data in Create options persistent even after Refresh button is hit. Here is an example 

Image 1 show defined data in Options panel:
image

Image 2 show the panel after refresh:
image

jakubjezek001 (approved):
Works well. The us is a huge improvement!

iLLiCiTiT:
> What is this button for? It seems to me it does nothing. Do we need it?

Fixed

mkolar (changes requested):
image
A few smaller comments. Also it should be `publishing went smoothly` or `publishing ran smoothly`.

Also the colored bar should always expand when some process is running I think. it's quite confusing otherwise, when publish is running without the progress bar.

mkolar (approved):
Solid update. Lots of work left, but that's adding new functionality and needs to be in a separate PRs. happy to merge

PR #3937, opened 2022-10-04
AttributeDefs: Fix crashing multivalue of files widget

iLLiCiTiT:
## Brief description
Widget for `FileAttrDef` can handle multivalue so multiselection in publisher don't crash on attributes creation.

## Description
This is a fast fix to have it "not crashing" we could maybe (in future) find better way how to handle multiselection in this widget.

## Testing notes:
1. Open TrayPublisher
2. Create at least 2 instances using same simple creator (enter files there)
3. Multiselect instances in UI
4. Files widget should not cause crash and be disabled


kalisp (approved):
Fixes the issue.

PR #3942, opened 2022-10-05
Maya + Yeti: Load Yeti Cache fix frame number recognition

BigRoy:
## Brief description

Previously a cache starting at frame 1005 would fail due to a "5" being present. This would only be noticable if the start frame used for detection included a digit that was not 0, 1, 2, 3 or 4. I guess we've had the luck to always have been starting with a frame number like 1001, 1000, etc. because if any number in 56789 would have been present then this would previously fail.

## Description

It's funny how long this must have been missed. We've been so lucky!

## Testing notes:

1. Publish a yeti cache with 5, 6, 7, 8 or 9 in the frame number (e.g. single frame at frame 1005)
2. Load yeti cache should work.

### DIFF
--- a/openpype/hosts/maya/plugins/load/load_yeti_cache.py
+++ b/openpype/hosts/maya/plugins/load/load_yeti_cache.py
@@ -250,7 +250,7 @@ class YetiCacheLoader(load.LoaderPlugin):
         """
 
         name = node_name.replace(":", "_")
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
Did a test with publishing just frame 1005. All went smoothly and later loaded well into shot scene as yeti single frame cache.

So its working fine.

PR #3943, opened 2022-10-05
Maya: fix regression of Renderman Deadline hack

antirotor:
## Fix

Fix regression of missing Renderman hack for Deadline.

For newer versions of Renderman, DL needs renderer to be set as `renderman22`. Refactoring of submission plugins removed this hack and this PR is putting it back.


BigRoy (approved):
Don't have renderman to test but code changes look good to me.

PR #3944, opened 2022-10-05
Maya: Set default value for RenderSetupIncludeLights option

antirotor:
## Fix

`RenderSetupIncludeLights` must be either set to 1 or 0 or not at all. This PR is adding default value for edge cases where this is not handled. Without it, DL is crashing.

### DIFF
--- a/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
+++ b/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
@@ -57,8 +57,9 @@ class MayaPluginInfo:
     RenderLayer = attr.ib(default=None)  # Render only this layer
     Renderer = attr.ib(default=None)
     ProjectPath = attr.ib(default=None)  # Resolve relative references
+    # Include all lights flag
     RenderSetupIncludeLights = attr.ib(
-        default="1", validator=_validate_deadline_bool_value)  # Include all lights flag
+        default="1", validator=_validate_deadline_bool_value)
 
 
 @attr.s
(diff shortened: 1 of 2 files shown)


m-u-r-p-h-y (commented):
submitted Maya render with old render instance with missing `Render Setup Include Lights` attribute.


Deadline job has the attribute in job properties with value 1 so Deadline wont crash anymore.

The question is, if we are really assigning default value from project settings
image


image

m-u-r-p-h-y:
another question is, for cases with the missing attribute on the instance and in settings, if the "safer" default value should be 0 instead of 1

antirotor:
> another question is, for cases with the missing attribute on the instance and in settings, if the "safer" default value should be 0 instead of 1

I think it should follow Maya default and that is `1`. Added code to pull default value from the Settings.

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
```diff
@@ -185,12 +198,25 @@ def get_plugin_info(self):
...
         context = instance.context
 
+        # Set it to default Maya behaviour if it cannot be determined
+        # from instance (but it should be, by the Collector).
+
+        default_rs_include_lights = instance.context.data\
+            .get('project_settings')\
+            .get('maya')\
+            .get('RenderSettings')\
+            .get('enable_all_lights')
```
BigRoy, line 208: This `.get().get().get()` feels wrong. As soon as one of the keys doesn't exist it'd be `None` and doing `.get()` would fail. So it's just as prone to error as just doing `["project_settings"]["maya"]["RenderSettings"]["enable_all_lights"]`

Also the fallback down below to the default value could thus still fall back to "None" otherwise if `enable_all_lights` wouldn't exist.

m-u-r-p-h-y:
> I think it should follow Maya default and that is `1`. Added code to pull default value from the Settings.

that's exactly what I meant. I thought our default 0 is Maya's default. You are correct it is 1, so we should change our defaults in `openpype\settings\defaults\project_settings\maya.json`  to 1 to follow Maya's default value.

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
antirotor: you are right, I've just copy&pasted it from somewhere. I guess it was like that because of readability but that is questionable anyway.

PR #3945, opened 2022-10-05
Fx collection custom prop del

Tilix4:
## Brief description
Because `del` removes from the memory it may lead to missing reference. `dict.pop()` is safer because local.


Tilix4:
Wrong PR, sorry.

PR #3947, opened 2022-10-06
General: Set root environments before DCC launch

iLLiCiTiT:
## Brief description
Root environments can be used in enviornments of applications now.

## Description
We set them after host launch so they were not available at the moment when system settings environment variables were resolved.

## Testing notes:
1. Use `OPENPYPE_ROOT_WORK` in DCC environments
    - e.g. `{"MY_ENV": "{OPENPYPE_ROOT_WORK}"}`
2. Launch the DCC
3. Check if the env is filled

### DIFF
--- a/openpype/lib/applications.py
+++ b/openpype/lib/applications.py
@@ -1403,6 +1403,7 @@ def get_app_environments_for_context(
 
         "env": env
     })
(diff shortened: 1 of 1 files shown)


BigRoy:
You're a hero!

PR #3949, opened 2022-10-06
Publisher: Pass instance to subset name method on update

iLLiCiTiT:
## Brief description
Creator has access to existing instance on it's subset name change.

## Description
Pass instance to `get_subset_name` on update of existing instance subset name. This gives option to use it but does not modify the code itself.

## Additional information
In some cases creators want to define additional data for subset name template (e.g. in pre-create attributes) and should have access to existing instance to receive the information from the instance.

This should give ability to Photoshop creators to fill `"layer"` key on [existing instances](https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/photoshop/plugins/create/create_image.py#L167).

## Testing notes:
Everything should work as before and nothing should crash.

### DIFF
--- a/website/docs/dev_publishing.md
+++ b/website/docs/dev_publishing.md
@@ -198,6 +198,37 @@ class RenderLayerCreator(Creator):
 
 - **`get_dynamic_data`** (method) - Can be used to extend data for subset templates which may be required in some cases.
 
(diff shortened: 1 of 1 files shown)


kalisp:
I think it would be beneficial to add some mention of this into documentation, or example creator.

PR #3950, opened 2022-10-07
Maya: Deadline OutputFilePath hack regression for Renderman

antirotor:
## Fix

During refactoring of Maya Deadline submitter, we've accidentaly dropped Renderman specific hacks. This is returning back one of them. 

Renderman integration in Deadline relies on plugin info `OutputFilePath` to be set differently than in other renderes. I has to point directly to the file output, unlike with other renderers where this can be further expanded by `OutputFilePrefix`

### Testing

Renderman renders should render and publish correctly


iLLiCiTiT (approved):
Looks ok

BigRoy (approved):
I agree - this looks ok. Odd, but ok.

PR #3951, opened 2022-10-07
General: Control Thumbnail integration via explicit configuration profiles

kalisp:
## Brief description
Thumbnail representations are needed for Ftrack integration to show something in Versions. This provides more explicit configuration if they should be integrated, eg. stored in DB.

## Description
This PR provides setting up explicit profiles (host, task name, task type, subset, family) to enable/disable integration (eg storing in DB AND storing thumbnail in 'publish' folder)

Storing in publish folder might be sometimes confusing for artists.

This adds explicit step, if no profile gets added, state as is still remains.

## Additional info
In most cases thumbnail representations shouldn't be integrated. There is no standardization right now if the thumbnail repre gets  stored or not. 
After customer issue it was decided that more control could be beneficial, It could be approached that configuration gets added to each host, but better solution would to have single point of configuration (in similar fashion as ftrack family is being added right now).

In the future it might be good to clean up all host implementations (which will require wider testing), this mostly solves important issue for a customer.

## Testing notes:
1. configure `project_settings/global/publish/PreIntegrateThumbnails/integrate_profiles/0/integrate_thumbnail`
2. publish some simple image through Webpublisher, without studio processing

### DIFF
--- a/openpype/plugins/publish/preintegrate_thumbnail_representation.py
+++ b/openpype/plugins/publish/preintegrate_thumbnail_representation.py
@@ -62,7 +62,7 @@ class PreIntegrateThumbnails(pyblish.api.InstancePlugin):
                 thumbnail_repre["tags"].append("delete")
         else:
             if "delete" in thumbnail_repre["tags"]:
-                thumbnail_repre["tags"].pop("delete")
+                thumbnail_repre["tags"].remove("delete")
 
         self.log.debug(
             "Thumbnail repre tags {}".format(thumbnail_repre["tags"]))
--- a/openpype/plugins/publish/preintegrate_thumbnail_representation.py
+++ b/openpype/plugins/publish/preintegrate_thumbnail_representation.py
@@ -19,7 +19,7 @@ from openpype.lib.profiles_filtering import filter_profiles
 class PreIntegrateThumbnails(pyblish.api.InstancePlugin):
     """Marks thumbnail representation for integrate to DB or not."""
 
-    label = "Override Integrate Thumbnails"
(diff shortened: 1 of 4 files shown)


openpype/plugins/publish/preintegrate_thumbnail_representation.py
```diff
@@ -0,0 +1,68 @@
...
+        host_name = instance.context.data["hostName"]
+
+        anatomy_data = instance.data["anatomyData"]
+        task = anatomy_data.get("task", {})
+
+        found_profile = filter_profiles(
+            self.integrate_profiles,
+            {
+                "hosts": host_name,
+                "tasks": task.get("name"),
```
iLLiCiTiT, line 49: Note:
We have combined usage of `"tasks"` and `"task_names"` in profiles and I would say that `"task_names"` is better (as is more explicit).

openpype/plugins/publish/preintegrate_thumbnail_representation.py
```diff
@@ -0,0 +1,68 @@
...
+
+    label = "Should Integrate Thumbnails"
+    order = pyblish.api.IntegratorOrder - 0.1
+    families = ["review"]
+
+    integrate_profiles = {}
+
+    def process(self, instance):
+        thumbnail_repre = None
+        for repre in instance.data["representations"]:
```
iLLiCiTiT, line 30:
```suggestion
        repres = instance.data.get("representations")
        if not repres:
            return

        thumbnail_repre = None
        for repre in repres:
```

64qam (approved):

PR #3952, opened 2022-10-07
TrayPublisher: Extract thumbnail for other families

iLLiCiTiT:
## Brief description
Created extract thumbnail for tray publisher purposes.

## Description
Global Extract thumbnail plugin has limitations to certain families and is based on representations on instances. Adding or removing families can cause huge issues on farm or in specific situations. In most of cases there is no need of existing representation on instance to create thumbnail and at the same time no family filter so key `"thumbnailSource"` can be added to instance which should point to a file on disk from which can be created thumbnail.

This is used for simple creators and batch mov creator. The extractor happens before the global extractor so the global should skip most of already processed instances but still can work for e.g. editorial instances which create thumbnail sources during extraction.

## Additional information
This practically adds a duplicate of thumbnail plugin. We should maybe add some thumbnail creation mechanism in pipeline api as common code for thumbnail generation.

## Testing notes:
1. Create a pointcache instance in TrayPublisher with review mov file
2. Publish the file
3. All should be integrated as expected
4. Thumbnail should be created and integrated (visible in Loader or e.g. in Ftrack)


kalisp (approved):
Replaces existing Thumbnail, provides same result.
I don't like this exception message, I am guessing it is caught exception, so it is just a warning, but got me spooked during regular skimming through.
exception

Plus docstring for `extract_thumnbail` seems to be cut off.

PR #3954, opened 2022-10-07
Maya: Hotfix for Deadline + Redshift renders without merge AOVs

### LINKED ISSUE #3953: Maya + Redshift: Deadline render duplicates <aov> in output render
**Running version**
3.14.4-nightly.1

**Describe the bug**

Redshift renders through Deadline with Merge AOVs (= Multilayer EXR) disabled will render with a double `<aov>` (RenderPass) in the filename. Like this:

afbeelding

Publishing will fail since the metadata will expect the AOV / RenderPass token to exist only once (and for the Beauty layer by default to not be there at all - hence why it has a single token in the screenshot)

This issue originates from the Deadline submitter taking the file prefix from the layer metadata from `lib_renderproducts` and [assign that as the file prefix for the Deadline submission](https://github.com/pypeclub/OpenPype/blob/develop/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py#L473-L476)

But the [Layer Metadata in `lib_rendersettings` collects the file prefix that matches what the renderer produces as output](https://github.com/pypeclub/OpenPype/blob/95b5f72852d9c933d402683b5a2403adc09be0a8/openpype/hosts/maya/api/lib_renderproducts.py#L983-L986) - and doesn't necessarily match the input setting.

Hence you get an `_<aov>` token forced in that:
afbeelding

The Deadline submitter thus should not take the prefix from the layer metadata but from the scene render settings instead.

**To Reproduce**

1. Render with Redshift from Maya without merge AOVs / multilayer through Deadline.
2. Publish of rendered images fails.

Example settings:
afbeelding

**Expected behavior**

Output from Deadline should not deviate from scene settings.
And above all, publishing must work.

 - OS: Windows
 - Host: Maya 2022.3, Redshift 3.5.06

**Additional context**

Issue likely has been introduced with the implementation of the new deadline submitter which uses the Abstract Deadline Submitter as base.

[cuID:OP-4208]

BigRoy:
Fixes #3953

## Brief description

Instead of taking the resolved output from Render Products which is the "image prefix" that matches the written output files for the renderer with the current layer's setting submit to Deadline with the actual "image prefix" value of that layer.

## Description

See #3953 for more details on why the error happened originally.

### Additional info

Would've preferred the code to be even cleaner than this by e.g. just having something like `imageFilePrefix` or `rawImageFilePrefix` stored in the `instance.data` so we could retrieve it directly instead of having to access the scene again.

## Testing notes:

1. Render with Redshift from Maya without merge AOVs / multilayer through Deadline.
2. Publish of rendered images fails.

### DIFF
--- a/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
+++ b/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
@@ -32,6 +32,9 @@ from maya import cmds
 
 from openpype.pipeline import legacy_io
 
+from openpype.hosts.maya.api.lib_rendersettings import RenderSettings
+from openpype.hosts.maya.api.lib import get_attr_in_layer
+
 from openpype_modules.deadline import abstract_submit_deadline
 from openpype_modules.deadline.abstract_submit_deadline import DeadlineJobInfo
 
@@ -471,9 +474,10 @@ class MayaSubmitDeadline(abstract_submit_deadline.AbstractSubmitDeadline):
             job_info.AssetDependency += self.scene_path
 
         # Get layer prefix
-        render_products = self._instance.data["renderProducts"]
-        layer_metadata = render_products.layer_data
-        layer_prefix = layer_metadata.filePrefix
+        renderlayer = self._instance.data["setMembers"]
+        renderer = self._instance.data["renderer"]
+        layer_prefix_attr = RenderSettings.get_image_prefix_attr(renderer)
+        layer_prefix = get_attr_in_layer(layer_prefix_attr, layer=renderlayer)
 
         plugin_info = copy.deepcopy(self.plugin_info)
         plugin_info.update({


BigRoy:
Bug was also present with V-Ray. This PR would also fix that case.

LiborBatek:
Hi, I will do it asap...we have some issues with Deadline submits causing immediete fails/errors so I cant test it right now, sry.

LiborBatek:
Ok I have tested it and it renders all out with correct naming like ```Main.####.exr``` or/and ```Main_N.####.exr``` (for normal AOV pass).

Not sure but Publish of the rendered imgs fails... not clear for me if its right or not by reading your description above...

Just clarify it and I can make Review to finnish it. Thx
L.

antirotor:
this was merged in with #3981

PR #3959, opened 2022-10-11
Settings: Move imageio from project anatomy to project settings [pypeclub]

jakubjezek001:
## Brief description

This implements #3865 by moving `imageio` settings per host from project anatomy to project settings.

Before:
```
project_anatomy/imageio/{app}
```

After:
```
project_settings/{app}/imageio
```

## Description

Currently untested - no backwards compatible or fallbacks implemented. Basically your old project anatomy customized settings will be lost. 🔥 - So as #3865 we'll need to look into how we'll want to transfer those settings over _automagically_ 🪄 .

## Additional info

Note that each `imageio` group has now set `is_group` to True as such changing one setting in the imageio for a host will now override that full imageio block of settings for the project. It's up for discussion whether we want that - if we want to change around we'll need to add some labels to some of the settings (required for `is_group` to be False) but that should be fine.

It could be that some documentation might also need to be updated along. 
- Even though admittedly the old [Color Managements docs are very slim](https://openpype.io/docs/admin_settings_project_anatomy#colour-management-and-formats) - it's just the header - ha! 
- But potentially a comment on and an update to description [of this Nuke Color Management tutorial video](https://openpype.io/docs/artist_hosts_nuke_tut/#nuke-color-management) would be worthwhile too to make users aware the settings have moved over in the future. 
- Hiero documentation will need to be updated around [apply colorspace project](https://openpype.io/docs/artist_hosts_hiero/#apply-colorspace-project)

#### Copying ImageIO settings from Anatomy to Project Settings

Any potential backwards compatible "updating" would only need to basically 'copy the settings values' from `project_anatomy/imageio/{app}` on top of `project_settings/{app}/imageio` to transfer the exact settings. So in essence it is:

```python
for app in anatomy["imageio"]:
    value = app.value
    project_settings[app.key]["imageio"].set(value)
    project_settings.save()
```

---

I'm not awaiting the moment this PR gets 'assigned' to me and I get all the blame in the future. 🗡️ 

## Testing notes:

1. 🔥 🚒 Get your fire extinguishers ready
2. Test color management related features across the hosts

- [x] Flame 
- [x] (and Flare?)
- [ ] Maya
- [x] Nuke
- [x] Hiero
- [x] (Did Resolve have any? I couldn't find it. If it didn't, let's check this box and continue our ignorance.)


openpype/hosts/maya/api/lib.py
```diff
@@ -23,7 +23,7 @@
     get_last_versions,
     get_representation_by_name
 )
-from openpype.api import get_anatomy_settings
+from openpype.api import get_project_settings
```
iLLiCiTiT, line 26:
```suggestion
from openpype.settings import get_project_settings
```

m-u-r-p-h-y (approved):
Maya settings are correctly applied from new location

image

iLLiCiTiT:
Conflict resolve is needed

openpype/hosts/nuke/api/lib.py
```diff
@@ -563,7 +563,15 @@ def get_node_path(path, padding=4):
 
 
 def get_nuke_imageio_settings():
-    return get_anatomy_settings(Context.project_name)["imageio"]["nuke"]
+    project_imageio = get_project_settings(
+        Context.project_name)["nuke"]["imageio"]
+
+    # backward compatibility for project started before 3.10
```
iLLiCiTiT, line 569:
```suggestion
    # backward compatibility for project started before 3.14.4
```

PR #3960, opened 2022-10-11
Settings: Added configuration of ValidateVersion

kalisp:
## Brief description
ValidateVersion controls if publishing version wasn't published before, eg. we are not trying to overwrite existing files. It should be configurable in which hosts plugin gets triggered now.

## Description
Added as defaults existing values `maya`, `nuke`, `blender`, `houdini`, `standalone`.

Added also `photoshop`, `aftereffects` as they should be there (AE was as a default in 2.x)

## Testing notes:
1. publish with any of the configured hosts and check that "Validate Version" is present

### DIFF
--- a/openpype/settings/defaults/project_settings/global.json
+++ b/openpype/settings/defaults/project_settings/global.json
@@ -31,7 +31,16 @@
         "ValidateVersion": {
             "enabled": true,
             "optional": false,
-            "active": true
+            "active": true,
+            "hosts": [
+                "aftereffects",
+                "blender",
+                "houdini",
+                "maya",
+                "nuke",
+                "photoshop",
+                "standalonepublisher"
+            ]
         },
         "ValidateIntent": {
             "enabled": false,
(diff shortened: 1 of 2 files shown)


BigRoy:
What's the use care where we DO want to overwrite over an existing version? :) Isn't that inherently dangerous?

m-u-r-p-h-y:
Maya

image

m-u-r-p-h-y:
> overwrit

m-u-r-p-h-y:
sorry

m-u-r-p-h-y:
> What's the use care where we DO want to overwrite over an existing version? :) Isn't that inherently dangerous?

I think there is just bad wording. It is meant to prevent the possible situation which could lead to (accidental) overwriting of published data.

kalisp:
After internal discussion, adding 'hosts' as a filtering device in Settings is not preferable option. Replaced with https://github.com/pypeclub/OpenPype/pull/3965

`ValidateVersion` should be updated to allow more elaborate configuration, see https://github.com/pypeclub/OpenPype/issues/3966

PR #3961, opened 2022-10-11
Webpublisher: use max next published version number  for all items in batch

kalisp:
## Brief description
With this PR it is possible to keep same version for all published items via WP (without studio processing).

## Description
Artist is uploading `.psd` and `.png`, it should look for highest published of `workfile (.psd)` or `image (.png)`, increase this by 1 and use this version number for BOTH workfile and image instances.

It looks for max version number, from both subsets, eg. if more `image` than `workfile` is published , workfile version number will follow image one.

## Testing notes:
1. set `Sync next publish version `project_settings/webpublisher/publish/CollectPublishedFiles`
Screenshot 2022-10-11 152401
3. publish only .psd (maybe couple of times)
4. publish both .psd and .png >> resulting version for .png should match to .psd (eg. there will be gaps in images versions)

### DIFF
--- a/openpype/hosts/webpublisher/plugins/publish/collect_published_files.py
+++ b/openpype/hosts/webpublisher/plugins/publish/collect_published_files.py
@@ -37,6 +37,15 @@ class CollectPublishedFiles(pyblish.api.ContextPlugin):
 
     This is not applicable for 'studio' processing where host application is
     called to process uploaded workfile and render frames itself.
+
+    For each task configure what properties should resulting instance have
+    based on uploaded files:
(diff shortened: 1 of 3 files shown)


64qam:
Looks good. However, publishing a single file in one batch results in the newer version (e.g. uploading a psd `v003`, results in `v004`). Then create a new batch with the corresponding png file (also `v003`) results also in `v004`. The correct action should be that the png from a new batch should end up with `v005`. So every new batch must look for the highest version in both workfile family folder and image/render family folder and version up files from wp batch.

64qam (approved):

PR #3964, opened 2022-10-11
General: Use direct import of resources

iLLiCiTiT:
## Brief description
Import `resources` directly instead of through api.

## Testing notes:
- [x] Launcher tool should work propely
- [x] OpenPype info widget should work propely (click on version in Tray)
- [x] StandalonePublisher should work propely
- [x] Maya should integrate properly
- [ ] Nuke should integrate properly


kalisp (approved):
Couldn't test in Nuke as I cannot get an license, but it work in other places.

PR #3970, opened 2022-10-12
nuke: fixing opening loader with context if shortcut is used

jakubjezek001:
## Brief description
Nuke is opening loader with context now.

## Description
Expected behaviour was open Loader in context wather it is open form menu or with shortcut. It didn't work.

## Testing notes:
1. Open Nuke in any shot context
2. press combination `CTRL` + `ALT` + `L`
3. it should open Loader in context of the shot


64qam:
The shortcut opens the Loader, but not in the shot context. When launching Loader from OP menu, the works correctly.
image

iLLiCiTiT:
I think it's issue in loader.

64qam:
When Nuke is launched and pressing Ctrl+Alt+L, the Loader is working now (shows subsets in a particular context). However, when you hit the shortcut for the second time, there are no data in Loader (Loader is not opened in the context). But Loader works correctly from OpenPype menu.

jakubjezek001:
This is issue related the QT implementation in loader. A race condition is different when started from Shortcut, then from a menu item. 
We are rewriting Loder for V4, so we will check later if the issue is still occurring.

PR #3973, opened 2022-10-12
Maya: add object to attr.s declaration

kalisp:
## Brief description
Older Maya (Python2) doesn't like declaration of attr.s without (object)

## Description
This handles issue when `Submit Render to Deadline` doesn't show up in Publish and when manually imported in Maya console it throws 
```
# Error: TypeError: file C:\Users\petrk\PycharmProjects\Pype3.0\pype\openpype\vendor\python\python_2\attr\_make.py line 1514: attrs only works with new-style classes. # 
```

## Additional info
Maybe this is not needed as new Maya's (with Python3) might not have an issue with it, but it just happened to me on Maya 2019 and on same day someone from community on Discord reported it too.
https://discord.com/channels/517362899170230292/517382145552154634/1029628902701805609

I don't see it to be too offensive to have object as an argument there for now.

## Testing notes:
1. Publish from Maya to DL
2. `Submit Render to Deadline` should be visible during

### DIFF
--- a/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
+++ b/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
@@ -48,7 +48,7 @@ def _validate_deadline_bool_value(instance, attribute, value):
 
 
 @attr.s
-class MayaPluginInfo:
+class MayaPluginInfo(object):
     SceneFile = attr.ib(default=None)   # Input
     OutputFilePath = attr.ib(default=None)  # Output directory and filename
     OutputFilePrefix = attr.ib(default=None)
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
Looks like duplicate of https://github.com/pypeclub/OpenPype/pull/3968

kalisp:
Duplicate of https://github.com/pypeclub/OpenPype/pull/3968

kalisp:
Actually I don't think empty () is enough, it should be (object) (or at least that was only working for me).

iLLiCiTiT (approved):
I think this is actually proper solution to fix the issue.

antirotor (approved):

PR #3975, opened 2022-10-12
Fix: Inventory action path wrong functions leads to errors

Tilix4:
## Brief description
This is wrong function for what it is supposed to do. It raises an error because a string is tested like a class during the process, which is not the correct behaviour.


BigRoy (approved):
Code changes look good to me.

PR #3979, opened 2022-10-13
Remove staging logic set by OpenPype version

antirotor:
## Refactor

This is removing logic on OpenPype version where staging was determined by `staging` string in version itself.

With  introduction of production/staging options in the Settings this became obsolete and confusing. This PR is removing all places where staging was determined by the version itself. It leaves commandline argumen `--use-staging` to start OpenPype with version specified in **staging** option in the Settings.

Settings handlers are not storing ordered lists of staging/production versions (they were not used anyway).

When `--use-staging` is entered an environment variable `OPENPYPE_USE_STAGING` is set to `"1"` so the staging is enabled for current and future processes even without passing `--use-staging`. Added function which check the value `is_staging_enabled`.

Changed how `is_running_staging` function works. The primary usage is to define icon of openpype tools and since staging is strongly settings related with this PR it also checks if current version is production version -> in case current version is same as production it resolve that it is in production version and standard icon is used -> this also trigger a dialog on tray start.

🔧 **TODO**:

- [x] fix Igniter tests
- [x] test it :)
- [ ] Update documentation


openpype/lib/openpype_version.py
```diff
@@ -60,9 +61,16 @@ def is_running_from_build():
 def is_running_staging():
     """Currently used OpenPype is staging version.
 
+    Deprecated:
+        Since 3.15
+
     Returns:
         bool: True if openpype version containt 'staging'.
     """
+    warnings.warn(
```
iLLiCiTiT, line 70: This function is used to determine what icon is used. We should be able somehow identify if staging is used or not...

antirotor: true but it does it currently from the version. Maybe it should check the staging version set in the settings, compare it with the one requested and if it matches use staging icon?

iLLiCiTiT: But there is no way how we can know that user used `--use-staging` command.

iLLiCiTiT:
Version update check is broken because we don't have way how to know if we should check for production or staging version the same for icon which can be fixed but will probably need more love as it require to use settings in lib and we should maybe the `openpype_version.py` move from lib to `openpype_common`.
Settings are broken because are using old function arguments for `get_remote_versions` (I can fix it).

64qam (approved):

PR #3980, opened 2022-10-13
Ftrack: Fix schema cache for Python 2

iLLiCiTiT:
## Brief description
Python 2 ftrack_api is caching server schema to user's directory so there are not issues with permissions on linux.

## Testing notes:
1. There should not be issues with caching server schema on linux machines


jakubjezek001 (approved):
seems to be working

PR #3981, opened 2022-10-13
Maya: remove invalid prefix token for non-multipart outputs

### LINKED ISSUE #3956: Maya: vray/redshift multichannel exr image output prefix
**Running version**
3.14.2

**Describe the bug**
when using multichannel exr output in vray (6.00.02) wrong image ouptut prefix is applied to deadline job

image

**Expected behavior**
with multichannel exr we should not use `_<aov>` image prefix

there is this check in `renderproduct.py`, but for whatever reason it is not working

[cuID:OP-4214]

antirotor:
## Fix

This is fixing image prefixes for Vray and Redshift to not add `<aov>` token for outputs that are multipart (merged aovs).

### Testing

Submit render to Deadline with merged AOVs from vray or redshift.

Close #3956

### DIFF
--- a/openpype/hosts/maya/api/lib_renderproducts.py
+++ b/openpype/hosts/maya/api/lib_renderproducts.py
@@ -260,20 +260,20 @@ class ARenderProducts:
 
         """
         try:
-            file_prefix_attr = IMAGE_PREFIXES[self.renderer]
+            prefix_attr = IMAGE_PREFIXES[self.renderer]
         except KeyError:
             raise UnsupportedRendererException(
                 "Unsupported renderer {}".format(self.renderer)
             )
 
-        file_prefix = self._get_attr(file_prefix_attr)
+        prefix = self._get_attr(prefix_attr)
 
-        if not file_prefix:
+        if not prefix:
             # Fall back to scene name by default
             log.debug("Image prefix not set, using <Scene>")
             file_prefix = "<Scene>"
 
-        return file_prefix
+        return prefix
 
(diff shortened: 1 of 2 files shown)


m-u-r-p-h-y (changes requested):
tested withm2023.2 and redshift

with multipart exr enabled it still sets `_<aov>` token for the output image 

image

LiborBatek:
Just tested it in with #3954 branch and it works normally (fixes it)....giving it just ```<name>.####.exr`` not nay <AOV> tokens when using multichannel exr format.

Im just not sure if these PRs should be joined together or?

BigRoy:
> Im just not sure if these PRs should be joined together or?

They should, yes.

This one fixes what the expected output is for the renderer in the situations where multipart EXR is enabled.
The other fixes what is passed to Deadline in the case multipart EXR is disabled.

antirotor:
I've merged it together. One PR to rule them all.

m-u-r-p-h-y:
Redshift has actually two multipart options, not very well documented and making the whole thing a true nightmare

image

redshiftOptions.exrForceMultilayer
redshiftOptions.exrMultipart

It seems we only take into account the first one, which is the only way as with all other combinations multipart + individual aov output settings it is nearly impossible to control.

m-u-r-p-h-y (approved):
It is good to go, In case of specific combinations of multipart option and per aov image tokens, user can always fall back with disabling the validator as it is really difficult to account for all possible combinations,

PR #3985, opened 2022-10-14
Fix - tags might be missing on representation

kalisp:
## Brief description
"tags" might be missing on representation, this fixes direct approach

## Description
Solves:
maya_error

## Testing notes:
1. publish review in Maya, shouldn't fail in `integrate_ftrack_component`


m-u-r-p-h-y (approved):
`Integrate Ftrack component` works now

image

PR #3989, opened 2022-10-14
Tests: add support for deadline for automatic tests

kalisp:
## Brief description
First sentence is brief description.

## Description

## Additional info
technical information.

## Testing notes:
1. start with this step
2. follow this step

### DIFF
--- a/tests/integration/hosts/aftereffects/test_deadline_publish_in_aftereffects.py
+++ b/tests/integration/hosts/aftereffects/test_deadline_publish_in_aftereffects.py
@@ -1,23 +1,25 @@
 import logging
 
 from tests.lib.assert_classes import DBAssert
-from tests.integration.hosts.aftereffects.lib import AELocalPublishTestClass
+from tests.integration.hosts.aftereffects.lib import AEDeadlinePublishTestClass
 
 log = logging.getLogger("test_publish_in_aftereffects")
 
 
-class TestPublishInAfterEffects(AELocalPublishTestClass):
-    """Basic test case for publishing in AfterEffects
+class TestDeadlinePublishInAfterEffects(AEDeadlinePublishTestClass):
+    """Basic test case for DL publishing in AfterEffects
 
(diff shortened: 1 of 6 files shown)


iLLiCiTiT:
We should maybe have some function that would return True/False if is in testing mode instead of checking `os.environ.get("IS_TEST")` in each individual piece of code.

openpype/plugins/publish/cleanup.py
```diff
@@ -44,6 +44,9 @@ class CleanUp(pyblish.api.InstancePlugin):
 
     def process(self, instance):
         """Plugin entry point."""
+        if os.environ.get("IS_TEST"):
```
iLLiCiTiT, line 47: There are 2 more cleanup plugins, do we want to skip them too?

tests/integration/hosts/aftereffects/test_deadline_publish_in_aftereffects.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    PERSIST = False
+
+    TEST_FILES = [
+        ("1xhd2ij2ixyjCyTjZgcJEPAIiBHLU1FEY",
+         "test_aftereffects_publish.zip",
+         "")
+    ]
+
+    APP = "aftereffects"
```
iLLiCiTiT, line 37: I would preffer `HOST_NAME` instead of `APP`.

iLLiCiTiT:
I would expect that argument `--automatic_tests` is flag that does not need a value.

kalisp:
> 

How could I highlight that there is an 'automatic_tests' argument available for commands? (It is actually applicable only for some commands, even I cannot see why it shouldn't be on all commands, it just won't make any sense there.)

tests/integration/hosts/aftereffects/test_deadline_publish_in_aftereffects.py
kalisp: Ok, renamed.

PR #3993, opened 2022-10-17
Publisher: Multiselection in card view

iLLiCiTiT:
## Brief description
Card view now supports multiselection of items.

## Description
Card view have multiselection and selectetion between views is consistent (or should be). Card view support Control + click to add/remove selection of single item and Shift+click to add selection from last selected to just selected item.

### Screenshot
image

## Testing notes:
1. Open TrayPublisher
2. Create some instances with different families so there is more groups
3. Play with the selection
4. Switch of view should keep the selection between

Depends on https://github.com/pypeclub/OpenPype/pull/3972


mkolar (approved):
Works nicely. thanks

jakubjezek001 (approved):
Works well!

PR #3994, opened 2022-10-17
TVPaint: Defined with HostBase

iLLiCiTiT:
## Brief description
TVPaint host is defined as `HostBase` class and small sleanup of imports.

## Description
TVPaint is defined and registered as object of `HostBase` class.

## Testing notes:
1. TVPaint should launch and be integrated
2. Creation of all families should work
3. Loading should work


iLLiCiTiT:
Should be fixed now

64qam (approved):

PR #4003, opened 2022-10-19
Renderman support for sample and display filters

moonyuet:
## Brief description
User can set up both sample and display filters in Openpype settings if they are using Renderman as renderer.

## Description
You can preset which sample and display filters for renderman , including the cryptomatte renderpass, in Openpype settings. Once you select which filters to be included in openpype settings and then create render instance for your camera in maya, it would automatically tell the system to generate your selected filters in render settings. 

The place you can find for setting up the filters: _Maya > Render Settings > Renderman Renderer > Display Filters/ Sample Filters_

image

image

## Additional info
I also edit the image_prefix for this ticket so that to unify the render settings a little bit. 
I set up the Image prefix template in the setting as if other renderers does. (see the screenshot of Openpype settings above)


antirotor (changes requested):
I am afraid that this is not complete yet - what it is missing is actually support for output of these filters. They need to be recognized by `lib_renderproducts` - for example:

adding `PxrCryptomatte` filter will add output (with it's own image prefix!) - by default `<imagedir>cryptomatte.<f4>.exr` - this needs to be added to render products so it is then collected and published.

Another thing is that image prefix you've added - that unfortunately works differently in Renderman than in other renderers. You see, Renderman has split image dir and image prefixes itself - setting this image prefix that holds dirs will result in doubling of the structure. THat's why renderman image prefixes are handled in OP somewhat differently.

moonyuet:
> I am afraid that this is not complete yet - what it is missing is actually support for output of these filters. They need to be recognized by `lib_renderproducts` - for example:
> 
> adding `PxrCryptomatte` filter will add output (with it's own image prefix!) - by default `<imagedir>cryptomatte.<f4>.exr` - this needs to be added to render products so it is then collected and published.
> 
> Another thing is that image prefix you've added - that unfortunately works differently in Renderman than in other renderers. You see, Renderman has split image dir and image prefixes itself - setting this image prefix that holds dirs will result in doubling of the structure. THat's why renderman image prefixes are handled in OP somewhat differently.
I have updated the PR request and I find out that the incorrect image prefix and image output directory are the main reason why the layers aren't being rendered correctly. 
The updated PR would add the correct directories for all the maps, including Cryptomatte. I also aligned the name similar to images prefix of other renderers,e.g. Main_beauty.exr.
image

image

antirotor (changes requested):
Sorry I've got to this after while. Setting of the render settings work great, but I think we still need to tweak how the Display and Sample filters are collected in render products. See this scenarion:

image

Simple scene, just one render layer, beauty and cryptomatte. Only beauty is collected though. This will result in correct render, but publishing script won't see cryptomattes.

I guess we need to add logic here to collect and process Display and Sample filter nodes?
https://github.com/ynput/OpenPype/blob/dbc34cc274a03dd02b51bc78fa3a07aee2d5a231/openpype/hosts/maya/api/lib_renderproducts.py#L1127-L1221

moonyuet:
> Sorry I've got to this after while. Setting of the render settings work great, but I think we still need to tweak how the Display and Sample filters are collected in render products. See this scenarion:
> 
> image
> 
> Simple scene, just one render layer, beauty and cryptomatte. Only beauty is collected though. This will result in correct render, but publishing script won't see cryptomattes.
> 
> I guess we need to add logic here to collect and process Display and Sample filter nodes?
> 
> https://github.com/ynput/OpenPype/blob/dbc34cc274a03dd02b51bc78fa3a07aee2d5a231/openpype/hosts/maya/api/lib_renderproducts.py#L1127-L1221

I have added cryptomatte and it will be included in the collect render layer.

antirotor (approved):
I think it works! There is a little issue that cryptomattes are still processed by review for some reason but this is not part of this PR

PR #4006, opened 2022-10-19
Nuke: refactor to latest publisher workfow

jakubjezek001:
## Brief description
Nuke is working with latest publisher workflow

## Description
All Creators plugins needed to be converted to the latest publisher API so they are recognized within the workflow. Also publishing plugins needed to be optimized so the new workflow is catching all as it should. Creator label names had been renamed for better consistency with family and plugin names.  

Openpype menu Create flagged as [depricated], since it had been merged into new publisher. 

## Additional info
- creator plugins had been converted to the new workflow
- Tool for conversion of old publishing instances had been added as instance item "Convert legacy instances"
- publisher plugins are converted to the new workflow 
- It is already having implemented new imageio settings position
- `CreateWriteStill` plugin had been renamed to `CreateWriteImage` to better reflect family use. Still is added as Variant name.

1. Open nuke workfile with most of created instances in old publishing workfow, so you can test the conversion tool
2. Go to Publisher and hit **Update subsets** button on `Convert legacy instances` item
3. You should be able to see instances collected into left frame
4. Hit publish and see what is going to happen ;)

PS: I apologize for the amount of changed files - there was no other way to split this into smaller chunks..


mkolar (changes requested):
Validation of current context in the publishable instance data cannot be repair. The repair button deletes the offending node instead. In case of read node, it crashed nuke completely
image

mkolar (changes requested):
After removing write node using the bin icon in the publisher, I can't re-create it. Use selection is turn on. 

image

iLLiCiTiT:
Nuke is crashing on `add_shortcuts_from_presets` if I have older settings overrides.
```python
try:
    menu = menubar.findItem(MENU_LABEL)
    item_label = menu_label_mapping[command_name] # On this line with KeyError
    menuitem = menu.findItem(item_label)
    menuitem.setShortcut(shortcut_str)
except AttributeError as e:
    log.error(e)
```

iLLiCiTiT:
Creators expect that there are not settings overrides from older versions in project settings.
```
  File "C:\Users\JakubTrllo\Desktop\Prace\openpype3_3\openpype\tools\publisher\window.py", line 363, in _on_show_restart_timer
    self.reset()
  File "C:\Users\JakubTrllo\Desktop\Prace\openpype3_3\openpype\tools\publisher\window.py", line 374, in reset
    self._controller.reset()
  File "C:\Users\JakubTrllo\Desktop\Prace\openpype3_3\openpype\tools\publisher\control.py", line 1680, in reset
    self._reset_plugins()
  File "C:\Users\JakubTrllo\Desktop\Prace\openpype3_3\openpype\tools\publisher\control.py", line 1696, in _reset_plugins
  File "C:\Users\JakubTrllo\Desktop\Prace\openpype3_3\openpype\pipeline\create\context.py", line 1124, in reset_plugins
    self.headless
  File "C:\Users\JakubTrllo\Desktop\Prace\openpype3_3\openpype\pipeline\create\creator_plugins.py", line 85, in __init__
  File "C:\Users\JakubTrllo\Desktop\Prace\openpype3_3\openpype\hosts\nuke\api\plugin.py", line 355, in apply_settings
    "instance_attributes"]
KeyError: 'instance_attributes'
```

openpype/hosts/nuke/api/pipeline.py
```diff
@@ -164,8 +226,13 @@ def _install_menu():
 
     menu.addSeparator()
     menu.addCommand(
-        "Create...",
-        lambda: host_tools.show_creator(parent=main_window)
+        "Create... [depricated]",
```
mkolar, line 229:
```suggestion
        "Create... [deprecated]",
```

openpype/hosts/nuke/api/pipeline.py
```diff
@@ -164,8 +226,13 @@ def _install_menu():
...
+        lambda: nuke.message(
+            "New position is in \"Publish...\" and \"Create\" tab")
```
mkolar, line 231:
```suggestion
            "Creator has been moved, you'll find it in the Publish tool -> Create tab")
```

openpype/hosts/nuke/plugins/publish/collect_workfile.py
```diff
@@ -0,0 +1,40 @@
+import os
+import nuke
+import pyblish.api
+
+
+class CollectWorkfile(pyblish.api.InstancePlugin):
+    """Collect current script for publish."""
+
+    order = pyblish.api.CollectorOrder - 0.498
```
mkolar, line 9: I'm a bit worried that all the collectors in Nuke need such explicit ordering. Is that still needed considering how the pre-collection has changed with the new publisher?

mkolar (changes requested):
Testing further, but a few small comments.

openpype/hosts/nuke/plugins/publish/collect_workfile.py
jakubjezek001: I think it is not necessary anymore on workfile collector.

jrsndl:
- Fails for me on collect slate node:

```
  File "C:\Users\robur\OpenPype\openpype\hosts\nuke\plugins\publish\collect_slate_node.py", line 38, in process
KeyError: 'families'
```

- Workafile that was named workfileComp (task comp) before is now named workfileMain. I assume that is by design.

- After conversion, the render and prerender nodes do not have the publishing knobs

render and prerender before conversion:
image

render aqnd prerender after conversion:
image

Tested with Nuke 13.2v4, win10

jakubjezek001:
>  Fails for me on collect slate node:

Thank you for report. This is simple fix

>  I assume that is by design.

Yes that is correct.

openpype/hosts/nuke/api/pipeline.py
```diff
@@ -1,21 +1,24 @@
...
-import nuke
-
 import pyblish.api
 
 import openpype
+from openpype.host import (
+    HostBase,
+    IWorkfileHost,
+    ILoadHost,
+    INewPublisher
```
iLLiCiTiT, line 14: Class `INewPublisher` was renamed to `IPublishHost`.

openpype/hosts/nuke/api/pipeline.py
```diff
@@ -58,6 +77,95 @@
     pyblish.api.register_gui(os.getenv("PYBLISH_GUI", None))
 
 
+class NukeHost(
+    HostBase, IWorkfileHost, ILoadHost, INewPublisher
```
iLLiCiTiT, line 81:
```suggestion
    HostBase, IWorkfileHost, ILoadHost, IPublishHost
```

openpype/hosts/nuke/api/pipeline.py
iLLiCiTiT:
```suggestion
    IPublishHost
```

jakubjezek001:
This is ready for merge

jrsndl (commented):
I've tested making and publishing camera, gizmo, prerender, render, still and workfile in Nuke 13.2v4. All fine, just two minor troubles.

**Optional Validate Version not available**
I have Validate Version set to optional and active in project_settings/global/publish/ValidateVersion, but there seem to be no way to temporarily disable validate version in new publisher.

**Still frame**
First I had trouble creating still instance. Renaming CreateWriteStill in Nuke imageio to CreateWriteImage fixed that. But I am not able to publish an image. Possibly just a setting problem?

  File "C:\Users\robur\OpenPype\.venv\lib\site-packages\pyblish\plugin.py", line 522, in __explicit_process
    runner(*args)
  File "C:\Users\robur\OpenPype\openpype\plugins\publish\integrate.py", line 161, in process
  File "C:\Users\robur\OpenPype\openpype\plugins\publish\integrate.py", line 225, in filter_representations
openpype.pipeline.publish.publish_plugins.KnownPublishError: Instance image has no representations to integrate_

I had to roll back [9522607], I assume it is just my git ignorance.

openpype/hosts/nuke/plugins/publish/collect_context_data.py
```diff
@@ -0,0 +1,69 @@
...
+            "handles": handle_start,
+            "handleStart": handle_start,
+            "handleEnd": handle_end,
+            "step": 1,
+            "fps": root_node['fps'].value(),
+
+            "currentFile": current_file,
+            "version": int(api.get_version_from_path(current_file)),
+
+            "host": pyblish.api.current_host(),
```
iLLiCiTiT, line 62: Should this be taken from `pyblish.api`?

64qam:
Also, an option for farm rendering is missing

jrsndl:
>

jrsndl:
> Also, an option for farm rendering is missing
Setting up the deadline URL in the project settings helped me to get the farm rendering options.
(project_settings/deadline/publish)
image

jrsndl (approved):
### Converting Instances
Nuke 12.2 and Nuke 13.2 correctly detected and converted legacy instances.

### Still publish
I was not able to create an image (still) publish from newly created or converted instance. 

### Optional Validate Version not available
I have Validate Version set to optional and active in project_settings/global/publish/ValidateVersion, but there seem to be no way to temporarily disable validate version in new publisher.

### Intent
There seem to be no way to set the Publish Intent, but you can Validate the Intent in project_settings/global/publish/ValidateIntent and then the publish fails.

openpype/hosts/nuke/api/plugin.py
```diff
@@ -1,27 +1,381 @@
...
+            project_name, instance_data["asset"], fields=["_id"]
+        )
+        subset_docs = get_subsets(
+            project_name, asset_ids=[asset_doc["_id"]], fields=["name"]
+        )
+        existing_subset_names_low = {
+            subset_doc["name"].lower()
+            for subset_doc in subset_docs
+        }
+        return subset_name.lower() in existing_subset_names_low
```
BigRoy, line 111: This seems like an overly verbose check. If a new nuke script is created after an initial publish and someone intentionally tries to publish into an existing entry they can't create it. Similary if somehow the output node got deleted in their nuke script and they want to recreate it again, they can't. 

Even someone else from another task in the same assets couldn't even try to do so? (or maybe they could if it was prefixed with the task actually; not sure if that's the case).

I feel like a check whether the current Nuke script already has an instance publishing to the same asset + subset name would be a decent validation but the current 'check' seems quite blocking for what can definitely be a production use case.

BigRoy: > We should only be checking if a node with particular subset name is not already created in workfile nodes, like you've said.

The next question is - is there ever a case where that might need to be allowed (outside of these creators, outside of the realm of Nuke). If not, then maybe should the publishing system/Creators in general already perform a check like that so that developers don't need to recreate that logic for each individual Creator plugin?

jakubjezek001: I presume this is more question for @iLLiCiTiT

iLLiCiTiT: To be honest I don't think it's job of creator to check it until it's required for his functionality. This looks like a validation before validators.

iLLiCiTiT: BTW you can change the context and variant after creation and where the `create` would not be called -> as the validation.

PR #4009, opened 2022-10-19
Nuke | Global: adding custom tags representation filtering

jakubjezek001:
## Brief description
Representation can be filtered by custom tagging in extract review

## Description
Extract review was filtering only on the publishing instance level and not able to aim directly at particular representation. We have to implement filtering method to be able to activate output preset only to particular representation.

## Additional info
- adding `custom_tag` as optional key to representation
- adding `custom_tags` to nuke extract review data movie plugin
- improving nuke's way of handling viewerProcess to include also display selector which is now pars-able from a string. The syntax is as follows:
	- add string input to `project_settings/nuke/publish/ExtractReviewDataMov/outputs/baking/viewer_process_override` in case you want to add bake stream overide of `project_settings/nuke/imageio/baking/viewerProcess` (notice we hade moved imagio from Anatomy settings to project settings)
	- syntax is either: `viewer_name (display_name)` or `display_name/viewer_name`
	- examples are: `sRGB (ACES)` or `ACES/sRGB`

## Matching mechanics:
image

## Testing notes:
1. Open Studio settings and go to your project settings
2. go to `project_settings/nuke/publish/ExtractReviewDataMov/outputs`
3. Add one more baking stream and set its Add custom tag to some value
4. Go to `project_settings/global/publish/ExtractReview/profiles`
5. Create a new profile, add it at first, set the host input to `nuke`
6. Add two output profiles > it could be h264 ffmpeg args copied form default preset. Set names to `h264first` and `h264first` two the same profiles, except one of them is having set Additional filtering to `custom tags` with the same input from step 3.
7. In the same project open Nuke and Create render write node.
8. Publish and notice that preset  defined filter on custom tag was run only once.

### DIFF
--- a/openpype/plugins/publish/extract_review.py
+++ b/openpype/plugins/publish/extract_review.py
@@ -128,6 +128,7 @@ class ExtractReview(pyblish.api.InstancePlugin):
         for repre in instance.data["representations"]:
             repre_name = str(repre.get("name"))
             tags = repre.get("tags") or []
+            custom_tags = repre.get("custom_tags")
             if "review" not in tags:
                 self.log.debug((
                     "Repre: {} - Didn't found \"review\" in tags. Skipping"
@@ -158,15 +159,18 @@ class ExtractReview(pyblish.api.InstancePlugin):
                 )
                 continue
 
-            # Filter output definition by representation tags (optional)
-            outputs = self.filter_outputs_by_tags(profile_outputs, tags)
+            # Filter output definition by representation's
+            # custom tags (optional)
+            outputs = self.filter_outputs_by_custom_tags(
+                profile_outputs, custom_tags)
             if not outputs:
                 self.log.info((
                     "Skipped representation. All output definitions from"
                     " selected profile does not match to representation's"
-                    " tags. \"{}\""
+                    " custom tags. \"{}\""
                 ).format(str(tags)))
                 continue
+
             outputs_per_representations.append((repre, outputs))
         return outputs_per_representations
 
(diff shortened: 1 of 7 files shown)


openpype/plugins/publish/extract_review.py
```diff
@@ -158,15 +159,18 @@ def _get_outputs_per_representations(self, instance, profile_outputs):
                 )
                 continue
 
-            # Filter output definition by representation tags (optional)
-            outputs = self.filter_outputs_by_tags(profile_outputs, tags)
+            # Filter output definition by representation's
+            # custom tags (optional)
+            outputs = self.filter_outputs_by_custom_tags(
```
iLLiCiTiT, line 164: I think we should still keep the filtering by tags.

openpype/plugins/publish/extract_review.py
```diff
@@ -1711,39 +1718,55 @@ def filter_output_defs(self, profile, subset_name, families):
...
+            # check if any repre tag is not in filter tags
+            for tag in repre_c_tags_low:
+                if tag in tag_filters_low:
+                    valid = True
+                    break
+
+            if not valid:
+                continue
+
+            filtered_outputs.append(output_def)
```
iLLiCiTiT, line 1765: Removed unnecessary loggers and do tags filtering only if they are set.
```suggestion
            else:
                # lower all filter tags
                tag_filters_low = [tag.lower() for tag in tag_filters]

                # check if any repre tag is not in filter tags
                for tag in repre_c_tags_low:
                    if tag in tag_filters_low:
                        valid = True
                        break

            if valid:
                filtered_outputs.append(output_def)
```

jrsndl (changes requested):
I tested the tags, and publish crashes on FFprobe in Extract Review. The same Shot Publishes OK on latest Develop.

02_MovExists
/45896205/197170720-a76fde38-e8b3-4dc6-b936-f5be251ab295.png)
01_FFprobeError

openpype/plugins/publish/extract_review.py
iLLiCiTiT: This will fix this issue too https://github.com/pypeclub/OpenPype/pull/4009#pullrequestreview-1150699781

64qam (approved):

jrsndl (approved):
The error above is not relevant to this PR. Sorry

I did more tests: with Nuke 12.2 and Nuke 13.2. With and without slate, with different burnin combinations.

PR #4015, opened 2022-10-21
Publisher: Catch creator errors

iLLiCiTiT:
## Brief description
Handle failed operations of creator plugins in Publihser UI to safe-guard failed state.

## Description
If creator failed during one of their operations the UI could stuck without any way how to rollback and user have no idea that something went wrong. For this purposes all oprations on create context raise special exceptions. Exception hold information about each failed creator with error message and optionally traceback. This can be showed in UI as a dialog. The dialog shows a tabs per each creator which failed. Exceptions are handled on collect, update, remove and create. Create widget does not handle exceptions on it's own because it would not be possible in remotelly controllerd UI instead a controller trigger events when dialog should be shown.

Reset of controller does not trigger save automatically. That is to prevent unwanted trigger e.g. when window is reopened, the previous state should not be saved because was already saved and the context could change meanwhile.

Fixed few smaller bugs. Tab `Details` crashed if was selected before publishing started. Animations were not working in older Qt bindings.

## Additional info
There is one remaining issue. We don't handle change of context/workfile in hosts. We should add that option in future.

The messages are really simple right now we could change them to something understandable by both users and developers. The raised error message should maybe contain information about all creators. Also the error dialog does not have title because it looked wierd above the tabs, could be added by changing the background of the title widget to same color as tabs have in empty space.

## Screenshots
image
image
image

## Testing notes:
1. Crashed created, collect, update, remove should show error dialog with information
2. Animations should work in Nuke

### DIFF
--- a/openpype/pipeline/create/context.py
+++ b/openpype/pipeline/create/context.py
@@ -1,6 +1,8 @@
 import os
+import sys
 import copy
 import logging
+import traceback
 import collections
 import inspect
 from uuid import uuid4
@@ -22,6 +24,7 @@ from .creator_plugins import (
     Creator,
     AutoCreator,
     discover_creator_plugins,
+    CreatorError,
 )
 
 UpdateData = collections.namedtuple("UpdateData", ["instance", "changes"])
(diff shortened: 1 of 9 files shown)


antirotor (approved):
It's throwing error dialogs at me so I guess it's working!

PR #4016, opened 2022-10-21
Houdini: fix wrong path in ASS loader

antirotor:
## Fix

ASS loader was filling in only file name, not the whole path. This is fixing the problem.

### Testing

Load ASS file with OpenPype in Houdini, everything should work as expected.

### DIFF
--- a/openpype/hosts/houdini/plugins/load/load_ass.py
+++ b/openpype/hosts/houdini/plugins/load/load_ass.py
@@ -32,7 +32,12 @@ class AssLoader(load.LoaderPlugin):
 
         # Create a new geo node
         procedural = obj.createNode("arnold::procedural", node_name=node_name)
-        procedural.setParms({"ar_filename": self.get_path(self.fname)})
+
+        procedural.setParms(
+            {
+                "ar_filename": self.format_path(
+                    self.fname, context["representation"])
+            })
 
         nodes = [procedural]
         self[:] = nodes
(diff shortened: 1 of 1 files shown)


openpype/hosts/houdini/plugins/load/load_ass.py
```diff
@@ -105,3 +67,23 @@ def remove(self, container):
...
+    @staticmethod
+    def format_path(path, representation):
+        """Format file path correctly for single bgeo or bgeo sequence."""
+        if not os.path.exists(path):
+            raise RuntimeError("Path does not exist: %s" % path)
+
+        is_sequence = bool(representation["context"].get("frame"))
+        # The path is either a single file or sequence in a folder.
+        if not is_sequence:
+            filename = path
```
iLLiCiTiT, line 80: Ridiculous question: Is it a filename or filepath?

openpype/hosts/houdini/plugins/load/load_ass.py
```diff
@@ -105,3 +67,23 @@ def remove(self, container):
...
+        else:
+            filename = re.sub(r"(.*)\.(\d+)\.(ass.*)", "\\1.$F4.\\3", path)
+
+            filename = os.path.join(path, filename)
```
iLLiCiTiT, line 84: Shouldn't it be like this?
```suggestion
            filename = os.path.join(os.path.dirname(path), filename)
```

iLLiCiTiT: Or to completelly simplify the function just
```
if is_sequence:
    dirpath, filename = os.path.split(path)
    new_filename = re.sub(r"(.*)\.(\d+)\.(ass.*)", "\\1.$F4.\\3", filename)
    path = os.path.join(dirpath, new_filename )

return os.path.normpath(path).replace("\\", "/")
```

iLLiCiTiT (approved):

mkolar:
@antirotor maybe we could add this request into the PR? https://discord.com/channels/517362899170230292/1035450698655088661/1035450701045846026

PR #4020, opened 2022-10-21
Publisher: Convertors for legacy instances

iLLiCiTiT:
## Brief description
Give ability to have convertor plugins for conversion from legacy instances to new creators.

## Description
Added Convertor plugins which can collect `ConvertorItem` into `CreateContext`. These items are shown in UI next to instances with a label that the convertor plugin defines. Convertor collect items after Creator's collection so there should not be a clash if any creator would convert something automatically and also has access to shared data. Has 2 methods `find_instances` and `convert` both without any arguments. Plugin must be in files next to creators (can be in file with creator). In UI is conversion button always before creator attributes. So multiple conversion plugins can be selected and UI does not break when instances are selected too.

## Screenshots
image
image

## Testing notes:
Create fake convertor in TrayPublisher e.g. `~/openpype/hosts/traypublisher/plugins/create/converter_test.py` with this content.
```python
from openpype.pipeline.create.creator_plugins import SubsetConvertorPlugin


class TestConverted(SubsetConvertorPlugin):
    identifier = "text.converter"

    def find_instances(self):
        # This is called during collection - plugin should create it's item based on some logic
        #  - for example find an instance without 'creator_identifier'
        self.add_convertor_item("My testing legacy item")

    def convert(self):
        # This is method which is triggered from the UI
        # - should update items for new creators and remove item if was successfull
        print("Fake converting. This is when would logic of conversion be.")
        self.remove_convertor_item()
```

### DIFF
--- a/openpype/pipeline/create/context.py
+++ b/openpype/pipeline/create/context.py
@@ -1195,9 +1195,7 @@ class CreateContext:
                 ))
                 continue
 
-            legacy_convertors[convertor_identifier] = (
-                convertor_identifier(self)
-            )
+            legacy_convertors[convertor_identifier] = convertor_class(self)
 
         self.legacy_convertors = legacy_convertors
 
--- a/openpype/pipeline/create/context.py
+++ b/openpype/pipeline/create/context.py
@@ -22,6 +22,7 @@ from .creator_plugins import (
     Creator,
     AutoCreator,
     discover_creator_plugins,
+    discover_legacy_convertor_plugins,
 )
 
 UpdateData = collections.namedtuple("UpdateData", ["instance", "changes"])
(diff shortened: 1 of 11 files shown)


antirotor:
Seems like it is working but I would add two things:

1) Reset publisher after conversion is done
2) add log to the converter (to access `self.log.*`)

antirotor (changes requested):
Seems like it is working but I would add two things:

1) Reset publisher after conversion is done
2) add log to the converter (to access `self.log.*`)

iLLiCiTiT:
> Seems like it is working but I would add two things:

Both should be available now.

antirotor (approved):

PR #4021, opened 2022-10-24
Maya: add OBJ extractor to model family

mkolar:
## Brief description
Adding a new extractor with its corresponding settings to allow exporting OBJ format from model. It is turned off by default. 

## Testing notes:
1. In the OP settings go to `project_settings/maya/publish/ExtractObj`  and enable it. 
2. Open a scene with publishable model in maya
3. Publish the model
4. check if OBJ was correctly created and that it can be imported into maya from OP loader


m-u-r-p-h-y (approved):
obj extractor works as expected.

Maya2023

image

PR #4026, opened 2022-10-25
General: More specific error in burnins script

iLLiCiTiT:
## Brief description
Raise more specific error in burnins script when input does not have streams before otio ffprobe is called.

## Description
It would crash anyway but with cryptic error that subprocess failed to run in otio adapters. This can happen when input does not have any streams (invalid movie file).

## Additional information
Not sure how the movie file can be invalid (seems like `skip-handles` tag on single frame can cause it?). Issue created https://github.com/pypeclub/OpenPype/issues/4028

## Testing notes:
1. Invalid input movie file in burnins should cause error with more specific information.


m-u-r-p-h-y (approved):
error output is more informative

image

PR #4037, opened 2022-10-27
Nuke - loader clip correct hash conversion in path

jakubjezek001:
## Brief description
Correctly converted hashes in path for loaded clip

## Description
technical information.

## Testing notes:
1. To test this you would need to set project name with use of first image sequence frame. Very unlikely. But even that can happen. This hash conversion is only used no on file name and not at full patch so it is enough just to check the code for logic.

### DIFF
--- a/openpype/hosts/nuke/plugins/load/load_clip.py
+++ b/openpype/hosts/nuke/plugins/load/load_clip.py
@@ -1,3 +1,4 @@
+import os
 import nuke
 import qargparse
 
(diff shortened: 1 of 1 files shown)


openpype/hosts/nuke/plugins/load/load_clip.py
```diff
@@ -84,14 +85,24 @@ def get_representations(cls):
             + plugin.get_review_presets_config()
         )
 
+    def _fix_path_for_knob(self, filepath, repre_cont):
+        basename = os.path.basename(filepath)
+        dirname = os.path.dirname(filepath)
+        frame = repre_cont.get("frame")
+        assert frame, "Representation is not sequence"
+
+        padding = len(frame)
```
iLLiCiTiT, line 94:
```suggestion
        padding = len(str(frame))
```

PR #4042, opened 2022-10-28
Publisher: Add thumbnail sources

iLLiCiTiT:
## Brief description
Added option to enter a thumbnail in Publisher UI.

## Description
Publisher UI has thumbnail widget in create and publish pages. By dropping image/video source file to the area is created thumbnail. In Publish view is the thumbnail assigned to all selected instances (or context) and in create view is passed to `create` method with pre create data. Creators can enable/disable option to drop thumbnail before creation which is disabled by default -> this is new feature which is not ready to use by default and creators must be ready for that. Thumbnails are kept until publisher UI is closed. At that moment are thumbnails removed (only thumbnails created by Publisher).

Thumbnails are stored to `CreatedContext` because it's only place that can be accessed in UI and during publishing. They're stored by instance id and each value should have path to a file.

It is possible to set thumbnail on context. In that case is the context thumbnail used as default for all instances. This logic is handled in integration part. Thumbnail integrator was changed to context plugin without any family filters.

## Additional information
Publisher UI can't know which instances after creation should have the thumbnail because creator can create 0-∞ instances.
Dropped files are converted using oiiotool or ffmpeg. If the conversion cannot be done an error message is shown.

For precreate data was prepared constant `PRE_CREATE_THUMBNAIL_KEY` under which is stored the path to thumbnail during creation. Creator (if want's to use the value) should call `set_instance_thumbnail_path` with instance id and thumbnail path where the thumbnail should be created.

Hover color of buttons is now defined with RGB instead of RGBA (alpha did some bad behavior).

## Testing notes:
1. Open TrayPublisher
2. Select one of creators which support thumbnail drop (e.g. Image)
3. Fill representation files (Do not fill review representation)
4. Drop the thumbnail to widget
5. Create instance
6. Hit publish
7. Check if thumbnail is integrated

The UI should be tested in Nuke and Houdini as the Thumbnail logic may not show right way.


64qam (approved):

iLLiCiTiT:
image
image

Changed background color of thumbnails widget to look like input (with dashes). Added checkerboard under transparent images (only under their size). Added clear button to "remove" thumbnail -> has same backgroung color as thumbnails widget (dark) and icon is selected card color (blue). The default image is scaled down.

mkolar (dismissed):
Looking good. We should however deal with the thumbnail integrator. it currently only triggers on review, which is really confusing with this update.

jakubjezek001 (changes requested):
If a reload button is pushed at publisher a previously dropped thumbnail is lost and it needs to be dropped again.

iLLiCiTiT:
> If a reload button is pushed at publisher a previously dropped thumbnail is lost and it needs to be dropped again.

Should be fixed

jakubjezek001 (approved):

PR #4045, opened 2022-10-31
Feature/3delight2

### LINKED ISSUE #2883: 3Delight Support for Maya
The OpenPype doesn't have support for 3Delight render.

Add render support for 3Delight.

This is additive, nothing alternative.

**Additional context**
This is needed by us, and we'd develop it and create PR for this feature.

dmo-j-cube:
## Brief description
Adding 3delight render support to OpenPype

## Description
Adding support to render via Maya on 3delight using the Deadline service. This PR adds the necessary structures and changes within OP to invoke rendering. 

Given that Deadline hasn't updated support for 3delight since 3delight switched to NSI, we had to add support for "render layers", via 3delight's dlRenderSettings via a pre-render MEL command, and also startFrame and endFrame too. This uses maya's render layers just for their names to map into the dlRenderSetting to use.

## Additional info
Given that 3delight uses dlRenderSettings instead of Maya's Render Layers to manage all rendering information, this poses an issue whenever a change in OP's way of handling renders changes. I have tried to get OP to ignore settings and valiations for 3delight when it's active, and hopefully no more changes will be needed.

Resolves #2883


antirotor (approved):
From my point of view it looks great. There are points that will need to be addressed later on during the switch to new publisher but right now it's fine. We'll need to test it functionally and of course.

m-u-r-p-h-y (commented):
I was testing Maya2022 with 3delight 2.9.14

- when creating a render instance, `dlRenderSettings1_RL` is created and correctly parented under the instance

- default path for render image is not following other renderers logic (we should add some 3delight render defaults to settings)

- default frame  `-#. ` token should be changed to `.#.`

after fixing these little things I was able to publish the scene and submit it to deadline

image


- rendering seems working, but I was not able to find the rendered files anywhere (is it possible to increase the verbosity of 3delight render?)

image


image

antirotor:
Resolves #2883

antirotor:
Closing in favour of #4319 which is the same but with resolved conflicts.

PR #4046, opened 2022-10-31
Feature/multiverse

dmo-j-cube:
## Brief description
Updating support for Multiverse/USD.

## Description
Adding OP-native support to USD Overrides - they load upon an existing USD
asset in the scene as a layer and they maintain layer order when switching
or updating. Also, when updating USD assets, if they have overrides added,
they will be maintained.

Fixed look collection when the nodes have mixed-type attributes - will skip
over them instead of crashing.

Texture support for mvLook has been updated and is more robust.

Updated Multiverse USD publish defaults to more sensible values.

Multiverse plugin loading is now more robust.

USD files are no longer `mvUsd`, just standard `usd`.


mkolar:
This is looking good in general, I'd like a review from @antirotor though on the code side and we should check the user side as well @m-u-r-p-h-y

antirotor (approved):

antirotor (approved):

PR #4047, opened 2022-10-31
Alembic Loader as Arnold Standin

moonyuet:
## Brief description
Adding Alembic Loader as Arnold Standin function as the Openpype Plugin

## Description
The user can load the published workfileswith abc format as Arnold Standin if the pusblished workfiles belong to model and pointcache family.
image
image

## Additional info
Doesn't have any for the current stage

## Testing notes:
1. Start maya in Openpype
2. Click "Load" from Openpype in maya
3. Right-click the workfiles which belongs to either model or pointcache family
4. Click Import Alembic as Standin
5. It will load the abc file as an arnold standin

### DIFF
--- a/openpype/hosts/maya/plugins/load/load_abc_to_standin.py
+++ b/openpype/hosts/maya/plugins/load/load_abc_to_standin.py
@@ -50,9 +50,9 @@ class AlembicStandinLoader(load.LoaderPlugin):
         if c is not None:
             cmds.setAttr(root + ".useOutlinerColor", 1)
             cmds.setAttr(root + ".outlinerColor",
-                (float(c[0])/255),
-                (float(c[1])/255),
-                (float(c[2])/255)
+                        (float(c[0])/255),
+                        (float(c[1])/255),
+                        (float(c[2])/255)
             )
 
         transform_name = label + "_ABC"
--- a/openpype/hosts/maya/plugins/load/load_abc_to_standin.py
+++ b/openpype/hosts/maya/plugins/load/load_abc_to_standin.py
@@ -98,6 +98,7 @@ class AlembicStandinLoader(load.LoaderPlugin):
         # Update the standin
         standins = list()
         members = pm.sets(container['objectName'], query=True)
+        self.log.info("container:{}".format(container))
         for member in members:
             shape = member.getShape()
             if (shape and shape.type() == "aiStandIn"):
@@ -105,8 +106,11 @@ class AlembicStandinLoader(load.LoaderPlugin):
 
         for standin in standins:
             standin.dso.set(path)
-            standin.useFrameExtension.set(0)
             standin.abcFPS.set(float(fps))
+            if "modelMain" in container['objectName']:
+                standin.useFrameExtension.set(0)
+            else:
+                standin.useFrameExtension.set(1)
 
         container = pm.PyNode(container["objectName"])
         container.representation.set(str(representation["_id"]))
--- a/openpype/hosts/maya/plugins/load/load_abc_to_standin.py
+++ b/openpype/hosts/maya/plugins/load/load_abc_to_standin.py
@@ -55,7 +55,8 @@ class AlembicStandinLoader(load.LoaderPlugin):
         transform_name = label + "_ABC"
 
         standinShape = cmds.ls(mtoa.ui.arnoldmenu.createStandIn())[0]
-        standin = cmds.listRelatives(standinShape, parent=True, typ="transform")
+        standin = cmds.listRelatives(standinShape, parent=True,
+                                     typ="transform")
         standin = cmds.rename(standin, transform_name)
         standinShape = cmds.listRelatives(standin, children=True)[0]
 
--- a/openpype/hosts/maya/plugins/load/load_abc_to_standin.py
+++ b/openpype/hosts/maya/plugins/load/load_abc_to_standin.py
@@ -50,10 +50,10 @@ class AlembicStandinLoader(load.LoaderPlugin):
         if c is not None:
             cmds.setAttr(root + ".useOutlinerColor", 1)
             cmds.setAttr(root + ".outlinerColor",
-                        (float(c[0])/255),
-                        (float(c[1])/255),
-                        (float(c[2])/255)
-            )
+                         (float(c[0])/255),
+                         (float(c[1])/255),
+                         (float(c[2])/255)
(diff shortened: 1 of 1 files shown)


openpype/hosts/maya/plugins/load/load_abc_to_standin.py
```diff
@@ -0,0 +1,114 @@
...
+from openpype.settings import get_project_settings
+
+
+class AlembicStandinLoader(load.LoaderPlugin):
+    """Load Alembic as Arnold Standin"""
+
+    families = ["model", "pointcache"]
+    representations = ["abc"]
+
+    label = "Import Alembic as Standin"
```
BigRoy, line 16: I feel like this label should at least mention Arnold.

openpype/hosts/maya/plugins/load/load_abc_to_standin.py
```diff
@@ -0,0 +1,114 @@
...
+    order = -5
+    icon = "code-fork"
+    color = "orange"
+
+    def load(self, context, name, namespace, options):
+
+        import maya.cmds as cmds
+        import pymel.core as pm
```
BigRoy, line 24: With issue https://github.com/pypeclub/OpenPype/issues/2353 open I'd recommend avoiding Pymel where possible and rewriting this using maya cmds

moonyuet: I agree. I also noticed another ArnoldStandin Loader would have an error message of `Error: Problem with the API object returned by __apiobject__ method` when loading ass file under some circumstance. I guess it's related to the pymel command too.

LiborBatek (changes requested):
I ve tested it ```maya2023``` with ```arnold 5.1.3.3```

Model imports as ASS file normally. Small note tho: use sequence is "ON" should be considered that model being static in general so should be "OFF" by nature

Speaking of Point cache at first everything seems fine and working however I ve got weird anim problems...double checked it and manually created ass.file from the same workfile and my version had no issues. 
Seems like some ASS creation params being wrong set. Need revision and testing.
Same frame different animation:
image

One more thing to consider why not allow user to "Import Alembic as ASS file" even when speaking about Animation caches not just Pointcaches in general (pyblished animation alembics) could be pretty handy and also there shouldnt be any reason to not allow it too...
so user could import these families as ASS:

```model``` (already present)
```pointcache``` (already present)
```animation``` (not present now)

moonyuet:
> I ve tested it `maya2023` with `arnold 5.1.3.3`
> 
> Model imports as ASS file normally. Small note tho: use sequence is "ON" should be considered that model being static in general so should be "OFF" by nature
> 
> Speaking of Point cache at first everything seems fine and working however I ve got weird anim problems...double checked it and manually created ass.file from the same workfile and my version had no issues. Seems like some ASS creation params being wrong set. Need revision and testing. Same frame different animation: image
> 
> One more thing to consider why not allow user to "Import Alembic as ASS file" even when speaking about Animation caches not just Pointcaches in general (pyblished animation alembics) could be pretty handy and also there shouldnt be any reason to not allow it too... so user could import these families as ASS:
> 
> `model` (already present) `pointcache` (already present) `animation` (not present now)

One of the possible reasons for the weird anim problem could be related to the fps of the animation. I find out the standin import the abc animation with 24 fps while the standard project setting is 25fps.  The abc fps is set up to 25 in the updated PR.
image
The updated PR also tells the systems not to load frame if the abc file doesn't contain any animation. And also I added the animation into families.

LiborBatek:
Now it works normally, did tests in ```maya2022``` and ```maya2023```. 

I think the problem with animation mismatch was caused by difference in framerates (25fps vs 29.97) even present in pyblished abc using OP tools. So it was not on your side!

Question is if special care should be taken. I mean taking note of fps from origin of the abc file (source animation maya scene). As I understand you read this from database and set it according to it right?

Also static models still have ```is sequence``` turned on in ASS properties. Should be addressed already right?

Thanks!

moonyuet:
> Now it works normally, did tests in `maya2022` and `maya2023`.
> 
> I think the problem with animation mismatch was caused by difference in framerates (25fps vs 29.97) even present in pyblished abc using OP tools. So it was not on your side!
> 
> Question is if special care should be taken. I mean taking note of fps from origin of the abc file (source animation maya scene). As I understand you read this from database and set it according to it right?
> 
> Also static models still have `is sequence` turned on in ASS properties. Should be addressed already right?
> 
> Thanks!

image
I think most of the issues I have discussed with you in discord. The abc file published from the model main wont have the sequence on as it doesn't contain any animation. So yes. the static models should be addressed already.

LiborBatek:
Made another go with testing...the static model like my table prop still gets the ```sequence``` option on. Also would be great if the loaded/imported ass files keep color coding in outliner (see img with outliner colors applied when assets loaded normally via asset load command using OP loader)
it means family animation being green colored in outliner and model dark orange color. Each family has its own.
image

moonyuet:
> Made another go with testing...the static model like my table prop still gets the `sequence` option on. Also would be great if the loaded/imported ass files keep color coding in outliner (see img with outliner colors applied when assets loaded normally via asset load command using OP loader) it means family animation being green colored in outliner and model dark orange color. Each family has its own. image
Hi I have made the outlinecolor for the group_name and I have set up the condition for not loading the sequence in model main. 
image

LiborBatek:
Have made another testing and all the features are already working fine! 

There is one thing which prevents the approval...
I have tried to fiddle with switching versions of loaded asset...first I must say that its great that user have this option!!! Its super useful and I didnt expect this would be possible (switching ass files versions!)

There is really minor issue that is ```Use File Sequence``` being unchecked afterwards even it was turned on before switching version of the asset.

image

Otherwise all works perfectly!

moonyuet:
> Have made another testing and all the features are already working fine!
> 
> There is one thing which prevents the approval... I have tried to fiddle with switching versions of loaded asset...first I must say that its great that user have this option!!! Its super useful and I didnt expect this would be possible (switching ass files versions!)
> 
> There is really minor issue that is `Use File Sequence` being unchecked afterwards even it was turned on before switching version of the asset.
> 
> ass_loading_issue
> 
> Otherwise all works perfectly!

Hi Libor. I think you can try with the latest code. it will update use file sequence according to which family subset it belongs to.
image

antirotor (approved):

PR #4049, opened 2022-11-01
Feature: Auto use last published workfile as first workfile

Tilix4:
## Brief description
Add a `PreLaunchHook` and a setting to use last published workfile as first workfile.

Capture d’écran 2022-11-01 à 10 08 21

## Description
1- Check if setting for this feature is enabled

It recovers Tray's `ModulesManager` and find initialized `sync_server` to access its `sync_server_thread`. An `Event` has been added to the `Thread` to be able to wait the reset loop has complete downloading.

### DIFF
--- a/openpype/hooks/pre_copy_last_published_workfile.py
+++ b/openpype/hooks/pre_copy_last_published_workfile.py
@@ -7,8 +7,10 @@ from openpype.client.entities import (
     get_subsets,
 )
 from openpype.lib import PreLaunchHook
+from openpype.lib.profiles_filtering import filter_profiles
 from openpype.modules.base import ModulesManager
 from openpype.pipeline.load.utils import get_representation_path
+from openpype.settings.lib import get_project_settings
 
 
 class CopyLastPublishedWorkfile(PreLaunchHook):
@@ -32,9 +34,45 @@ class CopyLastPublishedWorkfile(PreLaunchHook):
         Returns:
             None: This is a void method.
         """
-        # TODO setting
+        project_name = self.data["project_name"]
+        task_name = self.data["task_name"]
+        task_type = self.data["task_type"]
+        host_name = self.application.host_name
+
+        # Check settings has enabled it
+        project_settings = get_project_settings(project_name)
+        profiles = project_settings["global"]["tools"]["Workfiles"][
+            "last_workfile_on_startup"
+        ]
+        filter_data = {
+            "tasks": task_name,
+            "task_types": task_type,
+            "hosts": host_name,
+        }
+        last_workfile_settings = filter_profiles(profiles, filter_data)
+        use_last_published_workfile = last_workfile_settings.get(
+            "use_last_published_workfile"
+        )
+        if use_last_published_workfile is None:
+            self.log.info(
+                (
+                    "Seems like old version of settings is used."
+                    ' Can\'t access custom templates in host "{}".'
+                ).format(host_name)
+            )
+            return
+        elif use_last_published_workfile is False:
+            self.log.info(
+                (
+                    'Project "{}" has turned off to use last published workfile'
+                    ' as first workfile for host "{}"'
+                ).format(project_name, host_name)
+            )
+            return
+
         self.log.info("Trying to fetch last published workfile...")
 
+        # Check there is no workfile available
         last_workfile = self.data.get("last_workfile_path")
         if os.path.exists(last_workfile):
             self.log.debug(
@@ -43,9 +81,6 @@ class CopyLastPublishedWorkfile(PreLaunchHook):
                 )
             )
             return
-
-        project_name = self.data["project_name"]
-        task_name = self.data["task_name"]
 
         project_doc = self.data.get("project_doc")
         asset_doc = self.data.get("asset_doc")
@@ -65,6 +100,9 @@ class CopyLastPublishedWorkfile(PreLaunchHook):
                 None,
             )
             if not subset_id:
+                self.log.debug('No any workfile for asset "{}".').format(
+                    asset_doc["name"]
+                )
                 return
 
             # Get workfile representation
@@ -84,41 +122,46 @@ class CopyLastPublishedWorkfile(PreLaunchHook):
                 None,
             )
 
-            if workfile_representation:  # TODO add setting
-                # Get sync server from Tray, which handles the asynchronous thread instance
-                sync_server = next(
-                    (
-                        t["sync_server"]
-                        for t in [
-                            obj
-                            for obj in gc.get_objects()
-                            if isinstance(obj, ModulesManager)
-                        ]
-                        if t["sync_server"].sync_server_thread
-                    ),
-                    None,
-                )
+            if not workfile_representation:
+                self.log.debug(
+                    'No published workfile for task "{}" and host "{}".'
+                ).format(task_name, host_name)
+                return
 
-                # Add site and reset timer
-                active_site = sync_server.get_active_site(project_name)
-                sync_server.add_site(
-                    project_name,
-                    workfile_representation["_id"],
-                    active_site,
-                    force=True,
-                )
-                sync_server.reset_timer()
+            # Get sync server from Tray, which handles the asynchronous thread instance
+            sync_server = next(
+                (
+                    t["sync_server"]
+                    for t in [
+                        obj
+                        for obj in gc.get_objects()
+                        if isinstance(obj, ModulesManager)
+                    ]
+                    if t["sync_server"].sync_server_thread
+                ),
+                None,
+            )
 
-                # Wait for the download loop to end
-                sync_server.sync_server_thread.files_processed.wait()
+            # Add site and reset timer
+            active_site = sync_server.get_active_site(project_name)
+            sync_server.add_site(
+                project_name,
+                workfile_representation["_id"],
+                active_site,
+                force=True,
+            )
+            sync_server.reset_timer()
 
-                # Get paths
-                published_workfile_path = get_representation_path(
-                    workfile_representation, root=anatomy.roots
-                )
-                local_workfile_dir = os.path.dirname(last_workfile)
+            # Wait for the download loop to end
+            sync_server.sync_server_thread.files_processed.wait()
 
-                # Copy file and substitute path
-                self.data["last_workfile_path"] = shutil.copy(
-                    published_workfile_path, local_workfile_dir
-                )
+            # Get paths
+            published_workfile_path = get_representation_path(
+                workfile_representation, root=anatomy.roots
+            )
+            local_workfile_dir = os.path.dirname(last_workfile)
+
+            # Copy file and substitute path
+            self.data["last_workfile_path"] = shutil.copy(
+                published_workfile_path, local_workfile_dir
+            )
--- a/openpype/hooks/pre_copy_last_published_workfile.py
+++ b/openpype/hooks/pre_copy_last_published_workfile.py
@@ -30,6 +30,7 @@ class CopyLastPublishedWorkfile(PreLaunchHook):
         1- Check if setting for this feature is enabled
         2- Check if workfile in work area doesn't exist
         3- Check if published workfile exists and is copied locally in publish
+        4- Substitute copied published workfile as first workfile
 
         Returns:
             None: This is a void method.
--- a/openpype/hooks/pre_copy_last_published_workfile.py
+++ b/openpype/hooks/pre_copy_last_published_workfile.py
@@ -1,0 +1,124 @@
+import gc
+import os
+import shutil
+from openpype.client.entities import (
+    get_last_version_by_subset_id,
+    get_representations,
+    get_subsets,
+)
+from openpype.lib import PreLaunchHook
+from openpype.modules.base import ModulesManager
+from openpype.pipeline.load.utils import get_representation_path
+
+
+class CopyLastPublishedWorkfile(PreLaunchHook):
+    """Copy last published workfile as first workfile.
+
+    Prelaunch hook works only if last workfile leads to not existing file.
+        - That is possible only if it's first version.
+    """
+
+    # Before `AddLastWorkfileToLaunchArgs`
+    order = -1
+    app_groups = ["blender", "photoshop", "tvpaint", "aftereffects"]
+
+    def execute(self):
+        """Check if local workfile doesn't exist, else copy it.
+
+        1- Check if setting for this feature is enabled
+        2- Check if workfile in work area doesn't exist
+        3- Check if published workfile exists and is copied locally in publish
+
+        Returns:
+            None: This is a void method.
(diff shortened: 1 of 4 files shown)


openpype/hooks/pre_copy_last_published_workfile.py
```diff
@@ -0,0 +1,169 @@
...
+
+            if not workfile_representation:
+                self.log.debug(
+                    'No published workfile for task "{}" and host "{}".'
+                ).format(task_name, host_name)
+                return
+
+            # Get sync server from Tray,
+            # which handles the asynchronous thread instance
+            sync_server = next(
```
iLLiCiTiT, line 134: I would recommend not using this global variables approach. There may be multiple created instances of `ModulesManager` at the time this is running. Also consider that prelaunch hooks would be launched in separated process from the main thread (which is in plan). Please consider using rest api endpoint for local webserver rather then using this approach.

openpype/hooks/pre_copy_last_published_workfile.py
```diff
@@ -0,0 +1,169 @@
...
+                "Last workfile exists. Skipping {} process.".format(
+                    self.__class__.__name__
+                )
+            )
+            return
+
+        project_doc = self.data.get("project_doc")
+        asset_doc = self.data.get("asset_doc")
+        anatomy = self.data.get("anatomy")
+        if project_doc and asset_doc:
```
iLLiCiTiT, line 89: I would revert the condition -> there is no `else`...
```suggestion
        if not project_doc or not asset_doc:
            return
```

openpype/hooks/pre_copy_last_published_workfile.py
```diff
@@ -0,0 +1,169 @@
...
+                    'Project "{}" has turned off to use last published'
+                    ' workfile as first workfile for host "{}"'
+                ).format(project_name, host_name)
+            )
+            return
+
+        self.log.info("Trying to fetch last published workfile...")
+
+        # Check there is no workfile available
+        last_workfile = self.data.get("last_workfile_path")
```
iLLiCiTiT, line 77: I would move this condition as first thing in the prelaunch hook. It does not require any settings or data and in most of cases will be available.

openpype/hooks/pre_copy_last_published_workfile.py
```diff
@@ -0,0 +1,169 @@
...
+            # Get subset id
+            subset_id = next(
+                (
+                    subset["_id"]
+                    for subset in get_subsets(
+                        project_name,
+                        asset_ids=[asset_doc["_id"]],
+                        fields=["_id", "data.family"],
+                    )
+                    if subset["data"]["family"] == "workfile"
```
iLLiCiTiT, line 99: Warning: `subset["data"]["family"]` was added recently there is a lot of older projects where only `["data"]["families"]` is available.

openpype/hooks/pre_copy_last_published_workfile.py
```diff
@@ -0,0 +1,169 @@
...
+                return
+
+            # Get workfile representation
+            workfile_representation = next(
+                (
+                    representation
+                    for representation in get_representations(
+                        project_name,
+                        version_ids=[
+                            get_last_version_by_subset_id(
```
iLLiCiTiT, line 116: Last version access without validation of returned value. If something would go wrong in mongo there can be subset without versions s last version would be `None`.

openpype/hooks/pre_copy_last_published_workfile.py
Tilix4: While it only considers published files, every publish from now will have this key right?

iLLiCiTiT: From now on yes, but if I would run it on project which has older subsets it would crash and user can't do literally nothing with it.

Tilix4: Therefore you recommend using chained `get()` or to handle both use cases?

openpype/hooks/pre_copy_last_published_workfile.py
Tilix4: I understand but I cannot figure how to wait for the files copy to finish if I cannot access the asynchronous thread. If I start a `while`, it'll stop the futures stack from proceeding. 

I may try to get memory pointers from the REST API but I'm not sure how it could work.

Does @kalisp have any inputs regarding this issue?

iLLiCiTiT: It would require that sync server has 2 rest points:
- to "trigger sync" of representation
- to "check" if is representation synced

Tilix4: Yes that's what I'd have done, but it doesn't solve the concurrency issue. The server is built on an asynchronous model and not on a multiprocessing system. It means if something is running on the main thread, it'll block the server to execute anything. And that's waiting for a "check" will do. If I create a whole new asynchronous loop in the hook, the rest of the hooks will proceed and the file will be open before it's been downloaded.

kalisp: What about checking DB (periodically). It should fill 'created_dt' for particular site when it is finished.

Tilix4: In fact, I was wrong, the implementation I described works fine. Hope it suits your requirements better.

openpype/hooks/pre_copy_last_published_workfile.py
Tilix4: Ensured no error

openpype/hooks/pre_copy_last_published_workfile.py
Tilix4: Tried to ensure legacy

openpype/modules/sync_server/sync_server.py
```diff
@@ -396,6 +398,8 @@ async def sync_loop(self):
                                               representation,
                                               site,
                                               error)
+                    # Trigger files process finished
+                    self.files_are_processed = False
```
kalisp, line 402: I think this should be moved a tab left, here you are in loop for single project.

I am about bit confused about 'files_are_processed' set to False here, I would expect they would be set to True. (Logic seems to be in opposite direction just because of check of response.content in rest api).

And I don't think it will work as expected if you manage to trigger download of workfile if there is already loop that is synchronizing in progress. The loop will mark that is `files_are_processed` but these are completely different files, your workfile won't be being downloaded yet.

I think safer would be loop and check progress or finished synch for particular representation.

Tilix4: Do you think this check should be requested through REST API and executed by the server or only on the client side?

openpype/modules/sync_server/rest_api.py
```diff
@@ -0,0 +1,68 @@
...
+            )
+            self.log.error(msg)
+            return Response(status=400, message=msg)
+
+        # Add all sites to each representation
+        for representation_id in representations:
+            for site in sites:
+                self.module.add_site(
+                    project_name, representation_id, site, force=True
+                )
```
kalisp, line 58: You might want to bump up 'priority' higher to skip the queue of representations that should be synchronized, if there is one.

Unfortunately `add_site` cannot do it right now, but something like this should work:
```
            self.module.update_db(self.project, None, None,
                                       representation,  # this must be representation doc
                                       get_local_site_id(),
                                       priority=value)
```

openpype/modules/sync_server/sync_server.py
kalisp: I think only on client side. 
Eventually it will be changed to some kind of REST api call to v4 server, but I wouldn't worry about it in this PR.

Tilix4: Do you have any clue about where to listen for such information on the client side?

Tilix4: I added a `priority` field to the `add_site` function because `update_db` wasn't working and I had no clue why. This works pretty well.

openpype/client/entities.py
```diff
@@ -1367,6 +1368,33 @@ def get_representation_parents(project_name, representation):
     return parents_by_repre_id[repre_id]
 
 
+def get_representation_last_created_time_on_site(
```
iLLiCiTiT, line 1371: Point 1: This is not site sync (sync server) specific function (should not be in client code|).
Point 2: Python 2 compatibility breaking code.

openpype/hooks/pre_copy_last_published_workfile.py
```diff
@@ -0,0 +1,199 @@
...
+            },
+        )
+
+        # Wait for the download loop to end
+        last_created_time = get_representation_last_created_time_on_site(
+            workfile_representation, local_site_id
+        )
+        while (
+            last_created_time
+            >= get_representation_last_created_time_on_site(
```
iLLiCiTiT, line 179: I would actually expect that `requests.get(...)` is used here instead.

openpype/client/entities.py
Tilix4: 1. Could you make a conclusion up with @kalisp and tell me what to do in the end? He asked me to put it on the client side: https://github.com/pypeclub/OpenPype/pull/4049#discussion_r1014296215
2. I'll fix the python2

openpype/hooks/pre_copy_last_published_workfile.py
```diff
@@ -0,0 +1,199 @@
...
+    get_representation_by_id,
+    get_representation_last_created_time_on_site,
+    get_representations,
+    get_subsets,
+)
+from openpype.lib import PreLaunchHook
+from openpype.lib.local_settings import get_local_site_id
+from openpype.lib.profiles_filtering import filter_profiles
+from openpype.pipeline.load.utils import get_representation_path
+from openpype.settings.lib import get_project_settings
```
iLLiCiTiT, line 15:
```suggestion
from openpype.settingsimport get_project_settings
```

openpype/hooks/pre_copy_last_published_workfile.py
Tilix4: Come on, I changed it on the request from @kalisp...

openpype/client/entities.py
iLLiCiTiT: > He asked me to put it on the client side:

He was reffering that the code should run only on client side "a process" -> in tray or sync server processes. He didn't said that it should be in client code.

Tilix4: Thanks for precising.

Tilix4: But I don't understand which difference you make between client and server if you call 'client' what is run on sync server side using REST API (as you're suggesting [here](https://github.com/pypeclub/OpenPype/pull/4049#discussion_r1017676506))

iLLiCiTiT: There was a miscommunitaion on many fronts. We had an internal discussion about it and made a PR that should fix the issues we had (https://github.com/Tilix4/OpenPype/pull/5). Right now there is missing implementation of one method which will be filled soon.

Tilix4: Okay thanks, I've commented it a bit.

kalisp: Yeah, sorry, my bad, I am bit struggling with naming of `openpype.client`.

What I meant was that the method should be on sync server module, not necessarily on Rest API (that was the point I tried to make rather clumsily) , BUT @iLLiCiTiT described me reasons why Rest API is important.

(`ModuleManager.get_module_by_name('sync_server")` doesn't return singleton, there might be multiple instances of SyncModule, only one of those actually has access to `sync_server_thread`. On that thread there might be a timer that is delaying downloading. You need Rest API for instances of module to communicate between each other. Same API might be extended in v4 for additional functionality.)

So after discussion we will be doing some modifications, that you could test and hopefully bringing it home soon :).

Tilix4: No problem, we finally found our way out :)

I totally understand the power of the Rest API now and it completely makes sense against my first (inaccurate) proposal.

openpype/hooks/pre_copy_last_published_workfile.py
```diff
@@ -0,0 +1,174 @@
...
+            "tasks": task_name,
+            "task_types": task_type,
+            "hosts": host_name,
+        }
+        last_workfile_settings = filter_profiles(profiles, filter_data)
+        use_last_published_workfile = last_workfile_settings.get(
+            "use_last_published_workfile"
+        )
+        if use_last_published_workfile is None:
+            self.log.info(
```
kalisp, line 75:
```suggestion
            self.log.info(
                       (
                    "Seems like old version of settings is used."
                    ' Can\'t access custom templates in host "{}".'.format(host_name)
            )
```

openpype/hooks/pre_copy_last_published_workfile.py
```diff
@@ -0,0 +1,174 @@
...
+                ).format(host_name)
+            )
+            return
+        elif use_last_published_workfile is False:
+            self.log.info(
+                (
+                    'Project "{}" has turned off to use last published'
+                    ' workfile as first workfile for host "{}"'
+                ).format(project_name, host_name)
+            )
```
kalisp, line 88: I think it should be 
```
self.log.info(
                (
                    'Project "{}" has turned off to use last published'
                    ' workfile as first workfile for host "{}"' .format(project_name, 
                                                                                              host_name)
            )
            ```
Cannot persuade to add commit suggestion for multiple lines

File: openpype/hooks/pre_copy_last_published_workfile.py [hunk 13]
```diff
@@ -0,0 +1,174 @@
...
+                )
+                if representation["context"]["task"]["name"] == task_name
+            ),
+            None,
+        )
+
+        if not workfile_representation:
+            self.log.debug(
+                'No published workfile for task "{}" and host "{}".'
+            ).format(task_name, host_name)
```
kalisp, line 145: Same issue as higher, there shouldn't be ')' before `.format`

openpype/hooks/pre_copy_last_published_workfile.py
```diff
@@ -0,0 +1,174 @@
...
+        last_workfile_settings = filter_profiles(profiles, filter_data)
+        use_last_published_workfile = last_workfile_settings.get(
+            "use_last_published_workfile"
+        )
+        if use_last_published_workfile is None:
+            self.log.info(
+                (
+                    "Seems like old version of settings is used."
+                    ' Can\'t access custom templates in host "{}".'
+                ).format(host_name)
```
kalisp, line 79: Same issue as lower, there shouldn't be ')' before .format

kalisp (changes requested):
Couple of cases of wrong logging.

kalisp (approved):
It works! :)

kalisp:
Merged, thanks @Tilix4  for contribution and patience :)

PR #4058, opened 2022-11-01
Publisher: Easy access to publish page from create page

iLLiCiTiT:
## Brief description
Added button to go from create page to publish page.

## Description
Added button over instance view which simplify change of page from create page to publish page. The button shows on hover of instance view (in half width of the widget).

## Additional info
Color changes or timing changes can be done. Colors are now hardcoded.

### Screenshot
image

## Testing notes:
1. Open Publisher (e.g. in tray publisher)
2. Hover over right side of the window on create page
3. A button should show
4. Click should get you to publish page


jakubjezek001:
There might be still some room for improvements of UX. At first, the curve doesn't fit design language of openpype, secondly I would start the interaction from very close to `create` button already. 

Here is an idea of what might be closer to the design fit. If you switch Chrome to fullscreen and then hover on top edge of monitor you will see this full circle button with cross. We could do similar interaction on right side with right simple arrow. 

image

iLLiCiTiT:
Modified to circle with arrow.

mkolar (changes requested):
Much better now. I just change the colour to the same as buttons and make the animation a bit softer. It's quite jittery now because of the double bounce.

iLLiCiTiT:
> It's quite jittery now because of the double bounce.

You don't like bounce? Changed to cubic animation...

jakubjezek001 (approved):
looks awesome. I would even speed the animation little (my taste), but already works well now.

iLLiCiTiT:
> I would even speed the animation little (my taste), but already works well now.

I kept the animation duration from previous easing curve (made shorter).

PR #4060, opened 2022-11-02
General: Oiio conversion multipart fix

iLLiCiTiT:
## Brief description
Don't define input channels if input is multipart exr.

## Description
When converting multipart exr for ffmpeg it can't handle definition of channels on input because there are multiple subimages which cause crash if oiiotool.

Extended few types in xml info result parsing and added ability to receive information about every subimage in exr.

## Additional information
The output has all subimages (which can be big exr) that is because ffmpeg then created transparent image without any content. Not sure why but it's an issue. Also if any subimage in the exr has attributes with invalid characters or too long value it would cause issues in ffmpeg too -> future enhancements.

## Testing notes:
1. It should be possible to convert multipart exr for ffmpeg (e.g. in extract review)

### DIFF
--- a/openpype/lib/transcoding.py
+++ b/openpype/lib/transcoding.py
@@ -111,6 +111,7 @@ def get_oiio_info_for_input(filepath, logger=None):
 
 class RationalToInt:
     """Rational value stored as division of 2 integers using string."""
+
     def __init__(self, string_value):
         parts = string_value.split("/")
         top = float(parts[0])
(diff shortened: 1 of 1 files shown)


antirotor:
Looks good but since we are messing up now with xml, shouldn't we parse it and work with it that way instead of hacking it as string?

iLLiCiTiT:
> Looks good but since we are messing up now with xml, shouldn't we parse it and work with it that way instead of hacking it as string?

I don't know to which code is this related?

LiborBatek:
Did test  in ```maya2022``` and ```arnold``` it Deadline job finnished normally without any errors and published all.

here is the scrncap of published data
image

however when played the mp4 review even thumbnail looked fine the vid is completely black...
image

antirotor:
> > Looks good but since we are messing up now with xml, shouldn't we parse it and work with it that way instead of hacking it as string?
> 
> I don't know to which code is this related?

all lines like `if line == "</ImageSpec>":` - wouldn't be cleaner to use xml dom and not strings?

LiborBatek (changes requested):
Not working fully. Pls see comments.

PR #4063, opened 2022-11-03
Publisher: Prepared common function for instance data cache

iLLiCiTiT:
## Brief description
Prepared common function for caching instance data.

## Description
We've found out the same functionality if implemented across multiple hosts so it deserve already prepared function to achieve it.

## Testing notes:
1. TrayPublisher instances should be refreshed


kalisp (approved):
Works, but it is not much applicable for Adobe hosts as it doesn't handle legacy instances.

PR #4065, opened 2022-11-03
General: Direct interfaces import

iLLiCiTiT:
## Brief description
Replace import from dynamic python module `openpype_interfaces` with direct import `openpype.modules`.

## Description
Idea in past was that interfaces should be possible to add dynamically but after a time we got to point when it's not used anymore and only base/core interfaces are used thus dynamic module `openpype_interfaces` is not needed anymore.

Interfaces are imported to top level of `openpype.modules` to simplify module specific imports. All imports in existing modules are replaced with new source. Dynamic module `openpype_interfaces` still works but import from it will cause warning that it's using deprecated approach.

## Testing notes:
1. All modules are loaded correctly
2. All host integration are loaded correctly

### DIFF
--- a/openpype/hosts/aftereffects/addon.py
+++ b/openpype/hosts/aftereffects/addon.py
@@ -1,5 +1,4 @@
-from openpype.modules import OpenPypeModule
-from openpype.modules.interfaces import IHostAddon
+from openpype.modules import OpenPypeModule, IHostAddon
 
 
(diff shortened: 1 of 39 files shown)


kalisp (approved):
Works in Maya, Nuke, AE, PS.

But I am guessing it works everywhere as it is directly in addon and I didn't notice any errors during start of Tray, eg initialization of addons, imho.

PR #4066, opened 2022-11-03
Settings UI: Locked setting can't bypass lock

iLLiCiTiT:
## Brief description
Avoid to show unsaved changes dialog if settings are in view only mode.

## Description
Even if settings UI is in view only mode change of project may trigger unsaved changes dialog which can allow to save settings. This PR makes sure the dialog is not showed.

## Testing notes:
1. Open settings UI
2. Open second settings UI -> should ask to view only mode -> accept view only mode
3. Do some changes in project settings
4. Go to different project
5. No dialog should appear


kalisp (approved):
Solves the issue.

64qam (approved):

PR #4068, opened 2022-11-03
Fix for Workfile Widget Performance Issues when Shown

munkybutt:
## Brief description
Fix for workfile ui refresh issues

## Description
Fixed workfile ui refresh issues.
This was causing the file widget to refresh up to 10 times each time the workfile ui was shown.
This was due to over use of both signals and direct function calls to try and drive the refresh process across asset, task and file widgets.
It has been refactored to work off of signals only which should provide coverage for all use cases requiring the ui to refresh.
The ui will now only refresh twice on open which makes for a snappier ux.
This had significant performance implications, especially when working remotely as it queries the DB on each refresh.

This change has brought with it the `ItemViewWidget` class which is a base class for `AssetsWidget` and `TasksWidget` which will provide a unified search in later work.
For now the search bar is hidden in `TasksWidget` to maintain the appearance of the previous ui until other changes can be introduced so as to keep this change relatively small.

## Testing notes:
Changes tested in Maya
1. Once in DCC - open the workfiles ui from the OpenPype menu
2. Notice it doesn't take as long to load up and doesn't spam the output console with logged filter queries


openpype/tools/utils/tasks_widget.py
```diff
@@ -17,15 +31,48 @@
...
+        # type: (str, T_Dict, T_Dict, T_Dict | None, bool) -> None
+        task_type = task_info.get("type")  # type: str | None
+        label = (
+            "{} ({})".format(task_name, task_type or "type N/A")
+            if task_type_in_label else task_name
+        )
+        super(TaskViewItem, self).__init__(label)
+
+        task_assignees = set()  # type: set[str]
+        assignees_data = task_info.get("assignees") or []  # list[dict[str, str]]
```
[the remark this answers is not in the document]
munkybutt: breaking this into multiple lines doesn't help the readability of the code and the type annotation is beneficial so not worth removing.

openpype/tools/workfiles/model.py
```diff
@@ -15,12 +17,26 @@
...
 
+if six.PY2:
+    import pathlib2 as pathlib
+else:
+    import pathlib
+
+_typing = False
+if _typing:
+    from typing import Any
+    from Qt.QtGui import QIcon
```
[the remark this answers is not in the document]
munkybutt: This are type imports for static analysis, they will never run during the lifetime of the program.
Use of a `False` literal is cheaper than importing `typing` and using `typing.TYPE_CHECKING`
More info here: https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

openpype/tools/workfiles/model.py
```diff
@@ -15,12 +17,26 @@
...
 from openpype.pipeline import get_representation_path
 
+if six.PY2:
+    import pathlib2 as pathlib
+else:
+    import pathlib
+
+_typing = False
+if _typing:
+    from typing import Any
```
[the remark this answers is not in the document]
munkybutt: This are type imports for static analysis, they will never run during the lifetime of the program.
Use of a `False` literal is cheaper than importing `typing` and using `typing.TYPE_CHECKING`
More info here: https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

iLLiCiTiT:
Asset name filtering is missing and both asset and task widgets have refrehs button which does not seem right.
image

openpype/tools/workfiles/window.py
```diff
@@ -386,25 +384,25 @@ def _create_workfile_doc(self, filepath):
...
+        # type: (dict[str, Any]) -> None
+        if context is None:
             return
 
         if self.assets_widget.refreshing:
-            return
+            def _set_context():
+                self.set_context(context)
+
+            return QtCore.QTimer.singleShot(100, _set_context)
```
iLLiCiTiT, line 404: These objects (function and timer) can be garbage collected before execution in older python/pyside.

openpype/tools/utils/widgets.py
```diff
@@ -586,3 +592,204 @@ def set_orientation(self, orientation):
...
+    def __init__(self, placeholder_text="...", parent=None):
+        super(PlaceholderHistoryLineEdit, self).__init__(parent=parent)
+        self.setEditable(True)
+        self.setPlaceholderText(placeholder_text)
+
+
+class ToolbarButton(QtWidgets.QPushButton):
+    def __init__(self, *args, **kwargs):
+        super(ToolbarButton, self).__init__(*args, **kwargs)
+        self.setFixedWidth(35)
```
iLLiCiTiT, line 607: It had a reason the width and height are same. Also using fixed sizes cause issues when someone has high resolution and enabled scaling.

openpype/tools/workfiles/window.py
munkybutt: > Asset name filtering is missing and both asset and task widgets have refresh button which does not seem right.

I will resolve the ui layout so it looks and functions as before ( sloppy of me to not double check this ) but have the new components in place, just hidden.
The intention for a refresh button on tasks is to allow a localised refresh for the active asset, avoiding having to refresh the whole project which is much more expensive, especially for larger projects and remote setups. This was a request from our artists as tasks get added/updated in Jira arbitrarily.
If this doesn't make sense for some reason or limitation I haven't been able to discern from the code then that is fair enough and can be hidden permanently. 

> These objects (function and timer) can be garbage collected before execution in older python/pyside.

Ah interesting - I had wondered as to the fixed timer approach. I made this modification to test if the timer was causing the extra refreshes (it wasn't) but left it in as it was more concise. I will revert it to the previous setup to avoid the backwards compatibility issues.

iLLiCiTiT: > If this doesn't make sense for some reason or limitation I haven't been able to discern from the code then that is fair enough and can be hidden permanently.

I would say it's better to have single refresh button per tool. @mkolar opinions?

munkybutt: weird - something is going wrong as the asset filtering is visible for me:

image

Are you able to see anything that would be hiding it for you?

openpype/tools/utils/widgets.py
munkybutt: I can't see where the previous refresh button had it's width and height set?
Unless you have another suggestion, I will use `QtWidgets.QToolButton` instead as that maintains a square ratio?

openpype/tools/workfiles/window.py
BigRoy: > I would say it's better to have single refresh button per tool. mkolar opinions?

Personally I agree - this double refresh button feels too cluttered and confusing. Rather have one refresh that force refreshes all.

munkybutt: As mentioned, this was a request by my artists - so there is a valid use case.
I guess this is a case of form vs function?

Whilst this is on going, I am pushing updates which will include the task widget refresh button being hidden by default (for now).

iLLiCiTiT:
To be honest. I think you've did more unrelated changes in this PR then is necessary. At this moment we got to point we're not resolving the reason of this PR but changes of UX -> that should be totally different PR. At this moment there is more UX changes then handling slowdown.  Can we keep this PR as clean as possible without any UX changes and create issue for the UX part?

openpype/tools/utils/widgets.py
```diff
@@ -606,3 +612,245 @@ def set_orientation(self, orientation):
...
+        if not (show_search_bar and show_refresh_button):
+            height = self.height()
+            # these ratios were calculated using standard resolution
+            # and scaling:
+            top = int(height * 0.133333)
+            bottom = int(height * 0.533333)
+            self._header_layout.setContentsMargins(0, top, 0, bottom)
+
+
+class ItemViewWidget(QtWidgets.QWidget):
```
iLLiCiTiT, line 761: I quite don't like this approach. If we'll need to change how assets or tasks widgets works we'll be blocked by the fact they have this parent. I would keep assets widget and task widget as 2 separated classes (from top to bottom). If we'll need change one of them in some way it's more complicated. Also like I've mentioned I don't think tasks widget should have headers -> please create a ticket to propose that.

munkybutt: Fair request to split this up further - I just won't have the time to do it for a while as I need to get back to my in house work.
I will open a ticket about the changes we have made to the launcher and workfiles ui when I get a spare moment too.

I appreciate the time taken to review and give feedback.

iLLiCiTiT:
The PR is changing more than can be accepted. Workfiles tool will be rewriten for AYON and because the author of PR and Ynput team don't have time to modify the code, I'm closing the PR.

PR #4071, opened 2022-11-04
Get kitsu login to fill username burnin

friquette:
## Brief description
Get kitsu login to fill username burnin. when session username is not right.

## Description
Get kitsu login to fill username burnin when session username is not the right username (like multiple artists working on the same session).

### DIFF
--- a/openpype/modules/kitsu/plugins/publish/collect_kitsu_username.py
+++ b/openpype/modules/kitsu/plugins/publish/collect_kitsu_username.py
@@ -12,16 +12,17 @@ class CollectKitsuUsername(pyblish.api.ContextPlugin):
     label = "Kitsu username"
 
     def process(self, context):
+        kitsu_login = os.environ['KITSU_LOGIN']
+
+        if not kitsu_login:
+            return
+
+        kitsu_username = kitsu_login.split("@")[0].replace('.', ' ')
+        new_username = re.sub('[^a-zA-Z]', ' ', kitsu_username).title()
+
         for instance in context:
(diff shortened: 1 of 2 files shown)


iLLiCiTiT:
This is not how we add addon specific data to burnins, it's not job of ExtractBurnin to know that. To achive that you can fill `"customData"` key on instance in kitsu collector to pass the data to burnins. In burnins you can then use `{custom["kitsuUsername"]}` for instance.

friquette:
Thanks for the tips ! When you talk about kitsu collector, do you mean the collect_kitsu_entities file ?

iLLiCiTiT:
I would recommend special collector just for that, so it can have it's own settings.

openpype/modules/kitsu/plugins/publish/collect_kitsu_username.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class CollectKitsuUsername(pyblish.api.ContextPlugin):
+    """Collect Kitsu username from the kitsu login"""
+
+    order = pyblish.api.CollectorOrder + 0.499
+    label = "Kitsu username"
+
+    def process(self, context):
+        kitsu_login = os.environ['KITSU_LOGIN']
```
iLLiCiTiT, line 15: No sure if `KITSU_LOGIN` env is always filled?
```suggestion
        kitsu_login = os.environ.get('KITSU_LOGIN')
```

iLLiCiTiT (approved):

PR #4074, opened 2022-11-07
Nuke: resolve hashes in file name only for frame token

jakubjezek001:
## Brief description
Resolving path to image sequence with hashes only at frame token.

## Description
Solving issue with resolving name to hashes even in other parts of name.

## Testing notes:
1. Open your testing workfile in Nuke
2. Open Loader and load image sequence, and some video file
3. It should load as excepted

### DIFF
--- a/openpype/hosts/nuke/plugins/load/load_clip.py
+++ b/openpype/hosts/nuke/plugins/load/load_clip.py
@@ -1,5 +1,6 @@
 import nuke
 import qargparse
+from pprint import pformat
 from copy import deepcopy
 from openpype.lib import Logger
 from openpype.client import (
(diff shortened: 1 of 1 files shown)


openpype/hosts/nuke/plugins/load/load_clip.py
```diff
@@ -22,6 +22,8 @@
 )
 from openpype.hosts.nuke.api import plugin
 
+log = Logger.get_logger(__name__)
```
iLLiCiTiT, line 25: I would not change calling of `self.log` in the plugin. If you need to override logger please do it on class level rather.

64qam (approved):

PR #4076, opened 2022-11-07
Maya: fix rounding of fps

### LINKED ISSUE #2620: fps rounding issue in validate_maya_units.py

antirotor:
## Bug

This is fixing problem when fps is defined in greater precision than Maya is expecting (for example when it is defined as `23.9760246277` in db and Maya logic is expecting `23.98`)

Close #2620


m-u-r-p-h-y (commented):
In my case, everything seems correct.

- project set to 29.97 in ftrack and synced to avalon
- created shot in ftrack and synced to avalon
- manually changed the shot fps to `23.9760246277` in mongoDB
- started Maya (2022) and created animation instance with some content (Maya starts with 25fps by default)
- run publish and used automatic repair on Maya Unit validator. fps changed from 25fps to 23.967 in Maya
- run publish again
- everything is OK


image


so it seems it is fixing the issue of higher precision number in mongoDB (coming from hiero editorial for example)
but rounding not working in @LiborBatek case

m-u-r-p-h-y:
btw, should we fix project manager to show full length number for fps for better diagnostics?

m-u-r-p-h-y (changes requested):
just created another shot to confirm Libor's case and it is not handled correctly definitely

did not touch mongoDB, just let the ftrack sync to avalon

during the workfile save there is a warning about fps mismatch (we should use floats to see the issue properly)

image
 

after fixing the scene, the Units validator fails, as there is double rounding happening apparently

image

antirotor:
> after fixing the scene, the Units validator fails, as there is double rounding happening apparently

I don't think it is because of double rounding - it all stems from ftrack handling of decimal places. In ftrack, it will be **29.97**, but since maya stores it in much higher precision, it will correctly ceil the value to 29.98, because `ceil(29.968)` is **29.98**.

Since we are not able to tell during validation if that fps value in mongo is set by ftrack that is flooring the value, or it is set by any other means where the ceil is the rounding method, I am not really sure how to handle it except storing fps value as string on ftrack, or adding logic trying to guess if ftrack module is enabled...

antirotor:
I hope I've fixed it by comparing only numbers with the same precision.

LiborBatek:
So I ve made the test...

first setting the fps to 23.976024 in MongoDB also checked Project Manager to see if set accordingly...
PR#4076_scrnsht01

then I ve opened the shot and Maya told me to fix keyframes (I didnt go for fix and ignored it)
PR#4076_scrnsht01C
afterwards went for shot publish and got this error with wrong FPS set in the scene (also note small issue with incorrect info about actual fps seems rounded already but still being that 23.976)
PR#4076_scrnsht01B

When triggered Repair action all went good and shot was set to exact FPS 23.976 and Publish went ok! see maya fps in timeline...

PR#4076_scrnsht02

Seem fine to me besides small desinformation regarding 24fps instead of 23.976 in info/warning message.

m-u-r-p-h-y:
My testing scenario is (M2022)

sh001

set to `29.97` in mongoDB

image


opening sh001 workfile with fps set to 25fps

`currentUnit -l centimeter -a degree -t pal;`


which brings the popup message to fix the fps on file open

image

with sill not very precise message about project fps of 29 fps 


after pressing fix, OP is changing the scene fps:

`// openpype.hosts.maya.api.lib : Setting scene FPS to: '29.97fps' // `


but Maya Units validation is not passing

image

Repair function on the validator is not helping either, even though it says

`// openpype.hosts.maya.api.lib : Setting scene FPS to: '29.97fps' // `

but the validation fails with the above message again

LiborBatek:
To me it seems that Maya Fix FPS causes that weird FPS value...

Did u tried to not trigger that maya fix FPS? It could be it also it doesnt mean user wants to fix keyframe by moving them on the timeline to match new FPS but keep them as they are...

m-u-r-p-h-y:
> To me it seems that Maya Fix FPS causes that weird FPS value...
> 
> Did u tried to not trigger that maya fix FPS? It could be it also it doesnt mean user wants to fix keyframe by moving them on the timeline to match new FPS but keep them as they are...

user has to toggle the switch to modyfi the keyframes. The fix is only changing the scene fps by default.

antirotor:
solved by #4424

PR #4077, opened 2022-11-07
hiero: creator from settings with set maximum

jakubjezek001:
## Brief description
Fixing wrong value in ui set from settings

## Description
Settings value was not correctly transferred onto creator ui. 

## Testing notes:
1. Open Hiero workfile
2. select a clip and hit OP Creator
3. Look to down to workfile start frame and it should be 1001


jrsndl (approved):
Works as described. The project_settings/hiero/create/CreateShotClip/workfileFrameStart now doesn't max to 99.
BTW this is only problem for Hiero/NS 13+, settings works withouth issue in Hiero 12.2.

PR #4079, opened 2022-11-08
Flame: Loading clip with native colorspace resolved from mapping

jakubjezek001:
## Brief description
Flame is loading clips with correct colorspace mapping

## Description
Mapping can be added at Studio settings/Project tab: `project_settings/flame/imageio/profilesMapping`. 

## Testing notes:
1. Make sure mapping is set in `project_settings/flame/imageio/profilesMapping`
2. Load any published subset version 
3. See the colorspace is correctly set

### DIFF
--- a/openpype/hosts/flame/api/plugin.py
+++ b/openpype/hosts/flame/api/plugin.py
@@ -4,13 +4,13 @@ import shutil
 from copy import deepcopy
 from xml.etree import ElementTree as ET
 
+import qargparse
 from Qt import QtCore, QtWidgets
 
-import qargparse
 from openpype import style
-from openpype.settings import get_current_project_settings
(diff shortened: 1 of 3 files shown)


openpype/hosts/flame/api/plugin.py
```diff
@@ -690,6 +690,19 @@ class ClipLoader(LoaderPlugin):
         )
     ]
 
+    _mapping = None
+
+    def get_native_colorspace(self, input_colorspace):
```
iLLiCiTiT, line 695: I would recommend change this method to classmethod so the values are kept in plugin if multiple representations are loaded at a time.

iLLiCiTiT (approved):

PR #4082, opened 2022-11-09
TvPaint: Upload only main review file on kitsu from TvPaint

friquette:
## Brief description
When publishing, upload only the review file on kitsu.

## Description
We realized that when publishing exports from TvPaint, all the files are uploaded to Kitsu (all reviews for renderLayers and renderPasses) but the only one we really need is the main review. Is our logic correct or are all the files are necessary for Kitsu to work properly ?

### DIFF
--- a/openpype/hosts/tvpaint/plugins/publish/extract_sequence.py
+++ b/openpype/hosts/tvpaint/plugins/publish/extract_sequence.py
@@ -127,9 +127,9 @@ class ExtractSequence(pyblish.api.Extractor):
             output_frame_start
         )
 
-        # Fill tags and new families
+        # Fill tags and new families from project settings
         tags = []
-        if family_lowered in ("review", "renderlayer", "renderscene"):
+        if family_lowered in self.families_to_upload:
             tags.append("review")
 
         # Sequence of one frame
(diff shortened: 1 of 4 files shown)


iLLiCiTiT:
For shotgrid and ftrack there are tags in extract review for filtering of that. This filter is "very specific" in code. I expect you would break productions which don't produce `reviewMain` or `.mp4` for review.

openpype/hosts/tvpaint/plugins/publish/extract_sequence.py
```diff
@@ -127,9 +127,9 @@ def process(self, instance):
             output_frame_start
         )
 
-        # Fill tags and new families
+        # Fill tags and new families from project settings
         tags = []
-        if family_lowered in ("review", "renderlayer", "renderscene"):
+        if family_lowered in self.families_to_upload:
```
iLLiCiTiT, line 132: Those are not families to upload but families to add review tag.
```suggestion
        if family_lowered in self.families_to_review:
```

Also please define `families_to_review` as class attribute of the plugin with default value.

openpype/settings/defaults/project_settings/tvpaint.json
```diff
@@ -11,6 +11,11 @@
                 255,
                 255,
                 255
+            ],
+            "families_to_review": [
+                "review",
+                "renderpass",
+                "renderlayer"
```
iLLiCiTiT, line 18:
```suggestion
                "review",
                "renderlayer",
                "renderscene"
```

iLLiCiTiT (approved):

PR #4083, opened 2022-11-09
Maya: aov filtering

moonyuet:
## Brief description
fixing the bug for the regex filter not excluding the aov in RedShift
See(https://github.com/pypeclub/OpenPype/issues/4057)

## Description
Multipart boolean option of the renderers (RedShift and Arnold) is broken in lib_renderproduct.py, which submits a wrong information to collect_render.py(i.e. multipart is always True even when it is being turned off..), This cause the review option not working with regex filters, which depends on whether multipart is on /off.

For this fix, 
if “force combine beauty and AOVs into single file” or “multipart EXR” in Redshift Render Setting is off, it will filter out the renderpass according to the regex filter set in the Openpype Setting. (I set .*([Bb]eauty).* as filter)
image

If either one is on, it will show both beauty and Cryptomatte pass.

image

If mergeAOVs or multipart in Arnold Render Setting is off, it will filter out the renderpass in regards to the regex filter set in the Openpype setting. .

image

 If multipart is on, it will render all the passes for the multipart option (I am not sure if it’s a good idea. so let me know if you also want the filter option included here.) 
image

while it will just only render beauty pass for mergeAOVs
image

## Additional info
No Additional Info

## Testing Notes
1. check if you have disable the "multipart" or "force combine beauty and AOVs" in Redshift Render Setting  OR
   check if you have disable the "mutlipart" or "Merge AOVs" in Arnold Render Setting.
image
image

2. set your regex filter in deadline setting
image
3. add the AOV maps
image
4. Create the render instance and run the publisher

### DIFF
--- a/openpype/hosts/maya/api/lib_renderproducts.py
+++ b/openpype/hosts/maya/api/lib_renderproducts.py
@@ -1110,7 +1110,7 @@ class RenderProductsRedshift(ARenderProducts):
         if light_groups_enabled:
             return products
 
-        beauty_name = "Beauty_other" if has_beauty_aov else ""
+        beauty_name = "BeautyAux" if has_beauty_aov else ""
         for camera in cameras:
             products.insert(0,
                             RenderProduct(productName=beauty_name,
--- a/openpype/hosts/maya/api/lib_renderproducts.py
+++ b/openpype/hosts/maya/api/lib_renderproducts.py
@@ -536,7 +536,11 @@ class RenderProductsArnold(ARenderProducts):
 
         products = []
         aov_name = self._get_attr(aov, "name")
-        multipart = bool(self._get_attr("defaultArnoldDriver.multipart"))
+        multipart = False
+        multilayer = bool(self._get_attr("defaultArnoldDriver.multipart"))
+        merge_AOVs = bool(self._get_attr("defaultArnoldDriver.mergeAOVs"))
+        if multilayer or merge_AOVs:
+            multipart = True
         ai_drivers = cmds.listConnections("{}.outputs".format(aov),
                                           source=True,
                                           destination=False,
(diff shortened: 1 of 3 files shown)


m-u-r-p-h-y:
the render settings under `project_settings/maya/RenderSettings` should only be used as render presets, not for evaluation during rendering. We should always take the settings from the render scene itself.

moonyuet:
> the render settings under `project_settings/maya/RenderSettings` should only be used as render presets, not for evaluation during rendering. We should always take the settings from the render scene itself.

I find out that the boolean of multipart in ` collect_render.py` doesn't align with that of multipart in `lib_renderproduct.py`, I have edited the `lib_renderproduct.py` so that the multipart works properly in Redshift. For Arnold, I still want to see if you guys want to have the filtering even if the multipart is on  as it seems that it still renders all passes when the multipart is on.

m-u-r-p-h-y (dismissed):
tested the redshift and there is a problem with the code which was taking care of Redshift change of `Beauty` name to` BeautyAux`


image

you can see actually rendered files and the content of the JSON file with the wrong expected files

LiborBatek (approved):
So now it works also in Redshift. Seems ok to me. Not spotted any problems.

openpype/hosts/maya/api/lib_renderproducts.py
```diff
@@ -1044,7 +1054,7 @@ def get_render_products(self):
 
             # Any AOVs that still get processed, like Cryptomatte
             # by themselves are not multipart files.
-            aov_multipart = not multipart
+            # aov_multipart = not multipart
```
antirotor, line 1057: I would just remove this

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -494,12 +494,13 @@ def _create_instances_for_aov(self, instance_data, exp_files):
...
             aov_patterns = self.aov_filter
-            preview = match_aov_pattern(app, aov_patterns, render_file_name)
 
+            preview = match_aov_pattern(app, aov_patterns, render_file_name)
             # toggle preview on if multipart is on
+
             if instance_data.get("multipartExr"):
                 preview = True
-
+            self.log.info("preview:{}".format(preview))
```
antirotor, line 503:
```suggestion
            self.log.debug("preview:{}".format(preview))
```

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -542,7 +543,7 @@ def _create_instances_for_aov(self, instance_data, exp_files):
             if new_instance.get("extendFrames", False):
                 self._copy_extend_frames(new_instance, rep)
             instances.append(new_instance)
-
+            self.log.info("instances:{}".format(instances))
```
antirotor, line 546:
```suggestion
            self.log.debug("instances:{}".format(instances))
```

antirotor (commented):
Just some minor cosmetic changes

antirotor (approved):

PR #4087, opened 2022-11-10
Maya: keep existing AOVs when creating render instance

moonyuet:
## Brief description
Introduce an option to allow users to choose if existing AOVs should be kept when creating render instance in Openpype setting.
The setting is se to false by default 
image

## Description
The new options in Openpype Setting named "remove_AOVs" is avaliable in this PR. If this is set to be True, it would clean up the previous AOVs which is left by the deleted render instance. If this is set to be False, It would still store the existing AOVs as renderpasses.

## Additional info
Nothing specific

## Testing notes:
1. In Openpype Settings > Maya> Render Settings > Arnold Renderer, there is a on/off options of remove_AOVs
2. If you enable it, it would help you to clean up the existing AOVs if you create a new render instance.
3. if you disable it, it would keep the existing AOVs while creating new AOVs without duplication of existing AOVs.

resolves https://github.com/pypeclub/OpenPype/issues/4073


m-u-r-p-h-y (commented):
Tested with M2023 and Arnold 5.2.0.1

working perfectly as expected.


I would have some UI and useability questions though

1. I would recommend changing the label to a more descriptive `Remove existing AOVs`

2. This option should be available for all renderers (it is not Arnold specific)

should the button be placed in each renderer section or in the common, upper level?
image

In case we decide to keep the option per renderer we should just swap the order and keep all AOV related options together

image

@antirotor @mkolar what do you think?

antirotor:
I don't think it should be per renderer - and I agree with changing it to `Remove existing AOVs`

moonyuet:
> I don't think it should be per renderer - and I agree with changing it to `Remove existing AOVs`

I have renamed in the updated PR as well as swapping the position of `remove existing AOVs` attribute with the tiled attribute. The` remove existing AOVs` is now located right before the `AOVs to create` attribute

m-u-r-p-h-y:
so for now is it only valid for Arnold?

m-u-r-p-h-y:
> I don't think it should be per renderer - and I agree with changing it to `Remove existing AOVs`

so the settings should be like this now

image

and the setting of removing AOV's should be valid for all renderers below.

moonyuet:
> > I don't think it should be per renderer - and I agree with changing it to `Remove existing AOVs`
> 
> so the settings should be like this now
> 
> image
> 
> and the setting of removing AOV's should be valid for all renderers below.

I have moved the setting to the Render Settings. And also added the remove_aovs options for Redshift.

LiborBatek (approved):
tested in ```maya2022``` and ```maya2023``` with ```Arnold 5.1.33```
All working as expected! No issues found...

m-u-r-p-h-y (changes requested):
there is some issue with vray

`# Error: KeyError: file D:\REPO\OpenPype\openpype\hosts\maya\api\lib_rendersettings.py line 179: 'remove_aovs'`

moonyuet:
> there is some issue with vray
> 
> `# Error: KeyError: file D:\REPO\OpenPype\openpype\hosts\maya\api\lib_rendersettings.py line 179: 'remove_aovs'`
> 
> and with redshift nothing is happening when running Set Render Settings or creating a Render Instance

The updated PR has fixed the error of removing aovs as well as removing the light-select aov

LiborBatek:
Did run new tests in maya with ```arnold``` and ```redshift``` and lastly with ```vray```

Regarding OP settings for AOVs (creation / deleting existing ones) All works fine in all three renderers.

Succesfuly adds predefined AOVs in OP settings and also reacts to "Remove Existing AOVs" switch too...

antirotor (approved):
Looks good.

>**Warning**
> Merging this needs updating v4 settings

m-u-r-p-h-y (changes requested):
Vray works ok, but there is still an issue with Redshift.

tested with Maya2022.4 Redshift 3.5.09

It only breaks if one of the required AOVs is Beauty. Probably connected to `Beauty -> BeautyAux` name change in 3.0.67?
image

keepAOVSettings

Other render settings are not set correctly either

image

m-u-r-p-h-y (approved):
Redshift 3.5.12
Maya 2022.4

AOVs work as expected

PR #4089, opened 2022-11-10
General: Collect entities plugin minor changes

iLLiCiTiT:
## Brief description
Renamed collector `CollectAvalonEntities` to `CollectContextEntities` and changed how asset and task names are received.

## Description
Renamed the plugin and asset/task name is taken from `context.data` instead of `legacy_io.Session`.

## Additionl information
We should change all remaining publish plugins to use context from `instance.data` (or `context.data` if is context based) instead of using `legacy_io.Session`.

## Testing notes:
In theory nothing should change or break. I'm not worried about local publishing, issues could come on farm or webpublisher. Please check if integration is going to right context (asset and task) in:
- [x] farm publishing
- [ ] webpublisher


m-u-r-p-h-y (approved):
published Maya scene with Arnold render to Deadline

image

kalisp (approved):
Seems to work with basic and studio publishing.

PR #4092, opened 2022-11-11
AfterEffects: publish multiple compositions

kalisp:
## Brief description
This allows publishing of multiple compositions in one publish process.

## Description
Previously there was limitation on single composition, most probably because of Deadline, but that seems to be resolved.

Now multiple composition could be published.

There are still some assertions:
- only single item in Render Queue per composition
- only single Output Module in Render Queue per composition (eg. file format)
- 
## Additional info
!! Requires extension redeployment !!

There is still outstanding question about naming convention for created subsets. As previously there was only single composition, using composition name in subset name didn't make any sense. Now multiple subsets need to be differentiated.
Currently only option is to manually modify `Variant` (eg. instead of `Main` but there `Composition1Main` for example).

This limits a position of composition name in subset name. I could implement adding composition name to the subset similarly as in PS, where `{layer}` placeholder is possible (when multiple subsets are created).

I have two issues there:
- not particularly fond of having different value of placeholder in every DCC (`layer` vs `composition` vs `renderLayer` vs...). But that is probably bigger discussion for different day
- `{composition}` should be probably used only when there are multiple subsets created. (In other case there will be difference between 'old' (without) and 'new' (with) subsets. Solution could be have explicit toggle `Use composition name in subset` toggle (default to False) on Creator.

Note: This is built on top of `release/3.15.x` branch, eg. Python 3.9 and Publisher only.

## Testing notes:
1. create multiple subsets, remember to modify `Variant` to differentiate them
2. publish locally
3. publish on Deadline
4. publish some legacy file


64qam (approved):

kalisp:
Updated, there could be multiple output modules per composition (eg. output formats). Each of them is added as a representation.

64qam (approved):

openpype/hosts/aftereffects/plugins/create/create_render.py
```diff
@@ -102,6 +132,22 @@ def get_pre_create_attr_defs(self):
...
+    def get_dynamic_data(self, variant, task_name, asset_doc,
+                         project_name, host_name, instance):
+        dynamic_data = super(RenderCreator, self).get_dynamic_data(
+            variant, task_name, asset_doc, project_name, host_name,
+            instance
+        )
+
+        if instance is not None:
+            composition_name = instance.get("composition_name")
+            if composition_name:
```
iLLiCiTiT, line 144: I guess it's on purpose to not fill `"composition"` key at all if instance does not have it? (all old instances without `"composition"` will require recreate?).

iLLiCiTiT (approved):
One note regarding subset name creation, otherwise looks good.

openpype/hosts/aftereffects/plugins/create/create_render.py
kalisp: Yes, old instances need to be recreated. I think it is easier than renaming them.

PR #4094, opened 2022-11-11
Maya: Collect render preserve handles even if frameStart + frameEnd matches context

antirotor:
## Fix

This is complementary fix for #3437, see:
https://github.com/pypeclub/OpenPype/pull/3437#issuecomment-1277604005

> **Note**
> This should be merged with #3437

### DIFF
--- a/openpype/hosts/maya/plugins/publish/collect_render.py
+++ b/openpype/hosts/maya/plugins/publish/collect_render.py
@@ -225,7 +225,7 @@ class CollectMayaRender(pyblish.api.ContextPlugin):
             handle_end = None
             if context.data.get("handles"):
                 handle_start = context.data.get("handles")
-                handle_end  = context.data.get("handles")
+                handle_end = context.data.get("handles")
 
             if (int(context.data['frameStartHandle']) == frame_start_render
                     and int(context.data['frameEndHandle']) == frame_end_render):  # noqa: W503, E501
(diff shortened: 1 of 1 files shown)


BigRoy:
This seems invalid - this should be replaced with https://github.com/pypeclub/OpenPype/pull/3437 and not merged with it. The if statements involved with this PR don't exist in the other PR.

The other PR removes the behavior that if by chance the start and end frame of a publish match what the asset is set to that based on that it overrides other settings. Instead, the logic should just always rely on the settings OR have another setting that forces it to be 'live' to what the asset is set to, like a "Use asset frame range" toggle or "Use custom frame range".

openpype/hosts/maya/plugins/publish/collect_render.py
```diff
@@ -221,22 +221,28 @@ def process(self, context):
...
 
-                handle_start = context.data['handleStart']
-                handle_end = context.data['handleEnd']
+                handle_start = context.data.get('handleStart') or handle_start
+                handle_end = context.data.get('handleEnd') or handle_end
                 frame_start = context.data['frameStart']
                 frame_end = context.data['frameEnd']
-                frame_start_handle = context.data['frameStartHandle']
-                frame_end_handle = context.data['frameEndHandle']
+
```
BigRoy, line 237: Continuing on my comment. Basically this "if" branch - should not exist. And collect render should receive a way to potentially apply a custom handle start and end if needed.

antirotor:
I am not sure about this - currently there is no way to specify custom handles for render layers at all. There might be when we switch maya to new publisher, but at that point it will be handled differently.

As I am thinking about it, I'll close this one because all it does it is fixing only the backwards compatibility with `handles` key that is of no concern for render family anyway.

PR #4095, opened 2022-11-14
Webpublisher: extension is lowercased in Setting and in uploaded files

kalisp:
## Brief description
Configuration of extension in `project_settings/webpublisher/publish/CollectPublishedFiles/task_type_to_family` was case sensitive. (eg. `uploaded.PNG` wasn't recognized if there was only `png` in extensions.)

## Testing notes:
1. publish file with upper case version of extension which is configured in `project_settings/webpublisher/publish/CollectPublishedFiles/task_type_to_family`
2. it shouldn't fail

### DIFF
--- a/openpype/hosts/webpublisher/plugins/publish/collect_published_files.py
+++ b/openpype/hosts/webpublisher/plugins/publish/collect_published_files.py
@@ -247,11 +247,17 @@ class CollectPublishedFiles(pyblish.api.ContextPlugin):
         for config in families_config:
             if is_sequence != config["is_sequence"]:
                 continue
+            extensions = config.get("extensions") or []
+            lower_extensions = set()
+            for ext in extensions:
+                if ext:
+                    ext = ext.lower()
+                    if ext.startswith("."):
+                         ext = ext[1:]
+                    lower_extensions.add(ext)
 
-            lower_extensions = [ext.lower()
-                                for ext in config.get("extensions", [''])]
-            if (extension.lower() in lower_extensions or
-                    lower_extensions[0] == ''):  # all extensions setting
+            # all extensions setting
+            if not lower_extensions or extension in lower_extensions:
                 found_family = config["result_family"]
                 break
 
--- a/openpype/hosts/webpublisher/plugins/publish/collect_published_files.py
+++ b/openpype/hosts/webpublisher/plugins/publish/collect_published_files.py
@@ -247,9 +247,9 @@ class CollectPublishedFiles(pyblish.api.ContextPlugin):
                 continue
 
             lower_extensions = [ext.lower()
-                                for ext in config.get("extensions", [])]
+                                for ext in config.get("extensions", [''])]
             if (extension.lower() in lower_extensions or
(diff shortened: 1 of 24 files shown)


openpype/hosts/webpublisher/plugins/publish/collect_published_files.py
```diff
@@ -244,8 +247,11 @@ def _get_family(self, settings, task_type, is_sequence, extension):
         for config in families_config:
             if is_sequence != config["is_sequence"]:
                 continue
-            if (extension in config["extensions"] or
-                    '' in config["extensions"]):  # all extensions setting
+
+            lower_extensions = [ext.lower()
+                                for ext in config.get("extensions", [''])]
+            if (extension.lower() in lower_extensions or
+                    lower_extensions[0] == ''):  # all extensions setting
```
iLLiCiTiT, line 254:
```suggestion
            extensions = config.get("extensions")
            lower_extensions = None
            if extens:s
                lower_extensions  = {ext.lower() for ext in extensions}
                if "" in lower_extensions:
                    lower_extensions = None

            if not lower_extensions or extension.lower() in lower_extensions:
            # all extensions setting
```

openpype/plugins/publish/extract_review.py
```diff
@@ -152,7 +152,7 @@ def _get_outputs_per_representations(self, instance, profile_outputs):
             if input_ext.startswith("."):
                 input_ext = input_ext[1:]
 
-            if input_ext not in self.supported_exts:
+            if input_ext.lower() not in self.supported_exts:
```
iLLiCiTiT, line 155: I would say it should be already prefilled on representation as lowered...???

kalisp: For WP it should be lowercased, but it might come in through something different (Traypublisher ?).

I don't think explicit lowercase could hurt.

64qam (approved):

openpype/plugins/publish/extract_review.py
iLLiCiTiT: I'm a little bit concerned about the `repre["ext"]` lowering because it could be used somewhere as source of extension for input file (didn't find any)? But at the same time it's used as integrated extension so it will also be traversed to integrated file (not good)...

kalisp: Not sure I follow.
In WP I am lowercasing as customer confirmed it is accident.
In ExtractReview I see potential issue on line 937, but that's extension of review. So it could happen, that you have `image_Art_Main.PNG` (in WP it won't happen, but in Tray publisher probably might) and `review_Art_Main.png`, but I still don't see an issue there.

kalisp: After discussion removed lower here.
Using lower here would hide possibly broken representation, as we would expect both repre["ext"] and repre["name"] be lowercased. In case the aren't, review won't get created >> someone will notice and fix issues on source representation.

64qam (approved):

PR #4098, opened 2022-11-15
layout publish more than one container issue

moonyuet:
## Brief description
fix the assertion error shown in` extract_layout.py `when publishing layout.

## Description
fix the assertion error shown in` extract_layout.py `when publishing layout.
image

If the user creates instance for the content which is not loaded from the workfile loader, it will have an assertion error to warn the user to use the loaded content. Unlike the previous version, it wont show an asset with more than one containers as the assertion error.
image

## Additional info
N/A

## Testing notes:
1. load the content from any families
2. rename the content
3. create layout instance for the loaded content(s)
4. publish it
5. uses the loader to load the layout content.
### Testing Notes 2:
1. create a new geometry or load a scene with geometry
2. create layout instance for the new geometry
3. publish it
4. it will trigger the assertion issue to tell you to use the loaded content

### DIFF
--- a/openpype/hosts/maya/plugins/create/create_layout.py
+++ b/openpype/hosts/maya/plugins/create/create_layout.py
@@ -8,6 +8,7 @@ class CreateLayout(plugin.Creator):
     label = "Layout"
     family = "layout"
     icon = "cubes"
+
     def __init__(self, *args, **kwargs):
         super(CreateLayout, self).__init__(*args, **kwargs)
 
--- a/openpype/hosts/maya/plugins/create/create_layout.py
+++ b/openpype/hosts/maya/plugins/create/create_layout.py
@@ -8,3 +8,10 @@ class CreateLayout(plugin.Creator):
     label = "Layout"
     family = "layout"
     icon = "cubes"
+    def __init__(self, *args, **kwargs):
+        super(CreateLayout, self).__init__(*args, **kwargs)
+
+
+        # enable this when you want to
+        # publish group of loaded asset
+        self.data["groupLoadedAssets"] = False
--- a/openpype/hosts/maya/plugins/create/create_layout.py
+++ b/openpype/hosts/maya/plugins/create/create_layout.py
@@ -11,8 +11,6 @@ class CreateLayout(plugin.Creator):
 
     def __init__(self, *args, **kwargs):
         super(CreateLayout, self).__init__(*args, **kwargs)
-
-
         # enable this when you want to
         # publish group of loaded asset
         self.data["groupLoadedAssets"] = False
--- a/openpype/hosts/maya/plugins/publish/extract_layout.py
+++ b/openpype/hosts/maya/plugins/publish/extract_layout.py
@@ -38,14 +38,19 @@ class ExtractLayout(publish.Extractor):
             container_list = cmds.ls(project_container)
             assert len(container_list) == 1, \
                 "Please create instance with loaded asset!"
-            # list the children of the containers
-            grp_name = asset.split(':')[0]
-            container_sel = cmds.ls("{}*_CON".format(grp_name))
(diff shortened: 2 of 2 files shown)


LiborBatek:
Hi, I have tested it first with develop branch before switching to yours just to be sure whats happening before your fix. And it did warning as you mentioned "asset with more than one containers".

However when using your branch and tested it on the same scene it went smoothly through Publish and no errors showed.
When I have tested to load the ```layout``` from loader to new scene it all worked normally and the asset created in that layout scene were referenced in withou problems...which make sense. My ```layout asset``` consists of loaded assets and one mesh object not loaded but created in the layout scene (originating there).

So in the end I didnt obtain any warning to load asset using asset loader but it all went smooth without any trouble...

I will do some more tests with different scenarios like duplicate loaded assets so it creates new instances instead of creating brand new objects originating in the layout scene (not loaded). 
Will update on this soon. If you have any comment on this pls do. Thx!

moonyuet:
> Hi, I have tested it first with develop branch before switching to yours just to be sure whats happening before your fix. And it did warning as you mentioned "asset with more than one containers".
> 
> However when using your branch and tested it on the same scene it went smoothly through Publish and no errors showed. When I have tested to load the `layout` from loader to new scene it all worked normally and the asset created in that layout scene were referenced in withou problems...which make sense. My `layout asset` consists of loaded assets and one mesh object not loaded but created in the layout scene (originating there).
> 
> So in the end I didnt obtain any warning to load asset using asset loader but it all went smooth without any trouble...
> 
> I will do some more tests with different scenarios like duplicate loaded assets so it creates new instances instead of creating brand new objects originating in the layout scene (not loaded). Will update on this soon. If you have any comment on this pls do. Thx!

I have updated the testing notes in github which explains when the assertion error is triggered.Please let me know if you want further adjustment. Thanks!

LiborBatek:
So I have followed testing steps 02 (scene with local geometries put to ```layout``` )

it wont give me any warning or error during publish and all goes smooth (see pic)
image

So the outcome of the test is pretty much the same as when tested for a first time...no any error/warning msg...

moonyuet:
> So I have followed testing steps 02 (scene with local geometries put to `layout` )
> 
> it wont give me any warning or error during publish and all goes smooth (see pic) image
> 
> So the outcome of the test is pretty much the same as when tested for a first time...no any error/warning msg...

this is a good catch. I have updated PR for triggering the assertion error if they added both loaded and unloaded contents in the scene.

LiborBatek:
Ok, cool, but take a note this also counts when only local geometries are present in the layout subset (no warnings and layout published ok)

moonyuet:
image
image
image
image
I have tried 4 scanerios for extracting the layout from the loaded content. It wont be successfully published if you contain any geometries which are not in the "AVALON_CONTAINER" group

LiborBatek:
Ok I will pull your branch again and try it out... could be there some difference in your source code compared to one I pull?! I cant think of anything than our version somehow differ so behaving differently...

LiborBatek:
Ok have tested it with local geometry besides loaded assets in ```layout``` subset....gives error during publish but not as yours just this:
OP_Layout_Msg01

this counts also when local asset made by duplicating the loaded asset (basically same scenario as above) still gives me this error:
OP_Layout_Msg02

Seems odd that didnt trigger same ```error msg``` as you stated in your screens. I guess need revision still.

LiborBatek (changes requested):
Not fully working needs changes. See comments for more info.

LiborBatek (approved):
Tested with local assets and duplicates of the loaded ones and Publishes ok giving user warning of using local assets not being optimal for workflow in OP.

Seems ok now.

antirotor (approved):

PR #4102, opened 2022-11-16
Fix: Locally copied version of last published workfile is not incremented

Tilix4:
### Fix 1
When copied, the local workfile version keeps the published version number, when it must be +1 to follow OP's naming convention.

### Fix 2
Local workfile version's name is built from anatomy. This avoids to get workfiles with their publish template naming.

### Fix 3
In the case a subset has at least two tasks with published workfiles, for example `Modeling` and `Rigging`, launching `Rigging` was getting the first one with the `next` and trying to find representations, therefore `workfileModeling` and trying to match the current `task_name` (`Rigging`) with the `representation["context"]["task"]["name"]` of a Modeling representation, which was ending up to a `workfile_representation` to `None`, and exiting the process.

Trying to find the `task_name` in the `subset['name']` fixes it.

### Fix 4
Fetch input dependencies of workfile.


iLLiCiTiT:
Shouldn't the copied file actually have version 1?

Tilix4:
> Shouldn't the copied file actually have version 1?

Actually it's very confusing for users when the local version don't match the published ones.

iLLiCiTiT:
Am I wrong that the prelaunch hook is actually creating first workfile if there is not any? So it is actually creating first workfile version... ?

Tilix4:
> Am I wrong that the prelaunch hook is actually creating first workfile if there is not any? So it is actually creating first workfile version... ?

No you'are not, but it is very confusing if you know that 9 have been published and when opening it, you get a `v001`, but with the content of the `v009` :exploding_head:

iLLiCiTiT:
Ok but in case a workfile template should be used instead of just copying the file as is.

Tilix4:
> Ok but in case a workfile template should be used instead of just copying the file as is.

Isn't it handled by the workfile template hook?

Tilix4:
In fact, shouldn't use the `Save as` code and correcting the version number, because when copied it has the name of the published one and not the naming template of the workfile.

openpype/hooks/pre_copy_last_published_workfile.py
```diff
@@ -109,11 +114,14 @@ def execute(self):
...
                 )
-                if subset["data"].get("family") == "workfile"
-                # Legacy compatibility
-                or "workfile" in subset["data"].get("families", {})
+                if (
+                    subset["data"].get("family") == "workfile"
+                    # Legacy compatibility
+                    or "workfile" in subset["data"].get("families", {})
+                )
+                and task_name in subset["name"]
```
iLLiCiTiT, line 124: This condition is not right. You can "not use" task name in subset name, also in most of cases the task name have capitalized first letter if is used in subset name (so "animation" would become "Animation"). I think this requiers more complex logic that cannot be solved in single if statement inside loop...

openpype/hooks/pre_copy_last_published_workfile.py
```diff
@@ -171,7 +179,32 @@ def execute(self):
...
+            project_name, asset_name, task_name, host_name
+        )
+        local_workfile_path = get_last_workfile(
+            local_workfile_dir,
+            str(anatomy.templates["work"]["file"]),
+            workfile_data,
+            [os.path.splitext(published_workfile_path)[1]],
+            full_path=True,
+        )
+
```
iLLiCiTiT, line 193:
```
local_workfile_path["version"] = last_version_doc["name"] + 1
```

openpype/hooks/pre_copy_last_published_workfile.py
```diff
@@ -171,7 +179,32 @@ def execute(self):
...
+        local_workfile_version = get_version_from_path(local_workfile_path)
+        published_workfile_version = get_version_from_path(
+            published_workfile_path
+        )
+        local_workfile_path = version_up(
+            local_workfile_path.replace(
+                local_workfile_version, published_workfile_version
+            )
+        )
+
```
iLLiCiTiT, line 205:
```suggestion
        # Build workfile to copy and open's name from anatomy template settings
        workfile_data = get_template_data_with_names(
            project_name, asset_name, task_name, host_name
        )
        workfile_data["version"] = last_version_doc["name"] + 1
        template_key = get_workfile_template_key(
            project_name,
            task_type,
            host_name,
            project_settings
        )
        anatomy_result = anatomy.format(workfile_data)
        local_workfile_path = anatomy_result[template_key]["path"]

```

openpype/hooks/pre_copy_last_published_workfile.py
iLLiCiTiT: Something like this
```
        filtered_subsets = [
            subset
            for subset in get_subsets(
                project_name,
                asset_ids=[asset_doc["_id"]],
                fields=["_id", "name", "data.family", "data.families"],
            )
            if (
                subset["data"].get("family") == "workfile"
                # Legacy compatibility
                or "workfile" in subset["data"].get("families", {})
            )
        ]
        if not filtered_subsets:
            return

        matching_subset = None
        if len(filtered_subsets) > 1:
            low_task_name = task_name.lower()
            for subset in filtered_subsets:
                subset_name = subset["name"].lower()
                if low_task_name in subset_name:
                    matching_subset = subset
                    break

        if matching_subset is None:
            matching_subset = filtered_subsets[0]

        subset_id = matching_subset["_id"]
```

Tilix4: Changed

openpype/client/entities.py
```diff
@@ -518,6 +518,53 @@ def get_subset_families(project_name, subset_ids=None):
     return set()
 
 
+def match_subset_id(
```
iLLiCiTiT, line 521: Not sure if this function should be here. Also it's name does not tell much about what it does? Does it find subset based on family, asset and task? If yes then it make sense only for very limited use cases. I guess it's used to find workfile subset in that case I would mark it as helper function in openpype/pipeline/workfiles.

BTW without Python 3 syntax. We still must support Python 2

openpype/hooks/pre_copy_last_published_workfile.py
```diff
@@ -31,16 +32,12 @@ def execute(self):
         2- Check if workfile in work area doesn't exist
         3- Check if published workfile exists and is copied locally in publish
         4- Substitute copied published workfile as first workfile
+           with incremented version by +1
 
         Returns:
             None: This is a void method.
         """
 
-        sync_server = self.modules_manager.get("sync_server")
```
iLLiCiTiT, line 39: So this will happen all the time even if sync server is not available or enabled?

openpype/client/entities.py
iLLiCiTiT: As I'm looking for the usage I would say the functionality should be moved to sync server module and `pre_copy_last_published_workfile.py` should be launch hook inside sync server too (with lower order so it's processed before the global one).

openpype/hooks/pre_copy_last_published_workfile.py
Tilix4: I think this is a mistake.

openpype/client/entities.py
Tilix4: Okay, currently there are no hook for `sync_server` right? Therefore you expect me to set the `launch_hooks` up with this one?

openpype/client/entities.py
```diff
@@ -1032,6 +1079,31 @@ def get_representation_by_name(
     return conn.find_one(query_filter, _prepare_fields(fields))
 
 
+def get_representation_by_task(
```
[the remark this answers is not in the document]
Tilix4: @Sharkitty Are you using this one in #4176 ?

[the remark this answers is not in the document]
Sharkitty: Yes, it's called in the `DownloadLastWorkfile` Operator.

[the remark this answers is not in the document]
iLLiCiTiT: The `DownloadLastWorkfile` is fully dependent on sync server so if the functions are moved to sync server it should not block it.

openpype/client/entities.py
iLLiCiTiT: Yes, sync server must implement `get_launch_hook_paths` to return the path to it. Also the functions related to workfile workflow in `openpype/client/entities` should be moved to sync module too. Those functions are utils functions for workfile sync workflow and it would be confusing to have them in public api.

BTW keep the original `pre_copy_last_published_workfile.py` without sync server logic in global hooks please.

iLLiCiTiT: I would recommend renaming `match_subset_id` to `get_matching_subset_id` and `get_representation_by_task` should handle cases when representation does not have `"task"` key in `"context"`.

openpype/client/entities.py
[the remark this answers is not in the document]
Tilix4: Is it what we are supposed to do? Move `get_representation_by_task` and `get_matching_subset_id` to `sync_server.py`

LiborBatek:
Tried to test it out using older workfile than published workfile but no any pop up window appeared...not sure if I am doing something wrong but here are my workfiles with dates and location...
B3d_workfile_ver01

LiborBatek:
@Tilix4  opps. thanks, my bad! you are right!

openpype/client/entities.py
iLLiCiTiT: >  Also the functions related to workfile workflow in openpype/client/entities should be moved to sync module too.

This is still true, both functions do not belong to `entities.py`. They are both dependent on specific workflows and expectations that are not requirements (e.g. task name in subset name or representation with task in context). As I'm thinking about it they are just filtering functions and they don't have to query from server at all `get_representation_by_task` could be changed to `fitler_representations_by_task` with `representations` as an argument and `get_matching_subset_id` also another filtering function `filter_subsets_by_family_and_task` with `subsets` as an argument.

Tilix4: Okay, in fact this split into functions was made by anticipation for another feature #4176 but its implementation is still changing and uncertain, therefore we'll create these utils functions in this upcoming contribution if you don't mind, and keep this one light.

Tilix4: Okay, I removed them from this PR, and we'll find a better place in the upcoming PR if we still even need these functions.
Thanks for your advises.

openpype/modules/sync_server/sync_server.py
```diff
@@ -229,6 +244,220 @@ def _get_configured_sites_from_setting(module, project_name, project_setting):
...
+    # Get version patter for regex search
+    version_pattern = anatomy.templates[template_key]["version"]
+    version_pattern = re.sub(
+        r"{version.*?}",
+        r"([0-9]+)",
+        version_pattern,
+    )
+
+    # Get local workfile version number
+    last_local_workfile_version = None
```
iLLiCiTiT, line 413: There is function `get_last_workfile_with_version` in `openpype.pipeline.workfile` which does exactly this.

Tilix4: Changed with `get_last_workfile_with_version`

openpype/modules/sync_server/sync_server.py
```diff
@@ -229,6 +244,203 @@ def _get_configured_sites_from_setting(module, project_name, project_setting):
     return configured_sites
 
 
+def get_last_published_workfile_path(
```
iLLiCiTiT, line 247: There is too much arguments in this function. Also I'm not sure if it should be a function?
At the end its just this
```
path = None
if workfile_representation:
    path = get_representation_path_with_anatomy(
        workfile_representation, anatomy
    )
```

openpype/modules/sync_server/sync_server.py
```diff
@@ -229,6 +244,203 @@ def _get_configured_sites_from_setting(module, project_name, project_setting):
...
+
+    project_settings = get_project_settings(project_name)
+    template_key = get_workfile_template_key(
+        task_name, host_name, project_name, project_settings
+    )
+
+    # Get local workfile version number
+    _, last_local_workfile_version = get_last_workfile_with_version(
+        get_workdir(
+            get_project(project_name),
```
iLLiCiTiT, line 407: Reuse of project doc. Requires to replace import of `get_template_data_with_names` by `get_template_data`.
```suggestion
    project_doc = get_project(project_name)

    # Get workfile data
    workfile_data = get_template_data(
        project_doc, asset_doc, task_name, host_name
    )

    extension = last_published_workfile_path.split(".")[-1]

    project_settings = get_project_settings(project_name)
    template_key = get_workfile_template_key(
        task_name, host_name, project_name, project_settings
    )

    # Get local workfile version number
    _, last_local_workfile_version = get_last_workfile_with_version(
        get_workdir(
            project_doc,
```

openpype/modules/sync_server/sync_server.py
Tilix4: Done

kalisp:
Replaced by https://github.com/ynput/OpenPype/pull/4722

PR #4110, opened 2022-11-18
General: Collection Audio speed up

iLLiCiTiT:
## Brief description
Speed up Collect Audio publish plugin.

## Description
Collect audio did do a lot of database calls per instance which slowed down collection phase a lot if there was more instances. The plugin was converted from instance plugin to context plugin. That helps to query required data from database for all instances at once without doing it for each individual instance.

## Additional info
The plugin is not using `legacy_io.Session` to get context information. Instead of using `get_representation_path` is used `get_representation_path_with_anatomy`.

## Testing notes:
1. Create at least 10 instances each with different asset (e.g. in tray publisher)
2. Run collection
3. Compare speed with and without this PR


openpype/plugins/publish/collect_audio.py
```diff
@@ -1,21 +1,27 @@
...
+    get_assets,
+    get_subsets,
+    get_last_versions,
     get_representations,
 )
-from openpype.pipeline import (
-    legacy_io,
-    get_representation_path,
-)
+from openpype.pipeline import get_representation_path_with_anatomy
```
jakubjezek001, line 10:
```Filepath:
C:\CODE\__PYPE\OpenPype\openpype\plugins\publish\collect_audio.py

Traceback:
File "C:\CODE\__PYPE\OpenPype\openpype\pipeline\publish\lib.py", line 324, in publish_plugins_discover
six.exec_(f.read(), module.__dict__)
File "", line 10, in 
ImportError: cannot import name 'get_representation_path_with_anatomy' from 'openpype.pipeline' (C:\CODE\__PYPE\OpenPype\openpype\pipeline\__init__.py)```

jakubjezek001 (changes requested):

PR #4116, opened 2022-11-21
Publisher: Hide unknown publish values

iLLiCiTiT:
## Brief description
Hide unknown publish plugin values to avoid user's confusion.

## Description
We showed unknown attributes stored on instance to user which could easily cause confusion because the showed key would not tell user (or us) almost nothing. But it make sense to show the unknown attribute on instances thus there were kept untouched. Mostly visible after changing optional validators as required.

Added option to define attribute definition as "hidden" or "disabled" (both only for UI purposes). Also was added option to define hidden attribute definition that hold any value defined by code.

## Testing notes:
1. Set validator supported by Publisher tool as optional
2. Open a host which support Publisher tool and create an instance where the validator would show
3. Close the host and change settings of the validator to be required
4. Open host again and open Publisher tool
5. The `active` value of validator should not be visible on the instance anymore

Resolves https://github.com/pypeclub/OpenPype/issues/4099

### DIFF
--- a/openpype/lib/attribute_definitions.py
+++ b/openpype/lib/attribute_definitions.py
@@ -253,6 +253,26 @@ class UnknownDef(AbtractAttrDef):
         return value
 
 
+class HiddenDef(AbtractAttrDef):
+    """Hidden value of Any type.
+
+    This attribute can be used for UI purposes to pass values related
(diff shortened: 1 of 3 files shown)


jakubjezek001 (approved):
works well

PR #4117, opened 2022-11-22
General: Fix variable check in collect anatomy instance data

iLLiCiTiT:
## Brief description
Collect anatomy instance data is right variable to skip processing of instance.


kalisp (approved):
Seems legit.

PR #4120, opened 2022-11-22
Publisher: Open on specific tab

iLLiCiTiT:
## Brief description
Publisher has option to be opened on specific tab when showed.

## Description
It is possible to show publisher window on specific tab. Added option to open publisher on create/publish tab on first restart when there is at least one instance. Tasks combobox is extended to full width.

## Testing notes:
1. Pass `tab` keyword argument with value `"create"` or `"publish"` when opening publisher.
2. Check if the tab is set to what was passed


jakubjezek001 (approved):
works as advertised ;)

PR #4121, opened 2022-11-22
Photoshop: bug with pop-up window on Instance Creator

friquette:
## Brief description
In photoshop on Windows when we select multiple layers to create an instance, there's a pop-up window that appears behind the Instance Creator window, but we can't select this window.

## Description
The pop-up asks the users if they want to create one instance per layer or not, but we can't select what to do since it's not the active window (the Instance Creator window is). So the solution seems to make the Creator Instance window the parent of the pop-up window.

With the pop-up stuck behind the Instance Creator window:
image

With the Instance Creator window as parent of the pop-up:
image

### DIFF
--- a/openpype/hosts/photoshop/plugins/create/create_legacy_image.py
+++ b/openpype/hosts/photoshop/plugins/create/create_legacy_image.py
@@ -29,7 +29,8 @@ class CreateImage(create.LegacyCreator):
             if len(selection) > 1:
                 # Ask user whether to create one image or image per selected
                 # item.
-                msg_box = QtWidgets.QMessageBox()
+                active_window = QtWidgets.QApplication.activeWindow()
+                msg_box = QtWidgets.QMessageBox(parent=active_window)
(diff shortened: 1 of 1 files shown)


kalisp (approved):
Works.

kalisp:
Thank you for contribution :).
I was meaning to fix it, but totally forgot about it.

If your studio is using PS, you might be interested in checking `Publisher` to which we will be moving in next major release. 
It is already accessible in develop under Experimental tools. (For each artist configurable in Tray>>Settings>>Experimental Tools>>New Publisher). You would see new dialog under `Experimental tools` in PS OP menu.

For fans of bleeding edge there were some minor updates to PS (and AE) currently living in https://github.com/pypeclub/OpenPype/tree/release/3.15.x (extension reinstall is needed).

Here are some screenshot how Publisher looks like in PS https://github.com/pypeclub/OpenPype/blob/release/3.15.x/website/docs/artist_hosts_photoshop.md

Any feedback welcomed.

PR #4123, opened 2022-11-22
Create extract JSON file for TvPaint

friquette:
## Brief description
Create a JSON extractor for TvPaint in order to publish JSON file with all the data needed from each layer for compositing in After Effects (or any other compatible software).


iLLiCiTiT:
Both PRs (this and https://github.com/pypeclub/OpenPype/pull/4122) don't seem to target any specific layers but all of the exporting is happing on `renderPass` families. So there will be same output of psd (or json in the other PR) with exactly the same content for each instance which has `renderPass` family which doesn't seems to be right.

In this PR there is more issues. Like the json is created all the time but can be "not used" and does not represent the `renderPass` instance but all layers in the scene (no matter if are marked as `renderPass`). The json "creation" is based on `renderPass` family but "usage" is based on `renderLayer` -> there can be `renderLayer` without `renderPass`. Also when json is extracted it looks like the current rendereing is skipped.

I didn't test it so maybe I'm wrong but the approach does not seem to be 100% right. Can you more elaborate in description what is expected result and what will happen in all new cases? Because it's not clear from the code.

friquette:
> 

The goal for the json and psd is to generate a file with all the datas of each renderLayer in the published renderPass.

But the names of the families are a bit confusing in our side. Is the renderPass composed of several renderLayer, or is it the opposite ? Maybe the main issue comes from a misunderstanding of this.

iLLiCiTiT:
> But the names of the families are a bit confusing in our side. Is the renderPass composed of several renderLayer, or is it the opposite ? Maybe the main issue comes from a misunderstanding of this.

```
# Example layers in TVPaint
- Bob_arm
- Bob_head
- Bob_mouth
- Bob_eye
- Bob_mouth
- Carol_arm
- Carol_head
- Train
```

Render layer can be considered as a very specific object -> from example it would be `"Bob"`, `"Carol"` and `"Train"`. Result of render layer is "beauty" -> all layers marked with color group. Render layer may have passes (subparts) from example it could `"arm"` and `"head"` for both `"Bob"` and `"Carol"`. For render layer `"Bob"` would render pass `"head"` be combination of 3 layers (head, mouth, eye) -> how the passes are defined can be based on production usage. `"Train"` render layer can have 0 render passes because it's not needed for production to have it separated.

So render passes are more specific parts of render layer and you must have a render layer to be able define a render pass.

I'm not sure about the usage of the json. From the code it looks like you want to create single json for whole scene. In that case it should not be done this way, that's why I wanted to describe the usecase.

iLLiCiTiT:
>  Can you more elaborate in description what is expected result and what will happen in all new cases? Because it's not clear from the code.

I have to ask again. Can you describe more what is expected result? It's replacing current rendering of render passes but keeps render layers untouched. Also the work with json file as structure is confusing and unclear, should there be more then one json output: one json per render pass, one per render layer or one per publish? I still think that it should have separated extractor from the current one as render passes may contain more then one layer (I assume that is against the logic of json file) but I can't tell without the description and use cases.

friquette:
> > Can you more elaborate in description what is expected result and what will happen in all new cases? Because it's not clear from the code.
> 
> I have to ask again. Can you describe more what is expected result? It's replacing current rendering of render passes but keeps render layers untouched. Also the work with json file as structure is confusing and unclear, should there be more then one json output: one json per render pass, one per render layer or one per publish? I still think that it should have separated extractor from the current one as render passes may contain more then one layer (I assume that is against the logic of json file) but I can't tell without the description and use cases.

In TVPaint we can export a json file to After Effects. This is useful to keep the structure of the project as it is in TVPaint, by exporting the layers to keep working in After Effects. So we need a json file by publish, with all the renderPass. Here is an example of the json file we get:

```
{
    "version": {
        "major": 5,
        "minor": 1
    },
    "project": {
        "format": {
            "extension": "png"
        },
        "camera": {
            "width": 4240,
            "height": 2560,
            "field": 0,
            "framerate": 24.0,
            "pixelaspectratio": 1.0,
            "antialiasing": 1
        },
        "clip": {
            "name": "Sans Titre",
            "width": 4240,
            "height": 2560,
            "framerate": 24.0,
            "pixelaspectratio": 1.0,
            "image-count": 90,
            "bg": {
                "red": 255,
                "green": 255,
                "blue": 255
            },
            "markin": {
                "status": true,
                "value": 0
            },
            "markout": {
                "status": true,
                "value": 8
            },
            "camera": {
                "mode": 2,
                "profile": "2 0 1 0 0 1 1 ",
                "motion-blur": 0.0,
                "lock-all": false,
                "lock-x": false,
                "lock-y": false,
                "lock-angle": false,
                "lock-scale": false,
                "points": [],
                "positions": [
                    {
                        "x": 0.0,
                        "y": 0.0,
                        "angle": 0.0,
                        "scale": 1.0
                    },
                    
                ]
            },
            "layers": [
                [
                    {
                        "name": "SH400_CH_BANKSY_CU",
                        "visible": true,
                        "position": 0,
                        "opacity": 255,
                        "start": 0,
                        "end": 89,
                        "pre-behavior": 0,
                        "post-behavior": 3,
                        "blending-mode": "Color",
                        "group": {
                            "red": 214,
                            "green": 183,
                            "blue": 22
                        },
                        "link": [
                            {
                                "instance-index": 0,
                                "instance-name": "1",
                                "file": "M:\\/Test_CRCR_OpenPype/Shots/SQ01/SQ01_SH020/publish/renderPass/renderRough_Anim_Perso_Line/v125/test_crcr_openpype_SQ01_SH020_renderRough_Anim_Perso_Line_v125.0000.png",
                                "images": [
                                    0
                                ]
                            },
                            
                        ],
                        "repeat": []
                    },
                    {
                        "name": "SH400_CH_BANKSY_SH",
                        "visible": true,
                        "position": 1,
                        "opacity": 89,
                        "start": 0,
                        "end": 88,
                        "pre-behavior": 0,
                        "post-behavior": 3,
                        "blending-mode": "Color",
                        "group": {
                            "red": 214,
                            "green": 183,
                            "blue": 22
                        },
                        "link": [
                            {
                                "instance-index": 0,
                                "instance-name": "1",
                                "file": "M:\\/Test_CRCR_OpenPype/Shots/SQ01/SQ01_SH020/publish/renderPass/renderRough_Anim_Perso_Shadow/v125/test_crcr_openpype_SQ01_SH020_renderRough_Anim_Perso_Shadow_v125.0000.png",
                                "images": [
                                    0
                                ]
                            },
                        ],
                        "repeat": []
                    },
                    {
                        "name": "Character_Miror_SH",
                        "visible": true,
                        "position": 3,
                        "opacity": 89,
                        "start": 0,
                        "end": 88,
                        "pre-behavior": 0,
                        "post-behavior": 3,
                        "blending-mode": "Color",
                        "group": {
                            "red": 0,
                            "green": 0,
                            "blue": 255
                        },
                        "link": [
                            {
                                "instance-index": 0,
                                "instance-name": "1",
                                "file": "M:\\/Test_CRCR_OpenPype/Shots/SQ01/SQ01_SH020/publish/renderPass/renderRough_Anim_Perso_Miror_SH/v041/test_crcr_openpype_SQ01_SH020_renderRough_Anim_Perso_Miror_SH_v041.0000.png",
                                "images": [
                                    0
                                ]
                            },
                        ],
                        "repeat": []
                    },
                    {
                        "name": "Character_Miror_CL",
                        "visible": true,
                        "position": 4,
                        "opacity": 255,
                        "start": 0,
                        "end": 89,
                        "pre-behavior": 0,
                        "post-behavior": 0,
                        "blending-mode": "Color",
                        "group": {
                            "red": 0,
                            "green": 0,
                            "blue": 255
                        },
                        "link": [
                            {
                                "instance-index": 0,
                                "instance-name": "1",
                                "file": "M:\\/Test_CRCR_OpenPype/Shots/SQ01/SQ01_SH020/publish/renderPass/renderRough_Anim_Perso_Miror_CL/v041/test_crcr_openpype_SQ01_SH020_renderRough_Anim_Perso_Miror_CL_v041.0000.png",
                                "images": [
                                    0
                                ]
                            },
                        ],
                        "repeat": []
                    },
                ]
            ]
        }
    }
}
```
As you can see, each renderPass as a "position" key to place them correctly in the timeline and a "group" key to separate them into the right renderLayer based on the color.

PR #4124, opened 2022-11-23
Changed the UOpenPypePublishInstance to use the UDataAsset class

Casgen:
## Brief description
The UOpenPypePublishInstance now uses the UDataAsset to store references to saved assets.

## Description

- UOpenPypePublishInstance derives from the UPrimaryDataAsset.
- Assets are now either Internal (UOpenPypePublishInstance monitors it's own directory and subdirectories for changes) or external (Assets which can be added in from other locations).
- Created new functions to get external, internal or all assets combined.
- AssetData variables use TSets to ensure there are no duplicate assets being added. A check is also implemented between the external and internal TSets which also monitor, if a user isn't trying to add an duplicate asset to external container, if it already exists in the internal one.
- There is also a check to make sure the user doesn't include the UOpenPypePublishInstance itself.

## Additional info
To allow to edit the external asset container (TSet), user has to tick the 'Add External Assets' to allow him to expose the container. Note that externally added are not managed automatically whatsoever. They are managed manually! Also if the variable is unticked, the GetAllAssets() function behaves differently (It will not include the External Assets).

The TSets also use TSoftObjectPtrs. That means it references the assets on the drive, not the memory. The getter functions already take care of initializing into memory though.

## Testing notes:

Creation of the Publish Instance
1. Open the Unreal Project
2. Create a new Camera through the OpenPype Creator {A new UOpenPypePublishInstance should be created)
3. Check in the UOpenPypePublishInstance Asset, if the newly created "cameraDefault.uasset" is automatically included in the 'AssetDataInternal'. Make sure to also save the newly created asset "cameraDefault.uasset".

Monitoring of newly added assets
1. When a UOPPublishInstance is created, create a new asset in the same directory or subdirectory where the UOPPublishInstance asset is located (for instance Material)
2. Check in the details panel of UOPPublishInstance, if the newly created asset is added into the "AssetDataInternal"

Checking if the UOPPublishInstance can not be added into the "AssetDataExternal" Container.
1. Open the details panel in UOPPublishInstance
2. Try to add the the same or other UOPPublishInstance asset.
3. Editor shouldn't allow the user to add this kind of asset, and is warned by a notification.

Prohibition of adding a duplicate asset into 'AssetDataExternal', which is already contained in 'AssetDataInternal'.
1. UOPPublishInstance has to have at least one asset included in 'AssetDataInternal'.
2. Try to add a duplicate asset contained in 'AssetDataInternal' to 'AssetDataExternal'.
3. User should be notified by the editor, that he can not add Assets to the external container which are already contained in the Internal container.

### DIFF
--- a/openpype/hosts/unreal/integration/UE_5.0/Source/OpenPype/Private/OpenPypePublishInstance.cpp
+++ b/openpype/hosts/unreal/integration/UE_5.0/Source/OpenPype/Private/OpenPypePublishInstance.cpp
@@ -35,6 +35,7 @@ UOpenPypePublishInstance::UOpenPypePublishInstance(const FObjectInitializer& Obj
 	AssetRegistryModule.Get().OnAssetRemoved().AddUObject(this, &UOpenPypePublishInstance::OnAssetRemoved);
 	AssetRegistryModule.Get().OnAssetUpdated().AddUObject(this, &UOpenPypePublishInstance::OnAssetUpdated);
 	
+	
 }
 
 void UOpenPypePublishInstance::OnAssetCreated(const FAssetData& InAssetData)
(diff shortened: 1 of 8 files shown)


openpype/hosts/unreal/integration/UE_5.0/Source/OpenPype/Private/OpenPypePublishInstance.cpp
```diff
@@ -2,107 +2,151 @@
 
 #include "OpenPypePublishInstance.h"
 #include "AssetRegistryModule.h"
+#include "AssetToolsModule.h"
+#include "NotificationManager.h"
```
antirotor, line 6: for some reason it doesn't work for me but this does:
```suggestion
#include "Framework/Notifications/NotificationManager.h"
```

antirotor (changes requested):
Except the path in include, everything works! Thanks.

antirotor (approved):

PR #4126, opened 2022-11-23
Ftrack: Editorial asset sync issue

iLLiCiTiT:
## Brief description
Probably fixing most of current issue with editorial publishing. Auto sync is not turned off (and back to on) during ftrack hierarchy integration which caused that assets without unavailable entities were marked as archived and at the same time were created new assets which lead to double asset documents and subsets created during editorial publishing lead to the archived one.

## Description
Auto sync is not turned on/off. Entities for filling ftrack entities are prequeried for whole project instead of querying all entities separatelly. Custom attributes are "created" or "updated" to table `"CustomAttributeValue"` instead of `"ContextCustomAttributeValue"` (which could cause that entity did not have filled the value as expected. Recursive functions were modified to queu based. Avoid requery of data from ftrack.

## Additional info
There is still room for improvement in the hierarchy plugin and some change on ftrack event sync but for now this should fix most of current issues.

## Testing notes:
Editorial publishing should work the same or better way.
1. Publish editorial which would create at least 15 shots (assets)
2. None of that should cause issues and none of that should at the end be marked as archived in openpype mongo
3. Custom attributes should be propagated to new created entities on ftrack

### DIFF
--- a/openpype/modules/ftrack/plugins/publish/integrate_hierarchy_ftrack.py
+++ b/openpype/modules/ftrack/plugins/publish/integrate_hierarchy_ftrack.py
@@ -229,10 +229,10 @@ class IntegrateHierarchyToFtrack(pyblish.api.ContextPlugin):
 
     def import_to_ftrack(self, project_name, hierarchy_context):
         # Prequery hiearchical custom attributes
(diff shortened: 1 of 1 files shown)


jakubjezek001 (approved):
it worked

PR #4131, opened 2022-11-24
Maya: create and publish bounding box geometry

moonyuet:
## Brief description
A feature of creating instance for the animated model and converting it into bounding box before publishing it as a workfile in abc format

## Description
A feature of creating instance subset for the animated model and converting the model into the animated bounding box before publishing it as a workfile in Alembic format. After the publish, the workfile with bounding box can be loaded throughout the loader. 
image
image

## Additional info
n/a

## Testing notes:
1. create the instance of Proxy Alembic with your selected object
2. Publish it.
3. you can load the animated bounding box from the Proxy Alembic family from the loader.


LiborBatek:
Have tested it in my test anim scene with character rig animated and during publishing get these errors:
image

Briefly what I did... I have created  ```proxyAbcMain``` subset and put subgroup from my character rig containg just the model geometries.
image

LiborBatek (changes requested):
Didnt work as expected. Pls see comments.

LiborBatek:
still gives me this error...no matter if just the top GRP for geometries present in the subset for ```proxyAbcMain``` or just the actual geometries as on scrnshot
image

LiborBatek (changes requested):
Still not working and gives error msg.

LiborBatek (approved):
Works ok. Also tested to load it to the scene as ```proxyAlembic``` family.

image

moonyuet:
I updated the PR to update the codes for the option of removing bounding box after publish.

mkolar:
Have we tested with animated geometry?

LiborBatek:
Yes it was an animated character....its bounding boxes changed during playback. So working...

PR #4134, opened 2022-11-24
General: Legacy io not used in global plugins

iLLiCiTiT:
## Brief description
Session data from `legacy_io` are not used in most of global plugins.

## Description
Usage of `legacy_io.Session` was replaced with using data stored on pyblish context. Some plugins are still using legacy io but those plugins already have opened PR changing them or replacement in them would need more overall changes.

## Additional information
Replacement usage of session keys stored in context.data:
- `"AVALON_APP"` -> `"hostName"`
- `"AVALON_PROJECT"` -> `"projectName"`
- `"AVALON_ASSET"` -> `"asset"`
- `"AVALON_TASK"` -> `"task"`

## Testing notes:
- [ ] Farm cleanup plugin is working
- [ ] Collect anatomy instance data has right project locally and on farm
- [ ] Collect scene loaded versions  has right project locally and on farm
- [ ] Integrate is integrating to right project locally and on farm
- [ ] Validate editorial asset is validating correctly


kalisp (changes requested):
In AE it fails locally with 
2022-12-19 13_52_25-Testing Platform PC - AnyDesk

(It probably fails also in PS, Maya and Nuke - in automatic tests)

openpype/plugins/publish/integrate.py
```diff
@@ -244,7 +243,7 @@ def filter_representations(self, instance):
         return filtered_repres
 
     def register(self, instance, file_transactions, filtered_repres):
-        project_name = legacy_io.active_project()
+        project_name = instance.context["projectName"]
```
iLLiCiTiT, line 246:
```suggestion
        project_name = instance.context.data["projectName"]
```

iLLiCiTiT:
> In AE it fails locally with

Should be fixed now.

jakubjezek001 (approved):
Tested only Validate Editorial asset name and it worked

antirotor (approved):
tested in maya and it works

PR #4135, opened 2022-11-24
Maya: Looks - add all connections

antirotor:
## Bug

If file node was connected not using its "primary data" plug, look collector was ignoring it - for example file node driving only with `Alpha Out` something down the line was missed by collector.

### Fix

This is adding `allConnection` or `ac` flag to `cmds.listHistory` to get even those.

### Testing

Create shader with texture using `Alpha Out` to drive something in the material. Look collector should pick it up correctly.


LiborBatek (approved):
Everything works fine...
image
Succesfully published the look with only alpha component connected to the shader and loaded via OP loader for check....

PR #4136, opened 2022-11-24
General: Maketx executable issue

iLLiCiTiT:
## Brief description
Fix validation of maketx executable in maya.

## Description
Extract look validated if executable returned from `get_oiio_tools_path` exists which did not on windows because of missing extension but that should not cause issues.

## Additional info
We should fix the source issue in `get_oiio_tools_path` but that would require more time.

## Testing notes:
1. ExtractLook in Maya should work on windows


m-u-r-p-h-y (approved):
it works as expected on windows

image

PR #4137, opened 2022-11-24
General: Find executable enhancement

iLLiCiTiT:
## Brief description
Enhance how executables are found when `find_executable` is called which cause that oiiotools return full path to executable.

## Description
Modified function `find_executable` which can now also fill extension to path to executable if is missing. With this change oiiotools return full path to executable with extension.

## Additional information
Proper fix of previous issue from PR https://github.com/pypeclub/OpenPype/pull/4136 .

## Testing notes:
Open `Tray > Admin > Console` and run this script:
```
from openpype.lib import get_oiio_tools_path

print(get_oiio_tools_path("oiiotool"))
print(get_oiio_tools_path("maketx"))
```
Output on windows should return full path to executabls with extensions. It should also still work on linux and mac (if the executables are available there).

### DIFF
--- a/openpype/lib/vendor_bin_utils.py
+++ b/openpype/lib/vendor_bin_utils.py
@@ -60,9 +60,10 @@ def find_executable(executable):
             path to file.
 
     Returns:
-        str: Full path to executable with extension (is file).
-        None: When the executable was not found.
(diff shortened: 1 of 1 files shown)


m-u-r-p-h-y (approved):
works as expected

image

PR #4142, opened 2022-11-26
Maya: GLTF extractor

moonyuet:
## Brief description
Allows user to use the extrator to export static or animated glb with Maya2glTF

## Description
Allows user to use the extrator to export static or animated glb with Maya2glTF. After enabling the extractor, you can export glb by creating either model, animation or point cache instance.

You need to enable the extractor in Openpype settings
image

## Additional info
You need to install the maya2glTF first so that the extractor can run correctly.
See the link: https://github.com/iimachines/Maya2glTF 

## Limitation
The UI of maya2glTF would pop up for seconds and close because the system needs to open the UI first to allow run the mel scriptl library from maya2glTF

## Testing notes:
1. install the maya2glTF
2. enable the extractor in Openpype setting
3. create instance which is associated to the extractor(model, animation, point cache)
4. pubish it
5. you will find the glb file in the folder(s) of the related instance in the publish folder


BigRoy:
@moonyuet Nice. I actually have `maya2glTF` and the Collada exporter to GLTF logic implemented for OpenPype so I can probably merge in some logic. I'm swamped with work these weeks but hope to loop back soon with more details on our implementation.

I didn't need the UI to pop up for it to work, so at the very least that'd be a nice start. I also had support for exporting multiple clips in a single export.

PR #4143, opened 2022-11-26
Maya: Maintain time connections on Alembic update.

tokejepsen:
## Brief description
Maintain time connections on Alembic update.

## Description
This maintains the connections to the time attribute on the alembic node in Maya so you can for example animate the time.

## Additional info
technical information.

## Testing notes:
1. Load Alembic.
2. Disconnect time and key it.
3. Update Alembic.

### DIFF
--- a/openpype/hosts/maya/api/plugin.py
+++ b/openpype/hosts/maya/api/plugin.py
@@ -217,7 +217,7 @@ class ReferenceLoader(Loader):
 
         # Need to save alembic settings and reapply, cause referencing resets
         # them to incoming data.
-        alembic_attrs = ["speed", "offset", "cycleType"]
+        alembic_attrs = ["speed", "offset", "cycleType", "time"]
         alembic_data = {}
         if representation["name"] == "abc":
             alembic_nodes = cmds.ls(
@@ -226,7 +226,17 @@ class ReferenceLoader(Loader):
             if alembic_nodes:
                 for attr in alembic_attrs:
                     node_attr = "{}.{}".format(alembic_nodes[0], attr)
-                    alembic_data[attr] = cmds.getAttr(node_attr)
+                    connections = cmds.listConnections(node_attr, plugs=True)
+                    data = {
+                        "connected": False,
+                        "attribute": None,
+                        "value": cmds.getAttr(node_attr)
+                    }
+                    if connections:
+                        data["connected"] = True
+                        data["attribute"] = connections[0]
+
+                    alembic_data[attr] = data
             else:
                 self.log.debug("No alembic nodes found in {}".format(members))
 
@@ -263,8 +273,14 @@ class ReferenceLoader(Loader):
                 "{}:*".format(namespace), type="AlembicNode"
             )
             if alembic_nodes:
-                for attr, value in alembic_data.items():
-                    cmds.setAttr("{}.{}".format(alembic_nodes[0], attr), value)
+                for attr, data in alembic_data.items():
+                    node_attr = "{}.{}".format(alembic_nodes[0], attr)
+                    if data["connected"]:
+                        cmds.connectAttr(
+                            data["attribute"], node_attr, force=True
+                        )
+                    else:
+                        cmds.setAttr(node_attr, data["value"])
 
         # Fix PLN-40 for older containers created with Avalon that had the
         # `.verticesOnlySet` set to True.


LiborBatek (dismissed):
Been tested in maya ```2022``` and ```2023``` and works ok! Also tested with asset hero version switching. No problems found.

openpype/hosts/maya/api/plugin.py
```diff
@@ -263,8 +273,14 @@ def update(self, container, representation):
...
-                for attr, value in alembic_data.items():
-                    cmds.setAttr("{}.{}".format(alembic_nodes[0], attr), value)
+                for attr, data in alembic_data.items():
+                    node_attr = "{}.{}".format(alembic_nodes[0], attr)
+                    if data["connected"]:
+                        cmds.connectAttr(
+                            data["attribute"], node_attr, force=True
+                        )
+                    else:
+                        cmds.setAttr(node_attr, data["value"])
```
BigRoy, line 283: Doesn't updating the alembic automatically reconnect the `time` attribute and thus for us to set the value we'll first need to force disconnect it?

To test:
- Load alembic
- Disconnect time
- Set manually
- Update alembic with this PR

Does that work as expected?

openpype/hosts/maya/api/plugin.py
```diff
@@ -226,7 +226,17 @@ def update(self, container, representation):
...
-                    alembic_data[attr] = cmds.getAttr(node_attr)
+                    connections = cmds.listConnections(node_attr, plugs=True)
+                    data = {
+                        "connected": False,
+                        "attribute": None,
+                        "value": cmds.getAttr(node_attr)
+                    }
+                    if connections:
+                        data["connected"] = True
+                        data["attribute"] = connections[0]
```
BigRoy, line 237: I feel like this could be rewritten to:

```data = {
    "input": next(iter((cmds.listConnections(node_attr, plugs=True, destinations=False) or []), None))
    "value": cmds.getAttr(node_attr)
}
```

So basically no need to track `connected` + `attribute` separately since if input is None you know it wasn't connected. You can just store `input` (== `attribute`) as single value.

openpype/hosts/maya/api/plugin.py
```diff
@@ -226,7 +226,17 @@ def update(self, container, representation):
             if alembic_nodes:
                 for attr in alembic_attrs:
                     node_attr = "{}.{}".format(alembic_nodes[0], attr)
-                    alembic_data[attr] = cmds.getAttr(node_attr)
+                    connections = cmds.listConnections(node_attr, plugs=True)
```
BigRoy, line 229: This should have argument `destinations=False` since otherwise it could also list outputs of the attribute.

openpype/hosts/maya/api/plugin.py
BigRoy: Also, this would mean if you'd update without any changes to `.time` attribute it'd probably log a warning or error on update that `time` was already connected?

tokejepsen: You are right. Will have a look at this.

LiborBatek (approved):
Works as expected ...even when disconnecting time atribute and put just frame# manually (without setting anim key).  Keeps the value intact when changing asset version. No any errors given.

tokejepsen:
@BigRoy you happy with the PR?

openpype/hosts/maya/api/plugin.py
```diff
@@ -263,8 +271,19 @@ def update(self, container, representation):
...
             )
             if alembic_nodes:
-                for attr, value in alembic_data.items():
-                    cmds.setAttr("{}.{}".format(alembic_nodes[0], attr), value)
+                for attr, data in alembic_data.items():
+                    node_attr = "{}.{}".format(alembic_nodes[0], attr)
+                    if data["input"]:
+                        cmds.connectAttr(
+                            data["input"], node_attr, force=True
+                        )
```
BigRoy, line 279: Shouldn't this make sure that the connection to be made doesn't already exist? (is a warning issues when e.g. `time` was previously connected and you'd update (the update reconnects time by default) and thus this tries to connect time again even though already connected. Does that error or log a warning?

openpype/hosts/maya/api/plugin.py
```diff
@@ -226,7 +226,15 @@ def update(self, container, representation):
             if alembic_nodes:
                 for attr in alembic_attrs:
                     node_attr = "{}.{}".format(alembic_nodes[0], attr)
-                    alembic_data[attr] = cmds.getAttr(node_attr)
+                    inputs = cmds.listConnections(
+                        node_attr, plugs=True, destination=False
+                    )
```
BigRoy, line 231: I feel the code would be more readable if we'd move this to a simple separate function:
```python
def get_input(attr):
```

The code here could then just be:
```python
                    data = {
                        "input": get_input(node_attr),
                        "value": cmds.getAttr(node_attr)
                    }
```

And further down would just be:
```python
            if alembic_nodes:
                alembic_node = alembic_nodes[0]  # assume single AlembicNode
                for attr, data in alembic_data.items():
                    node_attr = "{}.{}".format(alembic_node, attr)
                    if data["input"]:
                        if data["input"] != get_input(node_attr):
                            cmds.connectAttr(
                                data["input"], node_attr, force=True
                            )
                    else:
                        input = get_input(node_attr)
                        if input:
                            cmds.disconnectAttr(input, node_attr)
                        cmds.setAttr(node_attr, data["value"])
```

Anyway, just feels more readable to only me maybe.

openpype/hosts/maya/api/plugin.py
LiborBatek: note from testing side...it didnt reconnect ```time``` by default after updating the asset so no chance for any warning or error. If I am correct.

tokejepsen:
@BigRoy what do you think now?

BigRoy (approved):
Haven't run the code, but code looks good to me.

PR #4144, opened 2022-11-28
Maya: Optional viewport refresh on pointcache extraction

tokejepsen:
## Brief description
Optional viewport refresh on pointcache extraction.

## Description
Sometimes you need the viewport to refresh for example when dealing with the bullet plugin.

## Additional info
technical information.

## Testing notes:
1. Create pointcache.
2. Enabled refresh.
3. See the viewport update as extraction is done.

### DIFF
--- a/openpype/hosts/maya/api/lib.py
+++ b/openpype/hosts/maya/api/lib.py
@@ -127,14 +127,14 @@ def get_main_window():
 
 
 @contextlib.contextmanager
-def suspended_refresh():
+def suspended_refresh(suspend=True):
     """Suspend viewport refreshes"""
-
+    original_state = cmds.refresh(query=True, suspend=True)
     try:
-        cmds.refresh(suspend=True)
+        cmds.refresh(suspend=suspend)
         yield
     finally:
-        cmds.refresh(suspend=False)
+        cmds.refresh(suspend=original_state)
 
 
 @contextlib.contextmanager
--- a/openpype/hosts/maya/plugins/create/create_pointcache.py
+++ b/openpype/hosts/maya/plugins/create/create_pointcache.py
@@ -28,6 +28,7 @@ class CreatePointCache(plugin.Creator):
         self.data["visibleOnly"] = False     # only nodes that are visible
         self.data["includeParentHierarchy"] = False  # Include parent groups
         self.data["worldSpace"] = True       # Default to exporting world-space
+        self.data["refresh"] = False       # Default to suspend refresh.
 
         # Add options for custom attributes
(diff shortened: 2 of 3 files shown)


LiborBatek (approved):
Everything seems ok. Pointcache being published and viewport updates as it goes when
```Refresh``` option on ```pointcache``` family turned on
image

**Comment by BigRoy** [comment]
What if we'd refactor `suspended_refresh` so that it takes an argument, and when False it also ensures/forces the viewport isn't suspended. That way it does still what you'd want even if the user had his viewport manually suspended - safer, and cleaner code?

@tokejepsen thoughts?

**Comment by tokejepsen** [comment]
> What if we'd refactor `suspended_refresh` so that it takes an argument, and when False it also ensures/forces the viewport isn't suspended. That way it does still what you'd want even if the user had his viewport manually suspended - safer, and cleaner code?
> 
> ```python
> @contextlib.contextmanager
> def suspended_refresh(suspend=True):
>     """Suspend viewport refreshes"""
>     original_state= cmds.refresh(query=True, suspend=True)
>     try:
>         cmds.refresh(suspend=suspend)
>         yield
>     finally:
>         cmds.refresh(suspend=original_state)
> ```
> 
> Otherwise a user that prior to publishing had his viewport disabled would've not gotten the same result.
> 
> @tokejepsen thoughts?

Good point. Will have a look.

**Review by antirotor — APPROVED** [review]

**Comment by tokejepsen** [comment]
@BigRoy looks we merged too fast here, cause `cmds.refresh`  is not queryable.

**Comment by BigRoy** [comment]
>  cause cmds.refresh is not queryable.

Looks like you're correct. 

### Option 1

As such that'd make `cmds.refresh(suspend=True)` context manager only able to do this if we'd really need it that way:
```python
@contextlib.contextmanager
def suspended_refresh(suspend=True):
    """Suspend viewport refreshes"""
    try:
        cmds.refresh(suspend=suspend)
        yield
    finally:
        cmds.refresh(suspend=False)
```

### Option 2

Other option for disabling the viewport as an optimization could be `cmds.ogs(pause=True`)` which is queryable and essentially boils down to almost the same thing? (Just pauses viewports). 

### Option 3

There's also the very old school:
```python
from maya import cmds
from maya import mel

main_pane = mel.eval("$gMainPaine=$gMainPane;")
cmds.paneLayout(main_pane, edit=True, manage=False)
```
To make Maya's main window unmanaged and disable and refreshes of it that way - however, torn off panels I believe would remain managed.

--

I'd recommend either Option 1 or 2.

PR #4145, opened 2022-11-28
Maya: Optional control of display lights on playblast.

tokejepsen:
## Brief description
Optional control of display lights on playblast.

## Description
Giving control to what display lights are on the playblasts.

## Additional info
technical information.

## Testing notes:
1. Create a review in Maya.
2. Modify the display lights attribute on the review object set.
3. Publish.


LiborBatek:
Did test it in ```maya2023``` but gives me this error msg during extraction of review

image

m-u-r-p-h-y:
> Did test it in `maya2023` but gives me this error msg during extraction of review
> 
did you create the fresh instance or used the old one from a previous scene? It may be missing some new attributes . .

tokejepsen:
> did you create the fresh instance or used the old one from a previous scene? It may be missing some new attributes . .

We should account for old instances, so will have a look at this.

openpype/hosts/maya/plugins/create/create_review.py
```diff
@@ -25,6 +25,7 @@ class CreateReview(plugin.Creator):
         "depth peeling",
         "alpha cut"
     ]
+    displayLights = ["default", "all", "selected", "active", "none"]
```
tokejepsen, line 29: This data should be shared between collector and extractor from a central location. Where would be the best place for it?

LiborBatek (changes requested):
Issues have been found.

openpype/hosts/maya/plugins/create/create_review.py
tokejepsen: @antirotor what do you think?

antirotor: hmm, maybe collector that can be used to collect stuff now processed by `extract_playblast` - like `collect_viewport`? It could put it under instace dictionary along with other related stuff.

tokejepsen: Ideally it should be available to the creator plugin as well, so maybe from a library?
Not sure whether OpenPype stores data in libraries?

tokejepsen:
@antirotor what do you think of this?

Have collected the display light list in the lib and referenced in the plugins. I've noticed a bit of data collected in the lib so mirrored that. Also noticed some data collected in lib that is not used.

openpype/hosts/maya/plugins/publish/extract_playblast.py
```diff
@@ -98,6 +98,10 @@ def process(self, instance):
         pm.currentTime(refreshFrameInt - 1, edit=True)
         pm.currentTime(refreshFrameInt, edit=True)
 
+        # Show lighting mode.
+        index = instance.data.get("displayLights", 0)
+        preset["viewport_options"]["displayLights"] = lib.DISPLAY_LIGHTS[index]
```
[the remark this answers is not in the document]
tokejepsen: If we have `display_lights = ["default", "all", "selected", "active", "none"]` in a collector to convert the enum to string, then we still need this list/data in the creator plugin;

https://github.com/tokejepsen/pype/blob/enhancement/maya_review/openpype/hosts/maya/plugins/create/create_review.py#L44

We could continue with having the lib storing the list/data, but do the enum to string conversion in a collector?

tokejepsen:
@BigRoy think I've addressed the issues now.

openpype/hosts/maya/plugins/publish/collect_review.py
```diff
@@ -139,3 +140,7 @@ def process(self, instance):
                         "filename": node.filename.get()
                     }
                 )
+
+        # Convert enum attribute to string.
```
BigRoy, line 144:
```suggestion
        # Convert enum attribute index to string.
```

openpype/hosts/maya/api/lib.py
```diff
@@ -113,6 +113,8 @@
 
 RENDERLIKE_INSTANCE_FAMILIES = ["rendering", "vrayscene"]
 
+DISPLAY_LIGHTS = ["default", "all", "selected", "active", "none"]
+
```
BigRoy, line 117: The only question I have is this. That **after** this PR the instance will always override the review Display Lights to `default` by default no matter what. As such, any setting in the Project Settings for Maya for Review is completely ignored.

Does it make more sense to EITHER:
1. Have the creator on create set the default value as defined in the Settings
OR
2. Add a first entry in the enum like "use from project settings" and then convert the index on collect accordingly.

My personal preference might be 1) so that if project settings change that the behavior of an existing scene doesn't suddenly change.

BigRoy (commented):
Two minor comments.

openpype/hosts/maya/api/lib.py
tokejepsen: Nothing about display lights are currently exposed in the OP settings, so do you mean that we should enforce the display lights to revert to `default` on create?

BigRoy: > Nothing about display lights are currently exposed in the OP settings,

Are you sure? I'm seeing this `project_settings/maya/publish/ExtractPlayblast/capture_preset/Viewport Options/displayLights` 

afbeelding

Could be that I've merged a PR that on `develop` isn't merged yet.

tokejepsen: Ahh, didnt see that. Seems a bit confusing to use the settings from the extractor plugin and apply to the creator. Should really be under `project_settings/maya/create/CreateReview`, or am I getting something wrong here?

BigRoy: I guess, yes - not sure about the thought process there. Moving it would influence existing productions so I'm not sure how to go about that. It'd be a separate issue to discuss where that should be I suppose.

tokejepsen: @antirotor got any direction suggestion here?

antirotor: I think it should be in extractor settings and not in review creator just because we might want to decouple playblast and review in the future (for example using some different method than maya playblast to generate review files, like different renderers that might not have the same options.

For now, would it make sense to have option on CreateReview like "Use Settings from extract playblast"?

tokejepsen:
@BigRoy what do you think of this now?

BigRoy:
I think functionally it looks ok. I personally would still prefer to 'simplify' if we can. I feel like we're putting much responsibility in the Extractors to get the right settings whereas I'd have expected the Collectors to do that.

For example, I'd expect the collectors to decide what to do with the display lights flag - and when set to 'project settings' to then maybe take the value from the project settings and store it in the instance directly. So basically do that conversion directly.

tokejepsen:
> I think functionally it looks ok. I personally would still prefer to 'simplify' if we can. I feel like we're putting much responsibility in the Extractors to get the right settings whereas I'd have expected the Collectors to do that.
> 
> For example, I'd expect the collectors to decide what to do with the display lights flag - and when set to 'project settings' to then maybe take the value from the project settings and store it in the instance directly. So basically do that conversion directly.

That sounds more like a bigger refactor that needs a concensus rather than doing it just for this PR.

openpype/hosts/maya/plugins/publish/collect_review.py
```diff
@@ -139,3 +140,7 @@ def process(self, instance):
                         "filename": node.filename.get()
                     }
                 )
+
+        # Convert enum attribute index to string.
+        index = instance.data.get("displayLights", 0)
+        instance.data["displayLights"] = lib.DISPLAY_LIGHTS[index]
```
BigRoy, line 146:
```suggestion
        # Convert enum attribute index to string.
        index = instance.data.get("displayLights", 0)
        display_lights = lib.DISPLAY_LIGHTS[index]
        if display_lights == "project_settings":
            # project_settings/maya/publish/ExtractPlayblast/capture_preset/Viewport Options/displayLights
            display_lights = instance.context.data["project_settings"]["maya"]["publish"]["ExtractPlayblast"]["capture_preset"]["Viewport Options"]["displayLights"]  # noqa
        instance.data["displayLights"] = display_lights
```

Wouldn't this work?

openpype/hosts/maya/plugins/publish/extract_playblast.py
```diff
@@ -98,6 +98,11 @@ def process(self, instance):
         pm.currentTime(refreshFrameInt - 1, edit=True)
         pm.currentTime(refreshFrameInt, edit=True)
 
+        # Show lighting mode.
+        display_lights = instance.data["displayLights"]
+        if display_lights != "project_settings":
+            preset["viewport_options"]["displayLights"] = display_lights
```
BigRoy, line 104:
```suggestion
        # Use displayLights setting from instance
        preset["viewport_options"]["displayLights"] = instance.data["displayLights"]
```

openpype/hosts/maya/plugins/publish/extract_thumbnail.py
```diff
@@ -105,6 +105,11 @@ def process(self, instance):
         pm.currentTime(refreshFrameInt - 1, edit=True)
         pm.currentTime(refreshFrameInt, edit=True)
 
+        # Show lighting mode.
+        display_lights = instance.data["displayLights"]
+        if display_lights != "project_settings":
+            preset["viewport_options"]["displayLights"] = display_lights
+
```
BigRoy, line 112:
```suggestion
        # Use displayLights setting from instance
        preset["viewport_options"]["displayLights"] = instance.data["displayLights"]
```

```

BigRoy (commented):
Wouldn't this work to move it to the collector?
_Haven't tested this code - consider it pseudocode_

antirotor:
> That sounds more like a bigger refactor that needs a concensus rather than doing it just for this PR.

I agree, that is something I would do once we move Maya to the new publisher. But the changes @BigRoy suggested make sense.

moonyuet:
I tested it out and it errors out when I choose no lights option. 
image

This can be probably fixed by changing the related attributes in  schema_maya_capture.json, it should be `none` instead of `nolights`
https://github.com/tokejepsen/pype/blob/ae496b9712bafc77a0d8350b92b0e84505eee512/openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_capture.json#L162-L171

jakubjezek001:
Could we please resolve the conflicts and merged latest develop. Thank you.

tokejepsen:
@LiborBatek @moonyuet could you test this again?

Successfully published here in 2023. For 2020 we'll need #4744, so maybe just ignore 2020.

moonyuet:
> @LiborBatek @moonyuet could you test this again?
> 
> Successfully published here in 2023. For 2020 we'll need #4744, so maybe just ignore 2020.

I have tested in my own machine with maya 2023, all the display options all works and publish without error. 
I especially tested with the none options twice (of which I requested for a change in the setting),  it works as expected now 
image

moonyuet (approved):
code looks good for me and the function works well after the change.

tokejepsen:

BigRoy (approved):
Code looks good to me except for the fact that it'll break for existing instances. The displayLights attribute will not exist for existing instances and thus `instance.data["displayLights"]` will fail.

I recall there being a Maya validator which checks whether instances have all their attributes which will invalidate it and can repair it so I'm approving anyway. But if that's not the case please correct the logic.

tokejepsen:
> I recall there being a Maya validator which checks whether instances have all their attributes which will invalidate it and can repair it so I'm approving anyway. But if that's not the case please correct the logic.

Yup, got the validator in place for this, so older instances will be repaired.

LiborBatek:
Still doesnt work for me... having old ```review``` instance also ```Repair``` func doesnt show up so practically no way to publish it.

image

tokejepsen:
> Still doesnt work for me... having old review instance also Repair func doesnt show up so practically no way to publish it.

Have you got latest commit from this PR?
You should get a validation failure for `Instance Attributes` which you can repair and get this `Display Lights` attribute on the instance objectset.

LiborBatek:
Did pull it like half an hour before....so I guess I have the latest

LiborBatek:
When creating new ```Review``` Instance with all params it get correctly validated but fails with identical error msg:

image

image

LiborBatek (changes requested):
Still some issues have been found.

tokejepsen:
@LiborBatek do you get any warning on `Collect Review Data`?

LiborBatek (changes requested):
Now I am able to create ```Review``` if obsolete Instance present ```Repair``` action correctly fixes it and user is able to ```Publish``` Review.

image

Problem being that it ignores the ```Display Lights``` settings and it uses ```All lights```. Also checked ```Extract Review``` and ```Override``` settings in OP...and it was turned ```Off``` so its not overriding my Review Settings anyhow.

image

LiborBatek:
Currently working ```Review``` and ```Display Lights```

- [ ] Project Settings
- [ ] Default
- [ ] All
- [ ] Selected
- [ ] Active
- [ ] None

tokejepsen:
@LiborBatek give this another go please.
I went through a more thorough testing now and found a couple of issues.

LiborBatek:
Sorry but still no luck :( 

Even on the first option ```Use project settings``` still takes the viewoprt active shading mode (whatever I have at the time on)

tokejepsen:
> Is there any specific OP Settings I need to set on or off ?? I went through all settings for Maya checking if anything not interferring but seems ok!

Hmm, do you have `project_settings/maya/publish/ExtractPlayblast/capture_preset/Viewport Options/override_viewport_options` enabled?

LiborBatek:
> > Is there any specific OP Settings I need to set on or off ?? I went through all settings for Maya checking if anything not interferring but seems ok!
> 
> Hmm, do you have `project_settings/maya/publish/ExtractPlayblast/capture_preset/Viewport Options/override_viewport_options` enabled?

This could be it! I have turn it ```Off``` ...I thought that when ```On``` no matter what I set in maya will override it with the settings below override!

So it should be ```On``` ?

tokejepsen:
> So it should be On ?

Yeah, it'll need to be enabled for the publishing to override the viewport changes else it'll just use what it setup in the workfile.

LiborBatek (approved):
All modes working fine and as expected.

PR #4150, opened 2022-11-30
Bugfix: Standalone Publish - ValidateTextureBatchWorkfiles settings

2-REC:
## Brief description
The ```ValidateTextureBatchWorkfiles``` plugin of the "Standalone Publish" tool expects an undefined setting for the main workfile extensions.

## Description
The class variable ```main_workfile_extensions``` is initialised to a specific value (```mra```), which cannot be overridden as no setting is defined for the plugin.
Additionally, the setting is already defined for the plugin ```CollectTextures``` and should thus be reused/shared ("Main workfile extensions"):
```
project_settings/standalonepublisher/publish/CollectTextures/main_workfile_extensions
```
Not reusing the same setting creates redundancy and potential issues as different values can then be set.
The ```ValidateTextureBatchWorkfiles``` plugin should thus get the setting value set from the ```CollectTextures``` plugin.


2-REC:
NOTE: I am not sure getting the setting from another plugin this way is the best way, as the name of the plugin and setting are hardcoded.

kalisp:
I think its fine, we are using this approach to get plugin settings from different plugin in multiple places.

Just please remove TODO comments and class variable, I don't think having it as a fallback is necessary and I will approve it and merge.

2-REC:
> Just please remove TODO comments and class variable, I don't think having it as a fallback is necessary and I will approve it and merge.

Should we still catch the exception if the setting is missing, or just remove the ```try/except``` block?

kalisp:
I would probably catch the exception and manipulate it into something human readable for artists.

kalisp (approved):

PR #4158, opened 2022-12-01
Global: adding imageio to settings

jakubjezek001:
## Brief description
OCIO config with file rules at global settings

## Description
Default path is pointing to our distributed aces_1.2 config. File rules are disabled by default as those are usually used on hosts, but times to times we could need to use them for some OIIO conversions. 

## Techical notes
This is a first step towards color managed OpenPype. There is at the moment none functionality yet connected to these settings. 

## Testing notes:
1. open settings and look to Global ImageIO

### DIFF
--- a/openpype/settings/defaults/project_settings/global.json
+++ b/openpype/settings/defaults/project_settings/global.json
@@ -1,4 +1,24 @@
 {
+    "imageio": {
+        "ocio_config": {
+            "enabled": true,
+            "filepath": {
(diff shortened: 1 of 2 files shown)


antirotor (approved):
Just to make sure it is clear - putting these configs to hosts that doesn't support OCIO is because the config defines rules to transform DCC native colorspace (via arbitrary mapping) to that OCIO config, right @jakubjezek001 ?

jakubjezek001:
> Just to make sure it is clear - putting these configs to hosts that doesn't support OCIO is because the config defines rules to transform DCC native colorspace (via arbitrary mapping) to that OCIO config, right @jakubjezek001 ?

Exactly ;)

PR #4165, opened 2022-12-02
Nuke: correct detection of viewer and display

jakubjezek001:
## Brief description
Regex pattern is now correctly detecting viewer

## Description
In case of Aces name convention if Rec.709 was used - regex pattern was falsely detecting Viewer as Rec only. It was causing wrong colorspace settings on OCIODisplay node.

## Testing notes:
1. Go to Nuke ImageIO settings and set `Rec.709 (ACES)` at your ACES project.
2. Publish and stop process during `Exctract Review Data Mov`. Look to OCIODisplay node and see there is Rec.709 set to Viewer.

### DIFF
--- a/openpype/hosts/nuke/api/lib.py
+++ b/openpype/hosts/nuke/api/lib.py
@@ -2961,7 +2961,7 @@ def get_viewer_config_from_string(input_string):
         viewer = split[1]
         display = split[0]
     elif "(" in viewer:
-        pattern = r"([\w\d\s]+).*[(](.*)[)]"
+        pattern = r"([\w\d\s\.\-]+).*[(](.*)[)]"
         result = re.findall(pattern, viewer)
         try:
             result = result.pop()


jrsndl:
I am not sure what to check. I use Rec709 for baking, and it works even without this PR.
No PR applied:

image
and the resulting OCIODisplay node:
image

jakubjezek001:
> I am not sure what to check. I use Rec709 for baking, and it works even without this PR. No PR applied:
> 
> image and the resulting OCIODisplay node: image

That is good. But you have different Baking input. `Rec.709 (ACES)` was the correct one. Before it was returning `Rec` and it randomly picked for example `Rec.2020`. Regex only works if `(ACES)` is in name.

And reason for this is that you could have in some productions `sRGB (DNeg)`.

jrsndl (approved):
I can confirm that using () in the name now doesn't break the baking.
Nuke 13.2v4

PR #4168, opened 2022-12-02
3dsmax: integration

antirotor:
## Feature

This is adding very basic support for Autodesk 3dsmax. It is mainly based around `pymxs` api so  it might work only on newer 3dsmax versions.

**Implemented:**
* new publisher - publishing of point caches (alembics) and workfiles
* loader - very simple alembic loader - limited due to the available api for working with abc
* workfile tool

## Testing

- Download and install 3dsmax (tested on 2023).
- Configure it in the Settings - be sure that `ADSK_3DSMAX_STARTUPSCRIPTS_ADDON_DIR` is set correctly
- Launch 3dsmax. You should see OpenPype menu next to the Help menu.


m-u-r-p-h-y (approved):
OP menu loads and basic functionality like workfile saving and opening works without a glitch.

image

PR #4169, opened 2022-12-02
Flame: vertical alignment of layers

jakubjezek001:
## Brief description
Creator is detecting properly vertical alignment

## Description
Shots above each other with the same name are now properly detected and only hero track defined in Creator is used for metadata collection during publishing.


iLLiCiTiT (approved):
One comment otherwise approved

PR #4174, opened 2022-12-05
Maya: Extractor for Unreal SkeletalMesh

simonebarbieri:
## Brief description
Implementation of the extractor of Unreal SkeletalMesh.

## Description
This PR implements an extractor of rigs as Unreal SkeletalMesh. As Unreal does not support meshes with faces with more than 4 vertices, it has been included a validator to verify that the mesh is triangulated.


LiborBatek (approved):
Works ok, validator works and repair action too. Model get triangulated and validator doesnt fail then.

LiborBatek:
not sure if somewhat connect to it but Extractor later fails (this PR is about triangulation/validation which is ok)

here is the error
image

simonebarbieri:
> I get this error... not sure if its somewhat connected to PR but Extractor later fails (this PR is about triangulation/validation which is working ok)
> 
> here is the error I get after validation which went ok image

I think it is caused by this commit: https://github.com/pypeclub/OpenPype/commit/a465315f034d9e297a27bdda9a9f37a52b21bc97

PR #4176, opened 2022-12-05
Draft: Feature: Handling out of date workfiles in Blender

Sharkitty:
Draft: This PR depends on this other PR: https://github.com/pypeclub/OpenPype/pull/4102

# Description
Added warning and validator when workfile is out of file (in Blender).
The warning shows up as a props dialog that user cannot ignore as it will be reinvoked on cancel (So if they click anywhere outside of the dialog, it reappears). This props dialog has an enum to select what action to perform, then they can click "OK" to execute.

There are three actions provided:
- Download last workfile
- Quit blender
- Proceed anyway

There is also a variable set when the workfile has been detected as out of date. The validator checks it, so if user decide to proceed anyway, they won't be able to publish, unless they explicitly untick this box.
A warning sign will appear next to the OpenPype menu as well, and there is also an Update workfile button in the OpenPype menu, which is grayed out if the workfile is up to date.
This system however presents a limitation: As the workfile is determined as up to date, or out of date based on modification time, if it is modified and saved then it will always be considered up to date. This brings a few unwanted behavior. For instance if the users proceeds anyway, then closes blender, as the file is saved when making paths absolute, it will always be considered up to date from now on.

# Testing
## General
- Have a local workfile (blender) with its datetime older than the last published workfile
- Open it in blender using OpenPype launcher
- Notice the props dialog popping, try to click somewhere else and see that it reappears whatever you do
- Select an option in the enum and click ok to test it.

## Download last workfile using the props dialog
- Select `Download last workfile` in the enum and click ok
- The last workfile should now download and be automatically opened, with the correct version number
- Confirm props dialog does not reopen
- Confirm OpenPype menu does not have a warning sign
- Confirm `Update workfile` in the OpenPype menu is grayed out
- Open pyblish and do a publish test (at least test the validators)

## Quit blender using props dialog
- Select `Quit blender` in the enum and click ok
- Blender should close

## Proceed anyway
- Select `proceed anyway` in the enum and click ok
- You should now be able to work
- OpenPype menu should have a warning sign
- Try to publish, the workfile up to date validator should fail
- Reset pyblish, untick the workfile up to date validator, and do a publish test again (at least test validators)
- Confirm `Update workfile` is available in the OpenPype menu and has the same behavior as the `Download last workfile` option in the props dialog

## Workfile is up to date case
- Confirm props dialog does not display
- Confirm OpenPype menu should does not have a warning sign
- Confirm `Update workfile` in the OpenPype menu is grayed out
- Confirm you can publish (at least test validators)


openpype/client/entities.py
```diff
@@ -1033,6 +1080,31 @@ def get_representation_by_name(
...
+    Returns:
+        str: Representation.
+    """
+    return next(
+        (
+            representation
+            for representation in get_representations(
+                project_name, version_ids=[version_doc["_id"]]
+            )
+            if representation["context"]["task"]["name"] == task_name
```
iLLiCiTiT, line 1102: There are representations without `"task"` in context.

openpype/client/entities.py
```diff
@@ -519,6 +519,53 @@ def get_subset_families(project_name, subset_ids=None):
     return set()
 
 
+def match_subset_id(
```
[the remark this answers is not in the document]
iLLiCiTiT: Not sure if this function should be here. Also it's name does not tell much about what it does? Does it find subset based on family, asset and task? If yes then it make sense only for very limited use cases. I guess it's used to find workfile subset in that case I would mark it as helper function in `openpype/pipeline/workfiles`.

openpype/client/entities.py
iLLiCiTiT: I have similar issue with this function as with `match_subset_id`. It make sense only in very specific use-cases and since it is in fact just a filtering function I don't see a benefit to have it in public api for entities.

openpype/modules/sync_server/sync_server.py
```diff
@@ -1,12 +1,23 @@
...
 import concurrent.futures
+from time import sleep
 from concurrent.futures._base import CancelledError
 
 from .providers import lib
+from openpype.client.entity_links import get_linked_representation_id
 from openpype.lib import Logger
+from openpype.lib.local_settings import get_local_site_id
+from openpype.modules.base import ModulesManager
+from openpype.pipeline import Anatomy
```
iLLiCiTiT, line 15: This is public definition of sync server which must be at this moment Python 2 compatible and should be clear without pipeline imports if possible.

BTW Even I understand the use case I don't think those functions should be called directly from blender as blender workflow is now dependent on sync server in it's core (which is not good). He should find if there is sync server addon, if it's enabled, and then call a method which does all the logic and return or raise result, but blender should not know anything about how it works (whole content of `DownloadLastWorkfile`).

openpype/hooks/pre_copy_last_published_workfile.py
```diff
@@ -31,16 +32,12 @@ def execute(self):
         2- Check if workfile in work area doesn't exist
         3- Check if published workfile exists and is copied locally in publish
         4- Substitute copied published workfile as first workfile
+           with incremented version by +1
 
         Returns:
             None: This is a void method.
         """
 
-        sync_server = self.modules_manager.get("sync_server")
```
iLLiCiTiT, line 39: So if sync server is not available or enabled it will happen anyway?

openpype/modules/sync_server/sync_server.py
Sharkitty: Okay, I see what you mean. I'll edit my implementation.

openpype/hooks/pre_copy_last_published_workfile.py
Sharkitty: This part comes from https://github.com/pypeclub/OpenPype/pull/4102 (which this PR depends on) so I think it's more appropriate to discuss it in #4102
But to answer your question the check still exists it's just been moved to `download_last_published_workfile` in `sync_server.py`

openpype/client/entities.py
Sharkitty: This comes from #4102 I'll look into it though, but as part of a fix for #4102

openpype/client/entities.py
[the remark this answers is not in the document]
Sharkitty: Same as the above

openpype/client/entities.py
iLLiCiTiT: I see. Almost none of the comments here are not related to this PR at all...

openpype/hooks/pre_copy_last_published_workfile.py
Tilix4: I think these kind of checks must remain in hooks to have explicit logging.

Tilix4 (changes requested):
Works fine, good job! (even if I'll review it after #4102 is merged and your branch will be rebased)

Doesn't work when local settings are both on `studio`. It holds forever.

openpype/hosts/blender/utility_scripts/is_workfile_out_of_date.py
```diff
@@ -0,0 +1,35 @@
+from pathlib import Path
+from datetime import datetime
+
+from openpype.lib.dateutils import get_timestamp
+from openpype.client.entities import get_last_version_by_subset_name
+from openpype.pipeline import legacy_io
+from openpype.hosts.blender.api.workio import current_file
+
+
+def is_work_file_out_of_date() -> bool:
```
Tilix4, line 10: This function doesn't belong here. It'd be more appropriate in `blender/pipeline.py`. `utility_scripts` is made for scripts to be run in blender using the `-P` CLI attribute.

openpype/hosts/blender/api/ops.py
```diff
@@ -344,6 +355,136 @@ def before_window_show(self):
...
+
+    bl_idname = "wm.workfile_out_of_date"
+    bl_label = "WARNING: Workfile is out of date"
+
+    action_enum: bpy.props.EnumProperty(
+        name="Action Enum",
+        items=(
+            ("DL", "Download last workfile", "Download last workfile"),
+            ("QUIT", "Quit blender", "Quit blender"),
+            ("PROC", "Proceed anyway", "Proceed anyway AT YOUR OWN RISK"),
```
Tilix4, line 452: Please write options in full words for better readability DOWNLOAD

openpype/hosts/blender/api/ops.py
```diff
@@ -344,6 +355,136 @@ def before_window_show(self):
...
+            )
+        )
+        return {"FINISHED"}
+
+    def invoke(self, context, event):
+        """Invoke this operator."""
+        return self.execute(context)
+
+
+class DrawWorkFileOutOfDateWarning(bpy.types.Operator):
```
Tilix4, line 439: Use Blender naming convention: `WM_OT_` as prefix.

openpype/hosts/blender/api/ops.py
```diff
@@ -344,6 +355,136 @@ def before_window_show(self):
...
+                f" of '{asset_name}'"
+            )
+            return {"CANCELLED"}
+
+        # Get workfile representation
+        last_version_doc = get_last_version_by_subset_id(
+            project_name, subset_id, fields=["_id", "name"]
+        )
+        if not last_version_doc:
+            print("Subset does not have any version")
```
Tilix4, line 397: Use Blender `self.report`

openpype/hooks/pre_copy_last_published_workfile.py
```diff
@@ -31,16 +32,12 @@ def execute(self):
         2- Check if workfile in work area doesn't exist
         3- Check if published workfile exists and is copied locally in publish
         4- Substitute copied published workfile as first workfile
+           with incremented version by +1
```
Tilix4, line 35:
```suggestion
           with incremented version by +1.
```

Tilix4 (changes requested):

openpype/hosts/blender/api/ops.py
```diff
@@ -384,12 +470,21 @@ def draw(self, context):
...
 
 
 def draw_avalon_menu(self, context):
     """Draw the Avalon menu in the top bar."""
 
-    self.layout.menu(TOPBAR_MT_avalon.bl_idname)
+    self.layout.menu(
+        TOPBAR_MT_avalon.bl_idname,
+        icon="ERROR"
+        if bpy.context.window_manager.is_workfile_out_of_date
```
Tilix4, line 484: `context` is passed as argument.
```suggestion
        if context.window_manager.is_workfile_out_of_date
```

openpype/hosts/blender/api/ops.py
```diff
@@ -344,6 +345,91 @@ def before_window_show(self):
...
+
+class WM_OT_DrawWorkFileOutOfDateWarning(bpy.types.Operator):
+    """Display a warning message about the workfile being out of date.
+    Either the user chooses to willingly proceed with their out of date
+    workfile, or blender will be closed after clicking on `OK`."""
+
+    bl_idname = "wm.workfile_out_of_date"
+    bl_label = "WARNING: Workfile is out of date"
+
+    action_enum: bpy.props.EnumProperty(
```
Tilix4, line 390: I'd rename it simply "action"

openpype/hosts/blender/api/pipeline.py
```diff
@@ -216,6 +230,34 @@ def _discover_gui() -> Optional[Callable]:
     return None
 
 
+def is_work_file_out_of_date() -> bool:
```
Tilix4, line 233: I think it's better to articulate concepts in a positive way, keep "workfile" a single word and explicitly say this is a function which "checks" things, not simply a boolean: 
```suggestion
def check_workfile_up_to_date() -> bool:
```

Therefore you'll need to reverse your logic.

openpype/hosts/blender/api/pipeline.py
```diff
@@ -216,6 +230,34 @@ def _discover_gui() -> Optional[Callable]:
...
+
+    # Getting date and time of the latest published workfile
+    last_published_time = get_last_version_by_subset_name(
+        legacy_io.active_project(),
+        f"workfile{session.get('AVALON_TASK')}",
+        asset_name=session.get("AVALON_ASSET"),
+    )["data"]["time"]
+
+    # Getting date and time of the latest locally installed workfile
+    # Time is converted to use the same format as for `last_publishd_time`
```
Tilix4, line 253:
```suggestion
    # Time is converted to use the same format as for `last_published_time`
```

openpype/hosts/blender/api/pipeline.py
```diff
@@ -216,6 +230,34 @@ def _discover_gui() -> Optional[Callable]:
...
+    workfile_time = get_timestamp(
```
Tilix4, line 254: You better keep the value of the last publish time in a `bpy.props.StringProperty` or maybe in a `FloatProperty` to have fewer conversions.

Tilix4 (changes requested):

LiborBatek:
@Tilix4 is it already ok to test it / review it?

Tilix4:
> @Tilix4 is it already ok to test it / review it?

@LiborBatek we will remove the `Draft` tag when it will!

LiborBatek (commented):
Did all testing with following findings...

1) Pop up win shows up when out of date workfile being present
B3d_OutdatedWorkfile_01

2) when selecting Proceed option , OP menu shows out of date workfile warning and Publish validator will fail correctly...
B3d_OutdatedWorkfile_02
B3d_OutdatedWorkfile_03

3) when selecting "Download workfile" blender freezes instantly (I guess its because absence of Sync Server on my machine or?!

B3d_OutdatedWorkfile_04

4) When correct up-to-date workfile exists in asset workfile folder the win pop up doesnt show up as expected and also OP menu doesnt have that workfile warning and all behaves normally in Blender.

B3d_OutdatedWorkfile_05


So my only concern is about that "download workfile" feature and how should I test it so its properly synced/downloaded... other than that all seems working fine.

openpype/hosts/blender/api/ops.py
```diff
@@ -344,6 +345,91 @@ def before_window_show(self):
...
+        """Execute this operator."""
+        # Get sync server module
+        sync_server = ModulesManager().modules_by_name.get("sync_server")
+        if not sync_server or not sync_server.enabled:
+            self.report(
+                {"WARNING"}, "Sync server module is disabled or unavailable."
+            )
+            return {"CANCELLED"}
+
+        last_workfile_path = download_last_workfile()
```
Tilix4, line 369: I think `download_last_workfile` must be changed to `get_last_workfile_representation(project_name, asset_name, task_name)` where you'd put almost everything from `download_last_workfile` instead of the last call `download_last_published_workfile`. This last call should be put right after using the gotten representation.

```suggestion
        session = legacy_io.Session
        project_name = session.get("AVALON_PROJECT")
        task_name = session.get("AVALON_TASK")
        asset_name = session.get("AVALON_ASSET")
        anatomy = Anatomy(project_name)
        last_workfile_representation = get_last_workfile_representation(project_name, asset_name, task_name)
        last_workfile_path = download_last_published_workfile(...)
```

Tilix4 (changes requested):

openpype/modules/sync_server/sync_server.py
```diff
@@ -229,6 +244,220 @@ def _get_configured_sites_from_setting(module, project_name, project_setting):
...
+    anatomy_result = anatomy.format(workfile_data)
+    local_workfile_path = anatomy_result[template_key]["path"]
+
+    # Copy last published workfile to local workfile directory
+    shutil.copy(
+        last_published_workfile_path,
+        local_workfile_path,
+    )
+
+    return local_workfile_path
```
Tilix4, line 458: In fact, `download_last_published_workfile` shouldn't return the local path but the last published wf which has been downloaded. The resolving of the `local_workfile_path`  must be put into another function and the copy made my the master process or into a utility function.

Tilix4 (changes requested):

PR #4186, opened 2022-12-06
Fix: Hero version not restored when downloading representation

### LINKED ISSUE #4154: Hero version not downloaded/created when downloading a subset representation
**Running version**
[ex. 3.14.1-nightly.2]

**Describe the bug**
Actual issue: When a representation is fetched by the sync server with all its dependencies, it does not fetch the hero version, while the representation's file might reference the hero version, which leads to broken libraries.

Minimal testing: Downloading a subset's representation doesn't download or recreate the hero version.

**To Reproduce**
1. Enable hero version in settings
2. Publish a subset
3. Make the subset not locally available
4. In Loader: Find the subset
5. Download a representation
6. The hero version is not available

[cuID:[OP-4544](https://app.clickup.com/t/37buk81)]

Tilix4:
## Brief description
Hero version is not created on remote site when a representation is downloaded. This copies the last version downloaded using the loader.

Fix #4154

### DIFF
--- a/openpype/modules/sync_server/sync_server.py
+++ b/openpype/modules/sync_server/sync_server.py
@@ -1,9 +1,13 @@
 """Python 3 only implementation."""
 import os
 import asyncio
+import shutil
 import threading
 import concurrent.futures
 from concurrent.futures._base import CancelledError
+from openpype.pipeline.anatomy import Anatomy
+
+from openpype.settings.lib import get_project_settings
 
 from .providers import lib
 from openpype.lib import Logger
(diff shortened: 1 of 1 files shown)


openpype/modules/sync_server/sync_server.py
```diff
@@ -138,6 +142,32 @@ async def download(module, project_name, file, representation, provider_name,
     module.handle_alternate_site(project_name, representation, local_site,
                                  file["_id"], file_id)
 
+    # Restore hero version in any case
+    project_settings = get_project_settings(project_name)
+    hero_enabled = project_settings["global"]["publish"][
```
iLLiCiTiT, line 147: I don't get how is this setting relevant to logic here? You can change it mid-production. It should not be source of truth. Also copying files based on directory doesn't seem right. All representations should have "files" on them.

Tilix4: Have I misundertood something from the original directives? https://discord.com/channels/517362899170230292/1047892401475305524/1049645280527663144

iLLiCiTiT: Unfortunatelly I can't as I'm also not very aknowledged with sync server and @kalisp is right now working on v4 version of it so he's busy too.

I would suggest to try find out why the representations of hero versions are not synced (why it does not work). I think it's not issue with sync server but with loader tool (which is in "duck taped" state at this moment).

We don't have time to do the research at this moment but if you'll find out where the issue is at the first place then we can spent some time on a fix.

kalisp: I messed up, for some unknown reason I assumed, that we are talking that loaded hero version is not downloaded with a published workfile, thats why I mentioned `data.inputLinks`. 

But the main issue is, that when you are downloading from Loader version of an asset which is ALSO a Hero version, only version is downloaded but not Hero version.
(Eg. I want to download `v010` of `Chair`, which is also Hero version of 'Chair', I click on Download in Loader (on v010), only v010 folder gets created, not 'hero' folder - with same content.)

I think it should be responsibility of `sync_server_module.add_site` (or actually `sync_server_module._update_site`). Let me think about it a bit.

kalisp: Actually did you engage `Download` explicitly on hero version (eg. version number must be in [])?

Screenshot 2022-12-06 194626

Tilix4: I'll try this out. 

But actually, what I've written in the issue is a minimal way to reproduce a behaviour I've experienced in another situation: when a representation is fetched by the sync server with all its dependencies, it does not fetches the hero version either, when the representation's file might reference the hero version when it was published, which is my main issue currently. I'll update the issue to explain this.

Tilix4:
New implementation proposal to #4191

PR #4188, opened 2022-12-06
Feature/op 3924 implement ass extractor

tokejepsen:
## Brief description
Reviewable ass extractor improvements.

## Description
I cant push directly to the OpenPype branch, so going through my personal fork instead. There are a lot of changes because I've merged in latest develop, so to get a better overview of the file changes we need to merge develop.

This is feature parallel to client extractor, with settings ported to OpenPype workflows with objectset and project settings.

## Discussion
I don't agree with having an `exportSequence` attribute because in OpenPype workflows you can define frame range easily, so it makes more sense to just support single frame publishes (which it already does). I propose to remove this attribute to simplify code and workflow.
So to note, current workflow of using current frame to export is prone to user errors and unpredictability.

## Testing notes:
1. Create and publish Ass instance.
2. Use Arnold's kick executable to validate renders.


tokejepsen:
It might also be useful to explore with the client whether using renderlayers is a better options than objectsets. The objectsets workflow is a bit like a blackbox, unless you are very confident in using the pipeline. Using renderlayers would mean that you could test the renders in Maya, then it's "just" a matter of implementation to convert renderlayers to ass files.

PR #4190, opened 2022-12-07
Maya: Include task name in render pass subset naming

tokejepsen:
## Brief description
Include task name in render pass subset naming

## Description
We include task name to avoid cross task overwriting.

## Testing notes:
1. Create a render in Maya with render pass.
2. Render and check subset name includes task name.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/collect_render_layer_aovs.py
+++ b/openpype/hosts/maya/plugins/publish/collect_render_layer_aovs.py
@@ -1,3 +1,5 @@
+import os
+
 from maya import cmds
 
 import pyblish.api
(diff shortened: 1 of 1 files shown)


antirotor:
Maybe we should add this to project settings so it can be changed per project? The default will be the current way so it won't break any existing productions but you can change it as needed on new projects...

iLLiCiTiT:
Isn't the task name inside the subset name by default (so it's double information)?

tokejepsen:
> Isn't the task name inside the subset name by default (so it's double information)?

The passes are hardcoded to not include it; https://github.com/pypeclub/OpenPype/blob/develop/openpype/hosts/maya/plugins/publish/collect_render_layer_aovs.py#L65

iLLiCiTiT:
> The passes are hardcoded to not include it;

But it's using `"subset"` which may already have task name in it.

tokejepsen:
@LiborBatek did you end up testing whether the task name was included in the render pass subset naming before this PR?

LiborBatek:
@tokejepsen if I am not mistaken yes the second screenshot is showing assets/renders before this PR. Or should I test it some other way? Let me know...

tokejepsen:
@LiborBatek great! Thanks for testing.

iLLiCiTiT:
The plugin is running only `if not instance.data("extendFrames", False):` how that can happen?

tokejepsen:
@iLLiCiTiT wrong issue?

PR #4191, opened 2022-12-07
Fix: Hero representation of version is not synchronized.

### LINKED ISSUE #4154: Hero version not downloaded/created when downloading a subset representation

Tilix4:
## Brief description
Hero version is not created on remote site when a representation is downloaded. This adds the related hero representation in `_update_site` when a version representation is fetched.

Fix #4154

Replaces #4186

## Testing notes:
1. Enable Hero version
1. Publish a workfile with a published subset as input link
2. Using the Loader, download the workfile
3. The hero representation is fetched along the version.

### DIFF
--- a/openpype/modules/sync_server/sync_server_module.py
+++ b/openpype/modules/sync_server/sync_server_module.py
@@ -16,7 +16,10 @@ from openpype.client import (
     get_representations,
     get_representation_by_id,
 )
-from openpype.client.entities import get_hero_version_by_subset_id, get_representation_parents
+from openpype.client.entities import (
+    get_hero_version_by_subset_id,
+    get_representation_parents,
+)
 from openpype.modules import OpenPypeModule, ITrayModule
 from openpype.settings import (
     get_project_settings,
@@ -1815,7 +1818,7 @@ class SyncServerModule(OpenPypeModule, ITrayModule):
             representation_ids.extend(
                 [repre["_id"] for repre in representations]
             )
-        
+
         query = {"_id": {"$in": representation_ids}}
 
         self.connection.database[project_name].update_many(
--- a/openpype/modules/sync_server/sync_server_module.py
+++ b/openpype/modules/sync_server/sync_server_module.py
@@ -16,6 +16,7 @@ from openpype.client import (
     get_representations,
     get_representation_by_id,
 )
+from openpype.client.entities import get_hero_version_by_subset_id, get_representation_parents
 from openpype.modules import OpenPypeModule, ITrayModule
 from openpype.settings import (
     get_project_settings,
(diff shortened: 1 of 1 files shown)


openpype/modules/sync_server/sync_server_module.py
```diff
@@ -1792,11 +1796,32 @@ def _update_site(self, project_name, representation_id,
...
+        )
+        hero_version = get_hero_version_by_subset_id(
+            project_name, representation_parents[1]["_id"], fields=["_id"]
+        )
+        if hero_version:
+            representations = get_representations(
+                project_name,
+                version_ids=[hero_version["_id"]],
+                fields=["_id"],
+                context_filters={"ext": [representation["context"]["ext"]]},
```
iLLiCiTiT, line 1816: I don't think it should be filtrered by extension in context but by `"name"`.

openpype/modules/sync_server/sync_server_module.py
```diff
@@ -1792,11 +1796,32 @@ def _update_site(self, project_name, representation_id,
...
 
-        self.connection.database[project_name].update_one(
+        # Add hero version
+        representation = get_representation_by_id(
+            project_name, representation_id
+        )
+        representation_parents = get_representation_parents(
+            project_name, representation
+        )
+        hero_version = get_hero_version_by_subset_id(
```
iLLiCiTiT, line 1808: You should check if hero version matches the version doc.
```
version_doc = get_version_by_id(representation["parent"])
hero_version = None
if version_doc["type"] == "hero_version":
    hero_version = version_doc

if not hero_version:
    hero_version = get_hero_version_by_subset_id(
        project_name, version_doc["parent"], fields=["_id", "version_id"]
    )
    if not hero_version or hero_version["version_id"] != version_doc["_id"]:
        hero_version = None

if hero_version:
    repre_doc = get_representation_by_name(
        project_name,
        representation["name"],
        hero_version["_id"]
    )
```

openpype/modules/sync_server/sync_server_module.py
kalisp: I don't think so, I have representations which have name 'thumbnail'. 
(But I have also some old representations without `representation["context"]["ext"]`)

iLLiCiTiT: So it should be filtered by representation name instead of extension. Extension on context can not be unique (there can be more then one representation with extension `.mov` under single version). So this is unsave and could potentially cause sync of more representations.

kalisp (changes requested):
I  would actually prefer to yank this logic out to `openpype/plugins/load/add_site` (maybe check if it couldnt be handled in `get_linked_representation_id`).

`_update_site` will go away in v4 and we would need to redo this for v4 again. 
I don't think also it should be responsibility of lowest method to do some complex logic.
(I understand that do this here would catch all possible ways this should get triggered, but I don't think it is too readable.)

Tilix4:
@kalisp I'm out of work for three weeks. Therefore if you have some time to fix this painful issue don't wait for me! I was proposing things but I don't have enough time to make it clean quickly.

kalisp:
Should be replaced with https://github.com/ynput/OpenPype/pull/4359

kalisp:
Replaced with https://github.com/ynput/OpenPype/pull/4359

PR #4192, opened 2022-12-07
Feature/op 4397 gl tf extractor for maya

moonyuet:
## Brief description
Allows user to use the extrator to export static or animated glb with Maya2glTF

## Description
Allows user to use the extrator to export static or animated glb with Maya2glTF. After enabling the extractor, you can export glb by creating either model, animation or point cache instance.

You need to enable the extractor in Openpype settings
image

## Additional info
You need to install the maya2glTF first so that the extractor can run correctly.
See the link: https://github.com/iimachines/Maya2glTF 

## Limitation
The UI of maya2glTF would pop up for seconds and close because the system needs to open the UI first to allow run the mel scriptl library from maya2glTF

## Testing notes:
1. install the maya2glTF
2. enable the extractor in Openpype setting
3. create instance which is associated to the extractor(model, animation, point cache)
4. pubish it
5. you will find the glb file in the folder(s) of the related instance in the publish folder

### DIFF
--- a/openpype/hosts/maya/api/gltf.py
+++ b/openpype/hosts/maya/api/gltf.py
@@ -7,34 +7,34 @@ from maya import cmds, mel  # noqa
 log = logging.getLogger(__name__)
 
 _gltf_options = {
-    "of" : str,                  # outputFolder
-    "cpr" : str,                 # copyright
-    "sno" : bool,                # selectedNodeOnly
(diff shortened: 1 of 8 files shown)


antirotor (changes requested):
apart of that `if` it looks good!

antirotor (approved):

PR #2604, opened 2022-01-28
Houdini: Implement Arnold ROP render publishing to Deadline from Houdini

BigRoy:
## Brief description

Implements Arnold rendering for Houdini in OpenPype.

## Description

- [x] ✅ Currently Tested solely with Deadline submissions
- [x] ✅ Publish job now gets created.

## Additional info

It seems the rendering logic, e.g. that was there for Redshift, never really worked in OpenPype. Likely no one actively uses it? Because some errors have been creeping into the code over time and it was unable to still function as it should.

Creating it as a Draft Pull Request to initiate discussion since there are still some TODOs plus some changes are in code also used by Redshift. Would love to know if those seem problematic to anyone.

## Testing notes:

- Tested in Houdini 19.0.455 with `htoa-6.0.2.0_r0ad447a_houdini-19.0.455.py3`
- This requires some tweaks in Deadline because latest Deadline does not come with Hou 19 support out of the box + there's a bug in Houdini that will ALWAYS try path remapping on some Arnold nodes (even if path remapping is disabled on the plug-in) which actually fails to work and errors with: `FailRenderException : Error: Caught exception: 'NoneType' object has no attribute 'instances'` - so if you hit that you'll need to just disable the full `pathmap_arnold_procedurals` function in `hrender_dl.py` in the Deadline Houdini plugin. I just added a `return` at the beginning of that function.
- This will also work fine without an Arnold license however render will be watermarked.

PR #4193, opened 2022-12-07
Houdini: Implement Arnold ROP render publishing to Deadline

antirotor:
## Brief description

> **Note**
> This is continuation of #2604  by @BigRoy so it can be changed for the new publisher.

Implements Arnold rendering for Houdini in OpenPype.

## Description

- [x] ✅ Currently Tested solely with Deadline submissions
- [x] ✅ Publish job now gets created.
- [ ] Convert to new publisher

## Additional info

It seems the rendering logic, e.g. that was there for Redshift, never really worked in OpenPype. Likely no one actively uses it? Because some errors have been creeping into the code over time and it was unable to still function as it should.

Creating it as a Draft Pull Request to initiate discussion since there are still some TODOs plus some changes are in code also used by Redshift. Would love to know if those seem problematic to anyone.

## Testing notes:

- Tested in Houdini 19.0.455 with `htoa-6.0.2.0_r0ad447a_houdini-19.0.455.py3`
- This requires some tweaks in Deadline because latest Deadline does not come with Hou 19 support out of the box + there's a bug in Houdini that will ALWAYS try path remapping on some Arnold nodes (even if path remapping is disabled on the plug-in) which actually fails to work and errors with: `FailRenderException : Error: Caught exception: 'NoneType' object has no attribute 'instances'` - so if you hit that you'll need to just disable the full `pathmap_arnold_procedurals` function in `hrender_dl.py` in the Deadline Houdini plugin. I just added a `return` at the beginning of that function.
- This will also work fine without an Arnold license however render will be watermarked.


BigRoy:
I've refactored the code to work on the latest `develop` and use the new publisher. I still need to double check whether the actual deadline submission and image publishing works - but the instance creation does already work.

BigRoy:
@fabiaserra I believe this is a PR you might want to give a go based on your needs for render submissions from Houdini. I know that @maxpareschi also has done some work on rendering from Houdini front, completely separate from this PR and he's submitting command line renders to Deadline for usd so his workflow will be quite different.

It'd be good to give this PR a test run @fabiaserra :)

BigRoy:
Closing this PR in favor of @moonyuet continuing the feature in this PR https://github.com/ynput/OpenPype/pull/4825

PR #4195, opened 2022-12-08
Colorspace Management and Distribution

jakubjezek001:
## Brief description
Adding colorspace management

## Description
Adding colorspace management functionality. Settings providing now to each host and global to set ImageIO config and file rules. Publishing is adding colorspace to each representation data so during loading we can maintain colorspace across all hosts connected to OpenPype. In case there is no colorspace  saved on representation file rules could be used and define colorspace per file path.

## Additional info
The plan is to integrate each pixel data representation with color space data in representation_db document. So loaders could set correct colorspace used during publishing.  Here is the example of colorspace data. 
```Python
{
    "colorspace": "linear",  # for other publish plugins and loaders
    "configData": {
        "path": "/abs/path/to/config.ocio",  # for future references in case need
        "template": "{project[root]}/path/to/config.ocio"  # for other plugins within remote publish cases
    }
}
```

## How to integrate it to your host
1. Each host setting schema should have following shemas
```json
{
    "key": "imageio",
    "type": "dict",
    "label": "Color Management (ImageIO)",
    "is_group": true,
    "children": [
        {
            "type": "schema",
            "name": "schema_imageio_config"
        },
        {
            "type": "schema",
            "name": "schema_imageio_file_rules"
        }

    ]
}
```
2. Use any mechanism to set OCIO config to host app resolved from `openpype\pipeline\colorspace.py:get_imageio_config`
	-	either set OCIO environment during host launching via pre-launch hook
	- or to set workfile ocio config path if host api is available
	
3. Each pixle related exporter plugins has to use parent class `openpype\pipeline\publish\publish_plugins.py:ExtractorColormanaged` and use it similarly as it is already implemented here `openpype\hosts\nuke\plugins\publish\extract_render_local.py`
- **get_colorspace_settings**: is solving all settings for the host context
- **set_representation_colorspace**: is adding colorspaceData to representation. If the colorspace is known then it is added directly to the representation with resolved config path. 

### Using OCIO config
Each path input is supporting [anatomy template key formating](https://openpype.io/docs/admin_settings_project_anatomy#available-template-keys) and also environment variables. Default distributed value is set to `{OPENPYPE_ROOT}/vendor/bin/ocioconfig/OpenColorIOConfigs/aces_1.2/config.ocio`. If any of host config is enabled and set to any path, global config path is overriden.

Config path can be defined to particular shot target with following path input `{root[work]}/{project[name]}/{hierarchy}/{asset}/config/aces.ocio`

### Using File rules
File rules are inspired by OCIO v2 configuration ([more info here](https://opencolorio.readthedocs.io/en/latest/guides/authoring/rules.html)). Since each rule is having unique name it can be overidden by host activated File rules (example: `project_settings/nuke/imageio/file_rules/rules`). Pattern input is using REGEX expression syntax (try [regexr.com](https://regexr.com/)). . Matching rules procedure's intention is to be used during publishing or loading of representation. Since the publishing procedure is run before integrator formates publish template path, make sure the pattern is working or any work render path. 

> Colorspace name value is string input and no validation is run after saving of changes. We recommand to open a config.ocio file and search exact colorspace names for validation.


antirotor (approved):

BigRoy:
Posting [the comment I made elsewhere](https://github.com/ynput/OpenPype/pull/4270#issuecomment-1377275232) to here since this appears to be the main PR:

---

After this PR - upon loading a publish. How would one retrieve the color space originally used for the file?
In other words, how to apply the same color space setting on load in Maya or alike. 

Looking over the code it seems to be in `representation["data"]["colorspaceData"]` and stored as:

```json
{
    "colorspace": "REC709",    // color space of the representation file
    "config": {
        "path": "/absolute/path/to/ocio.config",  // ocio config used
        "template": "/template/which/defined/path/to/ocio.config" // original template used to defined ocio config path?
    }
}
```

Would be great to see some **Color Space** documentation!

jakubjezek001:
> Would be great to see some **Color Space** documentation!
Yeah, for sure. I am working on it and will be part of this PR ;)

website/docs/project_settings/settings_project_global.md
```diff
@@ -13,6 +13,25 @@ Project settings can have project specific values. Each new project is using stu
...
+:::
+
+### Using OCIO config
+Each path input is supporting [anatomy template key formating](../admin_settings_project_anatomy#available-template-keys) and also environment variables. Default distributed value is set to `{OPENPYPE_ROOT}/vendor/bin/ocioconfig/OpenColorIOConfigs/aces_1.2/config.ocio`. If any of host config is enabled and set to any path, global config path is overriden.
+
+Config path can be defined to particular shot target with following path input `{root[work]}/{project[name]}/{hierarchy}/{asset}/config/aces.ocio`
+
+### Using File rules
+File rules are inspired by OCIO v2 configuration ([more info here](https://opencolorio.readthedocs.io/en/latest/guides/authoring/rules.html)). Since each rule is having unique name it can be overidden by host activated File rules (example: `project_settings/nuke/imageio/file_rules/rules`). Pattern input is using REGEX expression syntax (try [regexr.com](https://regexr.com/)). Matching rules procedure's intention is to be used during publishing or loading of representation. Since the publishing procedure is run before integrator formates publish template path, make sure the pattern is working or any work render path.
```
BigRoy, line 28:
```suggestion
### Using File rules
File rules are inspired by [OCIO v2 configuration]((https://opencolorio.readthedocs.io/en/latest/guides/authoring/rules.html)). Each rule has a unique name which can be overridden by host-specific _File rules_ (example: `project_settings/nuke/imageio/file_rules/rules`). 

The _input pattern_ matching uses REGEX expression syntax (try [regexr.com](https://regexr.com/)). Matching rules procedure's intention is to be used during publishing or loading of representation. Since the publishing procedure is run before integrator formate publish template path, make sure the pattern is working or any work render path.
```

I've restructured this somewhat. But still couldn't make any sense of this line:

> Matching rules procedure's intention is to be used during publishing or loading of representation. Since the publishing procedure is run before integrator formate publish template path, make sure the pattern is working or any work render path.

What? :D

website/docs/project_settings/settings_project_global.md
```diff
@@ -13,6 +13,25 @@ Project settings can have project specific values. Each new project is using stu
...
+
+:::warning Colorspace name input
+Colorspace name value is string input and no validation is run after saving of changes. We recommand to open a config.ocio file and search exact colorspace names for validation.
+:::
```
BigRoy, line 32:
```suggestion
:::warning Colorspace name input
The **colorspace name** value is a raw string input and no validation is run after saving project settings. We recommend to open the specified `config.ocio` file and copy pasting the exact colorspace names.
:::
```

openpype/pipeline/README.md
```diff
@@ -0,0 +1,29 @@
+# Colorspace
```
BigRoy, line 1: I feel this is the wrong place for this - this should just be in the DEV docs. No?

website/docs/project_settings/settings_project_global.md
```diff
@@ -13,6 +13,25 @@ Project settings can have project specific values. Each new project is using stu
...
+## Color Management (ImageIO)
+
+:::info Default OCIO config
+OpenPype is distributing own ocio configs. Those can be found in `{openpype install dir}/{version}/vendor/bin/ocioconfig/OpenColorIOConfigs`. Windows example: `C:\Program Files (x86)\OpenPype\3.14.0\vendor\bin\ocioconfig\OpenColorIOConfigs`
+:::
+
+### Using OCIO config
+Each path input is supporting [anatomy template key formating](../admin_settings_project_anatomy#available-template-keys) and also environment variables. Default distributed value is set to `{OPENPYPE_ROOT}/vendor/bin/ocioconfig/OpenColorIOConfigs/aces_1.2/config.ocio`. If any of host config is enabled and set to any path, global config path is overriden.
+
+Config path can be defined to particular shot target with following path input `{root[work]}/{project[name]}/{hierarchy}/{asset}/config/aces.ocio`
```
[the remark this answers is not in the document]
tokejepsen: Does it need to be for each shot? Or can you define it on a sequence level?

tokejepsen:
Is the intend still to have hierarchical colorspace data?
So project settings has the top level colorspace data but each visual parent can have their own colorspace data down to subset/version/representation.

website/docs/project_settings/settings_project_global.md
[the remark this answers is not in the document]
jakubjezek001: > Should this maybe additionally mention that when enabled like this EACH shot must have a config/aces.ocio of its own? (Does that make sense in production?)

@BigRoy Yeah, now when you mention it - I believe it will not be really solid feature for some. I presume some sort of falback, or multi path inputs with order should be implemented. I guess best approach on this would be to add all templated paths targeting all hierarchical paretns. So for example. 

1. `{root[work]}/{project[name]}/{hierarchy}/{asset}/config/aces.ocio`
2. `{root[work]}/{project[name]}/{hierarchy}/config/aces.ocio`
3. `{root[work]}/{project[name]}/config/aces.ocio`

Only that 2nd path will be a bit tricky. All parents will need to be iterated for possible paths to be tested.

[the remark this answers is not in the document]
jakubjezek001: > Does it need to be for each shot? Or can you define it on a sequence level?

@tokejepsen I guess the approach mentioned above would be addressing what you thing, right?

openpype/pipeline/README.md
jakubjezek001: Yeah, I believe you are right - for some reason I have missed that section of the documentation..

jakubjezek001:
> Is the intend still to have hierarchical colorspace data? So project settings has the top level colorspace data but each visual parent can have their own colorspace data down to subset/version/representation.
 
Sure it is. I will need to add that - current implementation is not providing the feature.

website/docs/project_settings/settings_project_global.md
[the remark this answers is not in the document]
tokejepsen: > @tokejepsen I guess the approach mentioned above would be addressing what you thing, right?

Yeah, I think going up the hierarchy from the current asset/version/representation would make sense and make it flexible enough for production.

openpype/pipeline/README.md
jakubjezek001: so I have move it to Dev documentation.

website/docs/dev_colorspace.md
```diff
@@ -0,0 +1,113 @@
...
+        "colorspace": "linear",
+        "configData": {
+            "path": "/abs/path/to/config.ocio",
+            "template": "{project[root]}/path/to/config.ocio"
+        }
+    }
+
+
+## How to integrate it into a host
+1. Each host setting schema should have following shemas. Ideally at top of all categories so it mimic already defined order at other hosts.
```
BigRoy, line 37:
```suggestion
1. The settings for a host should add the `imagio` schema. Ideally near the top of all categories in its `/settings/entities/schemas/system_scheams/host_settings/schema_{host}.json` so it matches the settings layout other hosts.
```

website/docs/dev_colorspace.md
```diff
@@ -0,0 +1,113 @@
...
+            "name": "schema_imageio_file_rules"
+        }
+
+    ]
+}
+```
+
+2. Use any mechanism to set OCIO config to host app resolved from `openpype\pipeline\colorspace.py:get_imageio_config`
+	-	either set OCIO environment during host launching via pre-launch hook
+	- or to set workfile ocio config path if host api is available
```
BigRoy, line 60:
```suggestion
2. Set the OCIO config path for the host to the path returned from `openpype.pipeline.colorspace.get_imageio_config`, for example:
	- set the `OCIO` environment variable before launching the host via a prelaunch hook
	- or (if the host allows) to set the workfile OCIO config path using the host's API
```

website/docs/dev_colorspace.md
```diff
@@ -0,0 +1,113 @@
...
+
+3. Each pixle related exporter plugins has to use parent class `openpype\pipeline\publish\publish_plugins.py:ExtractorColormanaged` and use it similarly as it is already implemented here `openpype\hosts\nuke\plugins\publish\extract_render_local.py`
+- **get_colorspace_settings**: is solving all settings for the host context
+- **set_representation_colorspace**: is adding colorspaceData to representation. If the colorspace is known then it is added directly to the representation with resolved config path.
```
BigRoy, line 64:
```suggestion
3. Each Extractor exporting pixel data (e.g. image or video) has to use parent class `openpype.pipeline.publish.publish_plugins.ExtractorColormanaged` and use `self.set_representation_colorspace` on the representations to be integrated.

The **set_representation_colorspace** method adds `colorspaceData` to the representation. If the `colorspace` passed is not `None` then it is added directly to the representation with resolved config path otherwise a color space is assumed using the configured file rules. If no file rule matches the `colorspaceData` is **not** added to the representation.

An example implementation can be found here: `openpype\hosts\nuke\plugins\publish\extract_render_local.py`

```

website/docs/dev_colorspace.md
```diff
@@ -0,0 +1,113 @@
...
+
+4. Implement the loading procedure. Each loader which needs to have colorspace (detected from representation doc) set to DCC reader nodes should implement following code.
```
BigRoy, line 66:
```suggestion
4. The Loader plug-ins should take into account the `colorspaceData` in the published representation's data to allow the DCC to read in the expected color space.
```

website/docs/dev_colorspace.md
```diff
@@ -0,0 +1,113 @@
...
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+## Introduction
+Defines the distribution of colors and OCIO config during publishing. Once colorspace data are captured and integrated into representation loaders could use them for loading image and video data with correct colorspace.
+
+:::warning Color Management (ImageIO)
+Settings schema is requred for any host or module which is processing pixel data.
+:::
```
BigRoy, line 15:
```suggestion
:::warning Color Management (ImageIO)
Adding the `imagio` settings schema is required for any host or module which is processing pixel data.
:::
```

website/docs/dev_colorspace.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Data model
+The *colorspaceData* are stored at root of representation dictionary during publishing. Once they are integrated into representation db document they are stored as *representation_doc.data["colorspaceData"]*
```
BigRoy, line 18:
```suggestion
Published representations that are extracted with color managed data store a **colorspaceData** entry in its data: `representation_doc["data"]["colorspaceData"]`.

It's up to the Host implementation to pre-configure the application or workfile to have the correct OCIO config applied.
It's up to the Extractors to set these values for the representation during publishing.
It's up tot he Loaders to read these values and apply the correct expected color space.
```

BigRoy (commented):
Some notes on the Dev documentation.

openpype/settings/defaults/project_settings/fusion.json
```diff
@@ -1,5 +1,13 @@
 {
     "imageio": {
+        "ocio_config": {
```
iLLiCiTiT, line 3: Isn't this duplicate of the `"ocio"` key here?

website/docs/dev_colorspace.md
iLLiCiTiT: Typo `It's up tot he Loaders ` -> `It's up to the Loaders `

BigRoy: Fixed the typo in the suggestion. Thanks.

tokejepsen:
Looks like the settings are not respecting the enabled state of the new OCIO settings.
Capture

Capture

I would imagine that if you have `project_settings/nuke/imageio/ocio_config/enabled` enabled then we can assume this as using OCIO with a custom path.

openpype/pipeline/publish/publish_plugins.py
```diff
@@ -280,3 +286,143 @@ def staging_dir(self, instance):
...
+            return
+
+        project_name = context.data["projectName"]
+        host_name = context.data["hostName"]
+        project_settings = context.data["project_settings"]
+
+        # get one filename
+        filename = representation["files"]
+        if isinstance(filename, list):
+            filename = filename.pop()
```
tokejepsen, line 402: `filename.pop()` will remove the last item from `representation["files"]` and modify the resulting representation for integration. Using `filename[0]` will fix this.

```suggestion
            filename = filename[0]
```

tokejepsen (changes requested):

openpype/pipeline/colorspace.py
```diff
@@ -0,0 +1,464 @@
...
+        config_data = _get_config_data(
+            config_host["filepath"], anatomy_data
+        )
+    else:
+        config_data = None
+
+    if not config_data:
+        # get config path from either global or host_name
+        config_global = imageio_global["ocio_config"]
+        print(config_global["filepath"])
```
BigRoy, line 360: Printing should be removed or should be debug logging.

openpype/pipeline/colorspace.py
```diff
@@ -0,0 +1,464 @@
...
+    anatomy_data["platform"] = platform.system().lower()
+
+    # get colorspace settings
+    imageio_global, imageio_host = _get_imageio_settings(
+        project_settings, host_name)
+
+    config_host = imageio_host["ocio_config"]
+
+    if config_host["enabled"]:
+        print(config_host["filepath"])
```
BigRoy, line 350: Printing should be removed or should be debug logging.

openpype/pipeline/colorspace.py
```diff
@@ -0,0 +1,464 @@
...
+        project_settings (dict): project settings.
+                                           Defaults to None.
+        host_name (str): host name
+
+    Returns:
+        tuple[dict, dict]: image io settings for global and host
+    """
+    # get image io from global and host_name
+    imageio_global = project_settings["global"]["imageio"]
+    imageio_host = project_settings[host_name]["imageio"]
```
BigRoy, line 462: I feel like this helper method might be better of 'safe-guarding' a host maybe not having implement `imagio` settings yet and do: `project_settings.get(host_name, {}).get("imageio", {})`

```suggestion
    imageio_host = project_settings.get(host_name, {}).get("imageio", {})
```

openpype/pipeline/colorspace.py
```diff
@@ -0,0 +1,464 @@
...
+        dict: file rules data
+    """
+    project_settings = project_settings or get_project_settings(project_name)
+
+    imageio_global, imageio_host = _get_imageio_settings(
+        project_settings, host_name)
+
+    # get file rules from global and host_name
+    frules_global = imageio_global["file_rules"]
+    frules_host = imageio_host["file_rules"]
```
BigRoy, line 437: As far as I know there is no hard requirement for any host - even if rudimentary to have implemented `imageio` in settings. Right? This might need to be me safeguarded to not exist - see comment below.

openpype/pipeline/colorspace.py
```diff
@@ -0,0 +1,464 @@
...
+        config_path (str): path leading to config.ocio file
+
+    Returns:
+        dict: `display/viewer` and viewer data
+    """
+    if sys.version_info[0] == 2:
+        return get_views_data_subprocess(config_path)
+
+    from ..scripts.ocio_wrapper import get_views_data
+    return get_views_data(config_path)
```
BigRoy, line 270: Should both the `get_views_data*` functions be made private? (Prefixed with an underscore). It seems like the intent is to always use `get_ocio_config_view(config_path)`

It's the same with the functions above for `get_colorspace_data` - they shouldn't be as "public" as these functions, no?

openpype/pipeline/colorspace.py
```diff
@@ -0,0 +1,464 @@
...
+        str: name of colorspace
+    """
+    if not config_data:
+        project_settings = project_settings or get_project_settings(
+            project_name
+        )
+        config_data = get_imageio_config(
+            project_name, host_name, project_settings)
+
+    config_path = config_data["path"]
```
BigRoy, line 155: Would this function make more sense when simplified/rewritten as just finding the first relevant color space from the list of color spaces of the config? It'd make it more versatile in use instead of it internally always needing to call `get_ocio_config_colorspaces`.

It feels to me that parsing from filenames like this might often be frequently done close after each other - e.g. parsing multiple files. Quering the color space files rules will be relatively slow since it requires accessing the OCIO config file each time.

```python
colorspaces = get_ocio_config_colorspaces(config_path)
for filepath in filepaths:
    colorspace = parse_colorspace_from_filepath(filepath, colorspaces=colorspaces)
```

This also looks to me like a much more friendly API to the user with much less arguments to pass. Plus it'd be much faster and a simpler function code body.

As a general recommendation I'd try to see if there are other areas where functions could be simplified like this. Needing to pass project name, host, project settings to all functions where they end up just parsing an OCIO config path I'd say should be rewritten to just take the config path.

openpype/pipeline/colorspace.py
```diff
@@ -0,0 +1,464 @@
...
+    if not colorspace_name:
+        log.info("No imageio file rule matched input path: '{}'".format(
+            path
+        ))
+        return None
+
+    # validate matching colorspace with config
+    if validate and config_data:
+        validate_imageio_colorspace_in_config(
+            config_data["path"], colorspace_name)
```
BigRoy, line 122: The body of `validate_imageio_colorspace_in_config` isn't too complex and doesn't really warrant being an own function.

I'd rewrite this to:
```python
# pseudocode
if colorspace_name not in get_ocio_config_colorspaces(config_path):
    raise
```

openpype/pipeline/colorspace.py
```diff
@@ -0,0 +1,464 @@
...
+            project_name, host_name, project_settings)
+
+    # match file rule from path
+    colorspace_name = None
+    for _frule_name, file_rule in file_rules.items():
+        pattern = file_rule["pattern"]
+        extension = file_rule["ext"]
+        ext_match = re.match(
+            r".*(?=.{})".format(extension), path
+        )
```
BigRoy, line 105: Should the extension potentially need to go through `re.escape()`?

openpype/pipeline/publish/publish_plugins.py
BigRoy: Can confirm this is an error and will break publishing for sequences.

openpype/pipeline/colorspace.py
```diff
@@ -0,0 +1,464 @@
...
+
+        # return all colorspaces
+        return_json_data = open(tmp_json_path).read()
+        return json.loads(return_json_data)
+
+
+def get_imageio_config(
+    project_name, host_name,
+    project_settings=None,
+    anatomy_data=None,
```
BigRoy, line 310: Should this be renamed to `template_data` maybe? Since the API functions to retrieve the data are also called: `get_template_data_from_session`, `get_template_data_with_names` and `get_template_data`?

website/docs/dev_colorspace.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Keys
+- **colorspace** - string value used in other publish plugins and loaders
+- **configData** - storing two versions of path.
```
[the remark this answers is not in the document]
tokejepsen: Yeah, I agree. Not sure `Data` is needed, but also why are we storing two paths?
`configData` could just be a string/path which always needs formatting to ensure its absolute.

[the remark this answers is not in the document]
jakubjezek001: No issue with renaming `configData` to `config` key. But my recommendation is to keep absolute path in the config for future reference in case something goes wrong. The same way as we store path in representations db document.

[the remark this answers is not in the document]
tokejepsen: Sounds good. Anything wrong with naming the entries `absolute` and `relative`?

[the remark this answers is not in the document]
BigRoy: One technically isn't relative but a template. And the absolute path might not be a valid absolute on your current machine if you're in a remote location or another OS. 

The `template` I think is nice. The `path` or `absolute` entries I'm not sure about. I would avoid giving to explicit of a meaning to it.

openpype/plugins/publish/integrate.py
```diff
@@ -529,6 +529,15 @@ def prepare_representation(self, repre,
...
         template_data["ext"] = repre["ext"]
 
+        # add template data for colorspaceData
+        if repre.get("colorspaceData"):
+            colorspace = repre["colorspaceData"]["colorspace"]
+            # replace spaces with underscores
+            # pipeline.colorspace.parse_colorspace_from_filepath
+            # is checking it with underscores too
+            colorspace = colorspace.replace(" ", "_")
+            template_data["colorspace"] = colorspace
```
tokejepsen, line 539: This seems a little hardcoded and not sure the reason for it?

jakubjezek001:
> I would imagine that if you have `project_settings/nuke/imageio/ocio_config/enabled` enabled then we can assume this as using OCIO with a custom path.

Yes your assumption is right, plan was to integrate hosts in next PR. But I had included the Nuke host in this PR. You can just enable OCIO config and add paths like in the example image.
image

openpype/pipeline/colorspace.py
```diff
@@ -0,0 +1,464 @@
...
+    formatting_data = deepcopy(anatomy_data)
+
+    # format the path for potential env vars
+    formatting_data.update(dict(**os.environ))
+
+    # format path for anatomy keys
+    formatted_path = StringTemplate(tempate_path).format(
+        formatting_data)
+
+    return os.path.abspath(formatted_path)
```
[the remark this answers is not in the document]
jakubjezek001: Yes that makes sense, I will change this.

openpype/settings/defaults/project_settings/fusion.json
jakubjezek001: I had implemented the new way into Nuke so inspiration of how this works could be taken from [here](https://github.com/ynput/OpenPype/pull/4195/files#diff-d496b587c94a631576b2d043b85b9c1b95394f41641e4681339c49d7a493176dR1888). Next step on the roadmap will be integrated in hosts.

tokejepsen:
> > I would imagine that if you have `project_settings/nuke/imageio/ocio_config/enabled` enabled then we can assume this as using OCIO with a custom path.
> 
> Yes your assumption is right, plan was to integrate hosts in next PR. But I had included the Nuke host in this PR. You can just enable OCIO config and add paths like in the example image. image

I think if this is to be merged, the Nuke implementation needs to be sorted. Currently if you enabled the new OCIO switch, the code wont use this. The new OCIO should be queried first for existence and have precedence over the old settings.

openpype/pipeline/colorspace.py
jakubjezek001: Honestly not sure how public and private works. From my perspective, this is public because it is imported into another module. But it is private form API perspective, yes.

openpype/pipeline/colorspace.py
jakubjezek001: The idea here was to make all DCC integrations with pixel data support required to have imageio shemas in settings. Weather they use them is other thing. But for example, even Photoshop would have available settings in config and file rules. Then it can use the config for some of its own config and file rules matching the config or file rules would be referencing the globally defined config. So I would rather keep this as required.

openpype/pipeline/colorspace.py
jakubjezek001: Yeah from the perspective of the actual state of code yes, but in the future, for example in settings we will want to validate user inputs.

openpype/pipeline/colorspace.py
jakubjezek001: Yeah, that could be also possible, but since I am using raw string this should be solid.

openpype/pipeline/colorspace.py
jakubjezek001: Yeah, it seems like you are right. We were previously using anatomy data and if you search trough openpype codebase this is always referred as `anatomy_data`. I will keep it this way.

openpype/plugins/publish/integrate.py
jakubjezek001: We are not supporting spaces in our path formatting. But If you check the code [here](https://github.com/ynput/OpenPype/pull/4195/files#diff-c4b595d393443040b1d7c9551351bfffe69f9c00ad87bd9b6243da47a643f337R163) were are supporting parsing from path on both variants, with or witout underscores. But the integreation will always convert spaces into underscores.

jakubjezek001:
> I think if this is to be merged, the Nuke implementation needs to be sorted. Currently if you enabled the new OCIO switch, the code wont use this. The new OCIO should be queried first for existence and have precedence over the old settings.

Perhaps I wasn't clear enough, but this is already merged. Please pull latest changes and try it.

openpype/pipeline/colorspace.py
tokejepsen: Unless you implement those settings in this PR, I think you should not have it as required but instead warn about missing settings.

openpype/pipeline/colorspace.py
tokejepsen: Think if the code is not validating the user inputs now, we should not anticipate it yet. This will just confuse later on.

jakubjezek001: We are validating it already but only in one function.

openpype/pipeline/colorspace.py
BigRoy: This comment wasn't resolved, was it?

openpype/pipeline/colorspace.py
BigRoy: @antirotor @mkolar thoughts on this? 

I'm mostly referring to the `get_views_data_subprocess(config_path)` which is the Py2 wrapper itself. Usually the user shouldn't need to call it explicitly because `get_ocio_config_view` itself decides based on Python version which to use.

openpype/pipeline/colorspace.py
BigRoy: Issue is still present in the code - not sure why it was marked as resolved?

openpype/pipeline/colorspace.py
```diff
@@ -0,0 +1,468 @@
...
+    colorspaces = get_ocio_config_colorspaces(config_path)
+    for colorspace_key in colorspaces:
+        # check underscored variant of colorspace name
+        # since we are reformating it in integrate.py
+        if colorspace_key.replace(" ", "_") in path:
+            colorspace_name = colorspace_key
+            break
+        if colorspace_key in path:
+            colorspace_name = colorspace_key
+            break
```
BigRoy, line 168: This seems like quite an unreliable parsing method. I feel like this should either use file rules or don't try this at all - or the logic should be made way more explicit.

A filepath with `aces` or `raw` anywhere in the path will match, but there might be other colorspaces that contain `aces` but are longer?

Some examples of those conflicts could be:
```
acesproxy
acesproxy_ap1
```
```
acescct
acescct_ap1
```
```
rlf_rc
rlf_rc2
rlf_rc3
rlf_rc4
```
```
out_rec709
out_rec709d60sim
```
```
Input - Sony - Linear - S-Gamut
Input - Sony - Linear - S-Gamut Daylight
```
```
Input - Sony - Linear - S-Gamut3
Input - Sony - Linear - S-Gamut3.Cine
```

I've just done a quick search through OCIO 1.0.3 where it could fail to find the right colorspace.

There's a couple of things we could try to do to perform better matching.

- Match only the filename, not the full path
- Take the LONGEST match that is found
- Or be even more explicit and e.g. ALWAYS expect an underscore before the colorspace or be in a certain location of the fname (e.g. before the extension, or before a dot, etc.). This is likely what'll make it most explicit.

openpype/pipeline/colorspace.py
jakubjezek001: Sorry, yeah my bad. I had converted it into issue https://github.com/ynput/OpenPype/issues/4348 but the automation is not adding the reference within this comment.

openpype/pipeline/colorspace.py
jakubjezek001: > - Match only the filename, not the full path
> - Take the LONGEST match that is found
> - Or be even more explicit and e.g. ALWAYS expect an underscore before the colorspace or be in a certain location of the fname (e.g. before the extension, or before a dot, etc.). This is likely what'll make it most explicit.

Yes that is very good point. Since it would be practical to keep full path, because some productions will wont to keep colorspace at folder level - and it is almost impossible to parse only part of name - we have only option to evaluate the longest match. 

I have add it into the next issue. I would address it in next patch release or do you suggest this feature will not be complete without this?

BigRoy: I'm not too bothered it being broken for now - the parsing will be somewhat unreliable no matter how far we get. :)

jakubjezek001: Well, we could at least try, but it will never be optimal for all, yeah.. Thanks man!

tokejepsen (approved):

openpype/settings/defaults/project_anatomy/attributes.json
```diff
@@ -19,8 +19,7 @@
         "blender/2-91",
         "harmony/20",
         "photoshop/2021",
-        "aftereffects/2021",
-        "unreal/4-26"
+        "aftereffects/2021"
```
iLLiCiTiT, line 22:
```suggestion
        "aftereffects/2021",
        "unreal/4-26"
```

iLLiCiTiT: Nevermind `unreal/4-26` is not in defaults...

openpype/plugins/publish/extract_colorspace_data.py
```diff
@@ -0,0 +1,47 @@
+import pyblish.api
+from openpype.pipeline import publish
+
+
+class ExtractColorspaceData(publish.ExtractorColormanaged):
```
BigRoy, line 5: ### Warning

This PR has broken publishing in Maya for a project that has NO color space config set due to this `ExtractColorspaceData` node.

I cannot publish e.g. a `pointcache` instance:

```python
  File "S:\openpype\OpenPype\.venv\lib\site-packages\pyblish\plugin.py", line 522, in __explicit_process
    runner(*args)
  File "S:\openpype\OpenPype\openpype\plugins\publish\extract_colorspace_data.py", line 46, in process
  File "S:\openpype\OpenPype\openpype\pipeline\publish\publish_plugins.py", line 376, in set_representation_colorspace
    colorspace_settings = self.get_colorspace_settings(context)
  File "S:\openpype\OpenPype\openpype\pipeline\publish\publish_plugins.py", line 331, in get_colorspace_settings
    anatomy_data=anatomy_data
  File "S:\openpype\OpenPype\openpype\pipeline\colorspace.py", line 365, in get_imageio_config
    "No OCIO config found in settings. It is "
FileExistsError: No OCIO config found in settings. It is either missing or there is typo in path inputs
```

This also goes to show this Extractor is INVALID.
It shouldn't enforce injecting colorspaceData for representations which are NOT a image/video/pixel format like e.g. a pointcache.

PR #4196, opened 2022-12-08
Maya: Enhanced ASS publishing

antirotor:
## Brief description
Reviewable ass extractor improvements.

## Description
This is feature parallel to client extractor, with settings ported to OpenPype workflows with objectset and project settings.

## Discussion
I don't agree with having an `exportSequence` attribute because in OpenPype workflows you can define frame range easily, so it makes more sense to just support single frame publishes (which it already does). I propose to remove this attribute to simplify code and workflow.
So to note, current workflow of using current frame to export is prone to user errors and unpredictability.

## Testing notes:
1. Create and publish Ass instance.
2. Use Arnold's kick executable to validate renders.


antirotor:
`exportSequence` is a matter of controversy indeed. I can imagine the need for quick option (from the user perspective) to simply check something and without having to set frame range and disabling frame range validator export just one frame for whatever reason...

This said, I would remove it from here because it will be much easier (again from the user perspective) to set it when we switch Maya to the new publisher.

tokejepsen (approved):

m-u-r-p-h-y (approved):
ASS instance publishing works as expected.


tested with Maya2023.3, MtoA 5.2.1.0

image

PR #4198, opened 2022-12-08
Unreal: Publishing and Loading for UAssets

simonebarbieri:
## Brief description
Implementation for the publishing and loading for UAssets in Unreal.

## Description
This PR implements the publishing and loading of UAssets in Unreal. At this moment, only a single UAssets can be published. Hence, it has been included a validator to check that the UAssets doesn't have any dependency before being published.


antirotor (changes requested):
I am not really sure if this will work as base for UAsset publishing yet - as we've discussed, it should be simple, checking for dependencies in validator, but look at what happened to me :) 

What I did was that I've just imported simple mesh from `.obj` - it correctly imported as **StaticMesh**. I can select it, create UAsset publishing instance and it will fail during the publishing on *Check no dependencies* with following:

```
Dependencies found: ['/Script/InterchangeCore', '/Script/InterchangeEngine', '/Script/NavigationSystem', '/Engine/EngineMaterials/WorldGridMaterial', '/Script/InterchangePipelines', '/Script/InterchangeFactoryNodes']
```
Question is how to deal with this - all of these are probably built-in asset. So they should be available on the project we are trying to load uaset in, right? So maybe we should allow these - and with it, what is the criteria for accepting dependencies and rejecting.

Can we say that descendants of `/Engine/*` path are all OK (because they are built in)? With `/Script/` it is complicated, but maybe we can say that only __direct__ descendant of `/Script` are OK, but all others are not (like `/Script/Foo` is OK, but `/Script/Bar/Alice` is not)?

antirotor:
So, I have this simple obj imported, I've created uasset (strangely, the mesh asset didn't moved to the instance folder, I had to do it manually). From that point everything seems fine, except publish can't find the instance of UAsset:

image

> **Note**
> Tested in Unreal 5.1

antirotor (approved):
works! thanks

PR #4201, opened 2022-12-09
Maya: pointcache

tokejepsen:
This is a fix for `cmds.refresh` not being queryable. See https://github.com/pypeclub/OpenPype/pull/4144

Gone with the `cmds.ogs` route but slightly confusing that its a toggle.


antirotor (approved):
tested in Maya 2020 and 2023 and it works.

PR #4202, opened 2022-12-09
Nuke: fixed imageio node overrides subset filter

jakubjezek001:
## Brief description
Subset filter works even mixture of letter sizes in subset name.

## Description
Once mixture of sizes was used (ex. `_this_ButThat`), subset filter was not able to detect correctly node override from imageio nodes preset.

## Testing notes:
1. Create new preset in `project_settings/nuke/imageio/nodes/overrideNodes`, which would override the **CreateWriteRender** plugin, with Node Node Class to `Write` from `project_settings/nuke/imageio/nodes/requiredNodes`
image

2. add subset name filter to `_bla_BlaBla`
3. add the `_bla_BlaBla` also to `project_settings/nuke/create/CreateWriteRender/defaults`
image

4. Open nuke, select a node, Create Render node and set Variant to _bla_BlaBla
5. created node should be pink

### DIFF
--- a/openpype/hosts/nuke/api/lib.py
+++ b/openpype/hosts/nuke/api/lib.py
@@ -611,7 +611,7 @@ def get_created_node_imageio_setting_legacy(nodeclass, creator, subset):
 
         if (
             onode["subsets"]
-            and not any(re.search(s, subset.lower()) for s in onode["subsets"])
+            and not any(re.search(s.lower(), subset.lower()) for s in onode["subsets"])
         ):
             continue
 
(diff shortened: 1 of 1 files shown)


64qam:
I can't approve it. The subset name I created was lowercase. However, the creator turned it into upper case. When I created an uppercase subset it didn't work either and still got not overridden node
image

jakubjezek001:
> I can't approve it. The subset name I created was lowercase. However, the creator turned it into upper case. When I created an uppercase subset it didn't work either and still got not overridden node image

Valid point, but it is the issue of backend I believe @iLLiCiTiT , right?

iLLiCiTiT:
To be honest I don't understand the PR so I can't tell.

What happened to @64qam is that his subset name template is `{family}_{Task}_{Variant}` (at least it looks like). In that case variant `"bla_bla"` must become `"Bla_bla"` and there is nothing I can do about it.

64qam (approved):

PR #4203, opened 2022-12-09
General: Default command for headless mode is interactive

iLLiCiTiT:
## Brief description
Print help instead of running `tray` by default if headless mode is enabled and no command is entered.

## Additional info
The only reason is that trying to launch tray when `headless` is explicitly defined is "not right".

The command won't do anything if `openpype_gui` executable is used on windows (there is no right command to run in that case).

## Testing notes:
1. Start openpype executable with only `--headless` argument


antirotor:
I guess that doing nothing is not good way to go. It might suggest something is wrong. I would print help instead

m-u-r-p-h-y (approved):
works both ways

image

image

PR #4204, opened 2022-12-09
Flame - loading multilayer with controlled layer names

jakubjezek001:
## Brief description
Layers can be renamed dynamically during loading to batch

## Description
Folowing settings were created for layer template and layer rename patterns:
- `project_settings/flame/load/LoadClipBatch/layer_rename_template`
- `project_settings/flame/load/LoadClipBatch/layer_rename_patterns`
image

Regular expression patterns could be used for detecting name patterns in loaded layers and if any match is found, then the layer is renamed to use of version context data and additional formatting keys found if the context of loading file and its layer. For example layer with name `rgba` and layer uid `BEAUTY:MasterBeauty` can be now renamed with template `{asset}_{subset} [{layerUID}]` to new name `shot01_renderMain [BEAUTY:MasterBeauty]`. 
image

This had been requested by a client who is using Flame for metadata slates. The metadata key layer will now show more than just `rgba`.

### DIFF
--- a/openpype/hosts/flame/plugins/load/load_clip.py
+++ b/openpype/hosts/flame/plugins/load/load_clip.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 import os
 import flame
 from pprint import pformat
@@ -25,6 +26,14 @@ class LoadClip(opfapi.ClipLoader):
     reel_name = "Loaded"
     clip_name_template = "{asset}_{subset}<_{output}>"
 
+    """ Anatomy keys from version context data and dynamically added:
+        - {layerName} - original layer name token
+        - {layerUID} - original layer UID token
+        - {originalBasename} - original clip name taken from file
+    """
+    layer_rename_template = "{asset}_{subset}<_{output}>"
+    layer_rename_patterns = []
+
     def load(self, context, name, namespace, options):
 
         # get flame objects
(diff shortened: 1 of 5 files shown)


openpype/hosts/flame/plugins/load/load_clip.py
```diff
@@ -38,8 +47,14 @@ def load(self, context, name, namespace, options):
         version_name = version.get("name", None)
         colorspace = self.get_colorspace(context)
 
+        # in case output is not in context replace key to representation
+        if not context["representation"]["context"].get("output"):
+            self.clip_name_template.replace("output", "representation")
```
iLLiCiTiT, line 52: There two lines won't do anything.  Replace called on string won't change the source string but will return new.

jakubjezek001: ah, yeah thanks

jakubjezek001:
Ready for merge ;)

openpype/hosts/flame/plugins/load/load_clip.py
```diff
@@ -38,8 +47,16 @@ def load(self, context, name, namespace, options):
         version_name = version.get("name", None)
         colorspace = self.get_colorspace(context)
 
+        # in case output is not in context replace key to representation
+        if not context["representation"]["context"].get("output"):
+            self.clip_name_template = self.clip_name_template.replace(
```
iLLiCiTiT, line 52: I just realized that if the load plugin is used for loading of 2+ representations and only one of them does not have `"output"` the template for others will be broken. My recommendation is to not change value on `self`.

iLLiCiTiT (approved):
Can't test if without flame but code looks ok.

PR #4207, opened 2022-12-12
Build:Revert .toml update of Gazu

kalisp:
## Brief description
This currently breaks build as `poetry.lock` doesn't match to `pyproject.toml`.

## Additional info
Change to .toml and .lock should not go directly to `develop`, but to `release/3.15.x`, as it would limit us to create additional bugfix release, if necessary.

Question is, if 0.8.32 is required for merged functionality.


iLLiCiTiT (approved):
Approved

PR #4209, opened 2022-12-12
Hiero: thumbnail from multilayer exr

jakubjezek001:
## Brief description
Hiero exctract thumbnail works on mutlilayer exr

## Description
Thumbnail creator needed to have specified `color` layer.

## Testing notes:
1. Open hiero workfile
2. add multilayer exr at timeline
3. Create publishable clip on the new clip
4. publish
5. see that all works well


jrsndl (changes requested):
This works for multilayer EXRs that have a colour channel [1].
Unfortunately it fails on EXRs with tech passes only, that often do not have any colour [2].

image

jakubjezek001:
> This works for multilayer EXRs that have a colour channel [1].
> Unfortunately it fails on EXRs with tech passes only, that often do not have any colour [2].
> 
> image

I wonder if this is not edge case nobody will hit in production. Yes we are having only `colour` layer set for thumbnail. I can do exception to soft fail in case no colour pass is present, because it is quite difficult to define which of technical passes should be used for  thumnails.

jrsndl:
>  exception to soft fail in case no colour pass is present, because it is quite difficult to define which of technical passes should be used for thumnails.
Either that, or if you can pull the list of layers from Hiero, just take a first one if colour is not present?

jakubjezek001:
> > exception to soft fail in case no colour pass is present, because it is quite difficult to define which of technical passes should be used for thumnails.
> 
> Either that, or if you can pull the list of layers from Hiero, just take a first one if colour is not present?

I would wrap this  known limitation and not to waste time on it, since this is quite unlike edge case.

jakubjezek001:
@jrsndl there is missing api support on Hiero to get list of available layers from a clip. I had tested the workaround with `nuke.layers(node)` but this works only in Nuke environment.

jrsndl (approved):
Works for multilayer EXRs that have colour (rgb/rgba) layer

PR #4210, opened 2022-12-12
General: Event callbacks pass event to callbacks as expected

iLLiCiTiT:
## Brief description
It seems that event callbacks were broken for all Python 2 hosts because of wrong check of function signature when a method was used.

## Description
To find out if arguments should be passed is tried to bound argumets to callback instead of checking the function specs on own. Also added option to pass event as kwarg to callback function (positional arguments have priority).

## Additional information
This PR was made because of issues in Publisher tool when combined with Nuke 12.* [in PR](https://github.com/pypeclub/OpenPype/pull/4006).

## Testing notes:
1. Try event callbacks in Python 2 host where are used for method bounding

### DIFF
--- a/openpype/lib/events.py
+++ b/openpype/lib/events.py
@@ -74,22 +74,52 @@ class EventCallback(object):
                 "Registered callback is not callable. \"{}\""
             ).format(str(func)))
 
-        # Collect additional data about function
-        #   - name
-        #   - path
-        #   - if expect argument or not
+        # Collect function name and path to file for logging
         func_name = func.__name__
         func_path = os.path.abspath(inspect.getfile(func))
+
+        # Get expected arguments from function spec
+        # - positional arguments are always preferred
+        expect_args = False
+        expect_kwargs = False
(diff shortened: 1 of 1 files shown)


jakubjezek001 (changes requested):
Tested in Nuke New publisher but something is still not clicking as it should. It is showing now workfile isntance and legacy convertor plugin but not the button in instance attributes. 

image

Once I click on any  instance it is doing also nothing and this is in terminal.
```
  File "C:\CODE\__PYPE\OpenPype\openpype\tools\utils\widgets.py", line 77, in mouseReleaseEvent
    self._mouse_release_callback()
  File "C:\CODE\__PYPE\OpenPype\openpype\tools\publisher\widgets\card_view_widgets.py", line 315, in _mouse_release_callback
    self.selected.emit(self._id, self._group_identifier, selection_type)
TypeError: selected(QString,QString) only accepts 2 argument(s), 3 given!
```

jakubjezek001 (approved):

PR #4211, opened 2022-12-12
Refactorization of folder coloring

Casgen:
## Brief description
Refactored the ability to alter folder colors.

## Description
In the Unreal host plugin, the SetFolderColor() function has been changed to alter the color on the fly when called. To distuingish the OpenPype folder from others, UOpenpypePublishInstances now have a function to color the paths in which they are located.

Since the color should be set firsthand, **a new UOpenPypeSettings class and .ini file has been created to alter the options by the user. That way we can set defaults without accesing the .cpp files directly. It can also be altered in the "Project Settings" tab in UE.** this could be used of course for future use.

## Additional info
The .ini file is included in the Config folder (Config/DefaultOpenPypeSettings.ini)

### Check whether the folder from OpenPype and it's subpaths are colored.
1. Create a new project.
2. After creating the project create a new asset within the OpenPype menu (for instance a camera).
3. After the creation, the newly generated folders should be colored in a sky blue color.

### Testing the `set_folder_color` function
1. Try to call the set_folder_color function within OpenPype (or within the cmd in UE) and choose a folder to color with a FLinearColor object (if you are not sure about the params. see with the 'help' function). Make sure that the directory is written correctly, ex. `unreal.OpenPypeLib.set_folder_color('/Game/OpenPype/Instances/cameraDefault', unreal.LinearColor(a=1.0,r=0.8,g=0.1,b=0.3), False)`
2. After executing the function (or executing the command in UE) the color should automatically change without reloading the engine. (A message will appear in the Output Log of UE if the color has been set correctly or not).

**NOTE: Since the OpenPypeLib.h is not a UObject anymore but a UBlueprintFunctionLibrary you can directly call the function without creating the UObject! (for instance, use `unreal.OpenPypeLib.set_folder_color(...)`**

### Changing the color in project settings

1. Go to "Project Settings" under the "Edit" tab.
2. Look for the OpenPype section and open its settings.
3. When opened, there should be an option for setting the folder color (If you don't see the option, dont forget to expand those tabs to show the options).
4. When you set a different color and reload UE, the folder should change their color to the newly set value.


simonebarbieri (approved):

openpype/hosts/unreal/integration/UE_5.0/Source/OpenPype/Private/OpenPypeSettings.cpp
```diff
@@ -0,0 +1,21 @@
+﻿// Fill out your copyright notice in the Description page of Project Settings.
+
+#include "OpenPypeSettings.h"
+
+#include "IPluginManager.h"
+#include "UObjectGlobals.h"
```
antirotor, line 6: At least in UE5.1 I had to change these includes like

```suggestion
#include "Interfaces/IPluginManager.h"
#include "UObject/UObjectGlobals.h"
```

openpype/hosts/unreal/integration/UE_5.0/Source/OpenPype/Public/OpenPypeSettings.h
```diff
@@ -0,0 +1,32 @@
+﻿// Fill out your copyright notice in the Description page of Project Settings.
+
+#pragma once
+
+#include "CoreMinimal.h"
+#include "Object.h"
```
antirotor, line 6: And again here:

```suggestion
#include "UObject/Object.h"
```

antirotor (changes requested):
It works great, apart of those includes in UE5.1 - thanks!

openpype/hosts/unreal/integration/UE_5.0/Source/OpenPype/Private/OpenPypeSettings.cpp
Casgen: Seems strange that it needs to be occasionally that specific. But changing it at to this in my workspace still works. I guess we just need to be specific around those include dirs.

antirotor (approved):

PR #4214, opened 2022-12-12
General: Extend 'IPluginPaths' to have more available methods

iLLiCiTiT:
## Brief description
Extended abilities of `IPluginPaths` to have methods for load, create and publish plugin paths. Use new methods to register plugins from addons.

## Description
Added more methods to collect paths to load, create and publish plugin paths. These methods are called on OpenPype plugins installation (was used only for create plugins before).

## Additional info
Renamed `collect_creator_plugin_paths` -> `collect_create_plugin_paths` on `ModulesManager` and `get_creator_plugin_paths` -> `get_create_plugin_paths` (added backwards compatibility and warning).

## Testing notes:
1. Loader plugins can be registered by addons

Resolves https://github.com/pypeclub/OpenPype/issues/4194

### DIFF
--- a/openpype/modules/interfaces.py
+++ b/openpype/modules/interfaces.py
@@ -70,6 +70,13 @@ class IPluginPaths(OpenPypeInterface):
             host_name (str): For which host are the plugins meant.
         """
 
+        if hasattr(self, "get_creator_plugin_paths"):
+            # TODO remove in 3.16
+            self.log.warning((
+                "DEPRECATION WARNING: Using method 'get_creator_plugin_paths'"
+                " which was renamed to 'get_create_plugin_paths'."
+            ))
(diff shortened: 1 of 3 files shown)


openpype/modules/interfaces.py
```diff
@@ -44,40 +44,78 @@ class IPluginPaths(OpenPypeInterface):
     def get_plugin_paths(self):
```
kalisp, line 44: I would like to see docstring here with an example as this method is required to be implemented.

kalisp (approved):

openpype/modules/interfaces.py
iLLiCiTiT: I would like to remove that method completelly in future.

PR #4215, opened 2022-12-12
General: Use qtpy in settings UI

iLLiCiTiT:
## Brief description
Settings UI is using `qtpy` instead of `Qt.py` module.

## Testing notes:
Settings tools should work on
- [x] Windows
- [ ] linux
- [ ] MacOS

### DIFF
--- a/openpype/tools/settings/settings/dict_mutable_widget.py
+++ b/openpype/tools/settings/settings/dict_mutable_widget.py
@@ -1,6 +1,6 @@
 from uuid import uuid4
 
-from Qt import QtWidgets, QtCore, QtGui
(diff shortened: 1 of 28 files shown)


antirotor (approved):
and the settings work - at least on windows...

openpype/tools/settings/settings/dict_mutable_widget.py
```diff
@@ -1,6 +1,6 @@
 from uuid import uuid4
 
-from Qt import QtWidgets, QtCore, QtGui
+from qtpy import QtWidgets, QtCore, QtGui
```
iLLiCiTiT, line 3:
```suggestion
from qtpy import QtWidgets, QtCore
```

PR #4217, opened 2022-12-12
Publish: toggle off burnins for hosts

kalisp:
## Brief description
Enhanced profiles for toggling burnin.

## Description
Current defaults configures default output for all hosts. If any host should be configured specifically (for example turned OFF), it should be added there explicitly.
(Eg. `webpublisher` burnins should be turned off >> set `webpublisher` host in Settins and put there no output.)

## Testing notes:
1. start with this step
2. follow this step


m-u-r-p-h-y:
There is some warning when starting OP

image

m-u-r-p-h-y (approved):
traypublisher publishes without burnins after configured in settings with empty burnin

image

image

PR #4222, opened 2022-12-13
bugfix/image plane load error

moonyuet:
## Brief description
Fix the image plane load error

## Description
Fix the bugs and remove some attributes according to https://github.com/pypeclub/OpenPype/issues/3256

## Additional info
No additional info

## Testing notes:
1. Openpype -> Load and select exr sequence
2. right mouse click to load the exrs as image plane sequence
3. create a new camera

### DIFF
--- a/openpype/hosts/maya/plugins/load/load_image_plane.py
+++ b/openpype/hosts/maya/plugins/load/load_image_plane.py
@@ -72,6 +72,8 @@ class CameraWindow(QtWidgets.QDialog):
 
     def on_cancel_pressed(self):
         self.camera = None
+        if self.camera:
+            return
         self.close()
 
 
--- a/openpype/hosts/maya/plugins/load/load_image_plane.py
+++ b/openpype/hosts/maya/plugins/load/load_image_plane.py
@@ -23,8 +23,6 @@ class CameraWindow(QtWidgets.QDialog):
         self.setWindowFlags(self.windowFlags() | QtCore.Qt.FramelessWindowHint)
 
         self.camera = None
-        # self.static_image_plane = False
-        # self.show_in_all_views = False
 
         self.widgets = {
             "label": QtWidgets.QLabel("Select camera for image plane."),
@@ -45,8 +43,6 @@ class CameraWindow(QtWidgets.QDialog):
         for camera in cameras:
             self.widgets["list"].addItem(camera)
 
-        # self.widgets["staticImagePlane"].setText("Make Image Plane Static")
-        # self.widgets["showInAllViews"].setText("Show Image Plane in All Views")
 
         # Build buttons.
         layout = QtWidgets.QHBoxLayout(self.widgets["buttons"])
@@ -57,8 +53,6 @@ class CameraWindow(QtWidgets.QDialog):
         layout = QtWidgets.QVBoxLayout(self)
         layout.addWidget(self.widgets["label"])
         layout.addWidget(self.widgets["list"])
-        # layout.addWidget(self.widgets["staticImagePlane"])
-        # layout.addWidget(self.widgets["showInAllViews"])
         layout.addWidget(self.widgets["buttons"])
         layout.addWidget(self.widgets["warning"])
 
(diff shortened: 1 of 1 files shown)


LiborBatek:
Following the testing steps, it shows this error msg:
image

however the image sequence get connected to the selected cam (chosen during the loading action co connect imagePlane to my shot cam)
image

LiborBatek:
gives me the error even when creating new camera during loading action of imagePlane not trying to attach it to the existing camera (creating new one)
image

moonyuet:
> gives me the error even when creating new camera during loading action of imagePlane not trying to attach it to the existing camera (creating new one) image
I have updated the PR to fix the setAttr issue. This could be because time1 node, which connects to the frameExtension of the image plane, doesn't allow to be deleted in maya. I can only disconnect the nodes instead of deleting it.

LiborBatek (changes requested):
The error with ```frame.extension``` is not showing anymore so its fixed but there is still other issue with ```frame offset``` being wrong set to ```1000```.
PR4222_imagePlane_offset01
When manually set to correct value (zero) the image sequence being correctly set and visible.
image
Could be also just an issue in ```maya2023``` not ```maya2022```. 
I think its necessary to keep default ```frame offset``` at zero value.

LiborBatek:
Tested it again  and  ```frame offset``` being set to ```1000``` in ```maya2022``` it shows image seq correctly and behaves differently than in ```maya2023``` were the image seq is not visible.  Seems like maya bug to me which was fixed in maya2023. I propose to keep the offset at zero by defaults.

m-u-r-p-h-y:
@LiborBatek can you check all items mentioned in the original issue https://github.com/pypeclub/OpenPype/issues/3256

moonyuet:
> Tested it again in `maya2022` `frame offset` being set to `1000` it shows image seq despite the offset value (it shouldnt be visible because of that offset) and behaves differently than in `maya2023` were the image seq is not visible as its out of frame range (seq 1001-1005 being offset by another 1000). Seems like maya2022 bug to me which was fixed in maya2023. Anyways I propose to keep/set the offset at zero by defaults. User can offset it manually when needed...

I will remove or set the frameoffset to zero but maybe need to wait because currently i can't really commit to the code and it seems that I dont have the access right to push the latest codes to the PR

LiborBatek (approved):
Did all tests in ```maya2022``` and ```maya2023``` now all correct.  ```frame offset``` being kept at ```zero``` value. Also tested to load multiple imagePlanes on the same ```camera``` and all worked fine.
Went through old PRs to see if anything was not left out for consideration. Seems ok to me.

m-u-r-p-h-y (changes requested):
The cancel button is still not working
image


```
  File "D:\REPO\OpenPype\openpype\tools\loader\widgets.py", line 1578, in _load_representations_by_loader
    options=options
  File "D:\REPO\OpenPype\openpype\pipeline\load\utils.py", line 299, in load_with_repre_context
  File "D:\REPO\OpenPype\openpype\hosts\maya\plugins\load\load_image_plane.py", line 112, in load
    camera = camera_names[window.camera]
KeyError: None
```

m-u-r-p-h-y (commented):
Now there is no error appearing after pressing the Cancel button, but nothing is happening, the dialog is still there.

m-u-r-p-h-y (approved):
the `Cancel` button works now . . .    :)

openpype/hosts/maya/plugins/load/load_image_plane.py
```diff
@@ -106,23 +97,21 @@ def load(self, context, name, namespace, data, options=None):
...
             window = CameraWindow(camera_names.keys())
             window.exec_()
-            camera = camera_names[window.camera]
-
-            is_static_image_plane = window.static_image_plane
-            is_in_all_views = window.show_in_all_views
+            try:
+                camera = camera_names[window.camera]
+            except KeyError:
+                pass
```
iLLiCiTiT, line 114:
```suggestion
            camera = camera_names.get(window.camera)
```

PR #4226, opened 2022-12-13
Blender: Fix Create Camera "Use selection"

simonebarbieri:
## Brief description
Creating "Camera" subset was creating a new camera even when the user selects one before create action, ignoring Use Selection option.

## Testing notes:
1. Create a new camera and select it.
2. From the menu `OpenPype`, select `Create`.
3. Select `Camera` family, keep the `Use selection` option ticked, and press `Create`.
4. It should add the previously created camera to the instance, and it shouldn't create any new camera.


LiborBatek (approved):
It works fine. I have also tested other option without "Use selection" and sucessfuly creates camera if not any present in the scene. So works in any circumstances.

PR #4227, opened 2022-12-14
Blender: Fix Layout Family Versioning

simonebarbieri:
## Brief description
This PR fixes a problem occurring when trying to update a layout in Blender. It also relink the updated objects in animation instances, if they exist.

### DIFF
--- a/openpype/hosts/blender/plugins/load/load_layout_blend.py
+++ b/openpype/hosts/blender/plugins/load/load_layout_blend.py
@@ -6,6 +6,7 @@ from typing import Dict, List, Optional
 
 import bpy
 
+from openpype import lib
 from openpype.pipeline import (
     legacy_create,
     get_representation_path,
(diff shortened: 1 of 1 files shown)


LiborBatek:
Ok, I have a layout asset loaded in version v01 (cave)
image

after hitting "update to latest" the layout asset dissapear
image

when trigger "switch to hero" it shows up again but not when switch to versioned again. When doing back and forward switching between versioned and hero it shows up again. But not as supposed to work.
image

LiborBatek (changes requested):
Need to be changed, pls see comments.

LiborBatek:
In short, when using "update to latest" action, it fails to update the asset, after that trying "change to hero" action works. However when trigger just the "change to hero" immedeately  without any previous tries it does not work and the asset dissapear too...

When playing with various actions later afterwards the asset suddenly shows up again. But in a weird and unexpected ways...in some circumstances yes and no. Sometimes triggering the update actions multiple times solves the issue and the asset show up in the viewport again.

PR #4228, opened 2022-12-14
Blender: Fix Layout Family Versioning

simonebarbieri:
## Brief description
This PR fixes a problem occurring when trying to update a layout in Blender. It also relink the updated objects in animation instances, if they exist.


LiborBatek:
The asset "layout" cannot be updated still. Gives following error...
image

LiborBatek (changes requested):
Need to be changed. Still not working.

LiborBatek (approved):
Its working ok. Been tested goin up / down in versions and set to hero version and back. All fine.

PR #4234, opened 2022-12-15
Blender: Workfile Loader

simonebarbieri:
## Brief description
Implemented the importer for other blend workfiles.

## Description
The workfile importer will import everything from another blend file, but it will not create any container for it. It will just merge the content in the current blend file.

### DIFF
--- a/openpype/hosts/blender/plugins/load/import_workfile.py
+++ b/openpype/hosts/blender/plugins/load/import_workfile.py
@@ -1,0 +1,38 @@
+from pathlib import Path
+
+import bpy
+
+from openpype.hosts.blender.api import plugin
+
+
+class ImportBlendLoader(plugin.AssetLoader):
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
It works as supposed to without any problem.

LiborBatek:
My only concern (not blocking merging of this PR) it would be better to call this function ```append file``` not ```link``` because its not basically linking...by blender terms.

LiborBatek (changes requested):
Necessary to adjust it as discussed together. Which means there will be 2 operation types:
append function (basically as blender which creates another scene in blender file)
import function (just importing workfile into existing blender scene creating new collection of imported objects)
The core function is working now but needs to be enhanced/polished.

LiborBatek (approved):
It has been tested both using ```append workfile``` and ```import workfile``` and both works ok!

PR #4238, opened 2022-12-16
Flame: Use qtpy in flame

iLLiCiTiT:
## Brief description
Use `qtpy` in flame host implementation instead of `Qt.py`.

## Additional information

## Testing notes:
Flame UIs should work as did before.


jakubjezek001 (approved):
it is working

PR #4242, opened 2022-12-18
Deadline: fix dependencies

### LINKED ISSUE #4164: Maya: Deadline job dependencies
**Running version**
 3.14.8-nightly.2

**Describe the bug**
Deadline job dependencies are not set correctly.

The main render job has file dependency set to two files. Workfile and published workfile. Only one is used for rendering though.
image

and if I remember correctly it was decided to disable this dependency as it was creating some issues in certain network configurations that lead to an infinite wait for the dependency to be fulfilled even though the render scene existed. 

another issue is the changed logic of tile assembly job dependency. It should wait for the main render job to finish rendering of individual tiles to be assembled to final image, Now the dependency is set to the same dependency as the main render job
image

causing to start assembly before the tile rendering is finished

I found some test jobs in my deadline repo from September when it was working ok ->
image

[cuID:[OP-4556](https://app.clickup.com/t/37hfex2)]

antirotor:
## Fix

This is fixing how job and asset dependencies were added to the jobs. First issue was that scene was added as asset dependency twice, but as original scene and published too. Second issue was that job dependencies for tile assembler job were added under wrong key and therefor they were not taken into consideration.

### Testing

Submit job from Maya having tiles. Tile assembly jobs should have proper Job dependency - pointing to render job. All jobs should have proper asset dependency - only one scene.

Resolves #4164

### DIFF
--- a/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
+++ b/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
@@ -183,7 +183,6 @@ class MayaSubmitDeadline(abstract_submit_deadline.AbstractSubmitDeadline):
         # Adding file dependencies.
         if self.asset_dependencies:
             dependencies = instance.context.data["fileDependencies"]
-            dependencies.append(context.data["currentFile"])
             for dependency in dependencies:
                 job_info.AssetDependency += dependency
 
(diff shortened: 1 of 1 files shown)


LiborBatek:
Seem that dependencies works ok

image

but there is some issue with tile assembly job "Empty file name is not legal" parameter ```sourceFileName```

image

@antirotor  So not sure if change needed or this PR is ok with its portion functional and these issues not connected to it.

antirotor:
> but there is some issue with tile assembly job "Empty file name is not legal" parameter `sourceFileName`

Yep, there is still bug for tile assembly but that is not connected to dependencies and deservers another PR

LiborBatek (approved):

PR #4246, opened 2022-12-20
Global: abstraction for color config and mapping

jakubjezek001:
## Brief description
Adding functions for config and mapping resolving from settings.

## Description
Colorspace settings are now fully supported in api. We can resolve host related color config path and color file rules mapping.

## Additional info
This is second PR from series of 4. Next we will focus more on publish and loading plugins.


openpype/pipeline/colorspace.py
```diff
@@ -0,0 +1,116 @@
...
+import platform
+import PyOpenColorIO as ocio
+from openpype.settings import get_project_settings
+from openpype.pipeline import legacy_io
+from openpype.lib.log import Logger
+
+log = Logger.get_logger(__name__)
+
+
+def get_imagio_colorspace_from_filepath(
```
iLLiCiTiT, line 12: Typo
```suggestion
def get_imageio_colorspace_from_filepath(
```

openpype/pipeline/__init__.py
```diff
@@ -8,6 +8,14 @@
 )
 from .anatomy import Anatomy
 
+from .colorspace import (
```
iLLiCiTiT, line 11: Content of `colorspace` requires specific dependencies thus should not be in `pipeline/__init__.py`.

openpype/pipeline/colorspace.py
```diff
@@ -0,0 +1,116 @@
...
+    path, host=None, project_name=None, validate=True
+):
+    project_name = project_name or legacy_io.Session["AVALON_PROJECT"]
+    host = host or legacy_io.Session["AVALON_APP"]
```
iLLiCiTiT, line 16: Rather use `host_name`. I would expect that `host` is an object of registered host.

openpype/pipeline/colorspace.py
```diff
@@ -0,0 +1,116 @@
...
+import PyOpenColorIO as ocio
+from openpype.settings import get_project_settings
+from openpype.pipeline import legacy_io
+from openpype.lib.log import Logger
+
+log = Logger.get_logger(__name__)
+
+
+def get_imagio_colorspace_from_filepath(
+    path, host=None, project_name=None, validate=True
```
iLLiCiTiT, line 13: I would like to be able pass project settings. Also it would be better to get the settings only once when this function is called. Both `get_imageio_file_rules` and `get_imageio_config` are calling `get_project_settings`.

openpype/pipeline/colorspace.py
```diff
@@ -0,0 +1,116 @@
...
+
+def validate_imageio_colorspace_in_config(config_obj, colorspace_name):
+    if not config_obj.getColorSpace(colorspace_name):
+        raise ocio.Exception(
+            "Missing colorspace '{}' in config file '{}'".format(
+                colorspace_name, config_obj.getWorkingDir())
+        )
+
+
+def validate_imageio_config_from_path(config_path):
```
iLLiCiTiT, line 58: I would expect the validation would return `True/False` instead of reraising (Possibly different error).
```
def create_imageio_config(config_path):
    return ocio.Config.CreateFromFile(config_path)


def validate_imageio_config_from_path(config_path):
    try:
        create_imageio_config(config_path)
    except ocio.Exception:
        return False
    return True
```

openpype/pipeline/colorspace.py
jakubjezek001: fixed in https://github.com/pypeclub/OpenPype/pull/4270

openpype/pipeline/colorspace.py
jakubjezek001: fixed in https://github.com/pypeclub/OpenPype/pull/4270

openpype/pipeline/colorspace.py
jakubjezek001: fixed in https://github.com/pypeclub/OpenPype/pull/4270

jakubjezek001:
All code is part of https://github.com/pypeclub/OpenPype/pull/4270

openpype/pipeline/__init__.py
jakubjezek001: This is solved in https://github.com/pypeclub/OpenPype/pull/4270 by bypassing it via subprocessing script file.

PR #4250, opened 2022-12-20
Maya: Use qtpy in Maya

iLLiCiTiT:
## Brief description
Use `qtpy` in maya host implementation instead of `Qt.py`.

## Additional information

## Testing notes:
Maya UIs should work as did before (even in Maya 2020).

### DIFF
--- a/openpype/hosts/maya/api/commands.py
+++ b/openpype/hosts/maya/api/commands.py
@@ -39,7 +39,7 @@ class ToolWindows:
 
 
 def edit_shader_definitions():
-    from Qt import QtWidgets
+    from qtpy import QtWidgets
     from openpype.hosts.maya.api.shader_definition_editor import (
         ShaderDefinitionsEditor
     )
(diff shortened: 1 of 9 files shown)


m-u-r-p-h-y (approved):
If it is only about loading OP menu and tools it works in all major Maya versions

image

BigRoy:
I believe it would also be for the "Scene has outdated content" popup that you get when opening a scene containing versions which are not the latest version.
It would also be good to double check things like the Workfile save version pop-up and right click in the Loader to load something.

Also double check the Look Assigner in maya maybe? (or isn't that influenced by changes in this PR?)

iLLiCiTiT:
> Also double check the Look Assigner in maya maybe? (or isn't that influenced by changes in this PR?)

Nope. Not yet. That will be big PR affecting all hosts. This is just for maya inner implementation.

antirotor (approved):
Tested all in Maya 2020 (py2) and Maya 2023 (py3), all dialogs I could produce, everything is working fine

PR #4251, opened 2022-12-20
Max: Use qtpy in Max

iLLiCiTiT:
## Brief description
Use `qtpy` in max host implementation instead of `Qt.py`.

## Additional information

## Testing notes:
Max UIs should work as did before.


antirotor (approved):
works as expected

PR #4252, opened 2022-12-20
Houdini: Use qtpy in Houdini

iLLiCiTiT:
## Brief description
Use `qtpy` in houdini host implementation instead of `Qt.py`.

## Additional information

## Testing notes:
Houdini UIs should work as did before.

### DIFF
--- a/openpype/hosts/houdini/api/usd.py
+++ b/openpype/hosts/houdini/api/usd.py
@@ -3,7 +3,7 @@
 import contextlib
 import logging
 
-from Qt import QtWidgets, QtCore, QtGui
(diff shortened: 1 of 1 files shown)


antirotor (approved):
Tested in Houdini 18.5 (py 2) and Houdini 19.5 (py 3) -UI works, trouble is that py 2 Houdini is currently broken (some use of f-strings, but more importantly Publisher creators are not shown). But that is not connected to this PR.

iLLiCiTiT:
> but more importantly Publisher creators are not shown

Should be fixed with this https://github.com/pypeclub/OpenPype/pull/4249

PR #4254, opened 2022-12-21
Resolve: Use qtpy in Resolve

iLLiCiTiT:
## Brief description
Use `qtpy` in resolve host implementation instead of `Qt.py`.

## Additional information

## Testing notes:
Resolve UIs should work as did before.


jakubjezek001 (approved):
works well

PR #2932, opened 2022-03-23
Maya: Do not pass `set` to maya commands (fixes support for older maya versions)

BigRoy:
## Implementation

This fixes support for Maya 2018

Older versions of Maya do not allow `set` type to be passed to Maya commands and will result in e.g. "RuntimeError: # Syntax error: unexpected end ( at position 4 while parsing"

E.g. error as [reported on discord](https://discord.com/channels/517362899170230292/563751989075378201/955419340826628157):

```
# Traceback (most recent call last):
#   File "C:\Users\22DOGS\OpenPype\openpype\tools\mayalookassigner\app.py", line 253, in on_process_selected
#     assign_look_by_version(nodes, version_id=version["_id"])
#   File "C:\Users\22DOGS\OpenPype\openpype\hosts\maya\api\lib.py", line 1600, in assign_look_by_version
#     shader_nodes = get_container_members(container_node)
#   File "C:\Users\22DOGS\OpenPype\openpype\hosts\maya\api\lib.py", line 1519, in get_container_members
#     for ref in cmds.ls(members, exactType="reference", objectsOnly=True):
# RuntimeError: # Syntax error: unexpected end ( at position 4 while parsing:
#     set([u'Props_lookMain_01_RN'])
#        ^
: set([u'Props_lookMain_01_RN'])
```

## Brief description

This fixes loaded version updating, shader assignments and more areas that relied on `get_container_members`. 

## Testing notes:
1. start maya versions
2. ensure updating versions of loaded content continues to work, and lookdev shader assignments still work

PR #4256, opened 2022-12-21
Maya: Xgen

tokejepsen:
## Brief description
Initial Xgen implementation.

## Description
Client request of Xgen pipeline.

## Additional info
Currently have publishing and loading, but with two issues:

- [x] Critical: Duplicate collection errors on launching workfile
- [x] Non-critical: errors on opening workfile and failed opening of published xgen.

The critical issue can be worked around by opening the workfile again once Maya is launched. It might be a better solution to provide a way for Maya to be launched without workfile as argument but load the workfile when Maya is initialized to work around these two issues.

## Testing notes:
1. Enable `project_settings/maya/open_workfile_post_initialization`, to work around the mentioned non-critical and critical errors.
2. Create Xgen collection and instance, then publish.
3. Load Xgen from Loader.
4. Load geometry animation from Loader (needs publishing of animation or pointcache of the animated geometry used to generate the xgen on).
5. Select the animation and xgen container in the inventory > Actions > Connect Geometry.
6. Load curves animation/fx from Loader (needs publishing of animation or pointcache of the animated curves used as guides in the xgen).
7. Select the animation/fx and xgen container in the inventory > Actions > Connect Xgen.
8. Create render instance and publish to Deadline.


tokejepsen:
Duplicate error might be due to taking all files from workfile xgen to publish resources. Will investigate more.

openpype/hosts/maya/plugins/publish/extract_xgen.py
```diff
@@ -0,0 +1,139 @@
...
+            pc.connectAttr(
+                "{}.matrix".format(duplicate_transform),
+                "{}.transform".format(node),
+                force=True
+            )
+            pc.connectAttr(
+                "{}.worldMesh".format(duplicate_shape),
+                "{}.geometry".format(node),
+                force=True
+            )
```
BigRoy, line 74: I would rewrite this code to not require `pymel`. Looking at how this is written you should be able to just replace it directly with `from maya import cmds` and replace `pc.` with `cmds.` except for the fact that it "reparents" and "duplicates" and thus the long name changes.
```suggestion
            # Duplicate_transform subd patch geometry.
            duplicate_transform = cmds.duplicate(transform_name)[0]
            duplicate_shape = cmds.listRelatives(
                duplicate_transform, 
                shapes=True, 
                fullPath=True
            )[0]

            cmds.connectAttr(
                "{}.matrix".format(duplicate_transform),
                "{}.transform".format(node),
                force=True
            )
            cmds.connectAttr(
                "{}.worldMesh".format(duplicate_shape),
                "{}.geometry".format(node),
                force=True
            )

            duplicate_transform = cmds.parent(duplicate_transform, world=True)[0]
```
I'd expect this to work.

openpype/hosts/maya/plugins/publish/extract_xgen.py
```diff
@@ -0,0 +1,139 @@
...
+
+        # Collect nodes to export.
+        duplicate_nodes = []
+        for node, connections in instance.data["xgenConnections"].items():
+            transform_name = connections["transform"].split(".")[0]
+
+            # Duplicate_transform subd patch geometry.
+            duplicate_transform = pc.duplicate(transform_name)[0]
+            pc.parent(duplicate_transform, world=True)
+            duplicate_transform.name(stripNamespace=True)
```
BigRoy, line 60: This query is unused. `.name()` as far as I know returns a value and doesn't change the actual node. So `stripNamespace=True` here does nothing to the node it's acting upon.

openpype/hosts/maya/plugins/publish/extract_xgen.py
tokejepsen: Any reason for not stripping the namespaces in your suggestion?

openpype/hosts/maya/plugins/publish/validate_xgen.py
```diff
@@ -0,0 +1,61 @@
...
+    order = pyblish.api.ValidatorOrder
+    host = ["maya"]
+    families = ["xgen"]
+
+    def process(self, instance):
+        # Validate only xgen collections are in objectset.
+        nodes = (
+            instance.data["xgenNodes"] +
+            cmds.ls(instance, type="transform", long=True)
+        )
```
BigRoy, line 22: Lookup `node in nodes` is being doen below. Turns this into a `set` for optimal lookup.

```suggestion
        nodes = set(
            instance.data["xgenNodes"] +
            cmds.ls(instance, type="transform", long=True)
        )
```

openpype/hosts/maya/plugins/publish/extract_xgen.py
tokejepsen: Good point.

openpype/hosts/maya/plugins/publish/extract_xgen.py
tokejepsen: nvm :)

openpype/hosts/maya/plugins/publish/validate_xgen.py
```diff
@@ -0,0 +1,61 @@
...
+            description = description.split("|")[-2]
+            modifier_names = xgenm.fxModules(palette, description)
+            for name in modifier_names:
+                attr = xgenm.getAttr("active", palette, description, name)
+                # Attribute value are lowercase strings of false/true.
+                if attr == "false":
+                    try:
+                        deactive_modifiers[description].append(name)
+                    except KeyError:
+                        deactive_modifiers[description] = [name]
```
BigRoy, line 53: I feel like this would be more readable using a `defaultdict(list)` type to begin with. I don't believe that's directly supported `json.dumps` so maybe even `deactive_modifiers.setdefault("description", list).append(name)` makes more sense?

Also **deactive** should be **inactive** in English.

openpype/hosts/maya/plugins/publish/extract_xgen.py
```diff
@@ -0,0 +1,139 @@
...
+        xgen_dir = os.path.join(
+            os.path.dirname(instance.context.data["currentFile"]), "xgen"
+        )
+        transfers = []
+        for root, dirs, files in os.walk(xgen_dir):
+            for file in files:
+                source = os.path.join(root, file)
+                destination = source.replace(
+                    xgen_dir, instance.data["resourcesDir"]
+                )
```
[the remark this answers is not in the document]
tokejepsen: Sure, we can do that :)

openpype/hosts/maya/plugins/publish/extract_xgen.py
BigRoy: Could we include in a comment why there is a need to duplicate the input transforms, etc? Why do we need that. Couldn't we just export a maya file with `preserveReferences = false` since duplicating breaks the reference connection anyway? What's the need for duplicating it?

openpype/hosts/maya/plugins/publish/validate_xgen.py
tokejepsen: Sorry, could you elaborate on why a set is needed here?

BigRoy: A check whether a string is inside the list is done down below:
```python
        for node in instance:
            if node in nodes:
                continue
```

This will be `On` depending on where it is in the list but the size of the list will have a heavy impact on the performance, whereas lookup in the set is `O1`. 

Some topics:
- https://stackoverflow.com/a/2831242
- https://www.reddit.com/r/Python/comments/j9ncmd/lookup_time_in_a_list_vs_set/
- https://lucasmagnum.medium.com/pythontip-list-vs-set-performance-experiments-dfbe4f72d47f

openpype/hosts/maya/plugins/publish/validate_xgen.py
tokejepsen: Hmm, feel like using `defaultdict(list)` might overcomplicate things rather than a try/except. Will change to `inactive`.

openpype/hosts/maya/plugins/publish/extract_xgen.py
tokejepsen: We need to strip out any connections the geometry has to the current scene. For example when using some rig geometry that is nested deep in the rig.
Up for suggestions on how to solve this without duplicating and parent to the world.

tokejepsen:
Elaboration of errors:

There is a non-critical errors when opening the workfiles. This is due to a workflow issue where we reference in the xgen collections which means initial the xgen is expecting to find a file by the xgFileName attribute from the referenced xgen. Later in the startup process the xgens xgFileName attribute is ref-editted so the xgen is imported correctly.

```
Importing XGen Collections...
# Error: XGen:  Failed to open file: P:/PROJECTS/OP01_CG_demo/shots/sh040/work/Lighting/cg_ball_xgenMain_v035__ball_rigMain_01___collection.xgen  # 
# Error: XGen:  Failed to import collection from file P:/PROJECTS/OP01_CG_demo/shots/sh040/work/Lighting/cg_ball_xgenMain_v035__ball_rigMain_01___collection.xgen  # 
Done
```

Loading duplicate xgen issue only happens when a new session of Maya and not when opening the Maya file again afterwards.

```
Importing XGen Collections...
# Error: XGen:  Failed to find description ball_xgenMain_01_:parent in collection ball_xgenMain_01_:collection. Abort applying delta: P:/PROJECTS/OP01_CG_demo/shots/sh040/work/Lighting/cg_sh040_Lighting_v001__ball_xgenMain_01___collection.xgen  # 
# Error: XGen:  Tried to import a duplicate collection, ball_xgenMain_02_:collection, from file P:/PROJECTS/OP01_CG_demo/shots/sh040/work/Lighting/cg_sh040_Lighting_v001__ball_xgenMain_02___collection.xgen. Aborting import.  # 
Done.
```

openpype/hosts/maya/plugins/publish/extract_xgen.py
BigRoy: > We need to strip out any connections the geometry has to the current scene. For example when using some rig geometry that is nested deep in the rig.
Up for suggestions on how to solve this without duplicating and parent to the world.

I see, thanks. I believe you should be able to add these arguments to the cmds file export `channels=False` and `constructionHistory=False`. However, whenever you do so any connections you want to actively persist in the export would need to be included in the "selection" that you've set to export.

So basically if you're able to identify which actual objects (including its history) you want to have end up in the maya file you should be able to add those arguments (make sure `preserveReference=False` otherwise any reference in your selection would still fully be included).

I believe if the selection actively includes animation curves (if you ever need them) then `channels=False` and `constructionHistory=False` will still include it in the export and connect them as they were as long as they are part of the active selection in an `exportSelected=True` export.

However, you might need to double test this. :)

BigRoy:
> Elaboration of errors:

Thanks - could those errors be worked around by somehow setting the default state of the XGEN export to inactive. Like disable it by default and actually also set the enabled state on load as a reference edit? So that it's only ever active once reference edits kicked in?

openpype/hosts/maya/plugins/publish/extract_xgen.py
tokejepsen: But this includes the namespaces of nodes, right?

tokejepsen:
> > Elaboration of errors:
> 
> Thanks - could those errors be worked around by somehow setting the default state of the XGEN export to inactive. Like disable it by default and actually also set the enabled state on load as a reference edit? So that it's only ever active once reference edits kicked in?

Thanks will look into this later.
I'm gunna investigate only publishing the xgen files we need. Think that could resolve the duplicate error issue.

openpype/hosts/maya/plugins/publish/extract_xgen.py
BigRoy: > But this includes the namespaces of nodes, right?

Yes, you are correct. It doesn't remove namespaces as far as I'm aware.

LiborBatek:
Did quick test with creating subset ```Xgen``` and tried to publish it.

Gives me this error during validate:
image

Note: the rigged char is referenced via Asset Loader and then Xgen created locally in the lookdev scene (basically the maya scene I gave you at the begining)

tokejepsen:
> Did quick test with creating subset `Xgen` and tried to publish it (first selecting the xgen description in outliner ant then creating the subset in OP menu)

Looks like you have everything in the objectset?
Should only be the collection that goes into the objectset, but the validation error is not very clear on that either so will improve on that now.

openpype/hosts/maya/plugins/publish/validate_xgen.py
```diff
@@ -0,0 +1,64 @@
...
+        for node in instance:
+            if node in nodes:
+                continue
+            remainder_nodes.append(node)
+
+        msg = (
+            "Only the collection is used when publishing. Found these invalid"
+            " nodes in the objectset:\n{}".format(remainder_nodes)
+        )
+        assert not remainder_nodes, msg
```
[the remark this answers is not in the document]
tokejepsen: No skin off my nose, but why not use assertion?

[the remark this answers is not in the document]
BigRoy: Not doing assertion was mostly because:
- the if statement helped to only format the string if the error occurred - it might actually do that too if the string formatting is moved to after the comma in the assertion, but I'm not entirely sure.
- to prepare the code for the new publisher which will require raising `KnownPublishError` errors as far as I know.
- because technically assertion errors can be disabled - [e.g. with `PYTHONOPTIMIZE`](https://realpython.com/python-assert-statement/#disabling-assertions-in-production-for-performance) environment variable. Also see [_When not to use assertions_](https://realpython.com/python-assert-statement/#when-not-to-use-assertions)

tokejepsen:
> > Elaboration of errors:
> 
> Thanks - could those errors be worked around by somehow setting the default state of the XGEN export to inactive. Like disable it by default and actually also set the enabled state on load as a reference edit? So that it's only ever active once reference edits kicked in?

Does not look like there is any way of disabling Xgen by default. The duplication error also occurs when we resolve the non-critical error cause it basically imports the xgen twice.
Might look into a workaround for now with opening workfiles after Maya is launched, cause alternative atm is not using referencing which makes it very hard to update the xgen attributes (for example attributes like the "renderer" is not registered when importing and making xgen delta files).

tokejepsen:
I can confirm that when we open the workfile post initialization of Maya, both the non-critical and critical errors are resolved.

tokejepsen:
For publishing xgen animated guide we can just do pointcaches or productions can make an FX rig (although I havent tested this workflow), of the hair system curves. This aligns with having alembic curves from other DCCs like houdini, so productions can do the guide FX in any DCC.

Loading animation can be done through using an alembic in the `Primitives > Guide Animation` but we have two options for this initially:

1. Absolute path to an alembic path.
2. Live mode where we can create a hair system and connect the curves from FX.

I'm leaning towards option 1 because its cleaner in the outliner and quite possibly less computation since Xgen is reading the alembic data directly rather than going through the hair system.
The best workflow I can think is this:

- Load in xgen from lookdev and pointcache from fx.
- Select both xgen and pointcache in the inventory > Actions > Connect Xgen.

This would set the absolute path in the xgen collection and somehow tag the pointcache container. When the pointcache container gets updated, the tag is queried and the xgen collection path is modified.

@BigRoy @antirotor @LiborBatek what do you think?

tokejepsen:
Running with option 1 cause option 2 is just too much computational overhead. Also option 1 is cleaner code wise.

LiborBatek:
Had tested it maya2022 publish of the Xgen went OK.
image

When loaded the Xgen asset also work fine:
image

Special occasion occurs when user tries to load published Xgen asset into unsaved workfile (untitled  new maya scene) this error shows up:
image

User should probably get sort of Warning that scene/workfile should be saved first (I quess its needed because of xgen folder structure being created and must be in scene context). Now its a bit confusing not informing whats goin on.

Note:
By loading published Xgen asset it is refferenced in the active workfile but its completely standalone and not connected anyhow to asset present in the active workfile . 

I guess its not the way we want it right?

Otherwise user would need to do extra steps manually to connect somehow the Xgen asset to the active character asset which he wants to apply it to. Or?

tokejepsen:
> Special occasion occurs when user tries to load published Xgen asset into unsaved workfile (untitled new maya scene) this error shows up

Good point. Will look into this.

LiborBatek:
@tokejepsen oops, Im afraid not followed testing steps, my bad! Will try better next time, sry!

Will do another go tommorow. Thx

LiborBatek:
Did tests in ```maya2023``` and publishing works ok
image

also loading and action ```connect to geometry``` works fine and Xgen get connected succesfully to animated geo
image

Still need to do animated guide test (action ```connect>Xgen``` )

Note on guides workflow in general : I agree on guide workflow using alembic not live hair dynamics. It should be resolved separately in FX scene or other DCC. 
In general it will be up to user to publish pointcache from selected guide curves and later trigger the connect Xgen action to use this guides pointcache in particular shot. If correctly using OP tools (published assets)  it will be just matter of using load and xgen actions. So no need to implement other workarounds.

tokejepsen:
@antirotor what is the workflow with documentation?
I would expect needing documentation for this feature else the workflow might get lost with this PR. Guessing we do documentation once the PR is merged?

BigRoy:
> Guessing we do documentation once the PR is merged?

What's the amount of work involved for documentation for this? I've seen features get lost many times due to PRs merged and 'docs later' but the docs rarely come.

It would be a good habit I think to see if we can work towards feature PRs to only get merged once docs are implemented.

tokejepsen:
> > Guessing we do documentation once the PR is merged?
> 
> What's the amount of work involved for documentation for this? I've seen features get lost many times due to PRs merged and 'docs later' but the docs rarely come.
> 
> It would be a good habit I think to see if we can work towards feature PRs to only get merged once docs are implemented.

Yeah, docs with feature/enhancement PRs would be my choice as well. @mkolar @antirotor can I work on this?

LiborBatek:
@tokejepsen speaking of closing this PR I guess we still need to resolve last but not least step to succesfuly publish the workfile for rendering with all the xgen data so it can be rendered on Deadline or other rendering sw and not being tight to workfiles/xgen data so it wont get overwritten by possible user interaction on groom/fx workfiles etc.

Am I correct? Or it will be separate PR??

tokejepsen:
> Am I correct? Or it will be separate PR??

Very good point! Will get to work on that now.

LiborBatek:
> > Am I correct? Or it will be separate PR??
> 
> Very good point! Will get to work on that now.

Also me thinks what happens if more than 1 Xgen present in groom/shot scene...we should bullet proof it too! Did you think about it or its for sure it will work too?

tokejepsen:
> Also me thinks what happens if more than 1 Xgen present in groom/shot scene...we should bullet proof it too! Did you think about it or its for sure it will work too?

I've tested with multiple xgens publishing and loading. Also duplicate xgens loaded into the workfile. Should all work so you are welcome to try and break it.

tokejepsen:
@LiborBatek support for rendering now, so I've edited the testing notes to include this.

openpype/hosts/maya/startup/userSetup.py
```diff
@@ -1,16 +1,67 @@
...
+            if source == destination:
+                continue
+
+            shutil.copy(source, destination)
+            cmds.setAttr(
+                palette + "." + attr, destination_basename, type="string"
+            )
+
+
+register_event_callback("save", xgen_on_save)
```
BigRoy, line 62: I feel like this logic shouldn't be directly in `userSetup.py` but instead be somewhere in `pipeline.py` or at least somewhere where it gets triggered through OpenPype host install. No?

Looking at most host implementations the actual 'startup.py' files remain rather minimal and just trigger the host install, which then sets up the rest.

openpype/hooks/pre_add_last_workfile_arg.py
```diff
@@ -40,5 +42,15 @@ def execute(self):
             self.log.info("Current context does not have any workfile yet.")
             return
 
+        # Determine whether to open workfile post initialization.
+        if self.data["app"].host_name == "maya":
```
BigRoy, line 46: I believe this can also be `self.host_name` 

See [here](https://github.com/ynput/OpenPype/blob/96719a39b4719dd6d23b638547825bca88fadc7a/openpype/lib/applications.py#L806-L808) - not sure if that'd be better.

openpype/hooks/pre_add_last_workfile_arg.py
```diff
@@ -40,5 +42,15 @@ def execute(self):
...
+            project_name = self.data["project_name"]
+            settings = get_project_settings(project_name)
```
BigRoy, line 48: No need to query project settings. It's readily available using `self.data["project_settings"]`

openpype/hosts/maya/startup/userSetup.py
```diff
@@ -1,16 +1,67 @@
...
+print("Starting OpenPype usersetup...")
+
+# Open Workfile Post Initialization.
+key = "OPENPYPE_OPEN_WORKFILE_POST_INITIALIZATION"
+if bool(int(os.environ.get(key, "0"))):
+    cmds.evalDeferred(
+        "cmds.file(os.environ[\"AVALON_LAST_WORKFILE\"], open=True,"
+        " force=True)",
+        lowestPriority=True
+    )
```
BigRoy, line 24: Eval deferred is capable of running from an actual function too, for example:
```python
from maya import cmds

def hello_world():
    print('hello')
    
cmds.evalDeferred(hello_world)
```

Thus here we could also do:
```python
from functools import partial
cmds.evalDeferred(partial(cmds.file, 
                          os.environ["AVALON_LAST_WORKFILE"],
                          open=True,
                          force=True),
                  lowestPriority=True)
```

openpype/settings/entities/schemas/projects_schema/schema_project_maya.json
```diff
@@ -5,6 +5,11 @@
     "label": "Maya",
     "is_file": true,
     "children": [
+        {
+            "type": "boolean",
+            "key": "open_workfile_post_initialization",
+            "label": "Open Workfile Post Initialization"
+        },
```
BigRoy, line 12: Should this maybe have a label in the documentation describing when you might want to use this?

website/docs/artist_hosts_maya_xgen.md
```diff
@@ -4,26 +4,92 @@ title: Xgen for Maya
...
+### Settings
 
-### Publishing Xgen Grooms
+Go to project settings > Maya > enable "Open Workfile Post Initialization";
 
-To prepare xgen for publishing just select all the descriptions that should be published together and the create Xgen Subset in the scene using - **OpenPype menu** → **Create**... and select **Xgen Interactive**. Leave Use selection checked.
+`project_settings/maya/open_workfile_post_initialization`
 
-For actual publishing of your groom to go **OpenPype → Publish** and then press ▶ to publish. This will export `.ma` file containing your grooms with any geometries they are attached to and also a baked cache in `.abc` format 
+This is due to two errors occurring when opening workfile contaiing referenced xgen nodes on launch of Maya;
```
BigRoy, line 15:
```suggestion
This is due to two errors occurring when opening workfile containing referenced xgen nodes on launch of Maya, specifically:
```

website/docs/artist_hosts_maya_xgen.md
```diff
@@ -4,26 +4,92 @@ title: Xgen for Maya
...
+```
 
-:::tip adding more descriptions
-You can add multiple xgen description into the subset you are about to publish, simply by
-adding them to the maya set that was created for you. Please make sure that only xgen description nodes are present inside of the set and not the scalp geometry. 
-:::
+### Workfile Incremental Save
 
-### Loading Xgen
+When you increment the Maya workfile to a new version, all `.xgen` and `.xgd` files referenced by the Xgen collection in the workspace is incremented as well.
```
BigRoy, line 32:
```suggestion
When you increment the Maya workfile to a new version, all `.xgen` and `.xgd` files referenced by the Xgen collection in the workspace are incremented as well.
```

BigRoy: Reading this sentence it leaves me wondering what does this means for my assets? Does it need to recache? Why does it need to increment along automatically?

website/docs/artist_hosts_maya_xgen.md
```diff
@@ -4,26 +4,92 @@ title: Xgen for Maya
...
 
-You can use published xgens by loading them using OpenPype Publisher. You can choose to reference or import xgen. We don't have any automatic mesh linking at the moment and it is expected, that groom is published with a scalp, that can then be manually attached to your animated mesh for example. 
+## Create
 
-The alembic representation can be loaded too and it contains the groom converted to curves. Keep in mind that the density of the alembic directly depends on your viewport xgen density at the point of export.
+Create an Xgen instance to publish. This needs to contain 1 Xgen collection only, but validation will check for this:
```
BigRoy, line 36:
```suggestion
Create an Xgen instance to publish. This needs to contain only **one Xgen collection**.
```

I feel there's no need to mention validation will check for it - we only need to explain what the user needs to do.

website/docs/artist_hosts_maya_xgen.md
```diff
@@ -4,26 +4,92 @@ title: Xgen for Maya
...
+- Makes a custom attribute on the Xgen collection, `float_ignore`, which can be seen under the `Expressions` tab of the `Xgen` UI. This is done to initialize the Xgen delta file workflow.
+- Setup an Xgen delta file (`.xgd`) to store any workspace changes of the published Xgen version.
+
+When the loading is done, Xgen collection will be in the Xgen delta file workflow which means any changes done in the Maya workfile will be stored in the current workspace. The published Xgen collection will remain intact, even if the user assigns maps to any attributes or otherwise modifies any attribute.
+
+### Updating
+
+When there are changes to the Xgen version, the user will be notified when opening the workfile or publishing. Since the Xgen is referenced, it follows the standard Maya referencing system and overrides.
+
+For example publishing `xgenMain` version 1 with the attribute `renderer` set to `None`, then version 2 has `renderer` set to `Arnold Renderer`. When updating from version 1 to 2, the `renderer` attribute will be updated to `Arnold Renderer` unless there is an override.
```
BigRoy, line 67:
```suggestion
For example publishing `xgenMain` version 1 with the attribute `renderer` set to `None`, then version 2 has `renderer` set to `Arnold Renderer`. When updating from version 1 to 2, the `renderer` attribute will be updated to `Arnold Renderer` unless there is a local override.
```

website/docs/artist_hosts_maya_xgen.md
```diff
@@ -4,26 +4,92 @@ title: Xgen for Maya
...
+
+The animation container gets connected with the Xgen container, so if the animation container is updated so will the Xgen container's attribute.
+
+## Rendering
+
+To render with Xgen, follow the [Rendering With OpenPype](artist_hosts_maya#rendering-with-openpype) guide.
+
+### Details
+
+When submitting a workfile with Xgen, all Xgen related files will be collected and published as the workfiles resources. This means the published workfile is no longer referencing the workspace Xgen files.
```
BigRoy, line 95: Is this always true? or only when use published file for rendering is enabled in the openpype settings?

BigRoy (commented):
Some notes. Very nice work on the documentation!

openpype/settings/entities/schemas/projects_schema/schema_project_maya.json
tokejepsen: Feel like this is generic enough to be use for other use cases, so I would not mention Xgen here specifically. But we should elaborate on what the setting does. Good point!

BigRoy: I meant more like - what's the reason the feature would not be enabled by default? Is there any? etc.

website/docs/artist_hosts_maya_xgen.md
tokejepsen: Yeah, I'm not sure this feature is actually needed in the end, or it needs to at least be configurable. There is no reason other than keeping the work files and area tidy.
The thinking was when publishing and the workfile gets incremented, we are still using the old versions xgen files so it would be hard (impossible?) to go back to an older workfile straight from the Workfile app.

The feature is actually also lacking because its only incrementing the `.xgen` and `.xgd` files, and not the side xgen folder structure containing `.ptx` and `.xuv`.

Question is whether this feature is worth finishing or abandon? @antirotor @mkolar @BigRoy

website/docs/artist_hosts_maya_xgen.md
tokejepsen: Hmm, this happens when there is an active renderlayer present in the context, but dont think it'll check for whether the rendering process is using the published file or not. Will investigate.

openpype/settings/entities/schemas/projects_schema/schema_project_maya.json
tokejepsen: Consulted with @antirotor about this and we cant think of a reason why it would be bad to enable this by default. We are just scared to change just a global setting...

LiborBatek (changes requested):
Edited: Sorry my setting didnt have turned on ```Open Workfile Post Init``` option ```On```

image

So now this works normally... 

problem arise when opening older workfile of the same shot (before it has just 1 Xgen asset + according char anim cache loaded)
```
Importing XGen Collections...
Done.
XgenLoader
ReferenceLoader
ReferenceLoader
# Warning: openpype.hosts.maya : Scene has outdated content.
// Error: XGen:  Failed to open file: D:/Projects/Libor_Tests/EP/sh004/work/lighting/resources/220901_sh004_lighting_v001__rayman_xgenMain_01__Hair_COL.xgen 
// Error: XGen:  Failed to import collection from file D:/Projects/Libor_Tests/EP/sh004/work/lighting/resources/220901_sh004_lighting_v001__rayman_xgenMain_01__Hair_COL.xgen 
```


also not sure when checking the folder files...its a bit weird to have several xgen files but with almost identical names (differences are with ```__``` in the names)

image

Is this ok? Seems somehow multiplied/duplicated, not sure.

LiborBatek:
First Note: The full procedure using Xgen worked at first instance.
 means... I did the whole process by publish groom asset > loading the groom and use actions to get connected to the animation cache of the character > publish render. It all worked fine. However It counts just for single xgen/char asset in the scene and no older versions existed (not any localized xgen folder/files) tho. But all the testing steps were functional and working ok.

But when I tried to
1) create new groom asset (publish new version)
2) multiply the amount of characters to 2
3)load the groom/update the groom to newest
4)publish the render

Two things happen:

The render finnished but without matching grooms (older groom ver applied on both characters) and when opening the workfile for the lighting scene it fails to load the Xgen assets giving those errors seen above.

Also its a bit weird that the work folder contains duplicate xgen files with slight name difference.

openpype/hooks/pre_add_last_workfile_arg.py
tokejepsen: Good point!

openpype/hooks/pre_add_last_workfile_arg.py
tokejepsen: Thanks!

openpype/hosts/maya/startup/userSetup.py
tokejepsen: > Also, I think there's a bug here @tokejepsen
> 
> If you disable launching last workfiles automatically globally but enable `maya/open_workfile_post_initialization` then it'd still try to read the `AVALON_LAST_WORKFILE` which I think might not exist - and even if it does.. it shouldn't load the file automatically because the pipeline had it disabled.

The logic for whether to open the last workfile is handled in [here](https://github.com/tokejepsen/pype/blob/feature/OP-3074_xgen/openpype/hooks/pre_add_last_workfile_arg.py#L36), so the enviroment variable `OPENPYPE_OPEN_WORKFILE_POST_INITIALIZATION` is never set and no workfile is launched.
Should this be more explicit and printed/logged?

BigRoy: I see, you're correct :) As long as it works I'm fine with it.

website/docs/artist_hosts_maya_xgen.md
tokejepsen: Will account for not using the published file now.

openpype/hosts/maya/startup/userSetup.py
tokejepsen: Have removed the `xgen_on_save` function now. Do you still think we should move the `OPENPYPE_OPEN_WORKFILE_POST_INITIALIZATION` code?

openpype/hosts/maya/api/plugin.py
```diff
@@ -300,6 +300,37 @@ def update(self, container, representation):
                      str(representation["_id"]),
                      type="string")
 
+        # Update any xgen containers.
+        compound_name = "xgenContainers"
+        object = "SplinePrimitive"
```
BigRoy, line 305: It seems the `object` "SplinePrimitive" isn't used until the `xgenm.setAttr` call. How about just putting that string directly into the function call?

LiborBatek:
did another go with testing...with fresh new shot (loaded animation + xgen asset and connected to geometry action triggered, ready for rendering) but...

when trying to publish for rendering (to deadline) from lighting workfile it gives me this error when extracting xgen workfiles:

image

tokejepsen:
> 

Nice catch. I'll amend the assumption on deadline.
What render farm are you using?

LiborBatek:
I did test with switching Xgen asset versions and was using only new published versions (originating from the same code branch) and it seems ok.

Asset got updated and is possible to switch versions back/forward with no issues. 

Will continue on more testing cases like multiple Xgens and again switching assets versions...

so far so good

LiborBatek:
After another round of tests, the core functionality is working!

From Xgen asset groom publishing to final rendered images / publish. Also tested on multiple instances of Xgen asset present in one maya workfile, switching versions and mixing different versions on different Xgen assets.

All works fine.
image

LiborBatek:
The probably last missing feature is rendering with ```motion blur``` 

Its neccessary to implement ```Export patches for batch render``` (basically alembic cache of the groom mesh animation) and guides. This could probably happen when publishing for render as a last step so the published workfile for render contains those connected and the Xgen render mode set to ```batch render``` not ```Live```.

tokejepsen:
@LiborBatek could you test this workaround out for motion blur?

- disable `project_settings/deadline/publish/MayaSubmitDeadline/use_published`
- publish xgen
- load xgen
- connect xgen to animation
- publish for rendering on Deadline

If it works for you, I'll document the workaround and finish up the updating code.

LiborBatek (approved):

LiborBatek:
Following the testing steps with workaround in OP Settings worked OK. All being rendered and fine.
image

image

here is also render log from Deadline after succesfully rendered frame 

image

tokejepsen:
@LiborBatek this is ready for a test run through. Anything you can find before we push this to the client.

LiborBatek:
Its all tested and working fine. Also finally used guided sim so nothing left for testing and can be merged.

All is OK.

tokejepsen:
@BigRoy @antirotor last comments before merging?

BigRoy:
> last comments before merging?

Are you really sure you want to be asking me this? Just before a merge! Let me get all over this bad boy!

openpype/hosts/maya/api/lib.py
```diff
@@ -3446,3 +3446,17 @@ def dgcontext(mtime):
...
+    with open(filepath, "r") as f:
+        for line in [line.rstrip() for line in f]:
+            for key, value in data.items():
+                if line.startswith("\t{}".format(key)):
+                    line = "\t{}\t\t{}".format(key, value)
+
+            lines.append(line)
+
+    with open(filepath, "w") as f:
+        f.write("\n".join(lines))
```
BigRoy, line 3462: Can we add a docstring to describe the usage of this `lib` function?

openpype/hosts/maya/plugins/publish/reset_xgen_attributes.py
```diff
@@ -0,0 +1,27 @@
...
+                self.log.info(
+                    "Setting \"{}\" on \"{}\"".format(value, node_attr)
+                )
+                cmds.setAttr(node_attr, value, type="string")
+
+            cmds.setAttr(palette + "." + "xgExportAsDelta", True)
+
+        if instance.data.get("xgenAttributes", {}):
+            self.log.info("Saving changes.")
+            cmds.file(save=True)
```
BigRoy, line 27:
```suggestion
        xgen_attributes = instance.data.get("xgenAttributes", {})
        if not xgen_attributes :
            return

        for palette, data in xgen_attributes.items():
            for attr, value in data.items():
                node_attr = "{}.{}".format(palette, attr)
                self.log.info(
                    "Setting \"{}\" on \"{}\"".format(value, node_attr)
                )
                cmds.setAttr(node_attr, value, type="string")
            cmds.setAttr(palette + ".xgExportAsDelta", True)
            
        self.log.info("Saving changes.")
        cmds.file(save=True)
```

Opt out earlier.

Also, we're changing the scene with the `xgExportAsDelta` attribute for enabled here? are we sure we want to leave the user's workfile in a changed state after publishing? This should be avoided unless there's a hard requirement why this needs to be the case and will ALWAYS be what the user wants for their current workfile.

openpype/hosts/maya/plugins/publish/validate_xgen.py
```diff
@@ -0,0 +1,59 @@
...
+                " Found:\n{}".format(set_members)
+            )
+
+        # Only xgen palette node is allowed.
+        node_type = cmds.nodeType(set_members[0])
+        if node_type != "xgmPalette":
+            raise KnownPublishError(
+                "Only node of type \"xgmPalette\" are allowed. Referred to as"
+                " \"collection\" in the Maya UI."
+                " Node type found: {}".format(node_type)
```
BigRoy, line 34: I believe these should be `KnownValidationError` or `PublishValidationError` or something.

openpype/hosts/maya/plugins/publish/validate_xgen.py
```diff
@@ -0,0 +1,59 @@
...
+                    except KeyError:
+                        inactive_modifiers[description] = [name]
+
+        msg = (
+            "There are inactive modifiers on the collection. "
+            "Please delete these:\n{}".format(
+                json.dumps(inactive_modifiers, indent=4, sort_keys=True)
+            )
+        )
+        assert not inactive_modifiers, msg
```
BigRoy, line 59: Do not assert, but raise the relevant `PublishValidationError`

openpype/hosts/maya/plugins/load/actions.py
```diff
@@ -93,7 +93,20 @@ class ImportMayaLoader(load.LoaderPlugin):
...
+        "animation",
+        "mayaAscii",
+        "mayaScene",
+        "setdress",
+        "layout",
+        "camera",
+        "rig",
+        "camerarig",
+        "staticMesh"
+    ]
```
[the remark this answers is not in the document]
tokejepsen: Yeah, we export the geometry patches (meshes) and xgen collection to Maya scene. Having the `import` option on this, does not have sense because you'll need to do extra setup to make it work.

openpype/hosts/maya/plugins/load/load_xgen.py
```diff
@@ -0,0 +1,174 @@
...
+
+        maya_file = get_representation_path(representation)
+        _, extension = os.path.splitext(maya_file)
+        new_xgen_file = maya_file.replace(extension, ".xgen")
+        data_path = ""
+        with open(new_xgen_file, "r") as f:
+            for line in [line.rstrip() for line in f]:
+                if line.startswith("\txgDataPath"):
+                    data_path = line.split("\t")[-1]
+                    break
```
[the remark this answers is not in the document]
tokejepsen: There are actually two instances of `xgDataPath` that needs updating.

openpype/hosts/maya/plugins/publish/extract_xgen.py
```diff
@@ -0,0 +1,158 @@
...
+            duplicate_transform = cmds.duplicate(transform_name)[0]
+
+            # Discard the children.
+            shapes = cmds.listRelatives(duplicate_transform, shapes=True)
+            children = cmds.listRelatives(duplicate_transform, children=True)
+            cmds.delete(set(children) - set(shapes))
+
+            duplicate_transform = cmds.parent(
+                duplicate_transform, world=True
+            )[0]
```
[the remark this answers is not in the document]
tokejepsen: That is true, it wont retain the input connections. But I would argue that you would not want to have it animated since you'll be later connecting it to animation caches.
We could validate against this, but think that is overkill for now.

openpype/hosts/maya/plugins/load/load_xgen.py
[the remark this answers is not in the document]
BigRoy: I think you're referring to a different part of the code with your comment. Nothing is being updated here in the loop at at all, so even without the break it's not updating both instances.

openpype/hosts/maya/plugins/publish/reset_xgen_attributes.py
tokejepsen: Maya/Xgen changes this attribute when the incremental save triggers. We are changing it back to what it was before publishing. Found that the scene was not marked as modified either, so the user could end up saving the scene then on launching again, the attributes would be different.
Will comment on this though for prosperity.

openpype/hosts/maya/api/lib.py
```diff
@@ -3446,3 +3446,17 @@ def dgcontext(mtime):
...
     return connections[0] if connections else None
+
+
+def write_xgen_file(data, filepath):
+    lines = []
+    with open(filepath, "r") as f:
+        for line in [line.rstrip() for line in f]:
+            for key, value in data.items():
+                if line.startswith("\t{}".format(key)):
+                    line = "\t{}\t\t{}".format(key, value)
```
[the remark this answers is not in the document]
tokejepsen: There are two instances of `xgDataPath` and `xgProjectPath` which needs changing. Ideally we should be probably parsing the Xgen file schema, but time does not allow for this.

openpype/hosts/maya/api/lib.py
tokejepsen: Good point!

openpype/hosts/maya/api/lib.py
```diff
@@ -3446,3 +3446,17 @@ def dgcontext(mtime):
...
     connections = cmds.listConnections(attr, plugs=True, destination=False)
     return connections[0] if connections else None
+
+
+def write_xgen_file(data, filepath):
+    lines = []
+    with open(filepath, "r") as f:
+        for line in [line.rstrip() for line in f]:
+            for key, value in data.items():
+                if line.startswith("\t{}".format(key)):
```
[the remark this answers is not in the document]
tokejepsen: Good point! Trailing with `\t\t` before the value so adding that.

openpype/hosts/maya/api/plugin.py
```diff
@@ -300,6 +300,36 @@ def update(self, container, representation):
...
+                            path.replace("\\", "/"),
+                            palette,
+                            description,
+                            "SplinePrimitive"
+                        )
+
+            # Refresh UI and viewport.
+            de = xgenm.xgGlobal.DescriptionEditor
+            de.refresh("Full")
+
```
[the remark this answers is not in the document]
tokejepsen: Good point!

openpype/hosts/maya/api/lib.py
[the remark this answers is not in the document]
BigRoy: Ok, if there are more instances to change - I'd recommend this:
```suggestion
        for line in f:
            for key, value in data.items():
                if line.startswith("\t{}".format(key)):
                    line = "\t{}\t\t{}".format(key, value)
```

I couldn't find any logic why the lines needed to have `.rstrip()` called on them since it didn't seem you did anything with the end of the line to begin with. Unless you explicitly didn't want to write those end line values back into the file for a reason?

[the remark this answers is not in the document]
BigRoy: Actually, NO! My previous suggestion was accurate. There might be more instances in the file, but the same line would never start with multiple keys - no? :) 

So it should still break out as soon as it did find a match and resolved it.

openpype/hosts/maya/plugins/inventory/connect_geometry.py
```diff
@@ -0,0 +1,139 @@
...
+            target = match[1]
+            blendshape = cmds.blendShape(source, target)[0]
+            cmds.setAttr(blendshape + ".origin", 0)
+            cmds.setAttr(blendshape + "." + target.split(":")[-1], 1)
+
+        # Update Xgen if in any of the containers.
+        if "xgmPalette" in node_types:
+            cmds.xgmPreview()
+
+    def get_container_data(self, container):
```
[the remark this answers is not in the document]
tokejepsen: Sure!

openpype/hosts/maya/plugins/inventory/connect_geometry.py
```diff
@@ -0,0 +1,139 @@
...
+        data = {"node_types": [], "ids": {}}
+        ref_node = cmds.sets(container, query=True, nodesOnly=True)[0]
+        for node in cmds.referenceQuery(ref_node, nodes=True):
+            node_type = cmds.nodeType(node)
+            data["node_types"].append(node_type)
+            if node_type == "mesh":
+                transform = cmds.listRelatives(node, parent=True)[0]
+                id = cmds.getAttr(transform + ".cbId")
+                data["ids"][id] = transform
```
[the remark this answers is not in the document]
tokejepsen: Cant rely on shape ids because they can change upstream due to deformation, but the transform id should match.

openpype/hosts/maya/plugins/inventory/connect_xgen.py
```diff
@@ -0,0 +1,168 @@
...
+
+        state = QtWidgets.QMessageBox.warning(
+            None,
+            "",
+            message,
+            buttons=buttons,
+            defaultButton=accept
+        )
+
+        return state == accept
```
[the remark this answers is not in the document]
tokejepsen: Not always;
https://github.com/tokejepsen/pype/blob/feature/OP-3074_xgen/openpype/hosts/maya/plugins/inventory/connect_xgen.py#L23
https://github.com/tokejepsen/pype/blob/feature/OP-3074_xgen/openpype/hosts/maya/plugins/inventory/connect_xgen.py#L98

openpype/hosts/maya/plugins/publish/extract_workfile_xgen.py
```diff
@@ -0,0 +1,217 @@
...
+
+import pyblish.api
+from openpype.hosts.maya.api import current_file
+from openpype.hosts.maya.api.lib import extract_alembic
+from openpype.pipeline import publish
+from openpype.lib import StringTemplate
+
+
+class ExtractWorkfileXgen(publish.Extractor):
+    """Extract Workfile Xgen."""
```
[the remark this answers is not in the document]
tokejepsen: Sure

openpype/hosts/maya/plugins/publish/extract_workfile_xgen.py
```diff
@@ -0,0 +1,217 @@
...
+            return
+
+        # We decrement start frame and increment end frame so motion blur will
+        # render correctly.
+        start_frame -= 1
+        end_frame += 1
+
+        # Extract patches alembic.
+        basename, _ = os.path.splitext(current_file())
+        dirname = os.path.dirname(current_file())
```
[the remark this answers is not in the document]
tokejepsen: Always forget about that collection. Thanks!

openpype/hosts/maya/plugins/publish/extract_workfile_xgen.py
```diff
@@ -0,0 +1,217 @@
...
+                    end_frame = i.data["frameEnd"]
+
+                break
+
+        if not renderlayer:
+            self.log.debug(
+                "No publishable renderlayers found in context. Abort Xgen"
+                " extraction."
+            )
+            return
```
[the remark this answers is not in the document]
BigRoy: Reading the logged message - maybe rephrase `abort xgen extraction` to `skipping xgen extraction` so it sounds less like an issue?

openpype/hosts/maya/plugins/inventory/connect_geometry.py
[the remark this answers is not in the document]
BigRoy: > Cant rely on shape ids because they can change upstream due to deformation, but the transform id should match.

There should be validators in place that shape ids should NOT change when publishing animation/pointcache if the shape id was different than on the 'intermediateMesh' in its history. See Validate Animation Out Set Node Ids.

I think we should either refactor it so that we NEVER rely on shape ids to begin with (if we feel that's problematically unsafe) or ALWAYS rely on it if we're doing it in other areas. OR we should document clearly why this particular case is a special case which makes the shape ids extra unsafe than e.g. when they are used for look assignments (which also rely on shape ids for the shader assignment by the way.)

openpype/hosts/maya/api/lib.py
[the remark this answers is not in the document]
BigRoy: By the way, if that's the case - we could rewrite this to something like this:

```python
def write_xgen_file(data, filepath):

    # Generate regex lookup for line to key basically
    # match any of the keys in `\t{key}\t\t`
    keys = "|".join(re.escape(key) for key in data.keys())
    re_keys = re.compile("^\t({})\t\t".format(keys))
    
    lines = []
    with open(filepath, "r") as f:
        for line in f:
            match = re_keys.match(line)
            if match:
                key = match.group(1)
                value = data[key]
                line = "\t{}\t\t{}".format(key, value)

            lines.append(line)

    with open(filepath, "w") as f:
        f.write("\n".join(lines))
```

This would be much faster if the file being iterated tends to exist of many lines and the amount of keys in data isn't super small.

openpype/hosts/maya/plugins/publish/extract_xgen.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def process(self, instance):
+        if "representations" not in instance.data:
+            instance.data["representations"] = []
+
+        staging_dir = self.staging_dir(instance)
+        maya_filename = "{}.{}".format(instance.data["name"], self.scene_type)
+        maya_filepath = os.path.join(staging_dir, maya_filename)
+
+        # Collect nodes to export.
```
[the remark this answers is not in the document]
tokejepsen: How would you clean up a failed scene change, if you dont know what nodes/files to clean up?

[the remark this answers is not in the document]
BigRoy: In some case it worked magically if I did:

```python
with undo_chunk():
    # Do all your changes here

maya.cmds.undo()
```
Hehe. But I'd say that if you do complex things that might be unsafe - however, if all commands you are doing are undoable I wouldn't know why this would fail.

Other ways would be to e.g. use the [delete_after](https://github.com/ynput/OpenPype/blob/2d574996c38e7990eb66c1dd4a98d8be30f5703c/openpype/hosts/maya/api/lib.py#L639-L676) context manager where you add the nodes you are generating as soon as possible after you've generated them so that IF an errors gets raised at least those nodes would get cleaned up.

It seems that in this case the `delete_after` workflow might be possible for your code?

openpype/hosts/maya/plugins/publish/validate_xgen.py
tokejepsen: Got `PublishValidationError` and `KnownPublishError`. @iLLiCiTiT got any insights on this?

openpype/hosts/maya/plugins/publish/validate_xgen.py
tokejepsen: Nice catch!

openpype/hosts/maya/plugins/publish/validate_xgen.py
iLLiCiTiT: `PublishValidationError` is for validation errors which can describe to user what happened, where and gives him option to run fix actions. `KnownPublishError` is for cases when publishing crashes but user can't do much about it.

openpype/hosts/maya/api/lib.py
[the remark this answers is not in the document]
tokejepsen: We should not break out cause the lines are read to be, potentially modified and, written back in?

openpype/hosts/maya/plugins/publish/extract_workfile_xgen.py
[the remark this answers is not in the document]
tokejepsen: Sure, can do

openpype/hosts/maya/api/lib.py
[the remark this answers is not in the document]
BigRoy: We're not breaking out of the outer for loop, just out of the inner for loop where we check each key in data. As soon as we've found one we've already defined the `line =` value and thus can break back into the outer loop which appends it into the `lines`

openpype/hosts/maya/plugins/inventory/connect_geometry.py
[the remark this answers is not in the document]
tokejepsen: Not sure what you want out of this. I need the name of the mesh transform from the container to compare but also to plug into the blendshape. How we archieve this, does not matter to me so up for any suggestions.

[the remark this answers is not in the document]
BigRoy: If this works for the use case it's fine. It just felt odd to me that somehow we're linking meshes but relying on transforms.

Anyway, would it be nicer to use `lib.get_id(transform)` just so we use the implementation that we might want to change up later?

openpype/hosts/maya/api/lib.py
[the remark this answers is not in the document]
tokejepsen: Cool, thanks. Will ingest and test

openpype/hosts/maya/plugins/inventory/connect_geometry.py
```diff
@@ -0,0 +1,139 @@
...
+        target_containers = []
+        for family, containers in containers_by_family.items():
+            if family in ["animation", "pointcache"]:
+                continue
+
+            target_containers.extend(containers)
+
+        source_data = self.get_container_data(source_object)
+        matches = []
+        node_types = []
```
BigRoy, line 67: `node_types` should really be a `set`. We're only using it to do a lookup later to see whether `xgmPalette` is one of the nodes types.

The same goes for `nodes_types` in the `get_container_data` result.

openpype/hosts/maya/plugins/load/actions.py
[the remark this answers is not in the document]
tokejepsen: True, we could but time is running out on this PR so cant wait on other PRs.

openpype/hosts/maya/plugins/publish/validate_xgen.py
tokejepsen: Thanks @iLLiCiTiT

openpype/hosts/maya/api/lib.py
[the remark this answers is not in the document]
tokejepsen: Cool, redundant now with this?
https://github.com/ynput/OpenPype/pull/4256#discussion_r1091851312

openpype/hosts/maya/plugins/inventory/connect_geometry.py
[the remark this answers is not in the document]
tokejepsen: Cool, have also documented why we are doing this as well.

openpype/hosts/maya/api/lib.py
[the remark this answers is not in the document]
BigRoy: I'd say so, yes - because it removes the for loop used here.

openpype/hosts/maya/plugins/inventory/connect_geometry.py
tokejepsen: Sure, can do

openpype/hosts/maya/plugins/publish/reset_xgen_attributes.py
```diff
@@ -24,6 +28,9 @@ def process(self, instance):
                 )
                 cmds.setAttr(node_attr, value, type="string")
             cmds.setAttr(palette + ".xgExportAsDelta", True)
-            
+
+        # Need to save the scene, cause the attribute changes above does not
+        # mark the scene as modified so user can exit without commiting the
+        # changes.
```
BigRoy, line 34: We could technically also set the scene as modified explicitly `cmds.file(modified=True)` or something like that. Not sure if that's what you'd want because you could still force quit the app so likely saving is what you want. The comment here however seemed to hint that if we were able to reset the modified state we'd be ok too.

openpype/hosts/maya/plugins/publish/extract_xgen.py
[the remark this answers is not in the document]
tokejepsen: Thanks for pointing me to `delete_after`. Will use that.

openpype/hosts/maya/plugins/publish/reset_xgen_attributes.py
BigRoy: Yes, I understand. Would be even nicer if the incremental save wouldn't end up changing the attributes to begin with. But hey, Xgen does its own. Ugh...

tokejepsen: Ugh indeed!

tokejepsen:
@BigRoy think I've addressed all your feedback now.

openpype/hosts/maya/plugins/publish/extract_workfile_xgen.py
```diff
@@ -0,0 +1,250 @@
...
+        transfers = []
+
+        # Validate there is any palettes in the scene.
+        if not cmds.ls(type="xgmPalette"):
+            self.log.debug(
+                "No collections found in the scene. Skipping Xgen extraction."
+            )
+            return
+        else:
+            import xgenm
```
BigRoy, line 87: I don't feel like this really needs to be in an `else` statement since we return in the if statement anyway. 
We could just do this import below the `get_render_max_frame_range` call below, no?

openpype/hosts/maya/plugins/publish/extract_workfile_xgen.py
```diff
@@ -0,0 +1,250 @@
...
+        template_data = copy.deepcopy(instance.data["anatomyData"])
+        published_maya_path = StringTemplate(
+            instance.context.data["anatomy"].templates["publish"]["file"]
+        ).format(template_data)
+        published_basename, _ = os.path.splitext(published_maya_path)
+
+        for source in alembic_files:
+            destination = os.path.join(
+                os.path.dirname(instance.data["resourcesDir"]),
+                os.path.basename(source.replace(basename, published_basename))
```
BigRoy, line 141: Reading this made me think "so wait, `basename` and `publish_basename` themselves aren't actually basenames"? They didn't go through `os.path.basename` but are the `root` of `os.path.splitext`. Do we have naming confusion here?

See this example of `os.path.splitext`:
```python
splitext('/foo/bar.exe')
# ('/foo/bar', '.exe')
```

openpype/hosts/maya/plugins/publish/extract_xgen.py
```diff
@@ -0,0 +1,142 @@
...
+                # Duplicate_transform subd patch geometry.
+                duplicate_transform = cmds.duplicate(transform_name)[0]
+                delete_bin.append(duplicate_transform)
+
+                # Discard the children.
+                shapes = cmds.listRelatives(duplicate_transform, shapes=True)
+                children = cmds.listRelatives(
+                    duplicate_transform, children=True
+                )
+                cmds.delete(set(children) - set(shapes))
```
BigRoy, line 66: Note: Older versions of maya don't support directly operating on `set` types. We did a bugfix some time ago for some other OP code.

It's Maya 2018 and below by the way, not sure if still relevant?

See PR: https://github.com/ynput/OpenPype/pull/2932

website/docs/artist_hosts_maya_xgen.md
```diff
@@ -4,26 +4,94 @@ title: Xgen for Maya
 sidebar_label: Xgen
 ---
 
-## Working with Xgen in OpenPype
+## Setup
 
-OpenPype support publishing and loading of Xgen interactive grooms. You can publish 
-them as mayaAscii files with scalps that can be loaded into another maya scene, or as
-alembic caches.
```
BigRoy, line 11: The new documentation gets technical rather quickly - should we maybe, like before have a few lines of 'introduction' of _what_ the Xgen functionality in OpenPype is intended to do. Not something long, but at least an artist looking at "hey what's supported" could find it in a quick read instead of first having to read about how they need to customize project settings.

BigRoy (commented):
Some minor notes.

openpype/hosts/maya/plugins/publish/extract_xgen.py
maxpareschi: Nope gone to 2022 and handed off grooming to Houdini sorry

openpype/hosts/maya/plugins/publish/extract_workfile_xgen.py
tokejepsen: yeah, good point. Going outside the if statement.

openpype/hosts/maya/plugins/publish/extract_workfile_xgen.py
tokejepsen: Have you got a suggestions of names/code?

openpype/hosts/maya/plugins/publish/extract_workfile_xgen.py
```diff
@@ -0,0 +1,250 @@
...
+        alembic_files = []
+        for palette in cmds.ls(type="xgmPalette"):
+            patch_names = []
+            for description in xgenm.descriptions(palette):
+                for name in xgenm.boundGeometry(palette, description):
+                    patch_names.append(name)
+
+            alembic_file = os.path.join(
+                dirname,
+                "{}__{}.abc".format(basename, palette.replace(":", "__ns__"))
```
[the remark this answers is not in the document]
tokejepsen: Its the `basename` of the current workfile, which needs to be included in the alembic file name.

openpype/hosts/maya/plugins/publish/extract_xgen.py
tokejepsen: Cool, gunna leave this.

website/docs/artist_hosts_maya_xgen.md
tokejepsen: Sure, will add a little intro.

openpype/hosts/maya/plugins/publish/extract_workfile_xgen.py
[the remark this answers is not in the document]
BigRoy: So `instance.context.data["currentFile"])` is the full path to the current file as far as I know: `C:/path/to/maya/scene.ma`
Doing `basename, _ = os.path.splitext(current_file)` on it makes `basename` `C:/path/to/maya/scene`.

As such we appear to be joining that absolute path onto the dirname here?

I think looking at the code you intended to originally call `os.path.basename` as well as `os.path.splitext` on the current file?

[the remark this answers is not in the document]
tokejepsen: Ahh I see, yeah you are right.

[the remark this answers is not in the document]
BigRoy: So it is intentional that you're joining the full path onto that dirname generating a path like:
```
C:/path/to/dirname/C:/path/to/maya/currentfile_{palette}.abc
```

Because:
```python
            alembic_file = os.path.join(
                dirname,
                "{}__{}.abc".format(
                    path_no_ext, palette.replace(":", "__ns__")
                )
            )
```
Because:
```python
dirname = "C:/path/to/current"
path_no_ext = "C:/path/to/current/scene"
```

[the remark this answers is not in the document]
tokejepsen: yeah, good point. Simplified now.

openpype/hosts/maya/api/plugin.py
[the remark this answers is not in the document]
tokejepsen: Sorry, I was too quick on this. The reason why its in `plugin.py` is because its being run on animation containers (`animation` or `pointcache`). I've update the comment to better explain the situation.

mkolar:
Gentlemen I believe this has gone through enough review torment now, and is ready for merge.

PR #4261, opened 2022-12-22
maya gltf texture convertor and validator

moonyuet:
## Brief description
Continuity of the gltf extractor implementation

## Description
Continuity of the gltf extractor https://github.com/pypeclub/OpenPype/pull/4192
UPDATE:
**Validator for GLSL Shader**:  Validate whether the mesh uses GLSL Shader. If not it will error out. The user can choose to perform the repair action and it will help to assign glsl shader. 
If the mesh with Stringray PBS, the repair action will also check to see if there is any linked texture such as Color, Occulsion, and Normal Map. If yes, it will help to relink the related textures to the glsl shader.

*****If the mesh uses the PBS Shader, 

## Additional info
N/A

## Testing notes:
1. Enable the extractor in Openpype setting
2. Create instance which is associated to the extractor(model, animation, point cache)
3. Assign the PBS Shaders with correct texture names into the asset
4. Validate it
5. If you are not using glsl shader, the validator will error out
6. click the repair function
7. it will assign a new glsl shader
8. if the material is stringray pbs, it will automatically relink all your connected texture to the newly assigned glsl shader.
9. Publish it
10. You will find the glb file in the folder(s) of the related instance in the publish folder
11. Test the glb file in https://sandbox.babylonjs.com/ and you will find the textured asset.
***** if you turn off the validator, you can still export but no guarantee that the output will be correct.


antirotor (changes requested):
Looks good, we just need to check the validator.

moonyuet:
> Looks good, we just need to check the validator.

I have created another PR for this branch mainly because I can't push the latest code within this PR for some reasons.
https://github.com/ynput/OpenPype/pull/4327

antirotor:
Closing this in favour of #4327 4327

openpype/hosts/maya/plugins/publish/convert_gltf_shader.py
```diff
@@ -0,0 +1,91 @@
...
+
+            # load the maya2gltf shader
+            maya_publish = instance.context.data["project_settings"]["maya"]["publish"] # noqa
+
+            ogsfx_path = maya_publish["ConvertGLSLShader"]["ogsfx_path"]
+            if not ogsfx_path:
+                maya_dir = os.getenv("MAYA_APP_DIR")
+                if not maya_dir:
+                    raise RuntimeError("MAYA_APP_DIR not found")
+                ogsfx_path = maya_dir + "/maya2glTF/PBR/shaders/"
```
BigRoy, line 42: How about not adding this extra logic but just setting the default settings path to `{MAYA_APP_DIR}/maya2glTF/PBR/shaders/`?

openpype/hosts/maya/plugins/publish/convert_gltf_shader.py
```diff
@@ -0,0 +1,91 @@
...
+import pyblish.api
+
+from openpype.pipeline import publish
+
+
+class ConvertGLSLShader(publish.Extractor):
+    """
+    Converting StingrayPBS material to GLSL Shaders
+    specially for the glb export through Maya2GLTF plugin
+    """
```
BigRoy, line 12: I've brought this up before - but this **should not** be done. A plug-in should not change the local scene and leave it in the changed state.

Publishing should be idempotent so that publishing the scene again would do the same thing as before.

If really required to change the scene for the user and leave the scene changed this should be an explicit repair action on a validator instead.

What if the user were to publish both a Look and a GLTF from the same scene? You'd unexpected results for the look publishing with this plug-in.

openpype/hosts/maya/plugins/publish/validate_gltf_textures_names.py
```diff
@@ -0,0 +1,118 @@
...
+class ValidateGLTFTexturesNames(pyblish.api.InstancePlugin):
+    """
+    Validate if the asset uses StingrayPBS material before conversion
+    of the GLSL Shader
+    Validate if the names of GLTF Textures follow
+    the packed ORM/ORMS standard.
+    The texture naming conventions follows the UE5-style-guides:
+    https://github.com/Allar/ue5-style-guide#anc-textures-packing
+    ORM: Occulsion Roughness Metallic
+    ORMS: Occulsion Roughness Metallic Specular
```
BigRoy, line 16:
```suggestion
    ORM: Occlusion Roughness Metallic
    ORMS: Occlusion Roughness Metallic Specular
```

openpype/hosts/maya/plugins/publish/validate_gltf_textures_names.py
```diff
@@ -0,0 +1,118 @@
...
+    Texture Naming Style:
+    Albedo/Diffuse: {Name}_D.{imageExtension} or
```
BigRoy, line 18: Is there a reason the `{Name}` key is uppercase?

Also, I feel `{imageExtension}` is overly verbose and `{ext}` or `{extension}` would do just as fine.

openpype/hosts/maya/plugins/publish/convert_gltf_shader.py
```diff
@@ -32,10 +32,12 @@ def process(self, instance):
...
-                raise RuntimeError("MAYA_APP_DIR not found")
-            ogsfx_path = maya_dir + "/maya2glTF/PBR/shaders/"
+            maya_publish = instance.context.data["project_settings"]["maya"]["publish"] # noqa
+
+            ogsfx_path = maya_publish["ConvertGLSLShader"]["ogsfx_path"]
+
+            if not os.path.exists(ogsfx_path):
+                raise RuntimeError("the ogsfx file not found")
 
             ogsfx = ogsfx_path + "glTF_PBR.ogsfx"
```
BigRoy, line 42: Wouldn't it make more sense if the `ogsfx_path` in settings would point to the actual file itself and NOT the folder?

openpype/hosts/maya/plugins/publish/convert_gltf_shader.py
```diff
@@ -32,10 +32,12 @@ def process(self, instance):
...
-            maya_dir = os.getenv("MAYA_APP_DIR")
-            if not maya_dir:
-                raise RuntimeError("MAYA_APP_DIR not found")
-            ogsfx_path = maya_dir + "/maya2glTF/PBR/shaders/"
+            maya_publish = instance.context.data["project_settings"]["maya"]["publish"] # noqa
+
+            ogsfx_path = maya_publish["ConvertGLSLShader"]["ogsfx_path"]
+
+            if not os.path.exists(ogsfx_path):
+                raise RuntimeError("the ogsfx file not found")
```
BigRoy, line 40:
```suggestion
            ogsfx_path = instance.context.data["project_settings"]["maya"]["publish"]["ConvertGLSLShader"]["ogsfx_path"]  # noqa
            if not os.path.exists(ogsfx_path):
                raise RuntimeError("The ogsfx shader file does not exist: {}".format(ogsfx_path))
```
I'd at least report what path you've checked.
Also, if you using ` # noqa` anyway I'd just take the whole settings path on one line instead.

LiborBatek (approved):
Been sucessfuly tested in # 4327 and works ok.

openpype/hosts/maya/plugins/publish/extract_gltf.py
```diff
@@ -45,6 +45,10 @@ def process(self, instance):
             "glb": True,
             "vno": True    # visibleNodeOnly
         }
+
+        # convert to gltf shader
+        self.convert_gltf_shader(instance)
+
```
BigRoy, line 51: This introduces the problem brought up [here](https://github.com/ynput/OpenPype/pull/4261#discussion_r1087555731) again.

Please do **not** merge this PR as is. Publishing should **NOT** automatically change the contents of your current scene. The workfile should remain untouched unless a user triggers an explicit action, like e.g. a Repair Action.

If converted shaders like these are required a Validator should be generated with a Repair action to allow the user to convert it.

openpype/hosts/maya/plugins/publish/extract_gltf.py
```diff
@@ -63,3 +67,80 @@ def process(self, instance):
...
+                                        renderable=True, noSurfaceShader=True)
+            cmds.connectAttr(glsl + ".outColor",
+                             glsl_shadingGrp + ".surfaceShader")
+
+            # load the maya2gltf shader
+            ogsfx_path = instance.context.data["project_settings"]["maya"]["publish"]["ExtractGLB"]["ogsfx_path"]  # noqa
+            if not os.path.exists(ogsfx_path):
+                self.log.info("Find the ogsfx shader file in "
+                              "default maya directory...")
+                # re-direct to search the ogsfx path in maya_dir
```
BigRoy, line 96: This log should describe WHY it falls back to the `MAYA_APP_DIR` - what file wasn't found, what was missing?
It'd be even simpler if it wouldn't even _try_ to fallback but just explicitly only allow to use the path from settings.

BigRoy (changes requested):
The publishing still changes the workfile 'automatically' and leaves it changed afterwards. That should be avoided.

Either the extractor changes it temporarily and reverts to the original state after the export was made (thus the state was only ever needed for export) OR an explicit user interaction like triggering a "repair" could be used to change the scene.

The problem we're introducing here is the same as e.g. having a model publish automatically changing the user's workfile by removing any non-manifold polygons. It's their model - they wouldn't even know that potentially vertices had disappeared - it's dangerous. The workfile should **not** be changed without the user's knowledge.

openpype/hosts/maya/plugins/publish/extract_gltf.py
moonyuet: This is done like this mainly because there is no guarantee that setting up custom path from settings is right. Or somehow the user will just leave the path there with blank as they install the plugin in maya directory. 
The documentation of maya2glTF asks the user to copy and paste the files within MAYA_APP_DIR. If the user fails to put into the custom mode, the system tried to find if the ogsfx path located after the installation of maya2glTF,  and use that instead. I can use self.log.warning instead of self.log.info. Maybe I am thinking too much, but if I follow what you suggested in the comment in the similar PR(only allow to use the path from settings.) somehow the system can't find the path and error out. Maybe it's just my computer.

BigRoy: I meant to say that IF the user had supplied a custom path in settings - the user would currently never be warned about that particular path not being found. It'd just see that it's searching the default maya directory.

E.g.

```python
            ogsfx_path = instance.context.data["project_settings"]["maya"]["publish"]["ExtractGLB"]["ogsfx_path"]  # noqa
            if not os.path.exists(ogsfx_path):
                if ogsfx_path:
                    # The settings that have a custom ogsfx_path set. Let's warn
                    # the user that it didn't find the path specified
                    self.log.warning("ogsfx shader file not found in {}".format(ogsfx_path))
                        
                self.log.info("Searching ogsfx shader file in "
                              "default maya directory...")
```

openpype/hosts/maya/plugins/publish/validate_gltf_textures_names.py
antirotor: > Is there a reason the `{Name}` key is uppercase?

This is probably because of PascalCase required by th [UE5 naming convention](https://github.com/Allar/ue5-style-guide#211-always-use-pascalcase)

antirotor (changes requested):
I agree with @BigRoy that workfile shouldn't be automatically changed. I propopse "simple" change - lets move `convert_gltf_shader` logic to repair action of validator, that will check if the shaders are GLSL shaders. That way artist might do it manually for whatever reason before publishing and we'll still provide tool for helping him with conversion. That logic of `convert_gltf_shader` can be even added to separate script and executed manually via OpenPype tool menu for example.

openpype/hosts/maya/plugins/publish/validate_glsl_material.py
```diff
@@ -0,0 +1,155 @@
...
+        # load the glsl shader plugin
+        cmds.loadPlugin("glslShader", quiet=True)
+
+        for mesh in meshes:
+            # create glsl shader
+            glsl = cmds.createNode('GLSLShader')
+            glsl_shadingGrp = cmds.sets(name=glsl + "SG", empty=True,
+                                        renderable=True, noSurfaceShader=True)
+            cmds.connectAttr(glsl + ".outColor",
+                             glsl_shadingGrp + ".surfaceShader")
```
antirotor, line 87:
```suggestion
            glsl_shading_grp = cmds.sets(name=glsl + "SG", empty=True,
                                        renderable=True, noSurfaceShader=True)
            cmds.connectAttr(glsl + ".outColor",
                             glsl_shading_grp + ".surfaceShader")
```

openpype/hosts/maya/plugins/publish/validate_glsl_material.py
```diff
@@ -0,0 +1,155 @@
...
+
+                        # connect nrm map if there is one
+                        nrm = cmds.listConnections(shader +
+                                                   ".TEX_normal_map")
+                        if nrm:
+                            nrm_output = nrm[0] + ".outColor"
+                            glsl_nrm = glsl + ".u_NormalTexture"
+                            cmds.connectAttr(nrm_output, glsl_nrm)
+
+            cmds.sets(mesh, forceElement=str(glsl_shadingGrp))
```
antirotor, line 155:
```suggestion
            cmds.sets(mesh, forceElement=str(glsl_shading_grp))
```

antirotor (approved):

antirotor:
some cosmetic issues, passing for one last functional test to @LiborBatek

LiborBatek (changes requested):
Did another tests.

First when using StingrayMat in ```maya2023``` with correctly connected textures (D,N,ORM) using ```color out``` for all of them ```publish``` went ok but the ```glb file``` has no textures.

4261_GLB_File_01

When using ```aiStandard``` material with connected textures again, then after successful publish from maya the ```glb file``` has textures correctly connected/exported.

4261_GLB_File_02

however seems like ```Transparency mode``` is not correctly set in the ```glb file```. Not sure but can be just a settings in the werb viewer. However the ```publish``` should check if ```albedo with alpha``` present/used in mat and set it during extraction correctly. 

Should be more like this in my case (no alpha present, opaque blending mode, or even none)

4261_GLB_File_03

There were no any warnings and ```publish``` went ok without any errors in both cases. Also noting no matter if ```stingray material``` or ```aiStandard``` were used. 

image

LiborBatek:
Not sure if it is correct behavior to give no warnings if using maya ```aiStandard``` surface or ```Stingray mat```. Both cases just went without any warnings.
Also note no any changes to the maya scene after publish (no alteration to maya scene data) ...which is ok I guess!

moonyuet:
The updated PR set the validator to be enabled by default, and it will yell if you dont assign the shader to the object.
image

image
if you are using the stringray PBS, it will relink the textures from PBS to the the glsl shader when you do repair function
image

if you are using the shader other than PBS and GLSL, it will just assign the glsl shader when you do repair function
image


But i think the shader needs to have double test, although the maps are all there
image

LiborBatek (changes requested):
Tested again with GLSL shader validator turned on in OP settings. The validator correctly kicks in when no GLSL shader present on the asset and ```repair``` action assigns the GLSL shader to the asset.

There are several issues though:

1) when converting to the GLSL some textures arent connected then (namely Metallness) see pic

image

leading to semi-correct output in GLB file (missing metallness information in the shader, the transparancy blending is correct however)

image

2) There is a missing validator for ```maya2gltf plugin``` being loaded so user is warned and cannot continue without loading it (right now the publish normally continue even the plugin isnt loaded which leads to wrong conversion to GLSL shader - textures arent connected to it then)

moonyuet:
the updated PR adds the new validator for maya2glTF plugin
image
and also it will re-linked the textures from aiStandardSurface to the newly assigned GLSL shader.
image

antirotor (changes requested):
Just a little error handling

LiborBatek (approved):
All components works ok.

Been tested

1) maya2gltf plugin ```validator``` correctly kicks in when no plugin loaded and ```repair``` action correctly loads the plugin
2) when no GLSL shader assigned ```repair``` action assigns GLSL shader with all textures correctly connected
3)when non GLSL shader (aiStandard) also repairs ok with all textures connected
4) GLB file looks ok in web Gltf viewer without any errors and with all tex maps

image

antirotor (approved):

PR #4268, opened 2022-12-28
Scene Manager: Fix variable name

iLLiCiTiT:
## Brief description
Fix access to width of geometry in `ScrollMessageBox`.

## Description
The `ScrollMessageBox` tried to access width of desktop by accessing `width` attribute but it is a method so comparison failed. 
Also renamed variable `max_len` to `message_len`.

## Additional info
We should maybe rewrite the whole dialog as the logic of size definition is not readable much. The dialog also contains "Copy to clipboard" button which is on right side next to OK and Cancel which is not very intuitive. Dialog does not have enabled close.

## Testing notes:
1. Use the dialog (Currently used only in Maya look loader)

Resolves https://github.com/pypeclub/OpenPype/issues/4243#event-8068289765

### DIFF
--- a/openpype/widgets/message_window.py
+++ b/openpype/widgets/message_window.py
@@ -113,7 +113,7 @@ class ScrollMessageBox(QtWidgets.QDialog):
             message_len = max(message_len, len(message))
 
         # guess size of scrollable area
-        max_width = QtWidgets.QApplication.desktop().availableGeometry().width
+        max_width = QtWidgets.QApplication.desktop().availableGeometry().width()
         scroll_widget.setMinimumWidth(
             min(max_width, message_len * 6)
         )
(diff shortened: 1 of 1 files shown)


LiborBatek:
From user POV seems normal. but will keep decision for ya devs...
image

BigRoy:
@LiborBatek is that the wrong screenshot you uploaded?

LiborBatek:
@BigRoy  well I guess it seems ok to me...I wasnt sure cos its probably more about code review I guess? The dialog works fine no any glitches,

To your question, no its not that wrong one (with issue) ...I just tested this PR in maya2023 and seems normal. 
So in the end I think its more of a review for you devs.

antirotor (approved):

PR #4270, opened 2022-12-30
Global: color-managed publishing workflow

jakubjezek001:
## Brief description
Color-managed publishing of picture data.
 
## Description
Implemetation of OpenColorIO workflow within publishing workflow. We have enabled to store colorspace metadata per representation. We are storing ocio config path and colorspace name for downstream processing and also for future references within loading worflow.

## Additional info
This PR has base branch from https://github.com/pypeclub/OpenPype/pull/4246 - so it will make sense to close it and continue with recommended improvement suggestions from @illicit within this PR.

## Testing notes:
use `poetry run python ./start.py runtests ../tests/unit/openpype/pipeline` in OpenPype root. All tests should pass.


openpype/pipeline/colorspace.py
```diff
@@ -0,0 +1,213 @@
...
+    run_openpype_process
+)
+from openpype.pipeline import Anatomy
+from openpype.lib.log import Logger
+
+log = Logger.get_logger(__name__)
+
+
+@contextlib.contextmanager
+def make_temp_file():
```
iLLiCiTiT, line 21: I would make it private and more explicit
```suggestion
def _make_temp_json_file():
```

openpype/pipeline/colorspace.py
```diff
@@ -0,0 +1,213 @@
...
+import contextlib
+import tempfile
+from openpype import PACKAGE_DIR
+from openpype.settings import get_project_settings
+from openpype.lib import (
+    StringTemplate,
+    run_openpype_process
+)
+from openpype.pipeline import Anatomy
+from openpype.lib.log import Logger
```
iLLiCiTiT, line 15: There is no need to import explicitly from `openpype.lib.log`
```suggestion
from openpype.lib import (
    StringTemplate,
    run_openpype_process,
    Logger
)
from openpype.pipeline import Anatomy
```

openpype/scripts/ocio_config_op.py
```diff
@@ -0,0 +1,52 @@
+import click
+import json
+from pathlib2 import Path
+import PyOpenColorIO as ocio
+
+
+@click.group()
+def config():
+    pass
```
iLLiCiTiT, line 9: With this change there is no need to call `cli.add_command(config)`. I would also recommend to rename `cli` function to `main` (it's not requirement, but convenience).
```suggestion

@click.group()
def cli():
    pass


@cli.command()
def config():
    pass
```

openpype/pipeline/publish/publish_plugins.py
```diff
@@ -13,6 +13,12 @@
     get_instance_staging_dir,
 )
 
+from ..colorspace import (
```
iLLiCiTiT, line 16:
```suggestion
from openpype.pipeline.colorspace import (
```

openpype/plugins/publish/integrate.py
```diff
@@ -529,6 +529,13 @@ def prepare_representation(self, repre,
         template_data["representation"] = repre["name"]
         template_data["ext"] = repre["ext"]
 
+        # add template data for colorspaceData
+        if repre.get("colorspaceData"):
+            colorspace = repre["colorspaceData"]["colorspace"]
+            # replace spaces with underscores
+            colorspace = colorspace.replace(" ", "_")
```
iLLiCiTiT, line 536: I'm not sure if the replacement of spaces should happen.

jakubjezek001: yes it should. I am retreaving the file colorspace with underscores here https://github.com/pypeclub/OpenPype/pull/4270/files#diff-c4b595d393443040b1d7c9551351bfffe69f9c00ad87bd9b6243da47a643f337R118

openpype/scripts/ocio_config_op.py
jakubjezek001: this is not working AttributeError: 'Command' object has no attribute 'command'

antirotor:
Great work! At first glance I would add docstrings and maybe just use python 3 syntax in host agnostic scripts? (like f-strings, etc.) Nothing crutial, but makes code more readable. I'll dig in deeper.

openpype/plugins/publish/integrate.py
antirotor: I would add it to the code comment then, otherwise it will get lost in time.

antirotor (approved):
I've fixed few typos, but otherwise it looks solid.

openpype/scripts/ocio_wrapper.py
```diff
@@ -0,0 +1,168 @@
+"""OpenColorIO Wrapper.
+
+Only to be interpreted by Python 3. It is run in subprocess in case
+Python 2 hosts needs to use it. Or it is used as modul for Python 3
```
BigRoy, line 4:
```suggestion
Python 2 hosts needs to use it. Or it is used as module for Python 3
```
typo

openpype/hosts/nuke/plugins/publish/extract_render_local.py
```diff
@@ -7,7 +7,7 @@
 from openpype.pipeline import publish
 
 
-class NukeRenderLocal(publish.Extractor):
+class NukeRenderLocal(publish.ExtractorColormanaged):
     # TODO: rewrite docstring to nuke
```
BigRoy, line 11: Just wondering - unrelated to this PR - isn't this `TODO` already done? :)

BigRoy:
After this PR - upon loading a publish. How would one retrieve the color space originally used for the file?
In other words, how to apply the same color space setting on load in Maya or alike.

PR #4271, opened 2022-12-30
Adding path validator for non-maya nodes

moonyuet:
## Brief description
Adding a path validator for filepaths from non-maya nodes, which are created by plugins such as Renderman, Yeti and abcImport.

## Description
As File Path Editor cannot catch the wrong filenpaths from non-maya nodes such as AlembicNodes, It is neccessary to have a new validator to ensure the existence of the filepaths from the nodes. 

## Additional info
There are some possibilities that some non-maya nodes with filepath attributes missing in the validator. 
Please suggest if you catch one or more missing.

## Testing notes:
1. Launch Maya from Openpype
2. Create non-maya nodes from the plugins(like importing abc) 
3. Run the publisher
4. It will error out and show that the nodes with wrong filepath if it does not exist.


m-u-r-p-h-y:
we should create some easy way to add new nodes with path attributes to the validation as we will miss some or there will be new nodes coming with every renderer in maya.

m-u-r-p-h-y:
**Redshift nodes with file path attribute**

```
RedshiftSprite.tex0
RedshiftBokeh.dofBokehImage
RedshiftLensDistortion.LDimage
RedshiftCameraMap.tex0
RedshiftEnvironment.tex0
RedshiftEnvironment.tex1
RedshiftEnvironment.tex2
RedshiftEnvironment.tex3
RedshiftEnvironment.tex4
RedshiftDomeLight.tex0
RedshiftDomeLight.tex1
RedshiftIESLight.profile
RedshiftLightGobo.tex0
RedshiftNormalMap.tex0
RedshiftProxyMesh.fileName
RedshiftVolumeShape.fileName
```

m-u-r-p-h-y:
**Vray nodes with file path attribute**

```
VRayTexGLSL.fileName
VRayMtlGLSL.fileName
VRayVRmatMtl.fileName
VRayPtex.ptexFile
VRayLightIESShape.iesFile
VRayMesh.fileName2
VRayMesh.overrideFileName
VRayMesh.alembicLayersFileName
VRayMesh.materialAssignmentsFile
VRayMtlOSL.fileName
VRayTexOSL.fileName
VRayTexOCIO.ocioConfigFile
VRaySettingsNode.causticsFile2
VRaySettingsNode.causticsAutoSaveFile2
VRaySettingsNode.lc_fileName
VRaySettingsNode.lc_autoSaveFile
VRaySettingsNode.imap_fileName2
VRaySettingsNode.imap_autoSaveFile2
VRaySettingsNode.pmap_file2
VRaySettingsNode.pmap_autoSaveFile2
VRayScannedMtl.file
VRayScene.FilePath
VRayScene.parameterOverrideFilePath
VRayMtlMDL.filename
VRayProxy.fileName
VRaySimbiont.file
```

moonyuet:
> we should create some easy way to add new nodes with path attributes to the validation as we will miss some or there will be new nodes coming with every renderer in maya.
Yes I agree I think it would be great to let the user just add nodes and the attributes in the openpype settings. I will add some default examples for references in the next update PR

moonyuet:
I have updated the options to allow users adding filename attributes and the related nodes in the setting. Like what you said, it is easier option to deal with the issue as the list of nodes with filename attributes are way too long. 
image

I will set the attributes below as default settings for the user references, it wont affect the validator works if the nodes do not exist. 
image

antirotor:
I agree with the solution, but lets add the nodes we know (and @m-u-r-p-h-y provided) as default settings so they are already present... No harm in doing that and it can save a lot of time configuring.

moonyuet:
> VRaySimbiont

I have added all the file attributes from the list @m-u-r-p-h-y provided into the settings in the updated PR.

openpype/hosts/maya/plugins/publish/validate_path_for_plugin.py
```diff
@@ -17,51 +17,29 @@ class ValidatePathForPlugin(pyblish.api.InstancePlugin):
...
+        )
+        file_attr = validate_path["ValidatePathForPlugin"]
+        if file_attr:
+            # get the nodes and file attributes
+            for node, attr in file_attr.items():
+                # check the related nodes
+                targets = cmds.ls(type=node)
+                path_attr = ".{0}".format(attr)
+
+                if targets:
```
BigRoy, line 37: No need to check `if targets`. If it's an empty `targets` list the for loop already wouldn't do anything.

openpype/hosts/maya/plugins/publish/validate_path_for_plugin.py
```diff
@@ -17,51 +17,29 @@ class ValidatePathForPlugin(pyblish.api.InstancePlugin):
...
-                red_fname = cmds.getAttr(redshift + ".fileName")
-                if red_fname and not os.path.exists(red_fname):
-                    invalid.append(redshift)
+
+        # get the project setting
+        validate_path = (
+            instance.context.data["project_settings"]["maya"]["publish"]
+        )
+        file_attr = validate_path["ValidatePathForPlugin"]
+        if file_attr:
```
BigRoy, line 30: I'd try to push for less indentation and change this to:

```python
if not file_attr:
    return []
```

openpype/hosts/maya/plugins/publish/validate_path_for_plugin.py
```diff
@@ -17,51 +17,29 @@ class ValidatePathForPlugin(pyblish.api.InstancePlugin):
...
+            # get the nodes and file attributes
+            for node, attr in file_attr.items():
+                # check the related nodes
+                targets = cmds.ls(type=node)
+                path_attr = ".{0}".format(attr)
```
BigRoy, line 35: Since we're formatting the string in the for loop anyway I see no reason why we need to prefix attr here already.

openpype/hosts/maya/plugins/publish/validate_path_for_plugin.py
```diff
@@ -17,51 +17,29 @@ class ValidatePathForPlugin(pyblish.api.InstancePlugin):
...
+
+                if targets:
+                    for target in targets:
+                        # get the filepath
+                        filepath = cmds.getAttr(target + path_attr)
```
BigRoy, line 40:
```suggestion
                        filepath = cmds.getAttr("{}.{}".format(target, attr))
```

openpype/settings/defaults/project_settings/maya.json
```diff
@@ -13,7 +13,9 @@
         },
         "colorManagementPreference": {
             "configFilePath": {
-                "windows": [],
+                "windows": [
+                    "C:\\Users\\Kayla\\Downloads\\OpenColorIO-Configs-master\\OpenColorIO-Configs-master\\aces_1.2"
+                ],
```
BigRoy, line 18: This change shouldn't be here.

openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_publish.json
```diff
@@ -313,6 +313,16 @@
...
+        {
+            "type": "dict-modifiable",
+            "collapsible": true,
+            "key": "ValidatePathForPlugin",
+            "label": "Non-existent Paths in Non-Maya Nodes",
+            "use_label_wrap": true,
+            "object_type": {
+                "type": "text"
+            }
+        },
```
BigRoy, line 325: I believe this is the wrong settings schema for this plugin. Instead it should be like the other plugins where it can also have `optional`, `enabled` , etc. for the `ValidatePathForPlugin` key.

As such, the `attributes` it should validate should be a subkey, like `project_settings/maya/publish/ValidatePathForPlugin/attributes`

@antirotor - thoughts?

BigRoy (commented):
Nice stuff - I have added some comments.

Another thing that stood out to me is that it doesn't explicitly tell the user which attributes failed. I think we might want to `self.log.error` the node attribute with the invalid value:

```
File does not exist: node.attr = "/path/to/file.png"
```

Should we log the specific attributes involved?

m-u-r-p-h-y:
OpenVDB
```
dlOpenVDBShape.filename
```
GPU Cache
```
<shape>.cacheFileName
```

Yeti
```
pgYetiMayaShape.cfn
pgYetiMayaShape.ofn
pgYetiMayaShape.labf
pgYetiMayaShape.trf
pgYetiMayaShape.liveABCFilename
```

BigRoy:
**Note**

Be aware that the logic involved is relatively rudimentary but the filepaths involved for some of these attributes can be quite different. E.g. some might support `<UDIM>`, `####` (frame padding?), `%04d` (frame padding, yeti?).

Then additionally if these would involved checking sequences - should it invalidate if ANY frame is missing inside the frame range? Or should we just check to see whether there's any image on disk in that range at all? If we were to check the full range of the sequence against e.g. `instance.data["frameStartHandle"` and `instance.data["frameEndHandle"]` then note that some of these attributes might also be influenced by additional attributes that could offset a sequence, or in some cases even have a custom `inputTime` attribute driving the node to what frames it would sample.

Currently if the logic here would check an attribute that has a dynamic file pattern like a sequence or a UDIM texture it would invalidate the publishing.

moonyuet:
> **Note**
> 
> Be aware that the logic involved is relatively rudimentary but the filepaths involved for some of these attributes can be quite different. E.g. some might support `<UDIM>`, `####` (frame padding?), `%04d` (frame padding, yeti?).
> 
> Then additionally if these would involved checking sequences - should it invalidate if ANY frame is missing inside the frame range? Or should we just check to see whether there's any image on disk in that range at all? If we were to check the full range of the sequence against e.g. `instance.data["frameStartHandle"` and `instance.data["frameEndHandle"]` then note that some of these attributes might also be influenced by additional attributes that could offset a sequence, or in some cases even have a custom `inputTime` attribute driving the node to what frames it would sample.
> 
> Currently if the logic here would check an attribute that has a dynamic file pattern like a sequence or a UDIM texture it would invalidate the publishing.

I think we can try to use clique library to query those files to see if possible to validate the files, but definitely needs time to write and test something out.

openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_publish.json
antirotor: sure, something like:

```javascript
{
            "type": "dict",
            "collapsible": true,
            "key": "ValidatePathForPlugin",
            "label": "Non-existent Paths in Non-Maya Nodes",
            "children": [
            {
                "type": "boolean",
                "key": "enabled",
                "label": "Enabled"
            },
            {
                "type": "boolean",
                "key": "optional",
                "label": "Optional"
            },
            {
                "type": "boolean",
                "key": "active",
                "label": "Active"
            },
            {
                "type": "dict-modifiable",
                "store_as_list": true,
                "key": "attributes",
                "label": "Attributes",
                "use_label_wrap": true,
                "object_type": {
                    "type": "text"
                }
         }
}
```

m-u-r-p-h-y:
Again we should first provide the feature and improve it over time (not waiting to be perfect from the first go)
:pray:
adding support to all tokens `<####>  <%04d> <F> <f> <UDIM> <udim> <frame>` can be another PR

moonyuet:
> Again we should first provide the feature and improve it over time (not waiting to be perfect from the first go) 🙏 adding support to all tokens `<####> <%04d> <F> <f> <UDIM> <udim> <frame>` can be another PR

 I agree we can provide the feature first, while we can work on validating the tokens in the next PR, as it needs some time to test and see if it works as expected.

openpype/hosts/maya/plugins/publish/validate_path_for_plugin.py
```diff
@@ -0,0 +1,51 @@
+import os
```
tokejepsen, line 1: Should we maybe rename this plugin to `validate_path_attributes`?

Whether its a native node in Maya or not, the path attributes still need to be correct. We can include any native Maya files in the plugin itself and leave the plugin nodes/attributes for the settings.

openpype/hosts/maya/plugins/publish/validate_path_for_plugin.py
```diff
@@ -0,0 +1,51 @@
...
+class ValidatePathForPlugin(pyblish.api.InstancePlugin):
+    """
+    Ensure Paths in Non-Maya Nodes(from plugins
+    such as Yeti, AbcExport) are correct
+    """
+
+    order = ValidateContentsOrder
+    hosts = ['maya']
+    families = ["workfile"]
+    label = "Non-existent Paths in Non-Maya Nodes"
```
tokejepsen, line 19: Seems like a mouthful. Maybe change to `Path Attributes( in Maya Plugin Nodes)`. (If we change the plugin name, we can skip the label in parentheses.)

openpype/settings/defaults/project_settings/maya.json
```diff
@@ -337,6 +337,64 @@
...
+            "VRayScene": "parameterOverrideFilePath",
+            "VRayMtlMDL": "filename",
+            "VRaySimbiont": "file",
+            "dlOpenVDBShape": "filename",
+            "pgYetiMayaShape": "cfn",
+            "pgYetiMayaShape": "ofn",
+            "pgYetiMayaShape": "labf",
+            "pgYetiMayaShape": "trf",
+            "pgYetiMayaShape": "liveABCFilename",
+            "gpuCache": "cacheFileName"
```
tokejepsen, line 395: Missing indentation for `attributes` entries.

openpype/settings/defaults/project_settings/maya.json
```diff
@@ -337,6 +337,64 @@
             "families": [
                 "rig"
             ]
+        },
+        "ValidatePathForPlugin": {
+            "enabled": true,
+            "optional": false,
+            "active": true,
+            "attribute": {
+            "AlembicNode": "abc_File",
```
tokejepsen, line 346: If possible it would be good to indicate to the user that its the node type and node attribute needed. If not possible in the settings view, then maybe in the documentation.

openpype/hosts/maya/plugins/publish/validate_path_for_plugin.py
BigRoy: Relabeling to "Path attributes" I think is really good idea. If it's plug-ins only "Plug-in node path attributes".
"Existing paths in attributes" could also work.

tokejepsen: Nice `Plug-in Path Attributes`

openpype/settings/defaults/project_settings/maya.json
moonyuet: I think we can add some comments here but it definitely needs more detail documentation as the user's guide

openpype/hosts/maya/plugins/publish/validate_path_attributes.py
```diff
@@ -0,0 +1,51 @@
+import os
+
+from maya import cmds
+
+import pyblish.api
+
+from openpype.pipeline.publish import ValidateContentsOrder
+
+
+class ValidatePathForPlugin(pyblish.api.InstancePlugin):
```
BigRoy, line 10: Now we still can - should we also refactor the class name to `ValidatePluginPathAttributes`?

openpype/hosts/maya/plugins/publish/validate_path_attributes.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    Ensure Paths in Non-Maya Nodes(from plugins
+    such as Yeti, AbcExport) are correct
+    """
```
BigRoy, line 14:
```suggestion
    """Validate plug-in path attributes point to existing file paths."""
```

Also `AbcExport`? I think you meant `AbcImport`? Anyway, My suggestion simplifies that away anyway.

openpype/settings/defaults/project_settings/maya.json
```diff
@@ -338,6 +338,45 @@
...
+              "VRayTexOSL": "fileName",
+              "VRayTexOCIO": "ocioConfigFile",
+              "VRaySettingsNode": "pmap_autoSaveFile2",
+              "VRayScannedMtl": "file",
+              "VRayScene": "parameterOverrideFilePath",
+              "VRayMtlMDL": "filename",
+              "VRaySimbiont": "file",
+              "dlOpenVDBShape": "filename",
+              "pgYetiMayaShape": "liveABCFilename",
+              "gpuCache": "cacheFileName"
```
tokejepsen, line 377: Less indentation than normal?

openpype/hosts/maya/plugins/publish/validate_path_attributes.py
```diff
@@ -0,0 +1,50 @@
+import os
```
tokejepsen, line 1: Think we should rename the file to `validate_plugin_path_attributes.py` to reflect the class name.

moonyuet: I got the name changed. thanks!

openpype/hosts/maya/plugins/publish/validate_plugin_path_attributes.py
```diff
@@ -0,0 +1,50 @@
...
+                targets = cmds.ls(type=node)
+
+                for target in targets:
+                    # get the filepath
+                    file_attr = "{}.{}".format(target, attr)
+                    filepath = cmds.getAttr(file_attr)
+
+                    if filepath and not os.path.exists(filepath):
+                        self.log.error("File {0} not exists".format(filepath)) # noqa
+                        invalid.append(target)
```
antirotor, line 41: A would use here the guard statement to ease up on nesting and making the code more readable:

```suggestion
        file_attr = validate_path["ValidatePathForPlugin"]["attribute"]
        if not file_attr:
            return invalid
        # get the nodes and file attributes
        for node, attr in file_attr.items():
            # check the related nodes
            targets = cmds.ls(type=node)

            for target in targets:
                # get the filepath
                file_attr = "{}.{}".format(target, attr)
                filepath = cmds.getAttr(file_attr)

                if filepath and not os.path.exists(filepath):
                    self.log.error("File {0} not exists".format(filepath)) # noqa
                    invalid.append(target)
```

antirotor (approved):
Apart of the guard statement that is in this case more of cosmetic issue, it looks good.

moonyuet:
> Apart of the guard statement that is in this case more of cosmetic issue, it looks good.

I have updated the PR with the fix on the cosmetic issue.

antirotor (approved):

openpype/hosts/maya/plugins/publish/validate_plugin_path_attributes.py
```diff
@@ -0,0 +1,52 @@
...
+    label = "Plug-in Path Attributes"
+
+    def get_invalid(self, instance):
+        invalid = list()
+
+        # get the project setting
+        validate_path = (
+            instance.context.data["project_settings"]["maya"]["publish"]
+        )
+        file_attr = validate_path["ValidatePathForPlugin"]["attribute"]
```
BigRoy, line 27: This PR has broken maya publishing because THIS `ValidatePathForPlugin` attribute does not exist in settings. It is `ValidatePluginPathAttributes` instead.

PR #4276, opened 2023-01-04
look publishing and srgb colorspace in maya 

moonyuet:
## Brief description
Check the OCIO color management is enabled before doing linearize colorspace for converting the texture maps into tx files.

## Description
Check whether the OCIO color management is enabled before the condition of converting the texture to tx extension.

## Additional info
N/A

## Testing notes:
1. launch Maya from Openpype
2. assign shaders with file textures into any mesh in your scene
3. create look instance with the meshes.
4. Publish
5. If the color management is enabled and makeTx in look instance is on, the validator will block before the extractors runs.
6. If the color management is disabled and makeTx in look instance is on, the validator won't block Openpype runs the extractors


antirotor:
I've didn't noticed it before but it seems that ignoring "linearize" flag is not solving all our problems:

So, imagine you have a texture loaded, as .png set to sRGB color space. You publish the look with *maketx* checked on. It will ignore the linearize flag and produce .tx file that is not linear.

Now when you load that look, Maya will use its file rules in Color Management settings:

image

So when it loads tx it automatically sets its colorspace to Raw, messing up the colors.

LiborBatek:
yeah thats true, the default behavior in maya and arnold tx convert utility / workflow is as (when auto convert to tx is enabled)

1- maya sets ```file_node``` automatically to proper color space by extension/filetype (for i.e. 8bit PNG being set to sRGB, EXR to linear) - or user can override it manually when necessary (valid for ie normal maps which are 8bit PNGs but must be set to "raw/linear")
2 - arnold tx utility converts to that color space so the outputted tx file has the color profile as being set in previuos step in maya 

3 - when tx files are used for rendering ,they are always set to "raw" or " linear" color in ```file_node``` by file rules preferences as @antirotor showed 
...however even when switching color settings doesnt affect the tx file and keeps orig tx color space intact (my findings from production)

It simply just matters in which color space was the file before conversion to tx and user can set it manually ignoring default color space behaviour set

LiborBatek:
@antirotor @moonyuet  looking on this discussion and my reply, to make resume

...in the end it doesnt matter that ```input color file rules``` is set to ```raw``` for ```.tx``` files if correctly converted originally :)

antirotor:
We can of course allow publishing tx files in whatever colorspace the texture is set, but the most common workflow with texture is still to use sRGB type of texture and have it linearized to tx format (and then the default rule in maya to set it to Raw makes of course sence). And that is exactly what we are not able to do until we properly implement color space management.

So I guess that warning (and that is actually what we are doing with the validator) is heard and that's why the validator is optional.

antirotor:
what do you guys @LiborBatek @m-u-r-p-h-y think? Mostly it is about workaround to be able to use ACES enabled maya with looks until we implement proper color management for looks...

LiborBatek:
I did a quick test in ```maya2023``` and it is as I have written...Arnold just cares about colorspace encoded inside .tx file. Always renders the texture in this color space no matter whats set in maya filenode. 

So the thing is proper color conversion to tx file is crucial and must be set before correctly. After that doesnt matter whats set in maya GUI.

antirotor:
> I did a quick test in `maya2023` and it is as I have written...Arnold just cares about colorspace encoded inside .tx file. Always renders the texture in this color space no matter whats set in maya filenode.
> 
> So the thing is proper color conversion to tx file is crucial and must be set before correctly. After that doesnt matter whats set in maya GUI.

ok, so based on this, we can allow maketx, but not linearize option.

antirotor (dismissed):
Looks good, @LiborBatek does it behave as you would expect?

LiborBatek (changes requested):
Unfortunatelly there are differences in color space of the converted .tx files. 

Here is a scrnshot (left monkey is correct colors but the right not) 
The correct one uses orig textures (PNG with autoconvert to tx) and the right bad colored one is using hero asset source files in .tx )
Asset_Tx_Conversion_01
its also obvious from the file sizes that there is difference in internal arnold auto conversion and OP conversion done during Publish...see file sizes (on left is workfiles on right published files)
image

moonyuet:
> Unfortunatelly there are differences in color space of the converted .tx files.
> 
> Here is a scrnshot (left monkey is correct colors but the right not) The correct one uses orig textures (PNG with autoconvert to tx) and the right bad colored one is using hero asset source files in .tx ) Asset_Tx_Conversion_01 its also obvious from the file sizes that there is difference in internal arnold auto conversion and OP conversion done during Publish...see file sizes (on left is workfiles on right published files) image
> 
> From this observation/testing its obvious that the conversion of the PNGs is not correct. Later on when used published converted `tx files` it really doesnt matter what color space the `file_node` is set to. Arnold just takes whats in the `tx file` as have already wrote above in discussion.
> 
> Do we take the `file node` color setting before converting or just file type ext (PNG, JPG, EXR etc) into account? It seems we fail on setting correct color space.

For this case, I think I will deal with it differently, I will implement a code to see if the color management is enabled before the tx conversion. If it isn't, it won't take the config file into the tx conversion in the current stage
I am thinking about customizing the config file for that, but not sure if it will complicate things.

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -535,6 +535,9 @@ def _process_texture(self, filepath, do_maketx, staging, linearize, force):
                 self.log.info("tx: converting sRGB -> linear")
                 additional_args.extend(["--colorconvert", "sRGB", "linear"])
 
+            self.log.info("Using nuke-default ocio config instead of maya ocio config!") # noqa
+            self.log.info("The tx conversion is different from the maya tx conversion!") # noqa
```
BigRoy, line 539: But why? Does it need to be? Couldn't it just use the OCIO path currently set in maya? Or?

moonyuet: i have tested and written to set up custom ocio config here. but it seems that if the users use aces_1.2 ocio.config, the args for` --colorconvert `needs to be also adjusted. While you are using maya-defaults, the args may also need some changes. I am not sure if the args will vary by version too. It definitely needs time to research. So that's why I set it info here first and the config itself.

antirotor: Yep, this is because we know that in nuke-default there is `sRGB` and `linear`. But without parsing and getting proper color spaces from ocio config we can't auto-determine what to use for linearizations.

BigRoy: I still think we're doing something wrong with our logic here. Arnold's built-in TX manager basically runs this over all images: `maketx -u -v --checknan --stats --unpremult --oiio %s`.

I've never really heard anyone complain about the results of that functionality. What's the reason we're instead trying to color convert it explicitly by default?

Seems like if _they_ (Autodesk/Arnold tools) don't do it by default then maybe we shouldn't either? And if we'd want to color convert (e.g. linearize) that maybe that logic should be achieved through settings?

BigRoy: I don't have much Color Space knowledge myself, but e.g. here they mention that if by chance an input image is 8bit to just keep the color values as they are: https://forum.isotropix.com/viewtopic.php?f=5&p=25452

Of course a logical issue if the input is 8 bit - but maybe we shouldn't assume that our texture inputs aren't 8 bit either?

openpype/hosts/maya/plugins/publish/validate_look_color_space.py
```diff
@@ -0,0 +1,25 @@
...
+    label = 'Maya Color Space'
+
+    def process(self, instance):
+        ocio_maya = cmds.colorManagementPrefs(q=True,
+                                              cmConfigFileEnabled=True,
+                                              cmEnabled=True)
+        maketx = instance.data["maketx"]
+
+        if ocio_maya and maketx:
+            raise Exception("Maya is color managed and maketx option is on. OpenPype doesn't support this combination yet.") # noqa
```
BigRoy, line 25: This should likely be a more specific exception than a bare `Exception` like e.g. `RuntimeError`

But, even better - to prepare for future publisher I'd already turn this into a `KnownPublishError` instead.

openpype/hosts/maya/plugins/publish/validate_look_color_space.py
```diff
@@ -0,0 +1,25 @@
...
+class ValidateMayaColorSpace(pyblish.api.InstancePlugin):
+    """
+        Check if the OCIO Color Management and maketx options
+        enabled at the same time
+    """
+
+    order = ValidateContentsOrder
+    families = ['look']
+    hosts = ['maya']
+    label = 'Maya Color Space'
```
BigRoy, line 16: I feel like this label is too generic - it doesn't actually validate any color space. The same goes for how it's listed in the OpenPype settings.

It should be clear that this relates at least to _Color Management_ and _maketx_
Would labeling it like `label = "Color Management with maketx" make more sense?

openpype/hosts/maya/plugins/publish/extract_look.py
moonyuet: > I still think we're doing something wrong with our logic here. Arnold's built-in TX manager basically runs this over all images: `maketx -u -v --checknan --stats --unpremult --oiio %s`.
> 
> I've never really heard anyone complain about the results of that functionality. What's the reason we're instead trying to color convert it explicitly by default?
> 
> Seems like if _they_ (Autodesk/Arnold tools) don't do it by default then maybe we shouldn't either? And if we'd want to color convert (e.g. linearize) that maybe that logic should be achieved through settings?

i have double checked the conversion again with maketx.exe and come up with the conclusion. The config file required for the --colorconvert command, otherwises it will error out.
image
I have practiced three different occassions with the maketx.exe in oiio and put them back into maya with Arnold renderer, finding out that the tx files without the color conversion along with the maya default ocio config, it will result different colors. The screenshot of the ipr render in Arnold kinda explains the case.
The sphere in the middle is with the shader where the tx file is autogenerated by Maya.The sphere in the left is with the shader where the albedo without` --colorconvert `args and the maya-defaults ocio config files added into args, i.e. `--colorconfig`.  The sphere in the right is with the shader where the albedo with both colorcovert args and maya-defaults ocio config set in the args during the tx conversion in maketx.exe. 
The color difference can also be seen when we are using different ocio config file, i.e. nuke-defaults.
image

moonyuet: If I don't get it wrong, file texture node in Maya automatically set the colospace to Raw if the image format is tx. The sphere in the far left is the shader with albedo texture in tx format,which the tx file is generated without both color convert and color config args. The sphere in the middle is with the shader which the albedo using autogenerated tx file by Maya. And you can see the color difference in-between. The sphere in the right side is using tx map as albedo which is generated by both color convert and Maya-default ocio color config.

BigRoy: Had a quick look on my side. Some notes:

- Whenever you directly load a `.tx` file into maya and render with arnold any color space settings you set on the `file` node do nothing whatsover. It seems Arnold will then always assume the texture itself is in the raw color space.
- The Maya viewport will change when you set the color space on a file node pointing to a `.tx` file, but arnold doesn't care at all. It takes in the `.tx` file as if it is raw.
- The auto-tx conversion does work.
- The TX manager conversion for me crashes in the latest version of arnold, yay. 

Tested in Maya 2022 + Mtoa 5.2.1.1

Might need to delve in deeper to really be able to help out. :)

BigRoy: So here's a function of one of the scripts that comes with Arnold, `scripts/mtoa/makeTx.py`:
_(There was no license in the file, but I suppose copyright Autodesk?)_

<details>
  <summary>Expand to see ``scripts/mtoa/makeTx.py` from Mtoa 5.2.2.1</summary>
```python
def makeTx(filename, colorspace='auto', arguments=''):
    '''Generate a TX texture with maketx
    '''

    # status[0] contains the amount of created tx files
    # status[1] the amount of skipped tx files
    # status[2] the amount of errors

    status = [0,0,0]
    if arguments == '':
        cmd = list(_maketx_cmd)
    else:
        cmd_str = _maketx_binary
        cmd_str += ' '
        cmd_str += arguments
        cmd = shlex.split(cmd_str, posix=False)

    #maya_version = utils.getMayaVersion()
    if mel.eval("exists \"colorManagementPrefs\""):

        if cmds.colorManagementPrefs(q=True, cmEnabled=True):
            if colorspace in cmds.colorManagementPrefs(q=True, inputSpaceNames=True):
                if cmds.colorManagementPrefs(q=True, cmConfigFileEnabled=True):
                    color_config = cmds.colorManagementPrefs(q=True, configFilePath=True)
                else:
                    color_config = cmds.internalVar(userPrefDir=True)

                render_colorspace = cmds.colorManagementPrefs(q=True, renderingSpaceName=True)
                
                if colorspace != render_colorspace:
                    cmd += ['--colorengine', 'syncolor', '--colorconfig', color_config, '--colorconvert', colorspace, render_colorspace]
            else:
                # FIXME what should we do in auto mode ?
                if colorspace != 'auto':
                    print('[maketx] Warning: Invalid input colorspace "%s" for "%s", disabling color conversion' % (colorspace, filename))

    for tile in expandFilename(filename):
        if os.path.splitext(tile)[1] == '.tx':
            print('[maketx] Skipping native TX texture: %s' % tile)
            status[1] += 1
            continue

        tile_info = imageInfo(tile)

        if colorspace == 'auto':
            colorspace = guessColorspace(tile_info)

        outputTx = os.path.splitext(tile)[0] + '.tx'
        AiTextureInvalidate(outputTx)

        AiMsgInfo( "[maketx] info : {}".format(','.join(["{}={}".format(k,v) for k, v in tile_info.items()])) )

        # promote 8-bit images to EXR half with DWAA compression to avoid quantization errors (#3919)
        if colorspace != 'linear' and tile_info['bit_depth'] <= 8:
            cmd += ['--format', 'exr', '-d', 'half', '--compression', 'dwaa']

        res = subprocess.Popen(cmd + [tile], stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE, startupinfo=_no_window).communicate()[0]

        if re.search(_maketx_rx_noupdate, res):
            # we don't want to get messages saying that texture is up-to-date, mainly
            # with auto-tx since this happens at every render.
            # If we really want it for txManager we can add an argument in this function
            # and re-introduce the print

            #print '[maketx] TX texture is up to date for "%s" (%s)' % (tile, colorspace)
            status[1] += 1
        else:
            mo = re.search(_maketx_rx_stats, res)
            if mo:
                print('[maketx] Generated TX for "%s" (%s) in %s seconds' % (tile, colorspace, mo.group(1)))
                AiTextureInvalidate(outputTx) 
                status[0] += 1
            else:
                print('[maketx] Error: Could not generate TX for "%s" (%s)' % (tile, colorspace))
                print(res)
                status[2] += 1
       
    return status
```
</details>

This has some nice information, like the 8 bit conversion they are doing, etc.

There's also the `scripts/mtoa/txManager/lib.py` which could be a good reference to check too?

### The Arnold Tx manager conversion

The TX Manager actually does the conversion from "node space" to "render space".
They do indeed do `--colorconvert`.

This is the Render Space:
```python
render_colorspace = cmds.colorManagementPrefs(query=True, renderingSpaceName=True)
```
And the texture's color space is:
```
color_space = cmds.getAttr(file_node + ".colorSpace")
```

As such, looking at that. The logic we should implement for maketx is to whenever Color Management is enabled to take the Color Space of the texture an convert it to the render space.

Et voila, issue solved.

TL;DR

For each node we should do:
The current node's color space to the render color space, that's it.

moonyuet: > @antirotor @moonyuet With above comment I feel we can solve this PR easily, yes? :) We'll just always need to do the color convert with the currently used OCIO config in maya and convert from the node's color space to the rendering space.

I am not quite sure if the "nuke-defaults" ocio config is set for some purpose in the code and let's assume yes. It would be great if there is an option to use the maya_defaults ocio config for tx conversion if it turns out it will give an accurate tx conversion.

BigRoy: > It would be great if there is an option to use the maya_defaults ocio config for tx conversion if it turns out it will give an accurate tx conversion.

Just to be exactly clear: I'm not saying to use Maya's default - but to use the config that maya is currently set to on the artist's machine. We can make the assumption that if color space management is enabled for Maya within OpenPype that OpenPype sets up Maya to be in the correct OCIO state.

moonyuet:
> There are some similarities between this PR and #4340 but not sure we can align while both are being developed so might be worth looking at a follow up issues of aligning some of the code.



> There are some similarities between this PR and #4340 but not sure we can align while both are being developed so might be worth looking at a follow up issues of aligning some of the code.



> There are some similarities between this PR and #4340 but not sure we can align while both are being developed so might be worth looking at a follow up issues of aligning some of the code.


> There are some similarities between this PR and #4340 but not sure we can align while both are being developed so might be worth looking at a follow up issues of aligning some of the code.

I just take a look at #4340 and maybe some of the codes from the PR can be re-used in this ticket to resolve the issue of colorspace caused by different config path and also the guessing of colorspace mentioned by @BigRoy

openpype/hosts/maya/api/lib.py
```diff
@@ -3379,3 +3380,45 @@ def dgcontext(mtime):
 def get_attribute_input(attr):
     connections = cmds.listConnections(attr, plugs=True, destination=False)
     return connections[0] if connections else None
+
+
+# Reference from Arnold
+# Get Image Information for colorspace
```
tokejepsen, line 3386: Put this in the methods doc string

openpype/hosts/maya/api/lib.py
```diff
@@ -3379,3 +3380,45 @@ def dgcontext(mtime):
...
+        img_info['format'] = "unknown"
+    return img_info
+
+
+def guess_colorspace(img_info):
+    ''' Take reference from makeTx.py
+    Guess the colorspace of the input image filename.
+    @return: a string suitable for the --colorconvert
+            option of maketx (linear, sRGB, Rec709)
+    '''
```
tokejepsen, line 3411: Instead of guessing, maybe we should expose some default settings so studios can override.

antirotor:
Could we fix the conflicts? I think it is almost ready to merge

openpype/hosts/maya/plugins/publish/validate_look_color_space.py
```diff
@@ -0,0 +1,25 @@
...
+    label = 'Color Management with maketx'
+
+    def process(self, instance):
+        ocio_maya = cmds.colorManagementPrefs(q=True,
+                                              cmConfigFileEnabled=True,
+                                              cmEnabled=True)
+        maketx = instance.data["maketx"]
+
+        if ocio_maya and maketx:
+            raise RuntimeError("Maya is color managed and maketx option is on. OpenPype doesn't support this combination yet.") # noqa
```
antirotor, line 25:
```suggestion
            raise PublishValidationError("Maya is color managed and maketx option is on. OpenPype doesn't support this combination yet.") # noqa
```

antirotor (changes requested):

openpype/hosts/maya/api/lib.py
```diff
@@ -16,6 +16,15 @@
...
 import maya.api.OpenMaya as om
+from arnold import (
+    AiTextureGetBitDepth,
+    AiTextureGetFormat,
+    AiTextureInvalidate,
+    # types
+    AI_TYPE_BYTE,
+    AI_TYPE_INT,
+    AI_TYPE_UINT
+)
```
BigRoy, line 27: Oh boy, we shouldn't be doing this! :) Arnold isn't a hard requirement to be installed with maya.

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -524,10 +535,35 @@ def _process_texture(self, filepath, do_maketx, staging, linearize, force):
...
+                                                "sRGB",
+                                                "Raw"])
+                    else:
+                        self.log.info("tx: texture's colorspace "
+                                      "is already linear")
+
+            config_path = cmds.colorManagementPrefs(query=True,
+                                                    configFilePath=True)
+            if not os.path.exists(config_path):
+                raise RuntimeError("No OCIO config path found!")
```
BigRoy, line 566: Seems odd we're doing this check after all the logic above, no?

Also, process texture is always called. Does this change now mean we always require a OCIO config?

moonyuet: The ocio config file is needed for the "--colorconvert" argument if i remember correctly, maybe I move the related code into the if-condtion of the colorspace

antirotor (approved):

mkolar:
@LiborBatek just waiting for a last functional review and we're good to go

LiborBatek (approved):
Works as expected...

if ```makeTX``` option on the ```look``` instance ```on``` it get validation error with following msg:

image

If ```makeTX``` option is ```off``` all validates ok and no warning/fail msg present

image

Issue #4125, opened 2022-11-23
SFTP connection failing when shouldn't be tested


Tilix4:
## Reproduce
1. Enable `Site Sync` module, create a `SFTP` site with wrong credentials and enable it for your production.
2. Change your local settings to `local` and `studio` (not to the `sftp` site).
3. Restart Tray, and look to the logs, it tries to connect when it shouldn't because the user settings don't ask for it.

The logs produced make bug hunting more tedious and slows the server to start because of attempts to connect.

## Lead
This is caused by `is_active` in `openpype\modules\sync_server\providers\sftp.py` which tries to get the connection to test if the sftp is available.

## Proposals
- Add a check of the current user settings, if the sftp site is not selected as remote, then `is_active` returns `False`.
- Don't even register the module if the sftp site is not selected as remote.

[cuID:OP-4490]

PR #4278, opened 2023-01-04
SiteSync: sftp connection failing when shouldnt be tested

kalisp:
## Brief description
Fix check if site is active, eg. correctly configured to be used, only for active and remote site actually set on a project.
Previously it went through all configured sites, even if they weren't used.

## Additional info
Site might be configured as possible target, but artist might not use it. They are in studio and not using SiteSync at all etc.

Closes https://github.com/ynput/OpenPype/issues/4125

## Testing notes:
- Enable Site Sync module, create a SFTP(or Gdrive without any values filled)  site with wrong credentials and enable it for your production.
- Change your local settings to local and studio (not to the sftp site).
- Restart Tray, and look to the logs, it tries to connect when it shouldn't because the user settings don't ask for it.

### DIFF
--- a/openpype/modules/sync_server/providers/dropbox.py
+++ b/openpype/modules/sync_server/providers/dropbox.py
@@ -165,7 +165,7 @@ class DropboxHandler(AbstractProvider):
         Returns:
             (boolean)
         """
-        return self.presets["enabled"] and self.dbx is not None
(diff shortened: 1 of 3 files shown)


m-u-r-p-h-y (approved):
with SiteSync enabled and sftp server defined it is not giving me the error anymore

seems to work fine

Issue #4272, opened 2023-01-02
Maya: Set frame range repair action doesnt set frame handles according to DB


LiborBatek:
When triggering repair action in maya scene the action doesnt set frame handles according to database values and set frame range globally (frame start/end).

For example:

database:
frame start: 1001
frame end: 1050
handle start: 10
handle end: 10

Repair action (maya):

frame start: 991
frame end: 1060
handle start: 0
handle end: 0

Repair action should keep the same values as in DB by putting correct values  for handles and not changing global frame range.

[cuID:[OP-4697](https://app.clickup.com/t/86771h246)]

LiborBatek:
When frame range differs (before repair action):
![image](https://user-images.githubusercontent.com/112623825/210253571-f6d04137-23a9-41d6-b7de-91f52cbd5c13.png)

and after repair action values arent set correctly in a way (absolute frame range set not handles only):

![image](https://user-images.githubusercontent.com/112623825/210253738-0484774a-124c-4dee-a27c-11611af3e4da.png)

Issue #3302, opened 2022-06-08
Maya: Validate frame range repair action


m-u-r-p-h-y:
**Describe the bug**
repair action on validate frame range fails with error:

```
// pyblish.ValidateMayaUnits : Units (linear): cm // 
// pyblish.ValidateMayaUnits : Units (angular): deg // 
// pyblish.ValidateMayaUnits : Units (time): 25.0 FPS // 
plugin is: <class 'D:\REPO\OpenPype\openpype\hosts\maya\plugins\publish\validate_frame_range.py.ValidateFrameRange'>
b"Preparing <class 'openpype.action.RepairAction'>.."
b'Action prepared.'
// pyblish.pyblish.plugin.Action : Finding failed instances.. // 
// Error: pyblish.plugin : Traceback (most recent call last):
  File "D:\REPO\OpenPype\.venv\lib\site-packages\pyblish\plugin.py", line 522, in __explicit_process
    runner(*args)
  File "D:\REPO\OpenPype\openpype\action.py", line 65, in process
    plugin.repair(instance)
  File "D:\REPO\OpenPype\openpype\hosts\maya\plugins\publish\validate_frame_range.py", line 90, in repair
RuntimeError: setAttr: No object matches name: Main.frameStart

Traceback (most recent call last):
  File "D:\REPO\OpenPype\.venv\lib\site-packages\pyblish\plugin.py", line 522, in __explicit_process
    runner(*args)
  File "D:\REPO\OpenPype\openpype\action.py", line 65, in process
    plugin.repair(instance)
  File "<string>", line 90, in repair
RuntimeError: setAttr: No object matches name: Main.frameStart
 // 
```
**Expected behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/352795/172612955-59b6b154-261d-4b13-beea-3710855fc3dc.png)

**Desktop (please complete the following information):**
 - OS: windows
 - Host: Maya 2022

[cuID:OP-3364]

m-u-r-p-h-y:
just tested latest develop and this bug is still there . .

m-u-r-p-h-y:
@BigRoy comment in another thread

> **Describe the bug**
> 
[...]

PR #4279, opened 2023-01-04
Maya: Fix validate frame range repair + fix create render with deadline disabled

BigRoy:
## Brief description

This fixes:
- #4272 - Now handles are set as expected if `handleStart` and `handleEnd` attribute exists on the instance
- #3302 - Render Layer frame ranges are now also repaired - also works with explicit render layer overrides on a layer
- Create Render globals not working (errors) when Deadline module is disabled

## Testing notes:

1. Test Create Render with Deadline module disabled and with it enabled in a scene that **does not yet** have a `renderingMain` globals.
2. Check with different start+end frame + handles settings for an asset on renderlayers, but also on other instances like pointcaches and cameras.

### DIFF
--- a/openpype/hosts/maya/plugins/create/create_render.py
+++ b/openpype/hosts/maya/plugins/create/create_render.py
@@ -72,15 +72,19 @@ class CreateRender(plugin.Creator):
     def __init__(self, *args, **kwargs):
         """Constructor."""
         super(CreateRender, self).__init__(*args, **kwargs)
-        deadline_settings = get_system_settings()["modules"]["deadline"]
-        if not deadline_settings["enabled"]:
-            self.deadline_servers = {}
-            return
+
+        # Defaults
         self._project_settings = get_project_settings(
             legacy_io.Session["AVALON_PROJECT"])
         if self._project_settings["maya"]["RenderSettings"]["apply_render_settings"]: # noqa
             lib_rendersettings.RenderSettings().set_default_renderer_settings()
+
+        # Deadline-only
         manager = ModulesManager()
+        deadline_settings = get_system_settings()["modules"]["deadline"]
+        if not deadline_settings["enabled"]:
+            self.deadline_servers = {}
+            return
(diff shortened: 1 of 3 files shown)


LiborBatek:
Did a first rum of test with creating ```Render``` subset. 

It works ok when deadline is turned on (deadlinewebservice is running) but gives errors when deadline module is off in OP settings and not running...

image

will send soon testing for the rest (Step - 02)

LiborBatek:
Test 02 - different frame start / end + frame handles setting and ```repair``` action

The ```repair``` action works on subsets (camera, review, animation etc) but it just set ```frame start``` and ```frame end``` according to absolute frame range as in DB (range + handles together) which sort of works but it doesnt respect the handle values in DB and the repair action doesnt put these values in. It behaves as in this PR #4272 aka same issue. So question is:
is it desirable to maintain frame start/end and frame handle start/end values exactly (as in DB) or just we care about absolute frame range so they match overall (maya scene / DB)

Another issue with ```repair``` action is when used on ```renderLayer``` and frame range overrides...it just fails and gives this error msg:
image

LiborBatek (changes requested):
Pls see comments, needs change doesnt work as expected.

BigRoy:
@LiborBatek I'm pretty sure you were NOT running those tests on the code in this PR but were running it on a different branch. The line numbers in the error of your first screenshot [here](https://github.com/ynput/OpenPype/pull/4279#issuecomment-1372146125) refer to very different lines in this PR. 

Also the second test you've done is exactly what this PR should also solve. Are you sure you tested correctly?

LiborBatek (approved):
All works as expected. Creating ```render``` subset in any circumstances (with/without deadline running). 
Repair action aslo works on all subsets and renderlayers as expected. Also sets correctly handle start and handle end. So it also resolves PR #4272

Here is scrncap before repair:
image

and after:

image

here is also DB values for check and all is ok...frame ranges are equal and handles set correctly in maya.
image

antirotor (approved):

Issue #3453, opened 2022-06-30
Maya: on exit there is a window left and maya is still running


m-u-r-p-h-y:
**Describe the bug**
Under certain conditions there is a window left open on Maya exit and the process is not finished, consuming a lot of memory (depending on the scene content before closing it)

**To Reproduce**

It is extremely random and very hard to consistently reproduce. So far it happens in the case of publish which ends up with fail and closing Maya.
It was reported by other users as well.

![image](https://user-images.githubusercontent.com/352795/176696409-5ff779ef-6875-4455-b395-405cf7232911.png)

**Expected behavior**
Maya should exit and close all subprocesses in a clean way

**Desktop (please complete the following information):**
 - OS: Windows
 - Host: Maya2022

[cuID:[OP-3510](https://app.clickup.com/t/2janfxk)]

BigRoy:
We've only ever had this occur with Maya 2022. Not sure if it happens in Maya 2023 too though.
But would definitely love this solved.

BigRoy:
Has anyone been able to solve this for Maya 2022?

Issue #4205, opened 2022-12-09
Substance Painter & Substance Designer integration


kekefreedog:
Hello !

## Need

Do you planned to integrate Substance Painter & Substance Designer into OpenPype ?

## Ressources

https://www.adobe.com/fr/products/substance3d-painter.html

https://substance3d.adobe.com/documentation/spdoc/scripts-and-plugins-197427392.html

[cuID:[OP-4612](https://app.clickup.com/t/382ke3d)]

BigRoy:
@kekefreedog As far as I'm aware there hasn't been huge community interest in a Substance Painter/Designer integration as of yet. It's nice seeing your request pop up and I hope there's more interest! :)

What particular features would you like in the integration?

- Workfiles support: save, open substance workfiles
- Launch Substance from OpenPype launcher
    - including start Substance with last workfile
- Importing published alembic models
- Exporting textures (and also publish them?)
- A published "package" for the textures to automatically be able to load import/load them for a renderer in a DCC of choice (massive undertaking likely)

### References

- [Shotgrid Toolkit Integration for Substance](https://community.shotgridsoftware.com/t/substance-designer-shotgun-toolkit-engine-released/9944)
- [Prism 2.0 Substance Integration Preview between 07:00 - 09:10](https://vimeo.com/551545616)

kekefreedog:
Hi @BigRoy !

Your list is perfectly matching with our need.

Then it could be nice to have those additional features :
- Have a color management using OCIO
- When Publish texture, option for generate equivalent of `.tx` for some render engine (like Arnold...) 
- Generate material already connected to mesh
- Be able to export USD layer with material and its assignations

BigRoy:
Those are some amazing stretch goals to the integration. Thanks for these notes - I don't expect these to be there for an initial integration but they are definitely nice-to-haves to look into once the bare necessities and first draft of the substance integration is tested.

kekefreedog:
I definitely agree with you. First integration will be awesome for lots of studio. 
Then if my dev department has free time, they could be interesting for contribute to it !
Let's keep in touch.
Thx ;)

BigRoy:
@kekefreedog I'm not sure if you had seen but I did a first prototype integration for Sustance Painter some time ago: https://github.com/ynput/OpenPype/pull/4283

It's not production-ready but any feedback is welcome of course. The biggest change that'll likely still need to be implemented is how texture files are output/published. (Like the naming conventions, etc.)

[...]

Issue #3800, opened 2022-09-07
Publish: Allow to disable printing logs to host


BigRoy:
**Is your feature request related to a problem? Please describe.**

I'd like to disable Pyblish (and potentially even OpenPype Logger logs during publishing) printing its logs into the Host DCC (e.g. Maya) as a setting when running the Pyblish UI.

**Describe the solution you'd like**

Be able to set e.g. a `OPENPYPE_PRINT_PYBLISH_LOGS=0` or an equivalent setting to disable the printing to `stderr`. The logs should still get collected and handled by the UI. I'd only like to have it **not** propagate its logs into the host DCC.

**Describe alternatives you've considered**

I've tried disabling the propagating of the Pyblish loggers like this:
```python
import logging

log = logging.getLogger("pyblish")
log.propagate = False
```

But that also avoids the Pyblish Pype UI to actually capture the logs. Note that the original Pyblish QML does work with propagate disabled - there the logs still get handled by the UI regardless.

**Additional context**

Printing in the Maya script editor has always been relatively slow - and during publishing there's relatively quite a high amount of logging/printing being done. I'd like to try and avoid the clutter in the Script Editor and at the same time likely optimize the publishing steps too.

Especially with the new publisher UI coming up the logs are much easier to debug per instance in the UI - better than the printed logs into the host DCC would allow for debugging.

[First mentioned on OpenPype discord here](https://discord.com/channels/517362899170230292/563751989075378201/1016985924392992828)

---

Of course would also love to know if this is a bad idea to begin with. :)

kalisp:
It would be great if we could set logging level for Pyblish plugins.

BigRoy:
Just wanted to chime in that this is very relevant for Substance Painter too. I had a case where a single `self.log.info("TEST")` took about 0.2 seconds if Substance's Painter logs were already quite full. Clearing the Substance Painter logs sped it up again. But that likely means not filling up its logs will be beneficial too.

As an example:
```python
class ExampleFast(pyblish.api.InstancePlugin):
    order = pyblish.api.CollectorOrder + 0.4

    def process(self, instance):
        return

class ExampleSlow(pyblish.api.InstancePlugin):
    order = pyblish.api.CollectorOrder + 0.4

    def process(self, instance):
        self.log.info("TEST")
```
The fast example here would run over 30 instances pretty much instantly, the second would then take about 0.2seconds per instance if Substance Painter has a lot of previous logs still in its logs (even if the log view is disabled).

[...]

PR #4283, opened 2023-01-06
Substance Painter Integration

BigRoy:
## Brief description

This implements a part of #4205 by implementing a Substance Painter integration

## Description

Status:
- [x] Implement Host
  - [x] start substance with last workfile using `AddLastWorkfileToLaunchArgs` prelaunch hook
- [x] Implement Qt tools
- [x] Implement loaders 
    - [x] Implemented a Set project mesh loader (this is relatively special case because a Project will always have exactly one mesh - a Substance Painter project cannot exist without a mesh).
- [x] Implement project open callback
    - [x] On project open it notifies the user if the loaded model is outdated
- [x] Implement publishing logic
    - [x] Workfile publishing 
    - [x] Export Texture Sets
- [x] Support OCIO using #4195 (draft brach is set up - [see comment](https://github.com/ynput/OpenPype/pull/4283#issuecomment-1383002885))
    - [ ] Likely needs more testing on the OCIO front
- [x] Validate all outputs of the Export template are exported/generated
    - [x] Allow validation to be optional **(issue: there's no API method to detect what maps will be exported without doing an actual export to disk)**
    - [x] Support extracting/integration if not all outputs are generated
- [x] Support multiple materials/texture sets per instance
    - [ ] Add validator that can enforce only a single texture set output if studio prefers that.
- [ ] Implement Export File Format (extensions) override in Creator
    - [ ] Add settings so Admin can choose which extensions are available.

### Substance Painter Qt Stylesheet

Substance Painter has quite a lot of styleSheet overrides on its own QApplication which do mess a bit with the one for the OpenPype tools as mentioned on [Ynput discord](https://discord.com/channels/517362899170230292/563751989075378201/1060583589487575143) - [here's a rough stylesheet export of what Substance Painter applies](https://gist.github.com/BigRoy/bd0dcd97a293d742d6fe914ddaa107f7?permalink_comment_id=4426415#gistcomment-4426415). 

Some style sheet changes were done in this PR to avoid the worst offenders to make the UI look quite similar to other DCCs - yet there are still some differences. And likely the changes made aren't done correctly either.

### Having Qt Windows open during shutting down of Substance Painter will shut down incorrectly

If any of the Qt tools remain open when shuttig down Substance 3D Painter then the next time it'll launch it will start with a Bug Report pop-up saying "Substance 3D Painter didn't shut down correctly."

afbeelding

Not entirely sure how to best close down all OpenPype tool UIs on Substance Painter shutdown to avoid this issue.

This can also be reproduced by running this in Substance's Python Console without OpenPype:

```python
from PySide2 import QtWidgets; import substance_painter.ui; dialog = QtWidgets.QDialog(parent=substance_painter.ui.get_main_window()); dialog.show()
```

### Publishing gets slower over time

It seems that Substance Painter and printing/logging gets much slower over time when there are many logs present. This means that when you've published many times in one session you'll definitely notice the publishing being slower. This is solved by opening a Log window and clearing the log, then everything is much faster again.

A potential workaround to look for in the future is to either Clear the logs regularly or just print/log much much less information into Substance Painter itself and keep it solely to the Publisher UI as described [here](https://github.com/ynput/OpenPype/issues/3800).

## Testing notes:

I tested on Windows with Adobe Substance Painter 8.2.0 build 1987

_Note: You might need to enable the OpenPype Integration plug-in at the top menu "Python > openpype_plugin"_

afbeelding

1. Launch Substance Painter
2. Use the Loader to 'start' a project with e.g. an FBX or Alembic mesh using "Load Mesh" **OR** open your own New Project template and then use the Loader to replace the mesh to that.


BigRoy:
### Preview screenshots

afbeelding

afbeelding

afbeelding

BigRoy:
> maybe silly question...I have only Steam edition of the Painter. Do you think it will work anyway or we need proper Adobe package?

Good question. I have no idea actually - I suspect it should. Steam mentions it only has updates until end of 2022. Painter build 8.2.0 I tested with came out beginning of December 2022 so I suppose it would have that update too.

LiborBatek:
Ok, I have just my private (for my projects) steam edition and what worse its even old one :)

heres a scrncap:
image

I have configured it in OP settings and it starts normally via OP Launcher however the OP menu doesnt show up in UI...

Is it chance to get it working even in this older ver or complete scifi and I need at least 8.x version of Painter?

Let me know and I will resolve this with @mkolar  if needed!

BigRoy:
Here are some previews of a very early WIP setup for publishing textures. Not fully functional yet:

create_textures_templates
create_textures_instance
extract_texture_sets

_Be aware that there's still the Qt StyleSheet issue so the UI might be a bit odd here and there!!!_

Still trying to find a better way in [Substance Painter on how to get all available export preset in Python](https://community.adobe.com/t5/substance-3d-painter-discussions/python-access-to-output-templates-in-texture-export-window/td-p/12984047).

Also still trying to define the right "family" for publishing textures/images as a group, somewhat related to this [Texture Sets publishing discussion.](https://github.com/ynput/OpenPype/discussions/4286) **Would love input on this!**

BigRoy:
@LiborBatek I believe it should work with that version too - however, you might need to enable the plugin first under the **Python** top menu entry. Does it show the plug-in there?

It's like this screenshot:

afbeelding

LiborBatek:
@BigRoy unfortunately not, heres my scrn grab (also tried to trigger those menu items if anything happen)

image

BigRoy:
Just wanted to follow up here - I've debugged with @LiborBatek to see if we could get the Steam version up and running - and got it to work ✅ 

The issue however is that Substance Painter inherits the environment variable from the Steam executable process instead of the one we're passing directly to the Substance Painter executable from `"C:\Program Files (x86)\Steam\steamapps\common\Substance Painter 2021\Adobe Substance 3D Painter.exe"`

The workaround for now is to close down Steam - then launch from Substance Painter from the OpenPype launcher. This will start the steam process _including the required environment_ for that particular process. However, this would mean Steam would need to be closed down _each time_ painter gets launched from OpenPype to ensure it has the right environment passed along.

There must be a better way! I've posted the question in some other areas to see if others might be familiar with working around this with Steam apps.

openpype/hosts/substancepainter/api/pipeline.py
```diff
@@ -0,0 +1,328 @@
...
+CREATE_PATH = os.path.join(PLUGINS_DIR, "create")
+INVENTORY_PATH = os.path.join(PLUGINS_DIR, "inventory")
+
+
+OPENPYPE_METADATA_KEY = "OpenPype"
+OPENPYPE_METADATA_CONTAINERS_KEY = "containers"  # child key
+
+self = sys.modules[__name__]
+self.menu = None
+self.callbacks = []
```
iLLiCiTiT, line 43: Why not store them to object of `SubstanceHost`?

LiborBatek:
Did testing in Substance 3D Painter (Steam Edition) Version 7.4.3. build 1610 and all works as expected.

Were testing workfiles saving/loading and publishing. Also tried switch asset versions. All fine.

Small note regarding OP menu, should contain ```create``` item and jump to create shelf in publish window 
image

openpype/hosts/substancepainter/api/pipeline.py
BigRoy: Should that store it to the class of `SubstanceHost` or can it store it on the instance of `SubstanceHost`? :) I don't the instance itself stays around - or does the actual Host instance itself get stored as the `registered_host`?

BigRoy:
> Small note regarding OP menu, should contain create item and jump to create shelf in publish window

Is that true? 

- I don't see it in Houdini - there is a create entry but it doesn't point to the new publisher (yet?)
- Code of Unreal Engine also seems to point to the old creator?
- `hosts/max/api/menu.py` doesn't have a create.. menu entry 
- Same goes for Nuke

Or am I misreading the codebase?

LiborBatek:
Well maya also have ```create``` menu item but originally this note came from @jakubjezek001  as he implemented it the mentioned Nuke PR. So maybe he should say?

BigRoy:
> Well maya also have create menu

Correct - but that's the old publishing behavior (prior to the new publisher). Let me know what the choice will be and I'll refactor the code.

BigRoy:
I've pushed some updates.

### First launch now loads the OpenPype plug-in

If Substance Painter's has found no stored settings about the preferred plug-in state for OpenPype's integration it will now load on first start-up. This should make the onboarding easier for new users since they wouldn't need to first load the plugin through the top Python menu as [explained here](https://github.com/ynput/OpenPype/pull/4283#issuecomment-1375985414)

If the user has ever disabled the plug-in previously it will however respect that. It might be worth it to add an extra "force load" setting to OpenPype to always load the plug-in. But if that's what we want we might just be better of changing the loading structure to a start-up plug-in (which can't be toggled off) instead. Thoughts?

### Add Substance Painter Asset library shelves through Project Settings

There is now the ability to define custom asset shelves locations for users in project settings. This will make it easier to share e.g. Brushes, Materials, Export Presets across the team in a production. You could set up one shelf for global studio assets and for example _one per project*_

*anatomy formatting isn't implemented yet - if someone could point me to an easy example on how to do so then I'd be happy to implement that.

afbeelding

**NOTE**
Since this is a dynamically loaded shelf, be aware of [these usability notes](https://github.com/ynput/OpenPype/blob/d2baa5ec4d9f92c143172f95719bb7b319ae79a2/openpype/hosts/substancepainter/api/lib.py#L12-L22).

Additionally be aware that if the path exists and gets loaded as a shelf that Substance Painter will automatically generate its assets folder structure into that folder.

BigRoy:
Have been working some more on getting publishing of textures working together with loading them in Maya.
Here's one of my test runs:

https://user-images.githubusercontent.com/2439881/212285701-c5351ba8-35cb-4015-9118-6552837ff5d6.mp4

BigRoy:
I've pushed the changes that I used to generate the publish in the [earlier demo video](https://github.com/ynput/OpenPype/pull/4283#issuecomment-1381533722) I shared. You're more than welcome than test it out!

It would be extra nice if someone could have a good look at the code changes to see what in particular might stand out. Definitely point out if some parts of the code seem confusing or need better documentation/docstrings, etc.

---

### Not ready for production

**Warning:** The subset names to be published from the textures are still WIP - the families, workflow and structure might likely still change so I'd recommend **NOT** using this in production yet because in time likely what you had published will become incompatible with what there is currently.

---

There's still loads of work to be done, but it does seem we're headed in a decent direction for a first draft state.

@iLLiCiTiT @mkolar A good glance over the code base would be nice - and also maybe come back with some feedback on the publishing of the texture. Currently the "Collector" makes the instance branch out into new instances of its own which get integrated as their own subsets. These separated instances don't show in the Create / Publish pages of the new publisher but they do show in the Details page if you'd want to debug things in that area. 

afbeelding


afbeelding


@LiborBatek are you able to set up a test run to see if any of the changes break things on your end? Your particular version of Substance Painter doesn't support OCIO yet for example and I wonder whether it'll automatically break support with that version with any changes I've made.

BigRoy:
Pushed some updates, some cleanup here and there. We're getting somewhere.

### Color Space Management test

I had also set up a test using the #4195 codebase, it's in a different branch: https://github.com/BigRoy/OpenPype/tree/substance_painter_ocio

@jakubjezek001 If you're interested in seeing how I ended up using your color space API check [the branch](https://github.com/BigRoy/OpenPype/tree/substance_painter_ocio) - but it's basically only [this commit](https://github.com/BigRoy/OpenPype/commit/33aafc3ff6f7e1b4f213345e7baa80f50d4e1f51) where I've put in the changes. Let me know if I did anything wrong.

_Note: I did have to fix [this bug](https://github.com/ynput/OpenPype/pull/4195#discussion_r1070241828) locally during testing otherwise it failed to publish. It'd pop my two images in the UDIM tile sequence in the integrator to one image - which the integrator didn't like._

LiborBatek:
Did one more test with my version of SubstancePainter (7.4.3 Steam ver) and it stopped working... when triggered the Substance in OP Launcher it pops up the splash scrn of Painter but in a while dissappear and nothing happens then - it just doesnt start.

Here is scrncap from my shell
image

P.S. I have turned off the Steam app before launching the app. When tested with Steam app running it starts but without OP menu again...

BigRoy:
> Did one more test with my version of SubstancePainter (7.4.3 Steam ver) and it stopped working... when triggered the Substance in OP Launcher it pops up the splash scrn of Painter but in a while dissappear and nothing happens then - it just doesnt start.
> 
> Here is scrncap from my shell image
> 
> P.S. I have turned off the Steam app before launching the app. When tested with Steam app running it starts but without OP menu again...

Thanks - unfortunately this will be hard to debug since the shell logs show no related output. @LiborBatek  Could you try and remove this [`openpype_load_on_startup.py`](https://github.com/ynput/OpenPype/blob/1c77d2b002527a450c8be21d93040bccd588413e/openpype/hosts/substancepainter/deploy/startup/openpype_load_on_first_run.py) file and run again? Maybe it has to do with that logic which runs during startup.

openpype/hosts/substancepainter/plugins/publish/collect_textureset_images.py
```diff
@@ -0,0 +1,152 @@
...
+        fnames = [os.path.basename(output["filepath"]) for output in outputs]
+        ext = os.path.splitext(first_filepath)[1]
+        assert ext.lstrip('.'), f"No extension: {ext}"
+        map_identifier = strip_template(template)
+
+        # Define the suffix we want to give this particular texture
+        # set and set up a remapped subset naming for it.
+        suffix = f".{map_identifier}"
+        image_subset = instance.data["subset"][len("textureSet"):]
+        image_subset = "texture" + image_subset + suffix
```
BigRoy, line 57: @mkolar Obviously this is a very nasty way of doing this since subset names can actually be dynamic based on the template - this breaks that logic.

Do you have any ideas on how to potentially solve this best?

My idea is that the instance itself is e.g. `textureSetMain` or `texturesMain` or `texturePackMain` (which would you prefer?) and that the individual images would be a 'sub-variant' of this like `textureSetMain.Roughness` (which is an Image family) or alike. However, if the template would be based on the family it'd likely become `imageMain.Roughness` which is likely confusing to the user too.

Could also supply the exact same 'dynamic subset name formatting' as the parent instance but ONLY add `.{map_identifier}` to **_variant_** and then using that let the templates do its thing.

If the user had `{Variant}_{family}` as template then it'd for example be `Main.Roughness_textureSet`.

openpype/hosts/substancepainter/api/pipeline.py
```diff
@@ -0,0 +1,347 @@
...
+    def _deregister_callbacks(self):
+        for event, callback in self.callbacks:
+            substance_painter.event.DISPATCHER.disconnect(event, callback)
+        self.callbacks.clear()
+
+    def _install_shelves(self, project_settings):
+
+        shelves = project_settings["substancepainter"].get("shelves", {})
+        for name, path in shelves.items():
+            # TODO: Allow formatting with anatomy for the paths
```
BigRoy, line 223: @antirotor Any pointers on how to best format a path using anatomy or alike during the host installation like this? :)

```python

from openpype.pipeline.template_data import get_template_data_with_names

anatomy_data = get_template_data_with_names(project_name, asset_name, task_name, system_settings)
path = anatomy.format(anatomy_data)
```
This doesn't seem like the right approach since it's not formatting against a template in anatomy but using a custom path. I need some way to specifically format the custom path. Some pseudocode:

```python
from openpype.pipeline.template_data import get_template_data_with_names

template_path = shelf["path"]
anatomy_data = get_template_data_with_names(project_name, asset_name, task_name, system_settings)
path = api.format(template_path, data=anatomy_data)
```

BigRoy:
@mkolar @iLLiCiTiT What state of this PR you feel would make this **Ready for Review** and basically be ready to get merged? Let me know if there's particular things I'd need to focus on. Some things are still draft, e.g. currently of texture set instances you can only create one persistently but that's just tweaking how to store the data.

- Do we need to solve any Steam-related Substance issues?
- Should it be OCIO ready? (It'd require #4195 to be merged first?)
- Are we expecting certain validations to be in-place?

Some other ideas that I thought of:

- Should we add a Project Settings which defines the **default** export preset shown in the TextureSet Creator?
- Should we add Project Settings so that a studio could filter what Export Presets are selectable to begin with (or only a Validator for that?), e.g. they could say _allow only RedshiftMaterial_template_ or alike.

LiborBatek:
> * Do we need to solve any Steam-related Substance issues?
> * Should it be OCIO ready? (It'd require [Colorspace Management and Distribution #4195](https://github.com/ynput/OpenPype/pull/4195) to be merged first?)


Speaking of OCIO color management being fully working  ...From user pov is not neccessary because the textures will be outputted in already set color space anyway (when OCIO and for ie ACES set properly in Painter)...so thet will be correct and ready to use down the pipeline no matter if fully OCIO enviroment working... so its not s Show stopper imho!

LiborBatek:
@BigRoy  Unfortunatelly removing ```openpype_load_on_first_run.py``` didnt help and Substance Painter didnt started up...

There is splash screen showing up and then dissappearing again and nothing happens then. Speaking of Substance 7.4.3 (Steam Edition)

jakubjezek001:
> Should it be OCIO ready? (It'd require https://github.com/ynput/OpenPype/pull/4195 to be merged first?)

So it is merged already so it would be good to support it.

BigRoy:
> So it is merged already so it would be good to support it.

I've merged in the OCIO test branch I had set up - as such this should now also register the color space with the extracted files I believe.

BigRoy:
Not sure who to best ping - but would love to get a review in on this code here.

moonyuet:
@BigRoy , can you update this branch also for https://github.com/ynput/OpenPype/pull/4313 so that i can do some test?

BigRoy:
@moonyuet I've merged in latest develop but I'm unable to test the branch myself now so let me know if anything is broken.

openpype/hosts/substancepainter/api/pipeline.py
```diff
@@ -0,0 +1,347 @@
...
+        metadata = substance_painter.project.Metadata(OPENPYPE_METADATA_KEY)
+        containers = metadata.get(OPENPYPE_METADATA_CONTAINERS_KEY)
+        if containers:
+            for key, container in containers.items():
+                container["objectName"] = key
+                yield container
+
+    @staticmethod
+    def create_context_node():
+        pass
```
moonyuet, line 145: I am not sure this function is similar to the `set_context_settings`. If yes, I am not certain what to set up within the limited arguments as far as I read the doc of SP's python api
```
substance_painter.project.Settings(default_save_path: Optional[str] = None, normal_map_format: Optional[[NormalMapFormat](https://github.com/ynput/OpenPype/pull/4283/files#substance_painter.project.NormalMapFormat)] = None, tangent_space_mode: Optional[[TangentSpace](https://github.com/ynput/OpenPype/pull/4283/files#substance_painter.project.TangentSpace)] = None, project_workflow: Optional[[ProjectWorkflow](https://github.com/ynput/OpenPype/pull/4283/files#substance_painter.project.ProjectWorkflow)] = None, export_path: Optional[str] = None, default_texture_resolution: Optional[int] = None, import_cameras: Optional[bool] = None, mesh_unit_scale: Optional[float] = None)
```
If we really follow the API, only thing we may think is to set up the resolution, but I dont really know.
```suggestion
        substance_painter.project.Settings(default_texture_resolution=2048)
```

openpype/hosts/substancepainter/api/pipeline.py
iLLiCiTiT: Anatomy in fact is just templates, roots and attributes + helpers methods to work with them. So I would just use `anatomy.roots` in fill data, but don't use anatomy to format the template.

```python
from openpype.lib import StringTemplate
from openpype.pipeline.template_data import get_template_data_with_names

template_path = shelf["path"]
fill_data = get_template_data_with_names(project_name, asset_name, task_name, system_settings)
fill_data["root"] = anatomy.roots
path = StringTemplate.format_template(template_path, anatomy_data)
```

openpype/hosts/substancepainter/plugins/create/create_workfile.py
```diff
@@ -0,0 +1,95 @@
...
+    def create(self):
+
+        if not substance_painter.project.is_open():
+            return
+
+        variant = self.default_variant
+        project_name = self.project_name
+        asset_name = legacy_io.Session["AVALON_ASSET"]
+        task_name = legacy_io.Session["AVALON_TASK"]
+        host_name = legacy_io.Session["AVALON_APP"]
```
iLLiCiTiT, line 34:
```suggestion
        asset_name = self.create_context.get_current_asset_name()
        task_name = self.create_context.get_current_task_name()
        host_name = self.create_context.host_name
```

BigRoy:
Sorry about that - had a little fun breaking my git setup here. :) The latest commit should be identical to the one before the force push.

moonyuet:
Publish function looks good with the updated collector with udim format
image
I have several comments on the loaders:
1. Do you think it's a good idea to introduce load_sbsar in the loader? Although it is still super early to ask(as the SD integration is not yet here), it is the way I used with my previous job when I built something from substance designer and sampler to substance painter. 
2. Or a loader to import baked maps from both SD and SP and an extractor for exporting baked maps from SP.

BigRoy:
> I have several comments on the loaders:
> 
>     1. Do you think it's a good idea to introduce load_sbsar in the loader? Although it is still super early to ask(as the SD integration is not yet here), it is the way I used with my previous job when I built something from substance designer and sampler to substance painter.
> 
>     2. Perhaps  a loader to import baked maps from both SD and SP and an extractor for exporting baked maps from SP?

I think those sound like great ideas - but could technically be done outside of this PR without requiring changes to the core logic of how Substance Painter integration behaves. If so, let's create separate issues for those.

I did have some open todos that I think should be solved:

1. [ ] When publishing an output template that exports maps for which you currently do not have the channels active in the project the integrator will fail since those textures are not exported and thus do not exist. This should still be fixed. (This was reported by @ddesmond)
    - The question here is do we want to validate it and invalidate the publish if some output maps won't be output or do we want to let it pass but just skip those maps that wouldn't be exported by Substance Painter. Do we want to force all the maps of output template to be exported or are we fine with e.g. a `normal` map not being included?
3. [ ] I believe the current logic does not support having different materials inside a single Substance Painter project to have them exported into the single texture set package. 
    - The question would be, do we want to split up the texture outputs even more so that it becomes `Main.{material}.{outputMap}` for the variant and thus by default the subset name becomes `textureMain.{material}.{outputMap}` which would generate massively long subset names. Or would we rather want each texture set to at most export a SINGLE material from the file so that you can explicitly choose the output name for each material. However, this would leave much more of a strain on the artist if you have a substance painter file with many separate objects/materials. 
 
@moonyuet @ddesmond @gabormarinov @LiborBatek @maxpareschi - any preferences on this? Also tagging you all since you've each shown interest in the Substance Painter integration so if you've any capability and availability to test the PR that'd be great.

@gabormarinov I believe you've had a workflow already where you've used the Standalone Publisher to publish textures that came from Substance Painter. What was the resulting subsets that it published for you? That'd be a great reference as to whether what this integration does matches what you already tried to 'mimic' in standalone.

I also saw this screenshot on discord from @maxpareschi on publishing "Texture Batches" (which is something I've not used so far) - so any production knowledge gained with using that would be welcome as input here too about how the output of substance painter should get integrated into OpenPype:

IMG_1604

gabormarinov:
@BigRoy great to see, that this is coming together! Regarding your questions:

- When publishing an output template that exports maps for which you currently do not have the channels active in the project the integrator: in my opinion it should not fail as we use a few templates, and a lot of times some of the channels not needed, so there's no map for them (eg. opacity, normal). So it would be the best if it could give you a warning, but should not stop the publishing process.

- Different materials inside a single Substance Painter project: we often use multiple materials in a single SP project, and would be very happy with splitting up the texture outputs as you suggested Main.{material}.{outputMap} for the variant and thus by default the subset name becomes textureMain.{material}.{outputMap} . It would easy the organization of the textures.

Answering your question about Standalone Publisher and Texture batches - we abandoned this approach, as it was to complex for the artists. Now we use the "trick" as I described on Discord naming SP project files v001, v002, etc. inside the texture task resources folder. Then we export textures into v001, v002, etc.  subfolders with a SP template which uses $project token (project name), so the folder name and version number in texture files generated automatically.  The Lookdev task loads textures from there, publishing the lookdev packs the textures into a versioned lookdev anyway. It is poor man's texture versioning, I hope this SP integration will replace it soon!

openpype/hosts/substancepainter/api/lib.py
```diff
@@ -0,0 +1,648 @@
...
+
+    print(f"Adding Shelf '{name}' to path: {path}")
+    substance_painter.resource.Shelves.add(name, path)
+
+    return name
+
+
+def _get_new_project_action():
+    """Return QAction which triggers Substance Painter's new project dialog"""
+    from PySide2 import QtGui
```
iLLiCiTiT, line 527:
```suggestion
    from qtpy import QtGui
```

iLLiCiTiT: BTW: Why are these imports inside functions?

openpype/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -0,0 +1,123 @@
...
+
+    families = ["*"]
+    representations = ["abc", "fbx", "obj", "gltf"]
+
+    label = "Load mesh"
+    order = -10
+    icon = "code-fork"
+    color = "orange"
+
+    options = [
```
iLLiCiTiT, line 27: You can use attribute definitoins from openpype.lib here (Not requirement).

openpype/hosts/substancepainter/api/lib.py
BigRoy: Haha good question. I might have tested building the functions elsewhere and copied them in. Will clean it up.

openpype/hosts/substancepainter/plugins/publish/extract_textures.py
```diff
@@ -0,0 +1,71 @@
...
+                continue
+
+            for representation in image_instance.data["representations"]:
+                self.set_representation_colorspace(representation,
+                                                   context=context,
+                                                   colorspace=colorspace)
+
+        # Add a fake representation which won't be integrated so the
+        # Integrator leaves us alone - otherwise it would error
+        # TODO: Add `instance.data["integrate"] = False` support in Integrator?
```
iLLiCiTiT, line 63: 👍

openpype/settings/defaults/project_settings/substancepainter.json
```diff
@@ -0,0 +1,13 @@
...
+            "enabled": true,
+            "filepath": []
+        },
+        "file_rules": {
+            "enabled": true,
+            "rules": {}
+        }
+    },
+    "shelves": {}
+}
```
iLLiCiTiT, line 13: We've changed it to have empty line at the end
```suggestion
}

```

openpype/hosts/substancepainter/plugins/create/create_textures.py
```diff
@@ -0,0 +1,118 @@
...
+    def get_instance_attr_defs(self):
+
+        return [
+            EnumDef("exportPresetUrl",
+                    items=get_export_presets(),
+                    label="Output Template"),
+            EnumDef("exportFileFormat",
+                    items={
+                        None: "Based on output template",
+                        # TODO: implement extensions
```
iLLiCiTiT, line 77: Could this be defined using settings? If substance have limitations, than maybe multiselection enum?

BigRoy: It could technically be implemented in settings. We could also always enforce it to be based on output template (which it is now due to not having the extensions enum) - which means the admin can just make sure the output templates are as required for production. 👍

openpype/hosts/substancepainter/deploy/plugins/openpype_plugin.py
```diff
@@ -0,0 +1,36 @@
...
+
+def cleanup_openpype_qt_widgets():
+    """
+        Workaround for Substance failing to shut down correctly
+        when a Qt window was still open at the time of shutting down.
+
+        This seems to work sometimes, but not all the time.
+
+    """
+    # TODO: Create a more reliable method to close down all OpenPype Qt widgets
```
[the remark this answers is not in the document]
BigRoy: If that works reliably without crashes let's give it a go. For what it's worth it might even resolve #3453 ?

antirotor:
Regarding this:

> The question here is do we want to validate it and invalidate the publish if some output maps won't be output or do we want to let it pass but just skip those maps that wouldn't be exported by Substance Painter. Do we want to force all the maps of output template to be exported or are we fine with e.g. a normal map not being included?

I think we should validate it, but with optional validator that could be disabled by artist if he really wants to continue even without all exports.

> The question would be, do we want to split up the texture outputs even more so that it becomes Main.{material}.{outputMap} for the variant and thus by default the subset name becomes textureMain.{material}.{outputMap} which would generate massively long subset names. Or would we rather want each texture set to at most export a SINGLE material from the file so that you can explicitly choose the output name for each material. However, this would leave much more of a strain on the artist if you have a substance painter file with many separate objects/materials.

This is long standing issue that we are constantly hitting. This could be potentially helped by #2724. Also by finally allowing representation to contain multiple types of files (I am wondering why we are missing issue for that on GH).

Long subset names are IMHO better than more strain on artist (and it can be alleviated by UI grouping or something like above mentioned #2724).

openpype/hosts/substancepainter/plugins/publish/save_workfile.py
```diff
@@ -0,0 +1,27 @@
...
+    """Save current workfile"""
+
+    label = "Save current workfile"
+    order = pyblish.api.ExtractorOrder - 0.49
+    hosts = ["substancepainter"]
+
+    def process(self, context):
+
+        host = registered_host()
+        if context.data['currentFile'] != host.get_current_workfile():
```
antirotor, line 19:
```suggestion
        if context.data["currentFile"] != host.get_current_workfile():
```

openpype/hosts/substancepainter/plugins/publish/extract_textures.py
```diff
@@ -0,0 +1,64 @@
...
+    Note:
+        This Extractor assumes that `collect_textureset_images` has prepared
+        the relevant export config and has also collected the individual image
+        instances for publishing including its representation. That is why this
+        particular Extractor doesn't specify representations to integrate.
+
+    """
+
+    label = "Extract Texture Set"
+    hosts = ['substancepainter']
```
antirotor, line 18:
```suggestion
    hosts = ["substancepainter"]
```

openpype/hosts/substancepainter/plugins/publish/collect_workfile_representation.py
```diff
@@ -0,0 +1,26 @@
...
+        current_file = context.data["currentFile"]
+
+        folder, file = os.path.split(current_file)
+        filename, ext = os.path.splitext(file)
+
+        instance.data['representations'] = [{
+            'name': ext.lstrip("."),
+            'ext': ext.lstrip("."),
+            'files': file,
+            "stagingDir": folder,
```
antirotor, line 25:
```suggestion
        instance.data["representations"] = [{
            "name": ext.lstrip("."),
            "ext": ext.lstrip("."),
            "files": file,
            "stagingDir": folder,
```

openpype/hosts/substancepainter/plugins/publish/collect_workfile_representation.py
```diff
@@ -0,0 +1,26 @@
+import os
+import pyblish.api
+
+
+class CollectWorkfileRepresentation(pyblish.api.InstancePlugin):
+    """Create a publish representation for the current workfile instance."""
+
+    order = pyblish.api.CollectorOrder
+    label = "Workfile representation"
+    hosts = ['substancepainter']
```
antirotor, line 10:
```suggestion
    hosts = ["substancepainter"]
```

openpype/hosts/substancepainter/plugins/publish/collect_textureset_images.py
```diff
@@ -0,0 +1,177 @@
...
+        # Clone the instance
+        image_instance = context.create_instance(instance.name)
+        image_instance[:] = instance[:]
+        image_instance.data.update(copy.deepcopy(instance.data))
+        image_instance.data["name"] = image_subset
+        image_instance.data["label"] = image_subset
+        image_instance.data["subset"] = image_subset
+        image_instance.data["family"] = "image"
+        image_instance.data["families"] = ["image", "textures"]
+        image_instance.data['representations'] = [representation]
```
antirotor, line 108:
```suggestion
        image_instance.data["representations"] = [representation]
```

openpype/hosts/substancepainter/plugins/publish/collect_textureset_images.py
```diff
@@ -0,0 +1,177 @@
...
+            project_name=context.data["projectName"],
+            host_name=context.data["hostName"],
+            project_settings=context.data["project_settings"]
+        )
+
+        # Prepare representation
+        representation = {
+            'name': ext.lstrip("."),
+            'ext': ext.lstrip("."),
+            'files': fnames if len(fnames) > 1 else fnames[0],
```
antirotor, line 83:
```suggestion
            "name": ext.lstrip("."),
            "ext": ext.lstrip("."),
            "files": fnames if len(fnames) > 1 else fnames[0],
```

openpype/hosts/substancepainter/plugins/publish/collect_textureset_images.py
```diff
@@ -0,0 +1,177 @@
...
+
+        The new instances will be of family `image`.
+
+        """
+
+        context = instance.context
+        first_filepath = outputs[0]["filepath"]
+        fnames = [os.path.basename(output["filepath"]) for output in outputs]
+        ext = os.path.splitext(first_filepath)[1]
+        assert ext.lstrip('.'), f"No extension: {ext}"
```
antirotor, line 58:
```suggestion
        assert ext.lstrip("."), f"No extension: {ext}"
```

But I am not sure about this assert, if there is a changes there might not be extension, shouldn't we throw some Exception instead?

openpype/hosts/substancepainter/plugins/publish/collect_textureset_images.py
```diff
@@ -0,0 +1,177 @@
...
+from openpype.client import get_asset_by_name
+
+
+class CollectTextureSet(pyblish.api.InstancePlugin):
+    """Extract Textures using an output template config"""
+    # TODO: Production-test usage of color spaces
+    # TODO: Detect what source data channels end up in each file
+
+    label = "Collect Texture Set images"
+    hosts = ['substancepainter']
```
antirotor, line 22:
```suggestion
    hosts = ["substancepainter"]
```

openpype/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -0,0 +1,123 @@
...
+        settings = substance_painter.project.MeshReloadingSettings(
+            import_cameras=container_options.get("import_cameras", True),
+            preserve_strokes=True
+        )
+
+        def on_mesh_reload(status: substance_painter.project.ReloadMeshStatus):
+            if status == substance_painter.project.ReloadMeshStatus.SUCCESS:
+                print("Reload succeeded")
+            else:
+                raise RuntimeError("Reload of mesh failed")
```
antirotor, line 108:
```suggestion
                raise LoaderError("Reload of mesh failed")
```
might be better here? (from `openpype.pipeline.load.LoadError`)

openpype/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -0,0 +1,123 @@
...
+            settings = substance_painter.project.MeshReloadingSettings(
+                import_cameras=import_cameras,
+                preserve_strokes=preserve_strokes
+            )
+
+            def on_mesh_reload(status: substance_painter.project.ReloadMeshStatus):  # noqa
+                if status == substance_painter.project.ReloadMeshStatus.SUCCESS:  # noqa
+                    print("Reload succeeded")
+                else:
+                    raise RuntimeError("Reload of mesh failed")
```
antirotor, line 65:
```suggestion
                    raise LoadError("Reload of mesh failed")
```

antirotor (commented):
Just a few quote-style notes and exception raising, otherwise the code looks solid. Can't test it as I don't have access to Substance Painter

antirotor (dismissed):
Looks cool. Can we test it @LiborBatek ?

LiborBatek:
@antirotor we need license for Adobe Substance to test it....I have only my Steam edition of older Painter 7.4.3 which doesnt work with OP. No OpenPype plugin shown in the addon menu...

Will be on it asap....

BigRoy:
@moonyuet @ddesmond @gabormarinov @maxpareschi  @fabiaserra If any of you are using Substance Painter in production or have experience with it then it would be super if you're able to test this PR whether it works as you'd need for your production cases. Being able to comment now before we get into merging this PR would make sure the core functionality fits your production needs.

LiborBatek:
There are still some questionable things too speaking of Substance integration... the core is functional tho!

1) TextureSet file naming differs from Substance Project and is hardcoded by OP?

image


2) Maya warns that color profile isnt available even tho its set to the same ```Aces cg``` via OCIO config?! Not sure whats goin on there

image

3) Its great user can ```Update Asset``` and gets warning of obsolete asset for ```TextureSet``` much appreciate feature! Great!
However its not later visible in ```Scene Manager``` ...bug? limitation?

BigRoy:
> TextureSet file naming differs from Substance Project and is hardcoded by OP?

Correct. The output behaves according to the output templates for the subset. If you need to preserve the source name I believe you should define a custom template and use the `{sourceName}` token or alike instead.

> Maya warns that color profile isnt available even tho its set to the same Aces cg via OCIO config?! Not sure whats goin on there

That's definitely something to investigate - thanks. I'll look into it. But in your Maya you are using the Maya OCIO config and that is almost 100% certain not the OCIO config you're using in Substance Painter. It seems your Maya isn't correctly color managed to the same OCIO config.

This only shows more and more that at least managing `OCIO` env var like @fabiaserra mentioned in #4700 [here](https://github.com/ynput/OpenPype/pull/4700#issuecomment-1485547686) and [here](https://github.com/ynput/OpenPype/pull/4700#issuecomment-1493436885) will result in a much more artist friendly behavior since it's much harder to get maya into a state where it won't be using the correct OCIO config since it'll always listen to the env var.

For now could you try manually setting the same OCIO path as Substance Painter ended up using in Maya to ensure it has access to the exact same color space names. I think maybe the color space is slightly different named in the Maya config, maybe `acescg` instead of `ACES - ACEScg`.

> Its great user can Update Asset and gets warning of obsolete asset for TextureSet much appreciate feature! Great!
However its not later visible in Scene Manager ...bug? My bad.....it works flawlessly! Had filter outdated on :) so nevermind!

What a rollercoaster of emotions there. Thanks for testing, and great to hear it's working.

LiborBatek:
So in the end, all core functions is working well! Now if any other users suggestions arises...

@BigRoy  Welldone, man!

BigRoy:
> I also understand the OCIO issue in maya as you described it. So I guess right now its more of a problem that maya host doesnt use OCIO config from the same place and setting (OP should dictate it right?)

It should be, yes. If not working correctly then it's an issue on the Maya side which MIGHT be solved with https://github.com/ynput/OpenPype/pull/4700 but it might be best to report it in that PR if the OP setting doesn't correctly propagate into Maya for you so it could be resolved with that PR.

mkolar (approved):
Seems to me that this might be ready for merge right?

openpype/hosts/substancepainter/plugins/publish/collect_textureset_images.py
```diff
@@ -0,0 +1,196 @@
...
+            # More than one stack, include stack name
+            suffix += f".{stack_name}"
+
+        # Always include the map identifier
+        map_identifier = strip_template(template)
+        suffix += f".{map_identifier}"
+
+        image_subset = get_subset_name(
+            # TODO: The family actually isn't 'texture' currently but for now
+            #       this is only done so the subset name starts with 'texture'
```
BigRoy, line 84: I'd say the only thing maybe @mkolar is this here. Any chance for subset names later will have backwards compatibility issues so wanted to note this now whether this needs changing.

openpype/hosts/substancepainter/plugins/publish/validate_ouput_maps.py
```diff
@@ -0,0 +1,110 @@
...
+
+            if allow_skipped_maps:
+                # TODO: This is changing state on the instance's which
+                #   should not be done during validation.
+                self.log.warning(f"Disabling texture instance: "
+                                 f"{image_instance}")
+                image_instance.data["active"] = False
+                image_instance.data["integrate"] = False
+                representation.setdefault("tags", []).append("delete")
+                continue
```
BigRoy, line 86: This is disabling maps that do not generate inputs if it's allowed for the instance to ignore those. It's a bit of an odd place for this logic. I'd say it's fine for a merge now since moving it shouldn't break backwards compatibility it would just end up being a bit cleaner.

BigRoy (commented):
@mkolar I think this is pretty much ready to merge. However I've placed some comments now that I thought would be good to highlight before we do so.

antirotor (approved):

openpype/hosts/substancepainter/plugins/publish/collect_textureset_images.py
antirotor: I vote for the *dynamic subset name formatting*, but that can be tweaked later on with additional PRs

PR #4291, opened 2023-01-06
Global: color v3 global oiio transcoder plugin

kalisp:
## Brief description
Implements possibility to use `oiiotool` to transcode image sequences from one color space to another(s).

## Description
Uses collected `colorspaceData` information about source color spaces, these information needs to be collected previously in each DCC interested in color management.
Uses profiles configured in Settings to create single or multiple new representations (and file extensions) with different color spaces.

New representations might replace existing one, each new representation might contain different tags and custom tags to control its integration step.

## Additional info
This PR is built on top of https://github.com/ynput/OpenPype/pull/4270.

## Testing notes:
1. Set File rules in DCC which has currently Color management implemented (Nuke??) `project_settings/nuke/imageio/file_rules/rules`
2023-01-06 18_33_01-Window
3. Configure output definitions for converted files `project_settings/global/publish/ExtractColorTranscode`

2023-01-06 18_39_40-OpenPype Settings (this says "For all representations from Nuke (which have `colorspaceData` prefilled) create new representation with files with same extension and convert it to `ACES` colorspace and send it to `ExtractReview`)

5.  Check if final files are in proper color space.


tokejepsen:
Is this dependent on #4195 ?

tokejepsen:
Sorry, nvm :) This PR goes into #4195

tokejepsen:
Why do we need the extra layer of the file rules?
Seems to me that the extractor will only run on representations that has `colorspace` data, so setting up the file rules seems redundant.

openpype/lib/transcoding.py
```diff
@@ -1037,3 +1037,59 @@ def convert_ffprobe_fps_to_float(value):
...
+
+        # Add last argument - path to output
+        base_filename = os.path.basename(input_path)
+        output_path = os.path.join(output_dir, base_filename)
+        oiio_cmd.extend([
+            "-o", output_path
+        ])
+
+        logger.debug("Conversion command: {}".format(" ".join(oiio_cmd)))
+        run_subprocess(oiio_cmd, logger=logger)
```
BigRoy, line 1095: This function makes more sense if it takes a single file instead of a list of filepaths.
It doesn't seem to do much optimization for running over multiple filepaths to warrant it to run over multiple files.

Additionally I think it's somewhat odd that you're specificying an output directory instead of an output filename, especially because it doesn't return the output filepaths I feel this is a bit odd.

Also other methods seem to name the color space argument as `colorspace` instead of `color_space` so I'd do that here too.

```python
convert_colorspace(filepath, out_filepath, config_path, source_colorspace, target_colorspace)
```

I'd also have it return the filepath it generated - just for clarity.

openpype/plugins/publish/extract_color_transcode.py
```diff
@@ -0,0 +1,212 @@
...
+                if not isinstance(files_to_convert, list):
+                    files_to_convert = [files_to_convert]
+
+                files_to_delete = copy.deepcopy(files_to_convert)
+
+                output_extension = output_def["output_extension"]
+                output_extension = output_extension.replace('.', '')
+                if output_extension:
+                    new_repre["name"] = output_extension
+                    new_repre["ext"] = output_extension
```
BigRoy, line 96: In some cases a representation `name` might not match exactly the `ext` - e.g. for burnins or alike. I think this is currently an oversimplification if it'd ever run over a representation whose name doesn't equal exactly the representation. Plus, there is an off chance the other representation we're changing too already existing in the instance's representation, if e.g. a particular instance would already be generating a `exr` and a `png` format.

openpype/plugins/publish/extract_color_transcode.py
```diff
@@ -0,0 +1,212 @@
...
+    def _rename_output_files(self, files_to_convert, output_extension):
+        """Change extension of converted files."""
+        renamed_files = []
+        for file_name in files_to_convert:
+            file_name, _ = os.path.splitext(file_name)
+            new_file_name = '{}.{}'.format(file_name,
+                                           output_extension)
+            renamed_files.append(new_file_name)
+        files_to_convert = renamed_files
+        return files_to_convert
```
BigRoy, line 142:
```suggestion
        return renamed_files
```
We are not assigning in-place here so the original assignment to `files_to_convert` has no use here.

openpype/plugins/publish/extract_color_transcode.py
```diff
@@ -0,0 +1,212 @@
...
+            return False
+
+        if not repre.get("files"):
+            self.log.debug((
+                "Representation \"{}\" have empty files. Skipped."
+            ).format(repre["name"]))
+            return False
+
+        if not repre.get("colorspaceData"):
+            self.log.debug("Repre has no colorspace data. Skipped.")
```
BigRoy, line 195: This log should also mention the name of the representation like the above - plus I wouldn't abbreviate `Representation` in the logged message to `Repre`.

openpype/plugins/publish/extract_color_transcode.py
```diff
@@ -0,0 +1,212 @@
...
+                output_extension = output_extension.replace('.', '')
+                if output_extension:
+                    new_repre["name"] = output_extension
+                    new_repre["ext"] = output_extension
+
+                    files_to_convert = self._rename_output_files(
+                        files_to_convert, output_extension)
+
+                files_to_convert = [os.path.join(original_staging_dir, path)
+                                    for path in files_to_convert]
```
BigRoy, line 102: Logic here is wrong. The `files_to_convert` would've already been remapped to a new extension above but the files on disk have not. So we'd from here on out be looking in the original staging dir for files which not exist.

BigRoy (commented):
Some notes. Let's see if we can still simplify this more. It seems we're introducing some edge cases that'll be hard to debug with the extension changing too, etc. Rather makes more sense to ONLY convert the color space - no?

jakubjezek001:
Task linked: [OP-4643 color-v3: global oiio transcoder exctractor plugin](https://app.clickup.com/t/6658547/OP-4643)

jakubjezek001:
Task linked: [OP-2479 Colorspace management and distribution](https://app.clickup.com/t/6658547/OP-2479)

kalisp:
> Some notes. Let's see if we can still simplify this more. It seems we're introducing some edge cases that'll be hard to debug with the extension changing too, etc. Rather makes more sense to ONLY convert the color space - no?

I think that the use case is to produce multiple extensions, each with different extension. At least that is what @jakubjezek001 mentioned.

openpype/plugins/publish/extract_color_transcode.py
kalisp: Makes sense.

openpype/plugins/publish/extract_color_transcode.py
```diff
@@ -0,0 +1,214 @@
...
+                instance.data["representations"].append(new_repre)
+
+    def _get_output_file_path(self, input_filepath, output_dir,
+                              output_extension):
+        """Create output file name path."""
+        file_name = os.path.basename(input_filepath)
+        file_name, input_extension = os.path.splitext(file_name)
+        if not output_extension:
+            output_extension = input_extension
+        new_file_name = '{}.{}'.format(file_name,
```
tokejepsen, line 141:
```suggestion
        new_file_name = '{}{}'.format(file_name,
```
Getting double `.` here from the format and `output_extension`.

tokejepsen (changes requested):

openpype/plugins/publish/extract_color_transcode.py
```diff
@@ -0,0 +1,214 @@
...
+                continue
+
+            colorspace_data = repre["colorspaceData"]
+            source_colorspace = colorspace_data["colorspace"]
+            config_path = colorspace_data.get("configData", {}).get("path")
+            if not os.path.exists(config_path):
+                self.log.warning("Config file doesn't exist, skipping")
+                continue
+
+            repre = self._handle_original_repre(repre, profile)
```
tokejepsen, line 78: This needs to come after copying the tags from original representation to the new representation, else the new representation will inherit the potential `delete` tag and no representations will be available for integration.

tokejepsen (changes requested):

tokejepsen:
Something that needs testing (just briefly tried it), is changing the output extension, for example `.exr` to `.png`, which currently errors out:

`"C:\Users\TOKEJE~1\AppData\Local\Temp\op_transcoding_vkwfcqlp\Main_beauty.0001.exr"`

Think there is some logic that does not change the extension on the new representation.

tokejepsen:
We'll need to add support for https://openimageio.readthedocs.io/en/latest/oiiotool.html#cmdoption-ociodisplay since this how its displayed in Maya.
Will return with some code/suggestions.

openpype/lib/transcoding.py
```diff
@@ -1037,3 +1037,56 @@ def convert_ffprobe_fps_to_float(value):
...
+    # Prepare subprocess arguments
+
+    oiio_cmd.extend([
+        input_arg, input_path,
+    ])
+
+    # Add last argument - path to output
+    oiio_cmd.extend([
+        "-o", out_filepath
+    ])
```
tokejepsen, line 1089:
```suggestion
    oiio_cmd = [
        get_oiio_tools_path(),
        input_path,
        # Don't add any additional attributes
        "--nosoftwareattrib",
        "--colorconfig", config_path,
        "--colorconvert", source_colorspace, target_colorspace,
        "-o", out_filepath
    ]
```

Dont see a reason build the `oiio_cmd` over several lines. Also better to have the input path first.

tokejepsen (changes requested):

tokejepsen:
Since the destination branch has been merged, this PR needs to be merged into `release/3.15.x`.

openpype/settings/entities/schemas/projects_schema/schemas/schema_global_publish.json
```diff
@@ -197,6 +197,103 @@
...
+                                "key": "outputs",
+                                "label": "Output Definitions",
+                                "type": "dict-modifiable",
+                                "highlight_content": true,
+                                "object_type": {
+                                    "type": "dict",
+                                    "children": [
+                                        {
+                                            "key": "output_extension",
+                                            "label": "Output extension",
```
tokejepsen, line 271:
```suggestion
                                            "key": "extension",
                                            "label": "Extension",
```
Think `output` is implicit enough.

openpype/settings/entities/schemas/projects_schema/schemas/schema_global_publish.json
```diff
@@ -197,6 +197,103 @@
...
+                                            "type": "text"
+                                        },
+                                        {
+                                            "key": "output_colorspace",
+                                            "label": "Output colorspace",
```
tokejepsen, line 276:
```suggestion
                                            "key": "colorspace",
                                            "label": "Colorspace",
```
Think `output` is implicit enough.

openpype/lib/transcoding.py
jakubjezek001: I agree that this would be better if file path would be a string with sequential format. And not the list of strings - the docstring is stating it is str but code is expecting list. 

here is an example form [oiiotool docs](https://openimageio.readthedocs.io/en/latest/oiiotool.html#frame-sequences):
```
oiiotool big.1-3#.tif --resize 100x100 -o small.1-3#.tif
oiiotool big.1-3%04d.tif --resize 100x100 -o small.1-3%04d.tif
oiiotool --frames 1-3 big.#.tif --resize 100x100 -o small.#.tif
oiiotool --frames 1-3 big.%04d.tif --resize 100x100 -o small.%04d.tif
```

openpype/plugins/publish/extract_color_transcode.py
jakubjezek001: This would be probably better to mimic what Exctract review does. So the name of representation is the inherited form the output key - it is at the moment ignored with `_` at iteration. Here is what I mean:
image

jakubjezek001 (changes requested):

jakubjezek001:
> > Some notes. Let's see if we can still simplify this more. It seems we're introducing some edge cases that'll be hard to debug with the extension changing too, etc. Rather makes more sense to ONLY convert the color space - no?
> 
> I think that the use case is to produce multiple extensions, each with different extension. At least that is what @jakubjezek001 mentioned.

Some productions were demanding to convert **exr** (aces_ap0) into **dpx** (aces_cct) so they can do grading on them (dpx is faster for some DCC). This way we would be able to create **dpx** and not include review tag to skip farther processing in Exctract review.

openpype/settings/entities/schemas/projects_schema/schemas/schema_global_publish.json
```diff
@@ -197,6 +197,103 @@
                 }
             ]
         },
+        {
+            "type": "dict",
+            "collapsible": true,
+            "key": "ExtractColorTranscode",
```
jakubjezek001, line 203:
```suggestion
	    "key": "ExtractOIIOTranscode",
            "label": "Extract OIIO Transcode",
```

openpype/settings/entities/schemas/projects_schema/schemas/schema_global_publish.json
```diff
@@ -197,6 +197,103 @@
...
+            "label": "ExtractColorTranscode (ImageIO)",
```
jakubjezek001, line 204:
```suggestion
            "key": "ExtractOIIOTranscode",
            "label": "Extract OIIO Transcode",
```

jakubjezek001:
> We'll need to add support for [openimageio.readthedocs.io/en/latest/oiiotool.html#cmdoption-ociodisplay](https://openimageio.readthedocs.io/en/latest/oiiotool.html#cmdoption-ociodisplay) since this how its displayed in Maya. Will return with some code/suggestions.

There are two ways to do this. Either to rewrite schema for outputs to be `dict-contitional` and let user to chose from two enum items > `colorspace` or `screenspace`. Each of them would have different user inputs. Screenspace would allow user to add display and viewer inputs. and colorspace only colorspace input. This way we could also switch the oiiotool arguments. 

Or the other solution is more implicit and we could use what is done in Nuke here https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/nuke/api/lib.py#L2942

Briefly it would let user to add colorspace input in two ways. Either standart (ex: _Rec.709_) or screenspace way with brackets **viewer (display)** (ex: `Rec.709 (ACES)`) . And if brackets in text input is found it would assume it is screenspace.

tokejepsen:
> There are two ways to do this. Either to rewrite schema for outputs to be `dict-contitional` and let user to chose from two enum items > `colorspace` or `screenspace`. Each of them would have different user inputs. Screenspace would allow user to add display and viewer inputs. and colorspace only colorspace input. This way we could also switch the oiiotool arguments.

I think we should be explicit cause it already quite confusing to work with. Also I think we should mirror what OCIO is doing naming wise, so `view` and `display` parameters rather than coming up with new terminology. The data member `view` and `display` would just be part of the `colorspaceData`:

```
colorspaceData = {
    "colorspace": "ACEScg",  # for other publish plugins and loaders
    "configData": {
        "path": "/abs/path/to/config.ocio",  # for future references in case need
        "template": "{project[root]}/path/to/config.ocio"  # for other plugins within remote publish cases
    }
    "view": "sRGB",
    "display": "ACES 1.0 SDR-video"
}
```

jakubjezek001:
>... so `view` and `display` parameters rather than coming up with new terminology. 

You are right Screen space is something we used to call it so sorry for misleading the terminology. I was looking into it now and it called differently, weather we want to use it or not:

**display-referred reference space**  > what I was calling screenspace

Anyway I agree we should add **view** and **display** keys to colorspace data but as optional. Question is how we collect them from user input. Perhaps we should just offer all three `colorspace`, `display`, `view`. Only issue I am having with this is that there is no validation to make sure they are filled correctly - meaning if display and view is filled then colorspace should be empty, etc.

tokejepsen:
> Anyway I agree we should add **view** and **display** keys to colorspace data but as optional. Question is how we collect them from user input. Perhaps we should just offer all three `colorspace`, `display`, `view`. Only issue I am having with this is that there is no validation to make sure they are filled correctly - meaning if display and view is filled then colorspace should be empty, etc.

The `colorspace`, `display` and `view` should always be collected by the DCC plugins and injected into the representations. If the settings for `display` and `view` are empty in the `Output Definitions`, the transcoder will inherit what the DCC collected.

openpype/lib/transcoding.py
```diff
@@ -1037,3 +1037,46 @@ def convert_ffprobe_fps_to_float(value):
...
+    """Convert source files from one color space to another.
+
+    Filenames of input files are kept so make sure that output directory
+    is not the same directory as input files have.
+    - This way it can handle gaps and can keep input filenames without handling
+        frame template
+
+    Args:
+        input_path (str): Paths that should be converted. It is expected that
+            contains single file or image sequence of samy type.
```
BigRoy, line 1059: Docstring is confusing? Is it a single path or multiple? Paths? or path?

Also, why is it `input_path` versus `out_filepath`

I'd expect either:
```
input_path
output_path
```
or
```
in_filepath
out_filepath
```

openpype/lib/transcoding.py
```diff
@@ -1037,3 +1037,46 @@ def convert_ffprobe_fps_to_float(value):
...
+    source_colorspace,
+    target_colorspace,
+    logger=None
+):
+    """Convert source files from one color space to another.
+
+    Filenames of input files are kept so make sure that output directory
+    is not the same directory as input files have.
+    - This way it can handle gaps and can keep input filenames without handling
+        frame template
```
BigRoy, line 1055: This part of the docstring doesn't seem to apply to what the command does or I'm not understanding what it's supposed to explain

kalisp:
I hope I fixed all issues mentioned in comments.

Implemented display and view keys, updated oiio cmd.

openpype/lib/transcoding.py
tokejepsen: This has not been implemented?
The method is expecting `input_paths` as list, but we need it as a single string, plus potential `--frames` argument for sequence starting at non-default start frame.

openpype/plugins/publish/extract_color_transcode.py
tokejepsen: This has not been implemented.

openpype/plugins/publish/extract_color_transcode.py
tokejepsen: This has not been implemented.

openpype/lib/transcoding.py
```diff
@@ -1037,3 +1037,65 @@ def convert_ffprobe_fps_to_float(value):
...
+        logger (logging.Logger): Logger used for logging.
+    Raises:
+        ValueError: if misconfigured
+    """
+    if logger is None:
+        logger = logging.getLogger(__name__)
+
+    oiio_cmd = [
+        get_oiio_tools_path(),
+        input_path,
```
tokejepsen, line 1078: We might need `--frames` argument here to account for sequence not starting at non-default value (think its 1 or 0).

openpype/lib/transcoding.py
```diff
@@ -1037,3 +1037,65 @@ def convert_ffprobe_fps_to_float(value):
...
+        # Don't add any additional attributes
+        "--nosoftwareattrib",
+        "--colorconfig", config_path,
+        "-o", out_filepath
```
tokejepsen, line 1082: Think the output argument needs to be the last in the oiio_cmd list.

openpype/plugins/publish/extract_color_transcode.py
```diff
@@ -0,0 +1,271 @@
...
+                target_colorspace = output_def["colorspace"]
+                view = output_def["view"] or colorspace_data.get("view")
+                display = (output_def["display"] or
+                           colorspace_data.get("display"))
+                # both could be already collected by DCC,
+                # but could be overwritten
+                if view:
+                    new_repre["colorspaceData"]["view"] = view
+                if display:
+                    new_repre["colorspaceData"]["view"] = display
```
tokejepsen, line 117:
```suggestion
                    new_repre["colorspaceData"]["display"] = display
```

openpype/plugins/publish/extract_color_transcode.py
```diff
@@ -0,0 +1,271 @@
...
+                if output_extension:
+                    if new_repre["name"] == new_repre["ext"]:
+                        new_repre["name"] = output_extension
+                    new_repre["ext"] = output_extension
+
+                    renamed_files = []
+                    _, orig_ext = os.path.splitext(files_to_convert[0])
+                    for file_name in files_to_convert:
+                        file_name = file_name.replace(orig_ext,
+                                                      "."+output_extension)
```
tokejepsen, line 104: Let me know if I'm wrong, but looks like this will generate a list of file name in `files_to_convert` of the output extension, which is later used as input path.
This means the input path files wont exist?

tokejepsen (changes requested):

openpype/lib/transcoding.py
BigRoy: I'm pretty sure this is the case yes. Code is still invalid.

openpype/lib/transcoding.py
```diff
@@ -1037,3 +1037,65 @@ def convert_ffprobe_fps_to_float(value):
...
+    if not target_colorspace and not all([view, display]):
+        raise ValueError("Both screen and display must be set.")
+
+    if target_colorspace:
+        oiio_cmd.extend(["--colorconvert",
+                         source_colorspace,
+                         target_colorspace])
+    if view and display:
+        oiio_cmd.extend(["--iscolorspace", source_colorspace])
+        oiio_cmd.extend(["--ociodisplay", display, view])
```
BigRoy, line 1093: Here these arguments almost appear 'optional' but the function signature doesn't hint that at all.

openpype/lib/transcoding.py
kalisp: Cleaned up.

openpype/lib/transcoding.py
kalisp: What does it mean? What will be example of frames (filenames)?

openpype/lib/transcoding.py
kalisp: Didn't seem so, but moved nevertheless.

openpype/plugins/publish/extract_color_transcode.py
kalisp: Cleaned up.

openpype/lib/transcoding.py
kalisp: Ok, put None there, fiddled with docstring.

openpype/lib/transcoding.py
tokejepsen: Actually we might not need it since you are specifying the range in the input path string;

https://openimageio.readthedocs.io/en/latest/oiiotool.html#frame-sequences

tokejepsen (changes requested):
Could we test line 211?

kalisp:
Tested, 
2023-01-27 14_58_05-Window

'.' character is cleaned up on [line 94](https://github.com/ynput/OpenPype/pull/4291/files#diff-19eb0ff46fbda7281bf0d947a160f8b50bd67c54929efa69e161074f9f78066eR94)

openpype/lib/transcoding.py
```diff
@@ -1037,3 +1037,59 @@ def convert_ffprobe_fps_to_float(value):
...
+
+        # Don't add any additional attributes
+        "--nosoftwareattrib",
+        "--colorconfig", config_path,
+        "--colorconvert", source_color_space, target_color_space
+    ]
+    for input_path in input_paths:
+        # Prepare subprocess arguments
+
+        oiio_cmd.extend([
```
iLLiCiTiT:
```suggestion
        # Create copy of args base
        oiio_cmd = list(oiio_cmd)
        oiio_cmd.extend([
```

openpype/lib/transcoding.py
```diff
@@ -1037,3 +1037,59 @@ def convert_ffprobe_fps_to_float(value):
...
+            input_arg, input_path,
```
iLLiCiTiT: The arguments would grouw with more then one input. Also I don't think it is needed to define `input_arg` variable :)
```suggestion
        # Create copy of args base
        oiio_cmd = list(oiio_cmd)
        oiio_cmd.extend([
            "-i", input_path,
```

tokejepsen (approved):

openpype/lib/transcoding.py
kalisp: Wtf, that code is not in the branch anymore, for some time actually. (I checked with completely new checkouted branch).
No idea why GH is showing `convert_colorspace_for_input_paths` , it is now called only `convert_colorspace`

iLLiCiTiT: GitHub magic?

openpype/plugins/publish/extract_color_transcode.py
```diff
@@ -0,0 +1,283 @@
...
+
+        return files_to_convert
+
+    def _get_output_file_path(self, input_path, output_dir,
+                              output_extension):
+        """Create output file name path."""
+        file_name = os.path.basename(input_path)
+        file_name, input_extension = os.path.splitext(file_name)
+        if not output_extension:
+            output_extension = input_extension
```
tokejepsen, line 210:
```suggestion
        file_name, input_extension = os.path.splitext(file_name)
        if not output_extension:
            output_extension = input_extension.replace(".", "")
```

If the `output_extension` is empty and we use the input extension, then we end up with double `.`.

tokejepsen (changes requested):

tokejepsen:
Is there a workflow for only transcoding the review representations of the renders?
Atm I'm needing to bake the colorspace into the renders or a duplicate of the renders (changing the output extension), but really we only need to color correct the review files.
@kalisp @jakubjezek001

jakubjezek001:
> Is there a workflow for only transcoding the review representations of the renders? Atm I'm needing to bake the colorspace into the renders or a duplicate of the renders (changing the output extension), but really we only need to color correct the review files. @kalisp @jakubjezek001

There is not yet any at the moment, but potentially might be. In Flame we are doing this by transcoding via Flame's exporter. But this could be used to bypass it. Best of course would be doing this at farm.

openpype/plugins/publish/extract_color_transcode.py
jakubjezek001: > This would be probably better to mimic what Exctract review does. So the name of representation is the inherited form the output key - it is at the moment ignored with `_` at iteration. Here is what I mean: image

@kalisp this hasn't been implemented, are you planning to add it? I would rather use the name as it will come handy. Here is an examle. I am having two output profiles with the same extentions but different configurations. The name allwos me to differentiate them. With current workfow this is not able to work. I would like the loader offer me two different representations.

jakubjezek001 (dismissed):
need to use output preset name as representation name

kalisp:
Output definition key is used as representation["name"] (unless it is set to 'passthrough' - cannot use any special characters in keys in Settings).

Updated plugin and schema documentation.

openpype/plugins/publish/extract_color_transcode.py
tokejepsen: I dont think its a good workflow to have implicit functionality through which name we use for the output definitions. If there is some functionality that an output definition needs, it should have a toggle in the settings.

openpype/lib/transcoding.py
```diff
@@ -1037,3 +1037,63 @@ def convert_ffprobe_fps_to_float(value):
...
+    view=None,
+    display=None,
+    logger=None
+):
+    """Convert source file from one color space to another.
+
+    Args:
+        input_path (str): Path that should be converted. It is expected that
+            contains single file or image sequence of same type
+            (sequence in format 'file.FRAMESTART-FRAMEEND#.exr', see oiio docs)
```
jakubjezek001, line 1057: I would also mention an example of input string to make it more clear `big.1-3#.tif`

openpype/lib/transcoding.py
```diff
@@ -1037,3 +1037,63 @@ def convert_ffprobe_fps_to_float(value):
...
+        output_path (str): Path to output filename.
```
jakubjezek001, line 1058: here I would mention that if sequence goes in we have to format output path likewise input path. And also an example would help: `output.1-3#.tif`

kalisp: Updated docstring

tokejepsen:
I think we need to provide a workflow to transcode only the review materials of a representation, then we can provide a default profile for any DCC that supports OCIO. The settings looking something similar to this:

Capture

kalisp:
> I think we need to provide a workflow to transcode only the review materials of a representation, then we can provide a default profile for any DCC that supports OCIO. The settings looking something similar to this:
> 

Not sure I follow. It is about committing default settings for some DCCs? (For which ones - Maya, Nuke?)
Shouldn't there be 'review' under Families in your example? 'review' in Output definitions key will be used as a name of new representation. 

There is Colorspace missing, so admin will need to provide it eventually. Is this about providing "helping hand" for basic workflow?

Isn't sufficient to set 'review' in Families and let old representation be deleted? That way 'files' of instance that should be pushed through ExtractReview (with a 'review' in families) will be transcoded, new repre created which will contain all propertes of old repre (eg. families etc), this one will be used for ExtractReview and old one will be skipped?

tokejepsen:
> Isn't sufficient to set 'review' in Families and let old representation be deleted? That way 'files' of instance that should be pushed through ExtractReview (with a 'review' in families) will be transcoded, new repre created which will contain all propertes of old repre (eg. families etc), this one will be used for ExtractReview and old one will be skipped?

I tried this and it does not work.

Up for trying any settings configuration that'll get us to only transcode the reviewable material but preserve the original renders colorspace Eg. ACEScg exr renders from Maya with sRGB transcoded mp4 for Ftrack.

kalisp:
Please try something like (ftrack tag not required, imho):
2023-02-10 17_38_08-Window

openpype/plugins/publish/extract_color_transcode.py
```diff
@@ -0,0 +1,325 @@
...
+                added_representations = True
+
+            if added_representations:
+                self._mark_original_repre_for_deletion(repre, profile,
+                                                       added_review)
+
+        for repre in tuple(instance.data["representations"]):
+            tags = repre.get("tags") or []
+            if "delete" in tags and "thumbnail" not in tags:
+                instance.data["representations"].remove(repre)
```
tokejepsen, line 183: @kalisp nice! It does work to have a "png" output, although you end up with "extra" PNGs. Ideally we need to introduce another tag that can delete the output later at the review extraction stage. Not sure what to call it, maybe `Delete output at review` or `Delete output later`. You can simulate this by having the tag `Delete output` and comment out these lines.

Capture

kalisp: Implemented logic to keep newly added representations for `extract_review`. 

We should design probably better way, but we need to define what "Delete output later" means.

Your Settings should allow create review, but delete source transcoded .pngs

tokejepsen: Noice! Tested and it works!

tokejepsen (approved):

jakubjezek001:
Could we also add support for argument inputs in settings presets? I had just tested conversion from exr to dpx and it turns out I need to add `-d uint10` so it is converted to integer 10bit per channel. 

Here is also another way we could use oiiotool to influance output. 

`oiiotool in.tif --text:x=400:y=600:color=1,0,0 "Hello" -o out.tif`

openpype/plugins/publish/extract_color_transcode.py
```diff
@@ -0,0 +1,333 @@
...
+                    transcoded_file_path = os.path.join(new_staging_dir,
+                                                        file_name)
+                    instance.context.data["cleanupFullPaths"].append(
+                        transcoded_file_path)
+
+                custom_tags = output_def.get("custom_tags")
+                if custom_tags:
+                    if not new_repre.get("custom_tags"):
+                        new_repre["custom_tags"] = []
+                    new_repre["custom_tags"].extend(custom_tags)
```
iLLiCiTiT, line 161:
```suggestion
                custom_tags = output_def.get("custom_tags")
                if custom_tags:
                    if "custom_tags" not in new_repre.get:
                        new_repre["custom_tags"] = []
                    new_repre["custom_tags"].extend(custom_tags)
```

openpype/plugins/publish/extract_color_transcode.py
```diff
@@ -0,0 +1,333 @@
...
+        if not repre.get("colorspaceData"):
+            self.log.debug("Representation '{}' has no colorspace data. "
+                           "Skipped.")
+            return False
+
+        return True
+
+    def _mark_original_repre_for_deletion(self, repre, profile, added_review):
+        """If new transcoded representation created, delete old."""
+        if not repre.get("tags"):
```
iLLiCiTiT, line 323:
```suggestion
        if "tags" not in repre:
```

openpype/plugins/publish/extract_color_transcode.py
```diff
@@ -0,0 +1,333 @@
...
+
+        if not is_oiio_supported():
+            self.log.warning("OIIO not supported, no transcoding possible.")
+            return
+
+        profile = self._get_profile(instance)
+        if not profile:
+            return
+
+        repres = instance.data.get("representations") or []
```
iLLiCiTiT, line 85:
```suggestion
        repres = instance.data.get("representations")
        if not repres:
            return
```

openpype/plugins/publish/extract_color_transcode.py
```diff
@@ -0,0 +1,333 @@
...
+                if not new_repre.get("tags"):
+                    new_repre["tags"] = []
+                for tag in output_def["tags"]:
+                    if tag not in new_repre["tags"]:
+                        new_repre["tags"].append(tag)
+
+                    if tag == "review":
+                        added_review = True
+
+                new_repre["tags"].append("newly_added")
```
iLLiCiTiT, line 173: What about create separated list of "new representations" to avoid using the tag?
```
new_representations = []
for repre in repres:
    ...
    new_representations.append(new_repre)
    
for repre in tuple(repres):
    if <remove condition>:
        repres.remove(repre)

repres.extend(new_representations)
```

openpype/plugins/publish/extract_color_transcode.py
kalisp: This syntax is just weird @_@. Any special reason why this is better?

iLLiCiTiT: Because `if not repre.get("custom_tags")` is checking output value, not existence of key.
```
repre = {"custom_tags": []}
# Even if the key is there and is set it will be set again
if not repre.get("custom_tags"):
    repre["custom_tags"] = []
```

It could be also changed to
```
if repre.get("custom_tags") is None:
    repre["custom_tags"] = []
```

openpype/plugins/publish/extract_color_transcode.py
kalisp: That is actually much nicer approach than what I concocted.

kalisp:
Added additional command line arguments for special cases (for example .dpx requires change of binary depth etc.)
2023-02-15 16_03_58-Window

openpype/plugins/publish/extract_color_transcode.py
```diff
@@ -0,0 +1,332 @@
...
+        task_data = instance.data["anatomyData"].get("task", {})
+        task_name = task_data.get("name")
+        task_type = task_data.get("type")
+        subset = instance.data["subset"]
+        filtering_criteria = {
+            "hosts": host_name,
+            "families": family,
+            "task_names": task_name,
+            "task_types": task_type,
+            "subset": subset
```
tokejepsen, line 276:
```suggestion
            "subsets": subset
```

`subset` vs `subsets` meant that subset filtering was not correct.

tokejepsen (changes requested):

tokejepsen:
We should also have some documentation on this, for example recommended workflow for transcoding only reviewables if we dont provide defaults for this.

openpype/modules/slack/manifest.yml
```diff
@@ -19,6 +19,8 @@ oauth_config:
       - chat:write.public
       - files:write
       - channels:read
+      - users:read
+      - usergroups:read
```
jakubjezek001, line 23:
```suggestion
```

This seems to be in wrong PR.

kalisp:
> I agree with @tokejepsen that part of this PR should be also documentation

Yes, but where? It is not hot specific, not a module...

kalisp:
Added documentation to `settings_project_global.md`

website/docs/project_settings/settings_project_global.md
```diff
@@ -45,6 +45,21 @@ The _input pattern_ matching uses REGEX expression syntax (try [regexr.com](http
 The **colorspace name** value is a raw string input and no validation is run after saving project settings. We recommend to open the specified `config.ocio` file and copy pasting the exact colorspace names.
 :::
 
+### Extract OIIO Transcode
+There is profile configurable (see lower) plugin which allows to transcode any incoming representation to one or multiple new representations (configured in `Output Definitions`) with different target colorspaces.
```
tokejepsen, line 49:
```suggestion
There is profile configurable plugin which allows to transcode any incoming representation to one or multiple new representations (configured in `Output Definitions`) with different target colorspaces.
```

Dont think we need to point people just below.

website/docs/project_settings/settings_project_global.md
```diff
@@ -45,6 +45,21 @@ The _input pattern_ matching uses REGEX expression syntax (try [regexr.com](http
...
+Plugin expects instances with filled dictionary `colorspaceData` on a representation. This data contains information about source colorspace and must be collected for transcoding.
+`oiiotool` is used for transcoding, eg. `oiiotool` must be present in `vendor/bin/oiio` or environment variable `OPENPYPE_OIIO_PATHS` must be provided for custom oiio installation.
+
+Notable parameters:
+- **`Delete Original Representation`** - keep or remove original representation. If old representation is kept, but there is new transcoded representation with 'Create review' tag, original representation looses its 'review' tag if present.
+- **`Extension`** - target extension, could be empty - original extension is used
```
tokejepsen, line 55:
```suggestion
- **`Extension`** - target extension. If left empty, original extension is used.
```

website/docs/project_settings/settings_project_global.md
```diff
@@ -45,6 +45,21 @@ The _input pattern_ matching uses REGEX expression syntax (try [regexr.com](http
...
+- **`Colorspace`** - target colorspace - must be available in used color config
```
tokejepsen, line 56:
```suggestion
- **`Colorspace`** - target colorspace, which must be available in used color config.
```

website/docs/project_settings/settings_project_global.md
```diff
@@ -45,6 +45,21 @@ The _input pattern_ matching uses REGEX expression syntax (try [regexr.com](http
...
+- **`Display & View`** - transcoding into colorspace OR into display and viewer space could be used. (It is disjunctive: Colorspace & nothing in Display and View or opposite)
```
tokejepsen, line 57:
```suggestion
- **`Display & View`** - transcoding into colorspace or into display and viewer space could be used. Cannot use both `Colorspace` and `Display & View` at the same time.
```

website/docs/project_settings/settings_project_global.md
```diff
@@ -45,6 +45,21 @@ The _input pattern_ matching uses REGEX expression syntax (try [regexr.com](http
...
+- **`Arguments`** - special additional command line arguments for `oiiotool`
```
tokejepsen, line 58:
```suggestion
- **`Arguments`** - special additional command line arguments for `oiiotool`.
```

website/docs/project_settings/settings_project_global.md
```diff
@@ -45,6 +45,21 @@ The _input pattern_ matching uses REGEX expression syntax (try [regexr.com](http
...
+
+
+Example here describes use case for creation of new color coded review of png image sequence. Original representation's files are kept intact, review is created from transcoded files, but these files are removed in cleanup process.
+![global_oiio_transcode](assets/global_oiio_transcode.png)
```
tokejepsen, line 62: Could we have another example here for transcoding just the review for a Maya beauty render?

tokejepsen (changes requested):

website/docs/project_settings/settings_project_global.md
```diff
@@ -45,6 +45,21 @@ The _input pattern_ matching uses REGEX expression syntax (try [regexr.com](http
 The **colorspace name** value is a raw string input and no validation is run after saving project settings. We recommend to open the specified `config.ocio` file and copy pasting the exact colorspace names.
 :::
 
+### Extract OIIO Transcode
+There is profile configurable plugin which allows to transcode any incoming representation to one or multiple new representations (configured in `Output Definitions`) with different target colorspaces.
+Plugin expects instances with filled dictionary `colorspaceData` on a representation. This data contains information about source colorspace and must be collected for transcoding.
+`oiiotool` is used for transcoding, eg. `oiiotool` must be present in `vendor/bin/oiio` or environment variable `OPENPYPE_OIIO_PATHS` must be provided for custom oiio installation.
+
```
jakubjezek001, line 52:
```suggestion
OIIOTools transcoder plugin with configurable output presets. Any incoming representation with `colorspaceData` is convertable to single or multiple representations with different target colorspaces or display and viewer names found in linked **config.ocio** file.

`oiiotool` is used for transcoding, eg. `oiiotool` must be present in `vendor/bin/oiio` or environment variable `OPENPYPE_OIIO_PATHS` must be provided for custom oiio installation.

```

website/docs/project_settings/settings_project_global.md
tokejepsen: Think the main things would be:

- no `Colorspace` or `Display & View` cause that is taken from the DCC (Maya).
- subset regex to only transcode "beauty" render layer.

tokejepsen (changes requested):
Could we have another transcoding example for Maya?

jakubjezek001 (dismissed):
I am having another issue. I was looking into the log and could not find in the command combination of both - colorspace convert, display and viewer. There is only colorspace convert, so there has to be some issue with the condition at `transcoding.py line 187`

```
INFO: Looking for matching profile for: hosts: "nuke" | families: "render" | task_names: "compositing" | task_types: "Compositing" | subsets: "renderCompositingMain"
INFO: Profile selected: {'families': [], 'hosts': ['nuke'], 'task_types': ['Compositing'], 'task_names': [], 'subsets': [], 'delete_original': False, 'outputs': {'dpx': {'extension': 'dpx', 'colorspace': 'compositing_log', 'display': '', 'view': '', 'oiiotool_args': {'additional_command_args': ['-d uint10']}, 'tags': [], 'custom_tags': []}, 'review': {'extension': 'png', 'colorspace': '', 'display': 'ACES', 'view': 'Rec.709', 'oiiotool_args': {'additional_command_args': []}, 'tags': ['review', 'ftrackreview', 'delete'], 'custom_tags': []}}}
DEBUG: profile: {'families': [], 'hosts': ['nuke'], 'task_types': ['Compositing'], 'task_names': [], 'subsets': [], 'delete_original': False, 'outputs': {'dpx': {'extension': 'dpx', 'colorspace': 'compositing_log', 'display': '', 'view': '', 'oiiotool_args': {'additional_command_args': ['-d uint10']}, 'tags': [], 'custom_tags': []}, 'review': {'extension': 'png', 'colorspace': '', 'display': 'ACES', 'view': 'Rec.709', 'oiiotool_args': {'additional_command_args': []}, 'tags': ['review', 'ftrackreview', 'delete'], 'custom_tags': []}}}
DEBUG: repre (1): `exr`
DEBUG: Conversion command: C:\CODE\__PYPE\OpenPype_3.15_up\vendor\bin\oiio\windows\oiiotool.exe C:/projects/test_surfing_giant_editorial_1/ep001_102b_Birthday/ep001_102b_shots/ep001_102b_sh0010/work/compositing/renders/nuke/renderCompositingMain\renderCompositingMain.1001-1120#.exr --nosoftwareattrib --colorconfig C:\CODE\__PYPE\OpenPype_3.15_up\vendor\bin\ocioconfig\OpenColorIOConfigs\aces_1.2\config.ocio -d uint10 --colorconvert ACES - ACES2065-1 compositing_log -o C:\Temp\op_transcoding_d8511_le\renderCompositingMain.1001-1120#.dpx
  File "C:\CODE\__PYPE\OpenPype_3.15_up\.venv\lib\site-packages\pyblish\plugin.py", line 522, in __explicit_process
    runner(*args)
  File "C:\CODE\__PYPE\OpenPype_3.15_up\openpype\plugins\publish\extract_color_transcode.py", line 153, in process
  File "C:\CODE\__PYPE\OpenPype_3.15_up\openpype\lib\transcoding.py", line 1088, in convert_colorspace
    raise ValueError("Colorspace and both screen and display"
ValueError: Colorspace and both screen and display cannot be set together.Choose colorspace or screen and display
```
My configuration is:
image

`project_settings/global/publish/ExtractOIIOTranscode/profiles/0/outputs`
[ExtractOIIOTranscode_profiles_0_outputs.zip](https://github.com/ynput/OpenPype/files/10802429/ExtractOIIOTranscode_profiles_0_outputs.zip)

kalisp:
My bad, got confused by 'no Colorspace or Display & View cause that is taken from the DCC (Maya)' and mistook source colorspace.
Fixed.

website/docs/project_settings/settings_project_global.md
kalisp: How this should work? I don't have `Colorspace` collected from DCC

kalisp:
Added new dropdown to differentiate between 'Colorspace' and 'Display' transcoding. It takes collected value from DCC if appropriate field not filled in.

Added use case for Maya to documentation.

tokejepsen (approved):

website/docs/project_settings/settings_project_global.md
```diff
@@ -45,6 +45,25 @@ The _input pattern_ matching uses REGEX expression syntax (try [regexr.com](http
 The **colorspace name** value is a raw string input and no validation is run after saving project settings. We recommend to open the specified `config.ocio` file and copy pasting the exact colorspace names.
 :::
 
+### Extract OIIO Transcode
+OIIOTools transcoder plugin with configurable output presets. Any incoming representation with `colorspaceData` is convertable to single or multiple representations with different target colorspaces or display and viewer names found in linked **config.ocio** file.
+
+`oiiotool` is used for transcoding, eg. `oiiotool` must be present in `vendor/bin/oiio` or environment variable `OPENPYPE_OIIO_PATHS` must be provided for custom oiio installation.
+
+Notable parameters:
+- **`Delete Original Representation`** - keep or remove original representation. If old representation is kept, but there is new transcoded representation with 'Create review' tag, original representation looses its 'review' tag if present.
```
BigRoy, line 54:
```suggestion
- **`Delete Original Representation`** - keep or remove original representation. If old representation is kept, but there is new transcoded representation with 'Create review' tag, original representation loses its 'review' tag if present.
```

jakubjezek001 (approved):
it works ;)

openpype/plugins/publish/extract_color_transcode.py
```diff
@@ -0,0 +1,365 @@
...
+
+        renamed_files = []
+        for file_name in files_to_convert:
+            file_name, _ = os.path.splitext(file_name)
+            file_name = '{}.{}'.format(file_name,
+                                       output_extension)
+            renamed_files.append(file_name)
+        new_repre["files"] = renamed_files
+
+    def _rename_in_representation(self, new_repre, files_to_convert,
```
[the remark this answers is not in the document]
tokejepsen: Could we have a follow PR resolving this?

PR #4293, opened 2023-01-07
Remove redundant export_alembic method.

tokejepsen:
## Brief description
`export_alembic` method is not used anywhere anymore.


moonyuet:
looks good to me based on the code as the function isn't applied into any scripts in the plugins folder.
Also tested with the code, it doesn't break anything too.

BigRoy:
Was a bit confused since I thought we're definitely using a method from `lib` to export the alembics, but apparently the one used is called `extract_alembic`. So `export_alembic` should definitely be removed.

BigRoy:
For what it's worth, the `_get_mel_global()` function in lib also appears to be unused?

moonyuet (approved):

PR #4295, opened 2023-01-09
General: Enhanced enum def items

iLLiCiTiT:
## Brief description
Atrribute definition `EnumDef` has more specifically defined value and label.

## Description
The previous implementation expected dictionary where key is value and value is label which is unintuitive. This logic was kept for backwards compatibility (at least for time there are opened PRs which use `EnumDef` that way. Added option to pass list where each item is dicrionary with `"value"` and `"label"` to make it more explicit.

Changed items in currently used `EnumDef` to use explicit items.

## Testing notes:
1. Create editorial should show correct enum items
2. Workfile builder should show corrent enum items

### DIFF
--- a/openpype/lib/attribute_definitions.py
+++ b/openpype/lib/attribute_definitions.py
@@ -488,8 +488,8 @@ class EnumDef(AbtractAttrDef):
 
         output = []
         if isinstance(items, dict):
-            for key, value in items.items():
-                output.append({"label": key, "value": value})
+            for value, label in items.items():
+                output.append({"label": label, "value": value})
 
         elif isinstance(items, (tuple, list, set)):
             for item in items:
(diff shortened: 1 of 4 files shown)


openpype/lib/attribute_definitions.py
```diff
@@ -433,38 +434,93 @@ def __init__(self, key, items, default=None, **kwargs):
...
+        output = []
+        if isinstance(items, dict):
+            for value, label in items.items():
+                output.append({"label": label, "value": value})
+
+        elif isinstance(items, (tuple, list, set)):
+            for item in items:
+                if isinstance(item, dict):
+                    # Test if value is available
+                    item["value"]
```
kalisp, line 498: What is this, this looks weird.

iLLiCiTiT: It would raise `KeyError` if `"value"` is missing.

kalisp: Yeah, I would prefer explicit.

kalisp (approved):
Other than implicit KeyError it seems to work. Tested on Transcoding addon, conversion from legacy works.

PR #4303, opened 2023-01-10
General: missing comment on standalone and tray publisher

kalisp:
## Brief description
Comment might have not been collected when inserted only after collection phase which resulted in not sending comment to Ftrack.

## Additional info
Clean-up of comment and using only comments on instances is good step, but as far as we are still using old Pyblish, context comment should be kept as a fallback (comment might be filled after collection phase, in this case it is present only on context level. Eg. artists knows/sees that it is filled, but integrator prints it is not there. Another approach would be to disable comments  field after publish is started, but I didn't want to mess with Pyblish.)

Comment field in Publisher is controlled by `project_settings/global/publish/collect_comment_per_instance`. To show comment field this should be set:
2023-01-10 17_47_24-CollectInstanceCommentDef

## Testing notes:
1. publish anything in Standalone Publisher, fill comment only after collecting phase
2. check `Integrate Ftrack note` and Ftrack for presence of comment
3. (btw default note template expects both `intent` and `comment`, skipping `intent` is quite easy in SP, in Publisher field for it is missing completely, imho)

### DIFF
--- a/openpype/modules/ftrack/plugins/publish/integrate_ftrack_note.py
+++ b/openpype/modules/ftrack/plugins/publish/integrate_ftrack_note.py
@@ -130,8 +130,8 @@ class IntegrateFtrackNote(pyblish.api.InstancePlugin):
             if not note_text.solved:
                 self.log.warning((
                     "Note template require more keys then can be provided."
-                    "\nTemplate: {}\nData: {}"
-                ).format(template, format_data))
+                    "\nTemplate: {}\nMissing values for keys:{}\nData: {}"
+                ).format(template, note_text.missing_keys, format_data))
                 continue
 
             if not note_text:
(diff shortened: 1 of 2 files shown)


openpype/modules/ftrack/plugins/publish/integrate_ftrack_note.py
```diff
@@ -45,7 +45,8 @@ def process(self, instance):
         host_name = context.data["hostName"]
         app_name = context.data["appName"]
         app_label = context.data["appLabel"]
-        comment = instance.data["comment"]
+        # context comment is fallback until old Pyblish is removed
+        comment = instance.data["comment"] or context.data.get("comment")
```
iLLiCiTiT, line 49: This is duplicate of what happens here https://github.com/ynput/OpenPype/blob/develop/openpype/plugins/publish/collect_comment.py#L99

jakubjezek001:
Task linked: [OP-4679 Issue Comment on Standalone and tray Publisher](https://app.clickup.com/t/6658547/OP-4679)

kalisp:
Pushed collector to Extract phase because of Pyblish.

Marked with TODO, should be refactored when Pyblish is gone.

iLLiCiTiT (approved):

m-u-r-p-h-y (approved):
tested Tray publisher

image

m-u-r-p-h-y:
there is just a slight inconsistency in terminology. 

There are three terms used and it may be a bit confusing.

We are writing a `Comment`, integrating a `Note` but publishing a `Description`

PR #4308, opened 2023-01-11
Nuke: new creators fixes

jakubjezek001:
## Brief description
New creators correctly checking nodes in scene for existing subset.

## Description
Latest publisher creator plugins are now correctly detecting if nodes in workfile node tree is having a node with the same subset name user wants to create. This way unique subset name is secured.

## Additional info
Error messages were having typos in all creators

## Testing notes:
1. open nuke in your testing workfile already converted with new publisher.
2. open Create from OpenPype menu and set family and variant matching any already created instance node. 
3. Error message should popup with clear message.

### DIFF
--- a/openpype/hosts/nuke/api/lib.py
+++ b/openpype/hosts/nuke/api/lib.py
@@ -896,7 +896,9 @@ def get_view_process_node():
     ipn_orig = None
     for v in nuke.allNodes(filter="Viewer"):
         ipn = v['input_process_node'].getValue()
-        if "VIEWER_INPUT" not in ipn:
+        if ipn:
+            if "VIEWER_INPUT" not in ipn:
+                return
             ipn_orig = nuke.toNode(ipn)
             ipn_orig.setSelected(True)
 
--- a/openpype/hosts/nuke/api/lib.py
+++ b/openpype/hosts/nuke/api/lib.py
@@ -897,7 +897,7 @@ def get_view_process_node():
     for v in nuke.allNodes(filter="Viewer"):
         ipn = v['input_process_node'].getValue()
         if ipn:
-            if "VIEWER_INPUT" not in ipn:
+            if "VIEWER_INPUT" in ipn:
                 return
             ipn_orig = nuke.toNode(ipn)
             ipn_orig.setSelected(True)
--- a/openpype/hosts/nuke/api/lib.py
+++ b/openpype/hosts/nuke/api/lib.py
@@ -893,17 +893,27 @@ def get_imageio_input_colorspace(filename):
 def get_view_process_node():
     reset_selection()
 
-    ipn_orig = None
-    for v in nuke.allNodes(filter="Viewer"):
-        ipn = v['input_process_node'].getValue()
-        if ipn:
-            if "VIEWER_INPUT" in ipn:
-                return
-            ipn_orig = nuke.toNode(ipn)
-            ipn_orig.setSelected(True)
+    ipn_node = None
+    for v_ in nuke.allNodes(filter="Viewer"):
+        ipn = v_['input_process_node'].getValue()
+        ipn_node = nuke.toNode(ipn)
+        if ipn_node:
+            if ipn == "VIEWER_INPUT":
+                # since it is set by default we can ignore it
+                # nobody usually use this
+                continue
+            else:
+                # in case a Viewer node is transfered from
+                # different workfile with old values
+                raise NameError((
+                    "Input process node name '{}' set in "
+                    "Viewer '{}' is does't exists in nodes"
+                ).format(ipn, v_.name()))
 
-    if ipn_orig:
-        return duplicate_node(ipn_orig)
+        ipn_node.setSelected(True)
+
+    if ipn_node:
+        return duplicate_node(ipn_node)
 
 
 def on_script_load():
--- a/openpype/hosts/nuke/api/plugin.py
+++ b/openpype/hosts/nuke/api/plugin.py
@@ -89,9 +89,17 @@ class NukeCreator(NewCreator):
         """)
         node.addKnob(info_knob)
 
-    def check_existing_subset(self, subset_name, instance_data):
-        """Check if existing subset name already exists."""
-        exists = False
+    def check_existing_subset(self, subset_name):
+        """Make sure subset name is unique.
+
+        It search within all nodes recursively
+        and checks if subset name is found in
+        any node having instance data knob.
+
+        Arguments:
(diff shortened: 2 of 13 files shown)


jakubjezek001:
Task linked: [OP-3845 nuke: convert to new publisher](https://app.clickup.com/t/6658547/OP-3845)

kalisp:
Still cannot create new render instance, if fails for me on `host.nuke.api.lib:1729`. Issue is with knob_name 'tile_color', which seems to be of knob_type '__legacy__' and knob_value "0xff0000ff".

I think I have default values in Settings, do not see any project overrides.

64qam (approved):

64qam:
Nuke 13.x is working well. However, in Nuke 12 publishing is finished successfully, but nothing is physically published (there are no render/workfiles in workfile folder nor publish folder, no reviews, nothing) 
image

openpype/hosts/nuke/api/lib.py
```diff
@@ -896,7 +896,9 @@ def get_view_process_node():
     ipn_orig = None
     for v in nuke.allNodes(filter="Viewer"):
         ipn = v['input_process_node'].getValue()
-        if "VIEWER_INPUT" not in ipn:
+        if ipn:
+            if "VIEWER_INPUT" in ipn:
+                return
```
BigRoy, line 901: Note: I know nothing of the nuke codebase.

This seems odd. Why would it iterate over all of them but return early as soon as it finds on viewer with an input process node with "VIEWER_INPUT" in it. Are you sure you didn't mean `continue` instead of `return`? Or maybe `return ipn`?

Additionally, this function really needs a docstring to explain what it does.

openpype/hosts/nuke/api/plugin.py
```diff
@@ -98,10 +93,13 @@ def check_existing_subset(self, subset_name, instance_data):
         """Check if existing subset name already exists."""
         exists = False
         for node in nuke.allNodes(recurseGroups=True):
+            # make sure testing node is having instance knob
+            if INSTANCE_DATA_KNOB not in node.knobs().keys():
+                continue
             node_data = get_node_data(node, INSTANCE_DATA_KNOB)
+            # test if subset name is matching
```
BigRoy, line 100: What happens if the current subset name is `renderMain` but the one on the node in `renderMain01`. Are we sure we want to do a `subset_name in `node_data.get("subset")` check.

Additionally, `subset_name` in `None` will error, so the `.get("subset")` will still error.

Did we mean to do `if subset_name == node.get("subset")` ?

openpype/hosts/nuke/api/plugin.py
```diff
@@ -316,8 +314,8 @@ def create(self, subset_name, instance_data, pre_create_data):
         # make sure subset name is unique
         if self.check_existing_subset(subset_name, instance_data):
             raise NukeCreatorError(
-                ("subset {} is already published"
-                 "definition.").format(subset_name))
+                ("Subset '{}' is already created "
+                 "in nodes! Change variant name!").format(subset_name))
```
BigRoy, line 318: I'd rephrase this as: `"A publish instance for '{subset}' in asset '{asset}' already exists in the scene. Please change the variant name to ensure unique output"`.

openpype/hosts/nuke/api/plugin.py
```diff
@@ -98,10 +93,13 @@ def check_existing_subset(self, subset_name, instance_data):
         """Check if existing subset name already exists."""
         exists = False
         for node in nuke.allNodes(recurseGroups=True):
```
BigRoy, line 95: Technically the artist could publish a `renderMain` for different assets from the same comp. As such - the check here is too strict. It should actually check whether an instance already exists that will publish the subset for the same asset as you're trying to create now.

BigRoy (changes requested):

openpype/hosts/nuke/api/plugin.py
iLLiCiTiT: I agree. I would move the validation to publishing. There is so much ways how to bypass this check and at the same time can cause so much headaches (and slowdown) I would remove it completelly.

BigRoy: Yes - a global plugin which validates this is the BEST case scenario since it's valid for basically ALL publishes - unless there are special cases I don't know about where two separate instances should somehow publish and integrate into the same asset+subset from within the same publish.

Here's an old basis for that logic: https://github.com/Colorbleed/colorbleed-config/blob/acre/colorbleed/plugins/global/publish/validate_unique_subsets.py

However as clearly can be seen from that todo it still has the bug that it doesn't check it per asset. I'd be happy to refactor it and set up a PR if we feel this is a worthwhile global plug-in.

openpype/hosts/nuke/api/lib.py
jakubjezek001: huh.. Well, that answer me the question `Is it good to code paralelly with client's weekly meeting?`
Fixed now.. Thanks!

openpype/hosts/nuke/api/plugin.py
jakubjezek001: > Did we mean to do if subset_name == node.get("subset") ?

Yeah good point, I will change this condition, with the exception that we need to do it on `node_data` rather on node name. It might happen that user will duplicate a node with name `renderMain` with node data `renderMain` and create second node with node name `renderMain01` but the node data will still remain on subset `renderMain`.  I see that more usefull might be to drive subset name from node name but since we store and collect the name from node data we should be checking it from there. We might change that in future perhpas but it will be done with different [issue ](https://github.com/ynput/OpenPype/issues/4329) and PR.

openpype/hosts/nuke/api/plugin.py
jakubjezek001: I had alternate it a bit, but yeah agree it sounded bit harsh. Also, I had reworked it completely because in fact we didn't need to have exceptions per plugin, but rather abstracted it into the function (I believe someone had already mentioned it so there it is.)

openpype/hosts/nuke/api/plugin.py
jakubjezek001: I like the approach of moving this to publishing. Each instance, should have been able to store the validator disabled value into instance data - so in case user disable it on a instance with different asset name he will still pass the validation phase. 

For now, since the validator is still not in place, lets keep the validation in clreators and remove it in the next PR.

kalisp (approved):
Solved my issue with creation of new subsets.

PR #4309, opened 2023-01-11
Feature: Add subprocess command before sync loop

Tilix4:
## Brief description
Add the possibility to run a subprocess command before the main sync loop is processed.

## Description
This could be very useful if you need to execute processes on published representations without slowing down the pyblish process too much.

## Testing notes:
In an integrator, add:
```
manager = ModulesManager()
sync_server_module = manager.modules_by_name["sync_server"]
sync_server_module.add_before_loop_cmd(["/path/to/blender.exe"])
```
Blender will open when the loop will be executed.

### DIFF
--- a/openpype/modules/sync_server/rest_api.py
+++ b/openpype/modules/sync_server/rest_api.py
@@ -29,9 +29,21 @@ class SyncServerModuleRestApi:
             self.prefix + "/reset_timer",
             self.reset_timer,
         )
+        self.server_manager.add_route(
+            "POST",
+            self.prefix + "/add_before_loop_cmd",
+            self.add_before_loop_cmd,
+        )
 
     async def reset_timer(self, _request):
         """Force timer to run immediately."""
         self.module.reset_timer()
+
+        return Response(status=200)
+
+    async def add_before_loop_cmd(self, request):
+        """Add a subprocess command to be run before sync loop."""
+        cmd = await request.json()
+        self.module.add_before_loop_cmd(cmd)
 
         return Response(status=200)
--- a/openpype/modules/sync_server/sync_server.py
+++ b/openpype/modules/sync_server/sync_server.py
@@ -206,6 +206,7 @@ class SyncServerThread(threading.Thread):
         self.is_running = False
         self.executor = concurrent.futures.ThreadPoolExecutor(max_workers=3)
         self.timer = None
+        self.before_loop_cmds = []
 
     def run(self):
         self.is_running = True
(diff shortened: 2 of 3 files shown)


Tilix4:
The current issue I figured out and I'm trying to solve is to wait for all before loop commands to complete before the rest of the tasks are executed.

iLLiCiTiT:
Apart from some smaller issues I don't understand how and why is the feature connected to sync server module?

Tilix4:
> Why don't you create the subprocesses directly in publish plugin?

The issue I'm trying to solve is that I need to run subprocesses before the sync server runs the synchronization without slowing down the pyblish by making it wait for the subprocess to end.

BigRoy:
I don't use the sync service myself so I might be off on workflow and terminology.

It sounds like you're trying to delay ANY sync server to sync the files because you want to do some on additional massaging to the file without delaying the current publish process.

As far as I know the sync server MIGHT not be running on your machine alone and might be on another headless server. For as far as I can see you're only really sending this 'delay' system to your local sync server, no? And as far as I know your local machine doesn't need to have direct access to the sync server itself - it could be behind a firewall or alike. All the sync server needs is the access to a file system to find the files to sync and access to the database to know what to sync - it doesn't have to have any connection with any artist's workstation, right?

I think what you might need instead is just add the file in the database **almost** ready to sync but mark it as "not ready" and then from the end of your own subprocess on your machine mark it ready.

Tilix4:
As explained, my goal is to run processes (several and diverse) on recently published representations before they get synched. That is what could slow down the publish because I need to ensure they are finished before the sync loop runs and as they could take some time, the next loop may occur before they're all done.

Actually, thanks for your comments, they helped me to explore other ways to do it and I found one which seems much better but locked because of a bug I describe here: https://github.com/ynput/OpenPype/issues/4311

iLLiCiTiT:
> Actually, thanks for your comments, they helped me to explore other ways to do it and I found one which seems much better but locked because of a bug I describe here: https://github.com/ynput/OpenPype/issues/4311

I still don't understand the source issue, but happy to hear there is a better solution :)

Is it that you're marking representations to be synced during publishing but you want to do something else, like "change paths in workfile" (this is of missing pieces in issue description), before the sync happens?

Tilix4:
> Is it that you're marking representations to be synced during publishing but you want to do something else, like "change paths in workfile" (this is one of missing pieces in issue description), before the sync happens?

That is exactly kind of this operation I want to execute before the sync happens yes! :)

PR #4310, opened 2023-01-11
General: Loader and Creator plugins can be disabled

iLLiCiTiT:
## Brief description
Loader and Creator plugins can be disabled.

## Description
Added `enabled` attribute to `LoaderPlugin` and `LegacyCreator` plugins. Check the value in Loader tool and in Creator tool.

#### Loader 

1. Set `enabled` to `False` on any Loader plugin
2. From that moment should not appear in loader tool

#### Creator

1. Set `enabled` to `False` on any Creator plugin
2. From that moment should not appear in creator tool


kalisp (approved):
Works (in Nuke)

PR #4313, opened 2023-01-13
Maya: Implement image file node loader

BigRoy:
## Brief description

Implements a loader for loading texture image into a `file` node in Maya.

## Description

Similar to Maya's hypershade creation of textures on load you have the option to choose for three modes of creating:
- Texture
- Projection
- Stencil

These should match what Maya generates if you create those in Maya.

- [x] Load and manage file nodes
- [x] Apply color spaces after #4195 
- [x] Support for _either_ UDIM or image sequence - currently it seems to always load sequences as UDIM automatically.
    - [ ] Add support for animation sequences of UDIM textures using the `<f>.<udim>.exr` path format?

## Additional info

This was implemented during the work on #4283 and tested against image output publishes.
Here's an example video where at the end I'm loading the images into maya for shading:

https://user-images.githubusercontent.com/2439881/212285701-c5351ba8-35cb-4015-9118-6552837ff5d6.mp4

Currently none. Should there be?

## Testing notes:

1. Make sure you have a published image (or udim sequence) with family "image", "plate" or "render" and an "exr", "tif", "png" or "jpg" representation.
2. Load away!


BigRoy:
It should be usable. I can try and pull latest develop into it today to resolve merge conflicts.

Other than that it's the open tasks in the first comment of the PR:

- I am not sure with all the PRs what the current color management state is in Maya (or what the way is to generate any color managed publish currently) to tick of the color management checkbox.

- And I am not sure what is the easiest way for a loader to identify whether what is being loaded is UDIM or a sequence. Any pointers?

LiborBatek (changes requested):
Did tests with loading EXR sequence as

- texture
- projection
- stencil

Works fine and loads as maya ```file node``` in ```Hypershade```

However needs solving the issue with ```file sequence``` being detected as ```UDIM``` probably because of existing file with frame ```1001``` acting as default ```UDIM``` and following frames as UDIM tiles (1002, 1003, 1004 etc.)

Need to be detected and probably switched to ```UV Tiling Mode``` to ```Off``` and set ```Use Image Sequence``` instead.

openpype/hosts/maya/plugins/load/load_image.py
```diff
@@ -0,0 +1,176 @@
...
+            cmds.setAttr(file_node + ".fileTextureName", path, type="string")
+
+            # Set UV tiling mode if UDIM tiles
+            # TODO: Detect UDIM tiles and set accordingly (also on update)
+            cmds.setAttr(file_node + ".uvTilingMode", 3)    # UDIM-tiles
+
+            # Enable sequence if publish has `startFrame` and `endFrame` and
+            # `startFrame != endFrame`
+            # TODO: Detect sequences (also on update)
+            # cmds.setAttr(file_node + ".useFrameExtension", True)
```
moonyuet, line 134: Meanwhile, i am thinking about something like this 
```
version = context["version"]
version_data = version.get("data", {})
frameStart = version_data.get("frameStart")
frameEnd = version_data.get("frameStart")
if frameStart != frameEnd:
      cmds.setAttr(file_node + ".useFrameExtension", 1)
else:
      cmds.setAttr(file_node + ".useFrameExtension", 0)
```
but also I am thinking about if udim and framerange will mix up at the same time.

BigRoy: I think we should somehow be able to do something like:

```python
is_frame = representation.get('frame') is not None
is_udim = representation.get('udim') is not None
```

I recall **udim** and **frame** getting stored with the representations - but not sure where it's stored. @iLLiCiTiT might know what _the_ way is to detect whether a publish is frame dependent and/or udim based or not.

moonyuet: As far as i know, the frame comes before the udim in terms of the file naming convention. 
Something like this : {published_file_name}.{frame}_{udim}.{ext}. 
e.g.: workfile_v001.1230_1001.exr

iLLiCiTiT: We store first frame and first udim (if were used). `repre_doc["context"].get("frame")` or `repre_doc["context"].get("udim")`

BigRoy: This should be implemented now - the code is a bit much maybe. Let me know what you think. It's trying to be as explicit as possible about udim and frames if they are defined in the original template upon publishing.

This logic would NOT work for any file that is published with a `{originalFilename}` template instead. @iLLiCiTiT @mkolar thoughts?

BigRoy:
Recent changes should now have added support for:
- UDIM tiles vs frame based images
    - Be aware: ONLY if publish is not published using `{originalFilename}` template. I'm not sure what to do in the other case?
- Setting colorspace value based on colorspace data of the publish and if not detected use OpenPype's imageio filerules for maya from openpype project settings.

Please review - the code is a bit on the heavier side now with that support added in.

LiborBatek (approved):
Now when I use ```Load File Node (exr)``` correctly sets ```Use Image Sequence``` not ```UDIM``` tiling and also sets colorspace according to maya file rules (RAW).

Note: Because not able to test ```File Texture``` publishing from ```Substance Painter``` (not functional unsupported Steam/old painter version)  I cant test it loading ```UDIM``` tiled file texture and if correctly set in ```file node``` in maya when loaded.

File sequences in exr work fine without any glitches.

image

BigRoy:
@moonyuet Any chance you're able to test some other cases for this PR?

moonyuet:
> @moonyuet Any chance you're able to test some other cases for this PR?

sorry, i arrived Prague in the late evening, and not able to test out anything yet. Will test it out tmr.

moonyuet (approved):

BigRoy:
@iLLiCiTiT Feel free to merge this if you think the code is ok.

mkolar:
Thank you @BigRoy looking good

PR #4315, opened 2023-01-13
Maya: Bug - Multiverse extractor executed on plain animation family

antirotor:
## Fix

Fix for when animation was published from rig, `Extract Multiverse USD` was triggered even if no usd extraction was required.

### Testing

Publish animation from rig.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/extract_multiverse_usd.py
+++ b/openpype/hosts/maya/plugins/publish/extract_multiverse_usd.py
@@ -4,6 +4,7 @@ import six
 from maya import cmds
 from maya import mel
 
+import pyblish.api
 from openpype.pipeline import publish
 from openpype.hosts.maya.api.lib import maintained_selection
 
@@ -156,13 +157,6 @@ class ExtractMultiverseUsd(publish.Extractor):
         return members
 
     def process(self, instance):
-        # check if there is "usd" in families because of
-        # `ExtractMultiverseUsdAnim` that inherits this and should
-        # run on animation family too.
-        families = set(instance.data["families"])
-        families.add(instance.data["family"])
-        if "usd" not in families:
-            return
 
         # Load plugin first
         cmds.loadPlugin("MultiverseForMaya", quiet=True)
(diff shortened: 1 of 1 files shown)


openpype/hosts/maya/plugins/publish/extract_multiverse_usd.py
```diff
@@ -156,6 +156,14 @@ def filter_members(self, members):
...
 
     def process(self, instance):
+        # check if there is "usd" in families because of
+        # `ExtractMultiverseUsdAnim` that inherits this and should
+        # run on animation family too.
+        families = set(instance.data["families"])
+        families.add(instance.data["family"])
+        if "usd" not in families:
+            return
+
```
tokejepsen, line 166: Should this not be happening anyways with the class attribute `families`?

antirotor: Problem is actually here:
https://github.com/ynput/OpenPype/blob/cce7c3be0839ff9eb2b603dd6239ce777ddaeec6/openpype/hosts/maya/plugins/publish/extract_multiverse_usd.py#L257

The second class defined there is using the same `process()` but defines class attribute families as just `["animation"]`

tokejepsen: Ah that makes more sense.

BigRoy: What about setting the other's class families like:
```
families = ["usd", "animation"]
```

And customizing its matching algorithm:
```python
matching = pyblish.api.Exact
```

Or even:
```python
matching = pyblish.api.Subset
```

---

Or if we're just trying to resolve for the other plug-in's process, how about actually doing that?

```python
...
    def process(self, instance):
        families = set(instance.data["families"])
        if "usd" not in families:
            return
    
        super(ExtractMultiverseUsdAnim, self).process(instance)
...
```

However, I'd recommend the solution mentioned above using another mathing algorithm.

antirotor: Cool, I've actually never used these, good to know it is there!

PR #4318, opened 2023-01-16
Slack: user groups speedup

kalisp:
## Brief description
Query user and group information only if necessary.

## Description
When message contains '@' character, translation between users and Slack user id is necessar. This operation might hit rate limit and slows down publish process.
Trigger this querying only if absolutely necessary.

## Testing notes:
1. publish something that should trigger Slack notification
2. compare messages with and without @ character

### DIFF
--- a/openpype/modules/slack/plugins/publish/integrate_slack_api.py
+++ b/openpype/modules/slack/plugins/publish/integrate_slack_api.py
@@ -62,8 +62,16 @@ class IntegrateSlackAPI(pyblish.api.InstancePlugin):
                     client = SlackPython3Operations(token, self.log)
 
                 if "@" in message:
-                    if users is None:
+                    cache_key = "__cache_ids"
+                    slack_ids = instance.context.data.get(cache_key, None)
+                    if slack_ids is None:
                         users, groups = client.get_users_and_groups()
+                        instance.context.data[cache_key] = {}
+                        instance.context.data[cache_key]["users"] = users
+                        instance.context.data[cache_key]["groups"] = groups
+                    else:
+                        users = slack_ids["users"]
+                        groups = slack_ids["groups"]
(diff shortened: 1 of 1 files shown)


64qam (approved):

openpype/modules/slack/plugins/publish/integrate_slack_api.py
```diff
@@ -60,8 +61,10 @@ def process(self, instance):
                 else:
                     client = SlackPython3Operations(token, self.log)
 
-                users, groups = client.get_users_and_groups()
-                message = self._translate_users(message, users, groups)
+                if "@" in message:
+                    if users is None:
+                        users, groups = client.get_users_and_groups()
```
BigRoy, line 66: What's the speed of `client.get_users_and_groups()`? Is it **very slow?**

Since this is an InstancePlugin would it be maybe worth caching the result in `context.data` on the first run and any consecutive instances would re-use the value.

For e.g. 10 instances that'd bring it back from 10 calls to a single call.

openpype/modules/slack/plugins/publish/integrate_slack_api.py
```diff
@@ -62,8 +62,16 @@ def process(self, instance):
                     client = SlackPython3Operations(token, self.log)
 
                 if "@" in message:
-                    if users is None:
+                    cache_key = "__cache_ids"
```
BigRoy, line 65:
```suggestion
                    cache_key = "__cache_ids"
```

Note that the context data is accessible by all plug-ins. It might be useful to avoid clashing with anything else that'll try to cache ids or alike like this.

Would: `self.__class__.__name__ + "__cache_ids` be a good candidate? or at least `__slack_user_cache`?

64qam (approved):

PR #4319, opened 2023-01-16
3delight support within OP, using most of its existing infrastructure.

### LINKED ISSUE #2883: 3Delight Support for Maya

antirotor:
This is #4045 with resolved conflicts. We're merging it even though there are some minor issues that can be patched later on.

---

## Brief description
Adding 3delight render support to OpenPype

## Description
Adding support to render via Maya on 3delight using the Deadline service. This PR adds the necessary structures and changes within OP to invoke rendering. 

Given that Deadline hasn't updated support for 3delight since 3delight switched to NSI, we had to add support for "render layers", via 3delight's dlRenderSettings via a pre-render MEL command, and also startFrame and endFrame too. This uses maya's render layers just for their names to map into the dlRenderSetting to use.

## Additional info
Given that 3delight uses dlRenderSettings instead of Maya's Render Layers to manage all rendering information, this poses an issue whenever a change in OP's way of handling renders changes. I have tried to get OP to ignore settings and valiations for 3delight when it's active, and hopefully no more changes will be needed.

- camera needs to be set under the dlRenderSettings
image

Resolves #2883

[cuID:OP-5491]


openpype/hosts/maya/api/lib_rendersettings.py
```diff
@@ -300,3 +306,79 @@ def _additional_attribs_setter(self, additional_attribs):
...
+            cmds.setAttr(
+                "{}.outputOptionsDefault".format(dl_render_setting), 2)
+
+        """
+        3delight doesn't use maya's render layers to indicate what to render,
+        instead, it uses dlRenderSettings as specified by the connection to
+        dlRenderGlobals1, of which there is only one. Since this is not the
+        "mayaonic" way, we must insert a pre-render MEL into maya's
+        "defaultRenderGlobals.preMel", which will help us work this out. We
+        wouldn't this this ... "hack" if deadline allowed us to insert commands
```
BigRoy, line 342:
```suggestion
        wouldn't do this ... "hack" if deadline allowed us to insert commands
```

openpype/hosts/maya/api/lib_rendersettings.py
```diff
@@ -300,3 +306,79 @@ def _additional_attribs_setter(self, additional_attribs):
...
+/*--**--*/"""
+
+        preMel = cmds.getAttr("defaultRenderGlobals.preMel")
+        print("This is our current 'preMel':[{}]".format(preMel))
+        if preMel is None:
+            print("  - we need to insert our own preMel")
+            preMel = PREMEL_TEMPLATE
+        else:
+            print("  - we have preMel update it")
+            preMel = re.sub(regex_match, PREMEL_TEMPLATE, preMel)
```
BigRoy, line 382: This is invalid - a premel script might already exist that MAY not contain the original premel script to update.
This should first do a `re.search` and if not found to APPEND the script.

openpype/hosts/maya/api/lib_rendersettings.py
```diff
@@ -300,3 +306,79 @@ def _additional_attribs_setter(self, additional_attribs):
...
+        wouldn't this this ... "hack" if deadline allowed us to insert commands
+        into its MayaBatch stuff, or, if it allowed setting up renderers in a
+        sensible way.
+        """
+
+        # We need the DOTALL since the regex will now match across newlines.
+        regex_match = re.compile(r"\/\*--\*3dl_v\d+\*--\*\/.*\/\*--\*\*--\*\/",
+                                 re.DOTALL)
+        PREMEL_TEMPLATE = """/*--*3dl_v4*--*/
+currentTime -e `getAttr ("defaultRenderGlobals.startFrame")`;
```
BigRoy, line 351: Why do we need to set the current time to the start frame as a pre-render script? Also, doesn't this logic enforce that it'll always render the full frame range somehow?

I feel like this particular pre-render snippet is lacking some comments to explain what the code itself is doing to resolve the issue at hand.

The lengthy comment above does tell us that there's something to work around, but it's not telling us how it's solving the problem at hand. From just reading the comment and reading the mel script I have still no idea why this works the way it does and how it solves what it intends.

I think we should clarify this code for future maintainers.

BigRoy (commented):
Some notes.

openpype/hosts/maya/api/lib_rendersettings.py
antirotor: I guess this works by updating only its own part using regex

BigRoy:
What's needed to get this merged from here? :)

This might need a new merge with latest develop and maybe some folks to test the functionality? Maybe if prepared @dmo-j-cube and @m-u-r-p-h-y could push it through a new test run?

LiborBatek (changes requested):
Did try to ```publish``` render to Deadline and get this error during publishing:

image

The scene contains 1 renderlayer after creating ```renderMain``` instance using OP (the name is as 3delight modify it?)

BigRoy:
> Did try to publish render to Deadline and get this error during publishing:

Seems like it'll need to be updated to support the changes made here: https://github.com/ynput/OpenPype/pull/4497

> The scene contains 1 renderlayer after creating renderMain instance using OP (the name is as 3delight modify it?)

Could you elaborate. What's the issue?

openpype/hosts/maya/api/lib_renderproducts.py
```diff
@@ -1421,9 +1422,161 @@ def get_render_products(self):
...
+
+    Warning:
+        This is very rudimentary and needs more love and testing.
+    """
+
+    renderer = "_3delight"
+
+    def __init__(self, layer, render_instance):
+        super(RenderProducts3Delight, self).__init__(layer, render_instance)
+
```
moonyuet, line 1439:
```suggestion
    def get_multipart(self):
        return False
  
```
I think you need to include this as the collector of render layer errors out if they find there is no get_multipart args

LiborBatek:
> > Did try to publish render to Deadline and get this error during publishing:
> 
> Seems like it'll need to be updated to support the changes made here: #4497
> 
> > The scene contains 1 renderlayer after creating renderMain instance using OP (the name is as 3delight modify it?)
> 
> Could you elaborate. What's the issue?

Well its just a bit weird renderLayer name by defaults ```dlRenderSettings1_RL``` but maybe its not an issue.

BigRoy:
> Well its just a bit weird renderLayer name by defaults `dlRenderSettings1_RL` but maybe its not an issue.

Ah yes, that doesn't sound like a very nice default renderlayer name.

antirotor:
It doesn't take frame range from render settings od 3delight, it takes it from the timeline settings.
image

LiborBatek (commented):
The ```renderMain``` got published ok and is successfully submitted to DL. Also ```validator``` for ```frame range``` kicks correctly in and ```Repair``` action fixes the frame range accordingly to DB.

image

The ```Repair``` action for ```Frame Range``` uses values from DB and not from ```dlRenderSettings1.frameRange``` as @antirotor  found out already. Once set via this repair action its valid for the shot but not set in the Render Settings.

LiborBatek:
Another 3delight integration issue is whenever user needs to create ```RenderMain``` instance again for some reason and there is already existing renderLayer called ```dlRenderSettings1_RL``` it creates a new one with ```RenderMain``` called ```dlRenderSettings1_RL1``` which is not a valid name when trying to ```Publish``` render.

image

OP expects name in different format ```RenderLayerName_RL``` instead. This logic should be changed or fixed the behavior of creating the render instance in preceding step.

image

moonyuet:
> Another 3delight integration issue is whenever user needs to create `RenderMain` instance again for some reason and there is already existing renderLayer called `dlRenderSettings1_RL` it creates a new one with `RenderMain` called `dlRenderSettings1_RL1` which is not a valid name when trying to `Publish` render. Seems more like a Duplicate of the layer not creating fresh new one (judging by the suffix).
> 
> image
> 
> OP expects name in different format `*putAnything*_RL` instead. This logic should be changed or fixed the behavior of creating the render instance in preceding step.
> 
> image

The frame range has been fixed. 
I find out that too and I am still exploring to force the naming to "_RL" . But it seems that not quite possible at the current stage as there is a parent class to set up the "_rendering" set under the rendering main. Will need time to investigate that.

LiborBatek (changes requested):
Now ```frame range``` being correctly taken from ```Render Settings``` 
image

but the ```Repair``` action when trying to ```Publish``` doesnt work 100%. It worked once but when trying changing ```frame range``` to break it again it doesnt repair it on second time...

Didnt found any exact way to replicate it other than trying different and invalid ```frame range``` values several times to trigger ```Repair``` action again....resulting of not repairing it at all.

openpype/hosts/maya/api/lib_renderproducts.py
BigRoy: This seems resolved?

openpype/hosts/maya/api/lib_rendersettings.py
BigRoy: But if there's no match than it doesn't add it? There's something invalid here. It's either expecting it to be there to get updated or for premel to be empty. It's not expecting the case the premel doesn't have the script yet but it does have some other value. Right?

openpype/hosts/maya/plugins/publish/validate_rendersettings.py
```diff
@@ -125,6 +125,10 @@ def get_invalid(cls, instance):
         layer = instance.data['setMembers']
         cameras = instance.data.get("cameras", [])
 
+        if renderer == "_3delight":
+            # 3delight is different enough to need it do it its own validation
+            return invalid
```
BigRoy, line 130: This validator is already massive, ha! But it stills feels weird we're completely skipping this without any debug log whatsoever.

BigRoy (commented):
Just commenting some things that currently stood out.

openpype/hosts/maya/api/lib_renderproducts.py
moonyuet: yes it is

LiborBatek (changes requested):
Now ```Set Frame Range``` works correctly, setting both ```Time line``` and ```Frame Range``` in Render Settings of 3delight.

When trying to delete and recreate ```Render``` Instance OP correctly connects the original (first) render layer called ```dlRenderSettings1_RL```.  But by this action (recreation of instance using OP Creator) it also creates new render layer each time user trigger this action.

Its not preffered should be checking if already renderlayer exists and connect it. If not existing any then just creating new one.

image

antirotor:
I've resolved the conflicts. I would merge it as it is if it is working and not breaking anything else and lets enhance it later when it gets some traction.

antirotor:
So I've added some additional small fixes, but there are issues with this. Apart from the specific step to get `dlSettings` node - like that render layer name has to end with `_RL` suffix and the stuff mentioned above by @BigRoy there seems to be issue with the support of 3delight within Deadline. `MayaBatch` plugin is calling non-existent functions from 3delight - at least that happened for me on DL 10.2 and 3delight 2.9.27.

I am still inclined to merge it after all just because it isn't touching much of the other code, even though there is a lot to improve in a future.

mkolar:
This poor PR has conflicts again. I reckon let's fix those and merge it. It is really additive and shouldn't affect other areas in a big way.

tokejepsen:
@antirotor what is the state of this PR?
Not needed anymore?

mkolar:
Because we're splitting OpenPype into ayon-core and individual host addons, this PR would have to be ported to target one of those.

PR #4321, opened 2023-01-16
Blender: Blend File Family

simonebarbieri:
## Brief description
Implementation of the Blend File family analogue to the Maya Scene one.

### DIFF
--- a/openpype/hosts/blender/plugins/publish/extract_blender_scene_raw.py
+++ b/openpype/hosts/blender/plugins/publish/extract_blender_scene_raw.py
@@ -1,0 +1,93 @@
+import os
+
+import bpy
+
+from openpype.pipeline import AVALON_CONTAINER_ID, publish
+from openpype.hosts.blender.api.workio import save_file
+
+
+class ExtractBlenderSceneRaw(publish.Extractor):
+    """Extract as Blender Scene (raw).
+
+    This will preserve all references, construction history, etc.
+    """
+
+    label = "Blender Scene (Raw)"
+    hosts = ["blender"]
+    families = ["blenderScene", "layout"]
+    scene_type = "blend"
+
+    def process(self, instance):
+        # Define extract output file path
+        dir_path = self.staging_dir(instance)
+        filename = "{0}.{1}".format(instance.name, self.scene_type)
+        path = os.path.join(dir_path, filename)
+
+        # We need to get all the data blocks for all the blender objects.
+        # The following set will contain all the data blocks from version
+        # 2.93 of Blender.
+        data_blocks = {
+            *bpy.data.actions,
+            *bpy.data.armatures,
+            *bpy.data.brushes,
+            *bpy.data.cache_files,
+            *bpy.data.cameras,
+            *bpy.data.collections,
+            *bpy.data.curves,
+            *bpy.data.fonts,
+            *bpy.data.grease_pencils,
+            *bpy.data.images,
+            *bpy.data.lattices,
+            *bpy.data.libraries,
+            *bpy.data.lightprobes,
+            *bpy.data.lights,
+            *bpy.data.linestyles,
+            *bpy.data.masks,
+            *bpy.data.materials,
+            *bpy.data.metaballs,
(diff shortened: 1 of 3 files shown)


openpype/hosts/blender/plugins/publish/extract_blender_scene_raw.py
```diff
@@ -0,0 +1,93 @@
...
+            *bpy.data.scenes,
+            *bpy.data.screens,
+            *bpy.data.shape_keys,
+            *bpy.data.speakers,
+            *bpy.data.sounds,
+            *bpy.data.texts,
+            *bpy.data.textures,
+            *bpy.data.volumes,
+            *bpy.data.worlds
+        }
```
BigRoy, line 64: Just wondering, wouldn't something dynamic like this work instead?

Like getting all members that are `bpy.types.struct` or even more specific are an [instance of `bpy.types.ID`](https://docs.blender.org/api/current/bpy.types.ID.html#bpy.types.ID) 

```python
data_blocks = set()
for attr in dir(bpy.data):
    data_block = getattr(bpy.data, attr)
    if isinstance(data_block, bpy.struct):
        data_blocks |= (*data_block)
```
Or:
```python
import inspect

def predicate(obj):
    return isinstance(obj, bpy.types.ID)

data_blocks = set()
for data_block in inspect.getmembers(byp.data, predicate):
    data_blocks |= (*data_block)
```

Please consider it pseudocode. :) I just wonder whether there's a way to get all the data dynamically?
I'm just wondering whether that might be more 'future-compatible'?

simonebarbieri: I've actually tried to get the list of collection of data blocks with blender built-in functions, and didn't think of using `getattr` 😅 I'm going to commit this in a few minutes

moonyuet:
So far so good. Looks good in my side for publishing blender scene in my side. I have something in my mind but I dont think it's neccessary for this extractor as it is not publishing and submitting to the renderfarm(the option for automatically packed resources , and it will increase the blend file size)
image
image

BigRoy:
> the option for automatically packed resources , and it will increase the blend file size

Not using blender myself - how does that work. Is that [what's roughly described here?](https://blender.stackexchange.com/a/143043/67296)

moonyuet:
> > the option for automatically packed resources , and it will increase the blend file size
> 
> Not using blender myself - how does that work? Is that [what's roughly described here?](https://blender.stackexchange.com/a/143043/67296)
> 
> If that's workable and option on the instance could be nice for that?

yes, the options for automatically packed resources is to pack all the textures into one scene. It can prevent the data loss of the texture files if you "carelessly" delete your texture sets. I usually used this to save the blend scene before submitting to the renderfarm but will make the file size big.

BigRoy:
@simonebarbieri what's needed to continue work on this and get it ready for review?

simonebarbieri:
Extremely late, but an update on this PR!

I'm restructuring a bit the PR. I've had another look at the handling of .blend files with OP, and realized that it shouldn't be needed a blend family, but a better handling of data published as blend, more or less as `load_reference` in Maya.

I'll probably get to the implementation of a `blenderScene` similar to the `mayaScene`, but I'll rework the loading of blend representations first.

LiborBatek (approved):
BlendScene family works as supposed.

I did try to Publish the ```blendScene``` without problems, load it via ```Loader``` and finally
manage its versions via ```Manage```...and all working fine.

image

openpype/hosts/blender/plugins/create/create_blendScene.py
```diff
@@ -0,0 +1,51 @@
...
+    label = "Blender Scene"
+    family = "blendScene"
+    icon = "cubes"
+
+    def process(self):
+        """ Run the creator on Blender main thread"""
+        mti = ops.MainThreadItem(self._process)
+        ops.execute_in_main_thread(mti)
+
+    def _process(self):
```
Minkiu, line 23: When does this one get executed? And should we have the `import bpy` here instead of at the top?

PR #4324, opened 2023-01-16
General: Functions for current context

iLLiCiTiT:
## Brief description
Defined more functions to receive current context information and added the methods to host integration so host can affect the result.

## Description
This is one of steps to reduce usage of `legacy_io.Session`. This change define how to receive current context information -> call functions instead of accessing `legacy_io.Session` or `os.environ` directly. Plus, direct access on session or environments is unfortunatelly not enough for some DCCs where multiple workfiles can be opened at one time which can heavily affect the context but host integration sometimes can't affect that at all.

`HostBase` already had implemented `get_current_context`, that was enhanced by adding more specific methods `get_current_project_name`, `get_current_asset_name` and `get_current_task_name`. The same functions were added to `~/openpype/pipeline/cotext_tools.py`. The functions in context tools are calling host integration methods (if are available) otherwise are using environent variables as default implementation does. Also was added `get_current_host_name` to receive host name from registered host if is available or from environment variable.

## Additional info
Didn't change usages of `legacy_io.Session` anywhere. We should make a decision and agree on an approach and then start replace current usages. Anyway there must be backwards compatibility for some time.

### Possible changes
- Use `legacy_io.Session` instead of `os.environ`. But I wanted to avoid that as the goal is to reduce usage of  `legacy_io`. Those functions are for "reading", "change" of current context in legacy_io is changing environemnts too (if does not then it should).
- Add `set_current_context` function to replace `change_current_context` and some logic in workfiles tool.
- The functionality is doubled in default implementation of host and in context tools (I didn't find out good or better solution).

## Testing notes:
It's more about decisions. Any comments are welcome

### DIFF
--- a/openpype/pipeline/context_tools.py
+++ b/openpype/pipeline/context_tools.py
@@ -306,6 +306,22 @@ def debug_host():
     return host
 
 
+def get_current_host_name():
+    """Current host name.
+
+    Function is based on currently registered host integration or environment
+    variant 'AVALON_APP'.
+
+    Returns:
+        Union[str, None]: Name of host integration in current process or None.
+    """
+
+    host = registered_host()
+    if host is not None and hasattr(host, "name"):
+        return host.name
+    return os.environ.get("AVALON_APP")
+
+
 def get_global_context():
     return {
         "project_name": os.environ.get("AVALON_PROJECT"),
--- a/openpype/pipeline/context_tools.py
+++ b/openpype/pipeline/context_tools.py
@@ -306,6 +306,42 @@ def debug_host():
     return host
 
 
+def get_global_context():
+    return {
+        "project_name": os.environ.get("AVALON_PROJECT"),
+        "asset_name": os.environ.get("AVALON_ASSET"),
+        "task_name": os.environ.get("AVALON_TASK"),
+    }
+
+
+def get_current_context():
+    host = registered_host()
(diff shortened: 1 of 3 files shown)


openpype/pipeline/context_tools.py
```diff
@@ -306,6 +306,58 @@ def ls():
...
+def get_current_context():
+    host = registered_host()
+    if host is not None and hasattr(host, "get_current_context"):
+        return host.get_current_context()
+    return get_global_context()
+
+
+def get_current_project_name():
+    host = registered_host()
+    if host is not None and hasattr(host, "get_current_project_name"):
```
kalisp, line 342: I would personally prefer:
```
if isinstance(host, HostBase):
    host.get_current_context()
```
but thats just my preference, couldn't find anything that this would be far superior to `hasattr` (just a bit less cluttered and not depending on a rigid method name, but I guess if wrapper name would need to be changed, we would need to change method names on the class).

And I would like docstring to highlight it coms from Host or from global context for code assist.

kalisp (approved):
Just nitpicking on condition format and missing docstring, other way it looks good.

(Maybe just one thing, shouldn't these be also in `pipeline/__init__.py` to highlight them a bit more?

iLLiCiTiT:
Agree with both comment and added them...

PR #4327, opened 2023-01-17
Maya: gltf textures and shaders export

moonyuet:
## Brief description
Continuity of the gltf extractor https://github.com/pypeclub/OpenPype/pull/4192
Validating the texture naming conventions from the StingrayPBS Material with the UE5 naming style and converting the material into glsl shaders with the maya2gltf very own pbr shader scripts.

## Description
**Validator for GLTF Texture Names**: validate the name of the texture file matches with the UE5 naming style
  The texture naming conventions follows the UE5-style-guide: https://github.com/Allar/ue5-style-guide#anc-textures-packing
*ORM: Occulsion Roughness Metallic Packed Textures
    Texture Naming Style:
    Albedo/Diffuse: {Name}_D.{imageExtension} or
                    {Name}_D.<UDIM>.{imageExtension}
    Normal: {Name}_N.{imageExtension} or
            {Name}_N.<UDIM>.{imageExtension}
    ORM: {Name}_ORM.{imageExtension} or
         {Name}_ORM.<UDIM>.{imageExtension}

For example: If the diffuse map is shirt_123_Default Fabric_FRONT_247310_D.png, it passes the validation.

**Texture Convertor**: It gets the related textures from the StingrayPBS material of an asset , connect to the newly created glsl shader with maya2gltf pbr shader script, and assign the glsl shader to the asset.
## Additional info
N/A

## Testing notes:
1. Enable the extractor in Openpype setting
2. Create instance which is associated to the extractor(model, animation, point cache)
3. Assign the PBS Shaders with correct texture names into the asset
4. Publish it
5. You will find the glb file in the folder(s) of the related instance in the publish folder
6. Test the glb file in https://sandbox.babylonjs.com/ and you will find the textured asset.


jakubjezek001:
Task linked: [OP-4668 Maya: glTF textures and shaders export](https://app.clickup.com/t/6658547/OP-4668)

openpype/hosts/maya/plugins/publish/validate_gltf_textures_names.py
```diff
@@ -0,0 +1,130 @@
...
+                # diffuse texture file node
+                albedo = cmds.listConnections(shader + ".TEX_color_map")
+                if not albedo:
+                    invalid.add(albedo)
+                orm_packed = cmds.listConnections(shader + ".TEX_ao_mapX")
+                if not orm_packed:
+                    invalid.add(orm_packed)
+                nrm = cmds.listConnections(shader + ".TEX_normal_map")
+                if not nrm:
+                    invalid.add(nrm)
```
antirotor, line 108: I think this is causing that if you don't have texture pack connected - ie. all of these file nodes to `TEX_color_map`, `TEX_ao_mapX`, `TEX_normal_map` then it will fail. My testing case is simple shader with only one diffuse texture `foo_D.png` connected to `TEX_color_map` and it is triggering this validator with `Related texture file nodes not connected`.

antirotor (changes requested):
The issue mentioned in the previous PR is resolved, but this another one  in fact impose the need to always  connect diffuse, AO and normal map.

antirotor (approved):

LiborBatek:
Did test and followed the steps 1-6. All worked and asset got published successfully.

here is my Stingray mat with textures connected:
image

and here in the GLTF viewer and publish folder:
image

image

The only thing which is not certain that in the GLTF viewer I just see DIFFUSE and NORMAL maps but no ORM map (Occlusion, Roughness, Metalness). Not sure if ```glb file``` is correctly exported. Please see my material editor screen for consideration.

openpype/hosts/maya/plugins/publish/convert_gltf_shader.py
```diff
@@ -0,0 +1,78 @@
...
+
+from openpype.pipeline import publish
+
+
+class ConvertGLSLShader(publish.Extractor):
+    """
+    Converting StingrayPBS material to GLSL Shaders
+    specially for the glb export through Maya2GLTF plugin
+
+    """
```
BigRoy, line 13: I personally think it's weird this is changing the local work scene. A plug-in shouldn't do that.
A validator could do that potentially through a repair method if needed.

I also feel that this plug-in on family `gltf` might be too specific to a certain workflow. I've done about 4-5 gltf projects using this maya2gltf plug-in and I've never had the need to use `StingrayPBR` so far. Should this be a `gltf.unreal` or whatever subfamily? 

I'm not sure what the naming conventions are based on, etc. but it might be good in the docstring to mention and clarify that. Is it unreal? Is it something else?

openpype/hosts/maya/plugins/publish/validate_gltf_textures_names.py
```diff
@@ -0,0 +1,126 @@
...
+        return list(invalid)
+
+    def shader_selection(self, instance):
+        shapes = cmds.ls(instance, type="mesh", long=True)
+        for shape in shapes:
+            shading_grp = cmds.listConnections(shape,
+                                               destination=True,
+                                               type="shadingEngine")
+
+            return shading_grp
```
BigRoy, line 126: Function name seems a bit weird. `get_material_from_shapes()` maybe more sensible?

openpype/hosts/maya/plugins/publish/validate_gltf_textures_names.py
```diff
@@ -0,0 +1,126 @@
...
+    def get_texture_shader_invalid(self, instance):
+
+        invalid = set()
+        shading_grp = self.shader_selection(instance)
+        for material in shading_grp:
+            main_shader = cmds.listConnections(material,
+                                               destination=True,
+                                               type="StingrayPBS")
+            if not main_shader:
+                invalid.add(material)
```
BigRoy, line 116: Also are you sure the logic here is correct? Usually the shadingGroup is on the source side of the shading groups.

Additionally it feels like you're mixing shading groups/engines and materials making the variables confusing.
The shading engine / shader group is usually the shader. (List under Shading Groups in Hypershade)
The material is what connects into it, e.g. the StingrayPBS material. (Listed under Materials in Hypershade) 

```suggestion
        shading_groups = self.shader_selection(instance)
        for shading_group in shading_groups :
            material = cmds.listConnections(shading_group,
                                            source=True,
                                            destination=False,
                                            type="StingrayPBS")
            if not material:
                invalid.add(material)
```

I think you're even looking for materials that are connected to the shading group's `surfaceShader` attribute so it might be good to explicitly check that toon, e.g. by doing this:

```python
material = cmds.listConnections("{}.surfaceShader".format(shading_group),
                                source=True,
                                destination=False,
                                type="StingrayPBS")
```

openpype/hosts/maya/plugins/publish/validate_gltf_textures_names.py
```diff
@@ -0,0 +1,126 @@
...
+    Validate if the asset uses StingrayPBS material before conversion
+    of the GLSL Shader
+    Validate if the names of GLTF Textures follow
+    the packed ORM/ORMS standard.
+
+    The texture naming conventions follows the UE5-style-guides:
+    https://github.com/Allar/ue5-style-guide#anc-textures-packing
+
+    ORM: Occulsion Roughness Metallic
+    ORMS: Occulsion Roughness Metallic Specular
```
BigRoy, line 19:
```suggestion
    ORM: Occlusion Roughness Metallic
    ORMS: Occlusion Roughness Metallic Specular
```

openpype/hosts/maya/plugins/publish/validate_gltf_textures_names.py
BigRoy: Also, it's adding `None` into `invalid`

I'd also, for these cases explicitly check for input connections (source=True) and disregard outgoing connections (destination=False). Rather be exact and explicit about what we're checking :)

openpype/hosts/maya/plugins/publish/convert_gltf_shader.py
```diff
@@ -0,0 +1,78 @@
...
+            # create glsl shader
+            glsl = cmds.createNode('GLSLShader')
+            glsl_shadingGrp = cmds.sets(name=glsl + "SG", empty=True,
+                                        renderable=True, noSurfaceShader=True)
+            cmds.connectAttr(glsl + ".outColor",
+                             glsl_shadingGrp + ".surfaceShader")
+
+            # load the maya2gltf shader
+            maya_dir = os.getenv("MAYA_APP_DIR")
+            ogsfx = maya_dir + "/maya2glTF/PBR/shaders/glTF_PBR.ogsfx"
```
BigRoy, line 37: This is incorrect - this would _only_ work if the user had a local install of `maya2glTF`. This should either be configurable in settings or in some other way be defined.

We tended to check the path relative to the maya2gltf module.

openpype/hosts/maya/plugins/publish/convert_gltf_shader.py
```diff
@@ -0,0 +1,78 @@
...
+    hosts = ["maya"]
+    label = "Convert StingrayPBS to GLTF"
+    families = ["gltf"]
+    optional = True
+
+    def process(self, instance):
+        meshes = cmds.ls(instance, type="mesh", long=True)
+        self.log.info("meshes: {}".format(meshes))
+        # load the glsl shader plugin
+        cmds.loadPlugin("glslShader.mll", quiet=True)
```
BigRoy, line 24: I believe this will be Windows-only due to the `.mll` suffix. You should be able to load these without the suffix. For other OS'es it would then load e.g. the `.so` file.

```suggestion
        cmds.loadPlugin("glslShader", quiet=True)
```

openpype/hosts/maya/plugins/publish/convert_gltf_shader.py
moonyuet: The purpose of this validator is to check the textures with the naming standard of ue5 style. We can set this validator as optional and allows users to enable if they want it here, similar to the gltf texture extractor

openpype/hosts/maya/plugins/publish/convert_gltf_shader.py
moonyuet: i am not sure about this actually, As I go through maya2glTF documentation and installation, it seems that it is recommended to be put into the default maya_app_dir plug-in. of course it can be configurable. and we can allow this configurable in the settings. This is the simplest way to define the environment for it but if the user doesn't define anything in the enivornment, I think I will just leave it as os.getenv("MAYA_APP_DIR") by default

openpype/hosts/maya/plugins/publish/validate_gltf_textures_names.py
moonyuet: I need to admit I need more coffee for this particular case, will do the adjustment regarding to that

moonyuet:
I have added the option to enable both validator and extractor for textures in the setting. 
image
image

openpype/hosts/maya/plugins/publish/validate_gltf_textures_names.py
```diff
@@ -0,0 +1,129 @@
...
+        return list(invalid)
+
+    def get_material_from_shapes(self, instance):
+        shapes = cmds.ls(instance, type="mesh", long=True)
+        for shape in shapes:
+            shading_grp = cmds.listConnections(shape,
+                                               destination=True,
+                                               type="shadingEngine")
+
+            return shading_grp
```
antirotor, line 129: wouldn't this return shading group for the first shape it founds?

BigRoy: It would yes. It might have even return None if the first shape had no shading engine assigned.

openpype/hosts/maya/plugins/publish/convert_gltf_shader.py
BigRoy: Publishing should be _idempotent_. Publishing again should do the exact same processing. Because this plugin changes the workfile that's not the case at all.

LiborBatek:
Did another go with tests this time connecting ```Out Color``` RGB channel directly to Occlusion, Roughness, Metallness not just single R,G,B separately.

image

when inspecting published ```glb file``` all textures are present.
image

openpype/hosts/maya/plugins/publish/validate_gltf_textures_names.py
```diff
@@ -0,0 +1,136 @@
...
+                                           destination=True,
+                                           type="StingrayPBS")
+        for shader in main_shader:
+            # diffuse texture file node
+            albedo = cmds.listConnections(shader + ".TEX_color_map")
+            if not albedo:
+                invalid.add(albedo)
+            orm_packed = cmds.listConnections(shader + ".TEX_ao_map")
+            if not orm_packed:
+                continue
```
BigRoy, line 104: Continuing here means we would not run any of the `TEX_normal_map` logic below.

However, looking at the code it seems that since we're just using `continue` here the logic basically does nothing of note at all.

openpype/hosts/maya/plugins/publish/validate_gltf_textures_names.py
```diff
@@ -0,0 +1,136 @@
...
+        shapes = cmds.ls(instance, type="mesh", long=True)
+        for shape in shapes:
+            shading_grp = cmds.listConnections(shape,
+                                               destination=True,
+                                               type="shadingEngine")
+
+            if shading_grp:
+                return shading_grp[0]
+            else:
+                raise RuntimeError("No Shading Engine found in the scene")
```
BigRoy, line 136: This logic will only ever return the material for the FIRST SHAPE in `get_material_from_shapes`. I think it's a bit too big of an assumption that there will only ever be one material assigned to multiple meshes.

This would return all shading engines on the selected objects:
```python
    def get_shading_engines_from_meshes(self, instance):
        shapes = cmds.ls(instance, type="mesh", long=True)
        shading_engines = cmds.listConnections(shapes,
                                               destination=True,
                                               type="shadingEngine")
        return shading_engines or []
```
_(Technically it would return any shadingEngine that is connected as an output from the given mesh shape - _it could potentially still NOT be assigned as a shader - it's an assumption that the shading engine is only ever connected to the mesh if were connected as a set member)_

It would not detect if any of the shapes did NOT have a shader assigned. (Not sure if this validator should check that?)

openpype/hosts/maya/plugins/publish/validate_gltf_textures_names.py
moonyuet: I added this mainly because NoneType Error.

LiborBatek (changes requested):
Works ok when all maps connected to StingrayPBS shader.

But Fails when only one texture is connected to the shader before publishing.

moonyuet:
sorry guys for causing your trouble. I have burnt my own fork and renew again and update the old branch https://github.com/ynput/OpenPype/pull/4261

LiborBatek (approved):
image

All working fine even when just single image texture connected to Stingray shader.
image

PR #4255, opened 2022-12-21
General: Update MacOs to PySide6

iLLiCiTiT:
## Brief description
Update PySide2 to PySide6 for MacOs and change all remaining imports of `Qt.py` module to `qtpy`. This change is needed for new MacOs distribution.

## Description
PR is combining change of Qt.py import with `qtpy` so all changes in this PR are affecting all hosts and it doesn't make sense to have it separated into multiple small PRs as most of them are connected to each other. Imports were changed in tools, utils widgets, launch script of non-python hosts, some vendor modules (`qargparse` and `scriptsmenu`) and global loader plugins. Added `QtPy` module to Python 2 vendor in latest available version that support `PySide` and `PyQt4`.

PySide6 changed few things here and there. E.g. how it is accessed to class methods (objects don't have access to class attributes) or renamed few methods. Some constants were removed and few classes were removed completelly.

We must make sure that all Qt tools in igniter are 100% right. Other possible bugs can be fixed later. Report any issue you will hit (make sure the issue was not already reported).

Checkstate constants are not integers anymore so output of `CheckStateRole` or `stateChanged` of checkbox can't be compared to `Checked`, `Unchecked` and `PartiallyChecked` without convertion to enum values first (big issue for backwards compatibility).

Issue with `qtpy` is that it always set `os.environ["QT_API"]` to best matching value. That causes issue if the value is set to `"pyside6"` because in that case older versions are not set. Currently was fixed for applications launch but it can cause more issues in future.

## TODOs
- try if PySide6 can be used on linux and MacOs

## Known issues
- [x] Loader does not show subsets -> Checkstate issue
- [ ] Settings UI stuck process -> Fix would be to "rewrite" settings, since the PR is affecting MacOs and Ayon switch will happen soon we're **ignoring this bug**
- [ ] Views have different background color of branch (probably stylesheets related) -> We'll have to revisit stylesheets in separated PR (there are reported other style issues)

## Testing notes:
All tools running in OpenPype process are potentially broken now and in most of cases the broken part is not recognized until someone hit exactly specific usecase. Because of `QtPy` module update and using it instead of `Qt.py` functionality in hosts should be checked too. The main focus is on tools running in OpenPype processes on MacOS.

It is required to install proper python and on MacOS also run `fetch_thirdpart_libraries` to install `PySide6`.

### Any possible UI tool executed from OpenPype tray
This consist of all tools from modules and in base of openpype.
- [x] LibraryLoader
- [x] Launcher
- [x] Local settings
[...]

PR #4332, opened 2023-01-17
General: Use class constants of QMessageBox

iLLiCiTiT:
## Brief description
PR fixing usage of constants with QMessageBox with combination with PySide6 (where constants are not available on objects).

## Testing notes:
1. Use workfiles tool where is save changes prompt that crashed
2. This dialog is shown if there are unsaved changes in host
3. The dialog should show and should be possible to select what should happen

Missed that bug in https://github.com/ynput/OpenPype/pull/4255 . There still can be more wrong usage of constants in the code, any reports welcomed.


kalisp (approved):
Works (in Workfile)

PR #4336, opened 2023-01-17
Global: Validate unique subsets

### LINKED ISSUE #4330: Global subset name unique validator publish plugin
Global plugin which validates this is the BEST case scenario since it's valid for basically ALL publishes - unless there are special cases I don't know about where two separate instances should somehow publish and integrate into the same asset+subset from within the same publish.

Here's an old basis for that logic: https://github.com/Colorbleed/colorbleed-config/blob/acre/colorbleed/plugins/global/publish/validate_unique_subsets.py

However as clearly can be seen from that todo it still has the bug that it doesn't check it per asset. I'd be happy to refactor it and set up a PR if we feel this is a worthwhile global plug-in.

_Originally posted by @BigRoy in https://github.com/ynput/OpenPype/pull/4308#discussion_r1071526144_
            

[cuID:[OP-4757](https://app.clickup.com/t/863fw69m7)]

BigRoy:
## Brief description

This implements a global validator whether the current publish will publish unique subsets (per asset).

This fixes #4330 

## Additional info

I've removed seemingly equivalent ones from Maya, Fusion and Photoshop - please double check.
There also appeared to be something relatively similar for Maya: 

It would be good to know whether there are any cases where this logic doesn't make sense. Would we ever allow a single publish have multiple instances set to the same asset+subset. Can we think of a reason?

## Testing notes:

I've only done a quick test in Maya.

1. Create some instances that publish to the same asset and subset names. (This should be **invalidated**!)
2. Create some instances that publish to different asset but match in subset names. (This should be allowed)
3. Create some instances that publish to same asset but differ in subset names. (This should be allowed)
4. Create some instances that publish to the same asset and subset names **but** during publishing deactive/disable the clashing ones. (This should be allowed)


mkolar (approved):
This makes good sense to me. I can't think of a scenario where it would be desirable to publish multiple instances into a same subset from one workfile.

BigRoy:
> I can't think of a scenario where it would be desirable to publish multiple instances into a same subset from one workfile.

Note that technically we're also talking about publishing from standalone publisher instead of 'a workfile' - not sure if that conceptually changes anything about potentially publishing into the same subset?

jakubjezek001 (approved):
I was testing on Nuke 12/13 at new and old publisher and all went smooth.

PR #4340, opened 2023-01-18
Maya: OP-2630 acescg maya

### LINKED ISSUE #2712: Quick way to produce reviews from ACEScg Maya renders
## Problem

Currently we don't handle colorspace information on review level - default ffmpeg profile is set to apply gamma 2.2 and thats it. It works for classic linear workflow, but not for anything else.

We now have requests to create reviews from ACEScg renders. Until #2579 (and discussion #2574) is properly resolved, we need quicker fix. Quick overview what can be done:

- mark project as using ACEScg for renders (collector should pick it up from Maya settings)
- using custom family name (as we currently have no easy mechism to do it differently) mark render instance as ACEScg
- in plugins generation reviews and thumbails, first feed exrs to OIIO and apply color transform and pipe it from OIIO to ffmpeg to encode it

### Questions

- how could that be re-used (with changes ofc) in implementing #2579
- how would it cope with super big exrs memory-wise

[cuID:25r576d]

tokejepsen:
## Brief description
Resolves #2712 

## Additional info
This grabs OCIO colour data and injects as `colorspaceData` to transcode between colourspaces.

This builds on #4195 and #4291 

## Testing Steps
1. Setup recommended output profile `project_settings/global/publish/ExtractOIIOTranscode/profiles`
Capture
2. Setup render in Maya.
3. Publish to Deadline.


jakubjezek001:
Task linked: [OP-2630 Quick way to produce reviews from ACEScg Maya renders](https://app.clickup.com/t/6658547/OP-2630)

BigRoy:
Is this PR to the wrong branch? There's a huge list of changes. :)

openpype/hosts/maya/api/lib.py
```diff
@@ -3446,3 +3447,47 @@ def dgcontext(mtime):
...
+
+    path = cmds.colorManagementPrefs(
+        query=True, configFilePath=True
+    )
+    # Resolve environment variables in config path. "MAYA_RESOURCES" are in the
+    # path by default.
+    for group in re.search(r'(<.*>)', path).groups():
+        path = path.replace(
+            group, os.environ[group[1:-1]]
+        )
```
BigRoy, line 3481: This breaks for me - `MAYA_RESOURCES` is not a defined environment variable on my end. The OCIO config does use it and also works fine in Maya, but I don't have it in my env vars.

Also watch out that the regex pattern here will match from the first `<` to the last `>` so will break on `<MAYA_RESOURCES>/hello/<SOME_OTHER_VAR>/world` since it matches `<MAYA_RESOURCES>/hello/<SOME_OTHER_VAR>` as one group.

Here's another way to substitute the groups:
```python
import re
import os

path = "<MAYA_RESOURCES>/OCIO-configs/Maya2022-default/config.ocio"

def subst_with_env_value(match):
    key = match.group(1)
    return os.environ.get(key, "")

path = re.sub(r'<([^>]+)>', subst_with_env_value, path)
print(path)
```
However due to the environment variable not existing for me that one still resolves to: `/OCIO-configs/Maya2022-default/config.ocio`

---

Unfortunately this also doesn't resolve the path:
```python
# Resolve environment variables in config path. "MAYA_RESOURCES" are in the
# path by default.

path = "<MAYA_RESOURCES%/OCIO-configs/Maya2022-default/config.ocio"

f = om.MFileObject()
f.setRawFullName(path)
print(f.resolvedFullName())
print(f.expandedFullName())
```

tokejepsen: Yeah, `MAYA_RESOURCES` is not an environment set when launching Maya natively so god knows why they included it in the default path for OCIO. We are account for the lacking environment variable here;
https://github.com/tokejepsen/pype/blob/enchancement/OP-2630_acescg_maya/openpype/hosts/maya/startup/userSetup.py#L7-L12

Good point about the regex. Didnt know that. Have replace it now.

BigRoy: Actually - let's just remove it from userSetup.py - let's try to keep that as clean as possible and not influence the actual environment where we don't need to. I don't see why we need to precache something like that - it's not heavy at all to retrieve it? So let's avoid cluttering different locations of the code if we can.

However, I've found a way using the API to get the absolute path to resources.

I'd say for the logic in this method - let's do this:
```python
    path = cmds.colorManagementPrefs(
        query=True, configFilePath=True
    )

    # The OCIO config supports a custom <MAYA_RESOURCES> token
    maya_resources_token = "<MAYA_RESOURCES>"
    resources_path = om.MGlobal.getAbsolutePathToResources()
    path = path.replace(maya_resources_token, maya_resources_path)

    # Should we expand any remaining vars?
    path = os.path.expandvars(path)
```

That seems to be the only 'absolute path' getter on the `MGlobal` so I assume it's the only really special one?

For what it's worth - it seems that V-Ray also does this hardcoded replacement of this token in function `vrayGetMayaOCIOConfig` in `vray_maya_utils.py` . I think we should do, especially because other tokens, e.g. `<MAYA_LOCATION>` don't work even if that is an environment variable. It seems `<MAYA_RESOURCES>` is just special.

openpype/hosts/maya/startup/userSetup.py
```diff
@@ -4,10 +4,16 @@
 from openpype.hosts.maya.api import MayaHost
 from maya import cmds
 
+# MAYA_RESOURCES enviornment variable is referenced in default OCIO path but
+# it's not part of the environment. Patching this so it works as expected.
+if "MAYA_RESOURCES" not in os.environ:
+    os.environ["MAYA_RESOURCES"] = os.path.join(
+        os.environ["MAYA_LOCATION"], "resources"
+    ).replace("\\", "/")
+
```
BigRoy, line 13:
```suggestion
```

Let's remove this. See [comment here](https://github.com/ynput/OpenPype/pull/4340#discussion_r1073969460) which apparently doesn't show on the PR :)

BigRoy:
@tokejepsen I saw this in the V-Ray code base and I suspect it's something we'd also need to watch out for if we are to support Maya older than 2022:

```python
def vrayGetMayaColorManagement():
    '''
    Returns a triple (rendering, display, view) for the active rendering colorspace to view transform
    '''
    rendering = cmds.colorManagementPrefs(q=True, renderingSpaceName=True)
    display = None
    view = None
    if vrayGetMayaApiVersion() >= 20220000:
        # Maya 2022 provides direct access to both vars
        display = cmds.colorManagementPrefs(q=True, displayName=True)
        view = cmds.colorManagementPrefs(q=True, viewName=True)
    else:
        # In previous versions we have to parse from the combined name
        # The format is "View (Display)", e.g. "sRGB (ACES)"
        viewdisplay = cmds.colorManagementPrefs(q=True, viewTransformName=True)
        vdmatch = re.match(r'(.*) \((.*)\)', viewdisplay)
        if vdmatch:
            view = vdmatch.group(1)
            display = vdmatch.group(2)
    return (rendering, display, view)




def vraySetMayaColorManagement(rendering, display, view):
    '''
    Set the display, view transform settings
    '''
    cmds.colorManagementPrefs(edit=True, renderingSpaceName=rendering)
    if vrayGetMayaApiVersion() >= 20220000:
        # Maya 2022 provides direct access to both vars
        display = cmds.colorManagementPrefs(edit=True, displayName=display)
        view = cmds.colorManagementPrefs(edit=True, viewName=view)
    else:
        # In previous versions we have to use the combined name
        # The format is "View (Display)", e.g. "sRGB (ACES)"
        viewdisplay = '{} ({})'.format(view, display)
        cmds.colorManagementPrefs(edit=True, viewTransformName=viewdisplay)
```

openpype/hosts/maya/api/lib.py
tokejepsen: Nice find!

Although what do you intent with using `path = os.path.expandvars(path)`?
Seems like we should need some logic to convert `<>` to OS native environment string variables.

tokejepsen:
> @tokejepsen I saw this in the V-Ray code base and I suspect it's something we'd also need to watch out for if we are to support Maya older than 2022.
> 
> ```python
> def vrayGetMayaColorManagement():
>     '''
>     Returns a triple (rendering, display, view) for the active rendering colorspace to view transform
>     '''
>     rendering = cmds.colorManagementPrefs(q=True, renderingSpaceName=True)
>     display = None
>     view = None
>     if vrayGetMayaApiVersion() >= 20220000:
>         # Maya 2022 provides direct access to both vars
>         display = cmds.colorManagementPrefs(q=True, displayName=True)
>         view = cmds.colorManagementPrefs(q=True, viewName=True)
>     else:
>         # In previous versions we have to parse from the combined name
>         # The format is "View (Display)", e.g. "sRGB (ACES)"
>         viewdisplay = cmds.colorManagementPrefs(q=True, viewTransformName=True)
>         vdmatch = re.match(r'(.*) \((.*)\)', viewdisplay)
>         if vdmatch:
>             view = vdmatch.group(1)
>             display = vdmatch.group(2)
>     return (rendering, display, view)
> 
> 
> 
> 
> def vraySetMayaColorManagement(rendering, display, view):
>     '''
>     Set the display, view transform settings
>     '''
>     cmds.colorManagementPrefs(edit=True, renderingSpaceName=rendering)
>     if vrayGetMayaApiVersion() >= 20220000:
>         # Maya 2022 provides direct access to both vars
>         display = cmds.colorManagementPrefs(edit=True, displayName=display)
>         view = cmds.colorManagementPrefs(edit=True, viewName=view)
>     else:
>         # In previous versions we have to use the combined name
>         # The format is "View (Display)", e.g. "sRGB (ACES)"
>         viewdisplay = '{} ({})'.format(view, display)
>         cmds.colorManagementPrefs(edit=True, viewTransformName=viewdisplay)
> ```
> 
> Basically View + Display settings are combined in older versions pre-Maya 2022.

Oh dear... Maya color management is a mess. Once we have something more solid in places we'll test in older versions of Maya to account for this.

openpype/hosts/maya/api/lib.py
BigRoy: > Seems like we should need some logic to convert <> to OS native environment string variables.

For as far as I can see the `<MAYA_RESOURCES>` is the only token supported in the ocio config path, e.g. `<MAYA_LOCATION>` (which is in my `os.environ`) does not resolve - as such, these tokens aren't converted to OS environment variables to begin with. It seems `<MAYA_RESOURCES>` is just special - which can also be see by the fact that it's the only path retrievable from MGlobal in the API like that.

The expand vars was just to potentially replace any environment variables a user might have put in the path to resolve the actual full path. Not sure if needed - could also start without that in. It would've been useful if e.g. you'd start subprocesses with only a part of the environment variables needing to resolve the OCIO config path.

tokejepsen: As a new user with the default `<MAYA_RESOURCES>` in the path, I would expect to be able to use `<>` for environment variables:

```python
import maya.api.OpenMaya as om
import os


path = "<MAYA_RESOURCES>/hello/<SOME_OTHER_VAR>"
os.environ["SOME_OTHER_VAR"] = "buddy"


# The OCIO config supports a custom <MAYA_RESOURCES> token
maya_resources_token = "<MAYA_RESOURCES>"
maya_resources_path = om.MGlobal.getAbsolutePathToResources()
path = path.replace(maya_resources_token, maya_resources_path)

# Should we expand any remaining vars?
path = os.path.expandvars(path)
print(os.environ["SOME_OTHER_VAR"])
print(path)
```
But maybe we should just fix `<MAYA_RESOURCES>` for now and see what pitfalls users get into?

tokejepsen:
Shit, sorry. Made a mess of this PR.

openpype/hosts/maya/api/lib.py
BigRoy: > As a new user with the default <MAYA_RESOURCES> in the path, I would expect to be able to use <> for environment variables.

You might think that but it's not how Maya works - so I wouldn't try to implement that. If the path would've contained `<SOME_OTHER_VAR>` it wouldn't work out of the box in maya.

tokejepsen:
@jakubjezek001 why is this closed?

jakubjezek001:
Fixed the target branch and reopening the PR - it was closed before due the removal of original target branch `release/3.15.x`.

openpype/hosts/maya/api/lib_renderproducts.py
```diff
@@ -625,12 +634,28 @@ def _get_aov_render_products(self, aov, cameras=None):
...
 
+    def _get_colorspace(self, attribute):
+        """Resolve colorspace from Arnold settings."""
+
+        def _view_transform():
+            preferences = lib.get_color_management_preferences()
+            return preferences["view_transform"]
+
+        resolved_values = {
+            "Raw": lambda: "Raw",
```
BigRoy, line 652: As far as I know `Raw` is not a valid colorspace. It might be for the default config but there's no specification requirement that a `Raw` key exists. I believe when an image is used 'raw' to Arnold what it means is that it's actually in Arnold rendering space. That is the `cmds.colorManagementPrefs(query=True, renderingSpaceName=True)`.

openpype/lib/transcoding.py
```diff
@@ -1037,3 +1037,56 @@ def convert_ffprobe_fps_to_float(value):
...
+    # Prepare subprocess arguments
+
+    oiio_cmd.extend([
+        input_arg, input_path,
+    ])
+
+    # Add last argument - path to output
+    oiio_cmd.extend([
+        "-o", out_filepath
+    ])
```
BigRoy, line 1089:
```suggestion
        # Input arguments
        "-i", input_path,

        # Output arguments
        "-o", out_filepath
    ]
```

Wouldn't this be just as readable if not more readable?

openpype/lib/transcoding.py
```diff
@@ -1037,3 +1037,56 @@ def convert_ffprobe_fps_to_float(value):
...
+    """Convert source files from one color space to another.
+
+    Filenames of input files are kept so make sure that output directory
+    is not the same directory as input files have.
+    - This way it can handle gaps and can keep input filenames without handling
+        frame template
+
+    Args:
+        input_path (str): Paths that should be converted. It is expected that
+            contains single file or image sequence of samy type.
```
BigRoy, line 1059: Paths or path? Is it a filepath? a folder? What happens if there are many different files in that folder?

It also feels a bit odd that this is called `input_path` but the output path is called `out_filepath`

openpype/lib/transcoding.py
```diff
@@ -1037,3 +1037,56 @@ def convert_ffprobe_fps_to_float(value):
...
+        # Don't add any additional attributes
+        "--nosoftwareattrib",
+        "--colorconfig", config_path,
+        "--colorconvert", source_colorspace, target_colorspace
+    ]
+    # Prepare subprocess arguments
+
+    oiio_cmd.extend([
+        input_arg, input_path,
+    ])
```
BigRoy, line 1084: Are we sure the colorconvert flag should not be in-between the input argument flag and output flag?

See [oiiotool command order](https://openimageio.readthedocs.io/en/latest/oiiotool.html#argument-order-matters)

The [color convert examples around here](https://openimageio.readthedocs.io/en/latest/oiiotool.html#dealing-with-multi-subimage-multi-part-files) also seem to put the input image before the colorconvert flag.

openpype/lib/transcoding.py
tokejepsen: Not part of this PR. Comes from; https://github.com/ynput/OpenPype/pull/4291

openpype/lib/transcoding.py
tokejepsen: Not part of this PR. Comes from; https://github.com/ynput/OpenPype/pull/4291

openpype/hosts/maya/api/lib_renderproducts.py
tokejepsen: Good point!

openpype/hosts/maya/api/lib.py
```diff
@@ -3402,13 +3402,22 @@ def get_color_management_preferences():
         )
     }
 
+    # Split view and display from view_transform. view_transform comes in
+    # format of "{view} ({display})".
+    display = data["view_transform"].split("(")[-1].replace(")", "")
+    data.update({
+        "display": display,
+        "view": data["view_transform"].replace("({})".format(display), "")[:-1]
+    })
```
BigRoy, line 3411: Wasn't this splitting only for need for older versions of Maya?

Also, what if a display or view has a `(` in it? I didn't see any for displays or views in Aces 1.0.3 but I did see them in names for Color spaces in that config. I think the logic is a bit to basic and we'd at least need to go regex.

```python
regex =re.compile(r"^(.+) \(.+\)$")
match = regex.match(combined_display_view)
display = match.group(1)
view = match.group(2)
```
Or better:
```python
regex =re.compile(r"^(?P<view>.+) \(?P<display>.+\)$")
match = regex.match(combined_display_view)
display = match.group("view")
view = match.group("display)

# Or:
print(match.groupdict())
# {"view": "My View", "display": "My Display"}
```

This works on these examples:
```
My View (My Display)
My View (Sexy) (My Display)
My View (Hello) (My Display-(Hello))
```
But fails as soon as the display contains a ` (Sub)` branch with a space before the `(`, like:
```
My - hello207 (-view) (my complex display (with brackets with space before it))
```
Which would parse it as:
```
view: My - hello207 (-view) (my complex display
display: (with brackets with space before it))
```

But at least it's an improvement? :)

openpype/settings/entities/schemas/projects_schema/schemas/schema_global_publish.json
```diff
@@ -267,13 +267,23 @@
...
+                                            "key": "display",
+                                            "label": "Display",
                                             "type": "text"
                                         },
                                         {
-                                            "key": "output_colorspace",
-                                            "label": "Output colorspace",
+                                            "key": "view",
+                                            "label": "View",
                                             "type": "text"
```
BigRoy, line 287: Was the name/definition wrong originally? I don't believe **View** is equivalent to what could be an **Output Colorspace**?

From the [OCIO Glossary](https://opencolorio.readthedocs.io/en/latest/concepts/overview/overview.html#glossary):

- **Transform** - a function that alters RGB(A) data (e.g transform an image from scene linear to sRGB)
- **Reference space** - a space that connects colorspaces
- **Colorspace** - a meaningful space that can be transferred to and from the reference space
- **Display** - a virtual or physical display device (e.g an sRGB display device)
- **View** - a meaningful view of the reference space on a Display (e.g a film emulation view on an sRGB display device)
- **Role** - abstract colorspace naming (e.g specify the “lnh” colorspace as the scene_linear role, or the color-picker UI uses color_picking role)
- **Look** - a color transform which applies a creative look (for example a per-shot neutral grade to remove color-casts from a sequence of film scans, or a DI look)

openpype/hosts/maya/api/lib.py
tokejepsen: Nice, thanks. Will test and use. My regex skills just werent up to the challenge :)

openpype/settings/entities/schemas/projects_schema/schemas/schema_global_publish.json
tokejepsen: Sorry, you'll have to redirect to https://github.com/ynput/OpenPype/pull/4291

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -537,7 +547,16 @@ def _create_instances_for_aov(self, instance_data, exp_files):
                 # If expectedFile are absolute, we need only filenames
                 "stagingDir": staging,
                 "fps": new_instance.get("fps"),
-                "tags": ["review"] if preview else []
+                "tags": ["review"] if preview else [],
+                "colorspaceData": {
+                    "colorspace": colorspace,
+                    "config": {
+                        "path": additional_data["colorspaceConfig"],
```
jakubjezek001, line 554: Isnt this absolute platform related path? I could track upstream coming form here https://github.com/ynput/OpenPype/pull/4340/files#diff-0f94b40dc0b348d7194c61cb01a57668eb3490d22cdbff89d37dd552287fccadR3422. Also presumably even platform absolute path could work with dirmaping, but just to make sure this works this way.

jakubjezek001:
Are we planning to update documentation too @tokejepsen ? This seems to be the feature which would need to be explained properly.

openpype/modules/deadline/plugins/publish/submit_publish_job.py
tokejepsen: Yeah, the path is platform specific. What are you suggesting to do about it?

tokejepsen:
> Are we planning to update documentation too @tokejepsen ? This seems to be the feature which would need to be explained properly.

Not sure it'll need a lot of documentation. Documentation should come from the transcoding plugin and its settings, plus the color management documentation.
This PR just focuses on making Maya work with the color management, so not sure there is any documentation to write?

LiborBatek:
Please put some additional info for testing steps. Thx

LiborBatek (approved):
Following the testing steps all works ok.

The review movie colors matches those in maya arnold render view (ACEScg) so the color transform works ok.

image

openpype/modules/deadline/plugins/publish/submit_publish_job.py
antirotor: actually I am not sure about this. If the path is platform specific and the job is published from maya on windows, but the node will create review for it on linux, dirmapping won't work, am I right?

tokejepsen: Not sure how dirmapping works. Got any suggestions to make it work?

antirotor: on some places we are storing path "rootless", like "`{root[work]}/some/path/foo/bar`" and those paths are resolved on destination platform at the time of use.

tokejepsen: We can try an do a template, but the config can also be one of Maya's native configs in which case the template will just be the absolute path.

antirotor (approved):
Looks great. Just one question - if we are adding colorspace information to instances (per AOV in Maya), how do we handle cases where color information is not needed or is not even desired. Like data type AOVs - point positions, normals etc. we don't want any conversions on them, yet if I am not mistaken, they'll inherit colorspace from the context? We don't produce reviewable from them, but thumbnails yes - if the file has data in rgba channels.

tokejepsen:
> Looks great. Just one question - if we are adding colorspace information to instances (per AOV in Maya), how do we handle cases where color information is not needed or is not even desired. Like data type AOVs - point positions, normals etc. we don't want any conversions on them, yet if I am not mistaken, they'll inherit colorspace from the context? We don't produce reviewable from them, but thumbnails yes - if the file has data in rgba channels.

Good point. I'll test a workflow for this, but think we can avoid conversion with a targeted profile.

tokejepsen:
> Looks great. Just one question - if we are adding colorspace information to instances (per AOV in Maya), how do we handle cases where color information is not needed or is not even desired. Like data type AOVs - point positions, normals etc. we don't want any conversions on them, yet if I am not mistaken, they'll inherit colorspace from the context? We don't produce reviewable from them, but thumbnails yes - if the file has data in rgba channels.

With targeted profiles we can avoid transcoding of the render passes. Something like this;

Capture

mkolar:
@tokejepsen please resolve the conflict and we can merge

tokejepsen:
> @tokejepsen please resolve the conflict and we can merge

Shouldnt merge until #4291 is merged.

tokejepsen:
@mkolar ready to merge.

PR #4343, opened 2023-01-18
Feature: Keep synced hero representations up-to-date.

### LINKED ISSUE #4331: Keep hero version up to date on remote sites
**It could be considered as bug, or at least as a missing feature for a reliable hero version system.**

Once a hero version has been downloaded on a remote site, if some other site publishes a new version and a new hero version is created, it is not automatically updated on sites which have downloaded the hero version earlier.

When a hero version is downloaded on a site, it must be kept up-to-date by the sync server.

You can re-download it by hand using the Loader but it becomes very tedious when you have to be careful about many of them.

#### Issue
Everything happens in `openpype/plugins/publish/integrate_hero_version.py`.

When new hero representation is created, if the old one is not archived but replaced, the new one starts with empty sites and has lost the synced sites information.

#### Proposal
Remove `old_repres_to_replace` system and make all repres to be processed like `old_repres_to_delete`. This way all hero representations will be archived and the sites assigned to them could be applied to the newly created hero representations.

Therefore `old_repres_to_delete`, could be renamed as `old_repres_to_archive`.

OR

We could apply the `sites` information, with removing the `created_at` field, from the old_repre to the new one.

[cuID:[OP-4759](https://app.clickup.com/t/863fw6fc8)]

Tilix4:
## Brief description
Keep previously synchronized sites up-to-date by comparing old and new sites and adding old sites if missing in new ones.

Fix #4331

## Testing notes:
1. Publish a version from a site A
2. Using the loader, download the hero version
3. Publish the same subset from site B
4. It's being synchronized in site A

### DIFF
--- a/openpype/plugins/publish/integrate_hero_version.py
+++ b/openpype/plugins/publish/integrate_hero_version.py
@@ -386,6 +386,25 @@ class IntegrateHeroVersion(pyblish.api.InstancePlugin):
                     repre["_id"] = old_repre["_id"]
                     update_data = prepare_representation_update_data(
                         old_repre, repre)
+
+                    # Keep previously synchronized sites up-to-date
+                    # by comparing old and new sites and adding old sites
+                    # if missing in new ones
+                    old_repre_files_sites = [
+                        f.get("sites", []) for f in old_repre.get("files", [])
+                    ]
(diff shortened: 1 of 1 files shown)


kalisp:
Thanks for contribution :).

PR #4349, opened 2023-01-18
General: Simplified addon distribution

iLLiCiTiT:
## Brief description
Simplified ayon distribution by using `ayon_api` module for most of server calls.

## Description
Use `ayon_api` to download files from server and to validated versions. Initialization of `AddonInfo` is not crashing if there is source with unknown type -> it causes crash of addon even if the source may not be important. Unknown source are stored to addon info. Initialization does not crash if there is none dependency package available on server. Addon code can expect type `"server"` to download package.

Renamed function `check_addons` -> `make_sure_addons_are_updated` and `check_venv` to `make_sure_venv_is_updated`. There will be functions in future to check addons updates but they should not download anything. Simplified argument of both functions.

Moved helper functions from `start.py` to openpype common as they'll be needed elsewhere. Argument `data` is required for downloaders but they don't contain `"headers"` or `"endpoint"` but what type of download file it is. Added autofix of `http` type to be converted to `server`. Fixed unzipping which was unzipping to wrong directory.

## Testing notes:
1. Everything should work as did before


jakubjezek001:
Task linked: [OP-3739 Connect OP tray to v4 server](https://app.clickup.com/t/6658547/OP-3739)

common/openpype_common/distribution/addon_distribution.py
```diff
@@ -99,23 +142,29 @@ class HTTPAddonDownloader(AddonDownloader):
     CHUNK_SIZE = 100000
 
     @classmethod
-    def download(cls, source, destination_dir, data=None):
+    def download(cls, source, destination_dir, data):
         source_url = source["url"]
```
Minkiu, line 146: Do we know if there could be an instance where `source ` is an empty dict?

common/openpype_common/distribution/addon_info.py
```diff
@@ -111,7 +124,9 @@ def from_dict(cls, package):
                 source_addon = ServerResourceSource(
                     type=source["type"], filename=source["filename"])
```
Minkiu, line 125: This could probably be:
```
source_addon = ServerResourceSource(
                    type=source_type, filename=source["filename"])
```

common/openpype_common/distribution/file_handler.py
```diff
@@ -111,13 +111,15 @@ def download_url(
...
-                url = url.replace('https:', 'http:')
-                print('Failed download. Trying https -> http instead.'
-                      ' Downloading ' + url + ' to ' + fpath)
+            if url[:5] == "https":
+                url = url.replace("https:", "http:")
+                print((
+                    "Failed download. Trying https -> http instead."
+                    f" Downloading {url} to {fpath}"
+                ))
                 RemoteFileHandler._urlretrieve(url, fpath,
```
Minkiu, line 123: Shouldn't we catch a possible Exception on the second try with `https` ?

Minkiu (approved):

common/openpype_common/distribution/file_handler.py
iLLiCiTiT: No, it's catched where the `download` is called (it would not know if and how it crashed).

common/openpype_common/distribution/addon_distribution.py
iLLiCiTiT: There can be, but in that case it is invalid source and would not get to `download` phase.

kalisp (approved):

PR #4352, opened 2023-01-19
Maya: Clean up redundant lib code.

tokejepsen:
## Brief description
Removing Maya library code that is redundant.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/submit_maya_muster.py
+++ b/openpype/hosts/maya/plugins/publish/submit_maya_muster.py
@@ -52,6 +52,11 @@ def _get_script():
     return module_path
 
 
+def get_renderer(layer):
+    with lib.renderlayer(layer):
+        return cmds.getAttr("defaultRenderGlobals.currentRenderer")
+
+
 def get_renderer_variables(renderlayer=None):
     """Retrieve the extension which has been set in the VRay settings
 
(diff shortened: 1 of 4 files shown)


moonyuet (approved):
The clean up looks good for me.

openpype/hosts/maya/plugins/publish/submit_maya_muster.py
```diff
@@ -52,6 +52,11 @@ def _get_script():
     return module_path
 
 
+def get_renderer(layer):
```
Minkiu, line 55: It might not be used by any other plugin now, but I'd say this is fine staying in `lib`.

tokejepsen: Should be expose in the api init then we expect to be used externally?

BigRoy: I would just keep it around in lib for if another plug-in might need it. No need to expose it further than this I think.

tokejepsen: I'm more thinking that exposing it in api ini would indicate for future reference that it needs to be kept for external use?

antirotor:
what about leaving the `get_render()` in lib but also modify other places querying `defaultRenderGlobals.currentRenderer` to use that? By quick search, there are just 8 places around so it would be unified...

openpype/hosts/maya/plugins/publish/submit_maya_muster.py
tokejepsen: ok, kept it in

tokejepsen:
> what about leaving the `get_render()` in lib but also modify other places querying `defaultRenderGlobals.currentRenderer` to use that? By quick search, there are just 8 places around so it would be unified...

I've just left it in the lib for now. The other instance of querying `defaultRenderGlobals.currentRenderer` were not using using renderlayers so think that is a bigger rewrite.

tokejepsen:
This can be merged now. @antirotor wanna do the honours?

antirotor (approved):

PR #4356, opened 2023-01-19
Unreal: Don't use mongo queries in 'ExistingLayoutLoader'

iLLiCiTiT:
## Brief description
Unreal load plugin `ExistingLayoutLoader` used direct mongo calls using `legacy_io.find_one` and imported from `openpype.api`. Both should not be used anymore.

## Description
Use `get_representations` function to get representations instead of direct queries. Replaced querying of representations one by one into 2 chunks. Added class attribute `delete_unmatched_assets` which stores value of `"delete_unmatched_assets"` from unreal settings (overriden classmethod `apply_settings`) -> removed import from `openpype.api`.

## Testing notes:
1. Loader should work as did before this PR (maybe faster)

### DIFF
--- a/openpype/hosts/unreal/plugins/load/load_layout_existing.py
+++ b/openpype/hosts/unreal/plugins/load/load_layout_existing.py
@@ -282,7 +282,7 @@ class ExistingLayoutLoader(plugin.Loader):
         actors_matched = []
 
         for (repr_data, lasset) in layout_data:
-             # For every actor in the scene, check if it has a representation in
+            # For every actor in the scene, check if it has a representation in
             # those we got from the JSON. If so, create a container for it.
             # Otherwise, remove it from the scene.
             found = False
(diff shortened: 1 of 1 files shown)


openpype/hosts/unreal/plugins/load/load_layout_existing.py
```diff
@@ -221,41 +222,66 @@ def _load_asset(self, representation, version, instance_name, family):
...
-                        "_id": ObjectId(element.get('representation'))
-                    }),
-                    element
-                ))
-
-        containers = []
-        actors_matched = []
+            repre_id = element.get('representation')
+            if repre_id:
+                repre_ids.append(repre_id)
```
simonebarbieri, line 253:
```suggestion
                repre_ids.add(repre_id)
```

Apart from this, it seems to be working fine!

simonebarbieri (changes requested):

simonebarbieri (approved):

PR #4359, opened 2023-01-20
Fix - addSite loader handles hero version

kalisp:
## Brief description
If adding site to representation presence of hero version is checked, if found hero version is marked to be donwloaded too.

Replacing https://github.com/ynput/OpenPype/pull/4191

## Testing notes:
1. locate subset which has some hero version in Loader
2.use latest version of said subset and `Download` representation (in right bottom)
3. check that both latest and hero version representation were downloaded

### DIFF
--- a/openpype/plugins/load/add_site.py
+++ b/openpype/plugins/load/add_site.py
@@ -2,12 +2,6 @@ from openpype.client import get_linked_representation_id
 from openpype.modules import ModulesManager
 from openpype.pipeline import load
 from openpype.modules.sync_server.utils import SiteAlreadyPresentError
-from openpype.client.entities import (
-    get_hero_version_by_subset_id,
-    get_representation_by_id,
-    get_version_by_id,
-    get_representation_by_name
-)
 
 
 class AddSyncSite(load.LoaderPlugin):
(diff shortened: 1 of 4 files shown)


Tilix4:
Works fine, thanks!

openpype/plugins/load/add_site.py
```diff
@@ -64,3 +78,32 @@ def load(self, context, name=None, namespace=None, data=None):
     def filepath_from_context(self, context):
         """No real file loading"""
         return ""
+
+    def _add_hero_representation_ids(self, project_name, repre_id):
+        """Find hero version if exists for repre_id.
+
```
Tilix4, line 84: Missing params in docstring

Tilix4 (approved):
Works well! Thanks!

Tilix4 (changes requested):
My bad, it doesn't work when subset are added to queue as input dependencies!

kalisp:
Updated handling of hero version.

Tilix4:
Works better for dependencies! but it doesn't fetch dependencies which also have their own dependencies. In the screenshot provided, the two missing ones are not atomic dependencies when the fetched ones have no inputs.
image

Tilix4 (approved):
LGTM now! Thanks

Tilix4 (approved):
Tested :ok: ! LGTM

openpype/plugins/load/add_site.py
```diff
@@ -64,3 +84,35 @@ def load(self, context, name=None, namespace=None, data=None):
     def filepath_from_context(self, context):
         """No real file loading"""
         return ""
+
+    def _add_hero_representation_ids(self, project_name, repre_id):
```
iLLiCiTiT, line 88: Unused method

iLLiCiTiT (approved):

PR #4360, opened 2023-01-20
General: Loader plugin define its own compatibility

iLLiCiTiT:
## Brief description
This is "idea", how to give loaders more power to decide if can be shown in UI by adding `is_compatible` method on it.

## Description
Add `is_compatible` class method to loader plugin which by default does what `is_compatible` function did but loader can override the method and be more precise in the decision.

## Additional information
There are also SubsetLoaders which didn't use `is_compatible` at all but probably should have that option too. In future we could add objects that would represent items in loader menu so loader does not have to calculate that on it's own. That is missing few more things (like unique identifier of loader plugins, etc.) -> doable, but more complicated.

I didn't test it...

## Testing notes:
1. Loader plugins should be filtered as they were before in Loader and LibraryLoader

### DIFF
--- a/openpype/pipeline/load/plugins.py
+++ b/openpype/pipeline/load/plugins.py
@@ -2,7 +2,10 @@ import os
 import logging
 
 from openpype.settings import get_system_settings, get_project_settings
-from openpype.pipeline import legacy_io
+from openpype.pipeline import (
+    schema,
+    legacy_io,
+)
 from openpype.pipeline.plugin_discover import (
     discover,
     register_plugin,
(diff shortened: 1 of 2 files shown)


openpype/pipeline/load/plugins.py
```diff
@@ -79,6 +82,36 @@ def apply_settings(cls, project_settings, system_settings):
...
+        has_family = (
+            "*" in cls.families or any(
+                family in cls.families for family in families
+            )
+        )
+        representations = cls.get_representations()
+        has_representation = (
+            "*" in representations
+            or representation["name"] in representations
+        )
```
BigRoy, line 112: Since we're tweaking anyway - I think we could optimize this by doing the lookup against a set.
This should make quite a difference for plugins with many valid representations but the context not being in it or for an instance with many families against a plugin with many valid families this should also speed things up.

```suggestion
        plugin_families = frozenset(cls.families)
        has_family = (
            "*" in plugin_families or any(
                family in plugin_families for family in families
            )
        )
        if not has_family:
            return False
        
        representation = context["representation"]
        representations = frozenset(cls.get_representations())
        has_representation = (
            "*" in representations
            or representation["name"] in representations
        )
        return has_representation
```
This also opts out from the representation check if the family already didn't pass.

BigRoy:
I like it! :) I haven't tested the code but the concept and logic appear sound.

antirotor:
Yep, I like it too. Need to test it before merging but the code looks good. First step for the better loaders, yay!

kalisp (approved):
Tested overriding `is_compatible_loader` on `AddSyncSite`, works.

PR #4362, opened 2023-01-24
Nuke: update rendered frames in latest version

kalisp:
## Brief description
Introduced new field to insert frame(s) to rerender only.

## Description
Rendering is expensive, sometimes it is helpful only to re-render changed frames and reuse existing.
Artists can in Publisher fill which frame(s) should be re-rendered.
2023-01-24 17_10_45-OpenPype publisher

If there is already published version of currently publishing subset, all representation files are collected (currently for `render` family only) and then when Nuke is rendering (locally only for now), old published files are copied into into temporary render folder where will be rewritten only by frames explicitly set in new field.

That way review/burnin process could also reuse old files and recreate reviews/burnins.

New version is produced during this process!

## Additional info

Currently only implemented for Nuke's local render as a proof of concept. Farm rendering will be implemented afterwards.

Only other host where this workflow can work is for now AfterEffects as they have Publisher implemented.

It might be useful to create abstract extractor class and use it in host extractor as multiple inheritance to highlight that this extractor implements this functionality (kinda trait - expect some specific fields, does something)...

Contains fix for AE found during testing, didn't want to create new PR for this (expected to implement AE first, decided for Nuke instead after commits were done.)

## Testing notes:
1. publish sequence of frames in Nuke
2. change specific frame in Nuke workfile (to see change)
3. fill frame of changed frame into text field
4. publish again - check published frame, review etc.


jakubjezek001:
Task linked: [OP-4617 Global: Update rendered frames in latest version](https://app.clickup.com/t/6658547/OP-4617)

kalisp:
Added new toggle to overwrite files of last version.

Fixed version overwriting in `integrate.py`!!!

(Currently it is not possible to add `UISeparatorDef` to plugin attributes to separate.)

jakubjezek001 (approved):
Nuke and integration into OP works well, but Ftrack reviewable component is not updating.

openpype/hosts/nuke/plugins/publish/extract_render_local.py
```diff
@@ -133,3 +151,60 @@ def process(self, instance):
...
+            # instance.context.data["cleanupFullPaths"].append(out_path)
+
+    def _get_frames_to_render(self, frames_to_fix):
+        """Return list of frame range tuples to render
+
+        Args:
+            frames_to_fix (str): specific or range of frames to be rerendered
+             (1005, 1009-1010)
+            first_frame (int): original first frame
+            last_frame (int)
```
BigRoy, line 191: Arguments don't exist.
```suggestion
```

BigRoy:
Why are we writing it into an existing version instead of generating a new version for which we ONLY render the 'to fix' frames and copy the other frames?

I feel like writing into an existing version will only start complicating thing more and feels kind of counter-intuitive to the whole versioning aspect we'd like to follow, but if we really want to write into an existing publish:

- What if the files are in use?
- Does site sync correctly identify that it needs to resync the files?
- Do hero versions get correctly updated too? (and also for hero version site sync?)

jakubjezek001:
> Why are we writing it into an existing version instead of generating a new version for which we ONLY render the 'to fix' frames and copy the other frames?
> 
> To me personally it'd make much more sense to increment a version number if we're changing it.
> 
> I feel like writing into an existing version will only start complicating thing more and feels kind of counter-intuitive to the whole versioning aspect we'd like to follow, but if we really want to write into an existing publish:
> 
> * What if the files are in use?
> * Does site sync correctly identify that it needs to resync the files?
> * Do hero versions get correctly updated too? (and also for hero version site sync?)

This was clients requested feature. Not something you want to use everyday. An issue they were facing were huge resolutions of rendered files - so they didn't want to accumulate frames which were not changed.

openpype/hosts/nuke/plugins/publish/extract_render_local.py
```diff
@@ -133,3 +151,60 @@ def process(self, instance):
...
+        Returns:
+            (list): [(1005, 1005), (1009-1010)]
+        """
+        frames_to_render = []
+
+        for frame_range in frames_to_fix.split(","):
+            if isinstance(frame_range, int) or frame_range.isdigit():
```
BigRoy, line 198: Should this do `frame_range = frame_range.strip()` too? Just so that if the user ever typed a space it'd be ignored?

PR #4363, opened 2023-01-24
AYON: Distribution with progress

iLLiCiTiT:
## Brief description
Made distribution object based so it can be used in UIs and progress can be tracked.

## Description
The object of `AyonDistribution` handle whole distribution logic. Create distribution items per dependency package and each addon to check if are available or should be distributed. The distribution can be process then in thread and it is possible to check the distribution progress overtime (for possible future UI). Renamed classes to define sources to match their new usage (not just addon specific).

Distributed items are stored to metadata file so we can keep track of what, when and which source was used to distribute the files. The metadata also contains file hash for future possible check of file changes of the same version (e.g. `dev` version).

## Testing notes:
1. Distribution should work as did before


jakubjezek001:
Task linked: [OP-3739 Connect OP tray to v4 server](https://app.clickup.com/t/6658547/OP-3739)

iLLiCiTiT:
> Could you please fix tests also (changed logic with missing update_addon_state).

Sure. I'll also update source info (server was enhanced by `path` attribute).

common/ayon_common/distribution/addon_distribution.py
```diff
@@ -312,185 +361,694 @@ def _try_convert_to_server_source(addon, source):
...
 
 
-def update_addon_state(
-    addon_infos,
-    destination_folder,
-    factory,
-    log=None
-):
-    """Loops through all 'addon_infos', compares local version, unzips.
+class DistributeTransferProgress:
```
kalisp, line 369: I don't understand what this is, shouldn't this inherit from `ayon_api.TransferProgress`?

I don't particularly like that 'Progress', I'll always imagine something like 0-100, not if if started, failed. I would prefer something like `DistributionItemStatus`

common/ayon_common/distribution/addon_distribution.py
```diff
@@ -312,185 +361,694 @@ def _try_convert_to_server_source(addon, source):
...
+            package.checksum,
+            self._dist_factory,
+            package.sources,
+            downloader_data,
+            package.name,
+            self.log,
+        )
+
+    def get_addons_progress(self):
+        if self._addons_progress is None:
```
kalisp, line 863: There is a lot of docstrings missing, that this method should be doing? I see some discrepancy between `progress` and `state` (`_prepare_current_addons_state`). Why was 'progress' is preferable from 'state', 'status'?
What it is supposed to be returning? There is something called 'DistributeTransferProgress` so I would expect that `get_addons_progress` returns these, but it is is actually returning dictionary of something called `DistributionItem`.

kalisp: Isn't this something like a queue of addons which should be downloaded?

common/ayon_common/distribution/addon_distribution.py
iLLiCiTiT: This object should help to track the progress of single distribution item source (progress not state). The progress consist of 3 parts: "transfer", "hashcheck", "unzip". All of these steps may take some time but may not have ability to track exact 0-100. Since it was created to have ability to use it in (future) Distribution UI it does not matter, important is knowledge in which part of progress we are and 0-100 is a bonus.

common/ayon_common/distribution/addon_distribution.py
iLLiCiTiT: Made this in rush and didn't think about it, will look into that.

kalisp (approved):
Test is working.

Tested connecting to ayon.dev, missing addon downloaded.

common/ayon_common/distribution/addon_info.py
```diff
@@ -114,17 +131,26 @@ class DependencyItem(object):
...
-                source_addon = ServerResourceSource(
-                    type=source_type, filename=source["filename"])
+                source_addon = ServerSourceInfo(
+                    type=source_type,
+                    filename=source.get("filename"),
+                    path=source.get("path")
+                )
             else:
                 print(f"Unknown source {source_type}")
                 unknown_sources.append(source)
```
BigRoy, line 156: Seeing this being duplicated logic from above I think we can have a helper function to `convert_source(source)` or `convert_sources(sources)` if we want to keep track of the unknown ones directly too.

common/ayon_common/distribution/tests/test_addon_distributtion.py
```diff
@@ -152,16 +153,31 @@ def test_update_addon_state(printer, sample_addon_info,
...
-        "Update should failed because of wrong hash"
-
+    distribution = AyonDistribution(
+        temp_folder, temp_folder, addon_downloader, [addon_info], None
+    )
+    distribution.distribute()
+    dist_items = distribution.get_addons_dist_items()
+    slack_state = dist_items["openpype_slack_1.0.0"].state
+    assert slack_state == UpdateState.UPDATE_FAILED, (
+        "Update should failed because of wrong hash")
```
BigRoy, line 165:
```suggestion
        "Update should fail because of wrong hash")
```

common/ayon_common/distribution/addon_distribution.py
```diff
@@ -44,6 +47,7 @@ def get_local_dir(*subdirs):
 
     local_dir = os.path.join(
         appdirs.user_data_dir("openpype", "pypeclub"),
+        # appdirs.user_data_dir("ayon", "ynput"),
```
BigRoy, line 50: This line shouldn't be here? Or maybe add a `todo` to it?

common/ayon_common/distribution/addon_distribution.py
```diff
@@ -112,7 +116,7 @@ def get_downloader(self, downloader_type):
 
     @classmethod
     @abstractmethod
-    def download(cls, source, destination_dir, data):
+    def download(cls, source, destination_dir, data, transfer_progress):
```
BigRoy, line 119: `transfer_progress` missing in docstring. Also unclear what it does.

common/ayon_common/distribution/addon_distribution.py
BigRoy: I agree with @kalisp - even from your extra explanation I feel this makes more sense as a status than progress?

BigRoy: Also, looking at the actual data that this hold? Should this maybe be brought back to being an explicit dataclass?

common/ayon_common/distribution/addon_distribution.py
```diff
@@ -294,203 +343,710 @@ def get_dependency_package(package_name=None):
...
+        return self._addons_dist_items
 
-    package = get_dependency_package()
-    if not package:
-        log.info("Server does not contain dependency package")
-        return
+    def get_dependency_dist_item(self):
+        if self._dependency_dist_item == -1:
+            self._dependency_dist_item = self._prepare_dependency_progress()
+        return self._dependency_dist_item
```
BigRoy, line 868: Missing docstring?

common/ayon_common/distribution/addon_distribution.py
```diff
@@ -294,203 +343,710 @@ def get_dependency_package(package_name=None):
...
+
+        if not os.path.exists(filepath):
+            dirpath = os.path.dirname(filepath)
+            if not os.path.exists(dirpath):
+                os.makedirs(dirpath)
+
+            if default_value is None:
+                default_value = {}
+            with open(filepath, "w") as stream:
+                json.dump(default_value, stream)
```
BigRoy, line 898: Wait, wait... why is this writing a file in a function called `read_metadata_file`? The docstring mentions it but boy, that's a confusing method name?

Why does it have to generate the file?

common/ayon_common/distribution/addon_distribution.py
iLLiCiTiT: We were discussing it and there may be confusion in state and status. But the name is not right. It should be something like `DistributeSourceProgress` or `DistributeSourceStatus`. I have opinion that it is progress but both of you had same idea to I'm willing to loose my opinion here.

> Should this maybe be brought back to being an explicit dataclass?

I personally don't like dataclasses as it's not clear what should be read-only.

I'm merging current state as we have to start with "prove of concept tests", but we can continue with comments here, it's not 100% finished anyway and there will be more PRs.

PR #4367, opened 2023-01-25
Maya: Fix template builder bug where assets are not put in the right hierarchy

friquette:
## Brief description
When buiding scene from template, the assets loaded from the placeholders are not put in the hierarchy. Plus, the assets are loaded in double.

## Testing notes:
1. In in maya: OpenPype > Template Builder > Build Workfile from template
2. The assets are not in the right hierarchy 
image

### DIFF
--- a/openpype/pipeline/workfile/workfile_template_builder.py
+++ b/openpype/pipeline/workfile/workfile_template_builder.py
@@ -521,6 +521,11 @@ class AbstractTemplateBuilder(object):
         if not level_limit:
             level_limit = 1000
 
+        placeholder_by_scene_id = {
+            placeholder.scene_identifier: placeholder
+            for placeholder in placeholders
+        }
+
         all_processed = len(placeholders) == 0
         # Counter is checked at the ned of a loop so the loop happens at least
         #   once.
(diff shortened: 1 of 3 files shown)


openpype/pipeline/workfile/workfile_template_builder.py
```diff
@@ -521,10 +521,6 @@ def populate_scene_placeholders(self, level_limit=None):
         if not level_limit:
             level_limit = 1000
 
-        placeholder_by_scene_id = {
```
iLLiCiTiT, line 524: We should fix the issue without getting rid of the recusive functionality.

friquette: The issue for now is that the recursive functionality is importing assets multiple times. I don't really understand the point of this functionality and its functioning. Could you explain it to me briefly ?

iLLiCiTiT: Few clients are using it in a way that they publish template workfile (or nodes). Then they define main template which has placehodlers that will load them. So first iteration would load other template placeholders and next iterations would load the placeholders from the previously loaded placeholders. We can't remove it...

iLLiCiTiT: Just looked at the code and if the `scene_identifier` of placeholder does not change after process, it should not process it again.

friquette: Ok I see. I will dig into it but I think the `scene_identifier` changes even though it shouldn't. Thank you !

iLLiCiTiT:
Is this ready?

friquette:
Yes it's ready for the hierarchy problem. I will do a separate PR for the references loaded in double because it is link to a problem with the recursivity.

iLLiCiTiT (approved):

PR #4370, opened 2023-01-25
Unreal: New Publisher

simonebarbieri:
## Brief description
Implementation of the new publisher for Unreal.

## Description
The implementation of the new publisher for Unreal. This PR includes the changes for all the existing creators to be compatible with the new publisher.
The basic creator has been split in two distinct creators:
- `UnrealAssetCreator`, works with assets in the Content Browser. 
- `UnrealActorCreator` that works with actors in the scene.

## Testing notes
You will need to have the new publisher enabled in the Experimental tools, in OP Settings. You will also need to import some assets in Unreal to test the creation of the instances.

1. Open the new publisher, from `Experimental tools -> New Publisher`.
2. Create at least one instance for each family.
3. Check if the created instance are collected correctly when closing and reopening the new publisher.
4. Try to publish the instances (note: there's no extractor for the `unrealStaticMesh`, so won't publish, but shouldn't give any error).


openpype/hosts/unreal/api/plugin.py
```diff
@@ -1,7 +1,251 @@
...
+)
+from openpype.pipeline import (
+    Creator,
+    LoaderPlugin,
+    CreatorError,
+    CreatedInstance
+)
+
+
+class OpenPypeCreatorError(CreatorError):
```
iLLiCiTiT, line 29: Why to create additional error? (with `OpenPype`)

openpype/hosts/unreal/api/plugin.py
```diff
@@ -1,7 +1,251 @@
...
+    pass
+
+
+@six.add_metaclass(ABCMeta)
```
iLLiCiTiT, line 33: `Creator` already has `ABCMeta` as metaclass.

openpype/hosts/unreal/plugins/create/create_look.py
```diff
@@ -1,56 +1,55 @@
...
-    def __init__(self, *args, **kwargs):
-        super(CreateLook, self).__init__(*args, **kwargs)
-
-    def process(self):
-        name = self.data["subset"]
-
+    def create(self, subset_name, instance_data, pre_create_data):
         selection = []
-        if (self.options or {}).get("useSelection"):
+        if pre_create_data.get("use_selection"):
```
iLLiCiTiT, line 22: So there is not an option to use this creator without selection? In that case the `"use_selection"` should not be in pre-create attributes. If it's disabled it crashes so it's pointless to have that attribute there.

openpype/hosts/unreal/plugins/create/create_camera.py
```diff
@@ -1,41 +1,13 @@
...
-        factory = unreal.LevelSequenceFactoryNew()
-        tools = unreal.AssetToolsHelpers().get_asset_tools()
-        tools.create_asset(name, f"{self.root}/{name}", None, factory)
-
-        asset_name = f"{self.root}/{name}/{name}.{name}"
-
-        data["members"] = [asset_name]
-
-        instantiate(f"{self.root}", name, data, None, self.suffix)
+    icon = "camera"
```
iLLiCiTiT, line 13:
```suggestion
    icon = "fa.camera"
```

LiborBatek:
Please add Testing Steps for our review. Thx

openpype/hosts/unreal/api/plugin.py
```diff
@@ -1,7 +1,251 @@
...
+                    else:
+                        shared_data[
+                            "unreal_cached_legacy_subsets"][family].append(i)
+                    continue
+
+                creator_id = i.get("creator_identifier")
+                if creator_id not in shared_data["unreal_cached_subsets"]:
+                    shared_data["unreal_cached_subsets"][creator_id] = [i]
+                else:
+                    shared_data["unreal_cached_subsets"][creator_id].append(i)
```
iLLiCiTiT, line 80: This could be simplified to
```suggestion
            unreal_cached_subsets = collections.defaultdict(list)
            unreal_cached_legacy_subsets = collections.defaultdict(list)
            for instance in lsinst():
                creator_id = instance.get("creator_identifier")
                if creator_id:
                    unreal_cached_subsets[creator_id].append(instance)
                else:
                    family = instance.get("family")
                    unreal_cached_legacy_subsets[family].append(instance)

            shared_data["unreal_cached_subsets"] = unreal_cached_subsets
            shared_data["unreal_cached_legacy_subsets"] = (
                unreal_cached_legacy_subsets
            )
```

openpype/hosts/unreal/api/plugin.py
```diff
@@ -1,7 +1,251 @@
...
+            instance_node = created_inst.get("instance_path", "")
+
+            if not instance_node:
+                unreal.log_warning(
+                    f"Instance node not found for {created_inst}")
+
+            new_values = {
+                key: new_value
+                for key, (_old_value, new_value) in _changes.items()
+            }
```
iLLiCiTiT, line 135: Changes object has changed
```suggestion
        for created_inst, changes in update_list:
            instance_node = created_inst.get("instance_path", "")

            if not instance_node:
                unreal.log_warning(
                    f"Instance node not found for {created_inst}")

            new_values = {
                key: changes[key].new_value
                for key in changes.changed_keys()
            }
```

openpype/hosts/unreal/api/plugin.py
```diff
@@ -1,7 +1,251 @@
...
+            created_instance = CreatedInstance.from_existing(instance, self)
+            self._add_instance_to_context(created_instance)
+
+    def update_instances(self, update_list):
+        unreal.log_warning(f"Update instances: {update_list}")
+        for created_inst, _changes in update_list:
+            instance_node = created_inst.get("instance_path", "")
+
+            if not instance_node:
+                unreal.log_warning(
```
iLLiCiTiT, line 129: Shouldn't be update skipped if `instance_node` is not available?

openpype/hosts/unreal/api/plugin.py
simonebarbieri: I've followed Houdini and Nuke as base, and they both do it 👀

iLLiCiTiT: :see_no_evil::hear_no_evil::speak_no_evil:

iLLiCiTiT: It's not needed to have custom exception. Use `CreatorError` as is.

LiborBatek:
Following the testing steps I have a hard time when creating ```uasset``` instance. Really confused what and where to select the object for creating the instance ```use selection``` gives me this error:
image

Whats the supposed workflow when creating instances for publishing? Should I use ```Outliner``` for selecting or ```Content Browser```?? seems the second one which is a bit odd (it is imported from outside DCC but actually doesnt need to be present in the actual UE level/map ?!

A lot of confusion...now

LiborBatek:
...when selected the asset for instance creation from ```content browser``` seems ok and ```uasset``` instance got created

image

openpype/hosts/unreal/api/pipeline.py
```diff
@@ -60,6 +62,23 @@ def show_tools_dialog(self):
 
         show_tools_dialog()
 
+    def update_context_data(self, data, changes):
+        unreal.log_warning("update_context_data")
+        unreal.log_warning(data)
+        content_path = unreal.Paths.project_content_dir()
+        op_ctx = content_path + CONTEXT_CONTAINER
+        with open(op_ctx, "w+") as f:
```
antirotor, line 70: I am thinking that this might cause potential issues with concurrency - imagine multiple artists working on the same project, updating the context file and reading it at the same time. I bet that this will result in `IOError`.

The concurrency issue in this case really scares me and we need to think about solution.

openpype/hosts/unreal/api/pipeline.py
```diff
@@ -133,6 +152,31 @@ def ls():
         yield data
 
 
+def lsinst():
```
antirotor, line 155: I would probably put underscore there
```suggestion
def ls_inst():
```

openpype/hosts/unreal/api/plugin.py
```diff
@@ -1,7 +1,242 @@
...
+)
+from openpype.pipeline import (
+    Creator,
+    LoaderPlugin,
+    CreatorError,
+    CreatedInstance
+)
+
+
+class UnrealBaseCreator(Creator):
```
antirotor, line 28: I guess it has to be still declared as abstract, because it is missing implementation of abstract property `family`

openpype/hosts/unreal/api/plugin.py
```diff
@@ -1,7 +1,242 @@
...
+    def remove_instances(self, instances):
+        for instance in instances:
+            instance_node = instance.data.get("instance_path", "")
+            if instance_node:
+                unreal.EditorAssetLibrary.delete_asset(instance_node)
+
+            self._remove_instance_from_context(instance)
+
+
+class UnrealAssetCreator(UnrealBaseCreator):
```
antirotor, line 143: again, this must be abstract because of the missing `family`

openpype/hosts/unreal/api/plugin.py
```diff
@@ -1,7 +1,242 @@
...
+                CreatorError(f"Creator error: {er}"),
+                sys.exc_info()[2])
+
+    def get_pre_create_attr_defs(self):
+        return [
+            BoolDef("use_selection", label="Use selection", default=True)
+        ]
+
+
+class UnrealActorCreator(UnrealBaseCreator):
```
antirotor, line 186: this needs to be abstract too

openpype/hosts/unreal/plugins/create/create_render.py
```diff
@@ -1,117 +1,137 @@
...
-                    "output": f"{ms_obj.get_name()}",
-                    "frame_range": (
-                        ms_obj.get_playback_start(), ms_obj.get_playback_end())
-                }
+        sel_objects = unreal.EditorUtilityLibrary.get_selected_assets()
+        selection = [
+            a.get_path_name() for a in sel_objects
+            if a.get_class().get_name() == "LevelSequence"]
+
+        if len(selection) == 0:
```
antirotor, line 29: more pythonic

```suggestion
        if not selection:
```

openpype/hosts/unreal/plugins/create/create_render.py
```diff
@@ -1,117 +1,137 @@
...
+            raise RuntimeError("Please select at least one Level Sequence.")
```
antirotor, line 30: There is `CreatorError` for it somewhere.

openpype/hosts/unreal/plugins/create/create_render.py
```diff
@@ -1,117 +1,137 @@
...
+            selected_asset = ar.get_asset_by_object_path(sel).get_asset()
+            selected_asset_path = selected_asset.get_path_name()
+
+            # Check if the selected asset is a level sequence asset.
+            if selected_asset.get_class().get_name() != "LevelSequence":
+                unreal.log_warning(
+                    f"Skipping {selected_asset.get_name()}. It isn't a Level "
+                    "Sequence.")
+
+            # The asset name is the the third element of the path which
```
antirotor, line 44:
```suggestion
            # The asset name is the third element of the path which
```

openpype/hosts/unreal/plugins/publish/collect_instance_members.py
```diff
@@ -0,0 +1,49 @@
...
+        if not pub_instance.get_editor_property("add_external_assets"):
+            # No external assets in the instance
+            return
+
+        assets = pub_instance.get_editor_property('asset_data_external')
+
+        members = []
+
+        for asset in assets:
+            members.append(asset.get_path_name())
```
antirotor, line 45:
```suggestion
        members = [asset.get_path_name() for asset in assets]
```

antirotor (changes requested):
Some little things about the code, I need to test it more functionally.

antirotor (changes requested):
Tested it:

**Camera**
It doesn't want to publish camera. Creator will not pick the correct one.

**Look**
It is not using the look from the current selection, it will set it to default material.

Also, I would change the menu so there is no `Create` and the `Publish` will open the new publisher.

antirotor (approved):
I think it's working!

PR #4380, opened 2023-01-26
CelAction: Collector CLI is now working as it should

jakubjezek001:
## Brief description
CelAction is now publishing as expected

## Description
Collector CLI was failing as it could not correctly parse arguments send form CelAction. It should be working fine now.

### DIFF
--- a/openpype/hosts/celaction/plugins/publish/collect_celaction_cli_kwargs.py
+++ b/openpype/hosts/celaction/plugins/publish/collect_celaction_cli_kwargs.py
@@ -1,5 +1,4 @@
 import pyblish.api
-import argparse
 import sys
 from pprint import pformat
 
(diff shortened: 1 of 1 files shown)


iLLiCiTiT (approved):

BigRoy:
> Collector CLI was failing as it could not correctly parse arguments send form CelAction

What in particular did it fail on originally?
Could you share an example of the values of `list(sys.argv[1:])`?

iLLiCiTiT:
> What in particular did it fail on originally? Could you share an example of the values of list(sys.argv[1:])?

Source issue was that the path is split into multiple arguments (by spaces). We tried to use `"` but it didn't help. This caused that `argparse` crashed -> and in process using `openpype_gui` it crashed even worse because it tried to `write` to `sys.stdout` which is not available in GUI base build.

BigRoy:
Would this workaround work too?

```python
parser.add_argument("--currentFile",
                    # The path could be split into separate arguments
                    # by spaces which argument parser doesn't like
                    # so we convert it ourselves
                    nargs="+",
                    help="Pass file to Context as `currentFile`")
```
                    
Then the result of `currentFile` of arg parse might need to be joined together to one string:

```python
opts = parser.parse_args()
current_file = ' '.join(opts.currentFile)
```

Or to stick to the old code:
```python
passing_kwargs["currentFile"] = " ".join(passing_kwargs["currentFile"])
```

iLLiCiTiT:
> But - this is only to workaround the issue where the path is basically incorrectly passed to the command shell to begin with because it SHOULD have been encapsulated in quotes.

That won't fix issue that argparse is trying to write to `sys.stdout` which is `None`.

BigRoy:
> That won't fix issue that argparse is trying to write to sys.stdout which is None.

That's true - I assume that's likely happening due to it trying to message about the error. Because it'll trigger the `.error()` method on the Argument parser if it fails. Which should write to `sys.stderr` by default by the way.

BigRoy:
So I suppose that means the alternative proposed didn't work? :)

PR #4383, opened 2023-01-26
Global: Do not force instance data with frame ranges of the asset

### LINKED ISSUE #4317: Collected instances without frame range data still publish with frame ranges
**Running version**

3.14.10-nightly.4

**Describe the bug**

Any collected instances without frame data still get the asset's frame data. This **should not** be the case.

It works fine for me on this particular use case if I remove the logic [here](https://github.com/ynput/OpenPype/blob/cf0cba1dba0d14b60ca1bff0f9d9170aff88bb43/openpype/plugins/publish/collect_anatomy_instance_data.py#L117-L134).

afbeelding

**To Reproduce**

1. Publish an instance that shouldn't have start/end frame.
2. It will show asset's start and end frame in the Loader instead.

**Expected behavior**

It should be up the instances themselves to define what their frame ranges should be - it shouldn't be globally forced to the asset's time.

A non-time based instance should not provide time data.

**Additional context**

Issue also came up on [discord here](https://discord.com/channels/517362899170230292/517382145552154634/1063506195656212480)

[cuID:[OP-4746](https://app.clickup.com/t/861m7rfpx)]

BigRoy:
## Brief description

This aims to resolve #4317 

## Additional info

The original code that forced the frame range data into it [was added about 6 months ago](https://github.com/ynput/OpenPype/commit/43d744b24b54f21ca2f7e622ca491af91833e689). It means that before that time everything worked splendidly without that logic :) This would mean it's been there since around 3.12.2

Looking back at our earlier projects I can clearly see that published Maya models, rigs, looks etc display no frame ranges in the Loader which is what I'd expect! At some points this was broken by the above mentioned commit.

## Testing notes:

1. Publish away your favorite instances!
2. Make sure frame data is as we'd like - those that set a frame range to extract should show a frame range in the loader, those that are not time-dependent but a single frame or full scene (workfile) should not show frame ranges.
3. Make sure loading and updating (setting version) of the publish still works as expected.


LiborBatek (dismissed):
It works as suppose to.

here is look asset before (note frame range 1001-1050)
4383_before

And after when using this PR branch all ok (again check fram range this time correct without frames)

4383_after

LiborBatek:
I have also tested to publish another family ```layout``` to be sure all works fine and it does!
4383_before2

and with this PR branch...
4383_after2

BigRoy:
I think this particular PR might benefit from testing in a DCC which might have been time dependent and might not have been explicitly setting frame data on the instances. I don't know of any, but maybe After Effects, TVPaint or Harmony is one of those?

Also, I believe when originally bringing up removing this logic is that there was some mention of "loaders" potentially relying on frame data being there too. Again, I'm not sure which those might be but that's what I recall came up in the discussion at the time.

iLLiCiTiT:
I'm afraid this is more complex change. What should be concidered as possible issue with this change:
- some extractors may be dependent on those values but their collectors don't fill the values because they expect that are already filled
- the missing values of frame start/end on version document can break some loaders which are dependent on frame start/end information

So the only way how to test this PR is to test all possible publishes that can produce image files, sequence or something that works with frames in all hosts from aftereffects to traypublisher.

BigRoy:
Yes, that's what I thought was the case. It's unfortunate this logic has been there for the past six months - if addressed earlier it might have been easier to only test the changes made since then.

A quick search in the codebase pointed me to about 20 files for Loaders which use `frameStart`
For files starting with `extract*` usage of `frameStartHandle` and `frameStart` is scattered over about 55 separate files.

So yes, these could be plenty. But then again, I'd call their current behavior "unexpected" since they should've been collecting the data intentionally to begin with?

Anyway, to fix this right - how should we go about it? Will we just test X random publishes across all hosts? or?

BigRoy:
We have actually pushed this into production on our end quite some time ago - and actually haven't had an issue with it since. However, it's good to note we solely use Maya, Houdini and Fusion and are not on latest develop.

Nonetheless, I'd say this is a "bug" / "feature" we should be solving rather sooner than later. The longer we wait the more likely it is that code has been expecting the data to be forced in there which some months ago wasn't the case yet and thus code from that time didn't expect it anyway.

The longer we wait to get to the bottom of this the harder it'll be to likely get us out of this sticky situation.

BigRoy:
@iLLiCiTiT We're using this in production but I'm not sure what would be needed to get this through review and the fix into OpenPype asap. Should we roll out a mass review with some random publishes across different hosts?

iLLiCiTiT:
I can think of Nuke, Hiero, Harmony, AfterEffects, TVPaint, Photoshop, TrayPublisher, some client custom publishers have to be tested. Both publishing, loading and check of output values (e.g. in burnins).

BigRoy:
> I can think of Nuke, Hiero, Harmony, AfterEffects, TVPaint, Photoshop, TrayPublisher, some client custom publishers (they're similar to standalone publisher so not necessary) have to be tested. Both publishing, loading and check of output values (e.g. in burnins).

@jakubjezek001 I assume you've also tried loading + updated the publishes?

So it seems remaining for review are:

- [ ] TVPaint
- [ ] Harmony
- [ ] Standalone Publisher

Who's up for that on the side of OpenPype?

iLLiCiTiT:
I had to fix TVPaint which was missing `"fps"` in https://github.com/ynput/OpenPype/pull/4674

BigRoy:
@iLLiCiTiT does that mean we can check of TVPaint and only need a test run in Harmony and Standalone Publisher? ❤️

64qam (approved):
Looks good in Harmony and Tray publisher

PR #4387, opened 2023-01-27
3dsmax: enhance alembic loader update function

antirotor:
## Enhancement

This PR is adding update/switch ability to pointcache/alembic loader in 3dsmax and fixing wrong tool shown when clicking on "Manage" item on OpenPype menu, that is now correctly Scene Inventory (but was Subset Manager).

Alembic update has still one caveat - it doesn't cope with changed number of object inside alembic, since loading alembic in max involves creating all those objects as first class nodes. So it will keep the objects in scene, just update path to alembic file on them.

### DIFF
--- a/openpype/hosts/max/plugins/load/load_pointcache.py
+++ b/openpype/hosts/max/plugins/load/load_pointcache.py
@@ -6,8 +6,10 @@ Because of limited api, alembics can be only loaded, but not easily updated.
 """
 import os
 from openpype.pipeline import (
-    load
+    load, get_representation_path
 )
+from openpype.hosts.max.api.pipeline import containerise
+from openpype.hosts.max.api import lib
 
 
 class AbcLoader(load.LoaderPlugin):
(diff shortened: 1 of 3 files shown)


openpype/hosts/max/plugins/load/load_pointcache.py
```diff
@@ -52,14 +54,47 @@ def load(self, context, name=None, namespace=None, data=None):
...
+
+    @staticmethod
+    def get_container_children(parent, type_name):
+        from pymxs import runtime as rt
+
+        def list_children(node):
+            children = []
+            for c in node.Children:
+                children.append(c)
+                children = children + list_children(c)
```
BigRoy, line 91:
```suggestion
                children += list_children(c)
```
Or
```suggestion
                children.extend(list_children(c))
```

Note that this could technically hit a recursion limit if a scene were ever nested deeper than 1000 levels.

openpype/hosts/max/plugins/load/load_pointcache.py
```diff
@@ -52,14 +54,47 @@ def load(self, context, name=None, namespace=None, data=None):
...
         from pymxs import runtime as rt
 
         node = container["node"]
         rt.delete(node)
+
+    @staticmethod
+    def get_container_children(parent, type_name):
+        from pymxs import runtime as rt
+
+        def list_children(node):
```
BigRoy, line 87: I feel `get_all_children` would be a more explicit name?

LiborBatek (approved):
The manage item in the OP menu correctly opens ```Scene Inventory``` window.

After loading abc animation cache in latest version  ```v007```  tested switching to older one ```v005``` sucesfully as seen on pic below
4387_ABC_versions01

Then saving and reopening the shot and update abc animation to latest ```v007``` via Scene Inventory worked ok as well.
4387_ABC_versions02

antirotor:
> But when I update the version to the latest in Scene Inventory, the object doesn't seem to update to the latest published abc as expected.

Might be related to one limitation I've mentioned - when you load alembic in 3dsmax for first time, it will create node for every geometry paths inside the alembic.  So if you have:
```
/obj/foo
/obj/boo
/bla
```

Then if you update version to alembic with different hierarchy, it will update the paths only on already existing nodes, not creating new one or deleting invalid. I left it like this semi-intentionally because the only option to do this properly would be to really delete everything and re-import it, but that would also delete any edits and associations with those nodes.

There might be way around this by crawling over the original nodes, serializing any changes and then reapplying them on newly imported nodes but hell.... that's for another PR.

moonyuet (approved):
I agree that there is some limitations with the abc import (but also others like fbx import), we can always go back to enhance the update function.

PR #4390, opened 2023-01-27
Global: extract colorspace is not failing

jakubjezek001:
## Brief description
Extract colorspace data is fixed

## Description
Plugin is not failing due missing config data. 

## Additional info
Order in `openpype\pipeline\publish\publish_plugins.py` had to be prioritised to first check supported file extension before a imageio settings should be equired. Also the host config settings are now optional since some older projects might have imageio group setting override with missing config and rules. 

## Testing notes:
1. open Maya and publish pointcache
2. all should be published without hiccups ;)

### DIFF
--- a/openpype/pipeline/colorspace.py
+++ b/openpype/pipeline/colorspace.py
@@ -344,9 +344,9 @@ def get_imageio_config(
     imageio_global, imageio_host = _get_imageio_settings(
         project_settings, host_name)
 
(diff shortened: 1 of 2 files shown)


LiborBatek (approved):
All seems ok.

Followed the testing steps and published ```pointcache``` family without any problems/errors.

image

Minkiu (approved):

Issue #4379, opened 2023-01-26
Houdini: New Publisher doesn't refresh attributes correctly


BigRoy:
**Running version**

3.15.0 (but also tested in 3.14.11-nightly.4, same issue)

**Describe the bug**

The New Publisher UI keeps data visible even on reset in a new scene. It'll also **infect** other instances in the scene (see Bug2). Only workaround is to close down houdini completely and restart with a non-infected scene.

As soon as you ever reset the Publisher UI with a scene that was 'infected' it'd infect all future scenes opened in the same session. Even when opening an existing scene that was fine before.

**To Reproduce**
1. Open Houdini
2. Open the new publisher
3. It should have generated a workfile instance, and thus have created a hidden `/obj/OpenPypeContext` node

Now run this python code:
```python
import hou
from openpype.hosts.houdini.api.lib import read, imprint

node = hou.node("/obj/OpenPypeContext")
data = read(node)

data["workfile"]["creator_attributes"]["my_fake_data"] = 10
imprint(node, data)
```
This code emulate storing creator attributes data for the workfile.

Now reset the publisher UI. It will show the fake data:
![afbeelding](https://user-images.githubusercontent.com/2439881/214835434-25332817-5d58-44c7-b3eb-063bfd4cced1.png)

**Bug 1: Data will always remain, it never goes away**

Now create a new Houdini file and reset the publisher UI.
The data still remains - which you wouldn't expect, because you're starting a new empty workfile.
The same happens if you explicitly delete the `/obj/OpenPypeContext` node in the scene to clear out any context data.

**Bug 2: Data will appear on ALL instances, not only the workfile**

Now create a pointcache instance, this will now suddenly also have this data:

![afbeelding](https://user-images.githubusercontent.com/2439881/214835724-05070366-9586-4410-82d2-2c69e1bf0c7e.png)

![afbeelding](https://user-images.githubusercontent.com/2439881/214835756-6c014183-dfcb-497c-9edb-e41941f36489.png)

But there's no reason for the pointcache to also get that data.

**Expected behavior**

It's like the UI just won't let go of the values that were visible in the UI before but are not relevant anymore.
I'd expect to just have a clean slate to work with when doing New File and resetting the publisher, but it doesn't.

**Additional context**

[My frustrations around this issue also came up on discord here](https://discord.com/channels/517362899170230292/517382145552154634/1068129295878344704) because it's hard to understand why the value remains.

[...]

PR #4392, opened 2023-01-27
General: Attribute definitions fixes

iLLiCiTiT:
## Brief description
Fix possible issues with attribute definitions in publisher if there is unknown attribute on an instance.

## Description
Source of the issue is that attribute definitions from creator plugin could be "expanded" during `CreatedInstance` initialization. Which would affect all other instances using the same list of attributes -> literally object of list. If the same list object is used in "BaseClass" for other creators it would affect all instances (because of 1 instance). There had to be implemented other changes to fix the issue and keep behavior the same.

Object of `CreatedInstance` can be created without reference to creator object. `CreatedInstance` is responsible to give UI attribute definitions (technically is prepared for cases when each instance may have different attribute definitions -> not yet).
Attribute definition has added more conditions for `__eq__` method and have implemented `__ne__` method (which is required for Py 2 compatibility). Renamed `AbtractAttrDef` to `AbstractAttrDef` (fix typo).

### Additional information
Currently multiselection of instances may cause that attributes are not loaded as fast (that because of very inefficient way how to find "common attributes" for all instances. That could be solved by implementing "create hash" method for attribute definitions. The result would be string which would contain all information which makes the attribute unique by combination of all definitions properties (could be huge for enums). We could look into that if we'll find out it's an issue.

```
# Example of attribute definition
NumberDef(
    "my_num",
    minimum=1.0,
    maximum=20.0,
    decimals=1,
    default=5.0,
    hidden=False,
    disabled=False
)

# Simplified possible tempalte for number definition
hast_str_template = "{class}|{key}|{hidde}|{disabled}|{default}|{minimum}|{maximum}|{decimals}"

# The result for the object above
hash_str = "NumberDef|my_num|0|0|5.0|1.0|20.0|1"
```

## Testing notes:
Hosts where is used New Publisher should work as before (or better).
1. Open one of the hosts e.g. TrayPublisher
2. Open publisher UI
3. Create part should show precreate attributes as expected
4. Create few instances
5. Instance attributes in `Publish` category should be shown correctly and they should work on multiselection

More complex testing:
Follow steps in this issue https://github.com/ynput/OpenPype/issues/4379

Resolves https://github.com/ynput/OpenPype/issues/4379

### DIFF
--- a/openpype/lib/__init__.py
+++ b/openpype/lib/__init__.py
@@ -30,7 +30,7 @@ from .vendor_bin_utils import (
 )
 
 from .attribute_definitions import (
-    AbtractAttrDef,
+    AbstractAttrDef,
 
     UIDef,
     UISeparatorDef,
@@ -246,7 +246,7 @@ __all__ = [
     "get_ffmpeg_tool_path",
     "is_oiio_supported",
 
-    "AbtractAttrDef",
+    "AbstractAttrDef",
 
     "UIDef",
     "UISeparatorDef",
--- a/openpype/lib/attribute_definitions.py
+++ b/openpype/lib/attribute_definitions.py
@@ -154,7 +154,12 @@ class AbtractAttrDef(object):
     def __eq__(self, other):
         if not isinstance(other, self.__class__):
             return False
-        return self.key == other.key
+        return (
+            self.key == other.key
+            and self.hidden == other.hidden
+            and self.default == other.default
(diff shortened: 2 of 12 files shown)


kalisp (approved):
Tested in Publisher, Nuke, AE, didn't notice irregularities from before.

(But I didn't have mentioned issues that this should fix.)

PR #4393, opened 2023-01-27
General: Simplified CreateContext api

iLLiCiTiT:
## Brief description
Method `create` on `CreateContet` is simplified version to trigger create plugins with standartized arguments.

## Description
There was not simple method as equivaleng of  `legacy_create` function for new creation system. The functionality is helpful for headless automations. That is implemented by adding new method. The method was named `create` which was already occupied so the origin was renamed to `create_with_unified_error`. Both of these methods returns whatever `create` method of creator plugin returns. The method will use current context context if asset name and task name are not passed.

Current context of `CreateContext` is not defined by object of `AvalonMongoDB` but context handle it on own.

### Additional information
Creator plugin does not require to return an instance from `create` method and `CreateContext` is not able to guess which instances were created during it's trigger so if the CreatePlugin does not return created instance it won't be returned by calling the method.

## Testing notes
Choose one of creator plugins in DCCs of your choice (there must be at leas one create for new publisher) and run code below.
```

variant = "Main"
# Fill creator identifier
creator_identifier = "put.here.creator.to.test"

host = registered_host()
context = CreateContext(host)
context.create(creator_identifier, variant)
```
This should create instance of your creator (or crash becase of create error).


openpype/pipeline/create/context.py
```diff
@@ -1117,11 +1107,20 @@ def host_is_valid(self):
 
     @property
     def host_name(self):
+        if hasattr(self.host, "name"):
+            return self.host.name
```
BigRoy, line 1111: Why was this added? Was this wrong originally?

openpype/pipeline/create/context.py
```diff
@@ -1458,6 +1458,65 @@ def create(self, identifier, *args, **kwargs):
...
+        """
+
+        if pre_create_data is None:
+            pre_create_data = {}
+
+        project_name = self.project_name
+        if asset_doc is None:
+            asset_name = self.get_current_asset_name()
+            asset_doc = get_asset_by_name(project_name, asset_name)
+            task_name = self.get_current_task_name()
```
BigRoy, line 1494: I feel like we might only want to do this if `task_name` is None?
Additionally however, it is technically possible to supply no task name to a creator too?

openpype/pipeline/create/context.py
```diff
@@ -1458,6 +1458,65 @@ def create(self, identifier, *args, **kwargs):
...
+                task_name,
+                asset_doc,
+                project_name,
+                self.host_name
+            )
+        instance_data = {
+            "asset": asset_doc["name"],
+            "task": task_name,
+            "variant": variant,
+            "family": family
```
BigRoy, line 1512: I think this is wrong - we shouldn't need to pass the family explicitly. The creator already produces that data, no? If not - that already seems invalid from the Creator's perspective, no?

openpype/pipeline/create/context.py
```diff
@@ -1458,6 +1458,65 @@ def create(self, identifier, *args, **kwargs):
...
+        if pre_create_data is None:
+            pre_create_data = {}
+
+        project_name = self.project_name
+        if asset_doc is None:
+            asset_name = self.get_current_asset_name()
+            asset_doc = get_asset_by_name(project_name, asset_name)
+            task_name = self.get_current_task_name()
+
+        creator = self.creators.get(creator_identifier)
```
BigRoy, line 1496: I feel like the check whether the Creator exists based on the identifier might need to be the first thing we check for? Everything else we're doing doesn't make any sense if we don't know what to create to begin with. It looks very odd that we're trying to continue with logic here even if the creator can't be identified.

openpype/pipeline/create/context.py
```diff
@@ -1458,6 +1458,65 @@ def create(self, identifier, *args, **kwargs):
...
+        pre_create_data=None
+    ):
+        """Trigger create of plugins with standartized
+
+        Args:
+            creator_identifier (str):
+            variant (str): Variant used for subset name.
+            asset_doc (Dict[str, Any]):
+            task_name (str): Name of task to which is context related.
+            pre_create_data (Dict[str, Any]): Pre-create attribute values.
```
BigRoy, line 1478: Docstring appears to be incomplete? Some descriptions are empty. Plus it should explain what is done/use if None is provided.

openpype/pipeline/create/context.py
```diff
@@ -1458,6 +1458,65 @@ def create(self, identifier, *args, **kwargs):
...
+
+    def create(
+        self,
+        creator_identifier,
+        variant=None,
+        asset_doc=None,
+        task_name=None,
+        pre_create_data=None
+    ):
+        """Trigger create of plugins with standartized
```
BigRoy, line 1471: Sentence incomplete/unclear.

openpype/tools/publisher/control.py
```diff
@@ -1589,20 +1589,19 @@ class PublisherController(BasePublisherController):
     Handle both creation and publishing parts.
 
     Args:
-        dbcon (AvalonMongoDB): Connection to mongo with context.
         headless (bool): Headless publishing. ATM not implemented or used.
     """
 
     _log = None
 
-    def __init__(self, dbcon=None, headless=False):
```
BigRoy, line 1598: Why is `dbcon` removed/redundant? Where is the connection/context of this particular context based on instead?

openpype/pipeline/create/context.py
```diff
@@ -1185,8 +1184,8 @@ def reset_finalization(self):
         self._collection_shared_data = None
         self.refresh_thumbnails()
 
-    def reset_avalon_context(self):
-        """Give ability to reset avalon context.
+    def reset_current_context(self):
+        """Refresh current context.
```
BigRoy, line 1188: Name is much better, nice!

I did notice that the docstring below is actually incomplete.

> tool but that change is not propagated to

To where?

openpype/pipeline/create/context.py
```diff
@@ -1458,6 +1458,65 @@ def create(self, identifier, *args, **kwargs):
...
+        variant=None,
+        asset_doc=None,
+        task_name=None,
+        pre_create_data=None
+    ):
+        """Trigger create of plugins with standartized
+
+        Args:
+            creator_identifier (str):
+            variant (str): Variant used for subset name.
```
BigRoy, line 1475: Is variant a requirement or not? :) I'm not sure. It seems optional currently due to being `variant=None`.

It'd be weird to define a subset name without a variant, no?

BigRoy (commented):
Decent draft, not perfect yet - here are some notes/questions.

openpype/pipeline/create/context.py
iLLiCiTiT: Based on this PR https://github.com/ynput/OpenPype/pull/4392 the instance does not need to have creator avalable.

BigRoy: > Based on this PR https://github.com/ynput/OpenPype/pull/4392 the instance does not need to have creator avalable.

Please explain. :) A Creator doesn't need to specify a family?

openpype/pipeline/create/context.py
iLLiCiTiT: It wasn't wrong, but if registered host is class of `HostBase` it's easier to use its `name` attribute.

BigRoy: If it's a special case, let's clarify in the code/docstring yes!

openpype/pipeline/create/context.py
BigRoy: If it's a special case, let's clarify in the code/docstring yes!

openpype/pipeline/create/context.py
BigRoy: Ok, check. To me personally this isn't _"easier"_. It's introducing more code and leaves me wondering why it's a 'special case'. I'd rather have it do one thing than branching between two things.

openpype/tools/publisher/control.py
iLLiCiTiT: This is just removing unnecessary middle-man. Connection from `dbcon` is not used at all and context is on `CreateContext`.

openpype/pipeline/create/context.py
iLLiCiTiT: It was required argument when the method was created (that's why it's first optional), then I changed it. But I don't have any valid reason why it should be optional > will change back to required.

openpype/pipeline/create/context.py
iLLiCiTiT: Or, make them required in this method and add new method where is explicitly used current context?

openpype/pipeline/create/context.py
iLLiCiTiT: > Please explain. :) A Creator doesn't need to specify a family?

No, `CreatedInstance` does not require to have passed `CreatorPlugin` in arguments.

iLLiCiTiT: Oh lord, the family is required argument for `CreatedInstance`. Will change.

openpype/pipeline/create/context.py
iLLiCiTiT: Ideal case is that only `self.host.name` is used all the time. The branching is because Publisher is in experimental tools so both must supported.

jakubjezek001 (changes requested):
I was testing in Nuke 13 (py3) create idetifier `create_model` and it worked as expected.

Then I had tried idetifier `create_write_render` and following occured. Nothing was created. I presume it has somethig to do with template for subset name using {task} or not sure what. There wasn't any more info to read from.

```
  File "<string>", line 10, in <module>
  File "C:\CODE\__PYPE\OpenPype_3.15_up\openpype\pipeline\create\context.py", line 1538, in create
    pre_create_data
  File "C:\CODE\__PYPE\OpenPype_3.15_up\openpype\pipeline\create\context.py", line 1466, in raw_create
    identifier, label, exc_info, add_traceback
openpype.pipeline.create.context.CreatorsCreateFailed: Faled to create instances
```

kalisp (changes requested):
Failing during publish in Nuke:
```
ERROR - Traceback (most recent call last):
  File "C:\Users\petrk\PycharmProjects\Pype3.0\pype\.venv\lib\site-packages\pyblish\plugin.py", line 522, in __explicit_process
    runner(*args)
  File "C:\Users\petrk\PycharmProjects\Pype3.0\pype\openpype\plugins\publish\collect_from_create_context.py", line 57, in process
AttributeError: CreateContext instance has no attribute 'dbcon'
  File "C:\Users\petrk\PycharmProjects\Pype3.0\pype\.venv\lib\site-packages\pyblish\plugin.py", line 522, in __explicit_process
    runner(*args)
  File "<string>", line 57, in process
AttributeError: CreateContext instance has no attribute 'dbcon'
ERROR 11:57:07.632:pyblish.plugin(23164): Traceback (most recent call last):
  File "C:\Users\petrk\PycharmProjects\Pype3.0\pype\.venv\lib\site-packages\pyblish\plugin.py", line 522, in __explicit_process
    runner(*args)
  File "C:\Users\petrk\PycharmProjects\Pype3.0\pype\openpype\plugins\publish\collect_from_create_context.py", line 57, in process
AttributeError: CreateContext instance has no attribute 'dbcon'
  File "C:\Users\petrk\PycharmProjects\Pype3.0\pype\.venv\lib\site-packages\pyblish\plugin.py", line 522, in __explicit_process
    runner(*args)
  File "<string>", line 57, in process
AttributeError: CreateContext instance has no attribute 'dbcon'
```

iLLiCiTiT:
Modified `create` to not call `raw_create` and modified `raw_create` to only pass args to create method -> which was irrelevant so the method was removed. Instead of previous `create` was implemented `create_with_unified_error` which is used in publisher.

jakubjezek001 (commented):
I have tried it again byt this happened on `create_write_render` idetifier
I am not sure if it is something nuke has to have fixed or it is failing of this PR.

```
  File "<string>", line 10, in <module>
  File "C:\CODE\__PYPE\OpenPype_3.15_up\openpype\pipeline\create\context.py", line 1884, in create
    pre_create_data
  File "C:\CODE\__PYPE\OpenPype_3.15_up\openpype\hosts\nuke\plugins\create\create_write_render.py", line 118, in create
    "render_target"
  File "C:\CODE\__PYPE\OpenPype_3.15_up\openpype\hosts\nuke\api\plugin.py", line 76, in pass_pre_attributes_to_instance
    creator_attrs[pass_key] = pre_create_data[pass_key]
KeyError: 'render_target'
```

iLLiCiTiT:
> I am not sure if it is something nuke has to have fixed or it is failing of this PR.

This means that your creator requires pre-create data which are not provided. That can't be handled by any logic in create context, it's job of the one who triggers the create method.

kalisp (approved):
Tested in AE, works.

jakubjezek001 (changes requested):
Nuke is not letting me to create `create_write_render` identifier.

```
  File "<string>", line 10, in <module>
  File "C:\CODE\__PYPE\OpenPype_3.15_up\openpype\pipeline\create\context.py", line 1884, in create
    pre_create_data
  File "C:\CODE\__PYPE\OpenPype_3.15_up\openpype\hosts\nuke\plugins\create\create_write_render.py", line 118, in create
    "render_target"
  File "C:\CODE\__PYPE\OpenPype_3.15_up\openpype\hosts\nuke\api\plugin.py", line 76, in pass_pre_attributes_to_instance
    creator_attrs[pass_key] = pre_create_data[pass_key]
KeyError: 'render_target'
```

iLLiCiTiT:
> Nuke is not letting me to create create_write_render identifier.


Again: There is nothing I can do anything about it. Your creator plugin requires to have passed `render_target` in precreate data. If you don't pass them then it crashes -> it's not job of this function to "magically" add it there.

iLLiCiTiT:
> That's what should be exposed to the user. It's the safest route.

So you suggest to get default precreate data from creator precreate attributes and "update" them with possibly passed precreate attributes and then call create?

jakubjezek001:
> > That's what should be exposed to the user. It's the safest route.
> 
> So you suggest to get default precreate data from creator precreate attributes and "update" them with possibly passed precreate attributes and then call create?

I like this solution. Althought it could be done in another PR.

jakubjezek001 (approved):
Issue with `create_write_render` could be addressed in another PR.

PR #4398, opened 2023-01-30
Global: supporting `OPENPYPE_TMPDIR` in staging dir maker

jakubjezek001:
## Brief description
Productions can use OPENPYPE_TMPDIR for staging temp publishing directory 

## Description
Studios were demanding to be able to configure their own shared storages as temporary staging directories. 
Template formatting is also supported with optional keys formatting and following anatomy keys:
    - root[work | <root name key>]
    - project[name | code]

## Documentation:
[future documentataion url](https://openpype.io/docs/admin_settings_system#environment)

## Testing notes:
1. configure `OPENPYPE_TMPDIR` on a testing project with `{root[work]}/{project[name]}/temp`
2. publish anything from any host. 
3. all should work as expected. 
4. notice temp foder in root of your project will appear.


openpype/pipeline/publish/lib.py
```diff
@@ -623,12 +641,58 @@ def get_instance_staging_dir(instance):
...
         )
-        instance.data["stagingDir"] = staging_dir
+    instance.data['stagingDir'] = staging_dir
+
+    instance.context.data["cleanupFullPaths"].append(staging_dir)
 
     return staging_dir
+
+
+def _formated_staging_dir(instance, openpype_temp_dir):
```
BigRoy, line 677: Currently has a typo in function name. `formated` -> `formatted`

I actually think however, we can also just refactor this to `_format_staging_dir`?

BigRoy (commented):
Nice feature! :)

Is it to be expected that there are no Studio Settings to set the temp dir and its up to the machine itself to "choose" the temp directory by setting a global `OPENPYPE_TEMP_DIR` variable from outside of OpenPype itself?

I like that a particular user/machine can set its own custom temporary directory, e.g. for if that machine has a strong local cache disk to use. But I felt that with the complex formatting that is possible using anatomy this feels more like a studio setting of sorts with a potential override in local settings?

Anyway, only other thing I'm worried about is that it breaks some logic I had set up for Substance Painter. In particular [this line of code](https://github.com/ynput/OpenPype/pull/4283/files#diff-4237e10cc78cc1f96af7ac5b06863eccc3aae714b5aa4fdcaac49e01622ad8f1R129) of using `publish.get_instance_staging_dir(instance)` from within a Collector because that is in a CollectorOrder BEFORE `instance.context.data["anatomy"]` already exists. Keeping that in mind this might be a change breaking other areas of code too.

**EDIT:**

### Missing documentation!

Also, this feature **must** have a mention in the documentation too. So please make sure the Admin Docs contain an explicit mention about the feature and how it's intended to be used by a studio or artist.

openpype/pipeline/publish/lib.py
```diff
@@ -623,12 +641,58 @@ def get_instance_staging_dir(instance):
...
+            )
+        )
+    else:
         staging_dir = os.path.normpath(
             tempfile.mkdtemp(prefix="pyblish_tmp_")
         )
-        instance.data["stagingDir"] = staging_dir
+    instance.data['stagingDir'] = staging_dir
+
+    instance.context.data["cleanupFullPaths"].append(staging_dir)
```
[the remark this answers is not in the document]
jakubjezek001: Yeah, that is a very good question. I believe this could stay here, but @iLLiCiTiT or @antirotor could have a different opinion?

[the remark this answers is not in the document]
iLLiCiTiT: Could the staging dir be output dir for farm or for after-publish purposes? If yes then it should not be auto-added.

openpype/pipeline/publish/lib.py
```diff
@@ -623,12 +641,58 @@ def get_instance_staging_dir(instance):
...
+                - root[work | <root name key>]
+                - project[name | code]
+                - asset
+                - hierarchy
+                - task
+                - username
+                - app
+            """
+    result = StringTemplate.format_template(openpype_temp_dir, anatomy_data)
+    result = os.path.normpath(result)
```
iLLiCiTiT, line 695:
```suggestion
    result = StringTemplate.format_template(
        openpype_temp_dir, anatomy_data).normalized
```

openpype/pipeline/publish/lib.py
```diff
@@ -623,12 +641,58 @@ def get_instance_staging_dir(instance):
...
+        return staging_dir
+
+    openpype_temp_dir = os.getenv("OPENPYPE_TEMP_DIR")
+    custom_temp_dir = None
+    if openpype_temp_dir:
+        if "{" in openpype_temp_dir:
+            custom_temp_dir = _formated_staging_dir(
+                instance, openpype_temp_dir
+            )
+        elif os.path.exists(openpype_temp_dir):
```
iLLiCiTiT, line 655: So if 'OPENPYPE_TEMP_DIR' is set but the directory does not exist (e.g. "F:/MyPath/temp") if won't be used?

BigRoy: That's seems odd, especially with a potentially formatted staging dir path I'd assume at least an `os.makedirs` would have been tried so that it could be dynamic folders?

jakubjezek001:
> Is it to be expected that there are no Studio Settings to set the temp dir and its up to the machine itself to "choose" the temp directory by setting a global `OPENPYPE_TEMP_DIR` variable from outside of OpenPype itself?

Yes that was the idea. Yes the documentation would be important for this.

openpype/pipeline/publish/lib.py
```diff
@@ -623,12 +641,58 @@ def get_instance_staging_dir(instance):
     Returns:
         str: Path to staging dir of instance.
     """
+    staging_dir = instance.data.get('stagingDir')
+    if staging_dir:
+        return staging_dir
+
+    openpype_temp_dir = os.getenv("OPENPYPE_TEMP_DIR")
```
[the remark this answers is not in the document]
jakubjezek001: yes, let's keep it in the same fashion ;)

openpype/pipeline/publish/lib.py
jakubjezek001: @iLLiCiTiT you are right, this use case can happen, but in case there is template `_format_staging_dir` is having its own `os.makedirs`

BigRoy: Ok, but then if it _were_ a path without template variables but a hardcoded path (e.g. a hardcoded local temp dir) it wouldn't get generated? Seems odd maybe?

iLLiCiTiT: If I would (as artist) set environment `OPENPYPE_TMPDIR` to `F:/MyPath/temp` but the folder does not exists and because it does not contain `{` it won't get to `_format_staging_dir` and thus won't be used.

BigRoy: But would you expect that, or not? I would be confused as an artist. :)

website/docs/admin_settings_system.md
```diff
@@ -13,18 +13,44 @@ Settings applicable to the full studio.
...
+    - username
+    - app
+
+#### OPENPYPE_DEBUG
+ - setting logger to debug mode
+ - example value: "1" (to activate)
+
+#### OPENPYPE_LOG_LEVEL
+ - stringified numeric value of log level. [Here for more info](https://docs.python.org/3/library/logging.html#logging-levels)
+ - example value: "10"
```
BigRoy, line 49: I would actually revert this docs change and move the special environment variables documentation into a page of its own. So that in Admin section for Configuration there's an "Environment Variables" page. This particular area of the docs could then just link to that page.

Why separate it? Because I'd actually argue that most of these environment variables mentioned here aren't supposed to be set in the global **environment** under system settings. But instead a particular machine could have a `OPENPYPE_LOG_LEVEL` env locally, or even have a "Environments" override in "Local Settings". There are many places where these are more likely to be used.

I'd just make a page dedicated for environment variables that are there to customize OP behavior. I'd also add `OPENPYPE_MONGO`, `OPENPYPE_USERNAME`, `OPENPYPE_LOCAL_ID`. Maybe even categorize some under **advanced** on that page.

Same goes for documenting `OPENPYPE_FFMPEG_PATHS` and `OPENPYPE_FFMPEG_PATHS`.

iLLiCiTiT:
Few comments:
- Current imlementation is using `OPENPYPE_TMPDIR` only at one place and there is no prepared function to get it from any other place. The implementation expects that it is used only for instance staging dir but it should be possible to get it out of instance scope. So it is not usable for other places where new temp directory should be created without using instance staging dir.
- The value is targeted for publishing and nothing else in that case it should be maybe called `OPENPYPE_PUBLISH_TMPDIR` because I would expect more usages then just publishing.
- The possible keys is overkill which is limiting usage of the value. I would limit possible keys only to project name/code and roots. I can't imagine any usecase where other keys from anatomy data would make sense.

- I would expect a function to just create tempdir (e.g. in `openpype/pipeline/tempdir.py` -> just made up) which is then used in publish staging dir.
```
def _format_tempdir(temp_dir, project_name, anatomy):
    if "{" not in temp_dir:
        return temp_dir

    if anatomy is None:

    # We still don't have `project_code` on Anatomy...
    project_doc = anatomy.get_project_doc_from_cache(project_name)
    data = {
        "root": anatomy.roots,
        "project": {
            "name": project_name,
            "code": project_doc["data"]["code"]
        }
    }
    return StringTemplate.format_template(temp_dir, data).normalized()


def create_tempdir(project_name, anatomy=None, prefix=None):
    if prefix is None:
        prefix = "op_tmp_"

    kwargs = {"prefix": prefix}
    openpype_temp_dir = os.getenv("OPENPYPE_TMPDIR")
    if openpype_temp_dir:
        openpype_temp_dir = _format_tempdir(
            openpype_temp_dir,
            project_name,
            anatomy
        )
        if not os.path.exists(openpype_temp_dir):
            os.makedirs(openpype_temp_dir)
        kwargs["dir"] = openpype_temp_dir

    return os.path.normpath(
        tempfile.mkdtemp(**kwargs)
    )
```

website/docs/admin_settings_system.md
jakubjezek001: Hey @BigRoy thanks for the suggestion. I have added this to backlog https://github.com/ynput/OpenPype/issues/4442

jakubjezek001: I have implemented the comment partly at this PR - I really liked the suggestion for separated Environment section. But I have added it under `Admin/Configuration`

jakubjezek001: @BigRoy If you are having suggestion for the Section introduction text, I am very happy to add it in it ;)

openpype/pipeline/temporarydir.py
```diff
@@ -0,0 +1,65 @@
+"""
```
iLLiCiTiT, line 1: I would rename it just to `tempdir.py`.

openpype/pipeline/publish/lib.py
```diff
@@ -623,12 +638,34 @@ def get_instance_staging_dir(instance):
...
+    if staging_dir:
+        return staging_dir
+
+    anatomy_data = instance.data.get("anatomy_data")
+    anatomy = instance.data.get("anatomy")
+
+    if anatomy_data:
+        project_name = anatomy_data["project"]["name"]
+    else:
+        project_name = os.getenv("AVALON_PROJECT")
```
iLLiCiTiT, line 651:
```suggestion
        project_name = instance.context.data["projectName"]
```

openpype/pipeline/publish/lib.py
```diff
@@ -623,12 +638,34 @@ def get_instance_staging_dir(instance):
     Returns:
         str: Path to staging dir of instance.
     """
+    staging_dir = instance.data.get('stagingDir')
+    if staging_dir:
+        return staging_dir
+
+    anatomy_data = instance.data.get("anatomy_data")
+    anatomy = instance.data.get("anatomy")
```
iLLiCiTiT, line 646: I think anatomy is only on context
```suggestion
    anatomy = instance.context.data.get("anatomy")
```

openpype/pipeline/temporarydir.py
```diff
@@ -0,0 +1,65 @@
...
+        data = {
+            "root": anatomy.roots
+        }
+        if formating_data is None:
+            # We still don't have `project_code` on Anatomy...
+            project_doc = anatomy.get_project_doc_from_cache(project_name)
+            data["project"] = {
+                "name": project_name,
+                "code": project_doc["data"]["code"],
+            }
```
iLLiCiTiT, line 45: After this [PR](https://github.com/ynput/OpenPype/pull/4445) can be modified to
```suggestion
            data["project"] = {
                "name": project_name,
                "code": anatomy.project_code,
            }
```

openpype/pipeline/temporarydir.py
```diff
@@ -0,0 +1,65 @@
+"""
+Temporary folder operations
+"""
+
+import os
+from openpype.lib import StringTemplate
+from openpype.pipeline import Anatomy
+
+
+def create_custom_tempdir(project_name, anatomy=None, formating_data=None):
```
iLLiCiTiT, line 10: There should be a function which create and return tempdir no matter if `OPENPYPE_TMPDIR` is set like in comment https://github.com/ynput/OpenPype/pull/4398#issuecomment-1422661943

iLLiCiTiT: This is not resolved. The function does nothing if `OPENPYPE_TMPDIR` is not set. Also `project_name` should be required.

iLLiCiTiT: You have to separate 2 things:
1. use different directory than default tmpdir by using `OPENPYPE_TMDIR` -> done
2. single function to create tempdir with prefix -> missing

openpype/pipeline/tempdir.py
```diff
@@ -0,0 +1,57 @@
+"""
+Temporary folder operations
+"""
+
+import os
+from openpype.lib import StringTemplate
+from openpype.pipeline import Anatomy
+
+
+def create_custom_tempdir(anatomy=None):
```
iLLiCiTiT, line 10: Project name should be requirement. It is the minimum required data for this function to work.

openpype/pipeline/tempdir.py
```diff
@@ -0,0 +1,57 @@
...
+    - optional key formating
+    - available keys:
+        - root[work | <root name key>]
+        - project[name | code]
+
+    Args:
+        anatomy (openpype.pipeline.Anatomy): Anatomy object
+
+    Returns:
+        bool | str: formated path or None
```
iLLiCiTiT, line 23:
```suggestion
        str | None: formated path or None
```

iLLiCiTiT (approved):

openpype/pipeline/tempdir.py
```diff
@@ -0,0 +1,58 @@
...
+import os
+from openpype.lib import StringTemplate
+from openpype.pipeline import Anatomy
+
+
+def create_custom_tempdir(project_name, anatomy=None):
+    """ Create custom tempdir
+
+    Template path formatting is supporting:
+    - optional key formating
```
antirotor, line 14:
```suggestion
    - optional key formatting
```

openpype/pipeline/tempdir.py
```diff
@@ -0,0 +1,58 @@
...
+    - available keys:
+        - root[work | <root name key>]
+        - project[name | code]
+
+    Args:
+        project_name (str): project name
+        anatomy (openpype.pipeline.Anatomy)[optional]: Anatomy object
+
+    Returns:
+        str | None: formated path or None
```
antirotor, line 24:
```suggestion
        str | None: formatted path or None
```

openpype/pipeline/tempdir.py
```diff
@@ -0,0 +1,58 @@
...
+        }
+        # path is anatomy template
+        custom_tempdir = StringTemplate.format_template(
+            openpype_tempdir, data).normalized()
+
+    else:
+        # path is absolute
+        custom_tempdir = openpype_tempdir
+
+    # create he dir path if it doesnt exists
```
antirotor, line 50:
```suggestion
    # create the dir path if it doesn't exists
```

openpype/pipeline/tempdir.py
```diff
@@ -0,0 +1,58 @@
...
+    if not os.path.exists(custom_tempdir):
+        try:
+            # create it if it doesnt exists
+            os.makedirs(custom_tempdir)
+        except IOError as error:
+            raise IOError("Path couldn't be created: {}".format(error))
```
antirotor, line 56: This should re-raise the exception either using python 3 syntax or use `six.reraise` if python 2 compatibility is required
```suggestion
        except IOError as error:
            raise IOError("Path couldn't be created: {}".format(error)) from error
```

openpype/pipeline/tempdir.py
```diff
@@ -0,0 +1,58 @@
...
+            openpype_tempdir, data).normalized()
+
+    else:
+        # path is absolute
+        custom_tempdir = openpype_tempdir
+
+    # create he dir path if it doesnt exists
+    if not os.path.exists(custom_tempdir):
+        try:
+            # create it if it doesnt exists
```
antirotor, line 53:
```suggestion
            # create it if it doesn't exists
```

openpype/pipeline/publish/lib.py
```diff
@@ -609,8 +615,17 @@ def context_plugin_should_run(plugin, context):
...
     """Unified way how staging dir is stored and created on instances.
 
-    First check if 'stagingDir' is already set in instance data. If there is
-    not create new in tempdir.
+    First check if 'stagingDir' is already set in instance data.
+    In case there already is new tempdir will not be created.
+
+    It also supports `OPENPYPE_TMPDIR`, so studio can define own temp
+    shared repository per project or even per more granular context.
+    Template formating is supported also with optional keys. Folder is
```
antirotor, line 623:
```suggestion
    Template formatting is supported also with optional keys. Folder is
```

openpype/pipeline/publish/lib.py
```diff
@@ -609,8 +615,17 @@ def context_plugin_should_run(plugin, context):
...
+    created in case it doesnt exists.
+
+    Available anatomy formatting keys:
+        - root[work | <root name key>]
+        - project[name | code]
 
     Note:
         Staging dir does not have to be necessarily in tempdir so be carefull
```
antirotor, line 631:
```suggestion
        Staging dir does not have to be necessarily in tempdir so be careful
```

antirotor: also the line below shoud be
```
about its use.
```
as it is possesive form. Can't add suggestions on the code not affected by this PR though :)

website/docs/admin_environment.md
```diff
@@ -0,0 +1,30 @@
...
+sidebar_label: Environment
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+## OPENPYPE_TMPDIR:
+ - Custom staging dir directory
+ - Supports anatomy keys formating. ex `{root[work]}/{project[name]}/temp`
+ - supported formating keys:
```
antirotor, line 13:
```suggestion
 - Supports anatomy keys formatting. ex `{root[work]}/{project[name]}/temp`
 - supported formatting keys:
```

website/docs/admin_settings_system.md
```diff
@@ -171,4 +176,4 @@ In the image before you can see that we set most of the environment variables in
 In this example MTOA will automatically will the `MAYA_VERSION`(which is set by Maya Application environment) and `MTOA_VERSION` into the `MTOA` variable. We then use the `MTOA` to set all the other variables needed for it to function within Maya.
 ![tools](assets/settings/tools_01.png)
 
-All of the tools defined in here can then be assigned to projects. You can also change the tools versions on any project level all the way down to individual asset or shot overrides. So if you just need to upgrade you render plugin for a single shot, while not risking the incompatibilities on the rest of the project, it is possible.
\ No newline at end of file
+All of the tools defined in here can then be assigned to projects. You can also change the tools versions on any project level all the way down to individual asset or shot overrides. So if you just need to upgrade you render plugin for a single shot, while not risking the incompatibilities on the rest of the project, it is possible.
```
antirotor, line 179:
```suggestion
All the tools defined in here can then be assigned to projects. You can also change the tools versions on any project level all the way down to individual asset or shot overrides. So if you just need to upgrade you render plugin for a single shot, while not risking the incompatibilities on the rest of the project, it is possible.
```

see https://proofed.com/writing-tips/word-choice-all-of/

antirotor (changes requested):
Some small typos.

Question - shouldn't this be set per context? I mean - it would be useful to influence the staging dir per family (or asset). Considering cases like special share for transient renders or big caches...

iLLiCiTiT:
> Question - shouldn't this be set per context? I mean - it would be useful to influence the staging dir per family (or asset). Considering cases like special share for transient renders or big caches...

It is still creating subdirectories in the temp dir, and must create them because of filename clashes.

antirotor (approved):

antirotor:
This is unfortunately breaking Python 2 hosts.

PR #4399, opened 2023-01-31
Batch script for running Openpype on Deadline.

tokejepsen:
## Description
Helps with development of render farm features by not having to build executable of openpype_console on Windows.

## Testing notes:
1. Add/replace `OpenPype Executable` in Deadline OpenPype plugin settings.
2. Submit render to Deadline.


kalisp:
Could some basic documentation be added to the file, please?

tokejepsen:
> Could some basic documentation be added to the file, please?

Sure, in what format? As a bulk comment or some `help` flag?

kalisp:
Just comments, similar to in existing .ps1 or .sh scripts.

tokejepsen:
> Just comments, similar to in existing .ps1 or .sh scripts.

done

kalisp:
Problem with this is that it still search for required version, as I understand it would work only if version env var is removed from job.

antirotor:
Could we also add *nix shell version?

tokejepsen:
> Could we also add *nix shell version?

Maybe for a different PR?
I've had a need for this, while testing farm rendering but if there is no need for Unix developers then dont invest the time?
I couldn't make Unix shell versions anyways as I'm on Windows.

tokejepsen:
> Problem with this is that it still search for required version, as I understand it would work only if version env var is removed from job.

@kalisp I've made this PR to tackle that issue; https://github.com/ynput/OpenPype/pull/4489

kalisp (approved):
Works.

mkolar:
Sounds like we can merge it?

PR #4400, opened 2023-01-31
Global: colorspace host settings are now optional

jakubjezek001:
## Brief description
Colorspace settings (imageio) are now optional

## Description
Settings might not be supported on some hosts and api was expecting it an all. 

## Additional info
API is now getting as fallback global imageio settings.

### DIFF
--- a/openpype/pipeline/colorspace.py
+++ b/openpype/pipeline/colorspace.py
@@ -465,6 +465,6 @@ def _get_imageio_settings(project_settings, host_name):
     # get image io from global and host_name
     imageio_global = project_settings["global"]["imageio"]
     # host is optional, some might not have any settings
-    imageio_host = project_settings[host_name].get("imageio", {})
+    imageio_host = project_settings.get(host_name, {}).get("imageio", {})
 
     return imageio_global, imageio_host
--- a/openpype/pipeline/colorspace.py
+++ b/openpype/pipeline/colorspace.py
@@ -438,7 +438,8 @@ def get_imageio_file_rules(project_name, host_name, project_settings=None):
 
     # get file rules from global and host_name
     frules_global = imageio_global["file_rules"]
-    frules_host = imageio_host["file_rules"]
+    # host is optional, some might not have any settings
+    frules_host = imageio_host.get("file_rules", {})
 
     # compile file rules dictionary
     file_rules = {}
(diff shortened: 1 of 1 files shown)


openpype/pipeline/colorspace.py
```diff
@@ -455,14 +456,15 @@ def _get_imageio_settings(project_settings, host_name):
...
         host_name (str): host name
 
     Returns:
         tuple[dict, dict]: image io settings for global and host
     """
     # get image io from global and host_name
     imageio_global = project_settings["global"]["imageio"]
-    imageio_host = project_settings[host_name]["imageio"]
+    # host is optional, some might not have any settings
+    imageio_host = project_settings[host_name].get("imageio", {})
```
iLLiCiTiT, line 468:
```suggestion
    imageio_host = project_settings.get(host_name, {}).get("imageio", {})
```

BigRoy: Yes please! Technically there's no **hard** requirement that the host has project settings at all.

jakubjezek001: I was thinking this might happen ;) Thanks!

openpype/pipeline/colorspace.py
```diff
@@ -438,7 +438,8 @@ def get_imageio_file_rules(project_name, host_name, project_settings=None):
 
     # get file rules from global and host_name
     frules_global = imageio_global["file_rules"]
-    frules_host = imageio_host["file_rules"]
+    # host is optional, some might not have any settings
+    frules_host = imageio_host.get("file_rules", {})
```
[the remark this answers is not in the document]
jakubjezek001: And I thought I've got it all.. Thanks!

iLLiCiTiT (approved):

openpype/pipeline/colorspace.py
```diff
@@ -438,13 +438,14 @@ def get_imageio_file_rules(project_name, host_name, project_settings=None):
...
 
     # compile file rules dictionary
     file_rules = {}
     if frules_global["enabled"]:
         file_rules.update(frules_global["rules"])
-    if frules_host["enabled"]:
+    if frules_host.get("enabled"):
```
BigRoy, line 448: How about that IF a host has imagio setting that it must have the enabled toggle? Or is there no default structure/schema we apply in settings to all hosts if it were to exist?
```suggestion
    if frules_host and frules_host["enabled"]:
```
This would at least error and tell us the `enabled` key doesn't exist. The current code would've instead just silently skipped the file rules.

jakubjezek001: Well, yes, this would also work. Yes, it uses default host schema and every file rule or config is having `enabled` toggle.

PR #4401, opened 2023-01-31
Maya: Fix Validate Attributes plugin

BigRoy:
## Brief description

Code was broken. So either plug-in was unused or it had gone unnoticed.

## Description

Looking at the commit history of the plug-in itself it seems this might have been broken somewhere between two to three years. I think it's broken since two years since [this commit](https://github.com/BigRoy/OpenPype/commit/4a300e0b087fb40e044a99f3a0d0d5c3835f0a19#diff-805f60be6c0433010b43f2495f96f7901397109d588b9b030dff36ccd24911aa).

Should this plug-in be removed completely?
@tokejepsen Is there still a use case where we should have this plug-in? (You created the original one)

## Testing notes:

1. Validate Attributes in maya, yay!

### DIFF
--- a/openpype/hosts/maya/plugins/publish/validate_attributes.py
+++ b/openpype/hosts/maya/plugins/publish/validate_attributes.py
@@ -58,23 +58,23 @@ class ValidateAttributes(pyblish.api.ContextPlugin):
             # Filter families.
             families = [instance.data["family"]]
             families += instance.data.get("families", [])
-            families = list(set(families) & set(self.attributes.keys()))
+            families = list(set(families) & set(cls.attributes.keys()))
             if not families:
                 continue
 
             # Get all attributes to validate.
             attributes = {}
             for family in families:
-                for preset in self.attributes[family]:
+                for preset in cls.attributes[family]:
                     [node_name, attribute_name] = preset.split(".")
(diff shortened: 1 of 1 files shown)


tokejepsen:
What is the alternative to validating attributes in Maya if we remove this?
We have been using it for production pipeline which is still on Pype 2.0, so around 2 years old seems about right.

BigRoy:
> What is the alternative to validating attributes in Maya if we remove this?

As far as I know there's no generic implementation which checks whether a node name exists and check its attribute. For rendering there are certain validators that check node type and their attributes, but only really for rendering per specific render engine instead of something as generic as this.

So if there's need for this - then let's keep the fixed version. It does mean however, we should get some automated tests in for this one so that ha! It doesn't stay broken for another two years?

tokejepsen:
> So if there's need for this - then let's keep the fixed version. It does mean however, we should get some automated tests in for this one so that ha! It doesn't stay broken for another two years?

Or we (Bumpybox) just need to hike up our boots and get the pipeline updated :)

Thanks for the fix though, we'll test it in 2 years time (Jokes! Hopefully...)

tokejepsen (approved):

PR #4402, opened 2023-01-31
Maya + Blender: Pyblish plugins removed unused `version` and `category` attributes

BigRoy:
## Brief description

Once upon a time in a land far far away there lived a few plug-ins who felt like they didn't belong in generic boxes and felt they needed to be versioned well above others. They tried, but with no success.

## Description

Even though they now lived in a universe with elaborate `version` and `category` attributes embedded into their tiny little plug-in DNA this particular deviation has been greatly unused. There is nothing special about the version, nothing special about the category.

It does nothing.

## Additional info

Having these attributes on these plug-ins just meant other developers would join in thinking _"hell yeah baby!"_ producing their own special plug-in babies with `category` and `version` attributes of their own. Well, they were fooled.

Their time spent on adding these precious attributes meant more lines of code that ended up doing nothing - except clutter and increase memory usage of machines across the world!

It's time to say, no more! This is where it stops.

Unless someone has a good use for these.

## Testing notes:

1. Test blender + maya publishing with these plug-ins, but they should remain unaffected.

### DIFF
--- a/openpype/hosts/blender/plugins/publish/validate_camera_zero_keyframe.py
+++ b/openpype/hosts/blender/plugins/publish/validate_camera_zero_keyframe.py
@@ -19,7 +19,6 @@ class ValidateCameraZeroKeyframe(pyblish.api.InstancePlugin):
     order = ValidateContentsOrder
     hosts = ["blender"]
     families = ["camera"]
-    version = (0, 1, 0)
     label = "Zero Keyframe"
     actions = [openpype.hosts.blender.api.action.SelectInvalidAction]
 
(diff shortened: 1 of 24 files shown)


iLLiCiTiT (approved):

BigRoy:
> BTW I don't know what category was meant for?

I guess the intent was to group plugins so the end user would see "these are for UVS, these are for meshes, these are for transforms" basically some sort of grouping to a category ai suppose. I believe it originated from our Colorbleed code originally - it was just always unused.

PR #4404, opened 2023-01-31
 3dsmax:  Add imageio settings

antirotor:
## Brief description
This is adding per project imageio settings for 3dsmax.

### DIFF
--- a/openpype/settings/defaults/project_settings/max.json
+++ b/openpype/settings/defaults/project_settings/max.json
@@ -1,0 +1,12 @@
+{
+    "imageio": {
+        "ocio_config": {
+            "enabled": false,
+            "filepath": []
+        },
+        "file_rules": {
+            "enabled": false,
+            "rules": {}
+        }
+    }
+}
--- a/openpype/settings/entities/schemas/projects_schema/schema_main.json
+++ b/openpype/settings/entities/schemas/projects_schema/schema_main.json
@@ -84,6 +84,10 @@
                 },
                 {
                     "type": "schema",
+                    "name": "schema_project_3dsmax"
+                },
+                {
+                    "type": "schema",
                     "name": "schema_project_maya"
                 },
                 {
--- a/openpype/settings/entities/schemas/projects_schema/schema_project_3dsmax.json
+++ b/openpype/settings/entities/schemas/projects_schema/schema_project_3dsmax.json
@@ -1,0 +1,26 @@
+{
+    "type": "dict",
+    "collapsible": true,
+    "key": "max",
+    "label": "3ds Max",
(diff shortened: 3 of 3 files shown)


BigRoy:
I'd say that adding these settings only makes sense if this also actually does something to set the OCIO config.

E.g. at the very least a prelaunch hook should be added which sets the OCIO config. Potentially the global prelaunch hook for OCIO I created for the #4283 would suffice already, it's [`openpype/hooks/pre_host_set_ocio.py`](https://github.com/ynput/OpenPype/blob/7cd999791296db63e21e84d5569ace97a5df0866/openpype/hooks/pre_host_set_ocio.py). But maybe 3DsMax requires more logic in that area too.

iLLiCiTiT (approved):

iLLiCiTiT:
> E.g. at the very least a prelaunch hook should be added which sets the OCIO config. Potentially the global prelaunch hook for OCIO I created for the https://github.com/ynput/OpenPype/pull/4283 would suffice already, it's [openpype/hooks/pre_host_set_ocio.py](https://github.com/ynput/OpenPype/blob/7cd999791296db63e21e84d5569ace97a5df0866/openpype/hooks/pre_host_set_ocio.py). But maybe 3DsMax requires more logic in that area too.

Maybe create separated PR just for the hook. I think the collection of information (ocio config path) could be done globaly but applying to environments should not.

BigRoy:
So then this PR adds settings which do nothing in the end? Or does something apply the OCIO config in 3dsmax already?

iLLiCiTiT:
> So then this PR adds settings which do nothing in the end? Or does something apply the OCIO config in 3dsmax already?

It's used for `ExtractColorspaceData` in `openpype/plugins/publish/extract_colorspace_data.py`.

BigRoy:
Sure. But that function would then be assuming an OCIO config from settings that 3dsmax isn't necessarily set to use.

antirotor:
As was stated, this is adding settings expected by `extract_colorspace_data`. 3dsmax support of OCIO (and color management) is ... lets say lacking, renderers are dealing with it their own way. I don't think it actually supports OCIO env var at all.

BigRoy:
I think what I'm getting at is that it seems invalid to assume we know _anything_ about the colorspace to extract if we don't even know what OCIO config is being used by whatever is generating the output. It makes what `extract_colorspace_data` tries to do 'automatically' just plain wrong information about the export.

Anyway - if it's what you need, go for it! Just wanted to address that particular point.

antirotor:
I suppose that with this #4400 it is not needed. Not until proper color management is implemented anyway.

PR #4405, opened 2023-01-31
Blender: fix import workfile all families

Tilix4:
## Brief description
Having this feature related to workfile available for any family is absurd.

### DIFF
--- a/openpype/hosts/blender/plugins/load/import_workfile.py
+++ b/openpype/hosts/blender/plugins/load/import_workfile.py
@@ -44,7 +44,7 @@ class AppendBlendLoader(plugin.AssetLoader):
     """
 
     representations = ["blend"]
-    families = ["*"]
+    families = ["workfile"]
 
     label = "Append Workfile"
     order = 9
@@ -68,7 +68,7 @@ class ImportBlendLoader(plugin.AssetLoader):
     """
 
     representations = ["blend"]
-    families = ["*"]
+    families = ["workfile"]
 
     label = "Import Workfile"
     order = 9


moonyuet (changes requested):
EDIT: Maybe you need to register in somewhere to make the import blend and append blend loader to show up, or use the original code.
image

moonyuet:
> EDITED: maybe the "workfile" also needs somewhere to load it right.
image

Tilix4:
> EDIT: Maybe you need to register in somewhere to make the import blend and append blend loader to show up, or use the original code. image

I'm sorry, I don't get your review. Could you please explain a bit more?

moonyuet:
> > EDIT: Maybe you need to register in somewhere to make the import blend and append blend loader to show up, or use the original code. image
> 
> I'm sorry, I don't get your review. Could you please explain a bit more?
when I was using your code with workfile added into families, the loader doesn't show the import option
But if I run the original code, it will be there.

Tilix4:
> > > EDIT: Maybe you need to register in somewhere to make the import blend and append blend loader to show up, or use the original code. image
> > 
> > 
> > I'm sorry, I don't get your review. Could you please explain a bit more?
> 
> when I was using your code with workfile added into families, the loader doesn't show the import option But if I run the original code, it will be there.

Tell me if I'm wrong, but it is actually the point. You mustn't see the loader for families other than `workfile`. And in your example you seem to test exactly the case where it must be disabled, confirming the change works.

moonyuet:
> > > > EDIT: Maybe you need to register in somewhere to make the import blend and append blend loader to show up, or use the original code. image
> > > 
> > > 
> > > I'm sorry, I don't get your review. Could you please explain a bit more?
> > 
> > 
> > when I was using your code with workfile added into families, the loader doesn't show the import option But if I run the original code, it will be there.
> 
> Tell me if I'm wrong, but it is actually the point. You mustn't see the loader for families other than `workfile`. And in your example you seem to test exactly the case where it must be disabled, confirming the change works.

ah i see. I misunderstand the intention of this PR and I resume my previous comment. If it is the case, i think you are right. It shouldn't load the workfile, instead there should be a family for raw blend scene included into the loader and the extractor for exporting the blend scene.

moonyuet (approved):

simonebarbieri (approved):

PR #4406, opened 2023-01-31
Tests: fix AE 

kalisp:
## Brief description
'png' extension have 2 representations instead 1. (One is regular, one is from review, with name 'png_png'


BigRoy:
> one is from review, with name 'png_png'

Shouldn't that be more appropriately named? Or is that like an `input -> output` type of naming convention?

64qam (approved):

kalisp:
> > one is from review, with name 'png_png'
> 
> Shouldn't that be more appropriately named? Or is that like an `input -> output` type of naming convention?

I think its input -> output, you could also have `png_mp4`, not sure origin of this logic though.

PR #4409, opened 2023-02-01
Maya: Change namespace for references

friquette:
## Brief description
Change namespace for references in Maya for a better readability. Proposed format: {asset}:{subset}_GRP


tokejepsen:
This is quite a fundamental change, so it'll need at least a setting that is off by default.

moonyuet:
For this case, it would be great if the users can set their own format for the namespace in the setting instead of hard-coding the format

LiborBatek:
If reviewable pls put some testing notes for user based testing, thx!

friquette:
Hi, I see you closed this PR without merging it. Will it be merged some day or do we need to modify things ?

PR #4411, opened 2023-02-01
Ftrack: Don't force ftrackapp endpoint

iLLiCiTiT:
## Brief description
Auto-fill of ftrack url don't break custom urls. Custom urls couldn't be used as `ftrackapp.com` is added if is not in the url.

## Description
The code was changed in a way that auto-fill is still supported but before `ftrackapp` is added it will try to use url as is. If the connection works as is it is used.

## Additional info
Can't test it as I wasn't able run ftrack server locally and couldn't run redirect proxy either.

## Testing notes:
1. Have ftrack server without `ftrackapp` in url
2. Set it in settings to be connected there
3. Run Tray and log in
4. Publish at least from one DCC

### DIFF
--- a/openpype/modules/ftrack/ftrack_module.py
+++ b/openpype/modules/ftrack/ftrack_module.py
@@ -510,7 +510,10 @@ def resolve_ftrack_url(url, logger=None):
         url = "https://" + url
 
     ftrack_url = None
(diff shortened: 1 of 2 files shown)


m-u-r-p-h-y:
do you have an idea how to test this one?

PR #4412, opened 2023-02-01
3dsmax: startup fixes

antirotor:
## Brief description
This is fixing various issues that can occur on some of the 3dsmax versions.

## Description
On displays with +4K resolution UI was broken, some 3dsmax versions couldn't process `PYTHONPATH` correctly. This PR is forcing `sys.path` and disabling `QT_AUTO_SCREEN_SCALE_FACTOR`

## Testing notes:
1. 3dsmax with OpenPype should start as expected
2. No UI issues on high resolution displays

### DIFF
--- a/openpype/hosts/max/startup/startup.py
+++ b/openpype/hosts/max/startup/startup.py
@@ -4,17 +4,12 @@ import sys
 
 # this might happen in some 3dsmax version where PYTHONPATH isn't added
 # to sys.path automatically
-try:
-    from openpype.hosts.max.api import MaxHost
-    from openpype.pipeline import install_host
-except (ImportError, ModuleNotFoundError):
+for path in os.environ["PYTHONPATH"].split(os.pathsep):
+    if path and path not in sys.path:
+        sys.path.append(path)
 
-    for path in os.environ["PYTHONPATH"].split(os.pathsep):
-        if path and path not in sys.path:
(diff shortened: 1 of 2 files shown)


openpype/hosts/max/startup/startup.py
```diff
@@ -1,6 +1,20 @@
...
+    from openpype.hosts.max.api import MaxHost
+    from openpype.pipeline import install_host
+except (ImportError, ModuleNotFoundError):
+
+    for path in os.environ["PYTHONPATH"].split(os.pathsep):
+        if path and path not in sys.path:
+            sys.path.append(path)
+
+    from openpype.hosts.max.api import MaxHost
+    from openpype.pipeline import install_host
```
iLLiCiTiT, line 17: Could be simplified to
```suggestion
for path in os.environ["PYTHONPATH"].split(os.pathsep):
    if path and path not in sys.path:
        sys.path.append(path)

from openpype.hosts.max.api import MaxHost
from openpype.pipeline import install_host
```

moonyuet:
I dont have the 4k monitor but I test the code, it doesn't break Openpype working in the max host.
Also tested the suggested code on the startup.py. It also works and doesn't break the plugin working in the host.

iLLiCiTiT (approved):

PR #4413, opened 2023-02-01
General: Remove 'openpype.api'

iLLiCiTiT:
## Brief description
PR is removing `openpype/api.py` file which is causing a lot of troubles and cross-imports.

## Description
I wanted to remove the file slowly function by function but it always reappear somewhere in codebase even if most of the functionality imported from there is triggering deprecation warnings. This is small change which may have huge impact.

There shouldn't be anything in openpype codebase which is using `openpype.api` anymore so only possible issues are in customized repositories or custom addons.

## Testing notes:
Nothing should crash on imports (easy to test, huh?).


jakubjezek001 (approved):
smells good to me ;)

kalisp (approved):
Seems to work fine.

PR #4414, opened 2023-02-01
Maya: Ignore workfile lock in Untitled scene

iLLiCiTiT:
## Brief description
Skip workfile lock check if current scene is 'Untitled'.

## Testing notes:
1. If Maya is opened in Untitled scene it should not crash on lock workfile check


moonyuet (approved):
Test it in my machine, works fine!

PR #4417, opened 2023-02-02
General: Use right environment variable for current task name

iLLiCiTiT:
## Brief description
Fix bug created in https://github.com/ynput/OpenPype/pull/4324 where for current task name is used `AVALON_ASSET` instead of `AVALON_TASK`.

## Description
Use `AVALON_TASK` instead of `AVALON_ASSET` in `get_current_task_name`.


kalisp (approved):
Fixes issue on DL with wrong task_name value.

PR #4418, opened 2023-02-02
Fix: DL on MacOS

kalisp:
## Brief description
This works if DL Openpype plugin Installation Directories is set to level of app bundle (eg. '/Applications/OpenPype 3.15.0.app')

### DIFF
--- a/openpype/modules/deadline/repository/custom/plugins/GlobalJobPreLoad.py
+++ b/openpype/modules/deadline/repository/custom/plugins/GlobalJobPreLoad.py
@@ -189,6 +189,12 @@ def get_openpype_executable():
     exe_list = config.GetConfigEntryWithDefault("OpenPypeExecutable", "")
     dir_list = config.GetConfigEntryWithDefault(
         "OpenPypeInstallationDirs", "")
+
+    # clean '\ ' for MacOS pasting
+    if exe_list:
+        exe_list = exe_list.replace("\\ ", " ")
+    if dir_list:
+        dir_list = dir_list.replace("\\ ", " ")
     return exe_list, dir_list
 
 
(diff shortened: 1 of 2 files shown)


kalisp:
Previous error was fixe by setting environment in System Settings:
```
'LC_ALL'='en_US.UTF-8'
'LANG'='en_US.UTF-8'
```
Not sure if is not only case of our test machine.

antirotor:
> Previous error was fixe by setting environment in System Settings:
> 
> ```
> 'LC_ALL'='en_US.UTF-8'
> 'LANG'='en_US.UTF-8'
> ```
> 
> Not sure if is not only case of our test machine.

I guess this is related to how Python is build using pyenv

antirotor (approved):
have not tested it, but these changes looks good

antirotor:
Now it's even better! :)

openpype/modules/deadline/repository/custom/plugins/GlobalJobPreLoad.py
```diff
@@ -189,6 +189,12 @@ def get_openpype_executable():
     exe_list = config.GetConfigEntryWithDefault("OpenPypeExecutable", "")
     dir_list = config.GetConfigEntryWithDefault(
         "OpenPypeInstallationDirs", "")
+
+    # clean '\ ' for MacOS pasting
+    if exe_list:
+        exe_list = exe_list.replace("\\ ", " ")
+    if dir_list:
+        dir_list = dir_list.replace("\\ ", " ")
```
iLLiCiTiT, line 197: If it's Mac specific it should happen on Mac. And I think the check of value is not needed as it should return always string.
```suggestion
    # clean '\ ' for MacOS pasting
    if platform.system().lower() == "darwin":
        exe_list = exe_list.replace("\\ ", " ")
        dir_list = dir_list.replace("\\ ", " ")
```

iLLiCiTiT (approved):

64qam (approved):

PR #4420, opened 2023-02-02
Deadline: Allow disabling strict error check in Maya submissions

kalisp:
## Brief description
DL by default has Strict error checking, but some errors are not fatal. 

## Description
This allows to set profile based on Task and Subset values to temporarily disable Strict Error Checks.

Subset and task names should support regular expressions. (not wildcard notation though).

## Testing notes:
1. Configure Settings in `project_settings/deadline/publish/MayaSubmitDeadline/strict_error_checking` - set to false
3. create new render instance, publish from Maya to DL, check `MayaBatch Settings` of job properties
2023-02-02 18_42_10-Testing Platform PC - AnyDesk

### DIFF
--- a/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
+++ b/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
@@ -35,6 +35,7 @@ from openpype.pipeline import legacy_io
 from openpype.hosts.maya.api.lib_rendersettings import RenderSettings
 from openpype.hosts.maya.api.lib import get_attr_in_layer
 
+from openpype.lib.profiles_filtering import filter_profiles
 from openpype_modules.deadline import abstract_submit_deadline
 from openpype_modules.deadline.abstract_submit_deadline import DeadlineJobInfo
 from openpype.tests.lib import is_in_tests
@@ -64,6 +65,7 @@ class MayaPluginInfo(object):
     # Include all lights flag
     RenderSetupIncludeLights = attr.ib(
         default="1", validator=_validate_deadline_bool_value)
+    StrictErrorChecking = attr.ib(default="1")
 
 
 @attr.s
@@ -104,6 +106,7 @@ class MayaSubmitDeadline(abstract_submit_deadline.AbstractSubmitDeadline):
     jobInfo = {}
     pluginInfo = {}
     group = "none"
+    disable_strict_check_profiles = []
 
     def get_job_info(self):
         job_info = DeadlineJobInfo(Plugin="MayaBatch")
@@ -219,6 +222,7 @@ class MayaSubmitDeadline(abstract_submit_deadline.AbstractSubmitDeadline):
             "renderSetupIncludeLights", default_rs_include_lights)
         if rs_include_lights not in {"1", "0", True, False}:
             rs_include_lights = default_rs_include_lights
+        strict_checking = self._get_strict_checking(instance)
         plugin_info = MayaPluginInfo(
             SceneFile=self.scene_path,
             Version=cmds.about(version=True),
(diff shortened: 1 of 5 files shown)


openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
```diff
@@ -64,6 +65,7 @@ class MayaPluginInfo(object):
     # Include all lights flag
     RenderSetupIncludeLights = attr.ib(
         default="1", validator=_validate_deadline_bool_value)
+    StrictErrorChecking = attr.ib(default="1")
```
iLLiCiTiT, line 68: Should it be string or int?
```suggestion
    StrictErrorChecking = attr.ib(default=1)
```

openpype/settings/entities/schemas/projects_schema/schema_project_deadline.json
```diff
@@ -195,6 +195,36 @@
...
+                        },
+                        {
+                            "type": "list",
+                            "collapsible": true,
+                            "key": "disable_strict_check_profiles",
+                            "label": "Disable Strict Error Check profiles",
+                            "use_label_wrap": true,
+                            "docstring": "Set profile for disabling 'Strict Error Checking'",
+                            "object_type": {
+                                "type": "dict",
```
iLLiCiTiT, line 207: I would add one boolean input which would tell if it should be strict or not. There is no way how to tell: By default enable strict error check, except one very specific case.

iLLiCiTiT: Or maybe for all `Scene` task types make it strict, but if task name is `TestScene` then ignore it.

iLLiCiTiT: Maybe: In that case the logic should be reversed to "enable" strict check profiles, so the profile would enable strict mode or disable strict mode?

kalisp: Default is Strict check enabled, that is DL default, even when we didn't sending anything.
This functionality is only as a workaround to push trough some critical publish which is failing on minor recoverable error.

We are now discussing with @m-u-r-p-h-y  if profiles are necessary, if simple boolean toggle wouldn't be enough.

m-u-r-p-h-y (changes requested):
I'm afraid we are overcomplicating this. Sorry for the confusion. I would suggest simplifying the whole process and creating simple default settings in 

`project_settings/deadline/publish/MayaSubmitDeadline` for Strict Error Checking (default on)

strictError

and reading the value when creating a render instance in Maya. This way user can decide to switch it off for particular scene which is failing on Deadline farm for a non fatal reason.

strictErrorMaya

kalisp:
Removed profiles, changed to boolean, set to render instance.

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
```diff
@@ -104,6 +105,7 @@ class MayaSubmitDeadline(abstract_submit_deadline.AbstractSubmitDeadline):
     jobInfo = {}
     pluginInfo = {}
     group = "none"
+    disable_strict_check_profiles = []
```
iLLiCiTiT, line 108:
```suggestion
```

iLLiCiTiT (approved):
One small suggestion, otherwise approved.

m-u-r-p-h-y (changes requested):
it does not work me.

Maya2022
Deadline 10.2.0.9

newly created Render instance is picking correct value from settings (false), but submitted job has StrictErrorChecking enabled anyway

image

kalisp:
Fixed propagation.

m-u-r-p-h-y (approved):
now everything works as expected.

tested a couple of scenarios.

Settings = true
Render instance = false

Settings = true
Render instance = true

Settings = false
Render instance = true

PR #4422, opened 2023-02-02
Maya: V-Ray Proxy - Update to latest fails - OP-2008

tokejepsen:
## Brief description
Updating the Vray proxy would fail.

Also added validator to ensure vray plugin is enabled since publishing could go past validation without needing vray and fail at extraction.

## Testing notes:
1. Create Vray Proxy in Maya.
2. Publish.
3. Load into Maya.
4. Publish a new version and update.

### DIFF
--- a/openpype/hosts/maya/plugins/load/load_vrayproxy.py
+++ b/openpype/hosts/maya/plugins/load/load_vrayproxy.py
@@ -81,10 +81,11 @@ class VRayProxyLoader(load.LoaderPlugin):
         c = colors.get(family)
         if c is not None:
             cmds.setAttr("{0}.useOutlinerColor".format(group_node), 1)
-            cmds.setAttr("{0}.outlinerColor".format(group_node),
-                (float(c[0])/255),
-                (float(c[1])/255),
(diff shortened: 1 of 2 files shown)


LiborBatek (changes requested):
All works as expected. Followed testing steps and all ok.

image


The only missing feature is Publishing also ```hero version``` now just versioned published assets. I guess should work as other ```instances``` like ```pointcache``` or ```animation``` families.

So it is needed to be implemented for vrayProxies too.

tokejepsen:
> The only missing feature is Publishing also `hero version` now just versioned published assets. I guess should work as other `instances` like `pointcache` or `animation` families.

I cant see any issue publish and updating to hero version. Do you have any error logs?

LiborBatek (approved):

LiborBatek:
After alterning the OP settings for ```Integrate Hero Version``` and adding ```vrayproxy``` family all works fine and publish makes version and hero asset too.
image

and in maya
image

So Approved.

antirotor (approved):

PR #4423, opened 2023-02-02
Fusion: Fix files collection and small bug-fixes

EmberLightVFX:
## Brief description
Fixed Fusion review-representation and small bug-fixes

## Description
This fixes the problem with review-file generation that stopped the publishing on second publish before the fix.
The problem was that Fusion simply looked at all the files in the render-folder instead of only gathering the needed frames for the review.

Also includes a fix to get the handle start/end that before throw an error if the data didn't exist (like from a kitsu sync).

### DIFF
--- a/openpype/hosts/fusion/api/lib.py
+++ b/openpype/hosts/fusion/api/lib.py
@@ -68,8 +68,8 @@ def set_asset_framerange():
     asset_doc = get_current_project_asset()
     start = asset_doc["data"]["frameStart"]
     end = asset_doc["data"]["frameEnd"]
-    handle_start = asset_doc["data"]["handleStart"]
-    handle_end = asset_doc["data"]["handleEnd"]
+    handle_start = asset_doc["data"].get("handleStart")
+    handle_end = asset_doc["data"].get("handleEnd")
     update_frame_range(start, end, set_render_range=True,
                        handle_start=handle_start,
                        handle_end=handle_end)
--- a/openpype/hosts/fusion/api/lib.py
+++ b/openpype/hosts/fusion/api/lib.py
@@ -68,8 +68,8 @@ def set_asset_framerange():
     asset_doc = get_current_project_asset()
     start = asset_doc["data"]["frameStart"]
     end = asset_doc["data"]["frameEnd"]
-    handle_start = asset_doc["data"].get("handleStart")
-    handle_end = asset_doc["data"].get("handleEnd")
+    handle_start = asset_doc["data"]["handleStart"]
+    handle_end = asset_doc["data"]["handleEnd"]
     update_frame_range(start, end, set_render_range=True,
                        handle_start=handle_start,
                        handle_end=handle_end)
--- a/openpype/hosts/fusion/api/lib.py
+++ b/openpype/hosts/fusion/api/lib.py
@@ -210,7 +210,7 @@ def switch_item(container,
     if any(not x for x in [asset_name, subset_name, representation_name]):
         repre_id = container["representation"]
         representation = get_representation_by_id(project_name, repre_id)
-        repre_parent_docs = get_representation_parents(representation)
+        repre_parent_docs = get_representation_parents(project_name, representation)
         if repre_parent_docs:
             version, subset, asset, _ = repre_parent_docs
         else:
--- a/openpype/hosts/fusion/api/lib.py
+++ b/openpype/hosts/fusion/api/lib.py
@@ -210,7 +210,8 @@ def switch_item(container,
     if any(not x for x in [asset_name, subset_name, representation_name]):
         repre_id = container["representation"]
         representation = get_representation_by_id(project_name, repre_id)
-        repre_parent_docs = get_representation_parents(project_name, representation)
+        repre_parent_docs = get_representation_parents(
+            project_name, representation)
         if repre_parent_docs:
             version, subset, asset, _ = repre_parent_docs
         else:
--- a/openpype/hosts/fusion/plugins/publish/render_local.py
+++ b/openpype/hosts/fusion/plugins/publish/render_local.py
@@ -37,7 +37,8 @@ class Fusionlocal(pyblish.api.InstancePlugin):
         basename = os.path.basename(path)
         head, ext = os.path.splitext(basename)
         files = [
-            f"{head}{str(frame).zfill(4)}{ext}" for frame in range(frame_start, frame_end + 1)
+            f"{head}{str(frame).zfill(4)}{ext}"
+            for frame in range(frame_start, frame_end + 1)
         ]
         repre = {
             'name': ext[1:],
--- a/openpype/hosts/fusion/plugins/publish/render_local.py
+++ b/openpype/hosts/fusion/plugins/publish/render_local.py
@@ -1,6 +1,4 @@
 import os
-import copy
-
 import pyblish.api
 from openpype.hosts.fusion.api import comp_lock_and_undo_chunk
 
@@ -19,7 +17,7 @@ class Fusionlocal(pyblish.api.InstancePlugin):
     families = ["render.local"]
 
     def process(self, instance):
-        
+
         # This plug-in runs only once and thus assumes all instances
         # currently will render the same frame range
         context = instance.context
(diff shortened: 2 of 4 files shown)


openpype/hosts/fusion/plugins/publish/render_local.py
```diff
@@ -23,42 +21,28 @@ def process(self, instance):
...
-            })
-
-        if "representations" not in instance.data:
-            instance.data["representations"] = []
-
-        collected_frames = os.listdir(output_dir)
+        basename = os.path.basename(path)
+        head, ext = os.path.splitext(basename)
+        files = [
+            f"{head}{str(frame).zfill(4)}{ext}" for frame in range(frame_start, frame_end + 1)
```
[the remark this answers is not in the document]
EmberLightVFX: I don't know how to make that string shorter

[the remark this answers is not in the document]
iLLiCiTiT:
```suggestion
            f"{head}{str(frame).zfill(4)}{ext}"
            for frame in range(frame_start, frame_end + 1)
```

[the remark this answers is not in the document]
iLLiCiTiT: Question is if there should be used hardcoded padding of `4`. If anyone would use frame start `10001` it would not work.

openpype/hosts/fusion/plugins/publish/render_local.py
```diff
@@ -23,42 +21,28 @@ def process(self, instance):
         # This plug-in runs only once and thus assumes all instances
         # currently will render the same frame range
         context = instance.context
-        key = "__hasRun{}".format(self.__class__.__name__)
```
iLLiCiTiT, line 26: I think this had a reason?

iLLiCiTiT: You've moved the logic into `render_once` but the logic before also skipped collection of filenames.

openpype/hosts/fusion/api/lib.py
```diff
@@ -68,8 +68,8 @@ def set_asset_framerange():
     asset_doc = get_current_project_asset()
     start = asset_doc["data"]["frameStart"]
     end = asset_doc["data"]["frameEnd"]
-    handle_start = asset_doc["data"]["handleStart"]
```
iLLiCiTiT, line 71: I would say that these should be required.

openpype/hosts/fusion/plugins/publish/render_local.py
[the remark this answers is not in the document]
EmberLightVFX: Currently the creator creates a saver node with the path `renderCompositingMain..exr`
Fusion will check before before the `.exr` if there is any numbers or `#`. If not it will add the frame number with a minimum of 4 numbers.
So as long as the creator doesn't change how the path looks like that zfill(4) will work what ever number you'll put in

openpype/hosts/fusion/api/lib.py
EmberLightVFX: I changed from `asset_doc["data"]["handleStart"]` to `asset_doc["data"].get("handleStart")` so as `update_frame_range()` defaults handleStart/End to 0 if there is no data being given. Else the code would fail if there were no handleStart/End being set.

iLLiCiTiT: From my point of view this is step back as `handleStart` and `handleEnd` should be required, but I'll let @jakubjezek001 and others to leave their opinions.

EmberLightVFX: This was a fix I made as the data from Kitsu didn't create any handleStart/End values for the asset. As update_frame_range() defaulted to 0 I thought an error in the code was made at some point.
But after doing some work on the Kitsu code it now creates all needed values.

As long as all other modules creates all required values this code could be removed.

openpype/hosts/fusion/plugins/publish/render_local.py
EmberLightVFX: Moved it back to the process function.

iLLiCiTiT: NOTE: Maybe it was right. I think @jakubjezek001 or @BigRoy could tell.

EmberLightVFX: Looking at how the old code worked, moving it back would result in the same behavior as before. But if Jakub och Roy says something else I'm more than happy to revert the change!

openpype/hosts/fusion/api/lib.py
jakubjezek001: Hey @EmberLightVFX, it is unfortunate that Kitsu is not yet integrating `handleStart` and `handleEnd` but all our DCC integrations are in consistency with those attributes to be required. Solution would be to fix this from above. Kitsu integration should add those attributes at least in 0 form by default. 

It would be cool to merge it soon as I think this fix is very beneficial. We could allow this to be merged temporarily. Favorabely to be cross referenced to the Kitsu integration issue you've mentioned. Also, you can add some sort of hack or to-do comment above the change, please.

openpype/hosts/fusion/plugins/publish/render_local.py
```diff
@@ -67,7 +58,23 @@ def process(self, instance):
         repre_preview["tags"] = ["review", "preview", "ftrackreview", "delete"]
```
jakubjezek001, line 58:
```suggestion
        repre_preview["tags"] = ["review",  "ftrackreview", "delete"]
```

This can also go. As far as I know `preview` tag is already gone for some time. Thanks.

openpype/hosts/fusion/plugins/publish/render_local.py
jakubjezek001: This seems to be ok with me - as far as I had tested it, it works and that is important! ;)

jakubjezek001 (changes requested):
Please just cross reference the Kitsu issue and this is good to go ;)

EmberLightVFX:
Fixed all commented problems and I would say that the code is ready to be merged :)

jakubjezek001 (approved):
good work, thank you!

openpype/hosts/fusion/plugins/publish/render_local.py
BigRoy: Actually no - I think this is now still broken. This logic should've moved to `render_once()`. Because as far as I can tell from the code, if you had multiple savers then all instances that are not the first processed saver instance would NOT have their representations added and thus would not be published/integrated.

@EmberLightVFX can you confirm the logic works for multiple savers from a single comp, do they all get published correctly?

Also, reading it over again - it should even store whether the render failed or not - so that the subsequent instances being processed now whether the render was completed or not. Currently it'd only raise the runtime error on the first instance.

EmberLightVFX: When reading the code again, yeah something doesn't feel complete correct.
It feels like some kind of for-loop is missing to go over each and every saver instance to append to data["representations"].

Right now it works the way the old code worked but updated. Did that work with comps with multiple savers?
I'll do some multi-saver tests when I get some minutes over!

BigRoy: This pyblish plug-in is an InstancePlugin, so it runs **per instance**. Which I think is good.
However, since currently Fusion integration is set to have all saver instances always render the fusion comp frame range it's much more optimized to render all at once instead of one by one (which will be massively slower if sharing parts of the comp).

As such, the intent was to:
- have something that ensures it only renders once, for all instances.
- have each instance check whether the render succeeded or not - if it did, have it add the representation into the instance.

The other solution would be to make it a Pyblish ContextPlugin, to then find the right instances, etc. - but I'd refrain from doing that personally because a next step would be making it so that - with the new publisher - we can also customize the frame ranges per saver we want to actually render. 💪 For which we'll end up needing InstancePlugins I'd say.

> Right now it works the way the old code worked but updated. Did that work with comps with multiple savers?

I've no idea what happened where but before OpenPype in our pipeline (on which this logic is based) this definitely worked just fine. But in OP I think it must have been broken for a while somehow. Or I'm massively misreading the code. :)

EmberLightVFX: > However, since currently Fusion integration is set to have all saver instances always render the fusion comp frame range it's much more optimized to render all at once instead of one by one (which will be massively slower if sharing parts of the comp).

Ooohhh! Now I see what you mean! Yeah, that makes total sense that the check would be in the render_once function!
Gonna make a new PR with the fix and add some comments about it in the file itself for future readers.

PR #4424, opened 2023-02-02
Maya: fps rounding - OP-2549

tokejepsen:
## Brief description
When FPS is registered in for example Ftrack and round either down or up (floor/ceil), comparing to Maya FPS can fail. Example:

23.97 (Ftrack/Mongo) != 23.976023976023978 (Maya)

## Description
Since Maya only has a select number of supported framerates, I've taken the approach of converting any fps to supported framerates in Maya. We validate the input fps to make sure they are supported in Maya in two ways:

Whole Numbers - are validated straight against the supported framerates in Maya.
Demical Numbers - we find the closest supported framerate in Maya. If the difference to the closest supported framerate, is more than 0.5 we'll throw an error.

If Maya ever supports arbitrary framerates, then we might have a problem but I'm not holding my breath...

## Testing notes:
1. Setup shot in Ftrack with fps 23.976023976023978, which will be rounded to 23.97.
2. Launch Maya.
3. Create Animation instance.
4. Publish.

There are other testing notes on https://github.com/ynput/OpenPype/pull/4076


m-u-r-p-h-y (dismissed):
I have a  project with fps set to 29.98

shot has fps set to 23.9760246277 (number coming from Hiero editorial publish)

a scene in Maya set to 25fps

during save, the popup informs about the wrong fps, but the numbers are not matching

image

after pressing fix, everything seems ok, saving works, no complaints anymore

but when you use OpenPype -> Reset frame range, the scene is set to 25fps with no error in the script editor

tokejepsen:
@m-u-r-p-h-y should be fixed now.

antirotor (commented):
It does work for me - the only question is the allowed error. I've tested it with frame rate `23.5` that doesn't make sense but  since the difference is just `0.4760...` and rounded `0` it is still accepted as `23.976023976023978`. And maybe it's good thing? But I feel that the tolerance should be less, maybe in order of `0.1`

tokejepsen:
> It does work for me - the only question is the allowed error. I've tested it with frame rate `23.5` that doesn't make sense but since the difference is just `0.4760...` and rounded `0` it is still accepted as `23.976023976023978`. And maybe it's good thing? But I feel that the tolerance should be less, maybe in order of `0.1`

Sure, I dont mind having a lower tolerance. We can maybe loosen it later if productions have issues with it. No point in having it a setting, right?

LiborBatek (approved):
Following testing steps and setting ```custom framerate``` in ftrack/mongoDB to ```23.976023976023978```

OP correctly sets the framerate according to DB
image

Did run the full test with creating / publishing /loading assets and all works as expected. When changing the shot to other framerate than in DB OP correctly fixes it back when re-opened.

antirotor (approved):

PR #4437, opened 2023-02-08
Fix - remove minor part in toml

kalisp:
## Brief description
Causes issue in create_env and new Poetry


kalisp:
Yeah, this is just temporary band aid until automation is fixed.

PR #4444, opened 2023-02-09
Maya: Arnold Scene Source overhaul - OP-4865

tokejepsen:
## Brief description
General overhaul of the Arnold Scene Source (ASS) workflow.

## Description
This originally was to support static files (non-sequencial) ASS publishing, but digging deeper that whole workflow needed and update to get ready for further issues. During this overhaul the following changes were made:

- Generalized Arnold Standin workflow to a single loader.
- Support multiple nodes as proxies.
- Support proxies for `pointcache` family.
- Generalized approach to proxies as resources, so they can be the same file format as the original.

This workflow should allow further expansion to utilize operators and eventually USD.

## Discussion
By supporting proxies for the `pointcache` family, we can also get rid of the specialized `proxyAbc` workflow. What do we think of that?
We could also implement/support proxy workflows when it comes to animation and rig, although not sure how much that is used outside of big studios.

## Testing notes:
1. Create and publish Arnold Scene Source or Pointcache instance.
2. Load and render.
3. Repeat with proxy workflow to see proxy in viewport and original at render time.


tokejepsen:
Closing this in a favour of a different fork origin.

tokejepsen:
https://github.com/ynput/OpenPype/pull/4449

PR #4446, opened 2023-02-09
General: Ignore decode error of stdout/stderr in run_subprocess

iLLiCiTiT:
## Brief description
Ignore decode errors and replace invalid character (byte) with escaped byte character.

## Description
Calling of `run_subprocess` may cause crashes if output contains some unicode character which (for example Polish name of encoder handler).

## Testing notes:
Issue was discovered by a client so testing notes are related to the case.
1. Open TrayPublisher
2. Select family (e.g. Render)
3. Drop video from client to both (source and reveiw) input fields
4. Publish
5. Publishing should went smoothly without issues in ExtractReview plugin

### DIFF
--- a/openpype/hosts/harmony/plugins/publish/extract_render.py
+++ b/openpype/hosts/harmony/plugins/publish/extract_render.py
@@ -108,9 +108,9 @@ class ExtractRender(pyblish.api.InstancePlugin):
         output = process.communicate()[0]
 
         if process.returncode != 0:
-            raise ValueError(output.decode("utf-8"))
+            raise ValueError(output.decode("utf-8", errors="backslashreplace"))
 
-        self.log.debug(output.decode("utf-8"))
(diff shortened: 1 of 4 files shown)


64qam:
Now it crashes on Extract burnins.

Minkiu (commented):
Maybe it should do de decoding on the `args` and `kwargs` of the subprocess?

iLLiCiTiT:
> Now it crashes on Extract burnins.

Should be fixed

> Maybe it should do de decoding on the args and kwargs of the subprocess?

It's not issue with args but with output from process.

64qam:
still crashing on burnins
[publish-report.txt](https://github.com/ynput/OpenPype/files/10706638/publish-report.txt)

iLLiCiTiT:
> still crashing on burnins

Forgot tu push one commit

64qam (approved):

PR #4448, opened 2023-02-10
Maya: Create Arnold options on repair.

tokejepsen:
## Brief description
When validating/repairing we previously required users to open render settings to create the Arnold options. This is done through code now.

## Testing notes:
1. Launch Maya and setup lighting scene without opening render settings.
2. Publish.


Minkiu (approved):

moonyuet:
I tested the texture with absolute path , it is errored out as expected
image

and when running the repair action, it works as expected
image

moonyuet (approved):

PR #4449, opened 2023-02-10
Maya: Arnold Scene Source overhaul - OP-4865

tokejepsen:
## Brief description
General overhaul of the Arnold Scene Source (ASS) workflow.

## Description
This originally was to support static files (non-sequencial) ASS publishing, but digging deeper whole workflow needed an update to get ready for further issues. During this overhaul the following changes were made:

- Generalized Arnold Standin workflow to a single loader.
- Support multiple nodes as proxies.
- Support proxies for `pointcache` family.
- Generalized approach to proxies as resources, so they can be the same file format as the original.

This workflow should allow further expansion to utilize operators and eventually USD.

## Discussion
By supporting proxies for the `pointcache` family, we can also get rid of the specialized `proxyAbc` workflow. What do we think of that?
We could also implement/support proxy workflows when it comes to animation and rig, although not sure how much that is used outside of big studios.

## Testing notes:
1. Create and publish Arnold Scene Source or Pointcache instance.
2. Load and render.
3. Repeat with proxy workflow to see proxy in viewport and original at render time.


openpype/hosts/maya/plugins/load/load_arnold_standin.py
```diff
@@ -0,0 +1,217 @@
...
+        )
+
+        # Root group
+        label = "{}:{}".format(namespace, name)
+        root = cmds.group(name=label, empty=True)
+
+        # Set color.
+        project_name = context["project"]["name"]
+        settings = get_project_settings(project_name)
+        colors = settings['maya']['load']['colors']
```
Minkiu, line 58: This could be something like:
```
color = settings['maya']['load']['colors'].get("ass")
```

Or if we know the default colour we could do something like:
```py
r, g, b = settings['maya']['load']['colors'].get(
    "ass",
    (0,0,0) # this should be the default
)
cmds.setAttr(root + ".useOutlinerColor", True)
cmds.setAttr( root + ".outlinerColor", r, g, b)
```

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        instance.data["representations"].append(representation)
+
+        self.log.info(
+            "Extracted instance {} to: {}".format(instance.name, staging_dir)
+        )
+
+        # Extract proxy.
+        kwargs["filename"] = file_path.replace(".ass", "_proxy.ass")
```
Minkiu, line 101: Would it be safer to do a copy of the `kwargs` just do do the `proxy` extraction?

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
```diff
@@ -0,0 +1,168 @@
...
+        filenames = self._extract(
+            instance.data["proxy"], attribute_data, kwargs
+        )
+
+        template_data = copy.deepcopy(instance.data["anatomyData"])
+        template_data.update({"ext": "ass"})
+        templates = instance.context.data["anatomy"].templates["publish"]
+        published_filename_without_extension = StringTemplate(
```
Minkiu, line 109: This could probably be:
```py
published_filename, published_extension = os.path.splitext(StringTemplate(templates["file"]).format(template_data))
```

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
```diff
@@ -0,0 +1,168 @@
...
+            templates["file"]
+        ).format(template_data).replace(".ass", "_proxy")
+        transfers = []
+        for filename in filenames:
+            source = os.path.join(staging_dir, filename)
+            destination = os.path.join(
+                instance.data["resourcesDir"],
+                filename.replace(
```
Minkiu, line 117: And here then:
```
filename, extension = os.path.splitext(filename)
destination = os.path.join(
    instance.data["resourceDir"].
    "{0}{1}".format(published_filename, extension)
```

openpype/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -112,6 +113,48 @@ def process(self, instance):
...
+                    file=path,
+                    startFrame=start,
+                    endFrame=end,
+                    **options
+                )
+
+        template_data = copy.deepcopy(instance.data["anatomyData"])
+        template_data.update({"ext": "abc"})
+        templates = instance.context.data["anatomy"].templates["publish"]
+        published_filename_without_extension = StringTemplate(
```
Minkiu, line 140: Same as for my commnets in `extract_arnold_scene_source.py`

openpype/hosts/maya/plugins/load/load_arnold_standin.py
```diff
@@ -0,0 +1,217 @@
...
+    load,
+    get_representation_path
+)
+from openpype.hosts.maya.api.lib import (
+    unique_namespace, get_attribute_input, maintained_selection
+)
+from openpype.hosts.maya.api.pipeline import containerise
+
+
+def is_sequence(files):
```
Minkiu, line 18: I wonder if this would be a good candidate to have in `openpype.lib`

Minkiu (commented):
Haven't tested the changes themselves, and I would be keen in adding more Docstrings, even if they seem redundant.

openpype/hosts/maya/plugins/load/load_arnold_standin.py
```diff
@@ -0,0 +1,217 @@
...
+            # Set the standin filepath
+            path, operator = self._setup_proxy(
+                standinShape, self.fname, namespace
+            )
+            cmds.setAttr(standinShape + ".dso", path, type="string")
+            sequence = is_sequence(os.listdir(os.path.dirname(self.fname)))
+            cmds.setAttr(standinShape + ".useFrameExtension", sequence)
+
+        nodes = [root, standin]
+        if operator is not None:
```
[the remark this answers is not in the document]
tokejepsen: Not sure. Its a variable here holding the name of the string replacement operator. Do you want it changed?

openpype/hosts/maya/plugins/load/load_arnold_standin.py
```diff
@@ -0,0 +1,217 @@
...
+        return containerise(
+            name=name,
+            namespace=namespace,
+            nodes=nodes,
+            context=context,
+            loader=self.__class__.__name__)
+
+    def get_next_free_multi_index(self, attr_name):
+        """Find the next unconnected multi index at the input attribute."""
+
```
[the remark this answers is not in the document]
tokejepsen: Sure, we can do that.

openpype/hosts/maya/plugins/load/load_arnold_standin.py
```diff
@@ -0,0 +1,217 @@
...
+            connection_info = cmds.connectionInfo(
+                "{}[{}]".format(attr_name, start_index),
+                sourceFromDestination=True
+            )
+            if len(connection_info or []) == 0:
+                return start_index
+            start_index += 1
+
+    def _get_proxy_path(self, path):
+        basename_split = os.path.basename(path).split(".")
```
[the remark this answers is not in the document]
tokejepsen: Not quite since the frame padding are using `.` as separator. For example: `assMain.0001.ass`.
Got any suggestion for this?

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
```diff
@@ -0,0 +1,174 @@
...
+        self.log.info(
+            "Extracted instance {} to: {}".format(instance.name, staging_dir)
+        )
+
+        # Extract proxy.
+        if not instance.data.get("proxy", []):
+            return
+
+        kwargs["filename"] = file_path.replace(".ass", "_proxy.ass")
+        filenames = self._extract(
```
moonyuet, line 102: Just curious: Is there any reason not to include the second publishing into the data of representations?

moonyuet:
I have tested loader and extractor, so far so good. Apart from the question about the second extraction in the extractor, I think it looks ok but it somehow breaks another loader due to the adjustment
image

This is something I saw when I run **reference .ass standin with proxy**
image

image

openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
```diff
@@ -1,35 +1,31 @@
...
-            if "content_SET" in objset:
-                instance.data['setMembers'] = members
-                self.log.debug('content members: {}'.format(members))
-            elif objset.startswith("proxy_SET"):
-                if len(members) != 1:
-                    msg = "You have multiple proxy meshes, please only use one"
-                    raise KnownPublishError(msg)
-                instance.data['proxy'] = members
-                self.log.debug('proxy members: {}'.format(members))
+            if objset.endswith("content_SET"):
```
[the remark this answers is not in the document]
tokejepsen: What logic is changing?

openpype/hosts/maya/plugins/publish/collect_pointcache.py
```diff
@@ -12,3 +14,31 @@ class CollectPointcache(pyblish.api.InstancePlugin):
...
+
+        proxy_set = None
+        for node in instance.data["setMembers"]:
+            if cmds.nodeType(node) != "objectSet":
+                continue
+            members = cmds.sets(node, query=True)
+            if members is None:
+                self.log.warning("Skipped empty objectset: \"%s\" " % node)
+                continue
+            if node.endswith("proxy_SET"):
```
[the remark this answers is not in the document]
tokejepsen: Not that I can think of. If you have a case in mind, welcome to reopen this conversation.

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
tokejepsen: Not sure why we would need to?

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
tokejepsen: Not really since we are doing an extension replacement `.replace(".ass", "_proxy")`.
Welcome to reopen the conversation if you have a different suggestion.

openpype/hosts/maya/plugins/load/load_arnold_standin.py
tokejepsen: Think it might get lost and by that I mean only used for this and no one will use it other places. Think there was a conversation with @BigRoy about needing some unified approach to this.

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
tokejepsen: I wanted to keep the original and proxy together. Think if I add them as representation, there will be conflict for destination file names since they share the same extension?

tokejepsen:
> I have tested loader and extractor, so far so good. Apart from the question about the second extraction in the extractor, I think it looks ok but it somehow breaks another loader due to the adjustment image
> 
> This is something I saw when I run **reference .ass standin with proxy** image
> 
> image

Not sure how you get that Loader since I've removing it with this commit?
https://github.com/ynput/OpenPype/pull/4449/commits/9733b07f6dd383a63c00d51312bf700a475aa5d8

moonyuet:
> I have tested loader and extractor, so far so good. Apart from the question about the second extraction in the extractor, I think it looks ok but it somehow breaks another loader due to the adjustment image
> 
> This is something I saw when I run **reference .ass standin with proxy** image
> 
> image



> > This is something I saw when I run **reference .ass standin with proxy**
> 
> @moonyuet Not sure how you get that Loader since I've removing it with this commit? [9733b07](https://github.com/ynput/OpenPype/commit/9733b07f6dd383a63c00d51312bf700a475aa5d8)

Sorry, maybe I mixed with other forks. my bad! Let me test again!

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
moonyuet: Maybe you can try this pseudocode
`           new_rep = {
            "name": "ass",
            "ext": "ass",
            "files": filenames,
            "stagingDir": staging_dir,
            "outputName": "proxy"
        }
`

moonyuet: Maybe you can try this pseduocode...
```
representation = {
            "name": "ass",
            "ext": "ass",
            "files": the_proxy_filepaths,
            "stagingDir": staging_dir,
            "outputName": "proxy"
}
```

moonyuet:
I have tested it again, the loader works fine. Sorry for the inconvenience!
image

The ass will be published successfully. Not sure if the ass with "_proxy" needed to be published. If yes,  the ass with "_proxy"  is not there yet. 
image

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
moonyuet: Or something like this
```
# .... first extraction

    representation = {
            "name": "ass",
            "ext": "ass",
            "files": filenames if len(filenames) > 1 else filenames[0],
            "stagingDir": staging_dir,
            "frameStart": kwargs["startFrame"]
    }

# ... second extraction
proxy_representation = {
            "name": "ass",
            "ext": "ass",
            "files": the_proxy_filepaths,
            "stagingDir": staging_dir,
            "outputName": "proxy"
}
instance.data["representations"].append(proxy_representation)
```

tokejepsen:
> The ass will be published successfully. Not sure if the ass with "_proxy" needed to be published. If yes, the ass with "_proxy" is not there yet.

@moonyuet the proxy version goes into the `resources` folder atm. Did you not get it publish and loaded?

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
tokejepsen: Nice one! Tested and it works. Will update now.

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
```diff
@@ -0,0 +1,160 @@
...
+        )
+
+        representation = {
+            "name": "proxy",
+            "ext": "ass",
+            "files": filenames if len(filenames) > 1 else filenames[0],
+            "stagingDir": staging_dir,
+            "frameStart": kwargs["startFrame"],
+            "outputName": "proxy"
+        }
```
moonyuet, line 111: The same name and ext in the second representation won't affect the publish of the first rep. And this may help the loader to load the published content and proxy. 
```
representation = {
            "name": "ass",
            "ext": "ass",
            "files": the_proxy_filepaths,
            "stagingDir": staging_dir,
            "outputName": "proxy"
}
```

tokejepsen: Having the same `name` means the Loader will think that the proxy representation can be loaded, so we need to differentiate between the original and proxy.

moonyuet: > Having the same `name` means the Loader will think that the proxy representation can be loaded, so we need to differentiate between the original and proxy.

oh i see. then it makes sense to have different `name`.

moonyuet (approved):

openpype/hosts/maya/plugins/load/load_arnold_standin.py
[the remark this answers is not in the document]
tokejepsen: Its declared here; https://github.com/tokejepsen/pype/blob/enhancement/OP-4865_static-ass-file-option/openpype/hosts/maya/plugins/load/load_arnold_standin.py#L76

openpype/hosts/maya/plugins/load/load_arnold_standin.py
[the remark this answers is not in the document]
Minkiu: Ah, you are right, i didn't though about that, in that instance we probably should be relying in something like `clique` or `pyseq`, it's fine as is then.

openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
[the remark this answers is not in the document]
Minkiu: We used to check if "content_SET" was in the string (in any position) and now we just check if it ends with that.

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
Minkiu: We are doing a second extract within the extract, if I understood correctly, if we were to be using kwargs somewhere else expecting the non-proxy values, by now changing it here we would be affecting the later stage.

Minkiu (approved):

openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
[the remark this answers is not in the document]
tokejepsen: Ahh, yeah. This is because if you want to have multiple ASS instances in the scene, we need to prefix the objectsets else they will end up with a number in the name cause of the duplicate names in the scene. Cant have objectsets with the same names even if they are nested.

antirotor (approved):
It works fine, just little style comment to accept or dismiss :)

PR #4450, opened 2023-02-10
Fix: Duplicate file to copy creates already exists error

Tilix4:
## Brief description
It may happen hero version tuple to copy is registered twice, in case of an Extractor listed by mistake for two families for example.
This is not clean indeed, but it shouldn't stop the process for a file already exists copy error.

## Description
By changing the paths mapping to `set` it ensures no duplicates and the already exists copy error won't be raised.

## Testing notes:
1. Add a family with hero version to two different extractors
2. Publish

### DIFF
--- a/openpype/plugins/publish/integrate_hero_version.py
+++ b/openpype/plugins/publish/integrate_hero_version.py
@@ -167,7 +167,7 @@ class IntegrateHeroVersion(pyblish.api.InstancePlugin):
         instance_publish_dir = os.path.normpath(
             instance.data["publishDir"]
         )
-        other_file_paths_mapping = []
+        other_file_paths_mapping = set()
         for file_path in all_copied_files:
             # Check if it is from publishDir
             if not file_path.startswith(instance_publish_dir):
@@ -179,7 +179,7 @@ class IntegrateHeroVersion(pyblish.api.InstancePlugin):
             dst_filepath = file_path.replace(
                 instance_publish_dir, hero_publish_dir
             )
-            other_file_paths_mapping.append((file_path, dst_filepath))
+            other_file_paths_mapping.add((file_path, dst_filepath))
 
         # Current version
         old_version, old_repres = self.current_hero_ents(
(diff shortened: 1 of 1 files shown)


tokejepsen:
@Tilix4 could you elaborate on the testing steps? For example are there already two families we can use or do we need to create some extra plugins to test?

Tilix4:
> @Tilix4 could you elaborate on the testing steps? For example are there already two families we can use or do we need to create some extra plugins to test?

@tokejepsen, your comment makes me reconsider things. I have a question before elaborating testing steps: Is a family supposed to have only a single `Extractor` at any time in OP philosophy?

kalisp:
I dont think it is required, there are some extractors that are doing work on top of previous extractor (I for example have `extract_local_render` for `render` family in AE, same instance will be pushed through `ExtractReview`, `ExtractThumbnail` etc.)

Tilix4:
Okay, therefore @tokejepsen, you can create a new plugin, inheriting from `ExtractBlend` in blender, and at it for family that is listed for `ExtractBlend` too. This way the representation creation process will be run twice.

That being said, I start to think my proposal is not good, because it'll allow something wrong and make it pass silently. Maybe we must raise an error if a file for copy is listed twice, and say that maybe two similar representations have been created and this is not right.

My main point is: the file copy already exists error is confusing and should be: 
1. More accurate, or
2. Pass without any problem (current code proposal)

kalisp:
I don't this is correct approach. 
Is there the issue that 2 extractors created completely SAME representation? In that case, this would be fine, but what if 2 extractors are creating slightly different representation (but with same name). In that case this would be wrong.

The problem is that this is probably already happening in regular integrate and `file_transaction.add` (which btw should be eventually used in `integrate_hero_version` too.)

Tilix4:
> I don't this is correct approach. Is there the issue that 2 extractors created completely SAME representation? In that case, this would be fine, but what if 2 extractors are creating slightly different representation (but with same name). In that case this would be wrong.

@kalisp The thing is, if two exactly same representations are created (the case I encountered), twice the same copy request is made, and when the first one is correctly executed, the second one raises the error, which is strange IMO, because it doesn't cause any trouble to OP. That's why I proposed this fix by having a `set` to avoid duplicates.

kalisp:
I am closing this for now as we discussed that this is silently hiding abnormal situation and it should rather fail.
Please free to comment or reopen if you would have any objections.

PR #4452, opened 2023-02-10
General: Fix original basename frame issues

iLLiCiTiT:
## Brief description
Treat `{originalBasename}` in different way then standard files processing. In case template should use `{originalBasename}` the transfers will use them as they are without any changes or handling of frames.

## Description
Frames handling is problematic with original basename because their padding can't be defined to match padding in source filenames. Also it limits the usage of functionality to "must have frame at end of fiename". This is proposal how that could be solved by simply ignoring frame handling and using filenames as are on representation. First frame is still stored to representation context but is not used in formatting part. This way we don't have to care about padding of frames at all.

### Additional information
This is a proposal, I could miss some other issues related to the feature.

## Testing notes:
1. Add templates to use `{originalBasename}`
2. Set to use the template for your usecase (could be e.g. for `render` family in `traypublisher`)
3. Publish something which should use the template
   - single file
   - sequence - different paddings
4. Validate if the filenames are correct and did not change
5. Validate if something else did not break

### DIFF
--- a/openpype/plugins/publish/integrate.py
+++ b/openpype/plugins/publish/integrate.py
@@ -506,6 +506,43 @@ class IntegrateAsset(pyblish.api.InstancePlugin):
 
         return version_doc
 
+    def _validate_repre_files(self, files, is_sequence_representation):
+        """Validate representation files before transfer preparation.
+
+        Check if files contain only filenames instead of full paths and check
+        if sequence don't contain more than one sequence or has remainders.
+
+        Args:
+            files (Union[str, List[str]]): Files from representation.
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
So the first issue is that loaders are using `{frame}` in template to fill frame padding on load of sequence. We'll probably have to add the frame back to template, but in that case it could be without padding in template and we would fill the padding in code.

To change the loaders logic we lack metadata on representation (like frame start/end). That can be guessed from `files` but there is no information about what files are what. There can be resources as sequence.

jakubjezek001 (approved):

PR #4453, opened 2023-02-10
Hiero/Nuke: originalBasename editorial publishing and loading

jakubjezek001:
## Brief description
Publishing and loading `originalBasename` is working as expected

## Description
Frame-ranges on version document is now correctly defined to fit original media frame range which is published. It means loading is now correctly identifying frame start and end on clip loader in Nuke.  

## Additional info
Loading is only fixed for Nuke but other DCC's should also be treated. Perhaps some abstracted implementation in wider scale should be done.

## Testing notes:
1. Set to use the template for your use case. `plate` family with `source` template.
Animation
2. Open Hiero and add a clip, create to publishable and publish 
3. Open Nuke Load published clip and all should play as it suppose to

### DIFF
--- a/openpype/hosts/nuke/plugins/load/load_clip.py
+++ b/openpype/hosts/nuke/plugins/load/load_clip.py
@@ -222,13 +222,12 @@ class LoadClip(plugin.NukeLoader):
         representation = deepcopy(representation)
         context = representation["context"]
         template = representation["data"]["template"]
-        frame = context["frame"]
-        hashed_frame = "#" * len(str(frame))
-
         if (
             "{originalBasename}" in template
             and "frame" in context
         ):
+            frame = context["frame"]
+            hashed_frame = "#" * len(str(frame))
             origin_basename = context["originalBasename"]
             context["originalBasename"] = origin_basename.replace(
                 frame, hashed_frame
--- a/openpype/hosts/nuke/plugins/load/load_clip.py
+++ b/openpype/hosts/nuke/plugins/load/load_clip.py
@@ -220,8 +220,21 @@ class LoadClip(plugin.NukeLoader):
             dict: altered representation data
         """
         representation = deepcopy(representation)
-        frame = representation["context"]["frame"]
-        representation["context"]["frame"] = "#" * len(str(frame))
+        context = representation["context"]
+        template = representation["data"]["template"]
+        frame = context["frame"]
+        hashed_frame = "#" * len(str(frame))
+
+        if (
+            "{originalBasename}" in template
+            and "frame" in context
+        ):
+            origin_basename = context["originalBasename"]
+            context["originalBasename"] = origin_basename.replace(
+                frame, hashed_frame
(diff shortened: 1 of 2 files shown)


openpype/plugins/publish/collect_otio_subset_resources.py
```diff
@@ -225,3 +237,26 @@ def _create_representation(self, start, end, **kwargs):
         if kwargs.get("trim") is True:
             representation_data["tags"] = ["trim"]
         return representation_data
+
+    def get_template_name(self, instance):
```
iLLiCiTiT, line 241: This is theorectical question which probably does not have solution at this moment:

Template should be calculated based on instance data, not on context data. For example during TrayPublisher publishing `context.data['anatomyData']` contains only project data (because there is not a global asset or task) -> and I know that this is not plugin which runs on traypublisher but still.

Additional theorectical question:
Is this is needed? Why is neededed to change of frames on version data here? I don't know where the files exactly are comming from but they could come from extraction? Could frame fill for version data rather happen during extraction?

openpype/hosts/nuke/plugins/load/load_clip.py
```diff
@@ -220,8 +220,21 @@ def _representation_with_hash_in_frame(self, representation):
...
-        representation["context"]["frame"] = "#" * len(str(frame))
+        context = representation["context"]
+        template = representation["data"]["template"]
+        frame = context["frame"]
+        hashed_frame = "#" * len(str(frame))
+
+        if (
+            "{originalBasename}" in template
+            and "frame" in context
+        ):
```
iLLiCiTiT, line 231: This would crash on single frame representation
```suggestion
        if (
            "{originalBasename}" in template
            and "frame" in context
        ):
            frame = context["frame"]
            hashed_frame = "#" * len(str(frame))
```

openpype/plugins/publish/collect_otio_subset_resources.py
jakubjezek001: I get your point. I would like to have rather instance anatomy data then context but the order is wrong here. I will try to shift order of this plugin after `collect_anatomy_instance_data.py` and see if it works. 

Regarding necesity - unfortunatelly there is not other way to find out if I should use media framerange for version or workfile framerange. This should have been done based on the condition if `"{originalBasename}"` is in template path and only way to find out is to get template set for the host/family presets.

iLLiCiTiT (approved):
Code looks ok

jrsndl (changes requested):
I have a problem loading published plate to Nuke. It looks like loadClip adds the hashes to read node path twice.

The plate name doesn't seem to make a difference. I have the same behavior with source plate named renderTest_v003.####.exr and boxed_mops_v003.Redshift_ROP1.####.exr

Plate published in Hiero 12.2v3, loading tested in Nuke 12.2v3 and 13.2v4

PublishA
PublishB

PR #4455, opened 2023-02-13
Option to start versioning from 0

maxpareschi:
## Brief description
Enables v000 as first version

## Description
Modified all bits i could find on version to allow for 0 as first version. workfile gui also shows 0 as an option.

## Additional info
Ftrack had problems with version 0 in its API, solution was to cast 0 as string before committing the assetversion to db. Seem to work as intended.

## Testing notes:
1. Open Nuke
2. see that the first version of workfile can be set at 0 or is already built as 0 if you have workfile builder active
3. Publish and see v000 on ftrack.
Tested on padding 3 and 4

## Tested hosts
- [x] Nuke (render)
- [x] Maya (pointcache, review)
- [ ] Houdini
- [ ] Fusion
- [ ] Nukestudio
- [ ] Blender
- [ ] 3dsmax
- [ ] Flame
- [ ] Resolve
- [ ] Harmony
- [ ] Tvpaint
- [ ] Photoshop
- [ ] AE
- [ ] Celaction
- [ ] Unreal
- [x] Tray Publisher

For now nuke, traypublisher and maya pass testing (limited to info in parenthesis)


maxpareschi:
does not work in maya, it does not increment version for publishing on collect. will try something. for now keep it blocked

tokejepsen (changes requested):
This will need some setting that is off by default since its quite a fundamental change.

maxpareschi:
> This will need extensive testing is all hosts so I would suggest making a list in the description we can tick off as we test the hosts. Something like this;
> 
> * [ ]  Maya
> * [x]  Nuke

Done!
Also version incrementing seem to work now.
Say, would we probably add some setting and the ability to actually choose the initial version? where would we put that? general? templates?

tokejepsen:
> Say, would we probably add some setting and the ability to actually choose the initial version? where would we put that? general? templates? Maybe family and host filters as well?

Yeah, I would say `project_settings/global/tools/Workfiles` would be a good place. Maybe look at how `project_settings/global/tools/Workfiles/open_workfile_tool_on_startup` is implemented for inspiration.

image

tokejepsen:
Closing this in favour of #5262

PR #4456, opened 2023-02-13
Work with a copy of Fusion preferences on launch

movalex:
## Brief description
Fusion preferences will be copied to the predefined `~/.openpype/hosts/fusion/prefs` folder (or any other folder set in system settings) on launch.

## Description
The idea is to create a copy of existing Fusion preferences, maintaining all path maps and interface tweaks, adding an OpenPype menu only for the instance launched with the OP Launcher.
By default the setting is turned off, so no file copy process is performed without asking. Instead the clean Fusion preferences file is created by the host app in the `~/.openpype/hosts/fusion/prefs` folder, then some modifications applied to this file: 
forced Python3, forced English interface, Openpype specific pathmaps set up).
If the `copy_prefs` is toggled, a copy of current Fusion preferences is created in the predefined location, then they are altered the same way as above. The operation is run only once, on the first launch, unless the `force_sync [Always sync preferences on launch]` is toggled.
English interface is forced because the `FUSION16_PROFILE_DIR` environment variable is not read otherwise (seems to be a Fusion bug).


openpype/hosts/fusion/hooks/pre_fusion_setup.py
```diff
@@ -16,19 +20,80 @@ class FusionPrelaunch(PreLaunchHook):
...
+    def get_copy_fusion_prefs_settings(self):
+        """Get copy prefserences options from the global application settings"""
+        copy_status = copy_path = None
+        try:
+            copy_status, copy_path, force_sync = get_system_settings()["applications"]["fusion"][
+            "copy_fusion_settings"].values()            
+        except ValueError:
+            self.log.error('Copy prefs settings not found')
+        finally:
+            return copy_status, Path(copy_path).expanduser(), force_sync
```
iLLiCiTiT, line 63: Try won't save you from crashing on `None` value passed to `Path`. Also explicit access to settings keys is preferred.
```suggestion
        copy_fusion_settings = (
            self.data
            ["system_settings"]
            ["applications"]
            ["fusion"]
            .get("copy_fusion_settings", {})
        )
        if not copy_fusion_settings:
            self.log.error("Copy prefs settings not found")
        copy_status = copy_fusion_settings.get("copy_status", False)
        force_sync = copy_fusion_settings.get("force_sync", False)
        copy_path = copy_fusion_settings.get("copy_path") or None
        if copy_path:
            copy_path = Path(copy_path).expanduser()
        return copy_status, copy_path, force_sync
```

openpype/hosts/fusion/hooks/pre_fusion_setup.py
```diff
@@ -16,19 +20,80 @@ class FusionPrelaunch(PreLaunchHook):
 
     """
     app_groups = ["fusion"]
+    PROFILE_NUMBER = 16
```
iLLiCiTiT, line 23: What is this number?

openpype/hosts/fusion/hooks/pre_fusion_setup.py
```diff
@@ -48,14 +113,21 @@ def execute(self):
...
+        # Add custom Fusion Master Prefs and the temporary profile directory variables
+        # to customize Fusion to define where it can read custom scripts and tools from
         self.log.info(f"Setting OPENPYPE_FUSION: {FUSION_HOST_DIR}")
         self.launch_context.env["OPENPYPE_FUSION"] = FUSION_HOST_DIR
 
-        pref_var = "FUSION16_MasterPrefs"   # used by Fusion 16, 17 and 18
-        prefs = os.path.join(FUSION_HOST_DIR, "deploy", "fusion_shared.prefs")
-        self.log.info(f"Setting {pref_var}: {prefs}")
-        self.launch_context.env[pref_var] = prefs
+        copy_status, openpype_fusion_profile_dir, force_sync = self.get_copy_fusion_prefs_settings()
```
iLLiCiTiT, line 123: Unhandled case when `openpype_fusion_profile_dir` is None. Also would make the variable shorter `fusion_profile_dir`?

openpype/settings/defaults/system_settings/applications.json
```diff
@@ -835,6 +835,11 @@
                 "linux": "/opt/Python/3.6/bin"
             }
         },
+        "copy_fusion_settings": {
+            "copy_prefs": false,
+            "prefs_path": "~/.openpype/hosts/fusion/prefs",
```
iLLiCiTiT, line 840: This should be maybe empty by default.
```suggestion
            "prefs_path": "",
```

openpype/hosts/fusion/hooks/pre_fusion_setup.py
```diff
@@ -48,14 +113,21 @@ def execute(self):
...
+        if copy_status:
+            prefs_source = self.get_profile_source()
+            self.copy_existing_prefs(prefs_source, openpype_fusion_profile_dir, force_sync)
+        fusion_profile_dir_variable = f"FUSION{self.PROFILE_NUMBER}_PROFILE_DIR"
+        master_prefs_variable = f"FUSION{self.PROFILE_NUMBER}_MasterPrefs"
+        openpype_master_prefs = Path(FUSION_HOST_DIR, "deploy", "fusion_shared.prefs")
```
iLLiCiTiT, line 129: There is no reason for prefix `openpype_` -> `master_prefs` should be fine

EmberLightVFX:
~/ doesn't point to anywhere on Windows.
Even if `Path().expanduser()` would resolve it it would confuse me as a windows user a lot thinking the feature would break even if it in the end would work.

EmberLightVFX:
I also think it would be better to move this feature from the System tab over to the Project tab.
That way we could build upon this PR to be able to include specific project paths that also would be added
Something like Nukes Gizmo menu:
image
There we could have each default prefs folder set already and under options we could add custom path-maps for each project.
That feature could come later on and not needed for this first PR 🙂
but could be good to just have everything in place for that kind of expansion.

openpype/hosts/fusion/hooks/pre_fusion_setup.py
movalex: This is an odd thing. I guess BlackmagicDesign devs intended to make the `FUSION_PROFILE_DIR` to be version specific, so from the v16 they made it to be `FUSION16_PROFILE_DIR`. But then they abandoned this, and despite it is already version 18, still `FUSION16_PROFILE_DIR` is used. I added the variable in case they would break this in future version (they really could), so we could easily change the version or disable it.

openpype/hosts/fusion/api/lib.py
```diff
@@ -302,10 +302,25 @@ def get_frame_path(path):
     return filename, padding, ext
 
 
-def get_current_comp():
-    """Hack to get current comp in this session"""
+def get_fusion():
```
iLLiCiTiT, line 305: This looks like a function from Nuclear power plant :)
```suggestion
def get_fusion_module():
```

openpype/hosts/fusion/api/lib.py
```diff
@@ -302,10 +302,25 @@ def get_frame_path(path):
...
+    """Get current Fusion instance"""
     fusion = getattr(sys.modules["__main__"], "fusion", None)
-    return fusion.CurrentComp if fusion else None
+    return fusion
+
+
+def get_current_comp():
+    """Get current comp in this session"""
+    fusion = get_fusion()
+    if fusion:
```
iLLiCiTiT, line 314: Who knows what is implemented under `__bool__`...
```suggestion
    if fusion is not None:
```

openpype/hosts/fusion/api/lib.py
```diff
@@ -302,10 +302,25 @@ def get_frame_path(path):
...
+        comp = fusion.CurrentComp
+        return comp
+
+
+def get_comp_filename():
+    """Get comp's Filename"""
+    comp = get_current_comp()
+    if comp:
```
iLLiCiTiT, line 322:
```suggestion
    if comp is not NOne:
```

openpype/hosts/fusion/api/lib.py
```diff
@@ -302,10 +302,25 @@ def get_frame_path(path):
...
+
+def get_current_comp():
+    """Get current comp in this session"""
+    fusion = get_fusion()
+    if fusion:
+        comp = fusion.CurrentComp
+        return comp
+
+
+def get_comp_filename():
```
iLLiCiTiT, line 319: Why was this function created? The logic could stay in `current_file` as it's only place where is used.

openpype/hosts/fusion/api/lib.py
EmberLightVFX: I guess you meant to write `if comp is not None:` and not `NOne`

openpype/hosts/fusion/hooks/pre_fusion_setup.py
iLLiCiTiT: Maybe add small comment above it so any future devs know "why the hell"...

openpype/settings/defaults/system_settings/applications.json
movalex: I thought about this string as a default value that would work out of the box on any host. Say, first setup, check the `Copy` checkbox, boom, now you have working preferences already. But probably I should work more to the @EmberLightVFX suggested way, and set the prefs location project-wise...

EmberLightVFX: It could place the preference file in a folder named after the project by default. That way nothing will be overwritten by different project.

openpype/hosts/fusion/api/lib.py
movalex: I used `get_comp_filename` [here](https://github.com/movalex/OpenPype/blob/807fd736fc7ce26b6e2553cfa9c1c53d7e13360f/openpype/hosts/fusion/api/workio.py#L28),  to maintain the .lib functions logic.

openpype/settings/entities/schemas/system_schema/host_settings/schema_fusion.json
```diff
@@ -19,6 +19,36 @@
             "label": "Environment",
             "type": "raw-json"
         },
+        {
+            "type": "splitter"
+        },
+        {
+            "type": "dict",
```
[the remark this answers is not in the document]
movalex: Do you think move it to the project setting only or leave it in system too? So if there's no per project project override, system would be used?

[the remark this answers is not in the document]
EmberLightVFX: > Do you think move it to the project setting only or leave it in system too? So if there's no per project project override, system would be used?

The project tab have settings that gets used by default by all projects if not changed so that would apply on all projects automatically. I don't see any need to have it in the system tabs too.

openpype/hosts/fusion/api/lib.py
iLLiCiTiT: That's my point. The function `get_comp_filename` is used only in `current_file` (which is requirement) so why to have 2 functions doing the same thing?

movalex: Yeah, I see. Probably a misunderstanding here. I though the `lib` is a functions hub, and the `workio` is one of the modules that is using it. So while the func is used only once now, it may be used later by another modules. But I can revert that for sure.

iLLiCiTiT: Well it's changing based on needs. Workio is/was categorized "lib hub" for workfiles. The lib.py is more likely uncategorized, or contains functions that are used on multiple places and there is no way how to put it to a better place to avoid cross imports. It matters on usage and host implementation (and developer). At this moment `get_comp_filename` is just a duplicate.

EmberLightVFX:
I wanted to try the PR out this weekend but I had some problems with my OP instance the few hours I had on me so I couldn't fully try the feature and/or write some own code.
Any how I do have some thoughts:

* `Copy Fusion settings on launch` should be renamed to `Copy local Fusion profile on launch` to make it more clear what it's actually doing. I think we should change settings to profile as profile is the correct Fusion terminology.
* Do we want to place the newly created profiles at `~/.openpype`? Does anything else in OPs ecosystem do that? For me it feels more logical if the default location is where Fusion stores its profiles. As it's still under a user-folder (%appdata% on Windows) there won't be any problems for multi-user computers doing this.
* On that node, I think `Local prefs directory` should be renamed to `Local profiles directory` as profile is the correct Fusion terminology. I can understand using `prefs` in the name as the pref file itself ends with `.prefs` but later on it would be nice if a project could for example set a custom `VidModes.def` that also lives in the profiles folder.
* I feel we need to have the default Linux, Mac and Win profile-dir location so it won't be confusing for any user, at least not for any windows user like myself. Especially if we will use the fusion profiles folder for the profile location.
* The location should be able to use template keys (https://openpype.io/docs/admin_settings_project_anatomy#available-template-keys), especially `{project[name]}` so the default path could be `%AppData%\Blackmagic Design\Fusion\Profiles\{project[name]}` so each project can get its own profiles folder and preferences.
* `Force sync preferences` should be renamed to `Resync local preferences on startup` to make it more clear what the option does. Currently for me `Copy Fusion settings on launch` and `Force sync preferences` mean the same without any context.

It's really sad I couldn't try the feature yet as mentioned above as this is the biggest problem Fusion in OP currently have and I love that it's about to finally be fixed!

For a future update I would love to add options to add custom pathmaps directly within OpenPype that those would be added together with the OpenPype pathmap in Fusion so all of those things would be setup easily by a TD :)

movalex:
I have also found one more inconsistency here. The local OP folder if used by the launcher disregarding if the `Copy local Fusion profile on launch` is enabled or not. If it is disabled, the empty profile will be created in the default folder anyway. So I guess we don't need the toggle there. So the settings may look like this:
image
As for the default folder location, I can't see how moving the copy of the profiles to the default profiles folder could be easily done for the multiplatform users. On Windows the profile is stored in `%AppData%`, on MacOS it is in the `~/Library/Application Support`, on Linux, I guess, it is in the `~/.fusion` folder. My intention was to make the default folder to work on any platform, so the user could just start Fusion with Launcher and see the OpenPype menu right away. Later they may specify any other paths they like. Also it would not be logical to store the profile in the default Fusion Profiles: folder, because it's location may be already overridden by the `FUSION16_PROFILE_DIR` environment variable by any reason. 
> Do we want to place the newly created profiles at ~/.openpype? Does anything else in OPs ecosystem do that?

I think one of the modules used to place stuff to the `~/.pypeclub` folder. I guess that location is somewhat deprecated now. I believe the `~/.openpype` location is pretty tenable. 
I wonder whether there should be any code changes so the `{project[name]}`  names would be supported, or it is parsed by default?

movalex:
> The problem with the current way is if a studio wants to place all the profiles in a specific location 

Right, I agree with that. But I wonder if adding three more paths would make the setup too complex. What if we just store the profile in the project's {root}? It is already set up, and it is both project and platform specific. Thus by specifying the profile location to the {root}/.fusion_profile/{username} we'd just keep the settings simple. And the profile settings will have just a `do copy / force re-sync` toggles. I would just need to know how to access these variables in the code...

EmberLightVFX:
> What if we just store the profile in the project's {root}?

I can go with that :) What does people at OP say about that? Is that an ok way to go for user settings?

> I would just need to know how to access these variables in the code...

Maybe best to as in the developer channel on their discord.

> Or even better, first check if the FUSION16_PROFILE_DIR exists, then create the OP Fusion profile folder there, otherwise do that in the project's folder.

The problem I see with this is that the check would be hidden so the user might expect it to end up in the specified folder in the settings but it ends up somewhere else. Maybe add a checkbox in the settings for that? But then again, that would then maybe mess up how a project is setup, if some have `FUSION16_PROFILE_DIR` setup on their computer and others doesn't.
I do think that `FUSION16_PROFILE_DIR` should be possible in the path, exactly like `%appdata%` is if so if a studio wants to use it they could just enter `FUSION16_PROFILE_DIR` in the path.

> I think we should not specify the {username} for the profile storage, because it is already regulated by the FUSION_PROFILE variable. If it is not set, Default folder is used.

If we don't, won't profiles be overwritten by all the different users? If all the files gets put in the same location.
Stuff like multi monitor setups would be messed up.

movalex:
> Is that an ok way to go for user settings?

I will ask on Discord

> The problem I see with this is that the check would be hidden so the user might expect it to end up in the specified folder in the settings but it ends up somewhere else.

Agree, it is not the best to do something silently, but I tend to log.info everything, including the paths for copying process. The `FUSION16_PROFILE_DIR` is a user variable, thus, if the studio is sophisticated enough to specify this folder for each user, we need to use their logic here. 

> If we don't, won't profiles be overwritten by all the different users? 

Same goes here. I wonder, do you assume there may be multiple OP users working under the same Windows account? I believe this is hardly possible. If the studio or the user had specified the profile directory or the profile name for the current user with the variable, we should respect that. Otherwise we just store the profiles in the project's folder, which, I assume frankly, would be so in 99.6% of the cases. 

> Maybe add a checkbox in the settings for that?

Yes, I think this is a way for that for sure. Like 2 more toggles: `Respect the FUSION16_PROFILE_DIR variable` and `Respect the FUSION16_PROFILE variable`. If only the the first is checked, the profile folder is created in the location specified by the variable, and the folder will be a {project[name]}. If only second is checked, the folder will appear in the project folder, and the parent name will be specified by the `FUSION_PROFILE`. And so on. 

P.S. This is all very boring, but I like it :)

EmberLightVFX:
> Agree, it is not the best to do something silently, but I tend to log.info everything, including the paths for copying process. The FUSION16_PROFILE_DIR is a user variable, thus, if the studio is sophisticated enough to specify this folder for each user, we need to use their logic here.

That's why the studio can place `{FUSION16_PROFILE_DIR}` in the profile path if they want to use it. Nothing will then be hidden from anyone and won't be "hidden" in the logs ether for more simpler users that doesn't check the massive amount of info the logs spit out.

>> If we don't, won't profiles be overwritten by all the different users?

> Same goes here. I wonder, do you assume there may be multiple OP users working under the same Windows account? I believe this is hardly possible.

But if we place the config in the same folder for everyone (as this is a project setting), then it would overwrite the profile for all machines, not just the local one as the projects folder isn't machine-account specific. That's why `{username}` would be needed so each OP user would get its own folder and its own config.
A nice bonus with this is for studios with a bunch of computers where users switching around. What ever computer you end up using you would have your config moved with it :)

> Yes, I think this is a way for that for sure. Like 2 more toggles: Respect the `FUSION16_PROFILE_DIR` variable and Respect the `FUSION16_PROFILE` variable.

I think it's better to just allow the OP admin to use those variables in the the path if they want to use it. Less obscured settings that could potentially confuse people not knowing about `FUSION16_PROFILE_DIR` and `FUSION16_PROFILE` the better.
Also, `FUSION16_PROFILE_DIR` and `FUSION16_PROFILE` would mostly be used within a pipeline and if a studio changes pipeline from X to OpenPype, maybe they just need to adjust to the way OpenPype works as you would need with a bunch of other options in OP.

openpype/settings/defaults/system_settings/applications.json
BigRoy: I'd prefer this in Project Settings instead. We'd likely have this value across all projects but it'd be very nice for a particular project to have very customized Fusion prefs on its own.

The path itself seems like a fine default path too me.

BigRoy:
> I think it's better to just allow the OP admin to use those variables in the the path if they want to use it. Less obscured settings that could potentially confuse people not knowing about FUSION16_PROFILE_DIR and FUSION16_PROFILE the better.
Also, FUSION16_PROFILE_DIR and FUSION16_PROFILE would mostly be used within a pipeline and if a studio changes pipeline from X to OpenPype, maybe they just need to adjust to the way OpenPype works as you would need with a bunch of other options in OP.

I agree with this. Let's just make it so that OpenPype has settings that specify the FUSION16_PROFILE_DIR (and if needed the FUSION16_PROFILE too) unless we think that gets in the way with a lot of things.

But I'd say since OpenPype manages environment variables, etc. too I feel like we shouldn't leave this open to something outside of OpenPype. 

> I can go with that :) What does people at OP say about that? Is that an ok way to go for user settings?

I personally wouldn't want the default to be per project. There's no other OpenPype application that does something along those lines as far as I know, other than maybe Unreal Engine integration which is a beast of its own. I'd set a default that wouldn't be unique per project but just the "OpenPype profile". If a studio wants per project profiles they can customize as they see fit.

EmberLightVFX:
> I personally wouldn't want the default to be per project. There's no other OpenPype application that does something along those lines as far as I know, other than maybe Unreal Engine integration which is a beast of its own. I'd set a default that wouldn't be unique per project but just the "OpenPype profile". If a studio wants per project profiles they can customize as they see fit.

So in short there would be only one OpenPype profile for all projects but if the studio wants to use per project they can just use `{project[name]}` to create one per project? That sounds logical and good.
For us we would need the `{project[name]}` as we use custom pathmaps per project. I can make an addition to this PR later to handle that.

So where do we land on the default profile location? Inside of the projects folder? Inside Fusions default profile folder?
@BigRoy @movalex

BigRoy:
> So in short there would be only one OpenPype profile for all projects but if the studio wants to use per project they can just use {project[name]} to create one per project? That sounds logical and good.
For us we would need the {project[name]} as we use custom pathmaps per project. I can make an addition to this PR later to handle that.

Actually, the user wouldn't even need `{project[name]}` since if we have this setting as a project setting they could even hardcode a project specific location. But yes, I'd expect that if they by default want per project unique/fresh Fusion preferences then it'd be `{project[name]}` or alike.

> So where do we land on the default profile location? Inside of the projects folder? Inside Fusions default profile folder?

To me [the `~/.openpype` one you had here](https://github.com/ynput/OpenPype/pull/4456#issuecomment-1437323044) seemed decent to me. If inside Fusion's default profiles directory which I think is also ok, we just need to make sure it doesn't interfere with the actual fusion default and still be unique to OpenPype, no?

EmberLightVFX:
> If inside Fusion's default profiles directory which I think is also ok, we just need to make sure it doesn't interfere with the actual fusion default and still be unique to OpenPype, no?

As long as we add {user} and/or {project[name]} to the path there shouldn't be any problems, so we don't overwrite the default preference.

> To me [the ~/.openpype one you had here](https://github.com/ynput/OpenPype/pull/4456#issuecomment-1437323044) seemed decent to me. If inside Fusion's default profiles directory which I think is also ok

My vote goes to Fusions default profiles directory, movalex vote to the ~/.openpype... dir and BigRoy is split 😅 
I can accept ~/.openpype as long as there are defaults for Linux, Mac and Windows as the studio I work for uses both Win and Mac and the dirs location won't have the same path for both systems.

BigRoy:
@movalex Do you need anything at this stage to move this forward?

movalex:
> @movalex Do you need anything at this stage to move this forward?

I think I fixed all the notes I've got so far. And it seems to work as expected. I went with default local Fusion prefs location to `~/.openpype` folder, this way the OP menu will appear to the user disregarding whether they pre-modified the settings or not, which I think is a good thing. 
The only part left is to make the copy_path string to comply with the [template keys convention](https://openpype.io/docs/admin_settings_project_anatomy/#available-template-keys). For instance, support the `{root[work]}` as a shortcut to the project folder, and so on. This would be an amazing addition, but also quite a task to do. As @iLLiCiTiT noted it is not possible to inherit the project anatomy functions (of this is how I got this), the template keys are supported only by the Project Anatomy, and not the Project Settings. And if they were, this would slow down the settings response, and they are already not too fast. Really, 10-15 seconds until the OP preferences dialogue opens and then another 10 sec while saving the changes is quite a bit of a wait already.
Probably making a local function that 'just' parses the input and replaces the {root} tag with the data info would be a solution, and I already tried this way, but this could potentially open some hell of the user parsing nuances I'm not quite ready to. 
TL:DR I'm a lazy ass.

BigRoy:
> Really, 10-15 seconds until the OP preferences dialogue opens and then another 10 sec while saving the changes is quite a bit of a wait already.

This delay is mostly UI actually - it's solved in Ayon (OP v4) as far as I know.

> The only part left is to make the copy_path string to comply with the [template keys convention](https://openpype.io/docs/admin_settings_project_anatomy/#available-template-keys). For instance, support the {root[work]} as a shortcut to the project folder, and so on. This would be an amazing addition, but also quite a task to do. As @iLLiCiTiT noted, it is not possible to inherit the project anatomy functions (or this is how I got this).

Since this is being formatted in the pre-launch hook that should be totally fine to accept anatomy formatting. Right @iLLiCiTiT. I recall the reply on the discussion before was more about supporting the formatting en masse on any/all path settings in OpenPype which would be much slower (so much settings!) and I'd say dangerous even because there might be entries where we wouldn't want that to happen at all.

movalex:
> here might be entries where we wouldn't want that to happen at all.
I guess to address that there might be added a special type in the schema, say `"type": "templated_path"` or something?

iLLiCiTiT:
> Since this is being formatted in the pre-launch hook that should be totally fine to accept anatomy formatting. 

It is as @BigRoy said, it's totally fine to do it. The issue is "auto-fill" of settings getters. There are so many potential usecases how the template could be filled that it is just not possible. Some settings should be filled with site roots some without, some are for totally different usage and have nothing to do with anatomy roots or the values are filled on different machine which have different roots. Also for load of settings you would need to get anatomy, somehow go through values and find out if they should be filled and do it -> all of this is massive slowdown.

BTW Since AYON will also settings work in slightly different way.

PR #4457, opened 2023-02-14
Strict Error Checking Default

tokejepsen:
## Brief description
Provide default of strict error checking for instances created prior to PR.

## Testing notes:
1. Create a render instance.
2. Delete the strict error checking attribute to simulate having an instance from before the PR was made.
3. Publish.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/collect_render.py
+++ b/openpype/hosts/maya/plugins/publish/collect_render.py
@@ -319,7 +319,7 @@ class CollectMayaRender(pyblish.api.ContextPlugin):
                     "renderSetupIncludeLights"
                 ),
                 "strict_error_checking": render_instance.data.get(
-                    "strict_error_checking", False
+                    "strict_error_checking", True
                 )
             }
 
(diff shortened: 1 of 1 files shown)


antirotor (approved):

Minkiu (approved):

kalisp (changes requested):
Actually True should be default. (That is DL default and that was triggered before the change.)

openpype/hosts/maya/plugins/publish/collect_render.py
```diff
@@ -320,7 +320,8 @@ def process(self, context):
                     "renderSetupIncludeLights"
                 ),
                 "strict_error_checking": render_instance.data.get(
-                    "strict_error_checking")
+                    "strict_error_checking", False
```
iLLiCiTiT, line 323:
```suggestion
                    "strict_error_checking", True
```

tokejepsen:
> Actually True should be default. (That is DL default and that was triggered before the change.)

Done

kalisp (approved):

antirotor (approved):

PR #4459, opened 2023-02-14
General: Don't use mongo call in load utils

iLLiCiTiT:
## Brief description
Replaced forgotten usage of `dbcon.parenthood` with function `get_representation_parents`.

## Testing notes:
Don't know if there is a way how to test this change. Technically it should never get to that point if the files from representation are available on the machine.


antirotor (approved):
seems like it's working

PR #4460, opened 2023-02-14
Maya: Cant assign shaders to the ass file - OP-4859 

### LINKED ISSUE #4535: Maya: Arnold Scene Source (ASS) publish fails
Running ver. 3.15.2. Nightly2

When trying to publish ```Arnold Standin``` OP gives validation errors:

image

So unable to publish anything as .ASS now

```

Instance: 
assMain
Message: 
'NoneType' object is not iterable
Line: 
39
Traceback: 
  File "D:\REPO\OP_3\.venv\lib\site-packages\pyblish\plugin.py", line 522, in __explicit_process
    runner(*args)
  File "D:\REPO\OP_3\openpype\hosts\maya\plugins\publish\validate_arnold_scene_source.py", line 48, in process
  File "D:\REPO\OP_3\openpype\hosts\maya\plugins\publish\validate_arnold_scene_source.py", line 39, in _get_nodes_data
TypeError: 'NoneType' object is not iterable

```

[cuID:[OP-5023](https://app.clickup.com/t/861me1da4)]

tokejepsen:
## Brief description
Support AiStandIn nodes for look assignment.

## Description
Using operators we assign shaders and attribute/parameters to nodes within standins. Initially there is only support for a limited mount of attributes but we can add support as needed;

```
primaryVisibility
castsShadows
receiveShadows
aiSelfShadows
aiOpaque
aiMatte
aiVisibleInDiffuseTransmission
aiVisibleInSpecularTransmission
aiVisibleInVolume
aiVisibleInDiffuseReflection
aiVisibleInSpecularReflection
aiSubdivUvSmoothing
aiDispHeight
aiDispPadding
aiDispZeroValue
aiStepSize
aiVolumePadding
aiSubdivType
aiSubdivIterations
```

## Testing notes
1. Load an alembic and lookdev, plus publish the look.
2. Load the same alembic and publish as Arnold Scene Source.
3. Load the publish Arnold Scene Source and assign the publish look through the LookManager.

Resolves #4535


tokejepsen:
Look at #4535

tokejepsen:
@LiborBatek give this another go. Think I've fixed #4535 now.

openpype/hosts/maya/tools/mayalookassigner/arnold_standin.py
```diff
@@ -0,0 +1,279 @@
...
+                    # https://arnoldsupport.com/2018/11/21/backdoor-setting-visibility/
+                    mapping = {
+                        "primaryVisibility": 1,  # Camera
+                        "castsShadows": 2,  # Shadow
+                        "aiVisibleInDiffuseTransmission": 4,
+                        "aiVisibleInSpecularTransmission": 8,
+                        "aiVisibleInVolume": 16,
+                        "aiVisibleInDiffuseReflection": 32,
+                        "aiVisibleInSpecularReflection": 64
+                    }
```
BigRoy, line 236: Maybe best to make this global to the file to use it here and above in the function instead of duplicating it?

**Edit:** Or actually, should this just be: `calculate_visibility_mask(edit["attributes"])`?

openpype/hosts/maya/tools/mayalookassigner/commands.py
```diff
@@ -107,6 +108,7 @@ def create_asset_id_hash(nodes):
     """
     node_id_hash = defaultdict(list)
     for node in nodes:
+        shapes = cmds.ls(cmds.listRelatives(node, shapes=True), long=True)
```
BigRoy, line 111:
```suggestion
        shapes = cmds.listRelatives(node, shapes=True, fullPath=True)
```

BigRoy (commented):
Some notes

openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
```diff
@@ -21,11 +22,13 @@ def process(self, instance):
...
-                self.log.debug("content members: {}".format(members))
+                set_members = get_all_children(cmds.ls(members, long=True))
+                instance.data["setMembers"] = set_members
+                self.log.debug("content members: {}".format(set_members))
             elif objset.endswith("proxy_SET"):
-                instance.data["proxy"] = cmds.ls(members, long=True)
-                self.log.debug("proxy members: {}".format(members))
+                set_members = get_all_children(cmds.ls(members, long=True))
+                instance.data["proxy"] = set_members
+                self.log.debug("proxy members: {}".format(set_members))
```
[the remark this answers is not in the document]
tokejepsen: Sorry, not entirely sure what you mean?

openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
```diff
@@ -21,11 +22,13 @@ def process(self, instance):
                 self.log.warning("Skipped empty instance: \"%s\" " % objset)
                 continue
             if objset.endswith("content_SET"):
-                instance.data["setMembers"] = cmds.ls(members, long=True)
-                self.log.debug("content members: {}".format(members))
+                set_members = get_all_children(cmds.ls(members, long=True))
+                instance.data["setMembers"] = set_members
+                self.log.debug("content members: {}".format(set_members))
```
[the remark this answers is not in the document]
tokejepsen: yeah, good point. Will change that to `contentMembers`

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
```diff
@@ -133,12 +141,37 @@ def _extract(self, nodes, attribute_data, kwargs):
...
-                cmds.rename(duplicate_transform, node.split("|")[-1])
-                duplicate_transform = "|" + node.split("|")[-1]
+                basename = node.split("|")[-1].split(":")[-1]
+                cmds.rename(duplicate_transform, basename)
+                duplicate_transform = "|" + basename
 
                 duplicate_nodes.append(duplicate_transform)
                 delete_bin.append(duplicate_transform)
 
+                # Copy cbId from original to mtoa_constant.
```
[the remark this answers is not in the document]
tokejepsen: Yup good point.

openpype/hosts/maya/plugins/publish/validate_arnold_scene_source_cbid.py
```diff
@@ -0,0 +1,70 @@
...
+    families = ["ass"]
+    label = "Validate Arnold Scene Source CBID"
+    actions = [RepairAction]
+
+    @staticmethod
+    def _get_nodes_data(nodes):
+        nodes_by_name = {}
+        for node in nodes:
+            node_split = node.split("|")
+            nodes_by_name[node_split[-1].split(":")[-1]] = node
```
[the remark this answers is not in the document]
tokejepsen: Not sure how you envisage to use `rsplit`?
```python
from maya import cmds
node = cmds.ls(selection=True, long=True)[0]
print(node)
print(node.split("|")[-1].split(":")[-1])
print(node.rsplit(":", 1)[-1])
```
```
|pointcache_GRP|pCylinder1
pCylinder1
|pointcache_GRP|pCylinder1
```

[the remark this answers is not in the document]
BigRoy: It's just optimization, same as your usage of split, but only split once on the right hand side.

So this becomes this:
```python
node_name = node.split("|")[-1].split(":")[-1]
```
Becomes this:
```python
node_name = node.rsplit("|", 1)[-1].rsplit(":", 1)[-1]
```

Difference being that the first will iterate the full string right to left splitting into a big list of things:
```
my|long|nested|path|with|groups|and|kids
```
The latter would just stop on the first split it finds from the right hand side.

openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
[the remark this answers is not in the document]
BigRoy: I meant to say to also not include all children here like with `setMembers` but since you renamed the other data to keep the logic like this you can ignore this comment.

openpype/hosts/maya/plugins/publish/validate_arnold_scene_source_cbid.py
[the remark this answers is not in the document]
tokejepsen: Ahh I see. Done.

openpype/hosts/maya/plugins/publish/validate_arnold_scene_source_cbid.py
```diff
@@ -0,0 +1,70 @@
...
+
+        return nodes_by_name
```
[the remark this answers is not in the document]
tokejepsen: Good point. Catching same named nodes in a different validator now.

openpype/hosts/maya/plugins/publish/validate_arnold_scene_source_cbid.py
```diff
@@ -0,0 +1,70 @@
...
+            )
+
+    @classmethod
+    def repair(cls, instance):
+        for content_node, proxy_node in cls.get_invalid_couples(cls, instance):
+            cmds.setAttr(
+                proxy_node + ".cbId",
+                cmds.getAttr(content_node + ".cbId"),
+                type="string"
+            )
```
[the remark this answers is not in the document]
tokejepsen: Nice, thanks.

openpype/hosts/maya/tools/mayalookassigner/app.py
```diff
@@ -240,18 +241,26 @@ def on_process_selected(self):
...
+                for vp in vray_proxies:
+                    if vp in nodes:
+                        vrayproxy_assign_look(vp, subset_name)
 
-                    nodes = list(set(item["nodes"]).difference(vray_proxies))
+                nodes = list(set(item["nodes"]).difference(vray_proxies))
 
-                # Assign look
+            # Assign Arnold Standin look.
+            arnold_standins = set(cmds.ls(type="aiStandIn", long=True))
```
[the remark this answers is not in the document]
tokejepsen: Will do.

openpype/hosts/maya/tools/mayalookassigner/arnold_standin.py
tokejepsen: Lol, yeah. Must have forgotten to complete the refactor.

LiborBatek:
When creating ```ASS``` instance from local geometries I can now ```publish``` the ass file but when following the test steps by first ```Load model as alembic``` and trying to then publish as ```ass``` instance it fails and creates some duplicate objects in the scene.

image

LiborBatek:
to your question: before trying publish the highlighted items in the outliner (as seen on the pic above) were not present.

so basically there was just the loaded asset (abc) and created ```ass``` subset for publishing

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
```diff
@@ -113,34 +111,62 @@ def process(self, instance):
...
-                children = cmds.listRelatives(
-                    duplicate_transform, children=True
+                duplicate_transform = "{}|{}".format(
+                    parent, duplicate_transform
                 )
-                cmds.delete(set(children) - set(shapes))
 
                 duplicate_transform = cmds.parent(
                     duplicate_transform, world=True
                 )[0]
```
BigRoy, line 144: This will fail if the original transform was already a root object (did not have a parent) so we should check for that.
```suggestion

                if cmds.listRelatives(duplicate_transform, parent=True):
                    duplicate_transform = cmds.parent(
                        duplicate_transform, world=True
                    )[0]
```

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
```diff
@@ -113,34 +111,62 @@ def process(self, instance):
...
 
-                cmds.rename(duplicate_transform, node.split("|")[-1])
-                duplicate_transform = "|" + node.split("|")[-1]
+                basename = node.split("|")[-1].split(":")[-1]
+                duplicate_transform = cmds.rename(
+                    duplicate_transform, basename
+                )
```
BigRoy, line 149: Since we're forcing it to all be root nodes this will not result in the exact same names if the original hierarchy had non-unique names.

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
```diff
@@ -113,34 +111,62 @@ def process(self, instance):
...
+
+                # Discard the children.
```
BigRoy, line 151:
```suggestion
                # Discard children nodes that are not shapes
```

openpype/hosts/maya/plugins/publish/validate_arnold_scene_source.py
```diff
@@ -35,17 +34,29 @@ def _get_nodes_data(self, nodes):
...
 
-            nodes_by_name[node_split[-1]] = node
-            for shape in cmds.listRelatives(node, shapes=True):
-                nodes_by_name[shape.split("|")[-1]] = shape
+            node_name = node.rsplit("|", 1)[-1].rsplit(":", 1)[-1]
+
+            # Check for same same nodes, which can happen in different
+            # hierarchies.
+            if node_name in nodes_by_name:
+                same_named_nodes.append((node, nodes_by_name[node_name]))
```
BigRoy, line 42: If the node name would e.g. appear five times then this would 4 times report each entry together with the first entry. Is that what is intended?

Might make more sense to log it as `Found nodes with the same name {name}: {nodes}`?

openpype/hosts/maya/plugins/publish/validate_arnold_scene_source_cbid.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+
+    order = ValidateContentsOrder
+    hosts = ["maya"]
+    families = ["ass"]
+    label = "Validate Arnold Scene Source CBID"
+    actions = [RepairAction]
+
+    @staticmethod
+    def _get_nodes_data(nodes):
```
BigRoy, line 21: Method name can be clearer: `get_nodes_by_name`

openpype/hosts/maya/plugins/publish/validate_arnold_scene_source_cbid.py
```diff
@@ -0,0 +1,73 @@
...
+            instance.data["contentMembers"]
+        )
+        proxy_nodes_by_name = self._get_nodes_data(
+            instance.data.get("proxy", [])
+        )
+
+        invalid_couples = []
+        for content_name, content_node in content_nodes_by_name.items():
+            proxy_node = proxy_nodes_by_name.get(content_name, None)
+
```
BigRoy, line 40: This logic would still allow proxy nodes to exist which do not exist in content.

We might want to do:
```python
names = set(content_nodes_by_name) + set(proxy_nodes_by_name)
for name in names:
    proxy_node = proxy_nodes_by_name.get(name, None)
    content_node = content_nodes_by_name.get(name, None)
```

And then check for both `content_node` and `proxy_node` whether they exist.

tokejepsen:
@LiborBatek could you send me the model workfile? I'll try and replicate the setup you have here.

openpype/hosts/maya/plugins/publish/validate_arnold_scene_source_cbid.py
```diff
@@ -0,0 +1,70 @@
...
+        if invalid_couples:
+            raise PublishValidationError(
+                "Found proxy nodes with mismatching cbid:\n{}".format(
+                    invalid_couples
+                )
+            )
+
+    @classmethod
+    def repair(cls, instance):
+        for content_node, proxy_node in cls.get_invalid_couples(cls, instance):
```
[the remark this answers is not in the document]
BigRoy: I see it now. You passed it `cls` explicitly to the function! Note that the behavior then is that `self` is NOT an instance of the class but is the actual class type itself. It doesn't introduce bugs currently because you're not doing anything with attrributes on `self` but I'd better clean this up and just make `get_invalid_couples` a classmethod so the behavior is clearer.

Some examples:

How you should call `fn`, but it doesn't work.
```python
class Test(object):
    
    def __init__(self):
        self.foobar = "a"
    
    def fn(self):
        print(self.foobar)
        
    @classmethod
    def class_fn(cls):
        cls.fn()
        
Test.class_fn()
# TypeError: fn() missing 1 required positional argument: 'self' # 
```

If I pass the class then unexpected things can happen - e.g. `__init__` was never triggered because we passed the class and not an instance of the class:
```python
class Test(object):
    
    def __init__(self):
        self.foobar = "a"
    
    def fn(self):
        print(self.foobar)
        
    @classmethod
    def class_fn(cls):
        cls.fn(cls)
        
Test.class_fn()
# AttributeError: type object 'Test' has no attribute 'foobar' # 
```

But worse it could introduce bugs like:
```python


class Test(object):
    
    value = 0
    
    def set_value(self, value):
        self.value = value
        
    @classmethod
    def class_fn(cls, value):
        cls.set_value(cls, value)
        

a = Test()
print('a.value', a.value)
# a.value 0

b = Test()
b.set_value(2)
print('b.value', b.value)
# b.value 2

c = Test()
print('c.value', c.value)
# c.value 0

# After calling this we've now actually changed the
# value on the class type and not the instance. As such
# now the default value is always changed.
Test.class_fn(3)

print('a.value', a.value)
# a.value 3
print('b.value', b.value)
# b.value 2
print('c.value', c.value)
# c.value 3

# Note how classes which never got `self.value` assigned explicitly
# are still set to the class value and thus now suddenly all got changed
# and worse, new class instances will now also default to the value.
d = Test()
print('d.value', d.value)
# d.value 3

# All because we ended up actually setting:
print("Test.value", Test.value)
```

openpype/hosts/maya/tools/mayalookassigner/app.py
```diff
@@ -240,18 +241,37 @@ def on_process_selected(self):
...
+                nodes = list(set(item["nodes"]).difference(vray_proxies))
+            else:
+                self.echo(
+                    "Could not assign to VRayProxy because vrayformaya plugin "
+                    "is not loaded."
+                )
+
+            # Assign Arnold Standin look.
+            if cmds.pluginInfo("mtoa", query=True, loaded=True):
+                arnold_standins = set(cmds.ls(type="aiStandIn", long=True))
```
BigRoy, line 262: This doesn't necessarily need to be a set. We're doing no lookups here.
And `.difference()` call down below can work against any iterable as far as I know.

openpype/hosts/maya/tools/mayalookassigner/vray_proxies.py
```diff
@@ -263,16 +180,16 @@ def vrayproxy_assign_look(vrayproxy, subset="lookDefault"):
...
+        shadernodes, _ = lib.load_look(version["_id"])
 
         # Get only the node ids and paths related to this asset
         # And get the shader edits the look supplies
         asset_nodes_by_id = {
             node_id: nodes_by_id[node_id] for node_id in node_ids
         }
         edits = list(
-            lib.iter_shader_edits(
+            api.lib.iter_shader_edits(
```
BigRoy, line 192: Question: Why would we want to directly turn this into a list instead of just iterating over it in the for loop?
Code is likely based on existing code - but just wondering.

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
tokejepsen: We are validating against same named nodes [here](https://github.com/tokejepsen/pype/blob/feature/OP-4859_cant-assign-shaders-to-the-ass-file/openpype/hosts/maya/plugins/publish/validate_arnold_scene_source.py#L47-L51)

openpype/hosts/maya/plugins/publish/validate_arnold_scene_source_cbid.py
tokejepsen: Done

openpype/hosts/maya/plugins/publish/validate_arnold_scene_source_cbid.py
tokejepsen: We validate we have the same amount of content to proxy nodes [here](https://github.com/tokejepsen/pype/blob/feature/OP-4859_cant-assign-shaders-to-the-ass-file/openpype/hosts/maya/plugins/publish/validate_arnold_scene_source.py#L84-L92)

openpype/hosts/maya/tools/mayalookassigner/app.py
tokejepsen: I was following the VrayProxy code [here](https://github.com/tokejepsen/pype/blob/feature/OP-4859_cant-assign-shaders-to-the-ass-file/openpype/hosts/maya/tools/mayalookassigner/app.py#L247). If there is no issues (?) then I would just keep it as is.

openpype/hosts/maya/plugins/publish/collect_look.py
```diff
@@ -556,7 +556,7 @@ def collect_attributes_changed(self, instance):
                     continue
                 if cmds.getAttr(attribute, type=True) == "message":
                     continue
-                node_attributes[attr] = cmds.getAttr(attribute)
+                node_attributes[attr] = cmds.getAttr(attribute, asString=True)
```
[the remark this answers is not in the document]
tokejepsen: This PR is for assign looks to loaded ass files.

This ensured that enum values are returned as strings rather than their index value.

tokejepsen:
Resolves #4535

tokejepsen:
@LiborBatek could you try again?
I've successfully published your workfile here, so fingers crossed!

openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
```diff
@@ -21,18 +22,20 @@ def process(self, instance):
                 self.log.warning("Skipped empty instance: \"%s\" " % objset)
                 continue
             if objset.endswith("content_SET"):
-                instance.data["setMembers"] = cmds.ls(members, long=True)
-                self.log.debug("content members: {}".format(members))
+                set_members = get_all_children(cmds.ls(members, long=True))
```
BigRoy, line 25: I believe this could just be `get_all_children(members)` - if not, then there's already something off with the members input since those should already have unique paths (not necessarily long paths).

It's good to note that `get_all_children` includes all children, but not the original members themselves. So it only includes the children of `members` not the original `members`.

You might want to do:
```python
members = cmds.ls(members, long=True)
children = get_all_children(members)
hierarchy = list(set(members + children))
```

_It is technically possible for some of the members to be present in the result of `get_all_children` if one of the members is a child of another member._

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
```diff
@@ -113,34 +111,63 @@ def process(self, instance):
...
+                # Only interested in transforms with shapes.
+                shapes = cmds.listRelatives(
+                    node, shapes=True, fullPath=True
+                ) or []
+                if not shapes:
+                    continue
+
+                parent = cmds.listRelatives(
+                    node, parent=True, fullPath=True
+                )[0]
```
BigRoy, line 136: If the original node has no parent this would fail.

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
```diff
@@ -113,34 +111,63 @@ def process(self, instance):
...
                 duplicate_transform = cmds.duplicate(node)[0]
+                duplicate_transform = "{}|{}".format(
+                    parent, duplicate_transform
+                )
+
+                if cmds.listRelatives(duplicate_transform, parent=True):
```
BigRoy, line 142: We're first prefixing with the parent name, but then checking whether it has a parent? Seems off?

ALso, should duplicate return the valid path? If not, then we should do `cmds.ls(duplicate_transform, long=True)[0]` instead because it'll otherwise still likely fail if the node name isn't unique. I'm pretty sure duplicate will then already be returning the shortest unique path.

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
```diff
@@ -113,34 +111,63 @@ def process(self, instance):
...
 
-                duplicate_transform = cmds.parent(
-                    duplicate_transform, world=True
-                )[0]
-
-                cmds.rename(duplicate_transform, node.split("|")[-1])
-                duplicate_transform = "|" + node.split("|")[-1]
-
                 duplicate_nodes.append(duplicate_transform)
+                duplicate_nodes.extend(shapes)
```
BigRoy, line 162: Aren't the shapes automatically deleted since they are children of the transform?

openpype/hosts/maya/tools/mayalookassigner/commands.py
```diff
@@ -122,7 +124,12 @@ def create_asset_id_hash(nodes):
                 pid = k.split(":")[0]
                 if node not in node_id_hash[pid]:
                     node_id_hash[pid].append(node)
-
+        elif shapes and cmds.nodeType(shapes[0]) == "aiStandIn":
+            path = arnold_standin.get_standin_path(shapes[0])
+            for id, _ in arnold_standin.get_cbid_by_node(path).items():
+                pid = id.split(":")[0]
+                if shapes[0] not in node_id_hash[pid]:
```
BigRoy, line 131: This not in node_id_hash[pid]` lookup will get very slow if there are many nodes per `pid`. Should we refactor this `node_id_hash` to be `defaultdict(set)`?

Also, I have no idea what `pid` stands for. It should be the asset's id. Why not pick `asset_id` as variable name? I know the v-ray logic above also names it `pid` but it's confusing what it refers to right?

openpype/hosts/maya/tools/mayalookassigner/vray_proxies.py
BigRoy, line 192: This looks confusingly close to the `lib.` above. Should we maybe just import this maya api lib as `maya_lib`?

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
tokejepsen: The `duplicate_nodes` variable is for adding `cbId`. Not for deletion.

openpype/hosts/maya/tools/mayalookassigner/vray_proxies.py
tokejepsen: Done

openpype/hosts/maya/tools/mayalookassigner/commands.py
tokejepsen: Yup, have replaced dict with defaultdict where I could find it needed.

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
tokejepsen: Yeah, have simplified this.

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
tokejepsen: Good point!

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
```diff
@@ -113,34 +111,63 @@ def process(self, instance):
...
+                # Only interested in transforms:
+                if cmds.nodeType(node) != "transform":
+                    continue
+
+                # Only interested in transforms with shapes.
+                shapes = cmds.listRelatives(
+                    node, shapes=True, fullPath=True
+                ) or []
+                if not shapes:
+                    continue
```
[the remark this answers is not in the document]
tokejepsen: Yup done

openpype/hosts/maya/tools/mayalookassigner/commands.py
```diff
@@ -107,6 +108,7 @@ def create_asset_id_hash(nodes):
     """
     node_id_hash = defaultdict(list)
     for node in nodes:
+        shapes = cmds.listRelatives(node, shapes=True, fullPath=True)
```
BigRoy, line 111: Why do we need the listRelatives call - the VRayProxy node I believe is also a Shape node right? Why did it work for that without requiring to go down to extra child nodes?

I believe however you're calling this `create_asset_id_hash` the nodes you're passing should already include the shape nodes too. You definitely don't want to be getting shapes in a for loop like this over what initially already could be a massive nodes list.

tokejepsen: `create_asset_id_hashes` is being used on the selection [here](https://github.com/ynput/OpenPype/blob/3da1f72f96ff8c633667211e5d0294a41a45a93d/openpype/hosts/maya/tools/mayalookassigner/widgets.py#L118-L123). Where do you see it including the shapes?

BigRoy: The `dag=True` flag includes the full hierarchy of all dag nodes. When selection is True there then it takes the one from commands that I passed above which explicitly takes the selection and its children. Both if and else branch there should include hierarchy downwards.

If they wouldn't that's a bug.

Also that, `cmds.ls` flag might need to also take `allPaths=True` if it requires all instances to be taken into account as well.

tokejepsen: Ok, think I've found the issue. This [method](https://github.com/ynput/OpenPype/blob/3da1f72f96ff8c633667211e5d0294a41a45a93d/openpype/hosts/maya/tools/mayalookassigner/commands.py#L78-L98) only lists the nodes added to the container and not its children which makes sense. Problem is that when loading in the aiStandIn, we only add the transform.
Question is whether we should add the shape or accommodate otherwise?

BigRoy: Interesting. 

This is a tricky situation. The container object set should include all the nodes.

I believe at some point in time shapes got excluded from being in object sets die to issues with shader assignments with instances if the shape node is in an objectSet. (This was for referenced objects actually)

Then at some point the referenced nodes in containers were replaced with just adding the reference node to the containers instead of all its nodes.

But I assume that the loader for e.g. aiStandin still excludes the shape from the container objectSet which is a bit counter intuitive because the shape IS actually part of that loaded data to begin with.

The best solution now would be that whenever we're retrieving the members from the container, in particular in the look assigner that we do a single `shapes = cmds.listRelatives(container_members, shapes=True, fullPath=True, noIntermediate=True)` then we just add that to the result of that function. We'd still need to ensure it's unique so `list(set(members))` or alike after that.

Then let's add a clear comment about what we are solving that way, e.g. **Some containers, like loaded aiStandin don't include the shape directly in the container's objectSet so we need to include the shapes here ourselves**

What do you think?

tokejepsen: I think maybe I should just add the shape to the container node. The shape is not needed for the vray implementation, so its most likely that I've misunderstood something like doing the [ASS overhaul](https://github.com/ynput/OpenPype/pull/4449).

BigRoy: It would make this logic incompatible however with any file that was loaded since that overhaul. I think we should be fine with including the shapes for now as I mentioned. We are triggering the call only once, shouldn't be too problematic.

BigRoy:
> @BigRoy does this make any sense to you?

Unfortunately not, sorry. This seems like an MtoA bug and might be best reported with autodesk or checked with their team.

It would be good however to maybe try debugging when it crashes - e.g. does it still crash if you remove some of the overrides on the `ass` file that the look assignment generates. It could be that one of those procedural overrides is configured wrong in a way that maybe makes arnold crash.

tokejepsen:
Think the issue is this [method](https://github.com/tokejepsen/pype/blob/feature/OP-4859_cant-assign-shaders-to-the-ass-file/openpype/hosts/maya/tools/mayalookassigner/arnold_standin.py#L61-L90). Will dig deeper tomorrow.

tokejepsen:
Looks like its not very safe to using Arnold Python API in Maya for whatever reason and I'm not holding my breath for any solution from Autodesk/Arnold, so I'll be going with a sidecar solution instead.
For replication:

1. Launch Maya.
2. Run in script editor:
```
import arnold


arnold.AiBegin()
arnold.AiEnd()
```
3. Crashing of Maya or Arnold stops working as expected.

BigRoy:
What version of Maya + MtoA are you using? [The example code](https://github.com/ynput/OpenPype/pull/4460#issuecomment-1491589442) runs perfectly fine for me in Maya 2022.3 + Mtoa 5.2.1.1

tokejepsen:
> What version of Maya + MtoA are you using? https://github.com/ynput/OpenPype/pull/4460#issuecomment-1491589442 runs perfectly fine for me in Maya 2022.3 + Mtoa 5.2.1.1

Maya 2023 and MTOA 5.3.1.

@LiborBatek what version of Maya and MTOA are you running?

LiborBatek:
> @LiborBatek what version of Maya and MTOA are you running?

maya 2023.3 + arnold 5.1.3.3

tokejepsen:
@LiborBatek could you try this again now?
I've replaced the ASS query with json sidecar, so it should be more stable now.

LiborBatek (approved):
Now its possible to assign ```Look``` asset to the ```Ass``` Instance loaded in scene as with other representations of the model.

So working ok.
image

Not sure why but I m not able to graph it in ```Hypershade``` editor its not shown as connected shader to the ass asset. Maybe its by its nature just a reminder what observed during testing....could be issue later on.

tokejepsen:
> Not sure why but I m not able to graph it in Hypershade editor its not shown as connected shader to the ass asset. Maybe its by its nature just a reminder what observed during testing....could be issue later on.

Yeah, assigning shaders to aiStandIn is done through operators, which do not follow any native material assignments workflows.; https://help.autodesk.com/view/ARNOL/ENU/?guid=arnold_user_guide_ac_operators_html

tokejepsen:

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
```diff
@@ -113,34 +125,58 @@ def process(self, instance):
...
             duplicate_nodes = []
             for node in nodes:
+                # Only interested in transforms:
+                if cmds.nodeType(node) != "transform":
+                    continue
+
+                # Only interested in transforms with shapes.
+                shapes = cmds.listRelatives(
+                    node, shapes=True, noIntermediate=True
+                ) or []
```
BigRoy, line 145: No need for `or []` because we're continuing on `not shapes` anyway. Not sure if we care, but if we do.. then this could be the change.
```suggestion
                shapes = cmds.listRelatives(
                    node, shapes=True, noIntermediate=True
                )
```

openpype/hosts/maya/tools/mayalookassigner/arnold_standin.py
```diff
@@ -0,0 +1,235 @@
...
+    for f in os.listdir(os.path.dirname(path)):
+        if f.endswith(".json"):
+            json_path = os.path.join(os.path.dirname(path), f)
+            break
+
+    with open(json_path, "r") as f:
+        return json.load(f)
+
+
+def shading_engine_assignments(shading_engine, attribute, nodes, assignments):
```
BigRoy, line 81: I feel like this might need a docstring. This likely came from elsewhere but I'm not entirely sure what this function does by a quick read of it.

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
```diff
@@ -113,34 +125,58 @@ def process(self, instance):
...
-
                 duplicate_nodes.append(duplicate_transform)
+                duplicate_nodes.extend(shapes)
                 delete_bin.append(duplicate_transform)
 
-            with attribute_values(attribute_data):
-                with maintained_selection():
+            # Copy cbId to mtoa_constant.
+            for node in duplicate_nodes:
+                nodes_by_id[lib.get_id(node)].append(node.replace("|", "/"))
```
BigRoy, line 176: Should this include a comment about `|` being the maya path separator and `/` being the ass/alembic separator? Or at least, I suspect that's why we are doing the replace on the string.

I wonder however, is that still the case now we're working with the sidecar `.json` file?

openpype/hosts/maya/tools/mayalookassigner/arnold_standin.py
tokejepsen: Yeah, fair.

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
tokejepsen: > I wonder however, is that still the case now we're working with the sidecar .json file?

This is because of the sidecar json that we need make it similar to how it look in `.ass` files. We could also convert `| > /` when loading the json?

BigRoy: Ah sorry, totally missed that this is in the extractor. The conversion seems fine to me 👍 . Just wondered whether a comment might help clarify that it's maya path separator to arnold/ass path separator.

tokejepsen:
```suggestion
                # Converting Maya hierarchy separator "|" to Arnold separator "/".
                nodes_by_id[lib.get_id(node)].append(node.replace("|", "/"))
```

openpype/hosts/maya/tools/mayalookassigner/arnold_standin.py
```diff
@@ -0,0 +1,235 @@
...
+
+    Returns:
+        (dict): Dictionary with node full name/path and id.
+    """
+    path = cmds.getAttr(standin + ".dso")
+    json_path = None
+    for f in os.listdir(os.path.dirname(path)):
+        if f.endswith(".json"):
+            json_path = os.path.join(os.path.dirname(path), f)
+            break
```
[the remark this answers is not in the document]
tokejepsen: > I'm just preparing for the possibility that maybe there's going to be another representation or an attached review that somehow starts having a json file of its own, etc.

Hoping this wont happen before another workflow takes over, like Material/USD etc.

> Also, should we account for the case that maybe a json path is NOT found for it? E.g. older publishes or maybe .ass files published through a standalone publisher? A warning should be logged maybe and the standin ignored?

Yeah, good point. Done.

tokejepsen:
@moonyuet or @LiborBatek could I get an approval on this?

LiborBatek (approved):
Seems ok. Being already tested.

antirotor (approved):

OlivierOsotimehin:
Hello,
This feature will be so great if it also worked on .abc loaded through the standin. This mean we could load animation caches directly in .abc and apply the look on them without the need to réexport in .ass

tokejepsen:
@OlivierOsotimehin it does work on abc representations :⁠-⁠)

PR #4461, opened 2023-02-14
Maya: Simplify pointcache proxy integration

tokejepsen:
## Brief description
Missing commit for #4449 

This makes pointcache proxy and Arnold Scene Source proxy use the same integration workflow.

## Testing notes
1. Create pointcache in Maya and add meshes to the proxy objectset.
2. Publish.
3. Check the proxy alembic is integrated rather than ending up in the `resources` folder.


LiborBatek:
Please provide some testing steps...thx

tokejepsen:
> Please provide some testing steps...thx

Sorry, my bad! Done now.

LiborBatek (changes requested):
Following testing steps it gives me this error during Publish:
4461_publishWithProxyGeo

when no proxy meshes added the pointcache publishes normally and can be loaded via asset loader...
4461_publishWithoutProxyGeo

I did try to add the asset geometries to the ```proxy_SET``` but gave me also that errors. So it do not work when any geometry added to the set.

tokejepsen:
@LiborBatek thanks for testing. Should be fixed now.

moonyuet:
The code looks okay to me. The only thing I suugested would be adding some info to inform the users about the proxy extraction(personal preference only)
```
        if not instance.data.get("proxy"):
            return

        self.log.info("Extracting proxy pointcache..")
```

moonyuet (approved):

LiborBatek:
Now it works when added some geometry to the ```proxy_SET```.

image

Also outputted the ```pointcache``` and ```pointcache_Proxy``` ok.

tokejepsen:
@mkolar ready to merge.

PR #4470, opened 2023-02-15
Update Asset field of creator Instances in Maya Template Builder

mre7a:
## Brief description
When we build a template with Maya Template Builder, it will update the asset field of the sets (creator instances) that are imported from the template. 

## Description
When building a template, we also want to define the publishable content in advance: create an instance of a model, or look, etc., to speed up the workflow and reduce the number of questions we are asked. After building a work file from a saved template that contains pre-created instances, the template builder should update the asset field to the current asset.

## Additional info
As an aside, for some instances that contain CbId, I didn't know if we should update the ID or not. 

## Testing notes:
1. Create a template base on the documentation.
2. Make sure that your template contains a creator instance. 
3. In a new asset, build workfile from template.
4. You should see that asset field is updated to the current one. 

image

### DIFF
--- a/openpype/hosts/maya/api/workfile_template_builder.py
+++ b/openpype/hosts/maya/api/workfile_template_builder.py
@@ -50,6 +50,7 @@ class MayaTemplateBuilder(AbstractTemplateBuilder):
             return True
 
         # update imported sets information
+        asset_name = get_current_asset_name()
         for node in imported_sets:
             if not cmds.attributeQuery("id", node=node, exists=True):
                 continue
@@ -57,9 +58,9 @@ class MayaTemplateBuilder(AbstractTemplateBuilder):
                 continue
             if not cmds.attributeQuery("asset", node=node, exists=True):
                 continue
-            asset = legacy_io.Session["AVALON_ASSET"]
 
-            cmds.setAttr("{}.asset".format(node), asset, type="string")
+            cmds.setAttr(
+                "{}.asset".format(node), asset_name, type="string")
 
         return True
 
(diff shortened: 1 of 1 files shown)


openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -41,10 +41,26 @@ def import_template(self, path):
...
+        for node in imported_sets:
+            if not cmds.attributeQuery("id", node=node, exists=True):
+                continue
+            if cmds.getAttr("{}.id".format(node)) != "pyblish.avalon.instance":
+                continue
+            if not cmds.attributeQuery("asset", node=node, exists=True):
+                continue
+            asset = legacy_io.Session["AVALON_ASSET"]
+
+            cmds.setAttr("{}.asset".format(node), asset, type="string")
```
iLLiCiTiT, line 62:
```suggestion
        # update imported sets information
        asset_name = get_current_asset_name()
        for node in imported_sets:
            if not cmds.attributeQuery("id", node=node, exists=True):
                continue
            if cmds.getAttr("{}.id".format(node)) != "pyblish.avalon.instance":
                continue
            if not cmds.attributeQuery("asset", node=node, exists=True):
                continue

            cmds.setAttr(
                "{}.asset".format(node), asset_name, type="string")
```

openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -2,7 +2,7 @@
 
 from maya import cmds
 
-from openpype.pipeline import registered_host
+from openpype.pipeline import registered_host, legacy_io
```
iLLiCiTiT, line 5:
```suggestion
from openpype.pipeline import registered_host, get_current_asset_name
```

antirotor (approved):
It works for me!

PR #4471, opened 2023-02-15
TVPaint: Use Publisher tool

iLLiCiTiT:
## Brief description
Use Publisher tool and new creation system in TVPaint integration.

## Description
Using new creation system makes TVPaint integration a little bit easier to maintain for artists. Removed unneeded tools Creator and Subset Manager tools. Goal is to keep the integration work as close as possible to previous integration. Some changes were made but primarilly because they were not right using previous system.

All creators create instance with final family instead of changing the family during extraction. Render passes are not related to group id but to render layer instance. Render layer is still related to group. Workfile, review and scene render instances are created using autocreators instead of auto-collection during publishing. Subset names are fully filled during publishing but instance labels are filled on refresh with the last known right value. Implemented basic of legacy convertor which should convert render layers and render passes.

## Additional info
Because of special subset name template for `render` families I had to add option to filter subset name templates by different family than is family to fill the subset name, which is kind anti-pattern -> fill family is `render` but for filtering is used `renderLayer`, `renderPass` or `renderScene` family. It can be kept as is now but it probably should be changed.
One option is to have same subset name template for all `render` families in TVPaint, which could work but could block other implementations of possible TVPaint render families -> this would break backwards compatibility for subset name templates.

## TODOs
Update documentation.

## Testing notes:
1. Start TVPaint
2. Make a scene
3. Open Publisher
4. Create a Render Layer/s (you can create it out of selection)
5. Reset publisher (this is to update Render Layers combobox)
6. Create a Render Pass/es - Based on selected TVPaint layers in scene
7. Enable/Disable instances
8. Publish

Legacy convertor
1. Open TVPaint file with legacy created instances
2. Run Convert
3. All instances should be converted to new creators


tokejepsen:
Get this when launching the publisher;
```
  File "C:\Users\tokejepsen\OpenPype\openpype\tools\publisher\window.py", line 456, in _on_show_timer
    self.reset()
  File "C:\Users\tokejepsen\OpenPype\openpype\tools\publisher\window.py", line 462, in reset
    self._controller.reset()
  File "C:\Users\tokejepsen\OpenPype\openpype\tools\publisher\control.py", line 1746, in reset
    self._reset_plugins()
  File "C:\Users\tokejepsen\OpenPype\openpype\tools\publisher\control.py", line 1764, in _reset_plugins
  File "C:\Users\tokejepsen\OpenPype\openpype\pipeline\create\context.py", line 1615, in reset_plugins
    self._reset_creator_plugins()
  File "C:\Users\tokejepsen\OpenPype\openpype\pipeline\create\context.py", line 1695, in _reset_creator_plugins
    creator = creator_class(
  File "C:\Users\tokejepsen\OpenPype\openpype\pipeline\create\creator_plugins.py", line 204, in __init__
  File "C:\Users\tokejepsen\OpenPype\openpype\hosts\tvpaint\plugins\create\create_workfile.py", line 21, in apply_settings
    self.active_on_create = plugin_settings["active_on_create"]
KeyError: 'active_on_create'
```

iLLiCiTiT:
> Get this when launching the publisher;

Had swapped code in workfile and review creators. Thanks

tokejepsen:
Getting this when launching the publisher now.

Capture

tokejepsen:
Small update to include `openpype\hosts\tvpaint\plugins\publish\validate_scene_settings.py`: `scene_data["resolutionWidth"]` > `scene_data["resolutionHeight"]`

```
                "current_height": scene_data["resolutionWidth"],
```

tokejepsen:
Publishing all seems to work well. Further thoughts:

- what is the reason for having a separate `Scene Render` instance? As far as I can see, we could just auto create an instance which includes all (enabled) layers in the scene, and include in the name something like `All_Layers`.
- is the token supposed to be replace here?
Capture
- too many entries to the group selection?
Screenshot 2023-02-17 125316
- think the client asked for a way to create passes from layers. Will this be included in this PR? We had to implement something similar for our last production (on Pype v2), where the artist gets asked whether they wanna create render passes for each tvpaint layer when creating a render layer.

iLLiCiTiT:
>  As far as I can see, we could just auto create an instance which includes all (enabled) layers in the scene, and include in the name something like All_Layers.

That's what `Scene Render` is for.

> is the token supposed to be replace here?

No, it could change on change of Render Layer on the instance.

> too many entries to the group selection?

I though I fixed that, will look into that.

> Will this be included in this PR? 

No, this is just conversion for publisher. That will be next PR.

tokejepsen (approved):

PR #4473, opened 2023-02-16
Publisher: Report also crashed creators and convertors

iLLiCiTiT:
## Brief description
Added crashes of creators and convertos discovery (lazy solution).

## Description
Report in Publisher also contains information about crashed files caused during creator plugin discovery and convertor plugin discovery. They're not separated into categroies and there is no other information in the report about them, but this helps a lot during development. This change does not need to change format/schema of the report nor UI logic.

## Testing notes:
1. Cause a crash in Creator plugin on import (e.g. add `import foobar` to file)
2. Open Publisher
3. Go to 'Report' tab
4. Go to crashed plugins information
5. The crashed file with creator should be there

### DIFF
--- a/openpype/pipeline/create/context.py
+++ b/openpype/pipeline/create/context.py
@@ -21,6 +21,7 @@ from openpype.lib.attribute_definitions import (
 )
 from openpype.host import IPublishHost
 from openpype.pipeline import legacy_io
+from openpype.pipeline.plugin_discover import DiscoverResult
 
 from .creator_plugins import (
     Creator,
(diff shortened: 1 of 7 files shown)


kalisp:
Publisher fails with:
2023-02-16 11_42_28-Window

iLLiCiTiT:
> Publisher fails with:

Should be fixed

kalisp:
Now it works, it shows failed creator in `Crashed plugins`.

Not sure if it because this PR but after fixing creator again and hitting Refresh:
2023-02-16 12_44_36-Window

iLLiCiTiT:
> Not sure if it because this PR but after fixing creator again and hitting Refresh:

Don't know either will discover

iLLiCiTiT:
> Don't know either will discover

Wasn't cause by this PR but is fixed.

kalisp (approved):

PR #4474, opened 2023-02-16
Publisher: Fix compatibility of QAction in Publisher

iLLiCiTiT:
## Brief description
Fix `QAction` for older version of Qt bindings where QAction requires a parent on initialization.

## Description
This bug was discovered in Nuke 11. Fixed by creating QAction when QMenu is already available and can be used as parent.

## Testing notes:
1. Open Publisher in Nuke 11
2. Should not crash


antirotor:
Not so easy to test because of #4475 🤣

antirotor (approved):
Works!

PR #4477, opened 2023-02-16
Unreal: Render Creator enhancements

simonebarbieri:
## Brief description
Improvements to the creator for render family

## Description
This PR introduces some enhancements to the creator for the render family in Unreal Engine:
- Added the option to create a new, empty sequence for the render.
- Added the option to not include the whole hierarchy for the selected sequence.
- Improvements of the error messages.

## Additional info
Requires #4370 

## Testing notes:
1. Try create a new render instance with the option of creating a new Level Sequence.
2. Try create a new render instance with and without using the hierarchy (to use the hierarchy, it will be necessary to have loaded a layout with a hierarchy).


antirotor (changes requested):
When I use it with  "add Level Sequence" it works, but it doesn't add the newly created sequencer to the rendering instance - I need to do it manually:

image

The **Render...** will not issue any warning, it just do nothing. I guess we need to handle that too. 

Also - we need to address the `renders` root for Anatomy as it is not in default Anatomy and needs to be configured to make rendering work. But that should be done probably in accordance with #4583 as separate PR

antirotor (changes requested):
I think we are almost there - the only issue I've found is this:

image

that without changing anything, the render creates 0-149 but everything is expecting 1-150

antirotor (approved):

PR #4480, opened 2023-02-16
Settings - first implementation of saving settings changes to DB

kalisp:
## Brief description
This logs all changes in Settings to separate collection ('setting_log') to help debug/trace changes. This functionality is pretty basic as it will be replaced by Ayon implementation in the future.

## Description
It saves only 'changes' (like diffs) not old values as resolution of that could get tricky (in complex dictionaries). 
No UI implemented, accessible only through Compass or other IDE.

## Additional info
It produces documents like this:
2023-02-20 13_33_51-Window

Changes to system/anatomy/project settings could be filtered by value in `settings_type`.

It would be theoretically possible to apply list of these changes on barebone Settings to create state at specific point in time. Not sure if we want to investigate this for v3 though.

## Testing notes:
1. Try to change something in settings, save, check `settings_log`

### DIFF
--- a/openpype/settings/handlers.py
+++ b/openpype/settings/handlers.py
@@ -238,7 +238,7 @@ class SettingsHandler(object):
     @abstractmethod
     def save_change_log(self, project_name, changes, settings_type):
         """Stores changes to settings to separate logging collection.
-        
+
         Args:
             project_name(str, null): Project name for which overrides are
                 or None for global settings.
--- a/openpype/settings/handlers.py
+++ b/openpype/settings/handlers.py
@@ -9,6 +9,7 @@ import six
 import openpype.version
 from openpype.client.mongo import OpenPypeMongoConnection
 from openpype.client.entities import get_project_connection, get_project
+from openpype.lib.pype_info import get_workstation_info
 
 from .constants import (
     GLOBAL_SETTINGS_KEY,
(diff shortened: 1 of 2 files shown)


iLLiCiTiT (approved):

openpype/settings/handlers.py
```diff
@@ -235,6 +235,18 @@ def save_project_anatomy(self, project_name, anatomy_data):
         """
         pass
 
+    @abstractmethod
+    def save_change_log(self, project_name, changes, settings_type):
+        """Stores changes to settings to separate logging collection.
+
+        Args:
+            project_name(str, null): Project name for which overrides are
```
Minkiu, line 243: Shouldn't then be `project_name=None` in the args?

kalisp: I copied docstring from different method, but `project_name` could actually be None, but adding this to signature would mean to add this to both `changes` and `settings_type` which are required. I wanted to keep `project_name` as first argument as it is in other methods.

m-u-r-p-h-y (commented):
I would add more human readable information about source of the changes. The user hash tag is not helpful here to find out who did the change.

something that we already have in OpenPype info tab
image

kalisp:
Added additional workstation info:
2023-02-20 13_33_51-Window

m-u-r-p-h-y (approved):
looks great. Will help us a lot!

PR #4481, opened 2023-02-16
General: Avoid popup console on windows build

iLLiCiTiT:
## Brief description
Avoid popup console on windows UI build.

## Description
Windows UI build may causes that subprocesses are opened using console (logic inside `subprocess` module). To avoid that is possible to pass different creation flags on windows which will cause that the process won't trigger a console or should not. This auto fill of creation flags happens only on windows machines. Added option to bypass creation flags if they are set to `0`. In that case is auto-fill disabled.

## Testing notes:
Oiiotool and ffmpeg subprocess should not raise a window if are executed on windows through `openpype_gui.exe`.

You can try that running this script in Tray > Admin > Console. It will run oiio tool 10x
```
from openpype.lib import run_subprocess, get_oiio_tools_path

src_path = r"<fill some source file path>"
dst_path = r"<fill output file path (different than source)>"
args = [
    get_oiio_tools_path(),
    "--nosoftwareattrib",
    "--compression", "none",
    "-i:ch=R,G,B,A", src_path,
    "--ch", "R=R,G=G,B=B,A=A",
    "--subimage", "0",
    "-o", dst_path
]
for _ in range(10):
    run_subprocess(args)
```


tokejepsen:
Getting console in Maya.

[Untitled_ Feb 17, 2023 8_15 AM (1).webm](https://user-images.githubusercontent.com/1860085/219589997-0a83d056-e8d2-4634-8f3e-b4ac3f361721.webm)

tokejepsen (changes requested):
Console appears in Maya.

tokejepsen (changes requested):
I get a console appearing on the first execution but subsequent executions do not have any console window.
[Untitled_ Feb 17, 2023 10_19 AM.webm](https://user-images.githubusercontent.com/1860085/219617487-0f89d0f3-9397-43e2-a3b2-565b2e48c773.webm)

openpype/lib/execute.py
```diff
@@ -74,18 +74,23 @@ def execute(args,
...
 
     Entered arguments and keyword arguments are passed to subprocess Popen.
 
+    Set 'creationflags' to '0' (int) if auto-fix for creation of new window
+    should be ignored.
+
     Args:
-        *args: Variable length arument list passed to Popen.
+        *args: Variable length argument list passed to Popen.
+        creationflags (int): Creation flags for 'subprocess.Popen'.
```
Minkiu, line 89: I was wondering why have a specific argument, instead of reading it from the kwargs?

iLLiCiTiT: You're right. Changed.

iLLiCiTiT:
> I get a console appearing on the first execution but subsequent executions do not have any console window.

Added `CREATE_NO_WIDOW` flag (if is available) so it should be ok now.

kalisp (changes requested):
In Nuke `ExtractReview` fails with
2023-02-20 11_15_31-OpenPype publisher

antirotor (approved):

antirotor (changes requested):
Indeed, there is no `subprocess.DETACHED_PROCESS` in Nuke subprocess

tokejepsen (approved):
Tested successfully in Maya.

kalisp:
It doesn't fail in Nuke anymore.

(Still opening windows in `extract_burnins.py` though.)

Minkiu (approved):

kalisp (approved):
It looks good in Nuke

antirotor (approved):

BigRoy:
Even after this PR I'm still getting tons of command prompts showing/hiding during publishing.

It happens the first time when opening the Publisher window in a DCC session (as if it's caching something?) which doesn't happen on "reset".
Then it also happens at the end of publishing per render instance for burnins and other things.

Here's what I'm visually seeing, keep an eye on the top left.
https://user-images.githubusercontent.com/2439881/226669310-58c13a40-de3e-4432-8b23-c9d3032ef889.mp4

It happens once on opening the publisher and then about 6 times or so at the end of publishing.

EmberLightVFX:
For me they don't flash as much as for you. This is how they look for me:

https://user-images.githubusercontent.com/49758407/226671061-7a5b578d-5975-4344-827f-2f993bcf3337.mp4

BigRoy:
> For me they don't flash as much as for you. This is how they look for me:

That looks the same to me because I was rendering two savers/renders but you're rendering a single instance so the post-render ones would pop-up about half as much so three times instead of six.

I also noticed the pop-up in yours on the opening of the publisher.

PR #4483, opened 2023-02-17
Ftrack: Upload reviewables with original name

iLLiCiTiT:
## Brief description
Ftrack can integrate reviewables with original filenames.

## Description
As ftrack have restrictions about names of components the only way how to achieve the result was to upload the same file twice, one with required name and one with origin name.

## Additional info
~~We could try "reuse" the same container id from the required file so the file does not have to uploaded twice.~~ I couldn't figura out how it could be possible.

## Testing notes:
Make sure you have enabled ftrack
1. Go to `project_settings/ftrack/publish/IntegrateFtrackInstance/upload_reviewable_with_origin_name` and enable it
2. Publish something with reviewable (e.g. trough tray publisher)
3. Check ftrack where should be on version `ftrackreview-mp4` component and component with original filename prepared to be downloaded

### DIFF
--- a/openpype/modules/ftrack/plugins/publish/integrate_ftrack_instances.py
+++ b/openpype/modules/ftrack/plugins/publish/integrate_ftrack_instances.py
@@ -56,6 +56,7 @@ class IntegrateFtrackInstance(pyblish.api.InstancePlugin):
         "reference": "reference"
     }
     keep_first_subset_name_for_review = True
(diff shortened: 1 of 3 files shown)


64qam (approved):

Minkiu (approved):

tokejepsen (approved):
Tested with review from Maya and downloads from components list and client review.

PR #4485, opened 2023-02-17
Harmony: missing set of frame range when opening scene

kalisp:
## Brief description
Frame range gets set from DB everytime scene is opened.

## Description
Added also check for not up-to-date loaded containers.

## Testing notes:
1. open scene in Harmony and check if frame duration matches to DB (Harmony starts from 1, so only duration is set
2. load older version of published asset


Minkiu (approved):
Looking at the comment, we haven't changed anything to fix the server client communication? 

Is moving them lower the fix?

64qam (approved):

PR #4488, opened 2023-02-20
Maya: Fix setting scene fps with float input

tokejepsen:
## Brief description
Returned value of float fps on integer values would return float.

## Description
This PR fixes the case when switching between integer fps values for example 24 > 25. Issue was when setting the scene fps, the original float value was used which makes it unpredictable whether the value is float or integer when mapping the fps values.

## Testing notes:
1. Set `project_anatomy/attributes/fps` to 24.
2. Launch Maya and save the workfile as 24 fps.
3. Set `project_anatomy/attributes/fps` to 25.
4. Publish or relaunch the workfile, and fix the fps validation.

### DIFF
--- a/openpype/hosts/maya/api/lib.py
+++ b/openpype/hosts/maya/api/lib.py
@@ -1998,7 +1998,7 @@ def set_scene_fps(fps, update=True):
         '48000': '48000fps'
     }
 
-    unit = fps_mapping.get(str(fps), None)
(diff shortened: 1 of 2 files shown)


antirotor (approved):

LiborBatek (approved):
All works fine. When FPS mismatch happens validator kicks in and Repair function fix the framerate so it all matches.

PR #4489, opened 2023-02-20
Deadline: Only submit version when running from build.

tokejepsen:
## Brief description
Only submit version when running from build.

## Description
When running from source code, we dont need a specific version of OP on the farm.

## Testing notes:
1. Run OP from source.
2. Submit a render to Deadline.
3. Validate no `OPENPYPE_VERSION` are in the jobs.


LiborBatek (approved):
Submitted ```Render Job``` to Deadline and checking ```Jobs Params``` and no ```OpenPype Version``` entry present.
image

All works as expected.

moonyuet (approved):
the code looks okay and I also tested it by myself. It works fine too.

antirotor (approved):

PR #4492, opened 2023-02-20
General: Support for extensions filtering in loaders

iLLiCiTiT:
## Brief description
Added extensions filtering support to loader plugins.

## Description
To avoid possible backwards compatibility break is filtering exactly the same and filtering by extensions is enabled only if class attribute 'extensions' is set.

## DCC with updated loaders
- [x] Nuke
- [x] Fusion
- [x] Hiero
- [x] Resolve
- [x] Flame
- [x] Harmony

## Testing notes:
1. Set extensions filter on loaders that can use that option
2. Open Loader tool where is the loader plugins available (in host or library loader)
3. Try if the loader will be shown


openpype/pipeline/load/plugins.py
```diff
@@ -109,16 +163,31 @@ def is_compatible_loader(cls, context):
...
         if maj_version < 3:
-            families = context["version"]["data"].get("families", [])
+            families = verison_doc["data"].get("families")
         else:
-            families = context["subset"]["data"]["families"]
+            families = context["subset"]["data"].get("families")
+            if families is None:
+                family = context["subset"]["data"].get("family")
+                if family:
+                    families = [family]
```
BigRoy, line 182: Personally I think the old style code was more readable - but maybe this is fixing something too?
Anyway, let's make sure to at least fix typo `verison` to `version`

openpype/pipeline/load/plugins.py
```diff
@@ -109,16 +163,31 @@ def is_compatible_loader(cls, context):
...
 
         plugin_families = set(plugin_families)
         return (
             "*" in plugin_families
-            or any(family in plugin_families for family in families)
+            or any(
+                family in plugin_families
+                for family in (families or [])
+            )
```
BigRoy, line 190: Was there a line length issue here? Don't see a reason why this needs to be split across multiple lines if it wasn't.

BigRoy (commented):
Looks like I didn't commit my review. Doing now.

openpype/pipeline/load/plugins.py
```diff
@@ -82,20 +84,72 @@ def apply_settings(cls, project_settings, system_settings):
...
 
         Returns:
-            bool
+            bool: Is loader compatible for context.
         """
 
         plugin_repre_names = cls.get_representations()
         plugin_families = cls.families
-        if not plugin_repre_names or not plugin_families:
+        if not plugin_repre_names or not plugin_families or not cls.extensions:
```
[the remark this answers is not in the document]
BigRoy: Already resolved. ✅

openpype/pipeline/load/plugins.py
```diff
@@ -82,20 +84,72 @@ def apply_settings(cls, project_settings, system_settings):
...
+                cls.log.info(
+                    "Representation doesn't have known source of extension"
+                    " information."
+                )
+                return False
+
+            cls.log.info("Using legacy source of extension on representation.")
+            ext = os.path.splitext(path)[-1]
+            while ext.startswith("."):
+                ext = ext[1:]
```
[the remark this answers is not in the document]
iLLiCiTiT: It's taken from `os.path.splitext` so it will 100% have dot.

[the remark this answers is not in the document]
BigRoy: Ah I see. I thouht it was taken from the database - didn't read right.

Still, `ext = ext.lstrip(".")` would avoid the need for the while loop.

```python
>>> s = "...hello"
>>> print(s.lstrip("."))
hello
```

openpype/pipeline/load/plugins.py
```diff
@@ -82,20 +81,67 @@ def apply_settings(cls, project_settings, system_settings):
...
+
+        if "*" in cls.extensions:
+            return True
+
+        # Get representation main file extension from 'context'
+        repre_context = repre_doc.get("context") or {}
+        ext = repre_context.get("ext")
+        if not ext:
+            # Legacy way how to get extensions
+            path = repre_doc["data"].get("path")
```
BigRoy, line 103: Only open issue I think is [this comment](https://github.com/ynput/OpenPype/pull/4492#pullrequestreview-1311019748) related to I believe this line of code. Somehow `repre_doc` doesn't have `data` in some cases?

Which might need to make this:
```suggestion
            path = repre_doc.get("data", {}).get("path")
```

BigRoy (commented):
Nice code. Only the one issue mentioned by @kalisp I think?

openpype/pipeline/load/plugins.py
BigRoy: It would be good however to double check whether `data` is ever available in `representation` in the `context`. It might very well be that it's a much more minimalized representation than the full document. I have no idea, but it did struck me as odd that a representation wouldn't have a `data` entry at all.

iLLiCiTiT: We've checked and the representation in database has `"data"` so it must be something in the code.

openpype/hosts/fusion/plugins/load/actions.py
```diff
@@ -15,6 +15,7 @@ class FusionSetFrameRangeLoader(load.LoaderPlugin):
                 "pointcache",
                 "render"]
     representations = ["*"]
+    extensions = {"*"}
```
BigRoy, line 18: This technically is a redundant addition, no? Since it defaults to that value.

openpype/hosts/fusion/plugins/load/load_sequence.py
```diff
@@ -129,6 +132,9 @@ class FusionLoadSequence(load.LoaderPlugin):
 
     families = ["imagesequence", "review", "render", "plate"]
     representations = ["*"]
+    extensions = set(
+        ext.lstrip(".") for ext in IMAGE_EXTENSIONS
+    )
```
BigRoy, line 137: Even though not entirely stable the Fusion Loaders do support video formats to as far as I know. We rarely use that however so I'm not sure it's needed here. Just wanted to mention it here.

@EmberLightVFX what do you think?

EmberLightVFX: I could see only allowing to load sequences as a problem. 

We sometimes work with r3d files directly.
Our stock footage library contains a lot of video files that we haven't converted to exr yet, and honestly probably won't.

jakubjezek001: I see, I was looking into the code and there was not mention of video files, also the fine name suggested image sequences. I can of course add video extension support too.

openpype/hosts/fusion/plugins/load/actions.py
jakubjezek001: Yes, it is but  I would rather keep it there for someone to quickly see what it suppose to support, without tracing inheritance.

jakubjezek001 (approved):
works well from Nuke, Hiero, Resolve, Flame, Fusion perspective.

kalisp (approved):

PR #4416, opened 2023-02-02
Maya: tile assembly fail in draft - OP-4820

tokejepsen:
## Brief description
Tile assembly in Deadline was broken.

## Description
Initial bug report revealed other areas of the tile assembly that needed fixing.

## Testing notes:
1. Update OpenPypeTileAssembler in Deadline.
2. Create render in Maya.
3. Enable `Tile Rendering` on the render objectset.
4. Publish to Deadline.

PR #4493, opened 2023-02-20
Deadline: Fix frames issues.

tokejepsen:
## Brief description
Fix frames issues.

## Description
When using start frames that are not 1 for example 1001, Deadline publishing errors.

## Testing notes:
1. Submit render with 1001 start frame to Deadline.
2. Validate publish is successful.


tokejepsen:
Not urgent, so will be part of #4416

PR #4494, opened 2023-02-21
Nuke: adding nukeassist 

jakubjezek001:
## Brief description
Adding support for NukeAssist

## Description
For support of NukeAssist we had to limit some Nuke features since NukeAssist itself Nuke with limitations. We do not support Creator and Publisher. User can only Load versions with version control. User can also set Framerange and Colorspace.

## Testing notes:
1. Set NukeAssist to your project
2. Open NukeAssist on task and see the OpenPype menu items have are redefined.

### DIFF
--- a/openpype/hosts/nuke/api/pipeline.py
+++ b/openpype/hosts/nuke/api/pipeline.py
@@ -60,7 +60,7 @@ from .workio import (
     work_root,
     current_file
 )
-from .utils import ASSIST
+from .constants import ASSIST
 
 log = Logger.get_logger(__name__)
 
--- a/openpype/hosts/nuke/api/pipeline.py
+++ b/openpype/hosts/nuke/api/pipeline.py
@@ -208,22 +208,25 @@ def get_workfile_build_placeholder_plugins():
 
 
 def _install_menu():
+    """Install Avalon menu into Nuke's main menu bar."""
 
     # uninstall original avalon menu
     main_window = get_main_window()
     menubar = nuke.menu("Nuke")
     menu = menubar.addMenu(MENU_LABEL)
 
-    label = "{0}, {1}".format(
-        os.environ["AVALON_ASSET"], os.environ["AVALON_TASK"]
-    )
-    Context.context_label = label
-    context_action = menu.addCommand(label)
 
     if not ASSIST:
+        label = "{0}, {1}".format(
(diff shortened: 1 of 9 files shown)


openpype/hosts/nuke/api/pipeline.py
```diff
@@ -218,7 +219,9 @@ def _install_menu():
     )
     Context.context_label = label
     context_action = menu.addCommand(label)
-    context_action.setEnabled(False)
+
+    if not ASSIST:
```
iLLiCiTiT, line 223: The context action probably should not be added at all when in assist.
```
if not ASSIST:
    context_action = menu.addCommand(label)
    context_action.setEnabled(False)
    menu.addSeparator()
```

openpype/hosts/nuke/hooks/pre_nukeassist_setup.py
```diff
@@ -0,0 +1,10 @@
+from openpype.lib import PreLaunchHook
+
+class PrelaunchNukeAssistHook(PreLaunchHook):
+    """
+    Adding flag when nukeassist
+    """
+    app_groups = ["nukeassist"]
+
+    def execute(self):
+        self.launch_context.env["NUKEASSIST"] = "1"
```
iLLiCiTiT, line 10:
```suggestion
        self.launch_context.env["NUKEASSIST"] = "1"

```

openpype/hosts/nuke/hooks/pre_nukeassist_setup.py
```diff
@@ -0,0 +1,10 @@
+from openpype.lib import PreLaunchHook
+
+class PrelaunchNukeAssistHook(PreLaunchHook):
```
iLLiCiTiT, line 3:
```suggestion


class PrelaunchNukeAssistHook(PreLaunchHook):
```

jakubjezek001: what is the difference? I am perhaps blind, but it seems to be the same.

openpype/hosts/nuke/api/pipeline.py
jakubjezek001: I believe only issue is that we are not able to disable the menu item. Not to create one.

PR #4496, opened 2023-02-21
TVPaint: Auto-detect render creation

iLLiCiTiT:
## Brief description
Create plugin which will create Render Layer and Render Pass instances based on information in the scene.

## Description
Added new creator that must be triggered by artist. The create plugin will first create Render Layer instances if were not created yet. For variant is used color group name. The creator has option to rename color groups by template defined in settings -> Template may use index of group by it's usage in scene (from bottom to top). After Render Layers will create Render Passes. Render Pass is created for each individual TVPaint layer in any group that had created Render Layer. It's name is used as variant (pass).

## Testing notes:
1. Open TVPaint
2. Create layers in scene
3. Mark some with color groups and some don't
4. Open Publisher
5. Select **Auto detect renders** plugin
- you can change some options there which may affect how will create instances
7. Hit `Create`
8. Validate created instances

The create plugin has settings that may change how renaming of groups work or if is even allowed.

There are many other options how this can be tested even I couldn't think about.


openpype/hosts/tvpaint/plugins/create/create_render.py
```diff
@@ -604,6 +607,356 @@ def get_instance_attr_defs(self):
...
+    This is auto-detection creator which can be triggered by user to create
+    instances based on information in scene. Each used color group in scene
+    will be created as Render Layer where group name is used as variant and
+    each TVPaint layer as Render Pass where layer name is used as variant.
+
+    Never will have any instances, all instances belong to different creators.
+    """
+
+    family = "render"
+    label = "Auto detect renders"
```
tokejepsen, line 622:
```suggestion
    label = "Render Layer/Passes"
```
This would need to be renamed throughout if you think its a good name.

openpype/settings/defaults/project_settings/tvpaint.json
```diff
@@ -40,6 +40,13 @@
             "mark_for_review": true,
             "default_variant": "Main",
             "default_variants": []
+        },
+        "auto_detect_render": {
+            "enabled": false,
```
tokejepsen, line 45: The state of the plugin is not respected in the publisher, ei. I've got it on default `false` and it still appears in the publisher.
Dont think we even need it to be togglable, since its a choice for the user to use the creator in the publisher.

openpype/hosts/tvpaint/plugins/create/create_render.py
iLLiCiTiT: I agree it should be renamed. Don't know if the suggested is good.

openpype/settings/defaults/project_settings/tvpaint.json
iLLiCiTiT: > The state of the plugin is not respected in the publisher, ei. I've got it on default false and it still appears in the publisher.

Issue of `CreateContext`. Will fix in different PR

> Dont think we even need it to be togglable, since its a choice for the user to use the creator in the publisher.

You're probably right

iLLiCiTiT: > Will fix in different PR

Created PR https://github.com/ynput/OpenPype/pull/4500

openpype/hosts/tvpaint/plugins/create/create_render.py
iLLiCiTiT: I just realized I didn't add description for user. So the label could be enough if there would be the description.

openpype/hosts/tvpaint/plugins/create/create_render.py
```diff
@@ -86,6 +89,25 @@
 )
 
 
+AUTODETECT_RENDER_DETAILED_DESCRIPTION = (
+    """Semi-automated Render Layer and Render Pass creation.
```
tokejepsen, line 93:
```suggestion
"""Semi-automated Render Layer and Render Pass creation.
```

tokejepsen (approved):

openpype/hosts/tvpaint/plugins/create/create_render.py
iLLiCiTiT: Hound didn't like it that way :)

tokejepsen: Bad doggy!

PR #4497, opened 2023-02-21
Maya: Multipart fix

tokejepsen:
## Brief description
Fix multipart logic in render products.

## Description
Each renderer has a different way of defining whether output images is multipart, so we need to define it for each renderer. Also before the `multipart` class variable was defined multiple times in several places, which made it tricky to debug where `multipart` was defined. Now its created on initialization and referenced as `self.multipart`

## Testing notes:
1. Create render in Maya with a supported renderer that support multipart image output (EXR).
2. Publish and inspect the produced json to confirm no AOV image outputs are expected.


BigRoy:
I have little knowledge about this. So a little question.

Isn't it the case that some renderers can have both multipart and non-multipart outputs from a single RenderLayer?

E.g. doesn't Redshift always separate Cryptomatte output into a render product of its own which itself is not a multipart file?

I know e.g. Arnold has some crazy per AOV driver settings in which you can override almost anything per AOV. Does that include toggling multipart for certain AOVs?

tokejepsen:
> I know e.g. Arnold has some crazy per AOV driver settings in which you can override almost anything per AOV. Does that include toggling multipart for certain AOVs?

This PR does not include logic for this. I would imagine when the need arises to repurpose the `get_multipart` for the specific renderer into taking AOV drivers into account so you could change the multipart per render product later in the execution.
As it stands now, the logic is broken cause the multipart data member is defined as `False` on initialization, which generates expected files that do not get rendered and hence the publishing breaks.
This originated from testing Vray.

LiborBatek:
for arnold with ```multipart``` on for ```.exr``` seems its working as expected. No trace of any ```AOV``` and only ```Main.####.exr``` present.

image

m-u-r-p-h-y (commented):
tested in Maya2023 and Vray and works as expected

image


did a quick test with Redshift and it seems off, so investigating . . .

Minkiu (approved):

openpype/hosts/maya/api/lib_renderproducts.py
```diff
@@ -1059,7 +1074,7 @@ def get_render_products(self):
                 continue
 
             aov_type = self._get_attr(aov, "aovType")
-            if multipart and aov_type not in self.unmerged_aovs:
+            if self.multipart and aov_type not in self.unmerged_aovs:
```
moonyuet, line 1077:
```suggestion
            multipart = self.get_multipart()
            if multipart  and aov_type not in self.unmerged_aovs:
```
If I dont get it wrong...

openpype/hosts/maya/api/lib_renderproducts.py
```diff
@@ -1108,7 +1123,7 @@ def get_render_products(self):
                 product = RenderProduct(productName=aov_name,
                                         aov=aov_name,
                                         ext=ext,
-                                        multipart=multipart,
+                                        multipart=self.multipart,
```
moonyuet, line 1126:
```suggestion
                                        multipart=multipart,
```

openpype/hosts/maya/api/lib_renderproducts.py
```diff
@@ -1094,7 +1109,7 @@ def get_render_products(self):
                             productName=aov_light_group_name,
                             aov=aov_name,
                             ext=ext,
-                            multipart=multipart,
+                            multipart=self.multipart,
```
moonyuet, line 1112:
```suggestion
                            multipart=multipart,
```

openpype/hosts/maya/api/lib_renderproducts.py
```diff
@@ -1124,7 +1139,7 @@ def get_render_products(self):
             products.insert(0,
                             RenderProduct(productName=beauty_name,
                                           ext=ext,
-                                          multipart=multipart,
+                                          multipart=self.multipart,
```
moonyuet, line 1142:
```suggestion
                                          multipart=multipart,
```

openpype/hosts/maya/api/lib_renderproducts.py
antirotor: > If I dont get it wrong...

I think:

```python
multipart = self.get_multipart()
```
is already done in the constructor of the abstract base class.

antirotor:
> I know e.g. Arnold has some crazy per AOV driver settings in which you can override almost anything per AOV. Does that include toggling multipart for certain AOVs?

currently we don't even support per-AOV prefix. In some renderers you can change the naming of files in individual AOVs. But I don't think that we should ever support it as I consider it from pipeline point of view as an anti-pattern.

openpype/hosts/maya/api/lib_renderproducts.py
moonyuet: i see. i will resume my statement.

tokejepsen:
> did a quick test with Redshift and it seems off, so investigating . . .

@m-u-r-p-h-y could you give Redshift another go now? Think I've fixes the expected files.

openpype/hosts/maya/api/lib_renderproducts.py
moonyuet: > @moonyuet sorry I'm not sure what this does?

just leaves this comment alone, I didn't notice there is the constructor of the base class done for the `multipart.`

m-u-r-p-h-y (changes requested):
Expected files in Redshift are not right imo.

image

it would be ok if `redshiftOptions.exrForceMultilayer` was 0, but it is not . . .

tokejepsen:
> Expected files in Redshift are not right imo.

@m-u-r-p-h-y good spot! I dont know what `multipart EXR` really does then, cause it seems the forcing option is the only way to get it properly multipart.

tokejepsen:
@m-u-r-p-h-y think we have it now. Give it another run.

m-u-r-p-h-y (approved):
looks perfect now!

main should contain all aov's and separate crypto 

image

openpype/hosts/maya/api/lib_renderproducts.py
```diff
@@ -196,12 +196,18 @@ def __init__(self, layer, render_instance):
...
         self.render_instance = render_instance
-        self.multipart = False
+        self.multipart = self.get_multipart()
 
         # Initialize
         self.layer_data = self._get_layer_data()
         self.layer_data.products = self.get_render_products()
 
+    def get_multipart(self):
+        raise NotImplementedError(
```
Minkiu, line 206: After some more though, maybe we could do:
```py
self.log.warn(
   "The render product implementation does not have a "
   "\"get_multipart\" method."
)
return False
```

that way we don't need to do https://github.com/ynput/OpenPype/pull/4497/files#diff-3b0eed55b2558684bcb1a84202dc932177842e1443bf2228020e1120305c57f0R1179

openpype/hosts/maya/api/lib_renderproducts.py
```diff
@@ -1283,6 +1319,9 @@ class RenderProductsMayaHardware(ARenderProducts):
         {"label": "EXR(exr)", "index": 40, "extension": "exr"}
     ]
 
+    def get_multipart(self):
+        return False
```
tokejepsen, line 1323:
```suggestion
    def get_multipart(self):
        # MayaHardware does not support multipart EXRs.
        return False
```

mkolar:
Thank you gentlemen. Merging

PR #4499, opened 2023-02-22
TVPaint: Ignore transparency in Render Pass

iLLiCiTiT:
## Brief description
It is possible to ignore layers transparency during Render Pass extraction.

## Description
Render pass extraction does not respect opacity of TVPaint layers set in scene during extraction. It can be enabled/disabled in settings.

## Additional info
I found out that `opacity` received using `tv_layerinfo` is always `0` which caused a troubles but `tv_layerdensity` does return previous value on change so I've used that option to change it to `100` and then back to previous value.

image

## Testing notes:
1. Enable `project_settings/tvpaint/publish/CollectRenderInstances/ignore_render_pass_transparency`
2. Open TVPaint scene
3. Create Render Layer
4. Create Render Pass on TVPaint layer
5. Change opacity of the layer (e.g. to 50%)
6. Publish
7. Output of Render Pass should have 100%

### DIFF
--- a/openpype/hosts/tvpaint/plugins/create/create_render.py
+++ b/openpype/hosts/tvpaint/plugins/create/create_render.py
@@ -660,7 +660,6 @@ class TVPaintAutoDetectRenderCreator(TVPaintCreator):
             ["create"]
             ["auto_detect_render"]
         )
-        self.enabled = plugin_settings["enabled"]
         self.allow_group_rename = plugin_settings["allow_group_rename"]
         self.group_name_template = plugin_settings["group_name_template"]
         self.group_idx_offset = plugin_settings["group_idx_offset"]
--- a/openpype/hosts/tvpaint/plugins/publish/collect_render_instances.py
+++ b/openpype/hosts/tvpaint/plugins/publish/collect_render_instances.py
@@ -9,6 +9,8 @@ class CollectRenderInstances(pyblish.api.InstancePlugin):
     hosts = ["tvpaint"]
     families = ["render", "review"]
 
+    ignore_render_pass_transparency = False
+
     def process(self, instance):
         context = instance.context
         creator_identifier = instance.data["creator_identifier"]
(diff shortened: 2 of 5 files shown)


tokejepsen (changes requested):
Think the branch needs updating?
```
  File "C:\Users\tokejepsen\OpenPype\openpype\tools\publisher\window.py", line 456, in _on_show_timer
    self.reset()
  File "C:\Users\tokejepsen\OpenPype\openpype\tools\publisher\window.py", line 462, in reset
    self._controller.reset()
  File "C:\Users\tokejepsen\OpenPype\openpype\tools\publisher\control.py", line 1762, in reset
    self._reset_plugins()
  File "C:\Users\tokejepsen\OpenPype\openpype\tools\publisher\control.py", line 1780, in _reset_plugins
  File "C:\Users\tokejepsen\OpenPype\openpype\pipeline\create\context.py", line 1623, in reset_plugins
    self._reset_creator_plugins()
  File "C:\Users\tokejepsen\OpenPype\openpype\pipeline\create\context.py", line 1703, in _reset_creator_plugins
    creator = creator_class(
  File "C:\Users\tokejepsen\OpenPype\openpype\pipeline\create\creator_plugins.py", line 210, in __init__
  File "C:\Users\tokejepsen\OpenPype\openpype\hosts\tvpaint\plugins\create\create_render.py", line 663, in apply_settings
    self.enabled = plugin_settings["enabled"]
KeyError: 'enabled'
```

iLLiCiTiT:
> Think the branch needs updating?

It's actually in develop?

tokejepsen (approved):
Works!

A bit OT but I find refreshing "unresponsive". Would be nice to have some initial indicator that the publisher is refreshing.

[Untitled_ Feb 23, 2023 10_01 AM.webm](https://user-images.githubusercontent.com/1860085/220876024-9213ee1e-37c1-4dc1-b1c2-34a8dcd2f58e.webm)

Also when repairing the UI does not indicate any progress, success or fail.

repair

Lastly I was wondering about the case for this enhancement? I was thinking it would be better to inherit the transparency override when creating the render pass and the user will have final say.

transparency

iLLiCiTiT:
> Lastly I was wondering about the case for this enhancement? I was thinking it would be better to inherit the transparency override when creating the render pass and the user will have final say.

From the request it is not an artist decision if the transparency is ignored or not.

PR #4501, opened 2023-02-22
TVPaint: Small enhancements

iLLiCiTiT:
## Brief description
Small enhancements in TVPaint integration which did not get to https://github.com/ynput/OpenPype/pull/4471.

## Description
It was found out that `opacity` returned from `tv_layerinfo` is always empty and is dangerous to add it to layer information. Added information about "current" layer to layers information. Disable review of Render Layer and Render Pass instances by default. In most of productions is used only "scene review". Skip usage of `"enabled"` key from settings in automated layer/pass creation.

## Testing notes:
1. Start TVPaint
2. Open Publisher
3. Render Layer and Render Pass creator should have `Mark for review` set to `False` by default
(Nothing should crash meanwhile)


tokejepsen (approved):
Works!

OT when creating a renderlayer followed by a renderpass you get an error message. I get why since the renderlayer is not persisted to TVP, but we should maybe inform the user about trying to refresh the publisher.

[Untitled_ Feb 23, 2023 10_15 AM.webm](https://user-images.githubusercontent.com/1860085/220879521-d68eaa36-2a2d-4c6a-b6f9-b470c524fd00.webm)

mkolar:
> Works!
> 
> OT when creating a renderlayer followed by a renderpass you get an error message. I get why since the renderlayer is not persisted to TVP, but we should maybe inform the user about trying to refresh the publisher.

I think we should also give more info in the message. `You cannot create a renderPass without a renderLayer. Please select one first`

iLLiCiTiT:
@tokejepsen added label to precreate attributes. Changed error message to @mkolar suggestion. And changed how Render Layers enum values are collected for precreate attributes.

tokejepsen:
> @tokejepsen added label to precreate attributes. Changed error message to @mkolar suggestion. And changed how Render Layers enum values are collected for precreate attributes.

Nice works for me!

PR #4503, opened 2023-02-22
global: source template fixed frame duplication

jakubjezek001:
## Brief description
Duplication is not happening.

## Description
Template is using `originalBasename` which already assume all necessary elements are part of the file name so there was no need for additional optional name elements. 

## Testing notes:
After you resaved project settings and made sure no project override is on `project_setitngs/anatomy/tempaltes/others/source`, follow testing notes from https://github.com/ynput/OpenPype/pull/4453


jrsndl (approved):
Works as advertised.

PR #4506, opened 2023-02-22
General: Default settings stored with newline

iLLiCiTiT:
## Brief description
Save default settings with newline character at the end of file.

## Description
Missing new line character at the end of settings default values is causing some issues especially now when pre-commit hook is validating it.

## Testing notes:
Nothing should really change.
1. Open settings UI in develop mode
2. Check "Modify defaults" checkbox
3. Change some values
4. Save defaults
5. Saved files should contain only changes of the values and not new line character change


kalisp (approved):
No change at the end of the file with defaults.

PR #2157, opened 2021-10-20
OpenPypeV3: add key task type, task shortname and user to path templating construction

ClementHector:
following: [old PR](https://github.com/pypeclub/OpenPype/pull/2119)

PR #4510, opened 2023-02-22
Create new variable for asset type from ftrack

friquette:
## Brief description
Create a new variable named asset[type] for templates in project settings. This variable is based on AssetBuild type from ftrack.

### DIFF
--- a/openpype/modules/ftrack/lib/avalon_sync.py
+++ b/openpype/modules/ftrack/lib/avalon_sync.py
@@ -1935,6 +1935,17 @@ class SyncEntitiesFactory:
                         avalon_id
                     )
                 )
+
+            # Add asset build type to entity
+            asset_build_type = None
+            if self.entities_dict[ftrack_id]["entity_type"] == "assetbuild":
+                asset_build_type = self.entities_dict[ftrack_id][
+                    'entity']['type']['name']
+
+                if asset_build_type:
+                    self.entities_dict[ftrack_id]['final_entity']['data'][
+                        'asset_type'] = asset_build_type
+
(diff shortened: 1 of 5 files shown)


iLLiCiTiT:
Hi, thanks for the PR, we would like to make this change too, but the change is not as easy as just these lines. All templates and places in code, that are using these data, must be prepared for this data structure change first, before any change of structure will happen. All the code must be backwards and forwards compatible so it does not matter what data are passed in or what template is passed in. We did it the same way with tasks.

As you can imagine that is a lot more job then few lines of code. We will have to do it, but it will probably happen before AYON release. At this moment we don't have capacity to make the changes, or do review of those changes. This is massivelly affecting every aspect of pipeline which is crutial to work and is important for use to not break productions.

If you would like to handle only the Asset Build from ftrack, it could be possible. I didn't test it, or looked into it closer, because of the changes in template data, so I don't know if the changes are right.

If you have opinions, write them down. For now the PR can be considered as on hold, and may be closed over time.

ClementHector:
I found the PR for adding tasks, even if it is dated, it gives an idea of the work to be done.
https://github.com/ynput/OpenPype/pull/2157

mkolar:
This is really tricky. Cannot be merged in this state for the reasons state by Jakub and at the same time with AYON being almost done, it's a bit of a wasted effort.

We won't be able to give it resources to properly test all the scenarios this might break at the moment. Sorry @ClementHector but I don't see how we could get this in at this point.

ClementHector:
No problem, we'll keep that to ourselves before moving on to Ayon :)

PR #4511, opened 2023-02-22
Maya: custom namespace for references

friquette:
## Brief description
Adding an option in Project Settings > Maya > Loader plugins to set custom namespace. If no namespace is set, the default one is used.


openpype/hosts/maya/plugins/load/load_reference.py
```diff
@@ -36,7 +36,7 @@ class ReferenceLoader(openpype.hosts.maya.api.plugin.ReferenceLoader):
     # Name of creator class that will be used to create animation instance
     animation_creator_name = "CreateAnimation"
 
-    def process_reference(self, context, name, namespace, options):
+    def process_reference(self, context, name, namespace, options, group_name=None):  # noqa
```
BigRoy, line 39: If we're changing the signature of this method. Shouldn't we be doing so on the base class too? Plus `load_look` would need to be adjusted too? (Even thought it doesn't generate a group?) There's also a `_load_animation.py` which likely needs to be refactored too?

Those would likely get called from the base class too, no?
This PR might currently be breaking those loaders? (not entirely sure)

openpype/hosts/maya/api/plugin.py
```diff
@@ -153,6 +155,25 @@ def load(
                 suffix="_",
             )
 
+            custom_naming = settings['maya']['load']['reference_loader']['naming']  # noqa
```
BigRoy, line 158: Should this logic for custom naming be above the 'add a counter suffix' above?
What happens with custom naming enabled if you were to load e.g. a pointcache with count = 5? They'd each get the same namespace assigned since there's nothing ensuring uniqueness. Seems breaking?

BigRoy (commented):
Overall I think it's a nice feature to allow the customization, but I do have some questions/notes.

website/docs/admin_hosts_maya.md
```diff
@@ -107,6 +107,18 @@ or Deadlines **Draft Tile Assembler**.
 This is useful to fix some specific renderer glitches and advanced hacking of Maya Scene files. `Patch name` is label for patch for easier orientation.
 `Patch regex` is regex used to find line in file, after `Patch line` string is inserted. Note that you need to add line ending.
 
+## Load Plugins
+
+### Reference Loader > Namespace
+Here you can create your own custom naming for the reference loader. <br>
+The custom naming is split into two parts separated by a ":" symbol. The first half is the namespace and the second half is the group name of the reference. If you don't set the namespace or the group name (or both), the default namespace and group name will be applied.
```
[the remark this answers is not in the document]
friquette: So if I understand correctly, you want this field to be filled by default with the originally generated namespace, and not blank ?

openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_load.json
```diff
@@ -91,6 +91,23 @@
...
+                {
+                    "type": "text",
+                    "label": "Naming",
+                    "key": "naming"
+                },
+                {
+                    "type": "label",
+                    "label": "Here's a link to the doc where you can find explanations about customing the naming of referenced assets: https://openpype.io/docs/admin_hosts_maya#load-plugins"
+                }
+            ]
```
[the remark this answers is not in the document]
friquette: Thank you.  Once I lost some time searching the variables I could use for a specific settings, so I decided to put here a direct link to the doc to avoid people losing time :)

openpype/hosts/maya/plugins/load/load_reference.py
```diff
@@ -45,14 +45,19 @@ def process_reference(self, context, name, namespace, options):
...
         attach_to_root = options.get("attach_to_root", True)
 
         with maintained_selection():
             cmds.loadPlugin("AbcImport.mll", quiet=True)
             file_url = self.prepare_root_value(self.fname,
                                                context["project"]["name"])
+
+            if not cmds.namespace(exists=namespace):
+                cmds.namespace(add=namespace)
+
```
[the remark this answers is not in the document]
friquette: Sometimes when I referenced assets (particularly with the template feature) I had namespace issues. Basically, Maya was telling me that it couldn't reference the asset because the namespace didn't exist.

website/docs/admin_hosts_maya.md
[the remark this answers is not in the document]
BigRoy: Yes, the result will be the same but then it could _always_ use the value from settings. Seems more explicit to me. No? I wonder what the OP crew thinks however.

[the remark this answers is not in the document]
friquette: Yes I agree with you, it could even be simpler. Do I wait for feedback from OP crew or do I make the modification now ?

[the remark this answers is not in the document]
BigRoy: @iLLiCiTiT @mkolar any thoughts on this by any chance? (not sure who best to tag, sorry)

openpype/hosts/maya/plugins/load/load_reference.py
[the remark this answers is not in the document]
friquette: Sadly I can't seem to reproduce the error. If I remember correctly, this happened when I loaded a reference for the first time, Maya was telling me that the namespace wasn't existing so it couldn't load the asset. But after that, if I loaded the same asset a second time, it worked because the namespace was already in the scene thanks to the first time (I don't know if that's clear).
But all of that was on the early stages of the PR, so I don't think it comes from it. Anyway, I think this is a small adding which could possibly prevent errors with namespaces. Even if the error is rare, I think it's important to take it into account.

openpype/hosts/maya/plugins/load/load_reference.py
friquette: I don't think this is a problem since the process_reference method is to be implemented by subclasses. If you look at the base class, it is clearly said in `openpype/hosts/maya/api/plugin.py` lines 212 to 214:
```
def process_reference(self, context, name, namespace, data):
    """To be implemented by subclass"""
    raise NotImplementedError("Must be implemented by subclass")
```

openpype/hosts/maya/api/plugin.py
friquette: I'm not sure what you are saying here. All I can tell you is that if multiple references are loaded with the same namespace, Maya will automatically add a number at the end of the namespace to ensure uniqueness.

BigRoy: Different references should have a unique namespace even if referencing the same file. The custom naming logic doesn't ensure the namespace is unique and thus problems will ensue.

openpype/hosts/maya/plugins/load/load_reference.py
BigRoy: Yes. But the method signature shouldn't change. If we are changing the signature all child classes will need to follow.

openpype/hosts/maya/plugins/load/load_reference.py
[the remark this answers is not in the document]
BigRoy: Actually. I'd actually argue that we don't want to ever reference into an existing namespace. So I'd actually remove this if it works without unless there's a clear problem it's solving.

Funnily enough the `maya.cmds.file` documentation states this for the `namespace` argument:

> The new namespace will be created by this command and can not already exist. The old namespace will be removed.

More reason for me to think we should have no need of generating it first?

openpype/hosts/maya/api/plugin.py
```diff
@@ -143,14 +143,45 @@ def load(
...
         loaded_containers = []
 
         count = options.get("count") or 1
+
         for c in range(0, count):
-            namespace = namespace or lib.unique_namespace(
-                "{}_{}_".format(asset["name"], context["subset"]["name"]),
-                prefix="_" if asset["name"][0].isdigit() else "",
-                suffix="_",
+            custom_naming = settings['maya']['load']['reference_loader']['naming']  # noqa
```
iLLiCiTiT, line 153: This line can be moved before the loop.

openpype/hosts/maya/api/plugin.py
```diff
@@ -143,14 +143,45 @@ def load(
         assert os.path.exists(self.fname), "%s does not exist." % self.fname
 
         asset = context['asset']
+        subset = context['subset']
+        settings = get_project_settings(os.environ['AVALON_PROJECT'])
```
iLLiCiTiT, line 147:
```suggestion
        settings = get_project_settings(context["project"]["name"])
```

openpype/hosts/maya/api/plugin.py
```diff
@@ -143,14 +143,45 @@ def load(
...
+                    "Wrong format for namespace, \
+missing content before ':' separator"
+                )
+            elif custom_naming.strip()[-1] == ':':
+                raise ValueError(
+                    "Wrong format for namespace, \
+missing content after ':' separator"
+                )
+
+            custom_naming = custom_naming.format(
```
iLLiCiTiT, line 171: I don't think it's good idea to give option to use any data in asset document and subset document. There should be defined subset of keys which should ideally match the rest of current templates.
```
{
    "asset": asset["name"],
    "subset": subset["name"],
    "family": subset["data"].get("family") or subset["data"]["families"][0]
}
```

e.g. `asset["type"]` will always be `"asset"`. Who set the settings should not know that family is in `subset["data"]["family"]`.

tokejepsen (changes requested):
I'm a bit confused by the code and the PR title/description. We get custom namespace but it looks like you also want control over the naming of the group that some loaders use to parent new nodes under?

If you need control over the group name as well, I would think that is a valid feature as well but in either case we need some code changes. Let me know what you want and I can go over the code in detail with the functionality in mind.

website/docs/admin_hosts_maya.md
```diff
@@ -110,40 +106,28 @@ or Deadlines **Draft Tile Assembler**.
...
-- **Background Top** the viewports background top color.
-- **Override display options** override the viewports display options to use what is set in the settings.
-- **Isolate view** isolates the view to what is in the review instance. If only a camera is present in the review instance, all nodes are displayed in view.
-- **Off Screen** records the playblast hidden from the user.
-- **2D Pan/Zoom** enables the 2D Pan/Zoom functionality of the camera.
-- **Renderer name** which renderer to use for playblasting.
-- **Width** width of the output resolution. If this value is `0`, the asset's width is used.
-- **Height** height of the output resolution. If this value is `0`, the asset's height is used.
-
-#### Viewport Options
```
BigRoy, line 134: This looks like you accidentally deleted some settings on a merge conflict. I suspect these shouldn't be removed. Right? There some other changes in this file too that I suspect shouldn't be here either.

friquette: Oh yeah I just realized I removed some things that I shouldn't have ! I will revert my changes

openpype/hosts/maya/api/plugin.py
```diff
@@ -143,14 +143,50 @@ def load(
...
+            elif custom_naming.strip()[0] == ':':
+                raise ValueError(
+                    "Wrong format for namespace, missing content before ':' "
+                    "separator"
+                )
+            elif custom_naming.strip()[-1] == ':':
+                raise ValueError(
+                    "Wrong format for namespace, missing content after ':' "
+                    "separator"
+                )
```
BigRoy, line 169: Can we move this logic out of the for loop? This check can be done much higher already - nearer to where we retrieve the setting.

openpype/hosts/maya/api/plugin.py
```diff
@@ -143,14 +143,50 @@ def load(
...
+
+            custom_naming = custom_naming.format(
```
BigRoy, line 171: This is overwriting the `custom_naming` variable in the for loop. :) This doesn't look intended.
If the custom naming formatting is unrelated to the for loop I'd move it to outside of the for loop as well.

tokejepsen:
> Yes I also want control over the group name. I think controlling the group name is important because the default "_GRP" isn't really explicit, especially if you have multiple references in multiple groups.

In this case I think we are better off having separate namespace and group setting. This is to not have any implicit logic in the settings where namespace also has the group naming in it.

Also `XgenLoader` needs some attention for this feature.

friquette:
> > Yes I also want control over the group name. I think controlling the group name is important because the default "_GRP" isn't really explicit, especially if you have multiple references in multiple groups.
> 
> In this case I think we are better off having separate namespace and group setting. This is to not have any implicit logic in the settings where namespace also has the group naming in it.
> 
> Also `XgenLoader` needs some attention for this feature.

So do I need to modify my code to separate namespace and group name ? Create 2 separate settings in the maya > loader plugins > reference loader ?

tokejepsen:
> So do I need to modify my code to separate namespace and group name ? Create 2 separate settings in the maya > loader plugins > reference loader ?

Yes, please.

website/docs/admin_hosts_maya.md
```diff
@@ -106,6 +106,31 @@ or Deadlines **Draft Tile Assembler**.
 This is useful to fix some specific renderer glitches and advanced hacking of Maya Scene files. `Patch name` is label for patch for easier orientation.
 `Patch regex` is regex used to find line in file, after `Patch line` string is inserted. Note that you need to add line ending.
 
+## Load Plugins
+
+### Reference Loader > Namespace and Group Name
```
tokejepsen, line 111:
```suggestion
### Reference Loader

#### Namespace and Group Name
```

tokejepsen (changes requested):
This is looking good atm. Will give it a test later.

openpype/hosts/maya/plugins/load/load_yeti_rig.py
```diff
@@ -17,8 +17,7 @@ class YetiRigLoader(openpype.hosts.maya.api.plugin.ReferenceLoader):
     color = "orange"
 
     def process_reference(
-        self, context, name=None, namespace=None, options=None
-    ):
+            self, context, name=None, namespace=None, options=None, group_name=None):  # noqa
 
         group_name = "{}:{}".format(namespace, name)
```
tokejepsen, line 22:
```suggestion
```

tokejepsen (changes requested):

tokejepsen (approved):

tokejepsen:
@BigRoy you happy with this?

openpype/hosts/maya/api/plugin.py
```diff
@@ -143,16 +143,49 @@ def load(
...
-        count = options.get("count") or 1
-        for c in range(0, count):
-            namespace = namespace or lib.unique_namespace(
-                "{}_{}_".format(asset["name"], context["subset"]["name"]),
-                prefix="_" if asset["name"][0].isdigit() else "",
-                suffix="_",
+        if not custom_naming['namespace']:
+            raise ValueError("No namespace specified")
+        elif not custom_naming['group_name']:
+            raise ValueError("No group name specified")
```
BigRoy, line 154: Should this error message be more descriptive errors, like at least mention that it's based on project settings or alike? Or is this fine as is?

Also should this be `openpype.pipeline.load.LoadError`?

openpype/hosts/maya/api/plugin.py
```diff
@@ -143,16 +143,49 @@ def load(
...
+                subset['data'].get('family') or
+                subset['data']['families'][0]
             )
+        )
 
+        namespace = lib.unique_namespace(
+            custom_namespace,
+            prefix="_" if custom_namespace[0].isdigit() else "",
+            suffix=""
+        )
```
BigRoy, line 180: This logic **must** remain in the for loop below otherwise it's not ensured that each loaded entry gets a unique namespace.

openpype/hosts/maya/api/plugin.py
```diff
@@ -187,7 +221,7 @@ def load(
 
         return loaded_containers
 
-    def process_reference(self, context, name, namespace, data):
+    def process_reference(self, context, name, namespace, data, group_name=None):  # noqa
```
BigRoy, line 224: I wanted to stress again that this could be a backwards incompatible change for any studio (or client?) which might have developed additional Loader plug-ins for Maya which inherit from this class because those will need to be refactored to also take this additional arguments.

A potential backwards compatible way would have been to maybe put the group name into the `data` dictionary as additional entry. Also, `data` here should be renamed to `options`. It was wrong before apparently but _luckily?_ was overridden on each inheriting plug-in to be `options` instead. With the argument being called `data` the call on line 200 (before line 167) wouldn't work because it explicitly passes `options` as a keyword.

The backwards compatible `_load_animation` plug-in would need to also be refactored `data` -> `options`. With it being broken currently I have a feeling that the old load animation hasn't been in use for quite some time otherwise we'd have seen this reported maybe?

openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_load.json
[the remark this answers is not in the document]
BigRoy: Still not sure if we should be adding this label like this. As far as I know no other settings do this either so I'm wondering whether it fits the "style" of OpenPype to start doing this now. @mkolar @iLLiCiTiT thoughts?

BigRoy (changes requested):
Very nice work, but unfortunately some issues still stood out. In particular `data` keyword argument to `options` and the unique namespace must be in the for loop. The rest of the notes are merely questions, not requirements.

openpype/hosts/maya/api/plugin.py
tokejepsen: Yeah, lets get it more descriptive.

@iLLiCiTiT do you have an input on what error to use?

openpype/hosts/maya/api/plugin.py
tokejepsen: > A potential backwards compatible way would have been to maybe put the group name into the data dictionary as additional entry. Also, data here should be renamed to options. It was wrong before apparently but luckily? was overridden on each inheriting plug-in to be options instead. With the argument being called data the call on line 200 (before line 167) wouldn't work because it explicitly passes options as a keyword.

Yup good point. Lets do that.

> The backwards compatible _load_animation plug-in would need to also be refactored data -> options. With it being broken currently I have a feeling that the old load animation hasn't been in use for quite some time otherwise we'd have seen this reported maybe?

Pretty sure any Pyblish plugins starting with `_` will be ignored, so it should actually be removed.

BigRoy: I think it might be there to allow 'managing' any loaded content which had previously been loaded using that loader - but not sure. :)

friquette: @BigRoy So if I understand correctly, you want to rename the argument `data` to `options` in the different `process_reference` methods?
And also pass the `group_name` as a new entry in this `options` so we don't need to refactor all classes that inherit this one? But the refactoring is inevitable because even if we pass the `group_name` as new entry, we need to modify all chidren classes to get the group name from the options, don't we?

BigRoy: > So if I understand correctly, you want to rename the argument data to options in the different process_reference methods?

Yes. This we should do either way since it's basically been invalid before.

> And also pass the group_name as a new entry in this options so we don't need to refactor all classes that inherit this one?

Yes. This is what I'm proposing. Whether we should do that maybe @tokejepsen and @antirotor have an opinion on the matter.

> But the refactoring is inevitable because even if we pass the group_name as new entry, we need to modify all chidren classes to get the group name from the options, don't we?

Sort of. Without refactoring it would not use the predefined group name, but at the same time it also wouldn't error. It would then at least not break existing custoom plugins someone might have in their studio.

tokejepsen: > Yes. This is what I'm proposing. Whether we should do that maybe @tokejepsen and @antirotor have an opinion on the matter.

@BigRoy Yeah, I think this is the best route to take. Wasnt keen on the idea of adding the extra keyword argument and its a good point about external people might be using it.

> But the refactoring is inevitable because even if we pass the group_name as new entry, we need to modify all chidren classes to get the group name from the options, don't we?

@friquette True we'll need to refactor anyways for this feature, but studios wont need to refactor if they use the method. They'll need to refactor if they want to use this feature, but that is expected for new features.

openpype/hosts/maya/api/plugin.py
iLLiCiTiT: The difference is how it's displayed to user. `LoadError` should show only the error message without traceback, so `LoadError` is suggested.

tokejepsen:
```suggestion
        if not custom_naming['namespace']:
            raise LoadError("No namespace specified")
        elif not custom_naming['group_name']:
            raise LoadError("No group name specified")
```

openpype/hosts/maya/api/plugin.py
```diff
@@ -143,15 +144,52 @@ def load(
...
+
+        custom_group_name = custom_naming['group_name'].format(
+            asset_name=asset['name'],
+            asset_type=asset['type'],
+            subset=subset['name'],
+            family=(
+                subset['data'].get('family') or
+                subset['data']['families'][0]
+            )
+        )
```
BigRoy, line 177: Not saying this is better - just wondering whether we should do this to avoid defining the formatting data twice. I personally don't mind the current one either.
```suggestion

        formatting_data = {
            "asset_name": asset["name"],
            "asset_type": asset["type"],
            "subset": subset["name"],
            "family": (
                subset['data'].get('family') or
                subset['data']['families'][0]
            )
        }
        custom_namespace = custom_naming['namespace'].format(
            **formatting_data
        )
        custom_group_name = custom_naming['group_name'].format(
            **formatting_data
        )
```

openpype/hosts/maya/api/plugin.py
```diff
@@ -143,15 +144,52 @@ def load(
...
+
         for c in range(0, count):
-            namespace = namespace or lib.unique_namespace(
-                "{}_{}_".format(asset["name"], context["subset"]["name"]),
-                prefix="_" if asset["name"][0].isdigit() else "",
-                suffix="_",
+            namespace = lib.unique_namespace(
+                custom_namespace,
+                prefix="_" if custom_namespace[0].isdigit() else "",
+                suffix=""
```
BigRoy, line 185: With the `"_"` not being set in the suffix does that mean that after this PR the naming convention will be slightly different? I'm personally fine with it being removed but it does mean that it'll change the behavior. So I just wanted to highlight that.

BigRoy (approved):
Solid work. I did have two notes that might still be worth checking but overall the code changes seem fine. There is one remark I made about backwards compatibility which @tokejepsen should maybe double check whether it's worth the change or not.

openpype/hosts/maya/api/plugin.py
tokejepsen: Yeah, nice. @friquette you can accept the suggestion here if you want to.

openpype/hosts/maya/api/plugin.py
tokejepsen: This is actually a good point. For reference I get `ball_rigMain01:_GRP` with this PR, whereas it was `ball_rigMain_01_:_GRP` before.
I wonder whether we should add another token called `amount` or `identifier` (dont like either of those names), so studios can control where the number gets inserted?
We could get a completely backwards compatible code plus added control.

openpype/hosts/maya/api/plugin.py
friquette: Yes this is a good idea, I will implement it !

openpype/hosts/maya/api/plugin.py
friquette: Can't we just do something like this ?
 ```
if custom_namespace[-1] != "_":
    custom_namespace += "_"
```
Let the user chose where the number gets inserterd isn't a bit risky ?

Comment (tokejepsen) [reply openpype/hosts/maya/api/plugin.py:185]: > Let the user chose where the number gets inserterd isn't a bit risky ?

If we provide a default value for the namespace that includes the number and warn in the documentation about removing it, I think that should be enough to reduce the risk.

Comment (BigRoy) [reply openpype/hosts/maya/api/plugin.py:185]: This should be pretty decent though:

```python
import re
from openpype.hosts.maya.api.lib import unique_namespace


def get_custom_namespace(custom_namespace):
    """Return unique namespace.

    The input namespace can contain a single group
    of '#' number tokens to indicate where the namespace's
    unique index should go.

    Warning: Note that a namespace will always be
    prefixed with a _ if it starts with a digit

    Examples:
    >>> get_custom_namespace("myspace_##_")
    # myspace_01_
    >>> get_custom_namespace("##_myspace")
    # _01_myspace
    >>> get_custom_namespace("myspace##")
    # myspace01

    """

    if not custom_namespace:
        raise ValueError("Empty custom namespace not supported")

    split = re.split("([#]+)", custom_namespace, 1)
    print(split)
    
    if len(split) == 3:
        # Custom namespace contains # for unique index position
        base, padding, suffix = split
        index_format = "%0{}d".format(len(padding))
    else:
        base = split[0]
        index_format = "%02d" # default behavior
        suffix = ""
        
    return unique_namespace(
        base, 
        format=index_format, 
        prefix="_" if not base or base[0].isdigit() else "",
        suffix=suffix
    )
 
# Testing
for test_custom_namespace in ["{asset_name}_{subset}_##", "{derp}_##_hello", "##hello", "hello"]:
    print(get_custom_namespace(test_custom_namespace))
```

tokejepsen: Works for me. @friquette how does that sound?

friquette: Yeah that's sounds great, I will try it :)

website/docs/admin_hosts_maya.md
```diff
@@ -106,6 +106,37 @@ or Deadlines **Draft Tile Assembler**.
...
+|---|---|
+|`{asset_name}` | Asset name |
+|`{asset_type}` | Asset type |
+|`{subset}` | Subset name |
+|`{family}` | Subset family |
+
+</div>
+</div>
+
+The namespace field can contain a single group of '#' number tokens to indicate where the namespace's unique index should go. The amount of tokens defines the zero padding of the number, e.g ### turns into 001.
```
BigRoy, line 132:
```suggestion
The namespace field can contain a single group of '#' number tokens to indicate where the namespace's unique index should go. The amount of tokens defines the zero padding of the number, for example `###` turns into `001`.
```

openpype/settings/defaults/project_settings/maya.json
```diff
@@ -1047,6 +1047,10 @@
                 125,
                 255
             ]
+        },
+        "reference_loader": {
+            "namespace": "{asset_name}_{subset}_##",
```
BigRoy, line 1052: To match the original behavior I believe this should still end with a `_` after the digits.

```suggestion
            "namespace": "{asset_name}_{subset}_##_",
```

tokejepsen:
@friquette let me know when I can test this again.

friquette:
> @friquette let me know when I can test this again.

You can test it, it should work :)

tokejepsen (approved):
@friquette Amazing work!

website/docs/admin_hosts_maya.md
```diff
@@ -170,6 +201,30 @@ These options are set on the camera shape when publishing the review. They corre
...
+|`{asset_name}` | Asset name |
+|`{asset_type}` | Asset type |
+|`{subset}` | Subset name |
+|`{family}` | Subset family |
+
+</div>
+</div>
+
+Example: `{asset_name}_{subset}_:_GRP`
+
```
tokejepsen, line 227:
```suggestion
```

tokejepsen (changes requested):
Sorry, looks like some old documentation snuck itself in.

antirotor (approved):
Looks great, thanks!

PR #4515, opened 2023-02-23
Start script: Change returncode on validate or list versions

iLLiCiTiT:
## Brief description
Change exit code from `1` to `0` when versions  are printed or when version is validated.

## Description
Return code `1` is indicating error but there didn't happen any error.

## Testing notes:
1. Run and check return code when you run `openpype_console --list-versions`


kalisp (approved):
`Process finished with exit code 0`

PR #4516, opened 2023-02-23
Maya: Passing custom attributes to alembic - OP-4111

tokejepsen:
## Brief description
Passing custom attributes to alembic

## Description
This PR makes it possible to pass all user defined attributes along to the alembic representation.

## Testing notes:
1. Create pointcache/animation instance in Maya.
2. Add some attributes to the output of the pointcache/animation.
3. Publish.
4. Load in the pointcache/animation, to validate the custom attributes exist.


openpype/hosts/maya/plugins/publish/collect_pointcache.py
```diff
@@ -42,3 +42,21 @@ def process(self, instance):
...
             instance.remove(proxy_set)
             instance.data["setMembers"].remove(proxy_set)
+
+        # Collect user defined attributes.
+        if not instance.data.get("includeUserDefinedAttributes", False):
+            return
+
+        all_nodes = (
+            instance.data["setMembers"] + instance.data.get("proxy", [])
+        )
```
BigRoy, line 52: Isn't all nodes supposed to be the nodes in the instance, like `nodes = list(instance)`?

Set members is only the direct member of the set. It doesn't include e.g. children (and thus also not e.g. children shapes). And thus that's why you also need to explicitly get the shapes. If you use `list(instance)` you shouldn't need to.

openpype/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -39,8 +39,7 @@ def process(self, instance):
         start = float(instance.data.get("frameStartHandle", 1))
         end = float(instance.data.get("frameEndHandle", 1))
 
-        attrs = instance.data.get("attr", "").split(";")
-        attrs = [value for value in attrs if value.strip()]
+        attrs = instance.data.get("userDefinedAttributes", [])
```
BigRoy, line 42: This is breaking the support of the `attr` setting that is on the instances. Functionality should be preserved.

BigRoy (commented):
It's too bad this functionality is created **before** PR #4388 

With that PR this whole functionality could have been in a single Collector, including the toggle in the UI. And I'd actually like to argue that if it could, that it should.

Only downside to that would be that you don't get the option per Creator to set the default in settings or not as far as I know.

openpype/hosts/maya/plugins/publish/collect_pointcache.py
tokejepsen: Yup good point.

openpype/hosts/maya/plugins/publish/extract_pointcache.py
tokejepsen: Yup good point.

website/docs/artist_hosts_maya.md
```diff
@@ -314,10 +314,18 @@ Example setup:
...
-:::
+#### Options
+
+- **Frame Start**: which frame to start the export at.
+- **Frame End**: which frame to end the export at.
+- **Handle Start**: additional frames to export at frame start. Ei. frame start - handle start = export start.
+- **Handle Start**: additional frames to export at frame end. Ei. frame end + handle end = export end.
+- **Step**: frequency of sampling the export. For example when dealing with quick movements for motion blur, a step size of less than 1 might be better.
+- **Refresh**: refresh the viewport when exporting the pointcache. For performance is best to leave off, but certain situations can require to refresh the viewport, for example using the Bullet plugin.
+- **Attr**: specific attributes to publish separated by `;`
```
BigRoy, line 325: There's also `attrPrefix` by the way.

Anyway, the new publisher #4388 exposes both of these nicely in the UI.

tokejepsen:
```suggestion
- **Attr**: specific attributes to publish separated by `;`.
- **AttrPrefix**: Prefix filter for determining which geometric attributes to write out, separated by `;`.
```

website/docs/artist_hosts_maya.md
```diff
@@ -322,7 +322,8 @@ Example setup:
 - **Handle Start**: additional frames to export at frame end. Ei. frame end + handle end = export end.
 - **Step**: frequency of sampling the export. For example when dealing with quick movements for motion blur, a step size of less than 1 might be better.
 - **Refresh**: refresh the viewport when exporting the pointcache. For performance is best to leave off, but certain situations can require to refresh the viewport, for example using the Bullet plugin.
-- **Attr**: specific attributes to publish separated by `;`
+- **Attr**: specific attributes to publish separated by `;`.
+- **AttrPrefix**: Prefix filter for determining which geometric attributes to write out, separated by `;`.
```
BigRoy, line 326: Sorry to be precise, but it's not only geometric - it'd be any attribute that matches it, also on transforms.

```suggestion
- **Attr**: specific attributes to publish separated by `;`.
- **AttrPrefix**: specific attributes which start with this prefix to publish separated by `;`.
```

Or:
```suggestion
- **Attr**: include attributes with these names in the publish separated by `;`, for example `my_attribute;other_attribute`
- **AttrPrefix**: include attributes starting with this prefix in the publish separated by `;`, for example `ai;vray`
```

LiborBatek (approved):
Works ok. Tested multiple types of Attribs and also add them to ```transformNode``` and ```ShapeNode``` both works well.

image

BigRoy (approved):
Code looks ok.

PR #4519, opened 2023-02-23
Nuke: creator farm attributes from deadline submit plugin settings

jakubjezek001:
## Brief description
Defaults in farm attributes are sourced from settings.

## Description
Settings for deadline nuke submitter are now used during nuke render and prerender creator plugins. 

## Testing notes:
1. Open Studio Project settings at your testing project at `project_settings/deadline/publish/NukeSubmitDeadline`
2. set following values to something like this (img) notice attributes moved to right panel
image

3. Open Nuke in testing project to some shot and create new render write node. 
4. Notice that created node is having defined values as it were in deadline plugin settings


openpype/hosts/nuke/plugins/create/create_write_render.py
```diff
@@ -47,6 +48,9 @@ def get_instance_attr_defs(self):
             self._get_reviewable_bool()
         ]
         if "farm_rendering" in self.instance_attributes:
+            render_farm_settings = RenderFarmSettings(
```
iLLiCiTiT, line 51: Do you know that this method is called for each individual instance? So, when you have 10 instances in scene, the exact same code will be called 10 times (create `RenderFarmSettings` -> get project settings -> recalculate values -> create attributes). Maybe having a cache of `RenderFarmSettings` would help a lot, especially when you have access to project settings in initialization part.

openpype/hosts/nuke/api/lib_rendersettings.py
```diff
@@ -0,0 +1,92 @@
...
+    log = Logger.get_logger("RenderFarmSettings")
+
+    _active_farm_module: str = None
+    _farm_modules: list = [
+        "deadline", "muster", "royalrender"]
+    _farm_plugins: dict = {
+        "deadline": "NukeSubmitDeadline"
+    }
+    _creator_farm_keys: list = [
+        "chunk_size", "priority", "concurrent_tasks"]
```
BigRoy, line 21: I feel like this should maybe in the actual Deadline Module on a plug-in, see for example [here](https://github.com/ynput/OpenPype/pull/4388/files#diff-dc5f13fcc4d3823ce88f1247d130582b4feaee7afe2a018d5fe2b562ac0e937aR680-R704).

These types of settings I feel should rather be at the farm renderer module level than here. No? Isn't that what we're trying to separate out to begin with using the new publisher?

openpype/hosts/nuke/api/lib_rendersettings.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class RenderFarmSettings:
+    """ Class for getting farm settings from project settings
+    """
+    log = Logger.get_logger("RenderFarmSettings")
+
+    _active_farm_module: str = None
+    _farm_modules: list = [
+        "deadline", "muster", "royalrender"]
```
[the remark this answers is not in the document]
jakubjezek001: I suppose I can strip the `muster` but we will be implementing `royalrender` soon.

openpype/hosts/nuke/plugins/create/create_write_render.py
jakubjezek001: I see, then I suppose it would be smart to cache. But isn't this called only during creation? Way you explain it, it seems like it is called every time you refresh publisher, right?

openpype/hosts/nuke/api/lib_rendersettings.py
jakubjezek001: Thanks for suggestion. It seems to be more sensible approach, thought I was looking into something more universal - more independent from a particular farm module. There is at the moment only Deadline but Royal Render will soon be fully integrated.

Nuke deadline submit plugin is not having implemented `AbstractSubmitDeadline` and it is not the part of this issue. 

Lets wait for your updates at the `AbstractSubmitDeadline` are merged and after that Nuke will follow.

openpype/hosts/nuke/plugins/create/create_write_render.py
iLLiCiTiT: Yes, it's called on refresh for each instane and then on create for each created instance.

jakubjezek001: done

BigRoy:
I think the issue with this isn't in the caching. But it is that it's trying to apply project + system setting defaults for the creators in `__init__`.
That logic is in the wrong place currently and should instead be done in the [Creator's `apply_settings` method](https://github.com/ynput/OpenPype/blob/093ffee97a11623e671953933877023a2d12cad3/openpype/pipeline/create/creator_plugins.py#L210-L215).

Also, since it's now intended for just a single render manager. I'd drop the complexity of the separate class. I personally don't see why it needs to be that length in code. All it's doing is just 'take the values from settings', right?

Additionally, I really feel we should take the chance now to just move these settings to the render farm module itself - maybe taking code from #4388 that does some of those changes for Deadline. I have a feeling they'd work 1-to-1 here and solve the issue at hand and actually move the issue to the actual render manager module as it should.

jakubjezek001:
> I think the issue with this isn't in the caching. But it is that it's trying to apply project + system setting defaults for the creators in `__init__`. That logic is in the wrong place currently and should instead be done in the [Creator's `apply_settings` method](https://github.com/ynput/OpenPype/blob/093ffee97a11623e671953933877023a2d12cad3/openpype/pipeline/create/creator_plugins.py#L210-L215).

I believe __init__ had been already cleared from it. Or perhaps it is not enought, right? Isn't apply_settings working only for settings in particular creator plugin. This are settings for module based plugins. 

> Also, since it's now intended for just a single render manager. I'd drop the complexity of the separate class. I personally don't see why it needs to be that length in code. All it's doing is just 'take the values from settings', right?

I know it is only single at the moment but we are preparing to merge soon Royal Renderer and droping it now would mean to implement it later again. Since it is done now already it feels like waste of time. 

Lets merge this now

iLLiCiTiT:
The issue is the slowdown of doing this all the time. The change is just addin this to creator plugin:
```
def apply_settings(self, project_settings, system_settings):
    self.render_farm_settings = RenderFarmSettings(project_settings, self.log)
    ...
```
And using `self.render_farm_settings` in rest of logic.

BigRoy:
> I know it is only single at the moment but we are preparing to merge soon Royal Renderer and droping it now would mean to implement it later again. Since it is done now already it feels like waste of time.

I really feel we should try and move this logic _to_ the module of the respective renderfarms. The logic shouldn't be in Nuke but in module/deadline and module/royalrender, no? if there's a base of setting that we suspect all renderfarms to at least have, then yes maybe a `openpype.lib.renderfarm` library with some re-usable logic is worth it. But I definitely feel we're now adding logic that'll somehow need to remain "backwards compatible" even though with the push to Ayon one of the major goals is separating out exactly these responsibilities to the addons themselves.

openpype/hosts/nuke/api/lib_rendersettings.py
```diff
@@ -0,0 +1,111 @@
...
+        return self._cached_system_settings
+
+    def _get_active_farm_module_from_system_settings(self):
+        """ Get active farm module from system settings
+        """
+        active_modules = [
+            module_
+            for module_ in self._farm_modules
+            if self.system_settings["modules"][module_]["enabled"]
+        ]
```
iLLiCiTiT, line 57: This should ideally use `ModulesManger` but if this is kept, it should use safe access to module keys.
```suggestion
        system_settings = self.system_settings
        active_modules = []
        for module_name in self._farm_modules:
            module_enabled = system_settings.get(module_name, {}).get("enabled")
            if module_enabled:
                active_modules .append(module_name
```

jakubjezek001:
> Resolving that now would mean we don't need to battle the technical debt and backwards compatibility in a few weeks. (and with that the support tickets maybe coming in too?)

I had looked again into your comment [here ](https://github.com/ynput/OpenPype/pull/4519#discussion_r1116114128) and have to admit for sure this makes more sense to do it your way. Thanks for insisting ;)

Look to latest commit if it is want you've meant @BigRoy .

64qam:
It crashes during publishing with an error
[publish-report-230301-10-23.txt](https://github.com/ynput/OpenPype/files/10859232/publish-report-230301-10-23.txt)

jakubjezek001:
> It crashes during publishing with an error [publish-report-230301-10-23.txt](https://github.com/ynput/OpenPype/files/10859232/publish-report-230301-10-23.txt)

This is not related to this PR. But yes there is an issue in deadline submiter commig from https://github.com/ynput/OpenPype/pull/4340
Also mentioned in this issue https://github.com/ynput/OpenPype/issues/4541

64qam (approved):

PR #4520, opened 2023-02-23
Maya: Set pool on tile assembly - OP-2012

tokejepsen:
## Brief description
Set pool on tile assembly

## Description
Pool for publishing and tiling jobs, need to use the settings (`project_settings/deadline/publish/ProcessSubmittedJobOnFarm/deadline_pool`) else fallback on primary pool (`project_settings/deadline/publish/CollectDeadlinePools/primary_pool`)

## Testing notes:
1. Set `project_settings/deadline/publish/ProcessSubmittedJobOnFarm/deadline_pool` to a valid pool on Deadline.
2. Set `project_settings/deadline/publish/CollectDeadlinePools/primary_pool` to a difference valid pool on Deadline.
3. Create render instance in Maya and publish.
4. Test various alternatives like nothing in `project_settings/deadline/publish/ProcessSubmittedJobOnFarm/deadline_pool` and/or `project_settings/deadline/publish/CollectDeadlinePools/primary_pool`.


LiborBatek:
I have set OP settings with valid deadline pools
image

when rendering tiles it uses "none" as set on ```renderMain``` instance in maya and when ```tile assemble``` job it uses correct deadline pool which was set to ```pool_tile```

image

LiborBatek:
Did another go this time with different OP settings and ```renderMain``` without any ```pool``` set to ```none```

image


When rendering it uses still ```none``` pool for any rendering/tile assembly job
image

LiborBatek:
I have end up with resume that when ```pool``` set to ```none``` in maya...

then it correctly takes whats set for ```proccessingJobs``` as a ```pool``` in my case ```pool_tile``` for ```tile assembly``` job



image

However when left this OP setting empty it doesnt take those neither ```primary pool``` nor ```secondary pool``` and sticks to ```none``` pool

tokejepsen:
@LiborBatek sorry I'm not 100% sure what your setup looks like when you say it does not work as expected?

The workflow is that publish/tile jobs take from `project_settings/deadline/publish/ProcessSubmittedJobOnFarm/deadline_pool` if it exists. If not it'll take what is on the render instance's `Primary Pool`.
The render instance's `Primary Pool` gets initially set to `project_settings/deadline/publish/CollectDeadlinePools/primary_pool` on creation.

LiborBatek:
Ok I see... now I get it the first Deadline pool option in OP Settings is for whats prefilled /preset on ```renderInstance``` and the other for ```publish/tile assembly jobs```.

So I should check if ```renderMain``` instance got prefilled with the ```primary pool``` string name from OP Settings and if ```tile assembly rendered using predefined ```pool``` from the second OP Setting.

Right?

LiborBatek (dismissed):
First catch is it seems that OP Settings and/or ```render``` instance' s ```primary``` and ```secondary``` pools are mismatched...check and compare scrngrab below

Here is the OP Settings and freshly created ```render``` instance's params
image

LiborBatek (approved):
All works fine.

m-u-r-p-h-y:
Just have a question, is it a good workflow to fallback to render pool if publish pool is not set?

The reason why we've introduced  `pool `for `ProcessSubmittedJobOnFarm`  (and now adding tile jobs to it) is to avoid blocking fast multicore machines with simple publish/tile jobs, generating, communicating with ftrack/shotgrid and other "simple" tasks., which can be handled by any other machine.

I would find it more suitable to set pool `None` if no `pool `for `ProcessSubmittedJobOnFarm` is defined in settings instead of pushing for primary or secondary pool used for rendering.

Of course we can argue, it is a misconfiguration just thinking of the most bulletproof approach.

tokejepsen:
> Of course we can argue, it is a misconfiguration just thinking of the most bulletproof approach.

@m-u-r-p-h-y dont mind whatever direction we take with this. I did the fallback because that was existing behaviour in code, ei. use instance `pool`.

mkolar:
I think that falling back to primary pool is reasonable actually. It is quite common to use pool for prioritizing shows for example in which case I'd expect for it to fall back there. It is called deadline/primary pool, rather than rendering pool and I think this PR is handling it absolutely correctly. 

Also it means no change of behaviour from previous setup, which is correct approach I'd say.

mkolar (approved):

antirotor (approved):

PR #4522, opened 2023-02-23
Fusion: Updated render_local.py to not only process the first instance

EmberLightVFX:
## Description
Moved the `__hasRun` to `render_once()` so the check only happens with the rendering. Currently only the first render node gets the representations added.

Critical PR


BigRoy (commented):
So the current logic does fix the rendering for multiple savers, but it doesn't catch for each instance whether the render succeeded or not. It'll only raise the Comp render failed error on the first instance.

I was thinking something like this would at least raise the error for all instances:
```python


    def render_once(self, context):
        """Render context comp only once, even with more render instances"""

        # This plug-in assumes all render nodes get rendered at the same time
        # to speed up the rendering. The check below makes sure that we only
        # execute the rendering once and not for each instance.
        key = f"__hasRun{self.__class__.__name__}"
        if key not in context.data:
            # We initialize as false to indicate it wasn't successful yet
            # so we can keep track of whether Fusion succeeded
            context.data[key] = False

            current_comp = context.data["currentComp"]
            frame_start = context.data["frameStartHandle"]
            frame_end = context.data["frameEndHandle"]
    
            self.log.info("Starting render")
            self.log.info(f"Start frame: {frame_start}")
            self.log.info(f"End frame: {frame_end}")
    
            with comp_lock_and_undo_chunk(current_comp):
                result = current_comp.Render({
                    "Start": frame_start,
                    "End": frame_end,
                    "Wait": True
                })
                
            context.data[key] = bool(result)

        if context.data[key] is False:
            raise RuntimeError("Comp render failed")
```

Thoughts?

jakubjezek001:
> So the current logic does fix the rendering for multiple savers, but it doesn't catch for each instance whether the render succeeded or not. It'll only raise the Comp render failed error on the first instance.

I agree that your approach would be more systematic. Woting for it. I have just tested your code and works fine. 

Perhaps I would add some log notifying that renders were done for each instance, whether it was rendering or not. At the moment it prints only once and it might be confusing for some (like me ;) )

openpype/hosts/fusion/plugins/publish/render_local.py
```diff
@@ -41,10 +34,10 @@ def process(self, instance):
             for frame in range(frame_start, frame_end + 1)
         ]
         repre = {
-            'name': ext[1:],
-            'ext': ext[1:],
-            'frameStart': f"%0{len(str(frame_end))}d" % frame_start,
-            'files': files,
+            "name": ext[1:],
+            "ext": ext[1:],
+            "frameStart": f"%0{len(str(frame_end))}d" % frame_start,
```
BigRoy, line 39: Anyone else feels this formatting line feels a bit unreadable?

BigRoy (approved):
Changes look good to me. 

I did have a small style/formatting comment that I've with this approval. I'm wondering whether both:
- that's the correct logic, does Fusion actually base the padding of the first frame on the length of the last frame?
- would there be a way to make the line more readable?

EmberLightVFX:
@BigRoy I just did some tests and

> does Fusion actually base the padding of the first frame on the length of the last frame?
No it doesn't I rendered from 9990 to 10010 and there is no z-fill in front of 999X.
If I change the render name from `render..exr` to `render.00000.exr` it does give z-fill in front of 999X
Fusion defaults to a 4 z-fill so frame 4 would always end up being 0004 if the render name doesn't specify the amount of numbers like I did above. (Currently OP names the render path to `render..exr` without any specification on the amount of z-fills)

> would there be a way to make the line more readable?
I don't see how but maybe someone else have a better eye for that kind of code

EmberLightVFX:
Added logging in this format in the latest push:
`Rendered '{name}' for asset '{asset}' under the task '{task}'`

Is something like that good? Or were you thinking something else?

openpype/hosts/fusion/plugins/publish/render_local.py
```diff
@@ -17,18 +17,20 @@ class Fusionlocal(pyblish.api.InstancePlugin):
...
 
+        # Start render
         self.render_once(context)
 
+        # Log render status
+        self.log.info(
+            "Rendered '{nm}' for asset '{ast}' under the task '{tsk}'".format(
+                nm=instance.data.name,
+                ast=instance.data.asset,
+                tsk=instance.data.task,
```
BigRoy, line 30: This doesn't work, right? It should be `instance.data["name"]`, `instance.data["asset"]`, etc. Right? :)
Or is this some legacy backwards compatible way of what Pyblish offers? Have never seen this.

@jakubjezek001 is this the kind of logic you'd have expected? Note that technically the second instance logging it didn't actually _just render it_ but the previous instance had already rendered the comp. But at least this shows a log per instance.

EmberLightVFX: No you're correct. idk how I could do that. Fixed it!

jakubjezek001:
Just to keep the conversation in the right place.. Lets wait for merge of https://github.com/ynput/OpenPype/pull/4556 and implement **mixin** color managed class rather then to use `ColormanagedExctractor`.

jakubjezek001:
Here is a confusion I am facing at the moment while I am testing this PR. I had set and activated new `OCIO config` settings input and left the old one activated. 
image

Now, this will be very confusing as the path had been set from the old depricated attribute. But since the render_local.py plugin and its mixin class functionality is getting config from the new `OCIO config`. 

How about to use always the new `OCIO config` and rather warn user somehow that the production is havign the old setting `configFilePath` still active and those has to be migrated?

BigRoy:
> Now, this will be very confusing as the path had been set from the old depricated attribute. But since the render_local.py plugin and its mixin class functionality is getting config from the new OCIO config.
>
> How about to use always the new OCIO config and rather warn user somehow that the production is havign the old setting configFilePath still active and those has to be migrated?

I recall having asked this before - so what should happen if the OCIO config setting for Fusion was disabled (setting: `project_settings/fusion/imageio/ocio_config/enabled`). Should it instead apply the global OCIO setting or should it disable OCIO color management in Fusion and NOT set the `OCIO` env value?

For example I could change the prelaunch hook to this:
```python
from openpype.lib import PreLaunchHook

from openpype.pipeline.colorspace import get_imageio_config


class FusionPreLaunchOCIO(PreLaunchHook):
    """Set OCIO environment variable for Fusion"""
    app_groups = ["fusion"]

    def execute(self):
        """Hook entry method."""

        config_data = get_imageio_config(
            project_name=self.data["project_name"],
            host_name=self.data["host_name"],
            project_settings=self.data["project_settings"],
            anatomy=self.data["anatomy"]
        )
        ocio_path = config_data["path"]

        self.log.info(f"Setting OCIO config path: {ocio_path}")
        self.launch_context.env["OCIO"] = ocio_path
```
So that it uses the `get_imageio_config` result which other areas of the code rely on too BUT I'm not entirely sure how I would e.g. disable OCIO config altogether for a host then - would that even be possible?

Looking at the logic of `get_imageio_config` it's never allowed to return `None` since it'd raise a `FileNotExistsError` then.

jakubjezek001:
> I recall having asked this before - so what should happen if the OCIO config setting for Fusion was disabled (setting: project_settings/fusion/imageio/ocio_config/enabled). Should it instead apply the global OCIO setting or should it disable OCIO color management in Fusion and NOT set the OCIO env value?

I thing you have found the gap in the worlfow. You are right we have to have some sort of color management per host switch which would be more explicit. At the moment it is broken in case you dont want to use any OCIO config an the host attribute is disabled, it is getting the config from global. 

Best would be to have activation toggle on `project_settings/fusion/imageio/enabled`
image

> For example I could change the prelaunch hook to this:

This way the hook would be working. But I am not sure about the issue with old and new ocio config activated.

BigRoy:
I've updated the hook now to use the logic from the `openpype.pipeline.colorspace` module just so it'll at least try to do what the color space functionality was intended to do.

Note: The older deprecated setting would thus be completely ignored.

jakubjezek001 (approved):
Awesome work! ;)

PR #4524, opened 2023-02-23
Scene Inventory: Avoid using ObjectId

iLLiCiTiT:
## Brief description
Avoid using conversion to ObjectId type in scene inventory tool.

## Description
Preparation for AYON compatibility where ObjectId won't be used for ids. Representation ids from loaded containers are not converted to ObjectId but kept as strings which also required some changes when working with representation documents.

## Testing notes:
1. Open DCC where is host integration with references
2. Reference (load) something into the scene
3. Open scene manager
4. You should see all loaded containers there
5. All actions on them should work
6. Try run Switch action
7. Switch dialog should show valid possible changes as before this PR

### DIFF
--- a/openpype/tools/sceneinventory/switch_dialog.py
+++ b/openpype/tools/sceneinventory/switch_dialog.py
@@ -2,7 +2,6 @@ import collections
 import logging
 from qtpy import QtWidgets, QtCore
 import qtawesome
-from bson.objectid import ObjectId
 
 from openpype.client import (
     get_asset_by_name,
@@ -161,7 +160,7 @@ class SwitchAssetDialog(QtWidgets.QDialog):
         repre_ids = set()
         content_loaders = set()
         for item in self._items:
-            repre_ids.add(ObjectId(item["representation"]))
+            repre_ids.add(str(item["representation"]))
             content_loaders.add(item["loader"])
 
         project_name = self.active_project()
@@ -170,7 +169,7 @@ class SwitchAssetDialog(QtWidgets.QDialog):
             representation_ids=repre_ids,
             archived=True
         ))
-        repres_by_id = {repre["_id"]: repre for repre in repres}
+        repres_by_id = {str(repre["_id"]): repre for repre in repres}
 
         # stash context values, works only for single representation
         if len(repres) == 1:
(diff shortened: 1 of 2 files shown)


openpype/tools/sceneinventory/switch_dialog.py
```diff
@@ -181,18 +180,18 @@ def _prepare_content_data(self):
         content_repres = {}
         archived_repres = []
         missing_repres = []
-        version_ids = []
+        version_ids = set()
```
BigRoy, line 183: With the conversion to `set` you can remove the `set()` from line 198.

openpype/tools/sceneinventory/view.py
```diff
@@ -146,25 +142,25 @@ def _build_item_menu_for_selection(self, items, menu):
...
         if has_loaded_hero_versions:
             def _on_switch_to_versioned(items):
-                repre_ids = []
+                repre_ids = set()
                 for item in items:
-                    item_id = ObjectId(item["representation"])
+                    item_id = item["representation"]
                     if item_id not in repre_ids:
-                        repre_ids.append(item_id)
+                        repre_ids.add(item_id)
```
BigRoy, line 149: Like above this could be something like:
```python
        repre_ids = {
            item["representation"]
            for item in items
        }
```

Right?

openpype/tools/sceneinventory/view.py
```diff
@@ -146,25 +142,25 @@ def _build_item_menu_for_selection(self, items, menu):
...
 
                 repre_docs = get_representations(
                     project_name,
                     representation_ids=repre_ids,
                     fields=["parent"]
                 )
 
-                version_ids = []
+                version_ids = set()
```
BigRoy, line 157: It's not a newly introduced bug but it does look like confusing code. On line 171 `version_ids` is redefined as a `set` and from a quick glance over the code I'm not entirely sure why it needs a completely new clear definition like that?

Should the attribute on 171 be called `hero_version_ids`, etc.?
Why does it need to be rebuild like that there from hero versions? A comment might be welcome?

iLLiCiTiT: Changed to `hero_src_version_ids`.

kalisp (approved):
Switching/updating assets works.

PR #4527, opened 2023-02-24
Maya: Extract review with handles

tokejepsen:
## Brief description
Review was not extracting properly with/without handles.

## Description
Review instance was not created properly resulting in the frame range on the instance including handles.

## Testing notes
1. Setup a shot with handles.
2. Setup `project_settings/global/publish/ExtractReview/profiles/0/outputs/h264/tags`  with the `Skip handle frames` tag.
3. Create an review instance in Maya and publish.
4. Validate review in Ftrack is without handles.


LiborBatek (changes requested):
When testing making review without the custom tag ```publish``` creates correctly ```review``` with frame handles (in my case 1001-1050 + 10 frames/start + 10 frames/end) resulting into 991-1060 frame range.

However when setting custom tag as seen on pic

image

```publish``` of the ```review``` outputs just single frame not even sequence of frames.
220901_sh003_reviewMain_v009

Resulting static frame without burnins.

**Comment by LiborBatek** [comment]
also gives me this non critical error during ```review``` extraction

image

File: openpype/hosts/maya/plugins/publish/validate_frame_range.py [hunk 1]
```diff
@@ -69,7 +73,7 @@ def process(self, instance):
         if [ef for ef in self.exclude_families
                 if instance.data["family"] in ef]:
             return
-        if(inst_start != frame_start_handle):
+        if (inst_start != frame_start_handle):
             errors.append("Instance start frame [ {} ] doesn't "
                           "match the one set on instance [ {} ]: "
                           "{}/{}/{}/{} (handle/start/end/handle)".format(
```
BigRoy, line 79: Not introduced by this PR, but reading this message now I think one of the words "instance" should either be `context` or `asset`, right? Same for the one on line 85.

openpype/hosts/maya/plugins/create/create_review.py
```diff
@@ -28,13 +28,13 @@ class CreateReview(plugin.Creator):
 
     def __init__(self, *args, **kwargs):
         super(CreateReview, self).__init__(*args, **kwargs)
+        data = OrderedDict(**self.data)
 
         # get basic animation data : start / end / handles / steps
-        data = OrderedDict(**self.data)
-        animation_data = lib.collect_animation_data(fps=True)
-        for key, value in animation_data.items():
+        for key, value in lib.get_frame_range().items():
```
BigRoy, line 34: 1. Could we add a comment here as to why we're using `get_frame_range` over `collect_animation_data`.
It's solely because we don't want to default it based on the timeline, but have it default to the asset?

2. Could we actually add this behind a Creator setting? Because our studio would prefer it to still be based on the current user timeline.

tokejepsen:
> 

@LiborBatek I'm not sure whether that is supposed to work?
To add the `Skip handle frames` tag, you'll need to use `project_settings/global/publish/ExtractReview/profiles/0/outputs/h264/tags` instead of `project_settings/global/publish/ExtractReview/profiles/0/outputs/h264/filter/custom_tags`

Capture

LiborBatek:
OK it seems I have set it wrong. It wasnt fully clear to me I guess. Will do another round of testing. Thx

openpype/hosts/maya/plugins/publish/validate_frame_range.py
tokejepsen: Yup, done.

openpype/hosts/maya/plugins/create/create_review.py
tokejepsen: Yeah, good point. Have done backwards compatibility now. New behaviour is behind a setting.

I was wondering though whether the timeline below, should result in a start handle of `25`?

Capture

BigRoy: I think so, yes. 1026-1001.

tokejepsen: It does not currently account for the Maya "handles" so we should have that behind a setting as well?

tokejepsen: > What does it do now, take the longest one? I think we've used it forever like that so that animators can focus on a part of the timeline but not have that interfere with what they are about to create to publish.

Yeah, it takes the longest timeline so its always without handles.

> To me the behavior where they don't turn into handles is just fine. What does the rest of OP crew think?

@antirotor @mkolar ?

LiborBatek (approved):

LiborBatek:
All seems fine when use the proper ```custom tag```

image


All published without frame handles.

image

LiborBatek:
Small note: also tested without the ```custom tag``` and it correctly outputs the ```review``` with frame handles.

openpype/hosts/maya/plugins/create/create_review.py
tokejepsen: > Good question. What does it do now, take the longest one? I think we've used it forever like that so that animators can focus on a part of the timeline but not have that interfere with what they are about to create to publish.

This is only on instance creation though. So the animator can have whatever timeline when the instances are setup.

mkolar (approved):

mkolar:
Makes good sense to me and looking all fine.

Issue #4316, opened 2023-01-13
Publisher: In create mode hide selection of Subsets


BigRoy:
**Is your feature request related to a problem? Please describe.**

When on the Create Page you still see the active selection in Subsets to publish on the right

![afbeelding](https://user-images.githubusercontent.com/2439881/212379611-31f443e5-8a30-46e8-b4e4-31b3db7cdb4e.png)

However, that makes it somewhat feel like that right hand side has something to do with the values I'm seeing to the left of it. But they aren't. I almost think that right hand side subsets to publish should be deactivated completely and just be visual reference of what you already have.

(The same goes for the + button on the right hand side underneath those Subset to Publish - it doesn't do anything in the Create page)

---

Also, selecting no asset/task on the left hand side still allows you to create an instance - which seems to be a bit confusing too.

**Describe the solution you'd like**

The create page to me still confuses me most of the time. So anything that tries to avoid clutter and confusion on the page will help.

**Describe alternatives you've considered**

It might be even worth deciding to NOT show the subset to publish at all. It's nice seeing what's already there currently but I feel like in the current state it's more confusing and clutter than what it adds to the end user.

[cuID:[OP-4745](https://app.clickup.com/t/861m7rdgd)]

iLLiCiTiT:
Accidentally marked as resolved in PR that fixed different issue

jakubjezek001:
Here are my perspective on this. I agree with @BigRoy that there are some confusing point. For me it is:

- The `Create >>` button as it is almost hidden and very small and it doesn't feel like next step to something. Also it would be good to make the button in deactivated mode in case something is missing in the family requirements. 
- Create tab doesn't need to have publishing comment and oher buttons, since those are more publishing related feaures. But Reload button is ofcourse related to Create so it could stay there.
- List of instances at Create, I understand why it is selectable - as user can multi/single select them and remove.
- Yes the plus sign at instance list at right side, is not realy any particular importance at Create tab. I would remove it completely. To make transition button similar as to go from Create to Publish with `>` right button we could create the same at left in Publish tab for going back to Creator with `<` button.
- There is the burger menu button which is not actually menu but display mode switch. It could be confusing for lot users. I would rather want to see diplay mode button on top of the list and action buttons at bottom - so for me Bin icon on bottom only.
- List of instances **Options** in Create tab, seems to be shown unnecessarily.

PR #4528, opened 2023-02-24
Publisher: Prevent access to create tab after publish start

iLLiCiTiT:
## Brief description
Prevent access to create tab after publish start.

## Description
Disable create button in instance view on publish start and enable it again on reset. Even with that make sure that it is not possible to go to create tab if the tab is disabled.

## Testing notes:
1. Open Publisher tool
2. Hit validate (you must have at least 1 created instance to be able do that)
3. Go to Publish tab
4. Minimize bottom overlay
5. `+` button in instance view should be disabled

Resolves https://github.com/ynput/OpenPype/issues/4316


jakubjezek001 (approved):
works as described

PR #4532, opened 2023-02-27
Data Exchanges: Point Cloud for 3dsMax

moonyuet:
## Brief description
Publish PRT format with tyFlow in 3dsmax

## Description
Publish PRT format with tyFlow in 3dsmax and possibly set up loader to load the format too.
- [x] creator 
- [x] extractor 
- [x] validator
- [x] loader

## Additional info
1. Users need to manually create export_particle operator for publishing to export the point cloud in PRT format as tyFlow API currently doesn't support creating any operator in editor.
2. Partition count and related set-up are not allowed due to the current renaming from integrate,py does not support weird naming in terms of parition count. 
3. currently supports prt format only.
4. Make sure you installed tyFlow plugin, otherwises it will error out if publish the instance.
5. Point Cloud would be loaded within tyCache in Loader.

## Testing notes:
1. launch 3dsmax via Openpype
2. Click Publish from Openpype
3. Create PointCloud instance with your tyFlow class
6. Before publishing, make sure you add export_particles operator into your tyFlow editor(it can be accessed through modifier)
7. If you forgot to add export_particles, the validator will remind you and it will error out
8. If you have any custom channel attributes needed to add, you can add through setting
image
9. If the custom attributes doesn't exist, the validator will also error out.
10. If you manually changed the partition settings, the validator will error out too.
11. Publish Point Cloud


LiborBatek:
Did test run of ```publish``` of ```pointCloud``` and it succesfully publishes the ```tyCache``` in ```PRT``` format.

image

I have also tested loading of ```pointCloud``` asset into new scene which seems also working fine.

Need to test more with display options of ```tyCache``` to be sure and also addition of ```custom attributes``` vie OP Settings 

When all tested will give it review.

LiborBatek:
So using ```Asset Loader``` it successfully loads ```tyCache``` in ```PRT``` format. As seen on pic there are also published all ```attributes```.

image

LiborBatek:
Not sure how it should work with ```attributes``` in OP Settings. 

I have removed ```Rotation``` entry in OP  Settings but during ```publish``` the validator went without any error.

It also seems that it exports the attribs as set in ```tyFlow``` ```Export Particles``` nodes and doesnt take OP Settings into account.

image

Not sure how it suppose to work with OP Settings and ```attributes```.

moonyuet:
> Not sure how it should work with `attributes` in OP Settings.
> 
> I have removed `Rotation` entry in OP Settings but during `publish` the validator went without any error.
> 
> It also seems that it exports the attribs as set in `tyFlow` `Export Particles` nodes and doesnt take OP Settings into account.
> 
> image
> 
> Not sure how it suppose to work with OP Settings and `attributes`.

The attributes slot may have some limitation, it is expected the prf format has exported all the channel attributes and it gives the log message, but for some reason they are not all loaded as expected. 
It is normal if you remove the attribute and still pass the validators. Only that when you include the invalid or non-exist attributes, it will error out.

LiborBatek:
When adding custom ```attribute``` into OP Settings which doesnt exists in tyFlow asset the ```validator``` correctly kicks in and fail on it which is desirable.

image

LiborBatek (approved):
Finally tested all aspects of Point cloud for 3dsmax

1) Publish of PRT files works ok

2) Loading of PRT works ok

3) Validation of custom attribs works ok

image

mkolar (approved):

PR #4536, opened 2023-02-27
Unreal: fix loaders because of missing AssetContainer

antirotor:
## Brief description
Fixing Unreal loaders, where changes in OpenPype Unreal integration plugin deleted AssetContainer.

## Description
`AssetContainer` and `AssetContainerFactory` are still used to mark loaded instances. Because of optimizations in Integration plugin we've accidentally removed them but that broke loader.

## Testing notes:
1. Start Unreal with OP
2. try to load something

> **Note**
> you need to force engine to rebuild the plugin - delete `C:\Program Files\Epic Games\UE_5.1\Engine\Plugins\Marketplace\OpenPype` (or analogous)


LiborBatek (approved):
Followed the testing steps and all works as expected.

Did load several assets via OP Loader and no issues found.

image

PR #4537, opened 2023-02-27
Enhancement kitsu note with exceptions

Sharkitty:
## Description

Adding a setting to choose some exceptions to IntegrateKitsuNote task status changes.

## Testing

- Go to project settings -> Kitsu -> Publish.
- With integrate kitsu note active, add entries to the exception list and set them up.
- Go on kitsu and set the status of the task you want to use for testing, so the exceptions would apply.
- Publish this task.
- The task status should remain unchanged.
- You can also test that in a case in which the exceptions should not trigger, then the status set as `Note Status Shortname` is applied.


EmberLightVFX:
If I understand this PR correct I can pick to not change the status to wip if the old statis is say wfa or some custom kind of task status?

openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
```diff
@@ -23,8 +24,21 @@ def process(self, context):
...
         # if it is not specified in the configuration
-        note_status = context.data["kitsu_task"]["task_status_id"]
-        if self.set_status_note:
+        kitsu_task = context.data["kitsu_task"]
+        shortname = kitsu_task["task_status"]["short_name"]
+        note_status = kitsu_task["task_status_id"]
+        if self.set_status_note and next(
+            (
+                False
+                for se in self.status_exceptions
```
Tilix4, line 33: call it `status_except`

openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
```diff
@@ -23,8 +24,21 @@ def process(self, context):
...
+                if shortname == se["short_name"]
+                and se["equality"].startswith("e")
```
Tilix4, line 35: `startswith` is slower than `==` because of the function overhead, you should change it.

Tilix4 (changes requested):
The code looks good apart from small optimization we could achieve.

Please add some documentation for this feature with a small screen capture.

website/docs/module_kitsu.md
```diff
@@ -38,6 +38,18 @@ This functionality cannot deal with all cases and is not error proof, some inter
...
 
+## Integrate Kitsu Note
+Task status can be automatically set during publish thanks to `Integrate Kitsu Note`. This feature can be configured in:
+
+`Admin -> Studio Settings -> Project Settings -> Kitsu -> Integrate Kitsu Note`.
+
+There are three settings available:
+- `Set status on note` -> turns on and off this integrator.
+- `Note shortname` -> Which status shortname should be set automatically (Case sensitive).
+- `Status exceptions` -> Exceptions for which task status should be left unchanged. You can add as many exceptions as you like. There are two fields to each exceptions: `Equality` (Whether current status should be equal or not equal to the exception status) and `Short name` (Shortname of the exception status).
```
Tilix4, line 49: I'd call it `Condition` instead of `Equality`

website/docs/module_kitsu.md
```diff
@@ -38,6 +38,18 @@ This functionality cannot deal with all cases and is not error proof, some inter
...
+
+![Integrate Kitsu Note project settings](assets/integrate_kitsu_note_settings.png)
```
Tilix4, line 51: Could you have the two examples `Equal` and `Not Equal` in you screenshot please?

Tilix4 (changes requested):
Nice! Almost there.

website/docs/module_kitsu.md
```diff
@@ -46,7 +46,7 @@ Task status can be automatically set during publish thanks to `Integrate Kitsu N
 There are three settings available:
 - `Set status on note` -> turns on and off this integrator.
 - `Note shortname` -> Which status shortname should be set automatically (Case sensitive).
-- `Status exceptions` -> Exceptions for which task status should be left unchanged. You can add as many exceptions as you like. There are two fields to each exceptions: `Equality` (Whether current status should be equal or not equal to the exception status) and `Short name` (Shortname of the exception status).
+- `Status exceptions` -> Exceptions for which task status should be left unchanged. You can add as many exceptions as you like. There are two fields to each exceptions: `Condition` (Whether current status should be equal or not equal to the exception status) and `Short name` (Shortname of the exception status).
```
Tilix4, line 49:
```suggestion
- `Status exceptions` -> Exceptions for which task status should be left unchanged. You can add as many exceptions as you like. There are two fields to each exception: `Condition` (Whether current status should be equal or not equal to the exception status) and `Short name` (Kitsu Shortname of the exception status).
```

Tilix4 (approved):
Well done! Last small change and LGTM

Tilix4 (approved):

mkolar (approved):

mkolar:
Thank you guys. Looking good.

Tilix4 (changes requested):
Sorry to interrupt, but I think turning it into a `Status conditions` instead of `exceptions` will be more intuitive. This way, in a glance we can figure for what condition it is supposed to work, and not for which one it is not...
IMO, invert the boolean logic.

Tilix4:
Hi @jakubjezek001, why did you close this? :slightly_smiling_face:

jakubjezek001:
> Hi @jakubjezek001, why did you close this? 🙂

Sorry, I had accidentally closed it by removing target branch.

mkolar:
@Tilix4 what do you think functionally? good to merge?

Tilix4:
> what do you think functionally? good to merge?

@mkolar as mentioned [here](https://github.com/ynput/OpenPype/pull/4537#pullrequestreview-1343248632), I recommend we turn it into a "Kitsu note with conditions" to be more explicit with a single glance. @Sharkitty is aware of that and has few changes to do (for readability and a bit more optimization, as discussed [here](https://github.com/ynput/OpenPype/pull/4551#discussion_r1139923609))

mkolar:
> > what do you think functionally? good to merge?
> 
> @mkolar as mentioned [here](https://github.com/ynput/OpenPype/pull/4537#pullrequestreview-1343248632), I recommend we turn it into a "Kitsu note with conditions" to be more explicit with a single glance. @Sharkitty is aware of that and has few changes to do (for readability and a bit more optimization, as discussed [here](https://github.com/ynput/OpenPype/pull/4551#discussion_r1139923609))

Aah. I completely missed that. Thanks

openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
```diff
@@ -11,6 +11,7 @@ class IntegrateKitsuNote(pyblish.api.ContextPlugin):
     families = ["render", "kitsu"]
     set_status_note = False
     note_status_shortname = "wfa"
+    status_exceptions = list()
```
iLLiCiTiT, line 14: I agree with https://github.com/ynput/OpenPype/pull/4537#pullrequestreview-1343248632 it should be conditions -> `status_conditions`. Exceptions are not intuitive and is confusing for coders.

Suggestion: Values of `status_exceptions` does not allow default behavior -> e.g. if `status_exceptions` is empty list. I would suggest to change it to profiles with filters and output value (`True` or `False`).

openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
```diff
@@ -27,9 +28,19 @@ def process(self, context):
 
             # Get note status, by default uses the task status for the note
             # if it is not specified in the configuration
-            note_status = kitsu_task["task_status"]["id"]
-
-            if self.set_status_note:
+            shortname = kitsu_task["task_status"]["short_name"].upper()
+            note_status = kitsu_task["task_status_id"]
+            if self.set_status_note and next(
```
iLLiCiTiT, line 33: It could be me, but this is hard to read code. I highly recommend to create separate method which returns `True` or `False`.
```python
def match_status_conditins(self, short_name):
    for status_except in self.status_exceptions:
        is_same = shortname == status_except["short_name"].upper()
        if status_except["condition"] == "equal" and is_same:
            return True
        if status_except["condition"] == "not_equal" and not is_same:
            return True
    return False
```

openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
iLLiCiTiT: In other words it looks like they must be set, and must be set for every possible short name which may be annoying if you want to change status all the time.

Sharkitty: I'm working on turning `exceptions` into `conditions` as requested. I'm not sure I understand very well your suggestion, this only meant to add conditions if you need them (For instance, if an asset has been validated on kitsu by the real or a lead, then don't change status), otherwise the list can be left empty.

openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
Sharkitty: I'm changing the implementation based on Roy's suggestion on [PR 4551](https://github.com/ynput/OpenPype/pull/4551#discussion_r1136731042), so it should be more readable on next commit.

iLLiCiTiT: Is this PR needed in that case?

Sharkitty: PR 4551 derives from this PR, but each introduce a different feature. Then if adding both features as a single PR is preferred, I don't mind.

openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
```diff
@@ -27,9 +28,19 @@ def process(self, context):
...
 
-            if self.set_status_note:
+            # Check if any status condition is not met
+            allow_status_change = True
+            for status_cond in self.status_conditions:
+                condition = status_cond["condition"] == "equal"
+                match = status_cond["short_name"].upper() == shortname
+                if match and not condition or condition and not match:
+                    allow_status_change = False
+                    break
```
Tilix4, line 41: I propose to turn it into validation to simplify
```suggestion
            allow_status_change = False
            for status_cond in self.status_conditions:
                condition = status_cond["condition"] == "equal"
                match = status_cond["short_name"].upper() == shortname
                if match and condition:
                    allow_status_change = True
                    break
```

Tilix4 (changes requested):
Small change proposed, we are very close! Good work

openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
Tilix4: Okay, my proposal is not good and stops if we don't set any condition, while we would expect it to pass if no any condition is set. Keep it your way.

openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
Tilix4: I think we could merge this one, now it's ready. https://github.com/ynput/OpenPype/pull/4551#discussion_r1136731042 will be rebased on it.

openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
Tilix4: @iLLiCiTiT, IMO the current implementation works the way you expect it to.

iLLiCiTiT: It was suggestion, not requirement :)

iLLiCiTiT:
Was the branch created from `develop` of `release/next-minor` branch?

Sharkitty:
> Was the branch created from `develop` or `release/next-minor` branch? It's targeted to `release/next-minor` which might not be if there were no changes in dependencies or in OpenPype bootstrap logic.

It was created from `develop`, then @jakubjezek001 changed the target branch to `release/next-minor`

iLLiCiTiT:
Ok, I've changed all unwated changes back (I hope). Please resolve the conflicts.

PR #4539, opened 2023-02-28
Maya: Validate model name - OP-4983

tokejepsen:
## Brief description
Validate model name issues.

## Description
Couple of issues with validate model name;

- missing platform extraction from settings
- map function should be list comprehension
- code cosmetics

## Testing notes:
1. Setup `simple_shader_list.txt` with:

```
fabricCotton
fabricCorduroy
fabricGeneric

rubberGeneric

IDorganicSkin
```

2. Setup settings:

- enable `project_settings/maya/publish/ValidateModelName`.
- disable `project_settings/maya/publish/ValidateModelName/database`
- `project_settings/maya/publish/ValidateModelName/regex` > `(.*)_ID(?P<shader>.*)_(GEO)`
- `project_settings/maya/publish/ValidateModelName/material_file/windows` > point to created `simple_shader_list.txt`.

3. Setup Maya scene with two meshes named `balloon_IDrubberGeneric_GEO` and `string_IDfabricGeneric_GEO`.
4. Publish.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/validate_model_name.py
+++ b/openpype/hosts/maya/plugins/publish/validate_model_name.py
@@ -2,9 +2,11 @@
 """Validate model nodes names."""
 import os
 import re
-from maya import cmds
-import pyblish.api
+import platform
 
+from maya import cmds
+
+import pyblish.api
 from openpype.pipeline import legacy_io
 from openpype.pipeline.publish import ValidateContentsOrder
 import openpype.hosts.maya.api.action
@@ -44,7 +46,7 @@ class ValidateModelName(pyblish.api.InstancePlugin):
                     if not cmds.ls(child, transforms=True):
                         return False
                 return True
-            except:
+            except Exception:
                 return False
 
         invalid = []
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
Followed the testing steps by settings OP with:

image


and tried to publish as in testing steps again...all works as expected.

image

m-u-r-p-h-y:
I'm a little bit confused here. Why do we read an external txt file instead of a simple text field stored in settings?

If it is a relict from v2 we should fix it . . .

tokejepsen:
> I'm a little bit confused here. Why do we read an external txt file instead of a simple text field stored in settings?

@m-u-r-p-h-y its a setting to read from a "material file" > `project_settings/maya/publish/ValidateModelName/database`. It was reported by client to fix. Whether its a v2 setting and we need to deprecate it, I'll leave to you and @antirotor

antirotor:
This is coming from the Pype 2, where settings were just json presets and there was no real good place to store shader names apart the txt files. It was used by one client, to location was so artists could add there on the need basis. I guess moving it to the Settings makes sense now, but I wouldn't do it right now anyway. This will be better handled by AYON, because of the user permission issues with Settings and so on. So I would address it in some later PR

antirotor (approved):

PR #4545, opened 2023-02-28
Nuke: Add option to use new creating system in workfile template builder

iLLiCiTiT:
## Brief description
Nuke workfile template builder can use new creators instead of legacy creators.

## Description
Modified workfile template builder to have option to say if legacy creators should be used or new creators. Legacy creators are disabled by default, so Maya has changed the value.

## Testing notes:
- [x] Nuke template builder works with creator plugins
- [x] Maya template builder works with creator plugins


antirotor (approved):
It does work but I have a one note probably not related to this PR but anyway - when runing the template build in the comp with something already selected, it will build the nodes from the placeholders correctly, but it will also try to connect it to the selected pre-existing nodes and in result it will wreak complete chaos to the node graph.

jakubjezek001 (approved):

PR #4550, opened 2023-02-28
AYON: Load local settings

iLLiCiTiT:
## Brief description
Added option to load some of site settings in client and to define site id.

## Description
Define AYON site id which is important part of connection to server. Tells server from where was connection created and from that moment it should be possible to change "studio" root overrides on server. Anatomy is using studio root overrides from server when in AYON mode.

## Additional info
There are unhandled cases of setting site id in ayon connection when non-ayon process is launched. This could cause that studio site overrides woult not work in DCCs. This has to be validated and fixed in this or following PR. Because it could be connected to changes in `ayon_api` it would be maybe better to do it in different PR. Or itcould be handled in host `install` logic?

## Note
Make sure you update venv.

## Testing notes:
1. Connection to server should still work and could be created
2. Change of user or server should be possible
3. Anatomy should return roots
4. If studio overrides are set on server for the site it should be propagated in Anatomy too


kalisp:
I had to stop OP and restart it and now I am getting:
2023-03-03 12_25_02-Pype3 0 – C__Users_petrk_AppData_Local_ynput_ayon_addons_sitesync_1 0 0_openpype

kalisp:
But after changing root on a site (`manageProjects/roots?project=demo_Commercial`), Anatomy in Tray returns correct root. So it's working!! :)
2023-03-03 13_36_31-Window

PR #4551, opened 2023-03-01
Kitsu: note family requirements

Sharkitty:
## Brief description
Allowing to add family requirements to `IntegrateKitsuNote` task status change.

## Description
Adds a `Family requirements` setting to `Integrate Kitsu Note`, so you can add requirements to determine if kitsu task status should be changed based on which families are published or not. For instance you could have the status change only if another subset than workfile is published (but workfile can still be included) by adding an item set to `Not equal` and `workfile`.

## Additional info
When adding an item to this setting, if `Equality` is set to `Equal`, task status will be updated if any published subset is in the indicated family. If it is set to `Not equal`, task status will be updated if any published subset is in another family than the one indicated. You can add as many suggestions as you need in this way.

## Testing notes
- In `Project settings -> Kitsu -> Integrate Kitsu Note`, set `Set status on note` to True, `Note shortname` to the status shorname you want, and set `Family requirements` the way you need. If you `Status exceptions`, make sure you do not trigger the exceptions you set to perform the next series of tests (Except the last one).
- Publish an asset in a way that would NOT fulfill the family requirements you set, and check that your task status has not been updated.
- Publish an asset in a way that would fulfill the family requirements you set, and check that your task status has been updated.
- Now you can also publish while triggering a Status exception and check that you task status has not been updated.


Sharkitty:
Can I get @Tilix4 as a reviewer please?

BigRoy:
Is this ready for review?

openpype/settings/defaults/project_settings/kitsu.json
```diff
@@ -7,7 +7,9 @@
     "publish": {
         "IntegrateKitsuNote": {
             "set_status_note": false,
-            "note_status_shortname": "wfa"
+            "note_status_shortname": "wfa",
```
Tilix4, line 10: I think we should stick to the naming convention with upper case

website/docs/module_kitsu.md
```diff
@@ -38,6 +38,19 @@ This functionality cannot deal with all cases and is not error proof, some inter
 openpype_console module kitsu push-to-zou -l me@domain.ext -p my_password
 ```
 
+## Integrate Kitsu Note
+Task status can be automatically set during publish thanks to `Integrate Kitsu Note`. This feature can be configured in:
+
+`Admin -> Studio Settings -> Project Settings -> Kitsu -> Integrate Kitsu Note`.
+
+There are four settings available:
+- `Set status on note` -> turns on and off this integrator.
```
Tilix4, line 47:
```suggestion
- `Set status on note` -> Turns on and off this integrator.
```

website/docs/module_kitsu.md
```diff
@@ -38,6 +38,19 @@ This functionality cannot deal with all cases and is not error proof, some inter
...
+- `Note shortname` -> Which status shortname should be set automatically (Case sensitive).
+- `Status exceptions` -> Exceptions for which task status should be left unchanged. You can add as many exceptions as you like. There are two fields to each exceptions: `Condition` (Whether current status should be equal or not equal to the exception status) and `Short name` (Shortname of the exception status).
+- `Family requirements` -> With this option you can add requirements to which families must be pushed or not in order to have the task status set by this integrator. There are two fields for each requirements: `Condition` (Same as the above) and `Family` (name of the family concerned by this requirement). For instance, adding one item set to `Not equal` and `workfile`, would mean the task status would change if a subset from another family than workfile is published (workfile can still be included), but not if you only publish the workfile subset.
```
Tilix4, line 50:
```suggestion
- `Family requirements` -> With this option you can add requirements to which families must be pushed or not in order to have the task status set by this integrator. There are two fields for each requirements: `Condition` (Same as the above) and `Family` (name of the family concerned by this requirement). For instance, adding one item set to `Not equal` and `workfile`, would mean the task status would change if a subset from another family than workfile is published (workfile can still be included), but not if you publish the workfile subset only.
```

openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
```diff
@@ -20,16 +22,54 @@ def process(self, context):
 
         self.log.debug("Comment is `{}`".format(publish_comment))
 
+        families = set(
+            instance.data.get("kitsu_task")
+            for instance in context
+            if instance.data.get("publish")
+        )
```
Tilix4, line 29: Please add comment to explain and I'd use the comprehension this way.
```suggestion
        families = {
            instance.data.get("kitsu_task")
            for instance in context
            if instance.data.get("publish")
        }
```

openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
```diff
@@ -20,16 +22,54 @@ def process(self, context):
...
+                    (
+                        False
+                        for status_except in self.status_exceptions
+                        if shortname == status_except["short_name"].upper()
+                        and status_except["condition"] == "equal"
+                        or shortname != status_except["short_name"].upper()
+                        and status_except["condition"] == "not_equal"
+                    ),
+                    True,
+                )
```
Tilix4, line 53: I'd turn it into positive way and simplify it like this:
```suggestion
                and next(
                    (
                        (status_except["condition"] == "equal"
                        and shortname == status_except["short_name"].upper())
                        and (status_except["condition"] == "not_equal"
                        and shortname != status_except["short_name"].upper())
                        for status_except in self.status_exceptions
                    ),
                    False,
                )
```

openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
```diff
@@ -20,16 +22,54 @@ def process(self, context):
...
+                                and family_req["family"].lower() != family
+                                or family_req["condition"] == "not_equal"
+                                and family_req["family"].lower() == family
+                            ),
+                            True,
+                        )
+                    ),
+                    False,
+                )
+            ):
```
Tilix4, line 72: Simplify it the same way here.

Tilix4 (changes requested):
Nice, thank you very much! Few changes but the whole system works fine

openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
BigRoy: Looking at this logic I'd say it'd actually be more readable in a for loop - no?

```python
# Must not match a status exception
allow_status = False
for status_req in self.status_exceptions:
     equals = status_req ["condition"] == "equal"
     match = status_req ["short_name"].upper() == shortname
     if match == condition:
         allow_status = True
         break

# Must match a family requirement
allow_family = False
for family_req in self.family_requirements:
     equals = family_req["condition"] == "equal"
     match = family_req["short_name"].lower() == family
     if match == condition:
         allow_family = True
         break

if (
    self.status_note and
    allow_status and
    allow_family
):
    ...
```

Consider this psuedocode because I just can't easily within a minute grasp what the original if statements here are trying to do. So likely my logic is off, but in essence that'd be close to what the next() branches, etc. are doing. Right?

Tilix4: Hey, your proposal is slightly more readable, but slower to execute. I always put speed first when I have the choice, but that's maybe not in the guideline.

You're right about the `lower()` though, it must be `upper()`.

BigRoy: The speed difference I think in this case will be more than neglible. In your case you're performing `.lower()` call twice and the lookup into the dict of `family_req` twice too.

---

By the way, I would love some references on `next` being that much faster in comparison if you have them? Isn't that just about it using iterators and breaking early which my code basically does the same as the iterator comprehension (or the `generator` that it creates). I thought it was more about don't use list comprehensions which pre-build full lists, etc.

Sharkitty: I do find Tilix proposition better than what I did, except it needs to be linted and there's a mistake in his code, I think it should look more like this:
```
                and next(
                    (
                        (
                            status_except["condition"] == "equal"
                            and shortname == status_except["short_name"].upper()
                        )
                        or (
                            status_except["condition"] == "not_equal"
                            and shortname != status_except["short_name"].upper()
                        )
                        for status_except in self.status_exceptions
                    ),
                    False,
                )
```

Sharkitty:
I have to set this PR back to draft because changes have been requested in [PR 4537](https://github.com/ynput/OpenPype/pull/4537), which this PR depends on. Although I'd appreciate if we can continue open discussions so we can win some time once 4537 is merged, as I hope we find a consensus. (In fact these discussions also are related to 4537).

openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
Tilix4: In fact, my proposal implies what I was suggesting here https://github.com/ynput/OpenPype/pull/4537#pullrequestreview-1343248632, when I propose to turn it in a positive way.

Also, what I propose in the current discussion is wrong if we want to match any listed status condition when the original proposal, with the `if` in the generator to actually hold a boolean if matched status, was better. Mine creates a boolean for each condition, even though they don't have anything to do with the current status, which is wrong because if the first one is `False`, we'll miss other matching conditions. For this use case we'd be tempted to use a `any()`.

About performances, here I've made a test script: [mytests.zip](https://github.com/ynput/OpenPype/files/10999239/mytests.zip)

Conclusions:
- `for` with `break` is the fastest,  @BigRoy's wins :tada: 
- `next()` is a very bit slower, probably due to the function's overhead
- `any` is far slower

Therefore, I'd recommend to use @BigRoy's proposal, in a positive way (see my comment)!

EmberLightVFX:
Isn't all this actually up to the instance object itself?
You pick in the publisher if you want to review the object or not.

EmberLightVFX:
> > Isn't all this actually up to the instance object itself? You pick in the publisher if you want to review the object or not. 
> 
> I'll look into it, the publisher isn't available yet for all hosts. Blender host currently uses pyblish and we don't have this option with it.

Hm... true. Does blender have OpenPype support @iLLiCiTiT ?
Having two ways to set options like this could really confuse the user why something might not be working as expected.
And if introducing two ways to do the same thing it will be hard to remove one later on as different users will pick different ways to handle this.

Thoughts?

Sharkitty:
> > > Isn't all this actually up to the instance object itself? You pick in the publisher if you want to review the object or not. 
> > 
> > 
> > I'll look into it, the publisher isn't available yet for all hosts. Blender host currently uses pyblish and we don't have this option with it.
> 
> Hm... true. Does blender have OpenPype support @iLLiCiTiT ? Having two ways to set options like this could really confuse the user why something might not be working as expected. And if introducing two ways to do the same thing it will be hard to remove one later on as different users will pick different ways to handle this.
> 
> Thoughts?

Blender support is in development. The point you're raising makes sense to me. I'll think about it, I'll have to ask my lead though.

Tilix4:
@jakubjezek001 please assign @Sharkitty

openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
```diff
@@ -62,13 +65,41 @@ def process(self, context):
...
 
+            if allow_status_change:
+                # Get families
+                families = {
+                    instance.data.get("family")
+                    for instance in context
+                    if instance.data.get("publish")
+                }
+
+                allow_status_change = False
```
Tilix4, line 85: This is weird that you don't use the same logic as for statuses, because doing this introduces the same default use case issue: if there is no any item in`family_requirements`, `allow_status_change` will remain `False` and the note status won't be changed, while we expect it to be if no any family is listed. I think you must stick to your design above.

Tilix4 (changes requested):

Tilix4 (approved):
LGTM now!

mkolar:
Approved the approval of the approval. Sounds like a solid agreement here :D.

PR #4552, opened 2023-03-01
Deadline: fix submit_publish_job

### LINKED ISSUE #4541: AE publish with rendering fails
**Running version**
[3.15.2-nightly.1]

**Describe the bug**
After this PR https://github.com/ynput/OpenPype/pull/4340 previously working AE and Nuke workfiles fails with:
2023-02-24_15_20_34-Window

tokejepsen:
## Brief description
Fix submit_publish_job

## Description
Resolves #4541

### DIFF
--- a/openpype/modules/deadline/plugins/publish/submit_publish_job.py
+++ b/openpype/modules/deadline/plugins/publish/submit_publish_job.py
@@ -933,8 +933,10 @@ class ProcessSubmittedJobOnFarm(pyblish.api.InstancePlugin):
 
         self.log.info(data.get("expectedFiles"))
 
-        additional_data = {}
-        if pyblish.api.current_host() == "maya":
+        if isinstance(data.get("expectedFiles")[0], dict):
+            # we cannot attach AOVs to other subsets as we consider every
+            # AOV subset of its own.
+
             config = instance.data["colorspaceConfig"]
             additional_data = {
                 "renderProducts": instance.data["renderProducts"],
@@ -945,10 +947,6 @@ class ProcessSubmittedJobOnFarm(pyblish.api.InstancePlugin):
                     str(context.data["anatomy"].roots["work"]), "{root[work]}"
                 )
             }
-
-        if isinstance(data.get("expectedFiles")[0], dict):
-            # we cannot attach AOVs to other subsets as we consider every
-            # AOV subset of its own.
 
             if len(data.get("attachTo")) > 0:
                 assert len(data.get("expectedFiles")[0].keys()) == 1, (
--- a/openpype/modules/deadline/plugins/publish/submit_publish_job.py
+++ b/openpype/modules/deadline/plugins/publish/submit_publish_job.py
@@ -588,7 +588,7 @@ class ProcessSubmittedJobOnFarm(pyblish.api.InstancePlugin):
             self.log.debug("instances:{}".format(instances))
         return instances
 
-    def _get_representations(self, instance, exp_files, additional_data):
+    def _get_representations(self, instance, exp_files):
         """Create representations for file sequences.
 
         This will return representations of expected files if they are not
--- a/openpype/modules/deadline/plugins/publish/submit_publish_job.py
+++ b/openpype/modules/deadline/plugins/publish/submit_publish_job.py
@@ -933,15 +933,18 @@ class ProcessSubmittedJobOnFarm(pyblish.api.InstancePlugin):
 
         self.log.info(data.get("expectedFiles"))
 
-        additional_data = {
-            "renderProducts": instance.data["renderProducts"],
-            "colorspaceConfig": instance.data["colorspaceConfig"],
-            "display": instance.data["colorspaceDisplay"],
-            "view": instance.data["colorspaceView"],
-            "colorspaceTemplate": instance.data["colorspaceConfig"].replace(
-                str(context.data["anatomy"].roots["work"]), "{root[work]}"
-            )
-        }
+        additional_data = {}
+        if pyblish.api.current_host() == "maya":
+            config = instance.data["colorspaceConfig"]
+            additional_data = {
+                "renderProducts": instance.data["renderProducts"],
+                "colorspaceConfig": instance.data["colorspaceConfig"],
+                "display": instance.data["colorspaceDisplay"],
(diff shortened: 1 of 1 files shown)


kalisp (changes requested):
Fails (on AE) with
2023-03-01 19_08_58-Testing Platform PC - AnyDesk.

I am not particularly fond of this `== "maya"` logic. Would we will be adding more and more hosts when "renderProducts"/colorspace are implemented for those, or what will be approach there?

jakubjezek001 (changes requested):
This PR is still broken. On this line `additional_data` are not used and braking the process `https://github.com/ynput/OpenPype/pull/4552/files#diff-2534ee7134eb5afe9462ddc39bdf86307c14a8592ff7c3deed8246041b7edb89R591`

jakubjezek001 (changes requested):
Also, this code had broken my tests. It is not adding OPENPYPE_VERSION into environment - but OpenPype.py plugin expecting it. I presume we had to either remove those lines or find different way. 
https://github.com/ynput/OpenPype/pull/4552/files#diff-2534ee7134eb5afe9462ddc39bdf86307c14a8592ff7c3deed8246041b7edb89R146-R148

tokejepsen:
> I am not particularly fond of this `== "maya"` logic. Would we will be adding more and more hosts when "renderProducts"/colorspace are implemented for those, or what will be approach there?

@kalisp Yeah, its not a very good solution, but I think we should get back to a working state first then look at making it host agnostic.

> Also, this code had broken my tests. It is not adding OPENPYPE_VERSION into environment - but OpenPype.py plugin expecting it. I presume we had to either remove those lines or find different way.

@jakubjezek001 this is outside of fixing colorspace issue. It originates from https://github.com/ynput/OpenPype/pull/4489. Could you check you have an executable in the OpenPype Deadline plugin settings?

@jakubjezek001 @kalisp could you give it another go now?

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -588,7 +588,7 @@ def _create_instances_for_aov(
             self.log.debug("instances:{}".format(instances))
         return instances
 
-    def _get_representations(self, instance, exp_files, additional_data):
+    def _get_representations(self, instance, exp_files):
```
BigRoy, line 591: If we're changing the signature isn't there a function call somewhere using this which needs to change along too?

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -933,15 +933,18 @@ def process(self, instance):
...
+        if pyblish.api.current_host() == "maya":
+            config = instance.data["colorspaceConfig"]
+            additional_data = {
+                "renderProducts": instance.data["renderProducts"],
+                "colorspaceConfig": instance.data["colorspaceConfig"],
+                "display": instance.data["colorspaceDisplay"],
+                "view": instance.data["colorspaceView"],
+                "colorspaceTemplate": config.replace(
+                    str(context.data["anatomy"].roots["work"]), "{root[work]}"
+                )
```
BigRoy, line 946: **Not an issue introduced by this PR**

What if the root is `/mnt/P` and the path had `/mnt/P/projects/mnt/P/files`? This should only replace the actual exact beginning? What if the root was just `/projects` or `/share` - no idea what someone comes up with on unix.
 
Also, what's ensuring the path dividers `/` or `\` are the same - and path case sensitivity on some platforms?

I feel like we might need a dedicated safer method here?

Should this instead be using [`anatomy.find_root_template_from_path`](find_root_template_from_path)? 

```python
anatomy = context.data["anatomy"]
result, template = anatomy.find_root_template_from_path(config)
```

openpype/modules/deadline/plugins/publish/submit_publish_job.py
jakubjezek001: As far as I could see in the code there, there is only one call of the function, and it was not calling it with third input `additional_data`.  Look to https://github.com/ynput/OpenPype/pull/4552/files#diff-2534ee7134eb5afe9462ddc39bdf86307c14a8592ff7c3deed8246041b7edb89R971 it is the only occurrence in the plugin.

BigRoy: 👍 So the signature was just wrong before, check.

openpype/modules/deadline/plugins/publish/submit_publish_job.py
jakubjezek001: > Should this instead be using [anatomy.find_root_template_from_path](https://github.com/ynput/OpenPype/pull/find_root_template_from_path)?

That is one point, yes. I agree that the case you've described might happen. 

> colorspaceTemplate only exists in this code - why not directly acquire the data in the way we're [setting them](https://github.com/ynput/OpenPype/blob/f037074747d040caf12b9af565650a8ae6d32bf3/openpype/modules/deadline/plugins/publish/submit_publish_job.py#L559-L567)?

Perhpas this should be moved to the code you are suggesting, but there was actually created abstracted functionality in here which could be used. There is only missing Display and View input.

jakubjezek001:
> @jakubjezek001 this is outside of fixing colorspace issue. It originates from https://github.com/ynput/OpenPype/pull/4489. Could you check you have an executable in the OpenPype Deadline plugin settings?

I am having executables in deadline OpenPype plugin, and yeah running it from source. Here is what I am having in there 
image

kalisp (approved):
This solves issue for AE, Nuke.

Move of additional_data to aov logic looks better.

tokejepsen:
@jakubjezek001 does this PR work for you for fixing the bug?

openpype/modules/deadline/plugins/publish/submit_publish_job.py
tokejepsen: @jakubjezek001 @BigRoy could we move this to a dedicated issue to tackle outside of this PR?

BigRoy: Fine with me.

jakubjezek001 (approved):

PR #4554, opened 2023-03-01
Maya: Yeti Validate Rig Input - OP-3454

tokejepsen:
## Brief description
Fix Yeti Validate Rig Input

## Description
Existing workflow was broken due to this #3297.

## Testing notes:
1. Follow documentation in PR for Yeti Rigs.


LiborBatek:
First test done and successfully published ```yeti_rig``` instance from workfile.

image

LiborBatek:
Also spotted that in the doc section for Yeti usage is a mistake when speaking about Yeti Instance creation...

Instead of creating ```Yeti Rig``` there is written ```Yeti Cache``` which is incorrect and misleading.

image

tokejepsen:
> As seeing the `Outliner` window at the bottom which clearly states `Yeti Rig Default` Instance has been created not `Yeti Cache`.

Good spot! Fixed.

LiborBatek (changes requested):
Following testing steps / OP user docs with these findings:

1) Publish of the ```Yeti_rig``` correctly outputs yeti data which can be then loaded via ```Asset Loader```

2) Publish doesnt output ```Sources``` folder with textures used in Yeti Graph into Publish folder

image

It should happen according to the last sentence in the user guides for Yeti:
image


3) When published Yeti asset loaded via ```Load Yeti Cache``` in Asset Loader it doesnt have correctly set on ```pgYetiMaya1Shape``` attrib for ```Other/Image Search Path``` to ```Publish Folder``` instead being set to orig file path to workfile area. * Note: I have inserted the file path manually when creating Yeti Setup as I was used to do it from production pov. I assume it would be left empty by OpenPype right now...no need to look after where it came from :)

image

There is also question how to handle publish/load of the Yeti setup according to Pergerine workflow guides. Will post more on this topic.

But generally it should be just:

- Select Yeti Setup and Publish it
- Select destination asset and Load Yeti Published Asset (will get connected to new selected Asset)
- Publish Yeti cache for animated sequence
- Load Yeti cache into lighting workfile scene for rendering (apply Look via Look Manager too)

Done.

openpype/hosts/maya/plugins/publish/collect_yeti_rig.py
```diff
@@ -31,6 +32,9 @@ def process(self, instance):
 
         input_connections = self.collect_input_connections(instance)
 
+        instance.extend(cmds.sets("input_SET", query=True))
+        instance.extend(get_all_children(cmds.sets("input_SET", query=True)))
```
BigRoy, line 36: Watch out.

`cmds.sets` returns None if the set is empty.

```python
from maya import cmds
my_set = cmds.sets(empty=True)
print(cmds.sets(my_set, query=True))
# None
```

BigRoy: Also, the `get_all_children` function itself does not filter to DAG nodes. So if the set in any way contains non dag nodes, like e.g. another set or dependency node the `get_all_children` function will likely error on `sel.getDagPath()` logic.

You'd need to make sure to pass only dagNode types.

```python
members = cmds.sets("input_SET", query=True) or []
children = get_all_children(cmds.ls(members, type="dagNode", long=True))
instance.extend(members)
instance.extend(children)
```

BigRoy: And another thing to watch out for. **If** the instance already contained any of the nodes that are in the `input_SET` then now the instance will have these nodes duplicated (or at least added more than once to the instance).

As such validators will check these nodes twice, and will likely report them more than once too.
It'll be slower + harder to read for the end user.

We might want to ensure uniqueness of what ends up in the instance here?

BigRoy: And actually reading in this now - why do we require all children to be collected of the `input_SET`. I recall that by original design that usage of that would not require that at all and wasn't intended like that **at all.** There is something fishy here.

A lot of discussion related to that I think is here: https://github.com/ynput/OpenPype/issues/3404

tokejepsen:
>There is also question how to handle publish/load of the Yeti setup according to Pergerine workflow guides. Will post more on this topic.

@LiborBatek think this need to be a separate issue cause this PR is just for fixing the broken pipeline. It'll need client support as well.

LiborBatek:
This is what happens when using ```Asset Loader``` and loading ```Yeti Rig``` asset in current state of things:

image

There is a whole character hierarchy and Yeti setup now. I guess its not desired to load it all despite missing shaders too.

openpype/hosts/maya/plugins/publish/collect_yeti_rig.py
BigRoy: It's best explained in [this comment](https://github.com/ynput/OpenPype/issues/3404#issuecomment-1238491559).

But in practice it means that the nodes in the `input_SET` are nodes that are **inside** your yeti rig.
So say it's `rig_GRP|input_GRP|rig_head_GEO`. When that particular node is found in the `input_SET` then it'll search for incoming connections to that transform and/or shape from any node outside of the rig.

The idea is that you e.g. build you yeti RIG like this:
- You load the referenced model for a character in pipeline, let's say that mesh is at `character/head_GEO`
- You create the yeti rig so that your yeti graph node uses a MESH inside the rig as input. This is `rig_GRP|input_GRP|rig_head_GEO`.
    - This gives the yeti rig full control on that geometry to do additional things on top of whatever input comes in, e.g. deleting faces, pushing faces in/out, subdividing, etc.
- You create a connection between the loaded character's head_GEO and the rig's rig_head_GEO. (e.g. `outMesh` -> `inMesh`).
- You add `rig_GRP|input_GRP|rig_head_GEO` to the `input_SET` because whenever loading the rig you want to re-apply its inputs.
- The yeti rig publisher will then collect the INPUTS found from the rig's head geo `rig_GRP|input_GRP|rig_head_GEO` so that it can later recreate that connection based on its `cbId` so that `{cbId}.output -> rig_head_GEO.input` gets recreated.

tokejepsen: So these docs are wrong?

https://openpype.io/docs/artist_hosts_maya_yeti#creating-and-publishing-yeti-rig

BigRoy: Yes. I'd say the the docs are wrong unless we change the logic of collecting the input set around.

At least currently the code logic doesn't match what the docs describe.

tokejepsen: Tried giving this workflow a go and something seems off. The publish works well and it produces a `.rigsettings` file with correct IDs and attributes but when loading, this information is taken into account. From the code it looks like IDs of the source and destination nodes need to be same and even then it'll only connect `outMesh` > `inMesh`; https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/maya/plugins/load/load_yeti_rig.py#L65-L77

BigRoy: That logic was implemented on OpenPype's side [here](https://github.com/ynput/OpenPype/commit/66544273fd08fc22aedc4267b5716d0dbb8abd23) and admittedly it's logic we (Colorbleed) thus had never used on our end on the intended implementation since it wasn't built to do so.

Instead it should be using the `sourceAttr` and `destAttr` stored in the rigSettings. I'd actually even go as far to by default NOT connect it on load unless someone ticks some "auto-connect" checkbox somewhere. We originally prototyped in our previous pipeline with a 'Connection Manager' which would show the possibilities of connections between A and B by the detected `cbId` entries.

BigRoy: > What do you think of using the same workflow as the Xgen implementation, where the "connection manager" is the inventory and the user selects which containers to connect together?

I think that's a solid first step after which we can look into the limitations of that workflow.

tokejepsen:
@BigRoy I've updated the docs now. Could you have a read through and let me know if that sounds correct?

openpype/hosts/maya/plugins/inventory/connect_yeti_rig.py
```diff
@@ -0,0 +1,132 @@
...
+                continue
+
+            cmds.connectAttr(
+                "{}.{}".format(source_node, input["connections"][0]),
+                "{}.{}".format(target_node, input["connections"][1])
+            )
+
+    def index_by_id(self, container):
+        reference_node = cmds.sets(container["objectName"], query=True)
+        members = cmds.referenceQuery(reference_node, nodes=True, dagPath=True)
```
BigRoy, line 99: Should this use [`openpype.hosts.maya.api.lib.get_container_members`](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/maya/api/lib.py#L1421-L1452)?

openpype/hosts/maya/plugins/inventory/connect_yeti_rig.py
```diff
@@ -0,0 +1,132 @@
...
+
+        ids = {}
+        for member in members:
+            if not cmds.objExists(member + ".cbId"):
+                continue
+            ids[cmds.getAttr(member + ".cbId")] = member
```
BigRoy, line 105: Use the [`openpype.hosts.maya.api.lib.get_id`](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/maya/api/lib.py#L1272-L1297) function.
```suggestion
            id = lib.get_id(member)
            if not id:
                continue
            ids[id] = member
```

Or:
```suggestion
            ids = {lib.get_id(member): member for member in members}
            ids.pop(None, None)  # ignore anything that didn't have an id
```

openpype/hosts/maya/plugins/inventory/connect_yeti_rig.py
```diff
@@ -0,0 +1,132 @@
...
+        return ids
+
+    def display_warning(self, message, show_cancel=False):
+        """Show feedback to user.
+
+        Returns:
+            bool
+        """
+
+        from Qt import QtWidgets
```
BigRoy, line 116: I believe this should be `qtpy` - also see #4250 
```suggestion
        from qtpy import QtWidgets
```

openpype/hosts/maya/plugins/inventory/connect_yeti_rig.py
```diff
@@ -0,0 +1,132 @@
...
+                    json.dumps(input, indent=4, sort_keys=True)
+                )
+                continue
+
+            cmds.connectAttr(
+                "{}.{}".format(source_node, input["connections"][0]),
+                "{}.{}".format(target_node, input["connections"][1])
+            )
+
+    def index_by_id(self, container):
```
BigRoy, line 97: `index_by_id` doesn't sound right. It's not returning indices right but members or nodes?

openpype/hosts/maya/plugins/inventory/connect_yeti_rig.py
```diff
@@ -0,0 +1,132 @@
...
+        reference_node = cmds.sets(container["objectName"], query=True)
+        members = cmds.referenceQuery(reference_node, nodes=True, dagPath=True)
+
+        ids = {}
+        for member in members:
+            if not cmds.objExists(member + ".cbId"):
```
BigRoy, line 103: Just for future reference, if you want to check for existence of an attribute explicitly then use `maya.cmds.attributeQuery` because `objExists` will also return True if you're checking for the attribute on a transform node which does not have the attribute but one of its child shapes does. Attribute query explicitly only checks the node you're passing in.

openpype/hosts/maya/plugins/inventory/connect_yeti_rig.py
```diff
@@ -0,0 +1,132 @@
...
+                self.log.debug(
+                    "Could not find nodes for input:\n" +
+                    json.dumps(input, indent=4, sort_keys=True)
+                )
+                continue
+
+            cmds.connectAttr(
+                "{}.{}".format(source_node, input["connections"][0]),
+                "{}.{}".format(target_node, input["connections"][1])
+            )
```
BigRoy, line 95: Should we check whether the attributes exist?
Please consider the suggestion psuedocode.

```suggestion
            source_attr, target_attr = input["connections"]
            
            if not cmds.attributeQuery(source_attr, node=source_node, exists=True):
                self.log.debug(
                    "Could not find attribute {} on node {} for input:\n{}".format(
                    source_attr, source_node, json.dumps(input, indent=4, sort_keys=True)
                )
                continue
            
            if not cmds.attributeQuery(target_node, node=target_attr, exists=True):
                self.log.debug(
                    "Could not find attribute {} on node {} for input:\n{}".format(
                    target_attr, target_node, json.dumps(input, indent=4, sort_keys=True)
                )
                continue
            
            cmds.connectAttr(
                "{}.{}".format(source_node, source_attr),
                "{}.{}".format(target_node, target_attr)
            )
```

openpype/hosts/maya/plugins/inventory/connect_yeti_rig.py
```diff
@@ -0,0 +1,132 @@
...
+        # Categorize containers by family.
+        containers_by_family = {}
+        for container in containers:
+            family = get_representation_context(
+                container["representation"]
+            )["subset"]["data"]["family"]
+            try:
+                containers_by_family[family].append(container)
+            except KeyError:
+                containers_by_family[family] = [container]
```
BigRoy, line 36: With a `defaultdict(list)` this could just be `containers_by_family[family].append(container)` - better?

openpype/hosts/maya/plugins/inventory/connect_yeti_rig.py
```diff
@@ -0,0 +1,132 @@
...
+
+        source_container = source_containers[0]
+        source_ids = self.index_by_id(source_container)
+
+        # Target containers.
+        target_ids = {}
+        inputs = []
+        for family, containers in containers_by_family.items():
+            if family in ["animation", "pointcache"]:
+                continue
```
BigRoy, line 62: Shouldn't this be like this?
```suggestion
            if family != "yetiRig":
                continue
```
Otherwise it'd also run over cameras, models, etc thinking they'd have ".rigsettings" or are we generalizing it in such a way that we suspect we'd have more families like this?
Maybe for now it's better to be explicit?

Do something like this:
```python
if family not in target_families:
    continue
```

And then have `target_families = {"yetiRig"}` on the class?

website/docs/artist_hosts_maya_yeti.md
```diff
@@ -44,34 +46,51 @@ You can now publish Yeti cache as any other types. **OpenPype → Publish**. It
...
+
+We'll need to prepare the scene a bit. We want some Yeti hair on the ball geometry, so duplicating the geometry, adding the Yeti hair and grouping it together.
 
-Yeti Rigs are working in similar way as caches, but are more complex and they deal with
-other data used by Yeti, like geometry and textures.
+![Maya - Yeti Hair Setup](assets/maya-yeti_hair_setup.png)
 
-Let's start by [loading](artist_hosts_maya.md#loading-model) into new scene some model.
-I've loaded my Buddha model.
+Now we need to connect the Yeti Rig with the animation rig. Yeti Rigs work by publishing the attribute connections from its input nodes and reconnect them later in the pipeline. This means we can only use attribute connections to get the animation from the rig to the Yeti Rig. Constraints or blendshapes are not supported. We'll connnect the Yeti Rig geometry to the animation rig, with the transform and mesh attributes.
```
BigRoy, line 79: When reading this my head goes "oh, boy.. that's limiting!" but in practice you can do whatever you want with constraints and blendshapes - those connections however would be internal to your rig.

So say you'd want to use an external mesh as blendshape in your rig you'd do:
- `.outMesh` -> `.inMesh` connection or alike from OUTSIDE the rig into a copy of the mesh INSIDE the rig.
- Then from that internal mesh you can do blendshapes just fine, etc.

I'm wondering whether we'd need to somehow elaborate on how that could work. I'm a tiny bit worried that someone starts to think they can't use e.g. blendshapes inside the yeti rig whatsoever.

website/docs/artist_hosts_maya_yeti.md
```diff
@@ -44,34 +46,51 @@ You can now publish Yeti cache as any other types. **OpenPype → Publish**. It
...
+### Loading
 
 You can load Yeti cache by **OpenPype → Load ...**. Select your cache, right+click on
 it and select **Load Yeti cache**. This will create Yeti node in scene and set its
 cache path to point to your published cache files. Note that this Yeti node will
 be named with same name as the one you've used to publish cache. Also notice that
 when you open graph on this Yeti node, all nodes are as they were in publishing node.
 
-### Creating and publishing Yeti Rig
+## Yeti Rigs
```
BigRoy, line 65: Should we document somewhere that the yetiRig can have any number of `pgYetiMaya` nodes and grooms as it'd like. It doesn't have to be just a single one. The nodes just must be uniquely named, that's all.

website/docs/artist_hosts_maya_yeti.md
```diff
@@ -81,28 +100,14 @@ the beginning of your timeline. It will also collect all textures used in Yeti
...
-under `input_SET`). This mechanism uses *cbId* attribute on those shapes.
-If match is found shapes are connected using their `outMesh` and `outMesh`. Thus you can easily connect existing animation to loaded rig.
-:::
+**Load Yeti Rig** item (among others).
+
+To connect the Yeti Rig with published animation, we'll load in the animation and use the Inventory to establish the connections.
+
+![Maya - Yeti Publish Setup](assets/maya-yeti_load_connections.png)
+
+The Yeti Rig should now be following the animation :tada:
```
BigRoy, line 113:
```suggestion
The Yeti Rig should now be following the animation. :tada:
```

BigRoy (commented):
Nice, definitely getting better with the docs now. Did have some pointers/questions.

openpype/hosts/maya/plugins/inventory/connect_yeti_rig.py
tokejepsen: Yup done

openpype/hosts/maya/plugins/inventory/connect_yeti_rig.py
tokejepsen: Cool done

openpype/hosts/maya/plugins/inventory/connect_yeti_rig.py
tokejepsen: yeah, good point

openpype/hosts/maya/plugins/inventory/connect_yeti_rig.py
tokejepsen: Thanks, will try and remember that

openpype/hosts/maya/plugins/inventory/connect_yeti_rig.py
tokejepsen: Yeah, nice. Less code.

openpype/hosts/maya/plugins/inventory/connect_yeti_rig.py
tokejepsen: Good point. Lets limit to not `yetiRig`.

website/docs/artist_hosts_maya_yeti.md
tokejepsen: Will elaborate on this.

website/docs/artist_hosts_maya_yeti.md
tokejepsen: Done

LiborBatek:
Following the User Docs in this PR I have succesfully tested the first part with ```YetiCache```  ```Publish``` and ```Load```

All files being published and can be loaded afterwards via ```Asset Loader```...

image

will follow with ```YetiRig``` instance workflow

LiborBatek:
Just before testing the ```YetiRig``` workflow after reading the User Docs for Yeti...

Have a few questions.

1) It seems user have to ```duplicate``` the geometry later used for ```pgYeti``` node as a groom mesh (also reconnect transforms and outmesh to the duplicated mesh)

What if it is desirable that the rigged asset already contains the geometry to be used for grooming or even directly the same geometry as for animation/rendering is used and not any separate geometry? Is it forbiden to use such a workflow?! Is user forced to do above mentioned manual setup?

@BigRoy how about Texture Reference objects? I dont see any mentions about them in our OP workflow but they are quite important as I m aware of standard Yeti workflows as stated in official docs. Yeti just need em to not jitter during animation right?

Just asking, will test it as it now... we will see where I get :)

BigRoy:
> What if it is desirable that the rigged asset already contains the geometry to be used for grooming or even directly the same geometry as for animation/rendering is used and not any separate geometry? Is it forbiden to use such a workflow?! Is user forced to do above mentioned manual setup?

Technically it CAN - but then the node in  `input_SET`  would be the pgYetiMaya node.

But really, don't look at it as "I can't connect my rig?" Sure you can!

```
Animation Rig mesh > Yeti Rig mesh > Yeti Rig pgYetiMaya node
```

It'd still be live

> @BigRoy how about Texture Reference objects? I dont see any mentions about them in our OP workflow but they are quite important as I m aware of standard Yeti workflows as stated in official docs. Yeti just need em to not jitter during animation right?

The texture reference object would be part of your `yetiRig` and thus would need no external connections. :) I think this workflow does require the duplicate meshes. In practice they shouldn't actually be duplicate meshes.

The **Yeti Rig mesh** would just have  Texture Reference object assigned and it would be part of your rig. So you're not assigning the Texture Reference object in the Animation Rig but in the Yeti Rig.

The Texture Reference object becomes a node of its own when creating it through the UI - and thus by having that in the rig you could even have that become an object of its own that is marked as input and then create complex rigs. (Not sure why you'd ever need that, but you could!)

tokejepsen:
> @tokejepsen are you able to follow along with my explanation here?

Pretty clear to me.

LiborBatek:
@BigRoy thx for in-depth explanation!

@tokejepsen  I think we will need to update the docs for both scenarios as @BigRoy outlined in his explanation and also put in the create ```texture reference object``` step in the tutorial ...now this step is missing. Also describing possible scenarios would be cool (yeti on the same geo, separate geo as we have now etc)

BigRoy:
Just to add as a note - would be good to mention in the docs explicitly too that the Yeti Rig can contain multiple yeti nodes/grooms - it doesn't have to be just a single one. The same goes for Yeti Cache publishes by the way.

tokejepsen:
@LiborBatek could you elaborate on what you want in the documentation?
The current state of the docs seem clear to me and I think would be clear for anyone coming from a rigging background.

LiborBatek:
Ok following the docs for creating ```YetiRig``` I have encountered an issue with unique IDs when ```texture Reference``` object is created... in your tutorial its a missing part so I have added it to my yeti setup  as @BigRoy  suggested its ok

image

BigRoy:
@LiborBatek thanks - I believe you should be able to "repair" and things would continue to work fine, but I can see how it's confusing. Since the geo here isn't to be rendered and used for lookdev it's fine to regenerate the IDs on them. It appears to me that Create Texture Reference object basically duplicates the geometry and thus would now have a non-unique id. The validator is correct to invalidate it.

Not sure how we could ease that for the artist to be less worrying to them.

LiborBatek:
Did more tests with ```load yetiRig``` into my scene and trying to use ```connect YetiRig``` but gives me errors 

image

here is the error I got after triggering the ```connect YetiRig``` action in Script editor:

```
#   File "D:\REPO\OP_3\openpype\tools\sceneinventory\view.py", line 502, in _process_custom_action
#     result = action.process(containers)
#   File "D:\REPO\OP_3\openpype\hosts\maya\plugins\inventory\connect_yeti_rig.py", line 105, in process
#     if not cmds.attributeQuery(
# TypeError: No object matches name: translate
```

BigRoy:
@LiborBatek could you try again?

LiborBatek:
@BigRoy thx, Works now ok!

Would be lovely to get some info popup msg about the connection has been made (and the entities?) so user has idea that it worked ok and is ready...

Speaking of those ```texture Reference``` objects should I give it another go with generating unique ```cbId``` or ?

BigRoy:
> Speaking of those texture Reference objects should I give it another go with generating unique cbId or ?

Yes, please - just to know whether that resolves the issue. Trigger the validation, right click "Repair" should work.
Then we can always look into streamlining the workflow later if it gets in the way too much or remains too confusing for production.

tokejepsen:
> Seems Like also when render we missing some pre render scripts...

@LiborBatek could you make an issue for this? Its a separate issue to this PR.

LiborBatek:
> > Speaking of those texture Reference objects should I give it another go with generating unique cbId or ?
> 
> Yes, please - just to know whether that resolves the issue. Trigger the validation, right click "Repair" should work. Then we can always look into streamlining the workflow later if it gets in the way too much or remains too confusing for production.

So yeah, the ```validator``` and the ```repair``` action works when using ```Texture Reference``` object without unique ID aka ```cbId```.

As you said its a bit confusing for user aka artist but once you are aware of it/reason its okeyish.

BigRoy:
> hmm seems like when opened the workfile with renderMain again it appeared in the rendersettings...

Yeti sometimes can be a bit picky on when exactly it sets the render callbacks. I believe it sets it on scene save too.

LiborBatek:
Another thing which is a bit weird is when scrubbing timeline checking ```pgYetiCache``` node on first frame aka ```1001``` its off the position... and its not just false in the viewport but also when rendered

image

when on frame ```1002``` all is ok

image

Not sure but maybe we should output the ```Yeti Cache``` in ```frame range``` with 1 frame handles for start/end  to be sure we have proper positions? But maybe its not related to frame range at all...

BigRoy:
> Not sure but maybe we should output the Yeti Cache in frame range with 1 frame handles for start/end to be sure we have proper positions? But maybe its not related to Yeti cache frame range at all...

It should work fine without handles as far as I know.

- Could you show your Yeti Cache instance settings from which you generated the publish?
- Could you show the Yeti Cache Subset in the loader with its start frame and end frame columns, etc?
- Could you show the publish UI? - _make it a bit wider; it should show the frame range it found for the instance to the right of the instance in the left column `yetiCacheMain (1001-1050)` or alike_
- What's your scene FPS? And workfile FPS?

tokejepsen:
> Not sure but maybe we should output the Yeti Cache in frame range with 1 frame handles for start/end to be sure we have proper positions? But maybe its not related to Yeti cache frame range at all...

Didnt notice any issues like that. Maybe its to do with the playback cache? Try disabling it.

LiborBatek:
Ok update here, it started to work with new published version! here it is with old one:

image

and then when updated to ver02:

image

Note: I didnt change any settings when doing ver02. Man soo weird to spent so much time realizing it suddenly works as expected!!!

BigRoy:
Interesting. Does that mean that currently the PR is working as expected?

LiborBatek (approved):
Did run all test all round from YetiRig publish including 2 Grooms on the mesh and ran through yeticaching and rendering it out.

All works as expected!

tokejepsen:
@BigRoy you happy with the code? Can merge this sucker then.

openpype/hosts/maya/plugins/inventory/connect_yeti_rig.py
```diff
@@ -0,0 +1,156 @@
...
+                        container["representation"]
+                    )["representation"]
+                )
+                _, ext = os.path.splitext(maya_file)
+                settings_file = maya_file.replace(ext, ".rigsettings")
+                if not os.path.exists(settings_file):
+                    continue
+
+                with open(settings_file) as f:
+                    inputs.extend(json.load(f)["inputs"])
```
BigRoy, line 77: For loop here now seems redundant

```suggestion
    yeti_rig_containers = containers_by_family.get("yetiRig")
    if not yeti_rig_containers:
        self.display_warning(
            "Select at least one yetiRig container"
        )
        return

    for container in yeti_rig_containers:
        target_ids.update(self.nodes_by_id(container))

        maya_file = get_representation_path(
            get_representation_context(
                container["representation"]
            )["representation"]
        )
        _, ext = os.path.splitext(maya_file)
        settings_file = maya_file.replace(ext, ".rigsettings")
        if not os.path.exists(settings_file):
            continue

        with open(settings_file) as f:
            inputs.extend(json.load(f)["inputs"])
```

openpype/hosts/maya/plugins/load/load_yeti_rig.py
```diff
@@ -22,72 +19,31 @@ class YetiRigLoader(openpype.hosts.maya.api.plugin.ReferenceLoader):
...
+                file_url,
+                namespace=namespace,
+                reference=True,
+                returnNewNodes=True,
+                groupReference=True,
+                groupName=group_name
+            )
 
-        c = colors.get('yetiRig')
+        settings = get_project_settings(os.environ["AVALON_PROJECT"])
```
BigRoy, line 39: Should we replace this with 
```python
from openpype.settings import get_current_project_settings
settings = get_current_project_settings()
```
To rely less on `os.environ['AVALON_PROJECT']` directly in the code?
I believe it's a goal to clean that up, right @iLLiCiTiT ?

openpype/hosts/maya/plugins/publish/validate_yeti_renderscript_callbacks.py
```diff
@@ -48,6 +48,11 @@ def get_invalid(cls, instance):
 
         yeti_loaded = cmds.pluginInfo("pgYetiMaya", query=True, loaded=True)
 
+        yeti_nodes = cmds.ls(type="pgYetiMaya")
+        if not yeti_nodes:
+            cls.log.info("No yeti nodes found.")
+            return False
```
BigRoy, line 54: If the yeti callback is set in your scene but yeti is NOT loaded then with this change the scene will not be invalidated even though I'd say the callback should be removed/cleaned up.

Additionally, this will always log a warning now in Maya's script editor if Yeti is not loaded because it'll warn that node type `pgYetiMaya` does not exist.

```suggestion
        if not yeti_loaded or not cmds.ls(type="pgYetiMaya"):
            cls.log.info("No yeti nodes found.")
            return False
```
This would be the same behavior as you had written but without the warning.
However, with this added code the `if not loaded` check below would never be used.

I personally feel we should be cleaning yeti callbacks from a scene that does not have yeti loaded.

BigRoy (commented):
Some small notes - and I still wonder whether we should mention in the docs that a yetiCache and yetiRig publish can contain multiple yeti nodes - just for clarity?

openpype/hosts/maya/plugins/inventory/connect_yeti_rig.py
```diff
@@ -0,0 +1,156 @@
...
+                        target_node,
+                        json.dumps(input, indent=4, sort_keys=True)
+                    )
+                )
+                continue
+
+            cmds.connectAttr(
+                "{}.{}".format(source_node, source_attr),
+                "{}.{}".format(target_node, target_attr)
+            )
```
BigRoy, line 121: I'm thinking now - what if you've updated your yetiRig which should now make new extra connections. You can't re-run this action because it'll error on some attributes already being connected. Should we first check if they are connected?

Plus we'd want to force override the connection if the target was already connected to another source.
```suggestion

            source_plug = "{}.{}".format(source_node, 
                                         source_attr)
            target_plug = "{}.{}".format(target_node, 
                                         target_attr)
            if cmds.isConnected(source_plug, 
                                target_plug,
                                ignoreUnitConversion=True):
                self.log.debug(
                    "Connection already exists: {} -> {}".format(
                        source_plug, target_plug
                    )
                )
                continue

            cmds.connectAttr(source_plug, target_plug, force=True)
            self.log.debug(
                "Connected attributes: {} -> {}".format(
                    source_plug, target_plug
                )
            )
```

LiborBatek:
> Some small notes - and I still wonder whether we should mention in the docs that a yetiCache and yetiRig publish can contain multiple yeti nodes - just for clarity?

Definetely yes (I think @tokejepsen already added it into docs), what comes to my mind is ```pgYeti node``` can have also multiple input geometries besides multiple groom nodes (which I succesfully tested already). Are we ok with this scenarios too? Maybe I should make more stress tests with more complex scenarios like this...

LiborBatek:
> Interesting. Does that mean that currently the PR is working as expected?

To your question: yes after that yeti cache publish v002 all worked ok (no that weird frame offset at the frame 1001)! I have then tested yeti_rig with multiple ```groom``` nodes and worked well too 

...but as I have mentioned above, I should also probably test scenario when more than one geometry is connected for grooming (which is perfectly ok in Yeti workflows)

tokejepsen:
> Some small notes - and I still wonder whether we should mention in the docs that a yetiCache and yetiRig publish can contain multiple yeti nodes - just for clarity?

@BigRoy I mention it [here](https://github.com/tokejepsen/pype/blob/bugfix/OP-3454_yeti-validate-rig-input/website/docs/artist_hosts_maya_yeti.md?plain=1#L80). Got anything to add to it?

openpype/hosts/maya/plugins/load/load_yeti_rig.py
tokejepsen: Done

openpype/hosts/maya/plugins/publish/validate_yeti_renderscript_callbacks.py
tokejepsen: > I personally feel we should be cleaning yeti callbacks from a scene that does not have yeti loaded.

Think that might be beyond the scope of this PR cause we might get into issues with how to differentiate between yeti callbacks and others.

BigRoy: Actually that logic is already in the validator. :)
This added code however breaks it because it never continues if the plugin is unloaded (and thus there are no yeti nodes)

openpype/hosts/maya/plugins/publish/validate_yeti_renderscript_callbacks.py
```diff
@@ -48,6 +48,10 @@ def get_invalid(cls, instance):
 
         yeti_loaded = cmds.pluginInfo("pgYetiMaya", query=True, loaded=True)
 
+        if not yeti_loaded or not cmds.ls(type="pgYetiMaya"):
+            cls.log.info("No yeti nodes found.")
+            return False
+
```
BigRoy, line 54:
```suggestion
        if not yeti_loaded and not cmds.ls(type="pgYetiMaya"):
            # The yeti plug-in is available and loaded so at
            # this point we don't really care whether the scene
            # has any yeti callback set or not since if the callback
            # is there it wouldn't error and if it weren't then
            # nothing happens.
            cls.log.info(
                "Yeti is loaded but no yeti nodes were found. "
                "Callback validation skipped.."
            )
            return False

```

openpype/hosts/maya/plugins/publish/validate_yeti_renderscript_callbacks.py
BigRoy: I think [this suggestion should work](https://github.com/ynput/OpenPype/pull/4554/files#r1140126733) to solve the issue for now.

BigRoy:
> on it [here](https://github.com/tokejepsen/pype/blob/bugfix/OP-3454_yeti-validate-rig-input/website/docs/artist_hosts_maya_yeti.md?plain=1#L80). Got anything to add to it?

Thanks - had missed that before. Looks good to me!

BigRoy (approved):
Solid work - code changes look good to me. However, **have not** done a test-run myself with the latest commits.


Also, I'd expect this PR to resolve this issue https://github.com/ynput/OpenPype/issues/3404 - Correct?

tokejepsen:
> Also, I'd expect this PR to resolve this issue https://github.com/ynput/OpenPype/issues/3404 - Correct?

Correct. Fixes #3404

tokejepsen:
> Solid work - code changes look good to me. However, have not done a test-run myself with the latest commits.

@LiborBatek would you mind doing one last test run on this?
My Yeti trial has just run out.

LiborBatek:
Yeah, Im on it! Will do another test run...

LiborBatek:
After testing round all seems fine even after newest commits....so go to go!

PR #4555, opened 2023-03-02
Data Exchange: Geometry in 3dsMax

moonyuet:
## Brief description
Introduces and updates a creator, extractors and loaders for model family
## Description
Introduces new creator, extractors and loaders for model family while adding model families into the existing max scene loader and extractor
- [x] creator
- [x]  adding model family into max scene loader and extractor
- [x]  fbx loader
- [x]  fbx extractor
- [ ]  usd loader
- [ ]  usd extractor
## Additional info
Make sure you have installed usd plugin 3dsMax through Autodesk Desktop App.

## Testing notes:
1. start with this step
2. follow this step

### DIFF
--- a/openpype/hosts/max/plugins/create/create_model.py
+++ b/openpype/hosts/max/plugins/create/create_model.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+"""Creator plugin for model."""
+from openpype.hosts.max.api import plugin
+from openpype.pipeline import CreatedInstance
+
+
+class CreateModel(plugin.MaxCreator):
+    identifier = "io.openpype.creators.max.model"
+    label = "Model"
+    family = "model"
+    icon = "gear"
+
+    def create(self, subset_name, instance_data, pre_create_data):
+        from pymxs import runtime as rt
+        sel_obj = list(rt.selection)
+        instance = super(CreateModel, self).create(
+            subset_name,
+            instance_data,
+            pre_create_data)  # type: CreatedInstance
+        container = rt.getNodeByName(instance.data.get("instance_node"))
+        # TODO: Disable "Add to Containers?" Panel
+        # parent the selected cameras into the container
+        for obj in sel_obj:
+            obj.parent = container
+        # for additional work on the node:
+        # instance_node = rt.getNodeByName(instance.get("instance_node"))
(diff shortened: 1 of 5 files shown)


LiborBatek:
Did test for ```model``` instance (max, ABC representations) 

Publishes ok also loads normally when loaded published model asset via ```Asset Loader```

Will continue testing with ```USD``` representation.

moonyuet:
The updated PR allows usd export with selected nodes only.

LiborBatek:
Also tested to ```publish``` a ```model``` in ```USD``` format.

All works as expected and ```model``` got exported in ```usd format```

image

Also tested to load in ```maya``` host and behave as expected

image

Not sure if ```triangulated``` model is desired and if correctly set during 3dsmax ```USD``` publish. Should not be triangulated by defaults.

moonyuet:
I tested the code a little bit in the spare time from the meeting yesterday. 
I find out that it is perhaps the limitation from the usdexporter in max. 
If I choose to export poly mesh and import it in maya, it can be loaded with some faces losing while some are triangulated with maya internal plugin. I will try to use multiverse usd to test and update you later after testing with multiverse usd plugin.
image

LiborBatek:
maybe @antirotor could tell us how it is with triangluation and USD...seems a bit odd (that we cant produce all quads geometry)

BigRoy:
> i will stick with quadmesh export option although the current usd exporter only consists of three according to [documentation](https://help.autodesk.com/view/MAXDEV/2023/ENU/?guid=MAXScript_USD_UsdExporter_html).

The `fromScene` option does seem to pretty much what you'd want right? Keep mesh/object data as close to the scene as possible? Instead of forcing quads or tris?

LiborBatek (approved):
All representations of the ```model``` successfully published.

image

If not just geometry inside ```Model``` Instance it correctly fails validations and gives user proper error msg:

image

mkolar (approved):

antirotor (changes requested):
Seems fine. I have only one note about the USD extractor. Since you need to install additional plugin to max to make it work (without it it fails on extractor with `AttributeError: 'pymxs.runtime' object has no attribute 'USDExporter'`), could we add validator for it? If there is no easy way to detect loaded plugin, even simple `try...except` on `pymxs.runtime.USDExporter` would do the trick probably.

antirotor (approved):

PR #4558, opened 2023-03-02
Editorial: Fix tasks removal

iLLiCiTiT:
## Brief description
Tasks won't disappear on editorial publish.

## Description
Editorial publishing which create tasks was always removing all other tasks if the asset already had some. With this PR, the other tasks kept.

## Testing notes:
1. Publish some editorial which creates assets and some tasks
2. Add any other task to a asset which was created using the publishing
3. Run the same editorial publishing again
4. The task from 2. should stay untouched


jakubjezek001 (approved):
Tested in Hiero and it worked as advertised!

PR #4563, opened 2023-03-02
SiteSync: host dirmap is not working properly

kalisp:
## Brief description
If artists uses SiteSync with real remote (gdrive, dropbox, sftp) drive, Local Settings were throwing error `string indices must be integers`.

## Description
Logic was reworked to provide only `local_drive` values to be overrriden by Local Settings. If remote site is `gdrive` etc. mapping to `studio` is provided as it is expected that workfiles will have imported from `studio` location and not from `gdrive` folder.

Also Nuke dirmap was reworked to be less verbose and much faster.

## Additional Info
This should be gone from Nuke logs:
2023-03-03 13_03_58-Window

## Testing notes:
1. Set SiteSync to be enabled on project
2. Set gdrive/dropbox/sftp remote site in Local Settings (might not be actually working)
3. Check in LS that no error is throwns
4. Download some Maya or Nuke workile from remote site, open it and check if dirmap doesn't fail.

### DIFF
--- a/openpype/host/dirmap.py
+++ b/openpype/host/dirmap.py
@@ -16,6 +16,7 @@ from openpype.lib import Logger
 from openpype.modules import ModulesManager
 from openpype.settings import get_project_settings
 from openpype.settings.lib import get_site_local_overrides
+from openpype.pipeline import Anatomy
 
 
 @six.add_metaclass(ABCMeta)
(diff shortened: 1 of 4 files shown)


jakubjezek001 (approved):
works

kalisp:
Confirmed by customer.

Issue #4479, opened 2023-02-16
Project Manager: fps input


antirotor:
**Running version**
3.15.1-nightly.4

**Describe the bug**
You cannot input fps in **OpenPype Project Manager** with more thatn 2 decimal precision. Also, the input for fps doesn't accept other decimal separator than the one set in system's locale.

**To Reproduce**
1. Go to Project Manage
2. Click on asset
3. Try to set fps to `23.976023976023978`, try to input with `,` as decimal separator instead.

**Expected behavior**
You should be able to input any arbitrary value in fps field. It should accept number with either `.` or `,` as decimal separators.

**Screenshots**
![image](https://user-images.githubusercontent.com/33513211/219399639-9a73c512-dc4b-4033-943e-9d2b1a589ab6.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Host: N/A

movalex:
Hey, check this out: https://github.com/movalex/OpenPype/commit/6fb1ff7a854394a293606ae73e216cae1aca56ef
I've set up the decimals number to 3 both in Project Manager and in the Project Anatomy settings.
I don't see the point to increase the decimals higher, because otherwise the fps counter will show 24.000000000 with whole numbers. There might be a [solution ](https://www.qtcentre.org/threads/44143-Hide-the-decimals-in-QDoubleSpinBox?p=201186#post201186) for this, but I'm not sure how to do that.
There's also added a step parameter for scroll adjustment of the decimals.

BigRoy:
@antirotor @iLLiCiTiT Does #4571 resolve this issue or are we still expecting more?

iLLiCiTiT:
It made it better, @antirotor can confirm if it's enough as we had issues with the small differences.

PR #4571, opened 2023-03-04
add up to 3 decimals precision to the frame rate settings

movalex:
## Brief description
Allows input 23.976 to the FPS settings both in Project Manager and the Project Anatomy.
Adds `setSingleStep` to the DoubleSpinBox editor with decimals.
Makes the FPS column in the Project Manager adjustable.

## Description
Based on this issue: https://github.com/ynput/OpenPype/issues/4479

## Testing notes:

* Go to Project Manager
* Click on asset
* Try to set fps to 23,976 with `,` as decimals separator.
* Select FPS cell and scroll the mouse, the value will be adjusted with 1 frame precision.
* Select Pixel Aspect cell and scroll the mouse, the value will be adjusted with 0.01 precision.
* Open Studio settings and check Project -> Anatomy -> Attributes. The Frame Rate field now have three decimals precision.

image
image

### DIFF
--- a/openpype/tools/project_manager/project_manager/delegates.py
+++ b/openpype/tools/project_manager/project_manager/delegates.py
@@ -83,15 +83,18 @@ class NumberDelegate(QtWidgets.QStyledItemDelegate):
         decimals(int): How many decimal points can be used. Float will be used
             as value if is higher than 0.
     """
-    def __init__(self, minimum, maximum, decimals, *args, **kwargs):
+    def __init__(self, minimum, maximum, decimals, step, *args, **kwargs):
         super(NumberDelegate, self).__init__(*args, **kwargs)
         self.minimum = minimum
         self.maximum = maximum
         self.decimals = decimals
+        self.step = step
 
     def createEditor(self, parent, option, index):
         if self.decimals > 0:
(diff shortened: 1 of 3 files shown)


mkolar (approved):
I've no quarrel with this. Should make live easier in some situations. @iLLiCiTiT could you please have a look at the code?

openpype/tools/project_manager/project_manager/delegates.py
```diff
@@ -83,15 +83,18 @@ class NumberDelegate(QtWidgets.QStyledItemDelegate):
...
         super(NumberDelegate, self).__init__(*args, **kwargs)
         self.minimum = minimum
         self.maximum = maximum
         self.decimals = decimals
+        self.step = step
 
     def createEditor(self, parent, option, index):
         if self.decimals > 0:
             editor = DoubleSpinBoxScrollFixed(parent)
+            editor.setSingleStep(self.step)
```
iLLiCiTiT, line 96: Step is set only for decimal and not for integers.

movalex: True, but I set the `step` to be a `setSingleStep` only if the editor is considered to be a DoubleSpinBox which works with decimals only.

PR #4572, opened 2023-03-04
Copy existing or generate new Fusion profile on prelaunch

movalex:
## Brief description
Fusion preferences will be copied to the predefined `~/.openpype/hosts/fusion/prefs` folder (or any other folder set in system settings) on launch.

## Description
The idea is to create a copy of existing Fusion profile, adding an OpenPype menu to the Fusion instance.
By default the copy setting is turned off, so no file copying is performed. Instead the clean Fusion profile is created by Fusion in the predefined folder. The default locaion is set to `~/.openpype/hosts/fusion/prefs`, to better comply with the other os platforms. After creating the default profile, some modifications are applied:

* forced Python3
* forced English interface
* setup Openpype specific path maps.

If the `copy_prefs` checkbox is toggled, a copy of existing Fusion profile folder will be placed in the mentioned location. Then they are altered the same way as described above. The operation is run only once, on the first launch, unless the `force_sync [Resync profile on each launch]` is toggled.
English interface is forced because the `FUSION16_PROFILE_DIR` environment variable is not read otherwise (seems to be a Fusion bug).

## Note
I accidentally deleted my previous PR, that has already had some conversation and reviews, because I'm a dummy. Sorry!
You can find it here: https://github.com/ynput/OpenPype/pull/4456


BigRoy:
I guess this new PR is intended to replace this one? https://github.com/ynput/OpenPype/pull/4456

movalex:
> I guess this new PR is intended to replace this one? #4456

exactly, the previous one deleted by mistake

BigRoy:
> exactly, the previous one deleted by mistake

Tried to match the reviewers, labels and milestone from the original PR just now. But can't seem to add back in @EmberLightVFX as reviewer. Other than that it seemed ok to set them for me.

openpype/hosts/fusion/deploy/fusion_shared.prefs
```diff
@@ -1,19 +1,20 @@
...
- },
- Script = {
-			PythonVersion = 3,
-			Python3Forced = true
-		},
+  Script = {
+    PythonVersion = 3,
+    Python3Forced = true
+    },
+  UserInterface = {
```
iLLiCiTiT, line 15: I know nothing about fusion but why is `UserInterface` filled? It looks like forcing a language and style.

movalex: Yeah, I noted that in description. If the profile is generated by fusion (i.e. the copy profile isn't toggled), all Fusion setting will be set to default, including the language interface. Menu localization is recent BMD `innovation`. I know very few people who uses native language UI interface. Besides, if the native locale is set up (in my case it is Russian), the custom Fusion profile directory variable is not read. This is probably a fusion bug. So this is seem to be an essential fix.

jakubjezek001:
> > exactly, the previous one deleted by mistake
> 
> Tried to match the reviewers, labels and milestone from the original PR just now. But can't seem to add back in @EmberLightVFX as reviewer. Other than that it seemed ok to set them for me.

I guess only @movalex can add @EmberLightVFX as reviewer.

movalex:
> > > exactly, the previous one deleted by mistake
> > 
> > 
> > Tried to match the reviewers, labels and milestone from the original PR just now. But can't seem to add back in @EmberLightVFX as reviewer. Other than that it seemed ok to set them for me.
> 
> I guess only @movalex can add @EmberLightVFX as reviewer.

I don't see such an option anywhere, should I look closer or do I need some special access rights for that?

BigRoy:
I think you can only add as reviewer registered contributors of the project and those who have replied to the PR. So I think @EmberLightVFX can be added as a reviewer only when he's commented on this PR. I want to emphasize that _I think_ that is the case.

EmberLightVFX:
Will take a look at this when I get some minutes over!

mkolar:
> I think you can only add as reviewer registered contributors of the project and those who have replied to the PR. So I think @EmberLightVFX can be added as a reviewer only when he's commented on this PR. I want to emphasize that _I think_ that is the case.

It is not the case. Only registered external contributors can be added as reviewers. And after the past few weeks i'd say tha @EmberLightVFX certainly earned that status. I'll sort it out.

openpype/hosts/fusion/deploy/fusion_shared.prefs
BigRoy: Since we're now copying the prefs anyway, should the master prefs maybe be 'generated' along with the copied prefs and maybe get generated e.g. per project from e.g. OpenPype Settings instead. Looking at what is being altered in this PR I have a feeling it might be worth being able to just customize the master prefs directly from OpenPype settings?
We would basically deploy from settings into the copied preferences folder.

Thoughts?

openpype/hosts/fusion/hooks/pre_fusion_setup.py
```diff
@@ -14,21 +17,108 @@ class FusionPrelaunch(PreLaunchHook):
     as set in openpype/hosts/fusion/deploy/fusion_shared.prefs to enable
     the OpenPype menu and force Python 3 over Python 2.
 
+    VERSION variable is used because from the Fusion v16 the profile folder
+    is project-specific, but then it was abandoned by devs,
+    and despite it is already Fusion version 18, still FUSION16_PROFILE_DIR
+    is used. The variable is added in case the version number will be
+    updated or deleted so we could easily change the version or disable it.
```
BigRoy, line 24: I feel like this comment should be in the docstring but directly above the `VERSION` variable. What do others think?

openpype/hosts/fusion/hooks/pre_fusion_setup.py
```diff
@@ -48,14 +138,26 @@ def execute(self):
...
+            force_sync,
+        ) = self.get_copy_fusion_prefs_settings()
+        if copy_status and fusion_profile_dir is not None:
+            prefs_source = self.get_profile_source()
+            self.copy_existing_prefs(prefs_source, fusion_profile_dir, force_sync)  # noqa
+        else:
+            fusion_profile_dir_variable = f"FUSION{self.VERSION}_PROFILE_DIR"
+            self.log.info(f"Setting {fusion_profile_dir_variable}: {fusion_profile_dir}")  # noqa
+            self.launch_context.env[fusion_profile_dir_variable] = str(fusion_profile_dir)  # noqa
+        master_prefs_variable = f"FUSION{self.VERSION}_MasterPrefs"
```
BigRoy, line 160:
```suggestion

        master_prefs_variable = f"FUSION{self.VERSION}_MasterPrefs"
```

I personally really prefer a little breathing space around some lines of code. For example here, to show that the master prefs logic here isn't one-to-one related directly the copying of the fusion profile dir.

Actually looking at the code involved with the copying of the Fusion profile dir I'm actually thinking of moving that full logic into a PreLaunchHook of its own.

So we have a dedicated one for setting the PYTHONHOME and a dedicated one for the potential copying of user prefs. So separating to a dedicated `FusionCopyPrefsPrelaunch` - anyone else feel like that separating those responsibilities would make for clearer code?

There's so much code here that looks intertwined and related to each other but in the end it doesn't rely on each other at all.

BigRoy (commented):
Overall it's looking really good already - I did have some remarks and I wonder what others think about it.

EmberLightVFX:
> Since we're now copying the prefs anyway, should the master prefs maybe be 'generated' along with the copied prefs and maybe get generated e.g. per project from e.g. OpenPype Settings instead. Looking at what is being altered in this PR I have a feeling it might be worth being able to just customize the master prefs directly from OpenPype settings?
> We would basically deploy from settings into the copied preferences folder.
> 
> Thoughts?

Yes please! That would be an awesome feature!!

movalex:
> > > exactly, the previous one deleted by mistake
> > 
> > 
> > Tried to match the reviewers, labels and milestone from the original PR just now. But can't seem to add back in @EmberLightVFX as reviewer. Other than that it seemed ok to set them for me.
> 
> I guess only @movalex can add @EmberLightVFX as reviewer.

I don't see such an option anywhere, should I look closer or do I need some special access rights for that?

> > Since we're now copying the prefs anyway, should the master prefs maybe be 'generated' along with the copied prefs and maybe get generated e.g. per project from e.g. OpenPype Settings instead. Looking at what is being altered in this PR I have a feeling it might be worth being able to just customize the master prefs directly from OpenPype settings?
> > We would basically deploy from settings into the copied preferences folder.
> > Thoughts?
> 
> Yes please! That would be an awesome feature!!

I'm not sure if that's already done by someone, but I recently did some rework on the lua-to-python project called [slpp](https://github.com/movalex/slpp), and now it converts the master prefs file to Python and vice versa. There's a [test file](https://github.com/movalex/slpp/blob/main/test_fusion_prefs.py) to try. I just wonder if we can incorporate this code here as is, or it should be rewritten. It would be awesome to be able to specify the preferences on prelaunch hook project-wise.

openpype/hosts/fusion/hooks/pre_fusion_setup.py
movalex: Do you mean of adding another prelaunch hook module just for the profile? I was thinkng about that, but could not get it to work. How should I trigger the new hook module? 
image

BigRoy:
> I'm not sure if that's already done by someone, but I recently did some rework (deleted a couple of symbols actually) on the lua-to-python project called [slpp](https://github.com/movalex/slpp). Now it converts the master prefs file to Python and vice versa. There's a [test file](https://github.com/movalex/slpp/blob/main/test_fusion_prefs.py) to try. I just wonder if we can incorporate this code here as is, or it should be rewritten. It would be awesome to be able to specify the preferences on prelaunch hook project-wise.

I was thinking we don't need `slpp` at all for this. We can just have a plain text field in the settings where you could put whatever content you want in the master prefs file. No need to encode/decode we can just write whatever is there into the master prefs file one to one. It'd mean that you could just copy whatever you in the .prefs file now directly into settings. No need for the Python lib dependency for that.

By the way, maybe a bit offtopic I had tried `slpp` in the past, I think five years ago and it basically failed at everything I threw at it. I had hoped to just throw e.g. Fusion `.comp` files at it directly but no luck at the time. Is that working better nowadays?

openpype/hosts/fusion/hooks/pre_fusion_setup.py
BigRoy: Yes. Just copy paste this file in the same folder, change the class name, etc. and separate the logic between the two. That should roughly do it.

movalex:
> By the way, maybe a bit offtopic I had tried `slpp` in the past, I think five years ago and it basically failed at everything I threw at it. I had hoped to just throw e.g. Fusion `.comp` files at it directly but no luck at the time. Is that working better nowadays?

I didn't try it with the comp file, but with the prefs file it did pretty well. I literally changed a single line for that. I wish I did a proper commit...

> No need for the Python lib dependency for that.

Well, pasting the whole master.prefs file in the text field would be much easier, however a bit more error prone. Miss a single curly brace, and the changes would not be applied with no notice.

BigRoy:
> Well, pasting the whole master.prefs file in the text field would be much easier, however a bit more error prone. Miss a single curly brace, and the changes would not be applied with no notice.

I'd say the idea is that a studio or admin eould have "their" master prefs already, or use the one from OpenPype or know how to manage them (or would need to look up how it works in Fusion anyway). Adding a layer on top would mean more logic to maintain, more logic to document and potentially even teach. I for example wouldn't want to even spend time rewriting my master prefs, and if I do I just secretly copy what I need from customizing my Fusion preferences through the UI and then taking those values into master prefs. Otherwise the admin would still first need to figure out the name of the key, the nesting, etc and then write that in Python logic and then test whether that actually encodes into something that works.

I think the entry point to failure is the same size if not even bigger due to the extra dependencies.

openpype/hosts/fusion/addon.py
```diff
@@ -3,6 +3,15 @@
 
 FUSION_HOST_DIR = os.path.dirname(os.path.abspath(__file__))
 
+FUSION_PROFILE_VERSION = 16
```
BigRoy, line 6: I'm not entirely sure about hardcoding this number? What happens if next version of Fusion they do remove the 16 or update it to 18 or 19. At that point in time, how do we support both versions of Fusion? We will still need to update the code logic, no? We would need be able to set both.

The logic here looks like it prepares for the variable being different/dynamic but the code doesn't really support that. As such, does it already make sense to add all these lines of code as 'preparation' where it doesn't really prepare is at all?

The code mentions it makes it easy to update the value. But if Fusion 19 comes out and they update it to 19 too then suddenly OpenPype will need to maintain both the 16 variant and the 19 variant to be backwards compatible. So it's not really as easy as just changing this single 16 to 19 since it's break all Fusion 18 projects.

What do you think?

openpype/hosts/fusion/addon.py
```diff
@@ -3,6 +3,15 @@
...
+
+# FUSION_PROFILE_VERSION variable is used by the pre-launch hooks.
+# Since Fusion v16, the profile folder became project-specific,
```
BigRoy, line 9: Do you mean version-specific instead of profile specific?

movalex:
@BigRoy 
This comment in `pre_fusion_setup.py` is not clear to me:
> # Fusion 18+ requires FUSION_PYTHON3_HOME to also be on PATH

It used to be more descriptive before:
```
Fusion 18+ does not look in FUSION16_PYTHON36_HOME anymore
but instead uses FUSION_PYTHON3_HOME and requires the Python to
be available on PATH to work. So let's enforce that for now.
```

So does Fusion18+ require BOTH variables to be set, or `FUSION_PYTHON3_HOME ` in the `Path` would be enough?

BigRoy:
> So does Fusion18+ require BOTH variables to be set, or FUSION_PYTHON3_HOME  in the Path would be enough?

Yes those two would be enough. This is what I meant. I couldn't get `FUSION_PYTHON3_HOME` to work of the path wasn't also added to `PATH`.

movalex:
According to our previous Discord conversation on how to manage the Fusion settings, so correct Fusion profile number would he used. I would simply restrict using other than default version convention in the `applications/fusion/variants` key. Of course, users still may change those default parameters, but I cannot imagine a single reason why they would want to. The version number is shown next to the app name in Applications attribute, and chaging the key in System settings, ditching the version number completely, would just would not make sense. Also this would require to change the corresponding project settings in `project_anatomy/attributes/applications`.  In a highly improbable situation, that the application key would not consist a valid Fusion version in any part of it, user will see this warning. What do you think? 

Screenshot 2023-03-16 024239

openpype/hosts/fusion/hooks/pre_fusion_profile_hook.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+
+    app_groups = ["fusion"]
+    order = 2
+
+    def get_fusion_profile_name(self, app_version) -> str:
+        # Returns 'Default', unless FUSION16_PROFILE is set
+        return os.getenv(f"FUSION{app_version}_PROFILE", "Default")
+
+    def check_profile_variable(self, app_version) -> Path:
```
BigRoy, line 20: Should this just be called `get_fusion_profile_dir`?

openpype/hosts/fusion/hooks/pre_fusion_profile_hook.py
```diff
@@ -0,0 +1,129 @@
...
+        fu_prefs_dir = f"Blackmagic Design/Fusion/Profiles/{fusion_profile}"
+        if platform.system() == "Windows":
+            profile_source = Path(os.getenv("AppData"), fu_prefs_dir)
+        elif platform.system() == "Darwin":
+            profile_source = Path(
+                "~/Library/Application Support/", fu_prefs_dir
+            ).expanduser()
+        elif platform.system() == "Linux":
+            profile_source = Path("~/.fusion", fu_prefs_dir).expanduser()
+        self.log.info(f"Got Fusion prefs file: {profile_source}")
```
BigRoy, line 49: The folder might not exist.

```suggestion
        self.log.info(f"Locating source Fusion prefs directory: {profile_source}")
```

openpype/hosts/fusion/hooks/pre_fusion_profile_hook.py
```diff
@@ -0,0 +1,129 @@
...
+    ) -> None:
+        """On the first Fusion launch copy the contents of Fusion profile
+        directory to the working predefined location. If the Openpype profile
+        folder exists, skip copying, unless re-sync is checked.
+        If the prefs were not copied on the first launch,
+        clean Fusion profile will be created in fu_profile_dir.
+        """
+        if copy_to.exists() and not force_sync:
+            self.log.info(
+                "Local Fusion preferences folder exists, skipping profile copy"
```
BigRoy, line 78: The preferences folder doesn't have be "Local". To me this is actually confusing me with the source directory.

```suggestion
                "Destination Fusion preferences folder exists, skipping profile copy"
```

openpype/hosts/fusion/hooks/pre_fusion_profile_hook.py
```diff
@@ -0,0 +1,129 @@
...
+            )
+            return
+        self.log.info(f"Starting copying Fusion preferences")
+        self.log.info(f"force_sync option is set to {force_sync}")
+        try:
+            copy_to.mkdir(exist_ok=True, parents=True)
+        except Exception:
+            self.log.warn(f"Could not create folder at {copy_to}")
+            return
```
BigRoy, line 87: This seems a bit too broad of an exception and should maybe debug log or alike the exception itself so we can debug what went wrong?

openpype/hosts/fusion/hooks/pre_fusion_profile_hook.py
```diff
@@ -0,0 +1,129 @@
...
+        if not copy_from.exists():
+            self.log.warning(f"Fusion preferences not found in {copy_from}")
+            return
+        for file in copy_from.iterdir():
+            if file.suffix in (".prefs", ".def", ".blocklist", ".fu"):
+                # convert Path to str to be compatible with Python 3.6+
+                shutil.copy(str(file), str(copy_to))
```
BigRoy, line 94: Out of curiosity - what's the reason we filter to only these extensions? Are there expected to be any files we explicitly don't want? Couldn't it just take the source folder and copy completely - or why shouldn't it?

BigRoy (commented):
@movalex I'm not sure what's going on but it looks like something went wrong with your PR. It seems to be reverting recent changes made in `develop`. Even if I check out your branch and pull in `develop` it doesn't seem to resolve the right changes.

For example, `develop` does not have the file `openpype/hosts/fusion/api/workio.py` anymore but your branch adds it back in. Additionally the GitHub diff also show changes from this PR https://github.com/ynput/OpenPype/pull/4571 and I'm not sure why because that should be in `develop` already.

openpype/hosts/fusion/api/workio.py
```diff
@@ -0,0 +1,38 @@
+"""Host API required Work Files tool"""
```
iLLiCiTiT, line 1: Why was this added back? Is was moved to `FusionHost`.

openpype/hosts/fusion/addon.py
```diff
@@ -1,7 +1,45 @@
...
+import re
 from openpype.modules import OpenPypeModule, IHostAddon
+from openpype.lib import Logger
 
 FUSION_HOST_DIR = os.path.dirname(os.path.abspath(__file__))
+FUSION16_PROFILE_VERSIONS = (16, 17, 18)
+FUSION9_PROFILE_VERSION = 9
+
+
+def get_fusion_profile_number(module: str, app_data: str) -> int:
```
iLLiCiTiT, line 11: Please remove Py3 syntax from addon public code.

BigRoy: Fusion does support Py2 but the logic implemented for it has been Py3-focused (e.g. f-strings) so has been Py3 only. Does that still make this problematic?

iLLiCiTiT: But this is public api of fusion addon (imported anywhere when `ModulesManager` is created).

BigRoy: I see - thanks!

openpype/hosts/fusion/hooks/pre_fusion_ocio_hook.py
```diff
@@ -7,6 +7,7 @@
 class FusionPreLaunchOCIO(PreLaunchHook):
     """Set OCIO environment variable for Fusion"""
     app_groups = ["fusion"]
+    order = 3
```
BigRoy, line 10: Why did this need an order?

openpype/hosts/fusion/hooks/pre_fusion_profile_hook.py
```diff
@@ -0,0 +1,152 @@
...
+
+class FusionCopyPrefsPrelaunch(PreLaunchHook):
+    """
+    Prepares local Fusion profile directory, copies existing Fusion profile.
+    This also sets FUSION MasterPrefs variable, which is used
+    to apply Master.prefs file to override some Fusion profile settings to:
+        - enable the OpenPype menu
+        - force Python 3 over Python 2
+        - force English interface
+        - force grey color scheme (because it is better that the purple one!)
```
BigRoy, line 21: This is a personal preference. We shouldn't make that a hardcoded setting.

openpype/hosts/fusion/hooks/pre_fusion_profile_hook.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def execute(self):
+        (
+            copy_status,
+            fu_profile_dir,
+            force_sync,
+        ) = self.get_copy_fusion_prefs_settings()
+
+        # Get launched application context and return correct app version
+        app_data = self.launch_context.env.get("AVALON_APP_NAME")
```
BigRoy, line 129: Personally I feel `app_name` would be a more readable variable name. Both here and in the function too. It's literally grabbing it from a value named app name.

openpype/settings/entities/schemas/projects_schema/schema_project_fusion.json
```diff
@@ -45,6 +45,29 @@
...
+                },
+                {
+                    "type": "boolean",
+                    "key": "copy_status",
+                    "label": "Copy profile on first launch"
+                },
+                {
+                    "key":"force_sync",
+                    "type": "boolean",
+                    "label": "Resync profile on each launch"
```
BigRoy, line 68: What happens if you were to enable always resync this but disable the option above? Would it still always sync?

Would this be easier as an enum setting `copy_local_prefs` or sync local prêts which has options "never", "first launch" or "always"?

And should first launch be more descriptive? It's not really only first launch. It's only if the destination settings folder does not exist. Can we come up with a short label that describes that better?

openpype/hosts/fusion/addon.py
```diff
@@ -1,8 +1,52 @@
...
+from openpype.lib import Logger
 
 FUSION_HOST_DIR = os.path.dirname(os.path.abspath(__file__))
 
+# FUSION_VERSIONS_DICT is used by the pre-launch hooks
+# The keys correspond to all currently supported Fusion versions
+# Each value is a list of corresponding Python home variables and a profile
+# number, which is used by the profile hook to set Fusion profile variables.
+FUSION_VERSIONS_DICT = {
+    9: ["FUSION_PYTHON36_HOME", 9],
```
BigRoy, line 13: Let's make the values tuples so they are immutable.

BigRoy (commented):
Very nice work on cleaning this up! Going to be a very nice feature. I'll be doing some test runs tomorrow.

openpype/hosts/fusion/deploy/fusion_shared.prefs
iLLiCiTiT: We do that with `workspace.mel` in Maya, it could have benefits. But in that case we should handle adding the core integration line:
```
["OpenPype:"] = "$(OPENPYPE_FUSION)/deploy",
```
For cases we would change name or path?

BigRoy (approved):
@movalex The functionality works for me. I've made some slight changes and pushed the commits into your branch. Please let me know if you dislike some of those changes made but it's mostly cleanup.

I still wonder whether we should [tweak the setting for enabling/force resync](https://github.com/ynput/OpenPype/pull/4572/files#r1139380617) but it does work fine as is and it's somewhat expected that it wouldn't sync if the top setting is disabled.

@jakubjezek001 @EmberLightVFX would be great if you can look at the code and test the functionality to see if this works for you.

---

The only feature that's not there is that the path can be dynamic based on Anatomy data since it doesn't get formatted but I feel like we could do that in a separate PR I'd say.

jakubjezek001 (approved):
LGTM ;)

PR #4573, opened 2023-03-06
Maya: SkeletalMesh family loadable as reference

simonebarbieri:
## Brief description
In Maya, fix the SkeletalMesh family not loadable as reference.

## Testing notes:
1. Have a rig published as SkeletalMesh.
2. Load that rig in the scene as reference.


LiborBatek (approved):
Did publish of ```skeletalMesh``` asset and then tried ```reference FBX``` the asset into shot scene.

image

It is possible to do so.

Note: Could be improved by coloring the asset name in the maya ```outliner```.

image

Issue #4553, opened 2023-03-01
Nuke: broken clip loader


jakubjezek001:
**Running version**
3.15.2-nightly.3

**Describe the bug**
```
Traceback (most recent call last):
  File "C:\CODE\__PYPE\OpenPype_3.15_up\openpype\tools\sceneinventory\view.py", line 241, in <lambda>
    lambda: _on_update_to_latest(items)
  File "C:\CODE\__PYPE\OpenPype_3.15_up\openpype\tools\sceneinventory\view.py", line 225, in _on_update_to_latest
    update_container(item, -1)
  File "C:\CODE\__PYPE\OpenPype_3.15_up\openpype\pipeline\load\utils.py", line 486, in update_container
    return loader.update(container, new_representation)
  File "C:\CODE\__PYPE\OpenPype_3.15_up\openpype\hosts\nuke\plugins\load\load_clip.py", line 255, in update
    representation
  File "C:\CODE\__PYPE\OpenPype_3.15_up\openpype\hosts\nuke\plugins\load\load_clip.py", line 237, in _representation_with_hash_in_frame
    representation["context"]["frame"] = hashed_frame
UnboundLocalError: local variable 'hashed_frame' referenced before assignment
```

**To Reproduce**
1. Go to Nuke test scene an load some image sequence subset version

**Expected behavior**
will normally load 

[cuID:[OP-5119](https://app.clickup.com/t/861meejzz)]

PR #4574, opened 2023-03-06
Nuke: fix clip sequence loading

jakubjezek001:
## Brief description
Nuke is loading correctly clip  from image sequence created without "{originalBasename}" token in anatomy template.

## Testing notes:
1. Open any testing workfile in Nuke
2. load any image sequence subset 
3. It should load without hiccups

# Resolving issue|
resolves https://github.com/ynput/OpenPype/issues/4553

### DIFF
--- a/openpype/hosts/nuke/plugins/load/load_clip.py
+++ b/openpype/hosts/nuke/plugins/load/load_clip.py
@@ -222,18 +222,21 @@ class LoadClip(plugin.NukeLoader):
         """
         representation = deepcopy(representation)
         context = representation["context"]
-        template = representation["data"]["template"]
+
+        # Get the frame from the context and hash it
+        frame = context["frame"]
+        hashed_frame = "#" * len(str(frame))
+
+        # Replace the frame with the hash in the originalBasename
         if (
-            "{originalBasename}" in template
-            and "frame" in context
+            "{originalBasename}" in representation["data"]["template"]
         ):
-            frame = context["frame"]
-            hashed_frame = "#" * len(str(frame))
             origin_basename = context["originalBasename"]
             context["originalBasename"] = origin_basename.replace(
                 frame, hashed_frame
             )
 
+        # Replace the frame with the hash in the frame
         representation["context"]["frame"] = hashed_frame
         return representation
 


openpype/hosts/nuke/plugins/load/load_clip.py
```diff
@@ -222,18 +222,21 @@ def _representation_with_hash_in_frame(self, representation):
         """
         representation = deepcopy(representation)
         context = representation["context"]
-        template = representation["data"]["template"]
+
+        # Get the frame from the context and hash it
+        frame = context["frame"]
```
iLLiCiTiT, line 227: The `"frame" in context` was there because if `{originalBasename}` is use `"frame"` key won't be on context. So this would crash with `KeyError`. So it should be rather:
```
        representation = deepcopy(representation)
        context = representation["context"]
        template = representation["data"]["template"]
        if (
            "{originalBasename}" in template
            and "frame" in context
        ):
            frame = context["frame"]
            hashed_frame = "#" * len(str(frame))
            origin_basename = context["originalBasename"]
            context["originalBasename"] = origin_basename.replace(
                frame, hashed_frame
            )

            representation["context"]["frame"] = hashed_frame
```

BigRoy:
```
        if (
            "{originalBasename}" in template
            and "frame" in context
        ):
```
With what you're saying then this would never be true, no? Unless the template had `{frame}` as well?

jakubjezek001: > The "frame" in context was there because if {originalBasename} is used, the "frame" key won't be on context.

@iLLiCiTiT  I haven't seen this usecase. Could you give me an example when this can happen? In Editorial it was always adding `frame` key to context data even `originalBasename` was set in publishing template. 

> With what you're saying then this would never be true, no? Unless the template had {frame} as well?
@BigRoy that is correct, this will never be true.

jakubjezek001: Actually @iLLiCiTiT has a point. Here is the code which is removing frame from the template, right? https://github.com/ynput/OpenPype/blob/767d6af84d5f9cc975093a1ac6093146ec685357/openpype/plugins/publish/integrate.py#L660

I presume it is only removed from formating data in integrating but not from context data, or am I wrong?

jakubjezek001: Anyway I hope it is clear that the pugins function `_representation_with_hash_in_frame` is only run in case the representation is image sequence. And to identify the representation as sequence there has to be more files under a representation. 

From the integration perspective it seems to me that we are adding representation with frame in context data allways in case if it is detected as image sequence.  So I do not see a problem with the code in the state as it is now.

BigRoy: Ah, quickly looking at the integrator code then `originalBaseName` publishes could indeed still have `frame` - interesting. And they could even have `udim` too? Do they? Where do those then get collected I wonder?

BigRoy: A udim sequence will get a value for `udim`, is still a sequence and would not have `frame`. I'd personally say the compositor should still be able to load it so that you could procedurally manipulate texture sets if needed.

64qam (approved):

openpype/hosts/nuke/plugins/load/load_clip.py
jakubjezek001: After personal discussion with @iLLiCiTiT we came to conclude that the case where `frame` would not in context.Data within clip loading might not happen and this is therefore mergable.

PR #4575, opened 2023-03-06
Unreal: Added settings for rendering

simonebarbieri:
## Brief description
Added settings for rendering in Unreal Engine.

## Description
Two settings has been added:
- Pre roll frames, to set how many frames are used to load the scene before starting the actual rendering.
- Configuration path, to allow to save a preset of settings from Unreal, and use it for rendering.

## Testing notes:
A layout will need to be loaded from Maya or Blender to be able to render the shot.
To render:
1. Create a render instance for a sequence.
2. Select the render instance and select `Render` from the OpenPype menu.

You can set the settings from the `Studio Settings/Project/Unreal Engine`.
For the `Pre-roll frames`, you can just set any number of frames. 
For the `Render Config Path`, it should work both with an empty setting (and in that case it will use some default settings), or by setting a path with the Unreal format (e.g. `/Game/Settings/Config.Config`).

### DIFF
--- a/openpype/hosts/unreal/api/rendering.py
+++ b/openpype/hosts/unreal/api/rendering.py
@@ -2,6 +2,7 @@ import os
 
 import unreal
 
+from openpype.settings import get_project_settings
 from openpype.pipeline import Anatomy
 from openpype.hosts.unreal.api import pipeline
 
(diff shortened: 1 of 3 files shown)


antirotor:
Since we are touching settings, shouldn't we add at least option to change file format? If I am not mistaken, now it renders to pngs hardcoded, but it should probably support more formats? Or is it a topic for another PR (along with the resolution that is now hardcoded?)

antirotor (approved):

antirotor (approved):

PR #4576, opened 2023-03-06
General: Input representation ids are not ObjectIds

iLLiCiTiT:
## Brief description
Don't use `ObjectId` as representation ids during publishing.

## Description
Representation ids are kept as strings during publishing instead of converting them to `ObjectId`. This change is pre-requirement for AYON connection.
Inputs are used for integration of links and for farm publishing (or at least it looks like).

### Links integration
1. Load (Reference) something into scene (fusion, houdini, maya)
2. Publish the scene
3. There should be integrated links in workfile version

### Farm publishing
1. Load (Reference) something into scene (fusion, houdini, maya)
2. Create a farm job
3. Input versions should be filled as expected

### DIFF
--- a/openpype/plugins/publish/collect_input_representations_to_versions.py
+++ b/openpype/plugins/publish/collect_input_representations_to_versions.py
@@ -22,8 +22,9 @@ class CollectInputRepresentationsToVersions(pyblish.api.ContextPlugin):
         # Query all version ids for representation ids from the database once
         representations = set()
         for instance in context:
-            inst_repre = instance.data.get("inputRepresentations", [])
-            representations.update(inst_repre)
+            inst_repre = instance.data.get("inputRepresentations")
+            if inst_repre:
+                representations.update(inst_repre)
 
         representations_docs = get_representations(
             project_name=context.data["projectEntity"]["name"],
@@ -31,17 +32,20 @@ class CollectInputRepresentationsToVersions(pyblish.api.ContextPlugin):
             fields=["_id", "parent"])
 
         representation_id_to_version_id = {
-            repre["_id"]: repre["parent"] for repre in representations_docs
+            str(repre["_id"]): repre["parent"]
+            for repre in representations_docs
         }
 
         for instance in context:
-            inst_repre = instance.data.get("inputRepresentations", [])
+            inst_repre = instance.data.get("inputRepresentations")
             if not inst_repre:
                 continue
 
-            input_versions = instance.data.get("inputVersions", [])
+            if "inputVersions" not in instance.data:
+                instance.data["inputVersions"] = set()
+            input_versions = instance.data["inputVersions"]
             for repre_id in inst_repre:
-                repre_id = ObjectId(repre_id)
-                version_id = representation_id_to_version_id[repre_id]
-                input_versions.append(version_id)
+                version_id = representation_id_to_version_id.get(repre_id)
+                if version_id:
+                    input_versions.add(version_id)
             instance.data["inputVersions"] = input_versions
--- a/openpype/plugins/publish/collect_input_representations_to_versions.py
+++ b/openpype/plugins/publish/collect_input_representations_to_versions.py
@@ -37,7 +37,7 @@ class CollectInputRepresentationsToVersions(pyblish.api.ContextPlugin):
         }
 
         for instance in context:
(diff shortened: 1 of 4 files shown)


BigRoy (commented):
Overall I like the tweak. However, I wanted to bring up a concern.

I'm not entirely sure but I believe that turning this into a `set` instead of `list` does lose the potential to know how many times the content was loaded. I believe previously it would have been possible to see that e.g. tree model v008 was loaded 10 times or once.

This was a discussion at the time of the implementation and I wonder whether we should lose that or not. I think it's actually worth something to know how many times the version was used in the content instead of only knowing whether it was used at all.

I'd love to know if e.g. 100 v003 were used of a tree and only 1 on v004 or vice-versa. It's good data, no?

I believe before this change it did actually collect the same loaded versions multiple times as an input link (though I'm not 100% sure)

iLLiCiTiT:
> I'm not entirely sure but I believe that turning this into a set instead of list does lose the potential to know how many times the content was loaded. 

Ah, I didn't know. Good point will change back.

openpype/plugins/publish/collect_input_representations_to_versions.py
```diff
@@ -22,26 +22,30 @@ def process(self, context):
         # Query all version ids for representation ids from the database once
         representations = set()
         for instance in context:
-            inst_repre = instance.data.get("inputRepresentations", [])
-            representations.update(inst_repre)
+            inst_repre = instance.data.get("inputRepresentations")
+            if inst_repre:
+                representations.update(inst_repre)
```
BigRoy, line 27: Anyone else prefers the old `.get()` style just because it hints what type we're expecting here?

openpype/plugins/publish/collect_input_representations_to_versions.py
```diff
@@ -22,26 +22,30 @@ def process(self, context):
...
         for instance in context:
-            inst_repre = instance.data.get("inputRepresentations", [])
+            inst_repre = instance.data.get("inputRepresentations")
             if not inst_repre:
                 continue
 
-            input_versions = instance.data.get("inputVersions", [])
+            if "inputVersions" not in instance.data:
+                instance.data["inputVersions"] = []
+            input_versions = instance.data["inputVersions"]
```
BigRoy, line 46: This would be better written as:
```suggestion
input_versions = instance.data.setdefault("inputVersions", [])
```

Also since we're now assigning it beforehand to `instance.data["inputVersions"]` I feel the `instance.data["inputVersions"] = input_versions` afterwards is redundant since we're basically assigning the same list again.

BigRoy (commented):
Haven't tested the code - but logically it looks ok to me. I did have some style related comments.

openpype/plugins/publish/collect_input_representations_to_versions.py
```diff
@@ -23,25 +23,27 @@ def process(self, context):
...
 
         representation_id_to_version_id = {
-            repre["_id"]: repre["parent"] for repre in representations_docs
+            str(repre["_id"]): repre["parent"]
+            for repre in representations_docs
         }
 
         for instance in context:
-            inst_repre = instance.data.get("inputRepresentations", [])
+            inst_repre = instance.data.get("inputRepresentations")
```
BigRoy, line 40:
```suggestion
            inst_repre = instance.data.get("inputRepresentations", [])
```
This one too then?

iLLiCiTiT: I wish we could use typehints...

openpype/plugins/publish/collect_input_representations_to_versions.py
```diff
@@ -23,25 +23,27 @@ def process(self, context):
...
         }
 
         for instance in context:
             inst_repre = instance.data.get("inputRepresentations", [])
             if not inst_repre:
                 continue
 
-            input_versions = instance.data.get("inputVersions", [])
+            instance.data.setdefault("inputVersions", [])
+            input_versions = instance.data["inputVersions"]
```
BigRoy, line 45:
```suggestion
            input_versions = instance.data.setdefault("inputVersions", [])
```

`dict.setdefault` already returns the value itself.

BigRoy (approved):
Haven't tested the code, but changes look good to me.

PR #4580, opened 2023-03-06
Fusion: Load sequence fix filepath resolving from representation

BigRoy:
## Brief description

[Resolves issue mentioned on discord](https://discord.com/channels/517362899170230292/563751989075378201/1082272423212224593) by @movalex:
> I launch openpype menu, choose Load - plateMain - right click - load sequence (mov) and instead of the plate the thumbnail file appears in the created loader. 
> there's two files in the publish folder.
> One is the plate, second is the thumbnail.

The loader was incorrectly trying to find the file in the publish folder which resulted in just picking 'any first file'.

## Description

This gets the filepath from representation instead of taking the first file from listing files from publish folder.

## Testing notes:
1. Load an image sequence (or video file)
2. Make sure to test a case where the publish folder also contains other files than the image sequence (or video file), like e.g. a thumbnail representation or a review file.

### DIFF
--- a/openpype/hosts/fusion/plugins/load/load_sequence.py
+++ b/openpype/hosts/fusion/plugins/load/load_sequence.py
@@ -1,11 +1,9 @@
-import os
 import contextlib
 
-from openpype.client import get_version_by_id
-from openpype.pipeline import (
-    load,
-    legacy_io,
-    get_representation_path,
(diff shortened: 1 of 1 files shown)


jakubjezek001 (approved):
works as advertized ;)

PR #4583, opened 2023-03-07
Global: persistent staging directory for renders

kalisp:
## Brief description
Allows configure if staging directory (`stagingDir`) should be persistent with use of profiles.

## Description
With this feature, users can specify a transient data folder path based on presets, which can be used during the creation and publishing stages. In some cases, these DCCs automatically add a rendering path during the creation stage, which is then used in publishing.

One of the key advantages of this feature is that it allows users to take advantage of faster storages for rendering, which can help improve workflow efficiency. Additionally, this feature allows users to keep their rendered data persistent, and use their own infrastructure for regular cleaning.

However, it should be noted that some productions may want to use this feature without persistency. Furthermore, there may be a need for retargeting the rendering folder to faster storages, which is also not supported at the moment.

It is studio responsibility to clean up obsolete folders with data.

Location of the folder is configured in `project_anatomy/templates/others`. ('transient' key is expected, with 'folder' key, could be more templates)

Which family/task type/subset is applicable is configured in:
`project_settings/global/tools/publish/transient_dir_profiles`

## Additional info
During publishing of renders, they are first rendered to work dir then publishing process will create reviews and thumbnails and only after that all of it is copied/moved to publish folder.

With renders, it will be much safer to render them to some common staging/transient folder and do the publishing from there. This way artists/other processes won't corrupt yet unpublished data so easily.

As example in Maya:
Renders in Maya are now done from published scene. We could change image prefix only in published scene to point to that staging directory, so farm will render there but scene that artist has open will point still to work dir.

## Testing notes:
1. Check configuration and modify folder path for transient stagingDir in `project_anatomy/templates/others`
2023-03-16 17_24_43-Window

3. Configure profiles in `project_settings/global/tools/publish/custom_staging_dir_profiles`
2023-03-21 18_23_54-Window

5. Publish and check that temporary files are in specific location configured in step 1


BigRoy:
Reading this PR I'm not entirely sure about how this would end up working. It does two things:

- It changes the staging dir to another folder
- It marks it as persistent.

However, not all extractors 'listen' to the the staging dir being the decision on where extracted files end up. Nuke render, Houdini extractors, Maya render, Fusion renders all take the output location from the actual settings of the scene or a node in the scene.
Some of those even update the `instance.data["stagingDir"]` at the extractor too, for example [here](https://github.com/ynput/OpenPype/blob/4fe053b109d892a5b5f3770be693ae72d1c19967/openpype/hosts/houdini/plugins/publish/extract_alembic.py#L25). As such, now suddenly that folder would be marked as persistent instead of the templated path from the settings?

As such, for those cases 'changing the staging dir' to a template would not work as expected, right? However it'd still get marked "persistent" but then for a different path then expected.

Also for the cleanup plug-in. More and more extractors end up setting `stagingDir` on representations. Should the cleanup plug-in check those too?

I feel like this PR is mixing up some of the responsibilities or might be relying on how things work in publishing where they don't actually always work like that.

jakubjezek001:
> Should this maybe just be setting cleanup's **skipCleanupFilepaths** context data to just ensure it's skipped? Or should there just be a **skipCleanupFolders** data instead to just skip any deletion of anything that's a file or folder that starts with that path?

This is certainly the option but I would rather recomand to add bool togle to preset in `CollectTransientStaging` for explicitly define if data are persistent or not. I beleve `Persistent` as name of attribute would be best.

BigRoy:
> This is certainly the option but I would rather recomand to add bool togle to preset in CollectTransientStaging for explicitly define if data are persistent or not. I beleve Persistent as name of attribute would be best.

I mean, isn't it all persistent if the cleanup wouldn't be triggered?

Somehow I feel this is much more about tweaking the cleanup plug-in to not clean up in certain cases than being about defining custom templates that get set that end up being persistent. I'm not entirely sure what that custom staging dir path would be for - why would it only be persistent if that path gets customized, etc.? I feel like two concepts are being mixed up here:
1. Set a custom output location in certain situations
2. Do or do not clean up in certain situations

Sure, the case of 1 + 2 could overlap, but to me they appear to be separate things - this PR seems to force them to get mixed together.

jakubjezek001 (changes requested):
> However, not all extractors 'listen' to the the staging dir being the decision on where extracted files end up. Nuke render, Houdini extractors, Maya render, Fusion renders all take the output location from the actual settings of the scene or a node in the scene.
Some of those even update the instance.data["stagingDir"] at the extractor too, for example [here](https://github.com/ynput/OpenPype/blob/4fe053b109d892a5b5f3770be693ae72d1c19967/openpype/hosts/houdini/plugins/publish/extract_alembic.py#L25). As such, now suddenly that folder would be marked as persistent instead of the templated path from the settings?

We are hitting this issue. I was testing on Nuke with Render family and the originally created path to `work/render` folder was used. Presumably if I create preset for overiding render family in Nuke, this plugin should do the trick, even the path had been forced in Create. In other way of implementation we would need to offer only limited amouth of families or hosts which would work, because this would be very confusing.

kalisp:
After discussion (Jezza, Ondrej, PK):
- this is not question only of collecting, but it should be applicable for Creators too(for DCC that store path value into read/write nodes - Nuke, Houdini etc.
- so logic to get profile should be yanked out into separate function (current collector will use it, creators could use it too)
- shouldn't be part of `pipeline.publish.lib.get_instance_staging_dir`, it should be used before, eg. to fill `stagingDir` on an instance explicitly
- this function will provide you with transient dir info (template, persistence)
- it is responsibility of collector or creator to fill the template
- we probably want to add new placeholder into templates `{scene_name}` to differentiate transient dir based on name of workfile (eg. version)
- each hosts which need to use this in Creators will use new function (issues will be created to track progress)
- configuration will be moved to `project_settings/global/tools/publish/transient_dir_profiles` (as it is not question of just collector anymore)

BigRoy:
> After discussion (Jezza, Ondrej, PK):

Thanks for concluding and sharing some notes, but as an outsider to the discussion I really can't follow along with what exactly the bullet points mean and with that what brought certain decisions and what will need to be done in the end. It seems to mention refactoring some logic but I don't see any point that tries to resolve the issues mentioned [here](https://github.com/ynput/OpenPype/pull/4583#issuecomment-1458228250).

If possible, please try to make sure the debrief in itself is clear for the community to understand.

Some notes/questions:

> this is not question only of collecting, but it should be applicable for Creators too(for DCC that store path value into read/write nodes - Nuke, Houdini etc.

This wouldn't work for a dynamic template though, like `{scene_name}`. I think if those apps require rendering into custom destination folders that it's the responsibility of the Extractor to adhere to it.

> so logic to get profile should be yanked out into separate function (current collector will use it, creators could use it too)

This describes to move the get profile logic into a separate function. What makes this getting profile so different from others that filter profiles?

> this function will provide you with transient dir info (template, persistence)

Which function? :)

> it is responsibility of collector or creator to fill the template

Why or why not? Unclear really what this means in the end.

> each hosts which need to use this in Creators will use new function (issues will be created to track progress)

Again, see my earlier point about how tokens might resolve to a new path if dynamic.

---

Getting back however to the original PR description:

> Transient staging dir could be useful in specific use cases where is desirable to have temporary renders in specific, persistent folders.

Could you provide a use case where this is a necessity in production? Why would we write a specific case/task/instance or whatever into a unique work output folder? 

If it's in particular about having "persistent data" somewhere that is not the publish location why not just have a dedicated integrator which also just copies the files to the persistent data location once its completed? It'd be the same result after publishing - files in custom location and persist. The staging directory then remains a temporary folder, files just become persistent elsewhere.

openpype/settings/defaults/project_settings/global.json
```diff
@@ -591,6 +591,21 @@
...
+                    "task_types": [],
+                    "task_names": [
+                        "transient_staging"
+                    ],
+                    "families": [],
+                    "subsets": [],
+                    "use_transient_dir": true,
+                    "transient_dir_persistent": true
+                }
             ]
```
BigRoy, line 609: I don't think we'd need a default setting like this, right?

openpype/settings/entities/schemas/projects_schema/schemas/schema_global_tools.json
```diff
@@ -408,6 +408,60 @@
...
+                                "object_type": "text"
+                            },
+                            {
+                                "type": "separator"
+                            },
+                            {
+                                "key": "use_transient_dir",
+                                "label": "Use Transient Folder",
+                                "type": "boolean"
+                            },
```
BigRoy, line 457: If we have a profile set, why would this ever be disabled?

openpype/settings/entities/schemas/projects_schema/schemas/schema_global_tools.json
```diff
@@ -408,6 +408,60 @@
...
+                            {
+                                "key": "transient_dir_persistent",
+                                "label": "Transient Folder Persistent",
+                                "type": "boolean"
+                            }
```
BigRoy, line 462: Is this missing a setting where it sets what template it should use - it appears hardcoded now to use the `transient` template. But what if we want VDB caches on a different disk than say renders?

We should be able to define multiple templates, right?

openpype/plugins/publish/collect_transient_folder.py
```diff
@@ -57,6 +57,9 @@ def process(self, instance):
         result_str = "Not adding"
         if transient_tml:
             anatomy_data["root"] = anatomy.roots
+            scene_name = instance.context.get("currentFile")
+            if scene_name:
+                anatomy_data["scene_name"] = scene_name
```
BigRoy, line 62: Note that `currentFile` is the full path.

openpype/settings/entities/schemas/projects_schema/schemas/schema_global_tools.json
kalisp: I think this is just question of preference, this way it might be easier to hot wire it and deactivate for debugging/production issue.

openpype/settings/defaults/project_settings/global.json
kalisp: Yeah, misfire.

openpype/settings/entities/schemas/projects_schema/schemas/schema_global_tools.json
BigRoy: I'd say if it's a global enable/disable for the entry in the list maybe it's better to name the key `enabled` like many other parts of settings do. And if somehow possible it'd be extra nice if it's like a header checkbox slider on this entry like seen in other areas of the UI.

jakubjezek001:
> And would the custom location override the OPENPYPE_TMPDIR variable or vice versa, which should take precedence?

They can coexist, but yes preference is on the transient data activated profile - in case it is set to instance.data["stagingDir"] via the collector plugin then `OPENPYPE_TMPDIR` is not gonna be used since it is only activated in case `stagingDir` key is missing on instance.data.

> Anyway, I still feel like this 'feature' is mixing two things:
> 1. A custom (default) output location for workfile output based on specific profile, say renders always go to drive X:/
> 2. Disabling the cleanup feature for specific profiles or locations (or vice-versa, enabling it for certain profiles/locations)

Don't disagree but I am not seing why we should not have two componets to be part of one design. Componet one is telling that preseted path should be used as stagingDir and with second component we can flag it for persistence. Would you rather spearated it into another layer of presets?

### Usecase from Nuke perspective for persistent transient data preset:
- setting transient data preset for:
   task: comp
   family: render
- Creator plugin will add resolved path into write node (without version or scene_name)
- preset is defining the folder as `persistent`
- user is overriding content of the folder anytime he renders - as it used to be untill now - except now it can be separated from work storage.

### Usecase from Nuke perspective for temporary transient data preset:
- setting transient data preset for:
   task: comp
   family: prerender
- Creator plugin will add resolved path into write node (without version or scene_name)
- preset is defining the folder as not `persistent`
- After rendering and publishing the `folder` will be romoved since it can be loaded from published version.

But yes if the cleanup plugin is not working now it should be fixed.

openpype/settings/entities/schemas/projects_schema/schemas/schema_global_tools.json
jakubjezek001: Here is some example from Flame and if it is really just about hot wiring, it might make more sense to make it this way
image

BigRoy:
> Custom Intermediate Output Directory

I'd personally pick Custom Staging Directory because it's doing exactly that, overriding the staging directory.

kalisp:
Renamed according to comments.

openpype/pipeline/publish/lib.py
```diff
@@ -690,3 +692,77 @@ def get_publish_repre_path(instance, repre, only_published=False):
...
+        subset_name (str)
+        anatomy (Anatomy) (optional)
+        log (Logger) (optional)
+
+    Returns:
+        (tuple)
+    Raises:
+        ValueError - if misconfigured template should be used
+    """
+    settings = get_project_settings(project_name)
```
BigRoy, line 717: As optimization we should make passing `project_settings` directly possible too so it doesn't require the query.

openpype/pipeline/publish/lib.py
```diff
@@ -690,3 +692,77 @@ def get_publish_repre_path(instance, repre, only_published=False):
...
+
+
+def _validate_transient_template(project_name, template_name, anatomy):
+    """Check that transient template is correctly configured.
+
+    Raises:
+        ValueError - if misconfigured template
+    """
+    if template_name not in anatomy.templates:
+        raise ValueError(("!!! Anatomy of project \"{}\" does not have set"
```
BigRoy, line 761: I'm seeing this often in the codebase that logs or messages get prefixed with things like `!!!` or `__ ` and it has bothered me so many times - haha. What's the reasoning behind this? It feels like the `!!!` mark something as 'hey something bad happened' but isn't a raised error exactly something bad that happened to begin with? Why do we need to add all these prefixes to our errors or logs?

[An example of the `__ ` is here](https://github.com/ynput/OpenPype/blob/e4085f768805e66c1ca58766101891922c0e963f/openpype/plugins/publish/extract_review_slate.py#L220-L231).

It has been making error messages hard to read/parse for me, but maybe there are reasons others like it.

openpype/settings/defaults/project_anatomy/templates.json
```diff
@@ -58,12 +58,16 @@
             "file": "{originalBasename}.{ext}",
             "path": "{@folder}/{@file}"
         },
+        "transient": {
+            "folder": "{root[work]}/{project[name]}/{hierarchy}/{asset}/work/{family}/{subset}"
+        },
```
BigRoy, line 63: Is it weird that we're adding this default but not have default custom staging dir profiles defined using this? I can see in the code that it defaults to `transient` if the template name setting is kept empty - but isn't that maybe a bit too implicit?

Maybe we should focus on creating good documentation for the feature as opposed to needing a default template key defined but enforce an explicit template value to be set with it.

openpype/settings/entities/schemas/projects_schema/schemas/schema_global_tools.json
BigRoy, line 1: Just wondering - should custom staging dir profiles maybe also be based on "sites"? So that e.g. a freelancer which might not have access to a certain custom staging dir location to ignore it? And if so, how do we feel about that data potentially not persisting?

I feel like that with remote working / site sync this feature can currently get in the way rather rapidly?

openpype/pipeline/publish/lib.py
kalisp: Copied & pasted from existing code, cleaned up.

openpype/pipeline/publish/lib.py
```diff
@@ -690,3 +692,79 @@ def get_publish_repre_path(instance, repre, only_published=False):
...
+        "task_names": task_name,
+        "task_types": task_type,
+        "subsets": subset_name
+    }
+    profile = filter_profiles(custom_staging_dir_profiles,
+                              filtering_criteria,
+                              logger=log)
+
+    if not profile or not profile["active"]:
+        return None, None
```
BigRoy, line 742: Shouldn't we be filtering the profiles on `active` state BEFORE filtering them since filtering profiles is based on finding the "best match". What if the first entry was a global matching profile (e.g. match anything) and it was set to inactive. Then it wouldn't ever match any of the profiles here because it could filter to that one and then ignore the others?

Same goes if e.g. I'd make a copy of one of my filter profiles that I like, but want to test a slightly different one - I'd deactivate the original, activate the new one but the profile filtering here might still be taking the inactive one.

BigRoy:
So the feature at this stage seems relatively clean. However I still feel like this should come with a major "watch out" label that it won't work for a lot of things where it'd break:
- Rendering from maya, rendering over the farm, extractors from Houdini and Fusion will fail to work with this. Since those each define staging dir depending on how the scene is set to output it.
- Site Sync with remote artists will get funky with this feature if the remote artist cannot access this custom staging dir location to write to for extraction.
- Extra worried for cases where it might trigger on a farm but the custom staging dir being set to some sort of "local directory" or alike where other machines might not be able to access the path.
- Since this is focused on writing to e.g. a custom drive because speed of writing might be much faster there extra care needs to be taken for defining a path that is accessible cross-OS, e.g. the path might need to be able to format correctly for both linux and windows. Is that possible with the path formatting that is currently possible, could we format it with e.g. an environment variable, or do we need to first set up a custom root and then use that root?

I think with this PR having these issues noted and maybe explained on how they could be worked around (e.g. the last bullet point) in the documentation is a must. We should have something we can point to that explains what people might be running into with this that we can anticipate.

website/docs/project_settings/settings_project_global.md
```diff
@@ -263,10 +263,25 @@ Template may look like `"{family}{Task}{Variant}"`.
...
+#### Custom Staging Dir Profiles
+With this feature, users can specify a transient data folder path based on presets, which can be used during the creation and publishing stages.
+
+![global_tools_custom_staging_dir](assets/global_tools_custom_staging_dir.png)
+
+Staging dirs are used as a destination for intermediate files (as renders) before they are renamed and copied to proper location during integration phase. They could be created completely dynamically in temp folders, or for some DCCs in `work` area.
+(Example could be Nuke where artist might want to temporarily render pictures into `work` area to check them before they get published.)
+One of the key advantages of this feature is that it allows users to repoint folder for such intermediate files to take advantage of faster storages for rendering, which can help improve workflow efficiency. Additionally, this feature allows users to keep their rendered data persistent, and use their own infrastructure for regular cleaning.
+
+In some cases, these DCCs automatically add a rendering path during the creation stage, which is then used in publishing. Creators of such DCCs need to use these profiles to use this functionality.
```
BigRoy, line 277: I'd argue this is maybe not entirely up to the creators. This might then also need validators, etc. I'd argue that the best case is being able to get the logic working in the Extractor (e.g. choosing the output path during extracting for e.g. a Houdini ROP node and then reverting it.)

website/docs/project_settings/settings_project_global.md
```diff
@@ -263,10 +263,25 @@ Template may look like `"{family}{Task}{Variant}"`.
...
-## Workfiles
+### Publish
+
+#### Custom Staging Dir Profiles
+With this feature, users can specify a transient data folder path based on presets, which can be used during the creation and publishing stages.
+
+![global_tools_custom_staging_dir](assets/global_tools_custom_staging_dir.png)
+
+Staging dirs are used as a destination for intermediate files (as renders) before they are renamed and copied to proper location during integration phase. They could be created completely dynamically in temp folders, or for some DCCs in `work` area.
+(Example could be Nuke where artist might want to temporarily render pictures into `work` area to check them before they get published.)
```
BigRoy, line 274: I don't see a reason why this needs to be between brackets in the documentation. It's an example use case which is what the documentation should explain if it's unclear without.

Additionally I have an issue with this example - because how does it allow them to "check the render" before they get published? The publisher still continues by default to actually publish and integrate the files, no? Unless maybe this is rendered on a farm and the submit publish job is suspended by default. To me this explanation in the docs hints a bit too much that it'll give me control to "check" before it integrates into the database which will not be the case for a lot of the publishes.

openpype/pipeline/publish/lib.py
kalisp: If the question is that  `filter_profiles` should filter (find best applicable) only in 'active' profiles is interesting question, but definitely for different PR/separate issue.

But I could totally see both use cases would be aplicable:
- I have general profile for `Nuke` - active
- I have more detailed profile for 'renders' in `Nuke`, I am publishing `render`, but custom dir doesnt' work, so I want to deactivate it for now to get `render` out, because chaos, deadline looming etc.etc.
- do I want to fall back to higher profile (I am still publishing from Nuke) and still run it or am I saying "for this context/profile doesn't use this" (because I think we are using latter use case in Settings to force some behavior on more descriptive profile - how would I really deactive it just for `render` family?)

website/docs/project_settings/settings_project_global.md
kalisp: Yeah, I am maybe using too much parentheses, even my czech language teacher was telling me that. That's Terry Pratchett fault ;)

openpype/pipeline/publish/lib.py
BigRoy: > If the question is that filter_profiles should filter (find best applicable) only in 'active' profiles is interesting question,

Not necessarily - it was in particular about this case but I didn't know there were more settings for profiles that had an "active" toggle. 

> 👉  **do I want to fall back to higher profile (I am still publishing from Nuke) and still run it** 👈  ~or am I saying "for this context/profile doesn't use this" (because I think we are using latter use case in Settings to force some behavior on more descriptive profile - how would I really deactive it just for render family?)~

That's what I would have expected to happen in this case based on the labels in settings, yes. I can understand your use case but that definitely isn't what I thought would happen.

It's more like an inclusive/exclusive filtering. Is active felt to me like it would just do nothing and not influence any of the other profiles defined.

website/docs/project_settings/settings_project_global.md
kalisp: Nuke can have that "Use existing frames" in write node, so I think Jezza was describing that as a use case, but maybe I misunderstood.

BigRoy: > Nuke can have that "Use existing frames" in write node, so I think Jezza was describing that as a use case, but maybe I misunderstood.

Ah, yes. That's an additional issue that could be problematic with this PR.

Are we sure this feature is worth the amount of 'likely won't work with X or Y' issues we're already assuming are there. 😄

openpype/pipeline/publish/lib.py
kalisp: Well, I had it before as `use_transient_dir` and you didn't fancy that either, so I think we are running out of options here :D.

I created issue for that here https://github.com/ynput/OpenPype/issues/4681 , please feel free to comment there behavior you would expect.

website/docs/project_settings/settings_project_global.md
kalisp: I think it is customer request and they will probably use it in pretty limited way, here we are trying to solve this for all the imaginary others and all the possible use cases, but deciding what feature and what uses cases it should cover is beyond my pay grade, I am afraid ;).

kalisp:
Updated documentation a bit.

jakubjezek001 (approved):
Tested and works as expected. Shouldn't we create also following dependent issues? Talking about DCC integrations.

kalisp:
Issues are already created in Clickup (with old name, will rename it):
2023-03-24 16_30_14-Staging directory for renders _ OP-4258

PR #4523, opened 2023-02-23
Fusion: New Publisher

EmberLightVFX:
## Brief description
This is an updated PR for @BigRoy 's old PR (https://github.com/ynput/OpenPype/pull/3892).
I have merged it with code from OP 3.15.1-nightly.6 and made sure it works as expected.

This converts the old publishing system to the new one. It implements Fusion as a new host addon.

## Description
- Create button removed in OpenPype menu in favor of the new Publisher
- Draft refactor validations to raise PublishValidationError
- Implement Creator for New Publisher
- Implement Fusion as Host addon

## Additional info
I have close to no knowledge of the new Publisher so there's probably a lot broken bits and pieces. Admittedly the API for Creator's and the UX for the new publisher can still use some work before doing this full code refactor and switching to the new publisher.

PR #4584, opened 2023-03-07
AttrDefs: Fix type for PySide6

iLLiCiTiT:
## Brief description
Use right type in signal emit for value change of attribute definitions.

## Description
Changed `UUID` type to `str`. This is not an issue with PySide2 but it is with PySide6.

## Testing notes:
1. Run some UI using AttrDefs (e.g. Publisher) in process using PySide6
2. Change value of the attribute
3. The change should not cause crash


BigRoy (approved):
This works me with the [issue I had in Fusion mentioned here](https://github.com/ynput/OpenPype/pull/4523#issuecomment-1458195711). I haven't tested it in PySide2 hosts however.

PR #4588, opened 2023-03-07
Houdini: simple bgeo publishing

### LINKED ISSUE #3098: Houdini bgeo publishing

antirotor:
## Brief description
Support for simple publishing of bgeo files.

## Description
This is adding basic support for bgeo publishing in Houdini. It will allow publishing bgeo in all supported formats (selectable in the creator options).  If selected node has `output` on sop level, it will be used automatically as path in file node.

## Additional info
For future work it might be good to rethink how "pointcache" family is implemented regarding to various format and validation needs. Maybe default option for file format can be added to settings, some additional validation, and so on.

## Testing notes:
1. Create some geometry
2. Select it
3. Create BGEO publishing instance
4. Publish

Resolves #3098


BigRoy (commented):
Some early comments

openpype/hosts/houdini/plugins/publish/extract_bgeo.py
```diff
@@ -0,0 +1,50 @@
...
+        instance.data["stagingDir"] = staging_dir
+
+        file_name = os.path.basename(output)
+
+        # We run the render
+        self.log.info("Writing bgeo files '%s' to '%s'" % (file_name,
+                                                        staging_dir))
+
+        # write files
+        ropnode.parm("execute").pressButton()
```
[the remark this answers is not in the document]
antirotor: Actually there was a reason as I've initially used `render_rop()`. And it didn't work for me, complaining that there is no `render()` method on the node. But since I've restarted everything it works again. Some dark Houdini magic there.

openpype/hosts/houdini/plugins/publish/extract_bgeo.py
```diff
@@ -0,0 +1,49 @@
...
+        # write files
+        render_rop(ropnode)
+
+        output = instance.data["frames"]
+
+        if "representations" not in instance.data:
+            instance.data["representations"] = []
+
+        representation = {
+            'name': 'bgeo',
```
fabiaserra, line 42: mixed single quotes for strings

openpype/hosts/houdini/plugins/create/create_bgeo.py
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        if self.selected_nodes:
+            parms["soppath"] = self.selected_nodes[0].path()
+
+            # try to find output node
+            for child in self.selected_nodes[0].children():
+                if child.type().name() == "output":
+                    parms["soppath"] = child.path()
+                    break
```
BigRoy, line 47: I'm not so sure about this workflow - it seems a bit too magical almost to only do it if the node itself happens to contain an output node. Is this to avoid the issue that an Object level path gets set instead of SOP path? If so, shouldn't we only ever do this if the node is an object level object.

Also, output nodes tend to have an Output Index set in Houdini, right? Shouldn't we be taking the one with the lowest index?

What if I happen to have selected a SOP node that is an HDA/Subnet which itself contains an Output node - I wouldn't want that output node to be set as path but the actual SOP path I had selected.

antirotor: I was thinking about it a lot. I've decided to go with this rather simple way to help set it up because I think there should be something working out-of-the-box and at the same time it shouldn't limit more advanced uses. There is probably no definitive way to handle all cases that can happen so it should either stop assuming anything at all, or implement very robust way of handling it even though there is a good chance it won't be ever used, or use something not perfect but simple to make you go. Honestly, I have no clear opinion on that especially in Houdini.

Good point with the index. It should probably behave like - use selected if it is SOP node, if not, try to find output with the lowest index available.

openpype/hosts/houdini/plugins/create/create_bgeo.py
```diff
@@ -0,0 +1,83 @@
...
+from openpype.hosts.houdini.api import plugin
+from openpype.pipeline import CreatedInstance
+from openpype.lib import EnumDef
+
+
+class CreateBGEO(plugin.HoudiniCreator):
+    """BGEO pointcache creator."""
+    identifier = "io.openpype.creators.houdini.bgeo"
+    label = "BGEO PointCache"
+    family = "bgeo"
```
fabiaserra, line 12: Out of scope for this PR but I was testing the Houdini publisher when I got confused (again, because I remember coming across this same confusion a couple weeks back) trying to find an Alembic cache family and realizing that it was the `Point Cache`. However, that got me thinking and made me remember this PR where you were creating a new family and I'm noticing there's quite semantic divergence on how you have been creating the more recent families and I don't understand why we are not making use of `representations` instead of creating new families for each file format.
Wouldn't it make more sense to have a single `Point Cache` family that allows us to define which representations we want for it (i.e., `Abc`, `Bgeo`, `ASS`, `USD`...) so it's up to the project settings to choose what data formats we want our families to be on? Same applies to `Model`, `Camera`... In my prior studio we published the different file formats as `components` on the same ftrack asset version, so sometimes you'd prefer to import the bgeo representation but some others the alembic cache (same for textures to store .exr and .tx or shaders for .mtlx, .ass...). Of course there will be occasions where there might be some slight differences of what's stored on the representation due to the missmatch of features but I think there's some big limitations on the alternative. I wouldn't want the artists to select every family we choose to support on the project, I would rather them just needing to know they have to publish a "Cache" and in our settings I could choose to not only create a `Bgeo` cache but perhaps a `USD` one as well, that way I could slowly migrate to a different format. Does this make any sense or I need some extra coffee?

BigRoy: > Wouldn't it make more sense to have a single Point Cache family that allows us to define which representations we want for it (i.e., Abc, Bgeo, ASS, USD...)

It's not **that** trivial due to them actually being separate ROP nodes (~and also requiring different validators to trigger in production~). So the complexity of maintaining that increases and the flexibility of the individual nodes somewhat go down I'd say. It's not impossible however but this would definitely require a custom HDA to make it maintainable.

I'm not entirely sure however about writing say `bgeo` + `alembic` + `usd` family into a single subset. Again, it's possible - just add the different families into a single instance and it does it. But does OpenPype / Ayon allow changing the family of an existing subset midway, like remove an `alembic` family and add a `bgeo` if the next version doesn't have it. Sounds dangerous if we start wiggling that around and someone starts ticking off and on a specific "output family" for the instance when publishing new versions?

Older iterations of OpenPype / Avalon had `families` on versions so they were per version, but at some point they were moved to Subsets to make them more static with a reason - to avoid the expected data to be versioned to to suddenly be something different. E.g. latest version 8 not having a USD file, so to what version do you update then when updating the loaded USD? Do you report an error? Take the last existing? etc. Complexity does quickly increase.

Another side note - for e.g. USD you tend to set a different "path" LOPs path instead of SOP path for Alembic, etc. So even setting what you're outputting could differ?

fabiaserra: Thanks @BigRoy for the quick response and explanation!! I definitely agree that the complexity would increase but knowing how many times on my prior studio we were able to fix production needs by having the flexibility of easily swapping representations/components of any existing asset version makes me worry whether that's something that could become a blocker with the current design.

Are the key concepts definitions from the OP docs outdated? I'm a bit confused when you mention "`bgeo` + `alembic` + `usd` **family** into a single subset". Wouldn't that be "`bgeo` + `alembic` + `usd` **representations** into a single subset"? Reading that glossary again (https://openpype.io/docs/artist_concepts#representation) this seems to describe exactly how I had envisioned the design to be, you choose the family, the subset and then the representations of that subset (family > subset > representation). But from what I'm seeing, "representations" aren't really used here, mostly just for storing a thumbnail of the same subset?

I think that creating custom HDAs might be a requirement to scale the OP integration and make it more Houdini native in the long run (allowing artists to create node networks directly and setting any dependencies for doing publishes instead of needing to go through the UIs, although that would still be possible). On my prior studio we had a system that hijacked the `hou.Node` python class by inserting it as an inherit into our custom python classes so you could write any python code and you could call it for any HDA or Houdini node directly from the node's python object (i.e., `hou.node("custom_node").custom_function()`, `hou.node("alembic_node").custom_function()`), not needing to write the code within the HDA. We also had custom nodes with very simple purposes (i.e., "OP Publisher", "OP Cache") so for TDs (and even the developers to provide new publishers) it made it very easy to build more complex functionality by combining the smaller nodes (think of Bifrost Compounds or Nuke gizmos). 

As for your worry about "representations" missing across different versions, the way we did it was very simple, if you try to update to a newer version that doesn't have the representation you are using, it just errors out saying that the "representation" doesn't exist. However, that was very uncommon as most artists didn't mess with the default job settings of which "representation/component" should be published by family. On the other hand, imagine the alternative situation where I might want to have "Bgeo" and "Alembic" caches throughout the entire production. How do you currently keep those separate families "on sync" with this design? If they don't **always** publish both, the versions get out of sync and all of a sudden it makes it very hard to know which family "represents" the other family. I think that's where "representations" should really be used for?

Anyway, I'm still very new to OpenPype and I still need to fight a lot more with the Houdini OP implementation to get all of these concepts wrapped in my brain, I might change my mind again then haha

Definitely agree to start a GH issue/discussion to discuss this further.

moonyuet:
Tested in my machine with some of my old project. 
It's kinda confusing me when I tested it for the first time but it works for exporting simple bgeo.
But it would be great if there is a validator to check to see if the user puts the SOP path into the bgeo instance.
If the user didn't put the SOP path, it won't export any bgeo path to the pyblish directory (and it can't pass the integrate.py)

image

website/docs/artist_hosts_aftereffects.md
```diff
@@ -15,18 +15,18 @@ sidebar_label: AfterEffects
 
 ## Setup
 
-To install the extension, download, install [Anastasyi's Extension Manager](https://install.anastasiy.com/). Open Anastasyi's Extension Manager and select AfterEffects in menu. Then go to `{path to pype}hosts/aftereffects/api/extension.zxp`. 
+To install the extension, download, install [Anastasyi's Extension Manager](https://install.anastasiy.com/). Open Anastasyi's Extension Manager and select AfterEffects in menu. Then go to `{path to pype}hosts/aftereffects/api/extension.zxp`.
```
moonyuet, line 18: I dont think this documentation is related to this PR.

antirotor:
> @antirotor Are you expected to do additional changes based on the open comments on this PR or does this just need some testing before it's ready to merge?

Yep, I would like to add the validator and logic around the output node index.

moonyuet (changes requested):
I tested the bgeo publisher with some of my old scene(which is quite huge), with two scanerios(with geometry node and sopoutput node) after studying the code. The publisher doesn't error out and bgeo is published successfully. The only thing is that it gets the error for increment and save since the geometry creation includes a merge node before the sopoutput as the node cook is in progress. We may need to introduce node cook (i.e. `node.cook(force=True)`) in _increment_current_file.py_ before performing save function.

image
image

antirotor:
> ... We may need to introduce node cook (i.e. `node.cook(force=True)`) in _increment_current_file.py_ before performing save function.

I see - but I think this should be handled by separate PR as this is not directly related to bgeo, but probably more Houdini workflows. What do you think?

moonyuet:
> > ... We may need to introduce node cook (i.e. `node.cook(force=True)`) in _increment_current_file.py_ before performing save function.
> 
> I see - but I think this should be handled by separate PR as this is not directly related to bgeo, but probably more Houdini workflows. What do you think?

if it's just for very simple file cache without merge node or any nodes needed to be cooked before save, it should be okay. But we need to see why the user wants to export the bgeo and load it back to file cache. File cache shortens the time of loading the geometry, which means the geometry usually consists of nodes taking time to process the data. 
For my personal user experience of Houdini, the data I cached usually has merge node. I used mutliple boolean nodes or self-created generators todo the fracture of the helipod. I cached the data and loads it back for the preparation of the stimulation(rigid body destruction). 
image

BigRoy:
> since the geometry creation includes a merge node before the sopoutput as the node cook is in progress. We may need to introduce node cook (i.e. node.cook(force=True)) in increment_current_file.py before performing save function.

Sorry, could explain more elaborately what happens? I don't understand why a scene save would break things there - and why is the problem with a merge node?

> For my personal user experience of Houdini, the data I cached usually has merge node. I used mutliple boolean nodes or self-created generators todo the fracture of the helipod. I cached the data and loads it back for the preparation of the stimulation(rigid body destruction).

That sounds totally fine - but I don't understand WHY it'd fail. Why do we need to cook it explicitly? When does it start cooking? It kind of sounds like something is triggering a 'cook' in houdini during publishing that we don't want to happen?

Maybe a screen recording could explain it better.

moonyuet:
> > since the geometry creation includes a merge node before the sopoutput as the node cook is in progress. We may need to introduce node cook (i.e. node.cook(force=True)) in increment_current_file.py before performing save function.
> 
> Sorry, could explain more elaborately what happens? I don't understand why a scene save would break things there - and why is the problem with a merge node?
> 
> > For my personal user experience of Houdini, the data I cached usually has merge node. I used mutliple boolean nodes or self-created generators todo the fracture of the helipod. I cached the data and loads it back for the preparation of the stimulation(rigid body destruction).
> 
> That sounds totally fine - but I don't understand WHY it'd fail. Why do we need to cook it explicitly? When does it start cooking? It kind of sounds like something is triggering a 'cook' in houdini during publishing that we don't want to happen?
> 
> It sounds more like we'd need to cancel any active cooking if it were cooking before saving - but I don't understand entirely why it should be cooking at that moment in time to begin with.
> 
> Maybe a screen recording could explain it better.

I restarted the computer and tried again i think it works now. maybe because of the old scene files. 
image

moonyuet:
> > since the geometry creation includes a merge node before the sopoutput as the node cook is in progress. We may need to introduce node cook (i.e. node.cook(force=True)) in increment_current_file.py before performing save function.
> 
> Sorry, could explain more elaborately what happens? I don't understand why a scene save would break things there - and why is the problem with a merge node?
> 
> > For my personal user experience of Houdini, the data I cached usually has merge node. I used mutliple boolean nodes or self-created generators todo the fracture of the helipod. I cached the data and loads it back for the preparation of the stimulation(rigid body destruction).
> 
> That sounds totally fine - but I don't understand WHY it'd fail. Why do we need to cook it explicitly? When does it start cooking? It kind of sounds like something is triggering a 'cook' in houdini during publishing that we don't want to happen?
> 
> It sounds more like we'd need to cancel any active cooking if it were cooking before saving - but I don't understand entirely why it should be cooking at that moment in time to begin with.
> 
> Maybe a screen recording could explain it better.

it can be because the first scene I used for test consists of tons of cooking processes It takes time to process all the geometry data. And when the extractor is doing export way faster than the cooking process, it may error out. 
Maybe I can send you guys the scene to investigate if you guys want to.

BigRoy:
> it can be because the first scene I used for test consists of tons of cooking processes It takes time to process all the geometry data. And when the extractor is doing export way faster than the cooking process, it may error out.
> Maybe I can send you guys the scene to investigate if you guys want to.

Yes, let's definitely get to the  bottom of this. Extracting could be long so we don't want it failing afterwards for users. Especially with an unclear error message where they don't know what succeeded or failed nor know what they could've done differently to resolve this.

I'd almost argue, let's not merge this until we solve it if it's reproducible with some scenes.

moonyuet:
> > it can be because the first scene I used for test consists of tons of cooking processes It takes time to process all the geometry data. And when the extractor is doing export way faster than the cooking process, it may error out.
> > Maybe I can send you guys the scene to investigate if you guys want to.
> 
> Yes, let's definitely get to the bottom of this. Extracting could be long so we don't want it failing afterwards for users. Especially with an unclear error message where they don't know what succeeded or failed nor know what they could've done differently to resolve this.
> 
> I'd almost argue, let's not merge this until we solve it if it's reproducible with some scenes. It would be good however to test whether that scene would also fail with e.g. publishing a vdb cache, or something else that goes through that same processing chain to learn whether the error is new to this PR or not.

Actually the scene I tested with has vdb creation(the mesh is a mix creation by vdb, scatter and copy). I tested the vdb cache publisher with the scene, it is successfully published without any error.
image
I loaded file cache with the published vdb cache and use bgeo publisher again for exporting bgeo, it still errors out.
image
I also did some research on this process. 
https://www.sidefx.com/docs/houdini/basics/cooking.html
When I set the cooking status to manual, it is successfully published without error. I resume my previous statement on adding `node.cook(force=True)` into the increment files. Maybe we need to include some checks on the cooking, and pehaps it needed to be manual instead of auto update before publishing bgeo. 
image

moonyuet:
One more thing to add, if the cooking process changes to Manual, the user needs to make sure loading all the nodes before publishing. 
image

fabiaserra:
> One more thing to add, if the cooking process changes to Manual, the user needs to make sure loading all the nodes before publishing. image

The ROP node that exports the SOP should already handle that?

moonyuet:
> > One more thing to add, if the cooking process changes to Manual, the user needs to make sure loading all the nodes before publishing. image
> 
> The ROP node that exports the SOP should already handle that?

the cooking process happens in not just only SOP but different nodes such as TOP and DOP, which means you will see that cooking process happens if you dont cache.
image
At the left hand bottom corner shows the cooking op process while the right shows how the cooking process works. If it is auto update, it always recook and update the view continuously as you change values and move sliders. When the bgeo publisher extracting the data, we are always doing frame range, which means moving sliders. so if we set the cooking process to auto update, it will always recook and update the view. This is okay for the data which is not big. But if we have really huge data to cache and the cooking process set to auto update, the exporting process may happen to be faster than the cooking process, the publisher possibly error out.

fabiaserra:
> > > One more thing to add, if the cooking process changes to Manual, the user needs to make sure loading all the nodes before publishing. image
> > 
> > 
> > The ROP node that exports the SOP should already handle that?
> 
> the cooking process happens in not just only SOP but different nodes such as TOP and DOP, which means you will see that cooking process happens if you dont cache. image At the left hand bottom corner shows the cooking op process while the right shows how the cooking process works. If it is auto update, it always recook and update the view continuously as you change values and move sliders. When the bgeo publisher extracting the data, we are always doing frame range, which means moving sliders. so if we set the cooking process to auto update, it will always recook and update the view. This is okay for the data which is not big. But if we have really huge data to cache and the cooking process set to auto update, the exporting process may happen to be faster than the cooking process, the publisher possibly error out(see my screenshot shown in the requested change's comment)

I'm aware of how cooking works in Houdini and no, it's not possible that the cooking and the caching in Houdini go separate, otherwise artists would constantly get inconsistent caching results based on what they have let cook. Houdini caching (or rendering or whatever process that requires a cook) cooks all the dependencies when it needs to output. I have created a GIF showing you how even when cooking is set to Manual, caching evaluates all the frames as expected:
houdinifx_jxfILYKE0U

So we shouldn't be needing to force cooks if we are using the vanilla Houdini evaluation on their nodes, only when you need to pull information that doesn't automatically cook

moonyuet:
> I'm aware of how cooking works in Houdini and no, it's not possible that the cooking and the caching in Houdini go separate, otherwise artists would constantly get inconsistent caching results based on what they have let cook. Houdini caching (or rendering or whatever process that requires a cook) cooks all the dependencies when it needs to output. 

I see your point and yeah, we need to be aware of the inconsistency of the workflow if the cooking process set it to manual. (We should take care of that as an artist's prespactive.) Maybe we can mention in the docs that if we have relatively huge data of bgeo cache, it is possible to crash the extractor and it is recommended that the artist set the cooking process to manual and make sure they force update the nodes before the extraction. 

> So we shouldn't be needing to force cooks if we are using the vanilla Houdini evaluation on their nodes, only when you need to pull information that doesn't automatically cook

that makes sense.

BigRoy:
> **When the bgeo publisher extracting the data, we are always doing frame range, which means moving sliders. so if we set the cooking process to auto update, it will always recook and update the view.** This is okay for the data which is not big. But if we have really huge data to cache and the cooking process set to auto update, the exporting process may happen to be faster than the cooking process, the publisher possibly error out(see my screenshot shown in the requested change's comment)

Isn't this the case for all publishes really? I wonder why we haven't seen this issue before with any of the other families. I'd love to have a simple reproducable scene. 

Even if it's just a cube of 1000x1000x1000 with a very slow cooking. Or maybe a time dependent Python node in SOPs that just does ```import time; time.sleep(2)``` so that it's just very slow.

Would to be able to reproduce this with a simple case so we can debug what's going on.

moonyuet:
> > **When the bgeo publisher extracting the data, we are always doing frame range, which means moving sliders. so if we set the cooking process to auto update, it will always recook and update the view.** This is okay for the data which is not big. But if we have really huge data to cache and the cooking process set to auto update, the exporting process may happen to be faster than the cooking process, the publisher possibly error out(see my screenshot shown in the requested change's comment)
> 
> Isn't this the case for all publishes really? I wonder why we haven't seen this issue before with any of the other families. I'd love to have a simple reproducable scene.
> 
> Even if it's just a cube of 1000x1000x1000 with a very slow cooking. Or maybe a time dependent Python node in SOPs that just does `import time; time.sleep(2)` so that it's just very slow.
> 
> Would to be able to reproduce this with a simple case so we can debug what's going on.

I have tested with a simple scene with cube of 1000x1000x1000 with a slow cooking process. It is published without any error. 
By this case, it can assume that the bgeo publisher works for the normal scene.
image

I also find out that the scene which errors the publisher out gets around 1x polygon than the simple cube scene.
image

It is more like the scene needs optimization.
image

BigRoy:
It's hard to tell. I wouldn't say creating very high res geometry breaking publishing is "ok" in production, but if it currently appears to be an edge case it's likely best to consider this PR ready for merge and create a separate issue (maybe with the problematic file attached?) so we can debug later and track the status of it.

openpype/hosts/houdini/plugins/publish/validate_bgeo_file_sop_path.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def process(self, instance):
+
+        import hou
+
+        node = hou.node(instance.data.get("instance_node"))
+        sop_path = node.evalParm("soppath")
+        if not sop_path:
+            raise PublishValidationError("Empty SOP Path found in "
+                                         "the bgeo isntance Geometry")
```
BigRoy, line 22:
```suggestion
            raise PublishValidationError("Empty SOP Path found in "
                                         "the bgeo instance Geometry")
```

Admittedly - the error message does look a bit vague to me. It should actually explain now where is the issue exactly and what could I do about it. We need to guide the artist in the right direction.

@antirotor could you improve both the validation messages?

openpype/hosts/houdini/plugins/publish/extract_bgeo.py
```diff
@@ -0,0 +1,49 @@
...
+        render_rop(ropnode)
+
+        output = instance.data["frames"]
+
+        if "representations" not in instance.data:
+            instance.data["representations"] = []
+
+        representation = {
+            "name": "bgeo",
+            "ext": instance.data["bgeo_type"],
```
BigRoy, line 43: What happens if the user had changed the filepath (e.g. extension) on the ROP node manually after creation. Then this would basically be invalid, right?

BigRoy (dismissed):
Validation messages can be improved.

openpype/hosts/houdini/plugins/publish/validate_bgeo_file_sop_path.py
mkolar: Agreed that this doesn't help the artist at all.

moonyuet:
I tested with the new codes, both bgeo publisher and loaders work. And the new one doesn't undergo caching which errors out the publisher due to the large scale caching.
image

image

I also did very simple test on point cache publishing, it works.
image

moonyuet (approved):

openpype/hosts/houdini/plugins/create/create_bgeo.py
```diff
@@ -0,0 +1,90 @@
...
+from openpype.hosts.houdini.api import plugin
+from openpype.pipeline import CreatedInstance
+from openpype.lib import EnumDef
+
+
+class CreateBGEO(plugin.HoudiniCreator):
+    """BGEO pointcache creator."""
+    identifier = "io.openpype.creators.houdini.bgeo"
+    label = "BGEO PointCache"
+    family = "pointcache"
```
BigRoy, line 12: With us now having two creators with the same family. What does that mean for the "Convert Legacy" logic for old scenes made prior to the new publisher. Will it correctly convert to the Alembic one instead of the BGEO one?

Please make sure this still updates legacy scenes fine.

openpype/plugins/publish/integrate.py
```diff
@@ -137,7 +137,8 @@ class IntegrateAsset(pyblish.api.InstancePlugin):
                 "mvUsdOverride",
                 "simpleUnrealTexture",
                 "online",
-                "uasset"
+                "uasset",
+                "bgeo"
```
BigRoy, line 141: Technically `bgeo` could be removed again right, since we now still rely on the `pointcache` family existing? If we expect `bgeo` to remain here, shouldn't we now also add the new `abc` family?

openpype/hosts/houdini/plugins/create/create_pointcache.py
```diff
@@ -19,6 +19,10 @@ def create(self, subset_name, instance_data, pre_create_data):
         instance_data.pop("active", None)
         instance_data.update({"node_type": "alembic"})
 
+        if not instance_data.get("families"):
+            instance_data["families"] = []
```
BigRoy, line 23: When would this occur? If this happens, doesn't that also mean it doesn't have `pointcache` in the families and thus we've a broken instance, or?

openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
```diff
@@ -21,15 +21,11 @@ def process(self, instance):
...
-                self.log.debug("content members: {}".format(children))
-            elif objset.endswith("proxy_SET"):
-                set_members = get_all_children(cmds.ls(members, long=True))
-                instance.data["proxy"] = set_members
-                self.log.debug("proxy members: {}".format(set_members))
+
+            instance.data["contentMembers"] = self.get_ass_data(
+                objset, members, "content_SET")
+            instance.data["proxy"] = self.get_ass_data(
+                objset, members, "proxy_SET")
```
BigRoy, line 28:
```suggestion
            if objset.endswith("content_SET"):
                instance.data["contentMembers"] = self.get_hierarchy(members)
            if objset.endswith("content_SET"):
                instance.data["proxy"] = self.get_hierarchy(members)
```

openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
```diff
@@ -48,3 +44,13 @@ def process(self, instance):
...
+
+    def get_ass_data(self, objset, members, suffix):
+        if objset.endswith(suffix):
+            members = cmds.ls(members, long=True)
+            if not members:
+                return
+            children = get_all_children(members)
+            members = list(set(members) - set(children))
+
+            return children + members
```
BigRoy, line 56: `get_ass_data` does not describe to me **what** it does.

Also, your logic above is invalid, it will now set `None` to one of the two entries since the object set will not match the suffix in that iteration, etc.

```suggestion
    def get_hierarchy(self, nodes):
        """Return nodes with all their children"""
        nodes = cmds.ls(nodes, long=True)
        if not nodes:
            return []
        children = get_all_children(nodes)

        # Make sure nodes merged with children only
        # contains unique entries
        return list(set(nodes) - set(children))

```

openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
moonyuet: I think this one is not related. I did a wrong branch unfortunately

openpype/hosts/houdini/plugins/create/create_pointcache.py
antirotor: I am paranoid. Couldn't find any point in code that is actually forcing "families" as list on `instance_data`

openpype/plugins/publish/integrate.py
antirotor: good catch! 👍🏻

openpype/hosts/houdini/plugins/create/create_pointcache.py
BigRoy: I see, but _if_ it doesn't enforce it - shouldn't it _also_ contain `pointcache` (or add it?)

openpype/hosts/houdini/plugins/create/create_bgeo.py
BigRoy: Was this checked? Since it's marked resolved.

openpype/hosts/houdini/plugins/create/create_pointcache.py
antirotor: Well since it is already in `family`, it doesn't have to be in `families`.  With this, it won't crash for artist because it doesn't have to. Of course, question is what is also missing if `families` is not there. I could just do:

```python
instance_data["families"] = ["abc"]
```
but what if logic calling `create()` will change and will start to add something there (not sure why though :)) So this was reasonably safe approach to it - append.

antirotor:
@BigRoy I think this is resolved, just want to check with you

BigRoy:
> I think this is resolved, just want to check with you

I still have some issues with adding the extra family on Create into `families` data. It's just so non-dynamic even though depending on the creator we **always** want the extra family so it seems. There are just so many edge cases that I can think of where this logic could break. The most apparent one currently is that I suspect that any **non-legacy** pointcache instances already created with the new publisher in Houdini prior this PR will now break due to the fact that those do not have the additional family stored on the node's data. To me this just doesn't seem "instance data" related since it's not a toggle within the creator but the creator itself should **always** contain that family. As such there's also no need to store that data on the actual node at all (I'd even say it's cluttered data there + a user could potentially break it by just removing the value).

A collector doing it would already be safer plus doesn't require us storing this basically non-dynamic/changeable data on the node (which makes no sense because it's just an implementation detail). A simple example collector could do:
```python
if creator_identifier == "io.openpype.creators.houdini.bgeo":
    families += ["bgeo"]
elif creator_identifier == "io.openpype.creators.houdini.pointcache":
    families += ["abc"]
```
Not perfect either because then the new publisher can't specify user parameters based on that collector because there's unfortunately no notion of **PreCollector** plug-ins which run before the user interaction but after the creator.

Anyway - let's at the very least fix up the (what I suspect will be an) issue with existing pointcache instances in the new publisher and I'd say this is an "ok" implementation.

openpype/hosts/houdini/plugins/create/create_bgeo.py
```diff
@@ -0,0 +1,90 @@
...
+    icon = "gears"
+
+    def create(self, subset_name, instance_data, pre_create_data):
+        import hou
+
+        instance_data.pop("active", None)
+
+        instance_data.update({"node_type": "geometry"})
+
+        if not instance_data.get("families"):
```
Minkiu, line 22: This could be done with:
```py
instance_data.setdefault("families", [])
```

openpype/hosts/houdini/plugins/publish/extract_bgeo.py
```diff
@@ -0,0 +1,55 @@
...
+    hosts = ["houdini"]
+    families = ["bgeo"]
+
+    def process(self, instance):
+
+        ropnode = hou.node(instance.data["instance_node"])
+
+        # Get the filename from the filename parameter
+        output = ropnode.evalParm("sopoutput")
+        staging_dir = os.path.dirname(output)
```
Minkiu, line 25: Cosmetic, but we could do:
```py
staging_dir, file_name = os.path.split(output)
```

website/docs/artist_hosts_houdini.md
```diff
@@ -132,3 +132,26 @@ switch versions between different hda types.
...
 When you load hda, it will install its type in your hip file and add published version as its definition file. When
 you  switch version via Scene Manager, it will add its definition and set it as preferred.
+
+## Publishing and loading BGEO caches
+
+There is a simple support for publishing and loading **BGEO** files in all supported compression variants.
+
+### Creating BGEO instances
+
+Just select your SOP node to be exported as BGEO. If your selection is object level, OpenPype will try to find there is `output` node inside, the one with the lowest index will be used:
```
Minkiu, line 142: I'd skip the "Just", and I think we are missing words here?
 `try to find **if** there is **an**`

website/docs/artist_hosts_houdini.md
```diff
@@ -132,3 +132,26 @@ switch versions between different hda types.
...
+
+![BGEO output node](assets/houdini_bgeo_output_node.png)
+
+Then you can open Publisher, in Create you select **BGEO PointCache**:
+
+![BGEO Publisher](assets/houdini_bgeo-publisher.png)
+
+You can select compression type and if the current selection should be connected to ROPs SOP path parameter. Publishing
```
Minkiu, line 150: Weird newline here?

Minkiu (approved):

openpype/hosts/houdini/plugins/create/create_bgeo.py
iLLiCiTiT: Why are the families filled in creation? Wouldn't be better to handle that in publishing?

BigRoy: > Wouldn't be better to handle that in publishing?

I agree that it would make much more sense - except that _then_ you can't have attribute definitions for the new publisher exposed to those families. It's an unfortunate downside to the design of the new publisher.

iLLiCiTiT: That would not be there even now.

BigRoy: It would be, right? Since it's added on create and thus is present in the instance directly after? Or doesn't it support `families` to begin with? :)

antirotor: I've moved the families to collector as there is no need to have it set by creator because there are no additional options set by plugins determined by those families.

It should be backwards compatible with instances created with previous creators.

I agree with storing of data on node, we should address it in another PR

antirotor:
> I still have some issues with adding the extra family on Create into families data. It's just so non-dynamic even though depending on the creator we always want the extra family so it seems. There are just so many edge cases that I can think of where this logic could break.

I've moved the logic to collector - it should handle legacy instances. The  problem with additional families, collection and plugin parameters is there and has no solution currently. In this particular case, it doesn't matter that much because it has no additional parameters defined down the line - but in theory, it should have - for example format specification for bgeo should be probably parameter of the bgeo extractor plugin. 

I don't mind having those families set by creator - you can argue that it is static, but moving it to collector (in this particular case) just move the static stuff down the line - testing `creator_identifier` and assigning family based on that is just the same piece of code just moved further. Yes, you can do lot of other things in collectors if you wish, but nothing is preventing doing so even when the `families` is already defined in the creator. Clear data driven workflow isn't there already because of the settings and various presumptions.

I agree totally on what we are storing on the instance itself and it should be as minimal as possible.

BigRoy:
> I tested the bgeo publisher with some of my old scene(which is quite huge), with two scanerios(with geometry node and sopoutput node) after studying the code. The publisher doesn't error out and bgeo is published successfully. The only thing is that it gets the error for increment and save since the geometry creation includes a merge node before the sopoutput as the node cook is in progress. We may need to introduce node cook (i.e. `node.cook(force=True)`) in _increment_current_file.py_ before performing save function.
> 
> image image

So we've just hit this error for the first time now - when publishing USD files on our end.
I wonder if anyone else has had this issue since and whether there are any suitable/save workarounds for the code?

@moonyuet have you had this since?

PR #4589, opened 2023-03-07
AYON: Integrate Hero version is more specific about data

iLLiCiTiT:
## Changelog Description
Hero versions integration is more specific about data integrated to ayon. The change was added to reduce time spent in requerying already known data. In AYON is hero version fully usable version without reference to source version entity (other than negative value of version).

## Additional info
During this process were discovered few bugs on server and python api. Some of bugfixes are in progress to be merged and released so it is possible that this PR won't work because of outdated versions (make sure you have run 'create_env' script).

## Testing notes:
1. Enabled Integrate Hero version on ayon server
2. Publish something (family must be in enabled families in settings)
3. Integration of hero version should be successful

### DIFF
--- a/openpype/client/server/conversion_utils.py
+++ b/openpype/client/server/conversion_utils.py
@@ -838,22 +838,46 @@ def convert_create_version_to_v4(version, con):
 
 
 def convert_create_hero_version_to_v4(hero_version, project_name, con):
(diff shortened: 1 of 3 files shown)


kalisp (approved):
Hero publishing works.

PR #4590, opened 2023-03-07
Fusion: Fix review burnin start and end frame

BigRoy:
## Changelog Description

Fix the burnin start and end frame for reviews. Without this the asset document's start and end handle would've been added to the _burnin_ frame range even though that would've been incorrect since the handles are based on the comp saver's render range instead.

## Additional info

[Mentioned on discord](https://discord.com/channels/517362899170230292/517382145552154634/1082747784492494958)

It does feel a bit odd that I'm required to ALSO set this explicitly like this because otherwise global plug-ins start messing with the data. There seems to be something fishy in logic elsewhere:

## Testing notes:
1. Create an asset with start + end handles of say 10
2. Create a comp with in/out points on 1001-1010. (including its global in/global out, so no handles)
3. Render with review burnin

Without this PR the burnin would say something like this screenshot:
afbeelding

After this PR it looks ok:
afbeelding

### DIFF
--- a/openpype/hosts/fusion/plugins/publish/collect_comp_frame_range.py
+++ b/openpype/hosts/fusion/plugins/publish/collect_comp_frame_range.py
@@ -39,3 +39,5 @@ class CollectFusionCompFrameRanges(pyblish.api.ContextPlugin):
         context.data["frameEnd"] = int(end)
         context.data["frameStartHandle"] = int(global_start)
         context.data["frameEndHandle"] = int(global_end)
+        context.data["handleStart"] = int(start) - int(global_start)
+        context.data["handleEnd"] = int(global_end) - int(end)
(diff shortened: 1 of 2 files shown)


jakubjezek001 (approved):
good catch! thanks a lot for the fix ;)

EmberLightVFX:
Got exactly this problem now so it became a perfect time to review it!
Solved the problem and everything looks good for me :)
It also seems to fix the bug where the footage gets places in the timeline in Fusion when importing it to the comp, so that's a great bonus!

Thanks a bunch for the fix!

EmberLightVFX (approved):
Got exactly this problem now so it became a perfect time to review it!
Solved the problem and everything looks good for me :)
It also seems to fix the bug where the footage gets places in the timeline in Fusion when importing it to the comp, so that's a great bonus!

Thanks a bunch for the fix!

PR #4594, opened 2023-03-08
Scene inventory: Fix code errors when "not found" entries are found

BigRoy:
## Changelog Description

Whenever a "NOT FOUND" entry is present a lot of errors happened in the Scene Inventory:

- It started spamming a lot of errors for the VersionDelegate since it had no numeric version (no version at all). 

[Error reported on Discord](https://discord.com/channels/517362899170230292/563751989075378201/1082727536523489341):
```python
  File "C:\Users\videopro\Documents\github\OpenPype\openpype\tools\utils\delegates.py", line 65, in paint
    text = self.displayText(
  File "C:\Users\videopro\Documents\github\OpenPype\openpype\tools\utils\delegates.py", line 33, in displayText
    assert isinstance(value, numbers.Integral), (
AssertionError: Version is not integer. "None" <class 'NoneType'>
```

- Right click menu would error on NOT FOUND entries, and thus not show. With this PR it will now _disregard_ not found items for "Set version" and "Remove" but still allow actions.

This PR resolves those.

## Additional info

I'd love for "remove" to still be possible for not found entries since the Loader logic for remove is just `Loader.remove(container)` and we do have a container so should be able to call that logic. However due to `Loader.__init__` currently requiring a full context that is not possible. 

[Also reported on Discord.](https://discord.com/channels/517362899170230292/563751989075378201/1082987059725221888)

## Testing notes:

1. Tweak the representation value on a container so it's broken
2. Use the scene inventory - it should now behave better.

### DIFF
--- a/openpype/tools/utils/delegates.py
+++ b/openpype/tools/utils/delegates.py
@@ -30,9 +30,11 @@ class VersionDelegate(QtWidgets.QStyledItemDelegate):
     def displayText(self, value, locale):
         if isinstance(value, HeroVersionType):
             return lib.format_version(value, True)
-        assert isinstance(value, numbers.Integral), (
-            "Version is not integer. \"{}\" {}".format(value, str(type(value)))
-        )
+        if not isinstance(value, numbers.Integral):
+            # For cases where no version is resolved like NOT FOUND cases
+            # where a representation might not exist in current database
+            return
+
         return lib.format_version(value)
 
     def paint(self, painter, option, index):
--- a/openpype/tools/sceneinventory/model.py
+++ b/openpype/tools/sceneinventory/model.py
@@ -327,7 +327,7 @@ class InventoryModel(TreeModel):
                 project_name, repre_id
             )
             if not representation:
-                not_found["representation"].append(group_items)
+                not_found["representation"].extend(group_items)
                 not_found_ids.append(repre_id)
                 continue
 
@@ -335,7 +335,7 @@ class InventoryModel(TreeModel):
                 project_name, representation["parent"]
             )
             if not version:
-                not_found["version"].append(group_items)
+                not_found["version"].extend(group_items)
                 not_found_ids.append(repre_id)
                 continue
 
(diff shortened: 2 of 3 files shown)


BigRoy:
> I'd love for "remove" to still be possible for not found entries since the Loader logic for remove is just Loader.remove(container) and we do have a container so should be able to call that logic. However due to Loader.__init__ currently requiring a full context that is not possible.

This would only be possible after #4602 - and it might be best kept for a separate PR afterwards to re-enable the "remove" menu entry for _"not found"_ containers

openpype/tools/utils/delegates.py
```diff
@@ -30,9 +30,11 @@ def __init__(self, dbcon, *args, **kwargs):
     def displayText(self, value, locale):
         if isinstance(value, HeroVersionType):
             return lib.format_version(value, True)
-        assert isinstance(value, numbers.Integral), (
-            "Version is not integer. \"{}\" {}".format(value, str(type(value)))
-        )
+        if not isinstance(value, numbers.Integral):
+            # For cases where no version is resolved like NOT FOUND cases
+            # where a representation might not exist in current database
+            return
```
iLLiCiTiT, line 36: Maybe we should display something?
```suggestion
            return "N/A"
```

BigRoy: I'll give it a go next week and take a screenshot of both empty column and N/A in column to see what looks best.

Side note: To me _not applicable_ doesn't sound entirely right because it's more a case of "was unable to resolve but it should have been applicable". :) It would be more like "unknown", "unidentified", "undetermined".

iLLiCiTiT: > To me not applicable doesn't sound entirely right because it's more a case of "was unable to resolve but it should have been applicable". :) It would be more like "unknown", "unidentified", "undetermined".

Too long for version column.

BigRoy: > Too long for version column.

Agreed. Was just trying to find the right terminology

---

Here's a visual comparsion.

**Empty version column**

afbeelding

**Version column with N/A**

afbeelding

---

I personally prefer the empty/cleaner one without N/A but would love to hear more opinions.

iLLiCiTiT: @mkolar @antirotor opinions? It can be easily changed when needed in future.

iLLiCiTiT (approved):
Looks good. Just one small discussion

openpype/tools/utils/delegates.py
antirotor: I think the empty is better because it is obvious from the first glance. Maybe even hyphen `-` could be used because it is visually different than any letters/numbers, but empty works for me.

antirotor (approved):

BigRoy:
Feel free to merge! :)

PR #4595, opened 2023-03-09
Fix: Loader doesn't filter asset list recursively

Tilix4:
## Changelog Description
Since OP 3.15.0 Loader doesn't filter asset list recursively

## Testing notes:
1. Open Loader
2. Try to filter by typing words

### DIFF
--- a/openpype/tools/utils/models.py
+++ b/openpype/tools/utils/models.py
@@ -240,7 +240,6 @@ class RecursiveSortFilterProxyModel(QtCore.QSortFilterProxyModel):
         )
 
 
-
 class ProjectModel(QtGui.QStandardItemModel):
     def __init__(
         self, dbcon=None, only_active=True, add_default_project=False,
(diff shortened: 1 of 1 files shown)


openpype/tools/utils/models.py
```diff
@@ -235,6 +240,7 @@ def filterAcceptsRow(self, row, parent_index):
         )
 
 
+
```
BigRoy, line 243:
```suggestion
```

iLLiCiTiT:
I found out it's broken on other places too. Would it be ok if I would create new PR?

Tilix4:
> I found out it's broken on other places too. Would it be ok if I would create new PR?

@iLLiCiTiT, sure, the goal is for it to work, no matter who does it :)

iLLiCiTiT:
Made PR fixing also other places https://github.com/ynput/OpenPype/pull/4597 @Tilix4 thanks for spotting and reporting the issue.

PR #4596, opened 2023-03-09
Kitsu: Sync OP with zou, make sure value-data is int or float

EmberLightVFX:
Currently the data zou pulls is a string and not a value causing some bugs in the pipe where a value is expected (like `Set frame range` in Fusion).

This PR makes sure each value is set with int() or float() so these bugs can't happen later on.

_(A request to cgwire have also bin sent to allow force values only for some metadata columns, but currently the user can enter what ever they want in there)_

### DIFF
--- a/openpype/modules/kitsu/utils/update_op_with_zou.py
+++ b/openpype/modules/kitsu/utils/update_op_with_zou.py
@@ -128,8 +128,8 @@ def update_op_assets(
             if frames_duration:
                 frame_out = frame_in + frames_duration - 1
             else:
-                frame_out = project_doc["data"].get("frameEnd", frame_in)
-        item_data["frameEnd"] = frame_out
+                frame_out = int(project_doc["data"].get("frameEnd", frame_in))
+        item_data["frameEnd"] = int(frame_out)
         # Fps, fallback to project's value or default value (25.0)
         try:
             fps = float(item_data.get("fps"))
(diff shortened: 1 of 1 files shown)


openpype/modules/kitsu/utils/update_op_with_zou.py
```diff
@@ -128,8 +128,8 @@ def update_op_assets(
             if frames_duration:
                 frame_out = frame_in + frames_duration - 1
             else:
-                frame_out = project_doc["data"].get("frameEnd", frame_in)
-        item_data["frameEnd"] = frame_out
+                frame_out = int(project_doc["data"].get("frameEnd", frame_in))
+        item_data["frameEnd"] = int(frame_out)
```
Tilix4, line 132: No need to cast it twice
```suggestion
                frame_out = project_doc["data"].get("frameEnd", frame_in)
        item_data["frameEnd"] = int(frame_out)
```

Tilix4 (changes requested):

tokejepsen:
> Currently the data zou pulls is a string and not a value causing some bugs in the pipe where a value is expected (like Set frame range in Fusion).

@EmberLightVFX you able to post an error message for these bugs?

EmberLightVFX:
> > Currently the data zou pulls is a string and not a value causing some bugs in the pipe where a value is expected (like Set frame range in Fusion).
> 
> @EmberLightVFX you able to post an error message for these bugs?
@tokejepsen 
Without this fix I'm getting this problem when trying to set the frame range in Fusion:
```
  File "C:\Users\jacob\Syncthing\Projekt\OpenPype\openpype\hosts\fusion\api\menu.py", line 181, in on_set_framerange_clicked
    set_asset_framerange()
  File "C:\Users\jacob\Syncthing\Projekt\OpenPype\openpype\hosts\fusion\api\lib.py", line 73, in set_asset_framerange
    update_frame_range(start, end, set_render_range=True,
  File "C:\Users\jacob\Syncthing\Projekt\OpenPype\openpype\hosts\fusion\api\lib.py", line 51, in update_frame_range
    "COMPN_GlobalStart": start - handle_start,
TypeError: unsupported operand type(s) for -: 'int' and 'str'
```

EmberLightVFX:
@Tilix4 Could you review to see if it's ok to merge?

Tilix4 (approved):
LGTM!

EmberLightVFX:
Hm, it still says review required

Tilix4:
> Hm, it still says review required

Because I don't have merge rights. This is up to @mkolar now

PR #4597, opened 2023-03-10
Tools: Fix recursive filtering

iLLiCiTiT:
## Changelog Description
Fix recursive filtering which is broken since PySide6 support was added.

## Additional info
Change order of available methods to check for legacy ones first to fix PySide2 support. Assets filtering if using Qt recursive filtering if is available.

## Testing notes:
The most changes can be visible in Loader, Scene Inventory and Settings tool.
### Loader
1. Open Loader
2. Try filter assets by nested asset name
3. The parents should stay visible
4. The same for Subsets

### Settings
1. Open settings
2. Press CTRL+F (or find shortcut of your OS)
3. Type some settings key nested in the search dialog
4. The filtering should not hide parents if they don't match the name but their children does

Enhanced [PR](https://github.com/ynput/OpenPype/pull/4595) which solved only one issue.

### DIFF
--- a/openpype/tools/loader/widgets.py
+++ b/openpype/tools/loader/widgets.py
@@ -295,10 +295,10 @@ class SubsetWidget(QtWidgets.QWidget):
             self.model.set_grouping(state)
 
     def _subset_changed(self, text):
-        if hasattr(self.proxy, "setFilterRegularExpression"):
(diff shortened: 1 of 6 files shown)


antirotor (approved):
Tested it with loader and it works well, with Settings, I wasn't able to replicate the issue

PR #4600, opened 2023-03-11
Maya Look Assigner: Don't rely on containers for get all assets

### LINKED ISSUE #4044: Maya: Look Manager Get All Assets on Layout not working
**Running version**
3.14.5

**Describe the bug**
In Maya's Look Manager, **Get All Assets** doesn't work on a referenced layout. Also when you click out of the gui, it disappears behind Maya's window.

**To Reproduce**
1. New Scene
2. Launch **Loader** 
3. Right click on Layout, select reference
4. Launch Look Manager, click on  **Get All Assets**
5. Nothing happens
6. In outliner, select the top group of the layout reference ( Look Manager disappears behind Maya) 
7. Alt tab to get look manager back
8. **Get Assets From Selection** lists assets

**Expected behavior**
Click **Get All Assets** and it lists all the assets. Window should stay in front of Maya.

Clicking **Get All Assets**:
image

Select top group:
image

Clicking  **Get Assets From Selection**:
image

**Desktop:**
 - OS: Windows 11
 - Host: Maya 2023

[cuID:OP-4330]

BigRoy:
## Changelog Description

This resolves #4044 by not actually relying on containers in the scene but instead just rely on finding nodes with `cbId` attributes. As such, imported nodes would also be found and a shader can be assigned (similar to when using get from selection).

**Please take into consideration the potential downsides below**

Potential downsides would be:
- IF an already loaded look has any dagNodes, say a 3D Projection node - then that will also show up as a loaded asset where previously nodes from loaded looks were ignored.
- If any dag nodes were created locally - they would have gotten `cbId` attributes on scene save and thus the current asset would almost always show?

## Testing notes:

1. In Maya, import unmanaged published geometry so there's no container (or use layouts?)
2. Open Look Manager, click "Get all assets" - it should show the asset of the imported geometry.


openpype/hosts/maya/tools/mayalookassigner/commands.py
```diff
@@ -80,21 +80,9 @@ def get_all_asset_nodes():
...
-            continue
-
-        # Gather all information
-        container_name = container["objectName"]
-        nodes += lib.get_container_members(container_name)
-
-    nodes = list(set(nodes))
-    return nodes
+    nodes = cmds.ls(dag=True, noIntermediate=True, long=True)
+    items = create_items_from_nodes(nodes)
```
iLLiCiTiT, line 84: Isn't the `"LookLoader"` filtering important?

BigRoy: The new way doesn't check against containers at all. But in a way somewhat important. This is what I mentioned in the PR description:

> IF an already loaded look has any dagNodes, say a 3D Projection node - then that will also show up as a loaded asset where previously nodes from loaded looks were ignored.
> If any dag nodes were created locally - they would have gotten cbId attributes on scene save and thus the current asset would almost always show?

It filters to dag nodes only which tend to not be present in many looks. It could do a more thorough check to filter out look nodes, however the "from selection" method also doesn't filter out look nodes. (But maybe you're not likely to select them.)

We could do a full container membership lookup to filter out look nodes again if needed - but it'd be quite a bit slower. I heard there was a UI redesign in the works for the Look Assigner (right @mkolar). I think for that design it might be nice to just filter by families of content (so e.g. it'd only show pointcaches if you wanted it to.) Which could then have disabled the `look` family by default. 👍

iLLiCiTiT:
@antirotor this is for you to validate.

LiborBatek (changes requested):
Doesnt work for me...did tests like

1) Imported .ma of several assets
2) checked the Scene Inventory that no present assets containers
3)  Get All Assets via Look Manager

Doesnt work no assets showed up in Look Manager

image


Doesnt get any sort of log in Script editor for more info too.

LiborBatek:
Well still didnt worked for me. Did u actually commit? I didnt see any when pulling the PR in Gitkraken.

image

However works with ```Get Assets from selection``` without being loaded and listed in ```Scene Inventory```....

image

LiborBatek:
Yeah, I did pull it again and even it sounds stupid...with the same result. Not working just when preselected in ```outliner``` and hit ```Get assets from selection``` btn

image


@m-u-r-p-h-y  can you try it also??! I did like 3 times with the same conclusion so not sure if not something wrong on my side?!

BigRoy:
@LiborBatek what version of Maya are you testing in?

LiborBatek:
> @LiborBatek what version of Maya are you testing in? I'll try to reproduce.

maya 2023 and our OP_Cg_Demo project assets....not sure if u have it but we have ZIP on our google drive

BigRoy:
> maya 2023 

Weird. It works fine for me in Maya 2023 and without an error message I'm not sure why it would differ on your end.

You should either be seeing the warning `No id hashes` or the warning `Id not found in the database` or alike.

What does this print for you:
```python
from maya import cmds
from openpype.hosts.maya.tools.mayalookassigner.commands import get_all_asset_nodes, create_asset_id_hash, create_items_from_nodes

print("NODES")
nodes = cmds.ls(dag=True, noIntermediate=True, long=True)
print(f"nodes: {nodes}")

print("NODES from commands")
all_asset_nodes = get_all_asset_nodes()
print(f"nodes: {all_asset_nodes}")
print("Match:", all_asset_nodes == nodes)

print("HASHES")
id_hash = create_asset_id_hash(nodes)
print(f"id_hash: {id_hash}")
for _id, _nodes in id_hash.items():
    print(_id, _nodes)
    
print("ITEMS")
items = create_items_from_nodes(nodes)
print(f"items: {items }")
for item in items:
    print(item)
```

LiborBatek:
I dont get it but it suddenly works ok! Maybe u fix something meanwhile...

image

openpype/hosts/maya/tools/mayalookassigner/commands.py
BigRoy: @antirotor could you still validate whether this is currently ok and still behaves in an acceptable manner for Get All Assets?

PR #3097, opened 2022-04-25
Maya: Render Creator has configurable options.

Allan-I:
## Brief description
Adds configurable render options in the settings.  Resolves #2282.

## Description
Adds schema specified render options to be triggered through the settings instead of hard-coded render options in the render creator plugins.

## Additional info
The intent is to have the configurable options grabbed through a function that's within the render creator class, and that will append the settings by being triggered within the render creator function that sets the renderer settings.

PR #4603, opened 2023-03-11
Maya: Arnold don't reset maya timeline frame range on render creation (or setting render settings)

### LINKED ISSUE #4429: Maya: Frame range reset on render creation
**Running version**
3.15.1-nightly.2

**Describe the bug**
When creating a render instance in Maya, the frame range is reset, which seems like unexpected behaviour. Frame range changes should be explicitly executed by the user and not implicit when creating.

**To Reproduce**
1. Launch Maya.
2. Set frame range to something else than what it was at launch.
3. Create render instance.

BigRoy:
## Changelog Description

Fix #4429: Do not reset fps or playback timeline on applying or creating render settings

## Additional info

I haven't tested this myself. :) But I tweaked it based on what I thought the issue was mentioned in the comment [here](https://github.com/ynput/OpenPype/issues/4429#issuecomment-1464983348):

> I suspect this only [happens for Arnold due to this line here](https://github.com/ynput/OpenPype/blob/47de52bb038de6e414e31008ce0b643396015554/openpype/hosts/maya/api/lib_rendersettings.py#L161) which was [committed here by](https://github.com/ynput/OpenPype/commit/a39eef07f4a91fe775d8c492fc0dbbf9502f4c2f) @Allan-I merged part of https://github.com/ynput/OpenPype/pull/3097 which was a big one. It could be that he knows more as to why it was needed but it could also have been a mistake.

## Testing notes:
1. Start maya
2. Tweak your timeline frame range
3. Create > Render
4. Your timeline should be preserved, render frame range should have been set to match the asset however? (or shouldn't it?)
5. (Additionally also test the _OpenPype > Set Render Settings_ option)


horsebrain:
Hello, I think it was asked for at the time to reset the frame range, but I'm not entirely sure.

BigRoy:
> Hello, I think it was asked for at the time to reset the frame range, but I'm not entirely sure.

Good to know. With this PR it does still reset the frame range, but only the one for the actual render.

tokejepsen (approved):

moonyuet (approved):

moonyuet:
It would be great if you can update your branch.

PR #4605, opened 2023-03-11
Maya: Unify menu labels for "Set Frame Range" and "Set Resolution"

### LINKED ISSUE #4109: Maya: unify menu items (Set/Reset)
We should unify the naming of menu times from current

image

To

Set Frame Range
Set Resolution
Set Colorspace
Set Render Settings

[cuID:[OP-4471](https://app.clickup.com/t/33a7xr3)]

BigRoy:
## Changelog Description

Fix #4109: Unify menu labels for "Set Frame Range" and "Set Resolution"
This also tweaks it in Houdini from Reset Frame Range to Set Frame Range.

## Additional info

Whether we prefer **set** over **reset** or not I'll let for others to decide. It's good to note currently that the functions in the code base are mostly called `reset_x` instead of `set_x` (from a coding perspective somewhat from an obvious perspective since `set_x` looks like you need to pass it the value.

## Testing notes:

1. Start Maya and Houdini
2. It should look super-unified.

### DIFF
--- a/openpype/hosts/blender/api/ops.py
+++ b/openpype/hosts/blender/api/ops.py
@@ -382,8 +382,8 @@ class TOPBAR_MT_avalon(bpy.types.Menu):
         layout.operator(LaunchLibrary.bl_idname, text="Library...")
         layout.separator()
         layout.operator(LaunchWorkFiles.bl_idname, text="Work Files...")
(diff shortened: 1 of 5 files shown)


m-u-r-p-h-y (approved):
Looks nice and tidy . . . 
image

image

PR #4607, opened 2023-03-12
Clockify: refresh and fix the integration

movalex:
## Changelog Description
Due to recent API changes, Clockify requires `user_id` to operate with the timers. 
I updated this part and currently it is a WIP for making it fully functional. Most functions, such as start and stop timer, and projects sync are currently working. For the rate limiting task new dependency is added: [https://pypi.org/project/ratelimiter/](url)

## Additional info
_TODO:_

- [x] ~~Fix the user permissions check. User validation is currently set to True for all users.~~
- [ ] Check for currently running timer, started with the App Launcher. The idea is to check for the currently running timer during application launch, if it is already running for the selected task. For some reason `get_in_progress` does not work in this case. This check works normally when the timer is started with `Clockify - Start Timer` action.
- [ ] Check if Clockify to Ftrack integration works. I currently have no Ftrack subscription, so I would be glad if someone could test it.

## Testing notes:
1. Enable Clockify module and enter workspace_name in the settings
2. Restart OP and fill the API key, generated in the Clockify Profile settings
3. Open Launcher and choose `Sync to Clockify`
4. Check that current OP projects now appear in the Clockify interface
5. Choose any OP task and choose `Clockify - Start Timer`
6. The timer will start, the timer description is in the format `asset_name/task_name`
7. Right click the OP tray and choose `Clockify` -> `Stop Timer`
8. Open Launcher and start any compatible application. 
9. The timer with the description set to the corresponding asset name will be started. If there's already running timer, it will be stopped. 
10. Launch OpenPype menu and switch to another workfile. The timer will change.


openpype/modules/clockify/clockify_module.py
```diff
@@ -63,7 +51,7 @@ def tray_init(self):
         self.bool_timer_run = False
         self.bool_api_key_set = self.clockapi.set_api()
 
-        # Define itself as TimersManager connector
+        # Define itself as ITimersManager connector
```
iLLiCiTiT, line 54:
```suggestion
        # Define itself as TimersManager connector
```

openpype/modules/clockify/clockify_module.py
```diff
@@ -74,6 +62,8 @@ def tray_start(self):
         self.bool_workspace_set = self.clockapi.workspace_id is not None
         if self.bool_workspace_set is False:
             return
+        self.workspace_id = self.clockapi.workspace_id
```
iLLiCiTiT, line 65: Attributes `workspace_id` and `user_id` are filled only if module is initialized in tray. This is by design not right. Ideally you should use the `clockapi` attribute when you need the values elsewhere.

If there is some reason, other than simplify access to them, you should add property methods that would return the value from `clockapi` instead. because I don't see any reason then that. Storing the value directly to module object can cause issue if something would change in clockify object.

```
@property
def workspace_id(self):
    return self.clockapi.workspace_id

@property
def user_id(self):
    return self.clockapi.user_id
```

openpype/modules/clockify/clockify_module.py
```diff
@@ -218,49 +203,55 @@ def register_timers_manager(self, timer_manager_module):
...
-            return
 
-        actual_timer = self.clockapi.get_in_progress()
+        # get running timer to check if we need to start it
+        # TODO: Check running is not working here
         actual_timer_hierarchy = None
         actual_project_id = None
-        if actual_timer is not None:
+        # current timer does not update here
+        actual_timer = self.currently_active_timer
```
iLLiCiTiT, line 225: This is based on module running in tray, which is not right. This method should be possible to execute anywhere where there is an access to the module and should always look to clockify what is the currently running timer. Storing it to module object is useless, user still can change the timer manually.

openpype/modules/clockify/clockify_module.py
```diff
@@ -218,49 +203,55 @@ def register_timers_manager(self, timer_manager_module):
...
         # Check project existence
         project_name = input_data["project_name"]
         project_id = self.clockapi.get_project_id(project_name)
         if not project_id:
-            self.log.warning((
-                "Project \"{}\" was not found in Clockify. Timer won't start."
-            ).format(project_name))
+            self.log.warning(
+                (
+                    f'Project "{project_name}" was not found in Clockify. '
```
iLLiCiTiT, line 244: Module must be Python 2 compatible so no f-strings.

openpype/modules/clockify/launcher_actions/ClockifyStart.py
```diff
@@ -6,9 +6,12 @@
 class ClockifyStart(LauncherAction):
     name = "clockify_start_timer"
     label = "Clockify - Start Timer"
-    icon = "clockify_icon"
+    icon = "app_icons/clockify.png"
     order = 500
     clockapi = ClockifyAPI()
+    clockapi.set_api()
```
iLLiCiTiT, line 12: This will trigger logic on import which is not good. The api should be initialized only when the action is used. Also the workspace of user may change between import and triggering the process. Also if the api initiaization would crash it would crash the action import and user can't know why he does not see the action.

iLLiCiTiT: This is for all the actions.

pyproject.toml
```diff
@@ -71,6 +71,7 @@ pysftp = "^0.2.9"
 dropbox = "^11.20.0"
 aiohttp-middlewares = "^2.0.0"
 opencolorio = "^2.2.0"
+ratelimiter = "^1.2.0.post0"
```
iLLiCiTiT, line 74: What is this for?

movalex: This is for the rate limiting. Clockify currently supports 10 requests per second. 
I guess this [function ](https://github.com/ynput/OpenPype/pull/4607/files#diff-77e6095f1cc4654b5d018ce050a6ccf789596314a1cf168835573a609477a2c2L15) was there for the same reason, although it is not clear how does it work (or whether it works at all) and also it lacks settings.

iLLiCiTiT: > and also it lacks settings.

It still lacks them.

> although it is not clear how does it work

Well using RateLimiter did not help much to be honest.

Dependency which looks like is not maintained to add one functionality is complication. We also need to limit dependencies as much as possible as the clockify module will be soon separated and less dependencies means less issues on deployment.

movalex: > It still lacks them.

By settings I meant some constants that would set the period and max retries. Sorry for not being clear. If these constants are set to some reasonable values, the result is definitely visible, especially for the loop that is checking for currently running timers. So do you suggest to remove it at all and hope this won't have any issues with the rate limiting, or rather incorporate it somewhere inside the module?

openpype/modules/clockify/clockify_api.py
```diff
@@ -236,231 +214,205 @@ def get_tag_id(self, tag_name, workspace_id=None):
...
+            CLOCKIFY_ENDPOINT + action_url, headers=self.headers, json=body
         )
-
         success = False
         if response.status_code < 300:
             success = True
         return success
 
-    def get_in_progress(self, workspace_id=None):
+    def get_current_timer_values(self, response=None):
```
iLLiCiTiT, line 283: This should be private method which always expect response. It is based on inner logic of the object.
```suggestion
    def _get_current_timer_values(self, response):
```

openpype/modules/clockify/clockify_api.py
```diff
@@ -236,231 +214,205 @@ def get_tag_id(self, tag_name, workspace_id=None):
...
+        if response is None:
+            return
+        output = response.json()
+        if isinstance(output, dict):
+            return
```
iLLiCiTiT, line 288: Why it returns `None` when output from server is dictionary?

movalex: it returns list is the timer is found. I've got dict returns when there's some errors, such `{'code': 1000, 'message': 'Multiple or none auth tokens present'}` when the headers was not found. 
I should probably parse the `code` message rather that ditching the dictionary at all...

openpype/modules/clockify/clockify_module.py
iLLiCiTiT: This is still an issue. Your storing values of clockapi object to clockify module object... That should not happen.

openpype/modules/clockify/clockify_api.py
```diff
@@ -6,34 +6,21 @@
...
 class ClockifyAPI:
     def __init__(self, api_key=None, master_parent=None):
         self.workspace_name = None
-        self.workspace_id = None
         self.master_parent = master_parent
         self.api_key = api_key
-        self.request_counter = 0
-        self.request_time = time.time()
-
+        self.__workspace_id = None
```
iLLiCiTiT, line 22: Please don't use double underscore. It's potentially dangerous as python interpreter resolves then in a different way than other attributes.

openpype/modules/clockify/clockify_api.py
```diff
@@ -139,18 +124,24 @@ def validate_workspace(self, name=None):
             return all_workspaces[name]
         return False
 
+    def set_user_id(self):
+        try:
+            user_id = self.get_user_id()
+        except Exception:
+            user_id = False
```
iLLiCiTiT, line 131:
```suggestion
            user_id = None
```

openpype/modules/clockify/clockify_api.py
```diff
@@ -236,231 +214,205 @@ def get_tag_id(self, tag_name, workspace_id=None):
...
+    def get_current_timer_values(self, response=None):
+        if response is None:
+            return
+        output = response.json()
+        if isinstance(output, dict):
+            return
+        current_timer = None
+        if output and isinstance(output, list):
+            (current_timer,) = output  # unpack list
+        return current_timer
```
iLLiCiTiT, line 292:
```suggestion
        if output and isinstance(output, list):
            return output[0]
        return None
```

openpype/modules/clockify/clockify_api.py
```diff
@@ -236,231 +214,205 @@ def get_tag_id(self, tag_name, workspace_id=None):
...
-        self, workspace_id=None, quantity=10
-    ):
+    def get_time_entries(self, workspace_id=None, user_id=None, quantity=10):
         if workspace_id is None:
             workspace_id = self.workspace_id
-        action_url = 'workspaces/{}/timeEntries/'.format(workspace_id)
-        time_check(self)
+        if user_id is None:
+            user_id = self.user_id
+        action_url = "workspaces/{}/user/{}/time-entries/".format(
```
iLLiCiTiT, line 331:
```suggestion
        action_url = "workspaces/{}/user/{}/time-entries".format(
```

openpype/modules/clockify/clockify_module.py
```diff
@@ -129,38 +114,30 @@ def check_running(self):
...
                     # Each pype timer must have at least 2 items!
                     if len(hierarchy_items) < 2:
                         continue
-                    task_name = hierarchy_items[-1]
-                    hierarchy = hierarchy_items[:-1]
 
-                    task_type = None
-                    if len(actual_timer.get("tags", [])) > 0:
-                        task_type = actual_timer["tags"][0].get("name")
+                    *hierarchy, task_name = hierarchy_items
```
iLLiCiTiT, line 134: This is invalid syntax for Python 2. Please keep it as it was or change it to
```
hierarchy = current_timer_hierarchy.split("/")
# Each pype timer must have at least 2 items!
if len(hierarchy_items) < 2:
    continue
task_name = hierarchy.pop(-1)
```

openpype/modules/clockify/clockify_module.py
```diff
@@ -234,33 +212,36 @@ def start_timer(self, input_data):
...
-        # Concatenate hierarchy and task to get description
-        desc_items = [val for val in input_data.get("hierarchy", [])]
-        desc_items.append(input_data["task_name"])
-        description = "/".join(desc_items)
+        if current_timer:
+            current_timer_hierarchy = current_timer.get("description")
+            current_project_id = current_timer.get("projectId")
 
+        # Concatenate hierarchy and task to get description
+        description_items = [val for val in input_data.get("hierarchy", [])]
```
iLLiCiTiT, line 228:
```suggestion
        description_items = list(input_data.get("hierarchy", []))
```

openpype/modules/clockify/ftrack/server/action_clockify_sync_server.py
```diff
@@ -48,13 +48,19 @@ def discover(self, session, entities, event):
...
             return {
                 "success": False,
                 "message": "Clockify Workspace or API key are not set!"
             }
 
-        if self.clockapi.validate_workspace_perm() is False:
+        if (
+            self.clockapi.validate_workspace_permissions(workspace_id, user_id)
+            is False
+        ):
```
iLLiCiTiT, line 63: Variable `user_id` is using `workspace_id`, but bot variables `workspace_id` and `user_id` are defined only because of this line which looks confusing especially when `user_id` is overriden at line 70 and `workspace_id` is not used at line 54 please change it to this:
```suggestion
        if not self.clockapi.validate_workspace_permissions(
            self.clockapi.workspace_id, self.clockapi.user_id
        ):
```

openpype/modules/clockify/launcher_actions/ClockifySync.py
```diff
@@ -4,19 +4,29 @@
...
+            return False
 
     def process(self, session, **kwargs):
+        self.clockapi.set_api()
+        workspace_id = self.clockapi.workspace_id
+        user_id = self.clockapi.user_id
+        if not self.clockapi.validate_workspace_permissions(
+            workspace_id, user_id
+        ):
+            self.log.info("Missing permissions for this action!")
```
iLLiCiTiT, line 28: User who clicked on the action will never know...

openpype/modules/clockify/constants.py
```diff
@@ -9,4 +9,8 @@
 )
 
 ADMIN_PERMISSION_NAMES = ["WORKSPACE_OWN", "WORKSPACE_ADMIN"]
-CLOCKIFY_ENDPOINT = "https://api.clockify.me/api/"
+CLOCKIFY_ENDPOINT = "https://api.clockify.me/api/v1/"
+
+# rate limiter
+MAX_CALLS = 10
+PERIOD = 2
```
iLLiCiTiT, line 16: These constants are not used.

openpype/modules/clockify/clockify_module.py
```diff
@@ -129,38 +112,33 @@ def check_running(self):
...
+                    project = self.clockapi.get_project_by_id(current_proj_id)
                     if project and project.get("code") == 501:
                         continue
 
-                    project_name = project["name"]
+                    project_name = project.get("name")
 
-                    actual_timer_hierarchy = actual_timer["description"]
-                    hierarchy_items = actual_timer_hierarchy.split("/")
+                    current_timer_hierarchy = current_timer.get("description")
```
iLLiCiTiT, line 128: If there can be missing description and it was changed to save access of `"description"` then the rest of logic should be skipped.
```suggestion
                    current_timer_hierarchy = current_timer.get("description")
                    if not current_timer_hierarchy:
                        continue
```

openpype/modules/clockify/clockify_api.py
```diff
@@ -236,231 +217,219 @@ def get_tag_id(self, tag_name, workspace_id=None):
...
+        current_timer = self.get_in_progress()
+        current_timer_hierarchy = None
+        current_project_id = None
+        current_task_id = None
 
         # Check if is currently run another times and has same values
-        current = self.get_in_progress(workspace_id)
-        if current is not None:
+        # DO not restart the timer, if it is already running for curent task
+        if current_timer and current_timer is not None:
```
iLLiCiTiT, line 256:
```suggestion
        if current_timer:
```

openpype/modules/clockify/clockify_api.py
```diff
@@ -236,231 +217,219 @@ def get_tag_id(self, tag_name, workspace_id=None):
...
+            current_timer_hierarchy = current_timer.get("description")
+            current_project_id = current_timer.get("projectId")
+            current_task_id = current_timer.get("taskId")
             if (
-                current.get("description", None) == description and
-                current.get("projectId", None) == project_id and
-                current.get("taskId", None) == task_id
+                current_timer
+                and current_timer is not None
```
iLLiCiTiT, line 262:
```suggestion
```

openpype/modules/clockify/clockify_api.py
```diff
@@ -236,231 +217,219 @@ def get_tag_id(self, tag_name, workspace_id=None):
...
+                and description == current_timer_hierarchy
```
iLLiCiTiT, line 263: It would not get to this point if current timer would not be set.
```suggestion
                description == current_timer_hierarchy
```

openpype/modules/clockify/launcher_actions/ClockifyStart.py
```diff
@@ -17,23 +17,35 @@ def is_compatible(self, session):
...
         project_name = session["AVALON_PROJECT"]
         asset_name = session["AVALON_ASSET"]
         task_name = session["AVALON_TASK"]
+        asset_doc = get_asset_by_name(project_name, asset_name)
+        task_info = asset_doc["data"]["tasks"][task_name]
+        task_type = task_info["type"]
 
         description = asset_name
-        asset_doc = get_asset_by_name(
+        parents_data = get_asset_by_name(
```
iLLiCiTiT, line 31: You already have asset document on line 26 why are you querying it again?

iLLiCiTiT:
Class `ClockifyApi` is using Python 3 syntax so it must be changed to Python 2 compatible syntax or change how it's used and imported in module. Instead of creating it in `initialize` it could be dynamically created object when needed. So the import would not be at the top but in method.

```
class ClockifyModule(...):
    def initialize(self, modules_settings):
        ...
        self._clockify_api = None
        ...

    @property
    def clockify_api(self):
        if self._clockify_api is None:
            from .clockify_api import ClockifyApi
            self._clockify_api = ClockifyApi(master_parent=self)
        return self._clockify_api
```

poetry.lock
```diff
@@ -1456,11 +1456,13 @@ python-versions = ">=3.6"
 files = [
     {file = "lief-0.12.3-cp310-cp310-macosx_10_14_arm64.whl", hash = "sha256:66724f337e6a36cea1a9380f13b59923f276c49ca837becae2e7be93a2e245d9"},
     {file = "lief-0.12.3-cp310-cp310-macosx_10_14_x86_64.whl", hash = "sha256:6d18aafa2028587c98f6d4387bec94346e92f2b5a8a5002f70b1cf35b1c045cc"},
+    {file = "lief-0.12.3-cp310-cp310-macosx_11_0_arm64.whl", hash = "sha256:d4f69d125caaa8d5ddb574f29cc83101e165ebea1a9f18ad042eb3544081a797"},
```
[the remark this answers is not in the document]
movalex: I think this update was generated by the poetry, I will reverse this.

iLLiCiTiT:
> Check for currently running timer, started with the App Launcher. The idea is to check for the currently running timer during application launch, if it is already running for the selected task. For some reason get_in_progress does not work in this case. This check works normally when the timer is started with Clockify - Start Timer action.

This looks like a bug (`start_time_entry` is for some reason called 3 times on application start). But I would look into that in next PR.

> Check if Clockify to Ftrack integration works. I currently have no Ftrack subscription, so I would be glad if someone could test it.

`TimersManager` cares about timer entries and their sync so it should be fine.

iLLiCiTiT (approved):
Looks good

PR #4611, opened 2023-03-13
Fusion publish existing frames

EmberLightVFX:
## Changelog Description
This PR adds the function to publish existing frames instead of having to re-render all of them for each new publish.
I have split the render_locally plugin so the review-part is its own plugin now.

I also change the saver-creator-plugin's label from Saver to Render (saver) as I intend to add a Prerender creator like in Nuke.

## Additional info
It also contains a validator to make sure that all expected files do exist before trying to publish them.

## Testing notes:
1. Create a saver
2. Render manually
3. Open the publisher
4. Set the render target of the saver to `Use existing frames`
5. Publish without re-rendering!


openpype/hosts/fusion/plugins/publish/validate_local_frames_existence.py
```diff
@@ -0,0 +1,71 @@
...
+                    "\n\n".join(invalid[2]),
+                ),
+                title=self.label,
+            )
+
+    @classmethod
+    def repair(cls, instance):
+        invalid = cls.get_invalid(instance)
+        for tool in invalid:
+            tool.SetInput("CreateDir", 1.0)
```
BigRoy, line 71: I don't think this is supposed to be here? :)
```suggestion

```

openpype/hosts/fusion/plugins/publish/validate_local_frames_existence.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class ValidateLocalFramesExistence(pyblish.api.InstancePlugin):
+    """Checks if files for savers that's set to publish existing frames exists"""
+
+    order = pyblish.api.ValidatorOrder
+    label = "Validate Existing Frames Exists"
+    families = ["render"]
+    hosts = ["fusion"]
+    actions = [RepairAction, SelectInvalidAction]
```
BigRoy, line 17:
```suggestion
    actions = [SelectInvalidAction]
```

openpype/hosts/fusion/plugins/publish/validate_local_frames_existence.py
```diff
@@ -0,0 +1,71 @@
+import os
+import pyblish.api
+
+from openpype.pipeline.publish import RepairAction
```
BigRoy, line 4:
```suggestion
```

openpype/hosts/fusion/plugins/publish/validate_local_frames_existence.py
```diff
@@ -0,0 +1,71 @@
...
+                f"{head}{str(frame).zfill(4)}{ext}"
+                for frame in range(frame_start, frame_end + 1)
+            ]
+
+            non_existing_frames = []
+
+            for file in files:
+                cls.log.error(file)
+                if not os.path.exists(os.path.join(output_dir, file)):
+                    non_existing_frames.append(file)
```
BigRoy, line 45:
```suggestion
                if not os.path.exists(os.path.join(output_dir, file)):
                    cls.log.error(f"Missing file: {file}")
                    non_existing_frames.append(file)
```

openpype/hosts/fusion/plugins/publish/validate_local_frames_existence.py
```diff
@@ -0,0 +1,71 @@
...
+
+            if len(non_existing_frames) > 0:
+                cls.log.error(
+                    "Some of {}'s files does not exist".format(tool.Name)
+                )
```
BigRoy, line 50:
```suggestion
                cls.log.error(
                    f"Some of {tool.Name}'s files does not exist"
                )
```

openpype/hosts/fusion/plugins/publish/extract_review_data.py
```diff
@@ -0,0 +1,55 @@
...
+):
+    """
+    Extract the result of savers by starting a comp render
+    This will run the local render of Fusion.
+    """
+
+    order = pyblish.api.ExtractorOrder - 0.1
+    label = "Extract Review Data"
+    hosts = ["fusion"]
+    families = ["review"]
```
BigRoy, line 19: Shouldn't this be on the "render" family? This doesn't feel like review data but this also now contains the actual file representation data of the render data itself.

The way this is separated into its own plugin feels quite weird to me. Need to read more thoroughly to really understand why it needs to be separate.

openpype/hosts/fusion/plugins/create/create_saver.py
```diff
@@ -205,11 +205,40 @@ def get_managed_tool_data(self, tool):
...
+        return EnumDef(
+            "render_target", items=rendering_targets, label="Render target"
+        )
+
+    def _get_reviewable_bool(self):
+        return BoolDef(
+            "review",
+            default=("reviewable" in self.instance_attributes),
+            label="Review",
+        )
```
BigRoy, line 244: I know the way this is structured in code comes from Nuke, but I honestly don't really understand why - if this is just for a single creator - we need to change this into separate methods for reviewable bool and render target enum. I feel like this could just be done directly in `get_instance_attr_defs` instead or alike. Anyway, not a huge issue with it myself but like the comment described - it doesn't feel in the right place currently.

BigRoy (commented):
Didn't thoroughly look into it - but here are some first things that stood out.

openpype/hosts/fusion/plugins/publish/validate_local_frames_existence.py
EmberLightVFX: I wish to add a repair option for the node. I just accidentally left the code I got from the `validate create dir` plugin.
So gonna review and add the correct code to it soon.

BigRoy: What would the repair button do? :) Render the missing frames?

openpype/hosts/fusion/plugins/create/create_saver.py
EmberLightVFX: Currently it's only 1 creator that uses it but I wish to also add a Rerender creator that also would need to access the `_get_render_target_enum` function and other creators in the future might also need that.

"Biggest" question is where to put that function.

openpype/hosts/fusion/plugins/publish/extract_review_data.py
EmberLightVFX: But if you publish using existing frames you need to get that file-array some how. Should that be in a third plugin then maybe?

openpype/hosts/fusion/plugins/publish/validate_local_frames_existence.py
EmberLightVFX: > What would the repair button do? :) Render the missing frames?

That's my wish :) or to auto-setup the properties to do that in the publisher.

openpype/hosts/fusion/plugins/publish/extract_review_data.py
BigRoy: Since you'd also need to validate they actually exist - I'd actually create a Collector which would collect the expected frames/files, the validator would validate those pre-collected frames and the extractor I guess even the one that's set to 'render local' could just add the representation with those frames (and only render if it's set to "render")?

openpype/hosts/fusion/plugins/create/create_saver.py
BigRoy: Rerender creator? How would it differ?

BigRoy: Same question really - how does a pre-render creator differ? It's still a saver node, with start and end frames?
Just wondering how the behavior differs for a "prerender".

openpype/hosts/fusion/plugins/publish/extract_review_data.py
EmberLightVFX: That's a better way actually! Implemented :)

openpype/hosts/fusion/plugins/publish/collect_existing_frames.py
```diff
@@ -0,0 +1,47 @@
+import pyblish.api
+import os
+
+
+class CollectFusionRenders(pyblish.api.InstancePlugin):
+    """Collect all frames needed to publish existing frames"""
+
+    order = pyblish.api.CollectorOrder + 0.5
+    label = "Collect Existing Frames"
```
BigRoy, line 9:
```suggestion
    label = "Collect Expected Frames"
```

openpype/hosts/fusion/plugins/publish/collect_renders.py
```diff
@@ -0,0 +1,30 @@
...
+
+    """
+
+    order = pyblish.api.CollectorOrder + 0.4
+    label = "Collect Renders"
+    hosts = ["fusion"]
+    families = ["render"]
+
+    def process(self, instance):
+        self.log.debug(pformat(instance.data))
```
BigRoy, line 20: Please do not log the full instance - it's very slow as data tends to grow large.
```suggestion
```

openpype/hosts/fusion/plugins/publish/collect_existing_frames.py
```diff
@@ -0,0 +1,47 @@
+import pyblish.api
+import os
+
+
+class CollectFusionRenders(pyblish.api.InstancePlugin):
```
BigRoy, line 5:
```suggestion
class CollectFusionExpectedRenders(pyblish.api.InstancePlugin):
```
Class name must be unique!

openpype/hosts/fusion/plugins/publish/collect_renders.py
```diff
@@ -0,0 +1,30 @@
...
+
+        render_target = instance.data["render_target"]
+        family = instance.data["family"]
+
+        # add targeted family to families
+        instance.data["families"].append(
+            "{}.{}".format(family, render_target)
+        )
+
+        self.log.debug(pformat(instance.data))
```
BigRoy, line 30:
```suggestion

```

openpype/hosts/fusion/plugins/publish/collect_renders.py
```diff
@@ -0,0 +1,30 @@
+import pyblish.api
+from pprint import pformat
```
BigRoy, line 2:
```suggestion
```

openpype/hosts/fusion/plugins/publish/validate_existing_frames_existence.py
```diff
@@ -0,0 +1,83 @@
...
+            data = invalid[0].GetData("openpype")
+
+            # Change render target to local to render locally
+            data["creator_attributes"]["render_target"] = "local"
+
+            invalid[0].SetData("openpype", data)
+            cls.log.error(
+                f"Reload the publisher and {invalid[0].Name} "
+                "will be set to render locally"
+            )
```
BigRoy, line 83: You should be able in fusion to directly set the value inside a nest dict with dot notation.

```suggestion
            tool = invalid[0]
            
            # Change render target to local to render locally
            tool.SetData(
                "openpype.creator_attributes.render_target", 
                "local"
            )

            cls.log.info(
                f"Reload the publisher and {tool.Name} "
                "will be set to render locally"
            )
```

I wonder whether the logged message should really be an error here - a bit unfriendly? Should that be `info`? I've taken the liberty to make it info in the suggestion.

openpype/hosts/fusion/plugins/publish/extract_render_local.py
```diff
@@ -33,6 +29,7 @@ def process(self, instance):
             )
         )
 
+        # Generate the frame list
```
BigRoy, line 32: Couldn't the logic here also just re-use the Expected Frame list from the Collector?

openpype/hosts/fusion/plugins/publish/validate_existing_frames_existence.py
```diff
@@ -0,0 +1,83 @@
...
+    actions = [RepairAction, SelectInvalidAction]
+
+    @classmethod
+    def get_invalid(cls, instance, non_existing_frames=None):
+        if non_existing_frames is None:
+            non_existing_frames = []
+
+        active = instance.data.get("active", instance.data.get("publish"))
+        if not active:
+            return []
```
BigRoy, line 28: This shouldn't run on an inactive instance as far as I know. If it does, that's a bug elsewhere.

openpype/hosts/fusion/plugins/publish/extract_render_local.py
EmberLightVFX: Yes, yes it could and should... Fixed!

openpype/hosts/fusion/plugins/publish/validate_existing_frames_existence.py
EmberLightVFX: It's actually code taken from `validate_create_folder_checked.py` so I took it as a needed part.
I just tried to comment that part out and it works as expected without that part. Gonna remove it from the code.

openpype/hosts/fusion/plugins/publish/validate_existing_frames_existence.py
EmberLightVFX: That's a better way. Added

EmberLightVFX:
@BigRoy  Could you review to see if it's ok to merge?

jakubjezek001 (approved):
LGTM ;)

PR #4615, opened 2023-03-14
MaxScene Family 

moonyuet:
## Changelog Description
Introduction of the Max Scene Family

## Additional info
Material and rigs would come with the mesh in the exported max scene

## Testing notes:
1. Launch 3dsMax via Openpype 
2. Click Publish....
3. Create Max Scene Instance
4. Publish it


tools/build.ps1
```diff
@@ -22,10 +22,10 @@ https://openpype.io/docs
 
 #>
 
-$arguments=$ARGS
-$disable_submodule_update=""
-if($arguments -eq "--no-submodule-update") {
-    $disable_submodule_update=$true
+$arguments = $ARGS
```
BigRoy, line 25: I don't think the changes in this file are intended for this PR?

moonyuet: I will merge the latest develop in this PR.

LiborBatek:
Did run test in ```3dsmax 2023``` and got this error...

I guess its probably the geometries already been in ```pointcache``` instance so they werent assigned to ```max file``` instance so its empty?!
Will test it when no other instances present and geometries doesnt belong to it.

image

LiborBatek:
So the same error happened even when no other instance present in the workfile (like ```pointcache``` etc) also tried with and without option ```Use Selection``` and end up with the error during ```publish```

image

LiborBatek (changes requested):
Not functional at the moment. Need changes.

openpype/hosts/max/plugins/publish/extract_max_scene_raw.py
```diff
@@ -19,7 +19,8 @@ class ExtractMaxSceneRaw(publish.Extractor,
 
     order = pyblish.api.ExtractorOrder - 0.2
     label = "Extract Max Scene (Raw)"
-    hosts = ["max"]
+    hosts = ["max",
+             "maxScene"]
     families = ["camera"]
```
BigRoy, line 24:
```suggestion
    hosts = ["max"]
    families = ["camera", "maxScene"]
```
I suspect this would fix the bug mentioned by @LiborBatek

tools/build.ps1
BigRoy, line 1: Somehow this file is still showing diffs on Github - I'd expect this PR to make no changes to this file whatsoever so I think something is going wrong.

openpype/hosts/max/plugins/publish/extract_max_scene_raw.py
moonyuet: my bad to set it wrong. will update in the next PR

LiborBatek:
Now all works fine...
image

moonyuet:
I have adjusted update function in the max loader while I have added the validator for just in case the user will create empty container for max scene family.

openpype/hosts/max/plugins/publish/validate_no_max_content.py
BigRoy, line 1: Wouldn't this validator be worth it even for the other families that are allowed to be published (maybe ignoring the workfile family however). But exporting an empty camera instance wouldn't make sense either, right? Or does that have a validator validating this too? Seems like there might be a useful overlap here between instances that could benefit from this logic triggering?

---

Additionally, looking at the design here - in Maya the nodes to be extracted or part of the instance are added into the instance during Collecting. Looking at how much logic involves the get node by name and iterating the children I'd almost want to say a collector should do:

```python
class CollectMembers
    def process(self, instance):
                container = rt.getNodeByName(instance.data["instance_node"])
                members = list(container.Children)
                instance.extend(members)
```

Then this validator would just be:
```python
if not list(instance):
    raise
```

moonyuet: > Wouldn't this validator be worth it even for the other families that are allowed to be published (maybe ignoring the workfile family however). But exporting an empty camera instance wouldn't make sense either, right? Or does that have a validator validating this too? Seems like there might be a useful overlap here between instances that could benefit from this logic triggering?
> 
> Additionally, looking at the design here - in Maya the nodes to be extracted or part of the instance are added into the instance during Collecting. Looking at how much logic involves the get node by name and iterating the children I'd almost want to say a collector should do:
> 
> ```python
> class CollectMembers
>     def process(self, instance):
>                 container = rt.getNodeByName(instance.data["instance_node"])
>                 members = list(container.Children)
>                 instance.extend(members)
> ```
> 
> Then this validator would just be:
> 
> ```python
> if not list(instance):
>     raise
> ```

perhaps yes, I guess this logic can be applied to another families too except from pointcache, it is only one of which the container is not parent to the target objects for extraction.

openpype/hosts/max/plugins/publish/validate_no_max_content.py
```diff
@@ -0,0 +1,23 @@
...
+class ValidateMaxContents(pyblish.api.InstancePlugin):
+    """Validates Max contents.
+
+    Check if MaxScene container includes any contents underneath.
+    """
+
+    order = pyblish.api.ValidatorOrder
+    families = ["camera",
+                "maxScene",
+                "maxrender"]
```
BigRoy, line 16: Wait, `maxrender` doesn't need children right? Or does it?
```suggestion
                "maxScene"]
```

LiborBatek:
Tested without selecting any objects for ```MaxScene``` family and validator correctly kicks in and give error message about it:

image

LiborBatek (approved):
All works as expected. Publish works ok and Validators also.

openpype/hosts/max/plugins/publish/validate_no_max_content.py
moonyuet: you need to have assigned the shot camera as the children of the container before rendering.

mkolar (approved):

PR #4617, opened 2023-03-14
Max: fix the bug of removing an instance

moonyuet:
## Changelog Description
fix the bug of removing an instance in 3dsMax

## Testing notes:
1. Launch 3dsMax via Openpype
2. Click Publish...
3. Create any instance for the obj
4. remove it with the button with the bin icon


openpype/hosts/max/api/plugin.py
```diff
@@ -101,7 +101,9 @@ def remove_instances(self, instances):
             instance_node = rt.getNodeByName(
                 instance.data.get("instance_node"))
             if instance_node:
-                rt.delete(rt.getNodeByName(instance_node))
+                rt.select(instance_node)
+                rt.execute(f'for o in selection do for c in o.children do c.parent = undefined')    # noqa
+                rt.delete(instance_node)
```
BigRoy, line 106: Just wondering, looking at [this](https://help.autodesk.com/view/MAXDEV/2022/ENU/?guid=MAXDEV_Python_using_pymxs_pymxs_scenegraph_html). Wouldn't this work?

```python
# Unparent all children so they are kept in the scene
node = rt.getNodeByName(instance_node)
for obj in node.Children:
    obj.parent = None
rt.delete(instance_node)
```

The parenting should be doable since there's an example [here](https://modelerrigs.wordpress.com/2019/02/26/3dsmaxpymxs-cn/) which does something similar
No idea if that'd work. Or whether executing maxscript is maybe faster? Have never used 3Ds Max really.

Also, a comment might be nice to add as to why we're unparenting objects (I've added one in my example, but not sure if accurate)

moonyuet: I have tried this code and it isn't working. Maybe it could be because of the Max Python API limitation. But the execute command of the MaxScript has done alrighty.

m-u-r-p-h-y (approved):
from user perspective it works as expected

image

PR #4619, opened 2023-03-14
Deactivate closed Kitsu projects on OP

friquette:
## Changelog Description
Deactivate project on OP when the project is closed on Kitsu.


openpype/modules/kitsu/utils/update_op_with_zou.py
```diff
@@ -267,12 +267,22 @@ def write_project_to_op(project: dict, dbcon: AvalonMongoDB) -> UpdateOne:
...
         gazu.project.update_project(project)
 
+    all_kitsu_projects_names = [project['name'] for project in gazu.project.all_projects()]
+    all_kitsu_open_projects_names = [project['name'] for project in gazu.project.all_open_projects()]
+    all_kitsu_closed_projects_names = [project for project in all_kitsu_projects_names if project not in all_kitsu_open_projects_names]
+
+    if project_name in all_kitsu_closed_projects_names:
+        project_status = False
+    else:
+        project_status = True
```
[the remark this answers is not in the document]
friquette: It seems that the closed projects on Kitsu are not synchronized, so we have to make a difference between all the projects and the open projects to see whether a project is open or not

openpype/modules/kitsu/utils/update_op_with_zou.py
```diff
@@ -267,12 +267,29 @@ def write_project_to_op(project: dict, dbcon: AvalonMongoDB) -> UpdateOne:
...
+    all_kitsu_projects_names = [
+        project['name'] for project in gazu.project.all_projects()
+    ]
+    all_kitsu_open_projects_names = [
+        project['name'] for project in gazu.project.all_open_projects()
+    ]
+    all_kitsu_closed_projects_names = [
+        project for project in all_kitsu_projects_names
+        if project not in all_kitsu_open_projects_names
+    ]
```
BigRoy, line 279:
```suggestion
    all_kitsu_projects_names = {
        project['name'] for project in gazu.project.all_projects()
    }
    all_kitsu_open_projects_names = {
        project['name'] for project in gazu.project.all_open_projects()
    }
    all_kitsu_closed_projects_names = (
        all_kitsu_projects_names  - all_kitsu_open_projects_names 
    )
```

I assume the lists of projects can grow quite large over time - especially if it includes the non-open ones. If we end up doing lookups like this I'd say we should go for `set` type.

openpype/modules/kitsu/utils/update_op_with_zou.py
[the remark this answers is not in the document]
BigRoy: [Looking at the logic of the sync all projects it wouldn't ever mark a project inactive](https://github.com/ynput/OpenPype/blob/eea058be0d984eb657d8c3c9ee40b958a9debc83/openpype/modules/kitsu/utils/update_op_with_zou.py#L341) because it'd never start and try to sync the project to begin with. As such, I think the way you're trying to work around it isn't the way to go. Right?

Maybe the way is that syncing all projects should iterate ALL projects including inactive ones, but for the inactive ones only set the status if the project already exists in OpenPype - if not, then ignore it?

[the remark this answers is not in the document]
friquette: So you think instead of modifying the write_project_to_op function, we should modify sync_all_projects to include all projects and not just the open ones ?

openpype/modules/kitsu/utils/update_op_with_zou.py
```diff
@@ -364,9 +372,17 @@ def sync_project_from_kitsu(dbcon: AvalonMongoDB, project: dict):
         if naming_pattern.match(item["name"])
     ]
 
+    all_status = gazu.project.all_project_status()
```
Tilix4, line 375: Comment explaining

openpype/modules/kitsu/utils/update_op_with_zou.py
```diff
@@ -267,12 +269,18 @@ def write_project_to_op(project: dict, dbcon: AvalonMongoDB) -> UpdateOne:
...
+    else:
+        project_status = True
+
     # Update data
     project_data.update(
         {
             "code": project_code,
             "fps": float(project["fps"]),
             "zou_id": project["id"],
+            "active": project_status,
```
Tilix4, line 283:
```suggestion
    # Update data
    project_data.update(
        {
            "code": project_code,
            "fps": float(project["fps"]),
            "zou_id": project["id"],
            "active": project['project_status_name'] != "Closed",
```

Tilix4 (changes requested):
Two small suggestions, otherwise LGTM

Tilix4 (approved):
LGTM!

openpype/modules/kitsu/utils/update_op_with_zou.py
```diff
@@ -364,9 +365,18 @@ def sync_project_from_kitsu(dbcon: AvalonMongoDB, project: dict):
...
+    for status in all_status:
+        if project['project_status_id'] == status['id']:
+            project['project_status_name'] = status['name']
+
     # Sync project. Create if doesn't exist
     bulk_writes.append(write_project_to_op(project, dbcon))
 
+    if project['project_status_name'] == "Closed":
+        return
+
```
BigRoy, line 379: With this PR we're now also writing **closed projects** into OpenPype database that have never been synced before. I wonder whether we should only sync a closed project _IF_ it exists in OpenPype **and** is currently active in OpenPype.

I'd then also move all the asset+shot+episode etc. querying to AFTER knowing whether we want to sync the project. That way we'd only sync to deactivate a project in OP once a project is closed, and otherwise SYNC everything again as soon as the project is active again.

This would make it much lighter, keep it faster as well as soon as you have e.g. 15 closed projects.

openpype/modules/kitsu/utils/update_op_with_zou.py
```diff
@@ -347,6 +347,15 @@ def sync_project_from_kitsu(dbcon: AvalonMongoDB, project: dict):
     if not project:
         project = gazu.project.get_project_by_name(project["name"])
 
+    # Get all statuses for projects from Kitsu
+    all_status = gazu.project.all_project_status()
+    for status in all_status:
+        if project['project_status_id'] == status['id']:
+            project['project_status_name'] = status['name']
+
+    if not get_project(project['name']) and project['project_status_name'] == "Closed":  # noqa
```
iLLiCiTiT, line 356: I would recommend this order (the project is queried only if project status is closed and not all the time).
```suggestion
    if (
        project['project_status_name'] == "Closed"
        and not get_project(project['name'])
    ):
```

iLLiCiTiT (dismissed):

openpype/modules/kitsu/utils/update_op_with_zou.py
```diff
@@ -346,6 +347,18 @@ def sync_project_from_kitsu(dbcon: AvalonMongoDB, project: dict):
     if not project:
         project = gazu.project.get_project_by_name(project["name"])
 
+    # Get all statuses for projects from Kitsu
+    all_status = gazu.project.all_project_status()
+    for status in all_status:
+        if project['project_status_id'] == status['id']:
+            project['project_status_name'] = status['name']
+
```
Tilix4, line 355: I'd add a `break` here for optimization
```suggestion
                break

```

BigRoy, line 355: Would this benefit from a comment?
```suggestion

    # Do not sync closed kitsu project that is not found in openpype
```

BigRoy (dismissed):
Haven't tested the code but changes look good to me.

Only optimization I see if that we could potentially query the `gazu.project.all_project_status()` only once and pass it to `sync_project_from_kitsu` instead of querying it per project. But I assume it's not a performance issue currently for anyone so I suppose it's fine for everyone. :)

openpype/modules/kitsu/utils/update_op_with_zou.py
```diff
@@ -346,6 +347,18 @@ def sync_project_from_kitsu(dbcon: AvalonMongoDB, project: dict):
...
+    if (
+        project['project_status_name'] == "Closed"
+        and not get_project(project['name'])
+    ):
```
Tilix4, line 359: To make it more explicit I'd switch statements and add a comment.
```suggestion
    # Do not create new project if closed
    if (
        not get_project(project['name'])
        and project['project_status_name'] == "Closed"
    ):
```

iLLiCiTiT:
Please update the branch
image

Tilix4 (changes requested):
With all this contributions, it gets very clean and straight forward, nice job!
One actual request lest and one suggestion for better understanding.

openpype/modules/kitsu/utils/update_op_with_zou.py
BigRoy: This would be slower however for all open projects since it'd first touch the OpenPype database. If kept as is then it only queries the OpenPype databse if the project is closed.

iLLiCiTiT: That is from my suggestion https://github.com/ynput/OpenPype/pull/4619#discussion_r1154242344 to avoid query of project all the time.

Tilix4: Ah you're right, disregard the switch then, but keep the comment, pick the one you prefer between @BigRoy's or mine ;)

iLLiCiTiT:
> Please update the branch image

Branch still have conflicts to resolve

BigRoy (approved):
Haven't tested the code, but changes look good to me. Great work!

PR #4621, opened 2023-03-14
fix the deadline pool settings in 3dsMax

moonyuet:
## Changelog Description
fix the bug of the deadline pools not being picked up from Openpype setting

## Testing notes:
1. Launch Openpype
2. Change the deadline pool settings through Openpype setting
3. launch 3dsMax in Openpype
4. Create Render Instance
5. Publish


BigRoy:
Now it ALWAYS picks what the openpype settings define, right? Then why would it ever need to look into `instance.data`? This seems like the wrong approach right?

moonyuet:
> Now it ALWAYS picks what the openpype settings define, right? Then why would it ever need to look into `instance.data`? This seems like the wrong approach right?

You are right....I notice that the group setting has been set similarly. 
Will fix it in the next commit

BigRoy:
3DsMax is using the new publisher, right? :) I think it's about time for having the [settings for the pools on the CollectDeadlinePools instance](https://github.com/ynput/OpenPype/pull/4388/files#diff-2e9b9251a56921f18fd21f17ca4f36791e022e97fd954cc0d9233d601329b6d9) so that for Deadline submissions it's the same logic across all different hosts. The setting could then just provide Host specific pool definitions, etc.

Less maintenance across different integrations if we can just move the logic into something uniform like that.

m-u-r-p-h-y:
Is it something I should wait for to be decided before doing a user test? @antirotor

jakubjezek001:
> 3DsMax is using the new publisher, right? :) I think it's about time for having the [settings for the pools on the CollectDeadlinePools instance](https://github.com/ynput/OpenPype/pull/4388/files#diff-2e9b9251a56921f18fd21f17ca4f36791e022e97fd954cc0d9233d601329b6d9) so that for Deadline submissions it's the same logic across all different hosts. The setting could then just provide Host specific pool definitions, etc.
> 
> Less maintenance across different integrations if we can just move the logic into something uniform like that.

This approach makes much more sense. Besides, I believe we should separate the changes for Deadline Pool collector as own PR and not to include it within the Maya new publisher PR.

BigRoy:
> I believe we should separate the changes for Deadline Pool collector as own PR and not to include it within the Maya new publisher PR.

Totally agree. It was just the easiest way for me to make it 'work' in a sensible way for the Maya PR. But would love to see a separate PR that I can align the Maya one with.

LiborBatek (changes requested):
When created ```RenderMain``` instance in max and then checking the job on DL ....the render pools doesnt bviously not match the ```render pool``` from OP Settings...

image

moonyuet:
> 

Instead of checking the 3dsMax Deadline Pool Setting, you should check the default Pool Setting. The 3dsmax deadline pool setting would be removed in #4658 to minimize the confusion. So basically I should close this ticket for #4658

PR #4622, opened 2023-03-14
Maya: Allow instances to have the same `cbId` attribute

### LINKED ISSUE #3991: Maya Instanced geometry support in ```model``` subset / Validator
When trying to publish ```model``` family I get validation errors because of non-unique IDs of the geometries which shares its ```ShapeNode``` due to instance nature of them using Maya ```DuplicateSpecial``` function with geometry type se to ```Instance```

image

Instancing in maya being crucial function and should be supported at least on a model level (look should be applied normally just multiple times I guess because of the same ```shapeNode``` and object ID...

BigRoy:
## Changelog Description

Allow instanced nodes to have the same `cbId` attribute.

## Additional info

This should fix #3991 

## Testing notes:

1. Publish a `model` with instanced nodes.

## Tested

- [ ] rigging with instances? (Or instanced control shapes like the mGear rig does)
- [ ] alembics with instances and publishing those again maybe
- [ ] lookdev done on published model with instances


BigRoy:
I've opened this as a draft because I haven't tested it on my end - there's likely typos. Anyway, I've written logic that I believe should do what came up in #3991:

- Allow instanced nodes in model sharing the same `cbId` nodes.
- Disallow multiple members that themselves instanced nodes but not of the same instance - so disallow "unique" members where unique is not the same instance.
- Log a warning when multiple nodes of the same instance are in the export

Currently it has a `allow_instances` class attribute which is currently unused - this could however be exposed in settings to maybe enable/disable it per project or even per family if needed? For now I've just enabled the allow instances to be able to test the logic.

LiborBatek:
Did tested in ```maya 2023``` on ```model``` instance publish

When creating multiple instances (dish models in my case) it gives me following errors:

image

I was not sure if I can make ```instance``` from the geometry without ```cbID``` so I have made it in 2 steps first publish with the single geometry (which didnt exist before in the ```model``` asset ) and later did create those other geometries as ```instance``` and tried to ```publish``` the ```model``` asset.

Will also do test do it in 1 step right away (new geometry added and create instances of it and then publish the model asset)

LiborBatek:
also gives me weird  Shader Connections error even its not related in ```model``` instance??

image

BigRoy:
Thanks, will look into it once I have some more time.

LiborBatek:
Small note/thought ...could be valid also for ```Texture Reference``` objects basically they can/should share ```cbId``` too and not causing any issue in OP. It would fix the problems when ```publish``` ```look``` and ```Yeti rig``` families too

BigRoy:
> Small note/thought ...could be valid also for Texture Reference objects basically they can/should share cbId too and not causing any issue in OP.

A texture reference object technically is just a separate mesh - I'm a bit hesitant about 'ignoring' that case explicitly. But thanks for noting it. Just for future reference - the [Texture Reference object use-case came from the comment here](https://github.com/ynput/OpenPype/pull/4554#issuecomment-1471645983)

tokejepsen:
What is the status on this? Ready for review?

LiborBatek (approved):
Did test it again and all seems ok and working.

When Validating the model and present instances with the same cbIds the validator kicks in and user can use ```Repair``` and regenerate Ids.

image

After that ```Publish``` succeeds...
image

BigRoy:
> But when using model.abc representation all instances being discreet shapes not instances anymore.
> Do we prefer this? I guess Alembic supports instances right? Not sure what this could cause, allowing Instancing when using Almebic representation tho.
> However this could be super handy in production....being able to use maya instances without problems lessening data/file sizes overhaul significantly!

The `.abc` file does contain the meshes as instances (the `.abc` file will be much smaller) but unfortunately the Maya Alembic importer does not preserve the instance upon loading. As such, yes - on load these are not instances anymore in Maya. If you load the alembic into an Arnold Proxy however it will correctly load the meshes as instances at rendertime. So there are benefits to it, but not when loading/importing alembics as geo into Maya.

You should be able to reproduce this behavior when generating Alembics manually too.

LiborBatek:
Wow, thx for explanation! Makes sense! So is this PR ready?! or still Draft? 

I guess can be merged or?

BigRoy:
It's one that I _think_ works ok, but there's so many unknown factors afterwards with looks and other publishing that I wonder whether it's worth it.

E.g. referencing the alembic natively back into Maya turns it into non-instances again making it still disallowed to have the same `cbId` if used in a rig or used in a `pointcache` down the line. Yes, it'd work for the model itself but I have a feeling it doesn't really hold up well through the pipeline (in Maya that is; renderers and Houdini will be much more capable however).

I've been tinkering whether it's worth the effort or not. Especially given also the confusion it gave e.g. @LiborBatek as he tested it. Others will likely hit the same issues.

tokejepsen:
> I've been tinkering whether it's worth the effort or not. Especially given also the confusion it gave e.g. @LiborBatek as he tested it. Others will likely hit the same issues.

Ok, will await your decision. Close or make the PR ready for review to continue.

tokejepsen:
@BigRoy did you get a chance to have a think about this PR?

BigRoy:
@LiborBatek reported a bug to me on Discord about MGear rigs not publishing as intended with tons of errors on **Missing node ids** or **Non-unique asset ids on nodes**.

After some debugging I noticed that MGear rigs use instanced curve shapes for controls - e.g. for "controlBuffer" controls. For example see:
```python
print(cmds.ls("arm_L0_upv_ctlShape"))
# ['arm_L0_upv_ctl_controlBuffer|arm_L0_upv_ctlShape']
print(cmds.ls("arm_L0_upv_ctlShape", allPaths=True))
# ['arm_L0_upv_ctl_controlBuffer|arm_L0_upv_ctlShape', 'arm_L0_upv_ctl|arm_L0_upv_ctlShape']
```

To avoid those particular bugs it's very likely we'll still need to start implementing this PR. @LiborBatek can you confirm whether you particular issue does work with this branch.

1. Take this branch
2. Open the problematic rig.
3. Save the scene (to allow it to generate the missing ids)
4. Publish rig

---

Some debug scripts for particular use case:

```python
# Regenerate any existing cbId
# (e.g. if you have a scene from somewhere else that has cbIds referring to asset ids not existing in your database)
from maya import cmds
import openpype.hosts.maya.api.lib as lib

nodes = cmds.ls("*.cbId", objectsOnly=True)
for node, id in lib.generate_ids(nodes):
    lib.set_id(node, id, overwrite=True)
```

```python
# Print nodes with missing cbId attributes
# This should print any (non-referenced) nodes without `cbId` attribute that are expected to have one
for node in lib.get_id_required_nodes():
    if cmds.attributeQuery("cbId", node=node, exists=True):
        continue
        
    print(node)
```

LiborBatek:
this particular PR helped me sort out issue with asset containing a lot of same cbIDs and succesfully publishes now...this not being possible in latest develop (OP 3.16.5.x)

Would be great to sort this PR out!

image

LiborBatek:
@BigRoy  speaking of possible issues....u mentioned possible future problems when publishing rigs. Could u elaborate more on that? In my case it was namely rig controls having the same cbIDs not geometry for actual pointcaches aka animation. Any possible issues still? thx

antirotor:
@BigRoy - We've tested it and I think that the code looks good. Is there anything you would like to add? Because I think it is ready to merge.

openpype/hosts/maya/plugins/publish/validate_mesh_shader_connections.py
```diff
@@ -35,14 +29,18 @@ def get_invalid_sets(shapes):
...
             members = cache.get(set_, None)
             if members is None:
-                members = set(cmds.ls(cmds.sets(set_,
-                                                query=True,
-                                                nodesOnly=True), long=True))
-                cache[set_] = members
+                # Bugfix: When querying `maya.cmds.sets` with `nodesOnly=True`
+                # instanced members will always return the first instance path
+                # even though that might not be the actual member of the set
+                # To resolve this we `maya.cmds.sets` regularly and filter to
```
BigRoy, line 39:
```suggestion
                # To resolve this we `maya.cmds.sets` as usual and filter to
```

LiborBatek:
@BigRoy  I am willing to undergo full circle of tests so we can find any possible negatives of this PR before merged (if someday). 

Could you help me with that by making a testing list for possible issues? Im mostly thinking about ```Look Assigner``` etc. Maybe preparing assets containing instances and checking if anything broken then.

tokejepsen:
@BigRoy you happy with this PR atm?
We'll do a round of testing when its marked ready to review by you, then hopefully a merge soon.

BigRoy:
I've marked this ready for review. We currently have no short-term need for this but I'm still quite wary about the potential new bugs this PR might allow where certain meshes may pass through unexpectedly.

If you all would love to do some good testing:

- rigging with instances? (Or instanced control shapes like the mGear rig does)
- alembics with instances and publishing those again maybe
- lookdev done on published model with instances

Feel free to pick up any code changes that might be required.

tokejepsen (changes requested):
I've added a testing list to the description so we can tick them off as we go.

When publishing a look with model instances, the `Look members unique` and `Non Duplicate Instance Members` plugins errors.

BigRoy:
> When publishing a look with model instances, the `Look members unique` and `Non Duplicate Instance Members` plugins errors.

There's two things to keep in mind here:

- Did you load the alembic or a maya ascii reference? If it's the Alembic, note that the loaded meshes are now _not_ instances anymore. Alembic does not load/import to maya native geometry back as instances - it's a limitation of the Alembic plug-in. (It will be instances in Houdini however or when used in rendering in e.g. an Arnold Standin)
- Look members requiring to be unique for this case I think makes sense with how the system works.  Look shader assignment is collected and re-assigned via the `cbId` attribute. This `cbId` attribute will not be unique per instance (since they share the exact same node and it's not a 'per-instance unique attribute' like e.g. the `worldMatrix` attribute is. As such, two instances will always have the same `cbId` value (they share it). This basically means they cannot have a unique shader assigned. (Technically we could implement the check so that _if_ two shapes are instances that we will _allow it_ but only if they share the exact same shader assignment.

tokejepsen:
> Did you load the alembic or a maya ascii reference? If it's the Alembic, note that the loaded meshes are now not instances anymore. Alembic does not load/import to maya native geometry back as instances - it's a limitation of the Alembic plug-in. (It will be instances in Houdini however or when used in rendering in e.g. an Arnold Standin)

Yeah, it was alembic. Tried with ma reference now as well.

> Look members requiring to be unique for this case I think makes sense with how the current look publishing system works. Look shader assignment is collected and re-assigned via the cbId attribute. This cbId attribute will not be unique per instance (since they share the exact same node and it's not a 'per-instance unique attribute' like e.g. the worldMatrix attribute is. As such, two instances will always have the same cbId value (they share it). This basically means they cannot have a unique shader assigned. (Technically we could implement the check so that if two shapes are instances that we will allow it but only if they share the exact same shader assignment.) A workaround that will already work for this is to not put the full top group in the look objectSet but instead the unique objects (excluding the duplicate entries for instances). It will end up assigning then for all instances just fine (since they have the same id). That is, if the look assigning code works with instances.

Yeah, it does make sense that same cbid on the shape node would receive the same look assignment. But would it make sense from a workflow perspective to allow assign unique materials per instance? Could see having the same model but variants in texture/material, would help a lot.
This could be possible if we consider the transform cbid as well when assigning looks.

tokejepsen:
@BigRoy I'm unsure about what needs to happen when you bring in instances to lookdev since the non-unique asset ids plugin errors?
Should we allow or disallow this? Should

LiborBatek:
@tokejepsen @BigRoy we were also question ourselves with @antirotor and came to a conclusion there should not be any issue with that so I would say to ```allow``` it ? or?

LiborBatek:
...lets wait for a decision made by @antirotor before any action happens then

antirotor:
If I understand it correctly, we can allow it and clearly state the fact in documentation, that they will recieve the same look. And then we'll need to work on enhancing look system as we've already started a long time ago? :)

tokejepsen (changes requested):
The duplicate cbid validation needs to be bypassed.

mkolar:

PR #4623, opened 2023-03-14
Global | Nuke: fixing farm publishing workflow

jakubjezek001:
## Changelog Description
After Nuke had adopted new publisher with new creators new issues were introduced. Those issues were addressed with this PR. Those are for example broken reviewable video files publishing if published via farm. Also fixed local publishing. 

## Additional info
Global plugins had to be addressed too so please test also other DCCs if those are still working - talking of AfterEffects, Maya mainly.

## Testing notes:
1. Open Nuke workfile and publish render family node to farm.
2. Ideally test also included slate workflow (you might need to add slate tag to your ExctractReview output preset)
3. After the publish is done check Ftrack if reviewable video (with slate) was published.

### DIFF
--- a/openpype/modules/deadline/plugins/publish/submit_publish_job.py
+++ b/openpype/modules/deadline/plugins/publish/submit_publish_job.py
@@ -756,6 +756,10 @@ class ProcessSubmittedJobOnFarm(pyblish.api.InstancePlugin):
             instance (pyblish.api.Instance): Instance data.
 
         """
+        if not instance.data.get("farm"):
+            self.log.info("Skipping local instance.")
+            return
+
         data = instance.data.copy()
         context = instance.context
         self.context = context
--- a/openpype/hosts/nuke/plugins/publish/extract_review_data.py
+++ b/openpype/hosts/nuke/plugins/publish/extract_review_data.py
@@ -25,7 +25,7 @@ class ExtractReviewData(publish.Extractor):
         # review can be removed since `ProcessSubmittedJobOnFarm` will create
         # reviewable representation if needed
         if (
-            "render.farm" in instance.data["families"]
+            instance.data.get("farm")
             and "review" in instance.data["families"]
         ):
             instance.data["families"].remove("review")
--- a/openpype/hosts/nuke/plugins/publish/extract_review_data_lut.py
+++ b/openpype/hosts/nuke/plugins/publish/extract_review_data_lut.py
@@ -49,7 +49,12 @@ class ExtractReviewDataLut(publish.Extractor):
                 exporter.stagingDir, exporter.file).replace("\\", "/")
             instance.data["representations"] += data["representations"]
 
-        if "render.farm" in families:
+        # review can be removed since `ProcessSubmittedJobOnFarm` will create
+        # reviewable representation if needed
+        if (
+            instance.data.get("farm")
+            and "review" in instance.data["families"]
+        ):
             instance.data["families"].remove("review")
 
         self.log.debug(
--- a/openpype/hosts/nuke/plugins/publish/extract_review_data_mov.py
+++ b/openpype/hosts/nuke/plugins/publish/extract_review_data_mov.py
@@ -105,10 +105,7 @@ class ExtractReviewDataMov(publish.Extractor):
                     self, instance, o_name, o_data["extension"],
                     multiple_presets)
 
-                if (
-                    "render.farm" in families or
-                    "prerender.farm" in families
-                ):
+                if instance.data.get("farm"):
                     if "review" in instance.data["families"]:
                         instance.data["families"].remove("review")
 
--- a/openpype/hosts/nuke/plugins/publish/extract_thumbnail.py
+++ b/openpype/hosts/nuke/plugins/publish/extract_thumbnail.py
@@ -31,7 +31,7 @@ class ExtractThumbnail(publish.Extractor):
 
 
     def process(self, instance):
-        if "render.farm" in instance.data["families"]:
+        if instance.data.get("farm"):
             return
 
         with napi.maintained_selection():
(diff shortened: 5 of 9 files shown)


openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -756,6 +756,10 @@ def process(self, instance):
             instance (pyblish.api.Instance): Instance data.
 
         """
+        if not instance.data.get("farm"):
+            self.log.info("Skipping local instance.")
+            return
```
tokejepsen, line 761: How come this is needed?
Ideally we should be filtering with families, for example `render.farm`, or at least have a comment about why this is needed.

tokejepsen (changes requested):

openpype/modules/deadline/plugins/publish/submit_publish_job.py
jakubjezek001: Yes, this is right, new publisher workflow had introduced this requirement. Is there any particular host you know about which is not adding `farm` key into instance.data?

jakubjezek001: We had just agreed with @antirotor @iLLiCiTiT @kalisp that it should be required for all DCC to collect farm key as True in case it needs to be published on a farm.

kalisp (approved):
Tested seems to work fine (even in Maya, AE)

kalisp:
After discussion with Jezza I enhanced details in `asserts` tests for Nuke. Name of representation looks kinda weird though.
2023-03-15 18_29_30-Testing Platform PC - AnyDesk

openpype/modules/deadline/plugins/publish/submit_publish_job.py
jakubjezek001: Adding here parallel conversation we had yesterday at discord. This is explained a lot and perhaps even addressing @tokejepsen and @BigRoy concerns. 

Quoting @iLLiCiTiT from [here](https://discord.com/channels/517362899170230292/1085580148629319680/1085636176788062320) :
- I don't think pyblish families would save us here. Family filtering requires only one of defined families on instance which can cause more issues and confusion then explicit check of a value -> plugin filter families = ["render", "farm"] would catch all render instances and all farm instances.
So a pyblish plugin which would care only about render going to farm would still have to do a check of families in process method, which is at the end more complex than simple check for a single key.
- Also the families filter is affecting submit publish job which probably should not care about families at all. It's current family filtering can break other hosts if we would add support for same families but without farm support.
- I understand that the plugins would not be filtered by pyblish logic, so they would show in details and their process would be triggered, but it doesn't mean that family filtering is a good idea at this point.
- At the same time I'm not saying that "farm" key is the best solution, but I really don't see how families would solve all the side effects

jakubjezek001:
> After discussion with Jezza I enhanced details in `asserts` tests for Nuke. Name of representation looks kinda weird though. 

Awesome, thanks a lot @kalisp!

openpype/modules/deadline/plugins/publish/submit_publish_job.py
tokejepsen: I guess the issue is with the matching algorithm on Pyblish. Think we can reduce the issue to this test;
```python
from pyblish import api, util


class RenderValidator(api.InstancePlugin):
    order = api.ValidatorOrder
    families = ["render"]
    match = api.Exact

    def process(self, instance):
        instance.data["validationCount"] += 1


class FarmValidator(api.InstancePlugin):
    order = api.ValidatorOrder
    families = ["farm"]
    match = api.Exact

    def process(self, instance):
        instance.data["validationCount"] += 1


class RenderFarmValidator(api.InstancePlugin):
    order = api.ValidatorOrder
    families = ["render", "farm"]
    match = api.Subset

    def process(self, instance):
        instance.data["validationCount"] += 1


plugins = [RenderValidator, FarmValidator, RenderFarmValidator]

context = api.Context()
instance = context.create_instance(
    name="Render", family="render", families=["render"], validationCount=0
)
util.publish(context, plugins)
assert instance.data["validationCount"] == 1

context = api.Context()
instance = context.create_instance(
    name="Farm", family="farm", families=["farm"], validationCount=0
)
util.publish(context, plugins)
assert instance.data["validationCount"] == 1

context = api.Context()
instance = context.create_instance(
    name="RenderFarm", families=["render", "farm"], validationCount=0
)
util.publish(context, plugins)
assert instance.data["validationCount"] == 1

context = api.Context()
instance = context.create_instance(
    name="RenderLayer", families=["layer", "render"], validationCount=0
)
util.publish(context, plugins)
assert instance.data["validationCount"] == 1
```
Not matter what matching algorithm combination we try here, we will always end up with an assertion error due to missing an exclusion matching algorithm, which might be a good idea to implement in Pyblish?

iLLiCiTiT: And how would that solve that the plugin should be executed by combination of `["render", "farm"]`? Also there already were ideas how to add that to pyblish and I don' think any were successful.

tokejepsen: > And how would that solve that the plugin should be executed by combination of ["render", "farm"]? Also there already were ideas how to add that to pyblish and I don' think any were successful.

Not sure. Does the above testing code describe the issue we have?
If so we could use it as basis to come up with ideas to solve the filtering problem.

iLLiCiTiT: Ok, that's reasonable for filtering of families but still some plugins must be able to find out if instance will go to farm or not.

Now practicalities. We need to merge this because it is just broken now and we definetely won't do suggested changes it in this PR. Moving the conversation here https://github.com/ynput/OpenPype/discussions/4663 please. We have to also care about cases when there is more than one farm addon + some changes related to Publisher attributes + families cleanup (exactly the conversation here).

tokejepsen (approved):

PR #4602, opened 2023-03-11
Loader: Remove `context` argument from Loader.__init__()

BigRoy:
## Changelog Description

Remove the previously required `context` argument.

## Additional info

Note: This currenty has a backwards compatibility implementation where the API's loading function (not on the Loader plug-in itself but in `load_with_repre_contexts`) that it'll still assign the `fname` variable on the Loader after initializing and before calling `Loader.load`.

This PR can help #4594 to make it possible to actually "Remove container" and "Switch container"  for containers in the scene marked **not found** that do not anymore have a valid representation, version, asset, subset or even project in de database because those methods do not need the original representation but only the container and/or new representation.

## Testing notes:

1. Test Loaders across DCCs.
2. Preferably also test a plug-in (not in this codebase anymore, but there could be extra loaders in someone else's for still relying on `self.fname` - using that in load _should_ still work due to the backwards compatibility.
3. Also double check the plug-ins that use Hiero and Resolve's `ClipLoader` class.

PR #4625, opened 2023-03-14
Redshift Proxy Support in 3dsMax

moonyuet:
## Changelog Description
Redshift Proxy Support for 3dsMax.
- [x] Creator
- [x] Loader
- [x] Extractor
- [x] Validator
- [x]  Add documentation

## Additional info
The loader will load the redshift with the subset instance parented to the CON
image

## Testing notes:
**Extractor**
1. launch 3dsMax via Openpype
2. Click Publish
3. Create Instance with Redshift Proxy
4. Publish

**Loader**
1. launch 3dsMax via Openpype
2. Click Load
3. RMB click redshiftproxymain
image

4. Redshift Proxy loaded

**Validator**
1. Run the Publisher
2. If there is no Redshift installed in Max, it will raise error
3. If the current renderer is not Redshift, it will raise error and you can repair this to set it back to Redshift before running the extractor


openpype/hosts/max/plugins/load/load_redshift_proxy.py
```diff
@@ -0,0 +1,59 @@
...
+    load,
+    get_representation_path
+)
+from openpype.hosts.max.api.pipeline import containerise
+from openpype.hosts.max.api import lib
+
+
+class RedshiftProxyLoader(load.LoaderPlugin):
+    """Redshift Proxy Loader"""
+
```
BigRoy, line 14:
```suggestion

    label = "Load Redshift Proxy"
```

openpype/hosts/max/plugins/load/load_redshift_proxy.py
```diff
@@ -0,0 +1,59 @@
...
+        files_in_folder = os.listdir(os.path.dirname(filepath))
+        collections, remainder = clique.assemble(files_in_folder)
+        if collections:
+            rs_proxy.is_sequence = True
+
+        container = rt.container()
+        container.name = f"{name}"
+        rs_proxy.Parent = container
+
+        asset = rt.getNodeByName(f"{name}")
```
BigRoy, line 36:
```suggestion
        asset = rt.getNodeByName(name)
```

openpype/hosts/max/plugins/load/load_redshift_proxy.py
```diff
@@ -0,0 +1,59 @@
...
+        filepath = os.path.normpath(self.fname)
+        rs_proxy = rt.RedshiftProxy()
+        rs_proxy.file = filepath
+        files_in_folder = os.listdir(os.path.dirname(filepath))
+        collections, remainder = clique.assemble(files_in_folder)
+        if collections:
+            rs_proxy.is_sequence = True
+
+        container = rt.container()
+        container.name = f"{name}"
```
BigRoy, line 33:
```suggestion
        container.name = name
```

openpype/hosts/max/plugins/load/load_redshift_proxy.py
```diff
@@ -0,0 +1,59 @@
...
+    families = ["redshiftproxy"]
+    representations = ["rs"]
+    order = -9
+    icon = "code-fork"
+    color = "white"
+
+    def load(self, context, name=None, namespace=None, data=None):
+        from pymxs import runtime as rt
+
+        filepath = os.path.normpath(self.fname)
```
BigRoy, line 24: It would be good maybe to prepare for this PR #4602 and use `self.filepath_from_context(context)` instead of `self.fname`

openpype/hosts/max/plugins/publish/extract_redshift_proxy.py
```diff
@@ -0,0 +1,77 @@
...
+            f"""
+fn ProxyExport fp selected:true compress:false connectivity:false startFrame: endFrame: camera:undefined warnExisting:true transformPivotToOrigin:false = (
+	if startFrame == unsupplied then (
+		startFrame = (currentTime.frame as integer)
+	)
+
+	if endFrame == unsupplied then (
+		endFrame = (currentTime.frame as integer)
+	)
+
```
BigRoy, line 46: This would never be the case, yes? Since we're passing `start` and `end` which are forced to `int` type?

openpype/hosts/max/plugins/create/create_redshift_proxy.py
```diff
@@ -0,0 +1,26 @@
...
+    def create(self, subset_name, instance_data, pre_create_data):
+        from pymxs import runtime as rt
+        sel_obj = list(rt.selection)
+        instance = super(CreateRedshiftProxy, self).create(
+            subset_name,
+            instance_data,
+            pre_create_data)  # type: CreatedInstance
+        container = rt.getNodeByName(instance.data.get("instance_node"))
+        # TODO: Disable "Add to Containers?" Panel
+        # parent the selected cameras into the container
```
BigRoy, line 22: Cameras? :) Probably a comment copied from elsewhere?

openpype/hosts/max/plugins/create/create_redshift_proxy.py
```diff
@@ -0,0 +1,26 @@
...
+        for obj in sel_obj:
+            obj.parent = container
+        # for additional work on the node:
+        # instance_node = rt.getNodeByName(instance.get("instance_node"))
```
BigRoy, line 26: Can be cleaned up?

```suggestion
```

openpype/hosts/max/plugins/publish/extract_redshift_proxy.py
moonyuet: i am still working on that so i just put the codes here first. I will do some adjustments.

openpype/hosts/max/plugins/create/create_redshift_proxy.py
moonyuet: I always forget about removing this comment:(, updated without that

LiborBatek:
I am missing RShift license these days so at least I have tested the ```Validator``` for running RS render plugin, seems fine and correctly kicks in when no RS loaded in Max.

image

openpype/hosts/max/plugins/publish/validate_renderer_redshift_proxy.py
```diff
@@ -0,0 +1,50 @@
...
+    def process(self, instance):
+        invalid = self.get_all_renderer(instance)
+        if invalid:
+            raise PublishValidationError("Please install Redshift for 3dsMax"
+                                         " before using this!")
+        invalid = self.get_current_renderer(instance)
+        if invalid:
+            raise PublishValidationError("Current Renderer is not Redshift")
+
+    def get_all_renderer(self, instance):
```
BigRoy, line 30: Function name does not describe what the function actually ends up doing.

openpype/hosts/max/plugins/publish/validate_renderer_redshift_proxy.py
```diff
@@ -0,0 +1,50 @@
...
+        renderer_class = get_current_renderer()
+        current_renderer = str(renderer_class).split(":")[0]
+        if current_renderer != "Redshift_Renderer":
+            invalid.append(current_renderer)
+
+        return invalid
+
+    @classmethod
+    def repair(cls, instance):
+        if "Redshift_Renderer" in str(rt.RendererClass.classes[2]()):
```
BigRoy, line 49: `.classes[2]` seems awfully hardcoded. Why do we need to take the third entry?

openpype/hosts/max/plugins/publish/validate_renderer_redshift_proxy.py
```diff
@@ -0,0 +1,50 @@
...
+        max_renderers_list = str(rt.RendererClass.classes)
+        if "Redshift_Renderer" not in max_renderers_list:
+            invalid.append(max_renderers_list)
+
+        return invalid
+
+    def get_current_renderer(self, instance):
+        invalid = list()
+        renderer_class = get_current_renderer()
+        current_renderer = str(renderer_class).split(":")[0]
```
BigRoy, line 41: Why do we need to split on `:`. What's the original string value returned by `get_current_renderer`?

Would be great if the [`lib.get_current_renderer` function](https://github.com/ynput/OpenPype/blob/9d3b788fb40137e651273f104682b31f907f2550/openpype/hosts/max/api/lib.py#L130) could maybe have some more info in the docstring too describing what output it generates, maybe with some examples.

openpype/hosts/max/plugins/publish/validate_renderer_redshift_proxy.py
```diff
@@ -0,0 +1,50 @@
...
+    families = ["redshiftproxy"]
+    hosts = ["max"]
+    label = "Redshift Renderer"
+    actions = [RepairAction]
+
+    def process(self, instance):
+        invalid = self.get_all_renderer(instance)
+        if invalid:
+            raise PublishValidationError("Please install Redshift for 3dsMax"
+                                         " before using this!")
```
BigRoy, line 25: This error message can be friendlier without the exclamation point. Also, _before using this_  might be slightly unclear. What is "this" when the artist sees this message?

openpype/hosts/max/plugins/publish/validate_renderer_redshift_proxy.py
```diff
@@ -0,0 +1,50 @@
...
+        invalid = self.get_current_renderer(instance)
+        if invalid:
+            raise PublishValidationError("Current Renderer is not Redshift")
```
BigRoy, line 28: This should likely highlight why it must redshift?

openpype/hosts/max/plugins/publish/extract_redshift_proxy.py
```diff
@@ -0,0 +1,64 @@
...
+            'ext': 'rs',
+            # need to count the files
+            'files': rs_filenames if len(rs_filenames) > 1 else rs_filenames[0],    # noqa
+            "stagingDir": stagingdir,
+        }
+        instance.data["representations"].append(representation)
+        self.log.info("Extracted instance '%s' to: %s" % (instance.name,
+                                                          stagingdir))
+
+        # TODO: set sequence
```
BigRoy, line 56: Is this todo wrongly indented? Or should we have an enter after this if it's not related to `get_rsfiles`?

openpype/hosts/max/plugins/publish/extract_redshift_proxy.py
```diff
@@ -0,0 +1,64 @@
...
+
+        self.log.info("Performing Extraction ...")
+
+        if "representations" not in instance.data:
+            instance.data["representations"] = []
+
+        representation = {
+            'name': 'rs',
+            'ext': 'rs',
+            # need to count the files
```
BigRoy, line 48: This comment is unclear. Why do we need to count the files?

openpype/hosts/max/plugins/load/load_redshift_proxy.py
```diff
@@ -0,0 +1,64 @@
...
+from openpype.pipeline import (
+    load,
+    get_representation_path
+)
+from openpype.hosts.max.api.pipeline import containerise
+from openpype.hosts.max.api import lib
+
+
+class RedshiftProxyLoader(load.LoaderPlugin):
+
```
BigRoy, line 13: This empty line should not be there before the docstring.
```suggestion
```

openpype/hosts/max/plugins/create/create_redshift_proxy.py
```diff
@@ -0,0 +1,23 @@
...
+
+class CreateRedshiftProxy(plugin.MaxCreator):
+    identifier = "io.openpype.creators.max.redshiftproxy"
+    label = "Redshift Proxy"
+    family = "redshiftproxy"
+    icon = "gear"
+
+    def create(self, subset_name, instance_data, pre_create_data):
+        from pymxs import runtime as rt
+        sel_obj = list(rt.selection)
```
BigRoy, line 15: Shouldn't the selection only be used whenever `Use selection` pre create checkbox was set?

BigRoy (commented):
Some notes

openpype/hosts/max/plugins/create/create_redshift_proxy.py
moonyuet: i can set `sel_obj = list(rt.selection) or list(self.selected_node)`, hopefully it won't error out

BigRoy: Actually, you don't need to do anything with the selection here, right?
It's already done in the [parent class](https://github.com/ynput/OpenPype/blob/9d3b788fb40137e651273f104682b31f907f2550/openpype/hosts/max/api/plugin.py#L52-L64).

BigRoy: If my previous comment is correct you don't need to implement the `create` method on this class here at all since the parent class already creates the correct implementation for you.

openpype/hosts/max/api/lib.py
```diff
@@ -128,7 +128,14 @@ def list_children(node):
...
 
 def get_current_renderer():
-    """get current renderer"""
+    """
+    Notes:
+        Get current renderer for Max
+
+    Returns:
+        "{Current Renderer}:{Current Renderer}"
+        e.g. "Redshift_Renderer:Redshift_Renderer"
```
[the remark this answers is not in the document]
moonyuet: > If this always returns the value like that, why wouldn't we split it directly as part of this function? Or do the values ever differ?

nothing different, but it should be the separate PR as it will affect the render instance.

openpype/hosts/max/plugins/publish/validate_renderer_redshift_proxy.py
```diff
@@ -0,0 +1,54 @@
...
+
+        return invalid
+
+    @classmethod
+    def repair(cls, instance):
+        renderer_count = len(rt.RendererClass.classes)
+        for r in range(renderer_count):
+            if "Redshift_Renderer" in str(rt.RendererClass.classes[r]()):
+                rt.renderers.production = rt.RendererClass.classes[r]()
+                break
```
BigRoy, line 54: Isn't this the same as this:
```suggestion
        for Renderer in rt.RendererClass.classes:
            renderer = Renderer()
            if "Redshift_Renderer" in renderer:
                rt.renderers.production = renderer
                break
```

openpype/hosts/max/plugins/publish/validate_renderer_redshift_proxy.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+
+    order = pyblish.api.ValidatorOrder
+    families = ["redshiftproxy"]
+    hosts = ["max"]
+    label = "Redshift Renderer"
+    actions = [RepairAction]
+
+    def process(self, instance):
+        invalid = self.get_all_renderer(instance)
```
BigRoy, line 22: Method `get_all_renderer` does not exist

moonyuet: nice catch, i definitley need to get some more coffee with bluberries.

LiborBatek (approved):
Publish went ok and loading too.

image

When Loaded into Scene, all working fine and can be rendered without any issues. Also shown in ```Scene Inventory``` fine.
image

antirotor (approved):
Looks good!

PR #4629, opened 2023-03-15
General: We should keep current subset version when we switch only the representation type

kaamaurice:
## Changelog Description
When we switch only the representation type of subsets, we should not get the representation from the last version of the subset. 

## Testing notes:
1. Load a subset with several representation(any host)
2. Set version that is not the last existing version.
3. Switch the subset representation using the scene inventory.
4. Version should not change to the latest.

## Additional suggestion
We could implement a version combobox selection to the switch dialog.

### DIFF
--- a/openpype/tools/sceneinventory/switch_dialog.py
+++ b/openpype/tools/sceneinventory/switch_dialog.py
@@ -1227,12 +1227,15 @@ class SwitchAssetDialog(QtWidgets.QDialog):
 
         version_ids = list()
 
-        version_docs_by_parent_id = {}
+        version_docs_by_parent_and_name = collections.defaultdict(dict)
+        last_version_docs_by_parent_id = {}
         for version_doc in version_docs:
             parent_id = version_doc["parent"]
-            if parent_id not in version_docs_by_parent_id:
-                version_ids.append(version_doc["_id"])
-                version_docs_by_parent_id[parent_id] = version_doc
+            version_ids.append(version_doc["_id"])
+            name = version_doc["name"]
+            version_docs_by_parent_and_name[parent_id][name] = version_doc
+            if parent_id not in last_version_docs_by_parent_id:
+                last_version_docs_by_parent_id[parent_id] = version_doc
 
         hero_version_docs_by_parent_id = {}
         for hero_version_doc in hero_version_docs:
@@ -1254,6 +1257,7 @@ class SwitchAssetDialog(QtWidgets.QDialog):
 
             container_version_id = container_repre["parent"]
             container_version = self.content_versions[container_version_id]
+            container_version_name = container_version["name"]
 
             container_subset_id = container_version["parent"]
             container_subset = self.content_subsets[container_subset_id]
@@ -1290,7 +1294,11 @@ class SwitchAssetDialog(QtWidgets.QDialog):
                         repre_doc = _repres.get(container_repre_name)
 
             if not repre_doc:
-                version_doc = version_docs_by_parent_id[subset_id]
+                version_by_name = version_docs_by_parent_and_name[subset_id]
+                if selected_asset or selected_subset:
+                    version_doc = last_version_docs_by_parent_id[subset_id]
+                else:
+                    version_doc = version_by_name[container_version_name]
                 version_id = version_doc["_id"]
                 repres_by_name = repre_docs_by_parent_id_by_name[version_id]
                 if selected_representation:
--- a/openpype/tools/sceneinventory/switch_dialog.py
+++ b/openpype/tools/sceneinventory/switch_dialog.py
@@ -1227,15 +1227,12 @@ class SwitchAssetDialog(QtWidgets.QDialog):
 
         version_ids = list()
 
-        version_docs_by_parent_and_name = collections.defaultdict(dict)
-        last_version_docs_by_parent_id = {}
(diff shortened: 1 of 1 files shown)


tokejepsen:
Tested in Nuke and it still updates to latest version.

@kaamaurice which hosts did you test this in?

openpype/tools/sceneinventory/switch_dialog.py
```diff
@@ -1290,7 +1294,11 @@ def _trigger_switch(self, loader=None):
                         repre_doc = _repres.get(container_repre_name)
 
             if not repre_doc:
-                version_doc = version_docs_by_parent_id[subset_id]
+                version_by_name = version_docs_by_parent_and_name[subset_id]
+                if selected_asset or selected_subset:
+                    version_doc = last_version_docs_by_parent_id[subset_id]
```
BigRoy, line 1299: Couldn't this just be 
```python
# Use last version
version_name = max(version_by_name)
version_doc = version_by_name[version_name]
```
So that we don't need to separately keep track of the `last_version_docs_by_parent_id`.

tokejepsen:
@kaamaurice any reason for closing this?

kaamaurice:
@tokejepsen Sorry, I have a very weird issue with branche names how start with `enhancement/`
VSCode add an upper case and I can't push commits after publishing the branch.

openpype/tools/sceneinventory/switch_dialog.py
```diff
@@ -1227,12 +1227,15 @@ def _trigger_switch(self, loader=None):
 
         version_ids = list()
 
-        version_docs_by_parent_id = {}
+        version_docs_by_parent_and_name = collections.defaultdict(dict)
```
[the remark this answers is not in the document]
kaamaurice: whoops, obviously, my bad.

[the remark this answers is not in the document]
kaamaurice: fixed

tokejepsen (dismissed):
Tested in Maya 2023 with `model` and it still updates to latest version.

openpype/tools/sceneinventory/switch_dialog.py
```diff
@@ -1294,17 +1291,32 @@ def _trigger_switch(self, loader=None):
...
                 else:
-                    repre_doc = repres_by_name[container_repre_name]
+                    repres_name = container_repre_name
+
+                # If selected representation is not available with container
+                # version, we use latest version.
+                if repres_name not in repres_by_name:
+                    repres_by_name = repre_docs_by_parent_id_by_name[
+                        version_by_name[max(version_by_name)]["_id"]
+                    ]
```
BigRoy, line 1317: The comment hints to me that this should be in the `if selected_representation:` branch?

Also I feel like this logic here is going to be confusing as to when this occurs. Should this at least log a debug or warning about this edge case?

kaamaurice: Ok I revert this change. This shouldn't be a big problem right now. I have to do some crash testing to be sure of this feature.

BigRoy:
Any chance we could see movement int his PR again @tokejepsen @kaamaurice ?

iLLiCiTiT:
Can someone test this with hero versions?

kaamaurice:
> Can someone test this with hero versions?
```python
  File "D:\Loops\Tools\OpenPype\openpype\tools\sceneinventory\switch_dialog.py", line 1181, in _on_accept
    self._trigger_switch()
  File "D:\Loops\Tools\OpenPype\openpype\tools\sceneinventory\switch_dialog.py", line 1257, in _trigger_switch
    container_version_name = container_version["name"]
KeyError: 'name'
```
Oups I have an issue, sorry, I fix this !

kaamaurice:
Works now as expected with hero version.

mkolar:
This is looking good, however whatever I do I can't checkout the branch. Actually i can't even find in when I add you remote @kaamaurice .

Tilix4:
Hi @mkolar, did you have time to review since the last changes?

openpype/tools/sceneinventory/switch_dialog.py
```diff
@@ -1227,12 +1227,12 @@ def _trigger_switch(self, loader=None):
 
         version_ids = list()
 
-        version_docs_by_parent_id = {}
```
Minkiu, line 1230: I would probably keep it as `version_docs_by_parent_id` and add a tuple in the value field:
```py
version_docs_by_parent_id = collections.defaultdict(dict)

for version_doc in versions_docs:
	version_ids.append(version_doc["id"])
	version_docs_by_parent_id[version_doc["parent"]] = (
		version_doc["name"],
		version_doc
	)
```


or even:
```py

versions_ids = [
	version_doc["id"]
	for version_doc in version_docs
]

version_docs_by_parent_id = {
	version_doc["parent"]: (version_doc["name"], version_doc)
	for version_doc in version_docs
}
```

Then down below:
```py
version_name, version_doc = version_docs_by_parent_id[subset_id]
...
```

iLLiCiTiT: That won't help, you can have multiple versions under single subset.

iLLiCiTiT (approved):
Changes make sense. LGTM

Issue #3890, opened 2022-09-21
Maya: ValidateRenderSettings can fail on float values


vonbehr:
**Running version**
3.12.2

**Describe the bug**
When validating Vray Render Attibutes that have a float value, the validation can fail because of precision errors.

Example error message:
`Invalid value 2.200000047683716 set on vraySettings.cmap_gamma. Expecting 2.2`

**To Reproduce**
1. Configure project_settings/maya/publish/ValidateRenderSettings/vray_render_attributes with VRaySettingsNode.map_gamma and a value of 2.2.
2. Create a render scene in Maya.
3. Create a render family...
4. ...and try to run validation.

**Expected behavior**
Scene should validate.

**Desktop (please complete the following information):**
 - OS: Windows
 - Host: Maya 2022
 - Plugin: V-Ray 5.20.02

BigRoy:
The question really would be, for those values - what precision would we like to compare float values with? Then if the value we're comparing with we could do the check like `math.abs(value-check_value) < precision` so that deviations of e.g. `1e-6` (= `0.000001`) are allowed. 

The deviation in the problem here is about 4e-8.

BigRoy:
Have looked into this a bit more and the `abs(value-check_value) < tolerance` with tolerance of `1e-6` works perfect.

However, it does leave the question with what we'd want to do with potential other attributes. 
How do we want to handle Enum attributes? (Do we allow the user to set it by its string value, or only by its index?)
How do we want to handle Boolean values? (Do we allow the settings to be "False", "True", "no", "yes", "off", "on", or just 1 or 0?

 - and it can be slightly more involved than we'd think to imThe settings for OpenPype will always store the settin

```python
import six
from maya import cmds


def _convert_value(value, value_type):
    """Convert a value to an explicit python type.
    
    This allows string conversions to boolean from
    "False", "True", "no", "off", "yes" and "1", "0"

    """
    
    if type(value) is value_type:
        # No conversion needed
        return value
    
    # Boolean
    if value_type is bool:
        if isinstance(value, str):
            if value.isnumeric():
                # Consider "0" or "0.0" as False
                value = bool(float(value))
            elif value.lower() in {"false", "no", "off"}:
                value = False
            else:
                value = bool(value)
        else:
            value = bool(value)
        return value
            
    # Integer, Float, String straight conversion
    if value_type in {int, float, str}:
        return value_type(value)
                
    # Passthrough any other type
[...]

PR #4631, opened 2023-03-15
Maya: Multiple values on single render attribute - OP-4131

tokejepsen:
## Changelog Description
When validating render attributes, this adds support for multiple values. When repairing first value in list is used.

## Testing notes:
1. Add attributes and values to test against in `project_settings/maya/publish/ValidateRenderSettings` (https://openpype.io/docs/admin_hosts_maya#render-settings-validator).
2. Publish a render instance in Maya.


BigRoy:
The PR description was slightly confusing me.

1. It's good to note: **This is a backwards incompatible change,** right? Old settings values will be lost after this PR due to it now being a list instead of text.
2. With this PR it's now possible to validate an attribute against multiple values. As such, it can now allow multiple values to be set - but if it's invalid it'll repair it to the first value in the list. Previously it was forced to one particular value that must've been set.

Could you describe a use case where we want one specific attribute to be allowed any of multiple values? E.g. based on what request was this implemented?

tokejepsen:
> It's good to note: This is a backwards incompatible change, right? Old settings values will be lost after this PR due to it now being a list instead of text.

@iLLiCiTiT you able to chime in here? Guessing there are no way of making transition smooth? We could make it comma separated and account for it in the code?

> Could you describe a use case where we want one specific attribute to be allowed any of multiple values? E.g. based on what request was this implemented?

Client requested.

"Setting up `project_settings/maya/publish/ValidateRenderSettings/vray_render_attributes` at the moment. It would be great if it would be possible to configure more than one valid value for an attribute. E.g. valid image formats would be exr, exr (multichannel) and vrimg."

openpype/hosts/maya/plugins/publish/validate_rendersettings.py
```diff
@@ -305,6 +325,10 @@ def repair(cls, instance):
             "{aov_separator}", instance.data.get("aovSeparator", "_")
         )
 
+        for attribute, data in cls.get_nodes(instance, renderer).items():
+            for node in data["nodes"]:
+                lib.set_attribute(attribute, data["values"][0], node)
```
moonyuet, line 330: I am not quite sure about this, does the setting only allow the attributes with string type? If not, the function of **set_attributes** needs some rewrites so that the render attributes with varied type can be repaired after the validation. RIght now it only allows the attributes with string type to be repaired. Otherwises, if I add something like `aiOptions.AA_samples`, it will give the RuntimeError and it can't perform the repairment as it is an attribute with integer type.
`
File "D:\toke\pype\build\exe.win-amd64-3.9\openpype\hosts\maya\api\lib.py", line 1395, in set_attribute
cmds.setAttr(node_attr, value, type=attr_type)
RuntimeError: setAttr: The attribute 'defaultArnoldRenderOptions.AASamples' does not accept data of type 'string'.
`

BigRoy: Actually, since you brought this up. Some time ago this got reported #3890 as an issue and I [wrote this comment](https://github.com/ynput/OpenPype/issues/3890#issuecomment-1255167358) with some examples on how it can tackle different cases but actually ALSO do the validation correctly with e.g. booleans, etc. and it will allow a tolerance for float values since 1.0 in settings might be 1.000000000000123 when doing `getAttr` on it in Maya. Since booleans are basically supported due to the string conversion I made a commit that seemed to work [here](https://github.com/BigRoy/OpenPype/commit/5cd0069f1500bb278fbb6e40d8502cf41d035da2) as [mentioned in a later comment](https://github.com/ynput/OpenPype/issues/3890#issuecomment-1255578184)

Would this be the moment to consider those additional types like enum attributes, booleans, float value precisions, etc in our tests?

openpype/hosts/maya/plugins/publish/validate_rendersettings.py
```diff
@@ -305,6 +352,12 @@ def repair(cls, instance):
             "{aov_separator}", instance.data.get("aovSeparator", "_")
         )
 
+        for attribute, data in cls.get_nodes(instance, renderer).items():
+            if not data["values"]:
+                continue
+            for node in data["nodes"]:
+                lib.set_attribute(attribute, data["values"][0], node)
```
moonyuet, line 359: Does the loop of the data["nodes"]do anything for the set_attribute function? Otherwise, this suggestion also applies although it is a bit with the ugly style(I know it's not quite readable).

```suggestion
            lib.set_attribute(attribute, data["values"][0],  data["nodes"][0])
```

moonyuet:
I tested with my machine, validateRenderSetting will error out if the attribute is not set to the same values as expected.
image
And tested with the repair function, it also works without the runtime error.
I only have one comment(or question) on the specific line of code, but that one is just optional. 
So far so good.

moonyuet (dismissed):

openpype/hosts/maya/plugins/publish/validate_rendersettings.py
tokejepsen: > Would this be the moment to consider those additional types like enum attributes, booleans, float value precisions, etc in our tests?

Sorry, I think that is beyond the scope of this PR.

openpype/hosts/maya/plugins/publish/validate_rendersettings.py
tokejepsen: We need to set the attribute on all nodes, since there could (but unlikely) be multiple nodes.

LiborBatek:
I have tested single value for attrib as @moonyuet  which worked for her too but now Im in progress of testing multiple values for single attrib as name of this PR suggests.

Need more testing.

LiborBatek (approved):
Set a multiple values for single render attribute for arnold:

image


then the Validator correctly kicks in when values not met (in my case ```exr``` not being listed) :

image

and ```Repair``` action correctly sets the first value to ```PNG``` in my case

image

After that ```validator``` is passed and ```publish``` is ok

moonyuet:
i have tried in Arnold. Collector for Render Layers in Redshift and Vray are not working, so I can't really test on these two.
image

tokejepsen:
> i have tried in Arnold, looks good. Collector for Render Layers in Redshift and Vray are not working, so I can't really test on these two(not related this PR)

@moonyuet What is your error message here?

BigRoy:
> Collector for Render Layers in Redshift and Vray are not working, so I can't really test on these two(not related this PR)

Would be good to report the issue separately with more details about the error.

moonyuet:
Issue created here: https://github.com/ynput/OpenPype/issues/4648

PR #4633, opened 2023-03-15
Ftrack: Ftrack additional families filtering

iLLiCiTiT:
## Changelog Description
Ftrack family collector makes sure the subset family is also in instance families for additional families filtering.

## Additional info
When additional filtering would look for combination of `"render"` and `"review"` families on instance it would not work when the `"render"` would be subset family.

## Testing notes:
1. Define additional filtering of families with `"plate"` and `"review"` for Collect Ftrack Familiy in `traypublisher` (in a way that you can validate it works)
2. Open Tray Publisher
3. Create Plate with review representation
4. Hit validate (you don't need to publish to validate the collector works)
5. The instance should have/not have `"ftrack"` family based on your settings

### DIFF
--- a/openpype/modules/ftrack/plugins/publish/collect_ftrack_family.py
+++ b/openpype/modules/ftrack/plugins/publish/collect_ftrack_family.py
@@ -70,7 +70,7 @@ class CollectFtrackFamily(pyblish.api.InstancePlugin):
             result_str = "Adding"
             self.log.debug("Adding ftrack family for '{}'".format(family))
             if "ftrack" not in families:
-                instance.data["families"].append("ftrack")
+                families.append("ftrack")
 
         self.log.info("{} 'ftrack' family for instance with '{}'".format(
             result_str, family
--- a/openpype/modules/ftrack/plugins/publish/collect_ftrack_family.py
+++ b/openpype/modules/ftrack/plugins/publish/collect_ftrack_family.py
@@ -11,18 +11,18 @@ from openpype.lib import filter_profiles
 
 
 class CollectFtrackFamily(pyblish.api.InstancePlugin):
+    """Adds explicitly 'ftrack' to families to upload instance to FTrack.
+
+    Uses selection by combination of hosts/families/tasks names via
+    profiles resolution.
+
+    Triggered everywhere, checks instance against configured.
+
+    Checks advanced filtering which works on 'families' not on main
+    'family', as some variants dynamically resolves addition of ftrack
+    based on 'families' (editorial drives it by presence of 'review')
     """
-        Adds explicitly 'ftrack' to families to upload instance to FTrack.
 
-        Uses selection by combination of hosts/families/tasks names via
-        profiles resolution.
-
-        Triggered everywhere, checks instance against configured.
-
-        Checks advanced filtering which works on 'families' not on main
-        'family', as some variants dynamically resolves addition of ftrack
-        based on 'families' (editorial drives it by presence of 'review')
-    """
     label = "Collect Ftrack Family"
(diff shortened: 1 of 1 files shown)


openpype/modules/ftrack/plugins/publish/collect_ftrack_family.py
```diff
@@ -34,68 +33,65 @@ def process(self, instance):
...
 
         if profile:
-            families = instance.data.get("families")
             add_ftrack_family = profile["add_ftrack_family"]
-
             additional_filters = profile.get("advanced_filtering")
             if additional_filters:
-                self.log.info("'{}' families used for additional filtering".
-                              format(families))
+                families_s = set(families) | {family}
```
BigRoy, line 58: I think we could actually just replace the old `families` variable we had directly with this instead of moving it and renaming it. It being a `set` also helps the `if "ftrack" not in families` below too because O1 lookup in sets?

iLLiCiTiT: Variable `families` is used to add `"ftrack"` family so we should not modify it, and we need to modify it to make sure `instance.data["family"]` is inside the family sequence. Thus new variable...

openpype/modules/ftrack/plugins/publish/collect_ftrack_family.py
```diff
@@ -34,68 +33,65 @@ def process(self, instance):
...
-                if families:
-                    if "ftrack" not in families:
-                        instance.data["families"].append("ftrack")
-                else:
-                    instance.data["families"] = ["ftrack"]
-
         result_str = "Adding"
-        if not add_ftrack_family:
+        if add_ftrack_family:
             result_str = "Not adding"
```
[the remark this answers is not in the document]
iLLiCiTiT: Yes

openpype/modules/ftrack/plugins/publish/collect_ftrack_family.py
BigRoy: But you are not using the mutable `families` variable to add `ftrack`. So current code doesn't really need it? Unless we expect `family` itself to be `ftrack` and not be in `families`?

iLLiCiTiT: I see, I forgot to change it to `families.append("ftrack")`.

openpype/modules/ftrack/plugins/publish/collect_ftrack_family.py
```diff
@@ -34,68 +33,65 @@ def process(self, instance):
...
-
-                if families:
-                    if "ftrack" not in families:
-                        instance.data["families"].append("ftrack")
-                else:
-                    instance.data["families"] = ["ftrack"]
+        result_str = "Not adding"
+        if add_ftrack_family:
+            result_str = "Adding"
+            self.log.debug("Adding ftrack family for '{}'".format(family))
```
kalisp, line 71: This shouldn't be here, imho. Line 75 handles both cases (addition/not addition).

iLLiCiTiT:
```suggestion
```

openpype/modules/ftrack/plugins/publish/collect_ftrack_family.py
jakubjezek001: it is a little thing, but could we rename `families_s` to `families_set`, it took me while to realize what it meant and it is only two letters long and saves time and mental power once is easier to read.

PR #4638, opened 2023-03-16
Create extract PSD sequence for TvPaint

friquette:
## Changelog Description
Create a PSD extractor for TvPaint in order to publish PSD sequences.


iLLiCiTiT:
Similar comments to https://github.com/ynput/OpenPype/pull/4639 . The PSD extraction logic must be explicit and not bounded to current render layer/pass logic. The usage is different and should be explicitly created.

PR #4639, opened 2023-03-16
Create extract JSON file for TvPaint

friquette:
## Changelog Description
Create a JSON extractor for TvPaint in order to publish JSON file with all the data needed from each layer for compositing in After Effects (or any other compatible software).

### DIFF
--- a/openpype/hosts/tvpaint/plugins/publish/extract_sequence.py
+++ b/openpype/hosts/tvpaint/plugins/publish/extract_sequence.py
@@ -1,8 +1,9 @@
 import os
-import copy
+import re
 import tempfile
 
 from PIL import Image
+from copy import deepcopy
 
 import pyblish.api
 
@@ -23,12 +24,14 @@ from openpype.hosts.tvpaint.lib import (
 
 
 class ExtractSequence(pyblish.api.Extractor):
+    order = pyblish.api.ExtractorOrder + 0.01
     label = "Extract Sequence"
     hosts = ["tvpaint"]
     families = ["review", "render"]
 
     # Modifiable with settings
     review_bg = [255, 255, 255, 255]
+    extract_json = False
 
     def process(self, instance):
         self.log.info(
@@ -76,11 +79,9 @@ class ExtractSequence(pyblish.api.Extractor):
 
         # Frame start/end may be stored as float
         frame_start = int(instance.data["frameStart"])
-        frame_end = int(instance.data["frameEnd"])
 
         # Handles are not stored per instance but on Context
         handle_start = instance.context.data["handleStart"]
-        handle_end = instance.context.data["handleEnd"]
 
         scene_bg_color = instance.context.data["sceneBgColor"]
 
@@ -111,11 +112,16 @@ class ExtractSequence(pyblish.api.Extractor):
             "Files will be rendered to folder: {}".format(output_dir)
         )
 
+        result = None
         if instance.data["family"] == "review":
             result = self.render_review(
                 output_dir, mark_in, mark_out, scene_bg_color
             )
-        else:
+        elif self.extract_json and instance.data['family'] == 'renderLayer':
+            self.add_render_layer_to_repre(instance)
+            return
+
+        if not result:
             # Render output
             result = self.render(
                 output_dir,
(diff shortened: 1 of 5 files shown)


openpype/hosts/tvpaint/plugins/publish/integrate_json.py
```diff
@@ -0,0 +1,119 @@
...
+
+        self.log.debug("Updated layer_data: {}".format(layer_data))
+        return layer_data
+
+    def _get_json_repre(self, context):
+        """Get the json representation of the instance.
+        Raises error if more than one representation is found.
+        """
+        json_repre = []
+        for repre in context.get("representations"):
```
BigRoy, line 101: If using `.get` because the key might not exist I think this needs to fall back to a list since you're iterating.
```suggestion
        for repre in context.get("representations", []):
```

openpype/hosts/tvpaint/plugins/publish/integrate_json.py
```diff
@@ -0,0 +1,119 @@
...
+        self.log.info(
+            "* Processing instance \"{}\"".format(instance.data["label"])
+        )
+
+        for repre in instance.data.get("representations"):
+            if repre['name'] != 'png' or 'review' in repre['tags']:
+                continue
+
+            if "renderLayer" in instance.data.get('families') and\
+                    'custom_published_path' not in instance.context.data.keys():
```
BigRoy:
```suggestion
                    'custom_published_path' not in instance.context.data:
```

openpype/hosts/tvpaint/plugins/publish/integrate_json.py
BigRoy, line 1: Just wondering - this seems to be re-writing into a single file to create at most a single `.json` file. Why shouldn't it instead be a `ContextPlugin` to run just once and then iterate over the instances manually to define what the json file should look like and write it once.

I'm also not entirely sure why this needs to be a custom integrator?

openpype/hosts/tvpaint/plugins/publish/extract_sequence.py
```diff
@@ -187,6 +193,60 @@ def process(self, instance):
...
+        for file_name in os.listdir(output_dir):
+            if file_name not in link_data:
+                continue
+
+            # Regex to search the frame number in the name of the file
+            frame_number = re.search(r'\.(\d*)\.png$', file_name)
+            if not frame_number:
+                continue
+
+            if frame_start <= int(frame_number.groups()[0]) <= frame_end:
```
BigRoy, line 233:
```suggestion
            if frame_start <= int(frame_number.group(0)) <= frame_end:
```

openpype/hosts/tvpaint/plugins/publish/extract_sequence.py
```diff
@@ -187,6 +193,60 @@ def process(self, instance):
...
+    def add_render_layer_to_repre(self, instance):
+        """ Add render layers to representation with only frameranged frames
+
+        Args:
+            instance (dict): the current instance of the sequence
+        """
+        json_output_dir = instance.context.data['json_output_dir']
+        layers = instance.data['layers']
+        if len(layers) != 1:
+            raise Exception("ERROR")
```
BigRoy, line 205: This would need a better error message and maybe a more specific Exception too?

openpype/hosts/tvpaint/plugins/publish/extract_sequence.py
```diff
@@ -23,12 +24,14 @@
 
 
 class ExtractSequence(pyblish.api.Extractor):
+    order = pyblish.api.ExtractorOrder + 0.01
```
BigRoy, line 27: Why is this needed? Wouldn't it make more sense to just move the ExtractJson to `-0.01`?

openpype/hosts/tvpaint/plugins/publish/extract_sequence.py
```diff
@@ -111,11 +112,16 @@ def process(self, instance):
...
 
+        result = None
         if instance.data["family"] == "review":
             result = self.render_review(
                 output_dir, mark_in, mark_out, scene_bg_color
             )
-        else:
+        elif self.extract_json and instance.data['family'] == 'renderLayer':
+            self.add_render_layer_to_repre(instance)
+            return
```
BigRoy, line 122: So whenever extract json is enabled it doesn't render but uses existing frames? Some logic here seems off?

It should just be able to render AND generate the JSON, right?

BigRoy (commented):
The PR description I think could use some more clarification on how this feature exactly works.

Whenever extracting a JSON it looks like it's **NOT** also rendering the output files. So it's like either render first, then publish the JSON. Or how is this intended to work?

I think it'd also be worth adding in the description an example JSON output so we know what to expect.

To me it's also a bit unclear as to why this needs a special custom integrator? It appears to be placing the JSON file in a very specific custom location. But why?

iLLiCiTiT:
We have a client reuest to do json extraction for AE, but we're ceratin that it must be separated from render layer/pass logic which current TVPaint implementation have. I also don't think it is possible, or good idea, to use render layer/pass output for the json output. It would at the end limit the output instead of giving more abilities.

PR #4649, opened 2023-03-17
Maya: GPU cache representation

tokejepsen:
## Changelog Description
Implement GPU cache for model, animation and pointcache.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/extract_gpu_cache.py
+++ b/openpype/hosts/maya/plugins/publish/extract_gpu_cache.py
@@ -1,0 +1,45 @@
+import os
+
+from maya import cmds
+
+from openpype.pipeline import publish
+from openpype.hosts.maya.api.lib import (
+    maintained_selection, attribute_values, delete_after
+)
+
+
+class ExtractGPUCache(publish.Extractor):
+    """Extract the content of the instance to an CPU cache file."""
+
+    label = "CPU Cache"
+    hosts = ["maya"]
+    families = ["model"]
+
+    def process(self, instance):
+        staging_dir = self.staging_dir(instance)
+        filename = "{}.abc".format(instance.name)
+
+        # Write out GPU cache file.
+        cmds.gpuCache(
+            instance[:],
+            directory=staging_dir,
+            fileName=filename,
+            saveMultipleFiles=False
+        )
+
+        if "representations" not in instance.data:
+            instance.data["representations"] = []
+
+        representation = {
+            "name": "gpu_cache",
+            "ext": "abc",
+            "files": filename,
+            "stagingDir": staging_dir,
+            "data": {"heroSuffix": "gpu_cache"}
+        }
+
+        instance.data["representations"].append(representation)
+
+        self.log.info(
+            "Extracted instance {} to: {}".format(instance.name, staging_dir)
+        )
--- a/openpype/plugins/publish/integrate_hero_version.py
+++ b/openpype/plugins/publish/integrate_hero_version.py
@@ -18,7 +18,7 @@ from openpype.client.operations import (
     prepare_hero_version_update_data,
     prepare_representation_update_data,
 )
-from openpype.lib import create_hard_link
+from openpype.lib import create_hard_link, StringTemplate
 from openpype.pipeline import (
     schema
 )
@@ -306,6 +306,23 @@ class IntegrateHeroVersion(pyblish.api.InstancePlugin):
                 anatomy_filled = anatomy.format(anatomy_data)
                 template_filled = anatomy_filled[template_key]["path"]
 
+                # For representations that have the same extension, an
+                # additional suffix can be available to make the destination
+                # filename different.
+                hero_suffix = repre_info["representation"]["data"].get(
+                    "heroSuffix"
+                )
+                if hero_suffix:
+                    fill_data = copy.deepcopy(template_filled.used_values)
+                    template_filled.template = template_filled.replace(
+                        "." + fill_data["ext"],
+                        "_{}.{}".format(hero_suffix, fill_data["ext"])
+                    )
+                    template_filled = StringTemplate(
+                        template_filled.template
+                    ).format(fill_data)
+
+                # Prepare new repre
                 repre_data = {
                     "path": str(template_filled),
                     "template": hero_template
@@ -316,7 +333,6 @@ class IntegrateHeroVersion(pyblish.api.InstancePlugin):
                     if value is not None:
                         repre_context[key] = value
 
-                # Prepare new repre
                 repre = copy.deepcopy(repre_info["representation"])
                 repre["parent"] = new_hero_version["_id"]
                 repre["context"] = repre_context
--- a/openpype/plugins/publish/integrate.py
+++ b/openpype/plugins/publish/integrate.py
@@ -398,7 +398,7 @@ class IntegrateAsset(pyblish.api.InstancePlugin):
         self.log.debug("{}".format(op_session.to_data()))
         op_session.commit()
 
-        # Backwards compatibility
+        # Backwards compatibility used in hero integration.
         # todo: can we avoid the need to store this?
         instance.data["published_representations"] = {
             p["representation"]["_id"]: p for p in prepared_representations


openpype/hosts/maya/plugins/publish/extract_gpu_cache.py
```diff
@@ -0,0 +1,45 @@
...
+from openpype.pipeline import publish
+from openpype.hosts.maya.api.lib import (
+    maintained_selection, attribute_values, delete_after
+)
+
+
+class ExtractGPUCache(publish.Extractor):
+    """Extract the content of the instance to an CPU cache file."""
+
+    label = "CPU Cache"
```
BigRoy, line 14:
```suggestion
    label = "GPU Cache"
```

openpype/hosts/maya/plugins/publish/extract_gpu_cache.py
```diff
@@ -0,0 +1,45 @@
...
+
+from maya import cmds
+
+from openpype.pipeline import publish
+from openpype.hosts.maya.api.lib import (
+    maintained_selection, attribute_values, delete_after
+)
+
+
+class ExtractGPUCache(publish.Extractor):
```
BigRoy, line 11: Probably best to allow this extractor to be enabled/disabled in settings.

openpype/plugins/publish/integrate_hero_version.py
```diff
@@ -306,6 +306,23 @@ def integrate_instance(
...
+                )
+                if hero_suffix:
+                    fill_data = copy.deepcopy(template_filled.used_values)
+                    template_filled.template = template_filled.replace(
+                        "." + fill_data["ext"],
+                        "_{}.{}".format(hero_suffix, fill_data["ext"])
+                    )
+                    template_filled = StringTemplate(
+                        template_filled.template
+                    ).format(fill_data)
```
BigRoy, line 323: Interesting - shouldn't this just be using the `name` of the representation then? Since that's required to be unique anyway?

openpype/hosts/maya/plugins/publish/extract_gpu_cache.py
tokejepsen: Good point. Got a load of other settings to implement later.

openpype/plugins/publish/integrate_hero_version.py
tokejepsen: Good point, would still need to enforce the name to be in the template cause current default settings do not have it in there.

tokejepsen: Actually it turns out the logic was not needed. The hero integration plugin was failing because its was checking for overwriting files, which happened when the alembic and gpu cache wanted to be the same file name. Since using `"outputName": "gpu_cache"` on the representation, the output file name is unique and the hero integration plugin succeeds.
Problem is actually with the integration plugin, which does not fail when trying to write to the same file. Have setup an issue for this https://github.com/ynput/OpenPype/issues/4656

openpype/hosts/maya/plugins/publish/extract_gpu_cache.py
```diff
@@ -0,0 +1,63 @@
+import json
+
+from maya import cmds
+
+from openpype.pipeline import publish
+
+
+class ExtractGPUCache(publish.Extractor):
+    """Extract the content of the instance to an CPU cache file."""
```
fabiaserra, line 9:
```suggestion
    """Extract the content of the instance to a GPU cache file."""
```

website/docs/admin_hosts_maya.md
```diff
@@ -110,6 +106,35 @@ or Deadlines **Draft Tile Assembler**.
 This is useful to fix some specific renderer glitches and advanced hacking of Maya Scene files. `Patch name` is label for patch for easier orientation.
 `Patch regex` is regex used to find line in file, after `Patch line` string is inserted. Note that you need to add line ending.
 
+### Extract GPU Cache
+
+![Maya GPU Cache](assets/maya-admin_gpu_cache.png)
+
+- **Evaluate Every** Specifies which samples are saved during cache creation. For example, a value of 2 specifies that only every other sample specified by the Evaluate every # frame(s) option is saved to your Alembic file.
+
+- **Save Every** Specifies how often samples are taken during file creation. By default, one sample of your object's transformations is taken every frame and saved to the Alembic file.
```
BigRoy, line 115: It feels like the descriptions here are reversed? **Evaluate Every** talks about **Save Every** and save talks about taking samples. Are these correct?

BigRoy: > **Step**
> 
> Specifies how often samples are taken during file creation. By default, one sample of your object's transformations is taken every frame and saved to the Alembic file.
> For example, a value of 2 caches the transformations of the current object at every other frame of the Cache Time Range.
>
> **Save every # evaluation(s)**
>     Specifies which samples are saved during cache creation. For example, a value of 2 specifies that only every other sample specified by the Evaluate every # frame(s) option is saved to your Alembic file.

So this does look swapped around, right?

LiborBatek (changes requested):
Got error when trying to output ```pointcache```. Also counts for ```animation``` instance...

image

openpype/settings/defaults/project_settings/maya.json
```diff
@@ -924,6 +924,21 @@
...
             "ogsfx_path": "/maya2glTF/PBR/shaders/glTF_PBR.ogsfx"
+        },
+        "ExtractGPUCache": {
+            "enabled": true,
+            "families": [
+                "model",
+                "animation",
+                "pointcache"
+            ],
+            "simulationRate": 0.0,
```
BigRoy, line 935:
```suggestion
            "simulationRate": 1.0,
```

openpype/settings/defaults/project_settings/maya.json
```diff
@@ -924,6 +924,21 @@
...
+            "sampleMultiplier": 0,
```
BigRoy:
```suggestion
            "simulationRate": 1.0,
            "sampleMultiplier": 1,
```

website/docs/admin_hosts_maya.md
tokejepsen: Yeah, confusing all around. Have change it to `step` and `stepSave`.

tokejepsen:
@LiborBatek sorry, could you try again now?

kalisp:
Does it make sense to implement this to be offloaded to the farm (similarly to `extract_pointcache.py`)?

I think lines with `targets` and first 3 in process are important there.

tokejepsen:
> Does it make sense to implement this to be offloaded to the farm (similarly to extract_pointcache.py)?

@kalisp Would make sense, but beyond the scope of this PR. Needs client support first.

LiborBatek:
> @LiborBatek sorry, could you try again now?

Yeah I will. Are there any special needs speaking of reviewing? There arent any testing steps so I will just publish the ```pointcache``` and try to ```load``` it later ok?

tokejepsen:
> Yeah I will. Are there any special needs speaking of reviewing? There arent any testing steps so I will just publish the pointcache and try to load it later ok?

Yeah, thats right. Have updated the description now.

LiborBatek:
Did another round of tests and got this error:
image

tokejepsen:
> Did another round of tests and got this error:
@LiborBatek yup, sorry. Fixed now.

LiborBatek:
So finally the ```animation``` and ```pointcache``` got published and can be imported as ```GPU cache```

Is it correct that just ```Import GPU Cache``` is the only option? There is no ```Load GPU Cache``` present so?

image

BigRoy:
> Is it correct that just Import GPU Cache is the only option? There is no Load GPU Cache present so basically "one way train" without possibilities to use Scene Inventory later on.

Actually, the label is wrong - it should be named `Load GPU Cache` because it can be managed in the Scene Inventory.

openpype/hosts/maya/plugins/load/load_gpucache.py
```diff
@@ -11,7 +11,7 @@ class GpuCacheLoader(load.LoaderPlugin):
     """Load Alembic as gpuCache"""
 
     families = ["model", "animation", "proxyAbc", "pointcache"]
-    representations = ["abc"]
+    representations = ["gpu_cache"]
```
BigRoy, line 14: We've actually in the past used regular alembic representations in gpuCaches. Actually we still do when e.g. loading an Alembic with certain types of data from Houdini when rendering with Arnold.

This would break that logic for us. I understand how having both `abc` and `gpu_cache` might look polluted/clumsy in the Loader so I understand the decision but we'd still need to be able to load regular alembic files too - for us it's actually more important than to be able to load a slightly optimized playback format that is an exported maya gpuCache (which is a maya output format only).

LiborBatek:
> > Is it correct that just Import GPU Cache is the only option? There is no Load GPU Cache present so basically "one way train" without possibilities to use Scene Inventory later on.
> 
> Actually, the label is wrong - it should be named `Load GPU Cache` because it can be managed in the Scene Inventory.

As I have tested it...it actually is ```Import``` because it was not listed then in the ```Scene Inventory```. so need to be changed to proper ```Load``` instead.

tokejepsen:
> As I have tested it...it actually is Import because it was not listed then in the Scene Inventory. so need to be changed to proper Load instead.

@LiborBatek you should have a container in the scene inventory which you can update to latest version. Let me know if you dont have this?

LiborBatek (dismissed):

openpype/settings/defaults/project_settings/maya.json
```diff
@@ -924,6 +924,21 @@
             "enabled": true,
             "active": true,
             "ogsfx_path": "/maya2glTF/PBR/shaders/glTF_PBR.ogsfx"
+        },
+        "ExtractGPUCache": {
+            "enabled": true,
```
BigRoy, line 929: I feel like this should be disabled by default.

BigRoy (dismissed):
Superb work! I'd expect however that the extraction of gpu caches is disabled by default.

openpype/settings/defaults/project_settings/maya.json
tokejepsen:
```suggestion
            "enabled": false,
```

BigRoy (approved):
Haven't tested it, but code looks good to me

fabiaserra:
looks good to me as well! thank you!

tokejepsen:
@LiborBatek could I get an approval on this?

LiborBatek (changes requested):
Unfortunatelly doesnt work for me...did try on multiple shots but same error found:

image

BigRoy:
@LiborBatek make sure you load the `gpuCache` plug-in in Maya.

openpype/hosts/maya/plugins/publish/extract_gpu_cache.py
```diff
@@ -0,0 +1,63 @@
...
+    step = 1.0
+    stepSave = 1
+    optimize = True
+    optimizationThreshold = 40000
+    optimizeAnimationsForMotionBlur = True
+    writeMaterials = True
+    useBaseTessellation = True
+
+    def process(self, instance):
+        staging_dir = self.staging_dir(instance)
```
BigRoy, line 23: Ensure `gpuCache` plug-in is loaded.
```suggestion
        cmds.loadPlugin("gpuCache", quiet=True)

        staging_dir = self.staging_dir(instance)
```

tokejepsen:
> @LiborBatek make sure you load the gpuCache plug-in in Maya.

Should be loading if extracting GPU cache now.

BigRoy (approved):
Nice work!

PR #4653, opened 2023-03-17
General: Use right plugin class for Collect Comment

iLLiCiTiT:
## Changelog Description
Collect Comment plugin is instance plugin so should inherit from `InstancePlugin` instead of `ContextPlugin`.

## Additional info
Pyblish will resolve the plugin as instance plugin because of `instance` argument in `process` method, but it can be confusing for developer who does not know that.


jakubjezek001 (approved):
LGTM ;)

PR #4655, opened 2023-03-17
Fusion: Write and use MasterPrefs from settings

BigRoy:
## Changelog Description

Fusion MasterPrefs are now generated/update at fusion launch with the value from project settings instead of hardcoded in the repository.

## Additional info

This currently writes the master prefs into the local profile directory from #4456 however since master pref content can be set per project it's extra important that the file contents remain unique per project.

As far as I'm aware the master prefs are never reloaded during the Fusion session but are only used on launch. So technically we could also just write it to a unique temporary file on each launch to never avoid a conflict with potential other projects. Thoughts?

It's functional but as can be seen from the question above it's not perfect yet. **For now consider it a quick prototype**

To add this functionality also came up [here](https://github.com/ynput/OpenPype/pull/4572#discussion_r1128423011)

## Testing notes:

1. Configure your masterprefs in Project Settings > Fusion > Masterprefs `project_settings/fusion/masterprefs`
2. Launch Fusion

---

@movalex I can't seem to add you as a reviewer - so just tagging you here to make you aware of the PR. You can make things grey again! :)


jakubjezek001 (dismissed):

openpype/settings/entities/schemas/projects_schema/schema_project_fusion.json
```diff
@@ -46,6 +46,13 @@
                 }
             ]
         },
+        {
+            "type": "text",
+            "multiline" : true,
+            "use_label_wrap": true,
+            "key": "masterprefs",
+            "label": "Fusion MasterPrefs"
```
jakubjezek001, line 54:
```suggestion
            "label": "Fusion Master Preferences"
```

BigRoy:
Converting to draft as I want to make some changes - in particular I want to try and write the master prefs file as a temporary file instead of a dedicated location from settings. That way it'd avoid renderfarm conflicts between the file and conflicts between different projects if an admin would set it to a non-dynamic path.

openpype/settings/entities/schemas/projects_schema/schema_project_fusion.json
```diff
@@ -46,6 +46,13 @@
                 }
             ]
         },
+        {
+            "type": "text",
+            "multiline" : true,
+            "use_label_wrap": true,
+            "key": "masterprefs",
+            "label": "Fusion Master Preferences"
+         },
```
BigRoy, line 55: @iLLiCiTiT is there any way to increase the size of the text input field in the settings UI?

afbeelding

This feels very cramped.

iLLiCiTiT: Not at this moment. Please create an issue, I may look into that during weekend.

BigRoy: Created https://github.com/ynput/OpenPype/issues/4774

openpype/hosts/fusion/hooks/pre_fusion_profile_hook.py
```diff
@@ -125,6 +125,8 @@ def execute(self):
             force_sync,
         ) = self.get_copy_fusion_prefs_settings()
 
+        self.log.info("Hello")
+
```
EmberLightVFX, line 129: Well hello there yourself 😉

EmberLightVFX:
The code looks good to me 🙂The only thing I'm thinking is if it would be good to move the `Fusion Master Preferences` page in the settings into the `Local Fusion profile settings` page. Currently they seem like two different options but one can't really live without the other so I would simply move `Fusion Master Preferences` into `Local Fusion profile settings`

EmberLightVFX:
> > The code looks good to me 🙂The only thing I'm thinking is if it would be good to move the Fusion Master Preferences page in the settings into the Local Fusion profile settings page. Currently they seem like two different options but one can't really live without the other so I would simply move Fusion Master Preferences into Local Fusion profile settings
> 
> @EmberLightVFX Could you explain why the one couldn't live without the other? :) The current logic for master prefs requires nothing of the local local fusion profile settings, especially now it writes to a temp file.

I took it as this PR is an extension of the profile settings part. Depending on the selection you do with `Copy profile on first launch` and `Resync profile on each launch` under `Local Fusion profile settings` you get different results from `Fusion Master Preferences`.

If that's not the case then I think we need to split these two features more so yours also have the `Copy profile on first launch` functionality on its side.
But that would end up with two very similar features that I'm sure would confuse people.
I would rather remove the `Local Fusion profile settings` page and move `Copy profile on first launch` to `Fusion Master Preferences`'s page.

Btw, after using the profile settings feature now on a couple of computer I never know a reason why you wouldn't want to copy the local profile on launch so I would set it to on as default.

movalex:
> the local profile on launch so I would set it to on as default.
Agreed. Also I usually set Force sync to always run always, which is not necessary in usual work, but helpful with testings. Probably it indeed should be not a preferences option, but an OpenPype menu button.

EmberLightVFX:
> > the local profile on launch so I would set it to on as default.
> 
> Agreed. Also I usually set Force sync to always run, which is not necessary in usual work, but helpful with testings. Probably it indeed should be not a preferences option, but an OpenPype menu button.

If the new masterprefs file is only temp, each time you run Fusion it would be the same as `Force sync` as the pref gets re-created each time. Right?
So only `Copy profile on first launch` would be needed. Maybe rename it to `Merge with local profile` to make more sense with how this PR works.

movalex:
Actually, I thought the idea of masterprefs is to only push the Openpype modifications, leaving all other settings as is, if possible. If the studio uses some preferences variations, they already have it managed with the FUSION_PROFILE variable. Masterprefs should be modifiable, because it sometimes still overwrites some settings. Thus we will avoid misunderstanding, because
1) local preferences file should me top of the hierarchy, and 
2) masterprefs is just the OP addon. 

And since the masterprefs brings minimal changes, I would still make it a json field, instead of text. Like so:

```json
{
    "Locked": false,
    "Global": {
        "Paths": {
            "Map": {
                "OpenPype:": "$(OPENPYPE_FUSION)/deploy",
                "Config:": "UserPaths:Config;OpenPype:Config",
                "Scripts:": "UserPaths:Scripts;Reactor:System/Scripts;OpenPype:Scripts"
            }
        },
        "Script": {
            "PythonVersion": 3,
            "Python3Forced": true
        },
        "UserInterface": {
            "Skin": "Neutral",
            "Language": "en_US"
        }
    }
}
```

Please check my latest updates of the slpp module, it now converts prefs to json with no indentation flaws.

https://github.com/movalex/slpp

EmberLightVFX:
> And since the masterprefs brings minimal changes, I would still make it a json field, instead of text. Like so:
> 
> ```json
> {
>     "Locked": false,
>     "Global": {
>         "Paths": {
>             "Map": {
>                 "OpenPype:": "$(OPENPYPE_FUSION)/deploy",
>                 "Config:": "UserPaths:Config;OpenPype:Config",
>                 "Scripts:": "UserPaths:Scripts;Reactor:System/Scripts;OpenPype:Scripts"
>             }
>         },
>         "Script": {
>             "PythonVersion": 3,
>             "Python3Forced": true
>         },
>         "UserInterface": {
>             "Skin": "Neutral",
>             "Language": "en_US"
>         }
>     }
> }
> ```
> 
> Please check my latest updates of the `slpp` module, it now converts `.prefs` to `json` with no indentation flaws.
> 
> https://github.com/movalex/slpp/blob/main/test_fusion_prefs.py

This could be a good addition to this PR as anyone might copy properties from their existing profile to populate it everywhere within the OP pipeline. Currently you have to convert it manually to python, right?

BigRoy:
I'm a bit confused by the discussion - I'm not entirely sure what I'm missing and even from reading the comments it seems to go over my head why you both think this feature is 1-to-1 related to local profiles.

> I took it as this PR is an extension of the profile settings part. Depending on the selection you do with Copy profile on first launch and Resync profile on each launch under Local Fusion profile settings you get different results from Fusion Master Preferences.

How come? 

> If the new masterprefs file is only temp, each time you run Fusion it would be the same as Force sync as the pref gets re-created each time. Right?

Well, no. Because the master prefs only override certain settings, not your full profile. They are the subset of settings that are to be static elements. They aren't "defaults" like a local profile, they are the hardcoded setting the user cannot customize on their end. Even if the master prefs wasn't temp, Fusion itself doesn't write into it as far as I know. It's the 'master preference' and has nothing to do with your local preferences except for that it overrides certain local settings to always match the master preferences.

If you want no master preference overrides you can just clear the setting and no master prefs will be generated for the fusion session. That way your local settings are maintained.

> This could be a good addition to this PR as anyone might copy properties from their existing profile to populate it everywhere within the OP pipeline. Currently you have to convert it manually to python, right?

No, currently it's in the default Fusion master prefs format that Fusion support. It's just like a regular master prefs file, nothing special. A user is fine with generating that using Python, etc. but the end result will always need to be the format that Fusion itself consumes as input, which is the `.prefs` file which I believe is a `lua` table. I'd rather not add another in-between package which also adds a dependency and more code we'd need to support and maintain.

> If the studio uses some preferences variations, they already have it managed with the FUSION_PROFILE variable.

I'd say this is not true, because the user is still free to "override" this. If you are using custom profiles with `force sync` then I'd say disable that and ONLY use a master prefs file. That'll force everyone to be on the same settings and disallow customizing the settings.

> The MasterPref overwrites the current profile so that's why we now make a temp profile and launch it. That way the users profile will never get changed.
Maybe Merge with local profile isn't the best name as it could indicate that the users normal profile would change. Maybe Use together with local profile would be better. Something like that at least.

There's no way to disable that behavior that master prefs override settings or not. If you don't want overridden behavior, then clear our the master prefs setting.

> This could be a good addition to this PR as anyone might copy properties from their existing profile to populate it everywhere within the OP pipeline.

Those existing profiles are `lua`, not `json`. They would be copy pastable currently and not when we turn it into a json table.

---

The way I see it:

1. Master Prefs is what dictates the MUST BE rules for the fusion settings.
2. The local prefs provide the sensible defaults for the fusion settings which the user at any point in time is free to override.

If local preferences are set to "always sync" then basically they start acting like master prefs but your master prefs then are the preferences you have when running fusion outside of the pipeline (since that's where it syncs from?) and you're actually mixing things up because you're then again basically _un-_pipelining things?

In short, the masterprefs allow "locking" attributes away from user customization [as seen here](https://www.steakunderwater.com/wesuckless/viewtopic.php?p=33726#p33726). More details [here](https://www.steakunderwater.com/VFXPedia/96.0.243.189/indexdca0.html?title=Environment_Variables):

> This variable allows additional "master" prefs files to be used in addition to the main profile prefs, useful for maintaining common settings in a facility. One or more full paths may be specified, separated by semicolons. These files will not be written to by Fusion, and may contain a subset of all available settings. If the line "Locked = true," appears in the main table of a master file, all settings in that file will override other prefs, and may not be altered by the user. An example file that turns off & locks the "Make this machine a Render Master" would look like:

```lua
{
	Global = {
		Network = {
			IsRenderServer = false,
		},
	},
	Locked = true,
}
```

As described the master prefs can also have "Locked = false" set and does support multiple master prefs files (e.g. one for locked settings, and one for unlocked preferences). Technically the 'unlocked' settings I suppose are like "default preferences" which the user can still override.

In that sense local prefs are somewhat similar but Master Prefs are what originally indicates what your settings should look like.

We ended up using custom profiles because whatever the master prefs dictates will still get saved into the local profile's preferences too. So if we wouldn't have local profiles it would then also start messing with preferences of Fusion outside of OpenPype which is what we avoided with the local profile copying.

EmberLightVFX:
I think I get the way you're seeing the feature.
OP will always run the OP local profile. It's up to you/the studio if you want to copy the users normal local profile to your OP local profile or not and then Fusion applies the MasterPrefs to it.
And your MasterPrefs feature is just to create the MasterPrefs dynamically.

The way I saw it was that the profile itself was a temp copy of the users local profile that your dynamic MasterPerfs got applied to by Fusion. 🤦‍♂️ My bad...

I still think merging the two pages into one could be good as they are both profile related. So all profile related stuff is under one and the same page.
The page could be renamed to something like `Profile Settings` + `Copy profile on first launch` and `Resync profile on each launch` be renamed to `Copy local profile...` and `Resync local profile...` to show better what they are related to.

What do you think about that @BigRoy ?

movalex:
> Fusion Master prefs > (Force) synced local profile > OP custom local profile

I think the misunderstanding was there. I could not get the idea why do you put masterprefs before the local prefs. I would not say it _dictates_, it just overrides certain fields, as you correctly mentioned. But the mere existence of masterprefs here is to add the openpype settings, nothing more.
Here's from the VXFPedia [article ](https://www.steakunderwater.com/VFXPedia/96.0.243.189/indexad6a.html?title=Per-User_Preferences_and_Paths#Setting_the_profile_directory)about `FUSION_MasterPrefs` environment variable:

> Administrators may choose to exert more control over the preferences used by artists. For this purpose, Fusion allows one or more "master" preferences file to be used in addition to the artist's local preferences, to initialize or lock down individual preference settings.

The key here is that there might be multiple masterprefs. And ours should have a single purpose. With this logic, the local profile should be left untouched as much as possibe, as it is the entity that is created and maintained by the artist and defines the experience. We should not replace settings there, just _add_ the corresponding OpenPype stuff. And since Masterprefs does the overwrite, ideally we should control, what was overridden or cleared by the Masterprefs, and if possible, bring those settings back.
Currently it is possible by changing default Masterprefs values. Here's what I mean:

If the user has a pathmap like this one:
`UserPaths:Scripts;Reactor:System/Scripts;OUR_BEAUTIFUL_SERVER:Scripts`, the custom path entry will be erased by this line in MasterPrefs: `["Scripts:"] = "UserPaths:Scripts;Reactor:System/Scripts;OpenPype:Scripts"`

This may be handled multiple ways:
1. Ask user to move the `OUR_BEAUTIFUL_SERVER` pathmap to the `UserPaths:`. Then `Scripts:` mapping could be easily overridden. This is the easiest way, and I did just that on my machine, but I guess this wont work in all cases.
2. Let the user edit the masterprefs and add the path they need there, so it would stay unchanged no matter what. That's the reason we added a customizable Masterprefs field. Although, this currently has disadvantages. The text field is unverifiable. So a single missed curly brace will result the OP settings silently ignored. It would be cool to edit the verifiable json field instead. And it also has more lines already.
3. We could check the modified settings and revert all incorrectly overridden values automatically.

Points 2 and 3 would be possible if we would work with json data, and then convert it to lua table. This has multiple advantages:
* JSON is default file format thorough entire OP ecosystem
* It is verifiable with current settings.
* It can be processed on a pre-launch step.

> Those existing profiles are lua, not json. They would be copy pastable currently and not when we turn it into a json table.

Sure, but since OP Masterprefs is a different thing rather than any other MP that may exist, we can treat it our way. The result will always be a Lua table, of course, but the OP settings are JSON. That's the idea.

> I'd rather not add another in-between package which also adds a dependency and more code we'd need to support and maintain.

I understand this concern. But I believe adding the code to the repo shoudl be a matter of weighting good and bad consequences, and I see only good here. Plus this is not a dependency anymore. The code I altered is now incompatible with the initial repo. It is now just a Python3 prefs parser we can add as a module. BTW, I recently tested it with my main prefs file, which is 5000 lines long (with all the comp history lists), and it converted it pretty well back and forth, and very fast.

Here, pls give it a chance, when you can: https://github.com/movalex/slpp

movalex:
Considering this:
> FUSION_MasterPrefs should contain the full paths to one or more prefs files, separated by semicolons. These prefs files will not be written to by Fusion, and may contain a subset of all available settings.
 
I guess (but I should check for sure later) this means that multiple Masterprefs files, defined in the `FUSION_MasterPrefs` environment variable, are applied sequentially, and each consequent file will override the previous settings if they are clashed. That means we should append the OP Masterprefs to the variable paths list.

BigRoy:
Thanks so much for taking the time to thoroughly reply and debate this. I appreciate it very much. Hopefully I'm not too stubborn but I still don't fully see why we need `json` format. 

There are advantages yes but for me it doesn't compare yet to the technical debt that we might need to maintain `slpp` library since we're now working with your particular fork which has 'fixed' what seems has been broken in `slpp` for years. It is now 310 lines of code which isn't crazy but it's a lua <-> json format parser. Your fork seems to have 'fixed' some things which apparently the original tests of the package didn't catch but there are no tests to confirm the fixes will be preserved. I'd say at least having that merged upstream with tests means we can 'rely' on the behavior being maintained in the community. But then again, it's been broken for a long time already so who's to say it's really that actively maintained. There are few commits from 2020 and then the next are from 2016.

> I understand this concern. But I believe adding the code to the repo shoudl be a matter of weighting good and bad consequences, and I see only good here. **Plus this is not a dependency anymore. The code I altered is now incompatible with the initial repo.** It is now just a Python3 prefs parser we can add as a module. BTW, I recently tested it with my main prefs file, which is 5000 lines long (with all the comp history lists), and it converted it pretty well back and forth, and very fast.

It being incompatible with the bigger, more popular repository with a bigger community (even though still not maintained) I'd say is actually a bad thing. We're now fully dependent on you maintaining the library over time, potentially for years to come. Or we'll have to take over and maintain it. 

It being incompatible with the upstream repo still makes this a dependency or at least code that needs to be maintained.

>  it converted it pretty well back and forth

Pretty well doesn't sound all too safe to me. We'll need 100% reliability since this OP launch hook would _always_ trigger and is the core part that currently sets up the openpype integration. I'm not saying it's broken currently, but maintaining that reliability with anything that Fusion starts adding or putting into the master preferences to me seems like much more work than just writing the Fusion native file.

There are definitely pros to managing it through JSON and maybe merging settings with JSON dynamically but to me 1) this PR isn't about those more complex cases, it's solely about making the master prefs functionality that's there configurable by the studio admin and 2) those potential pros still don't outweigh the cons I mentioned above.

---
_Answers to @EmberLightVFX _

> OP will always run the OP local profile. It's up to you/the studio if you want to copy the users normal local profile to your OP local profile or not and then Fusion applies the MasterPrefs to it. 

Yes. Studio defines **required preferences**. The copying of the local preferences is solely a 'helper' so that an artist loses as little as possible of what they might've already customized.

> And your MasterPrefs feature is just to create the MasterPrefs dynamically.

The feature is to allow a studio admin to customize the **required preferences** in their studio, per project if needed.

> The way I saw it was that the profile itself was now a temp copy of the users local profile that your dynamic MasterPerfs got applied to by Fusion. 🤦‍♂️ My bad...

Technically it does function like that. In OpenPype the local preferences are a copy in which the Master Prefs are forced.

> I still think merging the two pages into one could be good as they are both profile related. So all profile related stuff is under one and the same page.
> The page could be renamed to something like Profile Settings + Copy profile on first launch and Resync profile on each launch be renamed to Copy local profile... and Resync local profile... to show better what they are related to.
> 
> What do you think about that @BigRoy ?

I'm fine with it either way.

---
_Answers to @movalex_

>  But the mere existence of masterprefs here is to add the openpype settings, nothing more.

@movalex I wouldn't say so. It's configurable in settings now so I studio admin can specify their own rules.

> User may have other masterprefs variable defined, but we will override it anyway with our masterprefs, that would have a single purpose: bring the OpenPype stuff. With this logic, the local profile should be left untouched as much as possibe, as it is the entity that is created and maintained by the artist and defines the experience. 

The local profile of the user that is copied _is not a master preference_. See it more as the "default" for when you start Fusion through OpenPype. I've always seen it copying the default preferences from outside of OpenPype as a convenience for the artist, nothing more.

> We should not replace settings there, just add the corresponding OpenPype pathmaps and required Fusion settings overrides. And since Masterprefs does the overwrite, ideally we should control, what was cleared by the Masterprefs, and if possible, bring those settings back.

Well, there are some settings that a studio might definitely want to force - which master preferences allow.
But I'd say that for the path mapping of Fusion you'd rather not override to still allow an individual artist to add some extra path mapping. Unfortunately however, that's not how Fusion works. There's no way to have customizable preferences for path mapping but _enforce_ specific custom path mapping of OpenPype or the studio to be in there. You can run a script after Fusion runs and try it there, but you'd need to restart Fusion to have some things to take effect (like trigger startup `.fu` scripts from the path mapped folders). This has been a big debate before that we'd love to be able to avoid that but so far we haven't found a way where we can take that control without making things either way more complicated or keep the same level of control (we still don't want the user removing the values at runtime for example).

> Sure, but since OP Masterprefs is a different thing rather than any other MP that may exist, we can treat it our way. The result will always be a Lua table, of course, but the OP settings are JSON. That's the idea.

I don't see it like that. To me it's the Studio-specific overrides for the preferences. If they require a specific fusion locale set, they do it there, if they require a specific "CPU/GPU" setting or alike set, they do it there. I want to highlight that these are **studio settings maintained by an admin, they are not individual user setting**.

The only real "OP focused" thing there is that it's adding OpenPype to the path mapping which is the only way to have the startup scripts + menus available for Fusion. If this were possible without that and do it through e.g. just an env var like `FUSION_USER_PATH_MAP` we should to that instead but those don't seem to exist.

> I guess (but I should check that for sure) this means that multiple Masterprefs files, defined in the FUSION_MasterPrefs environment variable, are applied sequentially, and each consequent file will override the previous settings, if they are clashed. That means, we should append the OP Masterprefs to the variable paths list.

Yes, I saw that on that page too. If that's the case, we should definitely support appending it. Even more I'd even go as far as making the entry in settings a list so that you could define multiple of them, so you can create master prefs that are locked and some that specifiy unlocked prefs.

---

@mkolar @antirotor @jakubjezek001 would also love some opinions from your end.

EmberLightVFX:
Great writeup @BigRoy 

To sum up my final thoughts as for now I'm happy with how everything currently works but I personally would prefer to merge the two profile related functionalities into one page as I wrote about before, just to keep things as clear as possible.

Second I do like the idea to be able to create multiple masterprefs inside of the settings window. That way a studio could have a pref that is being used on every project but then a second one for that project specifically. Having them separately makes it super quick and easy to just copy-paste the base-preference for each new project.

Great work on this Roy!

openpype/hosts/fusion/hooks/pre_fusion_profile_hook.py
```diff
@@ -152,10 +152,31 @@ def execute(self):
...
+        master_pref_contents = self.data["project_settings"]["fusion"].get(
+            "masterprefs", ""
+        )
+        if not master_pref_contents:
+            self.log.debug("No Fusion master preferences contents set in "
+                           "project settings. Skipping setting of Fusion "
+                           "master preferences.")
+            return
+
+        with tempfile.NamedTemporaryFile(
```
iLLiCiTiT, line 169: Won't this spread preferences in temp folder? I mean, are they cleaned up in some way? Don't think it must be handled in this PR, but I can imagine that this could be an issue over time.

BigRoy: > Won't this spread preferences in temp folder? I mean, are they cleaned up in some way? Don't think it must be handled in this PR, but I can imagine that this could be an issue over time.

Technically it would yes. It'd keep around the estimated 1 kb files for each launch of fusion until temporary folder is cleared by the user or the OS.

movalex:
>  this PR isn't about those more complex cases, it's solely about making the master prefs functionality that's there configurable by the studio admin

I agree that adding the Lua parser would be a much more complex task than the initial PR goal.
I was just really excited that the parser finally worked, and with all the possibilities opened with that. This should be a matter of separate PR to discuss further. Although I'm glad this discussion happened.

I can also agree with the idea of the OP masterprefs as an instrument for maintaining the artist's preferences remotely by the manager. Although, as an artist, I would not be too happy, if it would silently overwrite some custom settings that are unique to my machine... Also I strongly believe that each instrument should have a single purpose. Currently we're using this settings as a preference setter and also as an administration tool. Instead, we could introduce two separate MP files:

1. The one that would force the OP settings (required).
2. Studio manager's settings overrides (optional and also editable by the manager).

movalex:
Regarding to the lua-json prefs parser, it indeed require some work. And I believe it can be very useful in the future. 
However, I've just tried that big prefs file that was generated by the script and it didn't work. It turned out, Fusion prefs has some exotic table elements, such as `ordered()` lists, which was parsed incorrectly. I will fix that. 

> we might need to maintain slpp library since we're now working your particular fork which has 'fixed' what seems has been broken in slpp for years.

I did't actually fix things, rather than forced dict -> lua encoding to comply more with Fusion prefs specifics (mainly regarding the integers key parsing). Also cleaned up a code a bit and removed legacy python dependencies. And I guess shortly this won't be an `slpp` anymore. Maybe `flpp`, for `Fusion Lua Python Parser`... 

> there are no tests to confirm the fixes will be preserved

Absolutely, and I'm going to add some tests to check that any prefs elements are correctly processed.

 > We're now fully dependent on you maintaining the library over time, potentially for years to come. Or we'll have to take over and maintain it.

Sure, agree. I'm surely going to stick around for some time, but there's always a bus factor... But Fusion preferences are in the current state for what, like 20-30 years? It seems we would not expect them to change much over time.

jakubjezek001:
Hey @BigRoy could you merge later develop into this PR? This seems to have still menu item `Set render target`

jakubjezek001:
So there are my two cents on the issue the feature. Take it from Nuke user perspective which might not always be in sync with Fusion mindset. First of all I agree with @EmberLightVFX on following

> I still think merging the two pages into one could be good as they are both profile related. So all profile related stuff is under one and the same page.
> The page could be renamed to something like Profile Settings + Copy profile on first launch and Resync profile on each launch be renamed to Copy local profile... and Resync local profile... to show better what they are related to.

Secondly rather then define settings schema for `project_settings/fusion/masterprefs` as multiline text field, I would rather recomand to have it divided into particular elements (e.g. like list imput for multi path inputs, or bool for `Locked = true`). So it would be something more user friendly to configure. Rest of the logic for the building up the pref file in correct syntax could be hidden in the prelaunch hook.

openpype/hosts/fusion/hooks/pre_fusion_profile_hook.py
EmberLightVFX: Would it be possible to overwrite the file itself on each launch? That way we'll only have one file lingering in the temp folder.

iLLiCiTiT: Is possible to have running more than one process of fusion? If yes, than that is not the way...

BigRoy: > Would it be possible to overwrite the file itself on each launch? That way we'll only have one file lingering in the temp folder.

&

> Is possible to have running more than one process of fusion? If yes, then that is not the way...

Yes, you can run multiple instances of Fusion. I agree with @iLLiCiTiT that thus writing into a single file thus is not the way to go about this.

mkolar (approved):
Gentlemen, I reckon let's get this bad boy in, so it can be used, tested in real world and we'll see if any of the alternative approaches might be needed in further PRs.

BigRoy:
🚨  I've marked it as draft again as I'd like to hold off merging this PR due to:

- I wanted to still check whether I should make it a list of pref entries instead of a single one since technically Fusion allows to have multiple prefs set/defined. 🙂 But I haven't looked into that more. A Downside with merging now would mean changing the setting later would make it backwards incompatible. As such, I would first like to implement the list behavior in settings and test it out.

mkolar:
Considering the ongoing discussion, that eventually turned this a bit stale and because we're splitting OpenPype into ayon-core and individual host addons, this PR would have to be re-created to target one of those. The PR has merit, the solution, not so much. Let's try to revisit in AYON

We're closing it down, but we'll he happy for a new PR to [ynput/ayon-fusion](https://github.com/ynput/ayon-fusion) repository once it's up.

PR #4657, opened 2023-03-19
Deadline: Remove unused FramesPerTask job info submission

BigRoy:
## Changelog Description

Remove unused `FramesPerTask` job info submission to Deadline.

## Additional info

Originally discussed on Discord [here](https://discord.com/channels/517362899170230292/1086736440777511065).

The intended attribute should be `ChunkSize` which is already passed along by the submissions.

**Note:** In Maya `ChunkSize` was set through the render globals so I've removed the `instance.data("chunkSize")` entry since it looked like it would use that but it wouldn't because afterwards it would [always be overridden with the render globals](https://github.com/ynput/OpenPype/blob/0cd7ddcef423003c1e40f29a88e6943c61ef7cdf/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py#L154-L156) which [were collected here](https://github.com/ynput/OpenPype/blob/f22ece73576aff4ee5828aa9aa1cba673953937e/openpype/hosts/maya/plugins/publish/collect_render.py#L412-L413).

## Testing notes:

1. Submit renders from Maya - specifically with 'framesPerTask' attribute tweaked (that's the attribute name on renderglobals node in Maya
2. Submit renders from Max - I don't believe Max actually supported setting the frames per task but I'm not sure; if it does, then test that.


fabiaserra:
> 

And do we prefer the chunkSize being set on the render globals rather than the instance?

BigRoy:
> And do we prefer the chunkSize being set on the render globals rather than the instance?

1. Do you mean from a coding perspective? 
2. Or that it's a setting for the whole scene instead of per RenderLayer?

If the first, I'd say not necessarily and we should clean that up. But #4388 is doing a lot of changes in that area and that's still quite a quick refactor from my end. Would like to do cleanup in that area so definitely please review/comment on that PR if you have the time.

If the second, having it per RenderLayer allows more granular customization but the downside that users will need to tweak more values. Anyway, #4388 actually splits all of the render settings for the publisher to the RenderLayer instances. With that PR you can set a different pool per layer if you need it and a different chunk size per layer too.

openpype/modules/deadline/plugins/publish/submit_max_deadline.py
```diff
@@ -74,7 +74,6 @@ def get_job_info(self):
         job_info.ChunkSize = instance.data.get("chunkSize", 1)
```
moonyuet, line 74:
```suggestion
        job_info.ChunkSize = instance.data.get("chunkSize", self.chunk_size)
```
it would take the OP setting of chunksize if I dont get it wrong.
The code below definitely works with the OP setting(and more precise).
```suggestion
        job_info.ChunkSize = self.chunk_size 
```

fabiaserra:
> > And do we prefer the chunkSize being set on the render globals rather than the instance?
> 
> 1. Do you mean from a coding perspective?
> 2. Or that it's a setting for the whole scene instead of per RenderLayer?
> 
> If the first, I'd say not necessarily and we should clean that up. But #4388 is doing a lot of changes in that area and that's still quite a quick refactor from my end. Would like to do cleanup in that area so definitely please review/comment on that PR if you have the time.
> 
> If the second, having it per RenderLayer allows more granular customization but the downside that users will need to tweak more values. Anyway, #4388 actually splits all of the render settings for the publisher to the RenderLayer instances. With that PR you can set a different pool per layer if you need it and a different chunk size per layer too.

Makes sense, thanks for pointing me to that PR. I wanted to make sure that we weren't solving this bug by regressing on how the chunk size was intended to be used. Thank you!

moonyuet (approved):
the code looks good for me but i haven't tested the code yet

tokejepsen (dismissed):

Will leave Max to someone else.

LiborBatek:
Im testing in 3dsmax because actually there is a ```chunk size``` attribute to fiddle with....however in this PR there are missing option for ```Render``` Instance in Publisher... see img

image

and here is the ```develop``` version in max with ```Frame Per Task``` attribute present:

image

So not really sure about this PR testing. @BigRoy should I wait?

BigRoy:
@LiborBatek please try again - the PR was just not up-to-date with latest develop, should be now.

BigRoy:
However I just saw this PR https://github.com/ynput/OpenPype/pull/4794 so it might be that 3Ds Max render publishing is broken in latest develop anyway. @moonyuet Right?

moonyuet:
> However I just saw this PR #4794 so it might be that 3Ds Max render publishing is broken in latest develop anyway. @moonyuet Right?

3dsmax rendering is fine while it fails in submitting publish job. This can be mainly because the 3dsMax collector doesn't have instance.data["farm"] introduced in the collector. 
The fix should be in the collector only so it won't do conlifct with this PR

LiborBatek:
@BigRoy yeah, I think I can test it now....just testing if the job will be set correctly rendering in frame chunks accordingly even that DL issue for 3dsmax being present as @moonyuet  mentioned its not interacting each other

LiborBatek (approved):
```Frame per Task``` is functional and correctly working.

image

And checking on DL matches the setting:
image

tokejepsen (approved):

BigRoy:
afbeelding

Hehe. That's confusing. :)

PR #4658, opened 2023-03-20
3dsmax: render instance settings in creator

moonyuet:
## Changelog Description
Allows user preset the pools, group and use_published settings in Render Creator in the Max Hosts.
User can set the settings before or after creating instance in the new publisher

image

image

## Testing notes:
If creating new render instance...
1. Choose Render from choose publish type
2.  Fill in the settings
3. Create the instance
4. Publish it

If there is render instance.....
1. Click Publish in the new Publisher
2. Click maxrenderMain
3. Edit the settings
5. Publish it

### DIFF
--- a/openpype/hosts/max/plugins/create/create_render.py
+++ b/openpype/hosts/max/plugins/create/create_render.py
@@ -1,8 +1,22 @@
 # -*- coding: utf-8 -*-
 """Creator plugin for creating camera."""
 from openpype.hosts.max.api import plugin
+from openpype.lib import (
+    TextDef,
+    BoolDef,
+    NumberDef,
+)
 from openpype.pipeline import CreatedInstance
 from openpype.hosts.max.api.lib_rendersettings import RenderSettings
+from openpype.settings import get_project_settings
+from openpype.pipeline import legacy_io
+
+
+def setting(project_setting=None):
+    render_setting = get_project_settings(
+        legacy_io.Session["AVALON_PROJECT"]
+    )
+    return render_setting["deadline"]["publish"]["MaxSubmitDeadline"]
 
 
 class CreateRender(plugin.MaxCreator):
@@ -31,3 +45,39 @@ class CreateRender(plugin.MaxCreator):
         RenderSettings().set_render_camera(sel_obj)
         # set output paths for rendering(mandatory for deadline)
         RenderSettings().render_output(container_name)
+
+    def get_instance_attr_defs(self):
+        return [
+            BoolDef("use_published",
+                    default=setting()["active"],
+                    label="Use Published Scene"),
+
+            NumberDef("priority",
+                      minimum=1,
+                      maximum=250,
+                      decimals=0,
+                      default=setting()["priority"],
+                      label="Priority"),
+
+            NumberDef("chunkSize",
+                      minimum=1,
+                      maximum=50,
+                      decimals=0,
+                      default=setting()["chunk_size"],
+                      label="Chunk Size"),
+
+            TextDef("group",
+                    default=setting()["group"],
+                    label="Group Name"),
+
+            TextDef("deadline_pool",
+                    default=setting()["deadline_pool"],
+                    label="Deadline Pool"),
+
+            TextDef("deadline_pool_secondary",
+                    default=setting()["deadline_pool_secondary"],
+                    label="Deadline Pool Secondary")
+        ]
+
+    def get_pre_create_attr_defs(self):
+        return self.get_instance_attr_defs()
--- a/openpype/modules/deadline/plugins/publish/submit_max_deadline.py
+++ b/openpype/modules/deadline/plugins/publish/submit_max_deadline.py
@@ -21,7 +21,7 @@ class MaxPluginInfo(object):
     SaveFile = attr.ib(default=True)
     IgnoreInputs = attr.ib(default=True)
 
-
+#TODO: add the optional attirbute
 class MaxSubmitDeadline(abstract_submit_deadline.AbstractSubmitDeadline):
 
     label = "Submit Render to Deadline"
@@ -49,11 +49,13 @@ class MaxSubmitDeadline(abstract_submit_deadline.AbstractSubmitDeadline):
 
         instance = self._instance
         context = instance.context
-
         # Always use the original work file name for the Job name even when
         # rendering is done from the published Work File. The original work
         # file name is clearer because it can also have subversion strings,
         # etc. which are stripped for the published file.
+        if not instance.data.get("publish"):
+            self.use_published = False
+
         src_filepath = context.data["currentFile"]
         src_filename = os.path.basename(src_filepath)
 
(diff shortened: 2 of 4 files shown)


openpype/hosts/max/plugins/create/create_render.py
```diff
@@ -31,3 +45,39 @@ def create(self, subset_name, instance_data, pre_create_data):
...
+                    label="Group Name"),
+
+            TextDef("deadline_pool",
+                    default=setting()["deadline_pool"],
+                    label="Deadline Pool"),
+
+            TextDef("deadline_pool_secondary",
+                    default=setting()["deadline_pool_secondary"],
+                    label="Deadline Pool Secondary")
+        ]
```
BigRoy, line 80: This should **really** be somewhere in the Deadline module and not on Create Render.

openpype/hosts/max/plugins/create/create_render.py
```diff
@@ -1,8 +1,22 @@
...
 from openpype.hosts.max.api.lib_rendersettings import RenderSettings
+from openpype.settings import get_project_settings
+from openpype.pipeline import legacy_io
+
+
+def setting(project_setting=None):
+    render_setting = get_project_settings(
+        legacy_io.Session["AVALON_PROJECT"]
+    )
+    return render_setting["deadline"]["publish"]["MaxSubmitDeadline"]
```
BigRoy, line 19: Ouch, do not doing this! :) You're requerying the database each time per attribute definition.

You should instead be setting the defaults using [`apply_settings` on the Creator like here](https://github.com/ynput/OpenPype/blob/173b404d0e2af1712f9e7cf4d9f60ac5ea030570/openpype/hosts/tvpaint/plugins/create/create_review.py#L15-L21)

openpype/modules/deadline/plugins/publish/submit_max_deadline.py
```diff
@@ -49,11 +49,13 @@ def get_job_info(self):
 
         instance = self._instance
         context = instance.context
-
         # Always use the original work file name for the Job name even when
         # rendering is done from the published Work File. The original work
         # file name is clearer because it can also have subversion strings,
         # etc. which are stripped for the published file.
+        if not instance.data.get("publish"):
+            self.use_published = False
```
BigRoy, line 64: This doesn't sound right. If the instance itself has `publish` set to something that equates to False than the instance shouldn't get processed to begin with.

openpype/modules/deadline/plugins/publish/submit_max_deadline.py
```diff
@@ -71,13 +73,10 @@ def get_job_info(self):
 
         job_info.Pool = instance.data.get("primaryPool")
         job_info.SecondaryPool = instance.data.get("secondaryPool")
-        job_info.ChunkSize = instance.data.get("chunkSize", 1)
+        job_info.ChunkSize = instance.data.get("chunkSize", self.chunk_size)
```
BigRoy, line 83: Since we're getting into attribute definitions now coupling this very tightly to how an instance is created/collected I feel like should we be getting pretty certain that `chunkSize` will be defined on the instance.

I wonder whether we should be getting into the habit of not building fallback value on fallback value. The Collect Render plug-in is currently always setting the chunk size. The logic here now falling back to `self.chunk_size` and maybe even doing anything with regards to that might just end up being more confusing to trace back where the value came from.

At the very best case - if the value itself would not need explicit validation or alike then the attribute definition could just be directly on this Submission plug-in and the value get used one to one without needing to trace back to the Creator, other Collectors, etc. The value would then only really be useful for this particular plug-in. Which I think would be the best case because then it also only exists in the Deadline module for which it is intended. For example look at some of the 'changes' [here](https://github.com/ynput/OpenPype/pull/4388/files#diff-148e36a4d12835986c861be64a4320853ba6f30a15b9cd2f91f49e00b301362e) to Submit Maya Deadline.

I've mentioned it before but I'd actually even like to argue that MOST of the attribute definitions (and its usage) should be directly in Abstract Submit Deadline so that not every host needs to 'reimplement' this logic which is basically doing the same thing.

openpype/hosts/max/plugins/create/create_render.py
moonyuet: Yes i know, but this is client's request so I need to put it there

LiborBatek (approved):
The ```Render``` instance get succesfully created and with settings for user.

image


Correctly published to DL also.

image

openpype/modules/deadline/plugins/publish/submit_max_deadline.py
```diff
@@ -216,3 +249,40 @@ def _clean_name(path):
...
+                    default=cls.group,
+                    label="Group Name"),
+
+            TextDef("deadline_pool",
+                    default=cls.deadline_pool,
+                    label="Deadline Pool"),
+
+            TextDef("deadline_pool_secondary",
+                    default=cls.deadline_pool_secondary,
+                    label="Deadline Pool Secondary")
```
BigRoy, line 285: I would avoid adding a `deadline` prefix. The attributes are stored as Plug-in Attributes on the instance per plug-in. So the attribute on the instance would already be e.g. `publish_attributes.MaxSubmitDeadline.pool` as far as I know.

LiborBatek:
Did last test and all seems normal and working ok.

image

openpype/modules/deadline/plugins/publish/submit_max_deadline.py
```diff
@@ -32,13 +41,24 @@ class MaxSubmitDeadline(abstract_submit_deadline.AbstractSubmitDeadline):
...
     tile_priority = 50
-    chunk_size = 1
+    chunkSize = 1
     jobInfo = {}
     pluginInfo = {}
     group = None
-    deadline_pool = None
-    deadline_pool_secondary = None
     framePerTask = 1
+    optional = True
```
BigRoy, line 49: Why is this optional?

openpype/modules/deadline/plugins/publish/submit_max_deadline.py
```diff
@@ -32,13 +41,24 @@ class MaxSubmitDeadline(abstract_submit_deadline.AbstractSubmitDeadline):
     use_published = True
     priority = 50
     tile_priority = 50
-    chunk_size = 1
+    chunkSize = 1
```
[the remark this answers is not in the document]
moonyuet: i dont think there is any tile_priority in max deadline, so I will also remove it in the next PR. Thanks for the catch!

BigRoy (commented):
I just wanted to say, nice work on getting the settings and attributes moved into the Deadline module. 💪 

Before we can merge this I suspect we'll need to test it in any application which uses the deadline module's Collect Deadline Pools collector, which is likely all applications that have any of these families:

```
"rendering",
                "render.farm",
                "renderFarm",
                "renderlayer",
                "maxrender"
```

Because now this collector will behave differently across all hosts. Nuke and After Effects come to mind, but also Maya which is still using the old publisher and I'm not sure how well that behaves with the `get_attr_values_from_data` logic, etc.

BigRoy:
@kalisp @jakubjezek001 Any chance you could test this PR with Nuke and After Effects to see if it affects anything from those hosts using new publisher that involve submissions to Deadline?

You'll need Deadline module enabled.

The issues I'm potentially expecting is:
- Multiple "pool" settings now appearing on the render instance
- The default pools of that particular host might not get set from settings
- It might be unclear which pool actually gets set for the submission to Deadline.

This is assuming After Effects has deadline submission logic.

antirotor (approved):
I've tested it in Nuke 11.3 on Deadline, seems it is working.

kalisp:
This will not work in AE as creator now doesn't provide `render.farm` family for now (it is actually set only in `collect_render`).
I can refactor AE to set `render.farm` directly in creator (in different PR though).

PR #4662, opened 2023-03-20
Integrator: Enforce unique destination transfers, disallow overwrites in queued transfers

### LINKED ISSUE #4656: IntegrateAsset: Allows overwriting representations.
**Describe the bug**
When having two representations with the same resulting file name, the plugin does not fail.

**To Reproduce**
1. Setup the below extractor in the pipeline;
```python
from maya import cmds

class ExtractExample(publish.Extractor):
    """Extract the content of the instance to an CPU cache file."""

    label = "ExtractExample"
    hosts = ["maya"]
    families = ["model"]

    def process(self, instance):
        staging_dir = self.staging_dir(instance)
        filename = "{}_gpu_cache".format(instance.name)

        # Write out GPU cache file.
        kwargs = {
            "directory": staging_dir,
            "fileName": filename,
            "saveMultipleFiles": False,
        }
        cmds.gpuCache(instance[:], **kwargs)

        representation = {
            "name": "gpu_cache",
            "ext": "abc",
            "files": filename + ".abc",
            "stagingDir": staging_dir
        }

        self.log.info(
            "Extracted instance {} to: {}".format(instance.name, staging_dir)
        )
```
2. Setup model instance in Maya and publish.
3. Check the output representations, which should be; `.ma` and `.abc`. (Missing an additional `.abc`)

**Expected behavior**
The `IntegrateAsset` plugin should fail or at least warn developers about conflicting file names.

[cuID:[OP-5282](https://app.clickup.com/t/85zrt8xjq)]

BigRoy:
## Changelog Description

Fix #4656 by enforcing unique destination transfers in the Integrator. It's now disallowed to a destination in the file transaction queue with a new source path during the publish.

## Additional info

Note that this is per instance - it doesn't validate cross-instance destinations in the context

## Testing notes:

1. Try integrating e.g. the example in #4656

### DIFF
--- a/openpype/lib/file_transaction.py
+++ b/openpype/lib/file_transaction.py
@@ -44,7 +44,7 @@ class FileTransaction(object):
     MODE_COPY = 0
     MODE_HARDLINK = 1
 
-    def __init__(self, log=None):
+    def __init__(self, log=None, allow_queue_replacements=False):
         if log is None:
             log = logging.getLogger("FileTransaction")
 
@@ -59,6 +59,8 @@ class FileTransaction(object):
 
         # Backup file location mapping to original locations
         self._backup_to_original = {}
+
+        self._allow_queue_replacements = allow_queue_replacements
 
     def add(self, src, dst, mode=MODE_COPY):
         """Add a new file to transfer queue.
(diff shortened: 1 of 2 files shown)


tokejepsen (approved):

openpype/lib/file_transaction.py
```diff
@@ -82,6 +84,14 @@ def add(self, src, dst, mode=MODE_COPY):
                         src, dst))
                 return
             else:
+                if not self._allow_queue_replacements:
+                    raise RuntimeError(
```
iLLiCiTiT, line 88: I would use different exception e.g. `ValueError` or custom. And maybe catch it in integrate.py and reraise as `KnownPublishError`.

BigRoy: I'll go for `DuplicateDestinationError` because it's not necessary that it's the same file. Anyway, thanks!

PR #4664, opened 2023-03-21
Maya: Set correct framerange with handles on file opening

friquette:
## Changelog Description
Set the range of playback from the asset data, counting handles, to get the correct data when calling the "collect_animation_data" function.


tokejepsen:
How does the timeline end up looking?

BigRoy:
Just want to link this [thread of comments on another PR](https://github.com/ynput/OpenPype/pull/4527#discussion_r1118561671) which feel slightly related. 

The PR description sounds a bit vague but looking at the code it does this:

- On "reset_framerange" it now sets the furthest timeline handles of Maya to include the handles of the asset data - previously it would **include** the handles completely for both the animation range and playback range.
    - Reset frame range is also what's triggered on "new scene" but **NOT** on "file open" so new scenes would now also have the handles set as described above.
- On creating new instances the collected default values now include the range outside of your timeline as handles which brings me to the [linked comment thread](https://github.com/ynput/OpenPype/pull/4527#discussion_r1118561671) which briefly discussed whether that's something we want or not.

friquette:
> How does the timeline end up looking?

So the timeline is looking something like this: 
image

The timeline is set from 101 to 222, but with 5 handles in start and end so we can see the animation goes from 96 to 227.

openpype/hosts/maya/api/lib.py
```diff
@@ -2097,13 +2102,11 @@ def reset_frame_range():
     if handle_end is None:
         handle_end = handles
 
-    frame_start -= int(handle_start)
-    frame_end += int(handle_end)
+    animation_start = frame_start - int(handle_start)
+    animation_end = frame_end + int(handle_end)
```
moonyuet, line 2106: Similar to the line above. Isn't `frame_end += int(handle_end)` equal to `frame_end = frame_end + int(handle_end`)?

moonyuet:
If the handles really needed to be included in the setting after creating new scene every single time or the reset frame range can perform to reset frame range with handles, maybe the best way to do is to include boolean options to allow users to enable/disable to set the handles in the callback and reset the handles by triggering set frame range in Openpype dropdown menu.

tokejepsen:
@moonyuet good point!
@friquette lets get this behind a setting that is off by default because it changes current behaviour.

friquette:
@tokejepsen So I added an option in the maya project settings, where you can choose in which task types you want the handles

openpype/hosts/maya/api/lib.py
```diff
@@ -2097,15 +2106,25 @@ def reset_frame_range():
     if handle_end is None:
         handle_end = handles
 
-    frame_start -= int(handle_start)
-    frame_end += int(handle_end)
+    if task_name in with_handles:
```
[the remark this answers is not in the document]
moonyuet: I actually suggest boolean type to enable/disable the setting. It's not the task-types-enum type. 
I agree that it is getting much more confused for the way it does to the frameHandle(the initial purposes seems to include the frame handle in timeline while the function actually excludes the handles)... and we must deal with the setting more carefully if we want to continue to make it as a part of the setting.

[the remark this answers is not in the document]
tokejepsen: Yeah, I agree that the naming is a bit confusing atm. How about `Use handles on timeline`?

Also agree that it should be a bit easier to use for all task types. How about the setting is a boolean (off by default)?
When enabled the functionality would be enabled for all tasks on the project. Then have a `Limit to Task Types` option, if a studio wanted to limit to a handful of task types.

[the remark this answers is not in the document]
BigRoy: I think @friquette you wanted this enabled always and others mentioned they preferred the old behavior? If so I maybe even wouldn't make it task dependent to begin with and just a project-wide setting? We can always add task overrides if people need it and have a use case for it later? Or do we really think people will end up customizing this per task?

Also, use handles on timeline still sounds confusing. The old behavior did include the handles too. It just wouldn't exclude the handles in the active timeline play area.
According to this [Maya documentation on the Range slider](
https://help.autodesk.com/view/MAYAUL/2022/ENU/index.html?guid=GUID-E9E054A1-A5D0-412C-976A-F57C0F6960B1) we can use the terminology Playback range versus Animation range. As such I would say the toggle does: *"Exclude handles from playback range** or **Set playback range to exclude handles**.

OlivierOsotimehin:
Hello, i'm cgsup from Fixstudio requesting this. I wanted to explain a bit more the use case. 

We have a shot that have a duration (101-200) and  handles (5 start and 5 end)

We want to publish animation abc family  with the handle (so we have extra anim, and correct motionblur ). 
We want to review without  (but sometime with if we have to check in the handle )
We want to render without (because render can be very expensive, but in some rare case with) 

So this lead to wanting a "with handle", or "include handle" booleen type to set this for each family

BigRoy:
> Hello, i'm cgsup from Fixstudio requesting this. I wanted to explain a bit more the use case.
> 
> We have a shot that have a duration (101-200) and handles (5 start and 5 end)
> 
> - We want to publish animation abc family with the handle (so we have extra anim, and correct motionblur ). 
> - We want to review without (but sometime with if we have to check in the handle ) 
> - We want to render without (because render can be very expensive, but in some rare case with)
> 
> So this lead to wanting a "with handle", or "include handle" booleen type to be able to generate the extra data that is in the handle or not

Thanks for the solid explanation. I actually totally forgot that `reset_framerange` by default also sets the scene's render frame range and thus it'd set it **without handle** completely with the new code.

I guess we should call the option: **Include/Exclude Handles in default playback & render range**. 

@friquette does this mean that with the notes here you're able to adjust this PR?

openpype/hosts/maya/api/lib.py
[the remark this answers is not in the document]
tokejepsen: > Set playback range to exclude handles

+1

> We can always add task overrides if people need it and have a use case for it later? Or do we really think people will end up customizing this per task?

Yup, think that is fair. Boolean type for project wide toggle. If another studio needs more granular control, then it can be implemented later.

friquette:
> 

So to summarize, I have to call the option something like **Include handles in default playback & render range**
This option must be a boolean, and the task type is not needed for now. Do I understand correctly ?

BigRoy:
> I have to call the option something like Include handles in default playback & render range

I think that'd work, yes.

> the task type is not needed for now. Do I understand correctly ?

Since @OlivierOsotimehin mentioned they would need that feature in production currently - I'd say let's directly implement it. I suspect that when @tokejepsen wrote [this comment](https://github.com/ynput/OpenPype/pull/4664#discussion_r1145766807) he might have missed [the comment by Olivier](https://github.com/ynput/OpenPype/pull/4664#issuecomment-1480261669).

tokejepsen:
> Since @OlivierOsotimehin mentioned they would need that feature in production currently - I'd say let's directly implement it. I suspect that when @tokejepsen wrote https://github.com/ynput/OpenPype/pull/4664#discussion_r1145766807 he might have missed https://github.com/ynput/OpenPype/pull/4664#issuecomment-1480261669.

I read that as @OlivierOsotimehin didnt need the task types limiting setting, but I think its a good feature to have so lets implement it.

@BigRoy how does `Limit to Task Types` sound for a label?

BigRoy:
> how does Limit to Task Types sound for a label?

So there's going to be one global toggle defining the "default value" for your productions and the task types you enter override that. I think it'll just be much clearer if then each task type comes with a checkbox of its own whether it enables/disables just to avoid confusing with the fact that otherwise it'd be like an "invert" of the global toggle just adding to confusion.

```
- include_handles_default = True (if no task type match below)
- per_task_type:
    [0]
    - task_type "animation"
    - include_handles = True
    [1]
    - task_type "rendering"
    - include_handles = False
```

tokejepsen (changes requested):
Works well for when the global toggle is disabled, but think we are missing logic for when the global toggled is enabled but disabled for specific task types. Example:

Capture

Also could you ensure `"` are used instead of `'`?

openpype/hosts/maya/api/lib.py
```diff
@@ -2097,15 +2106,33 @@ def reset_frame_range():
...
+                cmds.warning(
+                    "There's no task type in the \
+'Include by task type' setting."
+                )
+                return
+
+            if current_task['type'] in item['task_type'] \
+                    and item['include_handles']:
+                animation_start = frame_start - int(handle_start)
+                animation_end = frame_end + int(handle_end)
```
BigRoy, line 2128: I still feel like we could just simplify this to this:

```python
    animation_start = frame_start
    animation_end = frame_end

    include_handles = include_handles_settings['include_handles_default']
    for item in include_handles_settings['per_task_type']:
        if current_task["type"] == item["task_type"]:
            include_handles = item["include_handles"]
            break

    if include_handles:
        animation_start -= int(handle_start)
        animation_end += int(handle_end)
```

BigRoy (commented):
I see now that I never actually posted my review. Sorry about that.

openpype/hosts/maya/api/lib.py
```diff
@@ -2097,15 +2106,25 @@ def reset_frame_range():
...
 
-    cmds.playbackOptions(minTime=frame_start)
-    cmds.playbackOptions(maxTime=frame_end)
-    cmds.playbackOptions(animationStartTime=frame_start)
-    cmds.playbackOptions(animationEndTime=frame_end)
-    cmds.playbackOptions(minTime=frame_start)
-    cmds.playbackOptions(maxTime=frame_end)
+    include_handles = include_handles_settings["include_handles_default"]
+    for item in include_handles_settings["per_task_type"]:
+        print("ITEM: {}".format(item))
```
BigRoy, line 2114:
```suggestion
```

tokejepsen (changes requested):
Works great! Tested in Maya 2023.3.

Could we get some documentation for this feature?

website/docs/admin_hosts_maya.md
```diff
@@ -107,6 +107,18 @@ or Deadlines **Draft Tile Assembler**.
...
+
+The "Include handles by default" toggle is here to select weither you want handles or not in all task type. You can also chose in what task type you want handles or not.
+
+For example, in this image you can see that handles are included by default in all task types, except for the "Lighting" task, where the toggle is disabled.
+![Include/exclude handles](assets/maya-admin_exclude_handles.png)
+
+And here you can see thant handles are disabled by default, except in Animation task where it's enabled.
+![Custom menu definition](assets/maya-admin_include_handles.png)
+
+
```
tokejepsen, line 121:
```suggestion
## Include/exclude handles by task type
You can include or exclude handles, globally or by task type.

The "Include handles by default" toggle is here to select whether you want handles or not in all task type. You can also choose in which task types you want handles included.

For example, in this image you can see that handles are included by default in all task types, except for the `Lighting` task, where the toggle is disabled.
![Include/exclude handles](assets/maya-admin_exclude_handles.png)

And here you can see than the handles are disabled by default, except in `Animation` task where it's enabled.
![Custom menu definition](assets/maya-admin_include_handles.png)
```

tokejepsen (changes requested):

website/docs/admin_hosts_maya.md
```diff
@@ -107,6 +107,18 @@ or Deadlines **Draft Tile Assembler**.
 This is useful to fix some specific renderer glitches and advanced hacking of Maya Scene files. `Patch name` is label for patch for easier orientation.
 `Patch regex` is regex used to find line in file, after `Patch line` string is inserted. Note that you need to add line ending.
 
+## Include/exclude handles by task type
+You can include or exclude handles globaly or by task type.
+
+The "Include handles by default" toggle is here to select weither you want handles or not in all task type. You can also chose in what task type you want handles or not.
```
BigRoy, line 113:
```suggestion
The "Include handles by default" defines whether by default handles are included or not for all task type. Additionally you can add a per task type override whether you want to include handles or not.
```

website/docs/admin_hosts_maya.md
```diff
@@ -107,6 +107,18 @@ or Deadlines **Draft Tile Assembler**.
 This is useful to fix some specific renderer glitches and advanced hacking of Maya Scene files. `Patch name` is label for patch for easier orientation.
 `Patch regex` is regex used to find line in file, after `Patch line` string is inserted. Note that you need to add line ending.
 
+## Include/exclude handles by task type
+You can include or exclude handles globaly or by task type.
```
BigRoy, line 111:
```suggestion
You can include or exclude handles globally or by task type.

This setting is used to define the default frame range for new scenes or when using the **OpenPype > Set frame range** setting.
```

This made me think - does this actually also influence the **Validate Frame Range** validations or whatever it's called where it validates during publish whether an instance matches the asset document's frame range + handles?

friquette: Yes it does influence the Validate Frame Range validator. And the "Repair" button when the validator fails repair correctly the framerange and the handles in the failed instances.

website/docs/admin_hosts_maya.md
```diff
@@ -107,6 +107,18 @@ or Deadlines **Draft Tile Assembler**.
 This is useful to fix some specific renderer glitches and advanced hacking of Maya Scene files. `Patch name` is label for patch for easier orientation.
 `Patch regex` is regex used to find line in file, after `Patch line` string is inserted. Note that you need to add line ending.
 
+## Include/exclude handles by task type
+You can include or exclude handles, globally or by task type.
+
+The "Include handles by default" defines wether by default handles are included or not for all task type. Additionally you can add a per task type override wether you want to include handles or not.
```
tokejepsen, line 113:
```suggestion
The "Include handles by default" defines whether by default handles are included. Additionally you can add a per task type override whether you want to include or exclude handles.
```

tokejepsen (changes requested):

tokejepsen (dismissed):
Nice one @friquette !

mkolar:
@friquette this is ready for merge, could you have a look at the conflicts please?

BigRoy (approved):
Resolved merge conflicts seem fine to me.

tokejepsen (approved):
Amazing work @friquette !

openpype/hosts/maya/api/lib.py
```diff
@@ -2091,44 +2100,25 @@ def get_frame_range():
...
 
-    return {
-        "frameStart": frame_start,
-        "frameEnd": frame_end,
-        "handleStart": handle_start,
-        "handleEnd": handle_end
-    }
-
-
-def reset_frame_range(playback=True, render=True, fps=True):
```
antirotor, line 2102: This is still used on two places - without it Maya integration won't even start and reset frame range menu won't work.

**lib_rendersettings.py**
```
17: from openpype.hosts.maya.api.lib import reset_frame_range
161 reset_frame_range(playback=False, fps=False, render=True)
```

**menu.py**
```
command=lambda *args: lib.reset_frame_range()
```

openpype/hosts/maya/api/lib.py
```diff
@@ -2068,8 +2073,12 @@ def get_frame_range():
 
     # Set frame start/end
     project_name = legacy_io.active_project()
+    task_name = legacy_io.Session["AVALON_TASK"]
     asset_name = legacy_io.Session["AVALON_ASSET"]
```
antirotor, line 2077: Please be aware that we are trying to move from the use of `legacy_io`. This was already changed in `release/next-minor` branch like this:

```suggestion
     # Set frame start/end
    project_name = get_current_project_name()
    task_name = get_current_task_name()
    asset_name = get_current_asset_name()
```

antirotor (changes requested):
Deleted `reset_frame_range()` is breaking Maya integration now.

openpype/hosts/maya/api/lib.py
BigRoy: Thank you so much! Totally missed the fact that it somehow removed the function due to how it showed the changes on Github. Agreed, this should definitely be reverted.

tokejepsen:
@friquette what is the state of this PR? Ready to review again?

tokejepsen (changes requested):
Could you clean this PR up a bit?
There is a bunch of file changes unrelated to this PR.

friquette:
> Could you clean this PR up a bit? There is a bunch of file changes unrelated to this PR.

Sorry for that, I did a rebase on the wrong branch, it should be ok now :)

tokejepsen (approved):

openpype/hosts/maya/api/lib.py
```diff
@@ -2148,6 +2162,26 @@ def get_frame_range():
...
+        animation_start -= int(handle_start)
+        animation_end += int(handle_end)
+
+    cmds.playbackOptions(
+        minTime=frame_start,
+        maxTime=frame_end,
+        animationStartTime=animation_start,
+        animationEndTime=animation_end
+    )
+    cmds.currentTime(frame_start)
```
BigRoy, line 2183: **🚨  Oops!**

Ok. So this PR implemented this logic here - but that's wrong. This shouldn't be in `get_frame_range()` but in `reset_frame_range()`.

The `get_frame_range()` function should not be adjusting the scene state.

---

@friquette @tokejepsen

mkolar: Yeah needs to be fixed asap. Good that we had 3 reviews on this :D right?

PR #4652, opened 2023-03-17
Resolution settings referenced from DB record for 3dsMax

moonyuet:
## Changelog Description
1. Add Callback for setting the resolution according to DB after the new scene is created.
2. Add a new Action into openpype menu which allows the user to reset the resolution in 3dsMax

image

## Additional info
N/A

## Testing notes:
1. launch 3dsMax via Openpype
2. check the resolution in render setting
4. it should be set according to DB.
Reset Resolution
1. locate "Set Resolution" in openpype menu
2. click it
3. check the resolution in render setting to see if it is same with the setting shown in DB

PR #4665, opened 2023-03-21
scene length setting referenced from DB record for 3dsMax

moonyuet:
## Changelog Description
 Setting the timeline length based on DB record in 3dsMax Hosts
image

## Additional info
N/A

## Testing notes:
1. launch 3dsMax via Openpype
2. create new scene
3. check the timeline
4. locate "Set Frame Range" in openpype menu
5. click it
6. check the frame range or the timeline to see if it is same with the setting shown in DB


BigRoy:
This PR seems to also include #4652 - is that correct?

moonyuet:
> This PR seems to also include #4652 - is that correct?

yes, but they are in a separate ticket, I will wait the old one to merge and update this ticket.  I convert this to draft first

LiborBatek (approved):
All works as expected! Did several tests in multiple shots and all being ok.

image

openpype/hosts/max/api/lib.py
```diff
@@ -157,6 +165,125 @@ def get_multipass_setting(project_setting=None):
...
+    handle_end = asset["data"].get("handleEnd")
+    if handle_end is None:
+        handle_end = handles
+
+    return {
+        "frameStart": frame_start,
+        "frameEnd": frame_end,
+        "handleStart": handle_start,
+        "handleEnd": handle_end
+    }
```
antirotor, line 240: We should stop using `legacy_io`:

```suggestion
def get_frame_range() -> Dict:
    """Get the current assets frame range and handles.
    
    Returns:
    	Dict: with frame start, frame end, handle start, handle end.
    
    """
    # Set frame start/end
    asset = get_current_project_asset()

    frame_start = asset["data"].get("frameStart")
    frame_end = asset["data"].get("frameEnd")
    # Backwards compatibility
    if frame_start is None or frame_end is None:
        frame_start = asset["data"].get("edit_in")
        frame_end = asset["data"].get("edit_out")

    if frame_start is None or frame_end is None:
        return

    handles = asset["data"].get("handles") or 0
    handle_start = asset["data"].get("handleStart")
    if handle_start is None:
        handle_start = handles

    handle_end = asset["data"].get("handleEnd")
    if handle_end is None:
        handle_end = handles

    return {
        "frameStart": frame_start,
        "frameEnd": frame_end,
        "handleStart": handle_start,
        "handleEnd": handle_end
    }
```

openpype/hosts/max/api/lib.py
```diff
@@ -157,6 +165,125 @@ def get_multipass_setting(project_setting=None):
...
+        rt.frameRate = fps_number
+
+    frame_range = get_frame_range()
+
+    frame_start = frame_range["frameStart"] - int(frame_range["handleStart"])
+    frame_end = frame_range["frameEnd"] + int(frame_range["handleEnd"])
+
+    frange_cmd = f"animationRange = interval {frame_start} {frame_end}"
+
+    rt.execute(frange_cmd)
```
antirotor, line 266:
```suggestion
def reset_frame_range(fps: bool=True):
    """Set frame range to current asset.
    
    	This is part of 3dsmax documentation:

        animationRange: A System Global variable which lets you get and
        	set an Interval value that defines the start and end frames
        	of the Active Time Segment.
        frameRate: A System Global variable which lets you get
        	and set an Integer value that defines the current
        	scene frame rate in frames-per-second.
    """
    if fps:
        fps_number = float(get_current_project(fields=["data.fps"])["data"]["fps"])
        rt.frameRate = fps_number

    frame_range = get_frame_range()

    frame_start = frame_range["frameStart"] - int(frame_range["handleStart"])
    frame_end = frame_range["frameEnd"] + int(frame_range["handleEnd"])

    frange_cmd = f"animationRange = interval {frame_start} {frame_end}"

    rt.execute(frange_cmd)
```

openpype/hosts/max/api/lib.py
```diff
@@ -157,6 +165,125 @@ def get_multipass_setting(project_setting=None):
...
+
+    Returns:
+        None
+
+    """
+    rt.renderWidth = width
+    rt.renderHeight = height
+
+
+def reset_scene_resolution():
```
antirotor, line 183: Please see notes on #4652

openpype/hosts/max/api/lib.py
```diff
@@ -157,6 +165,125 @@ def get_multipass_setting(project_setting=None):
                            ["multipass"])
 
 
+def set_scene_resolution(width, height):
```
antirotor, line 168:
```suggestion
def set_scene_resolution(width: int, height: int):
```

antirotor (changes requested):
`legacy_io` deprecation notes (similar to #4652)

antirotor (approved):

PR #4667, opened 2023-03-21
General: Filter available applications

iLLiCiTiT:
## Changelog Description
Added option to filter applications that don't have valid executable available in settings in launcher and ftrack actions. This option can be disabled in new settings category `Applications`. The filtering is by default disabled.

## Testing notes:
1. Enable `project_settings/applications/only_available`
2. Make sure there is application which have set paths that don't lead to executable available on your machine
3. Add the application to a project
4. Open launcher and go to a task in project / Open actions on task in ftrack
5. The application should not be there

### DIFF
--- a/openpype/modules/ftrack/event_handlers_user/action_applications.py
+++ b/openpype/modules/ftrack/event_handlers_user/action_applications.py
@@ -124,6 +124,11 @@ class AppplicationsAction(BaseAction):
         if not avalon_project_apps:
             return False
 
+        settings = self.get_project_settings_from_event(
+            event, avalon_project_doc["name"])
+
+        only_available = settings["applications"]["only_available"]
+
         items = []
         for app_name in avalon_project_apps:
             app = self.application_manager.applications.get(app_name)
@@ -131,6 +136,10 @@ class AppplicationsAction(BaseAction):
                 continue
 
             if app.group.name in CUSTOM_LAUNCH_APP_GROUPS:
+                continue
+
+            # Skip applications without valid executables
+            if only_available and not app.find_executable():
                 continue
 
             app_icon = app.icon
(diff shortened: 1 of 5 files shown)


BigRoy:
Should this feature debug log what it filtered?

iLLiCiTiT:
> Should this feature debug log what it filtered?

Launcher actions are re-discovered each 10 seconds and ftrack actions discover is triggered on each click on `Actions`, so I'm not sure if it's a good idea. Maybe one log with all skipped app names?

BigRoy:
> Launcher actions are re-discovered each 10 seconds and ftrack actions discover is triggered on each click on Actions, so I'm not sure if it's a good idea. Maybe one log with all skipped app names?

Ah, I was thinking of logging it maybe when entering a project in the launcher only. I'm just wondering how as a dev we could best debug that it's just filtered due to an exe not found. I think this will get confusing quickly if there's no trace whatsoever.

iLLiCiTiT:
> I personally liked the old big ass pop-up saying "can't launch because it's not installed" :)

Me too, this is client request. It is possible to change it back to not filter them by default, so studio can enable the filtering if needed.

BigRoy:
> Me too, this is client request. It is possible to change it back to not filter them by default, so studio can enable the filtering if needed.

Let's disable it by default then. It'll be much clearer with the big pop-up to new users starting OpenPype for the first time too.

I still believe that it should leave a trace of what's going on when filtering is enabled, but it shouldn't get in the way of just having optimized code so I'll leave that comment here just as a note not as requirement.

iLLiCiTiT:
> Potentially the app executable not found pop-up could tell about the filtering possibility in settings too?

The popup is shown to artists. I would say the information about settings is not relevant for them.

64qam (approved):

PR #4670, opened 2023-03-21
General: Environment variable for default OCIO configs

iLLiCiTiT:
## Changelog Description
Define environment variable which lead to root of builtin ocio configs to be able change the root without changing settings. For the path in settings was used `"{OPENPYPE_ROOT}/vendor/bin/ocioconfig/OpenColorIOConfig"` which disallow to change the root somewhere else. That will be needed in AYON where configs won't be part of desktop application but downloaded from server.

## Testing notes:
1. Extraction with OCIO configs should work as did

### DIFF
--- a/openpype/settings/defaults/project_settings/global.json
+++ b/openpype/settings/defaults/project_settings/global.json
@@ -2,8 +2,8 @@
     "imageio": {
         "ocio_config": {
             "filepath": [
-                "{OPENPYPE_ROOT}/vendor/bin/ocioconfig/OpenColorIOConfigs/aces_1.2/config.ocio",
-                "{OPENPYPE_ROOT}/vendor/bin/ocioconfig/OpenColorIOConfigs/nuke-default/config.ocio"
+                "{BUILTIN_OCIO_ROOT}/aces_1.2/config.ocio",
+                "{BUILTIN_OCIO_ROOT}/nuke-default/config.ocio"
             ]
         },
         "file_rules": {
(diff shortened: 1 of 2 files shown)


iLLiCiTiT:
> Shouldn't the variable still contain OPENPYPE_ prefix? :)

I wanted to avoid it :) It's more preparation for AYON then need for openpype...

BigRoy:
Also, I'm not sure this should really be in `start.py` - feels a bit dirty. :) 
Additionally this still seems to basically be setting the env var using the OPENPYPE_ROOT - so how much does this actually solve with regards to forward compatibility?

iLLiCiTiT:
> Also, I'm not sure this should really be in start.py - feels a bit dirty. :)

There is not better place where it could be set.

> Additionally this still seems to basically be setting the env var using the OPENPYPE_ROOT - so how much does this actually solve with regards to forward compatibility?

Issue with this is that the path definition is in settings. If in settings is used `{BUILTIN_OCIO_ROOT}/aces_1.2/config.ocio` instead of `{OPENPYPE_ROOT}/vendor/bin/ocioconfig/OpenColorIOConfigs/aces_1.2/config.ocio`. We can just change value of `BUILTIN_OCIO_ROOT` pointing to different path and job done.

> By the way - we're still relying on [OPENPYPE_ROOT here too](https://github.com/ynput/OpenPype/blob/f96670d9dfa482174584b85f93f386aff472e841/openpype/lib/vendor_bin_utils.py#L137-L158) which is used to define the path for ffmpeg and ocio vendorized tools.

One by one. Also that logic can be changed with added condition.

PR #4671, opened 2023-03-21
Maya: Remove unused functions from Extract Look

BigRoy:
## Changelog Description

Remove unused functions from Maya Extract Look plug-in

## Testing notes:

1. Publish a look from maya


moonyuet (approved):
checked the code, get_ocio_config_path and escape_space both fucntions don't use within the code.
checked in Maya, it doesn't break the extractor
image

PR #4672, opened 2023-03-21
Publisher: Windows reduce command window pop-ups during Publishing

BigRoy:
## Changelog Description

Reduce the command line pop-ups that show on Windows during publishing.

## Additional info

This PR can be considered a continuation of PR #4481 
Also mentioned on discord [here](https://discord.com/channels/517362899170230292/517382145552154634/1087768816529985576)

Examples of the pop-ups can be found in the comments [here](https://github.com/ynput/OpenPype/pull/4481#issuecomment-1478126460):

https://user-images.githubusercontent.com/49758407/226671061-7a5b578d-5975-4344-827f-2f993bcf3337.mp4

#### The above video is from BEFORE this PR

---

The pop-ups:

- [x] This PR removes the pop-up shown when opening the Publisher first time
- [x] This PR removes one of the pop-up windows shown during extracting/integrating reviews.
- [x] [This PR fixes this Extract Review call window pop-up](https://github.com/ynput/OpenPype/blob/6754e97490c3462d4fbd27400bdcd9d24c20068d/openpype/plugins/publish/extract_review.py#L392)
- [x] [This PR fixes this Extract Burnin call window pop-up](https://github.com/ynput/OpenPype/blob/c8c31018d656ea206c42e4a083365e21b4252861/openpype/plugins/publish/extract_burnin.py#L344)

I'm not entirely sure how to resolve the last two. It shows both when running from source and also when running from a `openpype_gui.exe` in a build.

## Testing notes:

1. Publish away with reviews and the (new?) publisher

### DIFF
--- a/openpype/lib/execute.py
+++ b/openpype/lib/execute.py
@@ -105,7 +105,7 @@ def run_subprocess(*args, **kwargs):
         # shell=True already tries to hide the console window
         # and passing these creationflags then shows the window again
         # so we avoid it for shell=True cases
-        and not kwargs.get("shell", False)
+        and kwargs.get("shell") is not True
     ):
         kwargs["creationflags"] = (
             subprocess.CREATE_NEW_PROCESS_GROUP
--- a/openpype/lib/execute.py
+++ b/openpype/lib/execute.py
@@ -102,6 +102,10 @@ def run_subprocess(*args, **kwargs):
     if (
         platform.system().lower() == "windows"
         and "creationflags" not in kwargs
+        # shell=True already tries to hide the console window
+        # and passing these creationflags then shows the window again
+        # so we avoid it for shell=True cases
+        and not kwargs.get("shell", False)
     ):
         kwargs["creationflags"] = (
             subprocess.CREATE_NEW_PROCESS_GROUP
--- a/openpype/lib/vendor_bin_utils.py
+++ b/openpype/lib/vendor_bin_utils.py
@@ -224,18 +224,26 @@ def find_tool_in_custom_paths(paths, tool, validation_func=None):
 
 def _check_args_returncode(args):
     try:
-        # Python 2 compatibility where DEVNULL is not available
+        kwargs = {}
+        if platform.system().lower() == "windows":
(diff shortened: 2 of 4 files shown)


BigRoy:
> [This PR fixes this Extract Review call window pop-up](https://github.com/ynput/OpenPype/blob/6754e97490c3462d4fbd27400bdcd9d24c20068d/openpype/plugins/publish/extract_review.py#L392)
> [This PR fixes this Extract Burnin call window pop-up](https://github.com/ynput/OpenPype/blob/c8c31018d656ea206c42e4a083365e21b4252861/openpype/plugins/publish/extract_burnin.py#L344)

Apparently these were due to calls with `shell=True` which actually start showing a command prompt again as soon you pass the `creationflags` we were passing. (shell=True by itself already tries to hide the console that's to be started)

Anyway, with this PR all command line popups in my tests were gone **when running from a build**. 

NOTE: However, **when running from source code** or using `openpype_console` a console seems to still pop-up for the [`run_openpype_process`](https://github.com/ynput/OpenPype/blob/12032f408cb65e028e7b117453a43a91ec7ddbda/openpype/plugins/publish/extract_burnin.py#L340) call in Extract Burnin which seems to pop-up a "Python" window. Likely because its triggering the `sys.executable` (python.exe or openpype_console.exe) which themselves open a console.

openpype/lib/execute.py
```diff
@@ -102,6 +102,10 @@ def run_subprocess(*args, **kwargs):
     if (
         platform.system().lower() == "windows"
         and "creationflags" not in kwargs
+        # shell=True already tries to hide the console window
+        # and passing these creationflags then shows the window again
+        # so we avoid it for shell=True cases
+        and not kwargs.get("shell", False)
```
iLLiCiTiT, line 108: I would do explicit check for `True`
```suggestion
        and kwargs.get("shell") is not True
```

iLLiCiTiT (approved):
Looks ok

EmberLightVFX:
I were just about to say LGTM but you merged it before I could press comment 😅

PR #4673, opened 2023-03-22
Global: custom location for OP local versions

kalisp:
## Changelog Description
This provides configurable location to unzip Openpype version zips. By default, it was hardcoded to artist's app data folder, which might be problematic/slow with roaming profiles.

Location must be accessible by user running OP Tray with write permissions (so `Program Files` might be problematic)

## Additional info
It would be beneficial to allow artist to change it for themselves, but I am not sure if it makes sense to invest time to add it to Local Settings with Ayon impending. 

## Testing notes:
1. configure `Custom Local Versions Folder` in System Settings
2. provide zip to `Versions repository`
3. check that zip was consumed and unzipped in location configured in step 1


igniter/bootstrap_repos.py
```diff
@@ -289,6 +294,31 @@ def get_openpype_path(cls):
         """
         return os.getenv("OPENPYPE_PATH")
 
+    @classmethod
+    def get_local_openpype_path(cls, settings=None):
```
iLLiCiTiT, line 298: You're caching the value on class so if you'll don't pass settings on first call, and then pass it on second, it won't use the settings because it's already cached.

I don't think the function implementation should be classmethod of `OpenPypeVersion` and should cache the value if passed argument should affect output but does not.

openpype/settings/entities/schemas/system_schema/schema_general.json
```diff
@@ -138,6 +142,18 @@
...
+                {
+                    "type": "label",
+                    "label": "Define custom location for artist machine where to unzip versions of Openpype code. By default it is in user app data folder."
+                },
+                {
+                    "type": "path",
+                    "key": "local_openpype_path",
+                    "label": "Custom Local Versions Folder",
+                    "multiplatform": true,
+                    "multipath": true,
```
iLLiCiTiT, line 154: I don't think this should be enabled. If you want to override where versions are stored there should be only one path.

igniter/bootstrap_repos.py
```diff
@@ -319,8 +339,7 @@ def get_local_versions(cls) -> List:
             list: of compatible versions available on the machine.
 
         """
-        # DEPRECATED: backwards compatible way to look for versions in root
-        dir_to_search = Path(user_data_dir("openpype", "pypeclub"))
+        dir_to_search = OpenPypeVersion.get_local_openpype_path()
```
iLLiCiTiT, line 342:
```suggestion
        dir_to_search = cls.get_local_openpype_path()
```

openpype/settings/entities/schemas/system_schema/schema_general.json
kalisp: oh, I forgot

igniter/tools.py
```diff
@@ -188,6 +188,31 @@ def get_openpype_path_from_settings(settings: dict) -> Union[str, None]:
...
+        .get(platform.system().lower())
+    ) or []
+    # For cases when `local_openpype_path` is a single path
+    if paths and isinstance(paths, str):
+        paths = [paths]
+
+    overridden_path = next((Path(path)
+                            for path in paths
+                            if os.path.exists(path)), None)
+    return overridden_path
```
iLLiCiTiT, line 213:
```suggestion
    path = (
        settings
        .get("local_openpype_path", {})
        .get(platform.system().lower())
    )
    if path:
        return path
    return None
```

openpype/settings/defaults/system_settings/general.json
```diff
@@ -15,6 +15,11 @@
         "darwin": [],
         "linux": []
     },
+    "local_openpype_path": {
+        "windows": [],
+        "darwin": [],
+        "linux": []
```
iLLiCiTiT, line 21:
```suggestion
        "windows": "",
        "darwin": "",
        "linux": ""
```

igniter/bootstrap_repos.py
```diff
@@ -554,6 +571,14 @@ def empty_progress(x: int):
...
         self._progress_callback = progress_callback
 
+    def set_data_dir(self, data_dir):
+        if not data_dir:
+            self.data_dir = Path(user_data_dir("openpype", "pypeclub"))
+        else:
+            self._print(f"overriding local folder: {data_dir}")
+            self.data_dir = data_dir
+
+
```
iLLiCiTiT, line 581:
```suggestion
```

openpype/settings/entities/schemas/system_schema/schema_general.json
```diff
@@ -130,6 +130,10 @@
             "label": "OpenPype deployment control",
             "collapsible": false,
```
iLLiCiTiT, line 131: There is too much now
```suggestion
            "collapsible": true,
```

iLLiCiTiT: Damn, it is just a wrapped so it won't show overriden state... Maybe not.

iLLiCiTiT (approved):
I had to do changes from https://github.com/ynput/OpenPype/pull/4763. But other than that, works as expected.

PR #4674, opened 2023-03-22
TVPaint: Optional Validation plugins can be de/activated by user

iLLiCiTiT:
## Changelog Description
Added `OptionalPyblishPluginMixin` to TVpaint plugins that can be optional.

## Additional information
I've also fixed settings of auto detect creator which did not have settings for `enabled` and fixed `"fps"` collection.

## Testing notes:
1. Open TVPaint
2. Open Publisher in TVPaint
3. Context (Options) item should have Validator plugins that can be de/activated


BigRoy:
I believe these plug-ins are missing the following check in the `process` method because the optional toggle is now possible per instance instead of for the full plugin so the plugin will remain active.
```python
if not self.is_active(instance):
    return
```

iLLiCiTiT:
> I believe the refactor is missing the following check in the process method because the optional toggle is now possible per instance instead of for the full plugin so the plugin will remain active.

I totally forgot about that part. Thanks

64qam (approved):

PR #4675, opened 2023-03-22
fix bug for the update function for 3dsMax loaders

moonyuet:
## Changelog Description
Fix the bug of unabling to update/set the version of the published assets in 3dsmax
- [x] maxSceneloader
- [x] fbxloader for camera
- [x] abcloader
- [x] point cloud loader

## Additional Info
MaxScene Loader doesn't support downgrading version
if the version is downgraded, it will show the separate container unparented from the main "_CON" container.

## Testing notes:
1. Launch 3dsMax via Openpype
2. Click Load... in Openpype menu
3.  Load the elder version of the published assets
4. Click Manage... in Openpype menu
5. Click Update to the latest /Switch Version
6. The asset update to the latest version/set version as expected


LiborBatek (changes requested):
Gives me errors when trying update version for ```pointcache``` instance already loaded in the scene which is outdated and needs to update via ```Scene Inventory```. 

image

LiborBatek (changes requested):
Re-run the testing with simple ```model``` instance and it doesnt work as expected

Scene before ```Update to latest``` action
image

After that (updating the asset) ```Inventory``` still shows wrong version even tho the asset itself is updated in the container. 
image

This leads to user confusion which version is actually loaded in the scene.

There are also some errors in console during attempts of set different version:
```Traceback (most recent call last):
  File "D:\REPO\OpenPype\openpype\tools\sceneinventory\view.py", line 248, in <lambda>
    lambda: _on_update_to_latest(items)
  File "D:\REPO\OpenPype\openpype\tools\sceneinventory\view.py", line 232, in _on_update_to_latest
    update_container(item, -1)
  File "D:\REPO\OpenPype\openpype\pipeline\load\utils.py", line 486, in update_container
    return loader.update(container, new_representation)
  File "D:\REPO\OpenPype\openpype\hosts\max\plugins\load\load_max_scene.py", line 80, in update
    rt.select(prev)
RuntimeError: MAXScript exception raised.
-- Runtime error: Attempt to access deleted scene object
```

LiborBatek:
Whats also strange when using ```Load``` with latest version by default it loads normally and the ```modelMain``` Container has following attributes (note the OP Parameter section)
image

but when using ```Load``` with older version switched before loading after that the ```modelMain``` Containers attribs differ and completely miss the OP params section.
image

moonyuet:
> Whats also strange when using ```Load``` with latest version by default it loads normally and the ```modelMain``` Container has following attributes (note the OP Parameter section)
> 
> image
> 
> 
> 
> but when using ```Load``` with older version switched before loading after that the ```modelMain``` Containers attribs differ and completely miss the OP params section.
> 
> image
> 
> 


Maybe we should wait and finish https://github.com/ynput/OpenPype/pull/4931 first and then work on the loaders upversioning.

moonyuet:
The upversioning has been resolved in #4931, @antirotor should we close this PR?

mkolar:
Yes let's close this down if it's resolved elsewhere.

PR #4676, opened 2023-03-22
Publisher: Explicit save

iLLiCiTiT:
## Changelog Description
Publisher have explicit button to save changes, so reset can happen without saving any changes. Save still happens automatically when publishing is started or on publisher window close. But a popup is shown if context of host has changed. Important context was enhanced by workfile path (if host integration supports it) so workfile changes are captured too. In that case a dialog with confirmation is shown to user. All callbacks that may require save of context were moved to main window to be able handle dialog show at one place. Save changes now returns success so the rest of logic is skipped -> publishing won't start, when save of instances fails.

Save and reset buttons have shortcuts (Ctrl + s and Ctrls + r).

## Additional info
Messages shown in dialogs may be cryptic for artists.

## Testing notes:
1. Open host using Publisher
2. Open Publisher tool
3. Create an instance
4. Change any attribute
5. Hit refresh
6. Attribute values should be changed back to their origin values
7. Change any attribute
8. Hit save
9. Hit refresh
10. Attribute values should stay
11. Change any attribute (e.g. `review`)
12. Hit publish
13. Value of the attribute should be propagated to publishing (review is created)


openpype/tools/publisher/widgets/widgets.py
```diff
@@ -198,12 +198,20 @@ def __init__(self, parent=None):
...
+        self.setToolTip("Save changes (Ctrl + S)")
+
+
 class ResetBtn(PublishIconBtn):
     """Publish reset button."""
     def __init__(self, parent=None):
         icon_path = get_icon_path("refresh")
         super(ResetBtn, self).__init__(icon_path, parent)
-        self.setToolTip("Refresh publishing")
+        self.setToolTip("Refresh publishing (Ctrl + R)")
```
BigRoy, line 214: The shortcuts likely need their labels set from the OS respective key sequence string for it so that on mac it's still accurate?

openpype/tools/publisher/widgets/overview_widget.py
```diff
@@ -336,13 +337,16 @@ def _on_instance_context_change(self):
...
-        self._controller.trigger_convertor_items(convertor_identifiers)
+        self.convert_requested.emit()
 
     def get_selected_items(self):
         view = self._subset_views_layout.currentWidget()
         return view.get_selected_items()
 
+    def get_selected_convertors(self):
+        _, _, convertor_identifiers = self.get_selected_items()
+        return convertor_identifiers
```
BigRoy, line 348: What's a convertor? What is being converted? Looking around the codebase this appears to be related to converting legacy instances but I can tell from just reading the code here I was clueless - this might need some decent docstrings or should maybe include the word `legacy`? Or are convertors not only intended for legacy?

openpype/tools/publisher/window.py
```diff
@@ -388,7 +396,9 @@ def resizeEvent(self, event):
     def closeEvent(self, event):
         self._window_is_visible = False
         self._uninstall_app_event_listener()
-        self.save_changes()
+        # TODO capture changes and ask user if wants to save changes on close
```
BigRoy, line 399: Yes please! :)

openpype/tools/publisher/window.py
```diff
@@ -421,6 +431,21 @@ def keyPressEvent(self, event):
...
+        if event.matches(QtGui.QKeySequence.Save):
+            self._save_changes(True)
+            event.accept()
+            return
+
+        if (
+            event.modifiers() == QtCore.Qt.ControlModifier
+            and event.key() == QtCore.Qt.Key_R
+        ):
+            if not self.controller.publish_is_running:
```
BigRoy, line 444: Does this same check maybe need to be done for CTRL+S?

openpype/tools/publisher/window.py
```diff
@@ -455,8 +480,65 @@ def _on_show_timer(self):
...
+            message = (
+                "Context has changed since Publisher window was refreshed last"
+                " time.\n\nAre you sure you want to save changes?"
+            )
+        else:
+            message = (
+                "Your action requires save of changes but context has changed"
+                " since Publisher window was refreshed last time.\n\nAre you"
+                " sure you want to continue and save changes?"
+            )
```
BigRoy, line 511: I've never gotten these messages - when should I be seeing these?

BigRoy (commented):
First off, amazing work!

#### My first few seconds trying this out and already getting confused

1. When opening the publisher, for me the UI directly shows the save icon as clickable - it should be disabled when there's nothing to save.
2. After clicking save, it remains clickable which seems even weirder. Also, it doesn't seem to do anything.
3. In this same case, when clicking CTRL+S - I'd expect at least a log message saying "Nothing to save".

#### Show user that it saved changes

After changing values around then visually it doesn't seem to save something when I click the save icon or press CTRL+S. There's no visual clue that it started or finished saving. No log, no message in the UI and the save icon stays active. Only after resetting I knew it did in fact end up saving. Having a message like the "Refreshed." saying "Saved changes." would go a long way for the user. 

The same goes for when changing settings then not explicitly saving yet clicking Validate directly. If at the top then it'd just show "Saved changes." I would be much more comfortable to be sure it's doing what I wanted it to do.

And if there was nothing to save, then maybe show "Nothing to save."

#### Show user that it discarded changes

I think when resetting it'd also be nice if at the top of the UI it would show "Discarding unsaved changes..." or "Discarding changes..." when it detected it had unsaved settings but you decided to reset anyway.

#### Refresh -> Reset

Now that resetting actually _resets_ the changes in the UI I'd say that this message:
afbeelding
Should likely just read "Resetting.." or alike to make it clearer it's ignoring what you didn't save. Similarly goes for the the "reset" button I'd expect to the tooltip to say "Reset publishing". 

#### Changing asset/task on an instance is confusing

When changing asset/task on an instance it now becomes a bit confusing that you first need to CONFIRM the changes at the top but then afterwards still need to SAVE the changes. It got me confused a few times thinking it didn't work for me to change the asset but then I noticed I just forgot to save that change because I just clicked confirm already.

openpype/tools/publisher/window.py
iLLiCiTiT: Not this, but different.

openpype/tools/publisher/window.py
iLLiCiTiT: This is reason why the save button is enabled all the time...

iLLiCiTiT:
>  If at the top then it'd just show "Saved changes."

Doable

> Should likely just read "Resetting.." or alike to make it clearer it's ignoring what you didn't save. Similarly goes for the the "reset" button I'd expect to the tooltip to say "Reset publishing".

It refreshed create context and instances, it is shown even on first show. Not necessarilly related to click on reset.

> When changing asset/task on an instance it now becomes a bit confusing that you first need to CONFIRM the changes at the top but then afterwards still need to SAVE the changes. It got me confused a few times thinking it didn't work for me to change the asset but then I noticed I just forgot to save that change because I just clicked confirm already.

I'm still struggling how that should behave -> when the subset name should be re-calculated.

Most of other comments have pre-requirement that we know that something changed to be able do them.

> I think when resetting it'd also be nice if at the top of the UI it would show "Discarding unsaved changes..." or "Discarding changes..." when it detected it had unsaved settings but you decided to reset anyway.

Maybe (for now) change tooltip to `Reset & Discard changes (Ctrl + R)`?

openpype/tools/publisher/window.py
BigRoy: Was able to get the message - nice! :) 

In a demo video of Nuke + new publisher it was shown to after publishing tweak the instances to "save it" for the next time you'll run the publish. With this PR you can still after publishing go to the Publish tab, tweak the state of instances but there's no save button. There's no save button, but CTRL+S does still work (that should be disabled then too?) and worse when clicking CTRL+S it now also shows a context changed message because my scene was incrementally saved during publish - adding to the confusion. :)

I'm fine with it not being possible to save until a reset happened but I'd have expected I shouldn't be able to change the active state on instances then either nor that CTRL+S should be possible.

iLLiCiTiT: > that should be disabled then too?

Fixed

> I shouldn't be able to change the active state on instances

That is possible?

openpype/tools/publisher/widgets/widgets.py
iLLiCiTiT: Changed `Ctrl` to `Command` for MacOS

openpype/tools/publisher/window.py
BigRoy: > That is possible?

Sure was. After Publishing go to Publish tab.

iLLiCiTiT: Not anymore

BigRoy: Seems to work for me! :)

BigRoy: Actually here's a fun one. :) You can still toggle the instance state.
Go to this menu mode:
afbeelding

Now select the instance, press **Spacebar**.
Haha, not a big one however since I can't save the state at that point in time.

openpype/tools/publisher/widgets/widgets.py
```diff
@@ -198,12 +205,24 @@ def __init__(self, parent=None):
...
 
 
+class SaveBtn(PublishIconBtn):
+    """Save context and instances information."""
+    def __init__(self, parent=None):
+        icon_path = get_icon_path("save")
+        super(SaveBtn, self).__init__(icon_path, parent)
+        self.setToolTip(
+            "Save changes ({} + S)".format(get_control_str())
+        )
```
BigRoy, line 215: Should we use Qt to define the human readable name for it?
```suggestion
        shortcut = QtGui.QKeySequence(QtGui.QKeySequence.Save).toString()
        self.setToolTip(
            "Save changes ({})".format(shortcut)
        )
```

We can do the same for the reset button:

```python
key_sequence =QtGui.QKeySequence(QtCore.Qt.ControlModifier | QtCore.Qt.Key_R)
shortcut = key_sequence.toString()
```

You could use this same key sequence to match the event instead of checking control modifier and save key separately.

openpype/tools/publisher/widgets/widgets.py
```diff
@@ -198,12 +205,24 @@ def __init__(self, parent=None):
...
+
 class ResetBtn(PublishIconBtn):
     """Publish reset button."""
     def __init__(self, parent=None):
         icon_path = get_icon_path("refresh")
         super(ResetBtn, self).__init__(icon_path, parent)
-        self.setToolTip("Refresh publishing")
+        self.setToolTip(
+            "Refresh publishing ({} + S)".format(get_control_str())
+        )
```
BigRoy, line 225: `+ S` is definitely a typo here. 


```suggestion
        self.setToolTip(
            "Reset & discard changes ({} + R)".format(get_control_str())
        )
```

openpype/tools/publisher/window.py
```diff
@@ -421,6 +431,22 @@ def keyPressEvent(self, event):
...
+        if event.matches(QtGui.QKeySequence.Save):
+            if not self._controller.publish_has_started:
+                self._save_changes(True)
+            event.accept()
+            return
+
+        if (
+            event.modifiers() == QtCore.Qt.ControlModifier
+            and event.key() == QtCore.Qt.Key_R
+        ):
```
BigRoy, line 444: If you decide to refactor above to use `QKeySequence.toString()` to get the platform-specific human readable string for it then potentially we might want to do so here too?
```suggestion

        reset_sequence = QtGui.QKeySequence(QtCore.Qt.ControlModifier | 
                                            QtCore.Qt.Key_R)
        if (
            event.matches(reset_sequence)
        ):
```

iLLiCiTiT:
> Does anyone else feel like we should maybe move the Save button?

For it was more natural to most left.

> alhough that wouldn't be as pretty if we hide the save button every now and then

Save button can be added to publish overlay (but disabled).

openpype/tools/publisher/widgets/overview_widget.py
```diff
@@ -336,13 +340,30 @@ def _on_instance_context_change(self):
...
+            tuple[list[str], bool, list[str]]: Selected items. List of
+                instance ids, context is selected, list of selected legacy
+                convertor plugins.
+        """
+
         view = self._subset_views_layout.currentWidget()
         return view.get_selected_items()
 
+    def get_selected_legacy_convertors(self):
+        """Selected legacy convertor identifiers.
```
kalisp, line 358: Still not sure, what this actually returns. Maybe some example of legacy convertor identifier?

BigRoy:
> Refresh takes a second for me and for a while nothing is happening.

What DCC did you test it in?

kalisp:
Again probably not for this PR, but not sure what "Hide removed instances" should work. I disabled one instance, published, toggled "Hide..", expected that disabled instance would show up there (similarly as skipped plugins with empty circle).
2023-03-23 11_39_15-OpenPype publisher

kalisp:
> > Refresh takes a second for me and for a while nothing is happening.
> 
> What DCC did you test it in?

Nuke

kalisp (approved):
Only small comments to docstring and non relating issues LGTM.

BigRoy (approved):
I think this is a solid first version. We will however _if we get to merging what's there now_ create separate issues to track the comments that were still made:

- [ ] Publisher: Enable save icon and shortcut only when changes were actually made
    - This would also include potentially rephrasing some of the wording like "Saved changes." popup to show "No changes to save." or "Discarded changes..." when resetting but there were some active changes.
- [ ] Publisher: Improve UX for asset/task change confirm and explicit saving

Both were mentioned in [this comment](https://github.com/ynput/OpenPype/pull/4676#pullrequestreview-1353307206)

iLLiCiTiT:
Created issues
https://github.com/ynput/OpenPype/issues/4689
https://github.com/ynput/OpenPype/issues/4690

PR #4677, opened 2023-03-22
CelAction: conditional workfile parameters from settings

jakubjezek001:
## Changelog Description
Since some productions were requesting excluding some workfile parameters from publishing submission, we needed to move them to settings so those could be altered per project.

## Testing notes:
1. Open Studio settings at CelAction host and remove one of Workfile parameters.
2. Start Celaction on task and go to Submit render > Dialogue window will popup
3. Notice that parameters are not showing up. 
4. hit Submit and after publisher will open and Collect hit Publish
5. all should publish as usual

### DIFF
--- a/openpype/settings/defaults/project_settings/celaction.json
+++ b/openpype/settings/defaults/project_settings/celaction.json
@@ -10,12 +10,10 @@
         }
     },
     "workfile": {
-        "parameters": [
-            "--chunk *CHUNK*",
-            "--frameStart *START*",
-            "--frameEnd *END*",
-            "--resolutionWidth *X*",
-            "--resolutionHeight *Y*"
+        "submission_overrides": [
+            "render_chunk",
+            "frame_range",
+            "resolution"
         ]
     },
     "publish": {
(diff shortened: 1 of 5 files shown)


openpype/settings/defaults/project_settings/celaction.json
```diff
@@ -9,6 +9,15 @@
             "rules": {}
         }
     },
+    "workfile": {
+        "parameters": [
+            "--chunk *CHUNK*",
```
iLLiCiTiT, line 14: This won't work, `"--chunk *CHUNK*"` is in fact 2 arguments `"--chunk"` and `"*CHUNK*"`. So it must be that way. or you must handle it in the prelaunch hook.

iLLiCiTiT: BTW I think that we define what arguments are available. Maybe is should not be dynamic list but "dict" where we define  keys and someone can change value.
```
Chunk: *CHUNK*
Frame Start: *START*
...
```

jakubjezek001: Yeah, it makes sense. It is committed, have a look ;)

iLLiCiTiT (approved):

PR #4678, opened 2023-03-22
Kitsu: Slightly less strict with instance data

BigRoy:
## Changelog Description

- Allow to take task name from context if asset doesn't have any. Fixes an issue with Photoshop's review instance not having `task` in data.
- Allow to match "review" against both `instance.data["family"]` and `instance.data["families"]` because some instances don't have the primary family in families, e.g. in Photoshop and TVPaint.
- Do not error on Integrate Kitsu Review whenever for whatever reason Integrate Kitsu Note did not created a comment but just log the message that it was unable to connect a review.

## Additional info

These came from issues reported on Discord [here](https://discord.com/channels/517362899170230292/563751989075378201/1088047135938842664) and [here](https://discord.com/channels/517362899170230292/1088080160823193643)

## Testing notes:

1. Publish some things using Kitsu module with comments and without comments, with reviews and without reviews preferably from a few different DCCs

### DIFF
--- a/openpype/modules/kitsu/plugins/publish/collect_kitsu_entities.py
+++ b/openpype/modules/kitsu/plugins/publish/collect_kitsu_entities.py
@@ -29,7 +29,7 @@ class CollectKitsuEntities(pyblish.api.ContextPlugin):
             if not zou_asset_data:
                 raise ValueError("Zou asset data not found in OpenPype!")
 
-            task_name = instance.data.get("task")
+            task_name = instance.data.get("task", context.data.get("task"))
             if not task_name:
                 continue
 
--- a/openpype/modules/kitsu/plugins/publish/integrate_kitsu_review.py
+++ b/openpype/modules/kitsu/plugins/publish/integrate_kitsu_review.py
@@ -12,10 +12,9 @@ class IntegrateKitsuReview(pyblish.api.InstancePlugin):
     optional = True
 
     def process(self, instance):
-        task = instance.data["kitsu_task"]["id"]
-        comment = instance.data["kitsu_comment"]["id"]
 
         # Check comment has been created
+        comment = instance.data.get("kitsu_comment", {}).get("id")
         if not comment:
             self.log.debug(
                 "Comment not created, review not pushed to preview."
@@ -23,6 +22,7 @@ class IntegrateKitsuReview(pyblish.api.InstancePlugin):
             return
 
         # Add review representations as preview of comment
+        task = instance.data["kitsu_task"]["id"]
         for representation in instance.data.get("representations", []):
             # Skip if not tagged as review
             if "kitsureview" not in representation.get("tags", []):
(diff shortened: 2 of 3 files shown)


openpype/modules/kitsu/plugins/publish/integrate_kitsu_review.py
```diff
@@ -12,17 +12,17 @@ class IntegrateKitsuReview(pyblish.api.InstancePlugin):
...
         # Check comment has been created
+        comment = instance.data.get("kitsu_comment", {}).get("id")
         if not comment:
             self.log.debug(
                 "Comment not created, review not pushed to preview."
             )
             return
 
         # Add review representations as preview of comment
+        task = instance.data["kitsu_task"]["id"]
```
Tilix4, line 25: As we are dealing with IDs here, and refactoring for better understanding, I'd rename them for `comment_id` and `task_id`.

Tilix4 (changes requested):
Nice clean up! Added a very small suggestion.

openpype/modules/kitsu/plugins/publish/integrate_kitsu_review.py
EmberLightVFX: +1 for this

EmberLightVFX (commented):
Fix Félix's comment and then it LGTM :)

openpype/modules/kitsu/plugins/publish/integrate_kitsu_review.py
BigRoy: Done.

openpype/modules/kitsu/plugins/publish/collect_kitsu_entities.py
```diff
@@ -29,7 +29,7 @@ def process(self, context):
             if not zou_asset_data:
                 raise ValueError("Zou asset data not found in OpenPype!")
 
-            task_name = instance.data.get("task")
+            task_name = instance.data.get("task", context.data.get("task"))
```
iLLiCiTiT, line 32: We would like to get rid of using context task in these situations. Instance should know it's task. Was there a use-case when it was not filled?

EmberLightVFX: `Fixes an issue with Photoshop's review instance not having task in data.`

Tilix4 (approved):
LGTM! deal with @iLLiCiTiT now :)

openpype/modules/kitsu/plugins/publish/collect_kitsu_entities.py
BigRoy: Yes - it was in Photoshop [Collect Review which creates an instance during publishing](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/photoshop/plugins/publish/collect_review.py#L42-L52).

I've mentioned on Discord that it should've likely been refactored to a new style AutoCreator because before the refactor to new publisher I assume one would've been able to disable the review instance in the publisher. Looking at the code now I suspect that isn't possible.

(I haven't run Photoshop in OP myself)

iLLiCiTiT (approved):

PR #4682, opened 2023-03-22
Improve logging of used app + tool envs on application launch

BigRoy:
## Changelog Description

Improve logging of what apps + tool environments got loaded for an application launch.

#### Before

Before unordered (due to `set`) and unclear what belonged to what:
```
  - { GlobalHostDataHook }: [  Will add environments for apps and tools: mtoa, 2023, 3-2, 3-1, maya  ]
```
before

#### After

After order is preserved and app name / tool names match:
```
  - { GlobalHostDataHook }: [  Will add environments for apps and tools: maya/2023, mtoa/3-1, mtoa/3-2  ]
```
afbeelding

### Additional info

Additional benefit is that now you can also see/debug in what order the environments were build - so that if one env var ended up overriding one of another tool this might help to debug what could've overwitten what.

_Ignore the fact that for this test case in the screenshots/examples I added both mtoa 3.1 and 3.2 into a single asset_

## Testing notes:
1. Launch Tray (from code or openpype_console)
2. Launch an application, read the console log

### DIFF
--- a/openpype/lib/applications.py
+++ b/openpype/lib/applications.py
@@ -1508,8 +1508,8 @@ def prepare_app_environments(
         if key in source_env:
             source_env[key] = value
 
-    # `added_env_keys` has debug purpose
-    added_env_keys = {app.group.name, app.name}
+    # `app_and_tool_labels` has debug purpose
+    app_and_tool_labels = ["/".join([app.group.name, app.name])]
     # Environments for application
     environments = [
         app.group.environment,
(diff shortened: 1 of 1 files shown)


mkolar (approved):

openpype/lib/applications.py
```diff
@@ -1508,8 +1508,8 @@ def prepare_app_environments(
         if key in source_env:
             source_env[key] = value
 
-    # `added_env_keys` has debug purpose
-    added_env_keys = {app.group.name, app.name}
+    # `app_and_tool_labels` has debug purpose
+    app_and_tool_labels = ["/".join([app.group.name, app.name])]
```
iLLiCiTiT, line 1512: `"/".join([app.group.name, app.name])` could be replace by `app.full_name`

iLLiCiTiT: Milan's fury :)

PR #4687, opened 2023-03-23
Maya: enable 2D Pan/Zoom for playblasts - OP-5213

tokejepsen:
## Changelog Description
Setting for enabling 2D Pan/Zoom on reviews.

## Additional info
Also a bug fix where the review representation images were hardcoded to PNGs.

## Testing notes:
1. Enable `project_settings/maya/publish/ExtractPlayblast/capture_preset/Generic/pan_zoom`
2. Setup review in Maya with a camera that uses 2d Pan/Zoom.
3. Publish and validate the review is using the 2d Pan/Zoom.


website/docs/admin_hosts_maya.md
```diff
@@ -110,6 +110,39 @@ or Deadlines **Draft Tile Assembler**.
...
+- **Background Color** the viewports background color.
+- **Background Bottom** the viewports background bottom color.
+- **Background Top** the viewports background top color.
+- **Override display options** override the viewports display options to use what is set in the settings.
+- **Isolate view** isolates the view to what is in the review instance. If only a camera is present in the review instance, all nodes are displayed in view.
+- **Off Screen** records the playblast hidden from the user.
+- **2D Pan/Zoom** enables the 2D Pan/Zoom functionality of the camera.
+- **Renderer name** which renderer to use for playblasting.
+- **Width** width of the output resolution.
+- **Height** height of the output resolution.
```
BigRoy, line 130: Should we elaborate what happens if you set a value of zero? Which is when it defaults to the asset document's resolution IF the review width/height on the instance isn't explicitly overridden with the instance settings in the scene.

website/docs/admin_hosts_maya.md
```diff
@@ -110,6 +110,39 @@ or Deadlines **Draft Tile Assembler**.
...
+
+![Extract Playblast Settings](assets/maya-admin_extract_playblast_settings.png)
+
+- **Compression type** which file encoding to use.
+- **Data format** what format is the file encoding.
+- **Quality** lets you control the compression value for the output. Results can vary depending on the compression you selected. Quality values can range from 0 to 100, with a default value of 95.
+- **Background Color** the viewports background color.
+- **Background Bottom** the viewports background bottom color.
+- **Background Top** the viewports background top color.
+- **Override display options** override the viewports display options to use what is set in the settings.
```
BigRoy, line 124: Wait a second, isn't this missing a setting to decide on whether it should use a gradient background or not? :)

BigRoy (commented):
Nice!

> Also a bug fix where the review representation images were hardcoded to PNGs.

~Where's this fix?~ https://github.com/ynput/OpenPype/commit/236078d072bf1f3aff89970da9d52abac7e9e38a

openpype/hosts/maya/plugins/publish/extract_thumbnail.py
```diff
@@ -124,13 +124,16 @@ def process(self, instance):
 
         # Disable Pan/Zoom.
         pan_zoom = cmds.getAttr("{}.panZoomEnabled".format(preset["camera"]))
-        cmds.setAttr("{}.panZoomEnabled".format(preset["camera"]), False)
+        cmds.setAttr(
+            "{}.panZoomEnabled".format(preset["camera"]),
+            instance.data["panZoom"]
+        )
```
BigRoy, line 130: I feel like this should be set using a context manager. I suspect currently if you were to press escape during playblasting or any other error occurred during playblasting (e.g. no disk space or whatever) that the scene would be left in a changed state where the panZoomEnabled state of the camera might mismatch what the user had prior to that.

You should be able to use this [`attribute_values`](https://github.com/pypeclub/OpenPype/blob/a74e62a482ae0e8a1ef081cc3a85d3aa1a173258/openpype/hosts/maya/api/lib.py#L443) context manager and do:
```python
remap = {
    "{}.panZoomEnabled".format(preset["camera"]): instance.data["panZoom"]
}

with attribute_values(remap):
    ...
```

Then you also don't need to get the original value here first and revert the original value later manually too.

website/docs/admin_hosts_maya.md
```diff
@@ -110,6 +110,39 @@ or Deadlines **Draft Tile Assembler**.
 This is useful to fix some specific renderer glitches and advanced hacking of Maya Scene files. `Patch name` is label for patch for easier orientation.
 `Patch regex` is regex used to find line in file, after `Patch line` string is inserted. Note that you need to add line ending.
 
+### Extract Playblast Settings (review)
+These settings provide granular control over how the playblasts or reviews are produced in Maya.
```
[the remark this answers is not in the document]
tokejepsen: yeah, good point.

website/docs/admin_hosts_maya.md
tokejepsen: Did know that :)

website/docs/admin_hosts_maya.md
tokejepsen: Yeah, that seems like its completely inaccessible from the settings or anywhere. Outside of this scope unfortunately to fix so have made an issue for it; https://github.com/ynput/OpenPype/issues/4693

LiborBatek (approved):
Followed the testing steps and set ```Pan/Zoom``` to ```On``` in OP Settings

image


after creating ```review``` with turned on ```Pan/Zoom``` all worked ok when checking the output (was using Pan/Zoom) and it was correctly cropped to Pan/Zoom area:
image

tokejepsen:
@BigRoy you happy with this?

BigRoy (approved):
Solid work!

I am thinking since previously the codec was hardcoded to PNG but it now uses the settings which defaults to JPEG whether we should update the default to PNG so that it matches previous behavior?

Only issue with that might be that it'd change the default behavior for thumbnails in Maya maybe? Unless that's hardcoded to jpeg already?

tokejepsen:
>I am thinking since previously the codec was hardcoded to PNG but it now uses the settings which defaults to JPEG whether we should update the default to PNG so that it matches previous behavior?

Yeah, good point and done.

> I am thinking since previously the codec was hardcoded to PNG but it now uses the settings which defaults to JPEG whether we should update the default to PNG so that it matches previous behavior?

Its hardcoded to `jpeg` atm. What you do think? Follow playblast encoding or keep it hardcoded to `jpeg`.

BigRoy:
> Its hardcoded to jpeg atm. What you do think? Follow playblast encoding or keep it hardcoded to jpeg.

I'd keep it as is for now. PNGs create alpha and not sure whether we want that on the thumbnails for now.

PR #4692, opened 2023-03-23
Suggestion to change `integrations` to `hosts`.

tokejepsen:
## Changelog Description
Change sidebar section `Integrations` to `Hosts`.

## Additional info
The reason for this would be to have better terminology alignment with what is in code and daily chat on discord. Found myself searching for `Hosts` in the `Admin` page because the term `Integrations` arent used except maybe on the front page/promotions.

### DIFF
--- a/website/sidebars.js
+++ b/website/sidebars.js
@@ -32,7 +32,7 @@ module.exports = {
         {
             type: "category",
             collapsed: false,
-            label: "Integrations",
+            label: "Hosts",
             items: [
                 "artist_hosts_hiero",
                 "artist_hosts_nuke_tut",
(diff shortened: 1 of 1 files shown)


mkolar (changes requested):
Not the right approach I'm afraid. At the moment there are also other integrations like kitsu and ftrack. Also with the switch to AYON all of these will become addons, so if we change, that would be better term. 

Also Host is a completely alien term to artist, who is most likely going go browse the "artist" section. As of this week we're officially launching new docs for AYON, that need a lot of this type of cleanup, but in the new context and we'll spend the next few days working on structuring that.

mkolar:
Sorry @tokejepsen I'm closing this, for the reasons listed above.

PR #4694, opened 2023-03-23
Publisher: Fix reset shortcut sequence

iLLiCiTiT:
## Changelog Description
Fix bug created in https://github.com/ynput/OpenPype/pull/4676 where key sequence is checked using unsupported method. The check was changed to convert event into `QKeySequence` object which can be compared to prepared sequence.

## Testing notes:
1. Open Publisher (e.g. TrayPublisher)
2. Pres Ctrl + R
3. Publisher should reset

### DIFF
--- a/openpype/tools/publisher/window.py
+++ b/openpype/tools/publisher/window.py
@@ -439,7 +439,9 @@ class PublisherWindow(QtWidgets.QDialog):
             event.accept()
             return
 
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
LGTM.

No more errors whilst testing in my Substance Painter integration. Looks good, thanks. Sorry about misdirecting the original code.

PR #4695, opened 2023-03-23
3dsmax: make sure that startup script executes

antirotor:
## Changelog Description
Fixing reliability of OpenPype startup in 3dsmax.

## Additional info
Unfortunately using `ADSK_3DSMAX_STARTUPSCRIPTS_ADDON_DIR` environment variable to specify location of OpenPype startup script isn't very reliable, because of clash with other plugins, some strange race condition happening during the startup of max, or for whatever obscure reason. This fix is forcing max to execute OpenPype startup script via command line argument. Caveat is that it might be harder to connect it to existing pipeline if this functionality is already used.

## Testing notes:
Try to launch 3dsmax with OpenPype integration - it should always work.

### DIFF
--- a/openpype/settings/defaults/system_settings/applications.json
+++ b/openpype/settings/defaults/system_settings/applications.json
@@ -133,7 +133,7 @@
                     "linux": []
                 },
                 "arguments": {
-                    "windows": [],
+                    "windows": ["-U MAXScript {OPENPYPE_ROOT}\\openpype\\hosts\\max\\startup\\startup.ms"],
                     "darwin": [],
                     "linux": []
                 },


openpype/settings/defaults/system_settings/applications.json
```diff
@@ -133,7 +133,7 @@
                     "linux": []
                 },
                 "arguments": {
-                    "windows": [],
+                    "windows": ["-U MAXScript {OPENPYPE_ROOT}\\openpype\\hosts\\max\\startup\\startup.ms"],
```
BigRoy, line 136: If this is always needed, isn't this better to add in a prelaunch hook instead instead of relying on application settings that users tend to adjust?

antirotor: Since we've never found the exact reason, I think it is better to leave it here as the original method with `ADSK_3DSMAX_STARTUPSCRIPTS_ADDON_DIR` is imho more cleaner. I think that with 3dsmax 2023 it less of an issue, but this at least allows to fallback to environment variable way if studio would like to use their own startup script in command line arguments (not sure if you can chain them, probably not)

64qam (approved):

PR #4697, opened 2023-03-23
Maya: fix the bug of not being able to find ocio_path in extract look

moonyuet:
## Changelog Description
fix the bug of not being able to find ocio_path in extract look and errors out in the extractor

## Additional info
n/a

## Testing notes:
1. Create Look Instance
2. Disable Color Management in preference setting and swtich maketx setting on
3. Publish it


openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -552,8 +552,8 @@ def _process_texture(self, filepath, resource,
                         self.log.warning("cannot guess the colorspace"
                                          "color conversion won't be available!")    # noqa
 
-
-            additional_args.extend(["--colorconfig", config_path])
+            if config_path:
```
BigRoy, line 555: The variable `config_path` will still be undefined. You'd need to initialize it above to `config_path = None` to make this work

moonyuet: Yes, I just find out about that.

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -536,7 +536,7 @@ def _process_texture(self, filepath, resource,
                                             color_space,
                                             render_colorspace])
                 else:
-
+                    config_path = ""
```
BigRoy, line 539: This is fine to be `None` I'd say since we're checking with if statement before we use it anyway?

BigRoy (approved):
Haven't done a test run but code seems fine.

LiborBatek (changes requested):
Did follow the testing steps but gives me error during Publish:

image

LiborBatek (approved):
Tested in ```maya 2023``` without turned on Color Management and with ```maketx``` se to ```On```.

Validated ok without problems.

image

BigRoy (approved):
This seems fine to merge - but the issue this is trying to solve will also be solved by https://github.com/ynput/OpenPype/pull/4711 so we could also focus on merging that instead.

tokejepsen:
@moonyuet is this still valid with https://github.com/ynput/OpenPype/pull/4711 merged?

moonyuet:
> @moonyuet is this still valid with #4711 merged?

No, #4711 is the updated version so I will close it now.

PR #4699, opened 2023-03-23
AYON: Editorial hierarchy creation

iLLiCiTiT:
## Changelog Description
Implemented extract hierarchy to AYON plugin which created entities in AYON using ayon api.

## Additional info
I didn't test case when tasks should be created as I don't have testing data which would cause that. It was crashing because of missing OCIO configs for me now -> not fully related to this PR. I will try fix it later on.

## Testing notes:
1. Start editorial publishing which should create/update entities in project (e.g. Simple Editorial in TrayPublisher)
2. It should create entities in the project


jakubjezek001 (approved):
LGTM ;)

PR #4700, opened 2023-03-23
ImageIO: Adding ImageIO activation toggle to all hosts

jakubjezek001:
## Changelog Description
Colorspace management can now be enabled at the project level, although it is disabled by default. Once enabled, all hosts will use the OCIO config file defined in the settings. If settings are disabled, the system switches to DCC's native color space management, and we do not store colorspace information at the representative level.

## Additional info
Each host that publishes pixel data can now enable Color Management. This allows the publishing and loading process to run through file rules identification. 

## Hosts needs to be tested
- [x] Nuke
- [x] Hiero
- [ ] Resolve
- [x] Maya
- [ ] Fusion
- [x] Houdini
- [ ] After Effects
- [ ] 3D Max
- [x] Unreal
- [x] Blender
- [ ] Substance Painter

## Testing notes:
1. Open Studio Project Settings
2. Enable `project_settings/global/imageio/activate_global_color_management`
3. Go to any host and see there following  toggles each ImageIO category - except Global:
     - **Enable Color Management** - this overrides the globally enabled colormanagement.
4. each of `ocio_config` subcategory is holding now new toggle  **Override global OCIO config**
5. each of `file_rules` subcategory is holding now new toggle  **Override global file rules**

closes https://github.com/ynput/OpenPype/pull/4700

# Documentation
https://github.com/ynput/ayon-documentation/pull/42


BigRoy:
So the enabled setting currently is set to false for all hosts. Does that mean that with these settings that now all hosts should have *no* color management done by OpenPype by default at all? Seems like we want to default the opposite and have it enabled right?

Also, I have a feeling that potentially people might think that disabling the checkbox might mean "Do not override the global imagio settings" and thus think having it off would be "Use global imagio settings" instead of what you describe here where it means "Disable color management completely in the host"

Should we somehow make that difference clearer? And how?

openpype/hosts/flame/hooks/pre_flame_setup.py
```diff
@@ -47,6 +47,13 @@ def execute(self):
 
         imageio_flame = project_settings["flame"]["imageio"]
 
+        colormanaged = True
+        # check if host settings are having enabled key and if it is False
+        if imageio_flame.get("enabled") and imageio_flame["enabled"] is False:
+            # if host settings are disabled return False because
+            # it is expected that no colorspace management is needed
+            colormanaged = False
```
iLLiCiTiT, line 55:
```suggestion
        colormanaged = imageio_flame.get("enabled")
        if colormanaged is None:
            colormanaged = True
```

BigRoy:
To enumerate:
1. There's a global imageIO
2. There's a host imageIO
3. There's a way to toggle whether the host imageIO overrides the global imageIO
4. There's a way to toggle whether we want to enable or disable color management in the host.

We can also debate whether maybe OpenPype should ALWAYS color manage - but I'll leave that up to Ynput team to decide.
Anyway we'd need to be able to clearly differentiate between **enabling host imageIO overrides** and **disabling host color management completely**.

> You are rigth and the question is tough one. I am having personally this confusion now, since I am implementing it into some hosts. This enable toggle on ImageIO is disabling OCIO config - driven by OP. If it is enabled then I will get either Global or Host config, depends on the ocio_config enable key state. So if ImageIO is disabled then no ocio_config will be returned and OP will desable colormanagement features.

Makes sense. I still feel that the toggle isn't clear enough and it will be a source of confusion. Maybe only if that imageio config per host starts with a big checkbox "Override global imageIO settings" to enable/disable that then maybe it makes sense in a way?

> My concern is there might be following confusion in Nuke. If ImageIO is disabled then OP is not driving nukes colormanagement. But since Nukes colorspace configuration is placed under imagio key it looks like we are switching off ColorManagement, but we are not. We still need to set workfile defaults project_settings/nuke/imageio/workfile and nodes defaults project_settings/nuke/imageio/nodes. One can argue that this should not be under ImageIO in the first place, but it is very much related to the ColorManagement and it is embeded into Nuke integraton.

That actually sounds like disabling color management is actually not what should be allowed for Nuke since it'd break things?
Is it maybe the case that OpenPype should actually enforce color management and we should disallow disabling it in the hosts?

---

By the way, we could also just disable the header checkbox and make to clearly labeled checkboxes in host imageIO settings:

- Enable Color Management in host (enabled by default)
- Override global imageIO settings (disabled by default)

Then maybe that's the default values so that global imageio is the default setting for everything. Having labels instead of a header checkbox in this case might make it much clearer what they'll do. We can potentially even provide a more thorough tooltip too.

openpype/pipeline/colorspace.py
```diff
@@ -344,6 +345,12 @@ def get_imageio_config(
     imageio_global, imageio_host = _get_imageio_settings(
         project_settings, host_name)
 
+    # check if host settings are having enabled key and if it is False
+    if imageio_host.get("enabled") and imageio_host["enabled"] is False:
```
iLLiCiTiT, line 349: This condition can never happen.
```suggestion
    if "enabled" in imageio_host and imageio_host["enabled"] is False:
```

or maybe explicit, to give a hint why it is like that
```
imageio_enabled = imageio_host.get("enabled")
if imageio_enabled  is None:
    imageio_enabled = True

if not imageio_enabled :
    return False
```

openpype/hosts/fusion/hooks/pre_fusion_ocio_hook.py
fabiaserra, line 1: Could we make this generic (i.e., `PreLaunchOCIO`) and move it to `openpype/hooks/pre_ocio_hook.py`? I have done this on my fork of OP so `OCIO` is set on all hosts

openpype/hosts/nuke/api/lib.py
```diff
@@ -2001,18 +2001,18 @@ def set_viewers_colorspace(self, viewer_dict):
...
 
         '''
-        workfile_settings = nuke_colorspace["workfile"]
+        workfile_settings = imageio_host["workfile"]
 
-        # resolve config data if they are enabled in host
+        # get config data if imageio is enabled
         config_data = None
-        if nuke_colorspace.get("ocio_config", {}).get("enabled"):
+        if imageio_host.get("enabled") and imageio_host["enabled"] is True:
```
fabiaserra, line 2015: isn't this the same as just:
`if imageio_host.get("enabled") == True:`

fabiaserra:
Totally agree with @BigRoy on this one, I find it super confusing as it is and the distinction between both intentions needs to be done. I like his suggestion of using:

> * Enable Color Management in host (enabled by default)
> * Override global imageIO settings (disabled by default)

From my perspective, I don't foresee any situation where we'd want to do any color management overrides per DCC. In our current pipeline we manage everything through a single OCIO config file by setting $OCIO env var on all hosts environment and we haven't encountered any limitations with that workflow. I think that trying to abstract that simple logic away is counter-intuitive and can easily get over-engineered and harder to debug. As I said on my comment https://github.com/ynput/OpenPype/pull/4700#discussion_r1149417771, for now we have created a prelaunch hook for all hosts to simply set OCIO and let each host make use of that.

I think it's a nice addition to have the flexibility to override the color management per DCC but for us it's either been an entire show is OCIO managed or not. However, in the case where we wanted to not use OCIO on a DCC, I'd probably just do it by not setting OCIO on its environment and that's it.

Is the current expectation that you want to store OCIO data on the workfile itself and bake the setting on all the nodes of the scene for backwards compatibility? Is that a common requirement? Isn't it enough that we store the OCIO info as metadata on the published assets created from the workfile and version control the OCIO config file (in the future when we add that functionality to OP) itself if you ever want to reproduce the same publish? I'd hope we can set the OCIO on the whole workfile and let all the nodes default colorspace to be governed by the global OCIO instead of having to traverse all your scene for possible changes? My view is probably simplistic but my brain hurts a bit trying to reverse engineer the current Nuke colorspace workflow.

Could we do a clean cut and remove support of all the deprecated ImageIO settings? From the settings I find it hard to know which are the deprecated settings and which aren't so could we group all of the deprecated ones under a single entry?
image

openpype/hosts/nuke/api/lib.py
iLLiCiTiT: It should be if `imageio_host.get("enabled") is True`, but in this case only `if imageio_host.get("enabled"):` would work too.

openpype/hosts/fusion/hooks/pre_fusion_ocio_hook.py
iLLiCiTiT: Not all DCCs support OCIO. Since we want to split hosts and modules into separated repositories it is probably better to have it separated from core.

BigRoy: > Not all DCCs support OCIO. Since we want to split hosts and modules into separated repositories it is probably better to have it separated from core.

I think this will actually be the biggest downside to the separation, the amount of code duplication. But in this case I definitely agree with @fabiaserra - this makes much more sense on a global level instead of per host prelaunch hooks getting off sync. Hosts that do not support OCIO likely do not use the `OCIO` env flag at all anyway.

openpype/pipeline/colorspace.py
jakubjezek001: hey @fabiaserra it should be as it is said `check if host settings are having enabled key and if it is False`

We want ot make sure this is only happening if the bool is explicity set to False and exists in settings group. To explain it more: some productions might already set project overrides with older settings and since imageio category is group enabled, new schema is not effecting it and the `enable` key will not be part of settings data.

fabiaserra: sure, but I don't see how that is any different from what I'm saying?
`imageio_host.get("enabled") == False:` will only evaluate to True if `"enabled"` key exists and it's equal to False.

openpype/hosts/fusion/hooks/pre_fusion_ocio_hook.py
fabiaserra: exactly, $OCIO is an industry standard env that would be useful to have access on any host regardless of the host not using it

openpype/hosts/nuke/api/lib.py
fabiaserra: oh I see, `enabled` doesn't have a bool value

openpype/hosts/fusion/hooks/pre_fusion_ocio_hook.py
fabiaserra: awesome, glad to hear! thank you!

jakubjezek001: I have added set of hosts which are supporting somehow OCIO env var, but kept excluded Nuke Hiero and Maya - those are dealing with the ocio own native way. But actually I am not so sure about May. @BigRoy any idea?

jakubjezek001:
Lets break down your input @BigRoy:

> There's a way to toggle whether the host imageIO overrides the global imageIO

Yes that is done by host's `imageio/ocio_config/enable` == True

> There's a way to toggle whether we want to enable or disable color management in the host.

At the moment this is conficting a bit and it is done differently from host to host. But to be more specific at the moment of this PR we are having finally mechanism to do it. Once the `imageio/enabled` == False. There is no ocio_config returned and we can decide per host what does it mean. Mostly it means that we do not want to set custom ocio config path as it is done in Nuke. But we still want to control hosts colormanagement in native way. Like the example from Nuke we should be still able to set nuke default colormanagement and bit related colorspace names.

> ... to be able to clearly differentiate between enabling host imageIO overrides and disabling host color management completely.

I believe we want to control color management from settings all the time, except we want to decide if we want to switch to a custom OCIO configs.

> Enable Color Management in host (enabled by default)
> Override global imageIO settings (disabled by default)

I guess as much as I do not like this we have to do it this way. Except I cannot see `Enable Color Management in host` in Nuke at all. This would break Nuke integration completly.

jakubjezek001:
@fabiaserra thank you very much for your input (or ynput ;) ) Lets break down your comment here:

> Is the current expectation that you want to store OCIO data on the workfile itself and bake the setting on all the nodes of the scene for backwards compatibility? 

Not sure I understood correctly. What do you mean with `bake the settings on all the nodes`?

> Isn't it enough that we store the OCIO info as metadata on the published assets created from the workfile and version control the OCIO config file (in the future when we add that functionality to OP) itself if you ever want to reproduce the same publish? 

We are slowly getting there, but yes that is the goal.

> I'd hope we can set the OCIO on the whole workfile and let all the nodes default colorspace to be governed by the global OCIO instead of having to traverse all your scene for possible changes? My view is probably simplistic but my brain hurts a bit trying to reverse engineer the current Nuke colorspace workflow.

I am understanding your flustration but think of all the supported Nuke version we have to cover for all our clients (some of them are still on Nuke 11). This node colorspace driven from OCIO config's Roles is not working in all versions. Also we needed to create something which is more suitable for everyone, even those who are not that proficient in OCIO config reading and want to copy paste the colorspace names form Nuke script. But in future in Ayon we are plannig to push this farther and offer enumerators which will offer all the ocio config colorspaces/roles/aliases, the same way as Nuke is doing it. 

> Could we do a clean cut and remove support of all the deprecated ImageIO settings? From the settings I find it hard to know which are the deprecated settings and which aren't so could we group all of the deprecated ones under a single entry?

Fidling with Settings schemas are allowing it but we have to make sure all productions will be backward compatible and if we change structure of keys we could potentially harm some clients/projects.

openpype/hosts/flame/hooks/pre_flame_setup.py
```diff
@@ -47,6 +47,10 @@ def execute(self):
 
         imageio_flame = project_settings["flame"]["imageio"]
 
+        colormanaged = imageio_flame.get("enabled")
+        if colormanaged is None:
```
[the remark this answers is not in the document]
jakubjezek001: good point, thanks

openpype/hosts/nuke/api/lib.py
```diff
@@ -2001,18 +2001,18 @@ def set_viewers_colorspace(self, viewer_dict):
                 "Attention! Viewer nodes {} were erased."
                 "It had wrong color profile".format(erased_viewers))
 
-    def set_root_colorspace(self, nuke_colorspace):
+    def set_root_colorspace(self, imageio_host):
         ''' Adds correct colorspace to root
 
         Arguments:
-            nuke_colorspace (dict): adjustmensts from presets
+            imageio_host (dict): adjustments from presets
```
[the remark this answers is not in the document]
jakubjezek001: Yeah, it was some old spider webs and dust ;) but lets keep it on `imageio_host` since we are in Nuke host api lib, it might be explicit enough.

openpype/hosts/fusion/hooks/pre_fusion_ocio_hook.py
fabiaserra: what do you mean native way? Not sure about Hiero and Flame (although pretty sure it does as well) but Nuke and Maya make use of the $OCIO env var on launch - if it's set, it sets their color management to the $OCIO config path.

BigRoy:
> This seems to be some broken logic in the text right? Following is what I understand from first sentence : host.imageio.disable == global.imageio.enable + host.imageio.enable. That is not part of this PR and even cannot be read in comments. So I believe it might be some mistake in the text, right?

Let me just explain the way I think it should be:
```
enabled = global.imageio.enable
if host.imageio.override:
    enabled = host.imageio.enable
```
If I'm overriding the host I don't care what the global settings are. 
Or even better:
```
settings = global.imageio
if host.imageio.override:
    settings = host.imageio
```
When host override is enabled then it doesn't listen to anything from global.

> Isn't the text in Settings explaining enough. Seems to be clearly saying the key is depricated and not that it will be depricated. And those productions which didn't migrate the values into new keys will find out and do so. Honestly, I would also want to remove it and it would be a pleasure for me to have this cleared, but still I believe it is too soon.

Having a message in settings saying "your stuff is now broken" is not what backwards compatibility is. Anyway, I think we're currently saying the same thing. The old settings values there mean nothing and we need no compatibility for that. It's only there for someone to copy original settings from if they need.

> There are three host types we have to consider.
>     Host with OCIO and Native colorspace management
>    Host with Native colorspace management
>    Host without color management

Totally understand - but I'd say we then have per host the same settings (which are either read from global or overridden locally) + application specific imageio settings (which would always be used since global wouldn't provide that).

---

> I am having feeling that we might be repeating ourselves already

We have the same end goal and I feel we're approaching the same line of thought. However, just looking at the enabled toggles in settings I'm still unsure as a user how to now enable global but disable the host. Or how to disable global, but enable the host. Or how to enable global, but for the host override the global settings for filerules, etc. My head doesn't grasp what combinations of enabled here, enabled there I need to set to do what I described with:

```
settings = global.imageio
if host.imageio.override:
    settings = host.imageio
```

iLLiCiTiT:
> If I'm overriding the host I don't care what the global settings are.

I don't think `enabled` from global should have more power over hosts' `enabled`. If Blender has disabled imageio workflow I would expect it is not used at all in Blender. Use-case: You want to disable imageio for TVPaint and Celaction but let it enabled with the same rules for the rest of hosts.

The override is to override rules, not enabled state.

fabiaserra:
I agree with @BigRoy on this one, I think the current code doesn't fit the logic requirements yet. Let me try summarize my use cases to see if we are aligned:

1. Set global color management that affects ALL hosts.
2. Granular color management per host - completely ignoring the global settings and exclusively using the host imageio settings (I think if we try to do hierarchical overrides on top of the global ones the logic gets too complex and unnecessary). Using this setting, one could "enable" the host settings and leave them empty to disable color manegement (although I'm with Roy and I don't see much value on doing that, but to each their own).

---

In Maya we are currently completely ignoring any of the new host/global imageio settings, is that correct? From [here](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/maya/api/lib.py#L3078) I see that `set_colorspace` is only using the legacy settings. Could we deprecate those settings and make the $OCIO env var drive the OCIO settings just like we do on all the others? I feel like maybe we are trying to support all the legacy color workflows on the DCCs and that's making our lifes much harder. As I said on my prior comment, I personally don't see the value of using the `set_colorspace` menu from an artist POV (maybe as a developer yeah to debug color management but that's it) and I wouldn't want those options to be baked on my scene workfile, I want to be able to drive color management globally through the environment of the host.

fabiaserra:
> I agree with @BigRoy on this one, I think the current code doesn't fit the logic requirements yet. Let me try summarize my use cases to see if we are aligned:
> 
> 1. Set global color management that affects ALL hosts.
> 2. Granular color management per host - completely ignoring the global settings and exclusively using the host imageio settings (I think if we try to do hierarchical overrides on top of the global ones the logic gets too complex and unnecessary). Using this setting, one could "enable" the host settings and leave them empty to disable color manegement (although I'm with Roy and I don't see much value on doing that, but to each their own).
> 
> In Maya we are currently completely ignoring any of the new host/global imageio settings, is that correct? From [here](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/maya/api/lib.py#L3078) I see that `set_colorspace` is only using the legacy settings. Could we deprecate those settings and make the $OCIO env var drive the OCIO settings just like we do on all the others? I feel like maybe we are trying to support all the legacy color workflows on the DCCs and that's making our lifes much harder. As I said on my prior comment, I personally don't see the value of using the `set_colorspace` menu from an artist POV (maybe as a developer yeah to debug color management but that's it) and I wouldn't want those options to be baked on my scene workfile, I want to be able to drive color management globally through the environment of the host.

On any case, could we expose the `app_groups` from the [OCIOEnvHook](https://github.com/ynput/OpenPype/blob/833f4a843b0250f6e505b44ac9e6cfc5333602dc/openpype/hooks/pre_ocio_hook.py#L11) to the global imageio settings so each can choose which hosts to affect that env change?

LiborBatek (approved):
Turning ```Color Management``` ImageIO activation toggle  sets OCIO config file to OpenPype location in ```maya``` preferences.

So works as expected.

image

image

openpype/settings/entities/schemas/projects_schema/schema_project_global.json
```diff
@@ -27,8 +27,39 @@
                     ]
                 },
                 {
-                    "type": "schema",
-                    "name": "schema_imageio_file_rules"
```
BigRoy, line 31: Why did this have to change? The old schema seemed to be the fine. Right?

openpype/pipeline/colorspace.py
```diff
@@ -344,6 +345,17 @@ def get_imageio_config(
     imageio_global, imageio_host = _get_imageio_settings(
         project_settings, host_name)
 
+    # check if host settings group is having enabled key
+    imageio_enabled = imageio_host.get("enabled")
+    if imageio_enabled  is None:
+        # it it does not have enabled key, use global settings
+        imageio_enabled = True
```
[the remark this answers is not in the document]
jakubjezek001: This will not work in case someone had already saved settings in project overrides and the key was missing at the moment of saving.

openpype/settings/entities/schemas/projects_schema/schema_project_global.json
jakubjezek001: I am processing the more explicitly communicated attributes. This will be followed with code changes, of course. 
Proble here was that global project schema was also using the template `schema_imageio_file_rules` but we dont need to have global override toggle on global shema.

BigRoy:
Just want to say the new layout do seem to make sense about what their intent is now using the **Override** checkboxes and the explicit **Enable color management in host** checkboxes for the hosts.

### Global ImageIO

afbeelding

### Host ImageIO

afbeelding

So thanks for that. I think it's now just a matter of still aligning the host logic to adhere to it, yes? Or when is this solid for a new full review?

openpype/pipeline/colorspace.py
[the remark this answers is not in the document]
fabiaserra: I don't get it, can you expand what you mean? if you need a fallback if key is missing you can always do `imageio_global.get("enabled", True)`?

jakubjezek001:
Great! Thanks @BigRoy for the feedback and also for your helps via asking those questions! Those had helped me a lot to focus on UX. Yeah, I had switched it to Draft for now and will let you know as soon as it is ready for the next review.

openpype/hooks/pre_ocio_hook.py
```diff
@@ -0,0 +1,60 @@
...
+    is_host_use_ocio_config_activated
+)
+from openpype.pipeline.template_data import get_template_data_with_names
+
+
+class OCIOEnvHook(PreLaunchHook):
+    """Set OCIO environment variable for hosts that use OpenColorIO."""
+
+    order = 0
+    app_groups = [
```
fabiaserra, line 14: would it be possible to expose these on the settings so if we leave it empty it defaults to all apps?

openpype/pipeline/colorspace.py
```diff
@@ -343,17 +349,35 @@ def get_imageio_config(
     imageio_global, imageio_host = _get_imageio_settings(
         project_settings, host_name)
 
+    # check if host settings group is having activate_host_color_management
+    # it it does not have activation key then use global settings
```
fabiaserra, line 353:
```suggestion
    # it it does not have activation key then default it to True so it uses global settings
```

tests/unit/openpype/pipeline/publish/test_publish_plugins.py
```diff
@@ -37,7 +37,7 @@ class TestPipelinePublishPlugins(TestPipeline):
     # files are the same as those used in `test_pipeline_colorspace`
     TEST_FILES = [
         (
-            "1d7t9_cVKeZRVF0ppCHiE5MJTTtTlJgBe",
+            "1YinxOToVyAd3-jAMFgVf7EWQa2x8Ma-O",
```
fabiaserra, line 40: hopefully GH doesn't think this is a secret 😋

fabiaserra (commented):
looking much better, thanks for all these changes!
Should we update Maya's `set_colorspace` (https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/maya/api/lib.py#L3177) on this PR as well to use the new settings?

openpype/settings/entities/schemas/projects_schema/schemas/schema_nuke_imageio.json
```diff
@@ -6,8 +6,14 @@
...
-            "type": "label",
-            "label": "<b style='color:red'>'Custom OCIO config path'</b> has deprecated. <br> If you need to set custom config, just enable and add path into 'OCIO config'. <br>Anatomy keys are supported.</i>."
+            "type": "boolean",
+            "key": "activate_host_color_management",
+            "label": "Enable Color Management in host"
+        },
+        {
+            "type": "boolean",
+            "key": "use_ocio_config",
+            "label": "Use OCIO config file in host"
```
BigRoy, line 16: I don't understand the difference between these two options. Aren't these two mutually exclusive?

What happens if:
- `use_ocio_config` is true but `activate_host_color_management` is false - how is it using a ocio config if we consider the host not color managed?
- `use_ocio_config` is false but `activate_host_color_management` is true - what is it doing then?

Aren't these the exact same setting?

openpype/hooks/pre_ocio_hook.py
```diff
@@ -0,0 +1,60 @@
...
+            anatomy_data=template_data,
+            anatomy=self.data["anatomy"]
+        )
+
+        if config_data:
+            use_config_path = is_host_use_ocio_config_activated(
+                project_name=self.data["project_name"],
+                host_name=self.host_name,
+                project_settings=self.data["project_settings"]
+            )
```
BigRoy, line 52: Isn't it weird that if the host is somehow explicitly disabling the use OCIO config setting that `get_imageio_config` still returns a data for that host? Not sure what's going on here.

BigRoy (commented):
Code is looking much better, but I still have some questions!

openpype/hooks/pre_ocio_hook.py
jakubjezek001: Hey @fabiaserra thanks for the feedback. Yeah, I was having the same idea, but then I was also thinking why would anybody touch those settings - since it is possible either disable the color management for the host or disable `use_ocio_config` attribute in the host's image.

openpype/pipeline/colorspace.py
```diff
@@ -343,17 +349,35 @@ def get_imageio_config(
...
+        "activate_host_color_management", True)
+
+    if not activate_host_color_management:
+        # if host settings are disabled return False because
+        # it is expected that no colorspace management is needed
+        log.info(
+            "Colorspace management for host '{}' is disabled.".format(
+                host_name)
+        )
+        return False
```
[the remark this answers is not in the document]
jakubjezek001: yeah, good catch

openpype/settings/entities/schemas/projects_schema/schemas/schema_nuke_imageio.json
jakubjezek001: > use_ocio_config is true but activate_host_color_management is false - how is it using a ocio config if we consider the host not color managed?

This is something I am not able to make more interactive in the current Settigs gui. We do not have connected signals so once attribut would effect the other to be hidden or disabled. But to answer your question if host colormanagement is deactivated and use of config is activated the priority is on host management and use config is secondary. 

> use_ocio_config is false but activate_host_color_management is true - what is it doing then?

This is more itneresting. Since some hosts are having available own build in ocio configs (nuke, maya, etc) we can either have activated just the colormanagedment and take advantage of file_rules and colorspaceData distribution or we can activate `use_ocio_config` and push our ocio convig files into the host wheather it is global or host overrides. 

Notice that some hosts are not havin this attribute available and activated can be only host's colorspace management. This is here for hosts which are either not colormanaged internally at all or have only its own configs. This way we can wrap it during publishing with out colormanagement and with use of File rules to match colorspace names found in our configs.

openpype/hooks/pre_ocio_hook.py
jakubjezek001: This should not happen that is why there is `if config_data:` on the scene. If the host is disabling Colorspace management then `get_imageio_config` should return None. Or do you see anywhere in the code it is different?

openpype/settings/entities/schemas/projects_schema/schemas/schema_nuke_imageio.json
BigRoy: It's funny that even with the explanation I don't understand. @fabiaserra any chance you do and can dumb it down for me?

openpype/hooks/pre_ocio_hook.py
jakubjezek001: But anybody can always add OCIO env-var into global or application environments in settings. This would be unnecessary configuration. Why would anybody want to add OCIO ever to host and disable host color management? Every new host integration should also add host name into the hook as hardcoded rather than adding it via the settings.

openpype/hooks/pre_ocio_hook.py
BigRoy: Again, same question. If the host color management is enabled I am still not sure why we need to check 'use ocio config'. If it is color managed then it is using OCIO. If it is disabled it's not using OCIO (or at least not in a way that OpenPype should be handling it?)

openpype/settings/entities/schemas/projects_schema/schemas/schema_nuke_imageio.json
fabiaserra: I understood the first part yeah, that unfortunately the settings logic doesn't have UI conditionals that allows us to disable/hide options depending on some other value, so while you are right that those options are mutually exclusive... we have no alternative but to give priority to one setting over the other. On this case, I understand that `activate_host_color_management` takes priority and so no matter what you have on `use_ocio_config`, the setting is ignored as long as it's `False`?

On the contrary, if `activate_host_color_management` is `True`... some hosts might use the internal (in my opinion legacy) color management settings to do some color management without an OCIO file.


Personally I would rather not support legacy color management workflows and try enforce new adopters of OP to just use OCIO on all hosts that support it, but I don't know if that's too much of an ask

openpype/hooks/pre_ocio_hook.py
fabiaserra: > Every new host integration should also add host name into the hook as hardcoded rather than adding it via the settings.
With this you are expecting that every user needs to fork OP to change this file and add their custom flavor as opposed to allowing the customization on the settings level (not needing to touch any code)

openpype/hooks/pre_ocio_hook.py
jakubjezek001: Again, same answer. Colormanagement activated is not the same as using OCIO config. You don't need to use our OCIO config to be able use our color management. Someone might want to use DCC's internal color management. We need to be explicit abou this one since those DCC's are also explicit. 

Without `use_ocio_config` someone might wonder - in DCC like Flame which is not supported custom OCIO config path,  where is the color config used from.

openpype/settings/entities/schemas/projects_schema/schemas/schema_nuke_imageio.json
jakubjezek001: Thanks @fabiaserra for the interpretation. Yes it is correct :heart_eyes:

> Personally I would rather not support legacy color management workflows and try enforce new adopters of OP to just use OCIO on all hosts that support it, but I don't know if that's too much of an ask

Yes I see your point and say yes you can configure your studio this way, but someone might want to use it differently and this should be universal enoughh to give an option to them.

Farther details: https://github.com/ynput/OpenPype/pull/4700#discussion_r1173505722

openpype/hooks/pre_ocio_hook.py
jakubjezek001: > With this you are expecting that every user needs to fork OP to change this file and add their custom flavor as opposed to allowing the customization on the settings level (not needing to touch any code)

Thanks @fabiaserra for your point. Filtering by `app_groups` was not the optimal design, it would need to be done by forking in case you would need to add a new **app group** via the system settings. I have changed the filtering to be per hosts. This way if any new host integration is added, it would be always done via code changes and hardcoding it makes more sense.

openpype/hooks/pre_ocio_hook.py
BigRoy: Should we maybe name it `set_ocio_config` instead of `use_ocio_config`, because if color management is enabled in e.g. maya then it would still have a an OCIO config. We would just not be setting it?

fabiaserra: +1 on `set_ocio_config`

openpype/hooks/pre_ocio_hook.py
fabiaserra: Fair enough, then we certainly need some color management documentation that explains this clearly and how would you go on supporting it on new hosts.

Apologies as this is a lazy question that I could probably test myself but have you checked how having the $OCIO defined plays out with the `use_ocio_config` and `activate_host_color_management` from this PR?

openpype/hooks/pre_ocio_hook.py
jakubjezek001: Done ;)

openpype/hooks/pre_ocio_hook.py
jakubjezek001: > but have you checked how having the $OCIO defined plays out with the use_ocio_config and activate_host_color_management from this PR?

Did and it worked as expected. Please test it as that would be also very valuable input.

jakubjezek001:
> Should we update Maya's set_colorspace ([develop/openpype/hosts/maya/api/lib.py#L3177](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/maya/api/lib.py?rgh-link-date=2023-04-17T15%3A10%3A50Z#L3177)) on this PR as well to use the new settings?

Yes I agree this should also be implemented ideally within the scope of this PR. I am not Maya user myself but perhaps @tokejepsen might help on this one?

jakubjezek001:
@fabiaserra regarding the user documentation, you are right - I will include it within this OR for sure. There was not point to do it now since the attributes were not established properly.

openpype/hosts/fusion/hooks/pre_fusion_ocio_hook.py
jakubjezek001: Conversation was continuing here https://github.com/ynput/OpenPype/pull/4700#discussion_r1173505722 and this can be resolved.

openpype/hooks/pre_ocio_hook.py
fabiaserra: Tested it in Maya and unfortunately, with `activate_host_color_management` (we could call it `enable_color_management_host` so it's more aligned with the label or viceversa?) I'm seeing the default Maya OCIO enabled, shouldn't that be disabled then?:
image

However, for the workflow that I'm interested in, this does seem to work fine:
image

fabiaserra: What's the recommended workflow for cleaning up deprecated settings from older versions to avoid getting warnings like:
```
*** WRN: >>> { DictImmutableKeysEntity }: [  project_settings/fusion/imageio Unknown keys in default values: "ocio"  ] 
*** WRN: >>> { DictImmutableKeysEntity }: [  project_settings/global/imageio/file_rules Unknown keys in studio overrides: "enabled"  ] 
*** WRN: >>> { DictImmutableKeysEntity }: [  project_settings/fusion/imageio Unknown keys in default values: "ocio"  ] 
*** WRN: >>> { DictImmutableKeysEntity }: [  project_settings/global/imageio/file_rules Unknown keys in studio overrides: "enabled"  ] 
*** WRN: >>> { DictImmutableKeysEntity }: [  project_settings/global/imageio/file_rules Unknown keys in studio overrides: "enabled"  ]
```

fabiaserra:
> > Should we update Maya's set_colorspace ([develop/openpype/hosts/maya/api/lib.py#L3177](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/maya/api/lib.py?rgh-link-date=2023-04-17T15%3A10%3A50Z#L3177)) on this PR as well to use the new settings?
> 
> Yes I agree this should also be implemented ideally within the scope of this PR. I am not Maya user myself but perhaps @tokejepsen might help on this one?

Just to shed more light on this, here I'm showing how `Set Colorspace` of Maya breaks this PR. Maya launches as expected with the $OCIO config but then when clicking Set Colorspace it reverts to the legacy config workflow:
pcoip_client_QrApE2wBsj

jakubjezek001:
> Just to shed more light on this, here I'm showing how Set Colorspace of Maya breaks this PR. Maya launches as expected with the $OCIO config but then when clicking Set Colorspace it reverts to the legacy config workflow:

@fabiaserra try it with the new changes please.

jakubjezek001:
Also I was doing some tests with @LiborBatek and we found out following:
- even `Set OCIO config file in host` is confusing him. It was comunicating to him that it will configure internal configs rather then our OCIO config to be set into Maya. We were thinking that perhaps `Push OCIO config file to host` would comunicate it better. What do you thing @BigRoy and @fabiaserra ?

- Also need to mention that I didnt know that Maya is having `Enable Color management` bool attribute internally. That is changing the situation. Again it was also confusing @LiborBatek. We need to come with something more clear for `Enable Color Management in host` attribute. It should be comunicating that this is activating Wrapper functions for the host and that those could be used just as a Translator of inner colormanagement into pipeline. My suggestions are:
   - `Activate Colormanagement wrapper for host`
   - `Pipeline Colormanagement feature`
   - `Activate Colormanagement Distribution`

BigRoy:
> even **Set OCIO config file in host** is confusing him. It was communicating to him that it will configure internal configs rather then our OCIO config to be set into Maya. We were thinking that perhaps **Push OCIO config file to host** would comunicate it better. 

_Push_ to me is more confusing. It sounds like it's going to move files around instead of applying/setting some settings in the host.

> Also need to mention that I didnt know that Maya is having Enable Color management bool attribute internally. That is changing the situation. Again it was also confusing @LiborBatek. We need to come with something more clear for Enable Color Management in host attribute. It should be comunicating that this is activating Wrapper functions for the host and that those could be used just as a Translator of inner colormanagement into pipeline. My suggestions are:
>
>     Activate Colormanagement wrapper for host
>     Pipeline Colormanagement feature
>     Activate Colormanagement Distribution

Actually no, we're now over-engineering things. If OpenPype sets "use color management" to True then it should force enable the setting in Maya. If OpenPype sets "use color management" to False then it should force disable the setting in Maya. To me "use color management" setting OpenPype is 1 to 1 that exact setting in Maya. They are not two different things.

---

I'm extremely sorry if I come across as a bit dumb, but I don't feel like I can really follow along with the different use cases.

To me each of the offered alternative labels is much more unclear about what it does - which is enabling the color management features for the host or not. But then again, maybe my confusion comes from the fact that I'm misinterpreting what it sets out to do in OpenPype settings because maybe to you they are not 1 to 1 the settings in Maya.

I literally can't see how we can ever manage to get things working if we intend to support any combination of:

- **Enable Color Management in Host**: Use the imageio settings from project settings in the host which is basically like _Enable OpenPype OCIO color management for Host_ according to your description?
- **Enable Host color management features**: tick on/off the Color Management setting in e.g. Maya?
- **Set OCIO config file in host**: Set Use OpenPype defined OCIO config paths (overriding the ones the host might come with) - this enabled is also what sets the `OCIO` var.

I feel like there's just so much overlap between all of these. Enabling OpenPype color management makes no sense (to me) if you disable the OCIO configs and Host color management features. Sure there might be host that are not using OCIO itself, but then _if OpenPype enables color management_ then it enables whatever logic we can come up with that works for that host.

We should take a step back and simplify this, no? I've tried to read over the combinations of explanations about why we need separate options but it's not clicking for me why we need that.

I just see these two options (which is basically a single boolean):
- OpenPype color management for host enabled: OpenPype tries to manage the host's color management settings. Whether the host has OCIO or not we try to build the integration in such a way that it behaves like OCIO color management, so that we can still say the output file that might not come from an OCIO-compatible application is `sRGB` space in this OCIO config so that we still know how to load it in an application that is OCIO-compatible.
- OpenPype color management for host disabled: OpenPype tries to not touch the host's color settings and it is assumed that any file incoming/outgoing requires no processing on OpenPype's side - we pass it on 1-to-1.

In the case of someone **enabling OpenPype to do color management in Maya** but they preferring Maya's color management setting to be OFF then still SOMEWHERE OpenPype should be able to (from Maya settings or some part of project settings) say this file is OCIO colorspace X when rendered which could just mean: _if not color managed, output files are assumed to be sRGB in the OCIO config_ but this **requires** to at least be aware of an OCIO config because the `sRGB` value must be available in the config then.

#### Documentation please

Also, I wanted to highlight that **no matter what route we go down for the settings** this is clearly an area that requires thorough documentation with concrete practical examples saying "If you enable this but disable this, then this happens" so we can all relate it to real world cases and have expectations about the resulting combination of settings. Thinking about it now, maybe a first step is writing some sort of "Unit Test" documentation that describes how we WANT it to work. So write documentation that defines how you WANT the settings to be and then from there define how we can actually programmatically make it happen so that it works that way.

fabiaserra:
@jakubjezek001 unfortunately I find the labels that you are suggesting with @LiborBatek more confusing than what they are already and as @BigRoy said, over-engineering and complicating it even further. Let me try summarize what are all the options that I see necessary and how it makes sense to me:

* `Global` -> `Enable Color Management`: Enable/disable the full scope of color management that OP supports. This is a global switch, if you turn it off, **ALL** color management in OP is disabled.
* `Global` -> `OCIO configs`: List of OCIO configs with priority from top to bottom (if top exists, overrides bottom).
* `Global` -> `Input Color Space rules (OCIO v1)`*: Matches exactly the same parameter options that Maya exposes (https://help.autodesk.com/view/MAYAUL/2023/ENU/?guid=GUID-631C665F-3092-4B2B-90B7-2A94158870C9, https://user-images.githubusercontent.com/4348536/233613888-840d50b6-4af3-418b-8f55-ebbfd9fc0af9.gif) so you can set rules by extension or file pattern matching. Each host should try its best to support these rules.
* `Host` -> `Enable Color management` (defaults to True):  Enable/disable the full scope of color management that OP supports on the host (as long as the global option is enabled). Most of the relevant hosts that OP supports already have OCIO support so this should just be a matter of making sure the $OCIO env var is set on the host env and affects the workfile launched, nothing else, it should either be OCIO or nothing at this point. The only cases where we try support a color management that's not based on the OCIO config should be hosts that don't support OCIO but that's it.
* `Host` -> `Override global OCIO configs` (defaults to False) + `List of OCIO config paths`: Same as the global options but specific to the hosts. Very rarely someone would ever need to touch this and it should be recommended against as it would make it very easy to propagate non-matching color spaces between hosts. If for any reason you want to use the OCIO config from the host you can set the value of the host path directly (i.e., in Maya you would set it like `<MAYA_RESOURCES>/OCIO-configs/Maya${MAYA_VERSION}-default/config.ocio`).
* `Host` -> `Override global input color space rules (OCIO v1)`* (defaults to False) + `List of color space rules`: Same as the global ones but specific to the host. Again, you'd hope the global settings handles all cases and rarely one should need to touch this.

\* `Input Color Space rules` will be deprecated with OCIO v2 as it will be integrated on the config itself so I have doubts whether we should even try to support it now. Maybe it should only be exposed on the hosts that have currently try to support this workflow.

> That's 8 unique combination of settings that we'd need to consider each time we work on anything colors related.

@BigRoy not really because a lot of the settings depend on the other ones: if `Global` -> `Enable Color Management` is `False`, all of the other options can be ignored. Same as if  `Host` -> `Enable Color management` is false, all the below options can be ignored. Of course with this logic it could easily get confusing if someone set the `Global` setting to `False` and left the `Host` to `True` so there should be a clear warning when that's the case that all color management has been disabled.

And yeah, as I already said on my prior comment, documentation is key here

jakubjezek001:
Thank you very much for your feedback, @BigRoy. I can see your points. In fact, without those use cases, you wouldn't be able to understand our perspective or the full extent of the required configurability. Perhaps I didn't communicate this adequately, so I apologize for the lack of clarity. We have numerous clients whose projects in production would be negatively impacted by the implementation you're suggesting.

In any case, we're heading to FMX now, so this will have to wait. I will later map out all the use cases we need to consider for the client's deployment and also include some user documentation. Hopefully, this will make things clearer. Thank you for your patience and the time you've invested in this - it has been incredibly helpful! 👍

jakubjezek001:
> Global -> Enable Color Management: Enable/disable the full scope of color management that OP supports. This is a global switch, if you turn it off, ALL color management in OP is disabled.

@fabiaserra yes I agree that this is the missing bit. This way we could refrain from adding the `Set custom config in host` and still maintain backward compatibility on ongoing projects.

Just to be more clear about File rules. It is used for inputs and outputs so for publishing and loading. But I agree that since the ocio v2 is supporting it natively we I will add `(OCIO v1 only)` to the category label. Also Global File Rules should be swithable for those productions which are already working with OCIO v2.

jakubjezek001:
See documentation for more info about the concept https://github.com/ynput/ayon-documentation/pull/42

tokejepsen:
Currently getting an error when launching OP with this PR.

Capture

tokejepsen (changes requested):
Error when launching.

BigRoy:
I did a quick playing review of this to see how it works. The settings themselves seem ok and the labels help quite a bit too understand what it does. Great work! I haven't done much in-depth testing on whether the settings worked for all cases but for Fusion it seemed fine, in Resolve it did nothing whatsoever even though it does have the settings (so Resolve is basically not implemented correctly?), Houdini did set the OCIO env var which is great and Maya also did but within the application I think it still triggered code from the old settings, see what I mentioned below.

Below are my notes:

---

1. Default settings need to be regenerated due to the Substance Painter host now existing in recent develop to fix this pop-up:
invalid_settings

---

2. Isn't it weird that by default in OpenPype globally the color management is disabled? That seems wrong?
image

I like that there's a little helpful documentation with the settings and a link pointing to more details. Nice work!

_The related documentation link does not work in the documentation by the way._

> To avoid potential issues, ensure that the OCIO configuration path is not an absolute path and includes at least 
the root token (Anatomy).

Note: Would love to see an example of anatomy based formatting in that linked documentation too.

---

3. In Maya Settings the OCIO config is not at the top. Which seems weird consistency wise?

image

---

4. Side note: It seems that with this PR the OpenPype settings UI has gotten **even** slower. Oh boy. 😭 

---

5. I get this error whenever I ctrl + F in the settings UI (not sure if related to this PR):
```
  File "E:\openpype\OpenPype\openpype\tools\settings\settings\window.py", line 358, in keyPressEvent
    self._search_dialog.set_root_entity(widget.entity)
  File "E:\openpype\OpenPype\openpype\tools\settings\settings\search_dialog.py", line 100, in set_root_entity
    self._model.set_root_entity(entity)
  File "E:\openpype\OpenPype\openpype\tools\settings\settings\search_dialog.py", line 182, in set_root_entity
    for child in get_entity_children(parent_entity):
  File "E:\openpype\OpenPype\openpype\tools\settings\settings\search_dialog.py", line 13, in get_entity_children
    return entity.values()
  File "E:\openpype\OpenPype\openpype\settings\entities\item_entities.py", line 47, in values
    raise AttributeError(self.attribute_error_msg.format("values"))
AttributeError: 'PathEntity' has no attribute 'values' if is not set as multiplatform
```

---

6. When in Maya I still see this:
```
# openpype.hosts.maya.api.lib : Setting default Maya OCIO v2 config
```

But I wouldn't expect that to happen. Due to `OCIO` env var being set it still forces the OCIO config - so that's good but I think there's still invalid logic. Which maybe is similar to what @fabiaserra mentioned [here](https://github.com/ynput/OpenPype/pull/4700#issuecomment-1517621467). At the very least I think there's still some work for maya to be done before this can be merged.

BigRoy (changes requested):
Some code notes aside of my previous review comment [here](https://github.com/ynput/OpenPype/pull/4700#issuecomment-1547539443)

openpype/hosts/flame/hooks/pre_flame_setup.py
```diff
@@ -47,6 +47,13 @@ def execute(self):
 
         imageio_flame = project_settings["flame"]["imageio"]
 
+        # get host imageio settings enabled key if exists
```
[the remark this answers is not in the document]
BigRoy: I suppose this is resolved?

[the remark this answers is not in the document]
fabiaserra: I don't believe so but if you all understand why if key isn't found is set to True to ensure backwards compatible it might be just me missing some context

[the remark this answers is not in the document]
BigRoy: > I don't believe so but if you all understand why if key isn't found is set to True to ensure backwards compatible it might be just me missing some context

Ah I see, you're referring to this comment?

```python
        # if key was not found, set to True
        # ensuring backward compatibility
```

Also not sure why that's the case either.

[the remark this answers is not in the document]
fabiaserra: yeah exactly, we should expand the docstring to explain it then

openpype/settings/entities/schemas/projects_schema/schema_project_maya.json
```diff
@@ -101,31 +99,57 @@
                     ]
                 },
                 {
-                    "key": "colorManagementPreference",
-                    "type": "dict",
-                    "label": "Color Management Preference (legacy)",
+                    "type": "collapsible-wrap",
+                    "label": "<b>[Deprecated] please migrate all to 'Workfile' and enable it.</b>",
```
BigRoy, line 103: Does this mean that with Workfile setting enabled it does not support Maya v1 color management (which I believe was also OCIO but used OCIO v1.0) and thus does not support Maya below version 2022?

BigRoy:
@jakubjezek001 What's needed to continue the effort on this PR? Would be a great one to get through this week maybe?

jakubjezek001:
> Isn't it weird that by default in OpenPype globally the color management is disabled? That seems wrong?
@BigRoy this is the only way we can secure backward compatibility for our clients. Since the activation will set some hosts to use our custom ocio config path the activated by default would crash some ongoing productions.

openpype/pipeline/colorspace.py
```diff
@@ -17,6 +17,8 @@
 
 log = Logger.get_logger(__name__)
 
+class cashed_data:
+    remapping: dict = None
```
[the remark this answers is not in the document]
BigRoy: Also, the typing will break py2 host compatibility

openpype/pipeline/colorspace.py
```diff
@@ -335,25 +342,60 @@ def get_imageio_config(
         anatomy_data = get_template_data_from_session()
```
BigRoy, line 342: I think the docstring for this function **Returns** is wrong? It's not returning `bool`,  right? It also never returns None.

openpype/pipeline/colorspace.py
```diff
@@ -437,17 +479,72 @@ def get_imageio_file_rules(project_name, host_name, project_settings=None):
...
+        cashed_data.remapping[host_name] = {
+            "from_native": {
+                input["host_native_name"]: rule["ocio_name"]
+                for rule in remapping_rules
+            }
+        }
 
-    return file_rules
+    return cashed_data.remapping[host_name]["from_native"].get(
+        host_native_colorspace_name)
```
BigRoy, line 547: Even without caching thes efunctions don't seem inheritly slow. Actually, the call would just be:

```python
return imageio_host_settings["remapping"]["rules"].get(host_name, {}).get("from_native", {}).get(host_native_colorspace_name)
```
It's basically only a dictionary lookup. Why cache it?

jakubjezek001: This would not work. We are creating the dictionary from list and there is not `from_native` or `to_native` subcategories in settings.

openpype/pipeline/colorspace.py
```diff
@@ -437,17 +477,24 @@ def get_imageio_file_rules(project_name, host_name, project_settings=None):
 
     # get file rules from global and host_name
     frules_global = imageio_global["file_rules"]
+    activate_global_rules = frules_global.get(
+        "activate_global_file_rules", False)
+
+    if not activate_global_rules:
+        log.info(
+            "Global File Rules are disabled."
+        )
```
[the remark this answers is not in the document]
jakubjezek001: @BigRoy or @fabiaserra what is your point on this?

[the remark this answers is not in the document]
fabiaserra: Yeah makes sense for file rules that you'd want to have host specific ones even if global ones are disabled

fabiaserra:
A bit out of scope but I was just trying to set some file rule settings and I noticed they aren't sortable, making it quite annoying to create, specially because when you click the `+` sign it appends it on the list instead of prepending... and you would usually create the rules from bottom to top instead of viceversa
image
could you make the setting sortable and default to prepend?

antirotor:
> could you make the setting sortable and default to prepend?

We would have to change data model for this (change it to list of dict) nad that would break all existing settings. I would leave this change to AYON where it would be easier to handle.

antirotor (approved):

openpype/settings/entities/schemas/projects_schema/schema_project_maya.json
jakubjezek001: It will also support v1 settings. The code will solve this.

openpype/hosts/maya/api/lib.py
```diff
@@ -3199,102 +3200,90 @@ def iter_shader_edits(relationships, shader_nodes, nodes_by_id, label=None):
...
-        root_dict = imageio["colorManagementPreference_v2"]
-    else:
-        root_dict = imageio["colorManagementPreference"]
+    ocio_v2_support = use_ocio_v2 = maya_version >= ocio_v2_maya_version
+
+    root_dict = {}
+    use_workfile_settings = imageio.get("workfile", {}).get("enabled")
+
+    if use_workfile_settings:
+        # TODO: deprecated code from 3.15.5 - remove
```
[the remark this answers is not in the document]
jakubjezek001: Lets remove this in future releases

PR #4707, opened 2023-03-24
Maya: Xgen fixes

tokejepsen:
## Changelog Description
Fix for Xgen extraction of world parented nodes and validation for required namespace.

## Testing notes:
1. Setup Xgen on geometry without namespace that parent to the world.
2. Publish should fail.
3. Add namespace to Xgen and geometry.
4. Publish should succeed.


LiborBatek (approved):
When trying to publish ```Xgen``` family from asset being local and not reffered it correctly fails validations on missing "Namespace"

image

When properly asset refferenced into the scene all is fine and get Published. So all works as expected.

image

LiborBatek (approved):
All works as expected....when local model used it correctly fails

image

When using referenced model all good and Publish works fine.

PR #4711, opened 2023-03-25
Maya: Refactor Extract Look with dedicated processors for maketx

### LINKED ISSUE #2337: Maya: Look extraction and colorspace file rules
## Problem

Say you have a scene with some textures and some of those textures are **png** with linear color data (bumps or whatever). When you extract look, file nodes with pngs and `Raw` colorspace will be set to `sRGB` because of the default Maya colorspace rules (if `Ingore CS File Rules` is not checked).

Our problem is this generic `attribute_values()` context manager that is setting original attributes after some operation.

https://github.com/pypeclub/OpenPype/blob/09272ab963ded58fa5006e494ed2300e94e29049/openpype/hosts/maya/api/lib.py#L307-L335

It is trying to correctly restore original attribute but because it is not aware of order of things, it will set colorspace first and then file path. But setting file path will trigger Maya CS file rules and so Maya will switch it to `sRGB`.

```python
# what it internally does and is wrong
cmds.setAttr("file1.colorSpace", "Raw")
cmds.setAttr("file1.fileTextureName", "/some/texture.png")
```

This is correct order of operations:

```python
cmds.setAttr("file1.fileTextureName", "/some/texture.png")
cmds.setAttr("file1.colorSpace", "Raw")
```

### LINKED ISSUE #2599: Maya looks: support for native Redshift texture format

BigRoy:
## Changelog Description

Refactor Maya extract look to fix some issues:

- [x] Allow Extraction with maketx with OCIO Color Management enabled in Maya.
- [x] Fix file hashing so it includes arguments to maketx, so that when arguments change it correctly generates a new hash
- [x] Fix maketx destination colorspace when OCIO is enabled
- [x] Use pre-collected colorspaces of the resources instead of trying to retrieve again in Extract Look
- [x] Fix colorspace attributes being reinterpreted by maya on export (fix remapping) - goal is to resolve #2337
- [x] Fix support for checking config path of maya default OCIO config (due to using `lib.get_color_management_preferences` which remaps that path)
- [x] Merged in #2971 to refactor MakeTX into TextureProcessor and also support generating Redshift `.rstexbin` files. - goal is to resolve #2599
- [x] Allow custom arguments to `maketx` from OpenPype Settings like mentioned [here](https://discord.com/channels/517362899170230292/1088411966135083030/1088508351610093701) by @fabiaserra for arguments like: `--monochrome-detect`, `--opaque-detect`, `--checknan`.
- [x] Actually fix the code and make it work. :) (I'll try to keep below checkboxes in sync with my code changes)
    - [x] Publishing without texture processor should work (no maketx + no rstexbin)
    - [x] Publishing with maketx should work
    - [x] Publishing with  rstexbin should work
- [x] Test it. (This is just me doing some test-runs, please still test the PR!)

## Additional info

~**Please note:** It's currently WIP draft and will not work as is - hound will likely go crazy too if it sees some of my "WIP commits with todos" in them.~

~I'm only opening as draft so others know there's work in progress plus maybe get some comments in about features that I might have missed that should be in here.~

Current state should be usable but it's quite a big refactor I've done relatively quickly so I'd say definitely give it a thorough look over the code to comment on anything that stands out and test the different test cases of the testing notes below.

## Testing notes:
1. Start Maya
2. Extract Looks with textures, considering cases like:
- Extract without making tx files (with and without maya color management enabled)
- Extract with making tx files (with and without maya color management enabled)
- Extract with making rstexbin (with and without maya color management enabled)
- Test in Maya 2020 (Py2) and Maya 2022 or 2023 (Py3)
- Colors spaces and colors should be preserved or correctly set for the published texture (and also be correctly preserved in the look workfile after publishing)
- Colors should render as intended from the lookdev, should have no color shifts
- Try setting `project_settings/maya/publish/ExtractLook/maketx_arguments` to additional arguments for `maketx` and make sure they are used during the maketx conversion.


openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -46,59 +56,302 @@ def find_paths_by_hash(texture_hash):
...
+
+        This function requires the `REDSHIFT_COREDATAPATH`
+        to be in `PATH`.
+
+        Args:
+            source (str): Path to source file.
+
+        """
+        if "REDSHIFT_COREDATAPATH" not in os.environ:
+            raise RuntimeError("Must have Redshift available.")
```
moonyuet, line 124: it seems that this code is not working to check if REDSHIFT_COREDATAPATH exists in the environment. 
image
But it works when I run with the separate scripts, when checking the environment. 
image

Is it a good idea to include this in the validator instead of the extract to validate whether there is the reshift setup in the user varaible.

BigRoy: You might need to restart OpenPype tray or your machine completely before OP picks up that variable.

Also, haven't tested the rstex code yet and I expect it to error. However, it should pick up the tool fine I believe.

Could you check the actual error message?

BigRoy: Fixed with: https://github.com/ynput/OpenPype/pull/4711/commits/35f76a2365d76aa97ca32a5591e831662dcd4029

moonyuet (dismissed):

iLLiCiTiT:
Can't test it, but codewise looks much better (and explicit) than previous state.

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -1,18 +1,24 @@
 # -*- coding: utf-8 -*-
 """Maya look extractor."""
+import logging
+from abc import ABCMeta, abstractmethod
+
```
fabiaserra, line 5: any reasoning behind the separation of imports and not having them alphabetically ordered?

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -133,6 +89,307 @@ def no_workspace_dir():
...
+        return TextureResult(
+            path=destination,
+            file_hash=texture_hash,
+            colorspace=colorspace,
+            transfer_mode=COPY
+        )
+
+    @staticmethod
+    def get_extension():
```
fabiaserra, line 192: how about making it a class variable `MakeRSTexBin.extension` instead? I generally try to avoid staticmethod https://google.github.io/styleguide/pyguide.html#217-function-and-method-decorators

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -133,6 +89,307 @@ def no_workspace_dir():
...
+
+    def apply_settings(self, system_settings, project_settings):
+        # Allow extra maketx arguments from project settings
+        extra_args_dict = (
+            project_settings["maya"]["publish"]
+            .get("ExtractLook", {}).get("maketx_arguments", {})
+        )
+        extra_args = []
+        for flag, value in extra_args_dict.items():
+            extra_args.append(flag)
```
fabiaserra, line 244: tiny but:
```suggestion
            extra_args.extend([flag, value])
```

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -133,6 +89,307 @@ def no_workspace_dir():
...
+
+        if not maketx_path:
+            raise AssertionError(
+                "OIIO 'maketx' tool not found. Result: {}".format(maketx_path)
+            )
+
+        # Define .tx filepath in staging if source file is not .tx
+        fname, ext = os.path.splitext(os.path.basename(source))
+        if ext == ".tx":
+            # Do nothing if the source file is already a .tx file.
```
fabiaserra, line 283: what if we changed the colorspace though, would we expect it to regenerate the .tx?

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -133,6 +89,307 @@ def no_workspace_dir():
...
+            # We can't rely on the colorspace attribute when not in color
+            # managed mode because the collected color space is the color space
+            # attribute of the file node which can be any string whatsoever
+            # but only appears disabled in Attribute Editor. We assume we're
+            # always converting to linear/Raw if the source file is assumed to
+            # be sRGB.
+            # TODO Without color management do we even know we can do
+            #      "colorconvert" and what config does that end up using since
+            #       colorconvert is a OCIO command line flag for maketx.
+            #       Also, Raw != linear?
```
fabiaserra, line 324: from https://openimageio.readthedocs.io/en/latest/maketx.html:
> --colorconvert <inspace> <outspace>
Convert the color space of the input image from inspace to tospace. If OpenColorIO is installed and finds a valid configuration, it will be used for the color conversion. If OCIO is not enabled (or cannot find a valid configuration, OIIO will at least be able to convert among linear, sRGB, and Rec709.

seems like we would only be able to do from `sRGB` to `linear` and not `Raw`, I'd change it to:
```
args.extend(["--colorconvert", "sRGB", render_colorspace])
```

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -133,6 +89,307 @@ def no_workspace_dir():
...
+        # Exclude these additional arguments from the hashing because
+        # it is the hash itself
+        args.extend([
+            "--sattrib",
+            "sourceHash",
+            texture_hash
+        ])
+
+        # Ensure folder exists
+        destination = os.path.join(staging_dir, "resources", fname + ".tx")
```
fabiaserra, line 348:
```suggestion
        destination = os.path.join(staging_dir, "resources",  "{}.tx".format(fname))
```

fabiaserra: although I'd probably just split this in two to avoid doing the `os.path.dirname` again later:
```
resources_folder = os.path.join(staging_dir, "resources")
if not os.path.exists(resources_folder):
    os.makedirs(resources_folder)

texture_filename = os.path.join(resources_folder, "{}.tx".format(fname))
```

fabiaserra: And I'd probably move the declaration of `texture_filename`/`destination` closer to where the variable is actually used:
```
texture_filename = os.path.join(resources_folder, "{}.tx".format(fname))
subprocess_args.extend(["-o", texture_filename])
```

openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_publish.json
```diff
@@ -996,6 +996,22 @@
...
+        },
+        {
+            "type": "dict",
+            "collapsible": true,
+            "key": "ExtractLook",
+            "label": "Extract Look",
+            "children": [
+                {
+                    "key": "maketx_arguments",
+                    "label": "Extra arguments for maketx command line",
```
fabiaserra, line 1008: 💯

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -133,6 +89,307 @@ def no_workspace_dir():
...
+        hash_args = ["rstex"]
+        texture_hash = source_hash(source, *hash_args)
+
+        # Redshift stores the output texture next to the input but with
+        # the extension replaced to `.rstexbin`
+        basename, ext = os.path.splitext(source)
+        destination = "{}{}".format(basename, self.get_extension())
+
+        self.log.debug(" ".join(subprocess_args))
+        try:
```
fabiaserra, line 177: on this case we don't mind if the destination file already existed?

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -441,134 +718,128 @@ def process_resources(self, instance, staging_dir):
...
+                         filepath,
+                         processors,
+                         staging_dir,
+                         force_copy,
+                         color_management,
+                         colorspace):
+        """Process a single texture file on disk for publishing.
+        This will:
+            1. Check whether it's already published, if so it will do hardlink
+            2. If not published and maketx is enabled, generate a new .tx file.
```
fabiaserra, line 794: also `rstexbin`?

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -133,6 +89,307 @@ def no_workspace_dir():
...
+            if self._has_arnold():
+                img_info = image_info(source)
+                color_space = guess_colorspace(img_info)
+                if color_space.lower() == "sRGB":
+                    self.log.info("tx: converting sRGB -> linear")
+                    args.extend(["--colorconvert", "sRGB", "Raw"])
+                else:
+                    self.log.info("tx: texture's colorspace "
+                                  "is already linear")
+            else:
```
fabiaserra, line 335: if using `.tx` most likely arnold would already be included but I think my ordering would be something like this:
```
if ext.lower() in {".png", ".jpg", ".jpeg"}:
    input_colorspace = "sRGB"
elif self._has_arnold():
    img_info = image_info(source)
    input_colorspace = guess_colorspace(img_info)
else:
    self.log.warning(
        "tx: cannot guess the colorspace, color conversion won't be available!"
    )
    continue

if input_colorspace.lower() == "sRGB":
    self.log.info("tx: converting %s -> %s." % (input_colorspace, render_colorspace))
    args.extend(["--colorconvert", "sRGB", render_colorspace])
```
And I'd probably try store the `{".png", ".jpg", ".jpeg"}` set as some constant variable somewhere like `NONLINEAR_FILE_FORMATS`

openpype/hosts/maya/plugins/publish/extract_look.py
BigRoy: I've actually refactored this to support flags without values too, which `--opaque-detect` technically is. I'm wondering whether the `dict` structure in settings is actually best and whether it might make more sense to just make it a list for strings so that if values would be needed one could just add `--flag-with-value`, `1024` as separate keys. Then that might also make it work if there are ever flags for `maketx` that take more than one value which e.g. `--colorconvert` already is.

Thoughts?

openpype/hosts/maya/plugins/publish/extract_look.py
BigRoy: I'm not sure how to regenerate from a `.tx` file. In that case I think you'd regenerate from another source format, right? Is this something being done where tx files are regenerated?

openpype/hosts/maya/plugins/publish/extract_look.py
BigRoy: > if using .tx most likely arnold would already be included

`.tx` is also supported by other renderers (or anything supporting OpenImageIO I believe?). I know that e.g. V-Ray can render `.tx` files.

---

> ```
> if ext.lower() in {".png", ".jpg", ".jpeg"}:
>     input_colorspace = "sRGB"
> ```

I'm fine with adding this. @moonyuet @iLLiCiTiT thoughts? Can we assume any of those formats to ALWAYS be sRGB or do you predict any issues? Note that this logic is only used if Color Management is disabled in Maya.

moonyuet: > > if using .tx most likely arnold would already be included
> 
> `.tx` is also supported by other renderers (or anything supporting OpenImageIO I believe?). I know that e.g. V-Ray can render `.tx` files.
> 
> > ```
> > if ext.lower() in {".png", ".jpg", ".jpeg"}:
> >     input_colorspace = "sRGB"
> > ```
> 
> I'm fine with adding this. @moonyuet @iLLiCiTiT thoughts? Can we assume any of those formats to ALWAYS be sRGB or do you predict any issues? Note that this logic is only used if Color Management is disabled in Maya.

I am not sure about png as someone claims that there is also 16-bit or 32-bit png despite the rare existence(http://www.libpng.org/pub/png/png-rgba32.html). But adding jpg and jpeg are definitely fine for me

openpype/hosts/maya/plugins/publish/extract_look.py
BigRoy: Good question - in my tests it did not overwrite the `.rstexbin` file if it exists and matched the source file but it did overwrite it if the input file had changed. I'm fine with keeping the logic as is actually.

openpype/hosts/maya/plugins/publish/extract_look.py
iLLiCiTiT: Or simplify it to just argument and list of parameters
```json
{
    "type": "list",
    "key": "maketx_args",
    "object_type": {
        "type": "dict",
        "children": [
            {
                "key": "argument",
                "label": "Argument",
                "type": "text"
            },
            {
                "key": "list",
                "label": "Parameters",
                "object_type": "text"
            }
        ]
    }
}
```

This allows 0-n parameters for an argument

BigRoy: I like it, thanks - will play with it.

BigRoy: So this doesn't work:
```
{
                    "key": "maketx_arguments",
                    "label": "MakeTX Arguments",
                    "collapsible": true,
                    "type": "list",
                    "object_type": {
                        "type": "dict",
                        "children": [
                            {
                                "key": "argument",
                                "label": "Argument",
                                "type": "text"
                            },
                            {
                                "key": "list",
                                "label": "Parameters",
                                "type": "text"
                            }
                        ]
                    }
                }
```
The parameters don't turn into a list and the other example I couldn't get to work.

openpype/hosts/maya/plugins/publish/extract_look.py
BigRoy: @moonyuet @antirotor @fabiaserra  What do we want to do in this case?

Previously we'd still rely on using the OCIO config that was set in Maya even if maya color management was disabled. I'd say that's wrong because whatever the OCIO config is then set to in Maya would basically be unused. However, without color management we can't seem to 'linearize' through the maketx.

[This is what Arnold does internally in the tx manager (click on the **"Expand to see script"**)](https://github.com/ynput/OpenPype/pull/4276#discussion_r1071519771) which whenever color management is disabled comes down to only doing this:

```python
        # promote 8-bit images to EXR half with DWAA compression to avoid quantization errors (#3919)
        if colorspace != 'linear' and tile_info['bit_depth'] <= 8:
            cmd += ['--format', 'exr', '-d', 'half', '--compression', 'dwaa']
```

It basically does no color conversion whenever color management is disabled in Maya. Should we be doing the same?

---

Actually, I've decided for now to mimic what Arnold does in tx manager and whenever color management is disabled to do *no* color conversion on `.tx` generation.

openpype/hosts/maya/plugins/publish/extract_look.py
BigRoy: Not really - this particular one was just missed. I do tend to keep separate builtins versus additional dependencies and local imports. But as far as I know that's not really based on any style guide, just used to doing it that way.

I've refactored it a bit now.

openpype/hosts/maya/plugins/publish/extract_look.py
moonyuet: > @moonyuet @antirotor @fabiaserra What do we want to do in this case?
> 
> Previously we'd still rely on using the OCIO config that was set in Maya even if maya color management was disabled. I'd say that's wrong because whatever the OCIO config is then set to in Maya would basically be unused. However, without color management we can't seem to 'linearize' through the maketx.
> 
> [This is what Arnold does internally in the tx manager (click on the **"Expand to see script"**)](https://github.com/ynput/OpenPype/pull/4276#discussion_r1071519771) which whenever color management is disabled comes down to only doing this:
> 
> ```python
>         # promote 8-bit images to EXR half with DWAA compression to avoid quantization errors (#3919)
>         if colorspace != 'linear' and tile_info['bit_depth'] <= 8:
>             cmd += ['--format', 'exr', '-d', 'half', '--compression', 'dwaa']
> ```
> 
> It basically does no color conversion whenever color management is disabled in Maya. Should we be doing the same?
> 
> Actually, I've decided for now to mimic what Arnold does in tx manager and whenever color management is disabled to do _no_ color conversion on `.tx` generation.

We can still run the tx conversion without color config and color convert, right? I think we should be doing the same - the tx conversion only if there's no color config.  
We need to warn or remind the users the color management disabled and there won't be color conversion when making tx.

BigRoy: Thanks @moonyuet that's indeed what I ended up doing.

fabiaserra: surprise surprise, docs not being accurate haha 

and yeah, good reasoning on replicating Arnold's behavior, my +1 on that

openpype/hosts/maya/plugins/publish/extract_look.py
fabiaserra: yeah same, I usually have three blocks on this order:
<standard libraries>
<third-party libraries>
<local libraries>

openpype/hosts/maya/plugins/publish/extract_look.py
fabiaserra: +1 on the list of strings

openpype/hosts/maya/plugins/publish/extract_look.py
fabiaserra: oh right yeah makes sense

openpype/hosts/maya/plugins/publish/extract_look.py
BigRoy: Thanks for these references. They are good references for #4713 but since I've been able to now avoid the color space guessing since conversion couldn't happen anyway with color management disabled I suppose it's not needed for this particular PR.

openpype/hosts/maya/plugins/publish/extract_look.py
iLLiCiTiT: I don't like list of strings because 99% user have not idea how to split them into separated arguments.

> The parameters don't turn into a list

I made mistake it should be 
```
{
    "type": "list",
    "key": "maketx_args",
    "object_type": {
        "type": "dict",
        "children": [
            {
                "key": "argument",
                "label": "Argument",
                "type": "text"
            },
            {
                "type": "list",
                "label": "Parameters",
                "object_type": "text"
            }
        ]
    }
}
```

fabiaserra: is that any different than the existing ffmpeg args? I'd argue that that 99% of the users that don't know how to split CLI arguments won't/shouldn't be touching any of the arguments

iLLiCiTiT: Current ffmpeg arguments must be converted to string and the process is started using `shell=True` because of that -> that cause more issues than you would think.

iLLiCiTiT: This `'"/path/to/custom/font/with/space myfont.ttf"'` is actually invalid argument because it does not have spaces or special characters it will be passed as string `"/path/to/custom/font/with/space myfont.ttf"`, meaning the quotes will be part of the value passed to process.

iLLiCiTiT: EDITED: It has space, but in that case you must know when the quotes should be there or not.

iLLiCiTiT: At the end, I don't care, I'm just trying to explain why I think it is better to do it as I proposed.

With the proposed settings you can easily explanain how the arguments should be passed and how it works, instead of trying to figure out how to guess what user meant by passing values there.

Not sure if it's clear from previous messages but shlex and processing of subprocess arguments does not work the same way for all platforms.

BigRoy: For now I've decided to go for your proposal @iLLiCiTiT thanks. Commits coming up!

BigRoy: That looks like this:
afbeelding

fabiaserra (approved):
approving, thank you @BigRoy !! 🥇

antirotor (approved):
looks great :pray:

LiborBatek (commented):
Did try different scenarios with / without ```make tx``` on, also tested color management with ```AcesCg``` and ```Linear/sRGB``` and finally without any color management in ```maya 2023``` and all working fine!

Also compared the output for any color shift and all good!

PR #4714, opened 2023-03-27
Maya: Defining plugins to load on Maya start - OP-4994

tokejepsen:
## Changelog Description
Feature to define plugins to load on Maya launch.

## Additional info
All default plugins should be included plus some extras. Studio can add their own plugins to the list if needed.

## Testing notes:
1. Enable `project_settings/maya/explicit_plugins_loading`.
2. Launch Maya and validate only the plugins requested are loaded.


openpype/settings/entities/schemas/projects_schema/schema_project_maya.json
```diff
@@ -10,6 +10,41 @@
...
+                    "key": "plugins_to_load",
+                    "label": "Plugins To Load",
+                    "object_type": {
+                        "type": "dict",
+                        "children": [
+                          {
+                              "type": "boolean",
+                              "key": "enabled",
+                              "label": "Enabled"
+                          },
```
BigRoy, line 37: I personally think it's very weird this has an enabled state. I'd have expected any entry to be in there to just be intended for auto loading.

openpype/settings/defaults/project_settings/maya.json
BigRoy, line 1: I would really try and refrain from having this massive list as defaults in there, we should leave it to the admin to add what they need instead. I'd say it makes more sense to only have those in there that we really expect to need. This list is crazy, right? :)

If we want to ease the burden for an admin we can add a little python script to the documentation which an admin could run from within maya which could return a copy-pasteable entry for the OpenPype settings to auto-set it based on the plug-ins loaded at that moment.

BigRoy (commented):
Nice feature, had some comments.

I want to add that this feature will not work for e.g. a Deadline submission, right? So once documentation gets added it might be good to note that the renderfarm submission will not adhere to this functionality.

tokejepsen:
> I want to add that this feature will not work for e.g. a Deadline submission, right? So once documentation gets added it might be good to note that the renderfarm submission will not adhere to this functionality.

Good point! Have noted in the docs.

website/docs/admin_hosts_maya.md
```diff
@@ -172,3 +172,12 @@ Fill in the necessary fields (the optional fields are regex filters)
...
 ![maya build template](assets/maya-build_workfile_from_template.png)
+
+## Explicit Plugins Loading
+You can define which plugins to load on launch of Maya here; `project_settings/maya/explicit_plugins_loading`. This can help improve Maya's launch speed, if you know which plugins are needed.
+
+By default only the required plugins are enabled. You can also add any plugin to the list to enable on launch.
+
+:::note technical
+When enabling this feature, the workfile will be launched post initialization no matter the setting on `project_settings/maya/open_workfile_post_initialization`. This is to avoid any issues with references needing plugins.
+:::
```
BigRoy, line 183: As far as I'm aware those issues with plug-ins don't exist actually if maya got started with the explicit file argument. It was only in rendering with Deadline due to the fact they run Maya with a custom command which opens the file manually after launch.

As such, I believe the issue with plug-ins initializing too late only exists due to "post initialization" and a `maya.exe "/path/to/workfile.ma"` call doesn't suffer from the same issues.

_Yes, it'd load plug-ins after the scene had opened. But as far as I'm aware in that particular case it auto-loads the plug-ins for a file just fine._

Are we sure post initialization is needed?

Anyway, not a big issue. Just wondering.

openpype/hosts/maya/startup/userSetup.py
```diff
@@ -13,23 +12,40 @@
...
+                    print("Loading " + plugin["name"])
+                    try:
+                        cmds.loadPlugin(plugin["name"], quiet=True)
+                    except RuntimeError as e:
+                        print(e)
+
+    cmds.evalDeferred(
+        _explicit_load_plugins,
+        lowestPriority=True
+    )
```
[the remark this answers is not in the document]
tokejepsen: The `settings` was already queried for the later project shelf; https://github.com/ynput/OpenPype/blob/53361fe9dfae15bd4a1f318561dd4ea8393ce353/openpype/hosts/maya/startup/userSetup.py#L49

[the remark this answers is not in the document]
BigRoy: Ah yes, you're right - you just moved it.

openpype/settings/entities/schemas/projects_schema/schema_project_maya.json
tokejepsen: I experimented with just using a `dict-modifiable` but the plugin names got cut off which is a limitation of the current settings. Not something we should be spending time on implementing due to settings being transitioned to Ayon.
Capture (1)

website/docs/admin_hosts_maya.md
tokejepsen: Will investigate. Think it was just something in my Maya `precautions` telling me to do this.

openpype/settings/entities/schemas/projects_schema/schema_project_maya.json
BigRoy: I meant more - it could just be a list of strings which are just plug-in names right? It doesn't actually need the toggle. Don't want to autoload then remove the entry? Or is that maybe adding too much burden on the administrator? Hmm, maybe it is.

tokejepsen: The idea was to provide all the standard plugins, then if the studio has custom plugins, they should be technically minded enough to add the plugin.

@mkolar @antirotor what do you think?

m-u-r-p-h-y:
There are several aspects of this feature

- Maya startup time (loading all default plugins will make Maya to start 3 times longer)
- some plugins behave like a virus and break scenes (Turtle for example)
- nontechnical artists suffer from missing plugins they need for work
- some plugins can break on farm rendering


to fulfill all these scenarios, without too much explicit list of plugins, we could have an approach of two lists

**whitelist**
to specify plugins that have to be loaded on Maya start

**blacklist**
to specify plugins forbidden to load


this way you will only have short lists and plugins not included in none of the two lists will behave according to the user settings in `maya\<version>\prefs\pluginPrefs.mel`


what do you think @tokejepsen @BigRoy

BigRoy:
> this way you will only have short lists and plugins not included in none of the two lists will behave according to the user settings in maya\<version>\prefs\pluginPrefs.mel
> 
> what do you think @tokejepsen @BigRoy

I think the logic will get more complicated. We'll need to _find_ the pluginPrefs that the maya version will load and still pick a moment we want to load or unload the plug-ins. The current logic I think where the admin wants to override the local machine's behavior for pluginPrefs I feel is much more explicit and simple to implement.

> nontechnical artists suffer from missing plugins they need for work

I can definitely see this happening for smaller studios or freelancers just installing or having their own tools installed - but maybe those are cases where the admin setting for forcing the explicit plugin load to be disabled to begin with?

I like the blacklisting idea. However I think if we're to **blacklist** plugins then I feel like we want to be even stricter about that than just not having them autoload - in most cases we don't ever want them to be loaded. Like having a scene save or scene open command running which validates blacklisted plug-ins are not loaded or used. E.g. even with turtle autoload disabled opening any scene with turtle would still 'infect' your scene because it forces it to load. When blacklisting I expect to be so strict that I likely don't even want that to happen? Thoughts? (Sounds like that might be a different feature?)

tokejepsen:
@m-u-r-p-h-y 
> Maya startup time (loading all default plugins will make Maya to start 3 times longer)

Just to clarify, this PR will only load by default the plugins Maya natively loads. The native plugins are enabled by default in the settings.

> blacklist
to specify plugins forbidden to load

Not sure how you would enforce this. If a reference requires a plugin to load, it'll try to load that plugin. Only environment manipulation would prevent a plugin from loading.

The clients request is to speed up Maya launch as far as I know, which would happen if you only load plugins you need. The main issue to tackle is how to expose that control to the studio in the settings.

Let me know if I'm missing anything?

website/docs/admin_hosts_maya.md
tokejepsen: Here's a case I cant get to work "pre initialization":

1. Load GPU cache.
2. Have the GPU cache plugin disabled on launch.

@BigRoy unless you have a solution to this, I'll leave it as needing "post initialization".

ClementHector:
If I understood correctly this feature is only for plugins by default maya? Otherwise it conflicts with the Tools settings in the shot attributes
There is also a pyblish plugin that validates plugins in the maya scene.
image

ClementHector:
> you can add any plugin to load explicitly on launch.

If you don't put any plugin it doesn't load anything?

tokejepsen:
> If you don't put any plugin it doesn't load anything?

By default the feature it disabled. If you enable it, the default plugins that are enabled would be the ones you get loaded when using `-noAutoloadPlugins`:
```
Loading AbcExport
Loading AbcImport
Loading cacheEvaluator
Loading deformerEvaluator
Loading GPUBuiltInDeformer
Loading objExport
Loading renderSetup
Loading stereoCamera
```
You can enable any of the Maya native plugins, but also add any third party plugin.

LiborBatek (approved):
All works as expected. Did test with different plugins set to load and all worked ok.

image

openpype/settings/entities/schemas/projects_schema/schema_project_maya.json
mkolar: The toggles make sense here. It makes it easier to compare multiple projects. You can define all of the available plugins in the default studio settings and when you're overriding it for a project, it it much easier to toggle on and off, rather than delete and then go hunting for what it was that you want to add back. Ending up in a different order than other project increasing the complexity of the search.

tokejepsen:

mkolar:
Considering this is really advanced feature only for admins that hopefully know what they are doing, I'd say that the list should be as minimal as possible. I mean if you want to be explicit about exactly what you are loading, you should be able to make your own list of plugins you want. 

So I personally vote to match the list of `-noAutoLoadPlugins` and let the admin set their own.

BigRoy:
> So I personally vote to match the list of -noAutoLoadPlugins and let the admin set their own.

Only downside to the settings UI for this is that it's quite a task to say add 20 plug-ins there. Click "+", then type left side, then set the toggle, etc. As the list grows it also tends to get a bit slower as you click the "+" too. Not a super big issue, but it's something to keep in mind.

Would be very nice if there was just a very simple way to say add these 10 plugins from my clipboard or alike.

openpype/hooks/pre_add_last_workfile_arg.py
```diff
@@ -44,10 +44,15 @@ def execute(self):
 
         # Determine whether to open workfile post initialization.
         if self.host_name == "maya":
-            key = "open_workfile_post_initialization"
-            if self.data["project_settings"]["maya"][key]:
+            keys = [
+                "open_workfile_post_initialization", "explicit_plugins_loading"
+            ]
```
BigRoy, line 49: Still feels weird that `explict_plugins_loading` needs to be checked here but I can understand why it's needed.

openpype/hosts/maya/hooks/pre_auto_load_plugins.py
```diff
@@ -0,0 +1,15 @@
...
+
+    # Execute before workfile argument.
+    order = 0
+    app_groups = ["maya"]
+
+    def execute(self):
+        maya_settings = self.data["project_settings"]["maya"]
+        if maya_settings["explicit_plugins_loading"]["enabled"]:
+            self.log.debug("Explicit plugins loading.")
+            self.launch_context.launch_args.append("-noAutoloadPlugins")
```
BigRoy, line 15: Is there no way that maybe from this plug-in, if we order it earlier than the "add last workfile arg" hook that we can just pass something along the launch context's or alike so that the other hook knows "FORCE POST LAUNCH" hook.

E.g. we could maybe set something like `force_no_workfile_arg=True`. We could then do the same from a prelaunch hook for the post initialization option so that it doesn't need to be specifically checked for the maya settings in the add last workfile args prelaunch hook.

I think it could clean up the code nicely.

BigRoy (dismissed):
Code looks good to me. There is still some quirkiness with where `explicit_plugins_loading` is being checked but I understand the reasoning. I just wish that maybe it wouldn't needed to be as much all over the place.

openpype/hooks/pre_add_last_workfile_arg.py
tokejepsen: Yup agreed. The alternative would be have duplicated code in a `pre_explicit_plugins_loading.py` hook?

openpype/hosts/maya/hooks/pre_auto_load_plugins.py
tokejepsen: Lets continue this conversation in the other comment.

openpype/hooks/pre_add_last_workfile_arg.py
BigRoy: Sort of, yes. I was thinking doing this:

```python
from openpype.lib import PreLaunchHook


class MayaLaunchWorkfilePostInitialization(PreLaunchHook):
    """Define whether open last workfile should run post initialize"""

    # Before AddLastWorkfileToLaunchArgs
    order = 9
    app_groups = ["maya"]
    
    def execute(self):
    
        # Ignore if there's no last workfile to start
        if not self.data.get("start_last_workfile"):
            return
    
        
        maya_settings = self.data["project_settings"]["maya"]
        enabled = maya_settings["open_workfile_post_initialization"]
        if enabled:
            # Force disable the `AddLastWorkfileToLaunchArgs`
            self.data.pop("start_last_workfile")
            
            self.launch_context.env["OPENPYPE_OPEN_WORKFILE_POST_INITIALIZATION"] = "1"
            
            
class MayaPreAutoLoadPlugins(PreLaunchHook):
    """Define -noAutoloadPlugins command flag."""

    # Before AddLastWorkfileToLaunchArgs
    order = 9
    app_groups = ["maya"]
    
    def execute(self):
    
        # Ignore if there's no last workfile to start
        if not self.data.get("start_last_workfile"):
            return
    
        
        maya_settings = self.data["project_settings"]["maya"]
        enabled = maya_settings["explicit_plugins_loading"]["enabled"]
        if enabled:
            # Force disable the `AddLastWorkfileToLaunchArgs`
            self.data.pop("start_last_workfile")
            self.launch_context.env["OPENPYPE_OPEN_WORKFILE_POST_INITIALIZATION"] = "1"
```

And if you don't want _that_ code duplication then at least you could merge these two together to still keep the logic close to the Maya integration and not mixed into the `AddLastWorkfileToLaunchArgs` pre launch hook.

tokejepsen: Done

BigRoy: Did my pseudocode work out of the box? Haha, that's crazy! 😄

tokejepsen: Lol, yeah.

openpype/hosts/maya/startup/userSetup.py
```diff
@@ -13,23 +12,40 @@
 
 print("Starting OpenPype usersetup...")
 
+settings = get_project_settings(os.environ['AVALON_PROJECT'])
+
+# Loading plugins explicitly.
+if settings["maya"]["explicit_plugins_loading"]["enabled"]:
+    def _explicit_load_plugins():
+        project_settings = get_project_settings(os.environ["AVALON_PROJECT"])
```
BigRoy, line 20: This shouldn't get the project settings again but use the project settings that were already acquired outside of this function.

openpype/hosts/maya/startup/userSetup.py
```diff
@@ -13,23 +12,40 @@
...
+        _explicit_load_plugins,
+        lowestPriority=True
+    )
 
 # Open Workfile Post Initialization.
 key = "OPENPYPE_OPEN_WORKFILE_POST_INITIALIZATION"
 if bool(int(os.environ.get(key, "0"))):
+    def _log_and_open():
+        print("Opening \"{}\"".format(os.environ["AVALON_LAST_WORKFILE"]))
+        cmds.file(os.environ["AVALON_LAST_WORKFILE"], open=True, force=True)
```
BigRoy, line 42: Would this be more readable like this?
```suggestion
        path = os.environ["AVALON_LAST_WORKFILE"]
        print("Opening \"{}\"".format(path))
        cmds.file(path, open=True, force=True)
```

openpype/hosts/maya/startup/userSetup.py
```diff
@@ -13,23 +12,40 @@
...
+        maya_settings = project_settings["maya"]
+        explicit_plugins_loading = maya_settings["explicit_plugins_loading"]
+        if explicit_plugins_loading["enabled"]:
+            for plugin in explicit_plugins_loading["plugins_to_load"]:
+                if plugin["enabled"]:
+                    print("Loading " + plugin["name"])
+                    try:
+                        cmds.loadPlugin(plugin["name"], quiet=True)
+                    except RuntimeError as e:
+                        print(e)
```
BigRoy, line 30: This if statement is redundant right? Since we're already checking it above so it would never get here if it was disabled?
```suggestion
        for plugin in explicit_plugins_loading["plugins_to_load"]:
            if plugin["enabled"]:
                print("Loading plug-in: " + plugin["name"])
                try:
                    cmds.loadPlugin(plugin["name"], quiet=True)
                except RuntimeError as e:
                    print(e)
```

Also tweaked the print statement to mention it's loading a plug-in.

openpype/hosts/maya/startup/userSetup.py
BigRoy, line 35: Should there be a comment above this about why we're deferring the loading of the plug-ins? Why couldn't we load directly?

openpype/hosts/maya/startup/userSetup.py
tokejepsen: Yup, done

openpype/hosts/maya/startup/userSetup.py
tokejepsen: Yup will do

openpype/hosts/maya/startup/userSetup.py
```diff
@@ -13,24 +12,41 @@
...
+        for plugin in explicit_plugins_loading["plugins_to_load"]:
+            if plugin["enabled"]:
+                print("Loading plug-in: " + plugin["name"])
+                try:
+                    cmds.loadPlugin(plugin["name"], quiet=True)
+                except RuntimeError as e:
+                    print(e)
+
+    # We need to load plugins deferred as loading them directly does not work
+    # correctly due to Maya's initialization.
```
BigRoy, line 30: Can we make this more explicit. What doesn't work? Why doesn't it work? Doesn't it work due to a specific plug-in trying to access the UI even though it doesn't exist yet? Doesn't it work because Maya crashes?

If we know what happens or what fails it means we can come back here in 5 years time and read the comment to test whether it's still a potential issue instead of just relying on it 'not working'.

What do you think?

tokejepsen: Sorry, I dont know why its not working, but I'm not sure spending time on figure that out, is time well spent.

BigRoy: How did you noticed it didn't work? What happened? It didn't load the plug-ins? It failed to start? Even hinting at what actually occurred makes more sense to me. It's just that by reading the comment I currently have no idea what to potentially look out for.

tokejepsen:
```suggestion
    # We need to load plugins deferred as loading them directly does not work
    # correctly due to Maya's initialization. Maya launch crashes when for
    # example enabling the plugin "gpuCache".
```

Cool, have added some more information.

BigRoy (approved):
Code changes look good to me. Didn't do a test run.

LiborBatek (approved):
Did run tests both in ```maya 2022``` and ```maya 2023``` both are working and plugins loaded only on demand/by OP settings.

So all fine!

PR #4715, opened 2023-03-27
Nuke, DL: Returning Suspended Publishing attribute

jakubjezek001:
## Changelog Description
Old Nuke Publisher's feature for suspended publishing job on render farm was added back to the current Publisher.

## Additional info
This is how it suppose to look in publisher:
image

## Testing notes for Nuke:
1. Open Nuke in favorite testing workfile
2. Open publisher on render instance
3. Set the instance to Farm render
4. Set `Suspended publish` (img) as activated
5. Publish to farm (start your webservice)
6. Notice that python job is set to `suspended`


BigRoy:
Should we instead add the attribute on the Submit Publish Job plug-in since the data is really only needed there to begin with?

64qam (approved):

jakubjezek001:
> Should we instead add the attribute on the Submit Publish Job plug-in since the data is really only needed there to begin with?

Yes, that is a good idea to move it into its plugin (dono why was it not my idea! ;))

jakubjezek001:
There is known bug in publisher which needs to be solved so Hosts filtering is working as expected. https://github.com/ynput/OpenPype/issues/4726

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -20,7 +20,38 @@
...
+
+    @classmethod
+    def get_attribute_defs(cls):
+        return [
+            BoolDef(
+                "suspend_publish",
+                default=False,
+                label="Suspend publish"
+            )
+        ]
```
BigRoy, line 51: Could you elaborate as to why we still ended up separating this into a plugin of its own and why it can't be part of the `ProcessSubmittedJobOnFarm` plug-in?

jakubjezek001: Yes that is good question I was asking my self at the point of discussion with @iLLiCiTiT . But see it this way. If it was part of the plugin then you would not be able to exclude this interface for hosts you wish to exclude. Say you want to have it only for Maya and not Nuke. If it was part of the plugin you would need to disable plugin for the host or do it with some filtering. This way it seems to be quite good since you can also disable the plugin completely (after d2adc691e39a867822aef300c2bb0061425472f7).

BigRoy: I can see the reasoning, but is this really the way we want to go about this going forward?

1. From a code perspective it just looks weird.
- It has process which does nothing
- It has an order which has no relevance whatsoever, put this at integrator order or collector order, the functionality is the same.
3. There are more plugins to process and thus it'll be slower.
4. We need to get the data from the instance using the class, so now we're ending up with these multiple classes in the same file. But now if we want to access these settings from another plug-in we'll need to get the data by string/name instead (but there's no method on the openpype publisher mixin to do so). Which ends up just turning weird because usually these plug-ins aren't meant to cross reference each other like that
- At the very least I'd expect the plug-in itself to push the setting value into the instance. Right?
5. It's confusing in settings currently I'd expect the setting to be part of the SubmitPublishJob plug-in since that's what it is changing for.

I feel like if we really want to go this far to just get the attribute to only be available for a certain host/family that it almost starts making sense to just have a families + hosts filter per attribute definition. But yeah, please don't do that either.

Aren't we now setting up a coding practice that'll end up just being freakishly weird among the codebase. 
This looks like a massive amount of lines to _just_ add a `suspend_publish` boolean toggle? What if someone ends up adding the same for five more attribute definitions; the code will easily grow 500 lines?

Anyway, let me ask another question. Why would we want to **exclude** this particular setting for a particular host?

jakubjezek001: I will let @iLLiCiTiT to explain you most of your points since this was more his intention to do it this way.

> I feel like if we really want to go this far to just get the attribute to only be available for a certain host/family that it almost starts making sense to just have a families + hosts filter per attribute definition. But yeah, please don't do that either.

There is also another level here which we might need to implement in future and it is User permissions at attributes. Just to scare you little bit more.

BigRoy: > It's more about families than hosts. Host filter is a "bonus".

Sure, but as I'm saying. If we're also working towards a feature where some attributes might not be available to a user due to their permissions, or alike I feel like having host+families filtering on the attribute is about the same level of complexity in code to add it to the attribute definition.

It'd mean that per plug-in we'd have much less code to maintain than creating separate plug-ins as is done now to 

> (Have it on nuke submitter).

Again, we're ending up in a situation where we're reimplementing a core simple feature per host and need to maintain that logic for each of those plug-ins. We want to find the way where we need the least amount of duplicated code + preferably a way that the code doesn't require tons of line to just get an attribute on the instance the way we need it.

This setting is related to whether a publish job is to be submitted. This is the plug-in doing that. It can provide that setting. Whether that setting should be "hidden" for some hosts, families, instance names, users or what have you I guess can be discussed, there will be opinions and there will be cases we need a particular attribute definition to be only available in certain cases. As such, this seems like something we should be implementing on the attribute definition. But creating separate pyblish plug-in classes just to define an attribute like this means we're really hacking the current system in a way that I'm pretty sure we don't want to maintain down the line. This will just turn crazy if suddenly half of the attributes on a Creator need to be shown only in specific cases or half of the attributes on another plug-in needs to be filtered by host and user.

I really want to urge and push now to brainstorm and design it the simplest way possible instead of setting an example that this is the way we should do it going forward.

Again, if we do this there's almost no going back without breaking backwards compatibility for how the values are stored on the instance.

> Question now is, if we want to change it back, or keep it as it is. I don't think current state causes any trouble. This is optional feature that will probably be unused by most of studios/artists. But it nice to have option to decide for which families and for which host you want to allow suspend of publish job.

If we add it now and the settings is there, we should assume it's used to ensure we keep an eye on potential backwards compatibility. If not, we're just saying that we're disregarding backwards compatibility completely. So if we put it there now it makes it harder to change later.

iLLiCiTiT: That is not discussion for this PR and it is not a change that will happen here. Like I've mentioned few times, with publisher we didn't took out any ability that was there before. It may look like it with deadline, but unfortunatelly at the same time publisher came out we've start to work on royal render which changed how it worked before publisher.

So please, stop with comments on how publish attributes should work in this PR, because we won't change it here and the change will take some time + the conversation here will be lost...

Now, we need to resolve the changes made here, which have to work with what we have available now, because it is requirement for out client.

iLLiCiTiT: I would personally rollback to this commit https://github.com/ynput/OpenPype/pull/4715/commits/9fbd84b0ee5c620b4992bb104b72033042719f78 and merge it.

jakubjezek001: Agree that this is the least conflicting temporary solution.

iLLiCiTiT (approved):
Proposed solution works as expected

PR #4716, opened 2023-03-27
Maya: bug fix for passing zoom settings if review is attached to subset

antirotor:
## Changelog Description
Fix for attaching review to subset with pan/zoom option.

## Additional info
Recent changes allowed publishing of reviews with pan/zoom enabled, but introduced bug in case review is attached to subset. This is providing fix

## Testing notes:
1. Create review attached to something (model)
2. Publish

Publish should proceed normally.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/collect_review.py
+++ b/openpype/hosts/maya/plugins/publish/collect_review.py
@@ -79,6 +79,7 @@ class CollectReview(pyblish.api.InstancePlugin):
                 data['review_width'] = instance.data['review_width']
                 data['review_height'] = instance.data['review_height']
                 data["isolate"] = instance.data["isolate"]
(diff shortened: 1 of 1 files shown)


BigRoy (commented):
Fix seems fine for me.

However, the logic for `reviewable_subset` definitely feels off. Instead of `for inst in instance.context` with incrementing i += 1 and then accessing that instance again through `instance.context[i]` feels completely invalid. As far as I know `instance.name` isn't enforced to be unique at all in pyblish so the logic here would even break if it finds more matches since the `i` index would be off.

It feels more like this should've been:

```python
reviewable_inst = next(inst for inst in context if inst.name == reviewable_subset)
data = reviewable_inst.data
```

Also `instance.context[i].data.update(data)` is completely redundant since it's changing that data dictionary already in-place.

PR #4722, opened 2023-03-27
Fix: Locally copied version of last published workfile is not incremented

kalisp:
### Fix 1
When copied, the local workfile version keeps the published version number, when it must be +1 to follow OP's naming convention.

### Fix 2
Local workfile version's name is built from anatomy. This avoids to get workfiles with their publish template naming.

### Fix 3
In the case a subset has at least two tasks with published workfiles, for example `Modeling` and `Rigging`, launching `Rigging` was getting the first one with the `next` and trying to find representations, therefore `workfileModeling` and trying to match the current `task_name` (`Rigging`) with the `representation["context"]["task"]["name"]` of a Modeling representation, which was ending up to a `workfile_representation` to `None`, and exiting the process.

Trying to find the `task_name` in the `subset['name']` fixes it.

### Fix 4
Fetch input dependencies of workfile.

Replacing https://github.com/ynput/OpenPype/pull/4102 for changes to bring this home.


openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
```diff
@@ -94,38 +96,52 @@ def execute(self):
...
+            subset
+            for subset in get_subsets(
+                project_name,
+                asset_ids=[asset_doc["_id"]],
+                fields=["_id", "name", "data.family", "data.families"],
+            )
+            if (
+                subset["data"].get("family") == "workfile"
                 # Legacy compatibility
                 or "workfile" in subset["data"].get("families", {})
```
BigRoy, line 113: Families should be a list right?
```suggestion
                or "workfile" in subset["data"].get("families", [])
```

openpype/modules/sync_server/sync_server.py
```diff
@@ -189,6 +205,166 @@ def _site_is_working(module, project_name, site_name, site_config):
...
+                force=True,
+                priority=99
+            )
+    sync_server.reset_timer()
+
+    # While representation unavailable locally, wait.
+    while not sync_server.is_representation_on_site(
+        project_name, workfile_representation["_id"], local_site_id
+    ):
+        sleep(5)
```
BigRoy, line 308: How heavy is the `is_representation_on_site` check? I'd expect that it might make much more sense to prior to this just get the "expected files list" from the sync server and periodically just check whether the file exists on disk (avoiding any database calls, etc.)? And then maybe after do a database check if needed? I'm asking because 5 seconds seems quite long to be an interval checking at if this is used in e.g. a prelaunch hook where the user is actively waiting for a result to happen.

Also, this while loop might need a maximum timeout too?

openpype/modules/sync_server/sync_server.py
```diff
@@ -189,6 +205,166 @@ def _site_is_working(module, project_name, site_name, site_config):
...
+            task_name,
+            host_name,
+            anatomy=anatomy,
+            template_key=template_key,
+            project_settings=project_settings,
+        ),
+        anatomy.templates[template_key]["file"],
+        workfile_data,
+        [extension],
+    )
```
BigRoy, line 342: When does this happen through the prelaunch hook? The prelaunch hook only continues download if there's no local workfiles yet, right? So why does it try incrementing against local workfiles? Or what am I missing?

openpype/modules/sync_server/sync_server.py
kalisp: I don't think we can reliably test presence of a file on the disk as some OSs will create a file there immediately and file changes until is fully downloaded. We could query DB for file size and compare it against file metadata from the disk. But can we get them reliably from all OSs and would it be actually faster? I don't feel that this would happen too often and it is quick check against indexed `_id`.

Was thinking about maximum timeout too, but what is reasonable value? 2 minutes, 5 minutes, 10?

openpype/modules/sync_server/sync_server.py
kalisp: Hm, I think this might not be necessary, removed it.
Maybe @Tilix4 will have an idea why it is there.

openpype/modules/sync_server/sync_server.py
kalisp: Added check for failed downloads, it could happen that downloading fails, so there is a limit how many tries should be attempted.

openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
```diff
@@ -94,38 +96,52 @@ def execute(self):
...
-            ),
-            None,
-        )
-        if not subset_id:
+            )
+        ]
+        if not filtered_subsets:
             self.log.debug(
-                'No any workfile for asset "{}".'.format(asset_doc["name"])
+                "No any subset for asset '{}' with id '{}'.".format(
```
BigRoy:
```suggestion
                "No subset for asset '{}' with id '{}'.".format(
```

openpype/modules/sync_server/sync_server.py
Tilix4: I'm not sure the highlighted code is relevant. But the idea is that it takes the last version number and increments. This way, you don't get confused if you're downloading the `v003` for the first time, the opened workfile is a `v004`, and not a `v001`.

BigRoy: > I'm not sure the highlighted code is relevant. But the idea is that it takes the last version number and increments. This way, you don't get confused if you're downloading the v003 for the first time, the opened workfile is a v004, and not a v001.

Yes, that's what I'd expect too - but the logic that was here originally compared it with versions on disk locally which doesn't trigger since the logic wouldn't start if there were local files.

openpype/modules/sync_server/sync_server_module.py
```diff
@@ -838,6 +838,14 @@ def get_local_settings_schema(cls):
 
         return ret_dict
 
+    def get_launch_hook_paths(self):
+        """Implementation for applications launch hooks."""
```
Tilix4, line 842: Missing docstring of return

openpype/modules/sync_server/sync_server_module.py
```diff
@@ -1045,9 +1053,20 @@ def reset_timer(self):
...
+        Args:
+            project_name (str)
+            representation_id (str)
+            site_name (str)
+            max_retries (int) (optional) - provide only if method used in while
+                loop to bail out
+        Raises:
+              (ValueError)  Only If 'max_retries' provided if upload/download
+        failed too many times to limit infinite loop check.
+        """
```
Tilix4, line 1069: Missing docstring of return

Tilix4 (changes requested):

openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
```diff
@@ -92,86 +94,55 @@ def execute(self):
...
 
-        while not sync_server.is_representation_on_site(
-            project_name, workfile_representation["_id"], local_site_id
-        ):
-            sleep(5)
-
-        # Get paths
-        published_workfile_path = get_representation_path(
-            workfile_representation, root=anatomy.roots
+        workfile_representation = sorted_workfile_representations[-1]
```
BigRoy, line 129: This could also be `max(workfile_representations, key=lambda d: d["context"]["version"])`

openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
```diff
@@ -92,86 +94,55 @@ def execute(self):
...
+            host_name,
+            project_name,
+            asset_name,
+            task_name,
+            published_workfile_path,
+            workfile_representation,
+            max_retries,
+            anatomy=anatomy,
+            asset_doc=asset_doc,
         )
```
BigRoy, line 146: Should we at least log prior or after this that's it downloading a workfile representation?

BigRoy (dismissed):
I've posted some small comments. Note that I'm not using sync server module myself so I haven't tested this and also can't comment on whether this works as expected for someone using the module.

I'm solely approving based on that I think the logic seems ok.
It feels a bit weird that the copy last published workfile launch hook is set for blender, photoshop, tvpaint and aftereffects but not for many other apps like Maya, Max, Nuke, Fusion and I wonder why that is.

I'm not sure if there already is, but I'd expect that to be customizable for the project or module maybe? And additionally have some documentation explaining the feature. I don't see documentation about this [here](https://openpype.io/docs/module_site_sync).

Tilix4 (changes requested):
Tested and it doesn't work. It does not fetch the last published file if missing locally, it also stucks the loop and OP hangs.

Also I got this error constantly:
```
  File "/home/felix/Documents/Dev/OpenPype/openpype/modules/sync_server/sync_server.py", line 512, in sync_loop
    except asyncio.exceptions.CancelledError:
AttributeError: module 'asyncio' has no attribute 'exceptions'
```

openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
```diff
@@ -1,16 +1,17 @@
 import os
-import shutil
-from time import sleep
 from openpype.client.entities import (
     get_last_version_by_subset_id,
     get_representations,
     get_subsets,
```
Tilix4, line 5: Remove unused
```suggestion
    get_representations,
```

kalisp:
> module 'asyncio' has no attribute 'exceptions'

That doesn't happen to me. What version of Python are you using? 3.9 as it should be on develop?

openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
```diff
@@ -92,86 +92,55 @@ def execute(self):
...
-        local_site_id = get_local_site_id()
-        sync_server.add_site(
-            project_name,
-            workfile_representation["_id"],
-            local_site_id,
-            force=True,
-            priority=99,
-            reset_timer=True,
-        )
+        sorted_workfile_representations = sorted(workfile_representations,
```
iLLiCiTiT, line 123: Representations under hero versions don't have `["context"]["version"]` so it would probably crash with `KeyError`.

openpype/modules/sync_server/sync_server.py
```diff
@@ -189,6 +202,133 @@ def _site_is_working(module, project_name, site_name, site_config):
...
+    project_doc = get_project(project_name)
+
+    # Get workfile data
+    workfile_data = get_template_data(
+        project_doc, asset_doc, task_name, host_name
+    )
+
+    extension = last_published_workfile_path.split(".")[-1]
+
+    project_settings = get_project_settings(project_name)
```
iLLiCiTiT, line 311: Should this function copy the file to workfile directory? It looks like too much for `download_last_published_workfile`.

openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
iLLiCiTiT: It could be like this:
```
workfile_representation  = max(
    filter(
        lambda r: r["context"].get("version") is not None,
        workfile_representations
    ),
    key=lambda d: d["context"]["version"]
)
```

Or more readable:
```
filtered_repres = filter(
    lambda r: r["context"].get("version") is not None,
    workfile_representations
)
workfile_representation  = max(filtered_repres, key=lambda r: r["context"]["version"])
```

BigRoy: > Could there be a hero version of workfile?

If it's a published workfile, I'd say so yes.

openpype/modules/sync_server/sync_server.py
```diff
@@ -189,6 +202,133 @@ def _site_is_working(module, project_name, site_name, site_config):
...
+    template_key = get_workfile_template_key(
+        task_name, host_name, project_name, project_settings
+    )
+
+    workfile_data["version"] = (
+        workfile_representation["context"]["version"] + 1)
```
iLLiCiTiT, line 317: Also when talking about copying, this doesn't look right either. Publish and Workfile may have different versioning (there could be workfile v100 but only published workfile v1.

kalisp: This gets triggered only if there isn't any workfile present at all. It looks only for published.

We are not syncing work workfiles. so even if there would be v100 in studio's `work`, it gets ignored.

openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
kalisp: Ok, used recommended filtering.

Tilix4:
Tested :ok:

Last thing is I'd run `black --line-length 79` to the files before merging, but this is a must have.

kalisp:
I reformatted docs to line lenght, all others should ok.

website/docs/module_site_sync.md
Tilix4, line 1: I'm pretty sure this has no point to be linted. I'd `git checkout HEAD~1 website/docs/module_site_sync.md`

Tilix4 (approved):
LGTM, only one suggestion for markdown doc.

iLLiCiTiT:
The prelaunch hook does not handle cases when there are no profiles in settings, or when `filter_profiles` does return `None` -> BTW I think that profiles should be empty to be skipped  by default.

Tilix4:
Found a bug: If you delete all versions of a workfile subset for an asset and try to launch it after, it tries to find the last published workfile (which has been deleted) and hangs forever. It should not try to fetch it and open a new empty file.

kalisp:
Fixed issue with no profiles returned.

BigRoy:
> I tried to delete all published workfiles for particular tasks and open again, works.

How did you delete it? Did you use the loader action from the standalone loader? Because maybe @Tilix4 did use that which I believe 'archives' certain publishes instead of actually removing them from the database?

Tilix4:
> > I tried to delete all published workfiles for particular tasks and open again, works.
> 
> How did you delete it? Did you use the loader action from the standalone loader? (OP tray > Loader) Because maybe @Tilix4 did use that which I believe 'archives' certain publishes instead of actually removing them from the database?

Indeed that's what I did. That was my guess too.

kalisp:
Yeah, I used Delete version action in Loader.

Tilix4:
> Yeah, I used Delete version action in Loader.

The `Delete Old Versions` one?

kalisp:
Yeah, thats the one. I deleted all versions of "rigging" task for `Visitor` and then opened Blender without issue.
python_l3RODrvLzu

openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
```diff
@@ -92,86 +104,83 @@ def execute(self):
...
-            priority=99,
-            reset_timer=True,
-        )
+        project_doc = get_project(project_name)
 
-        while not sync_server.is_representation_on_site(
-            project_name, workfile_representation["_id"], local_site_id
-        ):
-            sleep(5)
+        project_settings = get_project_settings(project_name)
```
iLLiCiTiT, line 160:
```suggestion
        project_doc = self.data["project_doc"]
        project_settings = self.data["project_settings"]
```

openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
```diff
@@ -92,86 +104,83 @@ def execute(self):
             )
             return
 
+        max_retries = int((sync_server.sync_project_settings[project_name]
+                                                            ["config"]
+                                                            ["retry_cnt"]))
+
         self.log.info("Trying to fetch last published workfile...")
 
-        project_doc = self.data.get("project_doc")
```
iLLiCiTiT, line 97: This can still be used

openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
```diff
@@ -92,86 +104,83 @@ def execute(self):
...
-        local_site_id = get_local_site_id()
-        sync_server.add_site(
-            project_name,
-            workfile_representation["_id"],
-            local_site_id,
-            force=True,
-            priority=99,
-            reset_timer=True,
-        )
+        project_doc = get_project(project_name)
```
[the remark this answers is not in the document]
iLLiCiTiT: Or better keep this https://github.com/ynput/OpenPype/pull/4722/files#r1168391870

openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
```diff
@@ -92,86 +104,83 @@ def execute(self):
...
+                                                            ["retry_cnt"]))
+
         self.log.info("Trying to fetch last published workfile...")
 
-        project_doc = self.data.get("project_doc")
         asset_doc = self.data.get("asset_doc")
         anatomy = self.data.get("anatomy")
 
-        # Check it can proceed
-        if not project_doc and not asset_doc:
```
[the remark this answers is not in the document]
iLLiCiTiT: Nevermind this has app groups filtering

Tilix4:
> Yeah, thats the one. I deleted all versions of "rigging" task for `Visitor` and then opened Blender without issue. python_l3RODrvLzu

I had this issue by publishing several versions on the same subset and deleting them. It failed until I published the same amount again and them went fine.

That being said, don't block the PR for this as it is not a very frequent use case. We could deal with it later.

iLLiCiTiT (approved):

PR #4727, opened 2023-03-28
After Effects: fix handles KeyError

friquette:
## Changelog Description
Sometimes when publishing with AE (we only saw this error on AE 2023), we got a KeyError for the handles in the "Collect Workfile" step. So I did get the handles from the context if ther's no handles in the asset entity.

### DIFF
--- a/openpype/hosts/aftereffects/plugins/publish/collect_workfile.py
+++ b/openpype/hosts/aftereffects/plugins/publish/collect_workfile.py
@@ -49,8 +49,8 @@ class CollectWorkfile(pyblish.api.ContextPlugin):
         asset_entity = context.data["assetEntity"]
         project_entity = context.data["projectEntity"]
 
-        handle_start = asset_entity["data"].get("handleStart") or context.data.get("handleStart")
-        handle_end = asset_entity["data"].get("handeEnd") or context.data.get("handleEnd")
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
Main reviewer is @kalisp, but this looks like a bug that should be fixed in project management rather than in AE. It would crash in any other host too, so I don't think this is right fix. It should not look into context at all, rather use `0` if handles are not set on asset.

openpype/hosts/aftereffects/plugins/publish/collect_workfile.py
```diff
@@ -49,14 +49,20 @@ def _get_new_instance(self, context, scene_file):
...
         instance_data = {
             "active": True,
             "asset": asset_entity["name"],
             "task": task,
             "frameStart": asset_entity["data"]["frameStart"],
             "frameEnd": asset_entity["data"]["frameEnd"],
-            "handleStart": asset_entity["data"]["handleStart"],
-            "handleEnd": asset_entity["data"]["handleEnd"],
+            "handleStart": handle_start,
+            "handleEnd": handle_end,
```
kalisp, line 65: I think we could use only `context.data` where these keys should be validated and cleaned up.

```suggestion         
             "frameStart": context.data["frameStart"],
            "frameEnd": context.data["frameEnd"],
            "handleStart": context.data["handleStart"],
            "handleEnd": context.data["handleEnd"],
```

PR #4729, opened 2023-03-29
3dsMax: render dialogue needs to be closed

moonyuet:
## Changelog Description
Make sure the render setup dialog is in a closed state for the update of resolution and other render settings

## Testing notes:

**Set Resolution**
1. Launch 3dsMax via Openpype
2.  Open render setting
3.  Access Openpype menu and click Set Resolution
4. Render setup dialog is automatically closed and when re-opening the setting, the resolution setting is updated according to the DB.

**Render Instance**
1. Launch 3dsMax via Openpype
2. Access Publish in Openpype menu
3. Create Render Instance
4. If the render setup dialog is opened, it will be automatically closed. The render setting will be updated if the dialog reopened.


LiborBatek (changes requested):
Did test with following results:

Testing Case 01

All works as expected and is ok.

Testing Case 02 (Creating Render Instance)

It closes ```Render Settings``` dialog during creation of ```Publish Instance``` but when reopened the Render Settings resolution is not changed properly but same.

moonyuet:
> Did test with following results:
> 
> Testing Case 01
> 
> All works as expected and is ok.
> 
> Testing Case 02 (Creating Render Instance)
> 
> It closes `Render Settings` dialog during creation of `Publish Instance` but when reopened the Render Settings resolution is not changed properly but same.

Nice catch. fixed this in the updated PR

LiborBatek (changes requested):
Still doesnt work for me. When Render Settings remains opened it get closed during creation of ```Render``` instance but wont get changed to DB resolution and wrongly set resolution still remains in Settings.

When the scene is saved and re-opened it got the correct resolution from DB.
However its forced by OP and not in control of the user. Which is wrong.

I can imagine we need sort of ```resolution validator``` or as in maya ```Render Settings Validator``` being it also ```Optional``` so user can decide if want to keep it or change the resolution to DB project values.

OP should not force the change of ```Resolution``` itself in background manner...

@antirotor what u think?

LiborBatek (approved):
When wrong resolution set ```Validate Resolution``` kicks in correctly

image

When resolution correct all publishes ok.
image

LiborBatek:
> Looks good, the only thing I would change is the actual message to the artist. I think `DB` term is something we shouldn't mention even though it is technically correct that we store those informations in database. I would maybe rephrase it to something like:
> 
> ```
> Resolution Setting not matching resolution set on asset or shot.
> ```
> 
> What do you think @LiborBatek

I think you are right and sounds good to me!

PR #4735, opened 2023-03-29
Maya: Validate Render Settings.

tokejepsen:
## Changelog Description
Fixes error message when using attribute validation.

## Testing notes:
1. Add `aiOptions.AASamples` to `project_settings/maya/publish/ValidateRenderSettings/arnold_render_attributes` without any value.
2. Create render in Maya and publish.
3. Validate error message is correct.


openpype/hosts/maya/plugins/publish/validate_rendersettings.py
```diff
@@ -275,16 +275,16 @@ def get_invalid(cls, instance):
...
-                cls.log.error(
-                    "Settings for {}.{} is missing values.".format(
-                        node, attribute
+            for node in data["nodes"]:
+                # Validate the settings has values.
+                if not data["values"]:
+                    cls.log.error(
+                        "Settings for {}.{} is missing values.".format(
+                            node, attribute
+                        )
```
BigRoy, line 284: Seems a bit much that we're reporting this _per node_ just because the setting value is empty?

It might make much more sense to just IGNORE the value already in `get_nodes` method if it doesn't have any values and then log a warning/error once that it's skipping that particular attribute?

That'd also help to even avoid querying for the nodes too.

tokejepsen: Yup good point.

BigRoy (dismissed):
Didn't test it but code looks good to me.

LiborBatek (changes requested):
Did set OP settings with custom attrib ```aiOptions.AASamples``` without any preset value
Note: I have already had a two existing attribs present tho.

image

Later giving me Warning on invalid value for AASamples. 

image

Im not sure what kind of "Correct error message" shoud I get tho....I guess its not correct now to expect "" basically nothing/ no value?!

LiborBatek:
Note> Its alswo weird it gave an error on ```Exr``` format even when correctly set and was practically matching OP Settings...

BigRoy:
> Note> Its alswo weird it gave an error on Exr format even when correctly set and was practically matching OP Settings...

I'm seeing the same thing. I think you intended to use a different attribute name:

```
defaultArnoldDriver.ai_translator
```

Instead of:

```
defaultArnoldDisplayDriver.ai_translator
```
I'm not sure wha the difference is. But onee is connected to `defaultArnoldRenderOptions.driver` and the other to `defaultArnoldRenderOptions.drivers[0]`. The connections are:

```
defaultArnoldDriver.message -> defaultArnoldRenderOptions.driver
defaultArnoldDisplayDriver.message -> defaultArnoldRenderOptions.drivers[0]

```

---

Here's quickly printing both attributes:

```python
from maya import cmds
for attr in ["defaultArnoldDriver.ai_translator", "defaultArnoldDisplayDriver.ai_translator"]:
    value = cmds.getAttr(attr)
    print(attr, value)
```

tokejepsen:
@LiborBatek could you give this another go please?

tokejepsen:
> Im not sure what kind of "Correct error message" shoud I get tho....I guess its not correct now to expect "" basically nothing/ no value?!

The idea is you should be able to repair this validator and "only" get error logs if something is missing. Fixed now.

LiborBatek (approved):
Works ok. 

When adding render attributes to OP Settings

image

 and they does not match in the maya scene,  
```Validator``` correctly kicks in and ```Repair``` function set values accordingly so ```Publish``` happens ok. 

If missing ```Value``` for particular ```Attribute``` in OP Setting, user been warned by ```Validate Error``` and is unable to pass validation until the value is put in the OP Settings.

Works ok.

tokejepsen:
@BigRoy you happy with this?

BigRoy (approved):
Seems fine.

LiborBatek (approved):
Have been tested and all fine working.

PR #4740, opened 2023-03-29
Global: Fix Extract Burnin + Colorspace functions for conflicting python environments with PYTHONHOME

BigRoy:
## Changelog Description

This fixes the running of openpype processes from e.g. a host with conflicting python versions that had `PYTHONHOME` said additionally to `PYTHONPATH`, like e.g. Houdini Py3.7 together with OpenPype Py3.9 when using Extract Burnin for a review in #3839 

This fix applies to Extract Burnin and some of the colorspace functions that use `run_openpype_process`

## Additional info

I've removed the explicit passing of `env = {}` because previously it didn't do anything either since it would still take `os.environ` in `run_openpype_process`. On Windows you'll need certain basic environment variables to be able e.g. run Python. If I literally pass an empty environment I got errors like this:

```python
DEBUG: Executing: run S:\openpype\OpenPype\openpype\scripts\otio_burnin.py C:/Users/Roy/AppData/Local/Temp/tmpud_zmted.json
INFO: Fatal Python error: _Py_HashRandomization_Init: failed to get random numbers to initialize Python
Python runtime state: preinitialized


  File "S:\openpype\OpenPype\.venv\lib\site-packages\pyblish\plugin.py", line 522, in __explicit_process
    runner(*args)
  File "S:\openpype\OpenPype\openpype\plugins\publish\extract_burnin.py", line 86, in process
    self.main_process(instance)
  File "S:\openpype\OpenPype\openpype\plugins\publish\extract_burnin.py", line 344, in main_process
    run_openpype_process(*args, **process_kwargs)
  File "S:\openpype\OpenPype\openpype\lib\execute.py", line 210, in run_openpype_process
    return run_subprocess(args, env=env, **kwargs)
  File "S:\openpype\OpenPype\openpype\lib\execute.py", line 160, in run_subprocess
    raise RuntimeError(exc_msg)
RuntimeError: Executing arguments was not successful: "(['S:\\openpype\\OpenPype\\.venv\\Scripts\\python.exe', 'S:\\openpype\\OpenPype\\start.py', 'run', 'S:\\openpype\\OpenPype\\openpype\\scripts\\otio_burnin.py', 'C:/Users/Roy/AppData/Local/Temp/tmpud_zmted.json'],)"Error:
Fatal Python error: _Py_HashRandomization_Init: failed to get random numbers to initialize Python
Python runtime state: preinitialized
```

## Testing notes:

1. Trigger the processes with conflicting environments as described.
2. Publishing reviews and colorspaces from existing DCCs should still work as expected.


openpype/lib/execute.py
```diff
@@ -198,7 +200,7 @@ def run_openpype_process(*args, **kwargs):
     args = get_openpype_execute_args(*args)
     env = kwargs.pop("env", None)
     # Keep env untouched if are passed and not empty
-    if not env:
+    if env is None:
```
iLLiCiTiT, line 203: This should stay
```suggestion
    if not env:
```

BigRoy:
Another option could've been starting the subprocess [python with the `-E` flag](https://docs.python.org/3/using/cmdline.html#cmdoption-E) but I believe it might then still influence any offspring subprocesses triggered from there that _should_ use PYTHONHOME and PYTHONPATH so I've decided to not go down that route.

We might want to double check the list of [Python Environment Variables](https://docs.python.org/3/using/cmdline.html#environment-variables) whether there are others we want to exclude by default?

openpype/lib/execute.py
BigRoy: Why? :)

That way you could never pass an empty `env` which is exactly why passing `env = {}` looked weird in ExtractBurnin and color spaces file, it basically ended up still taking the local environment.

I've looked over the code base and those were the only places this function was being used.

I'd argue that passing `env = {}` should correctly run the subprocess with the empty environment.

iLLiCiTiT: Main reason is backwards compatibility (we use the function in custom client addons). But even then, empty dictionary should be filled with os.environ and remove possible clashes.

BigRoy: Why would an empty env be filled with `os.environ`. 

Why deviate from how subprocess works? Subprocess `env=None` copies `os.environ`. Passing `env={}` does exactly what you'd expect. It passes an empty environment.

iLLiCiTiT: > It passes an empty environment.

The behavior is not the same in all Python versions (e.g. Maya 2020), sometimes is different per platform (I think it was 3.7.5?).

BigRoy: I have yet to see a case where that happens, but even if it were it'd be a Python bug? The subprocess docs state that if it is not None it must be:

> If env is not None, it must be a mapping that defines the environment variables for the new process; these are used instead of the default behavior of inheriting the current process’ environment. It is passed directly to Popen. This mapping can be str to str on any platform or bytes to bytes on POSIX platforms much like os.environ or os.environb

Also there are many examples online showing subprocess with passing an empty dict as en to have an empty environment.

iLLiCiTiT (approved):

mkolar (approved):

PR #4749, opened 2023-03-30
General: Extract burnin hosts filters

iLLiCiTiT:
## Changelog Description
Removed hosts filter from ExtractBurnin plugin. Instance without representations won't cause crash but just skip the instance. We've discovered because Blender already has review but did not create burnins.

## Additional info
We came to conclusion that we don't know why it had hosts filters at all, only possible reasons were that missing hosts don't have review at all, so we don't want to show the plugin in plugins list and the plugin crashes if passed instance does not have representations but both reasons are not much relevant.

Hosts that were not in list `blender`, `celaction`, `houdini`, `max`, `resolve` and `unreal`.

## Testing notes:
1. Validate the code
2. Try to run publishing with some relevant instances in each missing host
- [ ] Celaction
- [ ] Houdini
- [ ] 3dsMax
- [ ] Resolve
- [ ] Unreal
3. They should not crash and should not create unwanted output


jakubjezek001 (approved):
LGTM

PR #4753, opened 2023-03-30
AYON: Vendorize ayon api

iLLiCiTiT:
## Changelog Description
Vendorize ayon api into openpype vendor directory. The reason is that `ayon-python-api` is in live development and will fix/add features often in next few weeks/months, and because update of dependency requires new release -> new build, we want to avoid the need of doing that as it would affect OpenPype development.

## Additional info
Ayon api is now added into separated vendor folder which is added to `sys.path` and `PYTHONPATH` when `ayon_start.py` is used.

## Testing notes:
1. AYON mode should still work

### DIFF
--- a/ayon_start.py
+++ b/ayon_start.py
@@ -96,6 +96,19 @@ else:
 sys.path.append(_dependencies_path)
 _python_paths.append(_dependencies_path)
 
+# -------------------------------------------------
+# Temporary solution to add ayon_api to python path
+# -------------------------------------------------
+# This is to avoid need of new build & release when ayon-python-api is updated.
+ayon_dependency_dir = os.path.join(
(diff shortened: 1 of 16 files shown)


BigRoy:
Admittedly this PR made me laugh a little. 😄 

I feel like this is just a tiny look into the future once everything is split into add-ons for each DCC in a separated codebase, each of which will likely have frequent contributions and changes with _also_ the core still growing and changing over time. I wonder if this just already shows that managing it separately is actually too big of a hurdle to keep the ship moving fast enough (at the current rate that core + apps are changing in conjunction still). Just look at the amount of PRs involving multiple hosts.

So I had a little laugh that we're now trying to avoid a test-run of actually developing these things in separation and start merging the into one big thing - which I'm fine with by the way. Just a fun little mind game.

PR #4757, opened 2023-03-30
Global: Merge review tags

### LINKED ISSUE #4569: Merge all review tags into one
Currently we have 3 different review tags. `ftrackreview`, `shotgridreview` and `kitsureview`.
I don't see a reason why we would have 3 different review tags instead of just one.

Maybe there are some extreme edge-case where a production uses two of theses review tools but I have never heard about it myself.
Thoughts?

[cuID:[OP-5182](https://app.clickup.com/t/861mer10j)]

BigRoy:
## Changelog Description

This removes the tags `kitsureview`, `ftrackreview` and `shotgridreview` and instead refactors it to just be the `review` tag instead.

## Additional info

Consider this a **WIP refactor** to quickly identify the areas of code this might touch.

⚠️ Note: I'm not sure how to refactor the `ftrackreview-mp4` naming, etc. so there are some todos that I added to look into.

There are also these strings that I saw in the codebase:
```
ftrackreview-image  Images reviewable in the browser
ftrackreview-mp4    H.264/mp4 video reviewable in browser
ftrackreview-webm   WebM video reviewable in browser
```
Which I think are names needed by the ftrack API maybe so I suppose it's needed to at least in the end get those names for the ftrack component names.

This fixes #4569 

## Testing notes:

1. Test the touched areas of code, it will mostly be related to "reviews" as both stills and videos.
2. Test publishing with production trackers and without:
- [ ] Shotgrid
- [ ] Kitsu
- [ ] Ftrack

### DIFF
--- a/openpype/hosts/flame/plugins/publish/extract_subset_resources.py
+++ b/openpype/hosts/flame/plugins/publish/extract_subset_resources.py
@@ -326,6 +326,7 @@ class ExtractSubsetResources(publish.Extractor):
             # HACK: `ftrackreview_withLUT` will result only in `ftrackreview`
             if (
                 "thumbnail" in unique_name
+                # TODO: this should not be hardcoded to ftrack
                 or "ftrackreview" in unique_name
             ):
                 repr_name = unique_name.split("_")[0]
--- a/openpype/hosts/flame/startup/openpype_babypublisher/modules/ftrack_lib.py
+++ b/openpype/hosts/flame/startup/openpype_babypublisher/modules/ftrack_lib.py
@@ -127,6 +127,7 @@ class FtrackComponentCreator:
         file = os.path.basename(file_path)
         _n, ext = os.path.splitext(file)
 
+        # TODO: This should not be hardcoded to ftrack
         name = "ftrackreview-mp4" if "mov" in ext else "thumbnail"
 
         component_data = {
@@ -136,6 +137,7 @@ class FtrackComponentCreator:
             "location": location
         }
 
+        # TODO: this should not be hardcoded to ftrack
         if name == "ftrackreview-mp4":
             duration = data["duration"]
             handles = data["handles"]
--- a/openpype/hosts/fusion/plugins/publish/render_local.py
+++ b/openpype/hosts/fusion/plugins/publish/render_local.py
@@ -63,7 +63,7 @@ class Fusionlocal(pyblish.api.InstancePlugin,
 
         # review representation
         if instance.data.get("review", False):
-            repre["tags"] = ["review", "ftrackreview"]
+            repre["tags"] = ["review"]
 
     def render_once(self, context):
         """Render context comp only once, even with more render instances"""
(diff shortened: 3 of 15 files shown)


iLLiCiTiT:
I actually think that `"review"` and `"ftrackreview"` tags are not interchangeable. Tag `"review"` is telling "this is reviewables representation", and tag `"ftrackreview"` say "this should be uploaded as reviewable on ftrack"`, but I agree that they should not be added automatically in host extractors. In other words you may want create review but not integrate it to ftrack.

Separation for different addons is matter of decision, that could explain @mkolar .

BigRoy:
> I actually think that "review" and "ftrackreview" tags are not interchangeable. Tag "review" is telling "this is reviewables representation", and tag "ftrackreview" say "this should be uploaded as reviewable on ftrack", but I agree that they should not be added automatically in host extractors. In other words you may want create review but not integrate it to ftrack.

This is exactly the discussion that was going on in https://github.com/ynput/OpenPype/issues/4569 so I'd say definitely add your arguments there to discuss with both @EmberLightVFX and @Tilix4 but of course I'd also love more input on that from @mkolar if he can take a look.

iLLiCiTiT:
Considering this comment https://github.com/ynput/OpenPype/issues/4569#issuecomment-1511986990 can we update this PR?

BigRoy:
> Considering this comment [#4569 (comment)](https://github.com/ynput/OpenPype/issues/4569#issuecomment-1511986990) can we update this PR?

I'm a bit unsure how to proceed in that scenario. That comments seem to hint that it might just be a matter of changing the tag `review` to `transcode` but keeping the others (`kitsureview`, `ftrackreview`, `shotgridreview`) separate and not merge them.

As such it might be easier to revert (or make a new PR that is only about changing the `review` family to `transcode`.

To me it makes more sense maybe to refactor the tags to:

- `review` -> `transcode`
- `kitsureview` -> `upload-kitsu`
- `ftrackreview` -> `upload-ftrack`
- `shotgridreview` -> `upload-shotgrid`

But then still, what does the `transcode` tag really do in the end? Is there still a way to decide which file to `transcode` and which file makes its way to kitsu, ftrack or shotgrid, etc.

I feel like we're missing a concrete example of how we we want the system to behave to decide what the best refactor approach would be. No?

iLLiCiTiT:
As I'm reading it, all what would change is that we would rename the tags. To be honest that doesn't look like usefull change, it will cause a lot of troubles, especially with studio addons. Also I'm not sure if the naming is right. But I agree that changes here are agains the decision so new PR would be better. To be honest I think `ftrackreview` is better than `upload-ftrack` and `transcode-source` would be better than `transcode`. As `transcode` is the same cryptic name as `review`.

mkolar:
Gentlemen I have to say that considering the conversation on all fronts i don't think this is actually adding anything or solving any problem. We could revisit at some point, but it will not me mergable in this state. 

Let's close this and build a proper case of why it should be done in the first place

Issue #4750, opened 2023-03-30
Maya: Validate Model Name definition source deprecated in AYON


antirotor:
**Problem**
In **Maya** `Validate Model Name` validator uses name definition sourced either from file or database. There is a bug with that logged as #2596 but it won't work in AYON like this anyway and its use is very limited. We should either deprecate the whole validator or at least change it's behavior so the definitions are taken from the Settings.

I vote for deprecation/removal.

[cuID:[OP-5488](https://app.clickup.com/t/863gc41n4)]

PR #4758, opened 2023-03-30
Maya: Remove deprecated Validate Model Name

BigRoy:
## Changelog Description

Remove unused and deprecated Validate Model Name plug-in.

## Additional info

Fixes issue #4750

## Testing notes:

1. Fires of hell should have burned this legacy 🔥 
2. But nothing of publishing `model` family in maya should break. 🧯


tokejepsen:
This was in use recently by a client and received a bug fix; https://github.com/ynput/OpenPype/pull/4539, so we cant remove it completely. The main issue is with having it reference external files for data used for settings.
We'll consult with client about maybe moving external data into settings.

BigRoy:
Would you like me to refactor this to take the `json` from settings instead?

tokejepsen:
> Would you like me to refactor this to take the json from settings instead? Or will we ignore until client decision is made on approach?

Lets hold off on development until client replies and we figure out how to handle it.

PR #4766, opened 2023-04-03
Workfiles: keep Browse always enabled

kalisp:
## Changelog Description
Browse might make sense even if there are no workfiles present, actually in that case it makes the most sense (eg. I want to locate workfile from outside - from Desktop for example).

## Testing notes:
1. Browse button in Workfiles should be enabled even in new tasks where no workfiles are present

### DIFF
--- a/openpype/tools/workfiles/files_widget.py
+++ b/openpype/tools/workfiles/files_widget.py
@@ -617,14 +617,24 @@ class FilesWidget(QtWidgets.QWidget):
         ext_filter = "Work File (*{0})".format(
             " *".join(self._get_host_extensions())
         )
+        dir_key = "directory"
+        if qtpy.API in ("pyside", "pyside2", "pyside6"):
+            dir_key = "dir"
+
+        workfile_root = self._workfiles_root
+        # Find existing directory of workfile root
+        #   - Qt will use 'cwd' instead, if path does not exist, which may lead
+        #       to igniter directory
+        while workfile_root:
+            if os.path.exists(workfile_root):
+                break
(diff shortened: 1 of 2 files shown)


BigRoy:
With this change the browse button is always enabled right? So does it still make sense to even keep the line around where it's setting the state?

openpype/tools/workfiles/files_widget.py
```diff
@@ -379,7 +379,7 @@ def _update_asset_task(self):
 
             # Disable/Enable buttons based on available files in model
             has_valid_items = self._workarea_files_model.has_valid_items()
-            self._btn_browse.setEnabled(has_valid_items)
+            self._btn_browse.setEnabled(True)
```
iLLiCiTiT, line 382:
```suggestion
```

iLLiCiTiT: > With this change the browse button is always enabled right? So does it still make sense to even keep the line around where it's setting the state?

I've checked the code and it's true.

iLLiCiTiT:
NOTE: When the workarea does not exist, it will use first existing directory. In extreme cases, when tha workdir path would look like this: `P:/projects/kuba_each_case/Shots/sq01/sq01_sh0010/work/compositing`, it could end up in `P:/projects/kuba_each_case` if `Shots` does not have any folders yet.
- this is probably not an issue, just pointing out

64qam (dismissed):

kalisp:
Kuba's changes work even on nonexistent folder (it creates it and browse is started from there).

64qam (approved):

PR #4768, opened 2023-04-03
Global: Improve speed of Collect Custom Staging Directory

BigRoy:
## Changelog Description

Improve speed of Collect Custom Staging Directory.

## Additional info

Collect Custom Staging Directory made publishing with many instances very slow, this resolve that issue.

- No requerying of project settings in `get_custom_staging_dir_info`
- Only deepcopy anatomy data once we need to edit and use it.

## Testing notes:

1. Publish with and without a custom staging dir


kalisp (approved):
Works as previous implementation, according to reading the code it should be faster.

PR #4724, opened 2023-03-28
Maya: Remove pymel dependency

BigRoy:
## Changelog Description

Refactors code written using `pymel` to use standard maya python libraries instead like `maya.cmds` or `maya.api.OpenMaya`

## Additional info

This should fix #2353 and would also [resolve the issues with Maya 2020 and recent Mtoa versions](https://github.com/ynput/OpenPype/issues/3233)

## Testing notes:

1. Test all the areas of the code this PR touches:

- [x] Set up render publishing with a few rendersetup layers with attribute overrides on different attribute types, e.g. single channel float (`translateX`), vector (`translate`) and maybe some other attributes from render settings. Publishing should be fine.
- [x] cbId generation on scene save should still work.
- [x] Load audio should work
- [x] Load image plane should work (this might've actually already been broken as I've seen it reported a few times)
- [x] Load references should work, especially check some referencing with the attached groups (default) with a few different family types, e.g. model, pointcache, look, rig, setdress
- [x] Collecting reviews and extracting playblasts + thumbnails should still work as expected.
    - [x] Review still works with audio as expected. I believe original logic came from @tokejepsen so maybe you could proofread the code or mention what might come up as an issue.
- Trigger these bunch of validations:
    - [x] [Validate Rig Output Ids](https://github.com/BigRoy/OpenPype/blob/dbb0f37fa3cc1762f72f473a7f97657200a9fc6d/openpype/hosts/maya/plugins/publish/validate_rig_output_ids.py)
    - [x] [Validate Model no namespaces](https://github.com/BigRoy/OpenPype/blob/dbb0f37fa3cc1762f72f473a7f97657200a9fc6d/openpype/hosts/maya/plugins/publish/validate_no_namespace.py)
    - [x] [Validate Mesh No Overlapping UVs](https://github.com/BigRoy/OpenPype/blob/dbb0f37fa3cc1762f72f473a7f97657200a9fc6d/openpype/hosts/maya/plugins/publish/validate_mesh_overlapping_uvs.py)
    - [x] [Validate Mesh Arnold Attributes](https://github.com/BigRoy/OpenPype/blob/dbb0f37fa3cc1762f72f473a7f97657200a9fc6d/openpype/hosts/maya/plugins/publish/validate_mesh_arnold_attributes.py) - note this plug-in is only available if your current renderer is arnold
[...]

PR #4771, opened 2023-04-04
Tray Publisher: Audio

tokejepsen:
## Changelog Description
Add initial support for publishing `.wav` files.

## Additional info
Had to ingest audio for #4724 and editorial ingesting was too complicated so ended in up just making the plugins necessary to publish `.wav` files.

Only initial working version so all comments are welcomed.

## Testing notes:
1. Launch Tray Publisher.
2. Publish Audio with a `.wav` file.

### DIFF
--- a/openpype/hosts/traypublisher/plugins/create/create_audio.py
+++ b/openpype/hosts/traypublisher/plugins/create/create_audio.py
@@ -1,0 +1,51 @@
+import os
+
+from openpype.lib.attribute_definitions import FileDef
+from openpype.pipeline import (
+    CreatedInstance,
+    CreatorError
+)
+from openpype.hosts.traypublisher.api.plugin import TrayPublishCreator
+
+
+class AudioCreator(TrayPublishCreator):
+    """Creates audio instance on asset."""
+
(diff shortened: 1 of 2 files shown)


openpype/hosts/traypublisher/plugins/create/create_audio.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class AudioCreator(TrayPublishCreator):
+    """Creates audio instance on asset."""
+
+    identifier = "io.openpype.creators.traypublisher.audio"
+    label = "Audio"
+    family = "audio"
+    description = "Publish audio files."
+    extensions = [".wav"]
```
BigRoy, line 18: Does it make sense to also add some other popular audio formats like `mp3`? Maybe even `aiff`, `flac`, `aac`, `ogg`?

iLLiCiTiT:
How is this different from simple creators?

tokejepsen:
> How is this different from simple creators?

Difference is that I didnt know about the simple creator :)

BigRoy:
Had a quick look. The documentation can definitely benefit from a Simple Creator section. I couldn't find how to do this from the documentation at least.

PR #4773, opened 2023-04-04
General: Anatomy templates formatting

iLLiCiTiT:
## Changelog Description
Added option to format only single template from anatomy instead of formatting all of them all the time. Formatting of all templates is causing slowdowns e.g. during publishing of hundreds of instances.

## Testing notes:
- Anatomy formatting should work as before
- It should be possible to format single template in anatomy
```python
# Example
from openpype.lib import Anatomy
from openpype.pipeline.template_data import get_template_data_with_names

project_name = "*fill*"
asset_name = "*fill*"
task_name = "*fill*"
fill_data = get_template_data_with_names(
    project_name,
    asset_name,
    task_name)
folder_template = anatomy.templates_obj["work"]["folder"]
folder_path = folder_template.format(fill_data)
print(folder_path)
```

### DIFF
--- a/openpype/lib/path_templates.py
+++ b/openpype/lib/path_templates.py
@@ -263,10 +263,10 @@ class TemplatesDict(object):
             ))
 
     def __getitem__(self, key):
-        return self.templates[key]
+        return self.objected_templates[key]
 
     def get(self, key, *args, **kwargs):
-        return self.templates.get(key, *args, **kwargs)
+        return self.objected_templates.get(key, *args, **kwargs)
 
     @property
     def raw_templates(self):
--- a/openpype/lib/path_templates.py
+++ b/openpype/lib/path_templates.py
@@ -256,7 +256,8 @@ class TemplatesDict(object):
         elif isinstance(templates, dict):
             self._raw_templates = copy.deepcopy(templates)
             self._templates = templates
-            self._objected_templates = self.create_objected_templates(templates)
+            self._objected_templates = self.create_objected_templates(
+                templates)
         else:
             raise TypeError("<{}> argument must be a dict, not {}.".format(
                 self.__class__.__name__, str(type(templates))
--- a/openpype/lib/path_templates.py
+++ b/openpype/lib/path_templates.py
@@ -280,8 +280,21 @@ class TemplatesDict(object):
     def objected_templates(self):
         return self._objected_templates
 
-    @classmethod
-    def create_ojected_templates(cls, templates):
+    def _create_template_object(self, template):
+        """Create template object from a template string.
+
+        Separated into method to give option change class of templates.
+
+        Args:
+            template (str): Template string.
+
+        Returns:
+            StringTemplate: Object of template.
+        """
+
+        return StringTemplate(template)
+
+    def create_ojected_templates(self, templates):
         if not isinstance(templates, dict):
             raise TypeError("Expected dict object, got {}".format(
                 str(type(templates))
@@ -297,7 +310,7 @@ class TemplatesDict(object):
             for key in tuple(item.keys()):
                 value = item[key]
                 if isinstance(value, six.string_types):
-                    item[key] = StringTemplate(value)
+                    item[key] = self._create_template_object(value)
                 elif isinstance(value, dict):
                     inner_queue.append(value)
         return objected_templates
--- a/openpype/lib/path_templates.py
+++ b/openpype/lib/path_templates.py
@@ -256,7 +256,7 @@ class TemplatesDict(object):
         elif isinstance(templates, dict):
             self._raw_templates = copy.deepcopy(templates)
             self._templates = templates
-            self._objected_templates = self.create_ojected_templates(templates)
+            self._objected_templates = self.create_objected_templates(templates)
         else:
             raise TypeError("<{}> argument must be a dict, not {}.".format(
                 self.__class__.__name__, str(type(templates))
@@ -294,7 +294,7 @@ class TemplatesDict(object):
 
         return StringTemplate(template)
 
-    def create_ojected_templates(self, templates):
+    def create_objected_templates(self, templates):
         if not isinstance(templates, dict):
             raise TypeError("Expected dict object, got {}".format(
                 str(type(templates))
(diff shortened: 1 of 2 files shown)


BigRoy:
I gave this a go by doing this:

```python
template_data = copy.deepcopy(instance.data["anatomyData"])
template = anatomy.templates["publish"]["folder"]
publish_folder = template.format(template_data)
```

Which results in:
```python
KeyError: 'root'
```

Even if I did:
```python
template_data["root"] = anatomy.roots
```
I get the same key error. 

Somehow I'm not getting any sensible stack trace in Maya.

```python
from openpype.pipeline.template_data import get_template_data
from openpype.pipeline.context_tools import get_current_project, get_current_project_asset, get_current_task_name, get_current_host_name
from openpype.settings import get_system_settings


project_entity = get_current_project()
asset_entity = get_current_project_asset()
host_name = get_current_host_name()
system_settings = get_system_settings()

anatomy_data = get_template_data(
    project_entity, asset_entity, task_name, host_name, system_settings
)
anatomy = Anatomy(project_entity["name"])
folder_template = anatomy.templates["work"]["folder"]
folder_path = folder_template.format(anatomy_data)
print(folder_path)
```

Results in:
```python
# Error: 'root'
# #   File "<maya console>", line 19, in <module>
# # KeyError: 'root'
```

BigRoy:
By the way, I suspect this is a perfect feature/enhancement that should get its own pytests to run? Is that something for @kalisp ?

iLLiCiTiT:
I had to do more changes than I expected. It could deserve more testing now.

openpype/lib/path_templates.py
```diff
@@ -280,8 +280,21 @@ def templates(self):
...
+        Args:
+            template (str): Template string.
+
+        Returns:
+            StringTemplate: Object of template.
+        """
+
+        return StringTemplate(template)
+
+    def create_ojected_templates(self, templates):
```
BigRoy, line 297: Since we're changing this line of code anyway, should we fix the typo?

```suggestion
    def create_objected_templates(self, templates):
```
Might need to refactor other areas of the code however.

openpype/pipeline/anatomy.py
```diff
@@ -908,7 +928,17 @@ def _keys_to_dicts(self, key_list, value):
             return {key_list[0]: value}
         return {key_list[0]: self._keys_to_dicts(key_list[1:], value)}
 
-    def _rootless_path(self, result, final_data):
+    @classmethod
+    def rootless_path_from_result(cls, result):
+        """Calculate rootless path from formatting result.
+
+        Args:
+            result (StringTemplate): Result of StringTemplate formatting.
```
BigRoy, line 936: Shouldn't the type here be `TemplateResult`?

openpype/pipeline/anatomy.py
```diff
@@ -606,6 +607,32 @@ def normalized(self):
...
+            data (dict[str, Any]): Formatting data for template.
+
+        Returns:
+            AnatomyTemplateResult: Formatting result.
+        """
+
+        anatomy_templates = self.anatomy_templates
+        if not data.get("root"):
+            data = copy.deepcopy(data)
+            data["root"] = anatomy_templates.anatomy.roots
```
BigRoy, line 630: I suspect that this `format` per string attribute is also called when doing a `anatomy.format()` call? Which makes it look like now we're deep copying the input data per string attribute if the data didn't include the `root` key which will be much slower since we have quite a few string attibutes. Right?

I'm also looking at the code that changed here but it didn't seem to do this "automatic ingesting of root" before either - it just required it. Should we keep it that explicit?

iLLiCiTiT: > Which makes it look like now we're deep copying the input data per string attribute if the data didn't include the root key which will be much slower since we have quite a few string attibutes. Right?

It does copy of data only if `"root"` is not in source data, when you call `anatomy.format(...)` it will fill the `"root"`. It is not explicit because of that.

Could be made explicit and faster by changing to this:
```python
if data.get("root") != anatomy_templates.anatomy.roots:
    data = copy.deepcopy(data)
    data["root"] = anatomy_templates.anatomy.roots
```

openpype/pipeline/anatomy.py
iLLiCiTiT:
```suggestion
            result (TemplateResult): Result of StringTemplate formatting.
```

openpype/lib/path_templates.py
```diff
@@ -256,17 +256,17 @@ def set_templates(self, templates):
         elif isinstance(templates, dict):
             self._raw_templates = copy.deepcopy(templates)
             self._templates = templates
-            self._objected_templates = self.create_ojected_templates(templates)
+            self._objected_templates = self.create_objected_templates(templates)
```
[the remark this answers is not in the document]
iLLiCiTiT:
```suggestion
            self._objected_templates = self.create_objected_templates(
                templates)
```

openpype/pipeline/anatomy.py
BigRoy: If `anatomy.format()` internally already pre-sets the `root` key then I'm fine with the current logic. I was just worried that now suddenly it'd start setting it for each string it's formatting.

BigRoy:
I'm still getting the same error as [here](https://github.com/ynput/OpenPype/pull/4773#issuecomment-1496220908).

For now also tried it in Houdini to see if it provides more stack trace but somehow the error is the same:

```python
  File "tool_2", line 20, in <module>
KeyError: 'root'
```

BigRoy:
Thanks. Yes, this works:

```python
from openpype.pipeline.template_data import get_template_data
from openpype.pipeline.context_tools import get_current_project, get_current_project_asset, get_current_task_name, get_current_host_name
from openpype.settings import get_system_settings


project_entity = get_current_project()
asset_entity = get_current_project_asset()
host_name = get_current_host_name()
system_settings = get_system_settings()

anatomy_data = get_template_data(
    project_entity, asset_entity, task_name, host_name, system_settings
)
anatomy = Anatomy(project_entity["name"])
folder_template = anatomy.templates_obj["work"]["folder"]
folder_path = folder_template.format(anatomy_data)
print(folder_path)
```

BigRoy (approved):

PR #4776, opened 2023-04-04
CreateContext: Added option to remove Unknown attributes

iLLiCiTiT:
## Changelog Description
Added option to remove attributes with UnkownAttrDef on instances. Pop of key will also remove the attribute definition from attribute values, so they're not recreated again.

## Additional info
This can be used for fix of instance data during collection after adding to context. The changes will be marked as changes for save but not yet saved to scene. The pop won't remove keys that have valid attribute definitions.

## Testing notes:
1. Create instance with unknown data in creator attributes
2. Remove the key when `CreatedInstance` object is created using `pop`
```python
# Example - this is fake mockup, and in this case 'new_key' should not show up in UI and should not be stored to scene
instance_data["creator_attributes"]["new_key"] = "value"
instance = CreatedInstance(instance_data, self)
instance["creator_attributes"].pop("new_key")
```

### DIFF
--- a/openpype/pipeline/create/context.py
+++ b/openpype/pipeline/create/context.py
@@ -596,7 +596,14 @@ class AttributeValues(object):
             self[_key] = _value
 
     def pop(self, key, default=None):
-        return self._data.pop(key, default)
+        value = self._data.pop(key, default)
+        # Remove attribute definition if is 'UnknownDef'
+        # - gives option to get rid of unknown values
+        attr_def = self._attr_defs_by_key.get(key)
+        if isinstance(attr_def, UnknownDef):
+            self._attr_defs_by_key.pop(key)
+            self._attr_defs.remove(attr_def)
+        return value
 
     def reset_values(self):
         self._data = {}


BigRoy:
> Yes. The idea is to do changes after instance initialization. If creator would do changes before instance object is initialized, the instance would have no idea that "something has changed". This way the changes should be captured so save of changes should trigger update of the instance -> question is if I'm right?

Thanks. But, that still doesn't tell me _**when**_ I want to initialize an instance but also directly want to mark it as "something has changed"? You are explaining what result it generates but I seem to miss why I'd want to generate such a case to begin with.

iLLiCiTiT:
Well only place where instances are initialized and changes can be applied is in `collect_instances` in creator. There is no other place where instances are initialized and it make sense to make changes. Collect instances should collect data from scene, create instance from unchanged data and then do the changes afterwards.

```python
# This is example of usage, not real use-case
def collect_instances(self):
    # cache instances  if missing
    self.cache_subsets(self.collection_shared_data)
    for instance in self.collection_shared_data[
        "houdini_cached_subsets"].get(self.identifier, []):
            created_instance = CreatedInstance.from_existing(
                read(instance), self
            )
            self._add_instance_to_context(created_instance)
            creator_attributes = created_instance["creator_attributes"]
            if "deadline" in creator_attributes:
                value = creator_attributes.pop("deadline")
                creator_attributes["farm"] = value
```

BigRoy:
```python
            creator_attributes = created_instance["creator_attributes"]
            if "deadline" in creator_attributes:
                value = creator_attributes.pop("deadline")
                creator_attributes["farm"] = value
```

Thanks, that helped clear things up. So it's solely for backwards compatibility once we start moving attributes around. 👍

BigRoy (approved):
Code changes seem fine to me.

Issue #4730, opened 2023-03-29
Maya: Extract Playblast in 2020.


tokejepsen:
**Running version**
develop

**Describe the bug**
Reported by @gabormarinov on [Discord](https://discord.com/channels/517362899170230292/517382145552154634/1090338762896195726).

Hi, I've found a bug regarding viewport settings and Maya 2020 (python2?). If project_settings/maya/publish/ExtractPlayblast/capture_preset/Viewport Options/override_viewport_options is set, then viewport setting for playblast fails.

```
Instance: 
modelProxy
Message: 
Cannot apply options Invalid flag ''
Plugin: 
pyblish.ExtractPlayblast
Path: 
C:\Users\gaborca\AppData\Local\pypeclub\openpype\3.15\openpype-v3.15.3-moonrock.1\openpype\vendor\python\common\capture.py
Line: 
748
Level: 
ERROR
Thread: 
MainThread
Millis: 
345.999956131
```

Playblast is extracted, but captured images are empty (Maya viewport is empty during extraction).

If I don't set project_settings/maya/publish/ExtractPlayblast/capture_preset/Viewport Options/override_viewport_options, then there is no error message, and the extraction works with the actual viewport settings as expected.

Bonus: Thumbnail extraction always fail with a similar message like above, I guess it always tries to override viewport settings.

I've attached two logs: one with viewport override on (which fails), and one without viewport override  (which is ok).
I've tried to further narrow down the problem, but I don't really have the tools or knowledge. Could you check it please, should create an issue?

In Maya 2022 the ˛project_settings/maya/publish/ExtractPlayblast/capture_preset/Viewport Options/override_viewport_options setting works ok.
The test was done with OP 3.15.3 in both Maya 2020 and 2022.

**To Reproduce**
@BigRoy investigated this more; https://github.com/ynput/OpenPype/pull/4531#issuecomment-1488194155

[cuID:[OP-5473](https://app.clickup.com/t/863gbt48b)]

BigRoy:
This reproduces it as well:
```python
import capture
import json
print(capture)

data = r"""
{
    "viewport_options": {
        "backfaceCulling": false, 
        "cameras": false, 
        "clipGhosts": false, 
        "controlVertices": false, 
        "deformers": false, 
        "dimensions": false, 
        "displayLights": "default", 
        "displayTextures": true, 
        "dynamicConstraints": false, 
        "dynamics": false, 
        "fluids": false, 
        "fogging": false, 
[...]

PR #4777, opened 2023-04-05
Maya: Playblast profiles

tokejepsen:
## Changelog Description
Support playblast profiles.

This enables studios to customize what playblast settings should be on a per task and/or subset basis. For example `modeling` should have `Wireframe On Shaded` enabled, while all other tasks should have it disabled.

## Additional info
Currently this a breaking change because of changing the settings from a single set of playblast settings to multiple. Let's discuss the smoothest way to transition.

## Testing notes:
1. Create 2 tasks of type `modeling` and `animation`.
2. Create a new profile under `project_settings/maya/publish/ExtractPlayblast/profiles` with `modeling` as task type. Modify the `modeling` profile to be different from the default profile. Suggestion is to enable `Wireframe On Shaded` (`project_settings/maya/publish/ExtractPlayblast/profiles/0/capture_preset/Viewport Options/wireframeOnShaded`).
3. Launch Maya and publish a review from both the `modeling` and `animation` task.
4. Validate the reviews are as the defined settings.


tokejepsen:
@iLLiCiTiT would like your input on how to best to tackle the transition.

LiborBatek (approved):
Did follow the testing steps and created 2 types of ```Review``` Instance one for ```model``` instance and other for ```animation``` instance.

They were correctly created with predefined settings in OP (model with wireframes on shaded, animation with defaults)

openpype/hosts/maya/plugins/create/create_review.py
```diff
@@ -32,6 +38,30 @@ def __init__(self, *args, **kwargs):
         super(CreateReview, self).__init__(*args, **kwargs)
         data = OrderedDict(**self.data)
 
+        project_name = legacy_io.Session["AVALON_PROJECT"]
```
BigRoy, line 41: We are trying to reduce the usage of legacy_io. Right @iLLiCiTiT 

This should be using `get_current_project()` from [here](https://github.com/ynput/OpenPype/blob/77f8a8ad0c36e2075a70132f963e8248f2a3407e/openpype/pipeline/context_tools.py#L300-L311)
```suggestion
        project_name = get_current_project()["name"]
```
It'd be much more optimal however if there was a `get_current_project_name` which didn't require the database query. I thought there was a function like that but I couldn't find it.

openpype/hosts/maya/plugins/create/create_review.py
```diff
@@ -48,4 +79,16 @@ def __init__(self, *args, **kwargs):
         data["transparency"] = self.transparency
         data["panZoom"] = self.panZoom
 
+        if preset:
+            self.log.info(
+                "Using preset: {}".format(
+                    json.dumps(preset, indent=4, sort_keys=True)
+                )
+            )
```
BigRoy, line 87: Are we sure we want to be this verbose?

openpype/hosts/maya/plugins/publish/validate_review.py
```diff
@@ -0,0 +1,35 @@
...
+                "No camera found in review instance: {}".format(instance)
+            )
+        elif len(cameras) > 2:
+            raise KnownPublishError(
+                "Only a single camera is allowed for a review instance but "
+                "more than one camera found in review instance: {}. "
+                "Cameras found: {}".format(instance, ", ".join(cameras))
+            )
+
+        camera = cameras[0]
```
BigRoy, line 34: This should actually also be validating `instance.data["review_camera"]` right to ensure that whatever might change in the collector with regards to how a camera is retrieved that it still correctly retrieves it?

We're now duplicating the logic and magically rely on the logic to stay in sync between here and the collector?

openpype/hosts/maya/plugins/publish/extract_thumbnail.py
```diff
@@ -110,8 +134,9 @@ def process(self, instance):
             preset["viewport2_options"]["transparencyAlgorithm"] = transparency
 
         # Isolate view is requested by having objects in the set besides a
-        # camera.
-        if preset.pop("isolate_view", False) and instance.data.get("isolate"):
+        # camera. If there is only 1 member it'll be the camera because we
+        # validate to have 1 camera only.
+        if instance.data["isolate"] and len(instance.data["setMembers"]) > 1:
             preset["isolate"] = instance.data["setMembers"]
```
BigRoy, line 140: This logic change should be highlighted in the PR I'd say - this changes how the isolate view logic works. Previously it'd have listened to the `isolate_view` setting regardless, now it silently ignores it if it thinks it didn't find anything other than the camera.

Not too bothered by it myself, but it does make the isolate view option less explicit.

openpype/hosts/maya/plugins/publish/extract_thumbnail.py
```diff
@@ -139,6 +164,12 @@ def process(self, instance):
                 preset.update(panel_preset)
                 cmds.setFocus(panel)
 
+            self.log.info(
+                "Using preset: {}".format(
+                    json.dumps(preset, indent=4, sort_keys=True)
+                )
+            )
```
BigRoy, line 171: Logging these things is slow, please refrain from doing so.

openpype/settings/defaults/project_settings/maya.json
```diff
@@ -789,129 +789,134 @@
...
+                            "displayResolution": false,
+                            "displayFilmGate": false,
+                            "displayFieldChart": false,
+                            "displaySafeAction": false,
+                            "displaySafeTitle": false,
+                            "displayFilmPivot": false,
+                            "displayFilmOrigin": false,
+                            "overscan": 1.0
+                        }
                     }
```
BigRoy, line 917: It seems you've resolved a merge conflict wrong? You're missing the `pluginObjects` entry and have the `gpuCacheDisplayFilter` back in the hierarchy above it.

openpype/settings/entities/input_entities.py
```diff
@@ -442,7 +442,9 @@ class TextEntity(InputEntity):
 
     def _item_initialization(self):
         self.valid_value_types = (STRING_TYPE, )
-        self.value_on_not_set = ""
+        self.value_on_not_set = self.convert_to_valid_type(
+            self.schema_data.get("default", "")
+        )
```
BigRoy, line 447: This shouldn't be in this PR I think?

openpype/settings/entities/color_entity.py
BigRoy, line 1: This shouldn't be in this PR I think?

openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_capture.json
```diff
@@ -5,627 +5,745 @@
...
-                            "type": "boolean",
-                            "key": "textures",
-                            "label": "Texture Placements"
-                        },
-                        {
-                            "type": "dict-modifiable",
-                            "key": "pluginObjects",
-                            "label": "Plugin Objects",
-                            "object_type": "boolean"
-                        }
```
BigRoy, line 569: This seems to be accidentally removed.

openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_capture.json
```diff
@@ -5,627 +5,745 @@
...
+                                        "key": "follicles",
+                                        "label": "Follicles",
+                                        "default": false
+                                    },
+                                    {
+                                        "type": "boolean",
+                                        "key": "gpuCacheDisplayFilter",
+                                        "label": "GPU Cache",
+                                        "default": false
+                                    },
```
BigRoy, line 513: This should not be there since #4730

openpype/hosts/maya/plugins/create/create_review.py
iLLiCiTiT: There is actually function to get only project name:
```suggestion
        project_name = get_current_project_name()
```

```python
# Import
from openpype.pipeline import get_current_project_name
```

openpype/hosts/maya/plugins/create/create_review.py
```diff
@@ -32,6 +38,30 @@ def __init__(self, *args, **kwargs):
...
+        profiles = get_project_settings(
+            project_name
+        )["maya"]["publish"]["ExtractPlayblast"]["profiles"]
+
+        preset = None
+        if not profiles:
+            self.log.warning("No profiles present for extract playblast.")
+        else:
+            asset_doc = get_asset_by_name(project_name, data["asset"])
+            task_name = legacy_io.Session["AVALON_TASK"]
```
iLLiCiTiT, line 51:
```suggestion
            task_name = get_current_task_name()
```

openpype/hosts/maya/plugins/publish/extract_thumbnail.py
```diff
@@ -23,25 +26,47 @@ class ExtractThumbnail(publish.Extractor):
     families = ["review"]
 
     def process(self, instance):
+        project_name = legacy_io.Session["AVALON_PROJECT"]
+        profiles = get_project_settings(
```
iLLiCiTiT, line 30:
```suggestion
        project_settings = instance.context.data["project_settings"]
        profiles = project_settings["maya"]["publish"]["ExtractPlayblast"]["profiles"]
```

openpype/hosts/maya/plugins/create/create_review.py
tokejepsen: Should be a debug log, but not sure whether that actually matters for performance? Are we disabling debugs calls by default?

openpype/settings/entities/color_entity.py
```diff
@@ -11,7 +11,9 @@ class ColorEntity(InputEntity):
 
     def _item_initialization(self):
         self.valid_value_types = (list, )
-        self.value_on_not_set = [0, 0, 0, 255]
+        self.value_on_not_set = self.convert_to_valid_type(
+            self.schema_data.get("default", [0, 0, 0, 255])
+        )
         self.use_alpha = self.schema_data.get("use_alpha", True)
```
iLLiCiTiT, line 17:
```suggestion
        self.use_alpha = self.schema_data.get("use_alpha", True)
        self.value_on_not_set = self.convert_to_valid_type(
            self.schema_data.get("default", [0, 0, 0, 255])
        )
```

openpype/settings/entities/color_entity.py
```diff
@@ -64,6 +66,6 @@ def convert_to_valid_type(self, value):
             new_value.append(item)
 
         # Make sure
-        if not self.use_alpha:
+        if hasattr(self, "use_alpha") and not self.use_alpha:
```
iLLiCiTiT, line 69:
```suggestion
        if not self.use_alpha:
```

iLLiCiTiT:
I think we have to expect that `profiles` are not available yet. Also I would keep `capture_preset` for some time, so studios won't loose previous values and can move them to profiles.

openpype/hosts/maya/plugins/create/create_review.py
BigRoy: No, and unfortunately we can't really because the UI would still need to 'collect it' for if the user does want to see the values. Additionally, even if there was no logging handler you're still doing the json dump so best case scenario you're still having to deal with that conversion in there.

openpype/hosts/maya/plugins/publish/extract_thumbnail.py
tokejepsen: This is technically a bug fix cause the code was not working as originally intended. The original [comment](https://github.com/ynput/OpenPype/blob/0019ff6f7da8f8bb21d7b4b88d3bb84f24bfdc90/openpype/hosts/maya/plugins/publish/extract_playblast.py#L123-L126), indicates this behaviour which I've elaborated on.

openpype/hosts/maya/plugins/create/create_review.py
tokejepsen: For this plugin especially I keep on needing to see the preset cause there is quite a lot of "magic" that happens before the preset is used, so I would like to keep this. I know this is lazy developer, but got any ideas to keep it in?
Maybe the performance loss is acceptable here for now?

BigRoy: If it's for development only I'd recommend putting this behind a `OPENPYPE_DEBUG` check.
I've done that in many cases, like:
```python
if os.environ.get("OPENPYPE_DEBUG") == "1":
```

I believe that value is set when running from source code.

openpype/settings/defaults/project_settings/maya.json
tokejepsen: Sorry, yup bad conflict merging by me.

openpype/settings/entities/input_entities.py
tokejepsen: It is necessary for adding defaults to the `text` field. We need defaults else the fields are empty when adding a new profile which is a lot of added admin work.
Do you want it in a separate PR?

openpype/settings/entities/color_entity.py
iLLiCiTiT: I'm sorry but I won't approve this. Just move initialization of `use_alpha` attribute before `self.value_on_not_set` and this is not needed.

iLLiCiTiT: Omg, stupid GitHub showing old code...

tokejepsen:
I have labelled the setting now as deprecated:

Capture

I've tested without profiles and it all works as before. I've removed the default profile so when studios update the pipeline everything is as before. If they want to start using the profile they just need to add one or more, to "enable" them.

@BigRoy @iLLiCiTiT what do you think?

openpype/hosts/maya/plugins/publish/extract_playblast.py
```diff
@@ -65,12 +68,38 @@ def process(self, instance):
...
-        # Grab capture presets from the project settings
-        capture_presets = self.capture_preset
+        if self.profiles:
+            capture_preset = filter_profiles(
+                self.profiles, filtering_criteria, logger=self.log
+            )["capture_preset"]
+            preset = lib.load_capture_preset(data=capture_preset)
+        else:
+            self.log.warning("No profiles present for Extract Playblast")
+
```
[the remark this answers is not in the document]
iLLiCiTiT: BTW profiles gives option that there won't be any matching preset -> `filter_profiles` will return `None`. That should be probably handled too...

tokejepsen, line 94:
```suggestion
        if self.profiles:
            profile = filter_profiles(
                self.profiles, filtering_criteria, logger=self.log
            )
            capture_preset = profile.get("capture_preset) or capture_preset
        else:
            self.log.warning("No profiles present for Extract Playblast")
        preset = lib.load_capture_preset(data=capture_preset)

```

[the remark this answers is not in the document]
tokejepsen: Ok, done

tokejepsen:
@BigRoy @iLLiCiTiT you happy with this PR?

openpype/hosts/maya/plugins/publish/extract_thumbnail.py
```diff
@@ -27,21 +29,48 @@ def process(self, instance):
...
-            preset = {}
-        self.log.info("Using viewport preset: {}".format(preset))
 
-        # preset["off_screen"] =  False
+        if plugin_settings["profiles"]:
+            capture_preset = filter_profiles(
+                plugin_settings["profiles"],
+                filtering_criteria,
+                logger=self.log
+            )["capture_preset"]
```
BigRoy, line 62: Filtered profiles might return `None` yes, if there's no matching profile?

openpype/hosts/maya/plugins/publish/extract_thumbnail.py
```diff
@@ -27,21 +29,48 @@ def process(self, instance):
...
+            "task_names": task_name,
+            "task_types": task_type,
+            "subset": subset
+        }
 
+        maya_settings = instance.context.data["project_settings"]["maya"]
+        plugin_settings = maya_settings["publish"]["ExtractPlayblast"]
+
+        capture_preset = plugin_settings["capture_preset"]
+        preset = {}
```
BigRoy, line 51: This is the _deprecated behavior_, right? Let's comment that this is deprecated fallback.
I feel like this should be in the `else` for when no profiles were found below?

openpype/hosts/maya/plugins/publish/validate_review.py
```diff
@@ -0,0 +1,30 @@
...
+        if len(cameras) == 0:
+            raise KnownPublishError(
+                "No camera found in review instance: {}".format(instance)
+            )
+        elif len(cameras) > 2:
+            raise KnownPublishError(
+                "Only a single camera is allowed for a review instance but "
+                "more than one camera found in review instance: {}. "
+                "Cameras found: {}".format(instance, ", ".join(cameras))
+            )
```
BigRoy, line 28: These should be `PublishValidationError` I believe. Right @iLLiCiTiT @mkolar ? Or would those be influenced by design changes for the new publisher in the coming days?

openpype/hosts/maya/plugins/create/create_review.py
```diff
@@ -32,6 +39,30 @@ def __init__(self, *args, **kwargs):
         super(CreateReview, self).__init__(*args, **kwargs)
         data = OrderedDict(**self.data)
 
+        project_name = get_current_project_name()
+        profiles = get_project_settings(
+            project_name
+        )["maya"]["publish"]["ExtractPlayblast"]["profiles"]
```
BigRoy, line 45: What happens if `get_project_settings` is called but there are no saved settings with the new settings schema, does it then also already have the `profiles` entry?

openpype/hosts/maya/plugins/publish/extract_thumbnail.py
tokejepsen: Yup, good spot.

openpype/hosts/maya/plugins/publish/extract_thumbnail.py
tokejepsen: Yeah, good point. Have added a comment to both extractors about deprecated settings.

openpype/hosts/maya/plugins/publish/extract_thumbnail.py
```diff
@@ -27,22 +29,46 @@ def process(self, instance):
...
+            "task_names": task_name,
+            "task_types": task_type,
+            "subset": subset
+        }
+
+        if self.profiles:
+            profile = filter_profiles(
+                self.profiles, filtering_criteria, logger=self.log
+            )
+            capture_preset = profile.get("capture_preset")
```
BigRoy, line 51: Does this `ExtractThumbnail` ever get the `self.profiles` value if the settings are on the `ExtractPlayblast` class.

This seems invalid/broken.

openpype/hosts/maya/plugins/publish/extract_thumbnail.py
```diff
@@ -27,22 +29,46 @@ def process(self, instance):
...
+        else:
+            self.log.warning("No profiles present for Extract Playblast")
```
BigRoy, line 53: Wrong plug-in name.
```suggestion
            self.log.warning("No profiles present for Extract Thumbnail")
```

openpype/hosts/maya/plugins/publish/extract_thumbnail.py
```diff
@@ -27,22 +29,46 @@ def process(self, instance):
...
 
-        maya_setting = instance.context.data["project_settings"]["maya"]
-        plugin_setting = maya_setting["publish"]["ExtractPlayblast"]
-        capture_preset = plugin_setting["capture_preset"]
+        host_name = instance.context.data["hostName"]
+        family = instance.data["family"]
+        task_data = instance.data["anatomyData"].get("task", {})
+        task_name = task_data.get("name")
+        task_type = task_data.get("type")
+        subset = instance.data["subset"]
```
BigRoy, line 37: These appear to be used once, and only in the `filtering_criteria`. I'd either do:

1. Move the `get_profile` logic into a method on the class so there's less clutter in the code and easier to see where the variables are used.
2. OR just put these directly into the filtering_criteria dict so it's e.g. `"hosts": context.data["hostName"]`

openpype/hosts/maya/plugins/publish/extract_playblast.py
```diff
@@ -65,12 +68,44 @@ def process(self, instance):
...
+            capture_preset = profile.get("capture_preset")
+        else:
+            self.log.warning("No profiles present for Extract Playblast")
+
+        # Backward compatibility for deprecated Extract Playblast settings
+        # without profiles.
+        if capture_preset is None:
+            capture_preset = self.capture_preset
+
+        preset = lib.load_capture_preset(data=capture_preset)
```
BigRoy, line 99: Should this maybe be moved to `lib`? It already appears to be duplicated and has quite some lines of code. By moving to lib it can be re-used across the creator and extractors like:

```python
def get_capture_profile(host_name, family, task_name, task_type, subset, project_settings=None):
    
    if not project_settings:

    playblast_settings = project_settings["maya"]["publish"]["ExtractPlayblast"]
    profiles = playblast_settings.get("profiles")
    preset = None
    if profiles:
        asset_doc = get_asset_by_name(project_name, data["asset"])
        task_type = asset_doc["data"]["tasks"][task_name]["type"]

        filtering_criteria = {
            "hosts": "maya",
            "families": "review",
            "task_names": task_name,
            "task_types": task_type,
            "subset": data["subset"]
        }
        profile = filter_profiles(
            profiles, filtering_criteria, logger=self.log
        )
        
        if profile is not None:
            return profile
        else:
            log.warning(
                "No profile for Extract Playblast capture"
                "with criteria: {}".format(
                    json.dump(filtering_criteria, indent=4)
                )
            )
            # TODO Do we want to raise an error here?
    else:
        # Backward compatibility for deprecated Extract Playblast settings
        # without profiles.
        log.debug("Falling back to deprecated Extract Playblast capture preset "
                  "because no new style playblast profiles are defined")
        return playblast_settings["capture_preset"]
```

That way you don't need to duplicate the backwards compatibility logic.

You could use it in the extractors like:

```python
task_data = instance.data["anatomyData"].get("task", {})
context = instance.context
preset = get_capture_profile_preset(
        host_name=context.data["hostName"],
        families=instance.data["family"],
        task_name=task_data.get("name"),
        task_type=task_data.get("type"),
        subset=instance.data["subset"],
        project_settings=context.data["project_settings"]
)
```

Note that it still can return None if profiles are defined but there's no match for the current profile criteria.

openpype/hosts/maya/plugins/publish/extract_playblast.py
```diff
@@ -65,12 +68,44 @@ def process(self, instance):
...
-        capture_presets = self.capture_preset
+        host_name = instance.context.data["hostName"]
+        family = instance.data["family"]
+        task_data = instance.data["anatomyData"].get("task", {})
+        task_name = task_data.get("name")
+        task_type = task_data.get("type")
+        subset = instance.data["subset"]
+
+        filtering_criteria = {
+            "hosts": host_name,
```
BigRoy, line 79: Does the profile filter criteria require `host`? seems a bit much for a feature we know only exists in Maya nor has any host setting in the profile's settings schema?

tokejepsen: @iLLiCiTiT ?

openpype/hosts/maya/plugins/publish/extract_thumbnail.py
tokejepsen: Good spot! Fixing

openpype/hosts/maya/plugins/publish/extract_thumbnail.py
tokejepsen: This should be `Extract Playblast` cause that is where the settings are.

openpype/hosts/maya/plugins/publish/extract_thumbnail.py
tokejepsen: Cool, will condense.

openpype/hosts/maya/plugins/publish/extract_playblast.py
tokejepsen: Yeah good idea.

openpype/hosts/maya/plugins/publish/extract_playblast.py
tokejepsen: Have refactored to `lib` and hardcoded host to `maya`.

openpype/hosts/maya/plugins/publish/validate_review.py
iLLiCiTiT: Yes, should be `PublishValidationError`. `KnownPublishError` would stop publishing.

tokejepsen:
@BigRoy @iLLiCiTiT hows this PR looking now?

iLLiCiTiT:
Note: With profiles there is option that any capture preset is used. Now is used "default" from deprecated settings, but when those are removed it could crash on many places.

iLLiCiTiT (approved):
LGTM

openpype/hosts/maya/api/lib.py
```diff
@@ -3855,3 +3856,47 @@ def get_all_children(nodes):
...
+            filtering_criteria,
+            logger=log
+        )
+        capture_preset = profile.get("capture_preset")
+    else:
+        log.warning("No profiles present for Extract Playblast")
+
+    # Backward compatibility for deprecated Extract Playblast settings
+    # without profiles.
+    if capture_preset is None:
```
BigRoy, line 3895: `capture_preset` might be undefined variable here if the above statement went into the `else` branch?

Are we missing a `capture_preset = None` at the top here?
Also, if no preset was found I'd still return an empty dict just to it always returns a dict. (That is for when we remove the deprecated settings)

openpype/hosts/maya/plugins/create/create_review.py
```diff
@@ -40,12 +63,13 @@ def __init__(self, *args, **kwargs):
             data[key] = value
 
         data["fps"] = lib.collect_animation_data(fps=True)["fps"]
```
BigRoy, line 65: Unrelated to this PR. Wasn't this attribute removed a very long time ago? :D Haha, does `fps` really make sense on a maya review instance right since there's a predefined scene fps that we can rely on?

mkolar: I'm almost certain this was used on a cartoon production that was animated on 2s at some point. But I can't for life of me remember the exact use case

openpype/hosts/maya/api/lib.py
```diff
@@ -3855,3 +3856,47 @@ def get_all_children(nodes):
...
+    profiles.
+
+    Args:
+        task_name (str): Task name.
+        take_type (str): Task type.
+        subset (str): Subset name.
+        project_settings (dict): Project settings.
+        log (object): Logging object.
+    """
+    filtering_criteria = {
```
tokejepsen, line 3874:
```suggestion
    """
    capture_preset = {}
    
    filtering_criteria = {
```

openpype/hosts/maya/api/lib.py
```diff
@@ -3855,3 +3856,49 @@ def get_all_children(nodes):
...
+
+    Args:
+        task_name (str): Task name.
+        take_type (str): Task type.
+        subset (str): Subset name.
+        project_settings (dict): Project settings.
+        log (object): Logging object.
+    """
+    capture_preset = {}
+
```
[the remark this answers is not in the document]
tokejepsen:
```suggestion

```

openpype/hosts/maya/api/lib.py
tokejepsen: Cool, done.

openpype/hosts/maya/api/lib.py
```diff
@@ -3855,3 +3856,48 @@ def get_all_children(nodes):
...
+            filtering_criteria,
+            logger=log
+        )
+        capture_preset = profile.get("capture_preset")
+    else:
+        log.warning("No profiles present for Extract Playblast")
+
+    # Backward compatibility for deprecated Extract Playblast settings
+    # without profiles.
+    if capture_preset is None:
```
BigRoy, line 3896: `capture_preset` can now also be dict `{}` instead of None above.

How about initializing it to `capture_preset = None`
And change the end return to: `return capture_preset or {}`?

openpype/hosts/maya/api/lib.py
```diff
@@ -3855,3 +3856,48 @@ def get_all_children(nodes):
...
+        log.debug(
+            "Falling back to deprecated Extract Playblast capture preset "
+            "because no new style playblast profiles are defined."
+        )
+        capture_preset = plugin_settings["capture_preset"]
+
+    return capture_preset
```
tokejepsen, line 3903:
```suggestion
    capture_preset = None
    filtering_criteria = {
        "hosts": "maya",
        "families": "review",
        "task_names": task_name,
        "task_types": task_type,
        "subset": subset
    }

    plugin_settings = project_settings["maya"]["publish"]["ExtractPlayblast"]
    if plugin_settings["profiles"]:
        profile = filter_profiles(
            plugin_settings["profiles"],
            filtering_criteria,
            logger=log
        )
        capture_preset = profile.get("capture_preset")
    else:
        log.warning("No profiles present for Extract Playblast")

    # Backward compatibility for deprecated Extract Playblast settings
    # without profiles.
    if capture_preset is None:
        log.debug(
            "Falling back to deprecated Extract Playblast capture preset "
            "because no new style playblast profiles are defined."
        )
        capture_preset = plugin_settings["capture_preset"]

    return capture_preset or {}
```

openpype/hosts/maya/api/lib.py
tokejepsen: Yeah, good idea.

openpype/hosts/maya/plugins/create/create_review.py
BigRoy: @mkolar I'd like to mark this conversation resolved so that the PR is ready to merge - however I'm still dieing to know more about that 2s scenario you described and why this was needed.

I'll still mark this resolved since it's not directly related to this PR, but if you have any time to elaborate it'd be lovely.

tokejepsen:
@LiborBatek would you mind giving this a run through again?

BigRoy (approved):
I didn't do a test run of the code, but this looks good to me!

LiborBatek (approved):
All working fine. Did create new profile with different settings for ```model``` family and compare it to the default profile for the rest of families. All as expected

image

image

BigRoy:
I'm almost sure that the next feature request will be - how can I generate two playblasts from modeling? On with wireframe and one with smooth shading, etc. haha. 

Anyway, solid work!

tokejepsen:
> I'm almost sure that the next feature request will be - how can I generate two playblasts from modeling? On with wireframe and one with smooth shading, etc. haha.

Lol, yeah. Probably not a huge amount of work, as you could have two profiles targetting `modeling` but thats famous last words.

antirotor (approved):

PR #4779, opened 2023-04-05
Global: label key in instance data is optional

jakubjezek001:
## Changelog Description
Collect OTIO review plugin is not crashing if `label` key is missing in instance data.

## Additional info
Label key is conditional. 

## Testing notes:
1. publish from Resolve clip which is having counterpart in reviewable layer
2. all should be published as expected

### DIFF
--- a/openpype/plugins/publish/collect_otio_review.py
+++ b/openpype/plugins/publish/collect_otio_review.py
@@ -87,7 +87,9 @@ class CollectOtioReview(pyblish.api.InstancePlugin):
                 otio_review_clips.append(otio_gap)
 
         if otio_review_clips:
-            instance.data["label"] += " (review)"
+            # add review track to instance and change label to reflect it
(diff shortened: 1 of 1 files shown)


iLLiCiTiT (approved):

openpype/plugins/publish/collect_otio_review.py
```diff
@@ -87,7 +87,9 @@ def process(self, instance):
                 otio_review_clips.append(otio_gap)
 
         if otio_review_clips:
-            instance.data["label"] += " (review)"
+            # add review track to instance and change label to reflect it
+            label = instance.data.get("label", instance.data["subset"])
+            instance.data["label"] = label + " (review)"
```
BigRoy, line 92: Isn't the default fallback for `label` the `instance.data["name"]` entry instead of subset?

PR #4746, opened 2023-03-29
Refactor: Remove `handles`, `edit_in` and `edit_out` backwards compatibility

BigRoy:
## Changelog Description

Removes backward compatibiliy fallback for data called `handles`, `edit_in` and `edit_out`.

## Additional info

This intends to fix #4709 

## Testing notes:

1. Test the areas of the code that were touched.
2. Use set frame range action in Maya, Fusion, Nuke
3. Load script precomp in Nuke
4. Publish reviews from a few hosts, like Maya + Nuke
5. Publish "reads" and "model" from Nuke
6. Publishes tested for this PR should still load and update as expected.
7. The loader tool should still nicely show the handles

Also please check whether I've found all cases of handles in the codebase :)

PR #4781, opened 2023-04-05
Loader: Fix missing variable

iLLiCiTiT:
## Changelog Description
There is missing variable `handles` in loader tool after https://github.com/ynput/OpenPype/pull/4746. The variable was renamed to `handles_label` and is initialized to `None` if handles are not available.

## Testing notes:
1. Make sure there is a version without handles information.
2. Open loader
3. The version should work and be shown in UI

### DIFF
--- a/openpype/tools/loader/model.py
+++ b/openpype/tools/loader/model.py
@@ -361,10 +361,11 @@ class SubsetsModel(TreeModel, BaseRepresentationModel):
             version_data.get("endFrame", None)
         )
 
+        handles_label = None
         handle_start = version_data.get("handleStart", None)
         handle_end = version_data.get("handleEnd", None)
         if handle_start is not None and handle_end is not None:
-            handles = "{}-{}".format(str(handle_start), str(handle_end))
+            handles_label = "{}-{}".format(str(handle_start), str(handle_end))
 
         if frame_start is not None and frame_end is not None:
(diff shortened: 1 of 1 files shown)


openpype/tools/loader/model.py
```diff
@@ -361,10 +361,11 @@ def set_version(self, index, version):
             version_data.get("endFrame", None)
         )
 
+        handles_label = None
         handle_start = version_data.get("handleStart", None)
         handle_end = version_data.get("handleEnd", None)
         if handle_start is not None and handle_end is not None:
-            handles = "{}-{}".format(str(handle_start), str(handle_end))
+            handles_label = "{}-{}".format(str(handle_start), str(handle_end))
```
BigRoy, line 368: Could we put the `handles_label = None` into the else statement instead? It's much more explicit then that it's THIS or THAT than the current implementation.

So basically bringing back the logic from before #4746 but setting the value to None.

iLLiCiTiT: > Could we put the handles_label = None into the else statement instead?

I would say this is more explicit in a way that the variable must be 100% available, so it's clear it must be available in the scope.

BigRoy: Fine with me.

BigRoy (approved):

PR #4784, opened 2023-04-05
Global: Optimize anatomy formatting by only formatting used templates instead

BigRoy:
## Changelog Description

Optimization to not format full anatomy when only a single template is used. Instead format only the single template instead.

## Additional info

This currently doesn't replace ALL cases yet where `anatomy.format` could be replaced with a more explicit `anatomy.templates_obj[template_key][name].format`

A call to `anatomy.format` took about 0.007 seconds for me. A call to a single template object `.format` takes 0.0 seconds. (or at least such a short time that printing time spend (end-start) would basically show in Python as `0.0`.

So say 5 instances each calling it 20 times would be about 0.7 seconds difference. In some areas it's called _per output file_. So say 5 instances of 200 frames could result in 7 seconds difference. (This example calculation is ignoring the fact that the single template object would actually not be 0.0, but likely 0.00001 or alike)

This optimization does would be mostly worth it in areas that are speed critical or might be called many times.

## Testing notes:

1. Test the areas of the code these changes touch


openpype/plugins/publish/integrate_legacy.py
```diff
@@ -600,9 +600,9 @@ def register(self, instance, repres):
 
             if not instance.data.get("publishDir"):
                 instance.data["publishDir"] = (
-                    anatomy_filled
-                    [template_name]
-                    ["folder"]
+                    anatomy.templates_obj[template_name]["folder"].format(
```
iLLiCiTiT, line 603:
```suggestion
                    anatomy.templates_obj[template_name]["folder"].format_strict(
```

openpype/plugins/publish/integrate.py
```diff
@@ -761,8 +759,8 @@ def prepare_representation(self, repre,
                     template_data["udim"] = index
                 else:
                     template_data["frame"] = index
-                anatomy_filled = anatomy.format(template_data)
-                template_filled = anatomy_filled[template_name]["path"]
+                template_obj = anatomy.templates_obj[template_name]["path"]
```
iLLiCiTiT, line 762: This could be moved before the huge condition, so the variable `template_obj` is defined only once?

iLLiCiTiT (approved):
LGTM

openpype/plugins/publish/integrate.py
iLLiCiTiT: Maybe even split it into 2 variables?
```
path_template = anatomy.templates_obj[template_name]["path"]
folder_template = anatomy.templates_obj[template_name]["folder"]
```

PR #4786, opened 2023-04-05
Global: Remove legacy integrator

BigRoy:
## Changelog Description

Remove the legacy integrator.

## Additional info

Also removes the parts of the new integrator that should be redundant without the legacy integrator, including its settings.

I didn't touch the old already deprecated legacy settings because those are [still called in a fallback here](https://github.com/ynput/OpenPype/blob/1531708236cf9f27e459c2b20cd642e5f08905af/openpype/pipeline/publish/lib.py#L69-L77). Other than that they are also still in settings to keep the deprecated setting values around I think.

Should we remove those fallbacks too? (Not sure how long those have been deprecated).

## Testing notes:

1. Publishing should still work (across all hosts)


iLLiCiTiT (approved):
LGTM

PR #4788, opened 2023-04-05
Unreal: Use common logic to configure executable

antirotor:
## Changelog Description
Unreal Editor location and version was autodetected. This easied configuration in some cases but was not flexible enought. This PR is changing the way Unreal Editor location is set, unifying it with the logic other hosts are using.

## Additional info
When defining location of Unreal, set executable path in Settings to point to UnrealEditor binary.

## Testing notes:
1. Set location to `UnrealEditor` binary in settings.
2. Unreal Editor should start-up with all new project machinery the same way as before.

> **Note**
> To test it properly, you might want to delete Integration Plugin from the Engine itself if you have one there: `Engine/Plugins/Marketplace/OpenPype`


LiborBatek (approved):
Successfully tested the run of UE 5.1 via OP Settings.

First deleting the OP plugin and then configuring path via OP Settings to UE editor executable...

image

And OP Plugin built successfully here:
image


Running UE with OP and some OP Assets loaded into the level:

image

simonebarbieri (approved):
Works for both 5.0 and 5.1!

PR #4790, opened 2023-04-05
Global: plugins cleanup plugin will leave beauty rendered files

kalisp:
## Changelog Description
Attempt to mark more files to be cleaned up explicitly in intermediate `renders` folder in work area for farm jobs.

## Additional info
Hosts are rendering files into intermediate folder `work/TASK/HOST/renders`
Currently nothing in `renders` folder gets deleted (at least on Deadline after publish). This PR collects all files in representations stored in metadata .json to be explicitly cleaned up.

Whole folder cannot be marked as there is `metadata.json` which needs to stay until the end (in case of multiple Publish jobs).

Attempt on mark reviews/thumbnails to be deleted as they are staying in `renders` folder too.

As data in `renders` are temporary, they should be cleaned up. This PR tries to be as explicit as possible to limit risk of purging something important.

I assume this might be much more complicated, so this PR is just starting point for a discussion.

!!! This works reliably for Farm publishes, it does not work for local rendering as local render plugins are not marking its products to be explicitly removed.  In some cases (Nuke?) it is desired behavior, but we should slowly update all local rendering plugins to mark its rendering outputs to be explicitly cleaned up.

Expected behavior:
`work/TASK/HOST/renders` should be cleaned up.


openpype/lib/file_transaction.py
```diff
@@ -212,3 +212,21 @@ def _same_paths(self, src, dst):
             return os.stat(src) == os.stat(dst)
 
         return src == dst
+
+
+def fill_cleanupFullPaths_for_representation(instance, repre):
```
iLLiCiTiT, line 217:
```suggestion
def add_expected_files_for_cleanup(instance, repre):
```

openpype/lib/file_transaction.py
```diff
@@ -212,3 +212,21 @@ def _same_paths(self, src, dst):
...
+    """ Explicitly mark repre files to be deleted.
+
+    Should be used on intermediate files (eg. review, thumbnails) to be
+    explicitly deleted.
+    """
+    files = repre["files"]
+    staging_dir = repre["stagingDir"]
```
iLLiCiTiT, line 224: Staging dir is not requirement on representation.

openpype/plugins/publish/extract_thumbnail.py
```diff
@@ -133,6 +137,7 @@ def process(self, instance):
                 "Adding thumbnail representation: {}".format(new_repre)
             )
             instance.data["representations"].append(new_repre)
+            fill_cleanupFullPaths_for_representation(instance, new_repre)
```
iLLiCiTiT, line 140: This is not needed at all. The directory is already added on line 82.

openpype/plugins/publish/extract_thumbnail_from_source.py
```diff
@@ -73,6 +77,7 @@ def process(self, instance):
             "Adding thumbnail representation: {}".format(new_repre)
         )
         instance.data["representations"].append(new_repre)
+        fill_cleanupFullPaths_for_representation(instance, new_repre)
```
iLLiCiTiT, line 80: Already happens on line 101 (96).

openpype/plugins/publish/extract_burnin.py
```diff
@@ -351,6 +354,8 @@ def main_process(self, instance):
                 # Add new representation to instance
                 instance.data["representations"].append(new_repre)
 
+                fill_cleanupFullPaths_for_representation(instance, new_repre)
```
iLLiCiTiT, line 357: I'm not sure if it's safe to do it all the time in extract burnin?

openpype/lib/file_transaction.py
iLLiCiTiT: If you would keep it, at least move it to `openpype.pipeline.publish` and rename it to key agnostic function name.

openpype/plugins/publish/collect_rendered_files.py
```diff
@@ -107,6 +111,8 @@ def _process_path(self, data, anatomy):
                 self._fill_staging_dir(repre_data, anatomy)
                 representations.append(repre_data)
 
+                fill_cleanupFullPaths_for_representation(instance, repre_data)
```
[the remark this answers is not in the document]
kalisp: The issue is that `renders` contains `metadata.json` which cannot be cleaned in first publish job (if there are multiple publish jobs).

openpype/lib/file_transaction.py
kalisp: Ok, moved.

openpype/pipeline/publish/lib.py
```diff
@@ -790,3 +790,23 @@ def _validate_transient_template(project_name, template_name, anatomy):
...
+    if not staging_dir:
+        return
+
+    if isinstance(files, str):
+        files = [files]
+
+    for file_name in files:
+        expected_file = os.path.join(staging_dir, file_name)
+        print("Explicitly marking {} for cleanup".format(expected_file))
+        instance.context.data["cleanupFullPaths"].append(expected_file)
```
iLLiCiTiT, line 812: The print would make it noisy and probably slow (100 frame sequence would print 100 lines).
```suggestion
    instance.context.data["cleanupFullPaths"].extend([
        os.path.join(staging_dir, file_name)
        for file_name in files
    ])
```

LiborBatek:
@kalisp please provide some testing steps and testing conditions needed (is it ok to submit to Deadline running on my machine for this PR? so you just meant to not render locally without using Deadline? Right?) Thx

kalisp:
This PR is meant more for Deadline publishes. It is fine to publish to your local Deadline, no changes in plugins in DL repository.

If your DCC of choice works with local publishing, you might try that too, but it is responsibility of DCC to mark its intermediate files to be cleaned up properly. (Which information if some DCC doesn't do it is also helpful.)

You should try to publish something via DL and then check `work/../renders` folder if it gets cleaned up properly (and if all things that you would expect got published).

jakubjezek001 (approved):
LGTM

openpype/pipeline/publish/lib.py
```diff
@@ -791,3 +791,22 @@ def _validate_transient_template(project_name, template_name, anatomy):
...
+
+
+def add_repre_files_for_cleanup(instance, repre):
+    """ Explicitly mark repre files to be deleted.
+
+    Should be used on intermediate files (eg. review, thumbnails) to be
+    explicitly deleted.
+    """
+    files = repre["files"]
+    staging_dir = repre.get("stagingDir")
```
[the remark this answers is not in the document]
BigRoy: Yes, can definitely happen since it used to originally be stored on the `instance.data["stagingDir"]` and wasn't always set on the representation. I believe that hasn't been changed since and in some cases it's still used from the instance.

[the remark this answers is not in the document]
iLLiCiTiT: > Can this actually happen that a representation is not having stagingDir?

I actually think that most of representations don't have it and are using staging dir from instance (not sure here...). I think that representation with staging dir is usually not pointing to same as staging dir on instance.

Issue #4789, opened 2023-04-05
Maya: focalLength should be collected also for render families


kalisp:
**Running version**
[ex. 3.15.3-nightly.2]

**Describe the bug**
`{focalLenght}` was added into burnin profiles, but currently is only collected in `collect_review.py`. Problem is that `render` family uses burnin profiles too. (At least noticed in Maya when rendering multiple render passes.)

**To Reproduce**
1. Create multiple render passes in Maya
2. Render on Farm
3. It will fail with:.
[focal_length_fail.txt](https://github.com/ynput/OpenPype/files/11160488/focal_length_fail.txt)

**Expected behavior**
Render family shouldn't fail

**Desktop (please complete the following information):**
 - OS: [e.g. windows]
 - Host: noticed in Maya, but would probably affect all hosts with 'render' family

[cuID:[OP-5551](https://app.clickup.com/t/861mmd6ex)]

tokejepsen:
> @tokejepsen what do you think?

Yeah lets try and isolate it to playblasts/reviews only. If a studio wants focal length support for renders then we can implement it properly (somehow).

BigRoy:
So the original production breaking bug has been resolved. However this issue is a decent remainder that at some point we might want to include the information for renders too, but I wouldn't say it's high priority.

Note also that I believe both Arnold and Redshift (as an example) write the focal length in the metadata of the output images. (Right?)

antirotor:
I am closing this, but it might still be an issue for additional hosts like Unreal. In that case, lets create host-specific issue.

PR #4791, opened 2023-04-05
Maya: Add family filter 'review' to burnin profile with focal length

BigRoy:
## Changelog Description

Avoid profile burnin with `focalLength` key for renders, but use only for playblast reviews.

## Additional info

@kalisp Any chance you could run this through automated tests to see if this would resolve the #4789 error you reported?
I have a feeling that additional filtering might not be doing what we need in this case, but I hope it would.

Also I believe Renders do actually get a `review` family as well, but about that I'm also not entirely sure.

## Testing notes:

1. Submit Maya render to deadline using default project settings (that includes the `focalLength` burnin profile)
2. The burnin for the render should NOT error and should exclude the focalLength burnin. The regular burnin should still get generated.

Tested:
1. [x] Playblast review instance should work fine and the burnin should have focalLength info at the top.
2. [x] Submit render + deadline render + publish job _with burnins!_ should work fine.


BigRoy:
@LiborBatek any chance you can confirm maya renders through deadline work ok with burnins from Maya? (make sure to work with default project settigns so it uses the profile settigns being updated here).

1. Playblast review instance should work fine and the burnin should have focalLength info at the top.
2. Submit render + deadline render + publish job _with burnins!_ should work fine.

tokejepsen (approved):
Code looks good and tested submitting renders from Maya 2023.

LiborBatek (changes requested):
Even tho ```Review``` has the ```Focal Length``` (img shown focal length in top center)

image

On DL it fails on ```Burnins``` after rendering the image sequence successfully before.

image

So still needs to address this issue.

tokejepsen:
@LiborBatek could you screengrab your burnin settings?
`project_settings/global/publish/ExtractBurnin/profiles`

LiborBatek:
> @LiborBatek could you screengrab your burnin settings? `project_settings/global/publish/ExtractBurnin/profiles`

image

BigRoy:
@LiborBatek could you right click it and "remove override"? It seems you have an older burnin override? Or did you add `focalLength` there yourself? There should be an additional profile underneath.

Also, this does make me think. Should we maybe just log an error/warning in the burnin generation if a key is invalid instead of actual erroring the burnin generation completely? I feel like we should allow it to pass but just log a message that it's skipping a certain key and will keep it empty instead? Thoughts?

tokejepsen:
> @tokejepsen Also, this does make me think. Should we maybe just log an error/warning in the burnin generation if a key is invalid instead of actual erroring the burnin generation completely? I feel like we should allow it to pass but just log a message that it's skipping a certain key and will keep it empty instead? Thoughts?

Yeah, that might be safer since missing burnins are not necessarily production stopper but failed renders are.

LiborBatek:
> could you right click it and "remove override"? It seems you have an older burnin override? Or did you add `focalLength` there yourself? There should be an additional profile underneath.

I had a hard time in OP Settings...I didnt see any entry containing ```focalLength``` in any ```Extract Burnins``` categories....so I have just add one then...

Should I see it somewhere without my entry then?! I coudlnt find it tho.

Note: the name of the attribute ```focalLength``` doesnt follow other naming conventions like ```frame_start``` or ```frame_end``` ...didnt have docs in hand I have first tried to format it like ```focal_length``` . Just a thought @tokejepsen

tokejepsen:
> I had a hard time in OP Settings...I didnt see any entry containing focalLength in any Extract Burnins categories....so I have just add one then...

Did you remove the overrides from the project?
Just checking but have you modified the default project?

LiborBatek:
> > I had a hard time in OP Settings...I didnt see any entry containing focalLength in any Extract Burnins categories....so I have just add one then...
> 
> Did you remove the overrides from the project? Just checking but have you modified the default project?

Yep, I did remove them and revert to ```default``` I also checked the ```default``` but no trace of any ```focalLength``` attrib...is that correct?? 
Show me a scrngrab of yours pls. So I know what to look and set. thx

BigRoy:
> Yep, I did remove them and revert to default I also checked the default but no trace of any focalLength attrib...is that correct??
> Show me a scrngrab of yours pls. So I know what to look and set. thx

There should be an extra profile below it.

tokejepsen:
> Show me a scrngrab of yours pls. So I know what to look and set. thx

Capture

LiborBatek (changes requested):
Ok so now I have it reset it so it mimmicks your OP Settings 

image



However still getting Error while running ```Publish job``` on DL with following error:

image

LiborBatek:
I have a question..is it possible that this PR interfere with others speaking of OP and functionality?! I cant get my head around it how it is possible but happening to me :)

Im trying to render some job on DL and getting the same errors on ```Publish Job``` because of ```focalLength``` even tho I dont have such a attribute in my OP settings and its non-existent.

image

tokejepsen:
@LiborBatek it could be that you'll need to save the settings to update the database.

Maybe try adding a task empty name `project_settings/global/publish/ExtractBurnin/profiles/1/task_names` > save > remove the empty task name > save. This should hopefully reset your project settings for burnins.

BigRoy:
@LiborBatek did we end up solving how/where you got stuck? Since this PR intends to fix Maya rendering functionality not working with default settings I feel like it's a critical bugfix.

LiborBatek:
> @LiborBatek did we end up solving how/where you got stuck? Since this PR intends to fix Maya rendering functionality not working with default settings I feel like it's a critical bugfix.

Yeah I have realized my Deadline was reffering wrong REPO for OP plugins / build ver. So I need to give it another go, now DL reffering to source code of OP not that obsolete build ver.

BigRoy:
Great! Let's get to merging.

antirotor (approved):

PR #4793, opened 2023-04-06
Maya: Update frame range of instances when clicking on "Reset frame range"

friquette:
## Changelog Description
Update "start frame", "end frame", "handle start" and "handle end" extra attributes of animation instances when using the "Reset frame range" of OP in Maya.


tokejepsen:
Interesting!

What is the use case for this?
Normally the frame range of the instances would be validated and repaired if they were out of sync from the asset.

BigRoy:
I agree with @tokejepsen - this doesn't seem to be the correct intended behavior. 

Also, be aware that the logic will not work for all instances. There will be instances who do not have frame range attributes. There will also be instances that are time-dependent but don't expose their time settings on the instance (e.g. renderlayers).

openpype/hosts/maya/api/lib.py
```diff
@@ -2111,6 +2111,34 @@ def reset_frame_range():
...
+    frames_attributes = {
+        'frameStart': frame_start,
+        'frameEnd': frame_end,
+        'handleStart': handle_start,
+        'handleEnd': handle_end
+    }
+
+    for instance in instances:
+        if not cmds.attributeQuery("id", node=instance, exists=True):
+            continue
```
BigRoy, line 2130: This check seems to be redundant since the `cmds.ls` call already filters to only nodes with an `.id` attribute.

openpype/hosts/maya/api/lib.py
```diff
@@ -2111,6 +2111,34 @@ def reset_frame_range():
...
+
+        id_attr = "{}.id".format(instance)
+        if cmds.getAttr(id_attr) != "pyblish.avalon.instance":
+            continue
+
+        for key, value in frames_attributes.items():
+            cmds.setAttr(
```
BigRoy, line 2137: There will be instance nodes that are not time dependent and don't have the relevant attributes. For those this will error.

BigRoy (commented):
As mentioned I suspect this PR won't get merged since it doesn't seem to be the intended behavior of the method. However, here are some purely code based notes.

OlivierOsotimehin:
Hello, 
This is was a feature requested to add fluidity to the artist work. If a supervisor tell artist there is a change of duration of a shot : 
[101-150] > [101-180] 

They will use the **rest frame range**  to get the last frame range from production tracking. This will get change for the frame in / out / handle of the scene itself but not on the exported entities (animation, camera , etc ) 

If publish as is, the sanity check is going to complain that things are not right.

Since we are already doing something (**reset frame range**) regarding the time we could ensure that all the entities that are dependent of time in the scene  are changed to the correct known value.

Otherwise, reset frame range should not be there and all should be repaired/ fix from the sanity check.

tokejepsen:
> If publish as is, the sanity check is going to complain that things are not right.

This validator [Validate Frame Range](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/maya/plugins/publish/validate_frame_range.py), should fail if the instance frame range is out of sync with the asset.

Does this not happen for your pipeline?

OlivierOsotimehin:
It does happen ! but we could make it faster to publish scene by changing also the instance to what we know to be the correct value when we use the "reset frame range" in the openpype menu. 

That's for sure a "comfort" request, that lead to less complaints of artist and more adhesion of pipeline

tokejepsen (changes requested):
@friquette this feature will need to be behind a setting

OlivierOsotimehin:
Hello, 
Yes the artist tend to click reset frame range when they are told that the frame range have change (which happens quite often in commercials).

And for sure we don't optimise cache time for exemple, we aim for stability (when the character is not visible there may be shadow, etc etc ) so we export always the full range with a little more even for motion blur raison

tokejepsen:
@BigRoy what you think of this?

This feature is behind a setting, so studios need to opt in now. I would also suggest to maybe widen the scope of this to no just `animation` instances, this could come as a `family` filter setting. We would need to be more explicit in the code though about which instances are updated and and maybe how since not all instances are made the same.
Lastly I think we should refactor the code for collecting the instances to `lib` so the collector and `reset frame range` share the same logic.

BigRoy:
> @BigRoy what you think of this?
> 
> This feature is behind a setting, so studios need to opt in now. 

Looks great to me. We would have it disabled but I can see it being useful for some.

We might want to add a `instances=True` argument to the `reset_frame_range` function because the function is used in other areas of the code to, e.g. on resetting render settings / creating a first render instance (if resetting on create is enabled in project settings, which I believe is the default behavior). As such, that would now ALSO reset the instances in the scene - making it very confusing. Then the calls that don't want to reset instances to begin with use `instances=False`.

> I would also suggest to maybe widen the scope of this to no just animation instances, this could come as a family filter setting.  > We would need to be more explicit in the code though about which instances are updated and and maybe how since not all instances are made the same.

It currently actually not only applies to `animation` family instances, but to any instance that has the relevant frame based attributes, so it'd also do it for pointcaches, yeti cache frame ranges, etc.

The logic wouldn't work for resetting frame ranges of renderlayer instances, because those attributes aren't on the renderlayer instance but set in the scene with potential renderlayer overrides, etc.

> Lastly I think we should refactor the code for collecting the instances to lib so the collector and reset frame range share the same logic.

That would make sense. It's good to note however that #4388 will already do that since there you'd get the instances from the Create Context instead. So I'm wondering whether it's worth it now. This also means that this feature will need refactoring in that PR to continue to work.

tokejepsen:
@friquette could you implement this?

- add keyword `instances=True` to `reset_frame_range` and make sure it does not run on creating render settings.
- widen the scope of to other instances that have frame ranges, but not `render` family.

BigRoy:
> widen the scope of to other instances that have frame ranges, but not render family.

As far as I know it already works for ANY instance that has the frame attributes - it doesn't seem to check anything related to animation instances or alike.

tokejepsen:
> As far as I know it already works for ANY instance that has the frame attributes - it doesn't seem to check anything related to animation instances or alike.

Yup, you are right.

openpype/settings/entities/schemas/projects_schema/schema_project_maya.json
```diff
@@ -153,6 +153,23 @@
         },
         {
             "type": "dict",
+            "collapsible": true,
+            "key": "update_instances",
+            "label": "Update animation instances on Reset Frame Range",
```
mkolar, line 158:
```suggestion
            "label": "Update publishable instances on Reset Frame Range",
```

Animation very much suggests that it only works for that family, but actually it updates all publishable instances

openpype/settings/entities/schemas/projects_schema/schema_project_maya.json
```diff
@@ -153,6 +153,23 @@
...
+            "checkbox_key": "enabled",
+            "children": [
+                {
+                    "type": "label",
+                    "label": "If enabled, the frame range and the handles of all the animation instances will be updated when using the 'Reset Frange Range' functionality"
```
mkolar, line 163:
```suggestion
                    "label": "If enabled, the frame range and the handles of all the publishable instances will be updated when using the 'Reset Frame Range' functionality"
```

openpype/settings/entities/schemas/projects_schema/schema_project_maya.json
```diff
@@ -153,6 +153,23 @@
         },
         {
             "type": "dict",
+            "collapsible": true,
+            "key": "update_instances",
```
mkolar, line 157: This is not explicit enough in maya context. It should be clearer what it relates too

```suggestion
            "key": "update_publishable_frame_range",
```

openpype/settings/defaults/project_settings/maya.json
```diff
@@ -47,6 +47,9 @@
             "destination-path": []
         }
     },
+    "update_instances": {
```
mkolar, line 50:
```suggestion
    "update_publishable_frame_range": {
```

mkolar (changes requested):
In general this makes pretty good sense in certain workflows. I'm not sure it actually add a whole lot of value compared to keeping identical functionality in the validators repair, but I have no major problem with it.

tokejepsen (changes requested):
Add keyword `instances=True `to `reset_frame_range` and make sure it does not run on creating render settings.

tokejepsen:
@friquette is this ready to test?

friquette:
> @friquette is this ready to test?

Yes it is

openpype/hosts/maya/api/lib.py
```diff
@@ -2199,6 +2199,8 @@ def reset_frame_range(playback=True, render=True, fps=True):
         render (bool, Optional): Whether to set the maya render frame range.
             Defaults to True.
         fps (bool, Optional): Whether to set scene FPS. Defaults to True.
+        instances (bool, Optional): Whether to update publishable instances.
+            Defaults to True.
```
BigRoy, line 2203:
```suggestion
        instances (bool, Optional): Whether to update publishable instances.
            Defaults to True. Even when True this will only reset instances if
            `update_publishable_frame_range` is also enabled in project settings.
```

Should we clarify that even when True it only does so if also enabled in settings?

tokejepsen (changes requested):
Getting this error when opening the settings:
```

  File "C:\Users\tokejepsen\OpenPype\openpype\tools\settings\settings\categories.py", line 564, in reset
    self._create_root_entity()

  File "C:\Users\tokejepsen\OpenPype\openpype\tools\settings\settings\categories.py", line 979, in _create_root_entity
    entity = ProjectSettings(

  File "C:\Users\tokejepsen\OpenPype\openpype\settings\entities\root_entities.py", line 654, in __init__
    super(ProjectSettings, self).__init__(schema_hub, reset)

  File "C:\Users\tokejepsen\OpenPype\openpype\settings\entities\root_entities.py", line 76, in __init__
    self._item_initialization()

  File "C:\Users\tokejepsen\OpenPype\openpype\settings\entities\root_entities.py", line 202, in _item_initialization
    self._add_children(self.schema_data)

  File "C:\Users\tokejepsen\OpenPype\openpype\settings\entities\root_entities.py", line 172, in _add_children
    child_obj = self.create_schema_object(children_schema, self)

  File "C:\Users\tokejepsen\OpenPype\openpype\settings\entities\root_entities.py", line 231, in create_schema_object
    return self.schema_hub.create_schema_object(

  File "C:\Users\tokejepsen\OpenPype\openpype\settings\entities\lib.py", line 358, in create_schema_object
    return klass(schema_data, *args, **kwargs)

  File "C:\Users\tokejepsen\OpenPype\openpype\settings\entities\base_entity.py", line 916, in __init__
    self._item_initialization()

  File "C:\Users\tokejepsen\OpenPype\openpype\settings\entities\dict_immutable_keys_entity.py", line 196, in _item_initialization
    self._add_children(self.schema_data)

  File "C:\Users\tokejepsen\OpenPype\openpype\settings\entities\dict_immutable_keys_entity.py", line 163, in _add_children
    child_obj = self.create_schema_object(children_schema, self)

  File "C:\Users\tokejepsen\OpenPype\openpype\settings\entities\base_entity.py", line 962, in create_schema_object
    return self.schema_hub.create_schema_object(*args, **kwargs)

  File "C:\Users\tokejepsen\OpenPype\openpype\settings\entities\lib.py", line 358, in create_schema_object
    return klass(schema_data, *args, **kwargs)

  File "C:\Users\tokejepsen\OpenPype\openpype\settings\entities\base_entity.py", line 916, in __init__
    self._item_initialization()

  File "C:\Users\tokejepsen\OpenPype\openpype\settings\entities\dict_immutable_keys_entity.py", line 196, in _item_initialization
    self._add_children(self.schema_data)

  File "C:\Users\tokejepsen\OpenPype\openpype\settings\entities\dict_immutable_keys_entity.py", line 146, in _add_children
    children_schemas = self.schema_hub.resolve_schema_data(

  File "C:\Users\tokejepsen\OpenPype\openpype\settings\entities\lib.py", line 308, in resolve_schema_data
    schema_type = schema_data["type"]

KeyError: 'type'
```

BigRoy:
Just noticed that there's actually logic implemented that is VERY CLOSE to this PR [here](https://github.com/ynput/OpenPype/blob/281f6186e44d476a45b52d1c2292ee85b2a93463/openpype/hosts/maya/api/lib.py#L3069-L3093) but that's only triggered on task changed. We could technically merge/clean up that functionality and also use it in reset frame range.

Also, according to the reset frame range logic implemented here I think that might be setting the wrong frame ranges for some things if handles needed to be included? It's not setting start and end handles for example.

Anyway, might be good to merge/cleanup since the logic is pretty much a duplicate.

tokejepsen:
> Just noticed that there's actually logic implemented that is VERY CLOSE to this PR [here](https://github.com/ynput/OpenPype/blob/281f6186e44d476a45b52d1c2292ee85b2a93463/openpype/hosts/maya/api/lib.py#L3069-L3093) but that's only triggered on task changed. We could technically merge/clean up that functionality and also use it in reset frame range.

Good spot! Yeah, lets merge the logic/code.

tokejepsen:
@friquette could you have a look at @BigRoy feedback [here](https://github.com/ynput/OpenPype/pull/4793#issuecomment-1516771857), then we can merge this PR?

friquette:
> @friquette could you have a look at @BigRoy feedback [here](https://github.com/ynput/OpenPype/pull/4793#issuecomment-1516771857), then we can merge this PR?

I don't really understand what I need to do here. Can you explain please ?

tokejepsen (changes requested):
@friquette I still get [this error](https://github.com/ynput/OpenPype/pull/4793#pullrequestreview-1393401921) when opening the settings.

friquette:
> @friquette I still get [this error](https://github.com/ynput/OpenPype/pull/4793#pullrequestreview-1393401921) when opening the settings.

It's fixed. I didn't see that I removed the 'type' key from a maya setting...

openpype/hosts/maya/api/lib.py
```diff
@@ -3053,32 +3061,48 @@ def remove_render_layer_observer():
         pass
 
 
-def update_content_on_context_change():
-    """
-    This will update scene content to match new asset on context change
-    """
-    scene_sets = cmds.listSets(allSets=True)
+def update_assets_frame_range():
```
tokejepsen, line 3064: Could we get doc string for this?

tokejepsen: Also maybe `update_instances_frame_range` would be more accurate?

openpype/hosts/maya/api/lib.py
```diff
@@ -3053,32 +3061,48 @@ def remove_render_layer_observer():
...
+    }
+
+    for instance in collected_instances:
+        family_attr = "{}.family".format(instance)
+        if family_attr == "render":
+            continue
+
+        id_attr = "{}.id".format(instance)
+        if cmds.getAttr(id_attr) != "pyblish.avalon.instance":
+            continue
```
tokejepsen, line 3091: Might need an attribute query here for `.id`.

tokejepsen (changes requested):
Tested in Maya 2023 successfully. Just some minor code feedback.

openpype/hosts/maya/api/lib.py
```diff
@@ -3053,32 +3061,48 @@ def remove_render_layer_observer():
...
+        'frameEnd': new_data["frameEnd"],
+        'handleStart': new_data["handleStart"],
+        'handleEnd': new_data["handleEnd"],
+        'asset': new_asset
+    }
+
+    for instance in collected_instances:
+        family_attr = "{}.family".format(instance)
+        if family_attr == "render":
+            continue
```
BigRoy, line 3087: This is never true because it's missing a `getAttr`?

tokejepsen (approved):

tokejepsen:

openpype/hosts/maya/api/lib.py
```diff
@@ -3053,31 +3061,52 @@ def remove_render_layer_observer():
...
+
+    for instance in collected_instances:
+        family_attr = "{}.family".format(instance)
+        if (cmds.attributeQuery('family', node=instance, exists=True) and
+                cmds.getAttr(family_attr) == "render"):
+            continue
+
+        id_attr = "{}.id".format(instance)
+        if cmds.getAttr(id_attr) != "pyblish.avalon.instance":
+            continue
```
BigRoy, line 3095:
```suggestion
        id_attr = "{}.id".format(instance)
        if cmds.getAttr(id_attr) != "pyblish.avalon.instance":
            continue
            
        if (
            cmds.attributeQuery('family', node=instance, exists=True) and
            cmds.getAttr("{}.family".format(instance)) == "render"
        ):
            continue
```

Re-ordered to check the `.id` attribute value first since if it's true it's also very likely it has the family attribute instead of first checking the family attribute on nodes with an `.id` attribute that might not be a publish instance.

I wonder if this way of writing is more readable for the family attribute if statement - so I've refactored that too. I believe this would also only format the string if the attribute existed to begin with.

Thoughts?

openpype/hosts/maya/api/lib.py
```diff
@@ -3053,31 +3061,52 @@ def remove_render_layer_observer():
...
+    collected_instances = cmds.ls(
+        "*.id",
+        long=True,
+        type="objectSet",
+        recursive=True,
+        objectsOnly=True
+    )
     asset_doc = get_current_project_asset()
     new_asset = asset_doc["name"]
     new_data = asset_doc["data"]
```
BigRoy, line 3077: With the function name refactored to `update_instances_frame_range` the `new_` suffix to this variable seems a bit odd. I feel like we could just refactor these variables to: `asset_name` and `asset_data`.

BigRoy (approved):
> you happy with this PR?

I am overjoyed. I like it!

I did have two minor cosmetic notes but they aren't critical.

tokejepsen (changes requested):
Lets get @BigRoy feedback in before merging.

tokejepsen (approved):
Just some code cosmetic suggestions.

BigRoy (approved):
Agree with the last cosmetics comment by @tokejepsen but other than that solid work.

openpype/hosts/maya/api/lib.py
```diff
@@ -354,13 +349,11 @@ def collect_animation_data(fps=False):
...
-        query=True, animationStartTime=True
-    )
-    frame_end_handle = cmds.playbackOptions(query=True, animationEndTime=True)
+    handle_start = cmds.playbackOptions(query=True, animationStartTime=True)
+    handle_end = cmds.playbackOptions(query=True, animationEndTime=True)
 
-    handle_start = frame_start - frame_start_handle
-    handle_end = frame_end_handle - frame_end
+    handle_start = frame_start - handle_start
+    handle_end = handle_end - frame_end
```
BigRoy, line 356: Can you revert this change so that we stick with the right naming as per this PR https://github.com/ynput/OpenPype/pull/4978#issuecomment-1551109533 ?

openpype/hosts/maya/api/lib.py
```diff
@@ -2225,42 +2212,35 @@ def get_frame_range(include_animation_range=False):
...
+        animation_start -= int(handle_start)
+        animation_end += int(handle_end)
+
+    cmds.playbackOptions(
+        minTime=frame_start,
+        maxTime=frame_end,
+        animationStartTime=animation_start,
+        animationEndTime=animation_end
+    )
+    cmds.currentTime(frame_start)
```
BigRoy, line 2233: This function is called `get_frame_range`. It should **not** adjust the scene state.

BigRoy (commented):
This PR is showing me many changes which seem to revert code that was originally better. Did resolving some merge conflicts maybe ended up the bad way?

openpype/hosts/maya/api/lib.py
```diff
@@ -3930,121 +3926,3 @@ def get_all_children(nodes):
...
-    creator_plugin = get_legacy_creator_by_name("CreateAnimation")
-    with maintained_selection():
-        cmds.select([output, controls] + roots, noExpand=True)
-        legacy_create(
-            creator_plugin,
-            name=namespace,
-            asset=asset,
-            options={"useSelection": True},
-            data={"dependencies": dependency}
-        )
```
BigRoy, line 4050: I suspect these shouldn't be removed?

BigRoy:
@tokejepsen Could you review again? There've been quite some commits with changing code.

mkolar:
Considering this is now behind a settings with and opt-in for a studio and that it's re-using existing code with some cleanup, I'm happy to merge it, provided someone can confirm the functionality itself is as intended.

mkolar (approved):

ClementHector:
Can anyone confirm that this works to be merged?

mkolar:
> Can anyone confirm that this works to be merged?

I was actually ready to merge it, but tons of commits came after the last approval, so we need to try it again.

openpype/hosts/maya/api/lib.py
tokejepsen: @friquette can we get this in?

tokejepsen (approved):
Minor docs request.

Tested successfully in Maya 2023.

tokejepsen:
@BigRoy what do you think of the code?

friquette:
New PR cleaner here: #5201

PR #3544, opened 2022-07-20
Nuke: Build workfile by template

friquette:
## Brief description
Following discussion https://github.com/pypeclub/OpenPype/discussions/2013
This PR depends on the Maya build workfile by template https://github.com/pypeclub/OpenPype/pull/3315

## Templated Build Workfile
Building a workfile using a template designed by users. Helping to assert homogeneous subsets hierarchy and imports. Template stored as file easy to define, change and customize for production needs.

**1. Make a template**

Make your template. Add nodes and everything needed for your tasks. Here is an example template for the compositing task using a placeholder.

image

The placeholder will be replaced by the nodes you want. To create a placeholder: **OpenPype > Template Builder > Create Place  Holder**

image

- **Configure placeholders**

Fill in the necessary fields (the optional fields are regex filters)

image

```
- Builder type: Wether the the placeholder should load current asset representations or linked assets representations

- Representation: Representation that will be loaded (ex: ma, abc, png, etc...)

- Family: Family of the representation to load (main, look, image, etc ...)

- Loader: Placeholder loader name that will be used to load corresponding representations

- Order: Priority for current placeholder loader (priority is lowest first, highet last)
```

- **Save your template**

**2. Configure Template**

- **Go to Studio settings > Project > Your DCC > Templated Build Settings**
- Add a profile for your task and enter path to your template

image

**3. Build your workfile**

- Open Nuke

- Build your workfile

image

PR #4797, opened 2023-04-06
Maya template builder - preserve all references when importing a template

mre7a:
## Changelog Description
When building a template with Maya template builder, we import the template and also the references inside the template file. This causes some problems:
- We cannot use the references to version assets imported by the template.
- When we import the file, the internal reference files are also imported. As a side effect, Maya complains about a reference that no longer exists. 

`// Error: file: /xxx/maya/2023.3/linux/scripts/AETemplates/AEtransformRelated.mel line 58: Reference node 'turntable_mayaSceneMain_01_RN' is not associated with a reference file.`

## Testing notes:
1. Create a template containing some references.
2. Build the workfile with template builder from openpype menu

### DIFF
--- a/openpype/hosts/maya/api/workfile_template_builder.py
+++ b/openpype/hosts/maya/api/workfile_template_builder.py
@@ -43,7 +43,13 @@ class MayaTemplateBuilder(AbstractTemplateBuilder):
             ))
 
         cmds.sets(name=PLACEHOLDER_SET, empty=True)
-        new_nodes = cmds.file(path, i=True, returnNewNodes=True)
+        new_nodes = cmds.file(
+            path,
+            i=True,
+            returnNewNodes=True,
+            preserveReferences=True,
+            loadReferenceDepth="all",
+        )
 
         cmds.setAttr(PLACEHOLDER_SET + ".hiddenInOutliner", True)
 


LiborBatek:
I think there shouldnt be any referenced files in ```Template File``` but only ```placeholders``` created via ```Template Creator```. 

So I guess its not valid request. @antirotor @iLLiCiTiT  can you confirm its like I said?

ClementHector:
@LiborBatek The purpose of the template file is to contain any type of data including placeholders. During the build phase the placeholders will be replaced by the data corresponding to the context.
We can say that it is a legitimate request.

LiborBatek:
> @LiborBatek The purpose of the template file is to contain any type of data including placeholders. During the build phase the placeholders will be replaced by the data corresponding to the context. We can say that it is a legitimate request.

My point is that template file should only/exclusively contain just placeholders....not any live reffed data. Tthas my point :)

ClementHector:
I understand your point of view but when we started developing this feature, the goal was not to have only placeholders. I invite you to consult the description of the first tickets and thoughts that led to the development of this feature.
https://github.com/ynput/OpenPype/discussions/2013
https://github.com/ynput/OpenPype/pull/2202
https://github.com/ynput/OpenPype/pull/3544
In your case it does nothing more than the `Build Workfile`

ClementHector:
If the template cannot integrate other things then for nuke we no longer have nodes that are connected to what is replaced by the placeholder.
And for maya, we can no longer put e which is replaced by the placeholder in the hierarchy.
And another point, maybe the reference didn't publish because it doesn't correspond to any asset, for example a lighting studio for asset reviews.

OlivierOsotimehin:
I agree that the best way  is to have proper loaded asset through Opype and not a "simple" reference. 
And that's a pull request that will be coming (able to load an asset from other project (library project), with the template builder).

Being able to reference an arbitrary scene enable us to quickly prototype template elements, test them through the pipeline without having to publish again, again and again a turntable remplate or a camera, etc we are building / finetuning

LiborBatek:
@mre7a I did try to test this but I have failed to replicate the conditions. Could elaborate more on it, describe your workflow/ testing steps.

for ie I dont really understand these steps you wrote:

_"When building a template with Maya template builder, we import the template and also the references inside the template file."_

If I am correct theres no any "import" step normally invovled when using ```template file```. Neither when creating the template itself nor when using that template file to create new workfile. 

Thanks.

mre7a:
@LiborBatek Sorry if I wasn't clear. 
Our template file contains not only placeholders (as expected), but also live references, and our problem is that the live references are imported into the scene and are no longer referenced.  
When building a template, the Maya template builder builds the template and also imports the references inside the template file. I use the word "import" because the link to the referenced file is broken and the references inside the model are completely integrated into the current scene. 
The template:
template_builder

The build scene:
template_builder_2

LiborBatek:
> When building a template, the Maya template builder builds the template and also imports the references inside the template file.

If I can reformat yout sentence it basically means this:

After using ```Create workfile from template file``` action, current maya ```workfile  file``` no longer contains maya references formerly present within the Template file. 

Btw thanks for the scrngrabs (these are very helpful! for debugging your issue). 

Also looking on your template setup (pic 1) you do use some sort of asset so it could be created via ```placeholders``` not standard maya reference. All should be working then. 

Does it have any special reason to create Template file like you do? Some benefits or any shortcommings when other method with placeholders?

mre7a:
@LiborBatek Thanks a lot for reformatting my sentences. 😄 
As @OlivierOsotimehin explained, I think the main reason here is that we can not use assets from other projects to build a template. 
Also 
> Being able to reference an arbitrary scene enable us to quickly prototype template elements, test them through the pipeline without having to publish again, again and again a turntable remplate or a camera, etc we are building / finetuning.

LiborBatek (approved):
After creating the workfile by ```template``` file it keeps the contained maya ```reference``` files in the scene as references.

image

LiborBatek:
There is one weird issue tho...I got each ```placeholder``` loaded 2x times with assets! 

This issue is not present in the latest ```develop``` tho. 
@mre7a  Could you try to merge develop into this PR?

LiborBatek (changes requested):
There is one weird issue tho...I got each ```placeholder``` loaded 2x times with assets! 

image

besides ```layout``` which is loaded just one time...

This issue is not present in the latest ```develop``` tho. 
@mre7a  Could you try to merge develop into this PR?

LiborBatek (approved):
Now all works as expected, references being preserved from ```Template file``` and
all assets correctly loaded...
image

openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -43,7 +43,13 @@ def import_template(self, path):
             ))
 
         cmds.sets(name=PLACEHOLDER_SET, empty=True)
-        new_nodes = cmds.file(path, i=True, returnNewNodes=True)
+        new_nodes = cmds.file(
+            path,
+            i=True,
```
tokejepsen, line 48:
```suggestion
            import=True,
```
Better to have full argument names, so there is less guess work.

tokejepsen (dismissed):
Tested in Maya 2023.

Just 1 small code cosmetic request.

openpype/hosts/maya/api/workfile_template_builder.py
mre7a: I'm afraid this one has no long name.

Screen Shot 2023-05-03 at 12 00 45

Issue #4796, opened 2023-04-06
Optional Publish Plugin: Misused active attribute


iLLiCiTiT:
**Describe the bug**
Mixin `OptionalPyblishPluginMixin` is misusing `active` attribute of pyblish plugin for default state of optional plugin. Because the logic how the optional plugins work in Publisher changed, we have to change the attribute name. That requires change of settings chemas in all plugins using `OptionalPyblishPluginMixin`.

Right now the attribute is set also on plugins which are used in hosts which do not yet use Publisher so we can't do the change now because it would break default active state of plugin in them.

**Expected behavior**
Change key in settings from `active` to something else -> my suggestion: `optional_state`. Change `OptionalPyblishPluginMixin` to look for the new attribute.

**Additional context**
When we change `active` attribute in settings we can actually use `enabled` value for `active` value on plugin. This changes will completelly break compatibility with other pyblish UIs (which probably already happened?).

Issue came up from https://github.com/ynput/OpenPype/issues/4770

[cuID:[OP-5554](https://app.clickup.com/t/863ge2yje)]

PR #4798, opened 2023-04-06
Publisher: Plugin active attribute is respected

iLLiCiTiT:
## Changelog Description
Publisher consider plugin's `active` attribute, so the plugin is not processed when `active` is set to `False`. But we use the attribute in `OptionalPyblishPluginMixin` for different purposes, so I've added hack bypass of the active state validation when plugin inherit from the mixin. This is temporary solution which cannot be changed until all hosts use Publisher otherwise global plugins would be broken. Also plugins which have `enabled` set to `False` are filtered out -> this happened only when automated settings were applied and the settings contained `"enabled"` key se to `False`.

## Additional info
This is hotfix for current state, so if it does not break anything, we can discuss full and proper fix [here](https://github.com/ynput/OpenPype/issues/4796).

## Testing notes:
- Enabled state
1. Choose/Create a context plugin which does not have settings
2. Add `enabled = False` to class definition of the plugin
3. Open Publisher (e.g. TrayPublisher) and run publishing
4. The plugin should not be processed and should not appear in UI at all

- Active state (without mixin)
1. Choose/Create a context plugin which does not inherit from `OptionalPyblishPluginMixin`
2. Add `active = False` to the plugin
3. Open Publisher (e.g. TrayPublisher) and run publishing
4. The plugin should not be processed and should be visible under skipped plugins

- Active state (with mixin)
1. Choose/Create a context plugin which does inherit from `OptionalPyblishPluginMixin`
2. Add `optional=True` and `active = False` to the plugin
3. Open Publisher (e.g. TrayPublisher)
4. The plugin should have checkbox when `Options` (`Context`) is selected in create mode
5. Run publishing
6. The plugin should be processed and should be visible in processed plugins

### DIFF
--- a/openpype/pipeline/publish/lib.py
+++ b/openpype/pipeline/publish/lib.py
@@ -354,6 +354,59 @@ def publish_plugins_discover(paths=None):
     return result
 
 
+def _get_plugin_settings(host_name, project_settings, plugin, log):
+    """Get plugin settings based on host name and plugin name.
+
+    Args:
+        host_name (str): Name of host.
+        project_settings (dict[str, Any]): Project settings.
+        plugin (pyliblish.Plugin): Plugin where settings are applied.
+        log (logging.Logger): Logger to log messages.
+
+    Returns:
+        dict[str, Any]: Plugin settings {'attribute': 'value'}.
+    """
+
+    try:
+        return (
+            project_settings
+            [host_name]
+            ["publish"]
+            [plugin.__name__]
+        )
+    except KeyError:
+        pass
+
+    # host determined from path
+    filepath = os.path.normpath(inspect.getsourcefile(plugin))
+    filepath = os.path.normpath(filepath)
+
+    split_path = filepath.split(os.path.sep)
+    if len(split_path) < 4:
+        log.warning(
+            'plugin path too short to extract host {}'.format(filepath)
+        )
+        return {}
+
+    host_from_file = split_path[-4]
+    plugin_kind = split_path[-2]
+
+    # TODO: change after all plugins are moved one level up
+    if host_from_file == "openpype":
+        host_from_file = "global"
+
+    try:
+        return (
+            project_settings
+            [host_from_file]
+            [plugin_kind]
+            [plugin.__name__]
+        )
+    except KeyError:
+        pass
+    return {}
+
+
 def filter_pyblish_plugins(plugins):
     """Pyblish plugin filter which applies OpenPype settings.
 
--- a/openpype/pipeline/publish/lib.py
+++ b/openpype/pipeline/publish/lib.py
@@ -425,21 +425,21 @@ def filter_pyblish_plugins(plugins):
     # TODO: Don't use host from 'pyblish.api' but from defined host by us.
     #   - kept becau on farm is probably used host 'shell' which propably
     #       affect how settings are applied there
-    host = pyblish.api.current_host()
+    host_name = pyblish.api.current_host()
     project_name = os.environ.get("AVALON_PROJECT")
 
-    project_setting = get_project_settings(project_name)
+    project_settings = get_project_settings(project_name)
     system_settings = get_system_settings()
 
     # iterate over plugins
     for plugin in plugins[:]:
+        # Apply settings to plugins
         if hasattr(plugin, "apply_settings"):
+            # Use classmethod 'apply_settings'
+            # - can be used to target settings from custom settings place
+            # - skip default behavior when successful
             try:
-                # Use classmethod 'apply_settings'
-                # - can be used to target settings from custom settings place
-                # - skip default behavior when successful
-                plugin.apply_settings(project_setting, system_settings)
-                continue
+                plugin.apply_settings(project_settings, system_settings)
 
             except Exception:
                 log.warning(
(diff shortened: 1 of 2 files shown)


openpype/pipeline/publish/lib.py
```diff
@@ -354,6 +354,61 @@ def publish_plugins_discover(paths=None):
...
+            ["publish"]
+            [plugin.__name__]
+        )
+    except KeyError:
+        pass
+
+    # Category determined from path
+    # - usually path is './<category>/plugins/publish/<plugin file>'
+    filepath = os.path.normpath(inspect.getsourcefile(plugin))
+    filepath = os.path.normpath(filepath)
```
BigRoy, line 384: We seem to be performing `os.path.normpath` twice. Can be skipped?

openpype/pipeline/publish/lib.py
```diff
@@ -354,6 +354,61 @@ def publish_plugins_discover(paths=None):
...
+            project_settings
+            [host_name]
+            ["publish"]
+            [plugin.__name__]
+        )
+    except KeyError:
+        pass
+
+    # Category determined from path
+    # - usually path is './<category>/plugins/publish/<plugin file>'
```
BigRoy, line 382: What's this `<category>`? Is this even still used anywhere? Or is that that intended to be the host name?

openpype/pipeline/publish/lib.py
```diff
@@ -354,6 +354,61 @@ def publish_plugins_discover(paths=None):
...
+    filepath = os.path.normpath(inspect.getsourcefile(plugin))
+    filepath = os.path.normpath(filepath)
+
+    split_path = filepath.split(os.path.sep)
```
BigRoy, line 386: This could be `.rsplit(os.path.sep, 5)` to skip splitting even further than five levels, very minor optimization.
I assume this is called quite often?

openpype/tools/publisher/control.py
```diff
@@ -2307,6 +2308,37 @@ def _publish_next_process(self):
...
+        """
+
+        if plugin.active:
+            return True
+
+        if not plugin.optional:
+            return False
+
+        if OptionalPyblishPluginMixin in inspect.getmro(plugin):
+            return True
```
BigRoy, line 2339: Wouldn't `issubclass` suffice here?

openpype/pipeline/publish/lib.py
iLLiCiTiT: I was thinking exactly the same, but I'm scared to remove.

openpype/pipeline/publish/lib.py
iLLiCiTiT: I've changed it to category (as settings category) because e.g. for deadline plugins it will be `deadline` and not host name.

openpype/tools/publisher/control.py
iLLiCiTiT: 100% yes in Python 3 DCCs. 50/50 in Python 2 DCCs.

BigRoy:
Tested this in Houdini. It seems to work as intended from a user perspective.

```python
import pyblish.api
from openpype.pipeline import OptionalPyblishPluginMixin

ORDER = pyblish.api.CollectorOrder


class PluginDisabled(pyblish.api.ContextPlugin):
    enabled = False
    order = ORDER

    def process(self, context):
        self.log.info(self.__class__.__name__)
        raise RuntimeError("SHOULD NOT RUN")


class PluginOptionalDisabled(pyblish.api.ContextPlugin):
    enabled = False
    optional = True
    order = ORDER

    def process(self, context):
        self.log.info(self.__class__.__name__)
        raise RuntimeError("SHOULD NOT RUN")


class PluginOptionalInactiveDisabled(pyblish.api.ContextPlugin):
    enabled = False
    optional = True
    active = False
    order = ORDER

    def process(self, context):
        self.log.info(self.__class__.__name__)
        raise RuntimeError("SHOULD NOT RUN")


class PluginOptionalInactiveDisabledMixin(pyblish.api.ContextPlugin,
                                          OptionalPyblishPluginMixin):
    enabled = False
    optional = True
    active = False
    order = ORDER

    def process(self, context):
        self.log.info(self.__class__.__name__)
        raise RuntimeError("SHOULD NOT RUN")


class PluginDisabledInApplySettings(pyblish.api.ContextPlugin):
    order = ORDER

    @classmethod
    def apply_settings(cls, system_settings, project_settings):
        cls.enabled = False

    def process(self, context):
        self.log.info(self.__class__.__name__)
        raise RuntimeError("SHOULD NOT RUN")


class PluginOptionalInactiveMixin(pyblish.api.ContextPlugin,
                                  OptionalPyblishPluginMixin):
    optional = True
    active = False
    order = ORDER

    def process(self, context):
        if not self.is_active(context.data):
            return
        self.log.info(self.__class__.__name__)


class PluginOptionalActiveMixin(pyblish.api.ContextPlugin,
                                OptionalPyblishPluginMixin):
    optional = True
    active = True
    order = ORDER

    def process(self, context):
        if not self.is_active(context.data):
            return
        self.log.info(self.__class__.__name__)
```

It's good to note that whenever **enabled=False** that the plug-in doesn't show at all **without warning**. It's not considered a Skipped plug-in, it just doesn't show.

The 2 optionals at the bottom can be toggled and work fine.

It would be nice if similar to the Crashed Plugins tab it might have a Disabled Plugins tab or alike so at least you can debug it.

BigRoy:
Also, a side note @iLLiCiTiT if a plug-in has no label it shows **blank** in the Details page: 

afbeelding

I believe in the old publisher it would just show `cls.__name__` or `plugin.__class__.__name__`.

PR #4811, opened 2023-04-06
Nuke: settings and optional attribute in publisher for some validators

jakubjezek001:
## Changelog Description
New publisher is supporting optional switch for plugins which is offered in Publisher in Right panel. Some plugins were missing this switch and also settings which would offer the optionality.  

## Testing notes:
1. in settings go to `project_settings/nuke/publish/ValidateBackdrop` and leave settings open. 
2. Open nuke in test workfile and create several nodes, select them and Create Nukenode family (backdrop) instance. 
3. switch to publisher and select created instance. 
4. notice in right panel is offered Validate Backdrop switch
5. Go to the settings and disable `Optional` toggle and save
6. refresh the nuke publisher and notice that the switch is removed


BigRoy (approved):
Don't use nuke myself and haven't tested it. But code changes look good to me.

64qam (approved):

PR #4815, opened 2023-04-07
Maya Template Builder - Remove default cameras from renderable cameras

mre7a:
## Changelog Description
When we build an asset workfile with build workfile from template inside Maya, we load our turntable camera. But then we end up with 2 renderables camera : **persp** the one imported from the template.
We need to remove the **persp** camera (or any other default camera) from renderable cameras when building the work file.

## Testing notes:
1. In Maya, build a scene from the template.
2. Check the render settings so that there is no persp camera in the list renderble cameras.

### DIFF
--- a/openpype/hosts/maya/api/workfile_template_builder.py
+++ b/openpype/hosts/maya/api/workfile_template_builder.py
@@ -49,7 +49,10 @@ class MayaTemplateBuilder(AbstractTemplateBuilder):
         default_cameras = [cam for cam in cmds.ls(cameras=True)
                            if cmds.camera(cam, query=True, startupCamera=True)]
         for cam in default_cameras:
-            if not cmds.objExists("{}.renderable".format(cam)):
+            if not cmds.attributeQuery("renderable", node=cam, exists=True):
+                self.log.debug(
+                    "Camera {} has no attribute 'renderable'".format(cam)
+                )
                 continue
             cmds.setAttr("{}.renderable".format(cam), 0)
 
--- a/openpype/hosts/maya/api/workfile_template_builder.py
+++ b/openpype/hosts/maya/api/workfile_template_builder.py
@@ -45,6 +45,13 @@ class MayaTemplateBuilder(AbstractTemplateBuilder):
         cmds.sets(name=PLACEHOLDER_SET, empty=True)
         new_nodes = cmds.file(path, i=True, returnNewNodes=True)
 
+        # make default cameras non-renderable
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
When creating new empty scene before ```build workfile from template``` action there is ```persp``` renderable camera.

image

After creating the workfile from template, the ```persp``` camera is not renderable anymore. Which is correct behavior.

image

tokejepsen (approved):

I'm curious about what the use case for this?
I'm guessing its because in the template file, there is already a non-default camera that enabled for rendering, so when building the workfile this camera will be the (only) renderable camera.

LiborBatek:
@tokejepsen Yeah I can image guys used templated workfiles but always had to remove those ```persp``` camera from render settings so just the one from ```template``` got rendered.

Thinking now about it if desirable that after building the workfile from template, theres no any camera set in the ```Render settings``` which would need manual labour then.

What you think guys?

tokejepsen:
> What you think guys?

Sorry, not sure what you are suggesting?
Currently this PR solves having default cameras renderable when building the workfile. In the template you can then setup which cameras should be renderable.

LiborBatek:
Well I think it doesnt set the camera properly...what it does now is to remove default persp camera from ```Renderable cameras```...

so it should set that one from ```template file``` no?

tokejepsen:
> Well I think it doesnt set the camera properly

What are you expecting?

Currently if there are any non-default camera in the template file that has its `renderable` attribute enabled, it'll be set to render in the built workfile. Setting the `renderable` attribute happens when you select the camera in the render settings.

LiborBatek:
Yeah, I re run the test and realized that ```Camera1``` in my ```Template file``` wasnt set as ```renderable camera``` so it wasnt set neither when building new workfile from that template.

When I have set it properly in the ```Template file``` all set accordingly in the new workfile...

All good! Nevermind.

ClementHector:
Is there something missing to be adopted?

LiborBatek:
> Is there something missing to be adopted?

I guess no, it can be merged! Will let any dev know so it finaly happens!

tokejepsen:
> Is there something missing to be adopted?

Just for clarification, is this the use case for this PR?

https://github.com/ynput/OpenPype/pull/4815#pullrequestreview-1405506443

ClementHector:
I guess it's good for everyone.

antirotor (approved):

PR #4818, opened 2023-04-07
Houdini: Validate VDB Input Node fix error

BigRoy:
## Changelog Description

Fix error report for Validate VDB Input node

## Additional info

Without this code change the plug-in would error when reaching the `raise` statement since the arguments weren't correct for the old class.

## Testing notes:

1. Publish VDB with invalid input node.


BigRoy:
Actually, no. Disregard this. This plug-in appears to be a duplicate of [`ValidateVDBOutputNode`](https://github.com/ynput/OpenPype/blob/d9c67a0bd50fb5c8625632d942c6bf4bf85eb908/openpype/hosts/houdini/plugins/publish/validate_vdb_output_node.py#L6-L5) but then with less validations?

PR #4819, opened 2023-04-07
Houdini: Fix Validate Output Node (VDB)

BigRoy:
## Changelog Description

- Removes plug-in that was a duplicate of this plug-in.
- Optimize logging of many prims slightly (can be improved much more, see todo)
- Fix error reporting like https://github.com/ynput/OpenPype/pull/4818 did

## Additional info

Whether this should refer to **input** or **output** node I'll leave up to discussion but all the plug-ins are named Output Node so I've kept that plug-in name for now.

## Testing notes:

1. Publish VDB with invalid input node, like e.g. a cube instead of VDB.


moonyuet:
I have done the simple test on the vdb cache main. It works well with VDB node.

image

But the validator doesn't error out as I thought when I set the SOP path to geometry

image

the validator works when the SOP path is empty
which also triggers another validator(validate output node) errors out.

image

openpype/hosts/houdini/plugins/publish/validate_vdb_output_node.py
```diff
@@ -56,12 +89,22 @@ def get_invalid(cls, instance):
         nr_of_prims = len(prims)
 
         # All primitives must be hou.VDB
-        invalid_prim = False
+        invalid_prims = []
```
moonyuet, line 92: maybe we need to introduce to check whether the node is VDB type. 
Consider the pseudo code below 
```
        main_node = hou.node(instance.data["instance_node"])
        vdb_path = main_node.evalParm("soppath")
        vdb = hou.node(vdb_path)
        if not isinstance(vdb, hou.VDB):
            cls.log.error("the output object is not VDB type")
            return [node]
```

BigRoy: Isn't that exactly what it's doing by checking whether the primitives are VDB types?
There isn't a "VDB" node type right? There's only the `hou.VDB` geometry type which is exactly what we're checking here.

However, your case of a "grid" passing just fine shouldn't happen so I'll investigate.

BigRoy:
I've made some changes @moonyuet could you try again. I've taken the liberty to improve the validation report itself a bit too. It will now also include the actual issue in the report too. Plus I've added a "Select invalid" action so you can select the ROP or SOP node directly. (It selects the ROP node if the issue is there like no valid sop path set. It selects the SOP node if the issue is in the geometry)

Here's a report screenshot:

afbeelding

I've also tweaked the logic that now it should correctly get the geometry at first frame to test against if your houdini was set to Manual mode for updating.

moonyuet:
the validator works now. 
image

The publish validation report shows why the instance doesn't pass the validator too. The select invalid option displays the debug message of 0 VDB primitive in the Houdini Console.
image

moonyuet (approved):

antirotor (approved):

PR #4821, opened 2023-04-08
Settings UI: Allow setting a size hint for text fields

iLLiCiTiT:
## Changelog Description
Text entity have `minimum_lines_count` which allows to change minimum size hint of UI input.

## Additional info
I've used line count as size hint definition because explicit size hint would not work for different resolution on different platforms. This can affect only multiline text.

## Testing notes:
1. Set `"minimum_lines_count"` on multiline text entity in settings schema
2. Open UI and validate the minim size hint is reflected


BigRoy (approved):
Looks good to me.

I wonder whether `minimum_line_count` is the best wording here. Is there a maximum? It kind of sounds like it's a validation rule to the text instead of a size hint for the UI. Should we find a better naming for the key?

iLLiCiTiT:
> Is there a maximum?

Nope, could be added too.

> Should we find a better naming for the key?

Suggestions are welcommed. E.g. `minimum_visible_line`?

BigRoy:
> Suggestions are welcommed. E.g. minimum_visible_lines?

It's not really a minimum right, because the field itself in the UI is of a fixed size. So it's really the line count that the UI just statically shows. E.g. `ui_line_count` of `display_line_count` or `ui_num_visible_lines` or whatever might make more sense? Thoughts?

iLLiCiTiT:
> OR, we should make the field resizable by the user if possible.

I was actually confused when you reported the issue because it does resize on my side when I add lines.

> It's not really a minimum right, because the field itself in the UI is of a fixed size

It is defines minimum of visible lines which affects minimum size hint. So it is minimum... + there can be added maximum of visible lines.

BigRoy:
Ah, if it acts as a minimum for you I suppose the name is fine. For me the multiline textbox UI is of fixed size, both before and after this PR the UI shows like this:

afbeelding

But after the PR the size hint can be used to make it bigger.

Looking at [this comment](https://github.com/ynput/OpenPype/pull/4655#issuecomment-1496124738) by @EmberLightVFX I assume it's also at that 3 line fixed size for him too.

BigRoy:
> With that in mind I could also add maximum lines too, to be able do more magic? Like resize up to maximum lines if minimum lines are not specified, and actually have some default like 30 lines? So it would be small if empty but could resize when needed.

I'd be fine with something like that I suppose. 50 already was quite long as you can see from my earlier screenshot, 25 or 30 should be a decent default max I suppose.

antirotor (approved):

PR #4822, opened 2023-04-08
Settings: Version settings popup fix

iLLiCiTiT:
## Changelog Description
Version completer popup have issues on some platforms, this should fix those edge cases. Also fixed issue when completer stayed shown fater reset (save).

## Additional info
Most important is that it does not make it worse for windows.

## Testing notes:
Windows & Linux & MacOs:
1. Open settings on windows
2. In System settings go to general in version selector
3. Click into input
4. The popup show up and should not cause a change of focus (infinite focus change from input to popup)
5. Reset settings UI when focus is in the input field


64qam (approved):
Works on Windows and Centos, not able to test it on Mac.

Issue #4035, opened 2022-10-27
Houdini: Switching context doesn't update variables


antirotor:
## Bug

When switching context in Houdini, sometimes `$HIP`/`$JOB` are not updated.

[cuID:[OP-4326](https://app.clickup.com/t/30u2296)]

BigRoy:
Could you elaborate on this issue?

- `$HIP` is always the workfile path and as such, how should this update?
- `$JOB` is a custom attribute definition which as far as I know OpenPype doesn't set it explicitly by default at all. In Houdini it's somewhat the equivalent to what setting workspace would be in Maya, [as be seen here for how it's used to "Set project"](https://www.sidefx.com/docs/houdini/basics/project.html). 
    - However, do note that `$JOB` defaults to the _current working directory_ that Houdini launched with which [this Houdini pre-launch hook](https://github.com/ynput/OpenPype/blob/b95e79e0d0374a3b95be5e5f21712975470dca93/openpype/hosts/houdini/hooks/set_paths.py#L17) ends up setting. Should we maybe make that more explicit by instead setting the `JOB` env var?

It's also good to note what [the docs mention about `$JOB`](https://www.sidefx.com/docs/houdini/basics/project.html#tips-and-notes):

> The value of $JOB is saved with the scene (.hip) file. So once you associate a given scene file with a project using File ▸ Set Project and save it, you don’t have to use File ▸ Set Project again the next time you open the scene file.

---

@antirotor So is this issue about updating JOB on Task change?

MustafaJafar:
I think this PR is related
https://github.com/ynput/OpenPype/pull/5532

MustafaJafar:
@BigRoy 
Hey, let's revisit this issue 

## Some Facts
### HIP
I saved Houdini file to desktop and published my job and Houdini addon oddly doesn't complain ?! 

![image](https://github.com/ynput/OpenPype/assets/20871534/a1aca95e-6ea6-42d1-a58b-ab9fca310749)


### JOB
as we discussed before, Houdini addon doesn't manage `$JOB` at all.
So, it's managed  according to Houdini's default which is as far as I know : 
1. set `$JOB` to `$HIP` 
2. users are allowed to set `$JOB` manually 
3. developers cam set `$JOB` programmatically by using `hou.hscript("set JOB=" + job)` 

----------
## Quick tests: 
### Launch Task in Houdini for the first time 
[...]

PR #4830, opened 2023-04-11
Houdini: Update $JOB variable on task change

BigRoy:
## Changelog Description

Update Houdini JOB variable on task change.

## Additional info

This intends to fix https://github.com/ynput/OpenPype/issues/4035

However, I've opened this as a **draft** because I'd like to open discussion on the usage of the `$JOB` variable. The Houdini docs for example describe to use the variable for the project and not per asset. The question becomes whether updating $JOB in OpenPype with this logic is what most studios would want or not. Or whether we should actually be setting it to a different value (e.g. project root) to begin with instead of per task context.

Tagging @gabormarinov @fabiaserra since I believe they are Houdini OpenPype users too.

## Testing notes:

1. [ ] Start Houdini (it should have `JOB` set on launch due to the `cwd` being the workdir on launch).
2. [ ] On task change (e.g. saving into another asset) it should set the JOB variable.
3. [ ] It might be good to also test loading from another asset's existing HIP file that originally had the wrong JOB variable because the task changed callback _might_ be called before the scene open and thus the opening of the scene might then overwrite the JOB var again. In that case we might want to set [`hou.allowEnvironmentToOverwriteVariable("JOB", True)`
](https://www.sidefx.com/docs/houdini/hom/hou/allowEnvironmentToOverwriteVariable.html).

### DIFF
--- a/openpype/hosts/houdini/api/pipeline.py
+++ b/openpype/hosts/houdini/api/pipeline.py
@@ -367,7 +367,7 @@ def on_task_changed():
 
     # Update Houdini JOB variable
     log.info("Updating JOB variable to: {}".format(workdir))
-    hou.hscript("set JOB={}".format(workdir.replace("\\", "/")))
+    hou.hscript("set -g JOB={}".format(workdir.replace("\\", "/")))
 
     # Dirty all nodes using the $JOB variable
     hou.hscript("varchange -V JOB")
--- a/openpype/hosts/houdini/api/pipeline.py
+++ b/openpype/hosts/houdini/api/pipeline.py
@@ -15,6 +15,7 @@ from openpype.pipeline import (
     register_creator_plugin_path,
     register_loader_plugin_path,
     AVALON_CONTAINER_ID,
+    legacy_io
 )
 from openpype.pipeline.load import any_outdated_containers
 from openpype.hosts.houdini import HOUDINI_HOST_DIR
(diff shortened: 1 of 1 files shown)


antirotor:
There is also question about `$JOB` and published workfiles used for render (as we are doing in Maya). Should it point to `work` or to `publish`? Currently when publishing workfile paths are checked for presence of variables like `$JOB` and `$HIP` by Validator, but there might be so many places other then file paths that use those variables...

BigRoy:
> Currently when publishing workfile paths are checked for presence of variables like $JOB and $HIP by Validator, but there might be so many places other then file paths that use those variables...

It's funny because this is the first thing we totally ripped out of OpenPype. We use these variables all the time - intentionally. We don't want to hardcode the workfile with them. It's fine for a published workfile to convert them to hardcoded values I suppose but for us it made no sense for actual workfile you're working in.

> There is also question about $JOB and published workfiles used for render (as we are doing in Maya). Should it point to work or to publish?

I'd say if we are programatically moving the file out of control of the artist then that's where any of the remapping needs to be considered - not in the workfile.

---

Anyway, with what @fabiaserra describes then `$JOB` is not related to the work folder or publish folder at all, it's e.g. the root of the asset or alike which _might_ be the same static path for publishes and work folders. (It doesn't have to be if someone has altered the anatomy templates to use `/publish/{asset}` and `/work/{asset}` instead.)

krishna8008:
There is this bug in 3.16.1.

The Houdini Job is changing when task changed,(verified using $JOB in parameter), but when we submit the scene file for rendering in deadline, it still consider the first task as a job. This issue is happening only when submitted to deadline, did something needs to be work on code still?

Issue #2721, opened 2022-02-14
Houdini: Creators as OpenPype nodes


antirotor:
## Feature

To align more with typical Houdini workflows it would be nice to have Creators accesible as other nodes in Houdini so they can be created without invoking menu.

🛠️ **Description work in progress**

[cuID:25xn68n]

PR #4831, opened 2023-04-12
Houdini: Allow creation of publish instances via Houdini TAB menu

BigRoy:
## Changelog Description

Register the available Creator's as houdini tools so an artist can add publish instances via the Houdini TAB node search menu from within the network editor.

afbeelding

## Additional info

- [x] They are dynamically generated on Houdini launch from the available Creator plug-ins
    - As a developer if you want to update the registered creators for the tab menu you can just trigger the install again:
      ```python
      from openpype.hosts.houdini.api.creator_node_shelves import install
      install()
      ```

- [x] They are not HDAs so don't require an HDA externally either - it just creates whatever the Creator plug-in does.
    - It's good to note that this also means when generated they are the exact publish instances as generated through the publisher UI. This PR does **not** implement new custom HDA ROPs.
- [ ] We can customize per Creator which gets shown where. E.g. Create Composite could only show when in a COP2 network.

Related to:

- This intends to fix https://github.com/ynput/OpenPype/issues/2721
- Might be related to the [Ynput forum's discussion on _Houdini Integration 2.0_](https://community.ynput.io/t/houdini-integration-2-0/123/1)

## Testing notes:

1. Launch Houdini
2. In the network editor in `/obj/geo` SOP or `/out` click TAB and search for "Create Point Cache" or alike (or click on the OpenPype submenu in the TAB menu)
3. Create some instances


antirotor (approved):
Perfect! Thanks.

PR #4832, opened 2023-04-12
Maya: Fix tile rendering with Vray - OP-5566

tokejepsen:
## Changelog Description
Fixes tile rendering with Vray.

## Testing notes:
1. Setup tile render in Maya with Vray.
2. Publish and validate render is assembled correctly.

### DIFF
--- a/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
+++ b/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
@@ -327,6 +327,11 @@ class MayaSubmitDeadline(abstract_submit_deadline.AbstractSubmitDeadline):
         job_info = copy.deepcopy(payload_job_info)
         plugin_info = copy.deepcopy(payload_plugin_info)
 
+        # Force plugin reload for vray cause the region does not get flushed
+        # between tile renders.
+        if plugin_info["Renderer"] == "vray":
+            job_info.ForceReloadPlugin = True
+
         # if we have sequence of files, we need to create tile job for
         # every frame
         job_info.TileJob = True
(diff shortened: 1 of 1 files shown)


openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
```diff
@@ -518,6 +527,11 @@ def _tile_render(self, payload):
 
         instance.data["assemblySubmissionJobs"] = assembly_job_ids
 
+        # Remove config files to avoid confusion about where data is coming
+        # from in Deadline.
+        for config_file in config_files:
+            os.remove(config_file)
```
BigRoy, line 533: We're also passing a config to `aux_files` for the payload. How do these config files differ from those? Should we have better variable naming distinction?

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
```diff
@@ -327,6 +327,11 @@ def _tile_render(self, payload):
         job_info = copy.deepcopy(payload_job_info)
         plugin_info = copy.deepcopy(payload_plugin_info)
 
+        # Force plugin reload for vray cause the region does not get flushed
+        # between tile renders.
+        if plugin_info["Renderer"] == "vray":
+            job_info.ForceReloadPlugin = True
```
BigRoy, line 333: This really sounds more like a Deadline Plugin issue which might be better solved there. Anyway, interesting catch!

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
tokejepsen: Those are the same config files, but the Draft Tile Assembler plugin only gets the config file from `aux_files`, so having it in the plugin parameters or on disk, actually makes more confusion.

openpype/modules/deadline/plugins/publish/submit_maya_deadline.py
tokejepsen: Yup, I agree but we are using built-in Draft Tile Assembler so we are at the hands of Thinkbox.

moonyuet (approved):
The code looks ok for me. Did some test with vray render and it doesn't have any error when submitting to deadline. The removal of config files doesn't affect the parameter of `aux-files` works in Job Info.
image

PR #4834, opened 2023-04-12
Houdini: Add null node as output indicator when using TAB search

BigRoy:
#### Add visual NULL node as indicator as to what the output will be.

With this PR when adding a Creator instance inside a SOP or COP2 context it now adds a null node as a placeholder for the selection instead of the actual selection.

The null in SOPs:

afbeelding

The publish instance then uses that node instead of 'selection':

afbeelding

This then now also allows to create the placeholder OUT node without having any selection to begin with.

#### Filter per creator in what node type categories it's shown 

Additionally this PR now implements filtering per Creator to decide where it is shown, for example:

- Create Point Cache shows in ROP + SOP context, not in COP2, LOPs, etc. since it's irrelevant there.
- Create Composite shows in ROP + COP2 context, not in SOP, etc. since it's irrelevant there.
- Create Camera (ABC) shows in ROP + OBJ context, not in others since it's irrelevant there.

#### Add use selection support to Create Composite

Create Composite creator didn't use the selection previously at all - it now does.

---

**Tl;DR**

1. Allow a creator plug-in to specify where it's shown for TAB search completion.
2. Add a feature where if generated in a SOP or COP2 context to add a null node as placeholder.

--- 

## Additional info

Here's a screen capture of this PR's functionality:

https://user-images.githubusercontent.com/2439881/231470018-9e827b48-53e0-4a52-959a-435656a64269.mp4

---

This is an enhancement to the feature added in #4831 

## Testing notes:

1. Launch Houdini
2. Create relevant publish instances using the TAB menu in the different use cases, e.g. in OBJ, SOP, COP2, ROP networks, etc.


openpype/hosts/houdini/api/creator_node_shelves.py
```diff
@@ -58,6 +67,33 @@ def create_interactive(creator_identifier):
...
+        )
+
+        tool_fn = {
+            hou.sopNodeTypeCategory(): soptoolutils.genericTool,
+            hou.cop2NodeTypeCategory(): cop2toolutils.genericTool
+        }.get(pwd.childTypeCategory())
+
+        if tool_fn is not None:
+            out_null = tool_fn(kwargs, "null")
+            out_null.setName("OUT_{}".format(subset_name), unique_name=True)
```
BigRoy, line 96: I think it's good to discuss whether this logic is in the correct place here. It's relatively hardcoded for all creators which could potentially want to do some other unique behavior for this case.

This logic might make more sense to be done directly in the `Creator.create` and then adding the `kwargs` as `pre_create_data` options so it could do the same filtering here from the houdini tool create scripts. 

However, I implemented it like this for now since it forced me to do less code duplication (or still putting it in the Houdini creator base class, etc.) **But** if the functionality this implements is actually preferred and we want to do more unique things per creator then likely it makes much more sense to implement this on the Creator plug-ins themselves.

openpype/hosts/houdini/api/creator_node_shelves.py
```diff
@@ -58,6 +67,33 @@ def create_interactive(creator_identifier):
...
+            task_name=context.get_current_task_name(),
+            asset_doc=get_asset_by_name(
+                project_name=context.get_current_project_name(),
+                asset_name=context.get_current_asset_name()
+            ),
+            project_name=context.get_current_project_name(),
+            host_name=context.host_name
+        )
+
+        tool_fn = {
```
fabiaserra, line 89: should we make this dictionary a module variable instead?

openpype/hosts/houdini/api/creator_node_shelves.py
```diff
@@ -135,12 +171,17 @@ def install():
 
     log.debug("Writing OpenPype Creator nodes to shelf: {}".format(filepath))
     tools = []
+
+    default_network_categories = [hou.ropNodeTypeCategory()]
```
fabiaserra, line 175: same comment as before, should it be a module variable?

openpype/hosts/houdini/plugins/create/create_composite.py
```diff
@@ -35,8 +37,20 @@ def create(self, subset_name, instance_data, pre_create_data):
             "copoutput": filepath
         }
 
+        if self.selected_nodes:
+            if len(self.selected_nodes) > 1:
+                raise CreatorError("More than one item selected.")
```
fabiaserra, line 42: could be a nice feature 😋

openpype/hosts/houdini/plugins/create/create_alembic_camera.py
```diff
@@ -47,3 +49,9 @@ def create(self, subset_name, instance_data, pre_create_data):
         self.lock_parameters(instance_node, to_lock)
 
         instance_node.parm("trange").set(1)
+
+    def get_network_categories(self):
```
fabiaserra, line 53: should we add `def get_network_categories(self)` on plugin.HoudiniCreator and only overload the creators that need some extra categories?
```
def get_network_categories(self):
   super().get_network_categories().extend(hou.objNodeTypeCategory())
```
Alternatively, I'm also thinking whether it'd be worth making these a class variable instead so you don't need to initialize the creator to get the info:
```
class CreateAlembicCamera(plugin.HoudiniCreator):
    """Single baked camera from Alembic ROP."""
    network_categories = plugin.HoudiniCreator.network_categories + [hou.objNodeTypeCategory()]

```

moonyuet:
i did some testings. They all works as expected. I do have one comment on the point cache and vdb creators although it isn't neccessary at all.
Is it possible to remove the null node with "out_{family_name}Main" when user deleting the instance through the rubbish bin button? 
When I test it and try to delete instance with the button, it seems the null node is still there.

https://user-images.githubusercontent.com/64118225/233288290-4afb04ed-9a4b-4be5-8273-a4dbba4e875a.mp4

BigRoy:
> i did some testings. They all works as expected. I do have one comment on the point cache and vdb creators although it isn't neccessary at all. Is it possible to also remove the null node with "out_{family_name}Main" when user deleting the instance through the rubbish bin button? When I test it and try to delete instance with the button, it seems the null node is still there.
> 2023-04-20.15-02-11.mp4

This is one of the reasons I wasn't too sure about this feature. Cleaning it up correctly means we'd need to track that this node was created by that creator and link it to the ROP node, etc. Not impossible, but with quite a few lines of code doable. For example we could add a parameter "created_by" on the null that is itself also linked to the ROP node and then find it through that. However, what if the user still has the ROP node but had changed the SOP path manually to something different than that `null`. Do we still want to remove it?

I've discussed it with our Lead Houdini FX artist and he said that he'd prefer the null to remain. For him it's more 'convenience' that the node gets created than really being a required part of the ROP node. As such to him it's just a regular null in his network - it just clarifies that a link is being generated to that.

If the feature that the null gets created to begin with is considered like a nice great idea. I'd say let's merge it like this and discuss maybe tweaking it later?

moonyuet:
> > i did some testings. They all works as expected. I do have one comment on the point cache and vdb creators although it isn't neccessary at all. **Is it possible to also remove the null node with "out_{family_name}Main" when user deleting the instance through the rubbish bin button?** When I test it and try to delete instance with the button, it seems the null node is still there.
> 
> This is one of the reasons I wasn't too sure about this feature. Cleaning it up correctly means we'd need to track that this node was created by that creator and link it to the ROP node, etc. Not impossible, but with quite a few lines of code doable. For example we could add a parameter "created_by" on the null that is itself also linked to the ROP node and then find it through that. However, what if the user still has the ROP node but had changed the SOP path manually to something different than that `null`. Do we still want to remove it?
> 
> I've discussed it with our Lead Houdini FX artist and he said that he'd prefer the null to remain. For him it's more 'convenience' that the node gets created than really being a required part of the ROP node. As such to him it's just a regular null in his network - it just clarifies that a link is being generated to that.
> 
> If the feature that the null gets created to begin with is considered like a nice great idea. I'd say let's merge it like this and discuss maybe tweaking it later?

I see your point and I agree let's get it merge and adjust later if there's issue.

moonyuet (approved):

openpype/hosts/houdini/plugins/create/create_composite.py
BigRoy: Haha, something for another time I'd say 🙉 🙈 🙊

openpype/hosts/houdini/plugins/create/create_alembic_camera.py
BigRoy: I've kept them functions for now but I did move the implementation.

openpype/hosts/houdini/plugins/create/create_usd.py
```diff
@@ -13,7 +15,6 @@ class CreateUSD(plugin.HoudiniCreator):
     enabled = False
 
     def create(self, subset_name, instance_data, pre_create_data):
-        import hou  # noqa
 
         instance_data.pop("active", None)
         instance_data.update({"node_type": "usd"})
```
moonyuet, line 20: i know it's not related to this PR. is the node_type "usd" from external plugin or not? Fir some reasons it's not working ang gives the error for both creation from TAB menu and that from the publisher.
image

BigRoy: Was able to reproduce here. The stack trace was:

```python
  File "S:\openpype\OpenPype\openpype\pipeline\create\context.py", line 1996, in _create_with_unified_error
    result = creator.create(*args, **kwargs)
  File "S:\openpype\OpenPype\openpype\hosts\houdini\plugins\create\create_usdrender.py", line 26, in create
    pre_create_data)  # type: CreatedInstance
  File "S:\openpype\OpenPype\openpype\hosts\houdini\api\plugin.py", line 191, in create
    self)
  File "S:\openpype\OpenPype\openpype\pipeline\create\context.py", line 914, in __init__
    data = copy.deepcopy(data or {})
  File "C:\PROGRA~1\SIDEEF~1\HOUDIN~1.435\python37\lib\copy.py", line 150, in deepcopy
    y = copier(x, memo)
  File "C:\PROGRA~1\SIDEEF~1\HOUDIN~1.435\python37\lib\copy.py", line 241, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "C:\PROGRA~1\SIDEEF~1\HOUDIN~1.435\python37\lib\copy.py", line 180, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "C:\PROGRA~1\SIDEEF~1\HOUDIN~1.435\python37\lib\copy.py", line 281, in _reconstruct
    state = deepcopy(state, memo)
  File "C:\PROGRA~1\SIDEEF~1\HOUDIN~1.435\python37\lib\copy.py", line 150, in deepcopy
    y = copier(x, memo)
  File "C:\PROGRA~1\SIDEEF~1\HOUDIN~1.435\python37\lib\copy.py", line 241, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "C:\PROGRA~1\SIDEEF~1\HOUDIN~1.435\python37\lib\copy.py", line 169, in deepcopy
    rv = reductor(4)
TypeError: can't pickle SwigPyObject objects
```

The issue is with this line:
```python
        instance_data["parent"] = hou.node("/stage")
```

`hou.Node` instances can't be deep copied and thus can't be added to the instance data for the `create` method.
It'd need to be set as transient data afterwards.

This should definitely be a separate issue and PR. Will take a quick look.

BigRoy: Created https://github.com/ynput/OpenPype/issues/4889 to track the issue.

fabiaserra: I see, sorry I should have checked this before responding to https://github.com/ynput/OpenPype/issues/4889. Couldn't we store the `/stage` string on `"parent"` instead, why do we need the `hou.Node` object?

openpype/hosts/houdini/plugins/create/create_composite.py
fabiaserra: hahaha totally

openpype/hosts/houdini/plugins/create/create_usd.py
BigRoy: > Couldn't we store the /stage string on "parent" instead, why do we need the hou.Node object?

In this case 100% we could, yes.

antirotor (approved):

PR #4841, opened 2023-04-13
AYON: Small fixes

iLLiCiTiT:
## Changelog Description
Bugsfixes and enhancements related to AYON logic. Define `BUILTIN_OCIO_ROOT` environment variable so OCIO configs are working. Use constants from ayon api instead of hardcoding them in codebase. Change process name from "openpype" to "ayon". Don't execute login dialog when application is not yet running but use `open` method instead. Fixed missing modules settings which were not taken from openpype defaults. Updated ayon api to `0.1.17`.

## Testing notes:
**Login dialog changes**
1. Login window should show if user is not yet logged in
2. It should be possible to change login when already logged in using tray

**OCIO**
Extractor for OCIO should not complain about missing path.

**Settings**
Modules which don't have enabled addon on AYON server, should not complain about missing keys in settings.


kalisp (approved):
Login/logout works.

No errors in the log noticed.

PR #4842, opened 2023-04-13
AYON: Implement integrate links publish plugin

iLLiCiTiT:
## Changelog Description
Implemented entity links get/create functions. Added new integrator which replaces v3 integrator for links.

## Additional info
Based on changes in https://github.com/ynput/OpenPype/pull/4841 . Modified ayon api with bugfix from develop.

## Testing notes:
**Publishing -> links integration**
1. Run publishing in a host with workfile instance
2. That should create links to the workfile

**Loader -> links getters**
1. Open Loader tool (e.g. from tray)
2. Go to asset with published subsets that were published with workfile (with links)
3. Select workfile subset
4. Change tab on right side of loader to `Dependency`
5. You should see dependent subsets there

### DIFF
--- a/openpype/plugins/publish/integrate_inputlinks_ayon.py
+++ b/openpype/plugins/publish/integrate_inputlinks_ayon.py
@@ -1,0 +1,159 @@
+import collections
+
+import pyblish.api
+from ayon_api import create_link, make_sure_link_type_exists
+
+from openpype import AYON_SERVER_ENABLED
+
+
+class IntegrateInputLinks(pyblish.api.ContextPlugin):
+    """Connecting version level dependency links"""
+
+    order = pyblish.api.IntegratorOrder + 0.2
+    label = "Connect Dependency InputLinks"
+
+    def process(self, context):
+        """Connect dependency links for all instances, globally
+
+        Code steps:
+        - filter instances that integrated version
+            - have "versionEntity" entry in data
+        - separate workfile instance within filtered instances
+        - when workfile instance is available:
+            - link all `loadedVersions` as input of the workfile
+            - link workfile as input of all other integrated versions
+        - link version's inputs if it's instance have "inputVersions" entry
+        -
+
+        inputVersions:
+            The "inputVersions" in instance.data should be a list of
+            version ids (str), which are the dependencies of the publishing
+            instance that should be extracted from working scene by the DCC
+            specific publish plugin.
+        """
+
+        if not AYON_SERVER_ENABLED:
+            self.log.info("Skipping, not in AYON mode")
+            return
+
+        workfile_instance, other_instances = self.split_instances(context)
+
+        # Variable where links are stored in submethods
+        new_links_by_type = collections.defaultdict(list)
+
+        self.create_workfile_links(
+            workfile_instance, other_instances, new_links_by_type)
+
+        self.create_generative_links(other_instances, new_links_by_type)
+
+        self.create_links_on_server(context, new_links_by_type)
+
+    def split_instances(self, context):
+        workfile_instance = None
+        other_instances = []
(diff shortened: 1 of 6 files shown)


openpype/plugins/publish/integrate_inputlinks_ayon.py
```diff
@@ -0,0 +1,160 @@
...
+        # link all loaded versions in scene into workfile
+        for version in loaded_versions:
+            self.add_link(
+                new_links_by_type,
+                "reference",
+                version["version"],
+                workfile_version_id,
+            )
+
+    def create_generative_links(self, other_instances, new_links_by_type):
```
mkolar, line 116: I'd say that this will be most of the time `reference` link. It's actually not 100% the case, but if a version is loaded and is tracked in the inventory, it is mose likely being pulled from the disk into current workfile, hence `reference`

`generative` is relationship between scene and it's render for example, where you don't need the scene to use the render once it's generated

mkolar (changes requested):
Looking ok codewise apart from the link type. Didn't test inside DCC

openpype/plugins/publish/integrate_inputlinks_ayon.py
iLLiCiTiT: To be honest I just "copied" code from previous integrator. If `"inputVersions"` represents loaded versions then what is different in that and these [here](https://github.com/ynput/OpenPype/pull/4842/files#diff-954c39e9258e836148aec91c00df5694700889aea4c2d1333bd09e4bc39bc308R86) ?.

iLLiCiTiT: Ok so input versions is more specific links based on scene. That is available only in fusion, houdini and maya.

BigRoy: Yes, input versions are what actually ends up being used for the instance - say I have loaded a cube and sphere in my scene but the pointcacheMain publish from my scene only contains the cube, then it doesn't have the sphere as input version.

The workfile publish however, would always have both. As such if you still want to know what was in the scene at the time you can go:
- `pointcacheMain` -> directly included content
- `pointcacheMain` -> `workfile` -> all contents

The workfile is always added into the published subset and thus is always accessible like this.

mkolar: Just a not for future this is actualy.

pointcacheMain **-gen->** directly included content
pointcacheMain **-ref->** workfile **-gen->** all contents

mkolar (approved):

openpype/plugins/publish/integrate_inputlinks_ayon.py
```diff
@@ -0,0 +1,160 @@
...
+        inputVersions:
+            The "inputVersions" in instance.data should be a list of
+            version ids (str), which are the dependencies of the publishing
+            instance that should be extracted from working scene by the DCC
+            specific publish plugin.
+        """
+
+        if not AYON_SERVER_ENABLED:
+            self.log.info("Skipping, not in AYON mode")
+            return
```
BigRoy, line 37: I know it might be a bit of a hack. But if at the end of this python file you'd do this:
```python
if not AYON_SERVER_ENABLED:
    del IntegrateInputLinksAYON
```
Then it wouldn't even show up in non-AYON mode.
We could do the same to the other integrator. Then at least you wouldn't be seeing the "other" plug-in in the UI and get confused?

(Plus processing is faster if less plug-ins hehe).

What do we think?

iLLiCiTiT: Will do separate PR.

PR #4846, opened 2023-04-14
Nuke: Legacy convertor skips deprecation warnings

iLLiCiTiT:
## Changelog Description
Nuke legacy convertor was triggering deprecated function which is causing a lot of logs which slows down whole process. Changed the convertor to skip all nodes without `AVALON_TAB` to avoid the warnings.

## Testing notes:
1. Legacy convertor should not trigger deprecation warnings if there are not any deprecated nodes in the scene
2. Legacy convertor should discover legacy nodes if they are there

### DIFF
--- a/openpype/hosts/nuke/api/lib.py
+++ b/openpype/hosts/nuke/api/lib.py
@@ -495,17 +495,17 @@ def get_avalon_knob_data(node, prefix="avalon:", create=True):
         data (dict)
     """
 
+    data = {}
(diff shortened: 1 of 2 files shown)


mkolar (approved):

mkolar:
@jakubjezek001 can you please have a look at this?

jakubjezek001 (approved):
LGTM ;)

PR #4849, opened 2023-04-17
3dsmax: move startup script logic to hook

antirotor:
## Changelog Description
Startup script for OpenPype was interfering with Open Last Workfile feature. Moving this loggic from simple command line argument in the Settings to pre-launch hook is solving the order of command line arguments and making both features work.

## Additional info
3dsmax has inconsistent behaviour about running custom startup script. The most reliable of them is passing `-U` command line argument with path to script file. But this argument can have multiple values so having it as the last one can make it work with specifying workfile.

## Testing notes:
1. Make sure there is no `-U ...` argument in Setting for 3dsmax.
2. Delete `ADSK_3DSMAX_STARTUPSCRIPTS_ADDON_DIR` env variable if present, it shouldn't be needed anymore
3. Run 3dsmax on task with already existing workfile

You should see OpenPype menu and the last workfile should open automatically.

### DIFF
--- a/openpype/hosts/max/hooks/force_startup_script.py
+++ b/openpype/hosts/max/hooks/force_startup_script.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+"""Pre-launch to force 3ds max startup script."""
+from openpype.lib import PreLaunchHook
+import os
+
+
+class ForceStartupScript(PreLaunchHook):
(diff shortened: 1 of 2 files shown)


LiborBatek (approved):
Did like 20 unique start of 3dsmax via OP Launcher with opening workfile directly and with skipping opening last workfile and all works flawlessly!

Approved.

PR #4853, opened 2023-04-18
Patchelf version locked

64qam:
For Centos dockerfile it is necessary to lock the patchelf version to the older, otherwise the build process fails.

### DIFF
--- a/Dockerfile.centos7
+++ b/Dockerfile.centos7
@@ -52,8 +52,7 @@ RUN yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.n
 
 # we need to build our own patchelf
 WORKDIR /temp-patchelf
-RUN git clone https://github.com/NixOS/patchelf.git . \
-    && git checkout 0.17.0 \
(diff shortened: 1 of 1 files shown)


Dockerfile.centos7
```diff
@@ -53,6 +53,7 @@ RUN yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.n
 # we need to build our own patchelf
 WORKDIR /temp-patchelf
 RUN git clone https://github.com/NixOS/patchelf.git . \
+    && git checkout 0.17.0 \
```
antirotor, line 56: This will do it in one command and will pull only necessary files, so it should be faster.

```suggestion
RUN git clone -b 0.17.0 --single-branch https://github.com/NixOS/patchelf.git . \
```

antirotor (changes requested):
This can make it faster, but it needs newer version of git. Please disregard if it doesn't work in centos 7 version of git

antirotor (approved):

PR #4858, opened 2023-04-18
Maya: Don't change time slider ranges in `get_frame_range`

BigRoy:
## Changelog Description

Don't change time slider ranges in `get_frame_range`

## Additional info

The bug was introduced [here](https://github.com/ynput/OpenPype/pull/4664/files#r1170416917).

It shouldn't have made much difference since `get_frame_range` appears to ONLY be used in `reset_frame_range` but at least this now is consistent with the function's name.

## Testing notes:

1. Launch Maya
2. Reset frame range should work fine for the different profile support implemented with https://github.com/ynput/OpenPype/pull/4664

### DIFF
--- a/openpype/hosts/maya/api/lib.py
+++ b/openpype/hosts/maya/api/lib.py
@@ -2153,17 +2153,23 @@ def set_scene_resolution(width, height, pixelAspect):
     cmds.setAttr("%s.pixelAspect" % control_node, pixelAspect)
 
 
-def get_frame_range():
-    """Get the current assets frame range and handles."""
+def get_frame_range(include_animation_range=False):
+    """Get the current assets frame range and handles.
+
+    Args:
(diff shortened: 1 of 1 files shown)


tokejepsen (approved):
Works!

openpype/hosts/maya/api/lib.py
```diff
@@ -2175,32 +2181,39 @@ def get_frame_range():
...
-    )
-    cmds.currentTime(frame_start)
-
-    return {
+    frame_range = {
         "frameStart": frame_start,
         "frameEnd": frame_end,
         "handleStart": handle_start,
-        "handleEnd": handle_end
+        "handleEnd": handle_end,
```
BigRoy, line 2188:
```suggestion
        "handleEnd": handle_end
```

PR #4859, opened 2023-04-18
Maya: Load Assembly fix invalid imports 

BigRoy:
## Changelog Description

Refactors imports so they are now correct.

## Additional info

Likely refactoring of the code disrupted this before.
The bug was reported on discord [here](https://discord.com/channels/517362899170230292/563751989075378201/1097953320338985121)

## Testing notes:

1. Publish and load setdress.
2. Publish and load assembly.


tokejepsen (commented):
The original issue reported has been resolved with this PR. It was only the `Assembly` family that was failing to load. `Setdress` loads successfully in latest `develop`.

But I'm getting the error below when loading the `Assembly`, but I think its unrelated to the original issue, so maybe we should tackle that in a different PR?
```
During load error happened on Subset: "assemblyMain" Representation: "json" Version: 1

Error message: No object matches name: code_review_02_:assemblyMain|code_review_02_:OP_4859_cant_assign_shaders_to_the_ASS_file_modelMain_01:_GRP

  File "C:\Users\tokejepsen\OpenPype\openpype\tools\loader\widgets.py", line 1590, in _load_representations_by_loader
    load_with_repre_context(
  File "C:\Users\tokejepsen\OpenPype\openpype\pipeline\load\utils.py", line 308, in load_with_repre_context
  File "C:\Users\tokejepsen\OpenPype\openpype\hosts\maya\plugins\load\load_assembly.py", line 32, in load
    containers = setdress.load_package(
  File "C:\Users\tokejepsen\OpenPype\openpype\hosts\maya\api\setdress.py", line 151, in load_package
    container = _add(instance=instance,
  File "C:\Users\tokejepsen\OpenPype\openpype\hosts\maya\api\setdress.py", line 220, in _add
    cmds.parent(loaded_root, parent, relative=True)
ValueError: No object matches name: code_review_02_:assemblyMain|code_review_02_:OP_4859_cant_assign_shaders_to_the_ASS_file_modelMain_01:_GRP
```

BigRoy:
Yes, no idea whether `assembly` should still work or not. This was just quickly fixing the reported issue, we can create an issue to track what the new error is and create a PR for that then? I don't have much incentive to really dig through `assembly` since we're not using it on our end but feel free to push fixes into this branch PR if you'd want to continue fixes here.

tokejepsen (approved):

LiborBatek (changes requested):
Did create ```Assembly``` Instance with loaded ```model``` instances and it seems due to GPR nodes doesnt have cbIds it fails during validation phase...when trying to Publish it

image

BigRoy:
> it seems due to GPR nodes doesnt have cbIds it fails during validation phase...when trying to Publish it

I believe this just means you'd need to re-save your scene and try again? :) IDs get added on scene save.
The documentation at the top of the validator, which you have collapsed for some reason should state that too.

LiborBatek:
> > it seems due to GPR nodes doesnt have cbIds it fails during validation phase...when trying to Publish it
> 
> I believe this just means you'd need to re-save your scene and try again? :) IDs get added on scene save. The documentation at the top of the validator, which you have collapsed for some reason should state that too.

What I remember now is that  I have triggered Repair Action for IDs and it didnt help either...

BigRoy:
The loaded cube models, where did they come from? It seems to complain about the referenced nodes to not have `cbId` attributes. In particular the couch (modelMain2).

LiborBatek:
They are all loaded via OP Loader so they have been properly published before...

As I understand an ```Assembly``` family being made from other loaded assets (something like multi user collaboration on model composed of multiple model parts aka published asset models.)

Am I correct?

LiborBatek:
Speaking of ```Assembly``` instance there is also ```Assembly Model Transforms``` validator for any transforms present on the Transform nodes. 

It doesnt seem much sensible to be forced to freeze all transform on any Assembly part. Imagine several artist collaborating on multiple parts of the Assembly being forced by OP to freeze all transforms when publishing the models. Pretty limiting imho.

@BigRoy  What do you think?

antirotor:
> Speaking of `Assembly` instance there is also `Assembly Model Transforms` validator for any transforms present on the Transform nodes.

I guess this is also for another issue.

tokejepsen:
@LiborBatek could you test for the original issue and report back?
We'll merge this PR if the original issue is resolved and tackle other issues separately.

LiborBatek (changes requested):
I was able to publish ```Assembly``` succesfully

image

But It failed during ```Load``` via Loader

image

tokejepsen:
@LiborBatek that issue has been logged with #4876.

Think we should merge this PR and tackle the load issue separately.

LiborBatek (approved):
Features valid for this PR working fine. Rest will be addressed in other PR.

PR #4872, opened 2023-04-19
Maya: Hide animation instance in creator

BigRoy:
## Changelog Description

- Hide animation instance in creator
- Add inventory action to recreate animation publish instance for loaded rigs

## Additional info

The animation family was often 'misused' by new users as a pointcache output but was originally designed for rigs only. This hides the animation instance but preserves the rig -> animation instance behavior.

It also adds an inventory action to recreate the animation instance if it was accidentally deleted.

afbeelding

## Testing notes:

1. Load a rig - animation instance should be created
2. Recreate animation instance through scene inventory / manager actions.


mkolar:
I like this! can't test right now, but it's a good idea

tokejepsen (changes requested):

Could we get some documentation as well?

tokejepsen (approved):
Great stuff!

antirotor (approved):

PR #4877, opened 2023-04-20
Load Rig References - Change Rig to Animation in Animation instance

mre7a:
## Changelog Description
We are using the template builder to build an animation scene. All the rig placeholders are imported correctly, but the automatically created animation instances retain the rig family in their names and subsets. In our example, we need animationMain instead of rigMain, because this name will be used in the following steps like lighting.

animation_instance (1)

animation_instance

Here is the result we need. 
image

I checked, and it's not a template builder problem, because even if I load a rig as a reference, the result is the same. For me, since we are in the animation instance, it makes more sense to have animation instead of rig in the name. The naming is just fine if we use create from the Openpype menu.

## Testing notes:
1. In Maya, in the animation task, use the template builder to build a template that contains rig placeholders. 
2. The created animation instance has Rig in its name instead of Animation. 

Note that you should add subset in the maya costum namespace settings.  
image

### DIFF
--- a/openpype/hosts/maya/api/lib.py
+++ b/openpype/hosts/maya/api/lib.py
@@ -3937,7 +3937,9 @@ def get_capture_preset(task_name, task_type, subset, project_settings, log):
     return capture_preset or {}
 
 
-def create_rig_animation_instance(nodes, context, namespace, options, log=None):
+def create_rig_animation_instance(
+    nodes, context, namespace, options, log=None
+):
     """Create an animation publish instance for loaded rigs.
 
     See the RecreateRigAnimationInstance inventory action on how to use this
@@ -3971,6 +3973,12 @@ def create_rig_animation_instance(nodes, context, namespace, options, log=None):
     asset = legacy_io.Session["AVALON_ASSET"]
     dependency = str(context["representation"]["_id"])
 
+    custom_subset = options.get("animationSubsetName")
+
+    if custom_subset:
+        rig_subset = context['subset']['name']
+        namespace = namespace.replace(rig_subset, custom_subset)
+
     if log:
         log.info("Creating subset: {}".format(namespace))
 
@@ -3982,6 +3990,6 @@ def create_rig_animation_instance(nodes, context, namespace, options, log=None):
             creator_plugin,
             name=namespace,
             asset=asset,
-            options=options.update({"useSelection": True}),
+            options={"useSelection": True},
             data={"dependencies": dependency}
         )
--- a/openpype/hosts/maya/api/lib.py
+++ b/openpype/hosts/maya/api/lib.py
@@ -190,6 +190,44 @@ def maintained_selection():
             cmds.select(clear=True)
 
 
+def get_custom_namespace(custom_namespace):
+    """Return unique namespace.
+
+    The input namespace can contain a single group
+    of '#' number tokens to indicate where the namespace's
+    unique index should go. The amount of tokens defines
+    the zero padding of the number, e.g ### turns into 001.
+
+    Warning: Note that a namespace will always be
+        prefixed with a _ if it starts with a digit
+
+    Example:
+        >>> get_custom_namespace("myspace_##_")
+        # myspace_01_
+        >>> get_custom_namespace("##_myspace")
+        # _01_myspace
+        >>> get_custom_namespace("myspace##")
+        # myspace01
+
+    """
+    split = re.split("([#]+)", custom_namespace, 1)
+
+    if len(split) == 3:
+        base, padding, suffix = split
+        padding = "%0{}d".format(len(padding))
+    else:
+        base = split[0]
+        padding = "%02d"  # default padding
+        suffix = ""
+
+    return unique_namespace(
+        base,
+        format=padding,
+        prefix="_" if not base or base[0].isdigit() else "",
+        suffix=suffix
+    )
+
+
 def unique_namespace(namespace, format="%02d", prefix="", suffix=""):
     """Return unique namespace
 
@@ -3974,10 +4012,21 @@ def create_rig_animation_instance(
     dependency = str(context["representation"]["_id"])
 
     custom_subset = options.get("animationSubsetName")
-
     if custom_subset:
-        rig_subset = context['subset']['name']
-        namespace = namespace.replace(rig_subset, custom_subset)
+        formatting_data = {
+            "asset_name": context['asset']['name'],
+            "asset_type": context['asset']['type'],
+            "subset": context['subset']['name'],
+            "family": (
+                context['subset']['data'].get('family') or
+                context['subset']['data']['families'][0]
+            )
+        }
+        namespace = get_custom_namespace(
+            custom_subset.format(
+                **formatting_data
+            )
+        )
 
     if log:
         log.info("Creating subset: {}".format(namespace))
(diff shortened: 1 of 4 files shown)


openpype/hosts/maya/plugins/load/load_reference.py
```diff
@@ -242,11 +242,15 @@ def _post_process_rig(self, name, namespace, context, options):
         creator_plugin = get_legacy_creator_by_name(
             self.animation_creator_name
         )
+        animation_subset_name = namespace.replace(
+            name,
+            name.replace('rig', 'animation')
+        )
```
BigRoy, line 248: Interesting, but also way too hardcoded I'd say. I wonder if there are better ways to find the right subset name for animation instances. @tokejepsen @mkolar any thoughts?

mre7a: @BigRoy Yes I know. You suggested I use something like this, but I haven't found the right place to set animationSubsetName in the options. :)
```
        name = options.get("animationSubsetName", namespace)
        with maintained_selection():
            cmds.select([output, controls] + roots, noExpand=True)
            legacy_create(
                creator_plugin,
                name=name,
                asset=asset,
                options={"useSelection": True},
                data={"dependencies": dependency}
            )
```

BigRoy: That's indeed what I recommended for the case you wanted to customize the naming purely from the workfile template system since I believe that system allows to pass custom options to the creators as it builds it. (I believe there's an options field that can take a JSON string with options?)

mre7a: Exactly. I found "Loader Arguments" for this purpose, but in the end, it was not what I was looking for to solve our subset name problem.

BigRoy: I'd say this is stuck due to it being hardcoded yes. However, others might have @antirotor @mkolar @jakubjezek001 decent solutions to propose for this?

ClementHector:
@BigRoy @mre7a How about give the choice to do the _post_process_rig (which is a specific exception)
To do this, let's add an option in the settings to run this post process (enable by default)
And @mre7a  could put an animation set in his template, could this solve the problem?

tokejepsen (changes requested):
Needs a more flexible workflow.

ClementHector:
@tokejepsen I really like this idea, I agree it seems to be the best solution and don't forget to keep the default behavior.

openpype/hosts/maya/api/lib.py
```diff
@@ -3982,6 +3982,6 @@ def create_rig_animation_instance(nodes, context, namespace, log=None):
             creator_plugin,
             name=namespace,
             asset=asset,
-            options={"useSelection": True},
+            options=options.update({"useSelection": True}),
```
BigRoy, line 3985: Update doesn't return the dict itself. This wouldn't work. Also the mutable `options` dict I believe is re-used for multiselection loading so I suspect this shouldn't be done like this.

````python
create_options = options.copy()
create_options["useSelection"] = True
```
And potentially a `copy.deepcopy` might be even safer.

mre7a: Thank you, @BigRoy. Finally, we do not need to change options here.

BigRoy: I'd argue the opposite actually. We still want to make sure `useSelection: True` is passed as an option even if the rig loader was called with it False.

mre7a:
@BigRoy @tokejepsen  At the end, it is similar to your proposal that was made from the beginning. 
I did some modifications to make it more flexible. However, I think it is possible to rebuild the namespace from the scratch using get_custom_namespace(custom_namespace) 
https://github.com/quadproduction/OpenPype/blob/develop/openpype/hosts/maya/api/plugin.py#L87

openpype/hosts/maya/api/lib.py
```diff
@@ -3971,6 +3973,12 @@ def create_rig_animation_instance(nodes, context, namespace, log=None):
     asset = legacy_io.Session["AVALON_ASSET"]
     dependency = str(context["representation"]["_id"])
 
+    custom_subset = options.get("animationSubsetName")
+
+    if custom_subset:
+        rig_subset = context['subset']['name']
+        namespace = namespace.replace(rig_subset, custom_subset)
```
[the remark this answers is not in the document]
mre7a: Oops, did not see this one. I have just proposed the same thing.

tokejepsen:
@mre7a flip this PR to "Ready to review" when you are done, so we can review it properly.

> However, I think it is possible to rebuild the namespace from the scratch using get_custom_namespace(custom_namespace)

I think you are right. Might be able to use it.

openpype/hosts/maya/api/plugin.py
```diff
@@ -84,44 +84,6 @@ def get_reference_node_parents(ref):
...
-        base = split[0]
-        padding = "%02d"  # default padding
-        suffix = ""
-
-    return lib.unique_namespace(
-        base,
-        format=padding,
-        prefix="_" if not base or base[0].isdigit() else "",
-        suffix=suffix
-    )
```
BigRoy, line 122: Did you forget to reimport this function into `plugin.py`? The `ReferenceLoader` in this file is using this function - with this moved I suspect this will now break.

openpype/hosts/maya/api/lib.py
```diff
@@ -3937,7 +3975,9 @@ def get_capture_preset(task_name, task_type, subset, project_settings, log):
     return capture_preset or {}
 
 
-def create_rig_animation_instance(nodes, context, namespace, log=None):
+def create_rig_animation_instance(
+    nodes, context, namespace, options, log=None
+):
```
BigRoy, line 3980: Can we add `options` argument to the docstring?

BigRoy (commented):
Almost there. Just found these notes.

openpype/hosts/maya/api/plugin.py
mre7a: I thought we import every thing from lib.
https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/maya/api/plugin.py#L19
That's one of the reasons I moved the function.

BigRoy: We import the `lib` but not the functions directly. You'll need to change the calls to it now by doing `lib.get_custom_namespace`. So you should refactor that lower in this file to that.

mre7a: Oops. Sorry. Fixed that one.

BigRoy (approved):
Code changes look good to me. @tokejepsen could you do a test run?

tokejepsen (changes requested):
Tested in Maya 2023 successfully!

Looks great :) Could we get some documentation so people can start to use this?

tokejepsen (changes requested):

tokejepsen (approved):
Nice work!

tokejepsen (changes requested):
Sorry, would still like to see some documentation for this, as stuff like what to input for the Loader arguments (`animationSubsetName`) might be forgotten.

mre7a:
> Sorry, would still like to see some documentation for this, as stuff like what to input for the Loader arguments (`animationSubsetName`) might be forgotten.

I'll do it for sure :)

tokejepsen (approved):

antirotor (approved):

PR #4878, opened 2023-04-20
Maya: Looks - calculate hash for tx texture

antirotor:
## Changelog Description
Texture hash is calculated for textures used in published look and it is used as key in dictionary. In recent changes, this hash is not calculated for TX files, resulting in `None` value as key in dictionary, crashing publishing. This PR is adding texture hash for TX files to solve that issue.

## Additional info
In case of existing TX files the hash should be taken from TX file itself? Also, this needs to be changed anyway for AYON as we can't store hash information on version the same way we are doing it in OpenPype.

## Testing notes:
1. In Maya, create look that is using already existing TX files
2. Publish the look, it should be published ok.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/extract_look.py
+++ b/openpype/hosts/maya/plugins/publish/extract_look.py
@@ -280,7 +280,7 @@ class MakeTX(TextureProcessor):
             # Do nothing if the source file is already a .tx file.
             return TextureResult(
                 path=source,
-                file_hash=None,     # todo: unknown texture hash?
+                file_hash=source_hash(source),
                 colorspace=colorspace,
                 transfer_mode=COPY
             )


antirotor:
> Seems like a hotfix and like a decent one I suppose.

Do you remember the reason for not setting the hash there? I am thinking about handling the hashes for tx files - they are using their own ID to know if they need to regenerate it, stored in tif/tx metadata. Maybe we should adopt that - anyway this is something we'll need to re-think in new look system and AYON.

BigRoy:
> Do you remember the reason for not setting the hash there? I am thinking about handling the hashes for tx files - they are using their own ID to know if they need to regenerate it, stored in tif/tx metadata

Well it was mostly anticipating that the `tx` might have been generated from a previous publish already (and then used) and that the `tx` file itself might already have a hash for that file contained in its metadata.

LiborBatek (approved):
All works as expected.

PR #4879, opened 2023-04-20
Project packager: Backup and restore can store only database

iLLiCiTiT:
## Changelog Description
Pack project functionality have option to zip only project database without project files. Unpack project can skip project copy if the folder is not found.

Added helper functions to `openpype.client.mongo` that can be also used for tests as replacement of mongo dump.

## Testing notes:
1. Run `openpype_console pack-project --project YourProjectName --dbonly`
2. Look to project root folder where should be a zip file with only `metadata.json` and `database.json`
3. Now try to reverse the project using `openpype_console unpack-project --zipfile C:/path/to/your/file.zip`
4. Project should be recreated in mongo

### DIFF
--- a/openpype/cli.py
+++ b/openpype/cli.py
@@ -415,11 +415,12 @@ def repack_version(directory):
 @main.command()
 @click.option("--project", help="Project name")
 @click.option(
-    "--dirpath", help="Directory where package is stored", default=None
-)
-def pack_project(project, dirpath):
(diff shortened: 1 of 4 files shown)


LiborBatek (approved):
Both commands ```pack_project``` and ```unpack_project``` with argument ```--dbonly```

works as expected producing just mongoDB files not project assets.

PR #4881, opened 2023-04-20
Houdini: ExtractOpenGL for Review instance not optional

### LINKED ISSUE #4868: Bug: Houdini Extract OpenGL is optional but doesn't make any sense

### Current Behavior:

The review publish instance implemented with #3839 has the ExtractOpenGL node set to "optional" but for a review instance that doesn't really make any sense since it's the only extractor for it and whenever you disable it then Integration fails because the instance has no representations to publish because it created nothing.

### Expected Behavior:

I think the extractor should not be optional.

### Version

3.15.4-nightly.3

Windows

### Steps To Reproduce:

1. Start Houdini
2. Create a review instance
3. On the review instance on publish tab disable "Extract OpenGL"
4. Publish
5. It will fail.

### Relevant log output:

_No response_

### Additional context:

_No response_

[cuID:[OP-5756](https://app.clickup.com/t/861mp7g8f)]

BigRoy:
## Changelog Description

Don't make ExtractOpenGL optional for review instance optional.

## Additional info

Fixes #4868

## Testing notes:

1. Publish review instance from Houdini
2. Extract OpenGL is not optional.
3. Publishing should work fine.


moonyuet (approved):
the code looks good. Also double check in the publisher, the publisher set without the optional as expected.
image

antirotor (approved):

PR #4885, opened 2023-04-20
Publisher: Clear comment on successful publish and on window close

### LINKED ISSUE #4867: Bug: Publisher comment field doesn't clear itself

### Current Behavior:

The new publisher UI always persists the value in the comment field.

On reset it remains, on closing the UI and reopening it the value remains and after a succesfull publish it remains.

I could have published something in SH010, then open SH020 to also publish it but the comment from SH010 is still there. Similarly it's weird that closing and reopening or even a successful publish keeps the comment value even though it was intended for that particular publish only.

### Expected Behavior:

The old publisher would clear out the comment field on close+reopen and on a successful publish. I'd expect that behavior now as well.

The comment field doesn't have to be cleared on reset if the current context hasn't changed (e.g. workfile is still the same or asset/task is still the same). If there were a switch I'd likely expect it to be cleared too.

### Version

3.15.4-nightly.3

Windows

### Steps To Reproduce:

1. Open publisher
2. Type a comment
3. Reset, publish, close and reopen. The UI will always persist the comment field value, which is unexpected.

### Relevant log output:

_No response_

### Additional context:

_No response_

[cuID:[OP-5755](https://app.clickup.com/t/861mp7bpj)]

BigRoy:
## Changelog Description

Clear comment text field on successful publish and on window close.

## Additional info

Resolves #4867

## Testing notes:

1. Open new publisher
2. Write a comment in field at bottom
3. On window close + reopen the comment should be removed
4. Write a comment in field at bottom
5. On successful publish the comment should be removed

### DIFF
--- a/openpype/tools/publisher/window.py
+++ b/openpype/tools/publisher/window.py
@@ -285,6 +285,9 @@ class PublisherWindow(QtWidgets.QDialog):
             "publish.has_validated.changed", self._on_publish_validated_change
         )
         controller.event_system.add_callback(
+            "publish.finished.changed", self._on_publish_finished_change
+        )
+        controller.event_system.add_callback(
             "publish.process.stopped", self._on_publish_stop
         )
         controller.event_system.add_callback(
@@ -400,6 +403,7 @@ class PublisherWindow(QtWidgets.QDialog):
         # TODO capture changes and ask user if wants to save changes on close
         if not self._controller.host_context_has_changed:
             self._save_changes(False)
+        self._comment_input.setText("")  # clear comment
         self._reset_on_show = True
         self._controller.clear_thumbnail_temp_dir_path()
         super(PublisherWindow, self).closeEvent(event)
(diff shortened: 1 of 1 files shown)


moonyuet:
I tested with houdini, looks good.
https://user-images.githubusercontent.com/64118225/233547780-9a2d03c8-7805-409f-86e2-489971778915.mp4

moonyuet (approved):

openpype/tools/publisher/window.py
```diff
@@ -400,6 +403,7 @@ def closeEvent(self, event):
         # TODO capture changes and ask user if wants to save changes on close
         if not self._controller.host_context_has_changed:
             self._save_changes(False)
+        self._comment_input.setText("")  # clear comment
```
iLLiCiTiT, line 406: Should it be cleared on close?

BigRoy: It got this reported from an artist, and I'd say yes. If it didn't it would also mean that when you reopen it later, which could be by the time you're working in another workfile in another shot then it'd still have the much older comment.

Aside of what the artist reported, I'd say so too, yes. It's weird that closing the UI and reopening it doesn't start it fresh I'd say.

PR #4886, opened 2023-04-20
Publisher: Make sure to reset asset widget when hidden and reshown

### LINKED ISSUE #4861: Bug: Available assets in context switch in new publisher don't update during session

### Current Behavior:

In the new publisher you can customize the asset/task for an instance.

afbeelding

However, the select asset pop-up doesn't update the asset list from the moment the publisher was first launched in the current session. For example, in Houdini, I open the publisher once.

In the meantime I, or someone else, creates a new asset in the pipeline. The new asset does NOT show up in that pop-up list until I completely restart the host to re-initialize the publisher completely.

Interestingly enough the Create tab does show the new asset left hand side - the issue seems to be mainly with the "Select asset" popup.

### Expected Behavior:

Asset list should be up-to-date

### Version

3.15.4-nightly.3

Windows

### Steps To Reproduce:

1. Launch Houdini, open publisher.
2. Create a new asset.
3. Go to an existing instance, customize the "asset" field by clicking on the right hand square icon to open the 'set asset' popup.
4. The pop-up does not show the new asset.

### Relevant log output:

_No response_

### Additional context:

_No response_

[cuID:[OP-5747](https://app.clickup.com/t/863gh9kmk)]

BigRoy:
## Changelog Description

Make sure to reset asset widget when hidden and reshown. Without this the asset list would never refresh in the set asset widget when changing context on an existing instance and thus would not show new assets from after the first time launching that widget.

## Additional info

Resolves #4861 

## Testing notes:

1. Open publisher
2. Change asset on an existing publish instance using the "select asset widget" in the publisher.
3. Create a new asset in the project (through e.g. project manager or from ftrack)
4. Try and change asset on existing instance again to the newly created asset.
  - Before this PR the new asset would not be listed. With this PR it will be refreshed.


openpype/tools/publisher/widgets/assets_widget.py
```diff
@@ -264,6 +264,11 @@ def reset(self, force=True):
 
         self._model.reset()
 
+    def hideEvent(self, event):
+        # Make sure to reset when we close and reopen
+        self._soft_reset_enabled = True
```
iLLiCiTiT, line 269: I would say it is overkill to reset model on each show event.

iLLiCiTiT: Made PR https://github.com/BigRoy/OpenPype/pull/12

BigRoy: Thanks, merged it. Looks good. Would require the publisher to "reset" before it refreshes the asset list, but I suppose that's totally fine as well and will result in less database calls.

iLLiCiTiT (approved):
LGTM

PR #4888, opened 2023-04-21
Add fps as instance.data in collect review in Houdini.

moonyuet:
## Changelog Description
fix the bug of failing to publish extract review in Houdini

Original error:
```python
  File "OpenPype\build\exe.win-amd64-3.9\openpype\plugins\publish\extract_review.py", line 516, in prepare_temp_data
    "fps": float(instance.data["fps"]),
KeyError: 'fps'
```

## Testing notes:
1. Launch Houdini
2. Create Review Instance
3. Publish

### DIFF
--- a/openpype/hosts/houdini/plugins/publish/collect_review_data.py
+++ b/openpype/hosts/houdini/plugins/publish/collect_review_data.py
@@ -17,6 +17,7 @@ class CollectHoudiniReviewData(pyblish.api.InstancePlugin):
         # which isn't the actual frame range that this instance renders.
         instance.data["handleStart"] = 0
         instance.data["handleEnd"] = 0
+        instance.data["fps"] = hou.fps()
 
         # Get the camera from the rop node to collect the focal length
         ropnode_path = instance.data["instance_node"]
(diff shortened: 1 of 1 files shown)


openpype/hosts/houdini/plugins/publish/collect_review_data.py
```diff
@@ -17,6 +17,7 @@ def process(self, instance):
         # which isn't the actual frame range that this instance renders.
         instance.data["handleStart"] = 0
         instance.data["handleEnd"] = 0
+        instance.data["fps"] = hou.fps()
```
BigRoy, line 20: Looks good to me! Note that there's also [CollectWorksceneFPS](https://github.com/ynput/OpenPype/blob/7b2932c9a78aa0e829a6bd06b148eb29a6f31e7d/openpype/hosts/houdini/plugins/publish/collect_workscene_fps.py#L5) collector and thus we could also do `instance.context.data["fps"]`.

I'm fine with either way however. Maybe @antirotor or @iLLiCiTiT has a preference.

BigRoy (approved):
Nice catch, thanks!

openpype/hosts/houdini/plugins/publish/collect_review_data.py
antirotor: I agree, in case of review we can safely use the value from context. I hope we'll never find use cases where you would need to set different fps in one work file per instance :)

BigRoy (approved):
❤️

Issue #4869, opened 2023-04-19
Bug: Houdini Review instance without valid camera errors on collect review data


BigRoy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

The publishing for a Houdini review instance will fail with "It is not your fault" whenever you have no valid camera set on the review node.

![image (1)](https://user-images.githubusercontent.com/2439881/233102802-0ae7123c-3813-46ad-a754-12e128786c6d.png)
![image](https://user-images.githubusercontent.com/2439881/233102805-0affda48-c219-41b6-9f30-3ce439a481e5.png)

### Expected Behavior:

The collectors should run fine and a validator should capture the issue with a nice validation report.

### Version

3.15.4-nightly.3

### What platform you are running OpenPype on?

Windows

### Steps To Reproduce:

1. Create review instance, do not set a valid camera
2. Run publish.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

[cuID:[OP-5757](https://app.clickup.com/t/861mp7jd4)]

antirotor:
This gets slightly more complicated when you switch it to LOPs mode. Then the camera parameter changes to `cameraprim` and is expected to hold camera primitive from stage.

It does work in slightly convoluted way that if you use **import camera prim** in object level and set it to OpenGL node as camera before switching it to LOPs mode, the collector will still collect proper camera attributes even though it comes in from unexpected way. I am not sure if there is easy way to get the same attributes we are getting from camera node from the camera primitive.

BigRoy:
Thanks, you are right. 

I hadn't even seen on the OpenGL node you could set it to LOPs too, that's perfect. I'd say it's a separate issue from this one but supporting LOPs should definitely be on the todo list. We definitely shouldn't be importing camera to object level since it just means we're not evaluating it in LOPs.

PR #4890, opened 2023-04-21
Houdini: OpenGL Review support LOPs/Solaris source mode

BigRoy:
## Changelog Description

Allow reviews using the openGL node to use LOP Paths.

## Additional info

Based on comment [here](https://github.com/ynput/OpenPype/issues/4869#issuecomment-1517572996) that OpenGL review node publishing didn't support Solaris reviews yet using LOP paths.

TODO:

- [ ] Update creator to allow 'use selection' for LOP paths and then have it auto-set to LOPs source mode.
- [x] Update collector to collect focal lengths for LOPs camera
  - [ ] Make sure it always collects the correct data (see todo in code)
- [ ] Update validator to allow LOPs scene path, etc.
- [ ] Test 

## Testing notes:

1. Set up openGL review instance to use LOP path
2. Publish review
3. Set up openGL review instance to use OBJ camera path
4. Publish review

Make sure to publish with reviews enabled so it also tests the focal length burnin.


moonyuet:
checked the collector, it seems that it doesn't collect the correct focal length from camera in LOP networks 
image

moonyuet:
@BigRoy I have updated the code for validating LOP networks and both add LOP networks and cameras into the opengl if the use selection enabled.
I also add an options for user to enable "LOP as source"(not sure if it is a good name). If they want to use LOP networks for review, they can enable it in the creator.
image

BigRoy:
> I also add an options for user to enable "LOP as source"(not sure if it is a good name). If they want to use LOP networks for review, they can enable it in the creator.

Can't we match what Houdini does with a "Source" enum attribute that is either "LOPs" or "SOPs"? Just so it's also familiar to those working with the OpenGL rop node?

openpype/hosts/houdini/plugins/create/create_review.py
```diff
@@ -56,30 +57,48 @@ def create(self, subset_name, instance_data, pre_create_data):
                 "aspect": pre_create_data.get("aspect"),
             })
 
+        lop_enabled = pre_create_data.get("lopEnabled")
```
BigRoy, line 60: Can't we just stick close to what Houdini uses internally and define a `opSource` or `source` type of data which defines either "SOPs" or "LOPs". That would also leave the path open to potential other sources that might ever appear in the future in Houdini? Plus it's also clearer that when LOPs is not chosen that it uses SOPs.

openpype/hosts/houdini/plugins/publish/validate_scene_review.py
```diff
@@ -37,36 +43,37 @@ def process(self, instance):
...
-        override = rop_node.parm("tres").eval()
+        override = node.parm("tres").eval()
         if not override:
             return
 
         invalid = []
-        res_width = rop_node.parm("res1").eval()
-        res_height = rop_node.parm("res2").eval()
+        res_width = node.parm("res1").eval()
+        res_height = node.parm("res2").eval()
```
BigRoy, line 76: Not sure why this was changed, but the old logic made more sense where it took the `rop_node` directly. Especially because the instance node already seems to be retrieved in `process`

moonyuet: maybe thid is the issue coming from merging develop to the branch. i agree that `rop_node` is more concise than `node`

mustafa-zarkash:
I tested it by modifying rop node manually and it successfully published with no errors 
image

I have an idea to update the creator, I'm going to try it. 
pseudocode something like this 
```
In LOPs Context

  case 1: No Selection 
    will only set `source` to `LOPs`
  
  case 2: one node selected 
    use its path as `loppath`
    if camera: 
      use its primpath as `cameraprim`
  
  case 3: two or more are selected 
    use camera's primpath as `cameraprim`
    pop camera from selection
    use first node as `loppath`
```

mustafa-zarkash:
I gave it a shot. 
I believe we will need at first to filter `selected_nodes` to sperate `hou.ObjNode`(s) from `hou.LopNode`(s) 
because user can accidently  have multiple selection from different contexts which can be confusing 

Imagine lop camera was the first node in selection, a user will complaint about selecting the obj camera but a lop camera was set! 

image



I can try again later. 

Here you are what I have achieved (with a lot of overthinking)
[create_review.zip](https://github.com/ynput/OpenPype/files/12114724/create_review.zip)

image

mkolar:

We're closing it down, but we'll he happy for a new PR to [ynput/ayon-houdini](https://github.com/ynput/ayon-houdini) repository once it's up.

PR #4891, opened 2023-04-21
TrayPublisher: Fill missing data for instances with review

iLLiCiTiT:
## Changelog Description
Fill required data to instance in traypublisher if instance has review family. The data are required by ExtractReview and it would be complicated to do proper fix at this moment! The collector does for review instances what did https://github.com/ynput/OpenPype/pull/4383

## Additional info
Ideally we should fill frame start/end from ranges of passed files, and use frame start from asset -> calculate frame end based on range from files. There is a lot of issues connected to this approach that may break other suff. We should also collect frame ranges for image frame sequences which are not right too (e.g. in burnins).

## Testing notes:
**Simple creators**
1. Open TrayPublisher
2. Select a creator (e.g. `render`)
3. Pass files to source file input and to review input
4. Publish
5. ExtractReview should not fail on missing data on instance

**Editorial**
1. Publish via TrayPublisher editorial which should also create reviews.

### DIFF
--- a/openpype/hosts/traypublisher/plugins/publish/collect_review_frames.py
+++ b/openpype/hosts/traypublisher/plugins/publish/collect_review_frames.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+import pyblish.api
+
+
+class CollectReviewInfo(pyblish.api.InstancePlugin):
+    """Collect data required for review instances.
+
+    ExtractReview plugin requires frame start/end, fps on instance data which
+    are missing on instances from TrayPublishes.
+
+    Warning:
+        This is temporary solution to "make it work". Contains removed changes
+            from https://github.com/ynput/OpenPype/pull/4383 reduced only for
+            review instances.
+    """
+
+    label = "Collect Review Info"
+    order = pyblish.api.CollectorOrder + 0.491
+    families = ["review"]
+    hosts = ["traypublisher"]
+
+    def process(self, instance):
+        asset_entity = instance.data.get("assetEntity")
+        if instance.data.get("frameStart") is not None or not asset_entity:
+            self.log.debug("Missing required data on instance")
+            return
+
+        asset_data = asset_entity["data"]
+        # Store collected data for logging
+        collected_data = {}
+        for key in (
+            "fps",
+            "frameStart",
+            "frameEnd",
+            "handleStart",
+            "handleEnd",
+        ):
+            if key in instance.data or key not in asset_data:
+                continue
+            value = asset_data[key]
+            collected_data[key] = value
+            instance.data[key] = value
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
It seems like an ok hotfix to at least get back the old behavior. 👍 🏅 

---

Consider below pseudocode/idea only.

However, admittedly it feels extremely weird that it's  not actually reading any information from the input files. Which is exactly why this logic was originally already weird too.

I had considered doing something like this:
```python
# This logic only works for instances that have representation files
representation_files = instance.data.get("representation_files", [])
for file_item in representation_files:

    if file_item.data.get("is_sequence", False):
        self.log.debug("Processing sequence..")

        frames = file_item.data.get("frames", [])
        if frames:
            frames = list(sorted(instance.data["frames"]))
            start = frames[0]
            end = frames[-1]

            instance.data["frameStart"] = start
            instance.data["frameEnd"] = end
            instance.data["frameStartHandle"] = start
            instance.data["frameEndHandle"] = end
            instance.data["handleStart"] = 0
            instance.data["handleEnd"] = 0

            self.log.debug(f"Collected frame range: {start}-{end}")

            # We only store the first sequence's frame range
            return

    else:
        # Not a sequence
        files = file_item.data.get("files", [])
        if files:
            if len(files) != 1:
                self.log.debug(
                    "More than one file found for an instance not marked "
                    "as sequence. Skipping detection of frame range..")
                return
            first_file = files[0]
            _, ext = os.path.splitext(first_file)
            if ext in transcoding.VIDEO_EXTENSIONS:
                probed_data = transcoding.get_ffprobe_data(
                    ext, logger=self.log
                )
                # TODO Get frame range from probed data

                # We only store the first file's frame range
                return
```

Which essentially could instead of returning on the first file instead set the instance.data frame range based on the earliest start of and latest end frame of all repre anyway, then the creator could have a handlesStart + handlesEnd field which would say how much of the frames should be considered start and end.

---

Another approach could've been to instead do it PER representation.
```python
for repre in instance.data.get("representations", []):
    files = repre.get("files")
    if not files:
        continue
    if isinstance(files, list):
        # sequence of files
        frame_range = get_sequence_frame_range(files)
    else:
        # single file
        frame_range = get_single_file_frame_range(files)

    if frame_range:
        repre["frameStart"] = frame_range["frameStart"]
        repre["frameEnd"] = frame_range["frameEnd"]
```

And for this we could do the same and set the instance's actual frame range to the earliest and last frame of the combined representation. Upside to this is that at least it has accurate duration / frames per representation as well.

Thoughts?

iLLiCiTiT:
I think the only way is doing it per representation. The biggest issue with that approach is that it would require to rewrite ExtractReview, probably ExtractBurnin and some more which don't expect that approach. That's a lot of work and a massive amount of work to review the PR.

PR #4893, opened 2023-04-22
Addons directory

tokejepsen:
## Changelog Description
This adds a directory for Addons, for easier distribution of studio specific code.

## Additional info
Discussion about how to distribute and maintain studio specific code is going on the forums; https://community.ynput.io/t/studio-specific-code/182/11

## Testing notes:
1. Add an addon to the `openpype/addons` directory.
2. Create a build, zip and installer.
3. Validate the addon files are included with the build, zip and installer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -112,3 +112,9 @@ tools/run_eventserver.*
 tools/dev_*
 
 .github_changelog_generator
+
+
+# Addons
+########
+/openpype/addons/*
+!/openpype/addons/README.md
--- a/openpype/addons/README.md
+++ b/openpype/addons/README.md
@@ -1,0 +1,3 @@
+This directory is for storing external addons that needs to be included in the pipeline when distributed.
+
+The directory is ignored by Git, but included in the zip and installation files.
--- a/openpype/modules/base.py
+++ b/openpype/modules/base.py
@@ -311,6 +311,7 @@ def _load_modules():
     # Look for OpenPype modules in paths defined with `get_module_dirs`
     #   - dynamically imported OpenPype modules and addons
     module_dirs = get_module_dirs()
+
     # Add current directory at first place
     #   - has small differences in import logic
     current_dir = os.path.abspath(os.path.dirname(__file__))
@@ -318,8 +319,11 @@ def _load_modules():
     module_dirs.insert(0, hosts_dir)
     module_dirs.insert(0, current_dir)
 
+    addons_dir = os.path.join(os.path.dirname(current_dir), "addons")
+    module_dirs.append(addons_dir)
+
     processed_paths = set()
-    for dirpath in module_dirs:
+    for dirpath in frozenset(module_dirs):
         # Skip already processed paths
         if dirpath in processed_paths:
             continue


BigRoy:
The code changes seem minimal which is nice - but I definitely feel this needs some extra explanation.

- If I for one wanted to distribute my own addons this way, how do I do that? Add your addon into the folder + enable the addon/module or whatever in your system settings?
- Why would I be adding my addon in `addons/` and not in `modules/` - isn't currently the behavior essentially the same?
- What do I set as "addon path" in settings for this path to get picked up? Do I need to set anything?

How are you git versioning your addons? You won't be just putting your git repositories into `/addons` since it would just include the full .git repository into the `.zip` right?

BigRoy:
I can understand your reasoning but I'm not seeing the direct benefit over instead just dropping it into the `modules/` folder if you start putting things inside the OpenPype folder anyway. It's good to see there's some thinking about solutions to this but I'm not sure this PR (in current state) is the best solution to it.

BigRoy:
In our case it's simple, we don't have a separate addon repository. We have our own fork and branch of OpenPype which contains our own "module"/"addon". Having it in `modules/` means we need to set up no addon paths, it's always available as a module like builtin modules but, like builtin modules, can be enabled/disabled.

Yes, it means our code is in the OpenPype folder (and also directly our fork/branch of the OpenPype repository).

tokejepsen:
> Yes, it means our code is in the OpenPype folder (and also directly our fork/branch of the OpenPype repository).

Does that mean your addon is publicly available?

BigRoy:
It is, yes. Here's [our colorbleed module](https://github.com/BigRoy/OpenPype/tree/colorbleed/openpype/modules/colorbleed). However, nothing is stopping you from working with your repository or branch in private I suppose.

tokejepsen:
> It is, yes. Here's [our colorbleed module](https://github.com/BigRoy/OpenPype/tree/colorbleed/openpype/modules/colorbleed). However, nothing is stopping you from working with your repository or branch in private I suppose.

I tend to work in private cause its normally a mess :)

I feel like there would be more management involve in having the addon in `modules`. For example personally I would need to have 2 OpenPype repositories; one private and one public (that is event possible).
Also although you could document how to gitignore your module within the repository, it would need a unique repository per studio.

I'll continue using this `addons` directory for now. Most likely there will be issues that I've not foreseen.

iLLiCiTiT:
> Assuming you can do a relative path to the addons directory, so {OPENPYPE_ROOT}/openpype/addons. But I haven't experimented that far yet.

We should find out if this is possible, or maybe change the behavior in `openpype/modules/base.py` to look into addons too.

> I can understand your reasoning but I'm not seeing the direct benefit over instead just dropping it into the modules/ folder if you start putting things inside the OpenPype folder anyway

One of reasons we changed in AYON that addons are not part of the base is different versioning. So I see the benefit in this approach in having different versions of addon in different versions of openpype, so you can easily change behavior without changes in main repository. And our current approach, using path to addons in System settings, does not allow to use different versions and also the artist must have access to addons directory.

iLLiCiTiT (approved):
Added addons dir automatically to modules discovery paths. LGTM

PR #4894, opened 2023-04-24
Publisher: Small style changes

iLLiCiTiT:
## Changelog Description
Small changes in styles and form of publisher UI.

## Additional info
- Attribute definitions have smaller spacing and padding, their labels are aligned to right.
- Cards in card view are a little bit smaller (6px).
- Icons on cards should have unified size -> pixmaps are not sometimes resized correctly but card size is unified.
- Change of active states on card view works the same as in list view.
- Renamed `Options` to `Context`.
- Border radius on inputs is a little bit smaller.
- Scroll area have explicitly set borders to none -> in Houdini were black.
- Implemente combobox which ignores mouse wheel event until has active focus (avoid accidental change of values)

## Testing notes:
1. Open Publisher in different DCCs
2. Validate it is not broken

### DIFF
--- a/openpype/pipeline/publish/publish_plugins.py
+++ b/openpype/pipeline/publish/publish_plugins.py
@@ -45,7 +45,7 @@ class PublishValidationError(Exception):
 
     def __init__(self, message, title=None, description=None, detail=None):
         self.message = message
-        self.title = title or "< Missing title >"
+        self.title = title
         self.description = description or message
         self.detail = detail
         super(PublishValidationError, self).__init__(message)
--- a/openpype/style/data.json
+++ b/openpype/style/data.json
@@ -48,7 +48,7 @@
         "bg-view-selection-hover": "rgba(92, 173, 214, .8)",
 
         "border": "#373D48",
-        "border-hover": "rgba(168, 175, 189, .3)",
+        "border-hover": "rgb(92, 99, 111)",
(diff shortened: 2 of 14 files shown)


BigRoy:
Tested in Houdini and Fusion and seems to work as intended.

Scrolling over comboboxes also doesn't change values anymore without being active.

**Houdini**

afbeelding
afbeelding

**Fusion**

afbeelding
afbeelding

Admittedly I feel the icons might be _slightly too big_ on the instances in Card view making it feel slightly cluttered / in your face.

---

Some other minor notes:

**Improve context entry distinction?**

Additionally, unrelated to this PR I think but I think in the non-card view (list view?) the **Context** entry might need to have some special design too. Maybe it could be slightly bigger in padding or have an icon, or get similar indenting as instances? Something about how it looks in the list doesn't make it look like a selectable entry:

afbeelding

In card view it might also be nice to have a default context icon? Which could technically for some applications be maybe the host application's icon?

afbeelding

**Disallow deleting auto creator entries**

As far as I know auto creator instances can't be deleted by their design. However, in the publisher it'll still show the garbage bin enabled when selecting e.g. the workfile instance:

afbeelding

Whereas if it can't be deleted I'd expect it to behave like the Context and have a disabled delete button:

afbeelding

BigRoy (approved):

Issue #4593, opened 2023-03-08
Scene Inventory: Cherry Picking hierarchy error for new style hosts


BigRoy:
**Running version**
3.15.2-nightly.4

**Describe the bug**

[This particular line of code](https://github.com/ynput/OpenPype/blob/develop/openpype/tools/sceneinventory/model.py#L211) for the Scene Inventory / Manager will fail for most hosts and especially new style hosts since `host.pipeline` isn't a thing on it. 🙂

You'll hit this error if you "Cherry Pick" in the loader. (I've personally never used that, but maybe others do?) 

The only host having a [`update_hierarchy` function implemented is Blender](https://github.com/ynput/OpenPype/blob/79e6de15b56ac9c3b2c60a1278b09958df1d67e3/openpype/hosts/blender/api/pipeline.py#L379-L406
). It's the only update_hierarchy function in the codebase 🙂

- Is cherry-picking still ever a thing?
- Are hierarchies like that in the scene inventory a thing? And where is it used in OP?

Doesn't seem like it's getting much use? It originates from Avalon, see the [Avalon Hierarchy View PR](https://github.com/getavalon/core/pull/353).

**To Reproduce**

1. Go to a new style host, e.g. houdini, nuke, after effects
2. Go to Scene Inventory
3. Right click on the child item, a specific container, and then click Cherry Pick.
4. Refreshing will then report an error

Example error:
```python
Traceback (most recent call last):
  File "S:\openpype\OpenPype\openpype\tools\sceneinventory\window.py", line 156, in refresh
    self._model.refresh(**kwargs)
  File "S:\openpype\OpenPype\openpype\tools\sceneinventory\model.py", line 211, in refresh
    if not hasattr(host.pipeline, "update_hierarchy"):
AttributeError: 'FusionHost' object has no attribute 'pipeline'
```

**Expected behavior**

No errors.

**Additional context**

Hierarchy views and cherry picking originate from Avalon, see [this Avalon Hierarchy View PR](https://github.com/getavalon/core/pull/353)

LiborBatek:
@iLLiCiTiT Can be closed then?

PR #4895, opened 2023-04-24
Scene inventory: Model refresh fix with cherry picking

iLLiCiTiT:
## Changelog Description
Fix cherry pick issue in scene inventory.

## Additional info
As described in [source issue](https://github.com/ynput/OpenPype/issues/4593), the functionality is implemented only in blender. I didn't find out what it actually does?

### Traceback
```
  File "S:\openpype\OpenPype\openpype\tools\sceneinventory\window.py", line 156, in refresh
    self._model.refresh(**kwargs)
  File "S:\openpype\OpenPype\openpype\tools\sceneinventory\model.py", line 211, in refresh
    if not hasattr(host.pipeline, "update_hierarchy"):
AttributeError: 'FusionHost' object has no attribute 'pipeline'
```

## Testing notes:
1. Open scene inventory
2. Cherry pick an item there

### DIFF
--- a/openpype/tools/sceneinventory/window.py
+++ b/openpype/tools/sceneinventory/window.py
@@ -107,8 +107,8 @@ class SceneInventoryWindow(QtWidgets.QDialog):
         view.hierarchy_view_changed.connect(
             self._on_hierarchy_view_change
         )
-        view.data_changed.connect(self.refresh)
-        refresh_button.clicked.connect(self.refresh)
+        view.data_changed.connect(self._refresh_callback)
+        refresh_button.clicked.connect(self._refresh_callback)
         update_all_button.clicked.connect(self._on_update_all)
 
         self._update_all_button = update_all_button
@@ -138,6 +138,11 @@ class SceneInventoryWindow(QtWidgets.QDialog):
         whilst trying to name an instance.
 
         """
+
+    def _refresh_callback(self):
+        """Signal callback to trigger 'refresh' without any arguments."""
+
+        self.refresh()
 
     def refresh(self, items=None):
         with preserve_expanded_rows(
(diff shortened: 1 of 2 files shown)


BigRoy (changes requested):
**Doesn't work yet** - this breaks refreshing (top right refresh button)in the scene inventory.

Not sure why, but in Fusion and Houdini when clicking "refresh" after the scene inventory was open it doesn't show anything in the scene inventory. Reopening the scene inventory however does show it again, then refresh again clears out the inventory.

openpype/tools/sceneinventory/window.py
```diff
@@ -139,6 +139,11 @@ def keyPressEvent(self, event):
 
         """
 
+    def _refresh_callback(self):
```
BigRoy, line 142: Can we rename this `_on_refresh` just so it aligns with the other callback `_on_update_all` and I believe also aligns more with how callbacks are more often set up in OpenPype in tools?

iLLiCiTiT: I wanted, but I would expect that 'on refresh' would be triggered during/before/after refresh, not to trigger refresh.

iLLiCiTiT: Maybe `_on_refresh_request`?

BigRoy: Yup, sounds like a good alternative to me.

BigRoy (approved):
Not getting any errors with cherry picking + refresh works. Would still recommend renaming the `refresh_callback` method to something that starts with `on_` as commented before. But aside of that this seems to at least resolve the source issue.

**However**, it'd still be worth having an issue to track whether "Cherry Picking" is worth anything in OpenPype - e.g. is it used?

PR #4896, opened 2023-04-24
Fusion: Add USD loader

BigRoy:
## Changelog Description

Add an OpenPype managed USD loader (`uLoader`) for Fusion. 

## Additional info

Fusion 18.5 beta adds USD tools, including a USD loader tool (`uLoader`).

_Note Fusion 18.5+ is currently still in public beta, so I'll keep this around as a draft for the time being._

## Testing notes:

1. Start Fusion 18.5 beta or a newer release once it's out
2. Load published USD files.
3. Also test whether updating and removing works with manage.. (scene inventory)

### DIFF
--- a/openpype/hosts/fusion/plugins/load/load_usd.py
+++ b/openpype/hosts/fusion/plugins/load/load_usd.py
@@ -1,0 +1,73 @@
+from openpype.pipeline import (
+    load,
+    get_representation_path,
+)
+from openpype.hosts.fusion.api import (
+    imprint_container,
+    get_current_comp,
+    comp_lock_and_undo_chunk
+)
+
+
+class FusionLoadAlembicMesh(load.LoaderPlugin):
+    """Load USD into Fusion
+
+    Support for USD was added since Fusion 18.5
+    """
+
+    families = ["*"]
+    representations = ["*"]
+    extensions = {"usd", "usda", "usdz"}
+
+    label = "Load USD"
+    order = -10
+    icon = "code-fork"
+    color = "orange"
+
+    tool_type = "uLoader"
+
+    def load(self, context, name, namespace, data):
+        # Fallback to asset name when namespace is None
+        if namespace is None:
+            namespace = context['asset']['name']
+
+        # Create the Loader with the filename path set
+        comp = get_current_comp()
+        with comp_lock_and_undo_chunk(comp, "Create tool"):
+
+            path = self.fname
+
+            args = (-32768, -32768)
+            tool = comp.AddTool(self.tool_type, *args)
+            tool["Filename"] = path
+
+            imprint_container(tool,
+                              name=name,
+                              namespace=namespace,
+                              context=context,
+                              loader=self.__class__.__name__)
+
+    def switch(self, container, representation):
+        self.update(container, representation)
+
+    def update(self, container, representation):
+
+        tool = container["_tool"]
+        assert tool.ID == self.tool_type, f"Must be {self.tool_type}"
+        comp = tool.Comp()
+
+        path = get_representation_path(representation)
+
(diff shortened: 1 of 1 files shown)


openpype/hosts/fusion/plugins/load/load_usd.py
```diff
@@ -0,0 +1,73 @@
...
+    imprint_container,
+    get_current_comp,
+    comp_lock_and_undo_chunk
+)
+
+
+class FusionLoadAlembicMesh(load.LoaderPlugin):
+    """Load USD into Fusion
+
+    Support for USD was added since Fusion 18.5
```
Minkiu, line 15: Is it possible to perform a check to see if we are on the correct versions and if not `pass` ?

openpype/hosts/fusion/plugins/load/load_usd.py
```diff
@@ -0,0 +1,73 @@
...
+    get_representation_path,
+)
+from openpype.hosts.fusion.api import (
+    imprint_container,
+    get_current_comp,
+    comp_lock_and_undo_chunk
+)
+
+
+class FusionLoadAlembicMesh(load.LoaderPlugin):
```
BigRoy, line 12:
```suggestion
class FusionLoadUSD(load.LoaderPlugin):
```

openpype/hosts/fusion/plugins/load/load_usd.py
BigRoy: I suppose we could do `fusion.GetVersion()`:
```python

Py3> print(fusion.GetVersion())
# {1: 18, 2: 0, 3: 2, 4: 6, 'App': 'Fusion'}
```

As such we could maybe do:
```python
version = fusion.GetVersion()
major = version[1]
minor = version[2]
is_usd_supported = (major, minor) < (18, 5)

class FusionLoadUSD(load.LoaderPlugin):
    # ...
    # ...
    # do not allow the loader
    enabled = is_usd_supported
```

Minkiu (approved):
Haven't tested it, but code looks good!

BigRoy:
For some reason I can't check out this branch anymore due to `Unreal` related files making it hard to test for me. I've had that before with 'older PRs' :D Anyone that has an IDE / GIT setup that does it better than mine and can test Fusion? @EmberLightVFX @jakubjezek001 maybe?

BigRoy:
Did a test run myself - version check seemed to work as intended + loading usd worked fine. Ready to merge.

antirotor (approved):
Code look great, can't test it though.

Minkiu (approved):
The code hasn't changed since my last review?
So with Roy's test, I think it's good to go!

openpype/hosts/fusion/plugins/load/load_usd.py
```diff
@@ -0,0 +1,87 @@
...
+                              loader=self.__class__.__name__)
+
+    def switch(self, container, representation):
+        self.update(container, representation)
+
+    def update(self, container, representation):
+
+        tool = container["_tool"]
+        assert tool.ID == self.tool_type, f"Must be {self.tool_type}"
+        comp = tool.Comp()
```
Minkiu, line 71: I guess this bit could be in a helper method:
```py
def _get_comp(self, container):
        tool = container["_tool"]
        assert tool.ID == self.tool_type, f"Must be {self.tool_type}"
        comp = tool.Comp()
        return comp
```

BigRoy: We still need `tool` variable outside of that scope too - meaning this would basically be `get_comp(tool)` and thus is just the same as `tool.Comp()` no?

PR #4899, opened 2023-04-24
Fusion: Simplify creator icons code

BigRoy:
## Changelog Description

Simplify code for setting the icons for the Fusion creators

## Additional info

Icons look like before:
afbeelding

There was no need to reimplement `get_icon` methods.

## Testing notes:

1. Open Publisher in Fusion
2. Icons should still look beautiful for the instances

### DIFF
--- a/openpype/hosts/fusion/plugins/create/create_saver.py
+++ b/openpype/hosts/fusion/plugins/create/create_saver.py
@@ -26,7 +26,7 @@ class CreateSaver(Creator):
     family = "render"
     default_variants = ["Main", "Mask"]
     description = "Fusion Saver to generate image sequence"
-    icon = "eye"
+    icon = "fa5.eye"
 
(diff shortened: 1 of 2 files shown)


EmberLightVFX:
Correct me if I'm wrong but isn't the qtawesome icons the latest/updated way to get icons in OP?

BigRoy:
> Correct me if I'm wrong but isn't the qtawesome icons the latest/updated way to get icons in OP?

I suspect @iLLiCiTiT could answer that. But if we are to replace the current usage of icon like this (which does use `qtawesome`, etc.) then we'd need to reimplement `get_icon` on all the creators that are currently in use. So I suspect this PR is a correct cleanup.

iLLiCiTiT:
I have one comment which is that there should be ideally full name of icon with prefix like `fa5.`.

iLLiCiTiT (approved):

PR #4901, opened 2023-04-25
Tray Publisher: User can set colorspace per instance explicitly

BigRoy:
## Changelog Description

With this feature a user can set/override the colorspace for the representations of an instance explicitly instead of relying on the File Rules from project settings or alike. This way you can ingest any file and explicitly say "this file is colorspace X".

## Additional info

I had a few cases where I just had some files laying around or received from a client that did not adhere to any of our file rules and I didn't want to fool around with the project settings to catch that particular colorspace. I just wanted to say "These files are colorspace X"

Warning: Due to how these settings are set per instance this will set **all of the representations** of an instance to be that colorspace. Thus if an instance had both `acescg` EXRs and a preview `srgb` h264 mov you'd be overriding the colorspace for both at once and thus can not differentiate between either.

#### Screenshots

afbeelding
afbeelding

### Alternatives

Instead of this implementation a workaround could've been to still set up explicit file rules in Project Settings but then rename the input files to adhere to those file rules so the colorspace is set according to that instead.

## Testing notes:

1. Start tray publisher.
2. On the publish tab each instance should have an "Override Colorspace" attribute, type a valid colorspace name from the OCIO config set in the Project Settings.
3. On publish the colorspace name is validated.

### DIFF
--- a/openpype/hosts/traypublisher/plugins/publish/collect_explicit_colorspace.py
+++ b/openpype/hosts/traypublisher/plugins/publish/collect_explicit_colorspace.py
@@ -1,0 +1,36 @@
+import pyblish.api
+
+from openpype.pipeline import publish
+from openpype.lib import TextDef
+
+
+class CollectColorspace(pyblish.api.InstancePlugin,
+                        publish.OpenPypePyblishPluginMixin,
+                        publish.ColormanagedPyblishPluginMixin):
+    """Collect explicit user defined representation colorspaces"""
+
+    label = "Choose representation colorspace"
+    order = pyblish.api.CollectorOrder + 0.49
+    hosts = ["traypublisher"]
+
+    def process(self, instance):
+        values = self.get_attr_values_from_data(instance.data)
+        colorspace = values.get("colorspace", None)
+        if not colorspace:
+            return
+
+        context = instance.context
+        for repre in instance.data.get("representations", {}):
+            self.set_representation_colorspace(
+                representation=repre,
+                context=context,
+                colorspace=colorspace
+            )
+
+    @classmethod
+    def get_attribute_defs(cls):
+        return [
+            TextDef("colorspace",
+                    label="Override Colorspace",
+                    placeholder="")
+        ]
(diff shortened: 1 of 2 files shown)


moonyuet:
I tested with exr and the validator works fine.
image
image
image

jakubjezek001:
Since `openpype\pipeline\colorspace.py` is having available function for getting all available colorspaces in an input config **get_ocio_config_colorspaces**, I wonder if it would be possible to offer then as enumerator in the field, rather then letting users to type with errors. @iLLiCiTiT

iLLiCiTiT:
> Since `openpype\pipeline\colorspace.py` is having available function for getting all available colorspaces in an input config **get_ocio_config_colorspaces**, I wonder if it would be possible to offer them as enumerated in the field, rather than letting users to type with errors. @iLLiCiTiT

It can be added as enum, but I don't know how? Function `get_ocio_config_colorspaces` expects config path, where is that taken from?

jakubjezek001:
> It can be added as enum, but I don't know how tu use function `get_ocio_config_colorspaces`. It expects config path, where is that taken from?

Better to wait for https://github.com/ynput/OpenPype/pull/4700 to be able to get it from correct context. At the moment you can get it from `openpype\pipeline\colorspace.py`:`get_imageio_config`.

Minkiu (approved):
Haven't tested, but code looks good

jakubjezek001 (approved):
LGTM. Thanks @BigRoy - this is an awesome feature!

BigRoy:
@LiborBatek could you give this a test run?

LiborBatek (approved):
I did setup 2 custom OCIO config files and varied the OP Color Management settings for all the cases

1) Global ocio config
2) Host specific config (Tray Publisher)
3) Without any ocio config / turned off CMS
4) Turned on CMS with Nuke Defaults

By observing those values during Publish for overrides of color space it all worked with corresponding OCIO config as defined in particular testing case.

So all good!

mkolar:
Thank you @BigRoy merging

PR #4902, opened 2023-04-25
General: Removed unused cli commands

iLLiCiTiT:
## Changelog Description
Removed `texturecopy` and `launch` cli commands from cli commands.


BigRoy:
I was wondering, since we are cleaning this up, is the `context_dialog` used together with the `contextselection` cli command?

iLLiCiTiT:
> I think it might be in use for [some royal render code](https://github.com/ynput/OpenPype/blob/8d45b649ec97d4dd041ab6617ba67658141d64a9/openpype/modules/royalrender/rr_root/plugins/control_job/perjob/m50__openpype_publish_render.py#L87-L97)?

Yes, it was implemented for royal render, and we have a client using it.

fabiaserra:
Repeating my comment from Discord.
Why removing these if they would both be very useful to have? I don't mind so much the `texturecopy` (although reading about it seems like it would be useful to have) but on my fork of OP we actually implemented `run_application` like this and `launch` CLI has been working fine. We use it pretty often for quickly launching DCCs without going through the Launcher:
```python
    def run_application(self, app, project, asset, task, tools, arguments):
        from openpype.lib.applications import (
            get_app_environments_for_context,
            ApplicationLaunchContext,
            ApplicationManager
        )
        from openpype.pipeline import install_openpype_plugins
        from openpype.lib import Logger

        log = Logger.get_logger("CLI-Launch")

        install_openpype_plugins()
        app_manager = ApplicationManager()
        if not app_manager.applications.get(app):
            log.warning("App not found: {}".format(app))
            log.info("All valid apps {}".format(list(app_manager.applications.keys())))
            return

        if all((project, asset, task, app)):
            env = get_app_environments_for_context(
                project, asset, task, app
            )
        else:
            env = os.environ.copy()
        
        os.environ.update(env)

        app = app_manager.applications[app]
        executable = app.find_executable()

        data = {
            "project_name": project,
            "asset_name": asset,
            "task_name": task,
            "app_args": arguments,
        }
        context = ApplicationLaunchContext(
            app, executable, **data,
        )

        context.launch()

        log.info("Application launched ...")
```

iLLiCiTiT:
The main reason at this moment is that it would not work the same way as from launcher, or ftrack. There might be missing some environment variables, that are set only in tray app, like ftrack related etc..

The command can be added back in the future, as command of applications addon. But at this moment it is not implemented and there is not 100% way how to make it work the same way as from launcher, until AYON realease we probably won't look into it much -> so there is command argument which is not implemented in the codebase. I don't know about anyone who would use it as noone ever complaint abou missing implementation, except you now. I would like to know the usecase, when you want to launch an app using cli?

BTW: Your implementation is doing a lot of thins that are not needed and could be simplified. You don't have to install plugins (that's what should happen in host, not before). Also you're duplicating a lot of code from Applications manager itself:
```python
app_manager = ApplicationManager()
if not app_manager.applications.get(app):
    log.warning("App not found: {}".format(app))
    log.info("All valid apps {}".format(list(app_manager.applications.keys())))
    return

kwargs = {}
for key, value in (
    ("project_name", project_name),
    ("asset_name", asset_name),
    ("task_name", task_name),
):
    if not value:
        break
    kwargs[key] = value
app_manager.launch(app, **kwargs)
```

fabiaserra:
> The main reason at this moment is that it would not work the same way as from launcher, or ftrack. There might be missing some environment variables, that are set only in tray app, like ftrack related etc..

Is it just environment variables difference? Where are those being set? Seems like it shouldn't be hard to align?
 
> The command can be added back in the future, as command of applications addon. But at this moment it is not implemented and there is not 100% way how to make it work the same way as from launcher, until AYON realease we probably won't look into it much -> so there is command argument which is not implemented in the codebase. I don't know about anyone who would use it as noone ever complaint abou missing implementation, except you now. I would like to know the usecase, when you want to launch an app using cli?

There's a few use cases where you'd want to launch an app from CLI without going through all the Launcher steps. We have a few artists that prefer launching apps from command line and sometimes they just want quick access to an app for sandboxing or R&D. With the launch CLI, I can create an alias of all the apps that sets some context defaults so we don't need to have a different environment logic separate from OP (we currently have a duplicate environment to launch from CLI without OP context but it's on my plans to replace that so we just need to maintain one). Similarly, from a dev POV, for quick development iterations for things that I know I need to relaunch the app or restart OP, I'd rather just do it directly with a single command than going through the launcher.

Finally, with this you could easily replace the launcher completely by having a `context` CLI tool that sets the environment on the terminal environment (i.e., `job <show> <seq>/<shot> -t task`) so any app you execute from that point runs in the environment of that context (this is how MPC Film works).

And of course, also the request from Discord that brought me to this topic: "does anyone know how to Open a Maya Scene in Batch Mode, within a OpenPype context?"

> BTW: Your implementation is doing a lot of thins that are not needed and could be simplified. You don't have to install plugins (that's what should happen in host, not before). Also you're duplicating a lot of code from Applications manager itself:

Yeah fair enough, I didn't implement this, it was my coworker, thus why I haven't tested it enough myself to be confident on creating a PR against your central repo... but thanks for the proposed changes, I can give it a try.

fabiaserra:
Is `applications.get_app_environments_for_context()` actually being called on the `ApplicationManager`?

iLLiCiTiT:
> Is it just environment variables difference? Where are those being set? Seems like it shouldn't be hard to align?

This would need confirmation, I am 99% sure that ftrack credentials are filled only if app is launched from tray, I know we can align it but that will take some time to do research and some implementation, which is not just glued in to "make it work because we want", and we don't have much spare time now. We have to think more about AYON preparation (separate the logic into addon) where the cli would look like `openpype modules applications launch --app maya/2020 ...`.

> Is applications.get_app_environments_for_context() actually being called on the ApplicationManager?

It is not called the same way, it's called by global prelaunch hook in `pre_global_host_data`.

antirotor (approved):
Battle plan is to have launch command in AYON application addon. If this needs to be back in OpenPype, lets do it in another PR with proper implementation.

PR #4825, opened 2023-04-11
Houdini farm publishing and rendering

moonyuet:
## Changelog Description
Deadline Farm publishing and Rendering for Houdini
- [x] Mantra 
- [x] Karma(including usd renders)
- [x] Arnold
- [x] Elaborate Redshift ROP for deadline submission
- [x]  fix the existing bug in Redshift ROP
- [x] Vray
- [x] add docs

## Additional info
If you turns "select" on and selected your camera node in the publisher, you need to set up the render camera manually through the related renderer node

Tested with Render
- [x] Mantra 
- [x] Karma(including usd renders)
- [x] Arnold (Tested in #4193 )
- [x] Redshift 
- [x] Vray

## Testing notes:

**General Renderer**
1. Launch Houdini via Openpype
2. (if you want to select camera, turns 'select' on and choose your camera node)
3. Create Render Instance 
4. Publish it

**Vray Render Element**
1.  Launch Houdini via Openpype
2. (if you want to select camera, turns 'select' on and choose your camera node)
3. Switch on 'Render Element''s button
4. Create Render Instance 
6.  Go to out/ and find the render element vop
7. Dive in to set your render element
8. Get back to the Publisher and publish it

PR #4903, opened 2023-04-25
Houdini: Farm caching submission to Deadline 

moonyuet:
## Changelog Description
Implements functionality to offload instances of the specific families to be processed on Deadline instead of locally. This increases productivity as artist can use local machine could be used for other tasks.
Implemented for families:
- [x] ass 
- [x] redshift proxy
- [x] ifd
- [x] abc
- [x] bgeo
- [x] vdb
## Additional info
Abc export via farm caching submission doesn't include any animation
Current version of deadline does not support vdb farm caching. (Tried with the manual deadline submission)

## Testing notes:
1. Launch Houdini via launcher
2. Enable "Submitting to Farm" and create instance with the families mentioned above
3. The selection of the object would be submitting to farm
4. After farm publishing, the loader would have subset parented to the families.

image


openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
```diff
@@ -0,0 +1,24 @@
...
+                "redshiftproxy"]
+    hosts = ["houdini"]
+    targets = ["local", "remote"]
+    label = "Collect Data for Cache"
+
+    def process(self, instance):
+        if instance.data.get("farm"):
+            instance.data["families"].append("publish.hou")
+            data = dict()
+            data["chunkSize"] = 100
```
BigRoy, line 21: This seems arbitrary? Surely there are better ways to make sure that the exports that run in one go actually produce a single task in deadline?

I recall seeing a "Submit as single task" or something setting in Deadline but couldn't quickly find it in the current arguments. But that's what I'd at least expect to create a workflow where we force it a single task in a clear distinct way (separate from ChunkSize I'd say unless we consider setting chunkSize to e.g. 999999 safe enough).

openpype/modules/deadline/plugins/publish/submit_houdini_cache_deadline.py
```diff
@@ -0,0 +1,133 @@
...
+import hou
+
+import pyblish.api
+
+from openpype.pipeline import legacy_io
+from openpype.tests.lib import is_in_tests
+from openpype.lib import is_running_from_build
+
+
+class HoudiniSubmitPublishDeadline(pyblish.api.InstancePlugin):
```
BigRoy, line 15: Shouldn't this be using the abstract deadline submitter base class?

openpype/hosts/houdini/plugins/publish/extract_vdb_cache.py
```diff
@@ -14,8 +14,12 @@ class ExtractVDBCache(publish.Extractor):
     label = "Extract VDB Cache"
     families = ["vdbcache"]
     hosts = ["houdini"]
+    targets = ["local", "remote"]
 
     def process(self, instance):
+        if instance.data.get("farm"):
+            self.log.debug("Should be processed on farm, skipping.")
+            return
```
BigRoy, line 22: If we have this `if instance.data.get('farm")` check then why should we still need to set the targets like this?

openpype/hosts/houdini/plugins/publish/extract_vdb_cache.py
```diff
@@ -14,8 +14,12 @@ class ExtractVDBCache(publish.Extractor):
     label = "Extract VDB Cache"
     families = ["vdbcache"]
     hosts = ["houdini"]
+    targets = ["local", "remote"]
 
     def process(self, instance):
+        if instance.data.get("farm"):
```
BigRoy, line 20: When this triggers on the farm what is the code that is setting `instance.data["farm"] = False`

BigRoy (commented):
Some questions

openpype/modules/deadline/plugins/publish/submit_houdini_cache_deadline.py
moonyuet: yes, it is. And you can see the creation from the base class in #4825 but there is a overlapping betweent the ticket. This is one of the reason why it is still in draft.

BigRoy: Actually, no. This has nothing to do with that particular PR since it's a different publish plug-in all together.
This PR can implement this logic on abstract deadline submitter just fine without getting in the way of the other PR.

Also, the filename being `submit_houdini_cache_deadline` and the class being `HoudiniSubmitPublishDeadline` feels a bit confusing. We should change one of them to match the other.

mkolar:
@mustafa-zarkash can you give it a test again please?

mkolar (approved):

mustafa-zarkash:
I was testing this PR as a user 
the good news:
- it submits to deadline and caches successfully. 

the bad news: 
- sometimes I can't find the cached files 
- some families need to be fixed for example there's a typo in redshift proxy ( extra  ` ) , _but that's a different PR_
https://github.com/ynput/OpenPype/blob/e917d2b91fa4b45612cca33ccc34a0d1bb237839/openpype/hosts/houdini/plugins/create/create_redshift_proxy.py#L36

image

moonyuet:
I am not sure if I understand the question correctly.
The output of the farm cache should be stored according to what you have assigned in the render output when you create the instances. (see the output)
image

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
moonyuet: I will make this into chuckSize: 9999999 just to be safe

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
```diff
@@ -0,0 +1,82 @@
...
+import hou
+from openpype.hosts.houdini.api import lib, colorspace
+from openpype.hosts.houdini.api.lib import get_color_management_preferences
+
+
+class CollectDataforCache(pyblish.api.InstancePlugin):
+    """Collect data for caching to Deadline."""
+
+    order = pyblish.api.CollectorOrder + 0.1
+    families = ["ass", "pointcache",
```
mustafa-zarkash, line 12: Regarding alembic and bgeo pointcache 
after the last update #4588 and #5324

I think we can use `pointcache` or we can use `abc` and `bgeo` separately.
however, if we used them separately a dependency may occur as it will be required to have this specific collector `collect_pointcache_type.py` to run before this collector. 
```suggestion
    families = ["ass", "abc",
```

mustafa-zarkash (commented):
about point cache alembic and bgeo

mustafa-zarkash:
So far from a user perspective, it submits cache job to deadline, but it doesn't submit the publish job yet.  
Good job! 

image

moonyuet:
Just find out that i haven't committed the script in right way, @mustafa-zarkash you can submit the job to publish now.

mustafa-zarkash:
yes, it submits publish job, Good Job! 

I can't tell whether the publish job will work fine or not because there may be an installation problem. 
image

mustafa-zarkash (changes requested):
I found these

openpype/hosts/houdini/plugins/publish/extract_vdb_cache.py
mustafa-zarkash: I think it was already implemented in #3225 

https://github.com/ynput/OpenPype/blob/40037b050ce287f503bf4214f63e9e9f1b8028f5/openpype/modules/deadline/plugins/publish/collect_publishable_instances.py#L36

openpype/hosts/houdini/plugins/publish/extract_vdb_cache.py
mustafa-zarkash: I think it was already implemented in #3225 

https://github.com/ynput/OpenPype/blob/40037b050ce287f503bf4214f63e9e9f1b8028f5/openpype/modules/deadline/plugins/publish/collect_publishable_instances.py#L36

mustafa-zarkash:
I believe there's something weird is happening. 
why `bgeo sop path` validator works with  alembic instances  after checkening "farm" parameter ?! 

image

image

openpype/hosts/houdini/plugins/publish/extract_vdb_cache.py
mustafa-zarkash: I think it was inspired from Maya extract point cache. 
does setting targets to remote cause some plugins to be skipped ? 

https://github.com/ynput/OpenPype/blob/40037b050ce287f503bf4214f63e9e9f1b8028f5/openpype/hosts/maya/plugins/publish/extract_pointcache.py#L30-L32

moonyuet:
> 
Dont think it is related to this PR at all.
it is supposed that it should check the bgeo sop path before farm submission, which is the same with the scanerio when you pulblish bgeo through extractor. So if this error happened, it would be also seen when you are not publishing to farm.

mustafa-zarkash:
It looks good to me, I was able to submit all these types to farm successfully. 

image

image

mustafa-zarkash (approved):
It looks good to me, I was able to submit all these types to farm successfully. 

image

image

krishna8008:
Hey!

I've tested this today, there are 2 faults I recognised, correct me If I'm wrong:
1. Job is succesfully submitted to farm thats good new, but Its not taking the proper houdini version, eg: I'm working with 19.5, but deadline job is getting failed and log says its using 19.0
2.  We need to have a "Frames per task" option coz all the frames are going as a single task for BGEO sequence, which is not totally a time consuming thing, Perhaps we might need single job submittion for tasks like abc, but not super helpful for sequences like BGEO

Thank You!

moonyuet:
> Hey!
> 
> I've tested this today, there are 2 faults I recognised, correct me If I'm wrong:
> 
> 1. Job is succesfully submitted to farm thats good new, but Its not taking the proper houdini version, eg: I'm working with 19.5, but deadline job is getting failed and log says its using 19.0
> 2. We need to have a "Frames per task" option coz all the frames are going as a single task for BGEO sequence, Perhaps we might need single job submittion for tasks like abc, but not super helpful for sequences like BGEO
> 
> Thank You!

Thanks for the testing!
1. For the Houdini version, @mustafa-zarkash has more maps on it as he is working on the patch version PRs to deal with the issue. You can take a look at the PR here: https://github.com/ynput/OpenPype/pull/5384
2. I will be working on that.

mustafa-zarkash:
> 1. Job is succesfully submitted to farm thats good new, but Its not taking the proper houdini version, eg: I'm working with 19.5, but deadline job is getting failed and log says its using 19.0

This is a little bit strange, because all houdini jobs (cache, render) get houdini version value in the same way! 

```python
hou_major_minor = hou.applicationVersionString().rsplit(".", 1)[0]
```
```python 
version = hou.applicationVersionString()
version = ".".join(version.split(".")[:2])
```
--- 
what version do you find in job parameters ? 
like this
image

krishna8008:
> > 1. Job is succesfully submitted to farm thats good new, but Its not taking the proper houdini version, eg: I'm working with 19.5, but deadline job is getting failed and log says its using 19.0
> 
> This is a little bit strange, because all houdini jobs (cache, render) get houdini version value in the same way!
> 
> ```python
> hou_major_minor = hou.applicationVersionString().rsplit(".", 1)[0]
> ```
> 
> ```python
> version = hou.applicationVersionString()
> version = ".".join(version.split(".")[:2])
> ```
> 
> what version do you find in job parameters ? like this image

Hey! Here is the snapshot
image

moonyuet:
> > 1. Job is succesfully submitted to farm thats good new, but Its not taking the proper houdini version, eg: I'm working with 19.5, but deadline job is getting failed and log says its using 19.0
> 
> This is a little bit strange, because all houdini jobs (cache, render) get houdini version value in the same way!
> 
> ```python
> hou_major_minor = hou.applicationVersionString().rsplit(".", 1)[0]
> ```
> 
> ```python
> version = hou.applicationVersionString()
> version = ".".join(version.split(".")[:2])
> ```
> 
> what version do you find in job parameters ? like this image

it's mainly because I hard-coded for houdini version which means it will always submit to Houdini 19.0. I will change it back in the upcoming PR but the user of 19.5 needs to make sure to add the param inside the Houdini executable for the system to find the right version.

moonyuet:
In the updated PR, I add a simple setting to set up customized chunk size(i.e. frame per task) for certain families. By default, the chunk size is 999999. (Of course we can do more concise settings which allows chunk size by each family.)

In AYON
image

In OP
image

krishna8008:
> In the updated PR, I add a simple setting to set up customized chunk size(i.e. frame per task) for certain families. By default, the chunk size is 999999. (Of course we can do more concise settings which allows chunk size by each family.) I have also updated the houdini version for deadline too and it is no longer hardcoded.
> 
> In AYON image
> 
> In OP image

Is it possible to set this number based on publish rather than project level
I believe some tasks might require 1frame pertask, some might require 2+ frames pertask based on artist requirement.
Is it possible to add this option in the Publish dialogue box? so we can choose the frames per task and submit for farm publishing

mustafa-zarkash:
I didn't test it yet, but quick note: 
please bump houdini addon version

https://github.com/ynput/OpenPype/blob/ae53caacc057e77271b2fc2389362bb04bb53e14/server_addon/houdini/server/version.py#L1

```sugesstion
 __version__ = "0.1.2" 
```

moonyuet:
@krishna8008 
I have added Frame per task into publisher's setting, which allows you to edit the setting. (By default it's 99999, as the realted OP/AYON setting sets as 999999 )
image

you can always preset the frame per task in AYON/OP setting first, if you dont want to edit in publisher.

AYON
image

OP
image

openpype/hosts/houdini/plugins/publish/collect_chunk_size.py
```diff
@@ -0,0 +1,40 @@
...
+    def get_attribute_defs(cls):
+        return [
+            NumberDef("chunkSize",
+                      minimum=1,
+                      maximum=999999,
+                      decimals=0,
+                      default=cls.chunkSize,
+                      label="Frame Per Task")
+
+        ]
```
mustafa-zarkash, line 40: Applying data is automated as long as you use the same key used in settings `"chunk_size"`

```suggestion
    label = "Collect Chunk Size"

    def process(self, instance):
        instance.data["chunkSize"] =  self.chunk_size

    @classmethod
    def get_attribute_defs(cls):
        return [
            NumberDef("chunkSize",
                      minimum=1,
                      maximum=999999,
                      decimals=0,
                      default=cls.chunk_size,
                      label="Frame Per Task")
        ]
```

moonyuet: > Applying data is automated as long as you use the same key used in settings `"chunk_size"`

OP Studio settings apply there too. so the apply_setting function needs to be included. But I do agree with instance.data["chunkSize"] =  self.chunk_size

moonyuet:
> Could you also update description and name of this PR for nicer looking Release notes? Thx.

I have updated the name and description. Let me know if you want more information for the releases. Thanks!

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -398,6 +408,13 @@ def process(self, instance):
                 instance, instance_skeleton_data,
                 self.aov_filter, self.skip_integration_repre_list,
                 do_not_add_review)
+            if instance.data.get("expected_families", []):
```
kalisp, line 411: I don't understand this logic, what is expected value in `instance.data.get("expectedFiles"instance.data.get("expectedFiles")` ? 
Aren't you re-creating first call of `create_instances_for_aov` again? Shouldn't be enough to add there proper value `farm_cache`, how do you decide it should be actually triggered? 
Arent 'expectedFiles' list all the time? Only for AOVs it contains dictionary, without AOVs just a list of files? (Not sure how does single file look like, but I am expecting ['file1.txt'])

moonyuet: 1. instance.data.get("expected_families") stores the family type which needs later for file transmission from the work folder (which stores the work cached in deadline) to the publish folder. I can add one more instance.data["farm_cache"] = True to replace the farm_cache == True if it is really
2. I am not quite sure about single file looks like so I follow the logic similar to the image render submission. Maybe I can edit the logic similar to the old implementation of Max deadline (which uses only the list) to see how it goes.

kalisp: Ohh, shh...my bad, got confused by `expectedFiles` and `expected_families`, stupid speedreading course :).

Shouldn't be there:
```
instances = create_instances_for_aov(
                    instance, instance_skeleton_data,
                    self.aov_filter,
                    self.skip_integration_repre_list,
                    do_not_add_review,
                    bool(instance.data.get("expected_families")))
```   
instead of double calls to `create_instances_for_aov`? You are rewriting `instances` from line 407, not extending them, right? (If there are no `expected_families` it will do original instances, if they do, it will create farm cached.)
(BTW all the farm cached families are producing `expectedFiles` as a dict, eg. AOVs, not as a regular list, right?)

moonyuet: Expected_files are all in dict and you can find these lines:)
https://github.com/ynput/OpenPype/blob/f963be7cd56db464f6b91c6c0dd6b19785a9c341/openpype/hosts/houdini/plugins/publish/collect_cache_farm.py#L52-L53

openpype/hosts/houdini/api/lib.py
```diff
@@ -138,6 +138,9 @@ def get_output_parameter(node):
             return node.parm("ar_ass_file")
     elif node_type == "Redshift_Proxy_Output":
         return node.parm("RS_archive_file")
+    elif node_type == "ifd":
```
kalisp, line 141: Is this meant for this PR? Just wondering.

kalisp:
I would also expect to see mentions of this functionality in documentation, most likely only for Artists

openpype/hosts/houdini/api/lib.py
moonyuet: yes, as this PR is also adding ifd implementation which is also needed for farm caching.....

moonyuet:
> I would also expect to see mentions of this functionality in documentation, most likely only for Artists

I will update the documentation in both ayon.doc and the script to explain how it works for the hosts later. The changelog description is relatively less info for use cases.

mustafa-zarkash (commented):
I thought about giving it a quick test, 
abc and Bgeo  okay
image

Mantra IFD didn't work because point cache validator failed! 
Therefore I printed `instance.data`  and found this: 
```
{'family': 'pointcache', 'name': 'mantraifdFarmIFD', 'subset': 'mantraifdFarmIFD'}
```

I can't test redshift and arnold because I don't have license.

moonyuet:
> I thought about giving it a quick test, abc and Bgeo okay image
> 
> Mantra IFD didn't work because point cache validator failed! Therefore I printed `instance.data` and found this:
> 
> ```
> {'family': 'pointcache', 'name': 'mantraifdFarmIFD', 'subset': 'mantraifdFarmIFD'}
> ```
> 
> I can't test redshift and arnold because I don't have license.

I fixed the issue in the updated commit. can you please check again? Thanks @mustafa-zarkash

moonyuet:
> I thought about giving it a quick test, abc and Bgeo okay image
> 
> Mantra IFD didn't work because point cache validator failed! Therefore I printed `instance.data` and found this:
> 
> ```
> {'family': 'pointcache', 'name': 'mantraifdFarmIFD', 'subset': 'mantraifdFarmIFD'}
> ```
> 
> I can't test redshift and arnold because I don't have license.

@mustafa-zarkash Can you please test it now? Sorry I have been mixing up with other PRs which is related to split render job in Houdini.
It should also increment the workfile version in the updated commit.
Thank you!

mustafa-zarkash:
Thanks, I can publish BGEO, ABC and IFD 

image

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
```diff
@@ -0,0 +1,87 @@
...
+    label = "Collect Data for Cache"
+    chunkSize = 999999
+
+    def process(self, instance):
+        if instance.data.get("farm", True):
+            for family in instance.data["families"]:
+                instance.data["family"] = family
+                if family == "bgeo" or "abc":
+                    instance.data["family"] = "pointcache"
+                instance.data["subset"] = instance.name
```
BigRoy, line 25: This is a mess of broken code. Sorry ;)

```python
if family == "bgeo" or "abc"
```
equals:
```python
if family == True:
```
Not what you intended ;)
```python
if family in {"bgeo", "abc"}
```
I think is what you meant.

Also, you're setting subset to `instance.name` during the for loop which is weird to begin with - since in the for loop it's always the same value. But why do you need to override subset anyway - seems like a bug?

It feels like all you were trying to do is that _if_ bgeo or abc is the family that you want to label it pointcache. But why is that needed? What's the goal?

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
```diff
@@ -0,0 +1,87 @@
...
+            # instance.data["expected_families"]
+            # only created for farm caching
+            instance.data["expected_families"] = list()
```
BigRoy, line 28: Why is this `expected_families` with an underscore even though we also use `expectedFiles` just below?

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
```diff
@@ -0,0 +1,87 @@
...
+                # list of files
+                instance.data["files"].extend(files)
+            else:
+                instance.data["files"].append(output_parm.eval())
+            cache_files = {"_": instance.data["files"]}
+            instance.data["expectedFiles"].append(cache_files)
+
+            # update the colorspace data
+            colorspace_data = get_color_management_preferences()
+            instance.data["renderProducts"] = colorspace.ARenderProduct()
```
BigRoy, line 59: Where is this used?

Also, if this is needed - I suspect it'd need to be a list of render products?
And why is `ARenderProduct` in the `colorspace` lib? :)

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
moonyuet: I am struggling a little bit as both bgeo and abc share the same pointcache families.

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
```diff
@@ -0,0 +1,87 @@
...
+            if "expectedFiles" not in instance.data:
+                instance.data["expectedFiles"] = list()
+            cache_files = dict()
+            if instance.data.get("frames"):
+                files = self.get_files(
+                    instance, expected_filepath)
+                # list of files
+                instance.data["files"].extend(files)
+            else:
+                instance.data["files"].append(output_parm.eval())
```
BigRoy, line 53: Note that the published representation will always then end up be integrated as a sequence since it's still a list. I believe if it's a single file than `instance.data["files"] = output_parm.eval()` the `files` data should be a single file. I'm not entirely sure but I've seen that so much in the code base that I think that's the case - it's one of the biggest annoying things haha.

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
BigRoy: Struggling why? What issues are you facing with it?

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
moonyuet: `expected_families `is not the same with the `expectedFile`. `expected_families `is to make sure the file transmission of submit job would be in the correct family such as `pointcache` and `ifd`

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -189,6 +189,16 @@ def _submit_deadline_post_job(self, instance, job, instances):
...
+        if instance.data.get("expected_families", []):
+            output_dir = self._get_publish_folder(
+                anatomy,
+                deepcopy(instance.data["anatomyData"]),
+                instance.data.get("asset"),
+                instance.data["subset"],
+                instance.context,
+                instance.data["family"],
+                override_version
+            )
```
BigRoy, line 201: Please add a comment WHY this is needed - also, if we're overidding `output_dir` anyway this should be the `if branch` and the other logic the `else` branch so we don't compute the other for no reason.

moonyuet: you need to get the correct filepath for the farm caching. otherwises it will go to the publish folder of render family.

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
BigRoy: They are two different variables, that's fine. I'm just saying why does the formatting differ? Why is one lowercase with an underscore and the other _lowerCamelCase_.

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
moonyuet: If there is no instance.data["renderProducts"], the submit publish job would give KeyError if that instance

openpype/pipeline/farm/pyblish_functions.py
```diff
@@ -623,6 +625,13 @@ def _create_instances_for_aov(instance, skeleton, aov_filter, additional_data,
         new_instance = deepcopy(skeleton)
         new_instance["subset"] = subset_name
         new_instance["subsetGroup"] = group_name
+        if farm_cache:
+            new_instance["subset"] = instance.name
+            new_instance["subsetGroup"] = instance.name
```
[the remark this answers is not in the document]
moonyuet: I write this to make sure the subset name won't be like `renderAssetPointCacheMain`....

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
BigRoy: By the way, the concept of defining in the data what the final published family (let's refer to it as the product-type) should be instead of inheriting all of the families during the Pyblish publish process is something that was done before here: https://github.com/ynput/OpenPype/blob/c10781a662e9dd85839d8eb786e471b0709cbccf/openpype/hosts/houdini/plugins/publish/collect_usd_bootstrap.py#L85 and https://github.com/ynput/OpenPype/blob/4c280328b85b8cccbe1f76bf8ffcfc26ae34df94/openpype/hosts/houdini/plugins/publish/collect_instances_usd_layered.py#L83 where it was named `publishFamilies`. It hasn't gotten much use in OpenPype and I have no idea if that logic actually works - but just so you know there is basically similar functionality as your `expectedFamilies` where it was named `publishFamilies` instead.

openpype/hosts/houdini/plugins/create/create_arnold_ass.py
```diff
@@ -52,3 +54,11 @@ def create(self, subset_name, instance_data, pre_create_data):
...
         to_lock = ["ar_ass_export_enable", "family", "id"]
         self.lock_parameters(instance_node, to_lock)
+
+    def get_pre_create_attr_defs(self):
+        attrs = super(CreateArnoldAss, self).get_pre_create_attr_defs()
+        return attrs + [
+            BoolDef("farm",
+                    label="Submitting to Farm",
+                    default=False)
+        ]
```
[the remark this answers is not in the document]
moonyuet: maybe we can implement similar attribute in the `collect_cache_farm`

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
BigRoy: I think we're mixing up two things too much - the **SubmitPublishJob** is about creating an additional 'publish job'  which publishes rendered outputs (renders/image sequences) from a host DCC after the render job itself has finished. The docstring is also quite specific about that it is for render jobs / image sequences.

There's assembly logic (for tiled rendering), there's aov logic, multiple cameras, multiple "separate publish jobs" from a single instance, etc. I don't see exactly why're ending up with quite a convoluted implementation as we currently are to publish instances.

I feel like 50% of the logic in this PR is about somehow 'wrestling' your new instances into that plug-in so that you can just get something as simple as a pointcache published on the farm.

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
moonyuet: > By the way, the concept of defining in the data what the final published family (let's refer to it as the product-type) should be instead of inheriting all of the families during the Pyblish publish process is something that was done before here:
> 
> https://github.com/ynput/OpenPype/blob/c10781a662e9dd85839d8eb786e471b0709cbccf/openpype/hosts/houdini/plugins/publish/collect_usd_bootstrap.py#L85
> 
> and
> https://github.com/ynput/OpenPype/blob/4c280328b85b8cccbe1f76bf8ffcfc26ae34df94/openpype/hosts/houdini/plugins/publish/collect_instances_usd_layered.py#L83
> 
> where it was named `publishFamilies`. It hasn't gotten much use in OpenPype and I have no idea if that logic actually works - but just so you know there is basically similar functionality as your `expectedFamilies` where it was named `publishFamilies` instead.

Will rename it to `publishFamilies` in the next PR

BigRoy:
> Current version of deadline does not support vdb farm caching. (Tried with the manual deadline submission)

We've done VDB caching just fine in Deadline. What issues did you get?

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
moonyuet: > I think we're mixing up two things too much - the **SubmitPublishJob** is about creating an additional 'publish job' which publishes rendered outputs (renders/image sequences) from a host DCC after the render job itself has finished. The docstring is also quite specific about that it is for render jobs / image sequences.
> 
> There's assembly logic (for tiled rendering), there's aov logic, multiple cameras, multiple "separate publish jobs" from a single instance, etc. I don't see exactly why're ending up with quite a convoluted implementation as we currently are to publish instances.
> 
> I feel like 50% of the logic in this PR is about somehow 'wrestling' your new instances into that plug-in so that you can just get something as simple as a pointcache published on the farm.

do you mean we need to have another submit publish job for farm cache?

openpype/pipeline/farm/pyblish_functions.py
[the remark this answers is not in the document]
BigRoy: I think we're better of solving why it gets a `instance.data["subset"]` named like that in the first place for a `pointcacheMain` instance. This feels like patching up a bug elsewhere.

moonyuet:
> > Current version of deadline does not support vdb farm caching. (Tried with the manual deadline submission)
> 
> We've done VDB caching just fine in Deadline. What issues did you get?



> > Current version of deadline does not support vdb farm caching. (Tried with the manual deadline submission)
> 
> We've done VDB caching just fine in Deadline. What issues did you get?

For some reason, I can't publish vdb into the deadline(get no vdb files), but maybe it's my issue, I can add it back laterwards

openpype/modules/deadline/plugins/publish/submit_publish_job.py
BigRoy: But the output of a pointcache is just whatever the ROP node is set to, right? What output dir is this that we need?

The way I see it all we need to do is:

- Submit a Houdini Job to Deadline
- Deadline launches Houdini and triggers OpenPype's regular "publish" (including all its validations!) but only for the instances the user has chosen on the submission.
    - It publishes and generates the output as usual, as regular publishing does.
    - Having the farm also validate the instances means it can actually confirm all loaded plug-ins still work, paths are valid, etc. - everything that we want to ensure it just doesn't go "render a rop node" and then producing different output.
    - We need almost no code changes to allow publishing on the farm.
    - Note: We don't need to keep track of any `.json` files or pre-processed data.

Or what else do we need?

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
```diff
@@ -0,0 +1,87 @@
...
+class CollectDataforCache(pyblish.api.InstancePlugin):
+    """Collect data for caching to Deadline."""
+
+    order = pyblish.api.CollectorOrder + 0.04
+    families = ["ass", "pointcache",
+                "mantraifd", "redshiftproxy"]
+    hosts = ["houdini"]
+    targets = ["local", "remote"]
+    label = "Collect Data for Cache"
+    chunkSize = 999999
```
BigRoy, line 17:
```suggestion
```

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
```diff
@@ -0,0 +1,87 @@
...
+            # update the colorspace data
+            colorspace_data = get_color_management_preferences()
+            instance.data["renderProducts"] = colorspace.ARenderProduct()
+            instance.data["colorspaceConfig"] = colorspace_data["config"]
+            instance.data["colorspaceDisplay"] = colorspace_data["display"]
+            instance.data["colorspaceView"] = colorspace_data["view"]
+            self.log.debug("{}".format(instance.data))
+
+        else:
+            self.log.debug("Caching at farm disabled!")
```
BigRoy, line 66: Here's basically this logic re-written somewhat (consider it pseudocode) but with some question comments added in there. There's a lot going on here that's no apparent at first sight WHY it's needed - so it'd be good to put some comments in to explain why certain things are needed. E.g. 
- Why do we need expectedFamilies? Why can't we rely on the instance's families? 
- Why are we remapping subsets? 
- That colorspace data being collected for the instance doesn't seem very "renderfarm" related. Why can't there just be a a single `CollectColorspaceConfig` which collects it maybe to the context? (collects it once instead of per instance), etc.

```suggestion
        if not instance.data.get("farm", True):
            self.log.debug("Caching on farm is disabled. "
                           "Skipping farm collecting.")
            return

        # Why do we need this particular collector to collect the expected
        # output files from a ROP node. Don't we have a dedicated collector
        # for that yet?
        # Collect expected files
        ropnode = hou.node(instance.data["instance_node"])
        output_parm = lib.get_output_parameter(ropnode)
        expected_filepath = output_parm.eval()

        instance.data.setdefault("files", list())
        instance.data.setdefault("expectedFiles", list())
        if instance.data.get("frames"):
            files = self.get_files(instance, expected_filepath)
            # list of files
            instance.data["files"].extend(files)
        else:
            # single file
            instance.data["files"].append(output_parm.eval())
        cache_files = {"_": instance.data["files"]}

        # Convert instance family to pointcache if it is bgeo or abc
        # because ???
        for family in instance.data["families"]:
            if family == "bgeo" or "abc":
                instance.data["family"] = "pointcache"
                break

        instance.data.update({
            "expectedFamilies": [instance.data["family"]],

            # Why are we overriding the subset?
            "subset": instance.name,

            # Why do we need these in `instance.data`?
            "plugin": "Houdini",
            "publish": True
        })
        instance.data["families"].append("publish.hou")
        instance.data["expectedFiles"].append(cache_files)

        # update the colorspace data
        colorspace_data = get_color_management_preferences()
        instance.data["renderProducts"] = colorspace.ARenderProduct()
        instance.data["colorspaceConfig"] = colorspace_data["config"]
        instance.data["colorspaceDisplay"] = colorspace_data["display"]
        instance.data["colorspaceView"] = colorspace_data["view"]
        self.log.debug("{}".format(instance.data))
```

openpype/modules/deadline/plugins/publish/submit_publish_job.py
moonyuet: > But the output of a pointcache is just whatever the ROP node is set to, right? What output dir is this that we need?
> 
> Houdini just renders the ROP - it renders to where the ROP node is set to; and that defines our staging dir, voila. Why does the farm need to be so special about this if the farm tries to basically do what the user would otherwise do on his own machine (except for being headless)
> 
> The way I see it all we need to do is:
> 
> * Submit a Houdini Job to Deadline
> * Deadline launches Houdini and triggers OpenPype's regular "publish" (including all its validations!) but only for the instances the user has chosen on the submission.
>   
>   * It publishes and generates the output as usual, as regular publishing does.
>   * Having the farm also validate the instances means it can actually confirm all loaded plug-ins still work, paths are valid, etc. - everything that we want to ensure it just doesn't go "render a rop node" and then producing different output.
>   * We need almost no code changes to allow publishing on the farm.
>   * Note: We don't need to keep track of any `.json` files or pre-processed data.
> 
> Or what else do we need?

You need to get the latest version of the published families folder in order to get the file transmission right.
let's say `publish\pointcache\pointcacheMain\v020`

BigRoy: We don't need to if the machine that "rendered the rop" itself would also publish it right.
What's the reason we explicitly want to separate on the farm the caching from the publishing [like in this screenshot](https://user-images.githubusercontent.com/20871534/265078060-9dcf1a48-d24d-409f-8129-6f2d7c07794e.png). Is that crucial?

I'm a bit wary that the implementation changes I'm seeing here will be total madness to get similar support across all families and all hosts - even though it should be as trivial as "just run the publishing on the farm without the integrating" and just separating the integrating part if we just want to have the 'publish' job separate from a cache job.

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
BigRoy: If that means we'll have to jump through 100 less hoops to get this feature I'd say so, yes. I feel like you're trying to wrangle your way through something that was never designed to do what you want it to do.

openpype/hosts/houdini/plugins/publish/collect_cache_farm.py
```diff
@@ -0,0 +1,77 @@
...
+        """Get the files with the frame range data
+
+        Args:
+            instance (_type_): instance
+            output_parm (_type_): path of output parameter
+
+        Returns:
+            files: a list of files
+        """
+        files = []
```
Minkiu, line 68: This could be:
```py
files = [
	os.path.join(directory, frame).replace("\\", "/")
	for frame in instance.data["frames"]
]
```

openpype/hosts/houdini/plugins/publish/extract_mantra_ifd.py
```diff
@@ -0,0 +1,51 @@
...
+            return
+
+        ropnode = hou.node(instance.data.get("instance_node"))
+        output = ropnode.evalParm("soho_diskfile")
+        staging_dir = os.path.dirname(output)
+        instance.data["stagingDir"] = staging_dir
+        file_name = os.path.basename(output)
+
+        files = instance.data["frames"]
+        missing = []
```
Minkiu, line 30: This could be:
```py
missing = [
	frame
	for frame in instance.data["frames"]
	if not os.path.exists(os.path.normpath(os.path.join(staging_dir, frame)))
]
```

openpype/modules/deadline/plugins/publish/submit_houdini_cache_deadline.py
```diff
@@ -0,0 +1,181 @@
...
+        job_info.SecondaryPool = instance.data.get("secondaryPool")
+
+        attr_values = self.get_attr_values_from_data(instance.data)
+
+        job_info.ChunkSize = instance.data["chunkSize"]
+        job_info.Comment = context.data.get("comment")
+        job_info.Priority = attr_values.get("priority", self.priority)
+        job_info.Group = attr_values.get("group", self.group)
+
+        keys = [
```
Minkiu, line 110: Maybe rename it to `third_party_api_keys` ?

openpype/modules/deadline/plugins/publish/submit_houdini_cache_deadline.py
```diff
@@ -0,0 +1,181 @@
...
+        ]
+
+        # Add OpenPype version if we are running from build.
+        if is_running_from_build():
+            keys.append("OPENPYPE_VERSION")
+        # Add mongo url if it's enabled
+        if self._instance.context.data.get("deadlinePassMongoUrl"):
+            keys.append("OPENPYPE_MONGO")
+
+        environment = dict({key: os.environ[key] for key in keys
```
Minkiu, line 123: Could we leverage `acre` here? Also, I think there's a flag for Deadline which will pass the current environment to the job (or I might be misremembering)

website/docs/artist_hosts_houdini.md
```diff
@@ -83,6 +83,24 @@ select your render camera.
 All the render outputs are stored in the pyblish/render directory within your project path.\
 For Karma-specific render, it also outputs the USD render as default.
 
+## Publishing cache to Deadline
+Five Instance (Arnold ASS, Pointcache(BGEO and ABC) and Redshift Proxy) are supported for
```
Minkiu, line 87: I can only count 4?
.ass, .bgeo, .abc and whatever redshit proxy extension is?

openpype/modules/deadline/plugins/publish/submit_houdini_cache_deadline.py
moonyuet: > a flag for Deadline which will pass the current environment to the job

do you mean `job_info.add_render_job_env_var()` ?

website/docs/artist_hosts_houdini.md
moonyuet: .ass, .bgeo, .abc, .rsproxy, and .vdb. I will update on that

Minkiu (approved):
Haven't tested but code looks good, I was also wondering if it's worth implementing both
`get_instance_attr_defs` and `get_pre_create_attr_defs` directly in the `plugin.HoudiniCreator` class?

openpype/modules/deadline/plugins/publish/submit_houdini_cache_deadline.py
Minkiu: That seems like single env vars, maybe I remember wrong, but when you submit to deadline, you have the option to pass the whole of the host environment as is

BigRoy: We don't want to pass the full environment because we do still want the deadline machine to 'build their OS-compatible environment' for the application, etc. - There are just specific vars that we DO want to pass along in certain situations. That's why we need to be explicit about it.

website/docs/artist_hosts_houdini.md
BigRoy: I feel like this is also better written as a clear list, e.g.

---

Caching on the farm is supported for:

- Arnold ASS
- Pointcache (BGEO and ABC)
- Redshift Proxy

BigRoy:
> I was also wondering if it's worth implementing both
get_instance_attr_defs and get_pre_create_attr_defs directly in the plugin.HoudiniCreator class?

As far as I know this PR requires special work for each particular creator/family to be supported for Farm publishing - as such, not ALL creators would support the farm toggle and thus it might not make sense to be on the HoudiniCreator class for ALL creators. Right?

Minkiu:
Hmm I see then, nevermind :D

openpype/settings/defaults/project_settings/deadline.json
MustafaJafar, line 1: Key `ProcessSubmittedCacheJobOnFarm` hasn't corresponding dictionary in the `schema_project_deadline.json`

moonyuet: nice catch, it could be that I miss copying the file when I tried to force push remove unnecessary files in this PR

MustafaJafar (commented):
I gave it a quick test, cache jobs failed  because of missing env vars


```
2023-09-21 14:44:00:  0: PYTHON: *** GlobalJobPreload start ...
2023-09-21 14:44:00:  0: PYTHON: >>> Getting job ...
2023-09-21 14:44:00:  0: PYTHON: >>> Injecting OpenPype environments ...
2023-09-21 14:44:00:  0: PYTHON: --- OpenPype executable: E:/Ynput/OpenPype/tools/openpype_console.bat
2023-09-21 14:44:00:  0: PYTHON: >>> Temporary path: C:\Users\MUSTAF~1\AppData\Local\Temp\20230921114400604761_28078634-5874-11ee-95d4-f02f74db125e.json
2023-09-21 14:44:00:  0: PYTHON: Traceback (most recent call last):
2023-09-21 14:44:00:  0: PYTHON:   File "C:\ProgramData\Thinkbox\Deadline10\workers\Major-Kalawy\plugins\650c2c5f4f74b3aca0aaa02a\GlobalJobPreLoad.py", line 354, in inject_openpype_environment
2023-09-21 14:44:00:  0: PYTHON:     "Missing required env vars: AVALON_PROJECT, AVALON_ASSET,"
2023-09-21 14:44:00:  0: PYTHON: RuntimeError: Missing required env vars: AVALON_PROJECT, AVALON_ASSET, AVALON_TASK, AVALON_APP_NAME
2023-09-21 14:44:00:  0: PYTHON: !!! Injection failed.
```

moonyuet:
@MustafaJafar can you test it again?

MustafaJafar:
@moonyuet 
It submits to farm and I assume it works fine. 

However, Cache jobs and Render Jobs doesn't inject environment variables  in the same way which is not compatible with this [PR](https://github.com/ynput/OpenPype/pull/5622)

Oh, Cache Jobs don't inject any environments at all! 

image

I think it's related to Job Environments 
Injection requires either `OPENPYPE_RENDER_JOB` or `OPENPYPE_REMOTE_PUBLISH` to be 1 
I think we can add one more `OPENPYPE_CACHE_JOB` and its Ayon equivalent `AYON_CACHE_JOB` then updating `GlobalJobPreLoad`

<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

----------

I have a little question, could you tell me the difference between these two `HoudiniCacheSubmitDeadline` and `ProcessSubmittedCacheJobOnFarm` as I'm a little confused ?

fabiaserra:
Hello, I'm sorry if this comes out as a bit harsh but I think the approach this PR is taking to support caching in the farm is wrong and over engineered.

First of all, caching in the farm (and rendering or any other Houdini processes) are already supported by third-party toolsets (Deadline, HQueue, Tractor...) and in WAY more powerful ways that this PR tries to accomplish and the OP plugin framework can manage. This is duplicating all of that logic in OP and adding 1,398 more lines to the already super complex code base!! Most Houdini TDs are already familiarized with those vanilla workflows and having them learn this other "black box" approach through OP is backwards and doesn't add any benefit in my opinion. You can see an example of a very normal submission to the farm here https://github.com/ynput/OpenPype/pull/5621#issuecomment-1732166830

OpenPype shouldn't try to orchestrate the extract/render dependencies of the Houdini node graph, that's already done by these schedulers/submitters, we just need means to be able to run OP publish tasks of the generated outputs, but without doing any gimmicks, just taking a path, a family, a few other data inputs and registering it to the DB so it runs the other integrate plugins of OP like publishing that to SG/ftrack as well (and ideally the API for doing that in OP should be super straightforward to call it from anywhere! the current json file input isn't the best access point to that funcionality). If we wanted to help the existing vanilla submission OP could provide a wrapper of the vanilla submitters so it sets some reasonable defaults and we could intersect the submission to run some pre-validations on the graph... set some parms that might be driven by the OP settings or create utility HDAs to facilitate the creation of the submitted graph so frame dependencies are set correctly and chunk sizes for simulations... but that's it, we don't need to reinvent the wheel by interpreting how the graph needs to be evaluated.

On the other hand, I still don't quite get why the existing `submit_publish_job` is limited to "render" families and why it's not abstracted in a simple reusable module that any plugins that require to submit to Deadline can reuse it. This PR showed how a lot of the code had to be duplicated again with most of the lines exactly the same, doubling in technical debt. This PR https://github.com/ynput/OpenPype/pull/5451/files goes in the right direction in abstracting some of those things but the right approach should be to remove all of the noise from submit publish job that's render specific and make use of the same util module every time we just need to run an OP publish task in the farm. However, as ï said initially, I don't think we should even take this approach for Houdini and we should just leverage the existing farm submitters code, but this is relevant for any other tasks that we choose to submit to the farm, we are going to have to write a "submit_<insert_family_type>_job" set of plugins for each?

Extra notes:

* I feel like I already mentioned this elsewhere in another PR but have you gotten anyone requesting to publish IFDs? Is there any use cases to load those after the fact? AFAIK those aren't really consumable by anything other than Houdini and just as a pre-process to then run the render, it's not like ASS files that can be used as an interchange format from the Arnold plugins.

moonyuet:
> Hello, I'm sorry if this comes out as a bit harsh but I think the approach this PR is taking to support caching in the farm is wrong and over engineered.
> 
> First of all, caching in the farm (and rendering or any other Houdini processes) are already supported by third-party toolsets (Deadline, HQueue, Tractor...) and in WAY more powerful ways that this PR tries to accomplish and the OP plugin framework can manage. This is duplicating all of that logic in OP and adding 1,398 more lines to the already super complex code base!! Most Houdini TDs are already familiarized with those vanilla workflows and having them learn this other "black box" approach through OP is backwards and doesn't add any benefit in my opinion. You can see an example of a very normal submission to the farm here [#5621 (comment)](https://github.com/ynput/OpenPype/pull/5621#issuecomment-1732166830)
> 
> OpenPype shouldn't try to orchestrate the extract/render dependencies of the Houdini node graph, that's already done by these schedulers/submitters, we just need means to be able to run OP publish tasks of the generated outputs, but without doing any gimmicks, just taking a path, a family, a few other data inputs and registering it to the DB so it runs the other integrate plugins of OP like publishing that to SG/ftrack as well (and ideally the API for doing that in OP should be super straightforward to call it from anywhere! the current json file input isn't the best access point to that funcionality). If we wanted to help the existing vanilla submission OP could provide a wrapper of the vanilla submitters so it sets some reasonable defaults and we could intersect the submission to run some pre-validations on the graph... set some parms that might be driven by the OP settings or create utility HDAs to facilitate the creation of the submitted graph so frame dependencies are set correctly and chunk sizes for simulations... but that's it, we don't need to reinvent the wheel by interpreting how the graph needs to be evaluated.
> 
> On the other hand, I still don't quite get why the existing `submit_publish_job` is limited to "render" families and why it's not abstracted in a simple reusable module that any plugins that require to submit to Deadline can reuse it. This PR showed how a lot of the code had to be duplicated again with most of the lines exactly the same, doubling in technical debt. This PR https://github.com/ynput/OpenPype/pull/5451/files goes in the right direction in abstracting some of those things but the right approach should be to remove all of the noise from submit publish job that's render specific and make use of the same util module every time we just need to run an OP publish task in the farm. However, as ï said initially, I don't think we should even take this approach for Houdini and we should just leverage the existing farm submitters code, but this is relevant for any other tasks that we choose to submit to the farm, we are going to have to write a "submit_<insert_family_type>_job" set of plugins for each?
> 
> Extra notes:
> 
> * I feel like I already mentioned this elsewhere in another PR but have you gotten anyone requesting to publish IFDs? Is there any use cases to load those after the fact? AFAIK those aren't really consumable by anything other than Houdini and just as a pre-process to then run the render, it's not like ASS files that can be used as an interchange format from the Arnold plugins.

I actually developed this PR some years ago and been struggling a lot but I think what you said are right.
If you talk about HQueue and Deadline or some other external plugins for queuing the cache, it is not possible by this case.
I saw the communitg

MustafaJafar (approved):
I've successfully published these types with the default deadline options settings. 
- ifd
- point cache abc
- point cache bgeo
- vdb
> I can't test the other two. 

OpenPype.
image

Ayon.
image

----

Could we add a validator that checks if camera exists ? 
I created a mantra IFD, set it to farm. 
and then my job failed with no clue in the log,
Although, it was my fault that I didn't update the camera path on the node, I think such a validator will help. 
we can create a ticket later. 
image

antirotor (approved):
I agree with @fabiaserra and I think we must take some effort to create more lightweight Deadline support. But that is more of a long-term plan because of the vanilla deadline support in DCCs differs vastly. There are multiple issues with the farm publishing currently, starting from farm specific attributes in the Publisher UI (and how are we handling local/farm rendering) to unifying deadline specific attributes on host/renderers. So I would merge it as it is adding functionality that is useful even if it is not "final" solution.

PR #4904, opened 2023-04-25
Maya Load References - Add Display Handle Setting

mre7a:
## Changelog Description
When we load a reference in Maya using OpenPype loader, display handle is checked by default and prevent us to select easily the object in the viewport. 
image

I understand that some productions like to keep this option, so I propose to add display handle to the reference loader settings. 
image

## Testing notes:
1. Using Openpype loader in Maya, load an asset as reference. 
2. Verify that Display Handle is following the settings.

### DIFF
--- a/openpype/hosts/maya/plugins/load/load_reference.py
+++ b/openpype/hosts/maya/plugins/load/load_reference.py
@@ -162,9 +162,12 @@ class ReferenceLoader(openpype.hosts.maya.api.plugin.ReferenceLoader):
                     with parent_nodes(roots, parent=None):
                         cmds.xform(group_name, zeroTransformPivots=True)
 
-                cmds.setAttr("{}.displayHandle".format(group_name), 1)
-
                 settings = get_project_settings(os.environ['AVALON_PROJECT'])
+
+                display_handle = settings['maya']['load'].get('reference_loader', {}).get(
+                    'display_handle', True)
+                cmds.setAttr("{}.displayHandle".format(group_name), display_handle)
+
                 colors = settings['maya']['load']['colors']
                 c = colors.get(family)
                 if c is not None:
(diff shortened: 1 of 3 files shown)


tokejepsen (changes requested):
Great work!

Just need those Hound fixes and we can merge.

tokejepsen:
> I'm wondering whether we should maybe also add the option to allow enabling/disabling it for certain families/profiles? We mostly disabled it because we really didn't like it on e.g. rigs but did like it on some others. Nevertheless we still ended up disabling it completely just it because was much simpler at the time.

Not sure we need that right now unless you need it?
If someone wants it, they can implement profiles later.

mre7a:
Thanks a lot. 

> Just need those Hound fixes
> Would need some cosmetics cleanup

Oops. Sorry.  vscode issues :)

openpype/hosts/maya/plugins/load/load_reference.py
```diff
@@ -169,9 +169,13 @@ def process_reference(self, context, name, namespace, options):
...
 
-                cmds.setAttr("{}.displayHandle".format(group_name), 1)
-
                 settings = get_project_settings(os.environ['AVALON_PROJECT'])
+
+                display_handle = settings['maya']['load'].get(
+                    'reference_loader', {}).get('display_handle', True)
+                cmds.setAttr(
+                    "{}.displayHandle".format(group_name), display_handle)
+
```
tokejepsen, line 178:
```suggestion

                display_handle = settings['maya']['load'].get(
                    'reference_loader', {}
                ).get('display_handle', True)
                cmds.setAttr(
                    "{}.displayHandle".format(group_name), display_handle
                )

```

openpype/hosts/maya/plugins/load/load_reference.py
```diff
@@ -181,7 +185,8 @@ def process_reference(self, context, name, namespace, options):
                                  (float(c[1]) / 255),
                                  (float(c[2]) / 255))
 
-                cmds.setAttr("{}.displayHandle".format(group_name), 1)
+                cmds.setAttr(
+                    "{}.displayHandle".format(group_name), display_handle)
```
tokejepsen, line 189:
```suggestion
                cmds.setAttr(
                    "{}.displayHandle".format(group_name), display_handle
                )
```

antirotor (approved):

BigRoy (approved):
Great work, looks good to me.

ClementHector:
Is there something missing to be adopted?

antirotor (approved):
Looks great!

PR #4909, opened 2023-04-27
Igniter: fix error reports in silent mode

### LINKED ISSUE #4908: Bug: Igniter and error silencing

### Current Behavior:

Igniter (mainly start.py) script has so called *silence_mode* used mostly in commands path where additional output isn't needed and can be hinderance. Problem is that when those paths ends up with fatal error, silence mode is still in place so you can't see errors (even though there are prints trying to show something). This complicate debugging troubles, mainly in Deadline log in case of missing OP versions, etc.

### Expected Behavior:

Just before commands fail with non-zero exit codes, at least on known errors, we should disable silence mode so the output is visible in the console.

### Version

3.15.5

Windows

### Steps To Reproduce:

Replicate it by forcing the silent commands to fail - for example `extractenvironments` with requested non-existent version.

### Relevant log output:

_No response_

### Additional context:

_No response_

[cuID:[OP-5812](https://app.clickup.com/t/863gkf39x)]

antirotor:
## Changelog Description
Some errors in silent mode commands in Igniter were suppressed and not visible for example in Deadline log.

## Additional info
This PR is forcing output just before non-zero (known error) exits from Igniter, so if there is fatal error, it is always shown.

## Testing notes:
1. Try to force condition like non-existent OP version on Deadline Job
2. OP should fail on the Deadline with "extract environments failed", but this time, you should see the reason in job log.

Resolves #4908

[cuID:OP-5812]


start.py
```diff
@@ -1049,8 +1050,8 @@ def boot():
         local_version = OpenPypeVersion.get_installed_version_str()
 
     if "validate" in commands:
-        _boot_validate_versions(use_version, local_version)
-        sys.exit(0)
+        valid = _boot_validate_versions(use_version, local_version)
+        sys.exit(1 if not valid else 0)
```
iLLiCiTiT, line 1054:
```suggestion
        sys.exit(0 if valid else 1)
```

iLLiCiTiT (approved):
LGTM

PR #4910, opened 2023-04-27
Deadline: escape rootless path in publish job

antirotor:
## Changelog Description
If the publish path on Deadline job contains spaces or other characters, command was failing because the path wasn't properly escaped. This is fixing it.

## Testing notes:
1. Try to submit publish job with space in the path
2. Everything should work as expected

### DIFF
--- a/openpype/modules/deadline/plugins/publish/submit_publish_job.py
+++ b/openpype/modules/deadline/plugins/publish/submit_publish_job.py
@@ -275,7 +275,7 @@ class ProcessSubmittedJobOnFarm(pyblish.api.InstancePlugin):
         args = [
             "--headless",
             'publish',
-            rootless_metadata_path,
+            '"{}"'.format(rootless_metadata_path),
             "--targets", "deadline",
             "--targets", "farm"
         ]


openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -275,7 +275,7 @@ def _submit_deadline_post_job(self, instance, job, instances):
         args = [
             "--headless",
             'publish',
-            rootless_metadata_path,
+            '"{}"'.format(rootless_metadata_path),
```
iLLiCiTiT, line 278: I think this will break if the space is not there. Rather use `subprocess.list2cmd(args)` when filling `"Arguments"`.

iLLiCiTiT (changes requested):
Rather use `subprocess.list2cmdline(args)` where arguments are filled to job than always fill quotes.

openpype/modules/deadline/plugins/publish/submit_publish_job.py
antirotor: Not sure about that. Apparently even Python isn't really sure if it should be used at all - https://discuss.python.org/t/why-is-subprocess-list2cmdline-not-documented/25044/11

antirotor: TL;DR 
`subprocess.list2cmdline()` is not multiplatfrom compatible, doesn't work with `shell=True`, might be deprecated in future, etc.

iLLiCiTiT: > subprocess.list2cmdline() is not multiplatfrom compatible

It is not, but it handles spaces the same way for all of them, you don't want to do multiplatform strange things.

> doesn't work with shell=True

From what I see in the forum, it's misunderstanding of function usage and purpose.

> might be deprecated in future, etc.

I don't think so. None of subprocess commands would work.

If you don't want to use it, then at least make sure the metadata path is stripped where is used, because arguments handler can pass `"c:/path_without_space/file"` as string `"c:/path_without_space/file"` instead of `c:/path with space/file`.

antirotor: As far as I know, this is used directly in Deadline API and who knows how it works there :) I suppose it needs to be tested. BTW notes like this in CPython subprocess code:

```python
# NOTE: We intentionally exclude list2cmdline as it is
# considered an internal implementation detail.  issue10838.
```

PR #4912, opened 2023-04-28
Fusion - Fixed "optional validation"

EmberLightVFX:
## Changelog Description
Added OptionalPyblishPluginMixin and is_active checks for all publish tools that should be optional

## Additional info
I did add the needed code to `increment_current_file.py` but the optional-switch doesn't show up in the publisher UI. Maybe it doesn't work for integrators?

## Testing notes:
1. Add a BG node set to something else than 32bit
2. Disable the 32bit check in publisher
3. Publish without getting stopped!


BigRoy:
The Increment file plugin is a **Context**Plugin and runs over the context and not per instance. As such the toggle is on the context (previously that item was called _Options_ but has recently been renamed to _Context_ again.)

afbeelding

---

I'm thinking that maybe if **NOTHING** is selected in the Publish tab then maybe it should auto-select the Context entry @iLLiCiTiT @mkolar what do you think? So that it defaults to this:
afbeelding

Instead of this:
afbeelding

BigRoy (approved):
Changes look good to me. I tested it in Fusion and both optional toggles seem to do what they intend to do.

EmberLightVFX:
As the `increment current file` toggle is located under Context there isn't totally clear what file it's talking to. Maybe changing its name to `increment current workfile version` would be better.

openpype/hosts/fusion/plugins/publish/increment_current_file.py
```diff
@@ -1,29 +1,38 @@
...
-            raise RuntimeError("Skipping incrementing current file because "
-                               "submission to render farm failed.")
+        if any(
+            plugin.__name__ == "FusionSubmitDeadline"
+            for plugin in errored_plugins
+        ):
+            raise RuntimeError(
+                "Skipping incrementing current file because "
+                "submission to render farm failed."
+            )
```
antirotor, line 35: Just two things - raising Exception here will stop the publish, so maybe "Skipping ..." isn't the right word here as (at least for me) it suggest it will continue, just with the warning. I know that this is at the Integration phase so all integrators before already did their job, but still - if we are rising error, maybe the message should be more imperative 🤔 

Also, it should be `KnownPublishError` probably (`from openpype.pipeline import KnownPublishError`)

antirotor (dismissed):
just a little comment on exception, otherwise looking great!

openpype/hosts/fusion/plugins/publish/increment_current_file.py
EmberLightVFX: Thanks for the note :) Updated to this!

BigRoy:
Can we merge this @antirotor @jakubjezek001 ?

PR #4913, opened 2023-04-28
Maya: Refactor Look assignment and support path based assignments

BigRoy:
## Changelog Description

Draft to refactor the look assignment functionality in Maya, including a start to support Path based assignments next to the _legacy_ `cbId` attribute based assignments.

Note, this is a **non-working draft!**

- [x] Store a new `source_id` attribute as replacement to `cbId`
  - [ ] Add validators 
- [ ] Refactor look lib
  - [x] Move look related code to dedicated `lib_looks`
  - [ ] Move vrayproxy + aistanding, etc. from tool to api lib too?
  - [ ] Allow the look assign functions to also assign to vrayproxy + aistandin even without the look manager tool. 
- [ ] Implement publishing looks with path based information
- [ ] Implement assigning looks with path based information
  - [ ] Implement for normal maya nodes (regular shader assignments)
  - [ ] Implement for VRayProxy
  - [ ] Implement for aiStandin

## Additional info

This is a draft PR to work towards shader assignments to be [Path based versus ID based like described in this discussion.](https://github.com/ynput/OpenPype/discussions/1559).

Currently it still stores metadata per mesh, in particular `{project_name}:{asset_id}` but no mesh related identifier that the `cbId` contained: `{mesh_id}`. As a comparison:

- `cbId` stores `{asset_id}:{mesh_id}`
- `root_id` stores `{project_id}:{asset_id}`.

In the current draft I've decided to include project as well to allow shader assignments to potentially allow assignments on library assets. Without storing at least the asset I wasn't sure how to detect what looks are applicable for that particular mesh.

I've been wondering whether maybe instead it would've been easier to store the `{mesh_path}` on the nodes as well so that even if at some point in the publishing the exact full path isn't a match you could still identify it. Or maybe we need to have marked `root` nodes so that paths were always identified as if that `root` was the top level. Anyway, these are all valid discussion points for the [ID vs Names discussion itself.](https://github.com/ynput/OpenPype/discussions/1559).

## Testing notes:
1. Launch maya
2. Publish a model
3. Publish a look for the model
4. Load the model, assign the look
5. Load an animation from a rig that uses the model, assign the look
6. Round-trip geometry from Maya -> Houdini -> Maya to test the look workflow


mkolar:

We're closing it down, but we'll he happy for a new PR to [ynput/ayon-core](https://github.com/ynput/ayon-core) repository.

PR #4914, opened 2023-04-28
Validators for Frame Range in Max

moonyuet:
## Changelog Description
Switch Render Frame Range Type to 3 for specific ranges (initial setup for the range type is 4)
Reset Frame Range will also set the frame range for render settings
Render Collector won't take the frame range from context data but take the range directly from render setting
Add validators for render frame range type and frame range respectively with repair action

## Additional info
Similar to maya, validator for frame range is optional. By default, it is off.
The one which check the render frame range type is mandatory.

## Testing notes:
1. Start 3dmax with Openpype
2. Run the publisher
3. if you have wrong render frame range type, it will error out and you can fix the type with repair action.
4. if you turn on the frame range validator, it will error out when the frame range from the render setting doesn't align with the database. You can fix it by performing repair action.


openpype/hosts/max/plugins/publish/validate_frame_range_type.py
```diff
@@ -0,0 +1,32 @@
...
+    label = "Validate Render Frame Range Type"
+    order = ValidateContentsOrder
+    families = ["maxrender"]
+    hosts = ["max"]
+    actions = [RepairAction]
+
+    def process(self, instance):
+        if rt.rendTimeType != 3:
+            raise PublishValidationError("Incorrect type of frame range"
+                                         " used in render setting..")
```
BigRoy, line 27: The validation error should report what value is wrong and what value it should be set to. I feel even though this has a repair action that the report itself is still too vague.

openpype/hosts/max/plugins/publish/validate_frame_range.py
```diff
@@ -0,0 +1,60 @@
...
+
+        if frame_start != inst_frame_start:
+            raise PublishValidationError(
+                "startFrame on instance does not match"
+                " with startFrame from the context data")
+
+        if frame_end != inst_frame_end:
+            raise PublishValidationError(
+                "endFrame on instance does not match"
+                " with endFrame from the context data")
```
BigRoy, line 55: This should actually report the values too so the user knows what to do.
Additionally I feel it's weird that it only reports the start frame differences if both start frame and end frame differ. I'd expect to get a report about both start and end frame if they are wrong, especially because repair would also be changing both.

openpype/hosts/max/plugins/publish/validate_frame_range.py
```diff
@@ -0,0 +1,60 @@
...
+
+    @classmethod
+    def repair(cls, instance):
+        rt.rendStart = instance.context.data.get("frameStart")
+        rt.rendEnd = instance.context.data.get("frameEnd")
```
BigRoy, line 60: Is it weird that this is an `InstancePlugin` even though apparently in repair it's setting (and getting?) the value from a global scene setting?

openpype/hosts/max/api/lib.py
```diff
@@ -150,10 +150,10 @@ def set_framerange(start_frame, end_frame):
     Todo:
         Current type is hard-coded, there should be a custom setting for this.
     """
-    rt.rendTimeType = 4
+    rt.rendTimeType = 3
     if start_frame is not None and end_frame is not None:
-        frame_range = "{0}-{1}".format(start_frame, end_frame)
-        rt.rendPickupFrames = frame_range
+        rt.rendStart = int(start_frame)
+        rt.rendEnd = int(end_frame)
```
BigRoy, line 156: Just going by the function's name and the docstring I wouldn't have expected this to only set the render frame range and not the animation playback frame range?

BigRoy (commented):
Some notes

openpype/hosts/max/api/lib.py
```diff
@@ -150,10 +150,10 @@ def set_framerange(start_frame, end_frame):
...
 
 
 def get_multipass_setting(project_setting=None):
```
BigRoy, line 159: Not related to this PR: This particular function appears to only be used in `submit_max_deadline`. Seems a bit off that this is in `lib`?

moonyuet: the original function is to set the frame range before the deadline submission. it is already there before the PR...

openpype/hosts/max/plugins/publish/validate_frame_range.py
moonyuet: this validator is set to be optional. If the user wants it to set the frame start and frame end from the global scene setting, they can activiate it.

BigRoy: I meant more that say if you have two instances, you deactivate it for one instance then technically the other instance not having it deactivated means you're still validating it for both instances since it's the same setting in the scene. As such, it's not really a setting per instance but a scene wide one, right?

openpype/hosts/max/api/lib.py
BigRoy: I understand it was already there - I'm just saying the the function name and its docstring are misleading. `set_render_frame_range` might make it more obvious.

BigRoy: I think your reply is regarding the other comment. This comment I made is about line 159 for function `get_multipass_setting`. It's definitely pre-existing code from outside of this PR like I mentioned. It just seems weird to be in the lib since it's just a dictionary lookup and used in once place only.

moonyuet: > I think your reply is regarding the other comment. This comment I made is about line 159 for function `get_multipass_setting`. It's definitely pre-existing code from outside of this PR like I mentioned. It just seems weird to be in the lib since it's just a dictionary lookup and used in once place only.

Yeah, sorry for the wrong spot!
yes it should be moved to somewhere else, but it is not quite related to this PR.

openpype/hosts/max/plugins/publish/validate_frame_range.py
BigRoy: There have been changes. I'd say however my comment here is still unresolved currently. In essence:

- Show both start / end frame mismatches instead of only reporting start frame mismatch if both differed.
- I'd say that it should include the current value and the expected value in the message too.

openpype/hosts/max/plugins/publish/validate_frame_range_type.py
```diff
@@ -0,0 +1,33 @@
...
+    order = ValidateContentsOrder
+    families = ["maxrender"]
+    hosts = ["max"]
+    actions = [RepairAction]
+
+    def process(self, instance):
+        if rt.rendTimeType != 3:
+            raise PublishValidationError("Incorrect type of frame range"
+                                         " used in render setting.."
+                                         "Repair action can help to fix it.")
```
BigRoy, line 28: Missing space between `setting..` and `Repair
```suggestion
                                         " used in render setting. "
                                         "Repair action can help to fix it.")
```

openpype/hosts/max/plugins/publish/validate_frame_range.py
```diff
@@ -0,0 +1,62 @@
...
+        frame_end = int(context.data.get("frameEnd"))
+
+        inst_frame_start = int(instance.data.get("frameStart"))
+        inst_frame_end = int(instance.data.get("frameEnd"))
+
+        if frame_start != inst_frame_start:
+            raise PublishValidationError(
+                "startFrame on instance does not match"
+                " with startFrame from the context data"
+                " You can use repair action to fix it")
```
BigRoy, line 51: Missing dot after `context data` and before ` You`.

```suggestion
            raise PublishValidationError(
                "startFrame on instance does not match"
                " with startFrame from the context data."
                " You can use repair action to fix it")
```
Same goes for the validation error below.

openpype/hosts/max/plugins/publish/validate_frame_range_type.py
```diff
@@ -0,0 +1,33 @@
...
+    order = ValidateContentsOrder
+    families = ["maxrender"]
+    hosts = ["max"]
+    actions = [RepairAction]
+
+    def process(self, instance):
+        if rt.rendTimeType != 3:
+            raise PublishValidationError("Incorrect type of frame range"
+                                         " used in render setting."
+                                         "Repair action can help to fix it.")
```
BigRoy, line 28: Still missing a space.

```suggestion
                                         " used in render setting."
                                         " Repair action can help to fix it.")
```
OR
```suggestion
                                         " used in render setting. "
                                         "Repair action can help to fix it.")
```

LiborBatek (dismissed):
Did try 2 times but even when recreated ```Render``` instance I cant locate any optional ```Frame range``` validator...is missing now

image

When trying to ```Publish``` I get frame range errors which is fine, I m just curious about these 2 validators ```Validate Frame Range``` and ```Validate Render Frame Range Type``` this one isnt im Maya and also have a bit weird naming.
image

moonyuet:
> When trying to `Publish` I get frame range errors which is fine, I m just curious about these 2 validators `Validate Frame Range` and `Validate Render Frame Range Type` this one isnt im Maya and also have a bit weird naming. image

The validator "Render Frame Range Type" only appears in 3dsmax as it has four scanerio to allow users to set up frame range.
But I can kill off this if it doesn't need here. 
image

LiborBatek:
Thats one thing, but the other is there is no any ```Optional Validator``` for frame range when ```Render``` Instance selected....also any in OP Settings for Max...

So I guess we miss something...

moonyuet:
In the updated PR you should see the vaidator along with other setting.
image
Also in the OP settings. I removed the frame type validator to ease the confusion
image

LiborBatek:
@moonyuet  @antirotor speaking of frame range, what comes to my mind...how we gonna treat those clients with needs of multiple but single frames acting as different shots? In terms of ```validator``` and ```render``` instances with different shot context.

Questions being:

1. Can we validate well, frame ranges and/or still frames via this PR?
2. How we deal with several frame ranges needed in single max scene using multiple ```render``` instances?

antirotor:
> 1. Can we validate well, frame ranges and/or still frames via this PR?

It is picking just the `Range` field (see image below)

> 2. How we validate multiple frame ranges needed in single max scene using several render instances acting in different shot context?

We don't, there is no way now to specify multiple frame ranges. It could be done using frame attribute per instance, be we are not there yet.

> 3. Speaking of Render instance is it capable of setting custom frame range at all?

Now you can set custom frame range in render settings and that will be used.

Nevertheless we will need to add support for *Every Nth Frame* and *Frames* - the first one is supported in Maya for example, but the second one isn't and we'll need to add some parsing there.

image

BigRoy:
I added deadline like frame parsing some time ago [here](https://github.com/Colorbleed/colorbleed-config/blob/acre/colorbleed/plugins/maya/publish/collect_renderlayer_custom_framelist.py#L16-L40), might be a good reference.

openpype/hosts/max/plugins/publish/validate_frame_range.py
```diff
@@ -0,0 +1,62 @@
...
+            raise PublishValidationError(
+                "startFrame on instance does not match"
+                " with startFrame from the context data."
+                " You can use repair action to fix it")
+
+        if frame_end != inst_frame_end:
+            raise PublishValidationError(
+                "endFrame on instance does not match"
+                " with endFrame from the context data."
+                " You can use repair action to fix it")
```
antirotor, line 57: I think the error messages should be more informative. What about this?

```suggestion
	errors = []
        if frame_start != inst_frame_start:
            errors.append(
                f"Start frame ({inst_frame_start}) on instance does not match "
                f"with the start frame ({frame_start}) set on the asset data. ")

        if frame_end != inst_frame_end:
            errors.append(
                f"End frame ({inst_frame_end}) on instance does not match "
                f"with the end frame ({frame_start}) from the asset data. ")
                
       if errors:
           errors.append("You can use repair action to fix it.")
           raise PublishValidationError("\n".join(errors))
```

antirotor (changes requested):
Just little comments.

antirotor (approved):

PR #4920, opened 2023-04-29
Feature/fusion loader plugins updates

EmberLightVFX:
## Changelog Description
Update to some Fusion loader plugins:
The sequence loader can now load footage from the image and online family.
The FBX loader can now import all formats Fusions FBX node can read.
You can now import the content of another workfile into your current comp with the workfile loader.

## Additional info
I would load to update the FBX and Alembic loader plugins. Currently they don't work the intended way to load geos. You should load them by the import menu instead of directly with a node.
I started a thread over at ynput's forum: https://community.ynput.io/t/fusion-fbx-abc-import-change/196 to see what the community thinks.

## Testing notes:
1. Import from the image and online family
1. Import an obj file
2. Import a published comp workfile

### DIFF
--- a/openpype/hosts/fusion/api/lib.py
+++ b/openpype/hosts/fusion/api/lib.py
@@ -309,6 +309,12 @@ def get_fusion_module():
     return fusion
 
 
+def get_bmd_library():
+    """Get bmd library"""
+    bmd = getattr(sys.modules["__main__"], "bmd", None)
+    return bmd
+
+
 def get_current_comp():
     """Get current comp in this session"""
     fusion = get_fusion_module()
--- a/openpype/hosts/fusion/plugins/load/load_fbx.py
+++ b/openpype/hosts/fusion/plugins/load/load_fbx.py
@@ -1,4 +1,3 @@
-
 from openpype.pipeline import (
     load,
     get_representation_path,
@@ -6,7 +5,7 @@ from openpype.pipeline import (
 from openpype.hosts.fusion.api import (
     imprint_container,
     get_current_comp,
-    comp_lock_and_undo_chunk
+    comp_lock_and_undo_chunk,
 )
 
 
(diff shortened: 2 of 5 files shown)


openpype/hosts/fusion/api/lib.py
```diff
@@ -309,6 +309,12 @@ def get_fusion_module():
     return fusion
 
 
+def get_bmd_library():
+    """Get bmd library"""
+    bmd = getattr(sys.modules["__main__"], "bmd", None)
```
Minkiu, line 314: Why not return it directly?

EmberLightVFX: I could but I kept the style of the other functions in there like `get_fusion_module` and `get_current_comp`
I could fix those also if you want?

openpype/hosts/fusion/plugins/load/load_fbx.py
```diff
@@ -15,7 +14,21 @@ class FusionLoadFBXMesh(load.LoaderPlugin):
...
+        "bvh",
+        "c3d",
+        "dae",
+        "dxf",
+        "fbx",
+        "htr",
+        "mcd",
+        "obj",
+        "trc",
+    }
```
BigRoy, line 31: Anyone else feels its weird that a `FusionLoadFBXMesh` loader now loads all these extensions? Does the `SurfaceFBXMesh` tool support all these extensions? 😮

openpype/hosts/fusion/api/lib.py
BigRoy: I'm fine with refactoring both, also fine with what it is now and moving 'cosmetics' to another PR to keep things simple. Feel free to mark resolved.

BigRoy (approved):
Just some minor cosmetic nodes - but functionally the code looks great. Nice work.

PR #4921, opened 2023-04-29
Fusion - Resolution and frame range validators

EmberLightVFX:
## Changelog Description
This PR adds two validators. One for each savers resolution and one for the comps frame range.

## Additional info
Both are optional as you might want/need to render out different resolutions and frame ranges.
I also made sure that when selecting problematic nodes in the publisher, the node-view now moves to that node.

## Testing notes:
1. Make a comp with wrong range range
2. Feed a saver footage with non-project resolution
3. The validator catches the problems.


openpype/hosts/fusion/plugins/publish/validate_frame_range.py
```diff
@@ -0,0 +1,79 @@
...
+
+        # Calcualte in/out points
+        range_start = start - handle_start
+        range_end = end - handle_start
+
+        invalid = self.get_invalid(context, range_start, range_end)
+        if invalid:
+            raise PublishValidationError(
+                "The comp's frame range isn't correct"
+                "compared to the projects properties."
```
BigRoy, line 50:
```suggestion
                "compared to the asset's properties."
```

Also should this report what the difference is? What are the current values and what are the asset's values?

openpype/hosts/fusion/plugins/publish/validate_frame_range.py
```diff
@@ -0,0 +1,79 @@
...
+                "\n\n{}".format("\n".join(invalid)),
+                title=self.label,
+            )
+
+    @classmethod
+    def get_invalid(cls, context, range_start, range_end):
+        invalid = []
+
+        if range_start != context.data["frameStartHandle"]:
```
BigRoy, line 59: Technically frameStartHandle could match but frameStart and handleStart could mismatch e.g.

frameStart 100 and handleStart 10
frameStart 90 and handleStart 0

Both result in the same range start but they aren't the same. I feel like this validator should more explicitly check both the handles and frame ranges.

openpype/hosts/fusion/plugins/publish/validate_saver_resolution.py
```diff
@@ -0,0 +1,96 @@
...
+        saver = instance[0]
+        firstFrame = saver.GetAttrs("TOOLNT_Region_Start")[1]
+        comp = get_current_comp()
+
+        # If the current saver hasn't bin rendered its input resolution
+        # hasn't bin saved. To combat this, add an expression in
+        # the comments field to read the resolution
+
+        # Create undo stack so we later can remove all changes
+        comp.StartUndo("Read resolution")
```
BigRoy, line 37: Probably better to use the undo chunk context manager from lib?

openpype/hosts/fusion/plugins/publish/validate_saver_resolution.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # Save old comment
+        oldComment = ""
+        hasExpression = False
+        if saver["Comments"][firstFrame] != "":
+            if saver["Comments"].GetExpression() is not None:
```
BigRoy, line 43: Isn't there a way to get the Image of a tool of which you can directly get its width and height?

I recall implementing a class for Image type in fusion in a library I wrote seven years ago here: https://github.com/Colorbleed/fusionless/blob/master/fusionless/core.py#L1671

I think you can get the value of the input link/parameter to get the image typed value for which you can then do `image.Width`

Which might be `saver.FindMainInput(1)` and then something. If that doesn't work you might need to get the output connected to that input to get its output value `.GetValue`.

I am not behind a PC for a week but I'm pretty sure this can be done through the Python API more directly as opposed to using expressions.

openpype/hosts/fusion/plugins/publish/validate_saver_resolution.py
```diff
@@ -0,0 +1,96 @@
...
+        if not self.is_active(instance.data):
+            return
+
+        wrong_resolution = []
+        invalid = self.get_invalid(instance, wrong_resolution)
+        if invalid:
+            entityData = instance.data["assetEntity"]["data"]
+            raise PublishValidationError(
+                "The input's resolution does not match"
+                " the project's resolution of {}x{}.\n\n"
```
BigRoy, line 89: It's not the project's resolution but the asset's resolution so please refactor project to asset where you mention it like that.

BigRoy (commented):
Some notes

openpype/hosts/fusion/plugins/publish/validate_frame_range.py
EmberLightVFX: Fixed.
It also do report the difference. You see "Inpoint is set to {} but should start at {}" at line 61 for example.
The way it's setup up you only get the needed info if only one of the in/out points are wrong.

openpype/hosts/fusion/plugins/publish/validate_frame_range.py
EmberLightVFX: The check is for the global start/end time and not the render start/end time as the render happens on the global range.
If the render range is different it doesn't matter as the render itself will use the global range

openpype/hosts/fusion/plugins/publish/validate_saver_resolution.py
EmberLightVFX: True. Updated the function so the undo chunk context manager's EndUndo accept a bool

openpype/hosts/fusion/plugins/publish/validate_saver_resolution.py
EmberLightVFX: From my tests I did to see what ways I could get the resolution, an expression was the only way I could find for a node that hasn't bin rendered yet.
If I read the nodes attributes the resolution variables doesn't exist until I render it.
If I go to the inputs/outputs and read its attributes I only get things like if the input is visible, what index it has and stuff not like. Nothing about the data passing through it.

As the nodes isn't rendered I think an expression is needed to "touch" the node-tree and get the correct resolution calculated. I don't think scripts does that.
I might be wrong but I didn't find any way to get what I needed.

openpype/hosts/fusion/plugins/publish/validate_saver_resolution.py
EmberLightVFX: Thanks for pointing it out!

EmberLightVFX:
A bunch of fixes and updates!
The biggest update is for the `validate frame range`. It checks the global in/out as expected. If it matches the assets range you're good to go and the render range will be fit to that range when the render starts.

But if you disable the validation Fusion will render the set render range.
One problem that comes after this is if you want to create a review file it will expect the assets full range and error out if not all files exist.
To solve this I have moved the `Collect expected frames` to the last step after all validations. I know it's wrong as it's not a validator but it was that or duplicate all its code inside of the `Validate frame range`. Any thoughts of that?

openpype/hosts/fusion/plugins/publish/validate_frame_range.py
BigRoy: I understand that it generates the same image frames, but the published instance will still have the frames and handles stored in the database which then still doesn't match the asset.

EmberLightVFX:
The latest commits fixes all problems mentioned above.

- When you're done rendering the render plugin now resets the render range to what it was originally, so the handles values will keep being correct.
- If you disable the check the handles will be set to 0 as the render now is disconnected from the asset's properties. 
- Burnins now act as expected!

The only "wrong" thing currently is that the validator also acts as a collector, lives in the collectors tab but is called validator as its job is still to validate. Thoughts about this?

EmberLightVFX:
Now when I could add a drop-down list to the collector the validator no longer need to also work as the collector!
So the user can now pick between:
```
"Asset's frame range" uses the frame range set to your opened asset.
"Fusion's current frame range" will render out the frame range you manually have selected.
The handles will both be set to 0 for this one.
```

The validator checks if the comp currently cover the savers expected frame range.

openpype/hosts/fusion/api/__init__.py
```diff
@@ -22,17 +22,15 @@
...
     "parse_container",
-
     # lib
     "maintained_selection",
     "update_frame_range",
     "set_asset_framerange",
     "get_current_comp",
     "comp_lock_and_undo_chunk",
-
+    "get_comp_render_range",
```
Minkiu, line 33: Let's keep the space in between for consistency's sake.

EmberLightVFX: Interesting... Black removes the new lines, didn't think about that. Will fix!

openpype/hosts/fusion/plugins/publish/collect_comp_frame_range.py
```diff
@@ -1,43 +1,98 @@
...
+    def get_attribute_defs(self):
+        frame_ranges = {
+            "asset_render": "Asset's frame range",
+            "freerange_render": "Fusion's current frame range",
+        }
+
+        return [
+            EnumDef(
+                "frame_range_type",
+                items=frame_ranges,
```
Minkiu, line 86: Maybe we could have the `dict` declaration in here directly, instead  of on line `78` ?

openpype/hosts/fusion/plugins/publish/collect_expected_frames.py
```diff
@@ -35,6 +35,8 @@ def process(self, instance):
             "stagingDir": output_dir,
         }
 
+        self.log.info(repre)
```
Minkiu, line 38: Should this be `debug` ?

openpype/hosts/fusion/plugins/publish/extract_render_local.py
```diff
@@ -71,7 +74,9 @@ def render_once(self, context):
 
         savers_to_render = [
             # Get the saver tool from the instance
-            instance[0] for instance in context if
+            instance[0]
```
Minkiu, line 77: I'd probably format it like this... find it quite odd to see a dangling `if` :
```py
savers_to_render = [
            instance[0]  # Get the saver tool from the instance
            for instance in context
            if (
                instance.data.get("publish", True)  # Only active instances
                and "render.local" in instance.data["families"]  # Only render.local instances
            )
]
```

openpype/hosts/fusion/plugins/publish/validate_frame_range.py
```diff
@@ -0,0 +1,101 @@
...
+            raise PublishValidationError(
+                message,
+                title=self.label,
+            )
+
+    @classmethod
+    def repair(cls, instance):
+        if instance.context.data["frame_range_type"] == "asset_render":
+            set_asset_framerange()
+        else:
```
Minkiu, line 100: I belive this:
```
else:
   pass
``` 
is redundant.

openpype/hosts/fusion/plugins/publish/validate_saver_resolution.py
```diff
@@ -0,0 +1,95 @@
...
+        saver = instance[0]
+        firstFrame = saver.GetAttrs("TOOLNT_Region_Start")[1]
+        comp = get_current_comp()
+
+        # If the current saver hasn't bin rendered its input resolution
+        # hasn't bin saved. To combat this, add an expression in
+        # the comments field to read the resolution
+
+        # False undo removes the undo-stack from the undo list
+        with comp_lock_and_undo_chunk(comp, "Read resolution", False):
```
Minkiu, line 40: I'm having a hard time understanding what is going on inside this `with` , probably cause i am not familiar with fusions API, but what confuses me is how in the same `saver["Comments"]` we get and set expressions.

EmberLightVFX: If the nodes isn't yet rendered (eg you just opened the comp) an expression is needed to "touch" the node-tree to get the calculated resolution.

`saver["Comments"].SetExpression("xxxx")` sets the expression to the comments controller. To read the result that the expression generates you need to do it at a frame (as the expressions result can change on different frames normally), thus `saver["Comments"][firstFrame]`

It's a bit of a hacky way but it's the only way I found with my extended research on how to get the resolution.

Minkiu: Thanks for explaining :D

Minkiu (approved):
Code wise, I'd say it looks good!

mkolar:
@BigRoy what's your take on this?

BigRoy:
> @BigRoy what's your take on this?

With recent changes to how frames are collected for rendering and now being able to set a frame range per instance (or at least tweak where it gets the frame range from) I suspect a lot of this PR is now obsolete or in a big conflict with the behavior.

I think the resolution validator is nice. We might need to just create a new PR with only related changes that still apply to current `develop`? What do you think @EmberLightVFX ?

EmberLightVFX:
Yeah I need to look over the frame-range validator. Does the new change add a frame-range validator?
I'm quite busy with a project currently but when I get some time over I'll jump back into this one

EmberLightVFX:
As the frame range validator isn't needed any longer I made a new PR that only contains the resolution validator here: https://github.com/ynput/OpenPype/pull/5325

PR #4923, opened 2023-04-30
Kitsu - Add "image", "online" and "plate" to review families

EmberLightVFX:
## Changelog Description
This PR adds "image", "online" and "plate" to the review families so they also can be uploaded to Kitsu.
It also adds the `Add review to Kitsu` tag to the default png review. Without it the user would manually need to add it for single image uploads to Kitsu and might confuse users (it confused me first for a while as movies did work).

## Testing notes:
1. Publish an image, video or image sequence as an image, online or plate and watch the review be uploaded to Kitsu

### DIFF
--- a/openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
+++ b/openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
@@ -9,7 +9,7 @@ class IntegrateKitsuNote(pyblish.api.ContextPlugin):
 
     order = pyblish.api.IntegratorOrder
     label = "Kitsu Note and Status"
-    families = ["render", "kitsu"]
+    families = ["render", "image", "online", "plate", "kitsu"]
 
     # status settings
     set_status_note = False
@@ -52,8 +52,9 @@ class IntegrateKitsuNote(pyblish.api.ContextPlugin):
         for instance in context:
             # Check if instance is a review by checking its family
             # Allow a match to primary family or any of families
-            families = set([instance.data["family"]] +
-                           instance.data.get("families", []))
+            families = set(
+                [instance.data["family"]] + instance.data.get("families", [])
+            )
             if "review" not in families:
                 continue
 
--- a/openpype/modules/kitsu/plugins/publish/integrate_kitsu_review.py
+++ b/openpype/modules/kitsu/plugins/publish/integrate_kitsu_review.py
@@ -8,11 +8,10 @@ class IntegrateKitsuReview(pyblish.api.InstancePlugin):
 
     order = pyblish.api.IntegratorOrder + 0.01
     label = "Kitsu Review"
-    families = ["render", "kitsu"]
+    families = ["render", "image", "online", "plate", "kitsu"]
     optional = True
 
     def process(self, instance):
-
         # Check comment has been created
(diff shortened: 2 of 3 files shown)


openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
```diff
@@ -52,8 +52,9 @@ def process(self, context):
         for instance in context:
             # Check if instance is a review by checking its family
             # Allow a match to primary family or any of families
-            families = set([instance.data["family"]] +
-                           instance.data.get("families", []))
+            families = set(
```
Minkiu, line 55: I feel like something like this is more readable:
```py
families = set(
    instance.data.get("families, []).append(instance.data["family"])
)
```

BigRoy: Not only is the old way more readable. The proposed way wouldn't even work because `append` doesn't return anything. Right?

If it however did work then you'd be editing the `families` list in place which is also NOT what we want. So I'd say the old logic is correct and perfectly readable to me.

If it's really currently unreadable to many I guess we could turn it into:
```python
families = set(instance.data.get("families", []))
```

But I don't see a direct reason to do so.

Tilix4 (approved):
I don't know these families and don't understand what they are supposed to do. But programmatically it looks fine!

Simply added one comment about the syntax you're discussing.

openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
BigRoy: Your example code doesn't work if `instance.data.get("families", {})` actually returns a list. It'd then be:

```python
set(["a"] | {"b"})
```

Resulting in:
```python
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for |: 'list' and 'set'
```

You'd still need to convert the first entry to a set since `families` is a list.

Tilix4: I'm sorry but that's exactly what it does :) `set` is only for "families"

Minkiu: Oh you are right, I assumed the append would be casted on the output of get, and that would be returned, I stand corrected.

EmberLightVFX:
> I don't know these families and don't understand what they are supposed to do. But programmatically it looks fine!
> 
@Tilix4 
They are from the Tray Publisher :)

openpype/modules/kitsu/plugins/publish/integrate_kitsu_note.py
BigRoy: > I'm sorry but that's exactly what it does :) set is only for "families"

I missed the closing `)` bracket for `set` there. Thanks for pointing it out! I'm not saying it's the most readable one but it should be functional.

iLLiCiTiT:
Looks ok for me. I would maybe consider to change all those families to `"review"`, or that would cause other issues?

iLLiCiTiT:
BTW @Tilix4 please approve if it's ok.

Tilix4 (approved):

iLLiCiTiT (approved):
LGTM

PR #4927, opened 2023-05-02
Add "shortcut" to Scripts Menu Definition

friquette:
## Changelog Description
Add the possibility to associate a shorcut for an entry in the script menu definition with the key "shortcut"

## Additional info
Example of an entry with a shortcut:
image

### DIFF
--- a/openpype/vendor/python/common/scriptsmenu/scriptsmenu.py
+++ b/openpype/vendor/python/common/scriptsmenu/scriptsmenu.py
@@ -19,9 +19,9 @@ class ScriptsMenu(QtWidgets.QMenu):
 
         Args:
             title (str): the name of the root menu which will be created
-        
+
             parent (QtWidgets.QObject) : the QObject to parent the menu to
-        
+
         Returns:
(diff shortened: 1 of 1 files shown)


BigRoy (commented):
Since Scriptsmenu is a vendorized package I feel like we should be adding this as a PR upstream so we stay in sync? Or have we already deviated to much from https://github.com/Colorbleed/scriptsmenu to begin with?

friquette:
> Since Scriptsmenu is a vendorized package I feel like we should be adding this as a PR upstream so we stay in sync? Or have we already deviated to much from https://github.com/Colorbleed/scriptsmenu to begin with?

I just compared what we have and the original repo, and it seems that this PR is the only difference between them, so maybe set this as a PR upstream is a good idea

iLLiCiTiT:
@BigRoy can we merge? Look ok for me.

iLLiCiTiT (approved):
LGTM

BigRoy (approved):

PR #4928, opened 2023-05-02
Drop-down menu to list all families in create placeholder

friquette:
## Changelog Description
Currently in the create placeholder window, we need to write the family manually. This replace the text field by an enum field with all families for the current software.

## Additional info
For Maya placeholder window:
image

For Nuke placeholder window:
image


iLLiCiTiT:
Since it is from "free text" to "explicit value", looking for families to Nuke creator plugins is limited as loaders can usually use other families that cannot be produced by host creators. So I would recommend to rather use `families` attribute from load plugins instead of create plugins...

openpype/pipeline/workfile/workfile_template_builder.py
```diff
@@ -1246,6 +1249,18 @@ def get_load_plugin_options(self, options=None):
...
+        # Get families and sort them alphabetically
+        all_loaders = discover_loader_plugins()
+        families = []
+        for loader in all_loaders:
+            families.extend(loader.families)
+
+        families = list(set(families))
+        if "*" in families:
+            families.remove("*")
+        families.sort()
```
BigRoy, line 1262: How about this?

```suggestion
		# Get combined families from all loaders excluding "*"
        families = set()
        for loader in all_loaders:
            families.update(loader.families)
		families.discard("*")
		
		# Sort for readability
		families = list(sorted(families))
```

openpype/pipeline/workfile/workfile_template_builder.py
```diff
@@ -43,8 +44,10 @@
     get_contexts_for_repre_docs,
     load_with_repre_context,
 )
+
 from openpype.pipeline.create import (
     discover_legacy_creator_plugins,
+    discover_creator_plugins,
```
BigRoy, line 50: This appears to be unused?
```suggestion
```

openpype/pipeline/workfile/workfile_template_builder.py
iLLiCiTiT: I like this more.

openpype/pipeline/workfile/workfile_template_builder.py
```diff
@@ -1246,6 +1249,18 @@ def get_load_plugin_options(self, options=None):
 
         loader_items = list(sorted(loader_items, key=lambda i: i["label"]))
         options = options or {}
+
+        # Get families and sort them alphabetically
+        all_loaders = discover_loader_plugins()
+        families = []
```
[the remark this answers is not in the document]
friquette: I can do it with a list comprehension, but honestly I don't know if it is really necessary. Are the 1 or 2 lines we gain here worth the lost in readability ?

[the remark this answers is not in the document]
iLLiCiTiT: I don't think this is needed with https://github.com/ynput/OpenPype/pull/4928#discussion_r1187602670

iLLiCiTiT:
Was possible to "not fill" the family input?

friquette:
> Was possible to "not fill" the family input?

Like an empty entry ? That maybe we can pass by default ?

iLLiCiTiT:
> Like an empty entry ? That maybe we can pass by default ?

It was meant as what happens if the family is empty? Would it work (somehow?). If yes, we should add empty string to enum, if it would crash then this is better.

friquette:
> > Like an empty entry ? That maybe we can pass by default ?
> 
> It was meant as what happens if the family is empty? Would it work (somehow?). If yes, we should add empty string to enum, if it would crash then this is better.

The family and representations are mandatory. If one is missing, the workfile builder will not find any asset

friquette:
> > Like an empty entry ? That maybe we can pass by default ?
> 
> It was meant as what happens if the family is empty? Would it work (somehow?). If yes, we should add empty string to enum, if it would crash then this is better.

The family and representations are mandatory. If one is missing, the workfile builder will not find any asset

openpype/pipeline/workfile/workfile_template_builder.py
```diff
@@ -24,6 +24,7 @@
     get_linked_assets,
     get_representations,
 )
+from openpype.pipeline.load.plugins import discover_loader_plugins
```
iLLiCiTiT, line 27:
```suggestion
```

iLLiCiTiT (approved):
Code looks ok. Can someone validate usability?

iLLiCiTiT (approved):
Code looks ok. Can someone validate usability?

LiborBatek (approved):
All works as expected in Maya Template Builder.
Instead of writting families manually rollout containing all families is present and working fine,

Did test to create several types of families and also ```Update Placeholder``` actions too. All good!

image

Minkiu (approved):

PR #4929, opened 2023-05-02
Maya: soft-fail when pan/zoom locked on camera when playblasting

antirotor:
## Changelog Description
When pan/zoom enabled attribute on camera is locked, playblasting with pan/zoom fails because it is trying to restore it. This is fixing it by skipping over with warning.

## Testing notes:
1. Create camera, lock pan/zoom attribute
2. Publish review from it
3. Publishing shouldn't fail


LiborBatek (approved):
All works as expected. Review with enabled Pan/Zoom and Locked option

image

tokejepsen (approved):

PR #4930, opened 2023-05-02
Feature/remove and load inv action

Sharkitty:
## Description
Added the ability to remove and load a container, as a way to reset it.
This can be useful in cases where a container breaks in a way that can be fixed by removing it, then reloading it.
Also added the ability to add `InventoryAction` plugins by placing them in `openpype/plugins/inventory`.

## Testing notes:
- Merge or cherry-pick this into main production branch.
- Open any asset with containers in the manage.
- Right click on the container that should be reset and select `Extra actions` -> `Remove and load`.


Sharkitty:
Can @Tilix4 be added to reviewers please?

openpype/plugins/inventory/remove_and_load.py
```diff
@@ -0,0 +1,43 @@
...
+            project_name = Session.get("AVALON_PROJECT")
+
+            # Get loader
+            loader_name = container["loader"]
+            for plugin in discover_loader_plugins(project_name=project_name):
+                if get_loader_identifier(plugin) == loader_name:
+                    loader = plugin
+                    break
+
+            assert (
```
BigRoy, line 29: Loader variable here might be undefined. In this case you might be better of doing a for..else statement.

So you could do like this psuedocode:
```
for plugin in plugins:
    if plugin == plugin_name:
        loader = plugin
        break
else:
    # Failed to get loader
```

openpype/plugins/inventory/remove_and_load.py
```diff
@@ -0,0 +1,43 @@
...
+
+class RemoveAndLoad(InventoryAction):
+    """Delete inventory item and reload it."""
+
+    label = "Remove and load"
+    icon = "refresh"
+
+    def process(self, containers):
+        for container in containers:
+            project_name = Session.get("AVALON_PROJECT")
```
iLLiCiTiT, line 20:
```suggestion
        project_name = get_current_project_name()
        for container in containers:
```

openpype/plugins/inventory/remove_and_load.py
```diff
@@ -0,0 +1,43 @@
+from openpype.pipeline import InventoryAction
+from openpype.pipeline.legacy_io import Session
```
iLLiCiTiT, line 2:
```suggestion
from openpype.pipeline import get_current_project_name
```

openpype/plugins/inventory/remove_and_load.py
```diff
@@ -0,0 +1,43 @@
...
+    label = "Remove and load"
+    icon = "refresh"
+
+    def process(self, containers):
+        for container in containers:
+            project_name = Session.get("AVALON_PROJECT")
+
+            # Get loader
+            loader_name = container["loader"]
+            for plugin in discover_loader_plugins(project_name=project_name):
```
iLLiCiTiT, line 24: I would recommend to discover load plugins only once before the loop.

openpype/plugins/inventory/remove_and_load.py
```diff
@@ -0,0 +1,43 @@
...
+            else:
+                raise RuntimeError(
+                    "Failed to get loader, can't remove and load container"
+                )
+
+            # Get representation
+            representation = get_representation_by_id(
+                project_name, container["representation"]
+            )
+            assert representation, "Representation not found"
```
iLLiCiTiT, line 37: Not sure if it should raise an exception?

openpype/plugins/inventory/remove_and_load.py
Sharkitty: May I ask for the reason of this suggestion? Perhaps `legacy_io` is deprecated? I'm fine with these changes, I am just curious because if what I did here should be avoided, then I need to change my practices.

openpype/plugins/inventory/remove_and_load.py
Sharkitty: I think this inventory action should not proceed if `representation` is `None`, because it would manage to remove the container, then fail to reload it. This is why I chose to raise an exception. Eventually I could simply print and return but I'm not too sure if that would be better.

openpype/plugins/inventory/remove_and_load.py
iLLiCiTiT: `legacy_io` is not yet deprecated, we want to make it deprecated, we can't until all usages are removed and replaced. Most of changes in core from last 2-3 releases reduced it's usage to minimum.

openpype/plugins/inventory/remove_and_load.py
iLLiCiTiT: Just continue before it's removed?

openpype/plugins/inventory/remove_and_load.py
Sharkitty: Okay, it makes sense. I'll try to reduce my use of `legacy_io` from now on.

openpype/plugins/inventory/remove_and_load.py
Minkiu: I believe he means to do the assert after the `remove_container`

Sharkitty: That would be problematic because if the load cannot be performed then we shouldn't remove in my opinion. That's why I chose to assert before remove.

BigRoy: I believe @iLLiCiTiT just means to not raise an assertion directly but to continue with the remainder of the containers and skipping the invalid ones with e.g. a `self.log.warning()` message as to why it's skipped that particular one.

So it would be like:
```suggestion
            if not representation:
                self.log.warning(
                    "Skipping remove and load because representation"
                    " id is not found in database: '{}'".format(container["representation"])
```

Sharkitty: Oh! I think I understand now, thanks you :smile:

Tilix4 (approved):
:ok: for me

openpype/plugins/inventory/remove_and_load.py
```diff
@@ -0,0 +1,50 @@
...
+
+    label = "Remove and load"
+    icon = "refresh"
+
+    def process(self, containers):
+        project_name = get_current_project_name()
+        for container in containers:
+            # Get loader
+            loader_name = container["loader"]
+            loaders = discover_loader_plugins(project_name=project_name)
```
iLLiCiTiT, line 23:
```suggestion
        project_name = get_current_project_name()
        loaders = discover_loader_plugins(project_name=project_name)
        for container in containers:
            # Get loader
            loader_name = container["loader"]
```

BigRoy: I'd even take this a step further and precache the by loader name identifier too

```suggestion
        project_name = get_current_project_name()
        loaders = discover_loader_plugins(project_name=project_name)
        loaders_by_name = {get_loader_identifier(plugin) for plugin in loaders}
        for container in containers:
            # Get loader
            loader_name = container["loader"]
            loader = loaders_by_name.get(loader_name, None)
            if not loader:
                raise RuntimeError(
                    "Failed to get loader '{}', can't remove "
                    "and load container".format(loader_name)
                )
```

Also added the loader name that was not found in the error message for clarity.

BigRoy: I'd even take this a step further and precache the by loader name identifier too

```suggestion
        project_name = get_current_project_name()
        loaders = discover_loader_plugins(project_name=project_name)
        loaders_by_name = {get_loader_identifier(plugin) for plugin in loaders}
        for container in containers:
            # Get loader
            loader_name = container["loader"]
            loader = loaders_by_name.get(loader_name, None)
            if not loader:
                raise RuntimeError(
                    "Failed to get loader '{}', can't remove "
                    "and load container".format(loader_name)
                )
```

Also added the loader name that was not found in the error message for clarity.

Sharkitty: @BigRoy I don't think your suggestion would work because we would only get a loader indentifier that way, but not the loader itself. `set` doesn't have a `get` method either but that's no big deal.

BigRoy: Sorry, @Sharkitty - typo. Fixed my suggestion. It was supposed to be a dict.

Sharkitty: Ah! Makes a lot more sense now ^^ Thanks you

iLLiCiTiT (approved):
LGTM

BigRoy (approved):
Did not test the plug-in, but code looks good to me.

Only thing I'm wondering if we'd want to play with ordering or coloring the action in such a way that it might not take precedence over other actions we might consider more important. Like should this have `order = 99` set maybe?

antirotor (approved):

Sharkitty:
> Did not test the plug-in, but code looks good to me.
> 
> Only thing I'm wondering if we'd want to play with ordering or coloring the action in such a way that it might not take precedence over other actions we might consider more important. Like should this have `order = 99` set maybe?

That's a good idea. The color is the default, so the icon in the mange window appears as an orange cogwheel, which is consistent with the rest of the UI and will follow any change to the default value. Do you think it needs a specific color or does that seem fine to you?

antirotor:
It looks like this now:

image

I don't have problem with it.

BigRoy:
Agree. Looks fine for now!

PR #4932, opened 2023-05-03
Fusion - Some small bugfixes/updates

EmberLightVFX:
## Changelog Description
This PR solves 3 small (but irritating) bugs in the Fusion integration. Instead of making 3 PRs I made this one as they are very small fixes.

1. The pyside2 installation now works for all Fusion versions from 16 and up (before it could only installed for Fusion 16).
2. _update_tool_with_data now generates the correct workdir if the user have changed the asset type in Publisher
3. If footage doesn't have Frames set (eg if you add from TrayPublisher) the footage now starts at Fusions global start instead of frame 0 (This is the way Fusion would do it)

## Testing notes:
1. Try install pyside2 from the script
2. Open Publisher and change the asset type of a saver
3. Import footage that doesn't have Frames assigned to it


openpype/hosts/fusion/plugins/create/create_saver.py
```diff
@@ -119,23 +117,26 @@ def _imprint(self, tool, data):
...
-        if "subset" not in data:
+
+        if data.get("subset") is None:
+            self.log.warning("No subset found for _update_tool_with_data")
             return
 
-        original_subset = tool.GetData("openpype.subset")
+        # Get current session but change the asset to the one from
+        # data in case a new asset has bin selected
+        session = legacy_io.Session
```
iLLiCiTiT, line 127: Please don't change `legacy_io.Session` for other purposes than real change of context. Also we're trying to remove `legacy_io` and `Session`. Rather use other ways how to get workdir path please.

iLLiCiTiT: Also this is producing a lot of calls to databases on save. I can imagine with 10 and more instances it could take tens of seconds.

openpype/hosts/fusion/plugins/create/create_saver.py
```diff
@@ -119,23 +121,27 @@ def _imprint(self, tool, data):
 
     def _update_tool_with_data(self, tool, data):
         """Update tool node name and output path based on subset data"""
-        if "subset" not in data:
-            return
 
-        original_subset = tool.GetData("openpype.subset")
+        if data.get("subset") is None:
```
Minkiu, line 125: Maybe we could do the variable assignation before the if, so we only access the dict ones:
```py
subset = subset = data.get("subset")

if subset is None:
    self.log.warning("No subset found for _update_tool_with_data")
    return
```

openpype/hosts/fusion/deploy/MenuScripts/install_pyside2.py
```diff
@@ -13,14 +25,17 @@
     print(f"Qt path: {mod.__file__}")
     print("Qt library found, nothing to do..")
 
-except ImportError:
+except Exception:
+    python_path = get_python_path()
+    if python_path is None:
+        print("Can't find any any python environment.")
```
BigRoy, line 31: Can't we just rely on `sys.executable` at this stage since we're already running from a Python process?

BigRoy: Had a quick look:
```python
import sys
print(sys.executable)
# C:\Program Files\Blackmagic Design\Fusion 18\Fusion.exe
```
So I guess that doesn't work.

However,
```
Py3> print(sys.base_prefix)
C:\Users\User\AppData\Local\Programs\Python\Python39
Py3> print(sys.base_exec_prefix)
C:\Users\User\AppData\Local\Programs\Python\Python39
Py3> print(sys.exec_prefix)
C:\Users\User\AppData\Local\Programs\Python\Python39
Py3> print(sys.prefix)
C:\Users\User\AppData\Local\Programs\Python\Python39
```
See also relevant documentation entries:
- https://docs.python.org/3/library/sys.html#sys.base_prefix
- https://docs.python.org/3/library/sys.html#sys.base_exec_prefix
- https://docs.python.org/3/library/sys.html#sys.exec_prefix
- https://docs.python.org/3/library/sys.html#sys.prefix

Could we rely on any of those?

BigRoy (approved):
Nice changes. Looks good to me!

openpype/hosts/fusion/deploy/MenuScripts/install_pyside2.py
```diff
@@ -13,14 +14,17 @@
     print(f"Qt path: {mod.__file__}")
     print("Qt library found, nothing to do..")
 
-except ImportError:
+except Exception:
+    python_path = sys.base_prefix
+    if python_path is None:
+        print("Can't find any any python environment.")
```
antirotor, line 20: Little typo here:

```suggestion
        print("Can't find any python environment.")
```

openpype/hosts/fusion/plugins/create/create_saver.py
```diff
@@ -119,23 +121,27 @@ def _imprint(self, tool, data):
...
-            filepath = os.path.join(workdir, "render", subset, filename)
-            tool["Clip"] = filepath
-
-            # Rename tool
-            if tool.Name != subset:
-                print(f"Renaming {tool.Name} -> {subset}")
-                tool.SetAttrs({"TOOLS_Name": subset})
+        if subset is None:
+            self.log.warning("No subset found for _update_tool_with_data")
+            return
```
antirotor, line 128: I think **subset** cannot be `None` in this case and if it is, it _must_ be bug somewhere.  In that case it should probably raise exception or at least log error. But it might be something Fusion integration specific?

antirotor (changes requested):
Found one typo and have one question, otherwise it looks good.

openpype/hosts/fusion/plugins/create/create_saver.py
iLLiCiTiT: Same question, how it can be `None`?

openpype/hosts/fusion/plugins/create/create_saver.py
```diff
@@ -119,23 +121,27 @@ def _imprint(self, tool, data):
...
+
+        context = get_current_context()
```
iLLiCiTiT, line 130: Why are you actually using current context when the instance may be in totally different context?
```
project_name = self.create_context.get_current_project_name()
asset_name = data["asset"]
task_name = data["task"]
...
```

openpype/hosts/fusion/plugins/create/create_saver.py
```diff
@@ -147,9 +153,10 @@ def _collect_unmanaged_saver(self, tool):
         comp = tool.Comp()
 
         # Allow regular non-managed savers to also be picked up
-        project = legacy_io.Session["AVALON_PROJECT"]
-        asset = legacy_io.Session["AVALON_ASSET"]
-        task = legacy_io.Session["AVALON_TASK"]
+        context = get_current_context()
```
iLLiCiTiT, line 156: Similar issue as here https://github.com/ynput/OpenPype/pull/4932/files#r1192409673 . Asset and task name should be probably taken from the tool attributes instead of from current context?

openpype/hosts/fusion/plugins/create/create_saver.py
```diff
@@ -137,12 +142,19 @@ def _imprint(self, tool, data):
...
-        if original_subset != subset:
+        original_data = tool.GetData("openpype")
+
+        # As data contains an extra key called active, add it to
+        # original_data to get a correct comparison
+        if data.get("active"):
+            original_data["active"] = data["active"]
+
+        if original_data != data:
             self._configure_saver_tool(data, tool, subset)
```
BigRoy, line 158: I think reconfiguring on "any change of data" isnt' the best way forward. That'd mean that technically if just toggle "Render local" to "render on farm" that it'd trigger the reconfigure logic which could technically change the output path, etc. even if for _whatever reason_ the user had it customized. I feel like we should be explicit about what data changed should result in a the output path to be reconfigured, like  a task change or subset change. Right?

movalex:
Hey, I just realized you also worked on the Install PySide script fix. Great work!

Please consider this additions: 

1. I added `install_pip_package`, so you can use it to install other packages too.
2. The script now finds macos python executable correctly.
3. macos requires pyside6, so this is also accounted.
4. Actual `python home` variables for all supported Fusion versions are stored in `openpype.hosts.fusion`. 
5. Not a big fan of wide exceptions like [this](https://github.com/movalex/OpenPype/blob/ecb4aa2574c2399219c90f139ab9081522060c59/openpype/hosts/fusion/deploy/MenuScripts/install_pyside2.py#L27), but I also did not find a better solution. Any ideas?

https://github.com/ynput/OpenPype/compare/develop...movalex:OpenPype:enhancement/fix-fusion-pyside-installation-script

openpype/hosts/fusion/plugins/create/create_saver.py
```diff
@@ -137,12 +142,19 @@ def _imprint(self, tool, data):
...
             return
 
-        original_subset = tool.GetData("openpype.subset")
-        subset = data["subset"]
-        if original_subset != subset:
+        original_data = tool.GetData("openpype")
+
+        # As data contains an extra key called active, add it to
+        # original_data to get a correct comparison
+        if data.get("active"):
```
iLLiCiTiT, line 154: Shouldn't this be like this? If `"active"` would be `False` it would not change it. (maybe it's desired, but from the comment it is not)
```suggestion
        if "active" in data:
```

EmberLightVFX:
I feel this PR is ready to be merge. One last review from @BigRoy maybe?

Minkiu:
Friendly ping for @BigRoy :D

openpype/hosts/fusion/deploy/MenuScripts/install_pip_package.py
```diff
@@ -0,0 +1,56 @@
...
+from pathlib import Path
+
+
+def get_python_home() -> Path:
+    python_home = Path(os.__file__).parent.parent
+    return python_home
+
+
+def get_python_executable(python_home):
+    if platform.system() == "Windows":
```
Minkiu, line 13: Wouldn't `shutil.which("python")` achieve the same?

Minkiu (approved):
Haven't tested it, but code looks good!

openpype/hosts/fusion/deploy/MenuScripts/install_pip_package.py
BigRoy: `shutil.which` would likely search `PATH` env var which isn't necessarily the case - also, as can be seen from the code we're actually looking for `python` on Windows and `python3` on other platforms, as such a single `which("python")` would not work, right?

Minkiu: I think most platforms do a symlink of `python` -> `python3`; but after testing it does pick up whatever is the first in `PATH` so it's not the same.

openpype/hosts/fusion/plugins/create/create_saver.py
BigRoy: This note is still relevant I believe @EmberLightVFX ?

openpype/hosts/fusion/plugins/create/create_saver.py
BigRoy: This note should still be applied?

openpype/hosts/fusion/deploy/MenuScripts/install_pip_package.py
BigRoy, line 1: Is there a reason this file needs to be separate like a library instead of being functions inside `install_pyside2`? It's only used there - right? Feels a bit odd to expose this 'library' inside the MenuScripts folder.

openpype/hosts/fusion/deploy/MenuScripts/install_pyside.py
```diff
@@ -0,0 +1,33 @@
...
+from install_pip_package import pip_install
+from openpype.hosts.fusion import FUSION_VERSIONS_DICT
+
+
+def get_pyside_version() -> str:
+    os_platform = platform.system()
+    pyside_version = "PySide2"
+    if os_platform == "Darwin":
+        pyside_version = "PySide6"
+    return pyside_version
```
BigRoy, line 15: Could this be PySide6 for all platforms by now? :)

openpype/hosts/fusion/deploy/MenuScripts/install_pip_package.py
BigRoy: I see - I think it's just a bit weird that it's in `MenuScripts` to begin with - if it's supposed to be used elsewhere I'd expect it in `lib` or alike. In this case - let's just still keep it inside `install_pyside` UNTIL we have another script that'll use it?

openpype/hosts/fusion/deploy/MenuScripts/install_pyside.py
```diff
@@ -0,0 +1,34 @@
...
+try:
+    from qtpy import API_NAME
+
+    print(f"Qt binding: {API_NAME}")
+    mod = importlib.import_module(API_NAME)
+    print(f"Qt path: {mod.__file__}")
+    print("Qt library found, nothing to do..")
+
+except Exception:
+    fusion_version = int(app.Version)
```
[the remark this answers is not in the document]
EmberLightVFX: Good idea! I'll add it

openpype/hosts/fusion/plugins/create/create_saver.py
BigRoy: Hmm - it's a matter of preference I suppose. I know for a fact that a lot of our artists here have been changing the paths a LOT since what they change it too the publisher doesn't really care as long as it can find the output files. :)

openpype/hosts/fusion/deploy/MenuScripts/install_pip_package.py
EmberLightVFX: Sure thing!

openpype/hosts/fusion/plugins/create/create_saver.py
EmberLightVFX: Should it maybe be a setting then? To allow the publisher to re-generate the save path or not.
I rather not to not end up with 1000 different settings but if it's a feature that's important for yours and maybe other studios I guess it could be a setting.

BigRoy: I don't mind it a crazy amount for it to be a dedicated feature. However, I personally would find it odd that if I just deactivated the instance for example that it'd suddenly start setting the path again. To me: asset, variant or subset changes it makes sense and on anything that tweaks the output. But e.g. disabling an optional validator or alike - then let's avoid touching it if we can would for me be the best solution.

@jakubjezek001 @antirotor what's the thought on this?

mkolar:
Because we're splitting OpenPype into ayon-core and individual host addons, this PR would have to be re-created to target one of those. It's also gotten a bit stale. 

We're closing it down, but we'll he happy for a new PR to [ynput/ayon-fusion](https://github.com/ynput/ayon-fusion) repository once it's up.

PR #4933, opened 2023-05-04
Feature/blender workfile out of date handling

Sharkitty:
## Changelog Description
Adding workfile out of date handling to Blender host. This feature is important to avoid users to work on an out of date version of their workfile without knowing it, then publishing. Which would imply that newly published subsets don't contain the latest changes their coworkers have made.

This comes with workfile up to date checks:
- On load (including on launch).
- Periodically.
- With a check workfile up to date button in the OpenPype menu.

If the open workfile is the latest, a popup will be displayed to notify the user that their workfile is up to date (which is discarded as soon as the cursor is moved away from the popup. On launch, you may not see it if you are multitasking).

If the open workfile is out of date, a popup will be displayed to notify the user about it, and propose them 3 actions:
- Download the latest workfile (and open it).
- Quit blender.
- Proceed anyway (knowingly using an out of date workfile).
This popup is made to not reappear if cancelled, so users can't ignore it, as this is important and should not be missed.

A warning sign will appear in the OpenPype menu with the indication that your workfile is out of date, so you'll always have a reminder if you are using an out of date workfile. The warning is clickable and has the same behavior as the check workfile up to date button (except it only appears when the workfile is already marked as out of date).

During publish, a validator checks if the workfile is up to date, and fails if it isn't. This validator is optional, enabled by default. If at any point a rollback is necessary (or if the workfile out of date handling bugs and wrongly marks a workfile as out of date), this validator can be ticked off, allowing to publish an out of date workfile on purpose. I see this manipulation as a way to confirm that "I know what I'm doing, I want to proceed with the publish".

## Additional info
This system stores last publish date in the workfile, so on a workfile that has been created without using this system, their may be issues with workfiles wrongly marked as out of date, which can be resolved by publishing once (since publish time is set in the workfile during integration). This implies that deploying this feature requires a conformation, but that can reasonably be done on the go as assets are being updated to meet production needs.

A publish failure during integration phase may cause the workfile to be marked as out of date. There are mitigations for this issue in place, but if it still happens it is safe to tick off the workfile up to date validator.

## Testing notes:
As explained in additional info, if this is tested on an existing asset, a publish may be necessary to have the expected behavior (so the workfile is conformed for this feature).
- Open a locally available workfile with blender host.
- If it's up to date, either nothing or the workfile up to date should appear.
- If it's out of date, the workfile out of date popup should appear.
- The button in the OpenPype menu should have the same behaviour.

If the workfile out of date popup appears:
- Clicking away from the popup should make the popup reappear under the pointer.
- Selecting quit blender should quit blender.
- Selecting proceed anyway should close the popup. The warning should be visible in the openpype menu, the check workfile up to date button in the openpype menu should reopen the workfile out of date popup.
- Workfile up to date validator should fail.
- Selecting download last workfile should succesfully download and load last workfile.

If the workfile is up to date it should be seamless to users, no warning and the validator should pass.
After performing a rollback by ticking off the workfile up to date validator, the workfile should be considered up to date at the end of the publish.

Testing tip: in the workfiles menu, copy and open an older workfile if you need an out of date workfile for testing.


Sharkitty:
Can @Tilix4 be added to reviewers please?

moonyuet:
I try to test it but I dont see the related validator and extractor show up in the publisher. 
image

moonyuet:
> > I try to test it but I dont see the related validator and extractor show up in the publisher. image
> 
> Hello! Is there any error message? And does the `Check workfile up to date` button appear in the OpenPype menu? I'm trying to find how this could happen, but I haven't found yet how to reproduce your issue.

Hi, nothing appears in my scene, I checked OP menu and OP launcher too. but it doesn't show.
image
image

Sharkitty:
> > > I try to test it but I dont see the related validator and extractor show up in the publisher. image
> > 
> > 
> > Hello! Is there any error message? And does the `Check workfile up to date` button appear in the OpenPype menu? I'm trying to find how this could happen, but I haven't found yet how to reproduce your issue.
> 
> Hi, nothing appears in my scene, I checked OP menu and OP launcher too. but it doesn't show. image image

Can you provide logs so I can have more information about the cause of your issue please?

moonyuet:
[blender_log.txt](https://github.com/ynput/OpenPype/files/11462851/blender_log.txt)

Sharkitty:
@moonyuet Can you try running `from openpype.hosts.blender.plugins.publish import validate_workfile_up_to_date` in blender's python console and tell me if it puts outputs an error please?

moonyuet:
> from openpype.hosts.blender.plugins.publish import validate_workfile_up_to_date

it shows the errror like this:
  from openpype.hosts.blender.plugins.publish import validate_workfile_up_to_date
    File "C:\Program Files\Blender Foundation\Blender 3.2\3.2\python\lib\code.py", line 90, in runcode
      exec(code, self.locals)
    File "<blender_console>", line 1, in <module>
  ImportError: cannot import name 'validate_workfile_up_to_date' from 'openpype.hosts.blender.plugins.publish' (unknown location)

Sharkitty:
> > from openpype.hosts.blender.plugins.publish import validate_workfile_up_to_date
> 
> it shows the errror like this: from openpype.hosts.blender.plugins.publish import validate_workfile_up_to_date Traceback (most recent call last): File "C:\Program Files\Blender Foundation\Blender 3.2\3.2\python\lib\code.py", line 90, in runcode exec(code, self.locals) File "<blender_console>", line 1, in ImportError: cannot import name 'validate_workfile_up_to_date' from 'openpype.hosts.blender.plugins.publish' (unknown location)

It's odd... I've never had any import issue with this file. Any chance you could try a blender 3.3LTS? I don't think blender version is the issue but I'd like to rule that out, as I developed and tested this feature with blender 3.3. You can use one of the `.zip` releases if you need something portable so it doesn't affect your installation.

Tilix4:
> > from openpype.hosts.blender.plugins.publish import validate_workfile_up_to_date
> 
> it shows the errror like this: from openpype.hosts.blender.plugins.publish import validate_workfile_up_to_date Traceback (most recent call last): File "C:\Program Files\Blender Foundation\Blender 3.2\3.2\python\lib\code.py", line 90, in runcode exec(code, self.locals) File "<blender_console>", line 1, in ImportError: cannot import name 'validate_workfile_up_to_date' from 'openpype.hosts.blender.plugins.publish' (unknown location)

@moonyuet could you please check `openpype/hosts/blender/plugins/publish/validate_workfile_up_to_date.py` actually exists?

openpype/hosts/blender/utility_scripts/__init__.py
Tilix4, line 1: I'm not sure this file is necessary.

Tilix4:
> > > > from openpype.hosts.blender.plugins.publish import validate_workfile_up_to_date
> > > 
> > > 
> > > it shows the errror like this: from openpype.hosts.blender.plugins.publish import validate_workfile_up_to_date Traceback (most recent call last): File "C:\Program Files\Blender Foundation\Blender 3.2\3.2\python\lib\code.py", line 90, in runcode exec(code, self.locals) File "<blender_console>", line 1, in ImportError: cannot import name 'validate_workfile_up_to_date' from 'openpype.hosts.blender.plugins.publish' (unknown location)
> > 
> > 
> > @moonyuet could you please check `openpype/hosts/blender/plugins/publish/validate_workfile_up_to_date.py` actually exists?
> 
> Yes. It exists in the openpype/plugins folder in the build version of openpype_console. image I installed blender 3.3 LTS and it seems that blender's python console still shows the same error. PYTHON INTERACTIVE CONSOLE 3.10.9 (main, Jan 10 2023, 11:01:25) [MSC v.1928 64 bit (AMD64)]
> 
> Builtin Modules: bpy, bpy.data, bpy.ops, bpy.props, bpy.types, bpy.context, bpy.utils, bgl, blf, mathutils Convenience Imports: from mathutils import *; from math import * Convenience Variables: C = bpy.context, D = bpy.data
> 
> from openpype.hosts.blender.plugins.publish import validate_workfile_up_to_date Traceback (most recent call last): File "C:\Program Files\Blender Foundation\Blender 3.3\3.3\python\lib\code.py", line 63, in runsource code = self.compile(source, filename, symbol) File "C:\Program Files\Blender Foundation\Blender 3.3\3.3\python\lib\codeop.py", line 153, in **call** return _maybe_compile(self.compiler, source, filename, symbol) File "C:\Program Files\Blender Foundation\Blender 3.3\3.3\python\lib\codeop.py", line 73, in _maybe_compile return compiler(source, filename, symbol) File "C:\Program Files\Blender Foundation\Blender 3.3\3.3\python\lib\codeop.py", line 118, in **call** codeob = compile(source, filename, symbol, self.flags, True) File "<blender_console>", line 1 from openpype.hosts.blender.plugins.publish import validate_workfile_up_to_date IndentationError: unexpected indent

This pretty much looks like a plugin loading issue more than a code one. Don't you have any setting error message at OP's initialization?

Sharkitty:
> > > > > > from openpype.hosts.blender.plugins.publish import validate_workfile_up_to_date
> > > > > 
> > > > > 
> > > > > it shows the errror like this: from openpype.hosts.blender.plugins.publish import validate_workfile_up_to_date Traceback (most recent call last): File "C:\Program Files\Blender Foundation\Blender 3.2\3.2\python\lib\code.py", line 90, in runcode exec(code, self.locals) File "<blender_console>", line 1, in ImportError: cannot import name 'validate_workfile_up_to_date' from 'openpype.hosts.blender.plugins.publish' (unknown location)
> > > > 
> > > > 
> > > > @moonyuet could you please check `openpype/hosts/blender/plugins/publish/validate_workfile_up_to_date.py` actually exists?
> > > 
> > > 
> > > Yes. It exists in the openpype/plugins folder in the build version of openpype_console. image I installed blender 3.3 LTS and it seems that blender's python console still shows the same error. PYTHON INTERACTIVE CONSOLE 3.10.9 (main, Jan 10 2023, 11:01:25) [MSC v.1928 64 bit (AMD64)]
> > > Builtin Modules: bpy, bpy.data, bpy.ops, bpy.props, bpy.types, bpy.context, bpy.utils, bgl, blf, mathutils Convenience Imports: from mathutils import *; from math import * Convenience Variables: C = bpy.context, D = bpy.data
> > > from openpype.hosts.blender.plugins.publish import validate_workfile_up_to_date Traceback (most recent call last): File "C:\Program Files\Blender Foundation\Blender 3.3\3.3\python\lib\code.py", line 63, in runsource code = self.compile(source, filename, symbol) File "C:\Program Files\Blender Foundation\Blender 3.3\3.3\python\lib\codeop.py", line 153, in **call** return _maybe_compile(self.compiler, source, filename, symbol) File "C:\Program Files\Blender Foundation\Blender 3.3\3.3\python\lib\codeop.py", line 73, in _maybe_compile return compiler(source, filename, symbol) File "C:\Program Files\Blender Foundation\Blender 3.3\3.3\python\lib\codeop.py", line 118, in **call** codeob = compile(source, filename, symbol, self.flags, True) File "<blender_console>", line 1 from openpype.hosts.blender.plugins.publish import validate_workfile_up_to_date IndentationError: unexpected indent
> > 
> > 
> > This pretty much looks like a plugin loading issue more than a code one. Don't you have any setting error message at OP's initialization?
> 
> No. If there is any setting error, you can't edit any project setting in Openpype setting. I can see and edit the setting. image
> 
> One thing I find is I tried to add `UpdateCurrentWorkfileLastPublishTime `to `blender.json` and `schema_blender_publish.json` for test(it's not in a right place!) , the integrator shows up in the publisher. Maybe the validator and the integrator dont show up mainly because they are not included in the settings in this PR? image
> 
> image

You could try to add settings for the validator as well, to confirm if it can be imported on your end. Then we should find why it's not loaded when there is no setting. I really wonder where the difference lies between environments we've been using this feature on, and your environment, if it's a setting or something else entirely :thinking:

openpype/hosts/blender/utility_scripts/__init__.py
Sharkitty: You're right, my bad.

Sharkitty:
> image image I have tried to add the validator to the setting, it shows up as expected after adding as a part of the setting. I guess it's a matter of the OP setting version issue. When I used the OP version ahead of yours, the validator and the integrator wont show up correctly as expected. When I tried to copy your script with the new branch created with the latest develop, the validator and the integrator show up in the blender's publisher. I guess this can be resolved by updating branch with the latest develop. (Not sure, but will see)

I updated the branch if you want to try again :smile:

moonyuet:
I have done simple test the code but not sure if I did something wrong. It seems that the button "check workfile up to date" doesn't really check the workfile version?
The video below shows I have used different versions of workfile(v1 and v4 workfile respectivaly). But the button can't identify which one is the latest version. 
Same thing happened to the validator of checking workfile up to date, it will still passes if you use elder version files. 
image

https://github.com/ynput/OpenPype/assets/64118225/34c73e7a-bfe3-440b-b854-777acac26f9c

Sharkitty:
> I have done simple test the code but not sure if I did something wrong. It seems that the button "check workfile up to date" doesn't really check the workfile version? The video below shows I have used different versions of workfile(v1 and v4 workfile respectivaly). But the button can't identify which one is the latest version. Same thing happened to the validator of checking workfile up to date, it will still passes if you use elder version files. image
> 2023-05-16.17-01-48.mp4

Did you publish all the workfiles you're testing with? This system is tied to publish time and not version number. The reason for this choice is that local workfiles version number may not match published version number, as users have the ability to increment their workfiles as much as they want to, without affecting published version numbers. And I do think they should absolutely keep this ability.

Also you will only get the expected behaviour once your workfiles has been published once with this system. This is due to the fact that datas must be set in the blend file by the update publish time plugin. So if a workfile has never been published with this workfile out of date handling system, then the data is just not there and the system can't make the difference between what's up to date what isn't.

moonyuet:
>>> Did you publish all the workfiles you're testing with? This system is tied to publish time and not version number. 
I have done two scanerios to ensure the validator works as expected. I published the workfile for v1 but i haven't published anything for v5.  

>>> The reason for this choice is that local workfiles version number may not match published version number
I dont think  `get_last_version_by_subset_name `will get any value even if the workfile has been published. I have made a simple script with validator for test and I just elaborate the check_workfile_up_to_date()  from workio.py to try to find to see if it can find the published version.  but it seems that the system can't find any last published version.

```
import pyblish.api

import bpy

    get_current_project_name,
    get_current_asset_name,
    get_current_task_name,
)

from openpype.client.entities import get_last_version_by_subset_name


class PublishedTimeDebug(pyblish.api.Validator):
    """Show The publish time differences"""

    hosts = ["blender"]
    families = ["workfile"]
    label = "Published Time Debug"
    # optional = True

    def process(cls, context):
        workfile_dict = cls.check_workfile_time_and_version()
        local_workfile_version = context.data("version")
        op_time = context.data("time")
        cls.log.debug("local_workfile_version:{}".format(local_workfile_version))
        cls.log.debug("op_setting_time:{}".format(op_time))

        cls.log.debug("published_version:{}".format(workfile_dict["published_version"]))
        if workfile_dict["last_published_time"]:
            cls.log.debug("last_published_time:{}".format(workfile_dict["last_published_time"]))

        if workfile_dict["op_published_time"]:
            cls.log.debug("op_published_time:{}".format(workfile_dict["op_published_time"]))
        if not workfile_dict:
            raise RuntimeError("No workfile data found!")


    @classmethod
    def check_workfile_time_and_version(cls) -> dict:

        """Check if the current workfile is out of date.

        This is based on last modification date, so if a user modifies an out of
        date workfile, this will return `False`. Also, in case of partial publish,
        this will return `True`.

        Returns:
            bool: True if the current workfile is up to date.
        """

        # Get date and time of the latest published workfile
        last_published_version = get_last_version_by_subset_name(
            get_current_project_name(),
            f"workfile{get_current_task_name()}",
            asset_name=get_current_asset_name(),
            fields=["data"]
        )

        scene = bpy.context.scene

        if not last_published_version:
            return {
                "op_published_time": scene.get("op_published_time"),
                "published_version": "N/A",
                "last_published_time": "N/A"
            }

        last_published_time = last_published_version["data"]["time"]
        if last_published_time:
            if scene.get("op_published_time"):
                return {
                    "op_published_time": scene.get("op_published_time"),
                    "published_version": last_published_version or "",
                    "last_published_time": last_published_time or ""
                }
            else:
                return {
                    "published_version": last_published_version,
                    "last_published_time": last_published_time
                }

```
Perhaps It will be great if you can do screencapture or recording to show how it works.

Sharkitty:
> 

If it's confusing for you, surely it will also be confusing for someone else. I'll add some much needed documentation then.

simonebarbieri:
@Sharkitty I am unable to make this work. I get the error sometimes when I open some old workfiles, but if I click on the error, or click to check again if it is out of date, it says that it's up to date.

I tried then publishing a new version of the workfile, just to test with two versions that have been published with this PR, and for both it says they're up to date. Although, from my understanding, the older one, that was published 5 minutes earlier, should result as outdated, or am I missing something?

simonebarbieri:
> * `wf_3` -> Last workfile, it's marked up to date.
> * `wf_2` -> Marked as out of date as it's no longer the last one.
> * `wf_1` -> Marked as out of date as there is no data to determine if it's up to date.

Thanks, this is perfectly clear :)
So, using this, `wf_1` is correctly marked as out of date when I open it, but if I click on "Check Workfile Up To Date", it is marked as up to date afterwards!
`wf_2` is marked as up to date on opening and after checking.
`wf_3` is correctly marked as up to date.

Sharkitty:
> > * `wf_3` -> Last workfile, it's marked up to date.
> > * `wf_2` -> Marked as out of date as it's no longer the last one.
> > * `wf_1` -> Marked as out of date as there is no data to determine if it's up to date.
> 
> Thanks, this is perfectly clear :) So, using this, `wf_1` is correctly marked as out of date when I open it, but if I click on "Check Workfile Up To Date", it is marked as up to date afterwards! `wf_2` is marked as up to date on opening and after checking. `wf_3` is correctly marked as up to date.

It's weird that `wf_1` passes from out of date to up to date like that, I'll look into it.

Sharkitty:
> > * `wf_3` -> Last workfile, it's marked up to date.
> > * `wf_2` -> Marked as out of date as it's no longer the last one.
> > * `wf_1` -> Marked as out of date as there is no data to determine if it's up to date.
> 
> Thanks, this is perfectly clear :) So, using this, `wf_1` is correctly marked as out of date when I open it, but if I click on "Check Workfile Up To Date", it is marked as up to date afterwards! `wf_2` is marked as up to date on opening and after checking. `wf_3` is correctly marked as up to date.

Hello, I rebased this branch and tried to look into your issue, although I'm unable to reproduce it. When I open an uncomformed workfile like `wf_1` in our previous example, it is marked as out of date both on launch and after reusing the operator as described. Can you try again? The rebase might have fixed it.

openpype/hosts/blender/api/workio.py
```diff
@@ -87,3 +94,35 @@ def work_root(session: dict) -> str:
...
+    this will return `True`.
+
+    Returns:
+        bool: True if the current workfile is up to date.
+    """
+
+    # Get date and time of the latest published workfile
+    last_published_version = get_last_version_by_subset_name(
+        get_current_project_name(),
+        f"workfile{get_current_task_name()}",
```
BigRoy, line 113: The subset naming can be customized as far as I know through settings - as such, it wouldn't be entirely safe to assume the subset name matches exactly this.

openpype/hosts/blender/plugins/publish/update_current_workfile_last_publish_time.py
```diff
@@ -0,0 +1,20 @@
...
+
+    order = pyblish.api.IntegratorOrder + 0.00002
+    hosts = ["blender"]
+    families = ["workfile"]
+    label = "Update Current Workfile Last Publish Time"
+    optional = False
+
+    def process(self, context):
+        scene = bpy.context.scene
+        scene["op_published_time"] = context.data["time"]
```
BigRoy, line 19: Storing into the file itself - doesn't that mean it'd basically become 'incorrect' as soon as someone would save the workfile into another asset (or another task) since the time stored in the scene wouldn't relate to that new context?

BigRoy (commented):
Interesting feature! :)

openpype/hosts/blender/plugins/publish/update_current_workfile_last_publish_time.py
Sharkitty: I guess so but is there a use case for that? I don't think that would be a very good practice, especially as there are other metadatas that would also be incorrect if you did that. I've already seen artists appending workfiles or elements of a workfile and ending up with a broken scene that could not be published (regardless of this feature). I think the same would happen with the scenario you are suggesting.

BigRoy: > I guess so but is there a use case for that? I don't think that would be a very good practice, especially as there are other metadatas that would also be incorrect if you did that. I've already seen artists appending workfiles or elements of a workfile and ending up with a broken scene that could not be published (regardless of this feature). I think the same would happen with the scenario you are suggesting.

Could you elaborate on the other type of metadata? We re-use workfiles all the time for other shots. If there are features currently disallowing that I'd say that's a bug and potentially a massive design flaw.

Would love to see reports of issues faced with that.

Sharkitty: In the current implementation of blender, metadata are stored in instance collections under the `avalon` key, aka `AVALON_PROPERTY`. They contain data such as the family, asset name, subset name, task name and variant.

To have more precise information about whether or not that would be problematic in the current version of openpype, I made a simple workfile with just a model subset containing the default cube. The asset name for this workfile is `Choochoo`. I saved as in another asset (`Pingoo`) folder using the correct syntax for the filename. Then I closed blender and reopened it as that other asset and I checked the metadata of the `modelMain` I created:
```
>>> for key, value in C.active_object.get('avalon').items():
...     print(f"{key}: {value}")
...     
id: pyblish.avalon.instance
family: model
asset: Choochoo
subset: modelMain
active: True
variant: Main
task: Fabrication
```
You can see that my `modelMain` is still recognized as part of the `Choochoo` asset. On one hand this is true. On the other hand, an artist who decides to use this subset as a base for a model in `Pingoo`, could publish its modified model thinking this is fine because they clicked on `Pingoo` earlier, yet the modified model will go in `Choochoo`. And that can be kinda dangerous. So I did it, and while the workfile was published in `Pingoo`, the model was published in `Choochoo`. In this case what should be done is to create a new instance and not publish the one already present in the workfile. In the context of a shot the same could happen with an animation subset for instance.

Does that work differently in other hosts?

All that being said, if you think this should remain doable in blender without confusing the workfile up to date system, I can look into it. Out of the blue I think I could check the asset name to see if the workfile has been moved around and adjust the behavior accordingly.

openpype/hosts/blender/api/ops.py
```diff
@@ -365,6 +373,108 @@ def execute(self, context):
...
+            # Get current project, asset and task names
+            project_name = get_current_project_name()
+            asset_name = get_current_asset_name()
+            task_name = get_current_task_name()
+
+            # Download, copy locally and open last workfile
+            last_workfile_path, last_published_time = download_last_workfile(
+                project_name, asset_name, task_name
+            )
+            if local_workfile_path := save_as_local_workfile(
```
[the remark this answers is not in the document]
Sharkitty: That line follows PEP8 convention afaik.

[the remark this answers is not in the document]
Minkiu: I think hound doesn't recognise the walrus operator, ignore it

openpype/hosts/blender/api/lib.py
```diff
@@ -315,3 +333,184 @@ def maintained_time():
...
+    ]
+    if not filtered_subsets:
+        raise RuntimeError(
+            "Not any subset for asset '{}' with id '{}'".format(
+                asset_doc["name"], asset_doc["_id"]
+            )
+        )
+
+    # Match subset wich has `task_name` in its name
+    low_task_name = task_name.lower()
```
Minkiu, line 387: Why not use it directly like `task_name.lower()` in line 390?

openpype/hosts/blender/api/ops.py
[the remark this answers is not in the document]
BigRoy: Walrus operator was added in Python 3.8+
Are we still supporting any Blender versions that might be using a lower Python version?

openpype/hosts/blender/api/lib.py
```diff
@@ -315,3 +333,184 @@ def maintained_time():
...
+    if len(filtered_subsets) > 1:
+        for subset in filtered_subsets:
+            if low_task_name in subset["name"].lower():
+                subset_id = subset["_id"]  # What if none is found?
+    else:
+        subset_id = filtered_subsets[0]["_id"]
+
+    if subset_id is None:
+        print(
```
Minkiu, line 396: Better to use the `log`

openpype/hosts/blender/api/lib.py
Sharkitty: Why did I do this in the first place <.<

Minkiu (approved):
haven't tested but code looks good

openpype/hosts/blender/api/workio.py
```diff
@@ -87,3 +97,51 @@ def work_root(session: dict) -> str:
...
+    if not last_published_version:
+        return True
+
+    scene = bpy.context.scene
+
+    last_published_time = last_published_version["data"]["time"]
+    if scene.get("op_published_time"):
+        return last_published_time <= scene["op_published_time"]
+    else:
+        return False
```
simonebarbieri, line 147: Wouldn't it be better to return `True` in this case? I guess this happens when users start using this feature in existing projects, and they have published workfiles already. I think it would be confusing to suddently get a message that their workfile is not Up to date.

Sharkitty: On one hand yes, and I agree that it could be confusing. On the other hand, in this fallback state where the system can't determine if the workfile is up to date or not, I think it's safer to wrongly mark it as out of date, in which case the worst thing that could happen is downloading a workfile while it wasn't necessary, than to mark it as up to date, in which case we would risk publishing out of date workfile, which is all that this system is trying to prevent.

This is why, in my opinion, returning `False` is the way to go. In practice, as I think downloading last workfile is what should generally be recommended to artists, and there is no risk in doing it. Artists will most likely just click ok and not worry too much about it.

Tilix4: I agree with Sharkitty. As this is a new feature, releasing a new OP version must come with information for artists about new behaviours they might encounter.  
Also, if we set it to `True`, we may leave issues to go through, like some one creates by mistake a new workfile from scratch (build or whatever), this workfile won't contain the key, and then he tries to publish. The user will have no any warning or anything to think about what is going on.

The fix could be to register a handler at the file opening to set this key with the appropriate value if it doesn't exist.

openpype/hosts/blender/api/ops.py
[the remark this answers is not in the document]
Tilix4: Also, this code is exclusively executed by Blender, which is now running only versions above py3.8

PR #4934, opened 2023-05-04
CI: WPS linting instead of Hound

antirotor:
## Changelog Description
Because Hound currently used to lint the code on GH ships with really old *flake8* support, it fails miserably on any newer Python syntax. This PR is adding WPS linter to GitHub workflows that should step in.

## Additional info
WPS linter suit is checking now even docstrings and it seems much more stricter than Hound, so it is a matter of discussion to set its limits.

## Testing notes:
Try this configuration file with your PRs and see its feedback. It should post code comments the same way as Hound does.


Minkiu:
Beyond the scope of this PR, but I was wondering if it would be beneficial to use `ruff` for linting: https://github.com/charliermarsh/ruff

antirotor:
> Beyond the scope of this PR, but I was wondering if it would be beneficial to use `ruff` for linting: [charliermarsh/ruff](https://github.com/charliermarsh/ruff)

so, the trouble with **ruff** is that it doesn't support posting review comments on GH as far as I know, the errors show only in process details and that is somewhat lost.

mkolar:
In that case, how about we get on with it and merge this?

PR #4935, opened 2023-05-04
Nuke: slates are published with old publisher comment

jakubjezek001:
## Changelog Description
Older publisher was storing comment on context level.

## Testing notes:
1. publishing should be working as it was

### DIFF
--- a/openpype/hosts/nuke/plugins/publish/extract_slate_frame.py
+++ b/openpype/hosts/nuke/plugins/publish/extract_slate_frame.py
@@ -298,7 +298,10 @@ class ExtractSlateFrame(publish.Extractor):
 
     def add_comment_slate_node(self, instance, node):
 
-        comment = instance.data["comment"]
+        comment = (
+            instance.context.data.get("comment")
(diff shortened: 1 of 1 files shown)


Minkiu (approved):

openpype/hosts/nuke/plugins/publish/extract_slate_frame.py
```diff
@@ -298,7 +298,10 @@ def _render_slate_to_sequence(self, instance):
 
     def add_comment_slate_node(self, instance, node):
 
-        comment = instance.data["comment"]
+        comment = (
+            instance.context.data.get("comment")
```
Minkiu, line 302: I was wondering, if we have a different comment per instance and a context comment, the context comment would take over the one in the instance, right?

jakubjezek001:
This was intended only to provide a backward compatibility. But I believe this is useless in New publisher, since comments are only stored in instance.data. Context.data were used previously. 

I am closing this PR since this doesnt make any sense to merge it.

PR #4940, opened 2023-05-05
General: fix build_workfile get_linked_assets missing project_name arg

kaamaurice:
## Changelog Description
Linked assets collection don't work within `build_workfile` because `get_linked_assets` function call has a missing `project_name `argument.
 - Added the `project_name` arg to the `get_linked_assets` function call.


iLLiCiTiT (approved):
LGTM

PR #4941, opened 2023-05-05
General: fix Scene Inventory switch version error dialog missing parent arg on init

kaamaurice:
## Changelog Description
QuickFix for the switch version error dialog to set inventory widget as parent.


BigRoy (approved):
Code changes look good to me. Didn't test the code.

PR #4943, opened 2023-05-05
Fusion: Saver creator settings

jakubjezek001:
## Changelog Description
Adding Saver creator settings and enhanced rendering path with template.

## Testing notes:
1. Open Fusion at test task 
2. Create Saver family
3. all should work as expected


openpype/hosts/fusion/plugins/create/create_saver.py
```diff
@@ -125,17 +135,34 @@ def _update_tool_with_data(self, tool, data):
...
+            "ext": "exr"
+        })
+
+        # build file path to render
+        filepath = self.temp_rendering_path_template.format(
+            **formatting_data)
+
+        # create directory
+        if not os.path.isdir(os.path.dirname(filepath)):
+            self.log.warning("Path does not exist! I am creating it.")
```
BigRoy, line 157: Logged message is very unclear. I would not auto-create these folders really.
I believe the Fusion saver has a setting to auto create directories if needed.

If we do want to create the folders then I'd at least log what path is being created.

openpype/settings/entities/schemas/projects_schema/schema_project_fusion.json
```diff
@@ -68,6 +68,50 @@
...
+                    "collapsible": true,
+                    "key": "CreateSaver",
+                    "label": "Create Saver",
+                    "is_group": true,
+                    "children": [
+                        {
+                            "type": "text",
+                            "key": "temp_rendering_path_template",
+                            "label": "Temporary rendering path template"
+                        },
```
BigRoy, line 89: Temporary rendering path feels like an odd wording for this. I can't think of anything else but it wasn't until I actually read what the default value is for this key that I understood what this tried to refer to. Can we think of anything that fits better?

BigRoy (commented):
Some notes

openpype/hosts/fusion/plugins/create/create_saver.py
EmberLightVFX: Yep, Fusions savers can create the folders when it starts to render. You even have a validator that checks that the `create folder` is checked

EmberLightVFX:
Great work on this!
I was thinking if it would be a good idea to add what format to render out to for the settings?
Like mov with h264/prores or exr with zip/piz/dwaa.
Currently it forces exr with zip but we mostly render out with dwaa ourself. Would be great to be able to set it for the saver and have a validator that checks it

openpype/settings/entities/schemas/projects_schema/schema_project_fusion.json
jakubjezek001: I am having no better name for now. This is now aligned with Nuke. But will add TODO note for later rename.

BigRoy (approved):
I tested this PR. It works for me.

However, I do wanted to highlight the a **backwards compatibility** change. Previously the savers by default would write to:
```
filename = f"{subset}..exr"
filepath = os.path.join(workdir, "render", subset, filename)
```
Which basically is:
```
"{workdir}/render/{subset}/{subset}.{frame}.{ext}"
```
_*But frame padding was always 4 and extension was always "exr"_

But now it writes to:
```
"{workdir}/renders/fusion/{subset}/{subset}.{frame}.{ext}"
```

As you can see the `fusion/` folder in there is a difference from the old behavior. It's customizable in settings but I just wanted to point it out.  Also the folder is now `renders` and previously appeared to be `render`?

We have `{app}` in our work folder templates so removing the explicit `fusion/` subfolder there is likely one of the first things we'd remove in settings to remain backwards compatible.

Should we change the default to be the same as before? Or is this change also done to make it more similar to what the maya defaults are?

BigRoy (changes requested):
Actually, no. Something is broken.

Whenever I create a saver instance, and then directly reset the publisher (without clicking save first) then this happens:

image

However, this **does not** happen when you click Create, then SAVE, and then reset the publisher. Cclicking Create should already persist the data as intended to the saver tool.

I'm not sure if it's due to this PR but it's definitely an issue we should solve.

EmberLightVFX:
> If we want to explicitly show only the tool name we can also set attribute: `tool.SetAttrs({"TOOLB_NameSet": True})` which makes the difference between the left entry (False) and the right entry (True):
> 
> image
> 
> With the explicit name set it works for both the backslashed paths and forward slashed paths.

I would personally go with render.####.exr if we now will be able to change the render type to say .mov or .tif an so on. That way the artist can quickly see the extension directly on the node.

openpype/hosts/fusion/plugins/create/create_saver.py
```diff
@@ -125,17 +138,35 @@ def _update_tool_with_data(self, tool, data):
...
+        # build file path to render
+        filepath = self.temp_rendering_path_template.format(
+            **formatting_data)
+
+        tool["Clip"] = filepath
+
+        # Rename tool
+        if tool.Name != subset:
+            print(f"Renaming {tool.Name} -> {subset}")
+            tool.SetAttrs({"TOOLS_Name": subset})
```
BigRoy, line 169:
```suggestion
            tool.SetAttrs({
                "TOOLB_NameSet": True,
                "TOOLS_Name": subset
            })
```
This fixes the issue described [here](https://github.com/ynput/OpenPype/pull/4943#issuecomment-1542176385) however it sets it without the extension which is something @EmberLightVFX mentioned he does like visually.

EmberLightVFX: One could add `_{ext}` after the name. Not the most beautiful but would show the extension in the name.

EmberLightVFX:
> It's a bit weird that Fusion shows the full path if the slashes don't match the current OS. If we'd want to show it like you said @EmberLightVFX that basically means to NOT set the name at all but do always set it with `os.path.normpath` applied to the filepath because an explicitly set name can't have `#` or `.` symbols in it. I'm not sure what that means for OS-mixed productions however.

Names on nodes can't contain some special characters and spaces. I'm guessing characters that could break scripting like dots and spaces.
I did a quick test and no matter if the path used `/` or `\` I would render out to the correct folder on PC.
I asked a friend on mac to do the same thing and no problems there ether. So shared comps between OSes will still work :)

BUT! For him it was reversed with the path vs filename. He saw the path when `\` was used and filename when `/` was used, so reverse from on Windows
I think using `os.path.normpath` is best to use ether way.

openpype/hosts/fusion/plugins/create/create_saver.py
BigRoy: @jakubjezek001 any preference on this?

BigRoy:
For what it's worth [this issue](https://github.com/ynput/OpenPype/pull/4943#pullrequestreview-1420557288) is due to `"id": "pyblish.avalon.instance"` not being set on create but only on updating because on create it doesn't seem to be in `instance_data` being passed along. Is this a bug? Or should we just always enforce that value to be added on create?

So on [`create`](https://github.com/ynput/OpenPype/blob/1a48ddefe4a665997dc9e4bac066200721949d3c/openpype/hosts/fusion/plugins/create/create_saver.py#L46) the `instance_data` does not contain `id`.

This is an example `instance_data` on create:

```
{'asset': 'asset1', 'task': 'modeling', 'variant': 'Main', 'family': 'render'}
```

@iLLiCiTiT @jakubjezek001 thoughts?

jakubjezek001:
> We're already explicitly adding subset into the data [here](https://github.com/ynput/OpenPype/pull/4943/files#diff-366db2d06eed704cb122f69f5c605fac2ff7a51c2fc93a6632bfc86a4ac15799R55). We can just add this line after and https://github.com/ynput/OpenPype/pull/4943#pullrequestreview-1420557288 is fixed:

Done

openpype/hosts/fusion/plugins/create/create_saver.py
jakubjezek001: I agree with @EmberLightVFX that it would be better to keep the name with extension.

openpype/hosts/fusion/plugins/create/create_saver.py
```diff
@@ -125,58 +138,54 @@ def _update_tool_with_data(self, tool, data):
...
+            "subset": ctx_data["subset"],
+            "task": ctx_data["task"],
+            "variant": ctx_data["variant"],
             "active": not passthrough,
             "family": self.family,
             # Unique identifier for instance and this creator
-            "id": "pyblish.avalon.instance",
+            "id": ctx_data["id"],
             "creator_identifier": self.identifier,
         }
```
BigRoy, line 190: This function should be removed completely, right? Managed savers are already being collected.

With this logic then a regular saver would start creating errors shown like this:
image

openpype/hosts/fusion/plugins/create/create_saver.py
```diff
@@ -78,7 +91,7 @@ def collect_instances(self):
         for tool in tools:
             data = self.get_managed_tool_data(tool)
             if not data:
-                data = self._collect_unmanaged_saver(tool)
+                data = self._collect_saver(tool)
```
BigRoy, line 94:
```suggestion
                continue
```

Unmanaged savers should just be ignored. Right?

openpype/hosts/fusion/plugins/create/create_saver.py
jakubjezek001: Yes, you are right. Removed.

antirotor (approved):

PR #4944, opened 2023-05-05
Feature: Copy resources when downloading last workfile

Sharkitty:
## Changelog Description
When the last published workfile is downloaded as a prelaunch hook, all resource files referenced in the workfile representation are copied to the `resources` folder, which is inside the local workfile folder.

## Testing notes:
- Open a workfile not locally available (so it is downloaded before launch), that references resource files.
- After launch, check the local workfile folder, it should have your workfile, as well as a `resources` folder with all resource files inside it.


Sharkitty:
Can @Tilix4 be added to reviewers please? :innocent:

openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
```diff
@@ -184,3 +185,22 @@ def execute(self):
...
+        for file in workfile_representation['files']:
+            resource_path = re.sub(
+                r"\{root\[main\]\}", str(anatomy.roots['main']), file['path']
+            )
+            # Only copy if the resource file exists, and it's not the workfile
+            if (
+                os.path.exists(resource_path)
+                and resource_path != last_published_workfile_path
+            ):
+                shutil.copy(resource_path, resources_dir)
```
BigRoy, line 206: Wouldn't this be quite dangerous in that it could override existing resources in the workdir if they did already exist? Not sure what the best UX is but this feels rather dangerous if that's the case. Should it either backup those files or maybe log a warning and skip the copy in that case?

Sharkitty: As this is hook is executed when no workfile is locally available, the only case I can think of in which that could happen is if a user goes into their task folder to delete all their workfiles but leaves the resource folder. Or if a user manually creates the task folder, the resource folder and manually copies the correct resources into it. But better safe than sorry. I think for now skipping with a warning might be more suitable, because local resource folders are not versioned so backing up resources would raise the question of "but what happens if the backup gets overwritten".

Tilix4: I think it should check if checksum is the same. If not, it should add `.old` to files with same name already present in directory (and if a `my_name.old` already exists, do `.old01`...).

Tilix4 (changes requested):
Nice! I think you should also make it work for `Copy and open` in `Workfiles` window.

mkolar:
This is pretty good, however, totally magical and should be documented somewhere. Otherwise it's completely arbitrary. Useful, but arbitrary.

Sharkitty:
> This is pretty good, however, totally magical and should be documented somewhere. Otherwise it's completely arbitrary. Useful, but arbitrary.

The copy last published workfile system is not documented in the first place so I guess I'll have to document that too ^^'

mkolar:
> The copy last published workfile system is not documented in the first place so I guess I'll have to document that too ^^'

Don't we all just love documentation. Best to go straight here. https://github.com/ynput/ayon-documentation

We'll be sunsetting openpype.io soon

Sharkitty:
> > The copy last published workfile system is not documented in the first place so I guess I'll have to document that too ^^'
> 
> Don't we all just love documentation. Best to go straight here. https://github.com/ynput/ayon-documentation
> 
> We'll be sunsetting openpype.io soon

So if I understand correctly, it's not necessary to update the documentation in OpenPype repo, but I should do it in the Ayon documentation repo instead, is that correct?

Sharkitty:
PR on ayon doc here: [PR 47](https://github.com/ynput/ayon-documentation/pull/47)

kalisp (changes requested):

Sharkitty:
Hello, any update on this PR?

kalisp:
You need to re-request review when you are done with fixes, otherwise PR won't show up in list of PRs to be reviewed.

openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
```diff
@@ -184,3 +185,53 @@ def execute(self):
...
+        for file in workfile_representation['files']:
+            resource_main_path = file["path"].replace(
+                "{root[main]}", str(anatomy.roots["main"])
+            )
+
+            # Only copy if the resource file exists, and it's not the workfile
+            if (
+                os.path.exists(resource_main_path)
+                and resource_main_path != last_published_workfile_path
+            ):
```
Tilix4, line 206: Instead of putting everything under a single condition, I'd create a `if not ... : continue`

openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
```diff
@@ -184,3 +185,53 @@ def execute(self):
...
+                resource_basename = os.path.basename(resource_main_path)
+                resource_work_path = os.path.join(
+                    resources_dir, resource_basename
+                )
+                if os.path.exists(resource_work_path):
```
Tilix4, line 211: Same, recommend `continue`

Tilix4 (commented):
The code looks fine to me, unless some things which could be cleaner IMO.

Also, this code is lacking code comments documentation, please add some to make your process more explicit.

openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
```diff
@@ -184,3 +185,53 @@ def execute(self):
...
+        # Get and make resources directory
+        resources_dir = os.path.join(
+            os.path.dirname(local_workfile_path), 'resources'
+        )
+        if not os.path.exists(resources_dir):
+            os.mkdir(resources_dir)
+
+        # Copy resources to the local resources directory
+        for file in workfile_representation['files']:
+            resource_main_path = file["path"].replace(
```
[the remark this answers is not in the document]
kalisp: Changed to list published `resource` folder directly as it should contain clean data and provides simpler logic.

kalisp:
Fixed some issues, changed looping through downloaded `publish` `resource` folder to simplify logic.

kalisp (approved):

openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
```diff
@@ -194,3 +195,71 @@ def execute(self):
...
+        # Make resource directory if it doesn't exist
+        if not os.path.exists(resources_dir):
+            os.mkdir(resources_dir)
+
+        # Copy resources to the local resources directory
+        for file in workfile_representation['files']:
+            # Get resource main path
+            resource_main_path = file["path"].replace(
+                "{root[main]}", str(anatomy.roots["main"])
+            )
```
BigRoy, line 212: @kalisp 

Why is this a hardcoded `root[main]`? Couldn't the studio template technically have _any_ root set? This should have been `file["path"].format(**anatomy.roots)` right?

Sharkitty: I thought I fixed that but I guess I must have made a mistake at some point because the commit is gone for some reason. Dunno what happened but I have other another fix to do anyway (Found cases in which workfiles are copied in the `resources` folder while they shouldn't) so I'll bring those changes back.

Sharkitty: Fix available in PR #5733

kalisp:
Wtf, my last commit with fixes disappeared :/.

Sharkitty:
> Wtf, my last commit with fixes disappeared :/.

I didn't see any, I was a bit confused

PR #4945, opened 2023-05-05
General: Applications addon

iLLiCiTiT:
## Changelog Description
Separate applications logic into an addon as preparation for AYON. This technically would allow in future replace the applications addon with different one.

## Additional info
The logic was located in `openpype.lib` which is wrong by design as it requires pipeline logic but lib content should not import from pipeline. Few parts of codebase are based on the applications logic which is also not right as it should be less dependent or the dependency should be abstracted or more explicit. Right now some of them stayed unchanged e.g. logic in ftrack module or launcher tool which should change over time, but that will require changes in those parts using them. There is a lot remaining but that require even more changes in more parts of codebase which would be hard to test in single PR.

## Testing notes:
1. Applications should launch as usually
2. Launch hooks should work


iLLiCiTiT:
> Just reading over this PR I'm unsure about why things like prelaunch hooks and applications are to be separated into addons?
Shouldn't we somehow at least document or identify what OpenPype or Ayon needs at its core for the pipeline to work?

An addon adds functionality, and addons can "be prepared" for other addons. Applications addon should define how applications are launched and also should define how other addons can affect that. Having a prelaunch hook for applications addon in different addon is ok. There are already addons with custom prelaunch hooks. Or addons which have ftrack actions, but they're not used until ftrack is available.

What are the benefits? For example possibility to replace current applications launch logic with e.g. rez specific, so instead of trying to bend current launch logic there would be rez addon which defines how other addons can affect the launch and how to communicate with it.

BigRoy:
> What are the benefits? For example possibility to replace current applications launch logic with e.g. rez specific, so instead of trying to bend current launch logic there would be rez addon which defines how other addons can affect the launch and how to communicate with it.

This is exactly what I'm getting at. I see why that might be a goal, but I don't see yet how this is solving it. Because if I wanted to swap out the launching logic I'd still need to update **all** addons that are built to rely on the Applications addon.

What if we were to keep the launch hooks separate from addons, like a built-in feature, but the application ends up just using that interface. Looking at the changes required in this PR something like that would make it much more viable to swap out Application with a Rez launch system if at least it can still trigger the same addons. Right?

Preferably the system is built in such a way that if I were to replace how an application is launched it doesn't mean that I'd need to replace all the required dependencies of all addons like Maya, Houdini, etc. Otherwise you'd just end up still forking ALL of those just to launch through your own custom application launcher instead of just saying "launch differently" but use the same maya addon.

I understand the goal, but this doesn't seem at a point where it's getting nearer to solving that yet to me.

BigRoy:
> What are the benefits? For example possibility to replace current applications launch logic with e.g. rez specific, so instead of trying to bend current launch logic there would be rez addon which defines how other addons can affect the launch and how to communicate with it.

This is exactly what I'm getting at. I see why that might be a goal, but I don't see yet how this is solving it. Because if I wanted to swap out the launching logic I'd still need to update **all** addons that are built to rely on the Applications addon.

What if we were to keep the launch hooks separate from addons, like a built-in feature, but the application ends up just using that interface. Looking at the changes required in this PR something like that would make it much more viable to swap out Application with a Rez launch system if at least it can still trigger the same addons. Right?

Preferably the system is built in such a way that if I were to replace how an application is launched it doesn't mean that I'd need to replace all the required dependencies of all addons like Maya, Houdini, etc. Otherwise you'd just end up still forking ALL of those just to launch through your own custom application launcher instead of just saying "launch differently" but use the same maya addon.

I understand the goal, but this doesn't seem at a point where it's getting nearer to solving that yet to me.

iLLiCiTiT:
> I understand the goal, but this doesn't seem at a point where it's getting nearer to solving that yet to me.

The issue is that we would like to do it at a certain point and it is not possible to do it in one PR. If we'll keep it in core addon it would look like we don't want to change it, which leads to 2 things: Developers will use it that way and expect it is always available and they'll start to bend it to misuse it (already happening). Longer it is in core the more problematic it will be to change it. Will cause bigger and bigger backwards compatibility troubles when we would like to change it. Changing it this way makes it a lot more clear and transparent. It already behaved as an addon even if it was in lib. Also in AYON it already is separated to an addon. We can separate the documentation for it and start to do things for which we didn't have usecases.

Long story short: If we won't start with smaller steps we may never change it, or it will be such big change that it will break other addons in a really bad way.

This PR should not break anything. The imports from lib, cli command should still work.

tokejepsen:
Will need to resolve the merge conflict and maybe a checklist of applications that has been tested.

mkolar:
@iLLiCiTiT I'm assuming we'll need to port this to ayon-core, so no point merging it here right?

iLLiCiTiT:
Yes

PR #4948, opened 2023-05-08
Global slate workflow from Max Pareschi

fabiaserra:
## Changelog Description

This is a clean up of @maxpareschi's Global Slate workflow https://github.com/ynput/OpenPype/pull/3931. I addressed the comments from that other PR, simplified the code, merged latest develop and adapted a few things to get it to work in our studio and for more people by adding the template resources inside OpenPype (so far it's working great so thanks again @maxpareschi!!).

This plugin allows us to generate Slate frames independent of the host using an HTML template:

image

> This enables a Project Settings > Global setting called ExtractSlateGlobal, which is toggleable and overrides the slate workflow in nuke without changing it, while enabling all other dccs to have that ability for sequences and reviewables. Tested in nuke and maya for now.
It uses Html2image to render a screen referred slate image which can then be colorconverted using oiio and its profiles.

## Additional info
I know that there's another internal Slate plugin from Ynput so I understand this won't get merged but it might be helpful for some other people. 

On the other hand, we are using Linux in our studio so we were able to install Chrome centrally on our shared server for the `html2image` module by setting it on the `OPENPYPE_CHROME_PATHS` but for Windows users it could make sense to add the binary as a `vendor` dependency as Chrome.

## Testing notes:
1. Install Chrome in your machine (needs to be version 110 of Chrome as version 111 has a bug when trying to set size resolution when using `html2image.screenshot) and set `$OPENPYPE_CHROME_PATHS` to point to the installation directory where `chrome` exists.
2. Turn on the plugin Extract Global Slate
3. Publish something


openpype/hosts/nuke/plugins/publish/extract_slate_frame.py
```diff
@@ -36,6 +36,12 @@ class ExtractSlateFrame(publish.Extractor):
 
     def process(self, instance):
 
+        # skip this plugin if the slate global one is active
+        if "slateGlobal" in instance.data:
+            self.log.info("Slate Global workflow is active, skipping Nuke Slate "
```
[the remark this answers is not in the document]
fabiaserra: damn hound, why not use the default `black` line width of 88 :( https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length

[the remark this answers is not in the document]
BigRoy: I believe it's solely due to being PEP based, not Black-based but @iLLiCiTiT or @mkolar can probably confirm.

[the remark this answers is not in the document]
fabiaserra: yeah I know, but 79 is very restrictive and I'm not the only one saying it https://www.youtube.com/watch?v=wf-BqAjZb8M&t=260s :)

openpype/plugins/publish/extract_slate_global.py
```diff
@@ -0,0 +1,538 @@
...
+        if subfolder:
+            staging_dir = os.path.join(staging_dir, subfolder)
+
+        if not os.path.exists(staging_dir):
+            os.makedirs(staging_dir)
+
+        self.staging_dir = staging_dir
+        self.log.debug("Staging dir set to: '%s'", staging_dir)
+
+    def read_template(self, template_path=None, resources_path=None):
```
BigRoy, line 100: Is it weird maybe that the class `__init__` allows to take in a template path and resources path to initialize but then these methods also do the same? I feel like with an instantiated object it makes more settings that it either takes the settings from the instance **or** that this method re-initializes the object to that new template, however this function does not seem to override the `self.template_path` or `self.resources_path` if it were supplied.

Should `read_template` either take the values from `self` OR should it be the one that is also storing/assigning it to `self` instead?

openpype/plugins/publish/extract_slate_global.py
```diff
@@ -0,0 +1,538 @@
...
+            staging_dir = tempfile.gettempdir()
+
+        if subfolder:
+            staging_dir = os.path.join(staging_dir, subfolder)
+
+        if not os.path.exists(staging_dir):
+            os.makedirs(staging_dir)
+
+        self.staging_dir = staging_dir
+        self.log.debug("Staging dir set to: '%s'", staging_dir)
```
BigRoy, line 98: Is this confusing maybe that this uses a different staging dir data than the one from the publishing instance?

openpype/plugins/publish/extract_slate_global.py
```diff
@@ -0,0 +1,538 @@
...
+
+        if self._slate_data_name not in instance.data:
+            self.log.warning(
+                "Slate Global workflow is not active, skipping slate extraction..."
+            )
+            return
+
+        if "representations" not in instance.data:
+            self.log.error("No representations to work on!")
+            raise ValueError("No items in list.")
```
BigRoy, line 371: Should this instead just log a warning and return early?

If not, let's at least report more clearly in the error what list is being referred to, etc.

openpype/plugins/publish/extract_slate_global.py
```diff
@@ -0,0 +1,538 @@
...
+                )
+
+        repre_match = None
+
+        # loop through representations and generate a slate frame for each
+        for repre in instance.data["representations"]:
+            if repre["name"] in repre_ignore_list:
+                self.log.debug(
+                    "Representation '{}' was ignored.".format(repre["name"])
+                )
```
BigRoy, line 410: Should this be present tense?
```suggestion
                self.log.debug(
                    "Representation '{}' is ignored.".format(repre["name"])
                )
```

openpype/plugins/publish/extract_slate_global.py
```diff
@@ -0,0 +1,538 @@
...
+                continue
+
+            # check if repre is a sequence
+            is_sequence = False
+            check_file = repre["files"]
+            if isinstance(check_file, list):
+                check_file = check_file[0]
+                is_sequence = True
```
BigRoy, line 418: Is this more readable?
```suggestion
            # check if repre is a sequence
            is_sequence = isinstance(check_file, list)
            check_file = repre["files"][0] if is_sequence else repre["files"]
```

openpype/plugins/publish/extract_slate_global.py
```diff
@@ -0,0 +1,538 @@
...
+
+            else:  # else find matching tags and transfer
+                frame_start = int(repre["frameStart"])
+                frame_end = int(repre["frameEnd"])
+                thumbnail_path = repre_thumbnail_path
+                output_name = "{}_slate_temp.png".format(repre["name"])
+                for tag in repre["tags"]:
+                    for profile in slate_data["slate_profiles"]:
+                        if tag in profile["families"]:
+                            repre_match = tag
```
BigRoy, line 449: I'm not entirely sure from reading the code, but what happens with single images as opposed to sequences? It does not get a slate? or it does?

openpype/plugins/publish/extract_slate_global.py
```diff
@@ -0,0 +1,538 @@
...
+                        "hosts": [],
+                        "oiio_args": {"input": [], "output": []},
+                    }
+                oiio_profile["oiio_args"]["output"].extend(
+                    [
+                        "--attrib:type=timecode",
+                        "smpte:TimeCode",
+                        '"{}"'.format(timecode),
+                    ]
+                )
```
BigRoy, line 471: 1. Should this break on the first match?
2. Since we have an `else` fallback doesn't it just always take the default if it doesn't have a match?

How about:

```suggestion
            # Default
            oiio_profile = {
                "families": [],
                "hosts": [],
                "oiio_args": {"input": [], "output": []},
            }
            for profile in slate_data["slate_profiles"]:
                if repre_match in profile["families"]:
                    # Use profile matching defaults
                    oiio_profile = profile
                    break
            
            # Add timecode attribute
            oiio_profile["oiio_args"]["output"].extend(
                [
                    "--attrib:type=timecode",
                    "smpte:TimeCode",
                    '"{}"'.format(timecode),
                ]
            )
```

Also, I feel like this behaves differently than other "profiles" with profile filtering in OpenPype. Should it behave more similar and use the regular filter_profiles function?

poetry.lock
BigRoy, line 1: This lock file seems to update much more than just the dependencies of `html2image` package, right? Is that intentional?

I also suspect since we're tweaking `poetry.lock` that likely this PR might need to be based to `next-minor` instead of `developed`

openpype/plugins/publish/collect_slate_global.py
```diff
@@ -0,0 +1,119 @@
...
+                        instance.data["slateGlobal"],
+                        indent=4,
+                        default=str
+                    )
+                )
+            else:
+                self.log.debug(
+                    "Task: %s is disabled for Extract Slate Global workflow, "
+                    "skipping slate extraction on review families...", task_type
+                )
```
BigRoy, line 119: 1. We can avoid the if nesting for whether settings exists by returning early
2. `instance.context.data` access is not needed if `context = instance.data` is already defined
3. Refactored logs to say "found in project settings" instead of "found on the project settings"

```suggestion
        context = instance.context
        slate_settings = (
            context.data["project_settings"]["global"]
            ["publish].get("ExtractSlateGlobal")
        )

        if not slate_settings:
            self.log.warning("No slate settings found. Skipping.")
            return
        
        if not slate_settings["enabled"]:
            self.log.warning("ExtractSlateGlobal is not active. Skipping.")
            return

        if context.data.get("host") == "nuke" and \
                "render.farm" in instance.data.get("families"):
            self.log.warning(
                "Skipping Slate Global Collect in Nuke context, defer to "
                "Deadline."
            )
            return

        self.log.info("ExtractSlateGlobal is active.")

        # Create dictionary of common data across all slates
        frame_padding = context.data["anatomy"]["templates"]["defaults"]\
            ["frame_padding"]
        slate_common_data = {
            "@version": str(instance.data["version"]).zfill(frame_padding),
            "frame_padding": frame_padding,
            "intent": {"label": "", "value": ""},
            "comment": "",
            "scope": "",
        }
        slate_common_data.update(instance.data["anatomyData"])
        if "customData" in instance.data:
            slate_common_data.update(instance.data["customData"])

        template_path = slate_settings["slate_template_path"].format(
            **os.environ
        )
        if not template_path:
            template_path = resources.get_resource(
                "slate_template", "generic_slate.html"
            )
            self.log.info(
                "No 'slate_template_path' found in project settings. "
                "Using default '%s'", template_path
            )

        resources_path = slate_settings["slate_resources_path"].format(
            **os.environ
        )
        if not resources_path:
            resources_path = resources.get_resource(
                "slate_template", "resources"
            )
            self.log.info(
                "No 'slate_resources_path' found in project settings. "
                "Using default '%s'", resources_path
            )

        slate_global = {
            "slate_template_path": template_path,
            "slate_resources_path": resources_path,
            "slate_profiles": slate_settings["profiles"],
            "slate_common_data": slate_common_data,
            "slate_thumbnail": "",
            "slate_repre_data": {},
            "slate_task_types": slate_settings["integrate_task_types"]
        }
        instance.data["slateGlobal"] = slate_global

        if "families" not in instance.data:
            instance.data["families"] = list()

        if "versionData" not in instance.data:
            instance.data["versionData"] = dict()

        if "families" not in instance.data["versionData"]:
            instance.data["versionData"]["families"] = list()

        task_type = instance.data["anatomyData"]["task"]["type"]
        if task_type in slate_settings["integrate_task_types"]:

            self.log.debug(
                "Task: %s is enabled for Extract Slate Global workflow, "
                "tagging for slate extraction on review families", task_type
            )

            instance.data["slate"] = True
            instance.data["families"].append("slate")
            instance.data["versionData"]["families"].append("slate")

            self.log.debug(
                "SlateGlobal Data: %s", json.dumps(
                    instance.data["slateGlobal"],
                    indent=4,
                    default=str
                )
            )
        else:
            self.log.debug(
                "Task: %s is disabled for Extract Slate Global workflow, "
                "skipping slate extraction on review families...", task_type
            )
```

openpype/plugins/publish/extract_review_slate.py
```diff
@@ -28,7 +28,26 @@ class ExtractReviewSlate(publish.Extractor):
...
+        "fusion",
+        "tvpaint",
+        "resolve",
+        "webpublisher",
+        "aftereffects",
+        "flame",
+        "unreal"
+    ]
+
     optional = True
```
BigRoy, line 51: Unrelated to this PR butl ikely a Nuke bug. The optional state here doesn't actually work in the new publisher. Right? @iLLiCiTiT

openpype/lib/vendor_bin_utils.py
```diff
@@ -390,6 +390,76 @@ def get_ffmpeg_tool_path(tool="ffmpeg"):
...
+
+    Returns:
+        bool: Filepath is valid executable.
+    """
+
+    filepath = find_executable(filepath)
+    if not filepath:
+        return False
+
+    return _check_args_returncode([filepath, "-version"])
```
BigRoy, line 420: > Install Chrome in your machine (needs to be version 110 of Chrome as version 111 has a bug when trying to set size resolution when using html2image.screenshot) and set $OPENPYPE_CHROME_PATHSto point to the installation directory wherechrome` exists.

Could this technically also log or report the error with that particular version?

BigRoy (commented):
Added some notes.

Aside of that, could someone explain why with this PR there are TWO Slate Extractors?
One is old, one is new. Which triggers when? Admittedly this looks relatively confusing.

I guess whenever you enable this one it overrides the other but I feel like shouldn't we somehow make one extractor that just does either the one or the other instead of having both? And couldn't we create a base class interface for the extractor that could work for both?

So there's just a default `SlateRenderer` class that just takes some data and has a render method to generate the output. The extractor then decides which `SlateRenderer` class to use and just initializes it and calls the `slate.render()` method? Or am I overcomplicating this?

openpype/plugins/publish/extract_review_slate.py
```diff
@@ -28,7 +28,26 @@ class ExtractReviewSlate(publish.Extractor):
...
+        "traypublisher",
+        "standalonepublisher",
+        "fusion",
+        "tvpaint",
+        "resolve",
+        "webpublisher",
+        "aftereffects",
+        "flame",
+        "unreal"
+    ]
```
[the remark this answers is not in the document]
fabiaserra: only reason being I didn't know that was possible :P but thanks to you now I know! Changing it!

openpype/hosts/nuke/plugins/publish/extract_slate_frame.py
```diff
@@ -36,6 +36,13 @@ class ExtractSlateFrame(publish.Extractor):
...
+                          "extraction..."
+            )
```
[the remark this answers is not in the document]
fabiaserra: yeah I like that formatting much more as well, Hound doesn't seem to like it though but I'm changing it, thanks!

openpype/plugins/publish/extract_slate_global.py
fabiaserra: Yeah I agree and I had the same thought while refactoring Max's code in a few other places with some of the variables stored on the instance. I did remove a few of them but in some places like this one it felt like someone could benefit from being able to override the template used after the SlateCreator object had been created... although I'm usually more of the philosophy of starting as simple as possible and only expand logic if needed.

So long story short, I'd be happy on removing the args from this function and just using the ones initialized on the instance, what do you think?

openpype/plugins/publish/extract_slate_global.py
fabiaserra: Yeah good call, changed to warning

openpype/plugins/publish/extract_slate_global.py
fabiaserra: fair yeah, changed

openpype/plugins/publish/extract_slate_global.py
fabiaserra: yes! I like it, changing it

openpype/lib/vendor_bin_utils.py
fabiaserra: yeah what would definitely be useful, although I'm not sure if I want to add the complexity of parsing the version string (different in chromium and chrome and who knows other builds) and perhaps introducing some bugs for the small wins of this...
Something like this could maybe work:
```python
version_str = _check_args_returncode([filepath, "-version"])
if not version_str:
    return False

# Check that the major version is not 112 as it has a bug on resolution
integers = re.findall(r"\d+", version_str )
if integers and integers[0] == '112':
    print(f"The major version 112 of Chrome has a bug, update to <112 or 113+.")
    return False

return True
```
what do you think?

poetry.lock
fabiaserra: yeah I think that's expected as we rely on `poetry` to handle the updates and not manually update `poetry.lock` but yeah, likely would need to be based against `next-minor` for this reason. As I said on the description though, I doubt Ynput will be happy to merge this for some of the reasons outlined there and not wanting to introduce more dependencies? So it's likely one of those that will be possible only with Ayon independent addons

openpype/plugins/publish/extract_slate_global.py
fabiaserra: oops, `check_file` isn't defined otherwise, changed to:
```python
is_sequence = isinstance(repre["files"], list)
check_file = repre["files"][0] if is_sequence else repre["files"]
```

openpype/lib/vendor_bin_utils.py
BigRoy: Ah yes, if version strings returned might differ it might not be worth the trouble. Your check seems fine to me however, hard to tell without test runs.

openpype/plugins/publish/extract_slate_global.py
```diff
@@ -0,0 +1,536 @@
...
+        # loop through representations and generate a slate frame for each
+        for repre in instance.data["representations"]:
+            if repre["name"] in repre_ignore_list:
+                self.log.debug(
+                    "Representation '{}' is ignored.".format(repre["name"])
+                )
+                continue
+
+            # check if repre is a sequence
+            is_sequence = isinstance(check_file, list)
```
BigRoy:
```suggestion
            is_sequence = isinstance(repre["files"], list)
```

BigRoy:
```suggestion
            is_sequence = isinstance(repre["files"], list)
```

fabiaserra:
> Added some notes.
> 
> Aside of that, could someone explain why with this PR there are TWO Slate Extractors? One is old, one is new. Which triggers when? Admittedly this looks relatively confusing.
> 
> I guess whenever you enable this one it overrides the other but I feel like shouldn't we somehow make one extractor that just does either the one or the other instead of having both? And couldn't we create a base class interface for the extractor that could work for both?
> 
> So there's just a default `SlateRenderer` class that just takes some data and has a render method to generate the output. The extractor then decides which `SlateRenderer` class to use and just initializes it and calls the `slate.render()` method? Or am I overcomplicating this?

Thank you very much @BigRoy for taking your time on reviewing this!!

And yeah, great points. With this global plugin I'd definitely mark the other one as deprecated but we probably wouldn't want to do that as a minor release, and some people might still prefer the control of the older one of just using Nuke for templating the Slate instead of HTML, even though this one is much more feature complete.

As for the common class... likely we could abstract a few things but really not that much as the former plugin is very tied to Nuke and it's probably not worth the effort if we'd deprecate it as you suggested

openpype/plugins/publish/extract_slate_global.py
BigRoy: Hard to tell what the intended behavior is there. I'd say if this is about matching families/profiles to instead rely on regular profiles filtering that OpenPype already has dedicated functions for just so it's using the same profiles filtering system as is used in other areas for consistency. That function instead takes the "best match" instead of the first.

iLLiCiTiT:
Few general notes that, I think, are for crutial.
- Any changes in dependencies must always go to next minor release branch, never to develop.
- I would not make it default or global, but as one of variants for slates, so I would rename it from "global" to "htmlSlate" and disable it by default. Studio can choose which you want to use. We know about some studios which already have their own.
- How studio can change the look of a slate? It looks like it is in a codebase so it can't be changed by production needs?

openpype/lib/vendor_bin_utils.py
```diff
@@ -390,6 +390,76 @@ def get_ffmpeg_tool_path(tool="ffmpeg"):
...
+
+
+def get_chrome_tool_path(tool="chrome"):
+    """Path to vendorized FFmpeg executable.
+
+    Args:
+        tool (string): Tool name (ffmpeg, ffprobe, ...).
+            Default is "ffmpeg".
```
iLLiCiTiT, line 428:
```suggestion
def get_chrome_tool_path():
    """Path to vendorized FFmpeg executable.
```

openpype/plugins/publish/collect_slate_global.py
```diff
@@ -0,0 +1,121 @@
+import os
+import json
+import pyblish.api
+from openpype import resources
+
+
+class CollectSlateGlobal(pyblish.api.InstancePlugin):
```
iLLiCiTiT, line 7: Is it necessary to have collector for the extraction? Content from here can be moved to extractor without any harm, or? Some custom data can be filled afterwards.

openpype/plugins/publish/collect_slate_global.py
```diff
@@ -0,0 +1,121 @@
...
+        if not slate_settings:
+            self.log.warning("No slate settings found. Skipping.")
+            return
+
+        if not slate_settings["enabled"]:
+            self.log.warning("ExtractSlateGlobal is not active. Skipping.")
+            return
+
+        if context.data.get("host") == "nuke" and \
+                "render.farm" in instance.data.get("families"):
```
iLLiCiTiT, line 31: Not sure what is reason of this condition?

But the key is `"hostName"` and is always filled.
```suggestion
        if (
            context.data["hostName"] == "nuke"
            and "render.farm" in instance.data["families"]):
```

openpype/plugins/publish/collect_slate_global.py
```diff
@@ -0,0 +1,121 @@
...
+            )
+            return
+
+        self.log.info("ExtractSlateGlobal is active.")
+
+        # Create dictionary of common data across all slates
+        frame_padding = context.data["anatomy"]["templates"]["defaults"]\
+            ["frame_padding"]
+        slate_common_data = {
+            "@version": str(instance.data["version"]).zfill(frame_padding),
```
iLLiCiTiT, line 44:
```suggestion
        slate_common_data = {
            "version": instance.data["version"],
            "@version": str(instance.data["version"]).zfill(frame_padding),
```

openpype/plugins/publish/collect_slate_global.py
```diff
@@ -0,0 +1,121 @@
...
+            "slate_thumbnail": "",
+            "slate_repre_data": {},
+            "slate_task_types": slate_settings["integrate_task_types"]
+        }
+        instance.data["slateGlobal"] = slate_global
+
+        if "families" not in instance.data:
+            instance.data["families"] = list()
+
+        if "versionData" not in instance.data:
```
iLLiCiTiT, line 92: Why are `"versionData"` filled here?

openpype/plugins/publish/extract_review_slate.py
iLLiCiTiT: Did you already created issue or should I create it?

openpype/plugins/publish/extract_slate_global.py
```diff
@@ -0,0 +1,541 @@
+import os
```
iLLiCiTiT, line 1: This plugin is using python modules that may cause crashes in some DCCs. Maybe some of the logic should be separated to a OpenPype script.

openpype/plugins/publish/extract_slate_global.py
```diff
@@ -0,0 +1,541 @@
...
+    def __init__(
+        self,
+        template_path,
+        resources_path="",
+        staging_dir="",
+        staging_subfolder="",
+        data=None,
+    ):
+        self.data = data.copy() if data else {}
+        self.log = logging.getLogger(self.__class__.__name__)
```
iLLiCiTiT, line 52: This should get logger from plugin probably, otherwise the logs won't be available in UI.

openpype/plugins/publish/extract_slate_global.py
```diff
@@ -0,0 +1,541 @@
...
+        cmd.extend(out_args or [])
+        cmd.extend(["-o", output])
+        try:
+            run_subprocess(cmd, logger=self.log)
+        except TypeError as error:
+            raise TypeError(
+                "%s: Error creating '%s' due to: %s", name, output, error
+            )
+
+    def get_timecode_oiio(self, input, timecode_frame=1001):
```
iLLiCiTiT, line 274: `input` is a builtin function I would recommend `src_path`.

openpype/plugins/publish/extract_slate_global.py
```diff
@@ -0,0 +1,541 @@
...
+            )
+            return
+
+        repre_ignore_list = ["thumbnail", "passing"]
+
+        slate_data = instance.data[self._slate_data_name]
+
+        # get pyblish comment and intent
+        common_data = slate_data["slate_common_data"]
+        common_data["comment"] = instance.context.data.get("comment") or "-"
```
iLLiCiTiT, line 380:
```suggestion
        common_data["comment"] = instance["comment"]
```

openpype/plugins/publish/extract_slate_global.py
```diff
@@ -0,0 +1,541 @@
...
+                repre_match = instance.data["family"]
+
+            else:  # else find matching tags and transfer
+                frame_start = int(repre["frameStart"])
+                frame_end = int(repre["frameEnd"])
+                thumbnail_path = repre_thumbnail_path
+                output_name = "{}_slate_temp.png".format(repre["name"])
+                for tag in repre["tags"]:
+                    for profile in slate_data["slate_profiles"]:
+                        if tag in profile["families"]:
```
iLLiCiTiT, line 448: Why is a `tag` checked if is in `families`? Those are 2 totally different things...

iLLiCiTiT: Profiles are usually instance data based, not representation based.

fabiaserra:
> Few general notes that, I think, are for crutial.
> 
> * Any changes in dependencies must always go to next minor release branch, never to develop.
Definitely yeah, I will update the target branch

> * I would not make it default or global, but as one of variants for slates, so I would rename it from "global" to "htmlSlate" and disable it by default. Studio can choose which you want to use. We know about some studios which already have their own.
I don't mind renaming it but I'm not sure if "htmlSlate" infers it's a DCC agnostic plugin and can be used for all host/applications? Ideally there would be one "global" supported plugin and deprecate the Nuke one which is more limited as default. If studios have their own plugin that should be the one that's specific and not "global"/"generic"

> * How studio can change the look of a slate? It looks like it is in a codebase so it can't be changed by production needs?
It's done all through the HTML template so yeah they can change it without changing the codebase (I don't consider HTML code :) ) at any project as a setting override here:
image

In the long run, we could expose some other common variations of the template to modify the HTML directly through the settings so they wouldn't need to set a different HTML template

openpype/plugins/publish/extract_slate_global.py
BigRoy: `instance["comment"]`? Did you mean `instance.data["comment"]`?

fabiaserra:
Thank you very much @iLLiCiTiT for reviewing and adding your feedback!

openpype/plugins/publish/extract_slate_global.py
fabiaserra: can confirm:
```
(Pdb) p instance
pyblish.plugin.Instance("renderLightingMain")
(Pdb) p instance["comment"]
*** TypeError: list indices must be integers or slices, not str
(Pdb) p instance.data["comment"]
*** KeyError: 'comment'
(Pdb) p instance.context.data["comment"]
'test'
```
`instance.data["comment"]` doesn't seem to exist either

openpype/plugins/publish/collect_slate_global.py
iLLiCiTiT: That's available only in pyblish pype UI, and does not reflect name of host integration but one of regitered host name in pyblish.

openpype/plugins/publish/collect_slate_global.py
```diff
@@ -0,0 +1,121 @@
...
+        slate_common_data = {
+            "@version": str(instance.data["version"]).zfill(frame_padding),
+            "frame_padding": frame_padding,
+            "intent": {"label": "", "value": ""},
+            "comment": "",
+            "scope": "",
+        }
+        slate_common_data.update(instance.data["anatomyData"])
+        if "customData" in instance.data:
+            slate_common_data.update(instance.data["customData"])
```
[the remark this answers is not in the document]
iLLiCiTiT: All custom data must be under `"custom"` key in fill data to match other templates for slates and burnins.
To use custom data in template there should be `{custom[my_custom_key]}`. That's to avoid clashes of keys.
```
{
    "project": {"name": "project", "code": "code"},
    "asset": "some asset",
    ...
    "custom": {
        ... custom data. ..
    }
}
```

openpype/plugins/publish/collect_slate_global.py
iLLiCiTiT: Version data are additional data for version document. Families are filled automatically in integrate.

openpype/plugins/publish/extract_slate_global.py
iLLiCiTiT: > If the latter, the plugin wouldn't show up on the DCC but that's it? I guess you are referring to html2image only?

Yes and yes.

openpype/plugins/publish/extract_slate_global.py
iLLiCiTiT: Or just pass it in arguments?

openpype/plugins/publish/extract_slate_global.py
iLLiCiTiT: > instance.data["comment"] doesn't seem to exist either

It is always available https://github.com/ynput/OpenPype/blob/develop/openpype/plugins/publish/collect_comment.py#L104

iLLiCiTiT:
> I don't mind renaming it but I'm not sure if "htmlSlate" infers it's a DCC agnostic plugin and can be used for all host/applications? Ideally there would be one "global" supported plugin and deprecate the Nuke one which is more limited as default. If studios have their own plugin that should be the one that's specific and not "global"/"generic"

I don't think this will make the other slate generation as deprecated. Nuke supports more, like colorspaces, and does not require chrome available. IMHO this has dependencies that may not be available and I don't think it should be named as `global` nor `generic` but more specific. But that's my opinion @mkolar , @jakubjezek001 , @antirotor can you share yours?

openpype/plugins/publish/collect_slate_global.py
fabiaserra: fair enough, good to know!

openpype/plugins/publish/collect_slate_global.py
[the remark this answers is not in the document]
fabiaserra: oh ok I see what you meant with your prior comment now, makes sense, I will update!

fabiaserra:
> > I don't mind renaming it but I'm not sure if "htmlSlate" infers it's a DCC agnostic plugin and can be used for all host/applications? Ideally there would be one "global" supported plugin and deprecate the Nuke one which is more limited as default. If studios have their own plugin that should be the one that's specific and not "global"/"generic"

I don't think this will make the other slate generation as deprecated. Nuke supports more, like colorspaces, and does not require chrome available. IMHO this has dependencies that may not be available and I don't think it should be named as `global` nor `generic` but more specific. But that's my opinion @mkolar , @jakubjezek001 , @antirotor can you share yours? 

Yeah fair enough, I don't mind renaming it

openpype/plugins/publish/collect_slate_global.py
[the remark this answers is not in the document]
fabiaserra: Looking at this better, I'm realizing that all the slate data is already under a `slateGlobal` key on the instance and not flat (`slateGlobal` -> `slate_common_data`). The commented line is only to update the slate common data with any possible `customData` so the suggested change is not equivalent

Having said that, as I'm going to rename the plugin I will also be renaming this key to `htmlSlate`

openpype/plugins/publish/collect_slate_global.py
fabiaserra: fair enough, thanks for the info, removed!

openpype/plugins/publish/extract_slate_global.py
fabiaserra: so it won't "crash" and it can be ignored?

Although as long as we add html2image as a dependency on poetry it won't error out just by importing it like any other dependencies. Not sure I follow on the advantage of separating into a script

openpype/plugins/publish/extract_slate_global.py
fabiaserra: done!

mkolar:
Folks I'm sorry but I don't see this being merged. In principle there's no problem, but the implementation doesn't lend itself very well to getting into the main repo. @fabiaserra is also correct that we have a similar thing internally, which is currently only available to clients. Not necessarily due to it being super secret, but it makes more sense as an addon. I'd recommend taking this code and turning it into addon itself actually

fabiaserra:
> Folks I'm sorry but I don't see this being merged. In principle there's no problem, but the implementation doesn't lend itself very well to getting into the main repo. @fabiaserra is also correct that we have a similar thing internally, which is currently only available to clients. Not necessarily due to it being super secret, but it makes more sense as an addon. I'd recommend taking this code and turning it into addon itself actually

Yeah I figured that would be the case, thank you @mkolar!

PR #4952, opened 2023-05-09
Drop-down menu to list all representations by family in create placeholder

friquette:
## Changelog Description
Currently in the create placeholder window, we need to write the representation manually. This replace the text field by an enum field which updates himself each time the family enum is changed. The new representation enum then contain all the representations for the selected family.
The representations are collected from the loaders of the software.

## Additional info
Here's an example on Maya.
For the "animation" family:
image
For the "render" family:
image

### DIFF
--- a/openpype/pipeline/workfile/workfile_template_builder.py
+++ b/openpype/pipeline/workfile/workfile_template_builder.py
@@ -47,7 +47,6 @@ from openpype.pipeline.load import (
 
 from openpype.pipeline.create import (
     discover_legacy_creator_plugins,
-    discover_creator_plugins,
     CreateContext,
 )
 
@@ -1250,16 +1249,15 @@ class PlaceholderLoadMixin(object):
         loader_items = list(sorted(loader_items, key=lambda i: i["label"]))
         options = options or {}
 
-        # Get families and sort them alphabetically
+        # Get families from all loaders excluding "*"
         all_loaders = discover_loader_plugins()
-        families = []
+        families = set()
         for loader in all_loaders:
-            families.extend(loader.families)
+            families.update(loader.families)
+        families.discard("*")
 
-        families = list(set(families))
-        if "*" in families:
-            families.remove("*")
-        families.sort()
+        # Sort for readability
+        families = list(sorted(families))
 
         return [
             attribute_definitions.UISeparatorDef(),
--- a/openpype/pipeline/workfile/workfile_template_builder.py
+++ b/openpype/pipeline/workfile/workfile_template_builder.py
@@ -24,6 +24,7 @@ from openpype.client import (
     get_linked_assets,
     get_representations,
 )
+from openpype.pipeline.load.plugins import discover_loader_plugins
 from openpype.settings import (
     get_project_settings,
     get_system_settings,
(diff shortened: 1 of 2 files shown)


friquette:
This PR is linked to https://github.com/ynput/OpenPype/pull/4928 and must be merged after it

openpype/pipeline/workfile/workfile_template_builder.py
```diff
@@ -1246,6 +1247,31 @@ def get_load_plugin_options(self, options=None):
...
+        attributes = ['representations', 'extensions', 'extension']
+        for loader in all_loaders:
+            for family in loader.families:
+                if family == (options.get('family') or families[0]):
+                    for attribute in attributes:
+                        if hasattr(loader, attribute):
+                            representations.update(getattr(loader, attribute))
+
+        representations.discard("*")
+        representations = list(sorted(representations))
```
BigRoy, line 1273: I would personally move `representations = set()` definition down closer to where it's used. But not sure if that's just my personal preference?

```suggestion
        for loader in all_loaders:
            families.update(loader.families)
        families.discard("*")

        # Sort for readability
        families = list(sorted(families))

        # Get representations for the default family,
        # or the first one if no default
        attributes = ['representations', 'extensions', 'extension']
        representations = set()
        for loader in all_loaders:
            for family in loader.families:
                if family == (options.get('family') or families[0]):
                    for attribute in attributes:
                        if hasattr(loader, attribute):
                            representations.update(getattr(loader, attribute))

        representations.discard("*")
        representations = list(sorted(representations))
```

Also, since we're re-using the logic. Shouldn't we move the `get_representation_by_family` logic into some sort of library function?

openpype/tools/workfile_template_build/window.py
```diff
@@ -240,3 +253,68 @@ def showEvent(self, event):
...
+                origin_widget_label,
+                target_widget_label
+            )
+        )
+
+        self._content_layout.addWidget(widget, 0)
+        self._content_layout.addStretch(1)
+        self._attr_defs_widget = widget
+
+    def get_representation_by_family(self, family):
```
BigRoy, line 297:
```suggestion
    def get_representations_by_family(self, family):
```
Does it make more sense here to keep `representations` plural?

openpype/tools/workfile_template_build/window.py
```diff
@@ -167,6 +168,18 @@ def _create_option_widgets(self, plugin, options=None):
...
+        for enum_widget in enum_widgets:
+            if enum_widget.attr_def.label == "Family":
+                enum_widget.value_changed.connect(
+                    lambda value: self.on_enum_change(
+                        value,
+                        plugin,
+                        "Family",
+                        "Representation name"
+                    )
+                )
```
BigRoy, line 181: @iLLiCiTiT what do you think of this logic? Is there a better way?

BigRoy (commented):
Some notes.

openpype/tools/workfile_template_build/window.py
friquette: Yes indeed it does

openpype/pipeline/workfile/workfile_template_builder.py
friquette: I agree with converting this method into a library function. But where could it fit in ?

iLLiCiTiT:
I'm afraid this PR won't be approved. I don't think it make sense to have representation as enum in these filters. Most of loaders have representations set to `*` which is not handled and is a huge limitation. Most of loaders are filtering by extension at this moment.

Note: We plan that loaders will have a method which returns true/false to filter representations so they won't have any filtering on their public attributes.

openpype/tools/workfile_template_build/window.py
iLLiCiTiT: I don't think it's good. It's loader plugin specific logic that won't work with other plugin types.

mkolar:
I agree that as much as this seems to make life easier, it takes away flexibility and functionality.

PR #4953, opened 2023-05-09
Add shortcut to open template for current context

ClementHector:
## Changelog Description
I added a shortcut to open the template corresponding to the current context. This allows for faster implementation, so template corrections are faster.
I also reorganized the maya menu so that it is in the same order as nuke.

## Testing notes:
1. Add a template for your context in the settings
2. Click on open template
image


openpype/hosts/nuke/api/workfile_template_builder.py
```diff
@@ -33,6 +34,30 @@
...
+            return
+
+        result = nuke.ask(
+            "This will replace current scene with template. Continue?"
+        )
+        if not result:
+            return
+
+        print("opening template {}".format(template_path))
+        nuke.scriptOpen(template_path)
```
tokejepsen, line 59: This does not clear the current scene. Maybe use [`open_file`](https://github.com/ynput/OpenPype/blob/bebaeee66a626196f4cb99df4728a7fa3f5b8010/openpype/hosts/nuke/api/workio.py#L22-L32)

tokejepsen (changes requested):

openpype/hosts/nuke/api/workfile_template_builder.py
ClementHector: Do you think we should do the same for Maya?

tokejepsen: Yeah, actually you could do this in `pipeline/workfile/workfile_template_builder.py` and use Qt for showing the dialogs, then its cross host and less code.

tokejepsen (changes requested):

tokejepsen:
@ClementHector when this is ready for review, just flip it to "Ready for review".

tokejepsen (approved):
Tested in Maya 2023 and Nuke 11 successfully.

openpype/pipeline/workfile/workfile_template_builder.py
```diff
@@ -37,7 +37,12 @@
     attribute_definitions,
 )
 from openpype.lib.attribute_definitions import get_attributes_keys
-from openpype.pipeline import legacy_io, Anatomy
+from openpype.pipeline import (
```
iLLiCiTiT, line 40: This should be tested in Nuke 12 or Maya 2020 as this will probably cause recursion import error.

iLLiCiTiT (changes requested):
Please move the UI handling from `openpype/pipeline/workfile/workfile_template_builder.py`. The function `open_template` should raise specific exception that can be easily handled.

openpype/pipeline/workfile/workfile_template_builder.py
```diff
@@ -533,6 +538,50 @@ def rebuild_template(self):
 
         self.clear_shared_populate_data()
 
+    def open_template(self):
```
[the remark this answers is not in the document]
iLLiCiTiT: Also the host handling like `openpype.hosts.{}.api.lib` is just wrong. It should not expect specific structure in host integration which is inside openpype/hosts. There are already some host integrations that are not there.

openpype/pipeline/workfile/workfile_template_builder.py
```diff
@@ -533,6 +538,50 @@ def rebuild_template(self):
...
+        result = message_window.message(
+            title="Opening template",
+            message="Caution! This will overwrite your current scene.\n"
+            "Do you want to continue?",
+            parent=main_window,
+            level="ask",
+        )
+
+        if result:
+            host = registered_host()
```
iLLiCiTiT, line 582: There is `self.host` on this class. Why to use `registered_host()`?

openpype/pipeline/workfile/workfile_template_builder.py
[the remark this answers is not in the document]
ClementHector: According to you, it is not possible to do this method in open/pipeline? It must be in each host?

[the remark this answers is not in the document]
iLLiCiTiT: I think there was a response here but disappeared. Question was if it should be handled in each host and answer is yes if it should show a Qt message. Not all hosts can use Qt directly where their backend lives.

mkolar:
@ClementHector this is looking great if we can address @iLLiCiTiT comments. Loving the idea!

iLLiCiTiT:
> Where should the common interface of the different DCCs be located?

For example in `openpype/tools/workfile_template_build`?

openpype/widgets/message_window.py
```diff
@@ -28,23 +32,32 @@ def _info(self):
...
     def _critical(self):
-        self.setWindowTitle(self.title)
         rc = QtWidgets.QMessageBox.critical(
             self, self.title, self.message)
         if rc:
             self.exit()
 
+    def _ask(self):
+        self.answer = None
+        rc = QtWidgets.QMessageBox.question(
```
BigRoy, line 48: Is there a  reason we choose `ask` over `question` even though the Qt API method is called `question`? I don't care too much. I just noticed the others do match the Qt counterpart and was wondering.

tokejepsen:
@ClementHector did you get a chance to progress with this?

openpype/widgets/message_window.py
tokejepsen: Yeah, think we should change it to `_question`.

ClementHector:
I’ll try ;)

ClementHector:
Before the first review, the interface was in each host and without Qt. I feel this version is closer to the expectations of @iLLiCiTiT ?

iLLiCiTiT:
> Before the first review, the interface was in each host and without Qt. I feel this version is closer to the expectations of @iLLiCiTiT ?

No it is not. There is still import of Qt inside pipeline content. There is used message widget inside pipeline function. All pipeline functionality must be headless. The handling of the exceptions must be inside the hosts, so they can define how the UI is shown.

tokejepsen:
@ClementHector have you need able to progress with this?

tokejepsen:
Closing this in favour of https://github.com/ynput/OpenPype/pull/6181

PR #4956, opened 2023-05-10
Unreal: Fix camera frame range

simonebarbieri:
## Changelog Description
Fix the frame range of the level sequence for the Camera in Unreal.

## Testing notes:
1. Load a camera in Unreal.
2. Open the level sequence that has been created.
3. Check if the frame range of the sequence is the one set in the project manager.


LiborBatek (approved):
After loading of ```Camera``` it correctly sets frame range according to Project DB.

in this case correctly set to frame range ```1001 - 1050```

image

PR #4958, opened 2023-05-10
Feature: Publish kitsu note even with no review setting + integrate kitsu note families settings

Sharkitty:
## Changelog Description
Added two settings to `Integrate Kitsu Note`:
- `Publish note even with no review` -> If true, it is not required to publish a review for kitsu notes to be published. If false, a review must be published to publish kitsu note, regardless on `Families` settings.
- `Families` -> Choose for which families the `Integrate Kitsu Note` will run.

## Testing notes:
- Go in project settings.
- Set `Families` and `Publish note even with no review` as required.
- Publish an asset.


EmberLightVFX:
Could give an example on when you would want to publish a note but not a review?
And how do you publish a note to Kitsu without a review?

Sharkitty:
> 

This could be used when an artists is doing a work in progress and wants to publish his workfile (with no other subsets) as a backup, but we still want a kitsu note to be posted in this case. There could also be cases in which a review aren't really necessary like a subset bank inside an asset. It would just look like a bunch of stuff stacked on each others on the review, yet we still need to track its progress.

To publish a kitsu note without review, you need to edit the `families` setting (which is part of this PR) and choose for what families kitsu note should be posted. With the example above you could set `workfile` as one of the families as well as setting `Publish note even with no review` to True, then you would always get a kitsu note when a workfile is published, even if there's no review.

Minkiu (approved):
Code looks good to me

EmberLightVFX:
There is never really a review file connected to the workfile so to post it you would always have to have `Publish note even with no review` turned on.
I don't see the need to have the switch `Publish note even with no review` at all really. To have options to pick what families should post notes and reviews to Kitsu would be enough.

Sharkitty:
> 

In the current implementation, if there is no `review` instance, even if you change `families` setting it won't publish a kitsu note. This happens because there's a check that skips publishing the kitsu note when there's no review instance. So changing families does not seem to be sufficient.

Therefore I'm trying to add the ability to do so, without completely removing this check, as I think it's here for a good reason and shouldn't be removed. Hence why I implemented it as a setting.

That being said I understand the skepticism, and criticism about my PR. I think I will need to discuss it with @Tilix4 and I hope we can either find a consensus or a better solution to the problem I'm trying to solve.

mkolar:
I don't know man. If you're just publishing the workfile and no review with it, but still want to make a comment about your progress, it feels that you should go straight to Kits web UI for that.

Sharkitty:
Considering the concerns regarding this PR, I discussed with @Tilix4 and we decided to close this PR.
My apologies for the inconvenience.

Issue #4949, opened 2023-05-08
Bug: Representations right click error - 3.15.6-nightly.1 next-minor


Jeremyball:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

Right click on subset produces an error.

### Expected Behavior:

Produce a list of available representations.

### Version

3.15.6-nightly.1

### What platform you are running OpenPype on?

Windows

### Steps To Reproduce:

Created new project on Ayon Server

In Editor Tab
- Create folder assets
- Create asset 
- Create modeling task

![image](https://user-images.githubusercontent.com/54645515/236957346-7bec9722-5193-42af-94cf-fa7234319093.png)

Launch Maya via Launcher

- Create sphere - ball_GEO
- Grouped under geo_GRP 
- Create modeling publish
- Add geo_GRP to set
- Publish(all checks green, publish is good)

New Maya scene

- Loader
- Navigate to asset
- Right click subset
- No menu - error

![image](https://user-images.githubusercontent.com/54645515/236957608-d6c1a29d-c980-4f51-af1e-f08659fce826.png)

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

### Relevant log output:

```shell
# Traceback (most recent call last):
#   File "E:\ayon\OpenPype-release-next-minor\build\exe.win-amd64-3.9\openpype\tools\loader\widgets.py", line 481, in on_context_menu
#     self._repre_contexts_for_loaders_filter(items)
#   File "E:\ayon\OpenPype-release-next-minor\build\exe.win-amd64-3.9\openpype\tools\loader\widgets.py", line 350, in _repre_contexts_for_loaders_filter
#     for repre_doc in repre_docs:
#   File "E:\ayon\OpenPype-release-next-minor\build\exe.win-amd64-3.9\openpype\client\server\entities.py", line 502, in get_representations
#     fields = representation_fields_v3_to_v4(fields, con)
#   File "E:\ayon\OpenPype-release-next-minor\build\exe.win-amd64-3.9\openpype\client\server\conversion_utils.py", line 603, in representation_fields_v3_to_v4
#     fields |= {
# TypeError: unsupported operand type(s) for |=: 'list' and 'set'
```

### Additional context:

Windows
3.15.6-nightly.1 
Banch: release/next-minor
Maya 2023.2

[cuID:[OP-5881](https://app.clickup.com/t/861mrwpjf)]

sjt-rvx:
[...]

PR #4959, opened 2023-05-10
bugfix - A collection of tiny tweaks to get ayon running with maya

sjt-rvx:
This is a collection of small tweaks to get maya (and ayon in general) to not error so I can test it. Most of this is names of things have changed just slightly.

### DIFF
--- a/openpype/client/server/conversion_utils.py
+++ b/openpype/client/server/conversion_utils.py
@@ -584,7 +584,9 @@ def representation_fields_v3_to_v4(fields, con):
     representation_attributes = con.get_attributes_for_type("representation")
 
     output = set()
-    for field in fields:
+    fields = set(fields)
+    iterfields = fields.copy()
+    for field in iterfields:
         if field in ("type", "schema"):
             continue
 
--- a/openpype/hosts/maya/api/lib.py
+++ b/openpype/hosts/maya/api/lib.py
@@ -3928,7 +3928,7 @@ def get_capture_preset(task_name, task_type, subset, project_settings, log):
     }
 
     plugin_settings = project_settings["maya"]["publish"]["ExtractPlayblast"]
-    if plugin_settings.get("profiles", []):
+    if plugin_settings["profiles"]:
         profile = filter_profiles(
             plugin_settings["profiles"],
             filtering_criteria,
--- a/openpype/hosts/maya/api/lib.py
+++ b/openpype/hosts/maya/api/lib.py
@@ -3981,7 +3981,7 @@ def create_rig_animation_instance(nodes, context, namespace, log=None):
     )
     assert roots, "No root nodes in rig, this is a bug."
 
-    asset = get_current_asset_name()  # remove legacy_io call
+    asset = get_current_asset_name()
     dependency = str(context["representation"]["_id"])
 
     if log:
--- a/openpype/hosts/maya/api/lib.py
+++ b/openpype/hosts/maya/api/lib.py
@@ -3928,7 +3928,7 @@ def get_capture_preset(task_name, task_type, subset, project_settings, log):
     }
 
     plugin_settings = project_settings["maya"]["publish"]["ExtractPlayblast"]
-    if plugin_settings["profiles"]:
+    if plugin_settings.get("profiles", []):
         profile = filter_profiles(
             plugin_settings["profiles"],
             filtering_criteria,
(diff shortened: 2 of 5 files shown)


openpype/client/server/conversion_utils.py
```diff
@@ -584,7 +584,9 @@ def representation_fields_v3_to_v4(fields, con):
     representation_attributes = con.get_attributes_for_type("representation")
 
     output = set()
-    for field in fields:
+    fields = set(fields)
+    iterfields = fields.copy()
+    for field in iterfields:
```
iLLiCiTiT, line 589: Why is this needed?

openpype/hosts/maya/api/lib.py
```diff
@@ -3981,7 +3981,7 @@ def create_rig_animation_instance(nodes, context, namespace, log=None):
     )
     assert roots, "No root nodes in rig, this is a bug."
 
-    asset = legacy_io.Session["AVALON_ASSET"]
+    asset = get_current_asset_name()  # remove legacy_io call
```
iLLiCiTiT, line 3984:
```suggestion
    asset = get_current_asset_name()
```

openpype/hosts/maya/api/lib.py
```diff
@@ -3928,7 +3928,7 @@ def get_capture_preset(task_name, task_type, subset, project_settings, log):
     }
 
     plugin_settings = project_settings["maya"]["publish"]["ExtractPlayblast"]
-    if plugin_settings["profiles"]:
+    if plugin_settings.get("profiles", []):
```
iLLiCiTiT, line 3931: Please don't make backwards compatibility because of AYON. AYON is not production ready and crashes because of inconsistency is only way how to find out we have something off...
```suggestion
    if plugin_settings["profiles"]:
```

openpype/lib/applications.py
```diff
@@ -1516,7 +1516,7 @@ def prepare_app_environments(
     tool_by_group_name = collections.defaultdict(dict)
     if asset_doc:
         # Make sure each tool group can be added only once
-        for key in asset_doc["data"].get("tools_env") or []:
+        for key in asset_doc["data"].get("tools") or []:
```
iLLiCiTiT, line 1519: This would break openpype, can't be used, we have to fix it other way.

openpype/hosts/maya/api/lib.py
iLLiCiTiT, line 3931: Please don't make backwards compatibility because of AYON. AYON is not production ready and crashes because of inconsistency is only way how to find out we have something off...
```suggestion
    if plugin_settings["profiles"]:
```

openpype/client/server/conversion_utils.py
BigRoy: I believe it was to resolve this bug reported here: https://github.com/ynput/OpenPype/issues/4949#issue-1701076658 which fails on [this line](https://github.com/ynput/OpenPype/blob/release/next-minor/openpype/client/server/conversion_utils.py#L603-L606) due to `fields` not being a `set`. 

A copy is likely made since it appears that the `fields` variable is being update as it is being iterated over. So it should basically be this:
```python
fields = set(fields)
for field in fields.copy():
    ...
```

iLLiCiTiT: Isn't the bug that there should be `output |= ...` instead of copying source fields?

openpype/hosts/maya/api/lib.py
tokejepsen, line 3931:
```suggestion
    if plugin_settings["profiles"]:
```

There is a fix coming for the addon settings [here](https://github.com/ynput/ayon-addons/pull/10)

openpype/hosts/maya/api/lib.py
tokejepsen, line 3984:
```suggestion
    asset = get_current_asset_name()
```

Dont think we need to comment here :)

tokejepsen (changes requested):

openpype/client/server/conversion_utils.py
iLLiCiTiT, line 589: @sjt-rvx please change this back
```suggestion
    for field in iterfields:
```

 and fix line 605 to:
```python
            output |= {
```

sjt-rvx:
Ok, sure thing. 

I didn't realize that my repo had disabled push; first time for me dealing with github pull requests.

PR #4963, opened 2023-05-11
Unreal: Fix render extraction

simonebarbieri:
## Changelog Description
Fix a problem with the extraction of renders in Unreal.

## Additional info
Render instances in unreal were already created in `collect_render_instances.py` as render and review instances, so the extraction of the render is handled by `Extract Review`. The `extract_render.py` extractor was just creating a duplicate instance.
Also fixed start and end frame to be `int`. That was creating another problem during the extraction of the review.

## Testing notes:
1. After loading a layout and a camera, create a render instance from the creator.
2. Render the scene.
3. Try publishing the render.


LiborBatek (approved):
Rendered images gets succesfully published. All works as expected.

image

Also the h264 mov file got created when published.

PR #4972, opened 2023-05-15
Unreal: Fix transform when loading layout to match existing assets

simonebarbieri:
## Changelog Description
Fixed transform when loading layout to match existing assets.

## Additional info
Changed the way the transform is applied to match the one for the regular layout loading.

## Testing notes:
1. In Unreal, load assets that have been published in a layout.
2. Add the assets to a level and save it.
3. Open the loader, select the layout, and select **Load Layout on Existing Scene**.
4. The transform for the assets should be correct.

### DIFF
--- a/openpype/hosts/unreal/plugins/load/load_layout_existing.py
+++ b/openpype/hosts/unreal/plugins/load/load_layout_existing.py
@@ -120,9 +120,9 @@ class ExistingLayoutLoader(plugin.Loader):
         transform = lasset.get('transform_matrix')
         basis = lasset.get('basis')
 
-        t = self._transform_from_basis(transform, basis)
+        computed_transform = self._transform_from_basis(transform, basis)
 
-        actor.set_actor_transform(t, False, True)
+        actor.set_actor_transform(computed_transform, False, True)
 
     @staticmethod
     def _get_fbx_loader(loaders, family):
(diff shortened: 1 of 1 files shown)


openpype/hosts/unreal/plugins/load/load_layout_existing.py
```diff
@@ -89,67 +89,40 @@ def _get_current_level():
...
-        filename = import_data.get_first_filename()
-        path = Path(filename)
 
-        transform = self._get_transform(
-            path.suffix, import_data, lasset)
+        transform = lasset.get('transform_matrix')
+        basis = lasset.get('basis')
 
-        actor.set_actor_transform(transform, False, True)
+        t = self._transform_from_basis(transform, basis)
```
Minkiu, line 123: I'd prefer to have  proper variable name instead of `t`.

Minkiu:
I was wondering, since we removed the `fbx` and `abc` matrix changes, if they are done somewhere else or they are no longer needed?

openpype/hosts/unreal/plugins/load/load_layout_existing.py
simonebarbieri: That is very fair! Fixed it!

Minkiu (approved):

LiborBatek (approved):
All works ok.

PR #4976, opened 2023-05-16
Enhancement: Resolve integration tweaks

BigRoy:
## Changelog Description

- Contains some PySide6 fixes
- Raise ApplicationLaunchError in prelaunchhook instead of assertion
- Add a Load Media loader which loads into the media pool for its metadata instead of storing it within a timeline like Load clip does.
    - As far as I know this is basically what setting the `lib.pype_marker_workflow` variable to `False` would do but I couldn't get that to work and I expected there's likely a reason for it to be True to begin with. (And I couldn't find anyway of disabling it without changing it hardcoded anyway).
    - There are multiple bugs with the existing "Load clip" logic. Since I didn't want it managed in the timeline but in the media pool for my use cases I started making my own loader anyways. Some of the bugs for load clip were:
      - it loads all files from the publish folder instead of just the one representation.
      - It uses deprecated API function `MediaStorage.AddItemsToMediaPool` in Resolve.
      - Whenever you delete the media pool entry then the timeline entry is still collected but fails to work correctly in manager
      - Timeline clip can't be removed via the scene inventory (not that hard too implement however)
    - [ ] **TODO:** Still need to fix the support for loading sequences correctly.
- Tweak OpenPype menu structure/order to match more with other hosts:
image

## Testing notes:
1. Launch Resolve
2. Use the new Load Media loader + use the old Load Clip loader


openpype/hosts/resolve/plugins/load/load_media.py
```diff
@@ -0,0 +1,177 @@
...
+            "schema": "openpype:container-2.0",
+            "id": AVALON_CONTAINER_ID,
+            "name": str(name),
+            "namespace": str(namespace),
+            "loader": str(self.__class__.__name__),
+        })
+
+        item.SetMetadata(lib.pype_tag_name, json.dumps(data))
+
+        timeline = options.get("timeline", self.timeline)
```
jakubjezek001, line 91: Are you planning to add the options? At the moment it is added to timeline by default - is happening to me once I am on Edit tab with an active timeline.

openpype/hosts/resolve/plugins/load/load_media.py
```diff
@@ -0,0 +1,177 @@
...
+                root=media_pool.GetRootFolder()
+            )
+
+        # TODO: Correctly set the media path for image sequences
+        # frames = [
+        #     "path/to/frame.1001.exr",
+        #     "path/to/frame.1002.exr"
+        # ]
+        path = get_representation_path(representation)
+        items = media_pool.ImportMedia(path)
```
jakubjezek001, line 71: Yes, this is a great approach. We should definitely fix this before the merge. I can help ;)

openpype/tools/creator/widgets.py
```diff
@@ -99,7 +99,7 @@ def __init__(self):
 
     def validate(self, text, pos):
         results = super(SubsetNameValidator, self).validate(text, pos)
-        if results[0] == self.Invalid:
```
jakubjezek001, line 102: @iLLiCiTiT can you confirm these changes are ok?

openpype/vendor/python/common/qargparse.py
```diff
@@ -629,7 +629,9 @@ def set_current(current):
                     )
 
             qindex = model.index(index, 0, QtCore.QModelIndex())
-            smodel.setCurrentIndex(qindex, smodel.ClearAndSelect)
```
jakubjezek001, line 632: @iLLiCiTiT an you confirm these changes are ok?

jakubjezek001 (changes requested):
Those are great weeks. Thank you for your time on this! Once we solve the sequence loading and add the solution also to the Timeline loader we are good to merge.

openpype/tools/creator/widgets.py
iLLiCiTiT: Yes, ok.

openpype/vendor/python/common/qargparse.py
iLLiCiTiT: Yes.

mkolar:

We're closing it down, but we'll he happy for a new PR to [ynput/ayon-resolve](https://github.com/ynput/ayon-resolve) repository once it's up.

PR #4977, opened 2023-05-16
fix the bug of fbx loaders in Max

moonyuet:
## Changelog Description
bug fix of fbx loaders for not being able to parent to the CON instances while importing cameras(and models) which is published from other DCCs such as Maya.

## Additional info
n/a

## Testing notes:
1. Launch Max via Openpype
2. Click "Load...."
3. Load Camera/Model Instance which is published from other DCCs such as maya
4. Also try to load Camera/Model Instance which is published within 3dsMax.
5. Both should free with error shown in the log in 3dsMax


LiborBatek (commented):
The ```Load``` works ok in general.

Problem being wrong Y axis orientation caused by different coord system in Maya (Y up).
Could be just setting the option in FBX importer.

The highlighted Camera is wrong oriented and the blue one being correct (originated from Alembic)
image

moonyuet:
> The `Load` works ok in general.
> 
> Problem being wrong Y axis orientation caused by different coord system in Maya (Y up). Could be just setting the option in FBX importer.
> 
> The highlighted Camera is wrong oriented and the blue one being correct (originated from Alembic) image

I updated the PR, the axis should be fixed!

antirotor (approved):
code seems great!

LiborBatek (approved):
Now all works as it should.

The camera in FBX perfectly matches the Alembic representation.

image

PR #4978, opened 2023-05-16
Bugfix/frame variable fix

Zipodod:
Renamed variables to match OpenPype terminology to reduce confusion and add consistency.

### DIFF
--- a/openpype/hosts/maya/api/lib.py
+++ b/openpype/hosts/maya/api/lib.py
@@ -319,15 +319,15 @@ def collect_animation_data(fps=False):
     handle_start = cmds.playbackOptions(query=True, animationStartTime=True)
     handle_end = cmds.playbackOptions(query=True, animationEndTime=True)
 
-    handle_start = frame_start - handle_start
-    handle_end = handle_end - frame_end
+    frame_start_handle = frame_start - handle_start
+    frame_end_handle = handle_end - frame_end
 
     # build attributes
     data = OrderedDict()
     data["frameStart"] = frame_start
     data["frameEnd"] = frame_end
-    data["handleStart"] = handle_start
-    data["handleEnd"] = handle_end
+    data["handleStart"] = frame_start_handle
+    data["handleEnd"] = frame_end_handle
     data["step"] = 1.0
 
     if fps:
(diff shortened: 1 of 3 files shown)


openpype/hosts/maya/api/lib.py
```diff
@@ -319,15 +319,15 @@ def collect_animation_data(fps=False):
     handle_start = cmds.playbackOptions(query=True, animationStartTime=True)
     handle_end = cmds.playbackOptions(query=True, animationEndTime=True)
 
-    handle_start = frame_start - handle_start
-    handle_end = handle_end - frame_end
+    frame_start_handle = frame_start - handle_start
+    frame_end_handle = handle_end - frame_end
```
BigRoy, line 323: This is actually incorrect naming. In this case it's actually the `handle_start` variables above that are `frame_start_handle`. So the names here are reversed.

This is especially noticeable in the assignment below because `handleStart` should not equal `frame_start_handle` but should equal (by name) with `handle_start`.

So I'd revert both these variable name changes here and below and then change the variables above queried from maya `animationStartTime` and `animationEndTime` to be the `frame_start_handle`, etc.

EmberLightVFX: By just looking at this, shouldn't it be `handle_end + frame_end`, + and not - ?

EmberLightVFX:
Is this something that should be looked into all DCCs?

tokejepsen (approved):
LGTM

EmberLightVFX:
My review is stuck as Pending so I'm writing it here as a comment so it can be seen.
In [openpype/hosts/maya/api/lib.py](https://github.com/ynput/OpenPype/pull/4978/files/987b7acf02e2744a136e511cd56424cc43c8cb0c#diff-0f94b40dc0b348d7194c61cb01a57668eb3490d22cdbff89d37dd552287fccad) line 325:
This should be `handle_end = frame_end_handle + frame_end`, `+` instead of `-`, right?

BigRoy:
> My review is stuck as Pending so I'm writing it here as a comment so it can be seen.
In [openpype/hosts/maya/api/lib.py](https://github.com/ynput/OpenPype/pull/4978/files/987b7acf02e2744a136e511cd56424cc43c8cb0c#diff-0f94b40dc0b348d7194c61cb01a57668eb3490d22cdbff89d37dd552287fccad) line 325:
This should be handle_end = frame_end_handle + frame_end, + instead of -, right?

No, the code looks ok.

Say your frame range is 1005 - 1015 and you have start handles of 5 and end handles of 5. These are the values:
```python
frame_start = 1005
frame_end = 1015
handle_start = 5
handle_end = 5
frame_start_handle = 1000
frame_end_handle = 1020
```

As such `handle_end = frame_end_handle - frame_end` to compute the handle value that is `5`.

PR #4983, opened 2023-05-17
AfterEffects: set frame range and resolution

kalisp:
## Changelog Description
Frame information (frame start, duration, fps) and resolution (width and height) is applied to selected composition from Asset Management System (Ftrack or DB) automatically when published instance is created.

It is also possible explicitly propagate both values from DB to selected composition by newly added menu buttons.

## Additional info
Paragraphs of text giving context of additional technical information or code examples.

!! Requires update of extension.

## Testing notes:
aftereffects_extension
1. set different values in `Composition Settings` in resolution and frame start and duration
2. create publishable instance from this updated composition
3. compare modified values, should match what is in DB
4. Do this also explicitly by newly added buttons.


openpype/hosts/aftereffects/api/extension/jsx/hostscript.jsx
```diff
@@ -347,38 +347,76 @@ function deleteItem(item_id){
...
     }
+
+    app.beginUndoGroup('change comp properties');
+        if (frameStart && framesCount && frameRate){
+            comp.displayStartFrame = frameStart;
+            comp.duration = framesCount / frameRate;
+            comp.frameRate = frameRate;
+        }
+        if (width && height){
+            var widthOld = comp.width;
```
iLLiCiTiT, line 390: Should these all be `let` instead of `var` (Global var states are dangerous)?

64qam (approved):

PR #4985, opened 2023-05-17
Blender: Fix Alembic loading

simonebarbieri:
## Changelog Description
Fixed problem occurring when trying to load an Alembic model in Blender.

## Testing notes:
1. Open Blender.
2. Try loading any alembic model.


LiborBatek (approved):
When loading ```Alembic``` representation of ```model``` and ```pointcache``` families all correctly loads into the scene.

image

PR #4986, opened 2023-05-17
Publish: Enhance automated publish plugin settings

iLLiCiTiT:
## Changelog Description
Added plugins option to define settings category where to look for settings of a plugin and added public helper functions to apply settings `get_plugin_settings` and `apply_plugin_settings_automatically`.

## Additional info
Plugin has no way how to tell automated settings apply function where to look for it's settings. We allow 2 options, one is to look for settings into current host (we should maybe remove that completelly?), second is to use 3rd parent folder name from plugin file -> we're lucky it somehow worked.

We already have option to implement `apply_settings` class method, but then developer looses the option to automatically apply the settings -> that's why those functions were made as public, so it is at least possible to reuse them.

In most of cases all we need to define is the first level of settings where to look, for that cases was added option to add `settings_category` attribute on a plugin, which tell where to look for settings. Other ways where to look for settings are ignored if this attribute is available on a plugin.
```python
class MyDeadlinePlugin(pyblish.api.ContextPlugin):
    settings_category = "deadline"
```
In this example values from `project_settings["deadline"]["plugins"]["publish"][MyDeadlinePlugin]` would be applied automatically.

### Why we need this
In AYON the folder of an addon does not match key in settings (e.g. addon is `ayon_ftrack` but key in settings is `"ftrack"`).

## Testing notes:
Nothing should really change in this PR.
- Change settings of any publish plugin and check if they are applied.

### DIFF
--- a/website/docs/dev_publishing.md
+++ b/website/docs/dev_publishing.md
@@ -506,6 +506,67 @@ or the scene file was copy pasted from different context.
 #### *Known errors*
 When there is a known error that can't be fixed by the user (e.g. can't connect to deadline service, etc.) `KnownPublishError` should be raised. The only difference is that its message is shown in UI to the artist otherwise a neutral message without context is shown.
 
+### Plugins
+Plugin is a single processing unit that can work with publish context and instances.
+
+#### Plugin types
+There are 2 types of plugins - `InstancePlugin` and `ContextPlugin`.  Be aware that inheritance of plugin from `InstancePlugin` or `ContextPlugin` actually does not affect if plugin is instance or context plugin, that is affected by argument name in `process` method.
+
+```python
+import pyblish.api
+
+
+# Context plugin
+class MyContextPlugin(pyblish.api.ContextPlugin):
+    def process(self, context):
+        ...
+
+# Instance plugin
+class MyInstancePlugin(pyblish.api.InstancePlugin):
+    def process(self, instance):
+        ...
+
+# Still an instance plugin
+class MyOtherInstancePlugin(pyblish.api.ContextPlugin):
+    def process(self, instance):
+        ...
+```
+
+#### Plugin filtering
+By pyblish logic, plugins have predefined filtering class attributes `hosts`, `targets` and `families`. Filter by `hosts` and `targets` are filters that are applied for current publishing process. Both filters are registered in `pyblish` module, `hosts` filtering may not match OpenPype host name (e.g. farm publishing uses `shell` in pyblish). Filter `families` works only on instance plugins and is dynamic during publish process by changing families of an instance.
+
+All filters are list of a strings `families = ["image"]`. Empty list is invalid filter and plugin will be skipped, to allow plugin for all values use a start `families = ["*"]`. For more detailed filtering options check [pyblish documentation](https://api.pyblish.com/pluginsystem).
+
+Each plugin must have order, there are 4 order milestones - Collect, Validate, Extract, Integration. Any plugin below collection order won't be processed. for more details check [pyblish documentation](https://api.pyblish.com/ordering).
+
+#### Plugin settings
+Pyblish plugins may have settings. There are 2 ways how settings are applied, first is automated, and it's logic is based on function `filter_pyblish_plugins` in `./openpype/pipeline/publish/lib.py`, second is explicit by implementing class method `apply_settings` on a plugin.
+
+
+Automated logic is expecting specific structure of project settings `project_settings[{category}]["plugins"]["publish"][{plugin class name}]`. The category is a key in root of project settings. There are currently 3 ways how the category key is received.
+1. Use `settings_category` class attribute value from plugin. If `settings_category` is not `None` there is not any fallback to other way.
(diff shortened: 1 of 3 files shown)


BigRoy (commented):
Nice. I think it's an _ok_ hack to what was already quite luckily working in a hacky way. 

---

But I have to admit I've been confused so many times by the 'automatic settings' being applied if it matched a plug-in somewhere in some area of the docs, etc. As a plug-in developer it'd be much clearer if _every_ plug-in that is setting values from settings to visually implement the `apply_settings` override. OR at the very least maybe inherit from a base plug-in that implements `apply_settings` in such a way.

Which made me think I would much rather see:
```python
def apply_settings(self, project_settings):
    settings = project_settings[x][y][z]
    apply_plugin_settings_automatically(plugin, settings):
```

Or even this explicitly on the plug-ins:
```python
def apply_settings(self, project_settings):
    settings = get_plugin_settings(self, project_settings)
    apply_plugin_settings_automatically(plugin, settings)
```

The `filter_pyblish_plugins` way of automatically assigning the settings is just to obscure and hidden away to the plug-in developer I'd say. I just wish there was a way to go upstream in the plug-ins inheritance to understand why it's applying settings that way.

With that in mind, does it make sense to expand this further instead of cleaning it up? Like, if we're calling this somewhat of a hack, do we still want to keep on expanding it.

iLLiCiTiT:
> The filter_pyblish_plugins way of automatically assigning the settings is just to obscure and hidden away to the plug-in developer I'd say. I just wish there was a way to go upstream in the plug-ins inheritance to understand why it's applying settings that way.
> 
> With that in mind, does it make sense to expand this further instead of cleaning it up? Like, if we're calling this somewhat of a hack, do we still want to keep on expanding it.

The only way how not do it in a hacky way would be to change `pyblish`. Any other option would be a hack. We could define our default plugins but that's not much better and would require to change all existing plugins to use them.

> Or even this explicitly on the plug-ins:

So every existing plugin which have settings would have exactly these 4 lines (you forgot classmethod decorator) -> that's an overkill and again would require to change it in every existing plugin...

BigRoy:
> We could define our default plugins but that's not much better and would require to change all existing plugins to use them.

Technically it would require only to change the plugins which would actually apply settings dynamically, making it quite explicit that the plug-in is automatically retrieving settings. Like a mixin almost:
```
class Validator(InstancePlugin, OpenPypeAutoApplySettings):
    settings_category = "deadline"

    def process(self, instance):
        # ...
```

Even though it's refactoring quite a few plug-ins I do feel it might be quite a decent improvement in clarity as to what's happening? At least it's more visible that something is happening.

I agree I'm not a fan of adding 4 lines of code across 50% of the plug-ins but if it makes the road simpler towards why a plug-in is behaving a certain way or not.

We could even just say the attribute is called `autoload_settings_category = "deadline"` and whenever it finds that attribute that it's used. Then we drop the old hack and just add that attribute explicitly to the existing plug-ins. Again, explicit versus implicit?

iLLiCiTiT:
We have to keep backwards compatibility for some time. A lot of our studios using OpenPype have their plugins with custom settings and we can't just stop supporting such important logic without any transition time.

If we would add something similar to proposed solutions, it must be done in a way that if you don't change anything in custom addons it still must work the same way. That's more decision making, before any implementation, as that's important change, first question would probably be: Does it worth it? :)

> Or is there anywhere a documentation page where we have clearly documented that these settings get auto-applied? Maybe I should just read that a few times so I never forget it again. Maybe it's just a matter of having this in clear bold text in the documentation and me reading it often. :)

This is more reasonable note.

website/docs/dev_publishing.md
```diff
@@ -506,6 +506,67 @@ or the scene file was copy pasted from different context.
 #### *Known errors*
 When there is a known error that can't be fixed by the user (e.g. can't connect to deadline service, etc.) `KnownPublishError` should be raised. The only difference is that its message is shown in UI to the artist otherwise a neutral message without context is shown.
 
+### Plugins
+Plugin is a single processing unit that can work with publish context and instances.
+
+#### Plugin types
+There are 2 types of plugins - `InstancePlugin` and `ContextPlugin`.  Be aware that inheritance of plugin from `InstancePlugin` or `ContextPlugin` actually does not affect if plugin is instance or context plugin, that is affected by argument name in `process` method.
```
jakubjezek001, line 513: Is it still necessary to differentiate between those plugin parent classes if only what matter is name of input argument in process method? It seems to be confusing and redundant to support both at least.

iLLiCiTiT: It's not something we can affect until we use `pyblish.base`. This is more likely warning for developers (as it is in dev docs)...

jakubjezek001 (approved):
LGTM

website/docs/dev_publishing.md
BigRoy: > actually does not affect if plugin is instance or context plugin, that is affected by argument name in `process` method.

I'd actually argue that THIS is what is confusing in Pyblish, not the class names. It's an unfortunate side effect of Pyblish being backwards compatible that way.

PR #4987, opened 2023-05-17
Blender: Added setting for base unit scale

simonebarbieri:
## Changelog Description
A setting for the base unit scale has been added for Blender.
The unit scale is automatically applied when opening a file or creating a new one.

## Testing notes:
1. Set a default unit scale in Studio Settings: `project_settings/blender/base_file_unit_scale`.
2. Open any task with Blender.
3. Check if the unit scale has been applied:
image

### DIFF
--- a/openpype/hosts/blender/api/pipeline.py
+++ b/openpype/hosts/blender/api/pipeline.py
@@ -26,6 +26,8 @@ from openpype.lib import (
     emit_event
 )
 import openpype.hosts.blender
+from openpype.settings import get_project_settings
+
 
 HOST_DIR = os.path.dirname(os.path.abspath(openpype.hosts.blender.__file__))
 PLUGINS_DIR = os.path.join(HOST_DIR, "plugins")
@@ -122,12 +124,23 @@ def set_start_end_frames():
     scene.render.resolution_y = resolution_y
 
 
+def set_base_file_unit_scale():
+    project = os.environ.get("AVALON_PROJECT")
+    settings = get_project_settings(project)
+
+    unit_scale = settings.get("blender").get("base_file_unit_scale")
+
+    bpy.context.scene.unit_settings.scale_length = unit_scale
+
+
 def on_new():
     set_start_end_frames()
+    set_base_file_unit_scale()
 
 
 def on_open():
     set_start_end_frames()
+    set_base_file_unit_scale()
 
 
 @bpy.app.handlers.persistent
--- a/openpype/hosts/blender/api/pipeline.py
+++ b/openpype/hosts/blender/api/pipeline.py
@@ -85,6 +85,31 @@ def uninstall():
         ops.unregister()
 
 
+def show_message(title, message):
+    from openpype.widgets.message_window import Window
+    from .ops import BlenderApplication
+
+    BlenderApplication.get_app()
+
+    Window(
+        parent=None,
+        title=title,
+        message=message,
+        level="warning")
+
+
+def message_window(title, message):
+    from .ops import (
+        MainThreadItem,
+        execute_in_main_thread,
+        _process_app_events
+    )
+
+    mti = MainThreadItem(show_message, title, message)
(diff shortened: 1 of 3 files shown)


openpype/hosts/blender/api/pipeline.py
```diff
@@ -122,12 +124,23 @@ def set_start_end_frames():
...
+
+
 def on_new():
     set_start_end_frames()
+    set_base_file_unit_scale()
 
 
 def on_open():
     set_start_end_frames()
+    set_base_file_unit_scale()
```
BigRoy, line 143: Is it dangerous to just change the unit settings for an existing file? I can imagine things might start behaving differently, like a simulation or alike. The user not getting notified on these changes seems quite intrusive and hidden away that whatever they are opening suddenly gets changed?

The same question for `set_start_end_frames` actually - less dangerous likely, but as an artist reopening my scene later I wouldn't want it to be different when opening it again without any notification whatsoever.

Also, should there be a publish validator confirming the state of this setting as well? If someone would've opened the scene, then changed the setting, then saves and publishes it would just pass. But a reopen of the scene would "force it back" potentially also making it very hard to debug why that particular publish might have acted different than when reopening the scene and publishing again (due to it auto-resetting the value without any hint to the user)

simonebarbieri: That's a fair point! I guess it shouldn't really be necessary to change already existing files anyway, as probably they're already set. I'll change that!

simonebarbieri: I added a setting to make it optional to change the unit scale on opening. I've also added a warning message if opening a file and the unit scale is changed.

LiborBatek: > Is it dangerous to just change the unit settings for an existing file? I can imagine things might start behaving differently, like a simulation or alike. The user not getting notified on these changes seems quite intrusive and hidden away that whatever they are opening suddenly gets changed?

@BigRoy  here is excerpt directly from Blender doc...should quite safe in that respect
image

LiborBatek (changes requested):
```Set Unit Scale``` settings is correctly aplied in the Blender host.

However the Settings cant be set to smaller values...should allow user to set it like

```0.001``` if wanted.

now this is not possible
image

LiborBatek (approved):
All works ok. Also user is able to input more decimals then just 2.

image

antirotor:
I think this doesn't handle loading files from various sources - it wil set the scale for everything. Lets say you have alembics coming from Maya and alembics coming from Blender, wouldn't this break one or the other?

We don't store the information about units on representation for now so it involves a lot of guessing. But the way to mitigate it would be by setting this scale per source DCC.

When you load something with loader, you know the source of that representation and from that you can at least determine the application that was used to produce this representation. You can't know what was the unit setting there but at least you can limit it per imported data?

What I am trying to say is, that we should maybe leave the scene settings alone, and only change the scale of the loaded data (if this option is enabled). This will not influence any existing data in the scene, and it actually mimics the actual workflow of artists having to deal with stuff coming from different DCCs. What do you think?

mkolar:
>What I am trying to say is, that we should maybe leave the scene settings alone, and only change the scale of the loaded data (if this option is enabled). This will not influence any existing data in the scene, and it actually mimics the actual workflow of artists having to deal with stuff coming from different DCCs. What do you think? @simonebarbieri

Even though that's valid suggestion, why don't we then handle it in all places and all DCCs. I'd argue that it's a bit beyond the current scope and goes completely against the fact, that loader should care at all where the data is coming from. Better place would be ensuring correct scales across the DCCs in a cross host production, which is really down to the studio and we should just provide the tools to allow it as much as possible. This is a simple setting that is reasonable to have, so I think we should merge this to allow easier unit management at least across full blender production and deal with various scales in a different PR

mkolar (approved):

Issue #4982, opened 2023-05-17
Enhancement: Prioritize correct pymxs implementation over rt.execute for improved safety and efficiency in 3ds Max


Sponge96:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Consider refactoring the existing implementation of 3dsmax to replace `rt.execute()` calls with the appropriate pymxs implementation. Although `rt.execute() `allows executing MAXScript code within pymxs, it is less efficient and less safe compared to using pymxs functions directly.` rt.execute()` involves parsing and evaluating MAXScript code, leading to a slight performance overhead. Additionally, and more importantly, it often bypasses crucial built-in safety mechanisms provided by the pymxs API, such as error handling. This can lead to silent fails making debugging more challenging.

### How would you imagine the implementation of the feature?

Refactoring the code would involve replacing `rt.execute()` calls with there pymxs implementation (where possible).
An example of this:

Current:
`frange_cmd = f"animationRange = interval {frame_start} {frame_end}"`
`rt.execute(frange_cmd)`

Refactor:
`
rt.animationRange = rt.interval(frame_start, frame_end)
`

This new implementation directly assigns the desired animation range to rt.animationRange using the rt.interval() constructor. By doing so, we eliminate the need to construct a MAXScript command string and execute it using rt.execute(). This refactor improves efficiency and eliminates potential safety concerns associated with using rt.execute() for this particular example.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

[cuID:[OP-5945](https://app.clickup.com/t/861mu44pg)]

antirotor:
Closing this as it is addressed by PRs mentioned above. If it is still issue, please re-open.

PR #4988, opened 2023-05-17
3dsmax: Refactored publish plugins to use proper implementation of pymxs

Sponge96:
### Refactored 3dsmax Publish Plugins
This pull request addresses part of the GitHub issue #4982 by refactoring most of the 3dsmax publish plugins. The primary focus of the refactoring is to replace the usage of `rt.execute()` with the appropriate implementation using pymxs.

This refactoring improves the overall code quality, performance, and safety of the 3dsmax publish plugins, bringing them in line with best practices and leveraging the full capabilities of the pymxs API.

The reasons for this refactoring are outlined in more detail in the GitHub issue #4982 .

## Additional Info:
Some rules of thumb to follow when trying to convert to undocumented pymxs function from MaxScript.

- If the function is not encapsulated behind an interface, it likely exists in the pymxs runtime (although possibly undocumented), allowing us to use rt. prefix to access it directly.

- In cases where a # symbol is used in the MaxScript function, the equivalent in pymxs can often be achieved using rt.name("foo") to handle the necessary conversion.

- Optional arguments are challenging to work with but often follow a very similar pattern to the MaxScript (example: 
pymxs = `rt.exportFile( path, rt.name("noPrompt"), selectedOnly=True, using=rt.AlembicExport)`
MaxScript = `exportFile @"{path}" #noPrompt selectedOnly:on using:AlembicExport`)


openpype/hosts/max/plugins/publish/extract_max_scene_raw.py
```diff
@@ -37,26 +28,23 @@ def process(self, instance):
...
         if "representations" not in instance.data:
             instance.data["representations"] = []
 
-        # saving max scene
-        with maintained_selection():
-            # need to figure out how to select the camera
-            rt.select(get_all_children(rt.getNodeByName(container)))
-            rt.execute(f'saveNodes selection "{max_path}" quiet:true')
+        nodes = get_all_children(rt.getNodeByName(container))
+        rt.saveNodes(nodes, max_path, quiet=True)
```
moonyuet, line 37: We implement` maintained_selection()` from context manager in other model extractors to restore the current selection before exporting. 
```suggestion

        with maintained_selection():
            nodes = get_all_children(rt.getNodeByName(container))
            rt.saveNodes(nodes, max_path, quiet=True)

```

moonyuet:
Great. I haven't fully tested it in my side but read the code. 
See comment above.

openpype/hosts/max/plugins/publish/extract_max_scene_raw.py
moonyuet: i see. that makes sense

moonyuet (approved):

antirotor (approved):
Looks great, thanks a lot! @LiborBatek could you quickly test it before the merge?

LiborBatek (approved):
I have tested all kinds of publishing and all good!

- [x] Camera
- [x] Max Scene
- [ ] Model
- [ ] PointCache
- [ ] PointCloud
- [ ] Render

PR #4989, opened 2023-05-17
AYON: Missing files on representations

iLLiCiTiT:
## Changelog Description
Fix integration of files into representation in server database.

## Additional info
The `files` were not integrated correctly. They were added to representation data instead of representation itself and the structure of files data were not right. Fixed the structure and fixed missing sites that were already handled, but not correctly. Also fixed update of hero version, where `"version_id"` change is skipped.

## Testing notes:
1. Integration in AYON mode should work and representations on server should have files
2. Update of representations should work
3. Update of hero version should work

NOTE: To test hero version enable the plugin in settings and add family you want to test to families filter in settings.

### DIFF
--- a/openpype/client/server/operations.py
+++ b/openpype/client/server/operations.py
@@ -359,8 +359,7 @@ def prepare_hero_version_update_data(old_doc, new_doc, replace=True):
     """
 
     changes = _prepare_update_data(old_doc, new_doc, replace)
-    if "version_id" in changes:
-        changes.pop("version_id")
+    changes.pop("version_id", None)
     return changes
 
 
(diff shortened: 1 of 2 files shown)


openpype/client/server/operations.py
```diff
@@ -358,7 +358,10 @@ def prepare_hero_version_update_data(old_doc, new_doc, replace=True):
         Dict[str, Any]: Changes between old and new document.
     """
 
-    return _prepare_update_data(old_doc, new_doc, replace)
+    changes = _prepare_update_data(old_doc, new_doc, replace)
+    if "version_id" in changes:
+        changes.pop("version_id")
```
BigRoy, line 363: Any reason not to just do this? Or is that slower?
```suggestion
    changes.pop("version_id", None)
```

kalisp (approved):
Seems to be working, was able to publish renders with representations to Ayon from Nuke.

PR #4991, opened 2023-05-18
General: Version attributes integration

iLLiCiTiT:
## Changelog Description
Implemented unified integrate plugin to update version attributes after all integrations for AYON. The goal is to be able update attribute values in a unified way to a version when all addon integrators are done, so e.g. ftrack can add ftrack id to matching version in AYON server etc.

The can be stored under `"versionAttributes"` key.

## Testing notes:
Since nothing is yet implementing `versionAttributes` the testing is hard.

### DIFF
--- a/openpype/plugins/publish/integrate_version_attrs.py
+++ b/openpype/plugins/publish/integrate_version_attrs.py
@@ -1,0 +1,86 @@
+import pyblish.api
+import ayon_api
+
+from openpype import AYON_SERVER_ENABLED
+from openpype.client.operations import OperationsSession
+
+
+class IntegrateVersionAttributes(pyblish.api.ContextPlugin):
+    """Integrate version attributes from predefined key.
+
+    Any integration after 'IntegrateAsset' can fill 'versionAttributes' with
+    attribute key & value to be updated on created version.
+
+    The integration must make sure the attribute is available for the version
+    entity otherwise an error would be raised.
+    """
+
+    label = "Integrate Version Attributes"
+    order = pyblish.api.IntegratorOrder + 0.5
+
+    def process(self, context):
+        available_attributes = ayon_api.get_attributes_for_type("version")
+        skipped_attributes = set()
+        project_name = context.data["projectName"]
+        op_session = OperationsSession()
+        for instance in context:
+            label = self.get_instance_label(instance)
+            version_entity = instance.data.get("versionEntity")
+            if not version_entity:
+                continue
+            attributes = instance.data.get("versionAttributes")
+            if not attributes:
+                self.log.debug((
+                    "Skipping instance {} does not have set attributes"
+                ).format(label))
+                continue
+
+            filtered_attributes = {}
+            for attr, value in attributes.items():
+                if attr not in available_attributes:
+                    skipped_attributes.add(attr)
+                else:
+                    filtered_attributes[attr] = value
+
+            if not filtered_attributes:
+                self.log.debug((
+                    "Skipping instance {} all attributes were filtered"
+                ).format(label))
+                continue
+
+            self.log.debug("Updated attributes on version {} to {}".format(
+                version_entity["_id"], str(filtered_attributes)
+            ))
+            op_session.update_entity(
+                project_name,
+                "version",
+                version_entity["_id"],
+                {"attrib": filtered_attributes}
+            )
+
+        if skipped_attributes:
+            self.log.warning(
+                "Attributes that are not available on a server {}".format(
+                    str(skipped_attributes)
+                )
+            )
+
+        if len(op_session):
+            op_session.commit()
+            self.log.info("Updated version attributes")
+        else:
+            self.log.info("There is nothing to update")
+
+    @staticmethod
+    def get_instance_label(instance):
+        return (
+            instance.data.get("label")
+            or instance.data.get("name")
+            or instance.data.get("subset")
+            or str(instance)
+        )
+
+
+# Discover the plugin only in AYON mode
+if not AYON_SERVER_ENABLED:
+    IntegrateVersionAttributes = object
--- a/openpype/plugins/publish/integrate_version_attrs.py
+++ b/openpype/plugins/publish/integrate_version_attrs.py
@@ -37,7 +37,8 @@ class IntegrateVersionAttributes(pyblish.api.ContextPlugin):
             attributes = instance.data.get("versionAttributes")
             if not attributes:
                 self.log.debug((
-                    "Skipping instance {} does not have set attributes"
+                    "Skipping instance {} because it does not have"
+                    " specified version attributes to set."
                 ).format(label))
                 continue
 
@@ -50,11 +51,11 @@ class IntegrateVersionAttributes(pyblish.api.ContextPlugin):
 
             if not filtered_attributes:
                 self.log.debug((
-                    "Skipping instance {} all attributes were filtered"
+                    "Skipping instance {} all attributes were filtered out."
                 ).format(label))
                 continue
 
-            self.log.debug("Updated attributes on version {} to {}".format(
+            self.log.debug("Updating attributes on version {} to {}".format(
                 version_entity["_id"], str(filtered_attributes)
             ))
             op_session.update_entity(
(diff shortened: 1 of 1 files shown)


jakubjezek001 (approved):
LGTM - code is looking good.

openpype/plugins/publish/integrate_version_attrs.py
```diff
@@ -0,0 +1,92 @@
...
+            instance.data.get("label")
+            or instance.data.get("name")
+            or instance.data.get("subset")
+            or str(instance)
+        )
+
+
+# Discover the plugin only in AYON mode
+if not AYON_SERVER_ENABLED:
+    IntegrateVersionAttributes = object
```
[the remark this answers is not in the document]
BigRoy: Doesn't just `del IntegrateVersionAttributes` work?
```suggestion
    del IntegrateVersionAttributes
```
That'd also shush the hound maybe?

openpype/plugins/publish/integrate_version_attrs.py
```diff
@@ -0,0 +1,92 @@
...
+                version_entity["_id"],
+                {"attrib": filtered_attributes}
+            )
+
+        if skipped_attributes:
+            self.log.warning(
+                "Attributes that are not available on a server {}".format(
+                    str(skipped_attributes)
+                )
+            )
```
BigRoy, line 72: I feel like this warning message can be clearer. _"On a server"_ what server? Did you mean to write Ayon server? "Attributes", what attributes or for what?

I imagine something like this might read clearer when just reading the logs: `"Attributes skipped due to not being available attributes for version entities on the Ayon server: {}"`

openpype/plugins/publish/integrate_version_attrs.py
```diff
@@ -0,0 +1,92 @@
...
+
+            if not filtered_attributes:
+                self.log.debug((
+                    "Skipping instance {} all attributes were filtered"
+                ).format(label))
+                continue
+
+            self.log.debug("Updated attributes on version {} to {}".format(
+                version_entity["_id"], str(filtered_attributes)
+            ))
```
BigRoy, line 59: Should this log appear AFTER the actual updating? Or should the log instead read "Updating" instead of "Updated"?

openpype/plugins/publish/integrate_version_attrs.py
```diff
@@ -0,0 +1,92 @@
...
+        for instance in context:
+            label = self.get_instance_label(instance)
+            version_entity = instance.data.get("versionEntity")
+            if not version_entity:
+                continue
+            attributes = instance.data.get("versionAttributes")
+            if not attributes:
+                self.log.debug((
+                    "Skipping instance {} does not have set attributes"
+                ).format(label))
```
BigRoy, line 41: Logged message doesn't read as a clear sentence. It's missing something like "due to" or "because of" or "because it" so it reads as "Skipping instance {} because it does not have version attributes to set specified"

openpype/plugins/publish/integrate_version_attrs.py
iLLiCiTiT: > Also, any reason we're logging this once for the full context instead of explicitly per instance?

For me it is the speed (single server call instead of one per instance). But I can change it to instance plugin if it's a big issue...

BigRoy: I didn't mean the actual committing of data, but I meant the logging of skipped attributes.

iLLiCiTiT: Aaah, I don't think it would give you more information. It is expected that there will be 1-3 attributes (max) that could be set during publishing and they would be the same for each instance -> report of the same information multiple times is useless.

iLLiCiTiT: > I imagine something like this might read clearer when just reading the logs: "Attributes skipped due to not being available attributes for version entities on the Ayon server: {}"

I had to read it twice (and I know what it should tell) :D

openpype/plugins/publish/integrate_version_attrs.py
```diff
@@ -0,0 +1,92 @@
...
+            for attr, value in attributes.items():
+                if attr not in available_attributes:
+                    skipped_attributes.add(attr)
+                else:
+                    filtered_attributes[attr] = value
+
+            if not filtered_attributes:
+                self.log.debug((
+                    "Skipping instance {} all attributes were filtered out."
+                ).format(label))
```
BigRoy, line 55: Sentence is still incorrect unfortunately. How about this?
```suggestion
                self.log.debug((
                    "Skipping instance {} because all version attributes were filtered."
                ).format(label))
```

openpype/plugins/publish/integrate_version_attrs.py
```diff
@@ -0,0 +1,92 @@
...
+            label = self.get_instance_label(instance)
+            version_entity = instance.data.get("versionEntity")
+            if not version_entity:
+                continue
+            attributes = instance.data.get("versionAttributes")
+            if not attributes:
+                self.log.debug((
+                    "Skipping instance {} because it does not have"
+                    " specified version attributes to set."
+                ).format(label))
```
BigRoy, line 42: Just thought of this shorter version that might be nice:
```suggestion
                self.log.debug((
                    "Skipping instance {} because it does not"
                    " specify version attributes to set."
                ).format(label))
```

openpype/plugins/publish/integrate_version_attrs.py
```diff
@@ -0,0 +1,92 @@
...
+                "version",
+                version_entity["_id"],
+                {"attrib": filtered_attributes}
+            )
+
+        if skipped_attributes:
+            self.log.warning((
+                "Skipped attributes because they're not available"
+                " on the server: {}"
+            ).format(str(skipped_attributes)))
```
BigRoy, line 72: I still feel this message can be improved. By just reading the log it's unclear to me where they are not available on the server. What does it mean if they are not available? Etc. With an effort to improve logged messages (with this one for example being visible to an artist) it'd be nice if these could point as close as possible to the source of the problem).

This at least mentions what attributes and where it's missing:
```suggestion
            self.log.warning((
                "Skipped version attributes because they're not available"
                " on the server for version entities: {}"
            ).format(str(skipped_attributes)))
```
Still quite technical, but least provides context to what it's referring to?

openpype/plugins/publish/integrate_version_attrs.py
```diff
@@ -0,0 +1,92 @@
...
+
+        if len(op_session):
+            op_session.commit()
+            self.log.info("Updated version attributes")
+        else:
+            self.log.info("There is nothing to update")
```
BigRoy, line 78: I feel like "nothing to update" isn't very helpful information. How about making this either a debug message or removing it?

We might want to also include **what** there is not to update because reading a log like [here](https://user-images.githubusercontent.com/2439881/237647133-ef3bf04e-af3c-4cd3-ae00-78d7be15d713.png) visually in no way points to or refers to the plug-in that made the log happen and as such to the end user there's no context that this is about the "Integrating of version attributes" to begin with.

openpype/plugins/publish/integrate_version_attrs.py
iLLiCiTiT: The plugin is named `Integrate Version Attributes`, why to repeat the "version attribute" information in each log?

BigRoy: Due to what I mentioned [here](https://github.com/ynput/OpenPype/pull/4991#discussion_r1200472956)

> We might want to also include what t[here](https://user-images.githubusercontent.com/2439881/237647133-ef3bf04e-af3c-4cd3-ae00-78d7be15d713.png) is not to update because reading a log like here visually in no way points to or refers to the plug-in that made the log happen and as such to the end user there's no context that this is about the "Integrating of version attributes" to begin with.

The UI nor logs clarify to which plug-in they are related.

iLLiCiTiT: Ok, now it make sense

PR #4992, opened 2023-05-19
AYON: Staging versions can be used

iLLiCiTiT:
## Changelog Description
Added ability to use staging versions in AYON mode.

## Additional info
It is possible to be in AYON mode with enabled staging (using `--use-staging` argument). We don't have yet staging icons ready -> maybe should be added in this PR?

## Testing notes:
1. Run `poetry run python ayon_start.py --use-staging` (or modify one of existing script tools)
2. Staging versions of addons should be loaded.
3. Enable addon in production and disable the same in staging -> it should load in production and should not in staging
4. Change setting in staging to somethign else than in production

NOTE: Once you start staging mode, an environment variable to use staging is set. If you're using terminal/cmd/powershell then don't use the same terminal window for launching production and staging.


antirotor (approved):
looks nice :)

kalisp (approved):
It correctly switches between addons that are enabled/disabled in production/staging.

PR #4993, opened 2023-05-19
General: Fix Python 2 vendor for arrow

iLLiCiTiT:
## Changelog Description
Moved remaining dependencies for arrow from ftrack to python 2 vendor.

## Additional info
This issue probably can happen only in AYON mode. Looks like the arrow is causing issues in Nuke 12 when builtins and functools are not available next to it. This is continuos fix to https://github.com/ynput/OpenPype/pull/4981 PR.

## Testing notes:
1. Run Nuke 12 (or older)
2. Startup should not crash


kalisp (approved):
Works with Nuke12

PR #4994, opened 2023-05-19
Unreal: Addon Py2 compatibility

iLLiCiTiT:
## Changelog Description
Fixed Python 2 compatibility of unreal addon.

## Additional info
Current addon implementation causes SyntaxError in Python 2 DCCs. Which is very scary for artists.

## Testing notes:
1. Unreal addon should not cause any crashes when ModulesManager is initialized in Python 2 DCC


antirotor (approved):
👍🏻 looks good!

PR #4995, opened 2023-05-21
Maya: Light family - OP-1430

tokejepsen:
## Changelog Description
Introduces new `light` family in Maya for publishing lights.

## Testing notes:
1. Add any light types into `project_settings/maya/publish/ValidateLightRequired/light_types` that needs to be considered as a `light` family.
2. Add any required attribute values into `project_settings/maya/publish/ValidateLightRequired/attribute_values`. First field is `{node_type}.{attribute_name}`, second field is the value expected.
3. Add any optional attribute values into `project_settings/maya/publish/ValidateLightOptional/attribute_values`. First field is `{node_type}.{attribute_name}`, second field is the value expected.
4. Publish and load.


openpype/hosts/maya/plugins/publish/validate_light.py
```diff
@@ -0,0 +1,111 @@
...
+
+            for shape in shapes:
+                file_nodes.extend(
+                    cmds.listConnections(
+                        shape,
+                        skipConversionNodes=True,
+                        type="file",
+                        destination=False
+                    )
+                )
```
BigRoy, line 41:
```suggestion
                file_nodes.extend(
                    cmds.listConnections(
                        shape,
                        skipConversionNodes=True,
                        type="file",
                        destination=False
                    ) or []
                )
```

openpype/hosts/maya/plugins/publish/validate_light.py
```diff
@@ -0,0 +1,111 @@
...
+
+            shape_types = {x: cmds.nodeType(x) for x in shapes}
+            invalid_types = set(shape_types.values()) - set(self.light_types)
+            invalid_nodes.extend(
+                [x for x in shape_types if shape_types[x] in invalid_types]
+            )
```
BigRoy, line 47: This making of a dict would've made much more sense if instead the dict lookup would be reversed since you're basically not using the key mapping to your benefit at all for the lookup requiring the re-iteration of all the shapes again.

```python
shapes_by_type = defaultdict(list)
for shape in shapes:
    nodetype = cmds.nodeType(shape)
    shapes_by_type[nodetype].append(shape)

invalid_types = set()
for nodetype, nodetype_shapes in shapes_by_types.items():
    invalid_types.add(nodetype)
    invalid_nodes.extend(nodetype_shapes)
```

But also since the collected invalid types don't appear to be logged further down this could've also been as simple as:
```python
for shape in shapes:
    if cmds.nodeType(shape) not in self.light_types:
        invalid_nodes.append(shape)
```

openpype/hosts/maya/plugins/create/create_light.py
```diff
@@ -0,0 +1,10 @@
+from openpype.hosts.maya.api import plugin
+
+
+class CreateLight(plugin.Creator):
+    """Light"""
```
BigRoy, line 5: "Light" seems awfully generic as a description of what this is.

Some things that crossed my mind:

- Is it only ever a single light? or possibly multiple lights?
- Are it always static lights, or also animated lights?
- Would it make more sense as "lights" as opposed to "light"?

LiborBatek (approved):
Tested publishing, Loading, Updating Versions of ```Light``` family.

All good and working now!

Small note: right now it only works when strictly light objects present in the ```Light``` instance but when putting GRP node (user groupped lights) it stops working and dont go through ```Validate```. Not sure how problematic would be to include GRP nodes too...would be handy to just select top GRP isntead of all lights in the scene used for ```Light``` instance.

image

openpype/hosts/maya/plugins/create/create_light.py
OlivierOsotimehin: maybe lightset, lightSet light_set ?

LiborBatek:
> we should also support light blocker and such from arnold (but i think they are transform)

@tokejepsen  @BigRoy User could also add any transfom nodes I guess ...this should be possible even now when set via OP Settings to the ```Validate Light``` settings list. Or not?

tokejepsen:
> > we should also support light blocker and such from arnold (but i think they are transform)
> 
> @tokejepsen @BigRoy User could also add any transfom nodes I guess ...this should be possible even now when set via OP Settings to the `Validate Light` settings list. Or not?

Yeah, havent tried but you could have `transform` as valid light type.

LiborBatek:
> Yeah, havent tried but you could have `transform` as valid light type.

Nope, seems like only ```shapes``` are the only valid types? Did try to add ```lights_GRP``` for my ```transform``` node and no luck.

@tokejepsen Can you confirm?

tokejepsen:
> Nope, seems like only `shapes` are the only valid types? Did try to add `lights_GRP` for my `transform` node and no luck.
> 
> @tokejepsen Can you confirm?

My bad. Yeah, only the shape types are being validated against atm.

Will have a think for next time I get back to this.

LiborBatek:
@BigRoy how about this PR? what u think? shouldnt we just merge it as is?? It seems quite ok to me but nothing happening here...

tokejepsen:
PR is in draft waiting for client funding.

LiborBatek:
@antirotor @BigRoy guys how about this kiddo? almost ready but on hold for such a long time. Would be cool to get it out from the doors!

moonyuet:
What is the stage of this PR? It seems to be left off for some time....

LiborBatek:
@moonyuet I think that this PR is functional at its core functionality (just for lights) however there been some proposal to broaden its function for other type of object classes like lightblocker etc. 
It really depends if these should be included within this PR or not. Maybe just review the code and test it as is should do....

tokejepsen:
Closing due to lack of client support.

PR #4996, opened 2023-05-22
Template Builder: update frames and handles of instances from Ftrack data when building the first scene

friquette:
## Changelog Description
When using the Build workfile from template, update the frames and handles for all instances by calling the reset_frame_range function.

## Additional info
https://github.com/ynput/OpenPype/pull/4793 needs to be merged before this one

### DIFF
--- a/openpype/hosts/maya/api/lib.py
+++ b/openpype/hosts/maya/api/lib.py
@@ -2299,6 +2299,24 @@ def reset_frame_range(playback=True, render=True, fps=True):
         cmds.setAttr("defaultRenderGlobals.startFrame", frame_start)
         cmds.setAttr("defaultRenderGlobals.endFrame", frame_end)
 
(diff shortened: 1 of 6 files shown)


iLLiCiTiT:
Not sure if the workflow is ok, but loaded references should be called `containers` not `instances`.

openpype/hosts/maya/api/lib.py
BigRoy, line 1: I'm pretty sure there are currently a lot of changes in this file that are actually reverting recent additions into `develop` - you might have rebased wrong or have reverted the wrong changes upon a merge conflict. For example `get_frame_range` is now suddenly changing the timeline again instead of just getting the values. I suspect something is off with this PR.

tokejepsen:
@friquette what is the status on this PR?

friquette:
New PR cleaner here: #5201

PR #5000, opened 2023-05-22
General: Fix new load plugins for next minor relase

iLLiCiTiT:
## Changelog Description
Fix access to `fname` attribute which is not available on load plugin anymore.

## Additional info
Continous task of https://github.com/ynput/OpenPype/pull/4602 after develop update.

## Testing notes:
1. Max load plugins should work
2. Load mesh in substance painter should work


antirotor (approved):
reasonable code!

PR #5002, opened 2023-05-22
Unreal: Fix the frame range when loading camera

simonebarbieri:
## Changelog Description
The keyframes of the camera, when loaded, were not using the correct frame range.

## Testing notes:
1. Load camera in Unreal.
2. Check in the sequencer if the frame range is right and the keyframes are in the right range.


LiborBatek (approved):
All keyframes are at correct ```Clip In``` and ```Clip Out``` positions...
image

PR #5006, opened 2023-05-22
Resolve: Use last workfile prelaunch hook

iLLiCiTiT:
## Changelog Description
Added `resolve` to last workfile prelaunch hook to launch last workfile on open.

## Testing notes:
1. Select a task in a project
2. Launch Resolve on the task
3. Save the workfile using workfiles tool
4. Close resolve
5. Launch it again on the same task
6. The workfile should be opened on resolve launch

### DIFF
--- a/openpype/hooks/pre_add_last_workfile_arg.py
+++ b/openpype/hooks/pre_add_last_workfile_arg.py
@@ -22,6 +22,7 @@ class AddLastWorkfileToLaunchArgs(PreLaunchHook):
         "houdini",
         "nukestudio",
         "fusion",
+        "resolve",
         "blender",
         "photoshop",
         "tvpaint",


BigRoy:
I tried this last week - this didn't work for me with Resolve at least. @iLLiCiTiT did you test this or did you just assume this might work?

iLLiCiTiT:
> @iLLiCiTiT did you test this or did you just assume this might work?

Just assumed. I don't have resolve.

BigRoy:
Pretty sure this doesn't work. @jakubjezek001 could you confirm?

jakubjezek001 (approved):
tested on Resolve 18 Studio and was working fine as described in Test notes

BigRoy:
> tested on Resolve 18 Studio and was working fine as described in Test notes

Ok, so this is what happens for me. It does seem to import the `.drp` file directly on launch. However if the project is already (by name) inside the local resolve database this is what I see:
image

Then if I were to save it with the "Copy of" name it opens as expected however the name will remain "Copy of" at the top of the Resolve window (even when using workfiles > save to increment).

Which admittedly is a bit of a confusing UX for the artist?

Anyway, aside of that it does seem to work to "import" (not open?) the last workfile on launch.

jakubjezek001:
> Which admittedly is a bit of a confusing UX for the artist?

I can confirm it is happening to me, after I was following your steps. Yes it is very confusing. 

It should be opening the found project in resolve db rather then importing .drp file into it.

openpype/hosts/resolve/api/lib.py
```diff
@@ -170,6 +170,8 @@ def create_media_pool_item(fpath: str,
     dirname, file = os.path.split(fpath)
     _name, ext = os.path.splitext(file)
 
+    print("Adding to media pool: {}".format(fpath))
+
```
BigRoy, line 174: Let's avoid this debug printing please.

openpype/hosts/resolve/api/lib.py
```diff
@@ -178,8 +180,10 @@ def create_media_pool_item(fpath: str,
         return False
 
     # if any are added then look into them for the right extension
-    media_pool_item = [mpi for mpi in media_pool_items
-                       if ext in mpi.GetClipProperty("File Path")]
+    media_pool_item = [
+        mpi for mpi in media_pool_items
+        if ext in mpi.GetClipProperty("File Path")
+    ]
```
BigRoy, line 186: Would `.endswith(ext)` make more sense?

openpype/hosts/resolve/api/workio.py
```diff
@@ -27,14 +27,16 @@ def save_file(filepath):
     pm = get_project_manager()
     file = os.path.basename(filepath)
     fname, _ = os.path.splitext(file)
+    dname, _ = fname.split("_v")
```
BigRoy, line 30: This is NOT safe. OpenPype doesn't require `_v` to be an existing prefix. Similarly, if a filename was named `myfile_variant_v001` this would error too. Do **not** do this! :)

I'm seeing now `dname` is also used in the logic below in `open_file`, but this looks like issues just waiting to happen!

openpype/hosts/resolve/api/workio.py
```diff
@@ -72,15 +75,19 @@ def open_file(filepath):
...
+
 def current_file():
     pm = get_project_manager()
     current_dir = os.getenv("AVALON_WORKDIR")
     project = pm.GetCurrentProject()
     name = project.GetName()
     fname = name + exported_projet_ext
     current_file = os.path.join(current_dir, fname)
+
+    # check if file exists
```
BigRoy, line 87: Comment is incorrect - there is no checking done here whether the file exists at all.

openpype/hosts/resolve/hooks/post_last_workfile.py
BigRoy, line 1: I know I shared a bit of this prototype with you, but I'd actually want to argue to not merge this logic. It's very unreliable.

The post launch hook is not reliable though, some reasons:
- For me when launching resolve if it's already open, it doesn't open it again. The postlaunch hook won't detect that. This is a big one since it could mean the script starts running against an active session!
  - And if there was a way to open Resolve twice then how do you know WHICH resolve session your post launch hook should run against?
- Resolve might fail to launch, the post launch hook as far as I know doesn't detect that.
- It basically means the post launch hook continues pinging for Resolve until it's opened - how long that takes can differ a lot per machine making it quite unsure whether it'll work as well elsewhere.
- Clicking launch of Resolve more than once would open only one Resolve session (due to only ever having one Resolve open) - but would run the post launch hook multiple times.
- It doesn't work with Resolve free as far as I know, however the post launch hook won't be able to detect whether it'd work or not.

I would either:
- Not do this
- Or find an alternate solution.

I would not consider this logic reliable enough for any production or pipeline whatsoever. As described above there are so many cases this _could_ break that it's basically a series of support tickets waiting to happen. I would not merge this as a temporary workaround.

openpype/hosts/resolve/hooks/post_last_workfile.py
```diff
@@ -0,0 +1,89 @@
...
+    while tried < {try_count}:
+        print("Waiting for Resolve to start ...")
+        try:
+            resolve = bmd.scriptapp("Resolve")
+            main()
+            break
+        except:
+            pass
+
+        tried += 1
```
BigRoy, line 41: As far as I know you don't need to `retry` yourself since you can just pass a `timeout` argument to the `scriptapp` function and set it to e.g. 30 seconds so it'll just try continuously during the 30 seconds automatically.

openpype/hosts/resolve/hooks/post_last_workfile.py
```diff
@@ -0,0 +1,89 @@
...
+        self.log.info(f"Found last workfile: {last_workfile}")
+
+        script = SCRIPT.format(
+            workfile_path=last_workfile,
+            openpype_menu_path=openpype_menu_path,
+            try_count=5
+        )
+
+        executable = str(self.launch_context.executable)
+        # executable = "C:/Program Files/Blackmagic Design/DaVinci Resolve/Resolve.exe"
```
BigRoy, line 81:
```suggestion
```

openpype/hosts/resolve/hooks/post_last_workfile.py
iLLiCiTiT: The timeout didn't affect the real timeout at all on our side.

iLLiCiTiT:
Closing in favor of https://github.com/ynput/OpenPype/pull/5047

PR #5008, opened 2023-05-22
Unreal: Implemented Creator, Loader and Extractor for Levels

simonebarbieri:
## Changelog Description
Creator, Loader and Extractor for Unreal Levels have been implemented.

## Additional info
This implementation has the same limitations that UAsset has. Levels that have external dependencies cannot be published. However, it works in case only internal assets have been used, which makes it perfect for publishing lighting setups.

## Testing notes:
1. In Unreal, create a level and save it.
2. Only include in the level default Unreal objects (lights, base models).
3. From the Content Browser, select the level.
4. Open the Publisher, and from the **Create** tab, select _Level_ and click **Create**.
5. Go to the **Publish** tab and click **Publish**.
6. Close the Publisher window and open the Loader.
7. Try load the published Level.
8. No errors should pop up during the whole process.


LiborBatek (changes requested):
I am able to create/publish UAsset but not being able to load it afterwards neither as ```level``` nor ```uasset```. However the asset being listed in ```Load```

There is also issue that user cannot change ```Variant``` to custom name and when try top change it and hit ```Refresh``` btn it got back to ```Main``` Variant.

image

LiborBatek:
After creation of the ```Level``` family and trying to ```Load Level``` via OP menu ....getting following error

image


I have also tried to ```Load``` as ```UAsset``` but it does nothing....

LiborBatek (approved):
All works as expected.

PR #5010, opened 2023-05-22
Unreal: Changed naming and path of Camera Levels

simonebarbieri:
## Changelog Description
The levels created for the camera in Unreal now include `_camera` in the name, to be better identifiable, and are placed in the camera folder.

## Additional info
To make this change, it has been necessary to add a visibility track for the camera level as well. This solves a problem with the rendering, which didn't load the camera correctly and kept it at the origin during rendering.
Because the generation of the sequences is now identical for camera and layout, the functions to handle it have been moved to `hosts/unreal/api/pipeline.py`, so they are accessible by both loaders.

## Testing notes:
1. In Unreal, try loading the camera.
2. Check name of the camera, and that it is placed correctly in the master level sequence.
3. Try rendering a scene.
4. Try updating the camera.
5. Try to remove the camera through the manager.

### DIFF
--- a/openpype/hosts/unreal/api/__init__.py
+++ b/openpype/hosts/unreal/api/__init__.py
@@ -22,6 +22,8 @@ from .pipeline import (
     show_tools_popup,
     instantiate,
     UnrealHost,
+    set_sequence_hierarchy,
+    generate_sequence,
     maintained_selection
 )
 
@@ -41,5 +43,7 @@ __all__ = [
     "show_tools_popup",
     "instantiate",
     "UnrealHost",
+    "set_sequence_hierarchy",
+    "generate_sequence",
     "maintained_selection"
 ]
--- a/openpype/hosts/unreal/api/pipeline.py
+++ b/openpype/hosts/unreal/api/pipeline.py
@@ -9,12 +9,14 @@ import time
 
 import pyblish.api
 
+from openpype.client import get_asset_by_name, get_assets
 from openpype.pipeline import (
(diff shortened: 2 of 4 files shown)


LiborBatek:
Cameras works ok. They are placed correctly in sequencer and according to clip in /out.

Also Render and Publish are fine and working.

Problem being when trying Remove via Manage. It seems only selected Camera asset/level is removed but all got removed from shots_map. They however still been visible as levels in Levels Editor for some reason.

We need to address this issue.

LiborBatek (commented):
Ok I have tested these:

- [x ] create camera
- [x ] correct time in sequencer
- [ ] render sequence
- [ ] update camera
- [ ] remove camera

LiborBatek:
When doing ```Update to Latest``` on shot camera (in my case sht020) the animation is not correctly placed in time in ```Sequencer``` and not respecting ```clipIn``` and ```ClipOut``` values.

image

LiborBatek:
So now when perform ```Update to Latest``` on ```Camera``` asset all correctly set to timeline.

Update of the Camera works.
image

LiborBatek (changes requested):
All works ok besides when Removing ```Camera``` from the ```Levels``` and perform ```Load``` via OP loader
the ```Camera``` got into ```Sequencer``` correctly but its not added to the ```Levels``` editor (see Levels List as it is missing ```sh010_map_camera``` after that.

image

LiborBatek (approved):
All works as expected.

Issue #5012, opened 2023-05-23
Bug: Fusion saver render publishing doesn't work


BigRoy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

So [this PR](https://github.com/ynput/OpenPype/pull/4955) has made its way into develop and there are now multiple issues it seems.

Currently the issues in **develop** are:
1. A published comp workfile now has the frame range of the asset listed, not the frame range of the comp itself. I'd say this is invalid. It makes much more sense that the published comp workfile states what the state fo the actual file was to begin with. 
2. I seem to be unable to publish newly created savers with anything other than the asset database frame range. It always seems to render that range, and never the comp timeline or render range.

_For more details see this screenshot:_

![image](https://github.com/ynput/OpenPype/assets/2439881/834db692-a3d9-42d4-a0db-33b0db68807c)

_It seems that the instance label (left hand side) correctly collected the frame range (1-10) it should be rendering however during "Render local" it seems to still render the other frame range of the asset entity in database (1001-1010)_

3. Existing savers will now by default render a different frame range than before (asset database instead of comp timeline) which makes old fusion files forward incompatible unless the setting is toggled by the user in the UI manually.
5. If the option "From asset database" is used and that range is **not** within the Fusion comp's current global in and out frame range it will still fail to render those frame ranges. As such, a validator should be in place to capture that case.

### Expected Behavior:

1. Published comp workfile should state the frame range of the actual workfile
2. Publishing a saver should use the frame range as stated in the publisher UI, either "From viewer render in/out", "From comp timeline". Both these options don't work, the render always seem to force the current context's frame range of the asset in the database.
3. (Nice to have): The savers should be backwards compatible - so it should likely default the above value to "from render range" instead of "Current asset context"

### Version

3.15.8-nightly.2

### What platform you are running OpenPype on?

Windows

### Steps To Reproduce:

1. Set comp range 1-10
2. Set asset database range 1001-1010
3. Create a new saver in the comp with a background as input
4. Set the saver frame range source to "From composition timeline"

The render will not succeed - will try and render the wrong frame range.
Also a published workfile of this comp will list the asset frame range in the loader, not the comp's actual frame range.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

[cuID:[OP-6027](https://app.clickup.com/t/863gu46mz)]

PR #5013, opened 2023-05-23
Fusion: fixing frame range targeting

jakubjezek001:
## Changelog Description
Frame range targeting at Rendering instances is now following configured options.

## Testing notes:
1. open fusion workfile and create render instance.
2. test all three options on render target switch. Make sure comp timeline is set to different frame range then asset db and set render range to some different range then comp range so you can test the calculation of handles.
3. ideally test it on multiple rendering targets

Closes https://github.com/ynput/OpenPype/issues/5012


EmberLightVFX:
> ...and have a validation for when they are outside of the comp's global in/out because Fusion seems to fail rendering for those cases anyway...

I have created a validator that might be of interest to modify and integrate in this PR: https://github.com/ynput/OpenPype/pull/4921

BigRoy:
> > ...and have a validation for when they are outside of the comp's global in/out because Fusion seems to fail rendering for those cases anyway...
> 
> I have created a frame-range validator that might be of interest to modify and integrate in this PR: #4921

Thanks. Took a quick look but unfortunately it's not a perfect fit, but it does show some ideas for improvements to the validator in this PR.

openpype/pipeline/publish/abstract_collect_render.py
```diff
@@ -167,16 +167,27 @@ def process(self, context):
 
             frame_start_render = int(render_instance.frameStart)
             frame_end_render = int(render_instance.frameEnd)
+            """ TODO: Needs to be refofactored because this
+                      seems to be very hacky
+            """
```
BigRoy, line 172:
```suggestion
            # TODO: Refactor hacky frame range workaround below
```

jakubjezek001:
We are having bug in the Render Local. It is not restoring original render range after it is rendered.

BigRoy:
@jakubjezek001 could you try again?

jakubjezek001:
> @jakubjezek001 could you try again?

works perfect now and ready to be Approved ;) Thanks @BigRoy

BigRoy (approved):
I did a lot of the commits so take my approval with a grain of salt please.
@EmberLightVFX anything that still stands out to you?

jakubjezek001:
> I did a lot of the commits so take my approval with a grain of salt please. @EmberLightVFX anything that still stands out to you?

Eh, sorry to be so quick. This was painful at one of our clients and there was a pressure to deploy it quickly. @EmberLightVFX you might want to give some feedback on this please.

PR #4938, opened 2023-05-05
Unreal: Fix missing 'object_path' property

simonebarbieri:
## Changelog Description
Epic removed the `object_path` property from `AssetData`. This PR fixes usages of that property.

Fixes #4936 

## Testing notes:
1. In Unreal, load a layout.
2. Try changing version of the layout.

PR #5014, opened 2023-05-23
Unreal: Fix usage of 'get_full_path' function

simonebarbieri:
## Changelog Description
This PR changes all the occurrences of `get_full_path` functions to alternatives to get the path of the objects.

## Additional info
`get_full_path()` returns a string that includes the name of the class of the objects. This was preventing the change of the level in most of the loaders, and to save some of the assets.
This function has been included to fix another error in PR #4938.

## Testing notes:
The function was not causing evident errors. But there were some messages in the output log that it was impossible to change level because the path was malformed.
One of the most evident problems was that updating a layout placed the new asset in the wrong level, so to test if the PR is working, try updating a layout and check if the assets are in the correct level.


LiborBatek (approved):
All good. Layout got updated and to correct Level.

PR #5015, opened 2023-05-23
Deadline: fix selection from multiple webservices

kalisp:
## Changelog Description
Multiple different DL webservice could be configured. First they must by configured in System Settings., then they could be configured per project in `project_settings/deadline/deadline_servers`.
Only single webservice could be a target of publish though.

## Testing notes:
1. set new DL webservice in System Settings and in  `project_settings/deadline/deadline_servers`. (it might not be working WS, collection phase will work without it, it should fail in Validator phase
2. check logging of `"Default Deadline Webservice"
AnyDeskMSI_bbehLEYOw6

### DIFF
--- a/website/docs/module_deadline.md
+++ b/website/docs/module_deadline.md
@@ -22,6 +22,9 @@ For [AWS Thinkbox Deadline](https://www.awsthinkbox.com/deadline) support you ne
 
 5. Install our custom plugin and scripts to your deadline repository. It should be as simple as copying content of `openpype/modules/deadline/repository/custom` to `path/to/your/deadline/repository/custom`.
 
+Multiple different DL webservice could be configured. First set them in point 4., then they could be configured per project in `project_settings/deadline/deadline_servers`.
+Only single webservice could be a target of publish though.
+
 
 ## Configuration
 


64qam (approved):

openpype/modules/deadline/plugins/publish/collect_default_deadline_server.py
```diff
@@ -23,3 +34,16 @@ def process(self, context):
...
+                            ["deadline"]
+                            ["deadline_servers"])
+        if deadline_servers:
+            deadline_server_name = deadline_servers[0]
+            deadline_webservice = deadline_module.deadline_urls.get(
+                deadline_server_name)
+            if deadline_webservice:
+                context.data["defaultDeadline"] = deadline_webservice
+                self.log.debug("Overriding from project settings with {}".format(  # noqa: E501
+                    deadline_webservice))
```
BigRoy, line 49: In the future please try to make the log clear as to what it's actually doing, like including _what_ is being overridden from project settings since visually in the logs it might not be directly clear that the log is related to this particular plug-in.

```python
self.log.debug("Overriding default deadline webservice from project settings to: {}".format(deadline_webservice))
```

Let's try and avoid things like logging just "Changing to: x" or "Updating with: y" as from reading the logs it's not instantly clear what is being altered.

PR #5018, opened 2023-05-23
Enhancement: Improve logging levels and messages for artist facing publish reports

BigRoy:
## Changelog Description

Tweak the logging levels and messages to try and only show those logs that an artist should see and could understand. Move anything that's slightly more involved into a "debug" message instead.

## Additional info

It's still unfortunate that some function calls log into `info` for what one could consider informational logs but for an artist facing report might not be so. Specifically subprocesses returning output through `stderr` gets redirected to `log.info` calls. This is noticable e.g. for FFMPEG conversions like in Extract Review:

image

Similarly profile filtering and colorspace logic calls log info messages like these:

image

Which I'd argue we could all consider debug messages - but I wasn't too sure whether I should start touching that parts of the code. What do we think?

_Also, I almost feel like we should also start considering having a logging level like _TRACE_ or something maybe even lower than debug or between debug and info since there's even some debugging information logged here and there that are basically literal variable value printing debugging - which is really verbose._ However, it seems Python has no default logging level that describes that type of extreme logging?

---

_I tested in Fusion for my current test runs._

## Testing notes:

1. Start publishing with the new publisher. Artist facing reports should be a bit clearer.
2. Anything that still does not feel right for an artist or is basically an unreadable message, please report and let's see how we want to proceed with cleaning up those.

### DIFF
--- a/openpype/hosts/fusion/plugins/publish/collect_inputs.py
+++ b/openpype/hosts/fusion/plugins/publish/collect_inputs.py
@@ -113,4 +113,4 @@ class CollectUpstreamInputs(pyblish.api.InstancePlugin):
 
         inputs = [c["representation"] for c in containers]
         instance.data["inputRepresentations"] = inputs
-        self.log.info("Collected inputs: %s" % inputs)
+        self.log.debug("Collected inputs: %s" % inputs)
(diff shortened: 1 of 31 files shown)


mkolar (approved):
I agree 100% with pretty much all of this. log.info has been long abused and should be reserved for only artist relevant information. 

regarding debug, it would be great to split it even more, but frankly I'm not sure it's worth the effort. If we hide the debug by default from the artist, this should be well sufficient and low hanging fruit to go through all info logs across the plugins and make sure they are reasonable. Great job

LiborBatek:
even tho its merged already as @BigRoy wrote to report any of such cases... I have always hated when the following one showed up! In multiple cases...wether being it wrong object inside subset or missing one,

image

PR #5021, opened 2023-05-24
Unreal: Fix sequence frames validator to use correct data

simonebarbieri:
## Changelog Description
Fix sequence frames validator to use clipIn and clipOut data instead of frameStart and frameEnd.

## Testing notes:
1. In Unreal, create a Render instance.
2. Render the sequence.
3. Try publishing. The validator should pass if data is correct in the Project Manager.


LiborBatek (approved):
All working fine, no Validate unnceccesary errors present. So all good.

image

Note: weird Publisher window UI render. But functional ok.

PR #5022, opened 2023-05-24
Settings: Added option to nest settings templates

iLLiCiTiT:
## Changelog Description
It is possible to nest settings templates in another templates.

## Additional info
Not sure if this is a bugfix or enhancement? I think it did work in past, but not really sure.

## Testing notes:
1. Nest template inside another template and use it in settings.
```
# template_1.json
[
   {"type": "boolean", "key": "test_bool"}
]

# template_2.json
[
    {"type": "template", "name": "template_1"}
]
```
2. UI Should not crash.

### DIFF
--- a/openpype/settings/entities/lib.py
+++ b/openpype/settings/entities/lib.py
@@ -323,7 +323,10 @@ class SchemasHub:
         filled_template = self._fill_template(
             schema_data, template_def
         )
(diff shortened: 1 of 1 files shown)


jakubjezek001 (approved):
LGTM!

PR #5023, opened 2023-05-24
Unreal: Fix render instances collection to use correct data

simonebarbieri:
## Changelog Description
Fix render instances collection to use `frameStart` and `frameEnd` from the Project Manager, instead of the sequence's ones.

## Testing notes:
1. In Unreal, create a Render instance and render it.
2. Publish the render.
3. Check in the published render files if the name uses the right frame range.


LiborBatek (approved):
The rendered image sequence got proper ```FrameStart``` and ```FrameEnd``` when ```Publish``` Finnished.

Small note on UI there is some weird display tho. But all finnished ok.

image

LiborBatek:
No errors during ```Publish``` shown...
image

PR #5024, opened 2023-05-24
General: delivery action add renamed frame number in Loader

moonyuet:
## Changelog Description
Frame Offset options for delivery in Openpype loader
image

## Additional info
n/a

## Testing notes:
1. Right Click Openpype Icon > Loader
2.  Right click and choose Deliver Versions
image
3.  Set your preferred Frame Offset
image
4. Deliver


openpype/plugins/load/delivery.py
```diff
@@ -88,6 +88,9 @@ def __init__(self, contexts, log=None, parent=None):
         template_label.setCursor(QtGui.QCursor(QtCore.Qt.IBeamCursor))
         template_label.setTextInteractionFlags(QtCore.Qt.TextSelectableByMouse)
 
+        frame_offset_value = QtWidgets.QSpinBox()
+        frame_offset_value.setMinimum(-99)
```
antirotor, line 92: I think this must be unlimited number - for example I might want to shift render from 1001 to 0001

antirotor: yep, that would be really useful

moonyuet:
i have added the new Checkbox for renumbering frame to 0001. The system won't see frame offset value and reset all the start frame begun from 0001 once the checkbox is checked.
image

BigRoy:
> i have added the new Checkbox for renumbering frame to 0001. The system won't see frame offset value and reset all the start frame begun from 0001 once the checkbox is checked.

I'd actually have expected to have a "Renumber frame" option PLUS a "Renumber start frame" value so that you could pick what start frame it should be for all delivered clips.

antirotor (changes requested):
for some reason it doesn't work for me as expected:

image

I have just two file sequence starting from 1001. With the settings, I would expect them to be renumbered like `sh010_lighting_v004.0000.exr`, `sh010_lighting_v004.0001.exr`

openpype/plugins/load/delivery.py
antirotor: > I feel that code suggestion should _definitely_ come with a comment to explain what it does. Would using [`sys.maxint`](https://docs.python.org/2/library/sys.html#sys.maxint) or alike work and be more readable? :)

oh, just saw your comment. `sys.maxint` didn't make it to python 3 and this is even more bizzare because this type size depends more on Qt binaries than Python. So this is just attempt to get 32bit size integer number divided by two to get maximum signed integer number, even though it might be totally off because you can build PySide2 with different sizes.

openpype/plugins/load/delivery.py
```diff
@@ -88,6 +88,11 @@ def __init__(self, contexts, log=None, parent=None):
         template_label.setCursor(QtGui.QCursor(QtCore.Qt.IBeamCursor))
         template_label.setTextInteractionFlags(QtCore.Qt.TextSelectableByMouse)
 
+        renumber_frame = QtWidgets.QCheckBox()
+
+        renumber_frame_value = QtWidgets.QSpinBox()
+        renumber_frame_value.setMinimum(-(1 << 32) // 2)
```
antirotor, line 94: We need to set max value to, otherwise it allows only `99`

```suggestion
        # set minimum and maximum values with the size of 32bit integer.
        max_int = (1 << 32) // 2
        renumber_frame_value.setRange(-max_int, max_int - 1)
```

antirotor (changes requested):
It works, I've found that we need to set maximum value on the re-number spinner (please see the suggestion above).

openpype/pipeline/delivery.py
```diff
@@ -120,7 +120,9 @@ def deliver_single_file(
     anatomy_data,
     format_dict,
     report_items,
-    log
+    log,
+    renumber_frame=False,
```
iLLiCiTiT, line 124: I'm not sure if it makes sense to have renumbering logic in deliver single file? We don't have any frame information about the file and it can easily try to change e.g. shot number instead of frame.

openpype/pipeline/delivery.py
```diff
@@ -299,10 +336,28 @@ def hash_path_exist(myPath):
             os.path.join(dir_path, src_file_name)
         )
 
-        dst_padding = dst_collection.format("{padding}") % index
-        dst = "{}{}{}".format(dst_head, dst_padding, dst_tail)
-        log.debug("Copying single: {} -> {}".format(src, dst))
-        _copy_file(src, dst)
+        if renumber_frame:
```
iLLiCiTiT, line 339: This way we would maybe not need `renumber_frame` argument at all? @antirotor thoughts?
```suggestion
        if re_frame_value != 0:
```

openpype/pipeline/delivery.py
```diff
@@ -299,10 +336,28 @@ def hash_path_exist(myPath):
...
+            dsp_index = (re_frame_value - int(index)) + 1
```
iLLiCiTiT, line 340: Typo in variable `dsp_index` -> `dst_index`.

And I'm not sure I understand the logic? If `re_frame_value` is `1001` and source frame is `1001` the output frame would be `1`.

openpype/pipeline/delivery.py
moonyuet: > I'm not sure if it makes sense to have renumbering logic in deliver single file? We don't have any frame information about the file and it can easily try to change e.g. shot number instead of frame.
When I first tried to implement  with `deliver_sequences` into delivery widget, it's not working as expected.  @antirotor told me that the related function is deprecated in Openpype 3 so I used `deliver_single_file` instead.

openpype/pipeline/delivery.py
antirotor: it is the *new frame start*

openpype/pipeline/delivery.py
```diff
@@ -162,8 +164,41 @@ def deliver_single_file(
...
         os.makedirs(delivery_folder)
 
-    log.debug("Copying single: {} -> {}".format(src_path, delivery_path))
-    _copy_file(src_path, delivery_path)
+    if has_renumbered_frame:
+        src_dir = os.path.dirname(src_path)
+        src_filepaths = os.listdir(src_dir)
+        src_collections, remainders = clique.assemble(src_filepaths)
+
+        src_collection = src_collections[0]
```
antirotor, line 172: I know we are doing it like so elsewhere, but as I am looking at it, I think this is rather dangerous - we assume that the sequence we want to renumber is the first collection, but there is no guarantee.

In ideal world, there should be just one collection and no reminder - and in that case it should work as expected. But in case there are more collections... Like for example in case of in-place publishing where you can have whatever content there can be?

We should use the files from representation - or at least issue warning that there is more than one sequence there? But the use of files from the selected representation would be the safest way probably. What do you think @iLLiCiTiT @moonyuet

moonyuet: i see your point and it is indeed way safer to the files from representation for the sequence part, especially when the user wants to enable to use the renumber option for the single frame.

Minkiu: I wanted to add that might as well keep it consistent with line `190`, whatever is that we end up going with.

openpype/pipeline/delivery.py
```diff
@@ -162,8 +164,41 @@ def deliver_single_file(
...
+        src_indexes = list(src_collection.indexes)
+
+        destination_indexes = [
+            renumber_frame + idx
+            for idx in range(len(src_indexes))
+        ]
+
+        dst_filepaths = []
+        for index in destination_indexes:
```
Minkiu, line 181: The idea here is to only modify the frame on the filename, right?

So if we have `beep_boop.1001.jpeg` we want to get `beep_boop.0001.jpeg` , correct?

IF the assumption above is correct, we could then do:
```py
collections, remainder = clique.assemble(["beep_boop.1001.jpg", "beep_boop.1002.jpg"])
source_collection = collections[0]
# <Collection "beep_boop.%d.jpg [1001-1002]">
destination_indexes = [
    renumber_frame + idx
    for idx in range(len(col.indexes))
]

destination_collection = clique.Collection(
    source_collection.head,
    source_collection.tail,
    source_collection.padding, # Unsure about this one,
    indexes=destination_indexes
)
# <Collection "beep_boop.%d.jpg [0-1]">
```

moonyuet: I am actually talking about other things. It will probably resolve if we are using representation data if a single frame has been published(if i dont get it wrong, it will publish in `{filename}.{frame}.{ext}` and the file transmission from integrate files will remove the frame if it is a single frame). It will probably break the delivery action if we still using `os.listdir(src_dir)` with single frame when the renumber option enabled.

BigRoy: Regarding @Minkiu 's example I don't think that's the perfect future-compatible fix either.  think at some point we'll need to support publishing sequences with holes in it - instead of force closing the gaps. However I do really like his code style, it's much more readable like that. 

I think this could be a quite a decent readable solution:

```python
def shift_collection(collection, start_offset, relative=True):
    """Shift frames of a clique.Collection.
    
    When relative is True `start_offset` will be an offset from
    the current start frame of the collection shifting it by `start_offset`.
    When relative is False `start_offset` will be the new start
    frame of the sequence - shifting the rest of the frames along.
    
    Arguments:
        collection (clique.Collection): Input collection.
        start_offset (int): Offset to apply (or start frame to set if
            relative is False)
        relative (bool): Whether the start offset is relative or the
            the absolute new start frame.
    
    Returns:
        clique.Collection: Shifted collection
    
    """
    first_frame = next(iter(col.indexes))
    shift = first_frame - start_offset if not relative else start_offset
    return clique.Collection(
        head=source_collection.head, 
        tail=source_collection.tail, 
        padding=source_collection.padding, 
        indexes={idx + shift for idx in collection.indexes} 
    )
```

Which means that if you have any renumbering you can just do:
```python
if has_shift:
    dst_collection = shift_collection(dst_collection, start_offset=start_offset, relative=relative)
```

Right?

moonyuet: I assume this function is prepare for the future use as there is `relative=True` as argument?

BigRoy: I thought this delivery action had a way to set the start frame explicitly OR shift the start frame. 

```
So input sequence is e.g. frame 1001-1100. 
Then renumber could either shift it by 100 (relative): 1101-1200
Or renumber it to a new start frame (absolute), e.g. frame 1: 0001-0100.
```

_Note:_ It's good to be a aware that if the input sequence is 1001-1100 that `clique.Collection` itself doesn't recognize a padding of the source sequence, so it'll have padding `0` and the output does becomes e.g. `1-100` not `0001-0100`.

If you want to have at least the padding width of the start frame of the source collection you'd still need to do:
```python
first_frame = next(iter(src_collection.indexes))
dst_collection.padding = len(str(first_frame))
```

openpype/pipeline/delivery.py
```diff
@@ -162,8 +164,48 @@ def deliver_single_file(
...
+        for src_file_name, dst in zip(src_collection, dst_collection):
+            src_path = os.path.join(src_dir, src_file_name)
+            delivery_path = os.path.join(delivery_folder, dst)
+            log.debug("Copying single: {} -> {}".format(
+                src_path, delivery_path))
+            _copy_file(src_path, delivery_path)
+    else:
+        log.debug("Copying single: {} -> {}".format(src_path, delivery_path))
+        _copy_file(src_path, delivery_path)
+
```
iLLiCiTiT, line 208: Single file cannot be reframed
```suggestion
    log.debug("Copying single: {} -> {}".format(src_path, delivery_path))
    _copy_file(src_path, delivery_path)
```

openpype/pipeline/delivery.py
```diff
@@ -120,7 +120,9 @@ def deliver_single_file(
     anatomy_data,
     format_dict,
     report_items,
-    log
+    log,
+    has_renumbered_frame=False,
+    renumber_frame=0
```
iLLiCiTiT, line 125:
```suggestion
    log
```

openpype/pipeline/delivery.py
moonyuet: but i have tried with `deliver_sequence_file` codes. it seems that the delivery action only works with `deliver_single_file`

openpype/pipeline/delivery.py
```diff
@@ -176,7 +218,9 @@ def deliver_sequence(
...
```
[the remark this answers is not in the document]
moonyuet: if the renumber frame is 12 and you have 100-120 frame range, the frame range starts from 12 and ends with 32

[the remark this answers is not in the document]
iLLiCiTiT: So it is "new_frame_start". I would never get it from `renumber_frame`.

openpype/pipeline/delivery.py
moonyuet: you can scroll up to see all the changes requested, it seems `deliver_sequence` is deprecated for some reasons. At the early commit I set it through `deliver_sequence`, the delivery action won't work. And it works after I set up the delivery action into `delivery_single_file` in the later commit.

openpype/pipeline/delivery.py
```diff
@@ -299,10 +343,64 @@ def hash_path_exist(myPath):
...
+            )
+
+            _copy_file(src, dst)
+
+        else:
+            dst_padding = dst_collection.format("{padding}") % index
+            dst = "{}{}{}".format(dst_head, dst_padding, dst_tail)
+            log.debug("Copying single: {} -> {}".format(src, dst))
+            _copy_file(src, dst)
+
```
iLLiCiTiT:
```suggestion
        dst_index = index
        if has_renumbered_frame:
            # Calculate offset between first frame and current frame
            # - '0' for first frame
            offset = first_frame - index
            # Add offset to new frame start
            dst_index = renumber_frame + offset
            if dst_index < 0:
                msg = "Renumber frame has a smaller number than original frame"     # noqa
                report_items[msg].append(src_file_name)
                log.warning("{} <{}>".format(msg, context))
                return report_items, 0

        dst_padding = dst_collection.format("{padding}") % index
        dst = "{}{}{}".format(dst_head, dst_padding, dst_tail)
        log.debug("Copying single: {} -> {}".format(src, dst))
        _copy_file(src, dst)

```

openpype/pipeline/delivery.py
iLLiCiTiT: But `delivery_single_file` by design cannot handle renumbering. Single file has no information about frame.

moonyuet: > But `delivery_single_file` by design cannot handle renumbering. Single file has no information about frame.

this is something i find really strange at first but this is something true. If you remove all the renumber args in the `deliver_single_files `and just add the args into `deliver_sequence`, and you run tray to do the delivery action you find out `deliver_sequence `doesn't renumber any frame.  you can try the code with this commit: https://github.com/ynput/OpenPype/pull/5024/commits/8030ee9c71fcbad57802aff9f8296569e8c00d4d and you will know.

iLLiCiTiT: Yes, but that should not be handled in this function. You have to change frame fill logic in `openpype/plugins/load/delivery.py`.

openpype/pipeline/delivery.py
```diff
@@ -165,6 +165,7 @@ def deliver_single_file(
     log.debug("Copying single: {} -> {}".format(src_path, delivery_path))
     _copy_file(src_path, delivery_path)
 
+
```
iLLiCiTiT, line 168:
```suggestion
```

openpype/plugins/load/delivery.py
```diff
@@ -222,7 +241,9 @@ def deliver(self):
                     args[0] = src_path
                     if frame:
```
iLLiCiTiT, line 242: The `deliver_sequence` is deprecated and should not be used for representations with `files`. I meant to recalculate the frames here and still use `deliver_single_file`.
```python
        if repre.get("files"):
            src_paths = []
            for repre_file in repre["files"]:
                src_path = self.anatomy.fill_root(repre_file["path"])
                src_paths.append(src_path)
            sources_and_frames = collect_frames(src_paths)

            frames = set(sources_and_frames.values())
            frames.discard(None)
            first_frame = None
            if frames:
                first_frame = min(frames)

            for src_path, frame in sources_and_frames.items():
                # Renumber frames
                if has_renumbered_frame and frame is not None:
                    # Calculate offset between first frame and current frame
                    # - '0' for first frame
                    offset = frame - first_frame
                    # Add offset to new frame start
                    dst_frame = renumber_frame + offset
                    if dst_frame < 0:
                        msg = "Renumber frame has a smaller number than original frame"     # noqa
                        report_items[msg].append(src_file_name)
                        self.log.warning("{} <{}>".format(msg, context))
                        continue
                    frame = dst_frame

                if frame is not None:
                    anatomy_data["frame"] = frame
                new_report_items, uploaded = deliver_single_file(*args)
                report_items.update(new_report_items)
                self._update_progress(uploaded)
```

openpype/pipeline/delivery.py
```diff
@@ -120,7 +120,9 @@ def deliver_single_file(
     anatomy_data,
     format_dict,
     report_items,
-    log
+    log,
+    has_renumbered_frame=False,
+    new_frame_start=0
```
iLLiCiTiT, line 125:
```suggestion
```

openpype/plugins/load/delivery.py
```diff
@@ -211,17 +223,49 @@ def deliver(self):
                 self.log
             ]
 
+            if renumber_frame:
+                optional_args = [
+                    renumber_frame,
+                    frame_offset
+                ]
+                args.extend(optional_args)
+
```
iLLiCiTiT, line 232:
```suggestion
```

openpype/plugins/load/delivery.py
```diff
@@ -211,17 +223,49 @@ def deliver(self):
...
                     src_paths.append(src_path)
                 sources_and_frames = collect_frames(src_paths)
 
+                frames = set(sources_and_frames.values())
+                frames.discard(None)
+                first_frame = None
+                if frames:
+                    first_frame = next(iter(frames))
+
+
```
iLLiCiTiT, line 246:
```suggestion

```

openpype/plugins/load/delivery.py
```diff
@@ -211,17 +223,49 @@ def deliver(self):
...
                 for repre_file in repre["files"]:
                     src_path = self.anatomy.fill_root(repre_file["path"])
                     src_paths.append(src_path)
                 sources_and_frames = collect_frames(src_paths)
 
+                frames = set(sources_and_frames.values())
+                frames.discard(None)
+                first_frame = None
+                if frames:
+                    first_frame = next(iter(frames))
```
iLLiCiTiT, line 244:
```suggestion
                    first_frame = min(frames)
```

openpype/plugins/load/delivery.py
```diff
@@ -211,17 +223,49 @@ def deliver(self):
...
-                        anatomy_data["frame"] = frame
+                    # Renumber frames
+                    if renumber_frame and frame is not None:
+                        # Calculate offset between
+                        # first frame and current frame
+                        # - '0' for first frame
+                        offset = frame_offset - int(first_frame)
+                        # Add offset to new frame start
+                        frame = int(frame)
+                        dst_frame = frame + offset
```
iLLiCiTiT, line 257:
```suggestion
                        dst_frame = int(frame) + offset
```

openpype/plugins/load/delivery.py
```diff
@@ -95,6 +95,12 @@ def __init__(self, contexts, log=None, parent=None):
         template_label.setCursor(QtGui.QCursor(QtCore.Qt.IBeamCursor))
         template_label.setTextInteractionFlags(QtCore.Qt.TextSelectableByMouse)
 
+        renumber_frame = QtWidgets.QCheckBox()
+
+        renumber_frame_value = QtWidgets.QSpinBox()
+        max_int = (1 << 32) // 2
+        renumber_frame_value.setRange(-max_int, max_int - 1)
```
iLLiCiTiT, line 102: If the result is new frame start we should 1.) rename the variables from `renumber...` to `frame_start...` (or something similar) and 2.) not allow to use less than `0` -> which makes it invalid.
```suggestion
        renumber_frame_value.setRange(0, max_int - 1)
```

openpype/plugins/load/delivery.py
```diff
@@ -211,17 +223,49 @@ def deliver(self):
...
+                            msg = "Renumber frame has a smaller number than original frame"     # noqa
+                            report_items[msg].append(src_path)
+                            self.log.warning("{} <{}>".format(
+                                msg, dst_frame))
+                            continue
+                        frame = dst_frame
+
+                    if frame is not None:
+                        if frame:
+                            anatomy_data["frame"] = frame
```
iLLiCiTiT, line 268:
```suggestion
                        anatomy_data["frame"] = frame
```

openpype/pipeline/delivery.py
```diff
@@ -294,17 +296,30 @@ def hash_path_exist(myPath):
     src_head = src_collection.head
     src_tail = src_collection.tail
     uploaded = 0
+    first_frame = next(iter(src_collection.indexes))
```
Minkiu, line 299: Maybe use `min()` here as well?

BigRoy: Using `min` will be slower however. The `indexes` are sorted so we should always be taking the first entry instead of requiring to iterate the full sorted set.

Minkiu: I suggested it so it's like in the `load/delivery` but maybe there it's required cause they might not be in order (?).

antirotor (approved):
it works as expected

mkolar:
Let's merge it then.... wooo

PR #5026, opened 2023-05-24
Houdini: Fix missing frame range for pointcache and camera exports

### LINKED ISSUE #5019: Bug: Houdini Pointcache and Camera publish don't collect frame range for database

### Current Behavior:

Whenever you publish a pointcache instance or camera instance the Loader will not show its frame range since it's not collected for the publish instances.

image

This is due to using the new Creator style instances and thus [_this logic_](https://github.com/ynput/OpenPype/blob/e0654ef9a4d787da9375fcf2e83bfeb211cc165f/openpype/hosts/houdini/plugins/publish/collect_instances.py#L73) in Collect Instances [is ignored since that plug-in only runs over legacy instances](https://github.com/ynput/OpenPype/blob/e0654ef9a4d787da9375fcf2e83bfeb211cc165f/openpype/hosts/houdini/plugins/publish/collect_instances.py#L50-L53).

There is however [this logic](https://github.com/ynput/OpenPype/blob/aa5ec57786442fc450c925ec9d5b2844574ba2cf/openpype/hosts/houdini/plugins/publish/collect_frames.py#L21-L22) for sequence based extractors like vdb and others where it does collect it fine. So the issue exists I suppose for any instance that is not in that plug-in's families for which we _do_ expect a certain frame range to be extracted.

### Expected Behavior:

Pointcache and camera instances should publish there frame ranges to the database.

#### Implementation?

I'd likely take out the `lib.get_frame_data` calls from both Collect Instances and the Collect Frames plug-ins and move it into a separate plug-in that runs over all instances, no matter the family since it should only collects the data if the ROP node has the time based attribute and `instance.data.get("instance_node")` exists (to avoid running over workfile instance).

### Version

3.15.8-nightly.2

Windows

### Steps To Reproduce:

1. Create a camera and pointcache instance
2. Publish them.
3. See no frame ranges in the loader.

### Relevant log output:

_No response_

### Additional context:

_No response_

[cuID:[OP-6031](https://app.clickup.com/t/861muwnuf)]

BigRoy:
## Changelog Description

Fix missing frame range for pointcache and camera exports on published version.

## Additional info

Resolves #5019 

## Testing notes:

1. Publish pointcache and camera instance
2. Loader should show start + end frame for it
3. Also publish other instances, e.g. workfile, vdbcache or alike - they should work as before.

### DIFF
--- a/openpype/hosts/houdini/plugins/publish/collect_frames.py
+++ b/openpype/hosts/houdini/plugins/publish/collect_frames.py
@@ -11,15 +11,13 @@ from openpype.hosts.houdini.api import lib
 class CollectFrames(pyblish.api.InstancePlugin):
     """Collect all frames which would be saved from the ROP nodes"""
 
-    order = pyblish.api.CollectorOrder
+    order = pyblish.api.CollectorOrder + 0.01
     label = "Collect Frames"
     families = ["vdbcache", "imagesequence", "ass", "redshiftproxy", "review"]
 
     def process(self, instance):
 
         ropnode = hou.node(instance.data["instance_node"])
-        frame_data = lib.get_frame_data(ropnode)
-        instance.data.update(frame_data)
 
         start_frame = instance.data.get("frameStart", None)
         end_frame = instance.data.get("frameEnd", None)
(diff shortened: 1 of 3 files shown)


mkolar (approved):
works well. Thank you

PR #5032, opened 2023-05-25
Publisher: UI works with instances without label

iLLiCiTiT:
## Changelog Description
Publisher UI does not crash if instance don't have filled 'label' key in instance data.

## Additional info
Issue was added with publish instances view in [this PR](https://github.com/ynput/OpenPype/pull/4915). Reported in UE where are instances without filled `"label"` key. Implemented helper function 'get_publish_instance_label' in publish lib, to get label from pyblish instance.

### Screenshot
image

## Testing notes:
1. Make sure you have publish instance without label filled
2. Use publisher UI to publish
3. Wait for successfull publishing, or stop publishing in middle
4. On left side should be instances view without any major issues

### DIFF
--- a/openpype/pipeline/publish/lib.py
+++ b/openpype/pipeline/publish/lib.py
@@ -30,6 +30,8 @@ from .contants import (
     TRANSIENT_DIR_TEMPLATE
 )
 
+_ARG_PLACEHOLDER = object()
+
 
 def get_template_name_profiles(
     project_name, project_settings=None, logger=None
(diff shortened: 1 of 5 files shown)


simonebarbieri (approved):
It works fine now!

image

openpype/pipeline/publish/lib.py
```diff
@@ -866,3 +868,33 @@ def add_repre_files_for_cleanup(instance, repre):
...
+    label = (
+        instance.data.get("label")
+        or instance.data.get("name")
+    )
+    if label:
+        return label
+
+    if default is _ARG_PLACEHOLDER:
+        return str(instance)
+    return default
```
BigRoy, line 900: Why can't default just be `None`
```suggestion
def get_publish_instance_label(instance, default=None):
    """Try to get label from pyblish instance.

    First are checked 'label' and 'name' keys in instance data. If are not set
    a default value is returned. Instance object is converted to string
    if default value is not specific.

    Todos:
        Maybe 'subset' key could be used too.

    Args:
        instance (pyblish.api.Instance): Pyblish instance.
        default (Optional[Any]): Default value to return if any

    Returns:
        Union[Any]: Instance label or default label.
    """

    label = (
        instance.data.get("label")
        or instance.data.get("name")
    )
    if label:
        return label

    if default is None:
        return str(instance)
    return default
```

BigRoy: Wouldn't you always want it to return a string anyway?

iLLiCiTiT: For most of cases yes, but I've used different default for debugging for example. Is it a harm to have this way?

BigRoy: Not huge harm - just more code complexity for something that currently in the code base is unused too. I'd even go as far as saying that currently the `default=` fallback is redundant since it's unused currently and thus there has not been a use case where we _required_ to deviate from just returning `instance._name` (which is what `str(instance)` does.)

iLLiCiTiT: Didn't realize `name` is required argument, added the change to PR https://github.com/ynput/OpenPype/pull/5034 here.

PR #5037, opened 2023-05-25
General: Publish plugin apply settings can expect only project settings

iLLiCiTiT:
## Changelog Description
Only project settings are passed to optional `apply_settings` method, if the method expects only one argument.

## Additional info
Previous implementation always passed project and system settings, but system settings are not used in 99.99% cases. Also is partially preparaion for AYON where system settings won't be a thing.

## Testing notes:
1. Make fake plugin with implemented classmethod `apply_settings`.
2. Make sure the plugin expects only one argument.
3. Only project settings should be passed to the method.
4. Change the arguments to expect both project and system settings.
5. Both should be passed to the method.

### DIFF
--- a/openpype/pipeline/publish/lib.py
+++ b/openpype/pipeline/publish/lib.py
@@ -12,7 +12,8 @@ import pyblish.api
 from openpype.lib import (
     Logger,
     import_filepath,
-    filter_profiles
+    filter_profiles,
+    is_func_signature_supported,
 )
 from openpype.settings import (
     get_project_settings,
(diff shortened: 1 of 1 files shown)


kalisp (approved):
Seems to work.
(I removed `system_settings` from arguments in `CollectPools` and it got triggered without issues.

BigRoy:
Did we try to profile or time the performance implications of this? There are tons of plugins and thus this runs a lot, basically on every reset or discover of plugins. Even if it adds for example 0.1s+ that'd be quite annoying.

iLLiCiTiT:
There are very few plugins that actually have implemented `apply_settings` so I would not worry about it so much.

PR #5038, opened 2023-05-25
AYON: Preparation for products

iLLiCiTiT:
## Changelog Description
Prepare ayon settings conversion script for `product` settings conversion.

## Additional info
Removed `IntegratAssetNew` settings (plugin was already removed) and TVPaint `families_to_review` unused settings. Conversion of AYON settings is converting product name to subset and product type to family.

### TODOs
- [x] Update ayon python api which has implemented products
- [x] Change entities conversion utils to be able use subset instead of product

## Testing notes:
This PR is based on changes in Sever. So make sure you're using server that is supporting products. This PR also changes template keys available in setting `{family}` -> `{product[type]}`, `{subset}` -> `{product[name]}`, `{asset}` -> `{folder[name]}`, etc.
It should work with old style templates too, but representations will have auto-converted template and context data to new style.
- Settings conversion should work with old addons and with addons that have already converted settings [this PR](https://github.com/ynput/ayon-addons/pull/12)
- Get functions and operations for subset should work as expected
1. Publishing should integrate correctly
2. Integration of the same version should work
3. Loading using Loader in DCC should work as expected

### Notes
PR should updated when [this PR](https://github.com/ynput/ayon-python-api/pull/52) is merged.


jakubjezek001 (changes requested):
There is still an issue in `collect_anatomy_instance_data.py`
image

jakubjezek001:
> There is still an issue in `collect_anatomy_instance_data.py`

This was caused by old db. After I had removed the DB and rebuilded server all was fixed.

jakubjezek001 (approved):

Issue #1780, opened 2021-06-30
Pre-flight validation


tokejepsen:
**Goal**

Generalize flagging updates to user on launch.

**Motivation**

There are a couple of crucial setups that needs to be flagged on launch rather than at publish;

- frame range update.
- resolution update.
- container update.

Currently there are no flagging of updates to asset attributes like frame range and resolution.
Container updates are handled different per host and more importantly flagging to the user about container updates are handled different. Generalizing this feedback to the user, would lower the pipeline specific knowledge for a host.

BigRoy:
An "outdated containers" pop-up does exist (Maya+Houdini that I know of at least) and for FPS I believe there's also a validation on scene open and save in Maya (and Houdini?).

Anyway, I'd love to see some UX proposals of something that is a solid small notification of "this and this and this might be something you want to look into" and potentially propose solution(s) to the user for each too.

It should be some non-blocking notification UX that the user can also "ignore" however still be notified of.

Some Notification UX related posts:
- [UI Cheat-Sheet In-App Notifications](https://uxdesign.cc/ui-cheat-sheet-in-app-notifications-fffeecf2bf38)
- [How To Design Notifications For Better UX](https://uxplanet.org/how-to-design-notifications-for-better-ux-6fb0711be54d)
- [UX Notification Design](https://www.toptal.com/designers/ux/notification-design)

I think these particular points are important

> **Non-interfering:** A notification is a timely alert. However, it can distract the user. Hence, the main characteristic of a notification is that it should be non-interfering. It should achieve the purpose of letting the user know that something important is on the way.
> **Small in size:** A good notification should be as small as possible but effective at the same time. An example of a simple unobtrusive notification is the calendar notification that usually slides at the top of your mobile screen. They are small in size but they serve the purpose well.

Would a small "stack" like this be nice?
![](https://bs-uploads.toptal.io/blackfish-uploads/uploaded_file/file/274440/image-1592224828601-435086140e9cfa92469cc6138ced91aa.png)

---

Question: A scene can take a long time to open and if a notification disappears in X seconds/minutes the user might still miss it if they were away from their desk or the application for a moment. Would you expect the pop-up to remain open? Or maybe be slightly smart that it does something like if app or user idle do not countdown timer (and is that even possible across all hosts to detect?)

Currently the notification is a tad big and stays around forever and ever until closed or acted upon by the user.

tokejepsen:
[...]

PR #5039, opened 2023-05-26
Nuke: Custom script to set frame range of read nodes

moonyuet:
## Changelog Description
Adding option to set frame range specifically for the read nodes in Openpype Panel. User can set up their preferred frame range with the frame range dialog, which can be showed after clicking `Set Frame Range (Read Node) ` in Openpype Tools
## Additional info
Settings:
image

Dialogs:
image

make sure you always choose to use selection(whether you use selected read node or not) before inserting the number of preferred frame range, otherwises it wont work as expected, 

## Testing notes:
1. Launch Nuke via Openpype
2. Create Multiple Read Nodes
3. Go to Openpype Tools > Set Frame Range(Read Node)
4. It will show a frame range dialog
5. Choose you use the selection 
6. Set your frame range
7. Hit "Execute"
8. Set up the frame range in the read nodes.


64qam:
The changed frames are not reflected in read nodes. E.g when I change the frame to 997-1024 (original frame 997-1034) the read node stays on 997-1034.
image

moonyuet:
> need to be added to OpenPype Tools items

i have updated code and changelog. and the panel is now in OP Tools in nuke

64qam:
Now, when I hit Set frame range in Openpype Tools menu it crashes with error:
```Traceback (most recent call last):
  File "C:\Users\qam\code\OpenPype\openpype\vendor\python\common\scriptsmenu\action.py", line 135, in run_command
    exec(self.process_command())
  File "<string>", line 1, in <module>
ImportError: No module named startup```

**Comment by moonyuet** [comment]
> Now, when I hit Set frame range in Openpype Tools menu it crashes with error:
> 
> ```
>   File "C:\Users\qam\code\OpenPype\openpype\vendor\python\common\scriptsmenu\action.py", line 135, in run_command
>     exec(self.process_command())
>   File "<string>", line 1, in <module>
> ImportError: No module named startup```
> ```

can you show me your script menu definition from nuke host in Openpype setting?

**Comment by 64qam** [comment]
> > Now, when I hit Set frame range in Openpype Tools menu it crashes with error:
> > ```
> >   File "C:\Users\qam\code\OpenPype\openpype\vendor\python\common\scriptsmenu\action.py", line 135, in run_command
> >     exec(self.process_command())
> >   File "<string>", line 1, in <module>
> > ImportError: No module named startup```
> > ```
> 
> can you show me your script menu definition from nuke host in Openpype setting?

```[
    {
        "type": "action",
        "sourcetype": "python",
        "title": "OpenPype Docs",
        "command": "import webbrowser;webbrowser.open(url='https://openpype.io/docs/artist_hosts_nuke_tut')",
        "tooltip": "Open the OpenPype Nuke user doc page"
    },
    {
        "type": "action",
        "sourcetype": "python",
        "title": "Set Frame Range (Read Node)",
        "command": "from openpype.hosts.nuke.startup import custom_popup;from openpype.hosts.nuke.startup.custom_popup import get_main_window;custom_popup.CustomScriptDialog(parent=get_main_window()).show();",
        "tooltip": "Set Frame Range for Read Node(s)"
    }
]```

File: openpype/hosts/nuke/startup/custom_popup.py [hunk 1]
```diff
@@ -0,0 +1,182 @@
...
+                widget.inherits("QMainWindow")
+                and widget.metaObject().className() == name
+            ):
+                main_window = widget
+                break
+    return main_window
+
+
+class CustomScriptDialog(QtWidgets.QDialog):
+    """A Popup that moves itself to bottom right of screen on show event.
```
BigRoy, line 26: Where is the logic that does this? Are we sure it's doing that?

If we would want it to do then then what's the reason we can't use the `openpype.widgets.popup.PopUp` class?

I feel like we should be working to having unified toast message pop-ups across all hosts whenever possible. So maybe there's a way that we can turn this into a dedicated Popup too or use a base class of sorts? Because somehow this somewhat sounds like [it's inline with the design proposals that came up here](https://github.com/ynput/OpenPype/issues/1780#issuecomment-1091077123).

However if it's not intended like that then I suppose just this part of the docstring is wrong?

openpype/hosts/nuke/startup/custom_popup.py
```diff
@@ -0,0 +1,182 @@
...
+                read_node["first"].setValue(frame_start)
+                read_node["last"].setValue(frame_end)
+
+        self.close()
+
+        return False
+
+    def showEvent(self, event):
+        # Position popup based on contents on show event
+        return super(CustomScriptDialog, self).showEvent(event)
```
BigRoy, line 169: This currently doesn't seem to be doing anything and thus can be removed?

openpype/hosts/nuke/startup/custom_popup.py
```diff
@@ -0,0 +1,182 @@
...
+    def setCancelText(self, text):
+        self.widgets["cancel"].setText(text)
+
+    def on_line_edit_changed(self):
+        line_edit = self.widgets['line_edit'].text()
+        self.on_line_changed.emit(line_edit)
+        return self.set_line_edit(line_edit)
+
+    def _on_clicked(self):
+        """Callback for when the 'show' button is clicked.
```
BigRoy, line 133: Docstring is wrong? There is no show button?

Doesn't raise the parent either?

openpype/hosts/nuke/startup/custom_popup.py
```diff
@@ -0,0 +1,182 @@
...
+        # Set default title
+        self.setWindowTitle("Custom Popup")
+
+    def update_values(self):
+        self.widgets["selection"].isChecked()
+
+    def emit_click_with_state(self):
+        """Emit the on_clicked signal with the toggled state"""
+        checked = self.widgets["selection"].isChecked()
+        return checked
```
BigRoy, line 112: Don't understand this function name - it doesn't really emit anything?

64qam (approved):

openpype/hosts/nuke/startup/custom_popup.py
```diff
@@ -0,0 +1,174 @@
+import sys
```
jakubjezek001, line 1: This is perhaps not the best name of the module. This seems to be something generic rather then specifically used for Read node start frame. How about to rename the module to something like `openpype_nuke_set_start_frame_to_reads.py`, this comunicate exacly the purpose of the script.

openpype/hosts/nuke/startup/custom_popup.py
```diff
@@ -0,0 +1,174 @@
...
+            if (
+                widget.inherits("QMainWindow")
+                and widget.metaObject().className() == name
+            ):
+                main_window = widget
+                break
+    return main_window
+
+
+class CustomScriptDialog(QtWidgets.QDialog):
```
jakubjezek001, line 25: Can we rewrite this it the way it uses nuke prefabricated ux or was there any particular reason we are not using them in this script? https://learn.foundry.com/nuke/developers/63/pythondevguide/custom_panels.html

jakubjezek001 (changes requested):

openpype/hosts/nuke/startup/custom_popup.py
moonyuet: We rewrite this with the nukescripts method for sure. At first I think it is possible to use `popup.py` panel in` openpype.widgets` as the base parent class for this case but it seems some of the functions dont expose and the position of the window can't be edited, so I abandoned to use "popup.py"

moonyuet:
Updated the changelog and the code for creating dialog through nukescripts instead of Qt.

64qam (approved):

PR #5042, opened 2023-05-26
Nuke: Allow for knob values to be validated against multiple values.

tokejepsen:
## Changelog Description
Knob values can now be validated against multiple values, so you can allow write nodes to be `exr` and `png`, or `16-bit` and `32-bit`.

## Testing notes:
1. Add `datatype` of `32 bit float` in `project_settings/nuke/imageio/nodes/requiredNodes/0/knobs/1/type`.
2. Create `render` instance in Nuke and modify the write node inside the output node, to use a different datatype.
3. Publish.

### DIFF
--- a/openpype/hosts/nuke/plugins/publish/validate_write_nodes.py
+++ b/openpype/hosts/nuke/plugins/publish/validate_write_nodes.py
@@ -1,3 +1,5 @@
+from collections import defaultdict
+
 import pyblish.api
 from openpype.pipeline.publish import get_errored_instances_from_context
 from openpype.hosts.nuke.api.lib import (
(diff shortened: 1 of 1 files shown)


moonyuet:
tried the validation with the reference targets , it works with `32 bit float` and some custom knob values.
image

Minkiu (approved):

openpype/hosts/nuke/plugins/publish/validate_write_nodes.py
```diff
@@ -105,28 +112,33 @@ def process(self, instance):
                 )
 
             key = knob_data["name"]
-            value = knob_data["value"]
+            values = values_by_name[key]
```
Minkiu, line 115: I'd probably change these to more meaningful names.

tokejepsen: Got any suggestions?

jakubjezek001 (approved):
CLG!

PR #5047, opened 2023-05-26
Feature: Resolve: Open last workfile on launch through .scriptlib 

BigRoy:
## Changelog Description

Added implementation to Resolve integration to open last workfile on launch.

## Additional info

This proposes an alternative implementation to what https://github.com/ynput/OpenPype/pull/5006 tries to solve.

Also added the option to directly launch the OpenPype menu on launch. Still have to add a toggle in settings to enable/disable that based on studio settings.

## Testing notes:

1. Launch resolve. It should open the last workfile directly if it exists (and open last workfile setting must be enabled, which is the default setting)

Or for a complete reproducable:

1. Select a task in a project
2. Launch Resolve on the task
3. Save the workfile using workfiles tool
4. Close resolve
5. Launch it again on the same task
6. The workfile should be opened on resolve launch


openpype/hosts/resolve/startup.py
```diff
@@ -0,0 +1,40 @@
...
+
+    # Activate resolve from openpype
+    install_host(openpype.hosts.resolve.api)
+
+    openpype.hosts.resolve.api.launch_pype_menu()
+
+
+def open_file(path):
+    # Avoid the need to "install" the host
+    openpype.hosts.resolve.api.bmdvr = resolve  # noqa
```
jakubjezek001, line 21: I am not sure about this. This seems to be quite hacky solution. Or can you explain more why do we need to do  it this way?

openpype/hosts/resolve/utility_scripts/openpype_startup.scriptlib
```diff
@@ -0,0 +1,22 @@
+-- Run OpenPype's Python launch script for resolve
+function file_exists(name)
+   local f = io.open(name, "r")
+   return f ~= nil and io.close(f)
+end
+
+
+openpype_root = os.getenv("OPENPYPE_ROOT")
+if openpype_root ~= nil then
+    script = openpype_root .. "/openpype/hosts/resolve/startup.py"
```
iLLiCiTiT, line 10: I would replce `OPENPYPE_ROOT` with resolve specific root, so the script is not based on location inside openpype.
```suggestion
resolve_root = os.getenv("OPENPYPE_RESOLVE_ROOT")
if resolve_root ~= nil then
    script = resolve_root .. "/startup.py"
```

This would require to set `OPENPYPE_RESOLVE_ROOT` e.g. in resolve addon.

```python
from .utils import RESOLVE_ROOT_DIR
...

class ResolveAddon(OpenPypeModule, IHostAddon):
    ...
    
    def add_implementation_envs(self, env, _app):
        env["OPENPYPE_RESOLVE_ROOT"] = RESOLVE_ROOT_DIR
```

openpype/hosts/resolve/startup.py
```diff
@@ -0,0 +1,40 @@
...
+# be opened at the minimum
+import openpype.hosts.resolve.api
+
+
+def launch_menu():
+    from openpype.pipeline import install_host
+    print("Launching Resolve OpenPype menu..")
+
+    # Activate resolve from openpype
+    install_host(openpype.hosts.resolve.api)
```
iLLiCiTiT, line 14: Shouldn't this happen in `main` before anything else?

BigRoy: I mostly _tried_ to delay it as much as possible since it's relatively slow and Resolve's startup UI is active during that time. So a user might be inclined to click to open a project even though the "open last workfile" bit might still.

It's good to note that this launch script runs in its own separate process - it doesn't influence the main Resolve process. So I'd recommend only ever triggering host install if whatever we're doing actually requires an installed host to begin with.

openpype/hosts/resolve/startup.py
BigRoy: I mostly _tried_ to speed up the launchijng as much as possible since it's relatively slow to trigger the install of the Resolve host and Resolve's startup UI is active during this time. So a user might be inclined to click to open a project even though the "open last workfile" bit might still.

As mentioned in a comment above even the importing of the python files alone took about 1.5 seconds for me (which might be due to the machine testing it from isn't running the code from an SSD)

openpype/hosts/resolve/startup.py
iLLiCiTiT: I though this scipt is actually triggered all the time, but it happens only if workfile should be opened right?

openpype/hosts/resolve/startup.py
BigRoy: I have now cleaned this up logically with a slower runtime as a result (so it'll take a bit longer before the workfile opens) but at least the code is somewhat simplified. Let me know how it works for you @jakubjezek001

openpype/hosts/resolve/startup.py
BigRoy: The startup script triggers on every start of Resolve - but workfile opening only happens if it's enabled, same for the startup menu start now as well.

jakubjezek001 (approved):
looks great thank you!

jakubjezek001:
It works as expected in case of the project is saved correctly at the project manager. It finds the project, open it and then starts the OpenPype menu. But once the project does not exists and it needs to be first started from Untitled project, then the OpenPype menu is not opening after the start. 

Lets merge this as it is and then I will create fix for this. I assume the best would be to separate those two scripts to do the two actions separatelly - open latest workfile, open menu after start. 

Also there is still the known bug which was solved in the https://github.com/ynput/OpenPype/pull/5006 PR. Workfile save file is not saving the file and only creates the project in project manager. Then the following issue is that the latest workfile is not created from the latest version, since no .drp is savet to the workfiles.

BigRoy:
> However, if the project doesn't exist and must initially be launched from an 'Untitled' project, the OpenPype menu fails to open post-startup.
> 
> We should proceed with merging this as it currently stands. Subsequently, I will address the issue. I believe the optimal solution would be to divide the actions into two separate scripts: one for opening the latest workfile and another for launching the menu after startup.

Interesting, because the menu opened fine for me even without it opening a workfile. (e.g. disable the opening of workfile explicitly in the launcher).

> Additionally, there remains a recognized bug that was previously addressed in https://github.com/ynput/OpenPype/pull/5006 PR. The issue lies in the fact that the workfile save file does not actually save the file, but only generates the project in the project manager. A subsequent issue arises as the latest workfile is not created from the latest version, since no .drp is saved to the workfiles.

Could you elaborate on **what** issue remains?

jakubjezek001:
> Could you elaborate on what issue remains?

It is explained in the linked issue.

PR #5051, opened 2023-05-26
Nuke: Create nodes with inpanel=False

Zipodod:
This PR is meant to remove the annoyance of the UI changing focus to the properties window just for the property window of the newly created node to disappear. Instead of using node.hideControlPanel I'm implementing the concealment during the creation of the node which will not change the focus of the current window.


Minkiu (approved):

mkolar (approved):

PR #5052, opened 2023-05-29
Maya: Deadline servers

tokejepsen:
## Changelog Description
Fix working with multiple Deadline servers in Maya.

- Pools (primary and secondary) attributes were not recreated correctly.
- Order of collector plugins were wrong, so collected data was not injected into render instances.
- Server attribute was not converted to string so comparing with settings was incorrect.
- Improve debug logging for where the webservice url is getting fetched from.

## Testing notes:
1. Add multiple webservices to `modules/deadline/deadline_urls`.
2. Setup render and use non-default webservice in Maya.
3. Publish.
4. Test enabling/disabling webservices in `project_settings/deadline/deadline_servers`.

### DIFF
--- a/openpype/hosts/maya/plugins/create/create_render.py
+++ b/openpype/hosts/maya/plugins/create/create_render.py
@@ -181,16 +181,34 @@ class CreateRender(plugin.Creator):
 
         primary_pool = pool_setting["primary_pool"]
         sorted_pools = self._set_default_pool(list(pools), primary_pool)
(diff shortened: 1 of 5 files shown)


LiborBatek (approved):
All works as expected.

When multiple deadline webservices defined its possible to choose in maya which one will be used during rendering.
image



image

Also tested with resulting full rendered finnished job via Deadline.

PR #5053, opened 2023-05-29
OCIO: Support working with single frame renders

tokejepsen:
## Changelog Description
When there is only 1 file, the datamember `files` on the representation should be a string.

## Testing notes:
1. Setup a profile for Maya in `project_settings/global/publish/ExtractOIIOTranscode/profiles`.
2. Setup a `render` in Maya and publish.

### DIFF
--- a/openpype/plugins/publish/extract_color_transcode.py
+++ b/openpype/plugins/publish/extract_color_transcode.py
@@ -184,6 +184,11 @@ class ExtractOIIOTranscode(publish.Extractor):
                     if tag == "review":
                         added_review = True
 
+                # If there is only 1 file outputted then convert list to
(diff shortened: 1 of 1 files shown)


moonyuet (approved):
code looks good, i also did some  simple test in review and rendering main, it works smoothly.
image

image

moonyuet:
> > code looks good, i also did some simple test in review and rendering main, it works smoothly.
> 
> Sorry, should have added to the testing notes that this fixes issues with single frame renders.

Also tested in the single frame. but just in case it will break something so I also test with multiple frame and different families.
image

iLLiCiTiT (approved):
LGTM

PR #5055, opened 2023-05-29
Maya: Repair RenderPass token when merging AOVs.

tokejepsen:
## Changelog Description
Validator was flagging that `<RenderPass>` was in the image prefix, but did not repair the issue.

## Testing notes:
1. Setup render in Maya with `Merge AOV` enabled.
2. Ensure `<RenderPass>` token is in image prefix.
3. Publish and repair.


moonyuet (approved):
code looks okay for me. Also did a simple test and it doesn't break thing so far and remove renderpass token once the repair action is performed.

LiborBatek (approved):
Set the name for renders so it contains ```<RenderPass>``` token

image

Validations correctly kicks in and ```Repair``` action works fine and correctly sets the naming when ```Merge AOVs``` in use.

image

antirotor (approved):

PR #5059, opened 2023-05-30
General: Qt scale enhancement

iLLiCiTiT:
## Changelog Description
Set ~~'QT_SCALE_FACTOR_ROUNDING_POLICY'~~ scale factor rounding policy of QApplication to `PassThrough` so the scaling can be 'float' number and not just 'int' (150% -> 1.5 scale).

## Additional info
This actually looks like very good solution sugested in [community forum](https://community.ynput.io/t/openpype-ui-scaling/237/3). We have to check if the variable affects Qt DCCs like Houdini, Nuke, Maya.

## Testing notes:
If you have scaling set in OS it should not make too large/small windows of tools.


movalex:
I realized that setting `QT_SCALE_FACTOR_ROUNDING_POLICY` globally may affect other apps' look, so no need to check is it is already set globally. I suggest setting this environment from local settings, so everyone could have their own rounding settings.

https://github.com/ynput/OpenPype/compare/release/next-minor...movalex:OpenPype:enhancement/qt_scale_enhancement

iLLiCiTiT:
To be honest I don't think this should be in local settings. Absolutelly 0 artists will be able to fill there something right + AYON local settings won't work the same way.

If you're worried about other apps, we can do it more explicitly. Not by using env variable, but using `setHighDpiScaleFactorRoundingPolicy`.

mkolar (approved):

BigRoy:
Does the `setHighDpiScaleFactorRoundingPolicy` call really help much with avoiding the issues with the `QT_SCALE_FACTOR_ROUNDING_POLICY` since the majority of the Qt-based applications are using the existing QApplication instance to begin with so even if we were triggering `setHighDpiScaleFactorRoundingPolicy` we'd still be setting it on the parent application. Right? :)

iLLiCiTiT:
> Which would also mean that for e.g. Maya and alike the setting of the attributes would never trigger and thus potentially not resolve the root of the current issue? (Or is that exactly the behavior we're now looking for so that it doesn't do anything in Maya, Houdini, etc.?)

I would not touch Qt definitions inside Maya na Nuke -> that's why the env is not set. The behavior is DCC version dependent and behaves in a different manner which is dangerous and should be handled in other way if needed. This PR is fixing OpenPype process UIs.

movalex:
> To be honest I don't think this should be in local settings. Absolutelly 0 artists will be able to fill there something right + AYON local settings won't work the same way.

Sure, you're right. My point was, if we make it a setting, some artists, maybe a few, can control the UI scale with this variable. Some would benefit from larger UI scale. But I agree this is not user friendly solution at all. 
PS: And I also love seeing how all this repeating code is melting away with this PR

PR #4854, opened 2023-04-18
Maya: Improve Templates

tokejepsen:
## Changelog Description
Use library method for fetching reference node and support parent in hierarchy.

## Discussion
The template code was not using library methods and was hardcoded to reference nodes ending with `_RN`. Now the issue comes with cyclic imports if using the `get_reference_node` from `maya.api.plugin` because it imports `maya.api.pipeline` which imports `maya.api.workfile_template_builder`. At the moment I have place a duplicate of the code in `maya.api.lib` but that is not ideal.

## Tasks

- [x] use library method for fetching reference node.
- [x] support parenting into hierarchy.
- [x] building template is not respecting settings like `Keep placeholders` or `Create first version`.
- [x] rename `cleanup_placeholder` to `post_representation_load` to reflect its place in the build process.

## Testing notes:
1. Create model (abc) placeholder.
2. Setup a placeholder in a hierarchy and save workfile.
3. Setup lookdev template with above workfile.
4. Build lookdev workfile from template.
5. Test settings `project_settings/maya/templated_workfile_build/profiles/0/keep_placeholder` and `project_settings/maya/templated_workfile_build/profiles/0/create_first_version`

PR #5061, opened 2023-05-30
Nuke: Fix workfile template builder so representations get loaded next to each other

fabiaserra:
## Changelog Description
Refactor when the cleanup of the placeholder happens for the cases where multiple representations are loaded by a single placeholder.

The existing code didn't take into account the case where a template placeholder can load multiple representations so it was trying to do the cleanup of the placeholder node and the re-arrangement of the imported nodes too early. I assume this was designed only for the cases where a single representation can load multiple nodes.

Screenshot 2023-05-23 170233

## Additional info
Removed the `failed` argument from `cleanup_placeholder` as it wasn't being used, we can add it if it's ever needed.

## Testing notes:
1. Create a Nuke workfile template with a placeholder that will load multiple representations (i.e. Using Load Clip loader for example)
2. Save workfile template at a place where project Nuke workfile template builder settings looks for
3. Create workfile from template


Minkiu (approved):
Haven't tested, code looks  good. I trust you replaced all instances of the changed methods hehe

fabiaserra:
> Haven't tested, code looks good. I trust you replaced all instances of the changed methods hehe otherwise might be worth keeping the arguments as optional and issue a warning of deprecation.

Yeah good call, apparently someone cared even less than me for backwards compatibility and changed the whole function name here https://github.com/ynput/OpenPype/pull/4854 haha so this comment no longer really applies...

I have now merged `develop` over and resolved the merge conflicts and I have kept the existing function signature with the redundant `failed` arg that no one is using...

openpype/pipeline/workfile/workfile_template_builder.py
```diff
@@ -1549,24 +1559,23 @@ def populate_load_placeholder(self, placeholder, ignore_repre_ids=None):
...
             except Exception:
-                failed = True
                 self.load_failed(placeholder, representation)
 
             else:
-                failed = False
                 self.load_succeed(placeholder, container)
-            self.post_placeholder_process(placeholder, failed)
+        # Run post placeholder process after load of all representations
+        self.post_placeholder_process(placeholder, failed)
```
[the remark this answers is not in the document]
fabiaserra: and yeah, this was going to be my next point, `failed` is undefined here but so it is on the next line that checks `if failed`

openpype/pipeline/workfile/workfile_template_builder.py
```diff
@@ -1549,24 +1559,23 @@ def populate_load_placeholder(self, placeholder, ignore_repre_ids=None):
...
 
         if failed:
```
fabiaserra, line 1580: here `failed` doesn't necessarily exist on this scope. @tokejepsen tagging you as you were the one who made that commit, what were you expecting this to be like? if the first representation fails but the next one works, the value of `failed` gets overridden

fabiaserra: I have now pushed here https://github.com/ynput/OpenPype/pull/5061/commits/ac7e53a44a038df8c31ad1fbd4d49c8c90c3a95b a change so `failed` is defined on the right scope and it only gets overridden to `True` if it fails for any of the load calls

jakubjezek001 (approved):
Tested and works well! Thanks @fabiaserra. Since Maya changes are only related to docstrings changes. I am assuming we can merge it!

openpype/pipeline/workfile/workfile_template_builder.py
jakubjezek001: It seems to be resolved @fabiaserra, right?

PR #5070, opened 2023-05-31
Unreal: Added verification for Unreal app name format

### LINKED ISSUE #5062: Unreal: Patch number in UE version breaks integration

### Current Behavior:

When Unreal Engine has patch version inside - like 5.1.1, it will  break logic trying to get compatible integration here:

https://github.com/ynput/OpenPype/blob/5fbae39a745b0016b2ada9d054b696aef2007fee/openpype/hosts/unreal/lib.py#L323-L361

### Expected Behavior:

This logic should work with unreal editor with patch version.

### Version

3.15.9-nightly.1

Windows

### Steps To Reproduce:

1. Install and configure Unreal Editor 5.1.1
2. Try to launch it on task without project

### Relevant log output:

_No response_

### Additional context:

_No response_

simonebarbieri:
## Changelog Description
The Unreal app name is used to determine the Unreal version folder, so it is necessary that if follows the format `x-x`, where `x` is any integer. This PR adds a verification that the app name follows that format.

## Additional info
Fixes #5062

## Testing notes:
1. Add a new version of Unreal in OpenPype System Settings, and add as key any string that does not respect the format.
2. Add that version to `project_anatomy/attributes/applications`.
3. Try creating a new Unreal project opening a task with that new Unreal version you added. It should show an error.

### DIFF
--- a/openpype/hosts/unreal/addon.py
+++ b/openpype/hosts/unreal/addon.py
@@ -1,5 +1,7 @@
 import os
+import re
 from openpype.modules import IHostAddon, OpenPypeModule
+from openpype.widgets.message_window import Window
 
 UNREAL_ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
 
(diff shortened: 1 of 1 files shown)


antirotor:
Maybe we should add where this name is set? Because it is a key in the settings. I need to check it out with AYON how it is namend there, but I think it is still a key, so maybe we can just change the message to:

`Unreal application key in the settings must be in format '5-0' or '5-1'`

What do you think?

PR #5071, opened 2023-05-31
Nuke: small tweeks for Ayon

jakubjezek001:
## Changelog Description
- renaming OpenPype menu items to Ayon
- removing instance node if removed via publisher
- removing info knob since it is not needed any more (was there only for the transition phase)

## Testing notes:
1. nuke should create instance nodes without info message
2. Nuke should remove node once is bin icon used in Publisher on particular instance
3. Menu `OpenPype Tools` is now showing as `Ayon Tools`


jakubjezek001:
implemented with https://github.com/ynput/OpenPype/pull/5083

PR #5072, opened 2023-05-31
CI: WPS linting instead of Hound (rebase)

antirotor:
## Changelog Description
Because Hound currently used to lint the code on GH ships with really old flake8 support, it fails miserably on any newer Python syntax. This PR is adding WPS linter to GitHub workflows that should step in.

## Additional info
WPS linter suit is checking now even docstrings and it seems much more stricter than Hound, so it is a matter of discussion to set its limits.

## Testing notes:
Try this configuration file with your PRs and see its feedback. It should post code comments the same way as Hound does.

> **Note**
> This is rebased PR #4934 without pre-commit hook.


jakubjezek001 (approved):
LGTM

jakubjezek001:
These need to be re-based (rather cherry-picked) and quickly merged, please > `next-minor` is really an unstable branch to work with @antirotor ;)

antirotor:
because of the rebase of the target branch, superseded by #5072

PR #5077, opened 2023-06-01
Auto generate first workfile with template builder

friquette:
## Changelog Description
When opening a task, if a workfile doesn't already exists, call the template builder to auto generate the first workfile.

## Additional info
Here's what I have for now. There's a cyclic import so I can't test the efficiency of the code, and I don't know how to solve this problem of cyclic imports.


OlivierOsotimehin:
There is already a settings "create first version" that does something else : it auto save the workfile after it have been built (the name is badly chosen)

So we should need a settings autobuild first version ?

tokejepsen:
@friquette what is the plan with this PR?
Sounds like it could do with waiting on #4854, to ensure the `create first version` actually works. Then we'll need a setting and the cyclic imports resolved.

ClementHector:
@friquette Can you test that https://github.com/ynput/OpenPype/pull/4854 solves this issue please.

friquette:
> @friquette what is the plan with this PR? Sounds like it could do with waiting on #4854, to ensure the `create first version` actually works. Then we'll need a setting and the cyclic imports resolved.

The point of my PR is to auto execute the template builder if there is no workfile for the current task. That's why I created a new setting "Autobuild first version".

I tested your PR #4854 and it seems that "Create first version" doesn't do what I expected from the name of the setting (in fact I don't see what it does at all). But if the point was to do the exact same thing as me, so we can wait for you to implement it :)

tokejepsen:
@friquette with #4854 merged, what is the plan with this PR? Is it needed anymore?

friquette:
> @friquette with #4854 merged, what is the plan with this PR? Is it needed anymore?

Yes it is needed. I tested with the #4854 merged and the `Create first version` is still not doing what we want.

OlivierOsotimehin:
If i followed well, the original create first version is supposed to save a version on disk when building the workfile ?. 
We(Fixstudio) want to autobuild a first version when there is no work file yet.

tokejepsen:
Sounds good.

@friquette is this ready to test? Cyclic import resolved?

friquette:
> Sounds good.
> 
> @friquette is this ready to test? Cyclic import resolved?
Yes it's all good, it should work fine now

friquette:
> Sounds good.
> 
> @friquette is this ready to test? Cyclic import resolved?

Yes it's ready to test, it should work fine now

openpype/hosts/maya/api/lib.py
```diff
@@ -2366,6 +2366,29 @@ def set_context_settings():
...
+    if build_workfile_profiles:
+        for profile in build_workfile_profiles:
+            is_task_name = task_name in profile['task_names']
+            is_task_type = current_tasks[task_name]['type'] in \
+                profile['task_types']
+            is_create_first_version = profile['autobuild_first_version']
+
+            if (is_task_name or is_task_type) and is_create_first_version:
+                from .workfile_template_builder import build_workfile_template
+                build_workfile_template()
```
BigRoy, line 2389:
```suggestion
    project_name = get_current_project_name()
    project_settings = get_project_settings(project_name)
    
    build_workfile_profiles = project_settings['maya']['templated_workfile_build'].get('profiles')  # noqa
    if not build_workfile_profiles:
        return

    asset_name = get_current_asset_name()
    asset = get_asset_by_name(project_name, asset_name)
    task_name = get_current_task_name()
    current_tasks = asset.get("data").get("tasks")
    task_type = current_tasks[task_name]['type']

    for profile in build_workfile_profiles:
        if not profile['autobuild_first_version']:
            # Do not use this profile for autobuild
            continue
    
        is_task_name = task_name in profile['task_names']
        is_task_type = task_type in profile['task_types']

        if is_task_name or is_task_type:
            # import here to avoid cyclic imports
            from .workfile_template_builder import build_workfile_template
            build_workfile_template()
```
Some optimization by opting out early whenever we can and hopefully restructures it to be more readable too.

Also, shouldn't this use regular profiles filtering for consistency? Using the `filter_profiles` function from OpenPype?

openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -14,7 +14,6 @@
     WorkfileBuildPlaceholderDialog,
 )
 
-from .lib import read, imprint
```
BigRoy, line 17: Where is this PR creating a circular/cyclic import even though the only import changing is this one? I don't see any imports being added and thus that means any cycling imports must have already existed prior to this PR; right?

As far as I know you can revert _this_ change if you only keep the `from .workfile_template_builder import build_workfile_template` import as you have refactored in the function in lib. Right?

BigRoy (commented):
Just want to add that I think it's weird that this implementation seems to alter something specific to maya even though technically the change in project settings adds the setting to ALL hosts with workfile template settings. Shouldn't we implement this way more agnostic than focused at maya? It doesn't seem to touch much maya code either.

openpype/pipeline/workfile/workfile_template_builder.py
```diff
@@ -1825,3 +1832,40 @@ def get_errors(self):
 
     def create_failed(self, creator_data):
         self._failed_created_publish_instances.append(creator_data)
+
+
+def build_first_workfile_from_template_builder():
+    print("#" * 100)
```
BigRoy, line 1838:
```suggestion
```

openpype/pipeline/workfile/workfile_template_builder.py
```diff
@@ -1825,3 +1832,40 @@ def get_errors(self):
 
     def create_failed(self, creator_data):
         self._failed_created_publish_instances.append(creator_data)
+
+
+def build_first_workfile_from_template_builder():
```
BigRoy, line 1837: This function name doesn't sound right? This returns whether it should build - it doesn't return a "builder"?
Additionally it should allow to take in settings, asset and task name, etc. so that if you want to optimize the call with settings etc that you might've already retrieved, then you can.

```
def should_create_first_workfile(
        project_name=None,
        project_settings=None,
        asset_doc=None,
        asset_name=None,
        task_name=None,
        host_name=Nome
    ):
    """Return whether first workfile should be created for given context"""
```
    project_name = project_name or get_current_project_name()
    if project_settings is None:
     etc.
```

I'd also move the `if build_workfile_profiles["profiles"]:` way higher up before retrieving more context things (directly after getting project settings) so that if no profiles are defined you don't need to do any other querying.

File: openpype/pipeline/workfile/workfile_template_builder.py [hunk 5]
```diff
@@ -1825,3 +1832,40 @@ def get_errors(self):
...
+    task_name = get_current_task_name()
+    current_tasks = asset.get("data").get("tasks")
+    task_type = current_tasks[task_name]['type']
+    host = get_current_host_name()
+
+    build_workfile_profiles = project_settings[host]['templated_workfile_build']  # noqa
+
+    filtering_criteria = {
+        "task_names": task_name,
+        "task_types": task_type
```
BigRoy, line 1853: Is it correct that the key names are plural but the variables are singular?

openpype/hosts/maya/api/workfile_template_builder.py
BigRoy: I don't think this moved import is still necessary, right? Or is it?

openpype/hosts/maya/api/pipeline.py
```diff
@@ -627,6 +627,14 @@ def on_new():
...
             "from openpype.hosts.maya.api import lib;"
             "lib.add_render_layer_change_observer()")
+        autobuild_first_workfile = """
+from openpype.pipeline.workfile import workfile_template_builder
+from openpype.hosts.maya.api.workfile_template_builder import build_workfile_template  # noqa
+builder = workfile_template_builder.build_first_workfile_from_template_builder()  # noqa
+if builder:
+    build_workfile_template()
+"""
+        cmds.evalDeferred(autobuild_first_workfile)
```
BigRoy, line 637: Oh boy. :) 

Let's expose a method directly that just does:
```
def auto_build_first_workfile():
    if should_build_first_workfile():
        build_workfile_template()
```
☝️ And that function I suspect should be in the global pipeline workfile.

And for readability just trigger it directly through `maya.utils.executeDeferred(auto_build_first_workfile)` so you don't need to work with re-importing and the whole messy string management. A similar thing is done [here](https://github.com/ynput/OpenPype/pull/5226/files#diff-8481a901810ec3f718a093664b6877423345730dc3719122846c4c83a9ff34d5R560-R572).

friquette: Thank you for your advice, I was kind of stuck on this part. I knew this was a little bit messed up, but I didn't found an other way to do it :)

openpype/pipeline/workfile/workfile_template_builder.py
friquette: In the settings we can add multiple task names and task types for a single profile. But here we are passing the current task name and the current task type, so there will be only single value

tokejepsen:
> Just want to add that I think it's weird that this implementation seems to alter something specific to maya even though technically the change in project settings adds the setting to ALL hosts with workfile template settings. Shouldn't we implement this way more agnostic than focused at maya? It doesn't seem to touch much maya code either.

I agree.
@friquette can we move the settings to the Maya section and implement the code inside of the Maya directory?

tokejepsen:
@friquette did you get a chance to look over this PR again?

friquette:
> > Just want to add that I think it's weird that this implementation seems to alter something specific to maya even though technically the change in project settings adds the setting to ALL hosts with workfile template settings. Shouldn't we implement this way more agnostic than focused at maya? It doesn't seem to touch much maya code either.
> 
> I agree. @friquette can we move the settings to the Maya section and implement the code inside of the Maya directory?

Isn't it the opposite of what BigRoy said ? I though he wanted the code to be less specific to maya ? I moved the function in another file to be more agnostic because this isn't really specific to maya and can be implemented for other dcc easily

openpype/hosts/maya/api/pipeline.py
```diff
@@ -625,10 +629,26 @@ def _on_show_inventory():
...
     with lib.suspended_refresh():
+        cmds.evalDeferred(
+            "from openpype.hosts.maya.api import lib;"
+            "lib.remove_render_layer_observer()")
+        cmds.evalDeferred(
+            "from openpype.hosts.maya.api import lib;"
+            "lib.add_render_layer_observer()")
+        cmds.evalDeferred(
+            "from openpype.hosts.maya.api import lib;"
+            "lib.add_render_layer_change_observer()")
```
tokejepsen, line 649: Dont think this is needed for this PR?

tokejepsen (changes requested):

tokejepsen:
https://github.com/ynput/OpenPype/pull/5077#pullrequestreview-1504851074
@BigRoy could you weigh in and clarify on this comment?

I assume you mean that the functional code in `openpype/hosts/maya/api/pipeline.py` is specific to Maya, but the settings is not specific to Maya. So instead of trying to implement it for all hosts, we'll limit it to Maya only and the settings should reflect that, to be only within the Maya settings.

openpype/hosts/maya/api/pipeline.py
BigRoy: Actually this should be removed because it's re-introducing legacy code that was removed on purpose.

openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -183,10 +184,18 @@ def create_placeholder(self, placeholder_data):
...
+        # TODO: this can crash if selection can't be used
+        cmds.parent(placeholder, selection[0])
+
+        # get the long name of the placeholder (with the groups)
+        placeholder_full_name = (
+            cmds.ls(selection[0], long=True)[0]
+            + "|"
+            + placeholder.replace("|", "")
+        )
+        from .lib import imprint
```
BigRoy, line 197: I still don't understand why this import needs to be moved here?

openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -183,10 +184,18 @@ def create_placeholder(self, placeholder_data):
         placeholder_name = self._create_placeholder_name(placeholder_data)
 
         placeholder = cmds.spaceLocator(name=placeholder_name)[0]
-        if parent:
-            placeholder = cmds.parent(placeholder, selection[0])[0]
 
-        imprint(placeholder, placeholder_data)
+        # TODO: this can crash if selection can't be used
+        cmds.parent(placeholder, selection[0])
```
BigRoy, line 189: This will indeed fail to work without `parent`. But why not just keep the `if statement` logic that was there before. That code seems just fine. 
Except that I'd write it as:

```suggestion
        if parent:
            placeholder = cmds.parent(placeholder, parent)[0]
```

openpype/hosts/maya/api/pipeline.py
BigRoy, line 649:
```suggestion
```

openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -14,7 +14,7 @@
     WorkfileBuildPlaceholderDialog,
 )
 
-from .lib import read, imprint, get_reference_node, get_main_window
+from .lib import get_reference_node, get_main_window
```
BigRoy, line 17:
```suggestion
from .lib import read, imprint, get_reference_node, get_main_window
```

openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -108,6 +108,7 @@ def _collect_scene_placeholders(self):
         return placeholder_nodes
 
     def _parse_placeholder_node_data(self, node_name):
+        from .lib import read
```
BigRoy, line 111:
```suggestion
```

openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -183,10 +184,18 @@ def create_placeholder(self, placeholder_data):
...
+        cmds.parent(placeholder, selection[0])
+
+        # get the long name of the placeholder (with the groups)
+        placeholder_full_name = (
+            cmds.ls(selection[0], long=True)[0]
+            + "|"
+            + placeholder.replace("|", "")
+        )
+        from .lib import imprint
+        imprint(placeholder_full_name, placeholder_data)
```
BigRoy, line 198:
```suggestion
        imprint(placeholder, placeholder_data)
```

openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -216,7 +226,7 @@ def update_placeholder(self, placeholder_item, placeholder_data):
 
         for key in new_values.keys():
             cmds.deleteAttr(node_name + "." + key)
-
+        from .lib import imprint
```
BigRoy, line 229:
```suggestion
```

openpype/pipeline/workfile/workfile_template_builder.py
```diff
@@ -1871,3 +1878,51 @@ def get_errors(self):
...
+        asset_doc=None,
+        asset_name=None,
+        task_name=None,
+        host_name=None
+):
+    """Return whether first workfile should be created for given context"""
+
+    project_name = project_name or get_current_project_name()
+    if project_settings is None:
+        project_settings = get_project_settings(project_name)
```
BigRoy, line 1895:
```suggestion
    if project_settings is None:
        project_name = project_name or get_current_project_name()
        project_settings = get_project_settings(project_name)
```

openpype/pipeline/workfile/workfile_template_builder.py
```diff
@@ -1871,3 +1878,51 @@ def get_errors(self):
...
+    )
+
+    if not profile or not profile.get("autobuild_first_version"):
+        return False
+
+    is_task_name = task_name in profile['task_names']
+    is_task_type = task_type in profile['task_types']
+
+    if not is_task_name and not is_task_type:
+        return False
```
BigRoy, line 1926: Shouldn't the `filter_profiles` logic itself already handle this? If there's specific reason to have this here we should add a comment why it's needed.

BigRoy:
> I assume you mean that the functional code in openpype/hosts/maya/api/pipeline.py is specific to Maya, but the settings is not specific to Maya. So instead of trying to implement it for all hosts, we'll limit it to Maya only and the settings should reflect that, to be only within the Maya settings.

Yes, somewhat. I do think however we can make it even further agnostic if for each host we just emit the `init` event or alike on launch. And then in core (outside of host logic) listen to that event to trigger the workfile template build - which I think can be done completely software agnostic from that point.

Similar to how `any_outdated` pop-up could potentially be made software agnostic if OpenPype would emit its own events per host, e.g. scene opened, and then in core just listens to that and then does:
```
if any_outdated():
    show_popup()
```

The _only_ issue there is that we don't have a host-agnostic means of detecting what the potential Qt parent might be for the current host. And there _may_ be specific hosts that require certain hacks for the UIs to work (blender maybe?).

We don't need to make it agnostic now, but it'd be much prefered if we implement this once in a way that it works just against the Host class API of OpenPype and thus works for any host implementing it. Then all we need to do for future hosts is emit the right event, instead of implementing the any outdated popup, the workfile build, etc. etc.

mkolar:
Because we're splitting OpenPype into ayon-core and individual host addons, this PR would have to be ported to target one of those.

PR #5078, opened 2023-06-01
Resolve: Make sure scripts dir exists

iLLiCiTiT:
## Changelog Description
Make sure the scripts directory exists before looping over it's content.

## Additional info
The issue appeared to me on first launch.

## Testing notes:
1. Remove utils script dir on your workstation
2. Launch Resolve -> That should not crash

### DIFF
--- a/openpype/hosts/resolve/utils.py
+++ b/openpype/hosts/resolve/utils.py
@@ -29,6 +29,10 @@ def setup(env):
     log.info("Utility Scripts Dir: `{}`".format(util_scripts_paths))
     log.info("Utility Scripts: `{}`".format(scripts))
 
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
The change makes sense. I assume this is the case if Resolve has never run before? (Like you run it the first time after install directly through OpenPype?)

I wonder if doing this might however make it harder to find cases where the scripts dir isn't set to a valid path to begin with. (Admittedly the startup logic already feels **very** dangerous if the admin by accident in system settings > applications or by some other means changes the value of `RESOLVE_UTILITY_SCRIPTS_DIR` env var and it just always killing whatever files and folders are in there. Definitely not set that path to the root of your project or production server hehe)

PR #5083, opened 2023-06-01
removing info knob from nuke creators

jakubjezek001:
## Changelog Description
- removing instance node if removed via publisher
- removing info knob since it is not needed any more (was there only for the transition phase)

## Testing notes:
1. nuke should create instance nodes without info message
2. Nuke should remove node once is bin icon used in Publisher on particular instance


barklaya (approved):
Everything works as it should

PR #5084, opened 2023-06-01
Nuke: custom settings for write node without publish

moonyuet:
## Changelog Description
Set Render Output and other settings to write nodes for non-publish purposes.

## Additional info.
Still work in progress

## Testing notes:
1. Launch Nuke via OP
2. Select some write nodes
3. Go to Openpype tool
4. Click Set Render Output for Write Node


openpype/hosts/nuke/startup/custom_write_node.py
```diff
@@ -0,0 +1,81 @@
...
+                255
+            ]
+        },
+        {
+            "type": "text",
+            "name": "channels",
+            "value": "rgb"
+        },
+        {
+            "type": "text",
```
jakubjezek001, line 48: Lets remove the colorspace knob. It would be better if the colorspace value will be taken from the workfile defaults. I had just tested it on a project set to ACES and this didn't work that well.

openpype/hosts/resolve/utility_scripts/openpype_startup.scriptlib
```diff
@@ -18,4 +18,4 @@ if openpype_startup_script ~= nil then
     else
         print("Launch script not found at: " .. script)
     end
-end
```
jakubjezek001, line 21: I presume this was committed by mistake.

jakubjezek001 (approved):
after the comments will be addressed it is ready to be merged ;)

openpype/hosts/resolve/utility_scripts/openpype_startup.scriptlib
moonyuet: This script has some conflict of missing the line with my branch after merging with the develop

jakubjezek001: but this should not be part of this PR. Please reverse the change back to the state before the change was made.

jakubjezek001 (changes requested):

moonyuet:
> @moonyuet the issue is coming from the following commit [7075d5c](https://github.com/ynput/OpenPype/commit/7075d5c4452dc0ea1c9f5addf3b96a75aef0ae55)

I have excluded the script from this PR

PR #5085, opened 2023-06-01
Tray: Fix restart arguments on update

iLLiCiTiT:
## Changelog Description
Fix arguments on restart.

## Additional info
First argument in `sys.argv` is always removed. When running from code, it contains `start.py` script path and when running from build it contains path to openpype executable. The argument of executable can be what was passed to terminal so it could be `./openpype_console` (on linux).

## Testing notes:
Update and restart from tray should work. To replicate
1. Make sure you have at least one version in update repository
2. Choose the latest version in update repository and make sure it is not available in your local versions in local app data
3. Run openpype with different version using `--use-version={lower version}` (this version must have changes from this PR)
4. It should ask to update and restart -> condifm to update and restart
5. OpenPype should start when update finished

### DIFF
--- a/openpype/tools/tray/pype_tray.py
+++ b/openpype/tools/tray/pype_tray.py
@@ -633,10 +633,10 @@ class TrayManager:
 
         # Create a copy of sys.argv
         additional_args = list(sys.argv)
-        # Check last argument from `get_openpype_execute_args`
-        # - when running from code it is the same as first from sys.argv
-        if args[-1] == additional_args[0]:
-            additional_args.pop(0)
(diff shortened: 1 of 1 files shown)


fabiaserra:
I can confirm I have been bit by this in the past and I ended up doing this hacky patch on my fork (as we launch openpype by having it on the $PATH with `openpype_console`):
```
                if arg in ["--use-staging", "openpype_console"] or arg.startswith("--use-version"):
                    continue
```
but that still didn't work for the run from source `start.py` so I'm glad to see this fix!

PR #5087, opened 2023-06-01
General: Show user who created the workfile on the details pane of workfile manager

fabiaserra:
## Changelog Description
Small addition to add name of the `user` who created the workfile on the details pane of the workfile manager:
Screenshot 2023-05-17 161521

## Testing notes:
1. Open workfile manager
2. You should see on the details pane a `User` entry showing who created that file


openpype/tools/workfiles/window.py
```diff
@@ -134,7 +135,9 @@ def set_context(self, asset_id, task_name, filepath, workfile_doc):
             "<b>Created:</b>",
             creation_time.strftime(datetime_format),
             "<b>Modified:</b>",
-            modification_time.strftime(datetime_format)
+            modification_time.strftime(datetime_format),
+            "<b>User:</b>",
+            pwd.getpwuid(filestat.st_uid).pw_name
```
BigRoy, line 140: Nice! Isn't pwd UNIX only which means this will fail on Windows? We might need to get a Windows equivalent solution as well?

antirotor (changes requested):
yes it doesn't work on windows and in fact, it is breaking OpenPype there because it will crash during host initialization on missing pwd module. Getting file ownership on Windows from Python is PITA. This is quick "multiplatform" snipped I've put together:

```python
def get_user_name(file: str) -> str:
    if platform.system() == "Windows":
        import win32security
        sd = win32security.GetFileSecurity(
            file, win32security.OWNER_SECURITY_INFORMATION)
        owner_sid = sd.GetSecurityDescriptorOwner()
        name, domain, type = win32security.LookupAccountSid(None, owner_sid)
        return f"{name}@{domain}"
    else:
        import pwd
        filestat = os.stat(file)
        return pwd.getpwuid(filestat.st_uid).pw_name
```

but the Windows part depends on `win32security` that is not standard part of Python and needs to be added to OP dependencies. 

In that case it needs to be added to `pyproject.toml` and `poetry.lock` and this PR should target for `release/next-minor`.

openpype/tools/workfiles/window.py
fabiaserra: > yes it doesn't work on windows and in fact, it is breaking OpenPype there because it will crash during host initialization on missing pwd module. Getting file ownership on Windows from Python is PITA. This is quick "multiplatform" snipped I've put together:
> 
> ```python
> def get_user_name(file: str) -> str:
>     if platform.system() == "Windows":
>         import win32security
>         sd = win32security.GetFileSecurity(
>             file, win32security.OWNER_SECURITY_INFORMATION)
>         owner_sid = sd.GetSecurityDescriptorOwner()
>         name, domain, type = win32security.LookupAccountSid(None, owner_sid)
>         return f"{name}@{domain}"
>     else:
>         import pwd
>         filestat = os.stat(file)
>         return pwd.getpwuid(filestat.st_uid).pw_name
> ```
> 
> but the Windows part depends on `win32security` that is not standard part of Python and needs to be added to OP dependencies.
> 
> In that case it needs to be added to `pyproject.toml` and `poetry.lock` and this PR should target for `release/next-minor`.

nice, thanks for the snippet! (damn windows..)

I will add this and target `next-minor` instead

fabiaserra: in the future it would be nice to add a way to filter workfiles by user as well and perhaps introduce a way to query the full user name from the SSO so it's more readable

fabiaserra: @antirotor done here https://github.com/ynput/OpenPype/pull/5087/commits/499071eda208606379dd3aab4dca3505790ff7e2 ! Tested in Linux and it still works fine, if you could try it on Windows please

openpype/tools/workfiles/window.py
```diff
@@ -29,11 +30,7 @@
 
 def file_size_to_string(file_size):
     size = 0
-    size_ending_mapping = {
-        "KB": 1024 ** 1,
-        "MB": 1024 ** 2,
-        "GB": 1024 ** 3
-    }
+    size_ending_mapping = {"KB": 1024**1, "MB": 1024**2, "GB": 1024**3}
```
fabiaserra, line 33: I ran `black` on the document and it automatically reformat some of these, I can revert it back

fabiaserra: Reverted so it only affects the lines I touched

antirotor (changes requested):
I've tested it and had hard times to get `win32security` to work in different python hosts (like Nuke 14). Seems that this module is heavily dependend on compatible python binaries, so I guess we'll have hard time to make it work across all hosts.

There are probably two ways:
1) Limit it so this featore only works on linux
2) Use `ctypes` to call windows api directly - that will get rid of the `win32security`

fabiaserra:
> I've tested it and had hard times to get `win32security` to work in different python hosts (like Nuke 14). Seems that this module is heavily dependend on compatible python binaries, so I guess we'll have hard time to make it work across all hosts.
> 
> There are probably two ways:
> 
> 1. Limit it so this featore only works on linux
> 2. Use `ctypes` to call windows api directly - that will get rid of the `win32security`

Your call, we use Linux so I don't mind much if it only works there. After a quick look at the implementation using `ctypes` it seems quite over-complicated... do you have some snippet?

What errors are you getting in Nuke 14? could we perhaps wrap it on a try/except and ignore the hosts that don't work?

fabiaserra:
> > What errors are you getting in Nuke 14? could we perhaps wrap it on a try/except and ignore the hosts that don't work?
> 
> Nuke 14 doesn't want to import `win32security` at all. Further digging leads to error that it cannot import `win32crypto` that is I think part of th `pywin`. All of these are distributed as **pyd** files bound to specific version of python. And even it I can use it without problems in PyCharm and Python coming with OpenPype, it doesn't work in Nuke 14 (with Python 3.9.10 - I haven't tried it in other hosts with different python version, but my guess is that we would have to vendor pywin and `win32security` for every python in every supported host).

Got it, makes sense.

I saw some online references installing `pywin32` instead of `win32security`... perhaps that would bring all the dependencies needed?

> So I think the overhead of making it work on Windows is big one and because this will be obsolete with AYON (with its user management system), I don't really mind having it working only on Linux (possibly mac) and not showing the user on Windows.

Yeah definitely, I will add the condition just for Linux then for now.

Thank you!!

fabiaserra:
@antirotor I changed it back to just work on Unix and reverted the changes to add the dependency, thus targeting `develop` again

fabiaserra:
> @antirotor I changed it back to just work on Unix and reverted the changes to add the dependency, thus targeting `develop` again

And stupid me to think git would just pick up my commits back to the target and not bring all of `next-minor`.... I will remove the merge commit from it

fabiaserra:
Ok this is messed up... I will close and re-open with just those changes sorry

fabiaserra:
New PR here https://github.com/ynput/OpenPype/pull/5093

PR #5089, opened 2023-06-02
General: Remove default windowFlags from publisher

fabiaserra:
## Changelog Description
The default windowFlags is making the publisher window (in Linux at least) only show the close button and it's frustrating as many times you just want to minimize the window and get back to the validation after. Removing that line I get what I'd expect.

**Before:**
image

**After:**
image

## Testing notes:
1. Open publisher window

### DIFF
--- a/openpype/tools/publisher/window.py
+++ b/openpype/tools/publisher/window.py
@@ -66,8 +66,7 @@ class PublisherWindow(QtWidgets.QDialog):
             on_top_flag = QtCore.Qt.Dialog
 
         self.setWindowFlags(
-            self.windowFlags()
-            | QtCore.Qt.WindowTitleHint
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
Hi, thanks. Have you tried it in Nuke or Maya? I think those are the only which set parent widget to the window and in that case is used `Qt.QDialog` flag which could cause the same issue.

iLLiCiTiT (approved):
LGTM

iLLiCiTiT (approved):
LGTM

fabiaserra:
> Hi, thanks. Have you tried it in Nuke or Maya? I think those are the only which set parent widget to the window and in that case is used `Qt.QDialog` flag which could cause the same issue.

Yeah I had tested it in Nuke, thanks for the approve and merge!

PR #5090, opened 2023-06-02
General: Lower unidecode version

iLLiCiTiT:
## Changelog Description
Use older version of Unidecode module to support Python 2.

## Additional info
The version definition `^1.2` does not specify that the version should be `1.2.*` so latest (`1.3.6`) version was used.

## Testing notes:
1. OpenPype should run in Python 2 DCCs


64qam (approved):
Looks good in Nuke

antirotor (approved):

PR #5092, opened 2023-06-02
Maya: Improve error feedback when no renderable cameras exist for ASS family.

tokejepsen:
## Changelog Description
When collecting cameras for `ass` family, this improves the error message when no cameras are renderable.

## Testing notes:
1. Setup `ass` instance in Maya.
2. Ensure no cameras are renderable.
3. Publish and validate collection error message.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
+++ b/openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
@@ -35,18 +35,16 @@ class CollectArnoldSceneSource(pyblish.api.InstancePlugin):
         # camera.
         cameras = cmds.ls(type="camera", long=True)
         renderable = [c for c in cameras if cmds.getAttr("%s.renderable" % c)]
-        if not renderable:
-            raise ValueError(
-                "No renderable cameras found, which is required for "
-                "publishing ASS."
-            )
-        camera = renderable[0]
-        for node in instance.data["contentMembers"]:
-            camera_shapes = cmds.listRelatives(
-                node, shapes=True, type="camera"
-            )
-            if camera_shapes:
-                camera = node
-        instance.data["camera"] = camera
+        if renderable:
+            camera = renderable[0]
+            for node in instance.data["contentMembers"]:
+                camera_shapes = cmds.listRelatives(
+                    node, shapes=True, type="camera"
+                )
+                if camera_shapes:
+                    camera = node
+            instance.data["camera"] = camera
+        else:
+            self.log.debug("No renderable cameraes found.")
 
         self.log.debug("data: {}".format(instance.data))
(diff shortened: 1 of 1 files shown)


openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
```diff
@@ -35,6 +35,11 @@ def process(self, instance):
         # camera.
         cameras = cmds.ls(type="camera", long=True)
         renderable = [c for c in cameras if cmds.getAttr("%s.renderable" % c)]
+        if not renderable:
+            raise ValueError(
+                "No renderable cameraes found, which is required for "
+                "publishing ASS."
+            )
```
BigRoy, line 42:
```suggestion
        if not renderable:
            raise ValueError(
                "No renderable cameras found, which is required for "
                "publishing ASS."
            )
```

Also, why is a camera required? Can't you export an ass file of just geometry?

BigRoy: Additionally, this error should actually not be occurring in collect but during validation.

mkolar: Totally. If we're touching it, let's convert it into a validation with `PublishValidationError` so it's compatible with the new publisher right away.

tokejepsen: Good point that is not required.

openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
```diff
@@ -35,13 +35,16 @@ def process(self, instance):
...
+            camera = renderable[0]
+            for node in instance.data["contentMembers"]:
+                camera_shapes = cmds.listRelatives(
+                    node, shapes=True, type="camera"
+                )
+                if camera_shapes:
+                    camera = node
+            instance.data["camera"] = camera
+        else:
+            self.log.debug("No renderable cameraes found.")
```
moonyuet, line 48: some little typos shown here
```suggestion
            self.log.debug("No renderable cameras found.")
```

moonyuet:
The collector run smoothly with the renderable camera but once it has been removed, the collector "strangely" errors out with IndexError.
image

The line which causes this issue is the list with renderable cameras. 
`renderable = [c for c in cameras if cmds.getAttr("%s.renderable" % c)]`

tokejepsen:
> maybe you are right, I find out that the code is built with the AppData folder. And it seems that the script doesn't sync with the script I built along with the system. Let me rebuild it to see how it looks.

Do you always rebuild for testing?

moonyuet:
> > maybe you are right, I find out that the code is built with the AppData folder. And it seems that the script doesn't sync with the script I built along with the system. Let me rebuild it to see how it looks.
> 
> Do you always rebuild for testing?

If you said run build.ps1, yes. But if you are talking about env and fetch library.ps, I dont do it always and I dont do it with your fork as often as I did for the main repo.
And I find out that it is actually env and fetch library.ps1. Maybe I need to add your fork remotely into my cloned of main repos and it would be easier to work together. 
It works after rebuilt. 
image

mkolar (approved):

tokejepsen:
Not sure whether you know, but you can run directly from source with `run_tray.ps1`. Then you dont need to rebuild and question the build.

PR #5094, opened 2023-06-02
Burnins: Refactored burnins script

iLLiCiTiT:
## Changelog Description
Refactored list value for burnins and fixed command length limit by using temp file for filters string.

## Additional info
Burnin script does not print irrelevant warning about not supported subdictionary formattings (it is supported). List values can have multiple list keys in single template and can expect other keys to fill (except timecode). FFmpeg filters are not passed to cli but through temp file and using `-filter_script` command to the file which should fix possible limit of command length.

## Testing notes:
1. Burnins should work with templates using dictionary subkeys (e.g. `{project[name]}`)
2. Burnins should work as expected with list values (currently supported only in focal length in maya)
    - make sure one of the burnin templates has `{focalLength}`
    - set keys for camera focal lenght in scene
    - publish
    - validate the burnin has right focal lenght values


LiborBatek (changes requested):
When adding those burnins for ```focalLength``` it gets too many decimals/digits. and when trying to limit those via OP Burnins settings ```focalLength:2f``` it gets ```N/A mm``` in the burnins.

image

image

image

iLLiCiTiT:
> When adding those burnins for focalLength it gets too many decimals/digits. and when trying to limit those via OP Burnins settings focalLength:2f it gets N/A mm in the burnins.

Should be resolved

openpype/scripts/otio_burnin.py
```diff
@@ -589,59 +661,50 @@ def burnins_from_data(
...
+
+                if isinstance(key_value, list):
+                    listed_keys[orig_key] = {
+                        "values": key_value,
+                        "keys": keys}
+                else:
+                    fill_values[orig_key] = orig_key.format(**data)
+            except (KeyError, TypeError):
+                missing_keys.append(orig_key)
+                continue
```
BigRoy, line 692: Readability wise I think this is better put in a function separate from the class with a clear docstring:
```python

def format_value_array(value, data):
    """Format string with key-values in data.
    
    Data may contain lists for values so that each entry gets
    formatted on its own with resulting values.
    
    Arguments:
        value (str): String containing formatting values
        data (dict): Key-values used for formatting.
    
    Returns:
        tuple: listed_keys, fill_values, missing_keys
    
    """
    missing_keys = []
    listed_keys = {}
    fill_values = {}
    for item in Formatter().parse(value):
        _, field_name, format_spec, conversion = item
        if not field_name:
            continue
        keys = [key.rstrip("]") for key in field_name.split("[")]
        conversion = "!{}".format(conversion) if conversion else ""
        format_spec = ":{}".format(format_spec) if format_spec else ""
        # Calculate original full key for replacement
        orig_key = "{{{}{}{}}}".format(
            field_name, conversion, format_spec)

        key_value = data
        try:
            for key in keys:
                key_value = key_value[key]

            if isinstance(key_value, list):
                listed_keys[orig_key] = {
                    "values": key_value,
                    "keys": keys}
            else:
                fill_values[orig_key] = orig_key.format(**data)
        except (KeyError, TypeError):
            missing_keys.append(orig_key)
            continue
            
    return listed_keys, fill_values, missing_keys
    
```

openpype/scripts/otio_burnin.py
```diff
@@ -263,6 +219,113 @@ def add_timecode(
...
+                    value = key.format(**sub_value)
+                except (TypeError, KeyError, ValueError):
+                    value = MISSING_KEY_VALUE
+                new_values[key] = value
+
+                value_len = len(value)
+                if value_len > key_max_len:
+                    key_max_value = value
+                    key_max_len = value_len
+            longest_value_by_key[key] = key_max_value
```
BigRoy, line 287: I feel this is at least lacking a comment here and there. It appears quite complex but there's no description as to what's going on here.

64qam (approved):
Looks good in Nuke

LiborBatek (changes requested):
The ```focalLength``` values differ from maya scene.

For example on frame #1046 instead being ```120.11mm``` it is ```122.64mm```

The same is valid for startframe and endframe. Values being off.

image

iLLiCiTiT:
> The focalLength values differ from maya scene.

Not an issue of this PR.

BigRoy:
That does sound like an issue of this PR since it's likely due to a time/frame mismatch of the data. It works fine in `develop` as far as I know.

LiborBatek (changes requested):
When tested in ```develop``` focal length is correct.

So its an issue within this PR and need to be addressed.

LiborBatek:
start frame being correct 

```focalLength``` being ```50```

but 

endframe does not match (should be 125mm)
image

and the same endframe in the review mov
image

**Comment by iLLiCiTiT** [comment]
So it was found out that there are some differences in fps, which cause that seconds are calculated wrong (24 vs 25 fps). Tried to fix it.

File: openpype/scripts/otio_burnin.py [hunk 3]
```diff
@@ -263,6 +220,139 @@ def add_timecode(
...
+                    new_text = new_text.replace(_key, str(_value))
+
+                new_text = (
+                    str(new_text)
+                    .replace("\\", "\\\\")
+                    .replace(",", "\\,")
+                    .replace(":", "\\:")
+                )
+                lines.append(
+                    f"{seconds} drawtext@{align} reinit text='{new_text}';")
```
BigRoy, line 338: We don't seem to write to the file as we generate the `lines` as such this can be outside of the `with` statement.

It seems we're doing "tempfile + write lines + add to cleanup files" quite a bit.
How about creating a method to replace that:

```python
path = self.as_temp_file("\n".join(lines))
```

```python
def as_temp_file(self, text):
    with tempfile.NamedTemporaryFile(mode="w", delete=False) as temp:
        path = temp.name
        temp.write(text)
    self.cleanup_path.append(path)
    return path
```

openpype/scripts/otio_burnin.py
```diff
@@ -512,17 +656,38 @@ def burnins_from_data(
...
+    if options.get("fps") is None:
+        # FPS May not be needed at all in some cases (e.g. sequence)
+        # - ignore the bug here, it will crash elsewhere when fps is needed
+        try:
+            burnin.options_init["fps"] = convert_ffprobe_fps_to_float(
+                r_frame_rate)
+        except ValueError:
+            print(
+                "Warning: Couldn't convert fps value: {}".format(r_frame_rate)
+            )
```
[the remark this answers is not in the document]
iLLiCiTiT: Even I understand, the question for me is, why ffprobe returns different fps than what is in data?

LiborBatek (approved):
Now all works as expected and 
```startframe``` and ```endframe```  values of ```focalLength``` matches with maya

```50mm``` and ```125mm``` both set correctly in burnins.
image

image

PR #5100, opened 2023-06-06
Loader: Hide inactive versions in UI

iLLiCiTiT:
## Changelog Description
Hide versions with `active` set to `False` in Loader UI.

## Additional info
This required changes in next-minor release too.

## Testing notes:
There is not a function to change or set active of version in our codebase.
1. Open Loader UI
2. Choose a subset with version
3. Find the version in Mongo
4. Set it's `data.active` to `False`
5. Refresh loader UI
6. The version should disappear from UI

### DIFF
--- a/openpype/client/entities.py
+++ b/openpype/client/entities.py
@@ -855,12 +855,13 @@ def get_output_link_versions(project_name, version_id, fields=None):
     return conn.find(query_filter, _prepare_fields(fields))
 
 
-def get_last_versions(project_name, subset_ids, fields=None):
+def get_last_versions(project_name, subset_ids, active=None, fields=None):
     """Latest versions for entered subset_ids.
 
     Args:
         project_name (str): Name of project where to look for queried entities.
         subset_ids (Iterable[Union[str, ObjectId]]): List of subset ids.
+        active (Optional[bool]): If True only active versions are returned.
         fields (Optional[Iterable[str]]): Fields that should be returned. All
             fields are returned if 'None' is passed.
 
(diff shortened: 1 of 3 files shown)


jakubjezek001:
Task linked: [OP-4916 unpublish state filtering on version in UI](https://app.clickup.com/t/6658547/OP-4916)

BigRoy:
Please also test the edge case where for an existing subset you set **all** versions to inactive - which I suppose the loader should then omit completely?

iLLiCiTiT:
> Please also test the edge case where for an existing subset you set all versions to inactive - which I suppose the loader should then omit completely?

Good point. Tested, and the subset is not shown in the UI at all in that case (without crashes).

kalisp (approved):
I am guessing that the check is only for explicit `active == False`, as I tried to change `data.active` to True, but that didn't bring version back.

kalisp:
Actually what is the reason for this? Wouldn't this have impact on "existence" of the version? 
Shouldn't this affect also `get_version_by_id`, `get_version_by_name`, `version_is_latest` and probably `get_versions` as currently no one is handling if that version is actually active.

BigRoy:
> as I tried to change `data.active` to `True`, but that didn't bring version back.

That should definitely bring it back. If it doesn't I'd consider it a bug.

iLLiCiTiT:
> I am guessing that the check is only for explicit active == False, as I tried to change data.active to True, but that didn't bring version back.

Resolved, a string was used instead of boolean.

> Actually what is the reason for this? Wouldn't this have impact on "existence" of the version?
Shouldn't this affect also get_version_by_id, get_version_by_name, get_versions, get_hero_versions, version_is_latest and probably get_versions as currently no one is handling if that version is actually active.

This is a client request to hide versions in UI, but don't reuse the same version on next published version. That's reason why the default behavior of functions didn't change -> integrateion will consider inactive versions as existing but UI won't show them.

PR #5106, opened 2023-06-06
General: Sort launcher actions alphabetically

fabiaserra:
## Changelog Description
The launcher actions weren't being sorted by its label but its name (which on the case of the apps it's the version number) and thus the order wasn't consistent and we kept getting a different order on every launch. From my debugging session, this was the result of what the `actions` variable held after the `filter_compatible_actions` function before these changes:
```
(Pdb) for p in actions: print(p.order, p.name)
0 14-02
0 14-02
0 14-02
0 14-02
0 14-02
0 19-5-493
0 2023
0 3-41
0 6-01
```
This caused already a couple bugs from our artists thinking they had launched Nuke X and instead launched Nuke and telling us their Nuke was missing nodes

**Before:**
image

**After:**
image

## Testing notes:
1. Start the launcher and check the order of the apps

### DIFF
--- a/openpype/tools/launcher/models.py
+++ b/openpype/tools/launcher/models.py
@@ -273,7 +273,7 @@ class ActionModel(QtGui.QStandardItemModel):
         # Sort by order and name
         return sorted(
             compatible,
-            key=lambda action: (action.order, action.name)
+            key=lambda action: (action.order, lib.get_action_label(action))
         )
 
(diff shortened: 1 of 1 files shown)


BigRoy:
Great to see you found the root of the issue! The old sorting definitely didn't make much sense if it was solely due to the version numbers of the apps.

iLLiCiTiT (approved):
LGTM. Thanks

PR #5107, opened 2023-06-07
Nuke: open_file function can open autosave script

moonyuet:
## Changelog Description
Fix the bug of the workfile dialog being unable to open autosave nuke script

## Additional info
n/a

## Testing notes:
1. Launch Nuke via Openpype
2. Go to Openpype menu and click work file
3. Open any nuke script with autosave
4. It will pop up nuke prompt to ask if you want to open autosave script.
5. It will load autosave script for you if you click yes; while it wont if you click no.


jakubjezek001:
Task linked: [OP-4096 Nuke: open_file function wouldn't open .autosave file](https://app.clickup.com/t/6658547/OP-4096)

openpype/hosts/nuke/api/workio.py
```diff
@@ -25,7 +25,13 @@ def open_file(filepath):
     # To remain in the same window, we have to clear the script and read
     # in the contents of the workfile.
     nuke.scriptClear()
-    nuke.scriptReadFile(filepath)
+    autosave = f"{filepath}.autosave"
+    autosave_prmpt = "Autosave detected.\nWould you like to load the autosave file?"       # noqa
+    if os.path.isfile(autosave) and nuke.ask(autosave_prmpt):
```
iLLiCiTiT, line 30: I'm not sure if `open_file` should trigger UI message? Thinking about automations.

moonyuet: it's weird but this UI implementation is to allow users to choose whether they want to use autosave script if there is one avaliable in the work directory. If it is removed and it will always load autosave script if there is one. 
maybe we can move this logic to `open_workfile` in `pipeline.py`.

openpype/hosts/nuke/api/workio.py
```diff
@@ -25,7 +25,13 @@ def open_file(filepath):
...
+        filepath = autosave
+        nuke.scriptReadFile(filepath)
+    else:
+        nuke.scriptReadFile(filepath)
```
iLLiCiTiT, line 34:
```suggestion

    nuke.scriptReadFile(filepath)
```

openpype/hosts/nuke/api/workio.py
BigRoy: By the way, another option _could_ be to allow `.autosave` extensions to be visible in the workfiles tool? (Less artist friendly I suppose?)

iLLiCiTiT: Maybe add kwargs to the function.
```python
def open_file(filepath, allow_autosave=True):
    ...
    if allow_autosave:
        autosave = f"{filepath}.autosave"
        autosave_prmpt = (
            "Autosave detected."
            "\nWould you like to load the autosave file?"
        )
        if os.path.isfile(autosave) and nuke.ask(autosave_prmpt):
            filepath = autosave
        nuke.scriptReadFile(filepath)
    ...
```
So it is at least possible in automations skip that option.

iLLiCiTiT: > We could just have **kwargs in all the open_file interfaces but it'd just start becoming confusing across the different hosts since we don't know which we should pass to enforce no prompts. Also, if we were to add a toggle for allow autosave I'd actually assume that False or True would both avoid the prompt since they are explicit values either way.

The api defines that there must be 1 positional required argument that is filepath, any additional kwargs can be host specific, if we consider that the kwarg could be used only from the host itself. If not we could change it to `headless`. But that's discussion for different place.

Ad this PR, I couldn't find way how to find out, if nuke is running in batch mode (I didn't do deep search). If you know how, please do a suggestion to make it quicker.

BigRoy: Couldn't find it quickly either - it seems in batch mode at least Nuke doesn't create a QApplication instance so you could check for the QApplication instance existing or not. But yeah, that's definitely not as clean as I'd expect.

openpype/hosts/nuke/api/workio.py
```diff
@@ -19,12 +19,18 @@ def save_file(filepath):
...
 
 
-def open_file(filepath):
+def open_file(filepath, allow_autosave=True):
     filepath = filepath.replace("\\", "/")
 
     # To remain in the same window, we have to clear the script and read
     # in the contents of the workfile.
     nuke.scriptClear()
+    if allow_autosave:
```
iLLiCiTiT, line 28:
```suggestion
def open_file(filepath):
    filepath = filepath.replace("\\", "/")

    # To remain in the same window, we have to clear the script and read
    # in the contents of the workfile.
    nuke.scriptClear()
    headless = QtWidgets.QApplication.instance() is None
    if not headless:
```

jrsndl:
May I sugest using Nuke preferences to get the proper autosave path?


`autosave = nuke.toNode("preferences")["AutoSaveName"].evaluate()`
will get the path for you from Nuke prefs:
image

Instead of just assuming that autosave is located where script is.

openpype/hosts/nuke/api/workio.py
```diff
@@ -25,6 +26,13 @@ def open_file(filepath):
     # To remain in the same window, we have to clear the script and read
     # in the contents of the workfile.
     nuke.scriptClear()
+    headless = QtWidgets.QApplication.instance() is None
```
jakubjezek001, line 29: This is quite interesting approach. Definitelly useful! Could we actually abstract this to api/lib.py as something like `is_headless` function?

moonyuet: I've added it in the latest commit.

openpype/hosts/nuke/api/lib.py
```diff
@@ -3155,3 +3155,12 @@ def get_viewer_config_from_string(input_string):
...
     return (display, viewer)
+
+
+def is_headless():
+    """
+    Returns:
+        bool: headless
+    """
+    headless = QtWidgets.QApplication.instance() is None
+    return headless
```
jakubjezek001, line 3166:
```suggestion
    return QtWidgets.QApplication.instance() is None
```

openpype/hosts/nuke/api/workio.py
```diff
@@ -25,6 +26,13 @@ def open_file(filepath):
     # To remain in the same window, we have to clear the script and read
     # in the contents of the workfile.
     nuke.scriptClear()
+    headless = is_headless()
+    if not headless:
```
jakubjezek001, line 30:
```suggestion
    if not is_headless():
```

openpype/hosts/nuke/api/lib.py
jakubjezek001: cosmetic change but less code to review ;)

jakubjezek001 (approved):
After the changes are applied this looks good to me and is will be ready for merge ;) Thanks @moonyuet !

openpype/hosts/nuke/api/workio.py
moonyuet: Updated in the latest commit! Thanks!

openpype/hosts/nuke/api/workio.py
```diff
@@ -1,6 +1,7 @@
 """Host API required Work Files tool"""
 import os
 import nuke
+from openpype.hosts.nuke.api.lib import is_headless
```
BigRoy, line 4: You've now introduced recurive imports. `workio` imports `lib` and `lib` imports `workio`. I assume it's currently broken.

moonyuet: Tested it just now. it is broken. maybe it is better to keep it in workio.py as the original. i.e. `headless = QtWidgets.QApplication.instance() is None`

openpype/hosts/nuke/api/lib.py
```diff
@@ -3155,3 +3155,11 @@ def get_viewer_config_from_string(input_string):
...
 
     return (display, viewer)
+
+
+def is_headless():
+    """
+    Returns:
+        bool: headless
+    """
+    return QtWidgets.QApplication.instance() is None
```
BigRoy, line 3165:
```suggestion
```
I suspect this can now be removed again?

jakubjezek001: yeah, it had been solved in [`c414c8c` (#5107)](https://github.com/ynput/OpenPype/pull/5107/commits/c414c8c6e933e82e6e7703b8e509b711e36f9d2e)

64qam (approved):

PR #5110, opened 2023-06-07
Enhancement:animation family loaded as standing (abc) uses "use file sequence"

FadyFS:
## Changelog Description
The changes are the following. We started by updating the the is_sequence(files) function allowing it to return True for a list of files which has only one file, since our animation in this provides just one alembic file. For the correct FPS number, we got the fps from the published ass/abc from the version data.

## Testing notes:
Load the animation with abc format with Maya

### DIFF
--- a/openpype/hosts/maya/plugins/load/load_arnold_standin.py
+++ b/openpype/hosts/maya/plugins/load/load_arnold_standin.py
@@ -16,10 +16,13 @@ from openpype.hosts.maya.api.pipeline import containerise
 
 def is_sequence(files):
     sequence = False
-    collections, remainder = clique.assemble(files)
+    print("here calling")
+    if len(files) == 1:
+        collections, remainder = clique.assemble(files, minimum_items=1)
+    else:
+        collections, remainder = clique.assemble(files)
     if collections:
         sequence = True
-
     return sequence
 
 
@@ -84,6 +87,7 @@ class ArnoldStandinLoader(load.LoaderPlugin):
             sequence = is_sequence(os.listdir(os.path.dirname(self.fname)))
             cmds.setAttr(standin_shape + ".useFrameExtension", sequence)
 
+
         nodes = [root, standin, standin_shape]
         if operator is not None:
             nodes.append(operator)
@@ -146,6 +150,10 @@ class ArnoldStandinLoader(load.LoaderPlugin):
             os.path.basename(path),
             type="string"
         )
+        print("#"*100)
+
+
+        # Object name value
 
         cmds.connectAttr(
             string_replace_operator + ".out",
--- a/openpype/hosts/maya/plugins/load/load_arnold_standin.py
+++ b/openpype/hosts/maya/plugins/load/load_arnold_standin.py
@@ -90,7 +90,7 @@ class ArnoldStandinLoader(load.LoaderPlugin):
             sequence = is_sequence(os.listdir(os.path.dirname(self.fname)))
             cmds.setAttr(standin_shape + ".useFrameExtension", sequence)
 
-            fps = float(version["data"].get("fps")) or get_current_session_fps()
+            fps = float(version["data"].get("fps"))or get_current_session_fps()
             cmds.setAttr(standin_shape + ".abcFPS", fps)
 
         nodes = [root, standin, standin_shape]
--- a/openpype/hosts/maya/plugins/load/load_arnold_standin.py
+++ b/openpype/hosts/maya/plugins/load/load_arnold_standin.py
@@ -27,6 +27,7 @@ def is_sequence(files):
         sequence = True
     return sequence
 
+
 def get_current_fps():
     session_fps = float(legacy_io.Session.get('AVALON_FPS', 25))
     return convert_to_maya_fps(session_fps)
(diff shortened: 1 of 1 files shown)


openpype/hosts/maya/plugins/load/load_arnold_standin.py
```diff
@@ -83,6 +85,7 @@ def load(self, context, name, namespace, options):
             cmds.setAttr(standin_shape + ".dso", path, type="string")
             sequence = is_sequence(os.listdir(os.path.dirname(self.fname)))
             cmds.setAttr(standin_shape + ".useFrameExtension", sequence)
+            cmds.setAttr(standin_shape + ".abcFPS", 30)
```
BigRoy, line 88: Wait a second - this doesn't sound right?

FadyFS: I am not sure about this one, but it works for this specific animation. Let me know what you think

FadyFS: Thank you for your reply, I will look it up more thoroughly !

openpype/hosts/maya/plugins/load/load_arnold_standin.py
```diff
@@ -84,6 +96,10 @@ def load(self, context, name, namespace, options):
             sequence = is_sequence(os.listdir(os.path.dirname(self.fname)))
             cmds.setAttr(standin_shape + ".useFrameExtension", sequence)
 
+            cmds.setAttr(standin_shape + ".dso", path, type="string")
+            matching_fps = get_fps(os.listdir(os.path.dirname(self.fname)))
+            cmds.setAttr(standin_shape + ".abcFPS", matching_fps)
```
moonyuet, line 101:
```suggestion
            fps = float(version["data"]["fps"])
            cmds.setAttr(standin_shape + ".abcFPS", fps)
```
you can get the fps from the published ass/abc from the version data.

openpype/hosts/maya/plugins/load/load_arnold_standin.py
```diff
@@ -6,22 +6,34 @@
...
+    if len(files) == 1:
+        collections, remainder = clique.assemble(files, minimum_items=1)
+    else:
+        collections, remainder = clique.assemble(files)
     if collections:
         sequence = True
-
     return sequence
 
+def get_fps(standin_shape):
```
moonyuet, line 31: You dont need this function if you get the fps data from the published object through the version data.

openpype/hosts/maya/plugins/load/load_arnold_standin.py
BigRoy: Note that technically a version **could** be lacking fps data, right? E.g. a publish from standalone publisher? (I suspect that doesn't know anything about its FPS?)

moonyuet: Something like this:(pesudo code only)
```
            has_published_fps = version["data"]["fps"]
            matching_fps = None
            if has_published_fps:
                matching_fps = float(version["data"]["fps"])
            else:
                matching_fps = get_fps(os.listdir(os.path.dirname(self.fname)))
            cmds.setAttr(standin_shape + ".abcFPS", matching_fps)
```

moonyuet (changes requested):
see comments above

moonyuet (approved):
the code looks good....it is ready for the test

ClementHector:
@FadyFS you can switch your PR to ready to review and set the finished comments to resolved 🥇

openpype/hosts/maya/plugins/load/load_arnold_standin.py
```diff
@@ -84,6 +93,11 @@ def load(self, context, name, namespace, options):
             sequence = is_sequence(os.listdir(os.path.dirname(self.fname)))
             cmds.setAttr(standin_shape + ".useFrameExtension", sequence)
 
+            fps = float(version["data"].get("fps"))
+            if fps is None:
+                fps = get_current_fps()
+            cmds.setAttr(standin_shape + ".abcFPS", fps)
```
ClementHector, line 99: it is equivalent to:
```suggestion
            fps = float(version["data"].get("fps"), get_current_fps())
            cmds.setAttr(standin_shape + ".abcFPS", fps)
```

ClementHector (changes requested):

moonyuet (commented):
image
looks good, i also did some simple test with the loader in maya(without fps data) (it might need some more complicated tests to see if it breaks anything)
it loaded with the same data regarding to the avalon.session.

openpype/hosts/maya/plugins/load/load_arnold_standin.py
```diff
@@ -6,23 +6,29 @@
...
-    collections, remainder = clique.assemble(files)
+    collections, remainder = clique.assemble(files, minimum_items=1)
     if collections:
         sequence = True
-
     return sequence
 
 
+def get_current_session_fps():
+    session_fps = float(legacy_io.Session.get('AVALON_FPS', 25))
```
BigRoy, line 29: @iLLiCiTiT is this `legacy_io` usage ok, or is there already an alternative to this?

BigRoy (approved):
Didn't do a test run, but the code changes seem fine to me. Nice work.

LiborBatek (approved):
Did test and loaded animation family as Ass(abc) and all good. 
It has correctly set "Use File Sequence" and updates fine when playback.

image

PR #5111, opened 2023-06-07
Global: replace dependencies methods

kalisp:
## Changelog Description
`dependencies` route got replaced with `desktop/dependency_packages` on the server.


kalisp:
Not needed now.

PR #5113, opened 2023-06-07
Nuke: connect custom write node script to the OP setting

moonyuet:
## Changelog Description
Allows user to customize the values of knobs attribute in the OP setting and use it in custom write node 

## Additional info
n/a

## Testing notes:
1. Launch Openpype
2. Go to Admin > Studio Settings
3. Project > Nuke > Imageio > Plugin's node overrides
4. 
image

5. Add a Plugin's node overrides and set values for Nuke Node Class, Subsets and Knobs overrides
6. Launch Nuke via Openpype
7. Go to Openpype Tools 
8. Create some write nodes
9. Click Set non publish outputs for Write Node
10. Select your preferred Subsets for setting up the knob values of the write nodes
11. If you dont set up any Plugin's node overrides in nuke via OP settings, it will use the default setting hardcoded in the `custom_write_node.py`


jakubjezek001:
Task linked: [OP-6149 nuke - custom write node script connected to settings](https://app.clickup.com/t/6658547/OP-6149)

openpype/hosts/nuke/startup/custom_write_node.py
```diff
@@ -53,24 +58,79 @@
...
+        self.addKnob(self.typeKnob)
+
+    def process(self):
+        """ Process the panel values. """
+        write_selected_nodes = [
+            s for s in nuke.selectedNodes() if s.Class() == "Write"]
+
+        node_knobs = self.typeKnob.value()
+        ext = None
+        knobs = None
```
Minkiu, line 81: I'd probably set `knobs` here with the defauult, and override `if node_knobs`:
```py
knobs = knobs_setting["knobs"]

if node_knobs:
    node_knobs_settings = self.get_node_knobs_setting(node_knobs)
    if not node_knobs_settings:
        nuke.message("No knobs value found in subset group..\nDefault setting will be used..") # noqa
    else:
        knobs = node_knobs_settings
```

openpype/hosts/nuke/startup/custom_write_node.py
```diff
@@ -53,24 +58,79 @@
...
+            if not knobs:
+                nuke.message("No knobs value found in subset group..\nDefault setting will be used..")  # noqa
+                knobs = knobs_setting["knobs"]
+        else:
+            knobs = knobs_setting["knobs"]
+
+        for knob in knobs:
+            if knob["name"] == "file_type":
+                ext = knob["value"]
+        for w in write_selected_nodes:
```
Minkiu, line 93: I'd do `write_node` instead of just `w`.

openpype/hosts/nuke/startup/custom_write_node.py
```diff
@@ -53,24 +58,79 @@
...
+                "subset": w["name"].value(),
+                "frame": "#" * frame_padding,
+                "ext": ext
+            }
+            file_path = temp_rendering_path_template.format(**data)
+            file_path = file_path.replace("\\", "/")
+            w["file"].setValue(file_path)
+            set_node_knobs_from_settings(w, knobs)
+
+    def get_node_knobs_setting(self, value):
```
Minkiu, line 106: I'd probably do a helper method that does :
```py
settings = [
            node
            for node in get_nuke_imageio_settings()["nodes"]["overrideNodes"]
        ]
        if not settings:
            return
```

or convert the two `get_node_knobs_setting` and `get_node_knobs_override` into one that returns a tuple of `(settings, overrides)`

openpype/hosts/nuke/startup/custom_write_node.py
```diff
@@ -1,6 +1,11 @@
...
 import nuke
-from openpype.hosts.nuke.api.lib import set_node_knobs_from_settings
+import nukescripts
+from openpype.hosts.nuke.api.lib import (
+    set_node_knobs_from_settings,
+    get_nuke_imageio_settings
+)
 
 
 frame_padding = 5
```
jakubjezek001, line 11: How about we take the frame padding value from Anatomy templates rather than to hardcode it?

openpype/hosts/nuke/startup/custom_write_node.py
```diff
@@ -53,24 +58,74 @@
 }
 
 
+class WriteNodeKnobSettingPanel(nukescripts.PythonPanel):
+    """ Write Node's Knobs Settings Panel """
+    def __init__(self):
+        nukescripts.PythonPanel.__init__(self, "Set Knobs Value(Write Node)")
+
+        knobs_value, _ = self.get_node_knobs_setting()
```
jakubjezek001, line 66: Shouldn't we name this `knobs_value` to something closer to the true meaning like `preset_name`?

openpype/hosts/nuke/startup/custom_write_node.py
```diff
@@ -53,24 +58,74 @@
...
+        # create knobs
+
+        self.typeKnob = nuke.Enumeration_Knob(
```
jakubjezek001, line 69: Here the name is also not very real. `typeKnob` should in fact be `preset_selector`. Also the consistency of names is not good. Lets keep all variable names in snake_case style. Please use some linter which will catch this in your IDE.

openpype/hosts/nuke/startup/custom_write_node.py
```diff
@@ -53,24 +58,74 @@
...
+            'override_subsets', 'override subsets', knobs_value)
+        # add knobs to panel
+        self.addKnob(self.typeKnob)
+
+    def process(self):
+        """ Process the panel values. """
+        write_selected_nodes = [
+            s for s in nuke.selectedNodes() if s.Class() == "Write"]
+
+        node_knobs = self.typeKnob.value()
```
jakubjezek001, line 79: Again, this is `selected_preset_name` right?

openpype/hosts/nuke/startup/custom_write_node.py
```diff
@@ -53,24 +58,74 @@
...
+        ext = None
+        knobs = knobs_setting["knobs"]
+
+        if node_knobs:
+            _, node_knobs_settings = self.get_node_knobs_setting(node_knobs)
```
jakubjezek001, line 84: Please make sure this will only return `Write` class presets. I am having two presets with the same subset name one for Write and another for Reformat and it is picking up the Reformat and fails to do the job as expected.

openpype/hosts/nuke/startup/custom_write_node.py
```diff
@@ -53,24 +58,74 @@
...
+            if not node_knobs_settings:
+                nuke.message("No knobs value found in subset group..\nDefault setting will be used..")  # noqa
+            else:
+                knobs = node_knobs_settings
+
+        for knob in knobs:
+            if knob["name"] == "file_type":
```
jakubjezek001, line 91: My guess is that we should require `file_type`, because without it it will fail. We should let know to user if this is failing due to the missing `file_type` knob in settings. Use `nuke.message` please.

jakubjezek001 (changes requested):
Nice first kick of but lets brush up all those fixes and merge it! ;)

openpype/hosts/nuke/startup/custom_write_node.py
jakubjezek001: is this really resolved? This seems like it is not addressed. Please change the variable name.

moonyuet: maybe i just rename it as `Set Preset(Write Node)`

jakubjezek001 (approved):
LGTM

PR #5116, opened 2023-06-08
Tray documentation

SherineKB:
## Changelog Description
I documented the OpenPype tray by writing about each category of it.


BigRoy:
I guess it's still a draft so my comments come way to early but this stood out to me:

Would it be visually nicer if the "Click on the Tray icon" actually shows an example of the icon in a tray instead of just the popup menu? :)

This didn't read very clearly due to it missing the tray icon itself to me.

image

---

The Ftrack section might be best marked clearly that it's only available when the Ftrack module is enabled (potentially linking to a page on how to set up Ftrack?)

---

I believe Logging is also only available if the module is enabled for it (which it is by default however)

---

I think the Settings link is invalid in this [chapter](https://github.com/ynput/OpenPype/blob/e9f82800705e725ce78a56e7bb006ff8df8b22e8/website/docs/artist_tools_tray.md#settings)

---

Services still appear to be undocumented?

---

Also, I believe the Tray can be in a state where it says something like Exit & Restart or something. I'd definitely love to see that visually documented and described as to when that happens. I believe it's when you have '

SherineKB:
Thanks for your suggestions, I updated the document. 

> Also, I believe the Tray can be in a state where it says something like Exit & Restart or something. I'd definitely love to see that visually documented and described as to when that happens. I believe it's when you have '

I only encountered the Exit button, can you let me know how to have the Exist and Restart please ?

website/docs/artist_tools_tray.md
```diff
@@ -0,0 +1,71 @@
...
+# Tray
+
+Click on the **OpenPype** logo to open the tray and find the major informations about your account and especially your studio files. You can customize your tray by activating options in the **Studio Settings** such ass adding **Ftrack** or **Kitsu**.
+
+![OpenPype Tray](assets/tools/tools_openpype_tray.png)
+
+## Ftrack
+*Enable it in the [Studio Settings](admin_settings_system.md).*
+### Credentials
+The **Credentials** contains the studio's **Ftrack URL**, your **Username** and the **API Key** given by your studio.
```
ClementHector, line 16: more information here

website/docs/artist_tools_tray.md
```diff
@@ -0,0 +1,71 @@
...
+ ![OpenPype Tray Ftrack](assets/tools/tools_openpype_ftrack.png)
+
+### Open Ftrack
+Access to **Ftrack**'s website and sign in.
+
+## Launcher
+Select your working context. Once choosen, you can see all the files, assets and its associated **Tasks**. You can also filter your assets for a more efficient research. Click on a task and select one of the softwares to open it.
+
+## Loader
+Loader loads published subsets into your current scene or script.
```
ClementHector, line 30:
```suggestion
The loader loads or performs an action on published subsets from outside DCC.
```

website/docs/artist_tools.md
```diff
@@ -8,6 +8,7 @@ sidebar_label: Tools
 
 OpenPype offers a collection of core tools in tandem with the Integrations:
 
+- [Tray](artis_tools_tray)
```
BigRoy, line 11:
```suggestion
- [Tray](artist_tools_tray)
```

website/docs/artist_tools_tray.md
```diff
@@ -0,0 +1,72 @@
...
+description: OpenPype general menu.
+---
+# Tray
+
+Click on the **OpenPype** logo to open the tray and find the major informations about your account and especially your studio files. You can customize your tray by activating options in the **Studio Settings** such ass adding **Ftrack** or **Kitsu**.
+
+![OpenPype Tray](assets/tools/tools_openpype_tray.png)
+
+## Ftrack
+*Enable it in the [Studio Settings](admin_settings_system.md).*
```
Minkiu, line 14: Can't help but think that maybe this should link to the ftrack docs, and have a more generic section such as "Project management credentials"
If you enabled [Ftrack](artist_ftrack.md) or [Kitsu](artist_ftrack.md) in the [Studio Settings](admin_settings_system.md) a menu item will be present where you can login into these.

ClementHector (changes requested):

website/docs/artist_tools_tray.md
ClementHector: To have a generic section, all the "Project management tools" plugins should have the same functioning. I think that as long as that's not the case, it's smarter to put one section per module.

mkolar:
Thank you @SherineKB for this and the other PRs improving the docs! Would you be willing to migrate them to the new documentation repo https://github.com/ynput/ayon-documentation ?

We'll be shutting down the old OpenPype docs page within a couple of weeks, redirecting it to the new ayon-docs, where we'll just keep a small openpype installation section. Vast majority of the user documentation are the same (well the settings have a bit of different look, but match almost 1-1) 

We'd like to avoid double work as much as possible for contributors and the core team

ClementHector:
@mkolar We can re-open this one because the Ayon menu is less complete, we will do another PR for Ayon.

PR #5117, opened 2023-06-08
3dsmax: Move from deprecated interface

antirotor:
## Changelog Description
`INewPublisher` interface is deprecated, this PR is changing the use to `IPublishHost` instead.

## Additional info
This shouldn't affect functionality and it should allow to remove `INewPublisher` from code as 3dsmax integration was its last use.

## Testing notes:
1. open 3dsmax
2. try to publish something

### DIFF
--- a/openpype/hosts/max/api/pipeline.py
+++ b/openpype/hosts/max/api/pipeline.py
@@ -6,7 +6,7 @@ from operator import attrgetter
 
 import json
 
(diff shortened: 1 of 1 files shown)


moonyuet (approved):
Tested with publisher. it doesn't break anything. 
image

BigRoy:
This was the last remainder still using the old interface. With this PR merged the old deprecated interface can be removed in next patch. (The class mentioned it was going to be removed 3.15.x)

PR #5123, opened 2023-06-08
Ftrack: Task status during publishing

iLLiCiTiT:
## Changelog Description
Added option to change task status during publishing for 3 possible cases: "sending to farm", "local integration" and "on farm integration".

## Additional info
First of all was removed hardcoded ftrack status change for celaction which was then applied in Integrate Ftrack Api. That was replaced by using default profile for celaction in `IntegrateFtrackFarmStatus` and logic of changing task status in Integrate Ftrack Api was removed.

Whole logic of task statuses was separated into 5 plugins which are in one file. First is collector which collects available task statuse for future usage and prepares data where are stored statuses that will be set on certain tasks. This way we make sure that task status change can happen only once during publishing -> First set value for task id will be considered as target status, if is set to `None` the status change won't happen.

Important are 3 plugins which collect status name in certain situations and in certain order. First is already existing plugin which collects status name for instances which will be sent to farm, next are instances that are integrated locally and finally instances that are integrated on farm. The final plugin will set all collected tasks and statuses and will promote the changes. The final plugin can be processed before, or after Integrate Ftrack Api, that is because of possible clashes with status automations running on event server -> this won't avoid them 100%, but gives a small option to reduce possible issues.

### Notes
I'm happy to hear ideas and suggestions about changes. I couldn't figure out how to keep it in single plugin because of filtering, or because of complexity of the settings.

### Celaction changes
Submit celaction to deadline should change status in ftrack. Make sure you're using new default values for `project_settings/ftrack/publish/IntegrateFtrackFarmStatus`.

### Task status changes
- To change task status of instances that will be sent to farm add profile to `project_settings/ftrack/publish/IntegrateFtrackFarmStatus`.
- To change task status of instances that are integrated locally add profile to `project_settings/ftrack/publish/ftrack_task_status_local_publish`
- To change task status of instances that are integrated on farm add profile to `project_settings/ftrack/publish/ftrack_task_status_on_farm_publish`

Each profile can define standard profile filtering and result is status name. Make sure the status name you fill can be actually set on the task. The easiest approach to test, is to set the same profile filters for each on the setting but define different status name for example:
- to farm: "Rendering"
- locally: "Ready to review"
- on farm: "Rendered"
Then publish localy -> validate status change (Ready to review) -> publish through farm -> validate status change (Rendering) -> wait for publish farm job to finish -> validate status change (Rebdered).

As a bonus it is possible to validate behavior when combined with ftrack event server status automations.


jakubjezek001:
Task linked: [OP-5481 Statuses in Ftrack](https://app.clickup.com/t/6658547/OP-5481)

openpype/modules/ftrack/plugins/publish/integrate_ftrack_status.py
```diff
@@ -0,0 +1,435 @@
...
+    iterable_size = len(tupled_iterable)
+    if chunk_size is None:
+        chunk_size = 200
+
+    if chunk_size < 1:
+        chunk_size = 1
+
+    for idx in range(0, iterable_size, chunk_size):
+        chunks.append(tupled_iterable[idx:idx + chunk_size])
+    return chunks
```
BigRoy, line 31: This should be in a lib I suspect. There's an entry [here](https://github.com/ynput/OpenPype/blob/5b163fe84d49719c0ebfc0e47142688dd551e647/openpype/modules/ftrack/lib/avalon_sync.py#L137-L161). It being in `avalon_sync` might feel weird but maybe that just means that that function needs to go into `lib` to begin with?

openpype/modules/ftrack/plugins/publish/integrate_ftrack_status.py
```diff
@@ -0,0 +1,435 @@
...
+    active = True
+    targets = ["farm"]
+    settings_key = "ftrack_task_status_on_farm_publish"
+
+
+class IntegrateFtrackTaskStatus(pyblish.api.ContextPlugin):
+    # Use order of Integrate Ftrack Api plugin and offset it before or after
+    base_order = pyblish.api.IntegratorOrder + 0.499
+    # By default is after Integrate Ftrack Api
+    order = base_order + 0.001
```
BigRoy, line 347: IntegratorOrder + 0.5 means it's basically not an integrator anymore right? It's in the next order range (which is not defined in Pyblish itself, but I think that's referred to as "Other" range.

openpype/settings/entities/schemas/projects_schema/schema_project_ftrack.json
```diff
@@ -1114,6 +1114,142 @@
                             }
                         }
                     ]
+                },
+                {
+                    "type": "dict",
```
BigRoy, line 1119: Is there a reason these need to be separate settings for Farm verus Local, etc. Would it make more sense to just have a single plugin that does the Ftrack status decisions which in it has a filter setting where a profile matches only IF it's set to farm or set to local, or set to whatever. I'm a bit worried we're going to get tons of plugins as the variety of cases start to increase?

openpype/settings/entities/schemas/projects_schema/schema_project_ftrack.json
```diff
@@ -1114,6 +1114,142 @@
...
+                    ]
+                },
+                {
+                    "type": "dict",
+                    "key": "IntegrateFtrackTaskStatus",
+                    "label": "Integrate Ftrack Task Status",
+                    "children": [
+                        {
+                            "type": "label",
+                            "label": "Apply collected task statuses. This plugin can run before or after version integration. Some status automations may conflict with status changes on versions because of wrong order."
```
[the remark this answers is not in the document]
iLLiCiTiT: Automation of propagation status from task -> version or version -> task. I can't affect if the automations will change the task, not matter if is set during publishing.

openpype/modules/ftrack/plugins/publish/integrate_ftrack_status.py
iLLiCiTiT:
```suggestion
    # Use order of Integrate Ftrack Api plugin and offset it before or after
    base_order = pyblish.api.IntegratorOrder + 0.499
    # By default is after Integrate Ftrack Api
    order = base_order + 0.0001
```

openpype/settings/entities/schemas/projects_schema/schema_project_ftrack.json
iLLiCiTiT: Like I've described in PR describtion. I don't know how would that be possible with different targets, also I'm not sure how would that setting look like, it would be massive headache to handle and set all 3 cases in single plugin.

BigRoy: On the plugin:
```
targets = ["local", "farm"]
```
Then in process you can just do however you like, right? (Including checking the set targets, etc.)

iLLiCiTiT: How would you find out that publishing is running on farm, or locally? It does matter, we have request that if is publishing local set status A, if is being send to farm set status B, if is published on farm se status C. All 3 cases can have different logic based on when/where happens and what task type/family it is.

openpype/settings/entities/schemas/projects_schema/schema_project_ftrack.json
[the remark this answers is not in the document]
iLLiCiTiT: Ftrack automations in general that are handling statuses in some way. In our ftrack event server handler are several event handlers that can change statuses of tasks and versions. Why are versions important? Because change of version status can trigger change of task status. We have several event handler in default ftrack integration for propagation of task/version statuses. They are just disable by default (check `project_settings/ftrack/events` settings).

openpype/settings/entities/schemas/projects_schema/schema_project_ftrack.json
BigRoy: I was thinking roughly like this:
```python
import pyblish.api

class Plugin(pyblish.api.InstancePlugin):
    order = pyblish.api.CollectorOrder + 0.499

    def process(self, instance):
    
        targets = pyblish.api.registered_targets()
        
        # Allow different status updates for what kind of publish
        # is currently occurring. We differentiate between a local
        # publish, a local submission for the instance to the farm
        # and a farm publish.
        if "farm" in targets:
            state = "farm"
        if "local" in targets and instance.data.get("farm"):
            state = "send_to_farm"
        else:
            # anything else we assume is just regular local
            state = "local"
            
        set_ftrack_status(instance, profiles=self.profiles, state=state)
```

iLLiCiTiT: But the profiles are not the same for all states. Each state has it's own profiles.

BigRoy: > Also the order of the plugins is also important, first find out which instances are going to farm, then which are local. 2 instances can lead to same task and only one status should be set. If workfile instance would be processed before render, a totally different status would be set.

Valid point.

openpype/settings/defaults/project_settings/ftrack.json
```diff
@@ -493,7 +493,29 @@
             "upload_reviewable_with_origin_name": false
         },
         "IntegrateFtrackFarmStatus": {
-            "farm_status_profiles": []
+            "farm_status_profiles": [
+                {
+                    "hosts": [
+                        "celaction"
```
Minkiu, line 499: Is this a typo or what is `celaction` ?

Minkiu (approved):

openpype/settings/defaults/project_settings/ftrack.json
BigRoy: It's one of the supported hosts of openpype, see [celaction](https://github.com/ynput/OpenPype/tree/develop/openpype/hosts/celaction) hosts folder.

I think it's this [celaction](https://www.celaction.com/).

64qam (approved):

PR #5127, opened 2023-06-12
Fix support for open last workfile with `AVALON_SCENEDIR`

### LINKED ISSUE #5126: Bug: Open last workfile does not work with AVALON_SCENEDIR

### Current Behavior:

OpenPype inherited the feature from Avalon to specify an `AVALON_SCENEDIR` which acts as the subfolder in which workfiles should be saved in the work directory for a shot. It's basically what [this issue describes](https://github.com/ynput/OpenPype/issues/2567).

For example as can be seen [here](https://github.com/ynput/OpenPype/blob/f36b8f49a2202b10e5f17fc4ecb26cfa1ab6e428/openpype/hosts/blender/api/workio.py#L82-L89) the workfiles directory then becomes: `AVALON_WORKDIR/AVALON_SCENEDIR`.

The "open last workfile on launch" feature does not work with `AVALON_SCENEDIR` subfolder for work files because it does not consider the subfolder at all. Any pointers on how this could best be fixed?

### Expected Behavior:

With `AVALON_SCENEDIR` defined in an application's environment the subfolder should be used to look for existing workfiles.

An alternative would be to completely deprecate the `AVALON_SCENEDIR` behavior to begin with and instead implement #2567 in such a way that the same feature can still be used but in a different manner, e.g. by having a scenedir template/profile per application.

### Version

3.15.11-nightly.1

Windows

### Steps To Reproduce:

1. Enable launch last workfile (= enabled by default)
2. Add `"AVALON_SCENEDIR": "scenes"` to e.g. the houdini or maya environment
3. Save a workfile with the Workfiles tool.
4. Restart the application, it should launch the existing workfile directly but it does not. (because it doesn't check in the subfolder but the work root instead)

### Relevant log output:

_No response_

### Additional context:

Issues with the `AVALON_SCENEDIR` have come up before so I'm linking theses issues as **related**:
- [Differentiate workdir and workfiles directory](https://github.com/ynput/OpenPype/issues/2567)
- [API: Query workfiles for an asset](https://github.com/ynput/OpenPype/issues/3389)
- [Workfiles tool to show all files](https://github.com/ynput/OpenPype/issues/4216#issuecomment-1468059376)

[cuID:[OP-6160](https://app.clickup.com/t/863gyuk25)]

BigRoy:
## Changelog Description

Fix support for open last workfile with `AVALON_SCENEDIR` + cosmetics

## Additional info

Resolves #5126 

## Testing notes:

1. Enable launch last workfile (= enabled by default)
2. Add `"AVALON_SCENEDIR": "scenes"` to e.g. the houdini or maya environment
3. Save a workfile with the Workfiles tool.
4. Restart the application, it should launch the existing workfile directly.

### DIFF
--- a/openpype/lib/applications.py
+++ b/openpype/lib/applications.py
@@ -1728,23 +1728,34 @@ def prepare_context_environments(data, env_group=None, modules_manager=None):
     data["env"]["AVALON_APP"] = app.host_name
     data["env"]["AVALON_WORKDIR"] = workdir
 
-    _prepare_last_workfile(data, workdir, modules_manager)
+    # Allow `AVALON_SCENEDIR` environment variable to define a subfolder
+    # for the work files to be stored in the work directory.
+    scene_dir = workdir
+    scene_subfolder = data["env"].get("AVALON_SCENEDIR")
+    if scene_subfolder:
+        scene_dir = os.path.join(workdir, scene_subfolder)
+
+    _prepare_last_workfile(data, scene_dir, modules_manager)
 
 
-def _prepare_last_workfile(data, workdir, modules_manager):
+def _prepare_last_workfile(data, scene_dir, modules_manager):
     """last workfile workflow preparation.
 
-    Function check if should care about last workfile workflow and tries
-    to find the last workfile. Both information are stored to `data` and
+    Function checks if last workfile workflow is enabled for this context and
+    tries to find the last workfile. Both information are stored to `data` and
     environments.
 
-    Last workfile is filled always (with version 1) even if any workfile
+    Last workfile is filled always (with version 1) even if no workfile
     exists yet.
+
+    This function also determines the value for OPENPYPE_WORKFILE_TOOL_ON_START
+    to define whether the workfiles tool should open on launch too.
 
     Args:
         data (EnvironmentPrepData): Dictionary where result and intermediate
             result will be stored.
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
Like I've mentioned in discord communication. We didn't want to support scenedir, only for host integrations that require it. In that case it is handled by the host addon. That discussion happend few months ago, so maybe we can reopen the discussion @antirotor @mkolar @jakubjezek001 . But with that in mind, I don't know if this will be approved.

BigRoy:
Thanks - just for context: we've been using `AVALON_SCENEDIR` all the time for Houdini, Fusion and Maya. So being able to have a subfolder for our workfiles for those applications will be a requirement for us for backwards compatibility and also a forward preference to still allow that behavior as it's preferred by our artists too.

BigRoy:
We are currently using this in production 👍

mkolar:
I agree with the functionality, but I think we need a slightly different approach. We have some requests to make the workfile handling a little more robust, and this should be included in there. We'll prep some proposal soon hopefully.

mkolar:
Because we're splitting OpenPype into ayon-core and individual host addons, this PR would have to be re-created to target one of those. The PR has merit, the solution, not so much. Let's try to revisit in AYON

We're closing it down, but we'll he happy for a new PR to [ynput/ayon-core](https://github.com/ynput/ayon-core) repository once it's up.

PR #5128, opened 2023-06-12
add label to matching family

ClementHector:
## Changelog Description
I added the possibility to filter the `family smart select` with the label in addition to the family.

## Additional info
The family smart select does not work with the default settings and it is not intuitive for users to have to know the family instead of the family label.

## Testing notes:

With the default settings (to test this new add):
1. open a light task in maya
2. open the creator interface
3. if it works Render is selected

To test the legacy code change the family smart select option from `render` to `rendering`:
1. open a light task in maya
2. open the creator interface
3. if it works Render is selected

### DIFF
--- a/openpype/tools/creator/model.py
+++ b/openpype/tools/creator/model.py
@@ -54,8 +54,8 @@ class CreatorsModel(QtGui.QStandardItemModel):
             item_id = index.data(ITEM_ID_ROLE)
             creator_plugin = self._creators_by_id.get(item_id)
             if creator_plugin and (
-                creator_plugin.label.lower() == family.lower() or
-                creator_plugin.family.lower() == family.lower()
+                creator_plugin.label.lower() == family.lower()
+                or creator_plugin.family.lower() == family.lower()
             ):
                 indexes.append(index)
         return indexes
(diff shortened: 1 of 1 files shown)


openpype/tools/creator/model.py
```diff
@@ -53,6 +53,9 @@ def get_indexes_by_family(self, family):
             index = self.index(row, 0)
             item_id = index.data(ITEM_ID_ROLE)
             creator_plugin = self._creators_by_id.get(item_id)
-            if creator_plugin and creator_plugin.family == family:
+            if creator_plugin and (
+                creator_plugin.label.lower() == family.lower() or
+                creator_plugin.family.lower() == family.lower()
```
iLLiCiTiT, line 58:
```suggestion
                creator_plugin.label.lower() == family.lower()
                or creator_plugin.family.lower() == family.lower()
```

iLLiCiTiT (approved):
I don't understand why not to use the family. But the creator tool is used only in Maya and Blender, and will probably be removed when they are converted to publisher, I'm not against.

Minkiu (approved):

PR #5132, opened 2023-06-12
Nuke: Allow for more complex temp rendering paths

fabiaserra:
## Changelog Description
When changing the temporary rendering template (i.e., add `{asset}` to the path) to something a bit more complex the formatting was erroring due to missing keys.

## Additional info
This also adds a small bug fix that I encountered when running latest develop where `customOCIOConfigPath` key was missing from the workfile_settings. Any idea why that's the case?

## Testing notes:
1. Change the Nuke temporary render path template and add `{asset}` key somewhere
2. Create a new render instance

### DIFF
--- a/openpype/hosts/nuke/api/lib.py
+++ b/openpype/hosts/nuke/api/lib.py
@@ -1302,10 +1302,7 @@ def create_write_node(
     fdir = str(anatomy_filled["work"]["folder"]).replace("\\", "/")
     fpath = data["fpath_template"].format(
         work=fdir,
-        version=data["version"],
-        subset=data["subset"],
-        frame=data["frame"],
-        ext=ext
+        **data,
     )
 
     # create directory
--- a/openpype/hosts/nuke/api/lib.py
+++ b/openpype/hosts/nuke/api/lib.py
@@ -2051,7 +2051,7 @@ class WorkfileSettings(object):
             self._root_node["colorManagement"].setValue("OCIO")
 
         # we dont need the key anymore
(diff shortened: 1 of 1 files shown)


Minkiu (approved):
Cod looks good.

fabiaserra:
> Cod looks good.

Thanks for reviewing @Minkiu , I have changed it to make use of `StringTemplate` now so it also supports optional keys

openpype/hosts/nuke/api/lib.py
```diff
@@ -1300,13 +1301,8 @@ def create_write_node(
 
     # build file path to workfiles
     fdir = str(anatomy_filled["work"]["folder"]).replace("\\", "/")
```
fabiaserra, line 1303: is the `.replace("\\", "/")` actually needed?

Minkiu (approved):
Much nicer with the `StringTemplate`

openpype/hosts/nuke/api/lib.py
jakubjezek001: Yeah `.replace("\\", "/")` is actually needed since Nuke is supporting only forward slashes internally.

jakubjezek001 (approved):
LGTM! Thanks @fabiaserra ;)

openpype/hosts/nuke/api/lib.py
fabiaserra: Oh yeah I was more questioning why it would have been "\" at first place (we use Linux so we never have "\") but that makes sense, thank you!

PR #5137, opened 2023-06-13
Blender: Add support for custom path for app templates

simonebarbieri:
## Changelog Description
This PR adds support for a custom [App Templates](https://docs.blender.org/manual/en/dev/advanced/app_templates.html) path in Blender by setting the `BLENDER_USER_SCRIPTS` environment variable to the path specified in `OPENPYPE_APP_TEMPLATES_PATH`. This allows users to use their own custom app templates in Blender.

## Testing notes:
1. Set the `OPENPYPE_APP_TEMPLATES_PATH` environment variable to a directory containing custom app templates.
2. Launch Blender and check that the custom app templates are available in the "New" menu.

### DIFF
--- a/openpype/hosts/blender/api/lib.py
+++ b/openpype/hosts/blender/api/lib.py
@@ -138,7 +138,18 @@ def set_app_templates_path():
     # Blender requires the app templates to be in `BLENDER_USER_SCRIPTS`.
     # After running Blender, we set that variable to our custom path, so
     # that the user can use their custom app templates.
-    app_templates_path = os.environ.get("OPENPYPE_APP_TEMPLATES_PATH")
+
+    # We look among the scripts paths for one of the paths that contains
+    # the app templates. The path must contain the subfolder
+    # `startup/bl_app_templates_user`.
+    paths = os.environ.get("OPENPYPE_BLENDER_USER_SCRIPTS").split(os.pathsep)
+
+    app_templates_path = None
+    for path in paths:
+        if os.path.isdir(
+                os.path.join(path, "startup", "bl_app_templates_user")):
+            app_templates_path = path
+            break
 
     if app_templates_path and os.path.isdir(app_templates_path):
         os.environ["BLENDER_USER_SCRIPTS"] = app_templates_path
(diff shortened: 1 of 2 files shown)


openpype/hosts/blender/api/lib.py
```diff
@@ -134,6 +134,16 @@ def append_user_scripts():
         traceback.print_exc()
 
 
+def set_app_templates_path():
+    # Blender requires the app templates to be in `BLENDER_USER_SCRIPTS`.
+    # After running Blender, we set that variable to our custom path, so
+    # that the user can use their custom app templates.
+    app_templates_path = os.environ.get("OPENPYPE_APP_TEMPLATES_PATH")
```
iLLiCiTiT, line 141: Maybe we could use first path from `OPENPYPE_BLENDER_USER_SCRIPTS` instead of new env variable?

iLLiCiTiT: And maybe first path that have existing subfolders `/startup/bl_app_templates_user` (from documentation that is requirement for app templates...)?

iLLiCiTiT (approved):
LGTM. Didn't test (I don't know what app templates are).

moonyuet:
image
tested in my side, it works.

moonyuet (approved):

BigRoy:
Out of curiosity, why wouldn't we just use `BLENDER_USER_SCRIPTS` in the environment and instead define an OpenPype variable which does basically the same thing?

moonyuet:
I actually want to suggest to include options in OP settings for users to put the paths of different templates too . Will be more convenient to store the path to access these templates.

alainxi:
Hi,
Sorry for the stupid question, but I don't find the OPENPYPE_APP_TEMPLATES_PATH environment variable anywhere in the source code.
Maybe it was renamed to OPENPYPE_BLENDER_USER_SCRIPTS ?

iLLiCiTiT:
The description does match first implementation, and was not updated afterwards. Because both addons and templates are (for blender) set with the same environment variable `BLENDER_USER_SCRIPTS`, and 100% of manuals and forums will tell you to use this variable for that purpose, we just reused the variable, which already is handled by us. 

`OPENPYPE_APP_TEMPLATES_PATH` was removed and `BLENDER_USER_SCRIPTS` is used instead. Both `BLENDER_USER_SCRIPTS` and `OPENPYPE_BLENDER_USER_SCRIPTS` can be used.

alainxi:
Thanks !

PR #5139, opened 2023-06-14
Save the filter selection for family in loader interface  only 

FadyFS:
## Changelog Description
We have changed the following file widgets.py to save the filter selection whenever you exit the OpenPype or the loader window. Using QSettings we save the user's last filter selection. We implemented as well a setter for the remaining checked families.

## Additional info
We set the new family list in the refresh method (i.e when we open the interface back up)
However, this new implementation only works on OpenPype and not on Maya as requested.

## Testing notes:
1. Open the OpenPype software, a loader window.
2. Choose a project and filter the family you want
3. Exit the software
4. Repeat step 1.

### DIFF
--- a/openpype/tools/loader/widgets.py
+++ b/openpype/tools/loader/widgets.py
@@ -1040,7 +1040,7 @@ class FamilyListView(QtWidgets.QListView):
             settings = None
 
         self._settings = settings
-        self._last_families_setting = self._settings.value('families') or None
+        self._last_families_setting = self._settings.value('enabled_families') or []
 
         print("settings: ", settings.allKeys())
         print("enabled_families", settings.value('enabled_families'))
@@ -1057,39 +1057,32 @@ class FamilyListView(QtWidgets.QListView):
         family_model.dataChanged.connect(self._on_data_change)
         self.customContextMenuRequested.connect(self._on_context_menu)
 
+
         self._family_model = family_model
         self._proxy_model = proxy_model
 
+        # self.set_last_families(self._last_families_setting)
+
 
     def set_enabled_families(self, families):
-        print("=> set_enabled_families", families)
-
-
         self._proxy_model.set_enabled_families(families)
-
         self.set_enabled_family_filtering(True)
 
     def set_enabled_family_filtering(self, enabled=None):
-        print("=> set_enabled_family_filtering", enabled)
         self._proxy_model.set_filter_enabled(enabled)
 
     def refresh(self):
-        print("=> refresh")
         self._family_model.refresh()
-
         self.active_changed.emit(self.get_enabled_families())
 
-        self.set_enabled_families_custom(self._last_families_setting)
+        if self._last_families_setting is not None:
+            self.set_last_families(self._last_families_setting)
+            self._last_families_setting = None
 
     def get_enabled_families(self):
         """Return the checked family items"""
-        print("=> get_enabled_families")
         model = self._family_model
         checked_families = []
-
-        # if self._last_families_setting:
-        #     families = self._last_families_setting
-        #     self._last_families_setting = None
 
         for row in range(model.rowCount()):
             index = model.index(row, 0)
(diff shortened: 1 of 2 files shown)


.github/workflows/sync_fork.yml
BigRoy, line 1: Pretty sure this change should not be here.

openpype/tools/loader/widgets.py
```diff
@@ -1033,6 +1033,18 @@ class FamilyListView(QtWidgets.QListView):
     def __init__(self, dbcon, family_config_cache, parent=None):
         super(FamilyListView, self).__init__(parent=parent)
 
+        from Qt import QtCore, QtWidgets
```
BigRoy, line 1036: Do not import from `Qt` here. If imports need to be done then import from `qtpy`.
Also, I suspect these imports should be at the top of file.

openpype/tools/loader/widgets.py
```diff
@@ -1033,6 +1033,18 @@ class FamilyListView(QtWidgets.QListView):
...
+        try:
+            settings = QtCore.QSettings("OpenPype", "FamilyList")
+        except Exception:
+            settings = None
+
+        self._settings = settings
+        self._last_families_setting = self._settings.value('enabled_families') or []
+
+        print("settings: ", settings.allKeys())
+        print("enabled_families", settings.value('enabled_families'))
```
BigRoy, line 1046: Debug prints can be removed
```suggestion
```

openpype/tools/loader/widgets.py
```diff
@@ -1033,6 +1033,18 @@ class FamilyListView(QtWidgets.QListView):
     def __init__(self, dbcon, family_config_cache, parent=None):
         super(FamilyListView, self).__init__(parent=parent)
 
+        from Qt import QtCore, QtWidgets
+        try:
+            settings = QtCore.QSettings("OpenPype", "FamilyList")
+        except Exception:
+            settings = None
```
BigRoy, line 1040: I'm pretty sure settings like these should be stored differently. @iLLiCiTiT likely has some decent input on how to expose this (and potentially even beyond the scope of just this setting).

Also, I'm pretty sure that using the "settings" value might interfere with e.g. the feature in OpenPype where you can define which default families are visible per host, etc. There's something in settings for that. (We prefer to actually have different families visible per host instead of a global setting value for a machine)

openpype/tools/loader/widgets.py
```diff
@@ -1075,9 +1088,21 @@ def get_enabled_families(self):
             if checked == QtCore.Qt.Checked:
                 family = index.data(QtCore.Qt.DisplayRole)
                 checked_families.append(family)
-
         return checked_families
 
+    def set_last_families(self, families):
+        if not families:
+            return
+
```
BigRoy, line 1096: Is this missing self.blockSignals(True)`  since we're setting it to false at the end?

iLLiCiTiT:
We actually have settings for family filtering that are base on current context. The logic for that is in `FamilyConfigCache` https://github.com/ynput/OpenPype/blob/develop/openpype/tools/utils/lib.py#L499 . This is kind of contra of that logic so it would require more logic changes (and probably settings) to avoid conflict in features.

When that is solved I would replace the `QSettings` with `JSONSettingRegistry` like in Launcher tool and use it inside `FamilyConfigCache` so the UI would not have any changes.

openpype/tools/loader/widgets.py
ClementHector: > Also, I'm pretty sure that using the "settings" value might interfere with e.g. the feature in OpenPype where you can define which default families are visible per host, etc. There's something in settings for that. (We prefer to actually have different families visible per host instead of a global setting value for a machine)

Can you tell me where is the option to define which family is visible in the loader? 
I don't think there is any interference with the other options because @FadyFS tries to store the checked families in the settings to put the checked back on when opening, it doesn't filter them.

iLLiCiTiT: > I don't think there is any interference with the other options because @FadyFS tries to store the checked families in the settings to put the checked back on when opening, it doesn't filter them.

That's exactly what the filters do too... `project_settings/global/tools/loader/family_filter_profiles`

BigRoy: > It's slightly different because the implementation of fady allows to remember the family check during the last sessions.

That's exactly what I tried to mention - is that if the last session overrides the settings, then basically the settings only ever apply once and don't differentiate between default visible families presets of different hosts.

PR #5141, opened 2023-06-14
General: Expose cleanup plugin settings

fabiaserra:
## Changelog Description
Expose the clean up plugins settings so they can be enabled/disabled.

This also exposes the `optional` key on all the clean up plugins so they could be optionally disabled during the publish. However, is the `optional` key having any effect in the new publisher?

**Before**
image

**After**
image

## Testing notes:
1. Open the Studio Settings and disable the `Explicit Clean up` plugin


openpype/plugins/publish/cleanup.py
```diff
@@ -41,8 +41,8 @@ class CleanUp(pyblish.api.InstancePlugin):
     active = True
 
     # Presets
-    paterns = None  # list of regex paterns
-    remove_temp_renders = True
+    paterns = None  # list of regex patterns
```
fabiaserra, line 44: I wanted to fix the typo on the variable as well but this would break backwards compatibility

openpype/plugins/publish/cleanup.py
```diff
@@ -41,8 +41,8 @@ class CleanUp(pyblish.api.InstancePlugin):
...
+    remove_temp_renders = False
```
fabiaserra, line 45: changed this so it matches the default setting

openpype/settings/defaults/project_settings/global.json
```diff
@@ -397,11 +397,18 @@
             ]
         },
         "CleanUp": {
```
iLLiCiTiT, line 399: Why are these optional? Optional means artist (user) can change it's 'active' state before publishing. Some of these are headless plugins -> there is no room to change it's state. And also I'm not sure if it's good idea to give user the power to affect that?

iLLiCiTiT: > Yeah good call, is optional even working in the new publisher though?

Yes they work, but in a different way.

> and made it consistent with the rest of clean up plugins

I don't know why this one is optional, but it does not make sence with the other plugins. Farm cleanup cannot be changed -> doesn't make sence to have it optional, and explicit cleanup is for files that are 100% to remove (they are in temp or cannot be reused).

fabiaserra: > Yes they work, but in a different way.

What different way?

> explicit cleanup is for files that are 100% to remove (they are in temp or cannot be reused) -> even advanced users would not want that.

In Nuke farm renders it's removing the files from the temporary staging directory, we don't want those removed so that's why we want to have control of disabling the plugin

I'm going to go ahead and remove the optionals

fabiaserra: Done here https://github.com/ynput/OpenPype/pull/5141/commits/08671cb49d830a7cdc0ca1413b1db59cde3ecef3

Minkiu (approved):
Haven't tested, but the code looks good.

openpype/settings/defaults/project_settings/global.json
iLLiCiTiT: > What different way?

Already described, one is on farm -> No way how to change state. Explicit Cleanup plugin is designed to explicitly remove files by pyblish plugins.

> In Nuke farm renders it's removing the files from the temporary staging directory, we don't want those removed so that's why we want to have control of disabling the plugin

That's a use-case that deservers changes, BUT that should be handled in Nuke plugin which is causing that, not in these plugins. These are just not meant to be optional. Single usecase should not affect every other. "Keep rendered files once done" should not cause that temp is flooded by other temporary files.

fabiaserra: > Already described, one is on farm -> No way how to change state -> no reason to have setting for that. Explicit Cleanup plugin is designed to explicitly remove files by pyblish plugins.

I meant how is the `Optional` working differently on the plugins on the new publisher, you are not responding to that. How is the user able to enable/disable the optional plugins on the newp ublisher?

> That's a use-case that deservers changes, BUT that should be handled in Nuke plugin which is causing that, not in these plugins. These are just not meant to be optional. Single usecase should not affect every other. "Keep rendered files once done" should not cause that temp is flooded by other temporary files from ExtractReview, ExtractBurnin and others.

Fair enough, I can investigate where are these files being explicitly added to be deleted but I think there's still benefit on exposing control on the plugins you want to execute or not, specially the ones that delete data

iLLiCiTiT: > I meant how is the Optional working differently on the plugins on the new publisher, you are not responding to that. How is the user able to enable/disable the optional plugins on the new publisher?

The plugin must inherit from `OptionalPyblishPluginMixin` and have to have this at the start of process.
```
def process(self, instance):
    if not self.is_active(instance.data):
        return
```

openpype/settings/defaults/project_settings/global.json
```diff
@@ -397,11 +397,15 @@
...
+            "enabled": true,
             "paterns": [],
             "remove_temp_renders": false
         },
         "CleanUpFarm": {
             "enabled": false
+        },
+        "ExplicitCleanUp": {
+            "enabled": false
```
iLLiCiTiT, line 408: It should be enabled by default.
```suggestion
            "enabled": true
```

openpype/settings/defaults/project_settings/global.json
fabiaserra: > > I meant how is the Optional working differently on the plugins on the new publisher, you are not responding to that. How is the user able to enable/disable the optional plugins on the new publisher?
> 
> The plugin must inherit from `OptionalPyblishPluginMixin` and have to have this at the start of process.
> 
> ```python
> def process(self, instance):
>     if not self.is_active(instance.data):
>         return
> ```

I will repeat, how is the **user** able to enable/disable the optional plugins on the new publisher? 

That would need to be through the UI

iLLiCiTiT: He can enabled/disable it per instance (if it's instance plugin) on Publish tab in right panel. If it's context plugin then he can enable/disable it when Context is selected.

iLLiCiTiT:
> Open the Studio Settings and disable the Explicit Clean up plugin

I'm sorry, I'm looking at this PR, and I forgot what was the reason to have option to disable the plugin?

fabiaserra:
> > Open the Studio Settings and disable the Explicit Clean up plugin
> 
> I'm sorry, I'm looking at this PR, and I forgot what was the reason to have option to disable the plugin?

Because I didn't want any plugins to "explicitly" delete anything, I wanted to be the one that "explicitly" chose when I actually want things to be deleted so I wanted to have the option to disable the plugin

fabiaserra:
> > > Open the Studio Settings and disable the Explicit Clean up plugin
> > 
> > 
> > I'm sorry, I'm looking at this PR, and I forgot what was the reason to have option to disable the plugin?
> 
> Because I didn't want any plugins to "explicitly" delete anything, I wanted to be the one that "explicitly" chose when I actually want things to be deleted so I wanted to have the option to disable the plugin

It's been that long though that I no longer know if it still has any impact on my fork but I do remember experiencing some files being deleted due to the Explicit Clean Up that I actually didn't want deleted

iLLiCiTiT:
> It's been that long though that I no longer know if it still has any impact on my fork but I do remember experiencing some files being deleted due to the Explicit Clean Up that I actually didn't want deleted

Can you please revisit that? Because I know we had issues and they should be resolved now. I don't think it's good idea to have setting to disable this plugin.

BigRoy:
Does this PR allow me to disable the _feature?_ that on a render publish from deadline that it DELETES the source render files? There've been quite a few times where one machine messed up in the render - the sequence got published; all source renders were deleted and we now have to re-render the full sequence instead of just 2 frames to publish it again. Drives us mad.

How can we disable the feature to remove the "render files" from the work/staging area - in our case referring to Maya + Fusion.

fabiaserra:
> Does this PR allow me to disable the _feature?_ that on a render publish from deadline that it DELETES the source render files? There've been quite a few times where one machine messed up in the render - the sequence got published; all source renders were deleted and we now have to re-render the full sequence instead of just 2 frames to publish it again. Drives us mad.
> 
> How can we disable the feature to remove the "render files" from the work/staging area - in our case referring to Maya + Fusion. Is there a setting for that somewhere?

Right? I feel the same way, I want to have full control of when files get deleted as I've been in the same situation where in the farm the staging temporary files get deleted and it would have been very useful to be able to toggle that off so I can debug what happened or pick it up from those intermediate files

fabiaserra:
> > Does this PR allow me to disable the _feature?_ that on a render publish from deadline that it DELETES the source render files? There've been quite a few times where one machine messed up in the render - the sequence got published; all source renders were deleted and we now have to re-render the full sequence instead of just 2 frames to publish it again. Drives us mad.
> > How can we disable the feature to remove the "render files" from the work/staging area - in our case referring to Maya + Fusion. Is there a setting for that somewhere?
> 
> Right? I feel the same way, I want to have full control of when files get deleted as I've been in the same situation where in the farm the staging temporary files get deleted and it would have been very useful to be able to toggle that off so I can debug what happened or pick it up from those intermediate files

I owe @iLLiCiTiT a test to confirm whether these changes actually have any effect

mkolar:
Boys, because we're splitting OpenPype into ayon-core and individual host addons, this PR would have to be re-created to target one of those. Also it probably needs to be properly re-validated :)

We're closing it down, but we'll he happy for a new PR to [ynput/ayon-blender](https://github.com/ynput/ayon-blender) repository once it's up.

PR #5142, opened 2023-06-14
TrayPublisher & StandalonePublisher: Specify version

iLLiCiTiT:
## Changelog Description
Simple creators in TrayPublisher can affect which version will be integrated. Standalone publisher respects the version change from UI.

## Additional info
Standalone publisher is considered as deprecated so TrayPublisher changes will be explained more. Simple creators allow to control integrated version by added bool ("Use next version") and number ("Version to use") inputs where user can enable if custom version is used abd define what version it will be. There is a setting to enable this option per creator. Added optional validator that makes sure if the version does not already exists to avoid possible conflicts with existing versions.

The next version on instance is prefilled on each publisher refresh, so it reflect what is the expected version to integrate. The version cannot be prefilled in create phase as subset name is not yet known.

## Testing notes:
1. Go to simple creator settings (`project_settings/traypublisher/simple_creators`).
2. Choose one of creators and change `Allow version control` to enabled state.
3. Open Publisher.
4. Create an instance with the creator where version control was enabled.
5. Go to publish page.
6. Select the instance.
7. There should be version inputs at the bottom (bool for enabling and number for the version).
8. Change the values with multiple combinations.
9. Publish

### What should happen with different combinations
- If the bool is enabled the version should not be propagated.
- If the bool is disable the version from the input should be propagated.
- If the version is higher than last published it should be just integrated.
- If the version is lower it should end on validation because the version already exists.
    - You can disable validation of version in Publish tab when instance is selected.
    - In that case version should be overriden.
- If version is set to `0`, the next available version should be used.
- Change of variant should update version after save and restart publish.

### DIFF
--- a/openpype/hosts/standalonepublisher/plugins/publish/collect_context.py
+++ b/openpype/hosts/standalonepublisher/plugins/publish/collect_context.py
@@ -222,7 +222,6 @@ class CollectContextDataSAPublish(pyblish.api.ContextPlugin):
                 "label": subset,
                 "name": subset,
                 "family": in_data["family"],
-                # "version": in_data.get("version", 1),
                 "frameStart": in_data.get("representations", [None])[0].get(
                     "frameStart", None
                 ),
@@ -232,6 +231,14 @@ class CollectContextDataSAPublish(pyblish.api.ContextPlugin):
                 "families": instance_families
             }
         )
+        # Fill version only if 'use_next_available_version' is disabled
+        #   and version is filled in instance data
+        version = in_data.get("version")
+        use_next_available_version = in_data.get(
+            "use_next_available_version", True)
+        if not use_next_available_version and version is not None:
+            instance.data["version"] = version
+
         self.log.info("collected instance: {}".format(pformat(instance.data)))
         self.log.info("parsing data: {}".format(pformat(in_data)))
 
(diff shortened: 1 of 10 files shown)


jakubjezek001:
Task linked: [OP-6172 Specify version in standalone publisher not respected](https://app.clickup.com/t/6658547/OP-6172)

64qam:
If the version is lower and the validation of the version is off it ends up with Version already exists error, I can only run the Repair action. Then when reset the publishing, it fails with Instance collection failed.
```Creator: Workfile (settings_workfile)

Error: 'ep001_102b_sh0020'

  File "C:\Users\qam\code\OpenPype\openpype\pipeline\create\context.py", line 2136, in reset_instances
    creator.collect_instances()
  File "C:\Users\qam\code\OpenPype\openpype\hosts\traypublisher\api\plugin.py", line 251, in collect_instances
    self._fill_next_versions(instances)
  File "C:\Users\qam\code\OpenPype\openpype\hosts\traypublisher\api\plugin.py", line 231, in _fill_next_versions
    version = subset_docs_by_asset_id[asset_name][subset_name]
KeyError: 'ep001_102b_sh0020'
```

64qam (approved):

PR #5145, opened 2023-06-15
Pack project: Raise exception with reasonable message

iLLiCiTiT:
## Changelog Description
Pack project crashes with relevant message when destination directory is not set.

## Additional info
It did crash with cryptic error when the logic tries to create a directory from `None`.

## Testing notes:
1. Run `pack-project --project {project name} --dbonly`
2. It should crash with more specific information what to do

### DIFF
--- a/openpype/lib/project_backpack.py
+++ b/openpype/lib/project_backpack.py
@@ -141,6 +141,16 @@ def pack_project(
     if not destination_dir:
         destination_dir = root_path
 
+    if not destination_dir:
+        if only_documents:
+            raise ValueError((
+                "Destination dir must be passed."
+                " Use '--dirpath {output dir path}' if using command line."
+            ))
+        raise ValueError(
+            "Project {} does not have any roots.".format(project_name)
+        )
+
     destination_dir = os.path.normpath(destination_dir)
     if not os.path.exists(destination_dir):
         os.makedirs(destination_dir)
--- a/openpype/lib/project_backpack.py
+++ b/openpype/lib/project_backpack.py
@@ -113,6 +113,12 @@ def pack_project(
             project_name
         ))
 
+    if only_documents and not destination_dir:
+        raise ValueError((
+            "Destination directory must be defined"
+            " when only documents should be packed."
+        ))
+
     root_path = None
     source_root = {}
     project_source_path = None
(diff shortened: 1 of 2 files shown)


openpype/lib/project_backpack.py
```diff
@@ -141,6 +141,16 @@ def pack_project(
     if not destination_dir:
         destination_dir = root_path
 
+    if not destination_dir:
+        if only_documents:
+            raise ValueError((
+                "Destination dir must be passed."
+                " Use '--dirpath {output dir path}' if using command line."
+            ))
```
BigRoy, line 149: Should this error message also mention that you need to supply `--dirpath` due to having only documents enabled?
Also, if the message mentions the command line flag, shouldn't it error like in the CLI part instead of the lib function itself. When calling `openpype.lib.project_backpack.pack_project` manually there is no `--dirpath` argument to provide at all and the error message will still be unclear.

iLLiCiTiT:
```suggestion
            raise ValueError((
                "Destination dir must be passed for documents only pack."
                " Use '--dirpath {output dir path}' if using command line."
            ))
```

Minkiu (approved):

openpype/lib/project_backpack.py
BigRoy: > It says if using command line. so I would say that make it clear.

I meant more that I'd rather know what I failed to do when calling that function myself too. It gives no hint whatsoever.
We're providing CLI responses in a function that itself doesn't seem to be a CLI function. It's fine with me, but it's definitely confusing.

> Like this?

If this is intended to only ever be used through CLI then the whole reporting can be CLI focused: 
```suggestion
            raise ValueError((
                "Destination dir must be defined when using --dbonly."
                " Use '--dirpath {output dir path}' flag to specify directory."
            ))
```
That to me is much clearer.

If it isn't only CLI, then this should **not** report the CLI flags. I'd say the error reporting then is too mixed (or at the very least the argument names differ way to much `destination_dir` and `--dirpath` seem like different arguments due to the command line flag not mentioning _destination_).

I'd rather do the check directly in the CLI method to report the CLI friendly message, and in the function itself report it friendly for the function arguments.

PR #5146, opened 2023-06-15
Allow "inventory" actions to be supplied by a Module/Addon.

### LINKED ISSUE #5121: Bug: IPluginPaths missing inventory plugins

### Current Behavior:

```
# Warning: ModulesManager : Expected keys from `get_plugin_paths` are "publish", "create", "load", "actions". Module: "bumpybox_addon" - got key "inventory"
```

No plugins can be registered for the inventory tool. Using `actions` does not work.

### Expected Behavior:

_No response_

### Version

3.15.10-nightly.2

Windows

### Steps To Reproduce:

Add a module or addon that registers an inventory action. For example:

```python
"""Addon definition is located here.

Import of python packages that may not be available should not be imported
in global space here until are required or used.
- Qt related imports
- imports of Python 3 packages
    - we still support Python 2 hosts where addon definition should available
"""

import os

from openpype.modules import (
    OpenPypeAddOn,
    IPluginPaths,
    ITrayAction
)


class BumpyboxAddon(OpenPypeAddOn, IPluginPaths, ITrayAction):
    """This Addon has defined its settings and interface.

    This example has system settings with an enabled option. And use
    few other interfaces:
    - `IPluginPaths` to define custom plugin paths
    - `ITrayAction` to be shown in tray tool
    """
    label = "Bumpybox Addon"
    name = "bumpybox_addon"

    def tray_init(self):
        """Implementation of abstract method for `ITrayAction`.

        We're definitely  in tray tool so we can pre create dialog.
        """
        pass

    def on_action_trigger(self):
        """Implementation of abstract method for `ITrayAction`."""
        pass

    def get_plugin_paths(self):
        """Implementation of abstract method for `IPluginPaths`."""
        path = os.path.join(
            os.path.dirname(os.path.abspath(__file__)), "plugins"
        )
        return {
            "publish": [path],
            "inventory": [path]
        }
```

### Relevant log output:

_No response_

### Additional context:

_No response_

[cuID:[OP-6158](https://app.clickup.com/t/861mwy7c9)]

BigRoy:
## Changelog Description

Adds "inventory" as a possible key to the plugin paths to be returned from a module.

## Additional info

Attempts to resolve #5121
It's untested :)

## Testing notes:

See steps to reproduce in  #5121 - it should now work

### DIFF
--- a/openpype/modules/interfaces.py
+++ b/openpype/modules/interfaces.py
@@ -33,8 +33,8 @@ class OpenPypeInterface:
 class IPluginPaths(OpenPypeInterface):
     """Module has plugin paths to return.
 
-    Expected result is dictionary with keys "publish", "create", "load" or
-    "actions" and values as list or string.
+    Expected result is dictionary with keys "publish", "create", "load",
+    "actions" or "inventory" and values as list or string.
     {
         "publish": ["path/to/publish_plugins"]
     }
@@ -108,6 +108,21 @@ class IPluginPaths(OpenPypeInterface):
         """
 
         return self._get_plugin_paths_by_type("publish")
+
+    def get_inventory_action_paths(self, host_name):
+        """Receive inventory action paths.
+
+        Give addons ability to add inventory action plugin paths.
+
+        Notes:
+           Default implementation uses 'get_plugin_paths' and always return
+               all publish plugin paths.
+
+        Args:
+           host_name (str): For which host are the plugins meant.
+        """
+
+        return self._get_plugin_paths_by_type("inventory")
 
 
 class ILaunchHookPaths(OpenPypeInterface):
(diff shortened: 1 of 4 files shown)


openpype/modules/interfaces.py
```diff
@@ -397,8 +412,8 @@ def set_service_idle_icon(self):
 class ISettingsChangeListener(OpenPypeInterface):
     """Module has plugin paths to return.
 
-    Expected result is dictionary with keys "publish", "create", "load" or
-    "actions" and values as list or string.
+    Expected result is dictionary with keys "publish", "create", "load",
+    "actions" or "inventory" and values as list or string.
     {
```
BigRoy, line 417: I added it to the docstring here, but I suspect this docstring was already wrong to begin with hehe.

iLLiCiTiT: Yes, looks like the docstring was copied from `IPluginPath`.

BigRoy: Any decent replacement docstring you could write up for `ISettingsChangeListener`? :) I can include it here if that makes sense.

iLLiCiTiT: Not decent at all, but don't have time for more.
```
"""Module want to catch settings changes.

Only changes that happened in current process are propagated. Changes made in other processes, or machines won't  trigger the callbacks.
```

Minkiu (approved):

tokejepsen (approved):

Issue #5108, opened 2023-06-07
Maya: arnoldExportAss errors out during extraction in Extract Arnold Scene Source


moonyuet:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

Extract Arnold Scene Source can't export the assets parented to ` assMain_content_SET` and `assMain_proxy_SET` from the assMain
![image](https://github.com/ynput/OpenPype/assets/64118225/f2e32ee0-b7d4-4ae6-9534-9fad63ba8ab5)

### Expected Behavior:

It should publish the set members from ` assMain_content_SET` and `assMain_proxy_SET` from the assMain

### Version

3.15.10-nightly.2

### What platform you are running OpenPype on?

Windows

### Steps To Reproduce:

1. Launch Maya via Openpype
2. Create Arnold Scene Source.
3. Publish
4. The extractor will error out

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

### Relevant log output:

```shell
Traceback (most recent call last):
  File "D:\kayla\OpenPype\.venv\lib\site-packages\pyblish\plugin.py", line 522, in __explicit_process
    runner(*args)
  File "D:\kayla\OpenPype\openpype\hosts\maya\plugins\publish\extract_arnold_scene_source.py", line 73, in process
  File "D:\kayla\OpenPype\openpype\hosts\maya\plugins\publish\extract_arnold_scene_source.py", line 191, in _extract
  File "D:\kayla\OpenPype\openpype\hosts\maya\plugins\publish\extract_arnold_scene_source.py", line 2, in arnoldExportAss
RuntimeError: arnoldExportAss -selected needs an active selection or a list of objects
```

[cuID:[OP-6150](https://app.clickup.com/t/863gxx5d3)]

PR #5150, opened 2023-06-16
Maya: bug fix arnoldExportAss unable to export selected set members

moonyuet:
## Changelog Description
See #5108 
fix the bug arnoldExportAss being not able to export and error out during extraction.

## Additional info
n/a

## Testing notes:
1. Launch Maya via Openpype
2. Create Ass Instance for both geometry and proxy objects
3. Publish

### DIFF
--- a/openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
+++ b/openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
@@ -24,7 +24,7 @@ class CollectArnoldSceneSource(pyblish.api.InstancePlugin):
                 continue
             if objset.endswith("content_SET"):
                 instance.data["contentMembers"] = self.get_hierarchy(members)
-            if objset.endswith("content_SET"):
+            if objset.endswith("proxy_SET"):
                 instance.data["proxy"] = self.get_hierarchy(members)
 
         # Use camera in object set if present else default to render globals


jakubjezek001:
Task linked: [OP-6150 Maya: arnoldExportAss errors out during extraction in Extract Arnold Scene Source](https://app.clickup.com/t/6658547/OP-6150)

LiborBatek (changes requested):
I have created geometries for assMain ```content_SET``` and for ```proxy_SET``` and tried to publish ```Arnold Scene Source```.

Went through the validation ok but fails on extract content with following errors

image

tokejepsen (changes requested):
I think the code solution could be simpler, so I've suggested with this PR https://github.com/ynput/OpenPype/pull/5168

Also a `.ass` file has snuck itself into this PR. Was that intentional?

moonyuet:
> I think the code solution could be simpler, so I've suggested with this PR #5168
> 
> Also a `.ass` file has snuck itself into this PR. Was that intentional?

No it is not...Checked the code...works in my side. Thank you!

openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
```diff
@@ -23,14 +23,45 @@ def process(self, instance):
...
+                members_list = []
+                for member in members:
+                    shape = cmds.listRelatives(
+                        member, shapes=True, fullPath=True)
+                    if not shape:
+                        continue
+                    members_list = members + shape
+                    group_name = "|{}".format(member)
+                    if group_name in members_list:
+                        members_list.remove(group_name)
```
BigRoy, line 35: This will be very very slow for large lists of members.  Also, the members_list is being overridden for EACH member and the previous shape will be lost. Are you sure this is doing what you want it to?

I'm pretty sure this is not doing what you expect it to.

First of:
1. It assumes the top level instance will only **ever** contain objectSets otherwise the code will fail on the `cmds.sets()` command (not a new issue in this PR)
2. The logic for both `proxy_SET` and `content_SET` is basically the same, but just stored elsewhere. We should deduplicate the code so that it's clear they are basically doing the same thing.
3. Why do we do a per member listRelatives check for a shape, then _something?_ with a group name that I have no idea what it's based on, to then remove the top level? But directly after we still extend all children? What is that we are trying to solve?

tokejepsen (changes requested):
Something is off here. The PR to simplify (https://github.com/ynput/OpenPype/pull/5168) has not changed the code for this PR. Maybe its better to just copy/paste the code and commit?

```python
        for objset in objsets:
            objset = str(objset)
            members = cmds.sets(objset, query=True)
            members = cmds.ls(members, long=True)
            if members is None:
                self.log.warning("Skipped empty instance: \"%s\" " % objset)
                continue
                children = get_all_children(members)
                instance.data["contentMembers"] = children + members
                self.log.debug(
                    "content members: {}".format(
                        instance.data["contentMembers"]
                    )
                )
            elif objset.endswith("proxy_SET"):
                children = get_all_children(cmds.ls(members, long=True))
                instance.data["proxy"] = children + members
                self.log.debug(
                    "proxy members: {}".format(instance.data["proxy"])
                )
```
Should reduce the files changed to just `collect_arnold_scene_source.py`.

openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
moonyuet: It seems that the updated version doesn't show up. it should look way much simpler than this.

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
```diff
@@ -69,7 +69,8 @@ def process(self, instance):
             "camera": instance.data["camera"],
             "mask": mask
         }
-
+        # TODO: dont use instance.data["contentMembers"]
+        # but use the new instance.data["contentMemberTransforms"]
```
tokejepsen, line 73:
```suggestion
```

openpype/hosts/maya/plugins/publish/extract_arnold_scene_source.py
```diff
@@ -109,6 +110,8 @@ def process(self, instance):
             return
 
         kwargs["filename"] = file_path.replace(".ass", "_proxy.ass")
+        # TODO: dont use instance.data["proxy"]
+        # but use the new instance.data["proxyTransforms"]
```
tokejepsen, line 114:
```suggestion
```

tokejepsen (changes requested):

tokejepsen:
@BigRoy you happy with these changes?

openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
```diff
@@ -18,18 +18,24 @@ def process(self, instance):
...
+                self.log.debug(
+                    "content members: {}".format(
+                        instance.data["contentMembers"]
+                    )
+                )
             elif objset.endswith("proxy_SET"):
-                set_members = get_all_children(cmds.ls(members, long=True))
-                instance.data["proxy"] = set_members
-                self.log.debug("proxy members: {}".format(set_members))
+                children = get_all_children(cmds.ls(members, long=True))
```
BigRoy, line 34: Members already consists of long paths.
```suggestion
                children = get_all_children(members)
```

BigRoy: Also be aware that the if the `members` also contains one of its own children then `children` will also contain that node. As such `members + children` can result in duplicate entries for the same node. It might be useful to still ensure they are unique afterwards if that's a case we might hit.

Additionally, the logging of the members (which was already there) can now be a much much larger list since it also includes ALL children which can make the logging heavy and slow. We might want to avoid the logging by default **or** like before only log the direct set members instead?

moonyuet: not sure if it is quite cliche to do but maybe we can have a function to store the data and use it in the main function.
```
def get_ass_data(objset, suffix):
    if objset.endswith(suffix):
        members = cmds.ls(members, long=True)
        if not members:
            return
        children = get_all_children(members)
        return children

```

tokejepsen (approved):

tokejepsen:
@BigRoy you happy?

BigRoy (approved):
Code seems readable and fine with me. I'm personally not a fan of the verbose logging but it's not a show stopper. Didn't test the code so will leave that to others.

I do want to point out that I'm **not** so sure what about the code changes fixes a bug. It seems to do pretty much the same as before?

openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
```diff
@@ -18,18 +18,14 @@ def process(self, instance):
...
-                members = cmds.ls(members, long=True)
-                children = get_all_children(members)
-                instance.data["contentMembers"] = children
-                self.log.debug("content members: {}".format(children))
-            elif objset.endswith("proxy_SET"):
-                set_members = get_all_children(cmds.ls(members, long=True))
-                instance.data["proxy"] = set_members
-                self.log.debug("proxy members: {}".format(set_members))
+                instance.data["contentMembers"] = self.get_hierarchy(members)
+            if objset.endswith("content_SET"):
```
tokejepsen, line 27:
```suggestion
            if objset.endswith("proxy_SET"):
```

tokejepsen (changes requested):

openpype/hosts/maya/plugins/publish/collect_arnold_scene_source.py
BigRoy: Well spotted! :)

tokejepsen: Flagged by testing :) I spot nothing.

tokejepsen (approved):

antirotor (approved):

PR #5152, opened 2023-06-16
Add collector to tray publisher for getting frame range data

### LINKED ISSUE #5136: Bug: missing time information on publishes from Tray Publisher

### Current Behavior:

Published instance no longer contain time data by default. This means that any family published with Simple Creators will miss it too, in effect breaking loading logic in loaders. For example:

Publishing `Plate` from Tray Publisher will publish stuff with missing frame range information on version. This version is then displayed incorrectly in the Loader and Clip Loader in Nuke will crash.

### Expected Behavior:

Tray Publisher - its collector - should add frame range information if it is publishing sequences. If files are list in representation, collect it using clique and set it. It is much more difficult to handle video files - in their case, it should either use something like ffprobe to get correct frame range (unreliable) or infer it from asset data (not precise enough).

### Version

3.15.9

Windows

### Steps To Reproduce:

1. Open Tray Publisher
2. Select Project
3. Select your context
4. Select `Plate` as family
5. Drop some image sequence
6. Publish
7. Open Nuke
8. In Loader, you won't see any time information
9. `Load as Clip` will crash

### Relevant log output:

```shell
  File "C:\Users\annat\Documents\Projects\OpenPype\sources\op-py39\openpype\tools\loader\widgets.py", line 1590, in _load_representations_by_loader
    load_with_repre_context(
  File "C:\Users\annat\Documents\Projects\OpenPype\sources\op-py39\openpype\pipeline\load\utils.py", line 308, in load_with_repre_context
  File "C:\Users\annat\Documents\Projects\OpenPype\sources\op-py39\openpype\hosts\nuke\plugins\load\load_clip.py", line 125, in load
    first -= self.handle_start
TypeError: unsupported operand type(s) for -=: 'NoneType' and 'int'
```

### Additional context:

_No response_

moonyuet:
## Changelog Description
Add collector to tray publisher to get frame range data. User can choose to enable this collector if they need this in the publisher.
Resolve #5136

## Additional info
n/a

## Testing notes:
1. Launch Openpype
2. Click Publisher...
3. Select Plate
4. Add Image Sequence
5. Publish
6. Launch Nuke via Openpype
7. Openpype -> Load...
8. plateMain -> Load Clip


jakubjezek001:
Task linked: [OP-6163 Bug: missing time information on publishes from Tray Publisher](https://app.clickup.com/t/6658547/OP-6163)

openpype/plugins/publish/collect_anatomy_frame_range.py
```diff
@@ -0,0 +1,36 @@
...
+        asset_data = asset_doc["data"]
+        for key in (
+            "fps",
+            "frameStart",
+            "frameEnd",
+            "handleStart",
+            "handleEnd"
+        ):
+            if key not in instance.data and key in asset_data:
+                value = asset_data[key]
```
Minkiu, line 33: We could assign the value directly on the following line instead of `value`

Minkiu (approved):

openpype/plugins/publish/collect_anatomy_frame_range.py
```diff
@@ -0,0 +1,36 @@
...
+                "vdbcache", "online",
+                "render"]
+    hosts = ["traypublisher"]
+
+    def process(self, instance):
+        self.log.info("Collecting Anatomy frame range.")
+        asset_doc = instance.data.get("assetEntity")
+        if not asset_doc:
+            self.log.info("Missing required data..")
+            return
```
BigRoy, line 22: I suspect these should be `debug` logs?
```suggestion
        self.log.debug("Collecting Anatomy frame range.")
        asset_doc = instance.data.get("assetEntity")
        if not asset_doc:
            self.log.debug("Missing required data..")
            return
```

openpype/plugins/publish/collect_anatomy_frame_range.py
```diff
@@ -0,0 +1,36 @@
...
+            "frameStart",
+            "frameEnd",
+            "handleStart",
+            "handleEnd"
+        ):
+            if key not in instance.data and key in asset_data:
+                value = asset_data[key]
+                instance.data[key] = value
+
+        self.log.info("Anatomy frame range collection finished.")
```
BigRoy, line 36: Debug logs
```suggestion
        self.log.debug("Anatomy frame range collection finished.")
```

openpype/plugins/publish/collect_anatomy_frame_range.py
```diff
@@ -0,0 +1,36 @@
...
+        for key in (
+            "fps",
+            "frameStart",
+            "frameEnd",
+            "handleStart",
+            "handleEnd"
+        ):
+            if key not in instance.data and key in asset_data:
+                value = asset_data[key]
+                instance.data[key] = value
```
BigRoy, line 34: Agree with @Minkiu here.
```suggestion
            if key not in instance.data and key in asset_data:
                instance.data[key] = asset_data[key]
```

openpype/plugins/publish/collect_anatomy_frame_range.py
BigRoy: Instead it'd would have been much clearer if instead we would track what keys were set in e.g. a `keys_set` variable and then logged what data was set:
```python
if keys_set:
    self.log.debug(
        "Set instance frame range data from asset"
        "entity for keys: {}".format(", ".join(keys_set)
    )
```
So that the logs actually provide debugging information of what really occurred.

openpype/plugins/publish/collect_anatomy_frame_range.py
```diff
@@ -0,0 +1,39 @@
...
+
+    order = pyblish.api.CollectorOrder + 0.491
+    label = "Collect Anatomy Frame Range"
+    families = ["plate", "pointcache",
+                "vdbcache", "online",
+                "render"]
+    hosts = ["traypublisher"]
+
+    def process(self, instance):
+        self.log.debug("Collecting Anatomy frame range.")
```
BigRoy, line 18:
```suggestion
```

openpype/plugins/publish/collect_anatomy_frame_range.py
```diff
@@ -0,0 +1,38 @@
+import pyblish.api
+
+
+class CollectAnatomyFrameRange(pyblish.api.InstancePlugin):
+    """Collect Frame Range specific Anatomy data.
+
+    Plugin is running for all instances on context even not active instances.
+    """
```
BigRoy, line 8: This isn't true right? It doesn't run on inactive instances since it's currently an `InstancePlugin`?

```suggestion
    """Collect Frame Range specific Anatomy data."""
```

moonyuet: This is started from stratch but anyways i have repharsed this docstring in the latest commit

BigRoy: It's still incorrect. It doesn't run over inactive ones, right? And if so, I don't see a reason why we'd want to specify it either.

Note that there's currently a double space in front of **data** in the docstring.
Also, the plugin name and file still refer to it as "Anatomy data" so I'd refactor those too. (and thus with that the settings as well)

openpype/plugins/publish/collect_frame_range_data.py
```diff
@@ -0,0 +1,36 @@
+import pyblish.api
+
+
+class CollectAnatomyFrameRange(pyblish.api.InstancePlugin):
```
BigRoy, line 4:
```suggestion
class CollectFrameRangeData(pyblish.api.InstancePlugin):
```

antirotor (changes requested):
Found one more instance the `Anatomy` word in log messages :).

But I have more fundemantal question: shouldn't this set `frameStart` and `frameEnd` actually from the representation itself and fallback to asset data only when this cannot be determined?

Because right know, when I drop sequence 1001-1050 to Tray Publisher and I'll publish it as a `plate` on asset with 1001-1100 frame range, it will get the `frameStart` and `frameEnd` wrong?

I think that when there is list of files, we currently consider it as file sequence and if it is file sequence, we can get start and end frame.

If it is single video file, it gets more difficult but maybe in that case we could fallback to asset data (or use ffprobe and if that fails we fall back to asset data)... What do you think @BigRoy @Minkiu

openpype/plugins/publish/collect_frame_range_data.py
```diff
@@ -0,0 +1,72 @@
+import pyblish.api
+import clique
+
+
+class CollectFrameRangeData(pyblish.api.InstancePlugin):
```
fabiaserra, line 5: could you align it with the existing `collect_review_frames` plugin?

openpype/plugins/publish/collect_frame_range_data.py
```diff
@@ -0,0 +1,67 @@
...
+
+        else:
+            self.log.info("No representation data.. "
+                          "\nUse Asset Entity data instead")
+            asset_doc = instance.data.get("assetEntity")
+            if instance.data.get("frameStart") is not None or not asset_doc:
+                self.log.debug("Instance has no asset entity set."
+                               " Skipping collecting frame range data.")
+                return
+            asset_data = asset_doc["data"]
```
BigRoy, line 52: I think we might want to put this fallback behavior behind a setting - I would want my image sequences to get their frame range collected (the if-branhc). But having an Alembic file suddenly getting the frame range of the asset doesn't always feel right to me (especially at the times that it's basically wrong). Maybe this fallback could be toggled by the artist in the publisher, like "Set asset frame range if no frame range detected in source files" or allowing an admin to enable/disable this fallback for specific extensions or alike?

openpype/plugins/publish/collect_frame_range_data.py
```diff
@@ -0,0 +1,67 @@
...
+
+            files = first_repre["files"]
+            repres_file = clique.assemble(
+                files, minimum_items=1)[0][0]
+            repres_frames = [frames for frames in repres_file.indexes]
+            last_frame = len(repres_frames) - 1
+            asset_data = {
+                "frameStart": repres_frames[0],
+                "frameEnd": repres_frames[last_frame],
+            }
```
BigRoy, line 42: It might be crucial to note that this - unlike the `asset` else branch does **NOT** set the FPS + handleStart and handleEnd values. I understand those can't basically be detected automatically - but just wanted to point that out.

openpype/plugins/publish/collect_frame_range_data.py
moonyuet: I agree that some specific files like those in the pointcache family are not quite logical for using asset_doc data. 
Maybe we can add settings to eanble/disable the fallback for pointcache family and vdb cache (or any non-image type) . @BigRoy

BigRoy: I prefer explicit over implicit so on our end we'd likely disable this always. I guess the exact implementation when to "fallback to asset data" would be up to others?

openpype/plugins/publish/collect_frame_range_data.py
```diff
@@ -0,0 +1,85 @@
...
+            if ext not in IMAGE_EXTENSIONS:
+                self.log.info("Collecting frame range data"
+                              " only supported for image extensions")
+                return
+
+            files = first_repre["files"]
+            repres_files, remainder = clique.assemble(files)
+            repres_frames = list()
+            for repres_file in repres_files:
+                repres_frames = list(repres_file.indexes)
```
BigRoy, line 39:
```suggestion
            collections, remainder = clique.assemble(files)
            if not collections:
                # No sequences detected so we can't retrieve
                # frame range
                return
                
            # Let's assume we only ever find a single collection
            # or we should error out otherwise? (or take min-max of ranges?)
            collection = collections[0]
            repre_frames = list(collection.indexes)
```

openpype/plugins/publish/collect_frame_range_data.py
```diff
@@ -0,0 +1,85 @@
...
+            asset_data = {
+                "frameStart": repres_frames[0],
+                "frameEnd": repres_frames[-1],
+            }
+
+        else:
+            self.log.info(
+                "No representation data.. Use Asset Entity data instead")
```
BigRoy, line 47:
```suggestion
            self.log.info(
                "No representation data. Using Asset Entity data instead")
```
No need for double dot.
Also changed to "Using" otherwise it sounds like something the user should be doing.

openpype/plugins/publish/collect_frame_range_data.py
```diff
@@ -0,0 +1,85 @@
...
+
+            attr_values = self.get_attr_values_from_data(instance.data)
+            if attr_values.get("setAssetFrameRange", True):
+                if instance.data.get("frameStart") is not None or not asset_doc:            # noqa
+                    self.log.debug("Instance has no asset entity set."
+                                   " Skipping collecting frame range data.")
+                    return
+                self.log.debug(
+                    "Falling back to collect frame range"
+                    " data from asset entity set.")
```
BigRoy, line 58:
```suggestion
                self.log.debug(
                    "Falling back to collect frame range"
                    " data from asset entity.")
```
This seems better grammar was.
If we want to include that it's the one that is set than it's likely better as:

```suggestion
                self.log.debug(
                    "Falling back to collect frame range"
                    " data from set asset entity.")
```
But I still find the `set` just confusing in there too since it COULD point to a collection of things definition wise.

openpype/plugins/publish/collect_frame_range_data.py
```diff
@@ -0,0 +1,85 @@
...
+        self.log.debug(f"Frame range data {key_sets} "
+                       "has been collected from asset entity.")
+
+    @classmethod
+    def get_attribute_defs(cls):
+        return [
+            BoolDef("setAssetFrameRange",
+                    label="Set Asset Frame Range",
+                    default=False),
+        ]
```
BigRoy, line 85: It must be much clearer that it ONLY does so if the file sequence has no frame range detected. (only for image formats).

Thinking about it now, do we even care whether it's an image format actually? The sequence frame range should also be computed for e.g. a VDB sequence, yes?

openpype/plugins/publish/collect_frame_range_data.py
```diff
@@ -0,0 +1,85 @@
...
+            first_repre = repres[0]
+            ext = ".{}".format(first_repre["ext"])
+            if "ext" not in first_repre:
+                self.log.warning(f"Cannot find file extension"
+                                 " in representation data")
+                return
+            if ext not in IMAGE_EXTENSIONS:
+                self.log.info("Collecting frame range data"
+                              " only supported for image extensions")
+                return
```
BigRoy, line 33: As stated [here](https://github.com/ynput/OpenPype/pull/5152/files#r1253309182) I don't think we care whether it's an image or not. E.g. OpenVDB caches should also be fine as sequences?

moonyuet: yes, i think we can just remove this and collect frame range data from file extension other than image ext.

openpype/plugins/publish/collect_frame_range_data.py
```diff
@@ -0,0 +1,85 @@
...
+    order = pyblish.api.CollectorOrder + 0.491
+    label = "Collect Frame Range Data"
+    families = ["plate", "pointcache",
+                "vdbcache", "online",
+                "render"]
+    hosts = ["traypublisher"]
+
+    def process(self, instance):
+        repres = instance.data.get("representations")
+        asset_data = None
```
BigRoy, line 21: The name `asset_data` is confusing. Because for the case of the image sequences it's  not actually the asset data. Can we make it `collected_data` or `frame_data` or whatever so it's clear that it's not the data from the asset entity itself.

openpype/plugins/publish/collect_frame_range_data.py
```diff
@@ -0,0 +1,85 @@
...
+        self.log.debug(f"Frame range data {key_sets} "
+                       "has been collected from asset entity.")
+
+    @classmethod
+    def get_attribute_defs(cls):
+        return [
+            BoolDef("setAssetFrameRange",
+                    label="Set Asset Frame Range",
+                    default=False),
+        ]
```
BigRoy, line 85:
```suggestion
        return [
            BoolDef("setAssetFrameRange",
                    label="Set Asset Frame Range",
                    tooltip="When enabled and the input is not a sequence with frame "
                                "numbers then fall back to setting the asset frame range."
                    default=False),
        ]
```
Adding a tooltip might help.

openpype/plugins/publish/collect_frame_range_data.py
```diff
@@ -0,0 +1,85 @@
...
+            self.log.info(
+                "No representation data. Using Asset Entity data instead")
+            asset_doc = instance.data.get("assetEntity")
+
+            attr_values = self.get_attr_values_from_data(instance.data)
+            if attr_values.get("setAssetFrameRange", True):
+                if instance.data.get("frameStart") is not None or not asset_doc:            # noqa
+                    self.log.debug("Instance has no asset entity set."
+                                   " Skipping collecting frame range data.")
+                    return
```
BigRoy, line 55: This logged message will be confusing for the cases where `frameStart` is None but an asset document is available. I'd say we should either just check the `asset_doc` and assume `frameStart` will be available in instance data or validate separately with a clearer logged message?

openpype/plugins/publish/collect_frame_range_data.py
```diff
@@ -0,0 +1,85 @@
...
+            "frameEnd",
+            "handleStart",
+            "handleEnd"
+        ):
+            if key not in instance.data and key in asset_data:
+                instance.data[key] = asset_data[key]
+                key_sets.append(key)
+
+        self.log.debug(f"Frame range data {key_sets} "
+                       "has been collected from asset entity.")
```
BigRoy, line 77:
```suggestion
        keys_set = []
        for key in (
            "fps",
            "frameStart",
            "frameEnd",
            "handleStart",
            "handleEnd"
        ):
            if key not in instance.data and key in asset_data:
                instance.data[key] = asset_data[key]
                keys_set.append(key)

        if keys_set:
            self.log.debug(f"Frame range data {keys_set} "
                           "has been set.")
```

This shouldn't log when no keys are set.
This isn't always setting them from asset entity, but sometimes from the sequence - so logged message can be clearer.

BigRoy (commented):
Sorry for yet again more comments - but it still feels off. I'm still wondering whether it might just make more sense to do this instead:


First try and set the data from the sequence
```
class CollectSequenceFrameData(pyblish.api.InstancePlugin):
    order = pyblish.api.CollectorOrder + 0.2

    def process(self, instance):
        frame_data = get_frame_data_from_repre_sequence(instance)
        if frame_data:
            for key, value in frame_data.items():
                if key not in instance.data:
                    instance.data[key] = value
```
Then later we have a collector that IF no frame data is available it'll allow to set it from the asset.

```python
class CollectFrameDataFromAssetEntity(pyblish.api.InstancePlugin):
    order = pyblish.api.CollectorOrder + 0.3

    def process(self, instance):
    
        # Check if instance.data is still missing any frame
        # data keys. If so, we'll look for them on the asset entity
        keys_missing = []
        for key in (
            "fps",
            "frameStart",
            "frameEnd",
            "handleStart",
            "handleEnd"
        ):
            if key not in instance.data:
                keys_missing.append(key)

        if not keys_missing:
            # All data already exists on instance
            return
        
        # Transfer frame data from asset entity
        keys_set = []
        for key in keys_missing:
            asset_data = instance.data["assetEntity"]["data"]
            if key in asset_data:
                instance.data[key] = asset_data[key]
                keys_set.append(key)
            
        if keys_set:
            self.log.debug(f"Frame range data {keys_set} "
                           "has been copied from asset entity.")
```

By separating them I feel like they become much clearer functionally as to what they do. Plus it separates the odd mixture of if not sequence frames detected then retrieve from the asset data.

In my example plugin case above the `CollectFrameDataFromAssetEntity` plugin could just be optional to decide whether or not to retrieve data for the instance or not.

The logging also becomes much clearer since it becomes clear exactly where it's taking the frame data from too.

moonyuet:
> Sorry for yet again more comments - but it still feels off.

Actually the way you suggested can be more "accurate" to grab the frame range data depending on user's preference. In the latest commit I did is just grabbing frameStart and frameEnd from the representation data. And we need to include fps or any other useful data such as handleStart and handleEnd for image sequences. sometimes the representation data would have included these data(but there is no guarantee they would be included).

openpype/plugins/publish/collect_frame_range_asset_entity.py
```diff
@@ -0,0 +1,39 @@
...
+                missing_keys.append(key)
+        key_sets = []
+        for key in missing_keys:
+            asset_data = instance.data["assetEntity"]["data"]
+            if key in asset_data:
+                instance.data[key] = asset_data[key]
+                key_sets.append(key)
+        if key_sets:
+            self.log.debug(f"Frame range data {key_sets} "
+                           "has been collected from asset entity.")
```
BigRoy, line 39:
```suggestion
        keys_set = []
        for key in missing_keys:
            asset_data = instance.data["assetEntity"]["data"]
            if key in asset_data:
                instance.data[key] = asset_data[key]
                keys_set.append(key)
        if keys_set:
            self.log.debug(f"Frame range data {keys_set} "
                           "has been collected from asset entity.")
```
I personally find **keys_set** clearer than **key_sets**.

openpype/plugins/publish/collect_sequence_frame_data.py
```diff
@@ -0,0 +1,47 @@
...
+
+    order = pyblish.api.CollectorOrder + 0.2
+    label = "Collect Sequence Frame Data"
+    families = ["plate", "pointcache",
+                "vdbcache", "online",
+                "render"]
+    hosts = ["traypublisher"]
+
+    def process(self, instance):
+        frame_data = self.get_frame_data_from_repre_sequence(instance)
```
BigRoy, line 17: Looking at your function implementation below this can return `None`

```suggestion
        frame_data = self.get_frame_data_from_repre_sequence(instance)
        if not frame_data:
            return
```

openpype/plugins/publish/collect_sequence_frame_data.py
```diff
@@ -0,0 +1,47 @@
...
+        for key, value in frame_data.items():
+            if key not in instance.data:
+                instance.data[key] = value
+                self.log.debug(f"Frame range data {key} has been collected ")
```
BigRoy, line 21: Since we are logging per key, would it be nice to also report the value?

```suggestion
                self.log.debug(f"Collected sequence frame range data '{key}': {value}")
```

BigRoy (commented):
This does look much clearer to me! ❤️

BigRoy (approved):
Didn't test but code looks good.

Regarding the testing notes:
> Openpype -> Load...
> plateMain -> Load Clip

I'd say that if that errors on no time data that it's maybe also a bug in the loader itself. It seems that apparently there is definitely a chance clips/media come in without frame data so it should be prepared to even allow loading that instead.

moonyuet:
@antirotor 's comment has been resolved and it can be tested and merged?

antirotor (approved):

PR #5157, opened 2023-06-19
Documentation project settings Houdini

SherineKB:
## Changelog Description
Documentation of the Houdini Project Settings.


.github/workflows/sync_fork.yml
BigRoy, line 1: File should not be part of this commit.

BigRoy (commented):
As @mkolar mentioned [here](https://github.com/ynput/OpenPype/pull/5116#issuecomment-1596651581) I suspect these documentation PRs to be a much better effort against the new documentation repo because by applying them here first we're doubling the work - so while you can, try to make the edits there if that's not an issue for you.

Other than that, of course nice to see work being done on the documentation to begin with! Great job.

PR #5158, opened 2023-06-19
Document project settings Photoshop

SherineKB:
## Changelog Description
Photoshop project settings' documentation.

### DIFF
--- a/.github/workflows/sync_fork.yml
+++ b/.github/workflows/sync_fork.yml
@@ -1,0 +1,20 @@
+name: 🔄 Sync Fork
+
+on:
+  schedule:
+    - cron: '0 0 * * 1-5' # Monday to friday at 00:00
+  workflow_dispatch:
+
+jobs:
+  sync:
+
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/sync_fork.yml
```diff
@@ -0,0 +1,20 @@
+name: 🔄 Sync Fork
```
kalisp, line 1: I don't think we need this file.

openpype/lib/project_backpack.py
```diff
@@ -125,6 +125,7 @@ def pack_project(
     if not only_documents:
         roots = project_doc["config"]["roots"]
         # Determine root directory of project
+        source_root = None
```
kalisp, line 128: I don't think we need this in this PR either.

kalisp (changes requested):

kalisp:
Actually let me internally discuss purpose and future of `https://openpype.io/docs/project_settings`.

Photoshop's plugins Settings are now described in https://openpype.io/docs/admin_hosts_photoshop . Any modifications there welcomed.

PR #5164, opened 2023-06-20
3dsmax: make code compatible with 3dsmax 2022

antirotor:
## Changelog Description
Python 3.7 in 3dsmax 2022 is not supporting walrus operator. This is removing it from the code for the sake of compatibility

## Testing notes:
3dsmax functionality should work the same in 2022 and 2023


moonyuet (approved):
code looks good to me. haven't done test yet.

LiborBatek (approved):
OP does start normally and multiple asset types publishes ok without any issues.

Rendering on DL also been tested successfully.

So all working as in max 2023 and no issues were found.

PR #5165, opened 2023-06-20
Publisher: Edge case fixes

iLLiCiTiT:
## Changelog Description
Fix few edge case issues that may cause issues in Publisher UI.

## Additional info
Conversion of exception to string may cause crashes, that lead to invalid state of publisher UI without any lead -> At least show message that the exception message couldn't be converted. Missing instance label caused crashes of whole publisher UI on cryptic error -> Use default label `No label` if the source label is `None`.

## Testing notes:
Both are random edge cases, usually happening because of wrong implementation.
### Exception
- Prepare custom exception that would crash during string conversion, and raise it during publishing.
```python
class MyException(Exception):
    def __str__(self):
        self._template.format(**self._data)
```
- Report and Details page should cause error.

### Instance without label
- Prepare creator that won't fill subset and label on new instance.
- In UI should be instance with `No label` label

### DIFF
--- a/openpype/tools/publisher/control.py
+++ b/openpype/tools/publisher/control.py
@@ -398,12 +398,22 @@ class PublishReportMaker:
         exception = result.get("error")
         if exception:
             fname, line_no, func, exc = exception.traceback
+
+            # Conversion of exception into string may crash
+            try:
+                msg = str(exception)
+            except BaseException:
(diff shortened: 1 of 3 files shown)


Minkiu (approved):
Haven't tested it, but looks good.

PR #5170, opened 2023-06-21
Unreal: get current project settings not using unreal project name

antirotor:
## Changelog Description
There was a bug where Unreal project name was used to query project settings. But Unreal project name can differ from the "real" one because of naming convention rules set by Unreal. This is fixing it by asking for current project settings.

## Testing notes:
Creatiing new Unreal project via AYON with different name from the main project should work as expected

### DIFF
--- a/openpype/hosts/unreal/hooks/pre_workfile_preparation.py
+++ b/openpype/hosts/unreal/hooks/pre_workfile_preparation.py
@@ -34,6 +34,7 @@ class UnrealPrelaunchHook(PreLaunchHook):
         super().__init__(*args, **kwargs)
 
         self.signature = f"( {self.__class__.__name__} )"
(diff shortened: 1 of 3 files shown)


openpype/hosts/unreal/hooks/pre_workfile_preparation.py
```diff
@@ -111,6 +112,7 @@ def exec_ue_project_gen(self,
         ue_project_worker = UEProjectGenerationWorker()
         ue_project_worker.setup(
             engine_version,
+            self.project_name,
```
iLLiCiTiT, line 115: You don't need to store it to attribute of the hook.
```suggestion
            self.data["project_name"],
```

iLLiCiTiT (commented):
LGTM

PR #5174, opened 2023-06-22
Maya: import workfile missing - OP-6233

tokejepsen:
## Changelog Description
Missing `workfile` family to import.

## Testing notes:
1. Publish a workfile.
2. Import the workfile through the Loader.

### DIFF
--- a/openpype/hosts/maya/plugins/load/actions.py
+++ b/openpype/hosts/maya/plugins/load/actions.py
@@ -93,21 +93,8 @@ class ImportMayaLoader(load.LoaderPlugin):
 
     """
     representations = ["ma", "mb", "obj"]
-    families = [
-        "model",
-        "pointcache",
-        "proxyAbc",
-        "animation",
-        "mayaAscii",
-        "mayaScene",
-        "setdress",
-        "layout",
-        "camera",
-        "rig",
-        "camerarig",
-        "staticMesh",
-        "workfile"
-    ]
+    families = ["*"]
+    excluded_families = ["xgen"]
 
     label = "Import"
     order = 10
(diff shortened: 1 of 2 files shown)


LiborBatek (approved):
All works as expected.

There is now ```Import workfile``` action available in the ```Loader``` and fully functional.

iLLiCiTiT:
I would rather avoid adding `exclude_families` into default filtering options of loader. @antirotor ideas?

tokejepsen:
> I would rather avoid adding `exclude_families` into default filtering options of loader. @antirotor ideas?

An alternative I thought about was to have some sort of exclusion syntax, like `!xgen`.

iLLiCiTiT:
There reversed filtering is breaking https://github.com/ynput/OpenPype/pull/4928 . Also I'm not sure if the action should be visible for all families except `xgen`, it will show for `image`, `render` or custom families that studio defined for their purposes.

tokejepsen:
Up to @m-u-r-p-h-y @iLLiCiTiT @antirotor to decide :)

BigRoy:
> We can do explicit families like in https://github.com/ynput/OpenPype/commit/43ece88f02bf1b35b22d35d3ceeef82241fe2ef3

I'd do that yes, and then discuss other changes like supporting `exclude_families` in a dedicated issue or on the forum. For now it's easiest to match all other plugins and just keep this PR about what it actually intended to solve?

mkolar:
> > We can do explicit families like in [43ece88](https://github.com/ynput/OpenPype/commit/43ece88f02bf1b35b22d35d3ceeef82241fe2ef3)
> 
> I'd do that yes, and then discuss other changes like supporting `exclude_families` in a dedicated issue or on the forum. For now it's easiest to match all other plugins and just keep this PR about what it actually intended to solve?

Agreed 100%

PR #5177, opened 2023-06-22
Webpublisher - headless publish shouldn't be blocking operation

kalisp:
## Changelog Description
`subprocess.call` was blocking, which resulted in UI non responsiveness as it was waiting for publish to finish.


iLLiCiTiT (approved):
Code is LGTM. Only future us will thank or hate us.

PR #5184, opened 2023-06-23
Publisher: Fix disappearing actions

iLLiCiTiT:
## Changelog Description
Pyblish plugin actions are visible as expected.

## Additional info
Action may not be visible in validation UI if the same action class was reused in multiple publish plugins. Store action ids by plugin ids to make sure they are stored in a different manner and store all action ids per plugin even if would already be in actions mapping.

## Source issue
The source issue came from Maya new publisher refactor where multiple pyblish plugins were using `RepairAction` but only one of them showed up in UI.

## Testing notes:
1. Have more than one validation plugin that is using `RepairAction`
2. Make both crash with validation error
3. Both should show Repair action in UI

### DIFF
--- a/openpype/tools/publisher/control.py
+++ b/openpype/tools/publisher/control.py
@@ -497,7 +497,9 @@ class PublishPluginsProxy:
         """
 
         return [
-            self._create_action_item(self._actions_by_id[action_id], plugin_id)
+            self._create_action_item(
+                self.get_action(plugin_id, action_id), plugin_id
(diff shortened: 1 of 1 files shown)


Minkiu (approved):
Haven't tested, code looks good.

BigRoy (approved):
Unable to test here now - but code changes make sense.

antirotor (approved):
works now

PR #5001, opened 2023-05-22
Review Family in Max

moonyuet:
## Changelog Description
Review Feature by creating preview animation in 3dsmax
(The code is still cleaning up so there is going to be some updates until it is ready for review)

## Additional info
The review animation `rt.CreatePreview()` function in pymxs api is still very limited without any arguments to set up the playblast setting so `rt.execute` with maxscript will be used for playblasting the animation. 

## Testing notes:
1. Create Review with your camera
2. If your selection is not camera type, it wont pass the validator
3. Publish it

PR #5187, opened 2023-06-24
Fix the reset frame range not setting up the right timeline in Max

### LINKED ISSUE #5181: Bug: 3dsmax - OP Menu >"Set Frame Range" doesnt work in Max 2024

### Current Behavior:

When setting frame range for asset/shot via OP menu ```Set Frame Range``` it doesnt work and frame range stays as is.

image

### Expected Behavior:

```Set Frame Range``` should set scene frame range according to Database.

### Version

3.15.11-nightly.4

Windows

### Steps To Reproduce:

1. Check if the ```frame range``` is not same as in Database and change it so they differ
2. use OP Menu > ```Set Frame Range``` action to fix it
3. The scene still have original wrong frame range without any change

### Relevant log output:

_No response_

### Additional context:

_No response_

[cuID:[OP-6244](https://app.clickup.com/t/863h2hqt3)]

moonyuet:
## Changelog Description
Resolve #5181 

## Additional info
also resolve the timeline issue at #5001 

## Testing notes:
1. Launch Max via OP
2. Openpype... -> Set Frame Range
3. Should be setting up the right timeline regarding to the DB

### DIFF
--- a/openpype/hosts/max/api/lib.py
+++ b/openpype/hosts/max/api/lib.py
@@ -250,10 +250,7 @@ def reset_frame_range(fps: bool = True):
         frame_range["handleStart"]
     )
     frame_end_handle = frame_range["frameEnd"] + int(frame_range["handleEnd"])
-    frange_cmd = (
-        f"animationRange = interval {frame_start_handle} {frame_end_handle}"
-    )
-    rt.Execute(frange_cmd)
+    set_timeline(frame_start_handle, frame_end_handle)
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
The ```Set Frame Range``` function works now as expected.

Comparing frame ranges before and after and it corelates with DB values.

Minkiu (approved):

LiborBatek:
> Haven't tested, but code looks good.

I did test it  so can be merged I guess. :)

PR #5189, opened 2023-06-26
Houdini: enhance abc publishing

mustafa-zarkash:
## Changelog Description
1) On point cache creation in Houdini :  Create plugin will 
      -  if `output` node doesn't exist , create one

2) On Publishing: `Validate Prims Hierarchy Path` 
    - added a repair action
      - if  `path` attribute doesn't exist,  create it with  a default path value `"{parent_node_name}/{output_node_name}"`
      - it will use the same path attribute  name as in the `ROP alembic node` 
      - it skips checking if output node is not a `SOP Output node`
    > This doesn't enforce a certain path attribute value, artists are free to use any values they would preferer.

3) On Publishing: `Validate Output Node`
    - added a repair action if `sop_path` in  the `ROP alembic node` doesn't point to a `SOP Output node`

4) On Publishing: `Validate Primitive to Detail (Abc)`
     I added a condition to skip this validation if  `output node` is not a sop node 
     it cause validator not to error at this line `geo = output_node.geometryAtFrame(frame)`

### Case 1: create point cache 
create any object (e.g. a sphere)
create point cache while use selection toggle `ON`

it adds only an output node 
if artist adds no path attribute, then on publishing it will redirect to Case 2

### Case 2: ROP's sop_path is set to a sop output node with no path attribute 
These validators will error: 
- `Validate Primitive to Detail (Abc)`
- `Validate Prims Hierarchy Path`

Repair Actions Response: 
- `Validate Prims Hierarchy Path`
  - log: *DEBUG - 'path' attribute was not found!*
  - it creats a `name` node, change `attribute` to the ROP's `path_attrib` parameter and set a default path value
  - log: *DEBUG - 'AUTO_PATH' was created, It adds 'path' with a default single value*

### case 3: Empty ROP 
These validators will error: 
- `Validate Output Node`
- `Validate Primitive to Detail (Abc)`
- `Validate Input Node (Abc)`
- `Validate Prims Hierarchy Path`

Repair Actions Response : 
- `Validate Output Node` 
  - log: *DEBUG - This ROP {rop path} has empty SOP Path*
  - log: *DEBUG - Try Setting SOP path to selected node*
- `Validate Prims Hierarchy Path`
  - log: *WARNING - No geometry output node found, skipping Repair Action..*

### case 4: ROP's sop_path is set to an Obj node
These validators will error: 
- `Validate Output Node`

Repair Actions Response : 
- `Validate Output Node` 
  - log: *DEBUG - SOP Path points to an Obj node*
  - It creates an output node and connects it to the node with render flag
  - log: *DEBUG - SOP Output node created and set.*

### case 5: ROP's sop_path is set to a sop node and it isn't a SOP Output node
it can happens when sop_path points to any sop node that isn't sop output node
whether the sop output node exists or not 

These validators will error: 
- `Validate Output Node`
- `Validate Primitive to Detail (Abc)`
- `Validate Prims Hierarchy Path`

Repair Actions Response : 
- `Validate Output Node` 
  - log: *DEBUG - SOP Path points to an Sop node and it isn't a SOP Output node*
  - It creates an output node and connects it to the sop node in the ROP's `sop_path` parameter
  - log: *DEBUG - SOP Output node created and set.*
- `Validate Prims Hierarchy Path`
  - log: *WARNING - No geometry output node found, skipping Repair Action..*

### Case 6: ROP's sop_path is set to a non existed node
A more general case of case 5 

Repair Actions Response : 
- `Validate Output Node` 
  - log: *DEBUG - '{output node}' doesn't exist*
  - log: *DEBUG - Trying setting sop path to the parent '{parent node}'*
  - if it isn't valid
    - log: *DEBUG - '{parent node}' is not valid or it doesn't exist!*
    - log: *DEBUG - Resetting SOP Path of this ROP '{rop node}'*
      - redirect to case case 3: Empty ROP

### DIFF
--- a/openpype/hosts/houdini/plugins/create/create_pointcache.py
+++ b/openpype/hosts/houdini/plugins/create/create_pointcache.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 """Creator plugin for creating pointcache alembics."""
 from openpype.hosts.houdini.api import plugin
-# from openpype.pipeline import CreatedInstance
 
 import hou
 
@@ -58,20 +57,6 @@ class CreatePointCache(plugin.HoudiniCreator):
 
                 child_output.setDisplayFlag(1)
                 child_output.setRenderFlag(1)
-                child_output.moveToGoodPosition()
-
-            paths = child_output.geometry().findPrimAttrib("path") and \
-                      child_output.geometry().primStringAttribValues("path")
-
-            # Create default path value if missing
-            if not paths:
-                path_node = parent.createNode("name", "AUTO_PATH")
-                path_node.parm("attribname").set("path")
-                path_node.parm("name1").set('`opname("..")`/`opoutput(".", 0)`') # noqa
-
-                path_node.setFirstInput(child_output.input(0))
-                path_node.moveToGoodPosition()
-                child_output.setFirstInput(path_node)
                 child_output.moveToGoodPosition()
 
             parms["sop_path"] = child_output.path()
--- a/openpype/hosts/houdini/plugins/create/create_pointcache.py
+++ b/openpype/hosts/houdini/plugins/create/create_pointcache.py
@@ -45,15 +45,15 @@ class CreatePointCache(plugin.HoudiniCreator):
 
             # try to find output node
             for child in parent.children():
-                if child.isGenericFlagSet( hou.nodeFlag.Render):
+                if child.isGenericFlagSet(hou.nodeFlag.Render):
                     child_render = child
                 if child.type().name() == "output":
-                    child_output =  child
+                    child_output = child
                     break
 
             # create output node if not exists
             if not child_output:
-                child_output = parent.createNode("output","OUTPUT")
+                child_output = parent.createNode("output", "OUTPUT")
                 child_output.setFirstInput(child_render)
 
                 child_output.setDisplayFlag(1)
@@ -64,10 +64,10 @@ class CreatePointCache(plugin.HoudiniCreator):
                       child_output.geometry().primStringAttribValues("path")
 
             # Create default path value if missing
-            if not paths :
+            if not paths:
                 path_node = parent.createNode("name", "AUTO_PATH")
                 path_node.parm("attribname").set("path")
-                path_node.parm("name1").set('`opname("..")`/`opoutput(".", 0)`')
+                path_node.parm("name1").set('`opname("..")`/`opoutput(".", 0)`') #noqa
 
                 path_node.setFirstInput(child_output.input(0))
                 path_node.moveToGoodPosition()
--- a/openpype/hosts/houdini/plugins/create/create_pointcache.py
+++ b/openpype/hosts/houdini/plugins/create/create_pointcache.py
@@ -67,7 +67,7 @@ class CreatePointCache(plugin.HoudiniCreator):
             if not paths:
                 path_node = parent.createNode("name", "AUTO_PATH")
                 path_node.parm("attribname").set("path")
-                path_node.parm("name1").set('`opname("..")`/`opoutput(".", 0)`') #noqa
+                path_node.parm("name1").set('`opname("..")`/`opoutput(".", 0)`') # noqa
 
                 path_node.setFirstInput(child_output.input(0))
                 path_node.moveToGoodPosition()
--- a/openpype/hosts/houdini/plugins/create/create_pointcache.py
+++ b/openpype/hosts/houdini/plugins/create/create_pointcache.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 """Creator plugin for creating pointcache alembics."""
 from openpype.hosts.houdini.api import plugin
-from openpype.pipeline import CreatedInstance
+# from openpype.pipeline import CreatedInstance
 
 import hou
 
@@ -37,14 +37,44 @@ class CreatePointCache(plugin.HoudiniCreator):
         }
 
         if self.selected_nodes:
-            parms["sop_path"] = self.selected_nodes[0].path()
+            parent = self.selected_nodes[0]
+            parms["sop_path"] = parent.path()
+
+            child_render = ""
+            child_output = ""
 
             # try to find output node
-            for child in self.selected_nodes[0].children():
+            for child in parent.children():
+                if child.isGenericFlagSet( hou.nodeFlag.Render):
+                    child_render = child
                 if child.type().name() == "output":
-                    parms["sop_path"] = child.path()
+                    child_output =  child
                     break
 
+            # create output node if not exists
+            if not child_output:
+                child_output = parent.createNode("output","OUTPUT")
+                child_output.setFirstInput(child_render)
+
+                child_output.setDisplayFlag(1)
+                child_output.setRenderFlag(1)
+                child_output.moveToGoodPosition()
+
+            paths = child_output.geometry().findPrimAttrib("path") and \
+                      child_output.geometry().primStringAttribValues("path")
+
+            # Create default path value if missing
+            if not paths :
+                path_node = parent.createNode("name", "AUTO_PATH")
+                path_node.parm("attribname").set("path")
+                path_node.parm("name1").set('`opname("..")`/`opoutput(".", 0)`')
+
+                path_node.setFirstInput(child_output.input(0))
+                path_node.moveToGoodPosition()
+                child_output.setFirstInput(path_node)
+                child_output.moveToGoodPosition()
+
+            parms["sop_path"] = child_output.path()
         instance_node.setParms(parms)
         instance_node.parm("trange").set(1)
 
--- a/openpype/hosts/houdini/plugins/publish/validate_abc_primitive_to_detail.py
+++ b/openpype/hosts/houdini/plugins/publish/validate_abc_primitive_to_detail.py
@@ -73,6 +73,13 @@ class ValidateAbcPrimitiveToDetail(pyblish.api.InstancePlugin):
         cls.log.debug("Checking Primitive to Detail pattern: %s" % pattern)
         cls.log.debug("Checking with path attribute: %s" % path_attr)
 
+        if not isinstance(output_node, hou.SopNode):
+            # In the case someone has explicitly set an Object
+            # node instead of a SOP node in Geometry context
+            # then for now we ignore.
+            cls.log.warning("No geometry output node found, skipping check..")
+            return
+
         # Check if the primitive attribute exists
         frame = instance.data.get("frameStart", 0)
         geo = output_node.geometryAtFrame(frame)
--- a/openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
+++ b/openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
@@ -119,9 +119,9 @@ class ValidatePrimitiveHierarchyPaths(pyblish.api.InstancePlugin):
 
         path_attr = rop_node.parm("path_attrib").eval() or "path"
         paths = geo.findPrimAttrib(path_attr) and \
-            geo.primStringAttribValues(path_attr)
+                    geo.primStringAttribValues(path_attr)
 
-        if not paths :
+        if not paths:
             output_node = instance.data.get("output_node")
             path_node = output_node.parent().createNode("name", "AUTO_PATH")
             path_node.parm("attribname").set(path_attr)
--- a/openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
+++ b/openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
@@ -64,6 +64,10 @@ class ValidatePrimitiveHierarchyPaths(pyblish.api.InstancePlugin):
 
         cls.log.debug("Checking for attribute: %s" % path_attr)
 
+        if not isinstance(output_node, hou.SopNode):
+            cls.log.warning("No geometry output node found, skipping check..")
+            return
+
         # Check if the primitive attribute exists
         frame = instance.data.get("frameStart", 0)
         geo = output_node.geometryAtFrame(frame)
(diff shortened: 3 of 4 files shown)


BigRoy (commented):
To me this PR implements some changes that I personally find more complex than what was there before - the logic is harder to follow due to how it's refactored into a `PointCachePathHelper` class.

Additionally, as explained in one of my comments forcing a single path value for `path` makes no sense for the majority of cases.

BigRoy:
Also just wanted to mention that you don't technically need a wrangle with VEX to set the path. You could just do:

image

And thus you could also just name e.g. a node using the Node's name with:

image

openpype/hosts/houdini/plugins/publish/validate_abc_path_attr.py
```diff
@@ -0,0 +1,36 @@
...
+class ValidateAbcPathAttr(pyblish.api.InstancePlugin):
+    """Validate Alembic Path attribute is correct.
+    Path attribute should be equal to "{parent_node_name}/{output_node_name}"
+    """
+
+    order = pyblish.api.ValidatorOrder + 0.1
+    families = ["pointcache"]
+    hosts = ["houdini"]
+    label = "Validate Path Attribute (Abc)"
+    actions = [FixPointCachePath]
```
moonyuet, line 19: you can take reference from other validators and use the Repair Action instead of your own implementation.

```suggestion
    actions = [RepairAction]
```
Make sure you import 
`from openpype.pipeline.publish import RepairAction`
By doing this,  you can implement repair function underneath the process function
You can take reference from the codes in other host if you are confused.
https://github.com/ynput/OpenPype/blob/53f5fc6d85cbb5b58cc624f58c760af628037fb3/openpype/hosts/max/plugins/publish/validate_frame_range.py#L61-L64C18

openpype/hosts/max/plugins/publish/validate_animation_timeline.py
```diff
@@ -0,0 +1,47 @@
...
+from pymxs import runtime as rt
+from openpype.pipeline.publish import (
+    RepairAction,
+    ValidateContentsOrder,
+    PublishValidationError
+)
+from openpype.hosts.max.api.lib import get_frame_range, set_timeline
+
+
+class ValidateAnimationTimeline(pyblish.api.InstancePlugin):
```
moonyuet, line 12: I dont think it is related to this PR  as it is from the max Host (also `extract_review_animation.py`  and `extract_thumbnailpy`)

mustafa-zarkash: Thanks for notifying me, I tried to revert to a clean commit and I think I reverted back a lot. 
My bad!

openpype/hosts/houdini/plugins/create/create_pointcache.py
```diff
@@ -37,14 +37,44 @@ def create(self, subset_name, instance_data, pre_create_data):
...
+                path_node = parent.createNode("name", "AUTO_PATH")
+                path_node.parm("attribname").set("path")
+                path_node.parm("name1").set('`opname("..")`/`opoutput(".", 0)`') # noqa
+
+                path_node.setFirstInput(child_output.input(0))
+                path_node.moveToGoodPosition()
+                child_output.setFirstInput(path_node)
+                child_output.moveToGoodPosition()
+
+            parms["sop_path"] = child_output.path()
```
BigRoy, line 77: I still don't think this is a good idea to **always** do on create.

It doesn't **require** an output node.
It doesn't **require** the path parameter to be `path` but just to be set. The user could still customize the `path` attribute (the name of it on the rop node) and have it publish fine as far as I know.

Since it's not a requirement it feels quite dirty to forcefully throw down these extra nodes all the time, especially also because deleting the publish instance does not delete these nodes.

I'd argue this is not the way to go unfortunately.

As such, I'd vote for not doing this on create.

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
```diff
@@ -104,3 +108,26 @@ def get_invalid(cls, instance):
...
+        if not paths:
+            output_node = instance.data.get("output_node")
+            path_node = output_node.parent().createNode("name", "AUTO_PATH")
+            path_node.parm("attribname").set(path_attr)
+            path_node.parm("name1").set('`opname("..")`/`opoutput(".", 0)`')
+
+            path_node.setFirstInput(output_node.input(0))
+            path_node.moveToGoodPosition()
+            output_node.setFirstInput(path_node)
+            output_node.moveToGoodPosition()
```
BigRoy, line 133: Can we add a little logging here so it states what it created and where?

openpype/hosts/houdini/plugins/create/create_pointcache.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 """Creator plugin for creating pointcache alembics."""
 from openpype.hosts.houdini.api import plugin
-from openpype.pipeline import CreatedInstance
+# from openpype.pipeline import CreatedInstance
```
BigRoy, line 4: If `CreatedInstance` is unusued then the import can just be removed.
```suggestion
```

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
```diff
@@ -104,3 +108,26 @@ def get_invalid(cls, instance):
                 "(%s of %s prims)" % (path_attr, len(invalid_prims), num_prims)
             )
             return [output_node.path()]
+
+    @classmethod
+    def repair(cls, instance):
+        output_node = instance.data.get("output_node")
```
BigRoy, line 114: Can we add a comment about what this repair is doing. Something like:
```python
# Add a "name" node before the the output node to automatically
# force a path attribute on the geometry. Be aware that this will
# create a _single name_ and thus will merge all geometry in one shape.
```

openpype/hosts/houdini/plugins/create/create_pointcache.py
mustafa-zarkash: Let me know know if I got it right : Do you think that implementing the repair action is sufficient ?

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
mustafa-zarkash: Good point!, 
Thanks for reminding me

openpype/hosts/houdini/plugins/create/create_pointcache.py
moonyuet: I agree with @BigRoy that he implementation of path parameter can be just in the repair function. User can add/customize the path attributes with some other different nodes knowing that there is not only name node allowing to add more attributes.
But I think the output node can be implemented when the user uses selection if there isn't one.

mustafa-zarkash: So, it's better to add `output node` if missing on creation 

And, repairing path attribute if missing on publishing

BigRoy: > when the user uses selection if there isn't one.

But if there's no selection - then _where_ do you add it?
Note that the tab menu creation for OpenPype creators in the network editor in Houdini actually already does exactly that.

openpype/hosts/houdini/plugins/create/create_pointcache.py
```diff
@@ -37,14 +36,30 @@ def create(self, subset_name, instance_data, pre_create_data):
...
+            # create output node if not exists
+            if not child_output:
+                child_output = parent.createNode("output", "OUTPUT")
+                child_output.setFirstInput(child_render)
+
+                child_output.setDisplayFlag(1)
+                child_output.setRenderFlag(1)
+                child_output.moveToGoodPosition()
+
+            parms["sop_path"] = child_output.path()
```
BigRoy, line 62: I still think this is an unexpected change - I wouldn't expect it to create a node, but just to pick the node I had selected.

openpype/hosts/houdini/plugins/publish/validate_abc_primitive_to_detail.py
```diff
@@ -73,6 +73,13 @@ def get_invalid(cls, instance):
         cls.log.debug("Checking Primitive to Detail pattern: %s" % pattern)
         cls.log.debug("Checking with path attribute: %s" % path_attr)
 
+        if not isinstance(output_node, hou.SopNode):
+            # In the case someone has explicitly set an Object
+            # node instead of a SOP node in Geometry context
+            # then for now we ignore.
+            cls.log.warning("No geometry output node found, skipping check..")
+            return
```
BigRoy, line 81: This doesn't seem entirely accurate right? Should it ever allow a non sop node?

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
```diff
@@ -104,3 +112,47 @@ def get_invalid(cls, instance):
...
+        It Adds a "name" node before the the output node to automatically
+        force a path attribute on the geometry. Be aware that this will
+        create a `default single name` and thus will merge all geometry in one shape.
+        """
+
+        output_node = instance.data.get("output_node")
+        rop_node = hou.node(instance.data["instance_node"])
+
+        if not (isinstance(output_node, hou.SopNode) and \
+                 output_node.type().name() == "output"):
```
BigRoy, line 129: I think the assumption that it'll be an output node is just not 100% accurate - the export works fine even if it's not an output node.

openpype/hosts/houdini/plugins/create/create_pointcache.py
moonyuet: maybe we can add another validator with repair action just for this output node.

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
```diff
@@ -104,3 +112,47 @@ def get_invalid(cls, instance):
...
+            return
+
+        frame = instance.data.get("frameStart", 0)
+        geo = output_node.geometryAtFrame(frame)
+
+        path_attr = rop_node.parm("path_attrib").eval() or "path"
+        paths = geo.findPrimAttrib(path_attr) and \
+                    geo.primStringAttribValues(path_attr)
+
+        if not paths:
```
BigRoy, line 144: Instead of basically reimplementing the validation check again I tend to write a `get_invalid` method that both `process` and `repair` use to detect whether it's currently still invalid.

Then repair could do:
```
if not cls.get_invalid(instance):
    # all ok
    return

# would probably move this into a method too so both `get_invalid` and this use the same logic 
path_attr = cls.get_path_attr(instance)  

output_node = instance.data.get("output_node")
path_node = output_node.parent().createNode("name", "AUTO_PATH")
path_node.parm("attribname").set(path_attr)
path_node.parm("name1").set('`opname("..")`/`opoutput(".", 0)`')

path_node.setFirstInput(output_node.input(0))
path_node.moveToGoodPosition()
output_node.setFirstInput(path_node)
output_node.moveToGoodPosition()

cls.log.debug(
    "'%s' was created. It adds '%s' with a default single value"
     % (path_node, path_attr)
)
```

openpype/hosts/houdini/plugins/publish/validate_sop_output_node.py
```diff
@@ -55,6 +57,16 @@ def get_invalid(cls, instance):
...
 
+        # Output node must be a Sop Output node.
+        if not output_node.type().name() == "output":
+            cls.log.error(
+                "Output node %s is not a SOP Output node. "
+                "SOP Path must point to a SOP Output node, "
+                "instead found sop type: '%s'"
+                % (output_node.path(), output_node.type().name())
+            )
+            return [output_node.path()]
```
BigRoy, line 68: I'd say this is not a requirement. All that is needed is that it's a SOP node. Doesn't have to be an output node.
```suggestion
```

Or why do you think it **must** be an output node for the publish to work?

openpype/hosts/houdini/plugins/create/create_pointcache.py
```diff
@@ -37,14 +36,30 @@ def create(self, subset_name, instance_data, pre_create_data):
...
-            parms["sop_path"] = self.selected_nodes[0].path()
+            parent = self.selected_nodes[0]
+            parms["sop_path"] = parent.path()
+
+            child_render = ""
+            child_output = ""
 
             # try to find output node
-            for child in self.selected_nodes[0].children():
+            for child in parent.children():
```
moonyuet, line 46: if user wants to select nodes(aka children nodes) inside geometry node and create instance, it will break at some point as it only allows users to select Geometry Nodes.

openpype/hosts/houdini/plugins/publish/validate_sop_output_node.py
```diff
@@ -81,3 +93,117 @@ def get_invalid(cls, instance):
                 "Output node `%s` has no geometry data." % output_node.path()
             )
             return [output_node.path()]
+
+    @classmethod
+    def repair(cls, instance):
```
BigRoy, line 98: This seems quite magical. A lot of different cases that might occur. Are we sure this is something you might want to rely on as an artist. I think the only one I like is the repair where it uses your current selection. If the repair label would then be "Set selected node as output node" that'd be super clear to the end user. For example:

```
class SetSelectedAction(RepairAction):
    label = "Set selected node as output node"
```

And then using that as an action.

openpype/hosts/houdini/plugins/create/create_pointcache.py
BigRoy: Still wouldn't know what it would "auto-set" too then.

openpype/hosts/houdini/plugins/create/create_pointcache.py
```diff
@@ -37,14 +36,30 @@ def create(self, subset_name, instance_data, pre_create_data):
...
+                if child.isGenericFlagSet(hou.nodeFlag.Render):
+                    child_render = child
                 if child.type().name() == "output":
-                    parms["sop_path"] = child.path()
+                    child_output = child
                     break
 
+            # create output node if not exists
```
moonyuet, line 53: perhaps consider this pesudo code for reference. I dont think you need to create both `child_render` and `child_output`
```
            for child in parent.children():
                # create output node if not exists
                if not child.type().name() == "output":
                    if child.isGenericFlagSet(hou.nodeFlag.Render):
                        child_output = parent.createNode("output", "OUTPUT")
                        child_output.setFirstInput(child)
                        child_output.setDisplayFlag(1)
                        child_output.setRenderFlag(1)
                        child_output.moveToGoodPosition()
                else:
                    parms["sop_path"] = child.path()
                    break
```

moonyuet:
I tried to create path attributes from attribute create node and string, it can pass the validation. When I remove the string, it fails the validation as expected but when I hit repair function it seems neither adding the related attribute back nor creating the new path attribute with correct string to replace it.
image

openpype/hosts/houdini/plugins/publish/validate_abc_primitive_to_detail.py
mustafa-zarkash: Would you recommend using this conditional because it is more general? 

```
if not hasattr(output_node, "geometry"):
  # In the case someone has explicitly set an Object
  # node instead of a SOP node in Geometry context
  # then for now we ignore - this allows us to also
  # export object transforms.
  cls.log.warning("No geometry output node found, skipping check..")
  return
```

> P.S. this file doesn't relate to the goal of this enhancement, I only added because it kept throwing error 
because of this line `geo = output_node.geometryAtFrame(frame)`

PR #5190, opened 2023-06-26
Enhancement: Cosmetics for Higher version of publish already exists validation error

BigRoy:
## Changelog Description

Fix double spaces in message.

Example output **after** the PR:

image

## Testing notes:

1. Publish from a lower workfile version than the latest publish with "Use workfile version" setting enabled.
2. Check the validation message in the publisher.

### DIFF
--- a/openpype/plugins/publish/validate_version.py
+++ b/openpype/plugins/publish/validate_version.py
@@ -25,16 +25,16 @@ class ValidateVersion(pyblish.api.InstancePlugin):
             # TODO: Remove full non-html version upon drop of old publisher
             msg = (
                 "Version '{0}' from instance '{1}' that you are "
-                " trying to publish is lower or equal to an existing version "
-                " in the database. Version in database: '{2}'."
+                "trying to publish is lower or equal to an existing version "
+                "in the database. Version in database: '{2}'."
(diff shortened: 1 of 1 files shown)


moonyuet (approved):

openpype/plugins/publish/validate_version.py
```diff
@@ -25,16 +25,16 @@ def process(self, instance):
             # TODO: Remove full non-html version upon drop of old publisher
             msg = (
                 "Version '{0}' from instance '{1}' that you are "
-                " trying to publish is lower or equal to an existing version "
-                " in the database. Version in database: '{2}'."
+                "trying to publish is lower or equal to an existing version "
+                "in the database. Version in database: '{2}'."
```
BigRoy, line 29: Haha, @kalisp it's only AFTER this is merged of course that I notice that this line should have a space after the dot at the end. :(

kalisp: Btw I myself would use `chore` label in branch as this is not much of an `enhancement`, but just house cleaning. But that just me. 
I think that `feature` and `enhancement` goes to Release notes, I don't think this needs to.

PR #5192, opened 2023-06-26
node references to check for duplicates in Max

moonyuet:
## Changelog Description
No duplicates for node references in Max when users trying to select nodes before publishing 

## Additional info
n/a

## Testing notes:
1. Launch Max via Openpype
2. Create Instance
3. Click "Add to Container" in OP Parameter in the created container
4. Add Items
5. And add the same items as 4
6.  The node references won't show the duplicated items anymore.

### DIFF
--- a/openpype/hosts/max/api/plugin.py
+++ b/openpype/hosts/max/api/plugin.py
@@ -42,6 +42,10 @@ MS_CUSTOM_ATTRIB = """attributes "openPypeData"
             (
                 handle_name = node_to_name c
                 node_ref = NodeTransformMonitor node:c
+                idx = finditem list_node.items handle_name
+                if idx do (
+                    return False
+                )
                 append temp_arr handle_name
                 append i_node_arr node_ref
             )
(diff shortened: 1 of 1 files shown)


openpype/hosts/max/api/plugin.py
```diff
@@ -42,6 +42,10 @@
             (
                 handle_name = node_to_name c
                 node_ref = NodeTransformMonitor node:c
+                idx = finditem list_node.items handle_name
+                if idx do (
+                    return False
+                )
```
BigRoy, line 48: This doesn't sound right. You wouldn't want to return directly but only continue the for loop, right? So that those in your selection that are **not** already on the list would still get added if they are in the current selection after the already existing ones?

moonyuet: At first, i tried with `if idx do continue` it doesn't work. But then I tried your suggestion of` if idx do (continue)` works smoothly. (It's quite strange but at least work). So thank you!

LiborBatek (approved):
Adding the same object to the OP container multiple times results in single list entry which is correct behavior.

Approved!
image

Minkiu (approved):

PR #5193, opened 2023-06-26
Resolve: after launch automatization fixes

jakubjezek001:
## Changelog Description
Workfile is no correctly created and aligned witch actual project. Also the launching mechanism is now fixed so even no workfile had been saved yet it will open OpenPype menu automatically.

## Additional info
Added testing script for startup.py  - it showing only if openpype has been run from code. 

## Testing notes:
1. make sure no workfile version is created in your task where resolve is launched. 
2. make sure `project_settings/resolve/launch_openpype_menu_on_start` in project settings are activated.
3. Openpype menu should popup after Resolve is active. 
4. Save as new workfile version from workfile tool and see that all is created and activated. 
5. Close Resolve and start it again. 
6. see that Resolve finds latest version and opens it with OpenPype menu
2. follow this step


openpype/hosts/resolve/startup.py
```diff
@@ -44,17 +46,22 @@ def open_file(path):
...
     # Open last workfile
     workfile_path = os.environ.get("OPENPYPE_RESOLVE_OPEN_ON_LAUNCH")
-    if workfile_path:
+
+    if workfile_path and os.path.exists(workfile_path):
+        log.info(f"Opening last workfile: {workfile_path}")
         open_file(workfile_path)
     else:
-        print("No last workfile set to open. Skipping..")
+        log.info("No last workfile set to open. Skipping..")
```
BigRoy, line 54: Do these logs even show anywhere? Last time I tried I didn't see this in the `console` output of Resolve.

openpype/hosts/resolve/utility_scripts/tests/testing_startup_script.py
BigRoy, line 1: Are we sure these tests make sense to be in the codebase? It feels like very local testing you're doing that's not suited to be in the main codebase. (However, I feel that's even more the case with the other test files).

openpype/hosts/resolve/api/workio.py
```diff
@@ -30,13 +31,21 @@ def save_file(filepath):
...
-        log.info("Saving project: `{}` as '{}'".format(name, file))
-        pm.ExportProject(name, filepath)
-    else:
-        log.info("Creating new project...")
-        pm.CreateProject(fname)
-        pm.ExportProject(name, filepath)
+    response = False
+    if name == "Untitled Project":
+        response = pm.CreateProject(fname)
+        # re-cash new current project after renaming
```
BigRoy, line 37:
```suggestion
        # recache new current project after creating new project
```

openpype/hosts/resolve/api/lib.py
```diff
@@ -89,11 +89,17 @@ def get_current_project():
     """Get current project object.
     """
     if not self.current_project:
-        self.current_project = get_project_manager().GetCurrentProject()
+        set_current_project()
 
     return self.current_project
```
BigRoy, line 94: Doesn't it just make sense to just never cache `current_project`?

Isn't it so that you can technically close a project and open a new one in the same resolve session meaning `self.current_project` would be inaccurate from that moment forward anyway?

openpype/hosts/resolve/utility_scripts/tests/testing_startup_script.py
jakubjezek001: Yeah you are right, all tests are dev tests but those are not showing anywhere once openpype is run from build and and those tests might be practical for anyone who is developing. Don't get me wrong on this - I do agree with you that they don't need to be part of codebase but also want share this somewhere and best place is the part of codebase.

openpype/hosts/resolve/api/workio.py
```diff
@@ -82,11 +85,9 @@ def current_file():
     current_dir = os.getenv("AVALON_WORKDIR")
     project = pm.GetCurrentProject()
     name = project.GetName()
-    fname = name + exported_projet_ext
+    fname = name + exported_project_ext
     current_file = os.path.join(current_dir, fname)
-    if not current_file:
-        return None
-    return os.path.normpath(current_file)
+    return os.path.normpath(current_file) if current_file else None
```
BigRoy, line 90: This is the same as:
```suggestion
    if current_file:
        return os.path.normpath(current_file)
```
Without the need to mention `None` explicitly. Not sure if better, just wanted to throw it out there.

openpype/hosts/resolve/api/workio.py
```diff
@@ -30,13 +29,19 @@ def save_file(filepath):
...
+    if name == "Untitled Project":
+        response = pm.CreateProject(fname)
+        log.info("New project created: {}".format(response))
+        pm.SaveProject()
+    elif name != fname:
+        response = project.SetName(fname)
+        log.info("Project renamed: {}".format(response))
+
+    if response:
+        # only export if project was saved or renamed
```
BigRoy, line 42: That doesn't sound right. Does it? Technically I would still expect to be able to save into the current file (and not incrementing). If my current file is `hello_v001` then I'd still be exporting it to the same. Right?

openpype/hosts/resolve/hooks/pre_resolve_launch_last_workfile.py
```diff
@@ -35,6 +37,7 @@ def execute(self):
         key = "OPENPYPE_RESOLVE_OPEN_ON_LAUNCH"
         self.launch_context.env[key] = last_workfile
 
+    def set_startup_script(self):
```
BigRoy, line 40: Seeing this being done I actually have the feeling that this logic should not at all be part of e launch hook named "ResolveLaunchLastWorkfile"? It's related to that, somewhat, but also does more than just that? Right?

64qam (approved):

openpype/hosts/resolve/api/workio.py
jakubjezek001: ah yeah, this was some dust from the development faze.

openpype/hosts/resolve/hooks/pre_resolve_launch_last_workfile.py
jakubjezek001: Good point. I will separate it from it.

openpype/hosts/resolve/api/workio.py
```diff
@@ -30,13 +27,17 @@ def save_file(filepath):
...
-        log.info("Creating new project...")
-        pm.CreateProject(fname)
-        pm.ExportProject(name, filepath)
+    response = False
+    if name == "Untitled Project":
+        response = pm.CreateProject(fname)
+        log.info("New project created: {}".format(response))
+        pm.SaveProject()
+    elif name != fname:
+        response = project.SetName(fname)
```
Minkiu, line 36: Should we be saving the file here again?

Minkiu (approved):

64qam (approved):

openpype/hosts/resolve/api/workio.py
```diff
@@ -30,13 +27,17 @@ def save_file(filepath):
...
     name = project.GetName()
 
-    if "Untitled Project" not in name:
-        log.info("Saving project: `{}` as '{}'".format(name, file))
-        pm.ExportProject(name, filepath)
-    else:
-        log.info("Creating new project...")
-        pm.CreateProject(fname)
-        pm.ExportProject(name, filepath)
+    response = False
```
BigRoy, line 30: This line could be removed by the way :) (I know it's merged already)

PR #5195, opened 2023-06-26
[DRAFT] TrayPublisher: Add support to submit publish jobs to the farm

fabiaserra:
## Changelog Description
Add support to submit OpenPype publishes to the farm through the TrayPublisher.

## Additional info
During our publishes we are running some plugins that are quite time-consuming (transcode + extract review) so having a way to run these in the farm from existing frames it's very powerful.

I had to change the order of some of the basic collect plugins as later plugins were erroring out due to missing data collected.

## Testing notes:
1. Start the TrayPublisher and select some context to publish
2. Drag and drop files you want to publish and set the render target to the farm
3. Hit publish


openpype/hosts/traypublisher/api/plugin.py
```diff
@@ -301,7 +303,21 @@ def get_pre_create_attr_defs(self):
                 single_item=True,
                 label="Reviewable representations",
                 extensions_label="Single reviewable item"
-            )
+            ),
+            TextDef(
+                "input_colorspace",
```
fabiaserra, line 308: I exposed this so we don't rely solely on colorspace input rules and the user has control to choose what's the colorspace of the given sequence

openpype/modules/deadline/plugins/publish/validate_expected_and_rendered_files.py
```diff
@@ -21,6 +21,10 @@ class ValidateExpectedFiles(pyblish.api.InstancePlugin):
 
     def process(self, instance):
         self.instance = instance
+        # TODO: Find a better way to check whether a job has been submitted with
```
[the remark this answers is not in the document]
BigRoy: Wasn't there a discussion some time ago about increasing the line character limit within OpenPype and thus tweaking the code style to allow slightly longer lines?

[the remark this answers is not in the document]
fabiaserra: I've suggested it a few times yeah but not sure if there was any traction... I think the choice of Black https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length is the best compromise

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -883,13 +891,15 @@ def process(self, instance):
         # transfer specific families from original instance to new render
         for item in self.families_transfer:
             if item in instance.data.get("families", []):
+                self.log.debug("Transfering '%s' family to instance.", item)
```
BigRoy, line 894: Reading this I was kind of thinking: "transfering from where?", "why does it need transferring?", "to where?"

Should this maybe mention it's transferring it to the farm instance?
```suggestion
                self.log.debug("Transfering '%s' family to farm instance.", item)
```
Also, for the other log.

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -1095,6 +1105,7 @@ def process(self, instance):
                 "FTRACK_SERVER": os.environ.get("FTRACK_SERVER"),
             }
 
+        deadline_publish_job_id = None
```
BigRoy, line 1108: Nice catch. However, looking at the other submission type "muster" I'm pretty sure this should **always** be True for a plugin in Deadline module hehe, and the muster stuff is waaaaaaaaay outdated.

openpype/modules/deadline/plugins/publish/validate_expected_and_rendered_files.py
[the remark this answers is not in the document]
fabiaserra: and I have since configured my IDEs to show the 79 char line and my Black extension to use the arg to 79 chars as well but I still miss it quite often

openpype/hosts/traypublisher/api/plugin.py
fabiaserra: Oh nice! I wasn't aware of that PR, indeed that serves the same purpose!

openpype/modules/deadline/plugins/publish/submit_publish_job.py
fabiaserra: Yeah fair, I added it for helping me figure out why some plugins weren't running remotely

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -883,13 +891,15 @@ def process(self, instance):
...
             if item in instance.data.get("families", []):
+                self.log.debug("Transfering '%s' family to farm instance.", item)
                 instance_skeleton_data["families"] += [item]
 
         # transfer specific properties from original instance based on
         # mapping dictionary `instance_transfer`
         for key, values in self.instance_transfer.items():
             if key in instance.data.get("families", []):
                 for v in values:
+                    self.log.debug("Transfering '%s' property to farm instance.", v)
```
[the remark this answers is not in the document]
fabiaserra: aaaghhh

openpype/modules/deadline/plugins/publish/submit_publish_job.py
fabiaserra: done

openpype/hosts/traypublisher/plugins/publish/collect_review_frames.py
```diff
@@ -16,7 +16,7 @@ class CollectReviewInfo(pyblish.api.InstancePlugin):
 
     label = "Collect Review Info"
     order = pyblish.api.CollectorOrder + 0.491
-    families = ["review"]
+    families = ["review", "render"]
```
iLLiCiTiT, line 19: You should add "review" family to instance if you want to trigger this plugin.
```suggestion
    families = ["review"]
```

openpype/hosts/traypublisher/plugins/publish/collect_simple_instances.py
```diff
@@ -65,6 +68,12 @@ def process(self, instance):
             representation_files_mapping
         )
 
+        # Hack to set env vars required to run in the farm
+        os.environ["AVALON_ASSET"] = instance.data["asset"]
+        os.environ["AVALON_TASK"] = instance.data.get("task")
+        legacy_io.Session["AVALON_ASSET"] = instance.data["asset"]
+        legacy_io.Session["AVALON_TASK"] = instance.data.get("task")
+
```
iLLiCiTiT, line 76: Why?

openpype/plugins/publish/collect_anatomy_context_data.py
```diff
@@ -42,7 +42,7 @@ class CollectAnatomyContextData(pyblish.api.ContextPlugin):
     }
     """
 
-    order = pyblish.api.CollectorOrder + 0.002
+    order = pyblish.api.CollectorOrder - 0.02
```
iLLiCiTiT, line 45: Change of this order makes this PR review complixity +1000 levels

openpype/hosts/traypublisher/plugins/publish/collect_simple_instances.py
```diff
@@ -95,6 +104,40 @@ def process(self, instance):
             )
         )
 
+        # Add targeted family to families
```
iLLiCiTiT, line 107: This whole content should be in different plugin. This would work only for simple creators.

openpype/hosts/traypublisher/plugins/publish/collect_simple_instances.py
```diff
@@ -95,6 +104,40 @@ def process(self, instance):
...
+        )
+
+        # Add render target specific data
+        if render_target == "farm":
+            # Farm rendering
+            instance.data["toBeRenderedOn"] = "deadline"
+            instance.data["transfer"] = False
+            instance.data["farm"] = True  # to skip integrate
+            if "review" in instance.data["families"]:
+                # to skip ExtractReview locally
```
iLLiCiTiT, line 122: We should maybe skip farm instances in ExtractReview than removing `review` family.

openpype/hosts/traypublisher/plugins/publish/collect_simple_instances.py
```diff
@@ -4,9 +4,12 @@
 
 import clique
 import pyblish.api
+from openpype.pipeline import publish, legacy_io
 
 
-class CollectSettingsSimpleInstances(pyblish.api.InstancePlugin):
+class CollectSettingsSimpleInstances(
+    pyblish.api.InstancePlugin, publish.ColormanagedPyblishPluginMixin
```
iLLiCiTiT, line 11: Why?

fabiaserra: Because we need to collect the colorspace data in order to run certain plugins (i.e., transcode).

openpype/hosts/traypublisher/plugins/publish/collect_simple_instances.py
fabiaserra: I just copied this from other places that do the same, I agree, but it's not my design choice so I don't know why it was done this way with the whole transfer of families in the submission, you tell me

openpype/hosts/traypublisher/plugins/publish/collect_simple_instances.py
fabiaserra: aren't all creators of Traypublisher simple creators?

openpype/plugins/publish/collect_anatomy_context_data.py
fabiaserra: Fair, I can remove the ordering from the PR if it still works for you without the order change, I will keep it in my fork. I don't quite get why collector plugins that can run almost independently aren't set to the earliest possible order...

openpype/hosts/traypublisher/plugins/publish/collect_simple_instances.py
fabiaserra: it's required so it gets written in the json file that's used in the headless publish so the task can be collected in the farm. The purpose of this PR being a draft though is to start discussion to figure out if there's a better way of doing this

openpype/hosts/traypublisher/plugins/publish/collect_review_frames.py
fabiaserra: One could argue that `plugin requires frame start/end, fps on instance data which are missing on instances from TrayPublishes.` isn't just for review purposes, what if I don't want a review?

BigRoy: I believe this issue is actually similar to another PR currently open. Does https://github.com/ynput/OpenPype/pull/5152 solve that too @fabiaserra ?

_Sorry if I'm incorrect - a bit in a hurry myself_

fabiaserra: yeah I believe so too, that's why I commented there this https://github.com/ynput/OpenPype/pull/5152#discussion_r1240099456 a few days ago as I had just encountered I needed this plugin, I think they should both be aligned for the same purpose

iLLiCiTiT: Ok then let's handle that in that PR and keep this PR without that?

openpype/hosts/traypublisher/plugins/publish/collect_simple_instances.py
iLLiCiTiT: No, there are by default editorial creators, online creator and we have many custom creators specific for client workflows.

iLLiCiTiT:
My thoughs about the PR. I think there should be used different submitter worklow. The current submitter workflow is to integrate output from different job. But this worklow requires to "replicate artist publishing on farm" -> send information from host on artist machine to farm and run publishing with the same data there. That's something we would like to call "remote publishing" (using workflile to mimic in-dcc publishing) -> that requires different validations and different launch logic and has missing implementation. The changes to push it to current workflow means that we have to lower requirements for farm publishing, which may cause some issues when something required for the workflow is missing.

The main differences are in requirements for the job, registered pyblish targets and registered pyblish host. For remote publishing should be registered only `traypublisher` host and not `shell`. Also the target should be different, nor `"local"` nor `"farm"`.

My thoughs conclusion is that we need to first make support for that in deadline addon, than making already overcomplicated code even more complicated.

Now, question is if it does worth to hack it in, because, it is possible, with sacrifices. Or implement base remote publishing logic to deadline. From my point of view it is high risk, but I have reputation of seeing issues that are behind corners, on other side of city.

## Note
There is also option n3: Write TrayPublisher Deadline plugin and add traypublisher submit plugin, which would not use submit publish job at all.

openpype/hosts/traypublisher/plugins/publish/collect_review_frames.py
fabiaserra: Sure

openpype/hosts/traypublisher/plugins/publish/collect_simple_instances.py
fabiaserra: still not following, this plugin is only running for traypublisher host? what's the difference of this and the `collect_writes` of Nuke for example? That's doing the same and not in a different plugin?

openpype/plugins/publish/collect_anatomy_context_data.py
fabiaserra: does it make sense that for a much more complex collect like `CollectSettingsSimpleInstances` it's set to `- 0.49` but for one as easy as context data is `+ 0.002`?

fabiaserra:
> My thoughs about the PR. I think there should be used different submitter worklow. The current submitter workflow is to integrate output from different job. But this worklow requires to "replicate artist publishing on farm" -> send information from host on artist machine to farm and run publishing with the same data there. That's something we would like to call "remote publishing" (using workflile to mimic in-dcc publishing) -> that requires different validations and different launch logic and has missing implementation. The changes to push it to current workflow means that we have to lower requirements for farm publishing, which may cause some issues when something required for the workflow is missing.

Why would we limit the current submitter plugin to only integrate output from a different job given that the tweaks required to not need a dependent job are rather small? There's already a catch on the plugin https://github.com/ynput/OpenPype/blob/develop/openpype/modules/deadline/plugins/publish/submit_publish_job.py#L1071 to handle the case where a render job isn't given and creating a different submitter would probably require a good deal of abstraction for both plugins to make it proper or duplicate a lot of code... at which point you might as well just spend the resources to make the submit_publish_job more friendly to handle both cases. Could you break down which different validations and launch logic are you expecting?

> The main differences are in requirements for the job, registered pyblish targets and registered pyblish host. For remote publishing should be registered only `traypublisher` host and not `shell`. Also the target should be different, nor `"local"` nor `"farm"`.

Why do you need a specific target for `traypublisher`? the logic is the same regardless of which host the output is coming from, you pass it a list of expected files and families and it runs the publish plugins remotely? I had to do very little tweaks to get the same working for Nuke https://github.com/ynput/OpenPype/pull/5196 with huge advantages

> My thoughs conclusion is that we need to first make support for that in deadline addon, than making already overcomplicated code even more complicated.

I generally agree with the theory of this so I'm all ears (eyes) to see a simplified proof of concept of how to do this, but having multiple logic branches for publishing remotely doesn't sound great either

> ## Note
> There is also option n3: Write TrayPublisher Deadline plugin and add traypublisher submit plugin, which would not use submit publish job at all.

And have scattered plugins for every tiny use case? :) Again, why do we need Traypublisher specific submission plugins if all it does is "publish"?

BigRoy:
>  The current submitter workflow is to integrate output from different job. But this worklow requires to "replicate artist publishing on farm" -> send information from host on artist machine to farm and run publishing with the same data there. That's something we would like to call "remote publishing" (using workflile to mimic in-dcc publishing) -> that requires different validations and different launch logic and has missing implementation. The changes to push it to current workflow means that we have to lower requirements for farm publishing, which may cause some issues when something required for the workflow is missing.

Either me or you (@iLLiCiTiT) are misunderstanding this PR. I read this PR description by @fabiaserra as if he didn't want to initate remote DCC publishing using a workfile.

The way I read it all he wants is:

1. Open Tray publisher
2. Configure it to publish something
3. Instead of publishing locally, trigger the publish on the farm.

I thought he solely wanted to do that because the publishing **in Tray Publisher** currently takes a long time (e.g. generating conversions for submitted rendered files). There's no workfile, no DCC, no host involved.

Of course there's the caveat that you might be dropping files into Tray Publisher which might not be available on another machine, etc.

I think the API he's really looking for is to serialize whatever the Tray Publisher's current create context (all the instances) and then do the publish headless (without GUI) on the farm.

> My thoughs conclusion is that we need to first make support for that in deadline addon, than making already overcomplicated code even more complicated.

I'd say no. The first step is being able to have a `.json` file and do a headless publish through CLI **without Deadline**. All we need to pass to deadline is the CLI command for this case.

Bonus points for being able to generate those JSON files easily programmatically making basically any "Tray publishing" programmatically for a CLI publish more trivial.

Or am I totally misunderstanding your point instead @iLLiCiTiT ?

iLLiCiTiT:
First of all, I'm not saying it does not work, I'm saying the approach is wrong. You had to modify collector and add a lot of (for me) unwanted logic there, to make it work. You're mismatching expected files with source files, it probably won't work if you publish from windows and have farm on linux, or not use simple creators but e.g. editorial creator.

## Why it is different in current deadline submitter vs. remote publishing?
Think about traypublisher as it is a DCC, like it is Maya or Nuke -> this is very important!

Current submitter tells farm (deadline) what instances and representations should be created on start of publishing, with expected files, and start publishing with `shell` host and `farm` target. There is None in-dcc logic in the plugins, a host can have logic for farm in plugins but they cannot use in-dcc logic.

Remote publishing idea is 1.) to open workfile in a DCC (in case of traypublisher it is `.json` file) and 2.) start the same logic (use same pyblish plugins) as would run on artist machine. With some smaller differences for `local` and `remote` targets, for example different validation plugins and collector that finds files required for processing. You want to run "traypublisher plugins" on "traypublisher instances" with it's logic, on farm.

But this approach would run all "farm plugins" on "traypublisher instances" + would run "in-dcc" plugins there. You may hit when the base logic of workflow with `shell` host and `farm` target would change. We would have to be careful about any changes in farm logic because there is traypublisher which is bending it, that's where I don't want to get.

> Why do you need a specific target for traypublisher? the logic is the same regardless of which host the output is coming from, you pass it a list of expected files and families and it runs the publish plugins remotely?

Target `remote` and host `traypublisher` (instead of target `farm` and host `shell`). But in case of traypublisher it is not expected files, this can work only for some of logic (e.g. simple creators), but not for complex instances like editorial or custom traypublisher logic.

> And have scattered plugins for every tiny use case? :) Again, why do we need Traypublisher specific submission plugins if all it does is "publish"?

Like I've described it is publish, but not the same publish workflow...


> 3. Instead of publishing locally, trigger the publish on the farm.
> I think the API he's really looking for is to serialize whatever the Tray Publisher's current create context (all the instances + checkboxes and entered values) is and then do the publish headless (without GUI) on the farm.

Which is not, what the current submitter is for.

> There's no workfile, no DCC, no host involved.

Like I've mentioned. Traypublisher is a host, it has own creators, own publish plugins, and addons are extending the creators and plugins. We actually planned to have option to save/open workfiles, but didn't finish it.

BigRoy:
> Like I've mentioned. Traypublisher is a host, it has own creators, own publish plugins, and addons are extending the creators and plugins. We actually planned to have option to save/open workfiles, but didn't finish it.

Actually - seeing it as a DCC I'd actually expect the same behavior. Any chance we can get a "Save Session" and "Open Session" for the standalone publisher? So that you could potentially start with some templates.

If a save and open session exists - then the standalone publishing through Deadline would just be creating a session file (`.json`?), load the session, publish.

> We actually planned to have option to save/open workfiles, but didn't finish it.

Yes, please. That basically solves the same thing right? If the session format is relatively light weight nothing is holding us from injecting something programmatically I suppose.

> > And have scattered plugins for every tiny use case? :) Again, why do we need Traypublisher specific submission plugins if all it does is "publish"?
> 
> Like I've described it is publish, but not the same publish workflow...

Not to sound harsh, but if the standalone publisher isn't intended for that. Then we'll still need an interface or method of publishing that does become that - which in essence means having another publishing method next to the standalone publisher and much more code to maintain within the same line of work. Which makes me think, _**why**_ wouldn't the standalone publisher way be that entry point? Unless you feel there's a way without Standalone Publisher to be involved to do the same thing _without the need of those 'scattered plugins'?_ If so, I feel that what @fabiaserra is basically asking is "how do I do **that** instead? (And I'm now wondering the same thing too).

> Remote publishing idea is 1.) to open workfile in a DCC (in case of traypublisher it is .json file) and 2.) start the same logic (use same pyblish plugins) as would run on artist machine. With some smaller differences for local and remote targets, for example different validation plugins and collector that finds files required for processing. You want to run "traypublisher plugins" on "traypublisher instances" with it's logic, on farm.

I see what you're getting at - there's two steps. The one for Tray Publisher that does NOT run on the farm, and some sort of result that is supposed to be farm only. However, I'm not seeing why step one of those couldn't just be also performed headless (and thus on the farm).

---

I think it's pretty clear from the discussion that the design that @fabiaserra has put in place in this PR is not the intended route. Maybe it's good to then maybe redirect to what would be the good or intended approach? Seemingly Fabia wants to put in effort to make it work but going solely based on this PR as someone on the sidelines it's quite unclear with the ongoing discussion which way he should work into instead. Leaving the code changes of this PR to the side for a bit - what do you feel would be the best way @iLLiCiTiT for @fabiaserra to implement something that solves what he's trying to do - which is performing a headless publish. Or I think what he's really looking for currently is not even that, but a headless "ingestion" of files for publishing.

fabiaserra:
Thank you guys for your feedback!!

> You're mismatching expected files with source files

I think of `expectedFiles` as a list of files that the publish process expects to have access at the point of publish in order to publish them, I don't see why we would want to do any other distinction

> it probably won't work if you publish from windows and have farm on linux, or not use simple creators but e.g. editorial creator.

Really nothing that could be sorted out if we need to make expectedFiles platform agnostic and add path mapping? You probably hit these platform difference errors constantly, it's not a flaw of the design.

> Current submitter tells farm (deadline) what instances and representations should be created on start of publishing, with expected files, and start publishing with shell host and farm target. There is None in-dcc logic in the plugins, a host can have logic for farm in plugins but they cannot use in-dcc logic

That creates two tasks in the farm, extraction and publish. At the end of the extraction step (in-DCC logic), the `expectedFiles` are files that exist on disk and then the publish runs with the pre-generated json file (nothing to do with Traypublisher or any DCC at that point), why not allowing to skip the extraction step completely by providing what the output is? That would expose the workflow so I can hijack the publish process from anywhere as long as I can provide the expected inputs, without requiring DCC specific logic. How do you intend to provide the "Use existing frames in the farm" publish functionality otherwise? Why do I need a "DCC extraction step" before I can run the publish if there's no "extraction" happening at all?

> Remote publishing idea is 1.) to open workfile in a DCC (in case of traypublisher it is .json file) and 2.) start the same logic (use same pyblish plugins) as would run on artist machine. With some smaller differences for local and remote targets, for example different validation plugins and collector that finds files required for processing. You want to run "traypublisher plugins" on "traypublisher instances" with it's logic, on farm.

The same .json file publish is used for all the other DCC publishes. I don't see the point on writing logic specific to the Traypublisher that I wouldn't want to run on any other publish process of other DCCs. In an ideal world, I would want my validation and publish code to be shared across all DCCs after the DCC logic is done (I don't want different results if I publish from Maya, Nuke or Traypublisher!) and in the case of Traypublisher, I would want it to be the thinnest layer possible to OP's publish process, there's no "DCC" logic of the Traypublisher that I shouldn't be able to run from any DCC, it's not constrained like DCCs with libraries or licenses or anything. I'm probably missing a lot of use cases you guys use the Traypublisher for where you want to do extract logic as well, but on all the families that I'm using it to publish, I just want it as a means to start the whole OP publish process, no extraction whatsoever (our publish plugins to create a slate, transcode and extract review is DCC agnostic). I just ended up writing what I need on the Traypublisher because the CLI options that I have of OP are very limited to do what I need (run the publish process headless). Like @BigRoy says, whether that's the Traypublisher or not, we need that functionality.

> But this approach would run all "farm plugins" on "traypublisher instances" + would run "in-dcc" plugins there. You may hit when the base logic of workflow with shell host and farm target would change. We would have to be careful about any changes in farm logic because there is traypublisher which is bending it, that's where I don't want to get.

I think this is exactly where we are misunderstading, that's what I want to do! There's nothing from the Traypublisher that I want to do differently from any other publish process, I want to be able to run what I do on the Traypublisher locally or in the farm. In fact, I don't even want to have to use the "Traypublisher" as a host to do that, I want a CLI where I give all the inputs that the publish process expects (i.e., through a .json file as it is right now for example, although there's much friendlier interfaces with CLI args) and trigger ALL the validation/publish plugins for that family. The Traypublisher to me should just be a UI on top of what I should be able to do exactly with the OP's publish CLI.

> but not for complex instances like editorial or custom traypublisher logic.

Isn't that why the plugin framework allows us to filter by family? For those cases we won't be submitting it to the farm the same way.

---

And yeah please, if this isn't the intended workflow, please guide me on how can I solve these use cases:

* Given a list of already rendered files start the same publish process that I can trigger from other DCCs (i.e., Nuke `use existing frames` for the `render` family)
* Run the "Use existing frames" from Nuke in the farm without the need of running Nuke remotely, just an OP process (https://community.ynput.io/t/nuke-publish-on-farm-with-existing-frames/334)

For those two use cases, I would expect a LOT of the code to be the same, I don't need Traypublisher or Nuke logic. During the publish process, I'm transcoding the frames, generating a slate and extracting reviews. All of those can run agnostic of the DCC but they can take a long time to run, thus why I need to be able to do these headless.

openpype/hosts/traypublisher/plugins/publish/collect_simple_instances.py
iLLiCiTiT: SimpleCreator is not the only creator that can be used in Traypublisher. We have other creators for "traypuslisher" host that create the same families as simple creators do, but with more sofisticated logic.

iLLiCiTiT:
> The same .json file publish is used for all the other DCC publishes. I don't see the point on writing logic specific to the Traypublisher that I wouldn't want to run on any other publish process of other DCCs.

I think that is the issue. The issue is that workfile for traypublisher can be a json, but it is NOT the same json as we send to current farm publish. The json from traypublisher is the same thing as `.nk` or `.ma` files.

>  I don't see the point on writing logic specific to the Traypublisher that I wouldn't want to run on any other publish process of other DCCs.

We actually already have remote publishing for other hosts like After Effects, Photoshop and TVPaint. We send a workfile of a DCC to farm. On farm a DCC is started, once launched a publishing is started "inside the DCC". That is what I mean by remote publishing: "Send a workfile to a farm to be processed inside the DCC". Not as "rendering job" & "integration job" but as single job that mimics local publishing. That has different workflow to current farm publishing and require different workflow than local publishing. Like different validations, more specific plugin filtering etc.

To make it properly, you should register traypublisher host in openpype and register it as pyblish host too, and NOT register `shell` host in pyblish, because it has more plugins to run than what you should use. But at the same time you have to register `shell`, to run validations of expected files -> result will be that you will start modifying current farm publish plugins because they don't match what you need to do, instead of implementing different plugins with different tartget that suits more the use case.

Like adding os.environ["AVALON_ASSET"] to collect simple instances plugin, which does not make sense, traypublisher does not have current context of an asset, but because farm publishing requires it, you've added it there. Nothing like that should happen in traypublisher, but it does.

> Run the "Use existing frames" from Nuke in the farm without the need of running Nuke remotely, just an OP process

The difference is that output of Nuke are image files. Output of traypublisher has much bigger scope. You can cover only simple creators, but the farm option would show even for complex creators where it would not work and cause confusion.

openpype/hosts/traypublisher/plugins/publish/collect_simple_instances.py
fabiaserra: exactly my point, if they aren't going through this plugin, why is it a worry that this plugin does this?

fabiaserra:
> We actually already have remote publishing for other hosts like After Effects, Photoshop and TVPaint. We send a workfile of a DCC to farm. On farm a DCC is started, once launched a publishing is started "inside the DCC". That is what I mean by remote publishing: "Send a workfile to a farm to be processed inside the DCC". Not as "rendering job" & "integration job" but as single job that mimics local publishing. That has different workflow to current farm publishing and require different workflow than local publishing. Like different validations, more specific plugin filtering etc.

I get that and that's great for DCC workflows that take benefit of **DCC-only** features... but I find treating the TrayPublisher with the same limitations and constraints as a DCC a bit backwards and adding complexity to the codebase without benefits that I can see. Last week I was trying to trigger the OpenPype publish pipeline headlessly through my own plugine and I found myself having to frankenstein and duplicate a lot of the submit publish plugin job and pieces from a lot of other plugins to get it to work. Why don't we start with the most bare bones publish pipeline that we can trigger from CLI or through a simple API? Why would we always need a host/DCC to start that?

> To make it properly, you should register traypublisher host in openpype and register it as pyblish host too, and NOT register `shell` host in pyblish, because it has more plugins to run than what you should use. But at the same time you have to register `shell`, to run validations of expected files -> result will be that you will start modifying current publish plugins because they don't match what you need to do, instead of implementing different plugins with different tartget that suits more the use case.

I get the theory but which shell plugins validations are currently redundant in this use case? Do you have some examples? For example, some different user stories:

* Choose a set of files to publish as a family on the TrayPublisher
* Run a CLI with an argument that passes a set of files and publishes it

What different validations would I want on those?

> Like adding os.environ["AVALON_ASSET"] to collect simple instances plugin, which does not make sense, traypublisher does not have current context of an asset, but because farm publishing requires it, you've added it there. Nothing like that should happen in traypublisher, but it does.

When you choose the context you are publishing an asset in the TrayPublisher, isn't that equivalent to the "current context"? What different ways are there to specify the publish context on farm submission? FWIW I probably don't need to set `os.environ["AVALON_ASSET"]` but just `legacy_io.Session["AVALON_ASSET"]`, I just copied it from other plugins that do the same but on any case, that's a very small argument to counter the idea behind this PR, there's many ways we can avoid having to add that?

> The difference is that output of Nuke are image files, and it does not matter if are created by render job or on artist machine, the different is in publish plugin which does not work with nuke. Output of traypublisher has much bigger scope and require to use in-DCC plugins on farm -> this is not possible for any other DCCs. You can cover only simple creators, but the farm option would show even for complex creators where it would not work and cause confusion.

Don't we have a way to specify the farm attributes only for Simple Creators and not the complex ones?


Again, I really don't mind if you guys don't want to support this workflow but then tell me how can we create a simple API that we can trigger the OP publish pipeline for the most common use cases. For example, in code, call a function, given a collection of files, that submits the publish process for a `render` family to the farm. With the plugins framework and how most functions are part of the plugin classes and not module functions, I find it's really hard to prepare all the inputs required to replicate the same code logic the Traypublisher does without a UI. I think that a lot of the code of `submit publish job` plugin could be abstracted and made generic as an API so others can benefit and build on top of.

fabiaserra:
> I think you still don't understand. We know it's bad, we know it's hard, and requires duplication, but at the same time if we add more and more bricks to a leaning tower, it's harder and harder to fix the leaning. We can't afford to just change everything what is currently farm related at once, it must be done in testable steps, at this moment we have to define what we actually need, because the needs we did have in past changed completelly. And we need a time for that. This is so different workflow to current submit publish job, it adds another bricks to the tower. ATM Submit publish job is for different purposes and handles more than should handle.

I think I will continue to not understand if you don't answer the questions with relatable examples... FWIW there's already two PRs from other people (https://github.com/ynput/OpenPype/pull/5451, https://github.com/ynput/OpenPype/pull/5409) making use of the exact same concepts I use here for reusing the `submit_publish_job` plugin for other "very different" workflows and I think that's going towards the good direction of cleaning up the code to make it more clear and reusable. If that's not how it should be done please make a proposal

iLLiCiTiT:
https://github.com/ynput/OpenPype/pull/5409 is reasonable, it skips render jobs and run integration job only -> that is expected behavior.

https://github.com/ynput/OpenPype/pull/5451 contains this PR, with small modifications. But it does not set global context, which is main blocker for me. Still have similar issues to this PR, we're discussing compromises, if it helps. The PR first needs separation into smaller PRs.

PR #5197, opened 2023-06-27
General: Revert backward incompatible change of path to template to multiplatform

kalisp:
## Changelog Description
Now platformity is still handed by usage of `work[root]` (or any other root that is accessible across platforms.) 

## Additional info
Previous change would broke backward compatibility as Settings are re-saving string value to dict and losing existing values.
In the case of customer request it needs to be reworked for proper multiplatformity.


jakubjezek001 (approved):
LGTM

PR #5199, opened 2023-06-27
AYON: OpenPype as server addon

iLLiCiTiT:
## Changelog Description
OpenPype repository can be converted to AYON addon for distribution. Addon has defined dependencies that are required to use it and are not in base ayon-launcher (desktop application).

## Additional info
The addon can be used with ayon-launcher or with OpenPype build with ayon executables. The logic for the addon creation is added to subfolder `./server_addon/` where is python script to create a package, pyproject.toml to define python dependencies required by the addon and server addon class definition. I'm not sure if that is the right place and if I should add helper shell/powershell scripts. If yes, should they be inside the `./server_addon/` folder or inside `./tools/`?

The README is mentioning bundles on server which are not yet available in current server develop, but will be requirement for full workflow, with installation of dependencies defined in pyproject.toml -> it won't be fully working with ayon-launcher now.

## Testing notes:
1. Create addon using `./.poetry run python ./server_addon/create_ayon_addon.py`
2. That should create `./server_addon/package/openpype/`
3. Copy `./server_addon/package/openpype/` to server addon directory
4. Restart AYON server
5. Addon should be visible in WebUI so you can enable it

### DIFF
--- a/server_addon/README.md
+++ b/server_addon/README.md
@@ -1,0 +1,21 @@
+# OpenPype addon for AYON server
+Convert openpype into AYON addon which can be installed on AYON server. The versioning of the addon is following versioning of OpenPype.
+
+## Intro
+As you might know, OpenPype is becoming AYON, which is moving from MongoDB to dedicated server with database. For some time, we'll keep OpenPype compatible with both MongoDB and AYON. But over time we'll start to change codebase to match AYON data and separate individual parts to addons.
+
+At this moment have OpenPype codebase AYON mode, which means that OpenPype is using AYON server instead of MongoDB by under-laying conversion utils. At first, we added AYON executable next to OpenPype executables to start in AYON mode. That works, but would be hard to update new versions of code (would require new full installed build). We've decided to rather create new repository where is only base desktop application logic, for now we call it AYON launcher which will replace executables created by OpenPype build, and convert openpype code to server addon which reduce size of updates.
+
+Because the overall implementation of ayon-launcher is not fully finished yet, we'll keep both ways how to start AYON mode for some time. Once the AYON launcher is finished, we'll remove AYON executables from OpenPype codebase completely.
+
+For some time this addon will be requirement as an entry point to use AYON launcher.
+
+## How to start
+There is a `create_ayon_addon.py` python file which contains logic how to create server addon from OpenPype codebase. Just run the code.
+```shell
+./.poetry/bin/poetry run python ./server_addon/create_ayon_addon.py
+```
+
+It will create directory `./package/openpype/<OpenPype version>/*` folder with all files necessary for AYON server. You can then copy `./package/openpype/` to server addons, or zip the folder and upload it to AYON server. Restart server to update addons information, add the addon version to server bundle and set the bundle for production or staging usage.
+
+Once addon is on server and is enabled, you can just run AYON launcher. Content will be downloaded and used automatically.
--- a/server_addon/create_ayon_addon.py
+++ b/server_addon/create_ayon_addon.py
@@ -1,0 +1,117 @@
+import os
+import re
+import shutil
+import zipfile
+import collections
+from pathlib import Path
+
+
+# Patterns of directories to be skipped for server part of addon
+IGNORE_DIR_PATTERNS = [
+    re.compile(pattern)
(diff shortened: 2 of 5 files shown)


server_addon/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+At this moment have OpenPype codebase AYON mode, which means that OpenPype is using AYON server instead of MongoDB by under-laying conversion utils. At first, we added AYON executable next to OpenPype executables to start in AYON mode. That works, but would be hard to update new versions of code (would require new full installed build). We've decided to rather create new repository where is only base desktop application logic, for now we call it AYON launcher which will replace executables created by OpenPype build, and convert openpype code to server addon which reduce size of updates.
+
+Because the overall implementation of ayon-launcher is not fully finished yet, we'll keep both ways how to start AYON mode for some time. Once the AYON launcher is finished, we'll remove AYON executables from OpenPype codebase completely.
+
+For some time this addon will be requirement as an entry point to use AYON launcher.
```
antirotor, line 11: I've refactored it a little bit with AI just for fun, then edited it a little bit, but I think it is in the end more refined, isn't it?

```suggestion
OpenPype is transitioning to AYON, a dedicated server with its own database, moving away from MongoDB. During this transition period, OpenPype will remain compatible with both MongoDB and AYON. However, we will gradually update the codebase to align with AYON's data structure and separate individual components into addons.

Currently, OpenPype has an AYON mode, which means it utilizes the AYON server instead of MongoDB through conversion utilities. Initially, we added the AYON executable alongside the OpenPype executables to enable AYON mode. While this approach worked, updating to new code versions would require a complete reinstallation. To address this, we have decided to create a new repository specifically for the base desktop application logic, which we currently refer to as the AYON Launcher. This Launcher will replace the executables generated by the OpenPype build and convert the OpenPype code into a server addon, resulting in smaller updates.

Since the implementation of the AYON Launcher is not yet fully completed, we will maintain both methods of starting AYON mode for now. Once the AYON Launcher is finished, we will remove the AYON executables from the OpenPype codebase entirely.

During this transitional period, the AYON Launcher addon will be a requirement as the entry point for using the AYON Launcher.
```

server_addon/create_ayon_addon.py
```diff
@@ -0,0 +1,117 @@
+import os
+import re
+import shutil
+import zipfile
+import collections
+from pathlib import Path
```
antirotor, line 6: doesn't look like `isort` processed? ;)

server_addon/create_ayon_addon.py
antirotor, line 1: since this is run by newer Pythons, I would add type hints there. Also maybe package level description on what does it do?

server_addon/create_ayon_addon.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def _value_match_regexes(value, regexes):
+    for regex in regexes:
+        if regex.search(value):
+            return True
+    return False
+
+
+def find_files_in_subdir(
```
antirotor, line 40: missing docstrings

server_addon/pyproject.toml
antirotor, line 1: I don't actually think that this is valid `pyproject.toml`. I know the rationale behind the use of it, but shouldn't we at least try to make it valid? PEP-517, PEP-518, PEP-621 and probably more? Found [this](https://validate-pyproject.readthedocs.io/en/latest/schemas.html#data-structure-for-pyproject-toml-files) as rather good source of information about its structure.

antirotor (changes requested):
Just some cosmetic comments, dismiss if you want.

server_addon/create_ayon_addon.py
iLLiCiTiT: I don't like `collections` at first place.

server_addon/pyproject.toml
iLLiCiTiT: I'm not sure what should be there. It is not full project. The project information would be duplicate of what is inside other files, but the values would not be used to anything (double maintain without any gain). To be honest I don't think it should be `pyproject.toml` at first place. All we need is list of python dependencies, the toml was probably used, because poetry expects that? Not sure...

antirotor (approved):

PR #5201, opened 2023-06-27
Merge feature from #4793 and #4996 to set frame range on instances

ClementHector:
## Changelog Description
New PR that brings together the work of:
https://github.com/ynput/OpenPype/pull/4793
https://github.com/ynput/OpenPype/pull/4996
The two PR is no longer functional, it now works with the latest update_frame_range changes.

## Testing notes:
1. Open Maya Scene with wrong frame range in instances
2. Click `set frame range` button to solve this problem
Or 
1. Build scene with `Build workfile template`
2. See all instances with the right frame range


tokejepsen:
Tested in Maya 2023 successfully. Code looks good to me.

tokejepsen (approved):

openpype/hosts/maya/api/lib.py
```diff
@@ -3129,31 +3137,52 @@ def remove_render_layer_observer():
         pass
 
 
-def update_content_on_context_change():
-    """
-    This will update scene content to match new asset on context change
+def update_instances_frame_range():
+    """Update 'frameStart', 'frameEnd', 'handleStart', 'handleEnd' and 'asset'
+    attributes of sets that got one, execpt if instance family is 'render'
```
BigRoy, line 3142:
```suggestion
    attributes of sets that got one, except if instance family is 'render'
```

openpype/hosts/maya/api/lib.py
```diff
@@ -3129,31 +3137,52 @@ def remove_render_layer_observer():
...
+    for instance in collected_instances:
+        id_attr = "{}.id".format(instance)
+        if cmds.getAttr(id_attr) != "pyblish.avalon.instance":
+            continue
+
+        if (
+            cmds.attributeQuery('family', node=instance, exists=True) and
+            cmds.getAttr("{}.family".format(instance)) == "render"
+        ):
+            continue
```
BigRoy, line 3171: Doesn't the render instance **not** have the attributes on the instance at all either and _thus_ this check is now redundant since we're doing the attribute query check below?

openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -251,6 +251,7 @@ def cleanup_placeholder(self, placeholder, failed):
         cmds.sets(node, addElement=PLACEHOLDER_SET)
         cmds.hide(node)
         cmds.setAttr(node + ".hiddenInOutliner", True)
+        reset_frame_range(playback=False, render=False, fps=False)
```
BigRoy, line 254: Should we add the `instances=True` argument here just to make clear by reading only this line that the sole focus is on resetting the instances. Just from reading this line (without knowing the function definition) it's not clear that the intent is to only update the instances in the scene.

openpype/hosts/maya/api/lib.py
ClementHector: You're right

openpype/hosts/maya/api/workfile_template_builder.py
ClementHector: The option says "_If enabled, the frame range and the handles of all the publishable instances will be updated when using the 'Reset Frame Range' functionality_" so it does not concern the creation of the workfile.

BigRoy: > so it does not concern the creation of the workfile.

Well, actually it does. Because that settings toggle influences this particular call for the workfile template too. If I disable that option in the settings than the workfile template wouldn't get the instances reset either. Not saying it's a problem in the end, just something to be wary of. I'll resolve the conversation leaving it at that.

ClementHector: You're absolutely right, I hadn't seen it that way, thank you.

Minkiu (approved):

openpype/hosts/maya/api/workfile_template_builder.py
ClementHector: I changed the logic so that the setting "_update instances on reset frame range_" has no influence on the templates.

openpype/hosts/maya/api/lib.py
```diff
@@ -2300,6 +2302,12 @@ def reset_frame_range(playback=True, render=True, fps=True):
         cmds.setAttr("defaultRenderGlobals.startFrame", animation_start)
         cmds.setAttr("defaultRenderGlobals.endFrame", animation_end)
 
+    project_name = get_current_project_name()
+    settings = get_project_settings(project_name)
+    if instances or settings["maya"]["update_publishable_frame_range"][
+            "enabled"]:
+        update_instances_frame_range()
```
BigRoy, line 2309: This won't work either - unfortunately, now there's no way to disable the behavior if user settings had it enabled. This is worse than what it was before unfortunately, since now resetting render settings will also reset all instances in the scene if it was enabled in settings - that's more confusing.

I think we have two options:
1. Revert to previous behavior, instances defaults to `True` but even when enabled only does something if it's enabled in settings and `False` always disables it.
2. Change `instances` default value to `None`, when `None` it uses the value from settings, when `True` it's force enabled, when `False` it's force disabled.

Better ideas are welcome of course.

Something like:

```python
if instances is None:
    # Allow default behavior to come from studio settings if True or False
    # is not explicitly set
    settings = get_project_settings(project_name)
    instances = settings["maya"]["update_publishable_frame_range"]["enabled"]

if instances:
    update_instances_frame_range()
```

Nice thing about the old behavior was sort of simplicity, if someone still wanted to reset just the instances no matter the setting you could still use: `update_instances_frame_range()` instead of `reset_frame_range`


@tokejepsen @antirotor Thoughts? Any preferences?

openpype/hosts/maya/api/lib.py
```diff
@@ -3129,31 +3137,46 @@ def remove_render_layer_observer():
...
-        except ValueError:
-            pass
+
+    asset_data = asset_doc["data"]
+    frames_attributes = {
+        'frameStart': asset_data["frameStart"],
+        'frameEnd': asset_data["frameEnd"],
+        'handleStart': asset_data["handleStart"],
+        'handleEnd': asset_data["handleEnd"],
+        'asset': asset_doc['name']
```
BigRoy, line 3159: **Wait a second!** I had totally missed this but this to me is confusing behavior now with this function name and it triggering on `reset_frame_range`. I would not expect _that_ to force reset the asset. We should probably change this, right?

@tokejepsen @antirotor Thoughts?

openpype/hosts/maya/api/lib.py
```diff
@@ -3129,31 +3137,46 @@ def remove_render_layer_observer():
         pass
 
 
-def update_content_on_context_change():
-    """
-    This will update scene content to match new asset on context change
+def update_instances_frame_range():
+    """Update 'frameStart', 'frameEnd', 'handleStart', 'handleEnd' and 'asset'
+    attributes of sets that got one, except if instance family is 'render'
     """
```
BigRoy, line 3143:
```suggestion
    """Update 'frameStart', 'frameEnd', 'handleStart', 'handleEnd' and 'asset'
    attributes of publishable instances (their objectSets) that got one.
    """
```

openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -251,6 +251,12 @@ def cleanup_placeholder(self, placeholder, failed):
         cmds.sets(node, addElement=PLACEHOLDER_SET)
         cmds.hide(node)
         cmds.setAttr(node + ".hiddenInOutliner", True)
+        reset_frame_range(
+            playback=False,
+            render=False,
+            fps=False,
+            instances=True
+        )
```
BigRoy, line 259: With these arguments I'd say at this stage you're just better off calling `lib.update_instances_frame_range()` to ignore the settings for reset frame range. Right?

openpype/hosts/maya/api/lib.py
tokejepsen: Yeah, that goes outside the scope of this method.
Also since we are looking at Maya here specifically we dont need to reset the asset cause that is validated by https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/maya/plugins/publish/validate_instance_in_context.py.
An OT question is whether we want to elevate the above plugin to be more universal. Especially with the new publisher, it would be very easy to make it universal (i think).

openpype/hosts/maya/api/workfile_template_builder.py
tokejepsen: Good point. Lets go through `update_instances_frame_range`

openpype/hosts/maya/api/lib.py
tokejepsen: I would opt for simpler solution of going with `instances=True` by default. Any forced update of the instances would happen through `update_instances_frame_range`

tokejepsen (changes requested):

openpype/hosts/maya/api/lib.py
iLLiCiTiT: I guess this would be much easier to resolve in new publisher where creators could handle that instead of "guessing" what and how should be updated?

BigRoy: Not sure how the new publisher would solve this particular case. The issue here is mostly about something that is studio configurable behavior for `reset_frame_range` functionality (a menu entry in Maya too) and how it influences instances + how that functionality is accessible in the codebase.

openpype/hosts/maya/api/lib.py
```diff
@@ -3129,31 +3137,53 @@ def remove_render_layer_observer():
...
-                                 new_data["frameEnd"],)
-        except ValueError:
-            pass
+    asset_data = asset_doc["data"]
+
+    set_attribute_instances("frameStart", asset_data["frameStart"])
+    set_attribute_instances("frameEnd", asset_data["frameEnd"])
+    set_attribute_instances("handleStart", asset_data["handleStart"])
+    set_attribute_instances("handleEnd", asset_data["handleEnd"])
+    set_attribute_instances("fps", asset_data['fps'])
```
BigRoy, line 3186: This will be very slow - we're querying all the instances every time. This is far from the recommended approach.

Here's a more optimal approach.

This only queries the asset if we need it.
This only queries the related data.
This only tries to find the instances once.
```python
def iter_publish_instances():
    for node in cmds.ls(
        "*.id",
        long=True,
        type="objectSet",
        recursive=True,
        objectsOnly=True
    )
        id_attr = "{}.id".format(instance)
        if cmds.getAttr(id_attr) != "pyblish.avalon.instance":
            continue

       yield node

def update_instances_frame_range():
    attributes = ["frameStart", "frameEnd", "handleStart", "handleEnd", "fps"]
    attrs_per_instance = {}
    for instance in iter_publish_instances():
        instance_attrs = []
        for attr in attributes:
            if cmds.attributeQuery(attr, node=node, exists=True):
                instance_attrs .append(attr)
        if instance_attrs :
            attrs_per_instance [instance] = instance_attrs

    if not attrs_per_instance:
        # no instances with any frame related attributes
        return

    fields = ["data.{}".format(key) for key in attributes]
    asset_doc = get_current_project_asset(fields=fields)
    asset_data = asset_doc["data"]
    for node, attrs in attrs_per_instance.items():
        for attr in attrs:
            plug = "{}.{}".format(node, attr)
            value = asset_data[attr]
            if cmds.getAttr(plug, type=True) == "string":
                cmds.setAttr(plug, value, type="string")
            else:
                cmds.setAttr(plug, value)
```

openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -251,6 +257,8 @@ def cleanup_placeholder(self, placeholder, failed):
         cmds.sets(node, addElement=PLACEHOLDER_SET)
         cmds.hide(node)
         cmds.setAttr(node + ".hiddenInOutliner", True)
+        update_instances_frame_range()
+        update_instances_asset_attribute()
```
BigRoy, line 261: Are we sure this is in the right place. `cleanup_placeholder` sounds like it runs _per placeholder_ instead of at the very end of workfile build. We only want to run this once, right? Not every single time.

I feel like this is probably in the wrong place.

ClementHector: Well seen, it is true that it is executed at each placeholder. I think we should move this to [here](https://github.com/ynput/OpenPype/blob/fe58cb3c7cc4520b61c8b73c166f6694c5d29e91/openpype/pipeline/workfile/workfile_template_builder.py#L712) with a `post_populate_scene_placeholders` method but we don't can't call maya commands here, maybe we should implement instance-related commands in `host` class? maybe it doesn't make sense with the new publisher (which I don't know)?
Tell me what would make more sense.

BigRoy: I feel like the abstract workfile builder is missing some methods that trigger before the build and after the build. We could expose a `after_build_process()` and `pre_build_process()` so that each DCC could implement its own callbacks that run before or after ALL placeholders. Not sure if the best approach and whether it's really missing - but couldn't quickly find a feature that does that.

Another way could be for the workfile building to emit events `emit_event("workfile.before.build")` and `emit_event("workfile.after.build")` so that any module could attach callbacks to the events using `register_event_callback` - which the host itself could do e.g. on `install()` etc.

Not sure which would be the best approach. @iLLiCiTiT @tokejepsen @antirotor thoughts?

openpype/hosts/maya/api/lib.py
```diff
@@ -3129,31 +3137,63 @@ def remove_render_layer_observer():
...
+
+    attrs_per_instance = {}
+    for instance in iter_publish_instances():
+        instance_attrs = [
+            attr for attr in attributes
+            if cmds.attributeQuery(attr, node=instance, exists=True)
+        ]
+
+        if instance_attrs :
+            attrs_per_instance [instance] = instance_attrs
```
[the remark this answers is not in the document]
BigRoy: That was likely my bad. :)
```suggestion
            attrs_per_instance[instance] = instance_attrs
```

openpype/hosts/maya/api/lib.py
```diff
@@ -3129,31 +3137,63 @@ def remove_render_layer_observer():
...
+    attributes = ["frameStart", "frameEnd", "handleStart", "handleEnd", "fps"]
+
+    attrs_per_instance = {}
+    for instance in iter_publish_instances():
+        instance_attrs = [
+            attr for attr in attributes
+            if cmds.attributeQuery(attr, node=instance, exists=True)
+        ]
+
+        if instance_attrs :
```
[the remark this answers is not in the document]
BigRoy:
```suggestion
        if instance_attrs:
```

openpype/hosts/maya/api/pipeline.py
```diff
@@ -653,7 +653,7 @@ def on_task_changed():
 
     with lib.suspended_refresh():
         lib.set_context_settings()
-        lib.update_content_on_context_change()
+        lib.update_instances_frame_range()
```
BigRoy, line 656: Should this also get the `lib.update_instances_asset_attribute()` call to behave as it did before?

openpype/hosts/maya/api/lib.py
```diff
@@ -3131,31 +3139,63 @@ def remove_render_layer_observer():
...
-                    cmds.setAttr("{}.frameEnd".format(s),
-                                 new_data["frameEnd"],)
-        except ValueError:
-            pass
+
+    for instance in iter_publish_instances():
+        if not cmds.attributeQuery("asset", node=instance, exists=True):
+            continue
+        attr = "{}.asset".format(instance)
+        cmds.setAttr(attr, get_current_asset_name(), type="string")
```
BigRoy, line 3166:
```suggestion
    asset_name = get_current_asset_name()
    for instance in iter_publish_instances():
        if not cmds.attributeQuery("asset", node=instance, exists=True):
            continue
        attr = "{}.asset".format(instance)
        cmds.setAttr(attr, asset_name, type="string")
```

openpype/hosts/maya/api/lib.py
```diff
@@ -3131,31 +3139,63 @@ def remove_render_layer_observer():
...
+
+    fields = ["data.{}".format(key) for key in attributes]
+    asset_doc = get_current_project_asset(fields=fields)
+    asset_data = asset_doc["data"]
+
+    for node, attrs in attrs_per_instance.items():
+        for attr in attrs:
+            plug = "{}.{}".format(node, attr)
+            value = asset_data[attr]
+            cmds.setAttr(plug, value)
```
BigRoy, line 3198: One minor note about this - say you have an instance that does **not** match the current shot. :) Would we want it to reset to the fream range of the asset it's publishing to, or the one of the currently publishing from?

Currently it does the latter.

ClementHector: In this case it seems that the option which deactivates the update on the instances is not sufficient! Should this option be a filter on the name of the instances?

BigRoy: > In this case it seems that the option which deactivates the update on the instances is not sufficient! Should this option be a filter on the name of the instances?

Sorry, could you elaborate? Not sure what you mean with it being not sufficient?

iLLiCiTiT:
I have maybe stupid comment, but isn't this PR outdated since this PR https://github.com/ynput/OpenPype/pull/4388 ?

BigRoy:
> I have maybe stupid comment, but isn't this PR outdated since this PR https://github.com/ynput/OpenPype/pull/4388 ?

Yes, I think so. Some of the logic still makes sense to keep - like updating the frame ranges, etc. but likely will need a decent overhaul to work with the new Creator system instead.

tokejepsen:
@ClementHector what is the status of this PR? Are you using the new Creator/Publisher?

mkolar:
@ClementHector we're closing this PR as it is now stale and obsolete. If there is any functionality from this you'd like to contribute again, feel free to re-open this again, targeting AYON

PR #5202, opened 2023-06-28
Max: Publisher instances dont hold its enabled disabled states when Publisher reopened again

### LINKED ISSUE #5183: Bug: 3dsmax : Publisher Instances dont hold its Enabled/Disabled states when Publisher reopened again

### Current Behavior:

When user turn off some instance(s) it doesnt keep its state when Publisher reopened. 

https://github.com/ynput/OpenPype/assets/112623825/31cdda5e-052c-4494-8537-c2a0dd2ca19b

Also doesnt work when workfile saved and reopened too. 

### Expected Behavior:

Publishers instances should keep their states (ON/OFF) if user changed it when reopened again. also valid if workfile saved and reopened later on.

### Version

3.15.11-nightly.4

Windows

### Steps To Reproduce:

1. Create any ```Instance``` in Publisher Create tab
2. Change the state of some Instances from On to Off
3. Close Publisher
4. Open Publisher

### Relevant log output:

_No response_

### Additional context:

_No response_

[cuID:[OP-6245](https://app.clickup.com/t/863h2jkpp)]

moonyuet:
## Changelog Description
Resolve #5183, general maxscript conversion issue to python (e.g. bool conversion, true in maxscript while True in Python)
(Also resolve the ValueError when you change the subset to publish into list view menu)

## Additional info
n/a

## Testing notes:
1. Launch Max via OP
2. Openpype -> Publish...
3. Create Instance
4. Disable it
5. Close and re-open the publisher
6. It should show the instance disabled instead of enabled

### DIFF
--- a/openpype/hosts/max/api/lib.py
+++ b/openpype/hosts/max/api/lib.py
@@ -78,13 +78,15 @@ def read(container) -> dict:
                 value.startswith(JSON_PREFIX):
             with contextlib.suppress(json.JSONDecodeError):
                 value = json.loads(value[len(JSON_PREFIX):])
-        if key.strip() == "active":
-            if value == "true":
-                data[key.strip()] = True
-            else:
-                data[key.strip()] = False
-        else:
-            data[key.strip()] = value
+
+        # default value behavior
+        # convert maxscript boolean values
+        if value == "true":
+            value = True
+        elif value == "false":
+            value = False
+
+        data[key.strip()] = value
 
     data["instance_node"] = container.Name
 
--- a/openpype/hosts/max/api/lib.py
+++ b/openpype/hosts/max/api/lib.py
@@ -78,7 +78,13 @@ def read(container) -> dict:
                 value.startswith(JSON_PREFIX):
             with contextlib.suppress(json.JSONDecodeError):
                 value = json.loads(value[len(JSON_PREFIX):])
-        data[key.strip()] = value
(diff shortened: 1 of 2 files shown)


LiborBatek (approved):
Did multiple tests with enabled / disabled instances and all works as expected.
Also when the workfile being saved and reopened.
image

All good.

openpype/hosts/max/api/lib.py
```diff
@@ -78,7 +78,13 @@ def read(container) -> dict:
...
             with contextlib.suppress(json.JSONDecodeError):
                 value = json.loads(value[len(JSON_PREFIX):])
-        data[key.strip()] = value
+        if key.strip() == "active":
+            if value == "true":
+                data[key.strip()] = True
+            else:
+                data[key.strip()] = False
+        else:
+            data[key.strip()] = value
```
BigRoy, line 87: You seem to always want to use the stripped key, so do it once.
```suggestion
        key= key.strip()
        
        if key== "active":
            # Special boolean conversion
            value = True if value == "true" else False
        
        data[key] = value
```
Are we sure there are no other boolean data entries that we want to convert? Why is only the `active` key special?

moonyuet: I am not quite sure but as far as I checked from the debug, it is only bool value in the dict

Issue #5138, opened 2023-06-14
Bug: OpenTimelineIO shouldn't be part of global environment


antirotor:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

We are installing OpenTimelineIO to OpenPype environment and then we are passing it to hosts. But there is a chance host will have its own OTIO or its own verison of Python with specific OTIO version. Only order of path resolution will then result either in success or failure when OTIO modules are imported.

Current case:

**Resolve with Python 3.6**
There are no existing wheels of OTIO for Python 3.6, so custom build must be done. When OTIO from OpenPype preceed in path variables OTIO for Python 3.6, OpenPype integration in Resolve will crash because of OTIO import.

Temporary solution:

Add to resolve settings `PYTHONPATH` like:

```
"PYTHONPATH": [
    "{LOCALAPPDATA}/Programs/Python/Python36/lib/site-packages",
    "{PYTHONPATH}"
]
```
*(on windows)*

That way OTIO installed to Python 3.6 takes priority over default OpenPype version, but it is far from ideal.

### Expected Behavior:

OTIO should be handled the same way as PySide2, OpenColorIO and other frameworks that might already be used in some form in hosts.

### Version

3.15.10

### What platform you are running OpenPype on?

Windows

### Steps To Reproduce:

This might  be hard to reporoduce because it depends on the specific studio configuration and environments. You can see it in effect in hosts with different python version than OpenPype has with their own OpenTimelineIO.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

BigRoy:
_Just referencing some related issues/PR/discussions._

Just linking [this hack from OpenRV PR which tries to workaround a similar issue but with `PyOpenColorIO`](https://github.com/ynput/OpenPype/pull/4839/files#diff-861536d4e90db05cb261cc42ec534924e072b3da2893ddb573fedfb0f621e642R19-R35) which **might?** have been solved since with https://github.com/ynput/OpenPype/pull/4946 because a similar issue occurred with [the Gaffer PR](https://github.com/ynput/OpenPype/pull/4267#issuecomment-1537422928).

The [OpenTimelineIO issue is also being faced with Nuke Studio (Hiero) integration as came up on Discord here](https://discord.com/channels/517362899170230292/563751989075378201/1123562258241421372) basically breaking publishing in versions not matching the right Python version it seems.

[...]

PR #5206, opened 2023-06-28
General: Runtime dependencies

iLLiCiTiT:
## Changelog Description
Defined runtime dependencies in pyproject toml. Moved python ocio and otio modules there.

## Additional info
This change makes sure that both modules won't be in PYTHONPATH thus not propagated to launched DCCs, which should fix possible issues e.g. in NukeStudio/Hiero.

## Testing notes:
1. Delete `./.venv` if you have any in repository
2. Run `./tools/create_env`
3. Run `./tools/fetch_thirdparty_libs`

Resolves https://github.com/ynput/OpenPype/issues/5138

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1895,46 +1895,6 @@ files = [
 setuptools = "*"
 
 [[package]]
-name = "opentimelineio"
-version = "0.14.1"
-description = "Editorial interchange format and API"
-category = "main"
-optional = false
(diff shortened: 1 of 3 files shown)


antirotor (changes requested):
Looks good, I would just change the print.

Minkiu (approved):

PR #5207, opened 2023-06-28
General: custom staging dir functionality migration

jakubjezek001:
## Changelog Description
The functionality of the staging directory has exceeded its original purpose of publishing. Considering its increased usage in the creation process, we need to migrate it to a dedicated module within the pipeline parent.

## Additional info
- adding functionality to Creator base class
- consistency of naming across all features
- making sure backward compatibility to older names and locations remain in settings and api
- anatomy templates with Staging Dir category

## Testing notes:
1. all should be working as it was before


openpype/pipeline/stagingdir.py
```diff
@@ -0,0 +1,146 @@
...
+)
+
+
+TRANSIENT_DIR_TEMPLATE = "transient"
+
+
+def get_custom_staging_dir_info(project_name, host_name, family, task_name,
+                                task_type, subset_name,
+                                project_settings=None,
+                                anatomy=None, log=None):
```
BigRoy, line 22: This returning some sort of "info" also sounds like this might expand over time. I'd say this could better return a `dict` with keys related to the info than a 2-tuple now which extends later breaking backwards compatibilities.

```python
return {
    "path": path,
    "persistent": persistent
}
```

openpype/pipeline/publish/lib.py
```diff
@@ -757,6 +698,7 @@ def get_publish_repre_path(instance, repre, only_published=False):
         str: Path to representation file.
         None: Path is not filled or does not exists.
     """
+    from .. import get_instance_staging_dir
```
BigRoy, line 701: Aren't we avoiding relative imports almost everywhere else?

openpype/pipeline/create/creator_plugins.py
```diff
@@ -611,6 +614,52 @@ def get_pre_create_attr_defs(self):
...
+            str: Task type.
+        """
+        create_context = self.create_context
+        project_name = create_context.get_current_project_name()
+        asset_name = create_context.get_current_asset_name()
+        task_name = create_context.get_current_task_name()
+
+        asset_doc = get_asset_by_name(project_name, asset_name)
+
+        return get_task_type(asset_doc, task_name)
```
BigRoy, line 661: If `get_task_type` takes the `asset_doc` shouldn't the call to `self.get_task_type` also actually use the asset from the instance instead of the global create context?

openpype/pipeline/create/creator_plugins.py
```diff
@@ -611,6 +614,52 @@ def get_pre_create_attr_defs(self):
...
+                staging dir.
+            task_type (str): Task type. If not set it will be queried from
+        Returns:
+            Tuple[Any, Any]: Tuple of staging dir and is_persistent or None
+        Raises:
+            ValueError - if misconfigured template should be used
+        """
+        from openpype.pipeline import get_custom_staging_dir_info
+
+        task_name = instance["task"]
```
BigRoy, line 631: Wasn't there a possibility in tray/standalone publisher to publish without a **task** set? And thus `instance["task"]` may not exist?

BigRoy (commented):
Not sure about the changes and whether it improves any readability, especially the circular imports (and it being worked around) feels problematic.

What stood out to me is that `get_instance_staging_dir` does not rely on the custom staging dir info at all nor the other way around. Which makes me wonder how do I actually get the correct staging directory for an instance. I'd have expected that `get_instance_staging_dir` would've considered also the custom transient dir setting, etc. but it seems that's still up to the implementation to check first whether an instance should have a custom staging dir, and if not.. then use `get_instance_staging_dir` which seems confusing from the name of that function.

openpype/pipeline/stagingdir.py
jakubjezek001: Yes I was considering this too. It had been inpremented.

openpype/pipeline/publish/lib.py
jakubjezek001: Is it problem?? Or is there a better way to do it?

openpype/pipeline/create/creator_plugins.py
jakubjezek001: Good point. Need to look into this, since it is really optional in TrayPublisher.

openpype/pipeline/create/creator_plugins.py
jakubjezek001: Good point. Need to look into this, since it is really optional in TrayPublisher.

openpype/pipeline/create/creator_plugins.py
jakubjezek001: Is the question related to editorial case? I presume we could face a problem there, but still it feels like you would want to use transient data linked to editorial task (and asset parent) rather than to each individual shot instances. Or is there another case you are having in mind?

openpype/pipeline/publish/lib.py
jakubjezek001: ok, so I have figured out with the assistance of @iLLiCiTiT ;)

openpype/pipeline/create/creator_plugins.py
jakubjezek001: fixed

openpype/pipeline/create/creator_plugins.py
```diff
@@ -611,6 +615,56 @@ def get_pre_create_attr_defs(self):
         """
         return self.pre_create_attr_defs
 
+    def get_transient_data_profile(self, instance, task_type=None):
```
Minkiu, line 618: I find it a bit confusing that they are called exactly the same :upside_down_face:

openpype/pipeline/stagingdir.py
```diff
@@ -0,0 +1,162 @@
...
+
+    Args:
+        instance (pyblish.lib.Instance): Instance for which we want to get
+            staging dir.
+
+    Returns:
+        str: Path to staging dir of instance.
+    """
+    staging_dir = instance.data.get('stagingDir')
+    if staging_dir:
```
Minkiu, line 140: Would it be worth checking if the `stagingDir` actually exists on disk?

openpype/plugins/publish/collect_transient_data_dir.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class CollectTransientDataDir(pyblish.api.InstancePlugin):
+    """Apply matching Transient data profile (custom stagingDir) to a instance.
+
+    Transient dir could be useful in specific use cases where is
+    desirable to have temporary renders in specific, persistent folders, could
+    be on disks optimized for speed for example.
+
+    It is studio responsibility to clean up obsolete folders with data.
```
Minkiu, line 27: `the studio's`

openpype/plugins/publish/collect_transient_data_dir.py
```diff
@@ -0,0 +1,105 @@
...
+        project_settings = instance.context.data["project_settings"]
+        anatomy = instance.context.data["anatomy"]
+        task = instance.data["anatomyData"].get("task", {})
+
+        transient_data_profile = get_transient_data_profile(
+            project_name, host_name, family, task.get("name"),
+            task.get("type"), subset_name, project_settings=project_settings,
+            anatomy=anatomy, log=self.log)
+
+        if transient_data_profile:
```
Minkiu, line 57: Shouldn't this be `if not` ? Otherwise the code at line 67 will never execute?  Or we'll pass an empty `transient_data_profile` to `_apply_transient_data` (and we are not doing any check there if it's empty or not)

jakubjezek001: huh.. yes that is correct

openpype/pipeline/stagingdir.py
jakubjezek001: At this stage path should exists. But yeah since I need it to prove it, I have found out that `CollectManagedStagingDir` plugin was missing the `os.makedirs`. Thanks @Minkiu !

openpype/pipeline/create/creator_plugins.py
jakubjezek001: @BigRoy how about to do it this way? https://github.com/ynput/OpenPype/pull/5207/commits/a31ce9ad14713cc70cdb17ed2739fd30d9c5cbf3

BigRoy: Actually yes, because to me this actually goes to show that `get_task_type` is not really needed as a method to have it on the `Creator` plugin to begin with since it's a very thin wrapper to `openpype.pipeline.template_data.get_task_type`. Or better yet, why do we need `get_staging_dir_profile` on the `Creator` plugin? Is it even used - it doesn't appear to be used in this PR?

openpype/pipeline/create/creator_plugins.py
```diff
@@ -611,6 +618,57 @@ def get_pre_create_attr_defs(self):
...
+        Returns:
+            str: Task type.
+        """
+        if not task_name:
+            return None
+
+        create_context = self.create_context
+        asset_name = asset_name or create_context.get_current_asset_name()
+        project_name = create_context.get_current_project_name()
+        task_name = create_context.get_current_task_name()
```
iLLiCiTiT, line 666:
```suggestion
        create_context = self.create_context
        if not task_name:
            task_name = create_context.get_current_task_name()
            if not task_name:
                return None

        asset_name = asset_name or create_context.get_current_asset_name()
        project_name = create_context.get_current_project_name()
```

openpype/pipeline/stagingdir.py
```diff
@@ -0,0 +1,212 @@
...
+    Args:
+        project_name (str)
+        host_name (str)
+        family (str)
+        task_name (str)
+        task_type (str)
+        subset_name (str)
+        project_settings(Dict[str, Any]): Prepared project settings.
+        anatomy (Dict[str, Any])
+        log (Logger) (optional)
```
iLLiCiTiT, line 33:
```suggestion
        log (Optional[logging.Logger])
```

openpype/pipeline/create/creator_plugins.py
```diff
@@ -611,6 +618,57 @@ def get_pre_create_attr_defs(self):
...
+            project_settings=self.project_settings,
+            anatomy=self.project_anatomy, log=self.log
+        )
+
+    def get_task_type(self, task_name=None, asset_name=None):
+        """Get task type from task name.
+
+        Args:
+            task_name (str)[optional]: Task name.
+            asset_name (str)[optional]: Asset name.
```
iLLiCiTiT, line 655:
```suggestion
            task_name (Optional[str]): Task name.
            asset_name (Optional[str]): Asset name.
```

openpype/pipeline/stagingdir.py
```diff
@@ -0,0 +1,212 @@
...
+
+    if template_name not in anatomy.templates[STAGING_DIR_TEMPLATES]:
+        raise ValueError((
+            "Anatomy of project \"{}\" does not have set"
+            " \"{}\" template key at Staging Dir section!").format(
+                project_name, template_name)
+        )
+
+
+def get_instance_staging_dir(instance):
```
iLLiCiTiT, line 121: Shouldn't this function stay in publish lib? It is directly connected to publish instance.

openpype/pipeline/stagingdir.py
```diff
@@ -0,0 +1,212 @@
...
+import tempfile
+from openpype.lib import (
+    Logger,
+    filter_profiles
+)
+from openpype import pipeline
+from openpype.settings import (
+    get_project_settings
+)
+
```
iLLiCiTiT, line 11: Use relative imports (would require to modify usages in code).
```suggestion
from openpype.settings import get_project_settings

from .anatomy import Anatomy
from .tempdir import create_custom_tempdir
```

openpype/pipeline/create/creator_plugins.py
```diff
@@ -611,6 +618,57 @@ def get_pre_create_attr_defs(self):
...
+            self.host_name,
+            self.family,
+            task_name,
+            task_type,
+            instance["subset"],
+            project_settings=self.project_settings,
+            anatomy=self.project_anatomy, log=self.log
+        )
+
+    def get_task_type(self, task_name=None, asset_name=None):
```
iLLiCiTiT, line 650: Shouldn't be actually all of that required?

openpype/pipeline/publish/lib.py
```diff
@@ -20,18 +19,17 @@
     get_system_settings,
 )
 from openpype.pipeline import (
-    tempdir,
-    Anatomy
+    get_staging_dir_profile,
+    get_instance_staging_dir as _get_instance_staging_dir
```
iLLiCiTiT, line 23:
```suggestion
    get_instance_staging_dir
```

openpype/pipeline/publish/lib.py
```diff
@@ -684,56 +682,21 @@ def context_plugin_should_run(plugin, context):
...
-        staging_dir = os.path.normpath(
-            tempfile.mkdtemp(prefix="pyblish_tmp_")
-        )
-    instance.data['stagingDir'] = staging_dir
 
-    return staging_dir
+# deprecated: backward compatibility only
+# TODO: remove in the future
+def get_instance_staging_dir(instance):
+    return _get_instance_staging_dir(instance)
```
iLLiCiTiT, line 701:
```suggestion
```

openpype/pipeline/create/creator_plugins.py
```diff
@@ -19,6 +19,13 @@
     deregister_plugin,
     deregister_plugin_path
 )
+from openpype.pipeline import (
+    get_staging_dir_profile as _get_staging_dir_profile
+)
```
iLLiCiTiT, line 24: Why `get_staging_dir_profile as _get_staging_dir_profile` ?
```suggestion
from openpype.pipeline import get_staging_dir_profile
```

openpype/pipeline/publish/lib.py
```diff
@@ -775,7 +738,7 @@ def get_publish_repre_path(instance, repre, only_published=False):
 
     staging_dir = repre.get("stagingDir")
     if not staging_dir:
-        staging_dir = get_instance_staging_dir(instance)
+        staging_dir = _get_instance_staging_dir(instance)
```
iLLiCiTiT, line 741:
```suggestion
        staging_dir = get_instance_staging_dir(instance)
```

openpype/pipeline/stagingdir.py
```diff
@@ -0,0 +1,212 @@
...
+    if not staging_dir:
+        staging_dir = get_staging_dir(
+            project_name=instance.context.data["projectName"],
+            anatomy=anatomy
+        )
+
+    instance.data['stagingDir'] = staging_dir
+
+    return staging_dir
```
iLLiCiTiT, line 141: Skip reassignment
```suggestion
    staging_dir = instance.data.get("stagingDir")
    if staging_dir :
        return staging_dir 

    staging_dir = get_staging_dir(
        project_name=instance.context.data["projectName"],
        anatomy=anatomy
    )

    instance.data["stagingDir"] = staging_dir

    return staging_dir
```

openpype/pipeline/stagingdir.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+def get_staging_dir(
```
iLLiCiTiT, line 144: Is this function bypassing the custom staging dir?

openpype/plugins/publish/extract_review.py
```diff
@@ -20,13 +20,13 @@
...
-    convert_input_paths_for_ffmpeg,
-    get_transcode_temp_directory,
+    convert_input_paths_for_ffmpeg
 )
-from openpype.pipeline.publish import (
-    KnownPublishError,
-    get_publish_instance_label,
+from openpype.pipeline import (
+    publish,
+    get_staging_dir
```
iLLiCiTiT, line 27: I would preffer explicit import if possible.
```suggestion
from openpype.pipeline import get_staging_dir
from openpype.pipeline.publish import (
    KnownPublishError,
    get_publish_instance_label,
)
    
```

openpype/pipeline/create/creator_plugins.py
jakubjezek001: Yeah, this is preparation for creators. Those will be using `get_staging_dir_profile` or better `apply_staging_dir` selectively. The idea is to call the `apply_staging_dir` method from particular creator and then have collected per host which will add it to stagingDir instance.data.

jakubjezek001: > Actually yes, because to me this actually goes to show that get_task_type is not really needed as a method to have it on the Creator plugin to begin with since it's a very thin wrapper to openpype.pipeline.template_data.get_task_type

You are right again. I have found better solution. Thanks for suggesting it ;)

openpype/pipeline/stagingdir.py
jakubjezek001: Not sure what do you mean. can you give some details?

openpype/pipeline/publish/lib.py
jakubjezek001: Are you suggesting we don't need to maintain compatibility for this function?

openpype/pipeline/stagingdir.py
```diff
@@ -0,0 +1,230 @@
...
+
+    # first try to get template data from documents then from names
+    if all([project_doc, asset_doc]):
+        # making fewer queries to database
+        ctx_data = get_template_data(
+            project_doc, asset_doc, task_name, host_name, system_settings
+        )
+    else:
+        ctx_data = get_template_data_with_names(
+            project_name, asset_name, task_name, system_settings
```
Minkiu, line 188: Should we be passing the `host_name` here as well?

openpype/pipeline/tempdir.py
```diff
@@ -3,11 +3,83 @@
...
+        return _create_local_staging_dir(prefix, suffix)
+
+    # make sure anatomy is set
+    if not anatomy:
+        anatomy = Anatomy(project_name)
+
+    # get customized tempdir path from `OPENPYPE_TMPDIR` env var
+    custom_temp_dir = _create_custom_tempdir(
+        anatomy.project_name, anatomy)
+
```
Minkiu, line 60: I think this could be only for both?:
```py
 return os.path.normpath(
            tempfile.mkdtemp(
                prefix=prefix,
                suffix=suffix,
                dir=custom_temp_dir
            )
        )
```

If `_create_custom_tempdir` returns None, so we don't need to do the conditional check.

openpype/pipeline/create/utils.py
```diff
@@ -0,0 +1,122 @@
...
+            with negative value if True. Otherwise None is used. -2 is used
+            for instances without filled asset or subset name. -1 is used
+            for missing entities.
+
+    Returns:
+        dict[str, Union[int, None]]: Last versions by instance id.
+    """
+
+    output = {
+        instance.id: -1 if use_value_for_missing else None
```
Minkiu, line 24: Here we are assuming all of them are missing, right? But later on I can't see where do we check for missing entities? So it could be that we return an instance with `-1` that has entities? i.e. Line 50

openpype/pipeline/__init__.py
```diff
@@ -194,5 +208,5 @@
 
     # Backwards compatible function names
     "install",
-    "uninstall",
+    "uninstall"
```
iLLiCiTiT, line 211: We add them so a future PRs which add after or move it does not have to add commit to the line.

openpype/pipeline/create/creator_plugins.py
```diff
@@ -611,6 +634,58 @@ def get_pre_create_attr_defs(self):
...
+            instance (CreatedInstance): Instance for which should be staging
+                dir applied.
+
+        Returns:
+            str: Path to staging dir.
+        """
+        create_ctx = self.create_context
+        asset_name = instance.get("asset")
+        subset = instance.get("subset")
+        if not any([asset_name, subset]):
```
iLLiCiTiT, line 652: This is actually faster and easier to read.
```suggestion
        if not asset_name or not subset:
```

openpype/pipeline/create/utils.py
```diff
@@ -0,0 +1,122 @@
...
+
+    last_versions = get_last_versions_for_instances(
+        project_name, instances, True)
+
+    output = {}
+    for instance_id, version in last_versions.items():
+        if version == -2:
+            output[instance_id] = None
+        elif version == -1:
+            output[instance_id] = 1
```
iLLiCiTiT, line 119: This has changed recently and requires `get_versioning_start` from `openpype.pipeline.version_start`.

PR #5209, opened 2023-06-28
AYON: Bundle distribution

iLLiCiTiT:
## Changelog Description
Since AYON server 0.3.0 are addon versions defined by bundles which affects how addons, dependency packages and installers are handled. Only source of truth, about any version of anything that should be used, is server bundle.

## Additional info
This approach required to change how we handle distribution and settings. The only information that is defined by server is bundle name, the rest must be done manually by distribution logic. These changes are breaking compatibility with older AYON server versions.

Added logic to run UIs in start logic in separated process to avoid glitches in UI after first login. One UI was added, that is used in cases a bundle that should be used is not available on server, or server does not have any bundle to use. In that case there is nothing to do and the process is killed.

## TODOs
- ~Update tests - the whole logic completelly changed and current tests are 100% wrong.~
- ~Handle cases when there is not any bundle to use from server.~
- ~Call right settings getters.~
- Auto update checks -> check producion/staging bundle changed -> Future PR

## Testing notes:
1. Make sure you have AYON server at least 0.3.1 (0.3.0 had issues with settings).
2. Run OpenPype in AYON mode.
3. Addons should be loaded by the bundle defined on server as production/staging. Or by bundle name defined in `AYON_BUNDLE_NAME` environment variable.
    - if there is any bundle on server marked as production a popup should show
    - if you define `AYON_BUNDLE_NAME` env to non-existing bundle a popup should show
4. Settings should be loaded as expected.
5. Try to logout (Tray > Login > Logout). Then run ayon mode again. Login and there should not be any glitches in UI (if you had any before).

### DIFF
--- a/openpype/vendor/python/common/ayon_api/__init__.py
+++ b/openpype/vendor/python/common/ayon_api/__init__.py
@@ -1,6 +1,8 @@
+from .version import __version__
 from .utils import (
     TransferProgress,
     slugify_string,
+    create_dependency_package_basename,
 )
 from .server_api import (
     ServerAPI,
@@ -183,8 +185,11 @@ from ._api import (
 
 
 __all__ = (
+    "__version__",
+
     "TransferProgress",
     "slugify_string",
+    "create_dependency_package_basename",
 
     "ServerAPI",
 
--- a/openpype/vendor/python/common/ayon_api/server_api.py
+++ b/openpype/vendor/python/common/ayon_api/server_api.py
@@ -61,6 +61,7 @@ from .utils import (
     entity_data_json_default,
     failed_json_default,
     TransferProgress,
+    create_dependency_package_basename,
 )
 
 PatternType = type(re.compile(""))
@@ -115,6 +116,10 @@ class RestApiResponse(object):
         self._data = data
 
     @property
+    def text(self):
+        return self._response.text
+
+    @property
     def orig_response(self):
         return self._response
 
@@ -150,11 +155,13 @@ class RestApiResponse(object):
     def status_code(self):
         return self.status
 
-    def raise_for_status(self):
+    def raise_for_status(self, message=None):
         try:
             self._response.raise_for_status()
         except requests.exceptions.HTTPError as exc:
-            raise HTTPRequestError(str(exc), exc.response)
+            if message is None:
+                message = str(exc)
+            raise HTTPRequestError(message, exc.response)
 
     def __enter__(self, *args, **kwargs):
         return self._response.__enter__(*args, **kwargs)
@@ -1303,13 +1310,15 @@ class ServerAPI(object):
             progress.set_transfer_done()
         return progress
 
-    def _upload_file(self, url, filepath, progress):
+    def _upload_file(self, url, filepath, progress, request_type=None):
+        if request_type is None:
+            request_type = RequestTypes.put
         kwargs = {}
         if self._session is None:
             kwargs["headers"] = self.get_headers()
-            post_func = self._base_functions_mapping[RequestTypes.post]
+            post_func = self._base_functions_mapping[request_type]
         else:
-            post_func = self._session_functions_mapping[RequestTypes.post]
+            post_func = self._session_functions_mapping[request_type]
 
         with open(filepath, "rb") as stream:
             stream.seek(0, io.SEEK_END)
@@ -1320,7 +1329,9 @@ class ServerAPI(object):
         response.raise_for_status()
         progress.set_transferred_size(size)
 
-    def upload_file(self, endpoint, filepath, progress=None):
+    def upload_file(
+        self, endpoint, filepath, progress=None, request_type=None
+    ):
         """Upload file to server.
 
         Todos:
@@ -1331,6 +1342,8 @@ class ServerAPI(object):
             filepath (str): Source filepath.
             progress (Optional[TransferProgress]): Object that gives ability
                 to track upload progress.
+            request_type (Optional[RequestType]): Type of request that will
+                be used to upload file.
         """
 
         if endpoint.startswith(self._base_url):
@@ -1349,7 +1362,7 @@ class ServerAPI(object):
         progress.set_started()
 
         try:
-            self._upload_file(url, filepath, progress)
+            self._upload_file(url, filepath, progress, request_type)
 
         except Exception as exc:
             progress.set_failed(str(exc))
@@ -1486,13 +1499,11 @@ class ServerAPI(object):
         """
 
         response = self.delete("attributes/{}".format(attribute_name))
-        if response.status_code != 204:
-            # TODO raise different exception
-            raise ValueError(
-                "Attribute \"{}\" was not created/updated. {}".format(
-                    attribute_name, response.detail
-                )
+        response.raise_for_status(
+            "Attribute \"{}\" was not created/updated. {}".format(
+                attribute_name, response.detail
             )
+        )
 
         self.reset_attributes_schema()
 
@@ -1732,8 +1743,9 @@ class ServerAPI(object):
         python_version,
         platform_name,
         python_modules,
+        runtime_python_modules,
         checksum,
-        checksum_type,
+        checksum_algorithm,
         file_size,
         sources=None,
     ):
@@ -1742,6 +1754,10 @@ class ServerAPI(object):
         This step will create only metadata. Make sure to upload installer
             to the server using 'upload_installer' method.
 
+        Runtime python modules are modules that are required to run AYON
+            desktop application, but are not added to PYTHONPATH for any
+            subprocess.
+
         Args:
             filename (str): Installer filename.
             version (str): Version of installer.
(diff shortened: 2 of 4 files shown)


common/ayon_common/distribution/file_handler.py
```diff
@@ -112,27 +118,27 @@ def download_url(
...
-                print((
-                    "Failed download. Trying https -> http instead."
-                    f" Downloading {url} to {fpath}"
-                ))
-                RemoteFileHandler._urlretrieve(url, fpath,
-                                               headers=headers)
-            else:
-                raise e
+        except (urllib.error.URLError, IOError) as exc:
+            if url[:5] != "https":
```
Minkiu, line 122: Maybe we could make `_get_redirect_url` return a tuple of `(protocol, url)` -> `("https", "https://ayon.cloud/bla/bleh")` so we can avoid this kind of string slices

common/ayon_common/distribution/file_handler.py
```diff
@@ -155,15 +162,15 @@ def download_file_from_google_drive(file_id, root,
 
         if os.path.isfile(fpath) and RemoteFileHandler.check_integrity(
                 fpath, sha256, hash_type="sha256"):
-            print('Using downloaded and verified file: ' + fpath)
+            print("Using downloaded and verified file: " + fpath)
```
Minkiu, line 165: Might as well use `f{}` or `{}.format()` here?

Minkiu (approved):

Minkiu:
Some methods are missing doc strings which would be good to have, otherwise code looks good.

common/ayon_common/distribution/downloaders.py
```diff
@@ -0,0 +1,244 @@
...
+    @classmethod
+    def download(cls, source, destination_dir, data, transfer_progress):
+        # OS doesn't need to download, unzip directly
+        addon_url = source["path"].get(platform.system().lower())
+        if not os.path.exists(addon_url):
+            raise ValueError(f"{addon_url} is not accessible")
+        return addon_url
+
+    @classmethod
+    def cleanup(cls, source, destination_dir, data):
```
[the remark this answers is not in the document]
iLLiCiTiT: Technically yes, but I would say it's better to have it here as "We really do not want to remove source file".

common/ayon_common/distribution/downloaders.py
```diff
@@ -0,0 +1,244 @@
...
+    @classmethod
+    def unzip(cls, addon_zip_path, destination_dir):
+        """Unzips local 'addon_zip_path' to 'destination'.
+
+        Args:
+            addon_zip_path (str): local path to addon zip file
+            destination_dir (str): local folder to unzip
+        """
+
+        RemoteFileHandler.unzip(addon_zip_path, destination_dir)
```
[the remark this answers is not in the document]
iLLiCiTiT: In that case it must crash ...

common/ayon_common/connection/credentials.py
```diff
@@ -267,12 +275,33 @@ def ask_to_login_ui(
...
         url = get_last_server()
     username = get_last_username_by_url(url)
-    return ask_to_login(url, username, always_on_top=always_on_top)
+    data = {
+        "url": url,
+        "username": username,
+        "always_on_top": always_on_top,
+    }
+
+    with tempfile.TemporaryFile(
```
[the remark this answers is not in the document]
iLLiCiTiT: Much easier way how to pass data to subprocess.

ayon_start.py
```diff
@@ -48,35 +49,55 @@
...
-    "extractenvironments",
-    "version"
-}
 if "--headless" in sys.argv:
+    os.environ["AYON_HEADLESS_MODE"] = "1"
     os.environ["OPENPYPE_HEADLESS_MODE"] = "1"
     sys.argv.remove("--headless")
-elif os.getenv("OPENPYPE_HEADLESS_MODE") != "1":
+
+elif (
```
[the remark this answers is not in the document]
iLLiCiTiT: We do in openpype logic...

common/ayon_common/distribution/file_handler.py
iLLiCiTiT: This is actually for cases when redirect won't happen inside `_get_redirect_url`. So not really...

common/ayon_common/distribution/data_structures.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+class UrlType(Enum):
+    HTTP = "http"
+    GIT = "git"
+    FILESYSTEM = "filesystem"
+    SERVER = "server"
+
+
+@attr.s
```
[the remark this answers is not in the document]
iLLiCiTiT: It was mutable for testing. I think it is not needed anymore, and maybe we should change it to dataclass, but not in this PR. Creating issue...

ayon_start.py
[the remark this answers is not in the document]
Minkiu: Oops, in my mind I though this was mostly for Ayon. You are right heh

common/ayon_common/distribution/file_handler.py
Minkiu: I was referring to do this for line `108` so that then when we reach this point, we already have the protocol in a variable, and we don't need to do `url[:5]` since we already had `urllib` do it for us (it could be empty for all I know)

iLLiCiTiT: I actually don't understand why this is here. It would crash in `_get_redirect_url` if the url would be wrong...

ayon_start.py
```diff
@@ -286,8 +311,44 @@ def _check_and_update_from_ayon_server():
...
+    bundle = None
+    bundle_name = None
+    try:
+        bundle = distribution.bundle_to_use
+        if bundle is not None:
+            bundle_name = bundle.name
+    except BundleNotFoundError as exc:
+        bundle_name = exc.bundle_name
+        if HEADLESS_MODE_ENABLED:
+            _print("!!! Server does not have defined bundle to use.")
```
mkolar, line 323:
```suggestion
            _print("!!! No release bundle is defined on AYON server")
```

ayon_start.py
```diff
@@ -286,8 +311,44 @@ def _check_and_update_from_ayon_server():
...
+            _print(
+                "!!! Make sure server has set bundle."
```
mkolar, line 325: when do we show this notice?

ayon_start.py
```diff
@@ -286,8 +311,44 @@ def _check_and_update_from_ayon_server():
...
+            sys.exit(1)
+
+    if bundle is None:
+        url = get_base_url()
+        if not HEADLESS_MODE_ENABLED:
+            show_missing_bundle_information(url, bundle_name)
+
+        elif bundle_name:
+            _print((
+                f"!!! Requested bundle '{bundle_name}'"
```
mkolar, line 336:
```suggestion
                f"!!! Requested release bundle '{bundle_name}'"
```

ayon_start.py
```diff
@@ -286,8 +311,44 @@ def _check_and_update_from_ayon_server():
...
+                " is not available on server."
+            ))
+            _print(
+                f"!!! Check if server '{url}' still have bundle available.")
```
mkolar, line 340:
```suggestion
                f"!!! Check if selected release bundle is available on the server '{url}'.")
```

common/ayon_common/distribution/ui/missing_bundle_window.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def _update_label(self):
+        self._info_label.setText(self._get_label())
+
+    def _get_label(self):
+        url_part = f" <b>{self._url}</b>" if self._url else ""
+
+        if self._bundle_name:
+            return (
+                f"Requested bundle <b>{self._bundle_name}</b>"
```
mkolar, line 95:
```suggestion
                f"Requested release bundle <b>{self._bundle_name}</b>"
```

common/ayon_common/distribution/ui/missing_bundle_window.py
```diff
@@ -0,0 +1,133 @@
...
+                f" is not available on server{url_part}."
+                "<br/><br/>Try to restart AYON applications. Please"
```
mkolar, line 97:
```suggestion
                "<br/><br/>Try to restart AYON desktop launcher. Please"
```

common/ayon_common/distribution/ui/missing_bundle_window.py
```diff
@@ -0,0 +1,133 @@
...
+                " contact your administrator if issue persist."
+            )
+        return (
+            f"Server{url_part} does not have set bundle to use."
+            "<br/><br/>There is nothing to launch. Please contact your"
+            " administrator to define addons in a bundle."
```
mkolar, line 103:
```suggestion
            f"No release bundle is set as production on the AYON server so"
            "so there is nothing to launch. "
            "Please contact your administrator to resolve the issue."
```

ayon_start.py
```diff
@@ -286,8 +311,44 @@ def _check_and_update_from_ayon_server():
...
+
+        else:
+            _print("!!! Server does not have defined bundle to use.")
```
mkolar, line 343:
```suggestion
            _print("!!! No release bundle is set as production on the AYON server.")
```

ayon_start.py
```diff
@@ -286,8 +311,44 @@ def _check_and_update_from_ayon_server():
...
+            _print(
+                f"!!! Make sure server '{url}' has set bundle."
```
mkolar, line 345:
```suggestion
                f"!!! Make sure there is a release bundle set as \"production"\ on the AYON server '{url}'."
```

mkolar (changes requested):
some wording changes

mkolar (approved):

PR #5212, opened 2023-06-28
Colorspace lib fix compatible python version comparison

### LINKED ISSUE #5211: Bug: Pipeline colorspace "compatible_python" function does not correctly detect python version

### Current Behavior:

The [`compatible_python` function](https://github.com/ynput/OpenPype/blob/82f9da2c8ee4cc1a8062737b8ce3bf99afde73f3/openpype/pipeline/colorspace.py#L240-L245) will currently always return false due to it comparing against `sys.version` which is a string like:
```python
3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:36:42) [MSC v.1929 64 bit (AMD64)]
```

As such a comparison `sys.version[1] >= 9` does not check whether the minor version is higher or equal to nine.

### Expected Behavior:

Function should likely be:

```python
def compatible_python():
    """Only 3.9 or higher can directly use PyOpenColorIO in ocio_wrapper"""
    compatible = False
    if sys.version_info.major == 3 and sys.version_info.minor >= 9:
        compatible = True
    return compatible
```

### Version

3.15.12-nightly.1

Windows

### Steps To Reproduce:

Run Tray Publisher from code on Windows - you'll briefly see a command prompt show up indicating that it's using the Subprocess route for getting OCIO configs + it's relatively slow.

### Relevant log output:

_No response_

### Additional context:

So this PR https://github.com/ynput/OpenPype/pull/4901 in Tray Publisher for me introduces quite a slow down on reset with color management enabled in 3.15.11 - Running from code I can visually see a command prompt pop up so it looks like it's still forcing a subprocess (command prompt?) to open to retrieve the colorspaces. My first thought was: shouldn't tray publisher be able to just get it through the OCIO Python API instead? @jakubjezek001 

Running this in the console of OP tray:
```python
import sys
print(sys.version)
import PyOpenColorIO
print(PyOpenColorIO)
```

Gives me:
```
3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:36:42) [MSC v.1929 64 bit (AMD64)]
<module 'PyOpenColorIO' from 'E:\\openpype\\OpenPype\\.venv\\lib\\site-packages\\PyOpenColorIO.pyd'>
```

So I'd expect it to just go through that in standalone publisher.

Then I noticed this

BigRoy:
## Changelog Description

Fix python version comparison.

## Additional info

Resolves #5211

**Note** that this is just fixing what the code seemed to intend to do - whether the colorspace logic actually works in e.g. 3.10+ or hosts like Maya, Nuke, Fusion, Houdini, etc. with more recent Python versions I'm not sure as I'm not sure whether those actually all have access to the `PyOpenColorIO` package. As such this might need testing across all hosts also whether `import PyOpenColorIO` actually works.

## Testing notes:

Run:
```python
import sys

print(sys.version)
print(sys.version_info)
print(colorspace.compatible_python())
```
Any python 3.9+ should return True on the last print.

**NOTE** We might also want to test across different DCCs whether the check is actually accurate, e.g.:
```python
try:
    import PyOpenColorIO
    ocio = True
except Exception:
    ocio = False
compatible = colorspace.compatible_python()
print("Compatible python function:", compatible)
print("Import PyOpenColorIO works:", ocio)
print(ocio == compatible)
```
Which makes me think the compatibility check might be easier including the import check itself as well to remain fully accurate?

### DIFF
--- a/openpype/pipeline/colorspace.py
+++ b/openpype/pipeline/colorspace.py
@@ -240,7 +240,7 @@ def get_data_subprocess(config_path, data_type):
 def compatible_python():
     """Only 3.9 or higher can directly use PyOpenColorIO in ocio_wrapper"""
     compatible = False
-    if sys.version[0] == 3 and sys.version[1] >= 9:
+    if sys.version_info.major == 3 and sys.version_info.minor >= 9:
         compatible = True
     return compatible
 


jakubjezek001 (approved):
LGTM

BigRoy:
I'm pretty sure with merging this PR so quickly you've broken some hosts.

**E.g. this in Maya 2024:**
```python
import PyOpenColorIO as ocio
```
Does not work, and thus even though it's now considered `compatible_python` it won't work with the colorspace functions.
```
# Error: DLL load failed while importing PyOpenColorIO: The specified module could not be found.
# #   File "<maya console>", line 1, in <module>
# # ImportError: DLL load failed while importing PyOpenColorIO: The specified module could not be found.
```

**Maya 2023**
```python
import PyOpenColorIO as ocio
```
Seems fine

@jakubjezek001 Did you do the test inside hosts as well?

Why isn't the compatible check just this:
```python
def compatible_python():
    try:
        import PyOpenColorIO
    except (ImportError, ModuleNotFoundError):
        return False
    return True
```

jakubjezek001:
Ah, yeah sorry I am just complete ignorant and looked only into the code. Thanks for poking me @BigRoy I will make PR with fix.

PR #5215, opened 2023-06-29
Bug: Maya - xgen sidecar files arent moved when saving workfile as an new asset workfile changing context - OP-6222

tokejepsen:
## Changelog Description
This PR manages the Xgen files when switching context in the Workfiles app.

## Testing notes:
1. Setup Xgen in Maya.
2. Switch context and validate xgen files are copied to new location.
3. Repeat steps 1-2 and validate overwrite dialog appears.
4. Load in Xgen in Maya.
5. Switch context and validate xgen still works after relaunch of Maya.

### DIFF
--- a/openpype/hosts/maya/api/pipeline.py
+++ b/openpype/hosts/maya/api/pipeline.py
@@ -712,6 +712,20 @@ def display_warning(message, show_cancel=False):
 
 
 def workfile_save_before_xgen(event):
+    """Manage Xgen external files when switching context.
+
+    Xgen has various external files that needs to be unique and relative to the
+    workfile, so we need to copy and potentially overwrite these files when
+    switching context.
+
+    Args:
+        event (Event) - openpype/lib/events.py
+    """
+    if not cmds.pluginInfo("xgenToolkit", query=True, loaded=True):
+        return
+
+    import xgenm
+
     current_work_dir = legacy_io.Session["AVALON_WORKDIR"].replace("\\", "/")
     expected_work_dir = event.data["workdir_path"].replace("\\", "/")
     if current_work_dir == expected_work_dir:
@@ -721,14 +735,13 @@ def workfile_save_before_xgen(event):
     if not palettes:
         return
 
-    import xgenm
-
     transfers = []
     overwrites = []
     attribute_changes = {}
     attrs = ["xgFileName", "xgBaseFile"]
     for palette in palettes:
-        project_path = xgenm.getAttr("xgProjectPath", palette.replace("|", ""))
+        sanitized_palette = palette.replace("|", "")
+        project_path = xgenm.getAttr("xgProjectPath", sanitized_palette)
         _, maya_extension = os.path.splitext(event.data["filename"])
 
         for attr in attrs:
@@ -743,7 +756,7 @@ def workfile_save_before_xgen(event):
             attr_value = event.data["filename"].replace(
                 maya_extension,
                 "__{}{}".format(
-                    palette.replace("|", "").replace(":", "__"),
+                    sanitized_palette.replace(":", "__"),
                     os.path.splitext(attr_value)[1]
                 )
             )
@@ -753,7 +766,7 @@ def workfile_save_before_xgen(event):
             attribute_changes[node_attr] = attr_value
 
         relative_path = xgenm.getAttr(
-            "xgDataPath", palette.replace("|", "")
+            "xgDataPath", sanitized_palette
         ).split(os.pathsep)[0]
         absolute_path = relative_path.replace("${PROJECT}", project_path)
         for root, _, files in os.walk(absolute_path):
@@ -765,23 +778,23 @@ def workfile_save_before_xgen(event):
                     overwrites.append(target)
 
     # Ask user about overwriting files.
-    msg = (
-        "WARNING! Potential loss of data.\n\n"
-        "Found duplicate Xgen files in new context.\n"
-        "Do you want to overwrite?\n\n{}".format("\n".join(overwrites))
-    )
     if overwrites:
+        msg = (
+            "WARNING! Potential loss of data.\n\n"
+            "Found duplicate Xgen files in new context.\n"
+            "Do you want to overwrite?\n\n{}".format("\n".join(overwrites))
+        )
         accept = display_warning(msg, show_cancel=True)
         if not accept:
             return
-
-    for attribute, value in attribute_changes.items():
-        cmds.setAttr(attribute, value, type="string")
 
     for source, destination in transfers:
         if not os.path.exists(os.path.dirname(destination)):
             os.makedirs(os.path.dirname(destination))
         shutil.copy(source, destination)
+
+    for attribute, value in attribute_changes.items():
+        cmds.setAttr(attribute, value, type="string")
 
 
 def after_workfile_save(event):
--- a/openpype/hosts/maya/api/pipeline.py
+++ b/openpype/hosts/maya/api/pipeline.py
@@ -686,30 +686,6 @@ def before_workfile_save(event):
         create_workspace_mel(workdir_path, project_name)
 
 
-def prompt_warning(message, show_cancel=False):
-    """Show feedback to user.
-
-    Returns:
-        bool
-    """
-
-    accept = QtWidgets.QMessageBox.Ok
-    if show_cancel:
-        buttons = accept | QtWidgets.QMessageBox.Cancel
-    else:
-        buttons = accept
-
-    state = QtWidgets.QMessageBox.warning(
-        None,
-        "",
-        message,
-        buttons=buttons,
-        defaultButton=accept
-    )
-
-    return state == accept
-
-
 def workfile_save_before_xgen(event):
     """Manage Xgen external files when switching context.
 
@@ -778,14 +754,13 @@ def workfile_save_before_xgen(event):
 
     # Ask user about overwriting files.
     if overwrites:
-        msg = (
+        log.warning(
             "WARNING! Potential loss of data.\n\n"
-            "Found duplicate Xgen files in new context.\n"
-            "Do you want to overwrite?\n\n{}".format("\n".join(overwrites))
+            "Found duplicate Xgen files in new context.\n{}".format(
+                "\n".join(overwrites)
+            )
         )
-        accept = prompt_warning(msg, show_cancel=True)
-        if not accept:
-            return
+        return
 
     for source, destination in transfers:
         if not os.path.exists(os.path.dirname(destination)):
--- a/openpype/hosts/maya/api/pipeline.py
+++ b/openpype/hosts/maya/api/pipeline.py
@@ -8,7 +8,6 @@ from maya import utils, cmds, OpenMaya
 import maya.api.OpenMaya as om
 
 import pyblish.api
-from qtpy import QtWidgets
 
 from openpype.settings import get_project_settings
 from openpype.host import (
--- a/openpype/hosts/maya/api/pipeline.py
+++ b/openpype/hosts/maya/api/pipeline.py
@@ -8,6 +8,7 @@ from maya import utils, cmds, OpenMaya
 import maya.api.OpenMaya as om
 
 import pyblish.api
+from qtpy import QtWidgets
 
 from openpype.settings import get_project_settings
 from openpype.host import (
@@ -691,9 +692,6 @@ def prompt_warning(message, show_cancel=False):
         Returns:
             bool
         """
-
-    from qtpy import QtWidgets
-
     accept = QtWidgets.QMessageBox.Ok
     if show_cancel:
         buttons = accept | QtWidgets.QMessageBox.Cancel
--- a/openpype/hosts/maya/api/pipeline.py
+++ b/openpype/hosts/maya/api/pipeline.py
@@ -685,7 +685,7 @@ def before_workfile_save(event):
         create_workspace_mel(workdir_path, project_name)
 
 
-def display_warning(message, show_cancel=False):
+def prompt_warning(message, show_cancel=False):
     """Show feedback to user.
 
         Returns:
@@ -784,7 +784,7 @@ def workfile_save_before_xgen(event):
             "Found duplicate Xgen files in new context.\n"
             "Do you want to overwrite?\n\n{}".format("\n".join(overwrites))
         )
-        accept = display_warning(msg, show_cancel=True)
+        accept = prompt_warning(msg, show_cancel=True)
         if not accept:
             return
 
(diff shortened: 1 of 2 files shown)


openpype/hosts/maya/api/pipeline.py
```diff
@@ -681,6 +685,105 @@ def before_workfile_save(event):
...
+        return
+
+    import xgenm
+
+    transfers = []
+    overwrites = []
+    attribute_changes = {}
+    attrs = ["xgFileName", "xgBaseFile"]
+    for palette in palettes:
+        project_path = xgenm.getAttr("xgProjectPath", palette.replace("|", ""))
```
BigRoy, line 731: I see a lot of `palette.replace("|", "")` - not sure why it;s needed for xgen but I feel like we can do that early once and have the rest of the code be more readable. Or is there a reason we're not doing that?

E.g.
```python
palette_attr = palette.replace("|", "")
palette_fname = palette_attr.replace(":", "__")
```

openpype/hosts/maya/api/pipeline.py
```diff
@@ -681,6 +685,105 @@ def before_workfile_save(event):
...
+        "",
+        message,
+        buttons=buttons,
+        defaultButton=accept
+    )
+
+    return state == accept
+
+
+def workfile_save_before_xgen(event):
```
BigRoy, line 714: Can we add a docstring as to what this does and why it's needed?

openpype/hosts/maya/api/pipeline.py
```diff
@@ -681,6 +685,105 @@ def before_workfile_save(event):
...
+    current_work_dir = legacy_io.Session["AVALON_WORKDIR"].replace("\\", "/")
+    expected_work_dir = event.data["workdir_path"].replace("\\", "/")
+    if current_work_dir == expected_work_dir:
+        return
+
+    palettes = cmds.ls(type="xgmPalette", long=True)
+    if not palettes:
+        return
+
+    import xgenm
```
BigRoy, line 724: Can we add a comment as to why we're importing no earlier than here - is it due to slowness? or required plugin load?

openpype/hosts/maya/api/pipeline.py
```diff
@@ -681,6 +685,105 @@ def before_workfile_save(event):
...
+    if overwrites:
+        accept = display_warning(msg, show_cancel=True)
+        if not accept:
+            return
+
+    for attribute, value in attribute_changes.items():
+        cmds.setAttr(attribute, value, type="string")
+
+    for source, destination in transfers:
+        if not os.path.exists(os.path.dirname(destination)):
```
BigRoy, line 782: Should we do the transfering _before_ changing the scene attributes to leave the scene state the same if for some reason the copying fails? (e.g. destination is locked or no disk space, etc.)

openpype/hosts/maya/api/pipeline.py
```diff
@@ -681,6 +685,105 @@ def before_workfile_save(event):
...
+    return state == accept
+
+
+def workfile_save_before_xgen(event):
+    current_work_dir = legacy_io.Session["AVALON_WORKDIR"].replace("\\", "/")
+    expected_work_dir = event.data["workdir_path"].replace("\\", "/")
+    if current_work_dir == expected_work_dir:
+        return
+
+    palettes = cmds.ls(type="xgmPalette", long=True)
```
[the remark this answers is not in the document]
tokejepsen: Yup, good point.

openpype/hosts/maya/api/pipeline.py
```diff
@@ -681,6 +685,105 @@ def before_workfile_save(event):
...
+                transfers.append((source, target))
+                if os.path.exists(target):
+                    overwrites.append(target)
+
+    # Ask user about overwriting files.
+    msg = (
+        "WARNING! Potential loss of data.\n\n"
+        "Found duplicate Xgen files in new context.\n"
+        "Do you want to overwrite?\n\n{}".format("\n".join(overwrites))
+    )
```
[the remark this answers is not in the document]
tokejepsen: 👍

openpype/hosts/maya/api/pipeline.py
```diff
@@ -681,6 +685,105 @@ def before_workfile_save(event):
...
+        accept = display_warning(msg, show_cancel=True)
+        if not accept:
+            return
+
+    for attribute, value in attribute_changes.items():
+        cmds.setAttr(attribute, value, type="string")
+
+    for source, destination in transfers:
+        if not os.path.exists(os.path.dirname(destination)):
+            os.makedirs(os.path.dirname(destination))
```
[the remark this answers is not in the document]
tokejepsen: Yeah, still have Python 2 AFAIK.

openpype/hosts/maya/api/pipeline.py
tokejepsen: `xgenm` does not like the `|` Maya path symbol, but its needed for unique path when fetching the node, but will do the xgenm sanitization once instead.

openpype/hosts/maya/api/pipeline.py
tokejepsen: 👍

openpype/hosts/maya/api/pipeline.py
tokejepsen: For ensuring the module is available, so will move to just below the plugin load check.

openpype/hosts/maya/api/pipeline.py
tokejepsen: Yeah, good idea.

LiborBatek (approved):
All seems working well!

I changed context without any issues with xgen folder/files. I did it multiple times too so there popped up Warning with possible data loss because of overwritting xgen files if already present some at the destination location - all good too!

I have loaded the Xgen published asset on to character in some anim/fx shot. Changed context to different shot and restarted Maya again to see if any problems. Not any.
I have double checked multiple time even the orig workfiles if any issues but not any.

All working well.

tokejepsen:
@BigRoy you happy with the code?

openpype/hosts/maya/api/pipeline.py
```diff
@@ -681,6 +685,118 @@ def before_workfile_save(event):
         create_workspace_mel(workdir_path, project_name)
 
 
+def display_warning(message, show_cancel=False):
```
BigRoy, line 688: Display warning doesn't really hint that it's to request an accept or cancel from the end user. The "Show feedback to user" also doesn't explain that in the docstring either. This is more of a prompt than a display.

BigRoy (commented):
Code looks good to me - only had one remaining comment on the choice of words for the `display_warning` function. It sounds incorrect? Also not too big of a fan of having it clutter up `pipeline.py` with something so mundane and likely to be used across different hosts.

@iLLiCiTiT - how do you feel about it raising a prompt to begin with here? Is that ok? 

It might be worrisome from a pure CLI/automation perspective if e.g. running maya in headless mode too.

openpype/hosts/maya/api/pipeline.py
tokejepsen: Changing name.

tokejepsen:
Will await @iLLiCiTiT feedback on where and how to display the warning information to the user.

openpype/hosts/maya/api/pipeline.py
```diff
@@ -681,6 +685,118 @@ def before_workfile_save(event):
         create_workspace_mel(workdir_path, project_name)
 
 
+def prompt_warning(message, show_cancel=False):
+    """Show feedback to user.
+
+        Returns:
+            bool
+        """
```
iLLiCiTiT, line 693:
```suggestion
    Returns:
        bool
    """
```

openpype/hosts/maya/api/pipeline.py
```diff
@@ -681,6 +685,118 @@ def before_workfile_save(event):
...
+
+    from qtpy import QtWidgets
```
iLLiCiTiT, line 695: Can be at the top.

iLLiCiTiT:
> It might be worrisome from a pure CLI/automation perspective if e.g. running maya in headless mode too.

I agree that headless mode must be handled in some way, auto-accept or just crash, no dialog.

> Code looks good to me - only had one remaining comment on the choice of words for the display_warning function. It sounds incorrect?

I have mixed thoughs about it. with cancel button it is warning+question at the same time. Not a fan, but not against. (Potential bug: It should always return `True` if there is not a cancel button?).

> Also not too big of a fan of having it clutter up pipeline.py with something so mundane and likely to be used across different hosts.

I do agree that there should be a function that would allow to show dialog easily, but that should probably be more complex than this one, to not block this PR I would keep it as it is now. The function could be elsewhere, but it's not draining my soul.

There are many ways how you may want the dialog to behave. Change of message type (question, warning, error),  dialog type (non modal/modal), etc.  We should prepare the logic in a way that we would not need to touch it for few months (or ever). We have to be very carefull about code in global (openpype core), every change there, will affect every addon/host (and it is not change only in OpenPype repo anymore).

tokejepsen:
@iLLiCiTiT @BigRoy what changes do you want on this?

openpype/hosts/maya/api/pipeline.py
```diff
@@ -681,6 +686,116 @@ def before_workfile_save(event):
...
+    # Ask user about overwriting files.
+    if overwrites:
+        msg = (
+            "WARNING! Potential loss of data.\n\n"
+            "Found duplicate Xgen files in new context.\n"
+            "Do you want to overwrite?\n\n{}".format("\n".join(overwrites))
+        )
+        accept = prompt_warning(msg, show_cancel=True)
+        if not accept:
+            return
```
iLLiCiTiT, line 788:
```suggestion
        if not lib.IS_HEADLESS:
            accept = prompt_warning(msg, show_cancel=True)
            if not accept:
                return
        else:
            print(msg)
```

tokejepsen: I'm up for overwriting by default, but just wanna make sure we consider other options.

If we raise an error, it would be up to the users to figure out why there are duplicate Xgen files and manage the files themselves.
If we built in some mechanism for accepting overwriting, it would have to come through some environment variable I think cause this is coming from an event emitted.

We happy with overwriting by default?

BigRoy: I'd likely do the opposite in headless mode and default to not overwriting, just due to safety concerns. But I'm not against this proposal either in headless mode for the time being. (As long as we are entirely sure that the folder and files we're touching are REALLY only xgen files and not by change accidentally the root path of a project or whatever, hehe) However, I'd use `log.warning` and also explain that we'll auto-overwrite it due to being in headless/batch mode so it's super clear what's going on. I had even considered that IF it overwrites that maybe it should log which files it's overwriting so the logs make clear that this process touched those files.

tokejepsen: Yeah, think we should maybe opt for caution here and raise an error explaining which files are in conflict. @iLLiCiTiT are you good with this?

iLLiCiTiT: It is just quick mock-up, I don't know the consequences. Question is if this can be somehow triggered in headless mode? It is just event callback and the even is triggered only by workflies tool. Maybe raising an exception is right way in this case?

BigRoy: > Good point! This only gets triggered from workfiles tool which is always a UI. @BigRoy what do you think of not accounting for headless and just prompt the user, aka discard this suggestion?

Doesn't that just point to a bug that it does not work without workfiles UI? It should, right?
You should instead be using event `before.save` right?

tokejepsen: > Doesn't that just point to a bug that it does not work without workfiles UI? It should, right?
You should instead be using event before.save right?

We'll need to know where the user is intending on saving the workfile to, because we need to move xgen files and modify attribute before saving.
I also think that modifying the behaviour of Maya native saving with xgen files management will limit the users too much, ei. they cant escape the xgen files management while Maya in launched through the pipeline.

openpype/hosts/maya/api/pipeline.py
```diff
@@ -681,6 +686,116 @@ def before_workfile_save(event):
...
+    if show_cancel:
+        buttons = accept | QtWidgets.QMessageBox.Cancel
+    else:
+        buttons = accept
+
+    state = QtWidgets.QMessageBox.warning(
+        None,
+        "",
+        message,
+        buttons=buttons,
```
Minkiu, line 706: Not sure if fond of ternary, but this could be like so:
```py
buttons = accept | QtWidgets.QMessageBox.Cancel if show_cancel else accept
```

openpype/hosts/maya/api/pipeline.py
```diff
@@ -681,6 +686,116 @@ def before_workfile_save(event):
...
+            attribute_changes[node_attr] = attr_value
+
+        relative_path = xgenm.getAttr(
+            "xgDataPath", sanitized_palette
+        ).split(os.pathsep)[0]
+        absolute_path = relative_path.replace("${PROJECT}", project_path)
+        for root, _, files in os.walk(absolute_path):
+            for f in files:
+                source = os.path.join(root, f).replace("\\", "/")
+                target = source.replace(project_path, expected_work_dir + "/")
```
Minkiu, line 774: Wouldn't this achieve the same?
```py
target = os.path.join(expected_work_dir, f).replace("\\", "/")
```

tokejepsen: Not quite, cause there can be subfolders between the `expected_work_dir` and the file.

openpype/hosts/maya/api/pipeline.py
tokejepsen: That is one mighty long line :)

Got a suggestion to break that up for line limit?

BigRoy:
```python
buttons = (
    QtWidgets.QMessageBox.Ok | QtWidgets.QMessageBox.Cancel 
    if show_cancel else QtWidgets.QMessageBox.Ok
)
```
Like that?

tokejepsen: @Minkiu does BigRoy suggestion work for you?

BigRoy (approved):
Didn't test the code, but the latest changes with regards to logging instead of overwriting seems fine for a first prototype. Nice work!

LiborBatek (approved):
There is no more Warning popup window when changing context of the workfile with Xgen present.

Only Warning of possible data loss in console...

image

PR #5221, opened 2023-06-30
Fix GDEFORM collection creation while first build workfile

BlueLag:
## Changelog Description
Fix an issue for gdeform collection which were created when we first build a workfile even if there are no characters with gdeformer modifiers in it. Fixed by adding a condition for the creation of the gdeform collection.

## Additional info
n/a

## Testing notes:
1. Open a ww's scene with OpenPype launcher in blender
2. In the OpenPype menu, do a first build workfile
3. Check in the blender outliner
- If the scene do not have a character, it should not have any GDEFORM collection
- If the scene do have a character with GDEFORM, the GDEFORM collection should be created and every GDEFORM should be linked to it

### DIFF
--- a/openpype/hosts/blender/api/action.py
+++ b/openpype/hosts/blender/api/action.py
@@ -96,3 +96,21 @@ class UpdateContainer(pyblish.api.Action):
         for action, containers in update_actions:
             if containers:
                 action().process(containers)
+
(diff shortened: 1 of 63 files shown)


BlueLag:
Sorry my Pr wasn't on the right repo :/

PR #5223, opened 2023-06-30
Colorspace: check PyOpenColorIO rather then python version

jakubjezek001:
## Changelog Description
Fixing previously merged PR (https://github.com/ynput/OpenPype/pull/5212) And applying better way to check compatibility with PyOpenColorIO python api.

## Additional info
- implementing suggestion from https://github.com/ynput/OpenPype/pull/5212#issuecomment-1614651292

## Testing notes:
1. Colorspace features should now work again in any host

### DIFF
--- a/openpype/pipeline/colorspace.py
+++ b/openpype/pipeline/colorspace.py
@@ -240,7 +240,7 @@ def compatibility_check():
     """Making sure PyOpenColorIO is importable"""
     try:
         import PyOpenColorIO  # noqa: F401
-    except (ImportError, ModuleNotFoundError):
+    except ImportError:
         return False
     return True
 
--- a/openpype/pipeline/colorspace.py
+++ b/openpype/pipeline/colorspace.py
@@ -1,7 +1,6 @@
 from copy import deepcopy
 import re
 import os
-import sys
 import json
 import platform
 import contextlib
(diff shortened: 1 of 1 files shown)


openpype/pipeline/colorspace.py
```diff
@@ -237,12 +236,13 @@ def get_data_subprocess(config_path, data_type):
...
-    """Only 3.9 or higher can directly use PyOpenColorIO in ocio_wrapper"""
-    compatible = False
-    if sys.version_info.major == 3 and sys.version_info.minor >= 9:
-        compatible = True
-    return compatible
+def compatibility_check():
+    """Making sure PyOpenColorIO is importable"""
+    try:
+        import PyOpenColorIO  # noqa: F401
+    except (ImportError, ModuleNotFoundError):
```
BigRoy, line 243: It seems [ModuleNotFoundError is a subclass of ImportError](https://docs.python.org/3/library/exceptions.html#ModuleNotFoundError) so we don't need it.
Which is nice because it also only [got added since Py3.6](https://docs.python.org/3/library/exceptions.html#ModuleNotFoundError) which would break compatibility for older hosts. :)
```suggestion
    except ImportError:
```

BigRoy (approved):
A bit rushed today so can't test myself - but the actual check seems much more sensible. :)

jakubjezek001:
I tested in Nuke 12 and TrayPublisher, both works as expected.

tokejepsen:
Not 100 % sure what to test here, but running this code;

```


print(colorspace.compatibility_check())
```

Maya 2023: `True`
Maya 2024: `False`

BigRoy:
That does seem to be solid. 👍

tokejepsen (approved):

PR #5225, opened 2023-06-30
Maya: fix save file prompt on launch last workfile with color management enabled + restructure `set_colorspace`

BigRoy:
## Changelog Description

- Only set `configFilePath` when OCIO env var is not set since it doesn't do anything if OCIO var is set anyway.
- Set the Maya 2022+ default OCIO path using the resources path instead of "" to avoid Maya Save File on new file after launch
   - **Bugfix: This is what fixes the Save prompt on open last workfile feature with Global color management enabled**
- Move all code related to applying the maya settings together after querying the settings
- Swap around the `if use_workfile_settings` since the check was reversed
- Use `get_current_project_name()` instead of environment vars

## Additional info

**Please note** that with the default settings **and** global color management enabled that you will after this PR now get logged errors on New Scene due to the default settings with color management setting OCIO configs using the `OCIO` env var:

image

However the default colorspaces for Maya do not existing in those configs (but they are based on the default OCIO configs of Maya):

image

Before this PR those values ALSO wouldn't be set however no error would have been logged due to the `colorManagementPrefs` config path to be set to Maya's default JUST before those calls and maya finding the views and displays there - however the OCIO env var takes precedence over it and thus overrides those prefs - and with that thus resets those settings again.

To avoid this you'll need to tweak the settings from the second screenshot into colorspaces that actually exist in the one from OpenPype.

## Testing notes:

1. Enable the global color management setting `project_settings/global/imageio/activate_global_color_management`
2. Open Maya 2022 and 2023 and 2024 with last workfile - it should not prompt to save the empty file for before opening the file on launch.
3. Color management should continue to work as expected in maya
    - As such, it might be good to also publish a look with some color management

Note: All this should also still continue to work with `project_settings/global/imageio/activate_global_color_management` disabled.

### DIFF
--- a/openpype/hosts/maya/api/lib.py
+++ b/openpype/hosts/maya/api/lib.py
@@ -3238,36 +3238,21 @@ def iter_shader_edits(relationships, shader_nodes, nodes_by_id, label=None):
 
 
 def set_colorspace():
-    """Set Colorspace from project configuration
-    """
+    """Set Colorspace from project configuration"""
 
-    # set color spaces for rendering space and view transforms
-    def _colormanage(**kwargs):
-        """Wrapper around `cmds.colorManagementPrefs`.
-
-        This logs errors instead of raising an error so color management
-        settings get applied as much as possible.
-
-        """
-        assert len(kwargs) == 1, "Must receive one keyword argument"
-        try:
-            cmds.colorManagementPrefs(edit=True, **kwargs)
-            log.debug("Setting Color Management Preference: {}".format(kwargs))
-        except RuntimeError as exc:
-            log.error(exc)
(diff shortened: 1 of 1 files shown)


LiborBatek (commented):
Seems working so far.
No errors during launch of Maya 2022. 2023 (need to install and test 20, 2024 tho)

Also no Saving prompts whatsoever...

image

When Global CMS is turned off it uses defaults of Maya which is ok

image

LiborBatek (approved):
All have been tested in maya 2022, 2023, 2024 and works ok.

Also did try to publish look with CMS enabled successfully.
image

PR #5226, opened 2023-06-30
Maya: Minor refactoring and code cleanup

BigRoy:
## Changelog Description

Some small cleanup and refactoring of logic. Removing old comments, unused imports and some minor optimization. Also removed the prints of the loader names of each container the scene in `fix_incompatible_containers` + optimizing by using `set` and defining only once. Moved some UI related code/tweaks to run `on_init` only if not in headless mode. Removed an empty `obj.py` file.

Each commit message kind of describes why the change was made.

## Additional info

Even though these are mostly cosmetic changes they might introduce subtle bugs (of course). So do a quick test run please.

## Testing notes:

1. Maya Look assigner should still open
2. Maya Creator UI should still work
3. Render Setup observers should still work, create a render instance - ensure it correctly catches adding new renderlayers or removing old ones and test the publishing of it.
4. Extracting `.obj` should still work.

### DIFF
--- a/openpype/hosts/maya/api/lib.py
+++ b/openpype/hosts/maya/api/lib.py
@@ -2811,19 +2811,22 @@ def get_attr_in_layer(attr, layer):
 
 def fix_incompatible_containers():
     """Backwards compatibility: old containers to use new ReferenceLoader"""
-
+    old_loaders = {
+        "MayaAsciiLoader",
+        "AbcLoader",
+        "ModelLoader",
+        "CameraLoader",
+        "RigLoader",
+        "FBXLoader"
+    }
     host = registered_host()
     for container in host.ls():
         loader = container['loader']
-
-        print(container['loader'])
-
-        if loader in ["MayaAsciiLoader",
-                      "AbcLoader",
-                      "ModelLoader",
-                      "CameraLoader",
-                      "RigLoader",
-                      "FBXLoader"]:
+        if loader in old_loaders:
+            log.info(
+                "Converting legacy container loader {} to "
+                "ReferenceLoader: {}".format(loader, container["objectName"])
+            )
             cmds.setAttr(container["objectName"] + ".loader",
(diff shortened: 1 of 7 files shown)


LiborBatek (commented):
Extracting model as OBJ works fine,
Creator can be normally opened/used,
Look assigner also works fine, 
Render instance got created and renderLayer also. 

However I am experiencing weird errors regarding file writting to disk (in this particular case is Look publish related (Extract Look) but I also seen it while publishing model...it helped to save and reopen the workfile but not for Extract Look case...

image

Not sure if related to this PR tho...

BigRoy:
That seems like a very odd place to save that file which usually ends up in stagingdir / temporary directory.
Do you have Custom Staging Directories enabled maybe? That'd be this feature: https://github.com/ynput/OpenPype/pull/4583

Could you try and reproduce the issue with latest `develop` branch?

LiborBatek:
> That seems like a very odd place to save that file which usually ends up in stagingdir / temporary directory. Do you have Custom Staging Directories enabled maybe? That'd be this feature: #4583
> 
> Could you try and reproduce the issue with latest `develop` branch?

Yeah u are right! I have defined custom staging dir as ```{root[work]}/{project[name]}/{hierarchy}/{asset}/work/{family}/{subset}```

and had profile for it as ```transient``` folder.

so I will disable custom staging directory.

BigRoy:
> Yeah u are right! I have defined custom staging dir as {root[work]}/{project[name]}/{hierarchy}/{asset}/work/{family}/{subset}

This does point to a bug/error with custom staging dir feature @kalisp - could you try and reproduce it in `develop` @LiborBatek and report it as a dedicated issue so we can track and solve it.

LiborBatek (approved):
All good!

Did successfuly all test from the list without any issues.

mkolar (approved):

Issue #4870, opened 2023-04-19
Enhancement: Using switch on a Loader which doesn't support it should provide better error message


BigRoy:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

(Tested in Houdini - however I suspect the error popup is universal)

Whenever a Loader does not implement the `switch` method then performing a Scene Inventory / Manage > Switch Asset on that loader will show an error like:

![image (3)](https://user-images.githubusercontent.com/2439881/233104272-41c1d829-2631-4c94-92e0-a7a21d9a55b1.png)

Note how the LOG is much clearer than the error message even though this is a perfect error that can be more informative to the artist directly without needing to consult logs.

### How would you imagine the implementation of the feature?

The base loader plugin could just raise a dedicated `SwitchNotImplementedError` by default from `switch` so that if it's raised and thus the plugins don't override the method then we can catch it and show a better error pop-up from the UI?

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

[cuID:[OP-5758](https://app.clickup.com/t/861mp7mtn)]

BigRoy:
This should be solved with #5227

PR #5227, opened 2023-07-02
Enhancement: More descriptive error messages for Loaders

BigRoy:
## Changelog Description

Tweak raised errors and error messages for loader errors.

## Additional info

Intends to solve #4870 

## Testing notes:

1. Try and hit the loader error cases, make sure the error message is clear.
   - perform switch asset on a loader that does not have `switch` implemented
   - change the 'loader' name on a maya container (e.g. change the attribute value to 'foobar') to something non-existent, loader messages should be clear when trying to interact with it.

### DIFF
--- a/openpype/tools/sceneinventory/switch_dialog.py
+++ b/openpype/tools/sceneinventory/switch_dialog.py
@@ -19,6 +19,9 @@ from openpype.pipeline.load import (
     switch_container,
     get_repres_contexts,
     loaders_from_repre_context,
+    LoaderSwitchNotImplementedError,
+    IncompatibleLoaderError,
+    LoaderNotFoundError
 )
 
 from .widgets import (
@@ -1300,7 +1303,18 @@ class SwitchAssetDialog(QtWidgets.QDialog):
 
             try:
                 switch_container(container, repre_doc, loader)
-            except Exception:
+            except Exception as exc:
+                if isinstance(exc, (LoaderSwitchNotImplementedError,
+                                    IncompatibleLoaderError,
+                                    LoaderNotFoundError)):
+                    # Show error message directly
+                    text = str(exc)
+                else:
+                    text = (
+                        "Switch asset failed. "
+                        "Search console log for more details."
+                    )
+
                 msg = (
                     "Couldn't switch asset."
                     "See traceback for more information."
@@ -1308,9 +1322,7 @@ class SwitchAssetDialog(QtWidgets.QDialog):
                 log.warning(msg, exc_info=True)
                 dialog = QtWidgets.QMessageBox(self)
                 dialog.setWindowTitle("Switch asset failed")
-                dialog.setText(
-                    "Switch asset failed. Search console log for more details"
-                )
+                dialog.setText(text)
                 dialog.exec_()
 
         self.switched.emit()
--- a/openpype/tools/sceneinventory/switch_dialog.py
+++ b/openpype/tools/sceneinventory/switch_dialog.py
@@ -1301,28 +1301,28 @@ class SwitchAssetDialog(QtWidgets.QDialog):
                 else:
                     repre_doc = repres_by_name[container_repre_name]
 
+            error = None
             try:
                 switch_container(container, repre_doc, loader)
-            except Exception as exc:
-                if isinstance(exc, (LoaderSwitchNotImplementedError,
-                                    IncompatibleLoaderError,
-                                    LoaderNotFoundError)):
-                    # Show error message directly
-                    text = str(exc)
-                else:
-                    text = (
-                        "Switch asset failed. "
-                        "Search console log for more details."
-                    )
-
-                msg = (
+            except (
+                LoaderSwitchNotImplementedError,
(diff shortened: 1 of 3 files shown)


openpype/tools/sceneinventory/switch_dialog.py
```diff
@@ -1300,17 +1303,26 @@ def _trigger_switch(self, loader=None):
 
             try:
                 switch_container(container, repre_doc, loader)
-            except Exception:
+            except Exception as exc:
+                if isinstance(exc, (LoaderSwitchNotImplementedError,
```
iLLiCiTiT, line 1307: Maybe there should be one exception that is used as base for the others?

iLLiCiTiT: So this code would be
```
try:
    switch_container(container, repre_doc, loader)
except SomeException as exc:
    text = str(exc)
except Exception:
    text = (
        "Switch asset failed. "
        "Search console log for more details."
    )

Comment (BigRoy) [reply openpype/tools/sceneinventory/switch_dialog.py:1307]: We could - but in your code example, both exceptions would still need to result in the full pop-up dialog (duplicating the code). Right? That's why I wrote it this way to re-use the logic.

Also, I didn't inherit from the same errors because some were previously RuntimeErrors and others were ValueErrors and making them now inherit from the same errors would mean that any try..except someone might have used to capture the previous RuntimeErrors or ValueErrors might then not catch the new errors. So I've tried to inherit from the original errors where I could. (Except for the AssertionError that was there)

Comment (iLLiCiTiT) [reply openpype/tools/sceneinventory/switch_dialog.py:1307]: > both exceptions would still need to result in the full pop-up dialog (duplicating the code).

It would just replace the `isintance` condition -> make more obvious different behavior.

> I didn't inherit from the same errors because some were previously RuntimeErrors and others were ValueErrors and making them now inherit from the same errors would mean that any try..except someone might have used to capture the previous RuntimeErrors or ValueErrors might then not catch the new errors.

To be honest I don't think we should keep that base exceptions. If anybody used `try: ... except RuntimeError:` in a custom logic (not sure where would someone used  custom logic for switching?) 1.) he should not do that 2.) he should know we added new exceptions.

This would work too.
```python
try:
    switch_container(container, repre_doc, loader)
except (
    LoaderSwitchNotImplementedError,
    IncompatibleLoaderError,
    LoaderNotFoundError,
) as exc:
    text = str(exc)
except Exception:
    text = (
        "Switch asset failed. "
        "Search console log for more details."
    )
```

BigRoy:
```python
try:
    switch_container(container, repre_doc, loader)
except (
    LoaderSwitchNotImplementedError,
    IncompatibleLoaderError,
    LoaderNotFoundError,
) as exc:
    text = str(exc)
except Exception:
    text = (
        "Switch asset failed. "
        "Search console log for more details."
    )
```

Yes, that would work. Except that instead it would become:
```python
try:
    switch_container(container, repre_doc, loader)
except (
    LoaderSwitchNotImplementedError,
    IncompatibleLoaderError,
    LoaderNotFoundError,
) as exc:
    text = str(exc)
        msg = (
        "Couldn't switch asset."
        "See traceback for more information."
    )
    log.warning(msg, exc_info=True)
    dialog = QtWidgets.QMessageBox(self)
    dialog.setWindowTitle("Switch asset failed")
    dialog.setText(text)
    dialog.exec_()
except Exception:
    text = (
        "Switch asset failed. "
        "Search console log for more details."
    )
    msg = (
        "Couldn't switch asset."
        "See traceback for more information."
    )
    log.warning(msg, exc_info=True)
    dialog = QtWidgets.QMessageBox(self)
    dialog.setWindowTitle("Switch asset failed")
    dialog.setText(text)
    dialog.exec_()
```
Right? Unless we move that logic into a re-useable function of its own?

iLLiCiTiT:
```
text = None
try:
    switch_container(container, repre_doc, loader)
except (
    LoaderSwitchNotImplementedError,
    IncompatibleLoaderError,
    LoaderNotFoundError,
) as exc:
    text = str(exc)
except Exception:
    text = (
        "Switch asset failed. "
        "Search console log for more details."
    )

if text:
    log.warning(msg, exc_info=True)
    dialog = QtWidgets.QMessageBox(self)
    dialog.setWindowTitle("Switch asset failed")
    dialog.setText(text)
    dialog.exec_()
```

BigRoy: Fine with me - added with https://github.com/ynput/OpenPype/pull/5227/commits/6c96c658e26c9c7cded3e124abcccf7c3ebf507b

iLLiCiTiT (approved):
LGTM

LiborBatek (approved):
When hitting some limitation in ```Manage``` when trying ```Switch``` in non compatible types of data it correctly shows Error message understandable to the user, like

image

Or
image

Issue #5162, opened 2023-06-20
Bug: Delivery action reports success even when there is none


antirotor:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

Using delivery action but not setting the UI correctly will result in "success message" even if nothing actually happens:

![delivery1 gif](https://github.com/ynput/OpenPype/assets/33513211/da8ab76e-b7e5-4183-b5b3-bcba0cf0eb5b)

here I haven't filled root and I didn't select representation, clicking on **Deliver** did nothing, but I still got "Delivery finished successfully".

### Expected Behavior:

I would expect some validation that at least one representation is selected and root is filled.

### Version

3.15.11-nightly.3

### What platform you are running OpenPype on?

Windows

### Steps To Reproduce:

1. Open Loader
2. Select sequence and right+click launch `Deliver Versions`
3. Leave it as it is and hit **Deliver** button

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

[cuID:[OP-6208](https://app.clickup.com/t/863h1vxpg)]

PR #5228, opened 2023-07-03
Validate delivery action representations status

movalex:
## Changelog Description

- disable delivery button if no representations checked
- fix macos combobox layout
- add error message if no delivery templates found

## Additional info
Addressed #5162

## Testing notes:

1. Open Loader
2. Select sequence and right+click launch Deliver Versions
3. Leave it as it is. Deliver button will not be activated
4. Delete all delivery templates in `project_anatomy` and open delivery dialogue. Error message will be shown in the `text_area` with corresponding error log message.

image


moonyuet (approved):
LGFM and the adjustment is also affecting #5024

mkolar:
Than you @movalex

PR #5229, opened 2023-07-03
Max: OP parameters only displays what is attached to the container

moonyuet:
## Changelog Description
The OP parameter in 3dsmax only displays what is currently attached to the container while deleting while you can see the items which is not added when you are adding to the container.

## Additional info
n/a
## Testing notes:
1. launch Max with OP
2. create instance
3. Go to OP Parameters -> Add to Container
4. Add something
5. Add again and it will only display the items which are not added into container
6. Go to OP Parameters -> Delete from Container
7. It will only display what is currently attached to the container


LiborBatek (approved):
All work as supposed to...

when adding items to OP Container
image

image

When Removing items from OP Container:
image

After removal:
image

All working fine...Add/remove window shows just items in the proper context not doubling them.

moonyuet:
@LiborBatek the update would filter out the container or any container objects.

mkolar (approved):
Code is looking good. @LiborBatek when you confirm function in DCC, we're ready for merge

LiborBatek:
> Code is looking good. @LiborBatek when you confirm function in DCC, we're ready for merge

yeah it works ok.

LiborBatek (changes requested):
There need to be resolved the issue when add to the container:
image

Steps to reproduce this error:

1) create some objects and select few and hit "Add" to Container (model Instance in my case)
2) repeat and hit "Add" for adding more objects
3) Error appears

moonyuet:
> There need to be resolved the issue when add to the container: image
> 
> Steps to reproduce this error:
> 
> 1. create some objects and select few and hit "Add" to Container (model Instance in my case)
> 2. repeat and hit "Add" for adding more objects
> 3. Error appears

It should be fixed now after removing the duplicated line in maxscript.

LiborBatek (approved):
No issues found. All works fine.

image

PR #5230, opened 2023-07-03
 Houdini: Add geometry check for pointcache family

mustafa-zarkash:
## Changelog Description
When `sop_path`  on ABC ROP node points to a  non `SopNode`, these validators `validate_abc_primitive_to_detail.py`, `validate_primitive_hierarchy_paths.py` will error and crash when this line is executed 
`geo = output_node.geometryAtFrame(frame)`

### DIFF
--- a/openpype/hosts/houdini/plugins/publish/validate_abc_primitive_to_detail.py
+++ b/openpype/hosts/houdini/plugins/publish/validate_abc_primitive_to_detail.py
@@ -73,6 +73,14 @@ class ValidateAbcPrimitiveToDetail(pyblish.api.InstancePlugin):
         cls.log.debug("Checking Primitive to Detail pattern: %s" % pattern)
         cls.log.debug("Checking with path attribute: %s" % path_attr)
 
+        if not hasattr(output_node, "geometry"):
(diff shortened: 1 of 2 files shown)


mustafa-zarkash:
> Are there other validators in place to ensure the publishing still gets invalidated?

yes, I faced this bug when testing wrong inputs on ROP `sop_path`
ROP `sop_path` is validated by `Validate Output Node`

mustafa-zarkash:
I'm afraid `validate_vdb_output_node` needs the same check! 
going to add it to this PR.

moonyuet:
> I'm afraid `validate_vdb_output_node` needs the same check! going to add it to this PR.

Isn't that only happened to vdbcache families?

mustafa-zarkash:
> Isn't that only happened to vdbcache family

you are right.  
I thought about adding it to this PR as it's the same bug but happening in a different family

BigRoy (approved):
Nice!

moonyuet (approved):
looks good for me

mkolar:
Thanks @mustafa-zarkash

PR #5231, opened 2023-07-03
Houdini: add select invalid action for ValidateSopOutputNode

mustafa-zarkash:
## Changelog Description
This PR adds `SelectROPAction` action to `houdini\api\action.py`
and it's used in `Validate Output Node` 

`SelectROPAction` is used to select the associated ROPs with the errored instances.

## Additional info
Testing this PR requires merging these [PR](https://github.com/ynput/OpenPype/pull/5230) and [PR](https://github.com/ynput/OpenPype/pull/5232)
Alternatively, you can add these lines before `geo = output_node.geometryAtFrame(frame)`
```
if not hasattr(output_node, "geometry"):
  # In the case someone has explicitly set an Object
  # node instead of a SOP node in Geometry context
  # then for now we ignore - this allows us to also
  # export object transforms.
  cls.log.warning("No geometry output node found, skipping check..")
  return
```

### DIFF
--- a/openpype/hosts/houdini/plugins/publish/validate_sop_output_node.py
+++ b/openpype/hosts/houdini/plugins/publish/validate_sop_output_node.py
@@ -6,8 +6,8 @@ from openpype.pipeline.publish import RepairAction
 import hou
 
 
-class SelectInvalidAction(RepairAction):
-    label = "Select Invalid ROP"
+class SelectROPAction(RepairAction):
+    label = "Select ROP"
     icon = "mdi.cursor-default-click"
 
 class ValidateSopOutputNode(pyblish.api.InstancePlugin):
@@ -26,7 +26,7 @@ class ValidateSopOutputNode(pyblish.api.InstancePlugin):
     families = ["pointcache", "vdbcache"]
     hosts = ["houdini"]
     label = "Validate Output Node"
-    actions = [SelectInvalidAction]
+    actions = [SelectROPAction]
 
     def process(self, instance):
 
@@ -89,10 +89,10 @@ class ValidateSopOutputNode(pyblish.api.InstancePlugin):
 
     @classmethod
     def repair(cls, instance):
-        """Select Invalid ROP.
+        """Select ROP.
 
-        It's used to select invalid ROP which tells the
-        artist which ROP node need to be fixed!
+        It's used to select the associated ROP for the selected instance
+        which tells the artist which ROP node need to be fixed!
         """
 
         rop_node = hou.node(instance.data["instance_node"])
--- a/openpype/hosts/houdini/plugins/publish/validate_sop_output_node.py
+++ b/openpype/hosts/houdini/plugins/publish/validate_sop_output_node.py
@@ -2,6 +2,7 @@
 import pyblish.api
 from openpype.pipeline import PublishValidationError
 from openpype.pipeline.publish import RepairAction
+from openpype.hosts.houdini.api.action import SelectInvalidAction
 
 import hou
 
@@ -26,7 +27,7 @@ class ValidateSopOutputNode(pyblish.api.InstancePlugin):
     families = ["pointcache", "vdbcache"]
     hosts = ["houdini"]
     label = "Validate Output Node"
-    actions = [SelectROPAction]
+    actions = [SelectROPAction, SelectInvalidAction]
 
     def process(self, instance):
 
@@ -48,7 +49,7 @@ class ValidateSopOutputNode(pyblish.api.InstancePlugin):
                 "Ensure a valid SOP output path is set." % node.path()
             )
 
-            return [node.path()]
+            return [node]
 
         # Output node must be a Sop node.
         if not isinstance(output_node, hou.SopNode):
@@ -58,7 +59,7 @@ class ValidateSopOutputNode(pyblish.api.InstancePlugin):
                 "instead found category type: %s"
                 % (output_node.path(), output_node.type().category().name())
             )
-            return [output_node.path()]
+            return [output_node]
 
         # For the sake of completeness also assert the category type
         # is Sop to avoid potential edge case scenarios even though
@@ -78,14 +79,14 @@ class ValidateSopOutputNode(pyblish.api.InstancePlugin):
             except hou.Error as exc:
                 cls.log.error("Cook failed: %s" % exc)
                 cls.log.error(output_node.errors()[0])
-                return [output_node.path()]
+                return [output_node]
 
         # Ensure the output node has at least Geometry data
         if not output_node.geometry():
             cls.log.error(
                 "Output node `%s` has no geometry data." % output_node.path()
             )
-            return [output_node.path()]
+            return [output_node]
 
     @classmethod
     def repair(cls, instance):
--- a/openpype/hosts/houdini/plugins/publish/validate_sop_output_node.py
+++ b/openpype/hosts/houdini/plugins/publish/validate_sop_output_node.py
@@ -1,15 +1,13 @@
 # -*- coding: utf-8 -*-
 import pyblish.api
 from openpype.pipeline import PublishValidationError
-from openpype.pipeline.publish import RepairAction
-from openpype.hosts.houdini.api.action import SelectInvalidAction
+from openpype.hosts.houdini.api.action import (
+    SelectInvalidAction,
+    SelectROPAction,
+)
 
 import hou
 
-
-class SelectROPAction(RepairAction):
-    label = "Select ROP"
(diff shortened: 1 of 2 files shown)


openpype/hosts/houdini/plugins/publish/validate_sop_output_node.py
```diff
@@ -1,7 +1,14 @@
...
 import pyblish.api
 from openpype.pipeline import PublishValidationError
+from openpype.pipeline.publish import RepairAction
 
+import hou
+
+
+class SelectInvalidAction(RepairAction):
+    label = "Select Invalid ROP"
+    icon = "mdi.cursor-default-click"
```
BigRoy, line 11: Nicely picked icon! We might want to push that through all `SelectInvalidAction` actions if that's a preferred icon.

Anyway, I believe in this case you should be able to use the [action already existing for Houdini here](https://github.com/ynput/OpenPype/blob/b3044398fc9181db2d2230f9f0f5cc1de7e9d297/openpype/hosts/houdini/api/action.py#L7).

The logic you have now also wouldn't work for multiple instances since it'd only select one rop since the actions run over all instances as far as I know ;)

BigRoy: > if multiple ROP nodes are selected, will artist figure out which ROP is associated with the selected validator error in publisher ?

It'll only select for invalid instances - **but** if there are two pointcache instances with the same invalidation then it'd select both. It's unfortunately how the Action system works, it triggers per plugin - not per instance.

> get_invalid return node paths instead of the node itself also it logs.

The logging is fine. Having it return the nodes instead of the paths seems fine to me if that's what SelectInvalidAction needs.

> I tried testing it first to learn how it works but for some reason it doesn't work on my machine, I suspect the process method is not even called!
> I added some log.debug inside process but no chance although get_invalid works fine in the mentioned validator above!

I think this is because `get_invalid` doesn't return a list of nodes but the path itself. It should [return a list of nodes to be correct inputs for `SelectInvalidAction`](https://github.com/ynput/OpenPype/blob/b3044398fc9181db2d2230f9f0f5cc1de7e9d297/openpype/hosts/houdini/api/action.py#L30-L35) if I look at this code. Looks like a bug in `validate_vdb_output_node.py`.

mustafa-zarkash: > Having it return the nodes instead of the paths seems fine
so, there two things here. 
1. get invalid return the associated `output` node
2. my action selects the associated `rop` node 
what about having both ? 
- select invalid 🔍 
- select ROP 


> a bug in `validate_vdb_output_node.py`.
I'm going to fix it in another PR

BigRoy: > select invalid 🔍
> select ROP

I guess it could be nice - but admittedly in that scenario I have no idea what "Select invalid" would select - haha. I'd just start off with solely implementing select invalid?

mustafa-zarkash: > I'd just start off with solely implementing select invalid?

`select invalid 🔍` will select one of these 
- `hou.node(instance.data["instance_node"])` if output_node is None 
- `output_node` if not isinstance(output_node, hou.SopNode)
- `output_node` if node failed to cook `output_node.cook()` 
- `output_node` if not output_node.geometry() 

unlike `Select Invalid ROP`
will only selects the associated ROP for the selected instance

BigRoy: Sure, let's give it a go.

openpype/hosts/houdini/plugins/publish/validate_sop_output_node.py
```diff
@@ -2,12 +2,13 @@
...
+from openpype.hosts.houdini.api.action import SelectInvalidAction
 
 import hou
 
 
-class SelectInvalidAction(RepairAction):
-    label = "Select Invalid ROP"
+class SelectROPAction(RepairAction):
+    label = "Select ROP"
     icon = "mdi.cursor-default-click"
```
BigRoy: Should we maybe change the `RepairAction` class so it allows customizing the method's name? Or even allows customizing that the method is on the action itself? A `SelectROPAction` would technically work for almost all Houdini validations because basically ALL of them have the `instance_node` available so we might even want to expose the `SelectROPAction` directly in `openpype.hosts.houdini.api.action` with selecting logic of its own.

E.g. this could be in the `action` lib:

```python
class SelectROPAction(pyblish.api.Action):
    on = "failed"  # This action is only available on a failed plug-in
    label = "Select ROP"
    icon = "mdi.cursor-default-click"

    def process(self, context, plugin):

        errored_instances = get_errored_instances_from_context(context)

        # Apply pyblish.logic to get the instances for the plug-in
        instances = pyblish.api.instances_by_plugin(errored_instances, plugin)

        # Get the invalid nodes for the plug-ins
        self.log.info("Finding invalid nodes..")
        rop_nodes = list()
        for instance in instances:
            path = instance.data.get("instance_node")
            if not path:
                continue

            node = hou.node(path)
            if not node:
                continue

            rop_nodes.append(node)

        hou.clearAllSelected()
        if invalid:
            self.log.info("Selecting ROP nodes: {}".format(
                ", ".join(node.path() for node in rop_nodes)
            ))
            for node in rop_nodes:
                node.setSelected(True)
                node.setCurrent(True)
        else:
            self.log.info("No ROP nodes found.")
```
And would be usable for all instances?

moonyuet: @BigRoy we can definitely store this piece of information in `houdini/api/lib.py` for future use in validation.

BigRoy: That'd be in `houdini/api/action.py` by the way, but yes. That was my intention.

mustafa-zarkash: > we might even want to expose the `SelectROPAction` directly in `openpype.hosts.houdini.api.action` with selecting logic of its own.

Sounds interesting! I going to give it a try. 

>  if there are two pointcache instances with the same invalidation then it'd select both. It's unfortunately how the Action system works, it triggers per plugin - not per instance.

I re-read it, then I checked the `RepairAction`  class which has similar code like `SelectInvalidAction` 
So, I modified the selection code in `repair` method to 
```python
rop_node.setSelected(True)
rop_node.setCurrent(True)
```

and I faced this weird thing! it's confusing 
even if there's one instance that errored, `repair` method still runs when clicking `Select ROP` Button
so it works with each validator _i.e. plugin_  separately for all instances ?  

image

BigRoy: Haha, wow. I'm quite blown away that this hasn't been found before. It looks like the call to `get_errored_instances_from_context` just filters whether that particular instance had ANY errors, not necessarily whether it had errors related to that particular plugin.

Could you try it so that the other instances don't have any OTHER errors too. Then it likely works. The fix shouldn't be that hard too.

mustafa-zarkash: I made a test, publish 3 `pointcaches`
1. `pointcacheMain` which has no errors 
2. `pointcacheMain2` which points to a `SopNode` but has no path attribute 
3. `pointcacheMain3` which points to an `ObjNode` 


No 1 is the only instance that didn't fail, but it wasn't published as others failed 
No 2 has failed for another reason other than `Validate Output Node`
No 3 is the only one that didn't pass `Validate Output Node`

however, when clicking `SelectROPAction` it selects the associated ROPs with all errored instances
whether they had errors related to the plugin.

image

BigRoy: Here's a draft PR which MIGHT solve the issue: https://github.com/ynput/OpenPype/pull/5240

mustafa-zarkash: I will check it ASAP. 
and I have moved the SelectRopAction to `openpype.hosts.houdini.api.action`

mustafa-zarkash:
the recent commit to make this PR compatible  with @BigRoy's [PR](https://github.com/ynput/OpenPype/pull/5240)

mustafa-zarkash:
the recent commit to make this PR compatible  with @BigRoy's [PR](https://github.com/ynput/OpenPype/pull/5240)

BigRoy (approved):
Code looks good to me. 👍

mkolar (approved):

PR #5232, opened 2023-07-03
Houdini: Add geometry check for VDB family

mustafa-zarkash:
## Changelog Description
When `sop_path` on Geometry ROP node points to a non SopNode, this validator `validate_vdb_output_node.py` will error and crash when this line is executed
`sop_node.geometryAtFrame(frame)`

## Additional info
The bug addressed here is the same as in this [PR](https://github.com/ynput/OpenPype/pull/5230)

### DIFF
--- a/openpype/hosts/houdini/plugins/publish/validate_vdb_output_node.py
+++ b/openpype/hosts/houdini/plugins/publish/validate_vdb_output_node.py
@@ -55,6 +55,8 @@ def get_geometry_at_frame(sop_node, frame, force=True):
     """Return geometry at frame but force a cooked value."""
     with update_mode_context(hou.updateMode.AutoUpdate):
         sop_node.cook(force=force, frame_range=(frame, frame))
+        if not hasattr(sop_node, "geometry"):
+            return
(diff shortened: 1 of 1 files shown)


openpype/hosts/houdini/plugins/publish/validate_vdb_output_node.py
```diff
@@ -55,6 +55,8 @@ def get_geometry_at_frame(sop_node, frame, force=True):
     """Return geometry at frame but force a cooked value."""
     with update_mode_context(hou.updateMode.AutoUpdate):
         sop_node.cook(force=force, frame_range=(frame, frame))
+        if not hasattr(sop_node, "geometry"):
+            return
```
BigRoy, line 59: Can we do this check outside of the with statement? That way we don't force a cook and change update states for no reason.

mustafa-zarkash: good observation! 
thank you

BigRoy (approved):
Looks good to me.

moonyuet (approved):
LGFM

PR #5233, opened 2023-07-03
Maya: aligning default settings to distributed aces 1.2 config

jakubjezek001:
## Changelog Description
Maya colorspace setttings defaults are set the way they align our distributed ACES 1.2 config file set in global colorspace configs.

## Additional info
Workfile category is disabled by default since we don't want to damage any productions which were having configured their settings already.

## Dependency
The issue was mentioned here https://github.com/ynput/OpenPype/pull/5225

### DIFF
--- a/openpype/settings/defaults/project_settings/maya.json
+++ b/openpype/settings/defaults/project_settings/maya.json
@@ -421,9 +421,9 @@
         },
         "workfile": {
             "enabled": false,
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
Yes please! ❤️ 🚀 However, merge it after https://github.com/ynput/OpenPype/pull/5225 because currently the `if workfiles` check is still reversed.

PR #5234, opened 2023-07-03
Substance Painter: Include the setting only in publish tab

### LINKED ISSUE #5172: Bug: Substance Painter: When creating TextureSet in Publisher, its params arent propagated to the Instance

### Current Behavior:

When user creates new ```Texture Set``` in ```Publisher``` and change its parameters like ```Output Template``` or ```File Type``` etc. those changes arent propagated to the newly created ```Texture Set``` instance and resets back to default values.

image

### Expected Behavior:

If any changes made to any of the parameters on ```Texture Set``` instance during creation it should propagate those to newly created ```Texture Set``` obviously, so user doesnt need to tweak those params again because of reverting to defaults.

### Version

3.15.11-nightly.4

Windows

### Steps To Reproduce:

1. Go to OP>Create
2. Create new ```Texture Set``` instance
3. Tweak its settings to your liking so it differ from defaults
4. Check newly created ```Texture Set``` if all changes were propagated/saved

### Relevant log output:

_No response_

### Additional context:

_No response_

[cuID:[OP-6232](https://app.clickup.com/t/863h2adfe)]

moonyuet:
## Changelog Description
Instead of having two settings in both create and publish tab, there is solely one setting in the publish tab for users to set up the parameters.
Resolve #5172

## Additional info
n/a

## Testing notes:
1. Launch SP via OP
2. Open project file
3. Create Texture Set
4. Go to Publish Tab -> Click TextureSetMain
5. It would show all the settings

### DIFF
--- a/openpype/hosts/substancepainter/plugins/create/create_textures.py
+++ b/openpype/hosts/substancepainter/plugins/create/create_textures.py
@@ -34,7 +34,6 @@ class CreateTextures(Creator):
         if not substance_painter.project.is_open():
             raise CreatorError("Can't create a Texture Set instance without "
                                "an open project.")
-
         instance = self.create_instance_in_context(subset_name,
                                                    instance_data)
         set_instance(
@@ -76,7 +75,6 @@ class CreateTextures(Creator):
         return instance
 
     def get_instance_attr_defs(self):
-
         return [
             EnumDef("exportPresetUrl",
                     items=get_export_presets(),
@@ -156,7 +154,3 @@ class CreateTextures(Creator):
             UILabelDef("*only used with "
                        "'Dilation + <x>' padding"),
         ]
-
-    def get_pre_create_attr_defs(self):
-        # Use same attributes as for instance attributes
-        return self.get_instance_attr_defs()
--- a/openpype/hosts/substancepainter/plugins/create/create_textures.py
+++ b/openpype/hosts/substancepainter/plugins/create/create_textures.py
@@ -35,6 +35,7 @@ class CreateTextures(Creator):
             raise CreatorError("Can't create a Texture Set instance without "
                                "an open project.")
         # Transfer settings from pre create to instance
+        creator_attributes = instance_data.setdefault("creator_attributes", dict())
         for key in [
             "exportPresetUrl",
             "exportFileFormat",
@@ -43,7 +44,7 @@ class CreateTextures(Creator):
             "exportDilationDistance"
         ]:
             if key in pre_create_data:
-                instance_data[key] = pre_create_data[key]
+                creator_attributes[key] = pre_create_data[key]
 
         instance = self.create_instance_in_context(subset_name,
                                                    instance_data)
--- a/openpype/hosts/substancepainter/plugins/create/create_textures.py
+++ b/openpype/hosts/substancepainter/plugins/create/create_textures.py
@@ -28,36 +28,22 @@ class CreateTextures(Creator):
     icon = "picture-o"
 
     default_variant = "Main"
-    image_format = None
-    exportPresetUrl = None
-    exportSize = None
-    exportPadding = "infinite"
-    exportDilationDistance = 16
 
     def create(self, subset_name, instance_data, pre_create_data):
 
         if not substance_painter.project.is_open():
             raise CreatorError("Can't create a Texture Set instance without "
                                "an open project.")
-        self.exportPresetUrl = pre_create_data.get("exportPresetUrl",
-                                                   self.exportPresetUrl)
-        instance_data["exportPresetUrl"] = self.exportPresetUrl
-
-        self.image_format = pre_create_data.get("exportFileFormat",
-                                                self.image_format)
-        instance_data["exportFileFormat"] = self.image_format
-
-        self.exportSize = pre_create_data.get("exportSize",
-                                              self.exportSize)
-        instance_data["exportSize"] = self.exportSize
-
-        self.exportPadding = pre_create_data.get("exportPadding",
-                                                 self.exportPadding)
-        instance_data["exportPadding"] = self.exportPadding
-
-        self.exportDilationDistance = pre_create_data.get("exportDilationDistance",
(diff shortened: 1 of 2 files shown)


BigRoy:
This doesn't sound right.

We'd still want to expose the parameters on both Create and on the instance itself - we should just need to make sure they persist correctly. They are not intended to be "two settings" - they are intended to be the exact same setting however you can already set them during Create.

The issue is that it doesn't propagate to the instance - not that the attributes exist on both pages.

moonyuet:
> This doesn't sound right.
> 
> We'd still want to expose the parameters on both Create and on the instance itself - we should just need to make sure they persist correctly. They are not intended to be "two settings" - they are intended to be the exact same setting however you can already set them during Create.
> 
> The issue is that it doesn't propagate to the instance - not that the attributes exist on both pages.

This bring up another question: why we need exactly same settings in both create and publish tabs? It is supposed that users know what to export and choose the settings before publishing...

BigRoy:
The idea is that on create you decide how you want to publish. Then if along the way you figured you wanted to tweak a setting then you can.

LiborBatek:
To add to this discussion...

I do agree with @BigRoy I wouldnt mind to have the params of the ```Texture Set``` exposed in both Publish tabs (Create and Publish). They just should match always.

Speaking of @moonyuet  proposal...if there are any technical reasons (you mentioned on discord there could be possible needs to change other OP modules too like collector) to do so we should consider how big impact this has.

I would also follow logic of other DCC and ```Publisher``` it should be implemented the same way.

BigRoy:
@iLLiCiTiT about this UX - would it otherwise be possible to "set any instance attribute defs" during Create.

So that maybe there's an extra divider that you can "open up" to even tweak the instance attributes for the to be created instance directly instead of having to do it afterwards on the Publish tab for the instance. It'd mean the code doesn't have to be duplicated in this case - plus it'd directly work the same across all hosts for all instances.

By default it then shows only pre create attributes, but there's like an extra vertical splitter that you can open up so that even instance attributes become settable during the creation? (Which would just trigger regular create, then after set the customized attributes (if any) and saves only those attributes for that instance?) 

---

Also, @mkolar what are your thoughts on the discussion regarding the attributes on create versus publish tab?

openpype/hosts/substancepainter/plugins/create/create_textures.py
```diff
@@ -28,12 +28,36 @@ class CreateTextures(Creator):
     icon = "picture-o"
 
     default_variant = "Main"
+    image_format = None
+    exportPresetUrl = None
+    exportSize = None
+    exportPadding = "infinite"
+    exportDilationDistance = 16
```
BigRoy, line 35: There's no need to store these as class attributes right? Especially because now it's changing for each created instance down below.

openpype/hosts/substancepainter/plugins/create/create_textures.py
```diff
@@ -28,12 +28,36 @@ class CreateTextures(Creator):
...
+                                              self.exportSize)
+        instance_data["exportSize"] = self.exportSize
+
+        self.exportPadding = pre_create_data.get("exportPadding",
+                                                 self.exportPadding)
+        instance_data["exportPadding"] = self.exportPadding
+
+        self.exportDilationDistance = pre_create_data.get("exportDilationDistance",
+                                                          self.exportDilationDistance)
+        instance_data["exportDilationDistance"] = self.exportDilationDistance
```
BigRoy, line 60:
```suggestion
    # Transfer settings from pre create to instance
    for key in [
        "exportPresetUrl",
        "exportFileFormat",
        "exportSize",
        "exportPadding",
        "exportDilationDistance"
    ]:
        if key in pre_create_data:
            instance_data[key] = pre_create_data[key]
```
Maybe better like this?

openpype/hosts/substancepainter/plugins/create/create_textures.py
```diff
@@ -145,13 +170,13 @@ def get_instance_attr_defs(self):
...
                     },
-                    default="infinite",
+                    default=self.exportPadding,
                     label="Padding"),
             NumberDef("exportDilationDistance",
                       minimum=0,
                       maximum=256,
                       decimals=0,
-                      default=16,
+                      default=self.exportDilationDistance,
```
BigRoy, line 179: I don't feel like we should update the defaults based on the pre create data at all. The defaults are the defaults. If someone tweaked it in pre create then it's still changed from the default value so I wouldn't bother with tweaking the defaults.

openpype/hosts/substancepainter/plugins/create/create_textures.py
mkolar: Yes!

openpype/hosts/substancepainter/plugins/create/create_textures.py
mkolar: Agreed.

openpype/hosts/substancepainter/plugins/create/create_textures.py
```diff
@@ -34,6 +34,18 @@ def create(self, subset_name, instance_data, pre_create_data):
...
+            "creator_attributes", dict())
+        for key in [
+            "exportPresetUrl",
+            "exportFileFormat",
+            "exportSize",
+            "exportPadding",
+            "exportDilationDistance"
+        ]:
+            if key in pre_create_data:
+                creator_attributes[key] = pre_create_data[key]
```
BigRoy, line 48: Oh yeah baby, I'd love it if this works! Since it's relatively clean and isolated where the logic is. Does it work? 😮

moonyuet: it does work and the parameter is propagated to the instance.

BigRoy (approved):
Of course there's still the debate whether the attribute should actually be exposed on both Create and Publish tab but this code looks good to me and according to @moonyuet fixes the issue it's intended to solve.

LiborBatek (approved):
All works as expected.
image

PR #5237, opened 2023-07-04
Enhancement/houdini add path action for abc validator

mustafa-zarkash:
## Changelog Description
Add a default path attribute Action.
it's a helper action more than a repair action, which used to add a default single value.

## Additional info
This validator is going to be skipped if the `sop_path` in the ROP doesn't point to a `SopNode` after merging this [PR](https://github.com/ynput/OpenPype/pull/5230)
otherwise the publisher will error 

## Testing notes:
use `pointcache` to export these examples, set the sop path explicitly
- test 1 : create `AUTO_PATH` node before `null` node 
- test 2 : create `AUTO_PATH` node before `output` node 
- test 3 : create `AUTO_PATH` node after `sphere` node and set it as output node and update ROP node
- test 4 :  create `AUTO_PATH` node after `merge` node and set it as output node and update ROP node
- test 5 :  create `AUTO_PATH` node after `copy` node and set it as output node and update ROP node
- test 6 : set `sop_path` explicitly to the parent geo node  of these nodes in screenshot, publish will crash
- test 6 : after merging the mentioned PR, publisher will only show `Validate Output Node` error 

Screenshot 2023-07-04 100221

Untitled

### DIFF
--- a/openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
+++ b/openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
@@ -150,9 +150,10 @@ class ValidatePrimitiveHierarchyPaths(pyblish.api.InstancePlugin):
             path_node, path_attr
         )
 
-        path_node.setGenericFlag(hou.nodeFlag.DisplayComment,True)
+        path_node.setGenericFlag(hou.nodeFlag.DisplayComment, True)
         path_node.setComment(
-            'Auto path node created automatically by "Add a default path attribute"'
+            'Auto path node was created automatically by '
+            '"Add a default path attribute"'
             '\nFeel free to modify or replace it.'
         )
 
--- a/openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
+++ b/openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
@@ -1,9 +1,17 @@
 # -*- coding: utf-8 -*-
 import pyblish.api
-from openpype.pipeline.publish import ValidateContentsOrder
 from openpype.pipeline import PublishValidationError
+from openpype.pipeline.publish import (
+    ValidateContentsOrder,
+    RepairAction,
+)
+
 import hou
 
+
+class AddDefaultPathAction(RepairAction):
+    label = "Add a default path attribute"
+    icon = "mdi.pencil-plus-outline"
 
 class ValidatePrimitiveHierarchyPaths(pyblish.api.InstancePlugin):
     """Validate all primitives build hierarchy from attribute when enabled.
@@ -18,6 +26,7 @@ class ValidatePrimitiveHierarchyPaths(pyblish.api.InstancePlugin):
     families = ["pointcache"]
     hosts = ["houdini"]
     label = "Validate Prims Hierarchy Path"
+    actions = [AddDefaultPathAction]
 
     def process(self, instance):
         invalid = self.get_invalid(instance)
@@ -104,3 +113,46 @@ class ValidatePrimitiveHierarchyPaths(pyblish.api.InstancePlugin):
                 "(%s of %s prims)" % (path_attr, len(invalid_prims), num_prims)
             )
             return [output_node.path()]
+
+    @classmethod
+    def repair(cls, instance):
+        output_node = instance.data.get("output_node")
+        rop_node = hou.node(instance.data["instance_node"])
+
+        path_attr = rop_node.parm("path_attrib").eval()
+
+        frame = instance.data.get("frameStart", 0)
+        geo = output_node.geometryAtFrame(frame)
+
+        paths = geo.findPrimAttrib(path_attr) and \
+                    geo.primStringAttribValues(path_attr)
+
+        if paths:
+            return
+
+        path_node = output_node.parent().createNode("name", "AUTO_PATH")
+        path_node.parm("attribname").set(path_attr)
+        path_node.parm("name1").set('`opname("..")`/`opname("..")`Shape')
+
+        cls.log.debug(
+            '%s node has been created'
+            % path_node
+        )
+
+        path_node.setGenericFlag(hou.nodeFlag.DisplayComment,True)
+        path_node.setComment(
+            'Auto path node created automatically by "Add a default path attribute"'
+            '\nFeel free to modify or replace it.'
+        )
+
+        if output_node.type().name() in ['null', 'output']:
+            # Connect before
+            path_node.setFirstInput(output_node.input(0))
+            path_node.moveToGoodPosition()
+            output_node.setFirstInput(path_node)
+            output_node.moveToGoodPosition()
+        else:
+            # Connect after
+            output_node.setFirstInput(path_node)
+            rop_node.parm('sop_path').set(path_node.path())
+            path_node.moveToGoodPosition()
--- a/openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
+++ b/openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
@@ -46,7 +46,7 @@ class ValidatePrimitiveHierarchyPaths(pyblish.api.InstancePlugin):
         if output_node is None:
             cls.log.error(
                 "SOP Output node in '%s' does not exist. "
-                "Ensure a valid SOP output path is set." % rop_node.path()
+                "Ensure a valid SOP output path is set.", rop_node.path()
             )
 
             return [rop_node]
@@ -68,7 +68,7 @@ class ValidatePrimitiveHierarchyPaths(pyblish.api.InstancePlugin):
             )
             return [rop_node]
 
-        cls.log.debug("Checking for attribute: %s" % path_attr)
+        cls.log.debug("Checking for attribute: %s", path_attr)
 
         # Check if the primitive attribute exists
         frame = instance.data.get("frameStart", 0)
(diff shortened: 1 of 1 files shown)


openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
```diff
@@ -104,3 +113,60 @@ def get_invalid(cls, instance):
...
+        output_node = instance.data.get("output_node")
+        rop_node = hou.node(instance.data["instance_node"])
+
+        path_attr = rop_node.parm("path_attrib").eval()
+
+        frame = instance.data.get("frameStart", 0)
+        geo = output_node.geometryAtFrame(frame)
+
+        paths = geo.findPrimAttrib(path_attr) and \
+                    geo.primStringAttribValues(path_attr)
```
BigRoy, line 134: It'd likely be easier to maintain to just do:
```python
if not cls.get_invalid(instance):
    return
```

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
```diff
@@ -104,3 +113,60 @@ def get_invalid(cls, instance):
...
+
+    @classmethod
+    def repair(cls, instance):
+        """Add a default path attribute Action.
+
+        it's a helper action more than a repair action,
+        which used to add a default single value.
+        """
+
+        output_node = instance.data.get("output_node")
```
BigRoy, line 125: Be aware that output node could be **not** set which will still fail in repair down below. Should we capture that case where no path is set (or not valid node path is set) and just log an error and return?

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
```diff
@@ -104,3 +113,60 @@ def get_invalid(cls, instance):
...
+        else:
+            # Connect after
+            path_node.setFirstInput(output_node)
+            rop_node.parm('sop_path').set(path_node.path())
+            path_node.moveToGoodPosition()
+            instance.data.update({"output_node" : path_node })
+
+            cls.log.debug(
+            "'%s' has set as the output node, "
+            "'%s' has updated"
```
BigRoy, line 170: This log message doesn't seem entirely right - has set _what_ as the output node?

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
```diff
@@ -1,10 +1,18 @@
...
+    RepairAction,
+)
+
 import hou
 
 
+class AddDefaultPathAction(RepairAction):
+    label = "Add a default path attribute"
+    icon = "mdi.pencil-plus-outline"
+
```
BigRoy, line 15: Isn't this missing a line of whitespace?
```suggestion


```

BigRoy (commented):
Very nice stuff! Just a few notes.

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
mustafa-zarkash: It's a nice idea. 
I think `get_invalid` will need to be refactored as the current logic doesn't support that. 

I think `add path action` button should only work if these checks evaluate to true 
`not attrib`
`not attrib.strings()`
`invalid_prims` 

and they do return `output_node.path()`  other checks return `None` or `rop_node.path()` 
so, this condition may be the one 
as well as moving logs to another method something similar to `get_invalid`  and `get_invalid_with_message`

```python
if  cls.get_invalid(instance) != [output_node.path()] :
    return
```

BigRoy: I think your check makes sense (with an added comment as to why we're checking that specifically.)
However, still be aware that `output_node` could be not set and thus `.path()` might not be available. ;) Which gets us into the unfortunate cases where we still need to perform additional checks.

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
mustafa-zarkash: what about this ? 
It will make us avoid a few things

```python
sop_path = rop_node.parm('sop_path').eval()
if not sop_path:
    return 
 output_node = hou.node(sop_path)
```

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
mustafa-zarkash: maybe "'%s' has been set as the output node" ? 
I should notify artist about `sop_path` of ROP (only happens when connecting path node after)

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
moonyuet: > what about this ? It will make us avoid a few things
> 
> ```python
> sop_path = rop_node.parm('sop_path').eval()
> if not sop_path:
>     return 
>  output_node = hou.node(sop_path)
> ```

maybe it's better to have log.message to tell user it will not perform repair action if there is no sop_path ?

BigRoy: You'll likely want this: `rop_node.parm("sop_path").evalAsNode()`
Right?

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
BigRoy: > SOP path on '{ropnode.path()}' updated to new output node {node.path()}

Something like that makes sense as a message?

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
mustafa-zarkash: well, I removed `.path()` which is more consistent with `SelectInvalidAction` 
also, if `output_node` could be not set, it will just be `None`

mkolar (approved):
On the code side this is looking solid. Functionality makes good sense to me as well and I'd say it covers the brief fairly well. @BigRoy i think it's worth merging and we'll see in practice

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
```diff
@@ -103,4 +113,62 @@ def get_invalid(cls, instance):
...
+        """
+
+        rop_node = hou.node(instance.data["instance_node"])
+        output_node = rop_node.parm('sop_path').evalAsNode()
+
+        if not output_node:
+            cls.log.debug(
+                "Action isn't performed, empty SOP Path on %s"
+                % rop_node
+            )
```
BigRoy, line 133: Technically it might not be empty - it could also point to a non-existent path and then return None. Right?
```suggestion
            cls.log.debug(
                "Action isn't performed, invalid SOP Path on %s"
                % rop_node
            )
```

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
```diff
@@ -103,4 +113,62 @@ def get_invalid(cls, instance):
...
+        else:
+            # Connect after
+            path_node.setFirstInput(output_node)
+            rop_node.parm('sop_path').set(path_node.path())
+            path_node.moveToGoodPosition()
+
+            cls.log.debug(
+                "SOP path on '%s' updated to new output node '%s'"
+                % (rop_node, path_node)
+            )
```
BigRoy, line 174: Since you're already using `%s` formatting I think you can actually allow the logger to format it only if a handler uses it by changing this to:
```suggestion
            cls.log.debug(
                "SOP path on '%s' updated to new output node '%s'",
                rop_node, path_node
            )
```
This would only really be faster if there's no logging handler using it however - but it might be more readable and due to it potentially being faster maybe a better standard too?

@iLLiCiTiT what do you think?

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
```diff
@@ -103,4 +113,62 @@ def get_invalid(cls, instance):
...
             )
-            return [output_node.path()]
+            return [output_node]
+
+    @classmethod
+    def repair(cls, instance):
+        """Add a default path attribute Action.
+
+        it's a helper action more than a repair action,
+        which used to add a default single value.
```
BigRoy, line 123:
```suggestion
        It is a helper action more than a repair action,
        used to add a default single value for the path.
```

BigRoy (approved):
Very nice work, just some cosmetic comments that I don't find crucial. So already approving it based on reading the code.

I did not do a test run however.

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
mustafa-zarkash: Technically, you are right 👌

openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
mustafa-zarkash: Interesting, I didn't know that!

mkolar:
We need to give it a test with another user, but looking great. We can merge it early next week.

mustafa-zarkash:
@moonyuet  
Could you give it a test  when you have time ?

moonyuet (approved):
image
Looks good for me too. 
This can be merged after fixing two cosmetic issues found by hound.

antirotor (approved):
Looks great!

Issue #5171, opened 2023-06-22
Bug: Substance Painter - Publish Fails on "Collect Texture Set images"


LiborBatek:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

When trying to publish texture set via Publisher it fails on collecting Texture set images.

### Expected Behavior:

Publish should pass normally without errors.

### Version

3.15.11-nightly.4

### What platform you are running OpenPype on?

Windows

### Steps To Reproduce:

1. Open any asset workfile with painted asset
2. Publish Texture Set
3. Error probably occurs.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

### Relevant log output:

```
DEBUG: Exporting using preset: resource://starter_assets/PBR Metallic Roughness
INFO: Processing testMat/
INFO: Processing $mesh_$textureSet_BaseColor(_$colorSpace)(.$udim)
INFO: Looking for matching profile for: families: "texture" | hosts: "substancepainter" | tasks: "lookdev" | task_types: "Lookdev"
DEBUG: "texture" not found in "families": ['workfile']
DEBUG: "texture" not found in "families": ['render']
DEBUG: "texture" not found in "families": ['renderLayer', 'renderPass']
DEBUG: "texture" not found in "families": ['review', 'workfile']
DEBUG: "texture" not found in "families": ['render']
DEBUG: "texture" not found in "families": ['staticMesh']
DEBUG: "texture" not found in "families": ['skeletalMesh']
INFO: Profile selected: {'families': [], 'hosts': [], 'task_types': [], 'tasks': [], 'template': '{family}{variant}'}
Traceback (most recent call last):
  File "D:\REPO\OpenPype\.venv\lib\site-packages\pyblish\plugin.py", line 522, in __explicit_process
    runner(*args)
  File "D:\REPO\OpenPype\openpype\hosts\substancepainter\plugins\publish\collect_textureset_images.py", line 43, in process
    self.create_image_instance(instance, template, outputs,
  File "D:\REPO\OpenPype\openpype\hosts\substancepainter\plugins\publish\collect_textureset_images.py", line 117, in create_image_instance
    image_instance.data.update(copy.deepcopy(instance.data))
  File "C:\Program Files\Adobe\Adobe Substance 3D Painter\resources\pythonsdk\lib\copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
[...]

PR #5238, opened 2023-07-04
Substance Painter: Fix Collect Texture Set Images unable to copy.deepcopy due to QMenu

BigRoy:
## Changelog Description

Fix `copy.deepcopy` of `instance.data`.

## Additional info

Resolves https://github.com/ynput/OpenPype/issues/5171 and issue reported on forum https://community.ynput.io/t/substance-painter-error-in-op-3-15-11/443/2

## Testing notes:

1. Start Substance Painter
2. Publish a texture set

### DIFF
--- a/openpype/hosts/substancepainter/plugins/publish/collect_textureset_images.py
+++ b/openpype/hosts/substancepainter/plugins/publish/collect_textureset_images.py
@@ -114,7 +114,7 @@ class CollectTextureSet(pyblish.api.InstancePlugin):
         # Clone the instance
         image_instance = context.create_instance(image_subset)
         image_instance[:] = instance[:]
-        image_instance.data.update(copy.deepcopy(instance.data))
+        image_instance.data.update(copy.deepcopy(dict(instance.data)))
         image_instance.data["name"] = image_subset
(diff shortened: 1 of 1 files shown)


BigRoy:
@iLLiCiTiT any idea why this might have become a new issue in recent versions of OpenPype. The issue originated from the fact that `instance.data` was of type 'pyblish.plugin._Dict' and deep copying it somehow tried to also deepcopy a `PySide2.QtWidgets.QMenu` which then failed to deep copy. 

It's likely the [`parent` attribute on the `pyblish.plugin._Dict`](https://github.com/pyblish/pyblish-base/blob/7e43e5663572a6d9a23ab9d6993c1708094cc5f8/pyblish/plugin.py#L672-L694) but I wonder why it's suddenly a new issue and hadn't occurred before. As [reported on the forum](https://community.ynput.io/t/substance-painter-error-in-op-3-15-11/443?u=bigroy) it:

> There is an error after updating Op 3.15.11 in Substance Painter, however it works fine with 3.15.10.

LiborBatek (approved):
All works as it suppose to do.

Successfully publish all texture from the set.

PR #5240, opened 2023-07-04
RepairAction and SelectInvalidAction filter instances failed on the exact plugin

BigRoy:
## Changelog Description

RepairAction and SelectInvalidAction actually filter to instances that failed on the exact plugin - not on "any failure"

## Additional info

This [comment](https://github.com/ynput/OpenPype/pull/5231#discussion_r1251954530) showcased that actually the repair and select actions were running on ANY plugin that errored, not necessarily the ones that errored on the plugin we're running the action on.

I haven't tried this code - so please test. But it showcases what I think could be a solution.

I've removed the `pyblish.api.instances_by_plugin(errored_instances, plugin)` filtering because whenever an error occurred for the instance for this particular plug-in then of course the instance was related to that plugin - at least that's what I assumed.

Likely this hasn't been spotted much earlier because MOST implementations of repair and select invalid relied on a `get_invalid` action that in itself would still be checking whether the Instance was invalid for that plug-in. But it would basically be uselessly running on instances that hadn't actually errored in the first place.

## Testing notes:

1. Reproduce a scenario like described [here](https://github.com/ynput/OpenPype/pull/5231#discussion_r1251954530)

### DIFF
--- a/openpype/pipeline/publish/lib.py
+++ b/openpype/pipeline/publish/lib.py
@@ -577,12 +577,14 @@ def remote_publish(log, close_plugin_name=None, raise_error=False):
                 raise RuntimeError(error_message)
 
 
-def get_errored_instances_from_context(context):
+def get_errored_instances_from_context(context, plugin=None):
     """Collect failed instances from pyblish context.
 
     Args:
         context (pyblish.api.Context): Publish context where we're looking
             for failed instances.
+        plugin (pyblish.api.Plugin): If provided then only consider errors
+            related to that plug-in.
 
     Returns:
         List[pyblish.lib.Instance]: Instances which failed during processing.
@@ -592,6 +594,9 @@ def get_errored_instances_from_context(context):
     for result in context.data["results"]:
         if result["instance"] is None:
             # When instance is None we are on the "context" result
+            continue
+
+        if plugin is not None and result.get("plugin") != plugin:
             continue
 
         if result["error"]:
--- a/openpype/action.py
+++ b/openpype/action.py
@@ -49,7 +49,7 @@ def deprecated(new_destination):
 
 
 @deprecated("openpype.pipeline.publish.get_errored_instances_from_context")
-def get_errored_instances_from_context(context):
+def get_errored_instances_from_context(context, plugin=None):
     """
     Deprecated:
         Since 3.14.* will be removed in 3.16.* or later.
@@ -57,7 +57,7 @@ def get_errored_instances_from_context(context):
 
     from openpype.pipeline.publish import get_errored_instances_from_context
 
-    return get_errored_instances_from_context(context)
+    return get_errored_instances_from_context(context, plugin=plugin)
 
 
 @deprecated("openpype.pipeline.publish.get_errored_plugins_from_context")
@@ -97,11 +97,9 @@ class RepairAction(pyblish.api.Action):
 
         # Get the errored instances
         self.log.info("Finding failed instances..")
-        errored_instances = get_errored_instances_from_context(context)
-
-        # Apply pyblish.logic to get the instances for the plug-in
-        instances = pyblish.api.instances_by_plugin(errored_instances, plugin)
-        for instance in instances:
+        errored_instances = get_errored_instances_from_context(context,
+                                                               plugin=plugin)
+        for instance in errored_instances:
             plugin.repair(instance)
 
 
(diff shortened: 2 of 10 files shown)


openpype/pipeline/publish/lib.py
```diff
@@ -594,6 +596,9 @@ def get_errored_instances_from_context(context):
             # When instance is None we are on the "context" result
             continue
 
+        if plugin is not None and result.get("plugin") != plugin:
+            continue
```
BigRoy, line 600: It could be that this should be:
```suggestion
        if plugin is not None and isinstance(result.get("plugin"), plugin):
            continue
```
If the result["plugin"] is an instance and `plugin` is the class. I haven't tested the code itself. If it fails, then try this change ;) @mustafa-zarkash

iLLiCiTiT: Why not to use `pyblish.api.instances_by_plugin(errored_instances, plugin)`?

BigRoy: Because that returns the related instances - but not if those errored. We need to find only those instances that errored for this particular plug-in.

iLLiCiTiT: I meant at the end of function (where are already filtered out).
```
    instances = list()
    for result in context.data["results"]:
        if result["instance"] is None:
            # When instance is None we are on the "context" result
            continue

        if result["error"]:
            instances.append(result["instance"])
    
    if plugin is None:
        return instances
    return pyblish.api.instances_by_plugin(instances, plugin)
```

mustafa-zarkash: @iLLiCiTiT  
I will use `SelectInvalidAction` 
This roughly the base code to get_invalid 
```python
errored_instances = get_errored_instances_from_context(context)
instances = pyblish.api.instances_by_plugin(errored_instances, plugin)
for instance in instances:
    invalid_nodes = plugin.get_invalid(instance)
``` 
I don't know why logging doesn't work in this action so, I added a line at the beginning of each `get_invalid`
`cls.log.debug("This is '%s' and '%s'" % ( cls.__name__ , hou.node(instance.data["instance_node"]) ) )` 

and here's my test case : 
I have 4 pointcaches: 
1) `pointcacheMain` has nor errors 
2) `pointcacheMain2` passed outputsop validation but did pass path validation
3) `pointcacheMain3` didn't pass outputsop (`sop_path` -> None)
4) `pointcacheMain4` didn't pass outputsop (`sop_path` -> `ObjNode`) 

--- 
at the beginning I thought that `Select Invalid` will be called for all plugins 
which is not true! 
however I have a different question now, 
why `Select Invalid` works with all errored instances that are associated with the plugin that action was triggered from 
even if that instance has passed that particular plugin but errored because of another plugin 
> i.e. I think `Select Invalid` on `Invalid output nodes(s)` should have worked only with the listed instances beneath it! 


image

iLLiCiTiT: Now I'm not sure if that is different issue or it is response to my previous comment?

mustafa-zarkash: I was writing that comment before but it took me some time to test so that I can have a good visual example of the problem 

> But I agree that there could be some changes that may resolve wrong instances. So it's probably not a good idea.
I agree with that, do you think we should put that example in an issue as a reminder?

iLLiCiTiT: @BigRoy Checked `pyblish.api` and the plugin from result is the class, so it can be
```
if plugin is not None and result.get("plugin") is not plugin:
```

mustafa-zarkash: my previous comment was irrelevant sorry,  
I tested it in houdini and it looks good!

iLLiCiTiT:
I think we can mark for ready for review?

openpype/pipeline/publish/lib.py
BigRoy: Thanks @mustafa-zarkash what you described as an issue here with it running over any instance that had **any error** even if not related to the plugin the action is for that is indeed what this PR intends to solve. So if that now works, all good! :)

BigRoy:
@LiborBatek can you run some "repair" and "select invalid" actions on publishing in Maya (and maybe Nuke if it has that) just to confirm this doesn't start breaking some other stuff in other areas?

mustafa-zarkash:
I tested it with Houdini and it looks good!

LiborBatek (approved):
I did multiple test with different ```Validators``` with ```Actions``` when erroring out and all working fine...

- [ x ] MeshHasUvs ```Select Invalid```
- [ x ] Unique IDs on mesh ```Repair```
- [ x ] Image Prefixes / RenderSettings ```Repair```

All works as before this PR so all good.

PR #5242, opened 2023-07-04
CreatePlugin: Get next version helper

iLLiCiTiT:
## Changelog Description
Implemented helper functions to get next available versions for create instances.

## Additional info
The helper functions are doing what they can do with available data on instances to find next possible available version for an instance. I would be scared to use the value for publishing as publishing may have different requirements in different cases so it should be always optional. This is a quick PR to unblock [custom staging directory PR](https://github.com/ynput/OpenPype/pull/5207).

Other possible usages need more brainstorm. About how to use it, before using it everywhere. e.g. to show the version in UI (and possibly change it?) would mean that the value must be in instance data.

## Testing notes:
Since the code is not used, it must be faked.
1. Add a creator that would call `self.get_next_versions_for_instances` on `collect_instances` and on `create`.
2. Validate that the functions are not crashing and are returning right values.

### DIFF
--- a/openpype/pipeline/create/__init__.py
+++ b/openpype/pipeline/create/__init__.py
@@ -4,6 +4,11 @@ from .constants import (
     PRE_CREATE_THUMBNAIL_KEY,
 )
 
+from .utils import (
+    get_last_versions_for_instances,
+    get_next_versions_for_instances,
+)
+
 from .subset_name import (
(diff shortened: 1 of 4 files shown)


jakubjezek001 (approved):
Works as described. LGTM ;)

PR #5244, opened 2023-07-05
Deadline: OPENPYPE_VERSION should only be added when running from build

tokejepsen:
## Changelog Description
When running from source the environment variable `OPENPYPE_VERSION` should not be added. This is a bugfix for the feature #4489

## Testing notes:
1. Run OP from source.
2. Setup render in Maya.
3. Submit to Deadline and verify there is no `OPENPYPE_VERSION` environment variable on the publish job.


LiborBatek (approved):
All works as supposed to work. Observing Jobs properties (Publish Job) there arent any traces of OP version so all good!

image

mkolar (approved):

PR #5245, opened 2023-07-05
Applications: Environment variables order

iLLiCiTiT:
## Changelog Description
Changed order of set environment variables. First are set context environment variables and then project environment overrides. Also asset and task environemnt variables are optional.

## Additional info
This change gives option to use context environment variables in project environment variables.

## Testing notes:
1. Add custom env variable to project environments with context variable (e.g. `"MY_ENV": "Test: {AVALON_PROJECT}")
2. Launch an application (e.g. nuke, maya)
3. Check if the env is there and has filled context variable


BigRoy:
I'm not sure the PR description is complete. The code changes look like it's trying to make a change so that application launching can work without requiring an asset and a task but the description doesn't seem to mention that at all. Maybe I'm reading the code changes wrong?

Could you try and explain once more what this intends to do?

---

Regarding the removal of asset+task for applications: As far as I know dependencies on asset + task are widespread across OpenPype or best-case scenario the behavior is undefined. So I suspect this will be a major change if we were to allow that.

If I launch an application without asset - do I never get any "tools" assigned? Or which ones do I get?
If I launch an application without asset/task - what should it base frame ranges, etc. on? I'm pretty sure most logic currently relies on asset document data to be valid.

iLLiCiTiT:
> The code changes look like it's trying to make a change so that application launching can work without requiring an asset and a task but the description doesn't seem to mention that at all.

Quote from description: "Also asset and task environemnt variables are optional."

> Regarding the removal of asset+task for applications: As far as I know dependencies on asset + task are widespread across OpenPype or best-case scenario the behavior is undefined. So I suspect this will be a major change if we were to allow that.

This change gives the option to launch application with project only (so only the environment variable for project is set)... -> might be handy in custom applications without integration or review applications, where project might be all you need. Is not meant for applications like maya or nuke (but even there it might be handy).

That is side effect change BTW, the main change is the order of applying of project environemnts which happens after context environments.

openpype/lib/applications.py
```diff
@@ -1649,11 +1649,7 @@ def prepare_context_environments(data, env_group=None, modules_manager=None):
     project_doc = data["project_doc"]
     asset_doc = data["asset_doc"]
     task_name = data["task_name"]
```
BigRoy, line 1651: Since these are now not checked - let's move them closer to where they are used lower down. Also, since `asset` now becomes optional it seems, shouldn't this be `data.get("asset_doc")` and `data.get("task_name")`?

openpype/lib/applications.py
```diff
@@ -1666,26 +1662,31 @@ def prepare_context_environments(data, env_group=None, modules_manager=None):
...
-    if not app.is_host:
+
+    # Apply project specific environments on current env value
+    # - apply them once the context environments are set
+    apply_project_environments_value(
+        project_name, data["env"], project_settings, env_group
+    )
+
+    if not app.is_host or not asset_doc or not task_name:
         return
```
BigRoy, line 1690: Is the idea that anything below this ONLY applies to any application that is a host? If so, it might be much clearer if we move that into a dedicated method `_prepare_host_context_environments` or alike so we can do:
```python
if app.is_host:
    self.prepare_host_context_environments(...)
```

Also, if asset_doc or task_name is not set but the application is a host we still intend to force those to be set, yes? If that's the case, then this should not just return but raise or log an error.
```python
if app.is_host:
    if not asset_doc or not task_name:
        raise RuntimeError("Missing required asset and task name for application launch with host support.")
    self.prepare_host_context_environments(...)
```

openpype/lib/applications.py
iLLiCiTiT: The keys are always set.

openpype/lib/applications.py
iLLiCiTiT: I think you're overthinking this change. It just fill environment variables if the values are available all the time. It is not a reason to crash, and I don't think that asset and task is a "must have" requirement for a host to launch.

BigRoy: > EDITED:
> The only change is, that if you would try to launch host without asset or task, you would at least know in which project it was launched.

Ok. In that case the testing notes should include a thorough check whether each DCC still works as expected for those cases. I'm pretty sure loaders, publishers, etc. will fail to work without asset + task available.

I'm pretty sure workfiles will fail for example. If that's the case I wouldn't expose it _without any warning or error_ as if things are all good just because in a far future "we might support it". To me it sounds like massive amounts of changes will be needed in the codebase to support a future where all DCCs can be launched without asset+task.. but I might be wrong, maybe it's just as easy as not setting `AVALON_ASSET` and `AVALON_TASK` and all will work. 🤞 

Also, as asked before - what's the expected behavior for Tools to be applied when launching an application without an asset/task since those are set per asset/folder? We'll need some fallback that in those cases then starts using the project defaults?

openpype/lib/applications.py
```diff
@@ -1657,28 +1653,42 @@ def prepare_context_environments(data, env_group=None, modules_manager=None):
...
+    apply_project_environments_value(
+        project_name, data["env"], project_settings, env_group
+    )
+
     if not app.is_host:
         return
 
+    data["env"]["AVALON_APP"] = app.host_name
+
+    if not not asset_doc or not task_name:
```
Minkiu, line 1685: is the double `not ` intentional?

openpype/lib/applications.py
```diff
@@ -1657,28 +1653,42 @@ def prepare_context_environments(data, env_group=None, modules_manager=None):
...
+        # QUESTION replace with log.info and skip workfile discovery?
+        # - technically it should be possible to launch host without context
```
Minkiu, line 1687: Does this affect the discoverability of the launcher? It would be cool to still allow people to launch hosts without a proper context

iLLiCiTiT: > Does this affect the discoverability of the launcher? 

No. Launcher is a different story...

openpype/lib/applications.py
iLLiCiTiT: Nope. Thanks

antirotor (approved):
make sense, didn't test it

PR #5249, opened 2023-07-05
General: add the os library before os.environ.get

moonyuet:
## Changelog Description
Adding os library into `creator_plugins.py` due to `os.environ.get` in line 667

## Additional info
n/a

## Testing notes:
1. Start Maya, or any dccs set as OP hosts
2. Load something from the loader


LiborBatek (approved):
All assets loads normally when using ```Load```.

No any errors seen.

PR #5252, opened 2023-07-05
Update scene inventory even if any errors occurred during update

BigRoy:
## Changelog Description

When selecting many items in the scene inventory to update versions and one of the items would error out the updating stops. However, before this PR the scene inventory would also NOT refresh making you think it did nothing.

Also implemented as method to allow some code deduplication.

## Additional info

We might want to change the `AssertionError` to show a popup to something more generic too to show ANY error as a dialog popup - but to remain consistent with old behavior I haven't done that. However, doing that might at least make it clearer to end user that something actually failed.

## Testing notes:

Starts updating and switching versions:
2. Update versioned containers randomly
3. Update versioned containers to explicit version (using Set version dialog)
4. Update containers to latest using top right update button
5. Update selected containers to latest using context menu "Update to latest"
5. Switch to hero versions and back
6. Use switch asset

### DIFF
--- a/openpype/tools/sceneinventory/view.py
+++ b/openpype/tools/sceneinventory/view.py
@@ -1,5 +1,6 @@
 import collections
 import logging
+import itertools
 from functools import partial
 
 from qtpy import QtWidgets, QtCore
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
Essentially it does what it should...updates assets even when some error occurs.

Question being if the error message shouldnt inform the user which asset caused the issue (in my case non existing [Hero ver] for Camera asset. See vid below

For you to consider / also not sure if its proper place to discuss or if its part of this PR.

https://github.com/ynput/OpenPype/assets/112623825/b81b8fbd-67cf-46e1-ac20-4049f8821148

iLLiCiTiT:
> I'd say if we can be more informative then, definitely we should do that! :) But better tracked as a separate issue and PR.

I do agree that the issue is not caused by this PR.

> We might want to change the AssertionError to show a popup to something more generic too to show ANY error as a dialog popup

100% agree but also task for future PR.

PR #5253, opened 2023-07-05
Maya: Reduce amount of code for Collect Looks

BigRoy:
## Changelog Description

- Refactor `get_file_node_files` because popping from `paths` by index should have been done in reversed order anyway. It's now changed to not need popping at all.
- Removed unused `RENDERER_NODE_TYPES` and if-branch which collected `node_attrs` list which was unused + collected members which was also done outside of the if branch and thus generated no extra data.
- Collected all materials from look set attributes at once instead of multiple queries
- Collected all file nodes in history from a single query instead of per type
- Restructured assignment of `instance.data["resources"]` to be more readable
- Cached `PXR_NODES` only ones (Note: plugin load is checked on discovery of the collect look plugin) instead of querying plugin load and its nodes per file node per attribute
- Removed some debug logs or combined some messages

## Additional info

Also mentioned on Discord [here](https://discord.com/channels/517362899170230292/517382145552154634/1126252092202430555)

## Testing notes:

Preferably publish looks with different file node types (e.g. `file` and `aiImage` with and without UDIMs or sequences) for each renderer type:

1. Publish looks
    - [x] arnold (mtoa)
    - [x] redshift
    - [x] renderman
    - [x] v-ray

2. Ensure look assignments still work as intended 
    - [x] shader assignments
    - [x] render stat attributes on meshes 
    - [x] (Redshift) mesh parameters object sets, etc.


LiborBatek:
All tested and functional!  
*missing test for Renderman which I m missing license for. Will resolve asap.

tokejepsen:
@LiborBatek did you manage to get a Renderman license?

tokejepsen:
@BigRoy could you resolve the conflict?

LiborBatek (changes requested):
~I have finally make Renderman working and facing an issue that renderman natively connects maya shader onto ShapeNode and connects ```pxrShader``` into dedicated slot similar to arnold etc,~

~When then publishing Look I will get the Lambert shaders instead of those Renderman ones...~

~I guess we should resolve it so user is not forced to reconnect anything manually each time assigning new material to ShapeNode.~

~Here is an example: ~
image

~When publish I get ```lambert2``` instead of ```PxrDiffuse1``` in the ```Look``` asset. Not sure if this behavior is newly introduced in Renderman (25.2) or been there even before.~

LiborBatek (approved):
Nevermind my last review...I didnt checked the shaders correctly...seem working ok.

Works also for ```Renderman```.
image

tokejepsen (approved):
Code looks good.

PR #5254, opened 2023-07-06
Missing os import

tokejepsen:
## Changelog Description
Missing os module import causing below error.

image

## Testing notes:
1. Open a DCC, for example Maya.
2. Load any subset, for example rig, in and verify no errors occur.


tokejepsen:
Already fixed in https://github.com/ynput/OpenPype/pull/5249

PR #5263, opened 2023-07-07
Error with multiple loads from 1 placeholder

SherineKB:
## Changelog Description

Bug : If we enable the set suppression, we have an error from the 2nd object because it deletes the set with it.
`TypeError: No object matches name: PLACEHOLDERS_SET`

To fixe it, we need to firstly remove the object from the set and secondly delete it, so it keeps the set.

**To test it :** 
Build workfile 
See if the rig assets linked to the shot are loaded (refer to Ftrack).

### DIFF
--- a/openpype/hosts/maya/api/workfile_template_builder.py
+++ b/openpype/hosts/maya/api/workfile_template_builder.py
@@ -254,8 +254,7 @@ class MayaPlaceholderLoadPlugin(PlaceholderPlugin, PlaceholderLoadMixin):
 
     def delete_placeholder(self, placeholder):
         """Remove placeholder if building was successful"""
-        cmds.sets(placeholder.scene_identifier, rm= PLACEHOLDER_SET )
-        print (placeholder.scene_identifier)
+        cmds.sets(placeholder.scene_identifier, rm= PLACEHOLDER_SET)
         cmds.delete(placeholder.scene_identifier)
 
     def load_succeed(self, placeholder, container):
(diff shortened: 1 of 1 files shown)


openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -254,6 +254,7 @@ def post_placeholder_process(self, placeholder, failed):
 
     def delete_placeholder(self, placeholder):
         """Remove placeholder if building was successful"""
+        cmds.sets(placeholder.scene_identifier, rm= PLACEHOLDER_SET)
```
BigRoy, line 257:
```suggestion
        cmds.sets(placeholder.scene_identifier, rm=PLACEHOLDER_SET)
```

Also, **why** does this fix the issue? What was the original error?

BigRoy:
I'm pretty sure someone might be getting issues with this branch name at some point:

image

fabiaserra:
this PR I created a while back https://github.com/ynput/OpenPype/pull/5061 should solve this

mustafa-zarkash:
@BigRoy good observation! 

May I suggest using branch names similar to gitflow.

```
enhancement/{issue_number}-{host or module}-{problem title}

feature/{issue_number}-{host or module}-{problem title}

bugfix/{issue_number}-{host or module}-{problem title}

documentation/{issue_number}-{host or module}-{problem title}

tests/{issue_number}-{host or module}-{problem title}
```

openpype/hosts/maya/api/workfile_template_builder.py
SherineKB: Because when you delete the only object in a set, the set is deleted. 
By removing the object and then deleting it, the set is kept.

PR #5264, opened 2023-07-07
Feature/blender handle q application

Laroustine:
## Changelog Description
This edit is to change the way the QApplication is run for Blender. It calls in the singleton (QApplication) during the register. This is made so that other Qt applications and addons are able to run on Blender. In its current implementation, if a QApplication is already running, all functionality of OpenPype becomes unavailable.

## Testing notes:
1. Start a QApplication in an other addon in Blender
2. Click on OpenPype and select any option


simonebarbieri:
I wasn't aware of this, especially because I have never used other Blender addons with Qt before. I've tried creating a new QApplication (in `develop`) and I can see the problem. But then I tried from this branch, and while it always creates the QApplication for OpenPype, I cannot create a new QApplication.

```
>>> from qtpy import QtWidgets, QtCore
>>> test = QtWidgets.QApplication()
  File "C:\Program Files\Blender Foundation\Blender 3.5\3.5\python\lib\code.py", line 90, in runcode
    exec(code, self.locals)
  File "<blender_console>", line 1, in <module>
RuntimeError: Please destroy the BlenderApplication singleton before creating a new QApplication instance.
```

From the description of the PR I assumed that it should work, though?

Laroustine:
This edit was made so that other Qt applications could use the instance of QApplication made by OpenPype with:
``
QtWidgets.QApplication.instance()
``
Otherwise we would need to kill the instance made by the other Application and then create the instance made by OpenPype.

simonebarbieri (approved):

PR #5265, opened 2023-07-07
Windows: Support long paths on zip updates.

tokejepsen:
## Changelog Description
Support long paths for version extract on Windows.

Use case is when having long paths in for example an addon. You can install to the C drive but because the zip files are extracted in the local users folder, it'll add additional sub directories to the paths and quickly get too long paths for Windows to handle the zip updates.

## Testing notes:
1. Setup a long path in for example `addons` folder with multiple subfolders.
2. Create zip and try to update with the zip.

### DIFF
--- a/igniter/bootstrap_repos.py
+++ b/igniter/bootstrap_repos.py
@@ -77,7 +77,6 @@ def sha256sum(filename):
 
 
 class ZipFileLongPaths(ZipFile):
-
     def _extract_member(self, member, targetpath, pwd):
         return ZipFile._extract_member(
             self, member, sanitize_long_path(targetpath), pwd
--- a/igniter/bootstrap_repos.py
+++ b/igniter/bootstrap_repos.py
@@ -34,6 +34,20 @@ LOG_WARNING = 1
 LOG_ERROR = 3
 
 
+def sanitize_long_path(path, encoding=None):
+    if platform.system().lower() == "windows":
+        path = os.path.abspath(path)
+
+        if path.startswith("\\\\"):
+            path = "\\\\?\\UNC\\" + path[2:]
+        else:
+            path = "\\\\?\\" + path
+
+        return path
+
+    return path
+
+
 def sha256sum(filename):
     """Calculate sha256 for content of the file.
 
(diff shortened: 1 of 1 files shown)


igniter/bootstrap_repos.py
```diff
@@ -34,6 +34,20 @@
 LOG_ERROR = 3
 
 
+def sanitize_long_path(path, encoding=None):
```
antirotor, line 37: `encoding` doesn't seem to be used. Also I would add docstring to explain what it does and why. And one last thing, I would change a little bit the logic, so it bails out from the function if platfrom isn't windows. It should help to make it more readable.

antirotor (changes requested):
just a little cosmetic issue

igniter/bootstrap_repos.py
tokejepsen: Not entirely sure what you mean by bailing out early to make it more readable, but I've tried something.

igniter/bootstrap_repos.py
```diff
@@ -34,6 +34,31 @@
...
+        return path
+    else:
+        path = os.path.abspath(path)
+
+        if path.startswith("\\\\"):
+            path = "\\\\?\\UNC\\" + path[2:]
+        else:
+            path = "\\\\?\\" + path
+
+        return path
```
iLLiCiTiT, line 59:
```suggestion
    path = os.path.abspath(path)

    if path.startswith("\\\\"):
        path = "\\\\?\\UNC\\" + path[2:]
    else:
        path = "\\\\?\\" + path
    return path
```

igniter/bootstrap_repos.py
```diff
@@ -53,6 +78,14 @@ def sha256sum(filename):
     return h.hexdigest()
 
 
+class ZipFileLongPaths(ZipFile):
+
+    def _extract_member(self, member, targetpath, pwd):
```
iLLiCiTiT, line 83:
```suggestion
    def _extract_member(self, member, targetpath, pwd):
```

iLLiCiTiT (approved):
LGTM (I should use it in ayon-launcher too).

antirotor (approved):

PR #5273, opened 2023-07-11
Colorspace: updating file path detection methods

jakubjezek001:
## Changelog Description
Support for OCIO v2 file rules integrated into the available color management API

## Additional info
- [x] support for ocio v2 file rules 
- [x] cashing python compatibility for unit testing switch
- [x] ocio config version compatibility check
- [x] parsing from file path improvements
- [x] merging parsing with file rules - since it is related
- [x] cashing available colorspace names from a ocio.config

## Testing notes:
1. unittesting `poetry run python ./start.py runtests [abs path to repo dir]\tests\unit\openpype\pipeline\test_colorspace.py`


openpype/pipeline/colorspace.py
```diff
@@ -118,6 +119,19 @@ def get_imageio_colorspace_from_filepath(
         if ext_match and file_match:
             colorspace_name = file_rule["colorspace"]
 
+    # if no file rule matched, try to get colorspace
+    # from filepath with OCIO v2 way
+    # QUESTION: should we override file rules from our settings and
+    #           in ocio v2 only focus on file rules set in config file?
```
BigRoy, line 125: What if we just have a toggle in settings that toggles "Use OpenPype file rules" - when enabled it uses the file rules from settings, when disabled it uses the file rules from the OCIO file if it's OCIO v2?

It would potentially have use do much fewer compatibility checks. Also, it's best to assume `get_imageio_colorspace_from_filepath` will be called many times. I assume the compatibility check can be quite slow, especially if we're going through the subprocess wrapper so we might want to optimize it so we don't need to query it per file and maybe store it with the data?

openpype/pipeline/colorspace.py
```diff
@@ -206,8 +251,9 @@ def validate_imageio_colorspace_in_config(config_path, colorspace_name):
     return True
 
 
+# TODO: remove this in future - backward compatibility
 def get_data_subprocess(config_path, data_type):
-    """Get data via subprocess
+    """[Deprecated] Get data via subprocess
```
BigRoy, line 256: Anything marked deprecated should have a deprecated function decorator added so that whenever it gets called it prints the deprecation message including information what the replacement function should be if that exists.

jakubjezek001: done ;)

openpype/pipeline/colorspace.py
jakubjezek001: I wonder if you could cache the version compatibility check in the `CashedData` class. This way, the check would only be performed once, assuming the context is not changed to another configuration with a different version. However, I believe this could be an edge case that we don't need to worry about.

jakubjezek001: @BigRoy look to ba237487a65176f2ee58e08533c65115b5a9c19d for more about the cashing of config version data.

jakubjezek001: I could ofcourse make a current context of asset name for chacking of the cache validity.

openpype/pipeline/colorspace.py
```diff
@@ -19,6 +22,54 @@
...
+                ).format(decorated_func.__name__, warning_message),
+                category=DeprecatedWarning,
+                stacklevel=4
+            )
+            return decorated_func(*args, **kwargs)
+        return wrapper
+
+    if func is None:
+        return _decorator
+    return _decorator(func)
```
BigRoy, line 72: It's so unfortunate that basically the same logic exists in `openpype.lib` and we need to duplicate it here because `openpype.lib` already imports `openpype.pipeline.colorspace`. I feel like that somehow this `deprecation` decorator logic should be in a module within openpype that we can import from everywhere else.

Would be nice if it didn't need to duplicate the code here.

openpype/pipeline/colorspace.py
```diff
@@ -105,7 +156,7 @@ def get_imageio_colorspace_from_filepath(
 
     # match file rule from path
     colorspace_name = None
-    for _frule_name, file_rule in file_rules.items():
+    for _, file_rule in file_rules.items():
```
BigRoy, line 159:
```suggestion
    for file_rule in file_rules.values():
```

openpype/pipeline/colorspace.py
```diff
@@ -132,52 +197,87 @@ def get_imageio_colorspace_from_filepath(
...
     Returns:
         str: name of colorspace
     """
-    if not config_data:
-        project_settings = project_settings or get_project_settings(
-            project_name
+    if not colorspaces and not config_path:
+        raise ValueError(
+            "You need to provide `config_path` if you don't "
+            "want to provide input `colorspaces`."
```
BigRoy, line 261:
```suggestion
        raise ValueError(
            "Missing either required input `colorspaces` or `config_path`."
```
_If you don't want_ doesn't really sound like what I'd expect in a developer message.

Or:
```suggestion
        raise ValueError(
            "Must provide at least one of `colorspaces` or `config_path` input arguments."
```

openpype/pipeline/colorspace.py
```diff
@@ -132,52 +197,87 @@ def get_imageio_colorspace_from_filepath(
...
         # check underscored variant of colorspace name
         # since we are reformatting it in integrate.py
-        if colorspace_key.replace(" ", "_") in path:
+        if colorspace_key.replace(" ", "_") in filepath:
             colorspace_name = colorspace_key
             break
-        if colorspace_key in path:
+        if colorspace_key in filepath:
             colorspace_name = colorspace_key
             break
```
BigRoy, line 276: Here's a fun trick that can be much faster potentially for many colorspaces if you need to process many files.

```python
import re


def get_colorspace_match_regex(colorspaces):
    """Return a regex which allows to search a colorspace match in a filename

        Args:
            colorspaces (list): List of colorspace names
            case_sensitive (bool, optional): Whether the match should
                be case sensitive or not. Defaults to True.

    """
    pattern = "|".join(
        # Allow to match spaces also as underscores because the
        # integrator replaces spaces with underscores in filenames
        re.escape(colorspace).replace("\ ", "[_ ]") for colorspace in 
        # Sort by longest first so the regex matches longer matches
        # over smaller matches, e.g. matching 'Output - sRGB' over 'sRGB'
        sorted(colorspaces, key=len, reverse=True)
    )
    return re.compile(pattern)


regex = get_colorspace_match_regex(colorspaces)
colorspaces = ["Utility - Raw",  'ACES - ACEScg', "Output - sRGB","acescg", "linear", "ACES - ACEScc", "sRGB"]
for filepath in ["file1.sRGB.jpg", "file2.ACES - ACEScg.exr", "/path/to/aces_testfile_acescg.exr"]:
    match = regex.search(filepath)
    colorspace = match.group(0) if match else None
    print("{: <25} --> {}".format(colorspace, filepath))
```

openpype/pipeline/colorspace.py
```diff
@@ -221,8 +323,46 @@ def get_data_subprocess(config_path, data_type):
             "config", data_type,
             "--in_path", config_path,
             "--out_path", tmp_json_path
+        ]
+        log.info("Executing: {}".format(" ".join(args)))
```
BigRoy, line 327: This is a debug log.
```suggestion
        log.debug("Executing: {}".format(" ".join(args)))
```

openpype/pipeline/colorspace.py
```diff
@@ -221,8 +323,46 @@ def get_data_subprocess(config_path, data_type):
...
+
+        process_kwargs = {
+            "logger": log
+        }
+
+        run_openpype_process(*args, **process_kwargs)
+
+        # return all colorspaces
+        return_json_data = open(tmp_json_path).read()
+        return json.loads(return_json_data)
```
BigRoy, line 337: This should be:
```suggestion
        # return all colorspaces
        with open(tmp_json_path, "r") as f:
            return json.load(f)
```
Also further down in next function.

openpype/pipeline/colorspace.py
BigRoy, line 327: Even though this function is deprecated you've basically duplicated the inner logic here into the `get_wrapped_with_subprocess` function down below so I'd recommend at least cleaning up this code to to start the re-use.

openpype/pipeline/colorspace.py
```diff
@@ -118,6 +169,20 @@ def get_imageio_colorspace_from_filepath(
...
+        compatibility_check_config_version(config_data["path"], major=2)
+        and not colorspace_name
+    ):
+        colorspace_name = get_colorspace_from_filepath(
+            config_data["path"], path)
+
+    # use parse colorspace from filepath as fallback
+    colorspace_name = colorspace_name or parse_colorspace_from_filepath(
+        path, config_path=config_data["path"]
+    )
```
BigRoy, line 184: By adding these into this function the docstring and function names feels incorrect - this isn't at all about imageio file rules in settings suddenly but it's doing way more fallbacks than that. Even worse is that with this here it's now even more unclear what the difference is between `get_colorspace_from_filepath`, `parse_colorspace_from_filepath` and `get_imageio_colorspace_from_filepath` since they start overlapping so much. I feel like adding these here is the wrong approach. If we want a "match any of the rules" function we'll likely just need a dedicated function for that which then combines all three in the right order.

With this added here it's now also to JUST compare `imageio filerules` with the filepath.

openpype/pipeline/colorspace.py
BigRoy: By the way, [here's more details as to why this could be faster](https://stackoverflow.com/a/22945732)

openpype/pipeline/colorspace.py
jakubjezek001: I agree but it is not issue for this PR

openpype/pipeline/colorspace.py
```diff
@@ -221,8 +323,46 @@ def get_data_subprocess(config_path, data_type):
...
+        }
+
+        run_openpype_process(*args, **process_kwargs)
+
+        # return all colorspaces
+        return_json_data = open(tmp_json_path).read()
+        return json.loads(return_json_data)
+
+
+def get_wrapped_with_subprocess(command_group, command, **kwargs):
```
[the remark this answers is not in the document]
jakubjezek001: Good question, I haven't found an alternative but @iLLiCiTiT might see it differently?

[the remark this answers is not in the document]
iLLiCiTiT: Don't understand the question? But this function should be 100% private (start with `_`)

openpype/pipeline/colorspace.py
```diff
@@ -237,18 +376,45 @@ def get_data_subprocess(config_path, data_type):
...
+    # compatible
+    return CachedData.python3compatible
+
+
+def compatibility_check_config_version(config_path, major=1, minor=None):
+    """Making sure PyOpenColorIO config version is compatible"""
+
+    if not CachedData.config_version_data:
+        if compatibility_check():
+            from openpype.scripts.ocio_wrapper import _get_version_data
```
iLLiCiTiT, line 398: The same as above. This function should be here.

openpype/pipeline/colorspace.py
iLLiCiTiT: I do agree with the suggestion, keep the name with underscode (even if is unused), or use `.values()`.

openpype/pipeline/colorspace.py
jakubjezek001: Agree with you on this. Look to a6b8fdfe33dc5dcf1eb1120a65f5ad264b3b1ef4 if it makes more sence this way.

openpype/pipeline/colorspace.py
```diff
@@ -132,52 +197,86 @@ def get_imageio_colorspace_from_filepath(
...
+        )
+        if result_data:
+            return result_data[0]
+
+    from openpype.scripts.ocio_wrapper import _get_colorspace_from_filepath
+
+    result_data = _get_colorspace_from_filepath(config_path, filepath)
+
+    if result_data:
+        return result_data[0]
```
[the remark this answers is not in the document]
jakubjezek001: I completely agree! This approach has always felt somewhat unclean. Let's refactor it, but in a separate pull request.

openpype/pipeline/colorspace.py
```diff
@@ -17,8 +20,56 @@
 log = Logger.get_logger(__name__)
 
 
-class CashedData:
+class CachedData:
     remapping = None
+    python3compatible = None
```
BigRoy, line 25: Admittedly this isn't necessarily about being Py3 compatible - it's just about whether `PyOpenColorIO` can be imported or not in the current running python session

```suggestion
    has_compatible_ocio_package = None
```

jakubjezek001: added via b8e054a50cac66c4e671da0261093be68db555b2

openpype/pipeline/colorspace.py
```diff
@@ -64,120 +115,263 @@ def get_ocio_config_script_path():
...
-            colorspace_name = colorspace_key
-            break
-        if colorspace_key in path:
-            colorspace_name = colorspace_key
-            break
+    colorspaces = colorspaces or get_ocio_config_colorspaces(config_path)
+    underscored_colorspaces = {
+        key.replace(" ", "_"): key for key in colorspaces
+        if " " in key
+    }
```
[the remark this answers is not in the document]
jakubjezek001: Yes it makes sense, haven't thought of the variant with half underscored names. I hope I had understood it correctly. Please have a look here da7ffb84fba2041f7415537e5031485f5835f8e8 for the changes related to this suggestion.

openpype/pipeline/colorspace.py
```diff
@@ -232,23 +450,53 @@ def get_data_subprocess(config_path, data_type):
         run_openpype_process(*args, **process_kwargs)
```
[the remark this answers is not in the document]
jakubjezek001: yep

openpype/pipeline/colorspace.py
```diff
@@ -69,124 +117,264 @@ def get_ocio_config_script_path():
...
+    )
+
+    if not colorspace_name:
+        log.info("No imageio file rule matched input path: '{}'".format(
+            filepath
+        ))
+        return None
+
+    # validate matching colorspace with config
+    if validate and config_data:
```
Minkiu, line 180: If `config_data` wasn't `True` we would have exited at line `151`

openpype/pipeline/colorspace.py
```diff
@@ -69,124 +117,264 @@ def get_ocio_config_script_path():
...
+        config_path (Optional[str]): path to config.ocio file
 
     Returns:
         str: name of colorspace
     """
-    if not config_data:
-        project_settings = project_settings or get_project_settings(
-            project_name
+    def _get_colorspace_match_regex(colorspaces):
+        """Return a regex patter
```
Minkiu, line 331: `patter` -> `pattern`

Minkiu (approved):

antirotor (approved):

openpype/pipeline/colorspace.py
```diff
@@ -20,10 +23,55 @@
 
 
 class CachedData:
-    remapping = {}
-    allowed_exts = {
```
BigRoy, line 24: ⚠️ The removal of this attribute has broken publishing since it's still used on [line 940 of this file](https://github.com/ynput/OpenPype/blob/5d1f2b0d9ed3ff44f1f924870e9a829c9ee4ee12/openpype/pipeline/colorspace.py#L940). ⚠️ 

@jakubjezek001

PR #5274, opened 2023-07-11
deadline shortcut added to settings with type error

FadyFS:
## Changelog Description
We added to json files new keys to deadline so it has a proper interface with its relatives paths on different OS.

## Additional info
While opening the settings you will have a Attribute error, if u have any clue to solve it I'd like your help
## Testing notes:
1. Launch openpype locally
2. Open studio settings,
3. Go to Modules and look for Deadline

### DIFF
--- a/openpype/settings/defaults/system_settings/modules.json
+++ b/openpype/settings/defaults/system_settings/modules.json
@@ -160,12 +160,14 @@
     },
     "deadline": {
         "enabled": false,
-        "deadline_urls": "http://deadline.prs.vfx.int:8082",
+        "deadline_urls": {
+            "default":"http://deadline.prs.vfx.int:8082"
+        },
         "deadline_paths": {
             "default": {
                 "windows": "",
                 "darwin": "",
-                "linux": ""                  
+                "linux": "/prod/softprod/apps/deadline/10.1.23.6/linux/deadlinemonitor10.desktop"
             }
         }
     },
--- a/openpype/settings/defaults/system_settings/modules.json
+++ b/openpype/settings/defaults/system_settings/modules.json
@@ -159,9 +159,14 @@
         "sites": {}
     },
     "deadline": {
-        "enabled": true,
-        "deadline_urls": {
-            "default": "http://127.0.0.1:8082"
+        "enabled": false,
+        "deadline_urls": "http://deadline.prs.vfx.int:8082",
+        "deadline_paths": {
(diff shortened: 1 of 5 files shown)


openpype/settings/defaults/system_settings/modules.json
```diff
@@ -159,9 +159,16 @@
...
+        "enabled": false,
         "deadline_urls": {
-            "default": "http://127.0.0.1:8082"
+            "default":"http://deadline.prs.vfx.int:8082"
+        },
+        "deadline_paths": {
+            "default": {
+                "windows": "",
+                "darwin": "",
+                "linux": "/prod/softprod/apps/deadline/10.1.23.6/linux/deadlinemonitor10.desktop"
```
ClementHector, line 170: replace the linux path by an empty string.

openpype/settings/defaults/system_settings/modules.json
```diff
@@ -159,9 +159,16 @@
         "sites": {}
     },
     "deadline": {
-        "enabled": true,
+        "enabled": false,
         "deadline_urls": {
-            "default": "http://127.0.0.1:8082"
+            "default":"http://deadline.prs.vfx.int:8082"
```
ClementHector, line 164: Put back the default value to default `"http://127.0.0.1:8082"`

openpype/settings/defaults/system_settings/modules.json
```diff
@@ -159,9 +159,16 @@
         "sites": {}
     },
     "deadline": {
-        "enabled": true,
+        "enabled": false,
```
ClementHector, line 162: Put back the value to true

kalisp:
I don't understand what this should be doing. Just adding Deadline Monitor icon (btw path to the icon is hardcoded)?
Could you please elaborate on its use case?

BigRoy:
This PR seems to try and expose a "launch deadline" from the OpenPype launcher?

The PR seems wrong in many ways - hardcoded paths, linux only, doesn't use application launching logic even though it tries to emulate the behavior somewhat with exposing `environment` etc.

I agree with @kalisp this PR is a bit unclear entirely why it's a needed feature - what does it offer and why do you need it like that? If it makes sense I suppose we can discuss cleaning up the code.

kalisp:
Closing for now. Please provide a clear use case for this and reopen.

PR #5275, opened 2023-07-11
Nuke: removing deprecated settings in baking

jakubjezek001:
## Changelog Description
Removing deprecated settings for baking with reformat. This option was only for single reformat node and it had been substituted with multiple reposition nodes.

# Dependency:
https://github.com/ynput/ayon-addons-settings/pull/18


iLLiCiTiT (approved):
LGTM

PR #5276, opened 2023-07-11
275 save the filter selection for family

FadyFS:
## Changelog Description
We implemented a code so the interface remembers the last selection of the family filtered in the loader.

## Testing notes:
1. Open the loader
2. Choose a project and filter families 
3. Launch the software again and go back to the loader


mkolar:

I'm closing it down, but we'll he happy for a new PR to https://github.com/ynput/ayon-core repository once it's up.

PR #5277, opened 2023-07-11
Nuke: auto apply all settings after template build

jakubjezek001:
## Changelog Description
Adding auto run of Apply All Settings after template is builder is finishing its process. This will apply Frame-range, Image size, Colorspace found in context of a task shot.

## Testing notes:
1. Run Template build and see that your workfile will be set to correct


iLLiCiTiT (approved):
Code looks ok, needs testing,

Issue #5280, opened 2023-07-12
Bug: Unable to publish renders in Houdini with AOVs added in OP Arnold ROP


krishna8008:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

Trying to publish renders through OP in Houdini 19.5.569, Publishing is not working with custom AOV's added in the arnold

### Expected Behavior:

Publish should be successful and must see Job in deadline submitted succesfully

### Version

3.15.12

### What platform you are running OpenPype on?

Windows

### Steps To Reproduce:

prepare a sample scene in houdini.
assign material standard surface and create custom aov plugged in it.
create a ROP using OP and add the custom AOV inside.
try to publish the ROP using OP.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

### Additional context:

![image](https://github.com/ynput/OpenPype/assets/61413765/fcb776d5-9c2b-4751-ba16-83281f723a89)

[cuID:[OP-6355](https://app.clickup.com/t/863h6kup4)]

mustafa-zarkash:
@BigRoy 
I have found out something 

https://github.com/ynput/OpenPype/blob/583c4fc15de9c16f9ed47f0e449101f00f6dea5c/openpype/hosts/houdini/plugins/publish/collect_arnold_rop.py#L53

I think it should be
```python
        if not rop.evalParm("ar_enable_aov{}".format(index)):
```


Reference, this is how `enable` parameter looks like in my arnold version 
![image](https://github.com/ynput/OpenPype/assets/20871534/ee39b221-c5d4-417a-89b7-c7ec68519f53)

BigRoy:
Yes, does look like it 👍 - must have been a typo I assume.

BigRoy:
Actually, yes. Pretty sure this was a typo by @antirotor when resolving conflicts - see [this commit](https://github.com/ynput/OpenPype/commit/d37ccb4718e3fa9299d098844ddb0a92c1a09552).

mustafa-zarkash:
@BigRoy  Thanks for confirming, I'm going to make PR if you hadn't do that already XD

PR #5281, opened 2023-07-12
Houdini: fix typo error in collect arnold rop

mustafa-zarkash:
## Changelog Description
Fixing a typo error in `collect_arnold_rop.py`

Reference: [#5280](https://github.com/ynput/OpenPype/issues/5280)

### DIFF
--- a/openpype/hosts/houdini/plugins/publish/collect_arnold_rop.py
+++ b/openpype/hosts/houdini/plugins/publish/collect_arnold_rop.py
@@ -50,7 +50,7 @@ class CollectArnoldROPRenderProducts(pyblish.api.InstancePlugin):
         num_aovs = rop.evalParm("ar_aovs")
         for index in range(1, num_aovs + 1):
             # Skip disabled AOVs
-            if not rop.evalParm("ar_enable_aovP{}".format(index)):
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
Seemed to be a typo indeed - didn't test but this change makes sense.

BigRoy:
> I dont have arnold for Houdini installed (as it needs to be paid), but the code looks good for me.

I believe you should be able to run `mtoa` and `htoa`, etc. without a license - you just won't be able to do (final) renders. I believe you can even render with a watermark if no license is available.

moonyuet:
> > I dont have arnold for Houdini installed (as it needs to be paid), but the code looks good for me.
> 
> I believe you should be able to run `mtoa` and `htoa`, etc. without a license - you just won't be able to do (final) renders. I believe you can even render with a watermark if no license is available.

didn't realize that arnold does have skip license check in htoa too.... I thought they only allow maya to do the rendering with skipping license check but I was wrong! ANd you can do final rendering but with the watermark too.

moonyuet (approved):
image
Looks good with the test

BigRoy:
@antirotor @mkolar ready to merge! :)

PR #5283, opened 2023-07-12
Maya: asset to fixed to render cam

FadyFS:
## Changelog Description
The implementation allow to center the renderable camera to the asset loaded with the workfile template builder. It excludes the  others cameras so it selects only the renderable one, so the specified model  is in the frame.

## Testing notes:
1. Use the branch specific to this issue [here.](https://github.com/quadproduction/OpenPype/pull/541)
2. Follow the same instructions as the PR related to it
3. Don't forget to `git checkout 9-add-center-loaded-asset-in-scene` in openpype-custom-plugins repo (Clone it if not installed yet)

### DIFF
--- a/openpype/hosts/maya/api/workfile_template_builder.py
+++ b/openpype/hosts/maya/api/workfile_template_builder.py
@@ -251,6 +251,32 @@ class MayaPlaceholderLoadPlugin(PlaceholderPlugin, PlaceholderLoadMixin):
         cmds.sets(node, addElement=PLACEHOLDER_SET)
         cmds.hide(node)
         cmds.setAttr(node + ".hiddenInOutliner", True)
+        model = cmds.ls(assemblies  = True)
+
+        filtered_model = []
+        exclude_list = ['persp', 'top', 'front',
+                         'side', 'root', 'wip']
+
+        for element in model:
+            if not any(exclude_elem in element for exclude_elem in exclude_list):
+                filtered_model.append(element)
+
+        cameras = cmds.ls(type = 'camera')
+        renderable_camera = None
+
+        for camera in cameras:
+            if cmds.getAttr(camera + ".renderable"):
+                shapes = cmds.listRelatives(camera,
+                                        parent=True,
+                                        fullPath=True,
+                                        type="transform")
+                if shapes:
+                    renderable_camera = shapes[0]
+                    break
+
+        cmds.lookThru(renderable_camera)
+        cmds.viewFit(filtered_model, f=0.6)
+
 
     def delete_placeholder(self, placeholder):
         """Remove placeholder if building was successful"""
--- a/openpype/hosts/maya/api/workfile_template_builder.py
+++ b/openpype/hosts/maya/api/workfile_template_builder.py
@@ -254,7 +254,7 @@ class MayaPlaceholderLoadPlugin(PlaceholderPlugin, PlaceholderLoadMixin):
         model = cmds.ls(assemblies=True)
 
         filtered_model = []
-        exclude_list = ['persp','top','front','side','root','wip']
+        exclude_list = ['persp', 'top', 'front', 'side', 'root', 'wip']
 
         for element in model:
             if not any(exclude_elem in element for exclude_elem in exclude_list):
@@ -265,10 +265,7 @@ class MayaPlaceholderLoadPlugin(PlaceholderPlugin, PlaceholderLoadMixin):
 
         for camera in cameras:
             if cmds.getAttr(camera + ".renderable"):
-                shapes = cmds.listRelatives(camera,
-                                        parent=True,
-                                        fullPath=True,
-                                        type="transform")
+                shapes = cmds.listRelatives(camera,parent=True,fullPath=True,type="transform")
                 if shapes:
                     renderable_camera = shapes[0]
                     break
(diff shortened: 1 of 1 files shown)


openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -142,7 +139,7 @@ def _create_placeholder_name(self, placeholder_data):
         # add loader arguments if any
         loader_args = placeholder_data["loader_args"]
         if loader_args:
-            loader_args = json.loads(loader_args.replace('\'', '\"'))
+            loader_args = json.loads(loader_args.replace("'", '"'))
```
BigRoy, line 142: This PR seems to implement tons of cosmetic style changes (some if which the Hound disagrees with too). Anyway, the cosmetic changes make it hard to see what actually changed for this PR. A quick read on my end makes me assume that **only** this line really was a change to fix an issue. Is that correct?

At this point in time I'm not sure why we need to even do the replace thing to begin with. Why is that?

FadyFS: Yes it does, I make bad manipulation while using a formatter to correct a specific line. Let me correct that

openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -244,13 +233,36 @@ def get_placeholder_options(self, options=None):
...
         cmds.sets(node, addElement=PLACEHOLDER_SET)
         cmds.hide(node)
         cmds.setAttr(node + ".hiddenInOutliner", True)
+        model = cmds.ls(assemblies=True)
+
+        filtered_model = []
+        exclude_list = ["persp", "top", "front", "side", "root", "wip"]
+
+        for element in model:
+            if not any(exclude_elem in element for exclude_elem in exclude_list):
```
BigRoy, line 248: This will fail to detect the default cameras if they were parented to something since they wouldn't be assemblies.
You might be better of doing the `cmds.camera(cam, query=True, startupCamera=True)` query on the cameras to filter.

Also, instead of filtering the cameras you might be better of just searching for only the visible shapes? `cmds.ls(type=("mesh", "nurbsCurve", "nurbsSurface"), visibleOnly=True)` (even though that doesn't account for shapes that become visible over time?

openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -244,13 +233,36 @@ def get_placeholder_options(self, options=None):
...
+                shapes = cmds.listRelatives(
+                    camera, parent=True, fullPath=True, type="transform"
+                )
+                if shapes:
+                    renderable_camera = shapes[0]
+                    break
+
+        cmds.lookThru(renderable_camera)
+        print("filtered_model: ", filtered_model)
+        cmds.viewFit(filtered_model)
```
BigRoy, line 266: The filter model logic only applies to this - and not to the renderable camera loop above. As such I'd move the logic closer together to make it clear that it's only needed for this and doesn't have anything to do with the renderable camera.

openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -244,13 +233,36 @@ def get_placeholder_options(self, options=None):
...
```
BigRoy, line 265: I'd argue that it doesn't make sense to do this in `post_placeholder_process` if that's called PER placeholder - since we'd be doing this many many times. We should be doing this at most ONCE after the whole build process completed - and better yet, I'd even argue that this shouldn't be in the default template logic - this should be something that can be added to a build process separately instead of forced upon.

openpype/hosts/maya/api/workfile_template_builder.py
```diff
@@ -244,13 +233,36 @@ def get_placeholder_options(self, options=None):
...
+        """Hide placeholder, add them to placeholder set"""
         node = placeholder.scene_identifier
 
         cmds.sets(node, addElement=PLACEHOLDER_SET)
         cmds.hide(node)
         cmds.setAttr(node + ".hiddenInOutliner", True)
+        model = cmds.ls(assemblies=True)
+
+        filtered_model = []
+        exclude_list = ["persp", "top", "front", "side", "root", "wip"]
```
BigRoy, line 245: "root" and "wip" I assume are very likely hardcoded examples from your workflow?

openpype/hosts/maya/api/workfile_template_builder.py
BigRoy: I'd say a direct hard replacement of quotes isn't safe to begin with - but I can see how an artist might think using single quotes is fine even though they should've been writing double quotes for JSON.

However, a JSON string like `"my neighbour's cat" would now break due to this string replacement - breaking esstentially what is valid JSON to begin with.

openpype/hosts/maya/api/workfile_template_builder.py
FadyFS: yes they are

openpype/hosts/maya/api/workfile_template_builder.py
FadyFS: We will add this specifically as action, it won't stay in part of the code. So as you say, it will be called per placeholder as you say.

openpype/hosts/maya/api/workfile_template_builder.py
FadyFS: it's visible=True not visbleOnly, this flag does not exist. Thanks for your advice they work well !

BigRoy: > it's visible=True not visbleOnly, this flag does not exist.

Thanks, must've mixed it up with another flag or command. Anyway, it was just intended to be an idea - not sure if that's the best workflow either.

tokejepsen:
@FadyFS when this PR is ready for review again, just flip the switch and we can progress with it.

tokejepsen:
@FadyFS is this still relevant for you?

mkolar:
@FadyFS we're closing this PR as it is now stale. If there is any functionality from this you'd like to contribute again, feel free to re-open this again, targeting AYON

PR #5284, opened 2023-07-12
Deadline: add OCIO environment to a job

jakubjezek001:
## Changelog Description
OCIO environment variable is now added to job submission in case it is used. 

## Additional info
- also included some clean-up of old env var configuration for OPENPYPE_VERSION

## Testing notes:
1. Enable Global colorspace management on your project with config priority on nuke-default (since those are default configuration at Nuke imageio settings)
2. Open nuke at some task and publish a Render write instance to Deadline.
3. Open Deadline Monitor and look to submitted job properties to Environment section and find the OCIO key

## Dependencies:
including https://github.com/ynput/OpenPype/pull/5302


openpype/modules/deadline/plugins/publish/submit_houdini_render_deadline.py
```diff
@@ -88,7 +88,7 @@ def get_job_info(self):
             "AVALON_APP_NAME",
             "OPENPYPE_DEV",
             "OPENPYPE_LOG_NO_COLORS",
-            "OPENPYPE_VERSION"
```
jakubjezek001, line 91: @antirotor is this ok to be removed. It seemed like some old code to me.

openpype/modules/deadline/plugins/publish/submit_max_deadline.py
```diff
@@ -110,9 +111,14 @@ def get_job_info(self):
             "AVALON_TASK",
             "AVALON_APP_NAME",
             "OPENPYPE_DEV",
-            "OPENPYPE_VERSION",
```
jakubjezek001, line 113: @antirotor is this ok to be removed?  I have added the code which was used in other hosts.

BigRoy:
What's the source issue this is trying to solve?

I feel like this logic currently is invalid due to a few reasons:
- Isn't the launching of an application already setting `OCIO` in the prelaunch hook? 
    - If it is - wouldn't that also override the `OCIO` that deadline tries to set too?
    - If it isn't - shouldn't it? 
- The OCIO env var will be platform specific - we'll need to get the platform-specific path in Deadline, right?

openpype/modules/deadline/plugins/publish/submit_houdini_render_deadline.py
tokejepsen: Yup, should be removed cause it'll be added if running from a build.

openpype/modules/deadline/plugins/publish/submit_max_deadline.py
tokejepsen: Yeah, looks good to remove.

tokejepsen (changes requested):
Could we get the removal of `OPENPYPE_VERSION` in a different PR?

Its polluting the intent of the PR and will be lost if the PR is not merged.

jakubjezek001:
> Its polluting the intent of the PR and will be lost if the PR is not merged.

good point! It had been recreated here https://github.com/ynput/OpenPype/pull/5302

jakubjezek001:
I definitely agree with you @BigRoy. This PR needs more thinking. At the moment the process is broken only if multiple platforms are used within production. Single platform should be working as expected since the current config path is saved into workflows.

### Here are some current implementation concerns:

Maya and Nuke (at the moment only hosts I am monitoring) are included OCIO path content into workflows but they are platform specific. Those will work only in case Deadline is having path remapper configured. 

> * Isn't the launching of an application already setting `OCIO` in the prelaunch hook?

No the prelaunch hook is not started at Deadline. We are only resolving host's environment and it is not launched wrapped with openpype. 

> * The OCIO env var will be platform specific - we'll need to get the platform-specific path in Deadline, right?

Adding the OCIO environment variable to the submitted job environment is not a solution here because it only supports a single path. Therefore, it is still platform-specific and no solution here. 

>   * If it is - wouldn't that also override the `OCIO` that deadline tries to set too?
>   * If it isn't - shouldn't it?

We need to lock the job config dependencies to a specific ocio config path to prevent job failures if someone changes the ocio path during ongoing production.

### Ideas to fix this:
1. Introduce a new variable, OPENPYPE_OCIO, which will be appended with the host submitter to job environments.
	a. during publishing to farm a process will store the de-root-ified configuration path. This variable will only be added if the OCIO path was configured in the running environment host that was used during submission. This variable will be  supporting multipath so in case the config path will not be de-root-ifiable and some other paths will be found in imageio settings under ocio_config, those will be added here.  
	b. Our Deadline plugin sets up environment variables during the process. It takes the OPENPYPE_OCIO and resolves root-ified paths to the appropriate platform root from Anatomy. Alternatively, it checks all working paths and sets the first working path as the active config in OCIO.

2. Set OCIO  with platform specific path to job environment variable and resolve it with OpenPype plugin before host is launched.
	a. The process will be similar to the previous idea, but we will now proceed with the prelaunch processing of the Deadline OpenPype plugin.

#### There are downsides of second idea:
it is too dynamic and paths are resolved form settings during job processing and those can be changed in settings and not correct towards workfile configurations (if used path was not de-root-ifiable, and production is using multiple paths for individual platforms instead of root token)

openpype/modules/deadline/plugins/publish/submit_houdini_render_deadline.py
tokejepsen: This should be removed as its a separate PR now.

openpype/modules/deadline/plugins/publish/submit_max_deadline.py
tokejepsen: This should be removed as its a separate PR now.

tokejepsen (commented):
Not sure why we need to submit `OCIO` environment variable?
It should be baked in to the published workfile anyways?

BigRoy:
> We need to lock the job config dependencies to a specific ocio config path to prevent job failures if someone changes the ocio path during ongoing production.

Actually, that's an interesting point. Since we now have versioned settings I suppose we could provide an env var to lock the settings to a particular point in time or state of settings? That could fix the OCIO issue but potentially a lot of others too?

> No the prelaunch hook is not started at Deadline. We are only resolving host's environment and it is not launched wrapped with openpype.

I'd actually argue this could be an issue. There are some hosts that actually have pre launch hooks that have quite an influence on how the application behaves, e.g. for Fusion it defines how it initializes with settings which could influence e.g. even what plugins are available in Fusion and thus might be required for a correct render.

jakubjezek001:
> Actually, that's an interesting point. Since we now have versioned settings I suppose we could provide an env var to lock the settings to a particular point in time or state of settings? That could fix the OCIO issue but potentially a lot of others too?

That is a good idea and I would love to have the feature available already, but not a problem to be addressed in this PR. We are talking about different feature request. 

> I'd actually argue this could be an issue. There are some hosts that actually have pre launch hooks that have quite an influence on how the application behaves, e.g. for Fusion it defines how it initializes with settings which could influence e.g. even what plugins are available in Fusion and thus might be required for a correct render.

No argument about this but again not within the scope of this PR.

tokejepsen:
@jakubjezek001 could you resovle the conflicts?

jakubjezek001:
closing this PR in preference to https://github.com/ynput/OpenPype/pull/5379

PR #5286, opened 2023-07-13
Max: Optional Renderable Camera Validator for Render Instance

moonyuet:
## Changelog Description
Optional validation to check on renderable camera being set up correctly for deadline submission.
If not being set up correctly, it wont pass the validation and user can perform repair actions. 

## Additional info
1. Render instance won't include the max content validator
2. there would be a change after #5124 when batch renderlayer being introduced.

## Testing notes:
1. Launch 3dsmax via OP
2. Create empty render instance
3. If no renderable camera set, it won't pass the validation
4. Repair it
5. it pops up the menu to allow you to choose your preferred camera and pass the validation


LiborBatek (changes requested):
Did follow the testing steps and got correct validate error
image


and my render settings before ```Repair``` action

image

After it stays the Render Settings same but validates ok and can be Published (I have added the default ```Camera001``` to the scene tho.
image

Seems it checked the camera exists in the scene aka workfile but was not set correctly as rendering camera tho.
Here are render settings afterwards:
image
Still not set as render cam.

LiborBatek:
During testing it again Im getting for some reason this error during Publish:
image

Not sure whats going on...

moonyuet:
> During testing it again Im getting for some reason this error during Publish: image
> 
> Not sure whats going on...

there is a repair action for this to fix the naming of how the renderlayer sets.

moonyuet:
@LiborBatek this is helpful for your case if you have more than one validator errored out.
https://github.com/ynput/OpenPype/assets/64118225/9513d9cb-9d54-4911-84e7-624b51bd6ed9

LiborBatek (approved):
So now when no camera present in the scene the validator correctly fails and if Repair action triggered warning arise if no camera was found.
image

If Camera exists in the scene but isnt set as ```View to Render``` then ```Repair``` action will set it as renderable and ```Publish``` finnishes ok.
image

moonyuet:
@antirotor would you mind to check on the code before the merge? Thanks!

openpype/hosts/max/plugins/publish/validate_viewport_camera.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def process(self, instance):
+        if not self.is_active(instance.data):
+            return
+        cameras_in_scene = [c for c in rt.Objects
+                            if rt.classOf(c) in rt.Camera.Classes]
+        if rt.viewport.getCamera() not in cameras_in_scene:
+            raise PublishValidationError(
+                "Cameras in Scene not used as viewport camera"
+            )
```
BigRoy, line 35: Wording is a bit unclear? Did you mean something like this.

```suggestion
        if rt.viewport.getCamera() not in cameras_in_scene:
            raise PublishValidationError(
                "Viewport camera is not a scene object"
            )
```
And this report should contain information what the artist should do and what it actually means ;)

openpype/hosts/max/plugins/publish/validate_viewport_camera.py
```diff
@@ -0,0 +1,48 @@
...
+
+from pymxs import runtime as rt
+
+
+class ValidateViewportCamera(pyblish.api.InstancePlugin,
+                             OptionalPyblishPluginMixin):
+    """Validates Viewport Camera
+
+    Check if the renderable camera in scene used as viewport
+    camera for rendering
```
BigRoy, line 17: Wording is a bit unclear?

openpype/hosts/max/plugins/publish/validate_viewport_camera.py
```diff
@@ -0,0 +1,48 @@
...
+    OptionalPyblishPluginMixin)
+from openpype.pipeline.publish import RepairAction
+from openpype.hosts.max.api.lib import get_current_renderer
+
+from pymxs import runtime as rt
+
+
+class ValidateViewportCamera(pyblish.api.InstancePlugin,
+                             OptionalPyblishPluginMixin):
+    """Validates Viewport Camera
```
BigRoy, line 14: Why do we need to validate the viewport camera? for what?

BigRoy: It seems very odd that the viewport must be set to a viewport to be able to do a 'batch' render.
I see in other areas of the code that you're able to set e.g. an explicit "Camera" option for Mtoa - which as such seems like it wouldn't need the viewport to specify a camera, yes?

Similarly the 3dsmax Deadline Submitter has an explicit 'Camera' field:
image

Why can't we use that plug-in info in the submission to specify the camera for the batch render? Would mean there's no need to tweak the scene at all. Win-win?

openpype/hosts/max/plugins/publish/validate_viewport_camera.py
moonyuet: I have edited the docstring a little bit

openpype/hosts/max/plugins/publish/collect_render.py
```diff
@@ -34,6 +34,11 @@ def process(self, instance):
         aovs = RenderProducts().get_aovs(instance.name)
         files_by_aov.update(aovs)
 
+        camera = rt.viewport.GetCamera()
+        instance.data["cameras"] = None
```
Minkiu, line 38: This could be:
```py
instance.data["cameras"] = [camera.name] if camera else None
```

openpype/modules/deadline/plugins/publish/submit_max_deadline.py
```diff
@@ -154,7 +155,8 @@ def get_plugin_info(self):
             SceneFile=self.scene_path,
             Version=instance.data["maxversion"],
             SaveFile=True,
-            IgnoreInputs=True
+            IgnoreInputs=True,
+            Camera=instance.data["cameras"][0]
```
Minkiu, line 159: Are we sure that this will be an empty list?

Minkiu (approved):
Haven't tested it but code looks good.

moonyuet:
@LiborBatek Can you test again for double check before merging?

LiborBatek (approved):
Validator works fine
image

And when just single camera present, ```Repair``` will correctly set it as ```Renderable``` if multiple cameras present, pop up window will show up so user can set the one he want to be set as renderable.

All working fine.

PR #5292, opened 2023-07-13
Maya: Removed unnecessary import of pyblish.cli

kalisp:
## Changelog Description
This import resulted in adding additional logging handler which lead to duplication of logs in hosts with plugins containing `is_in_tests` method. Import is unnecessary for testing functionality.


BigRoy (approved):
This worked perfectly for me in 3.15 (however didn't test on latest develop)

PR #5296, opened 2023-07-13
Maya: fix validate frame range on review attached to other instances

mkolar:
## Changelog Description
Fixes situation where frame range validator can't be turned off on models if they are attached to reviewable camera in Maya. 

## Testing notes:
1. Create model publish instance in maya
2. create review publish instance in maya 
3. parent model maya set under the review maya set based on this **https://ayon.ynput.io/docs/artist_hosts_maya#publishing-model-with-review**
4. turn off `validate frame range` on the review in the publisher GUI. 
5. Everything should go through correctly and you should end up with review attached to the model Product.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/collect_review.py
+++ b/openpype/hosts/maya/plugins/publish/collect_review.py
@@ -107,10 +107,8 @@ class CollectReview(pyblish.api.InstancePlugin):
             data["displayLights"] = display_lights
             data["burninDataMembers"] = burninDataMembers
 
-            publish_attributes = data.setdefault("publish_attributes", {})
             for key, value in instance.data["publish_attributes"].items():
-                if key not in publish_attributes:
-                    publish_attributes[key] = value
+                data["publish_attributes"][key] = value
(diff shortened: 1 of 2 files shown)


tokejepsen (changes requested):
I'm getting validation passed with the `Validate Frame Range` enabled as well as disabled.

cap1
cap2

LiborBatek (commented):
Strangely it works no matter if ```Validate Frame Range``` being ```ON``` or ```OFF```  even when asset DB ```frame range```  ```1001-1001``` and review differ by ```1001-1050```

image

tokejepsen:
For testing this we also need to remove the `model` family from `project_settings/maya/publish/ValidateFrameRange/exclude_families`, which is quite confusing given the label of the settings as well.

I've suggested a code fix for this PR [here](https://github.com/ynput/OpenPype/pull/5340)

tokejepsen (approved):
Tested in Maya 2023 successfully. Code looks good to me as well.

PR #5299, opened 2023-07-13
Traypublisher: Fix plugin order

iLLiCiTiT:
## Changelog Description
Frame range collector for traypublisher was moved to traypublisher plugins and changed order to make sure `assetEntity` is filled in `instance.data`.

## Testing notes:
1. Open TrayPublisher
2. Create `render` instance
3. Hit Publish (or Validate)
4. `Collect Frame Data From Asset Entity` should not crash


moonyuet (approved):
image
Looks good and doesn't break anything.

mkolar (approved):

PR #5300, opened 2023-07-13
AYON: 3rd party addon usage

iLLiCiTiT:
## Changelog Description
Prepare OpenPype code to be able use `ayon-third-party` addon which supply ffmpeg and OpenImageIO executables. Because they both can support to define custom arguments (more than one) a new functions were needed to supply.

New functions are `get_ffmpeg_tool_args` and `get_oiio_tool_args`. They work similar to previous but instead of string are returning list of strings. All places using previous functions `get_ffmpeg_tool_path` and `get_oiio_tool_path` are now using new ones. They should be backwards compatible and even with addon if returns single argument.

## Testing notes:
This is tricky one. Since dependency packages are not yet available there are 2 ways how to be able start executable that would be able to use `ayon-third-party` addon:
1. Install/Copy needed dependencies to ayon-launcher so it is possible to launch it with openpype addon.
2. Copy `common` folder from `ayon-launcher` to `OpenPype` repository, so `ayon-third-party` addon has needed requirements.

Prerequirements:
1. Working executable
2. openpype addon on AYON server
3. `ayon-third-party` addon on AYON server

Testing:
1. Start tray
2. 3rd part addon should download ffmpeg and oiio to your machine
3. Launch any host of your preference that is working with ffmpeg or oiio
4. Run publishing which will use them (e.g. review)
5. Validate in output that ffmpeg/oiio from 3rd party addon was used (`...\addons\ayon_third_party_0.0.1\ayon_third_party\downloads\ffmpeg\bin\ffmpeg.exe`)

### DIFF
--- a/openpype/hosts/hiero/plugins/publish/extract_frames.py
+++ b/openpype/hosts/hiero/plugins/publish/extract_frames.py
@@ -2,7 +2,7 @@ import os
 import pyblish.api
 
 from openpype.lib import (
-    get_oiio_tools_path,
+    get_oiio_tools_args,
     run_subprocess,
 )
 from openpype.pipeline import publish
@@ -18,7 +18,7 @@ class ExtractFrames(publish.Extractor):
     movie_extensions = ["mov", "mp4"]
 
     def process(self, instance):
-        oiio_tool_path = get_oiio_tools_path()
+        oiio_tool_args = get_oiio_tools_args()
         staging_dir = self.staging_dir(instance)
         output_template = os.path.join(staging_dir, instance.data["name"])
         sequence = instance.context.data["activeTimeline"]
@@ -36,7 +36,7 @@ class ExtractFrames(publish.Extractor):
             output_path = output_template
             output_path += ".{:04d}.{}".format(int(frame), output_ext)
 
-            args = [oiio_tool_path]
+            args = list(oiio_tool_args)
 
             ext = os.path.splitext(input_path)[1][1:]
             if ext in self.movie_extensions:
--- a/openpype/hosts/maya/plugins/publish/extract_look.py
+++ b/openpype/hosts/maya/plugins/publish/extract_look.py
@@ -16,7 +16,7 @@ import pyblish.api
 from maya import cmds  # noqa
 
 from openpype.lib.vendor_bin_utils import find_executable
-from openpype.lib import source_hash, run_subprocess, get_oiio_tools_path
+from openpype.lib import source_hash, run_subprocess, get_oiio_tools_args
 from openpype.pipeline import legacy_io, publish, KnownPublishError
 from openpype.hosts.maya.api import lib
 
@@ -267,11 +267,10 @@ class MakeTX(TextureProcessor):
 
         """
 
-        maketx_path = get_oiio_tools_path("maketx")
-
-        if not maketx_path:
+        maketx_args = get_oiio_tools_args("maketx")
+        if not maketx_args:
             raise AssertionError(
-                "OIIO 'maketx' tool not found. Result: {}".format(maketx_path)
+                "OIIO 'maketx' tool not found."
             )
 
         # Define .tx filepath in staging if source file is not .tx
@@ -328,8 +327,7 @@ class MakeTX(TextureProcessor):
 
         self.log.info("Generating .tx file for %s .." % source)
 
-        subprocess_args = [
-            maketx_path,
+        subprocess_args = maketx_args + [
             "-v",  # verbose
             "-u",  # update mode
             # --checknan doesn't influence the output file but aborts the
--- a/openpype/hosts/photoshop/plugins/publish/extract_review.py
+++ b/openpype/hosts/photoshop/plugins/publish/extract_review.py
@@ -1,10 +1,9 @@
 import os
-import shutil
 from PIL import Image
 
 from openpype.lib import (
     run_subprocess,
-    get_ffmpeg_tool_path,
+    get_ffmpeg_tool_args,
 )
 from openpype.pipeline import publish
 from openpype.hosts.photoshop import api as photoshop
@@ -85,7 +84,7 @@ class ExtractReview(publish.Extractor):
             instance.data["representations"].append(repre_skeleton)
             processed_img_names = [img_list]
 
-        ffmpeg_path = get_ffmpeg_tool_path("ffmpeg")
+        ffmpeg_path = get_ffmpeg_tool_args("ffmpeg")
 
         instance.data["stagingDir"] = staging_dir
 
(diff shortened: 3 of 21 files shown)


openpype/hosts/hiero/plugins/publish/extract_frames.py
```diff
@@ -36,7 +36,7 @@ def process(self, instance):
             output_path = output_template
             output_path += ".{:04d}.{}".format(int(frame), output_ext)
 
-            args = [oiio_tool_path]
+            args = list(oiio_tool_args)
```
antirotor, line 39: isn't this list anyway, as `get_oiio_tools_args()` always returns list?

openpype/lib/vendor_bin_utils.py
```diff
@@ -321,7 +354,29 @@ def get_oiio_tools_path(tool="oiiotool"):
...
+
+    if AYON_SERVER_ENABLED:
+        args = _get_ayon_oiio_tool_args(tool_name)
+        if args:
+            return args
+
+    path = get_oiio_tools_path(tool_name)
+    if path:
+        return [path]
+    return []
```
antirotor, line 376: I am wondering if we should allow this. All the places that are using it will actually fail if this is empty, so I am not sure if we shouldn't rise exception here or handle it in the place where this is used.

antirotor (commented):
code looks good, have not tested it yet

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -267,11 +267,10 @@ def process(self,
...
         """
 
-        maketx_path = get_oiio_tools_path("maketx")
-
-        if not maketx_path:
+        maketx_args = get_oiio_tools_args("maketx")
+        if not maketx_args:
             raise AssertionError(
-                "OIIO 'maketx' tool not found. Result: {}".format(maketx_path)
+                "OIIO 'maketx' tool not found."
```
Minkiu, line 273: This probably should be raise inside `get_oiio_tools_args` ?

openpype/hosts/photoshop/plugins/publish/extract_review.py
```diff
@@ -85,7 +84,7 @@ def process(self, instance):
             instance.data["representations"].append(repre_skeleton)
             processed_img_names = [img_list]
 
-        ffmpeg_path = get_ffmpeg_tool_path("ffmpeg")
+        ffmpeg_path = get_ffmpeg_tool_args("ffmpeg")
```
Minkiu, line 87: `ffmpeg_path` -> `fmpeg_args`

openpype/hosts/photoshop/plugins/publish/extract_review.py
```diff
@@ -142,8 +149,9 @@ def _generate_mov(self, ffmpeg_path, instance, fps, no_of_frames,
             "tags": self.mov_options['tags']
         })
 
-    def _generate_thumbnail(self, ffmpeg_path, instance, source_files_pattern,
-                            staging_dir):
+    def _generate_thumbnail(
+        self, ffmpeg_args, instance, source_files_pattern, staging_dir
```
Minkiu, line 153: I was wondering if we can skip passing `ffmpeg_args` altogether and do:
```py
ffmpeg_args = get_ffmpeg_tool_args("ffmpeg")
```

here?

Maybe have a `ffmpeg_extra_args=None` in case we want to add more...

openpype/lib/vendor_bin_utils.py
```diff
@@ -270,32 +272,63 @@ def _oiio_executable_validation(filepath):
...
+        filepath = find_executable(args)
+        if not filepath:
+            return False
+        args = [filepath]
+    return _check_args_returncode(args + ["--help"])
+
+
+def _get_ayon_oiio_tool_args(tool_name):
+    try:
+        # Use 'ayon-third-party' addon to get ffmpeg arguments
```
Minkiu, line 295: Type, this should be `oiio` instead of `ffmpeg`

openpype/plugins/publish/extract_review.py
```diff
@@ -788,7 +786,9 @@ def ffmpeg_full_args(
                     audio_filters.append(arg)
 
         all_args = []
-        all_args.append(path_to_subprocess_arg(self.ffmpeg_path))
+        all_args.append(
```
Minkiu, line 789: This could be:
```py
all_args = [
    subprocess.list2cmdline(get_ffmpeg_tool_args("ffmpeg")),
    *input_args
]
```

Minkiu (changes requested):
I find changing to `*_tools_args` and still getting a `path` to an executable plus arguments a bit confusing.

openpype/hosts/hiero/plugins/publish/extract_frames.py
iLLiCiTiT: The `args` list is modified through the loop. It would keep arguments from previous loop on next run.

openpype/plugins/publish/extract_review.py
iLLiCiTiT: I think tuple expanding into list did not work for some DCCs. I'm not saying it is still an issue, but I'm scared to change it.

iLLiCiTiT:
> I find changing to *_tools_args and still getting a path to an executable

You mean the the variable where is stored is path?

> plus arguments a bit confusing.

Not sure I understand this one?

kalisp:
`ayon-thirt-party` addon is missing json manifest, imho

iLLiCiTiT:
> ayon-thirt-party addon is missing json manifest, imho

Urelated to this PR.

mkolar (approved):

mkolar (approved):

mkolar:
Works as expected with AYON-launcher.

Minkiu:
So clarifying my previous comment:
if I have a variable named `foo_tools_args` I expect i to have a string or a list of arguments, not `path_to_executable` + `list_of_arguments`, at some point I would see `foo_executable` and `foo_tools_args` passed to the sub process or similar.

iLLiCiTiT:
> if I have a variable named foo_tools_args I expect i to have a string or a list of arguments, not path_to_executable + list_of_arguments, at some point I would see foo_executable and foo_tools_args passed to the sub process or similar.

But you can define custom launch arguments in settings of ayon third party addon, where, for REZ, it could be `rez env ffmpeg -- ffmpeg` -> `["rez", "env", "ffmpeg", "--", "ffmpeg"]` thus it may not be just a path...

Minkiu:
Yeah no, it was more naming thing than anything, for example at some point we have:
`ffmpeg_args = get_ffmpeg_tool_args("ffmpeg")`

And then somewhere else we have:
`ffmpeg_executable_args = get_ffmpeg_tool_args("ffmpeg")`

I get what you mean and you are correct; the thing is that regardless of `rez` or `/opt/rez/rez` the first argument will be a an executable, and that's where the "confusing" part is; but there's no point in getting bogged down on this, it's looks good and does the job :D

Minkiu (approved):
Code looks good, haven't tested it.

PR #5303, opened 2023-07-14
Maya: Apply project settings to creators

tokejepsen:
## Changelog Description
Project settings were not applied to the creators.

## Testing notes:
1. Make a change from the default settings on any of the Maya creators; `project_settings/maya/create`
2. In Maya create the modified creator.
3. Verify the settings from the project settings are applied.

### DIFF
--- a/openpype/hosts/maya/api/plugin.py
+++ b/openpype/hosts/maya/api/plugin.py
@@ -181,6 +181,8 @@ class MayaCreatorBase(object):
 @six.add_metaclass(ABCMeta)
 class MayaCreator(NewCreator, MayaCreatorBase):
 
+    settings_name = None
+
     def create(self, subset_name, instance_data, pre_create_data):
 
         members = list()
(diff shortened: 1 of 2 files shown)


openpype/hosts/maya/api/plugin.py
```diff
@@ -238,6 +240,24 @@ def get_pre_create_attr_defs(self):
                     default=True)
         ]
 
+    def apply_settings(self, project_settings, system_settings):
```
iLLiCiTiT, line 243: I didn't implement this by default because I wanted force developers to implement explicit settings apply. So any future developer can know what and which settings are applied, when and how, by looking into the file where the creator is, instead of magically applying them. That's my personal opinion how I think it should work, so don't take my word and let @antirotor and @BigRoy decide on this.

tokejepsen: Personally I think it'll lower the barrier of entry for contributions, but let's get some more opinions on this.

antirotor: I agree that it will lower the entry point barrier - for example in Maya case (but others too) it would mean explicitly adding tons of settings. But if we stick to the system, that most of the creator settings are in `project_settings["maya"]["create"]` and whatever else the creator needs will pull it explictitly, I think it is well-enough defined.

tokejepsen: @iLLiCiTiT you happy to continue with this then?

antirotor: From my point of view yes. I would be good to clearly state soemwhere (in docstrings?) how the logic works and where to define the settings, but overall it is responsibility of the creator developer to document it properly inside the plugin itself

iLLiCiTiT: > @iLLiCiTiT you happy to continue with this then?

@antirotor has spoken :) not a blocker for me.

simonebarbieri (approved):
It seems to be working fine!

mkolar (approved):

Issue #4614, opened 2023-03-14
Nuke: extract review slate does not respect ExtractReview pixel aspect scaling


jeangjenq:
**Running version**
3.14.7

**Describe the bug**
[extract_review_slate.py](../blob/develop/openpype/plugins/publish/extract_review_slate.py) follows pixel aspect ratio in instance data. If `scale_pixel_aspect` is turned off in `global/publish/ExtractReview/profile`, slate frame will have different pixel aspect ratio to the rest of the frames.

**To Reproduce**
1. Go to a shot with pixel aspect ratio of 2.
2. Make sure `scale_pixel_aspect` is on in ExtractReview profile.
3. Render and publish.
4. Slate and shot appears consistent.
5. Turn off "Scale pixel aspect" in ExtractReview and publish again.
6. Slate frame now appears different compared to the rest of shot frame range.

**Expected behavior**
Turning off `scale_pixel_aspect` in ExtractReview profile should apply for `extract_review_slate` as well.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Host: Nuke

[cuID:[OP-5225](https://app.clickup.com/t/861mgg2fn)]

jakubjezek001:
already merged by https://github.com/ynput/OpenPype/pull/5304

PR #5304, opened 2023-07-14
Nuke Anamorphic slate - Read pixel aspect from input

jrsndl:
## Changelog Description
When asset pixel aspect differs from rendered pixel aspect, Nuke slate pixel aspect is not longer taken from asset, but is readed via ffprobe.

## Testing notes:
1. Setup shot with 1828x1556, pixel aspect 2.0
2. Publish render from Nuke (with Nuke slate)
3. Change render output to be 1920x1080, pixel aspect 1.0
4. Publish again

With this PR, slate pixel aspect should not be squeezed in step 4

Might solve Issue #4614


openpype/plugins/publish/extract_review_slate.py
```diff
@@ -438,6 +442,16 @@ def _get_video_metadata(self, streams):
             input_width = width
             input_height = height
 
+            input_pixel_aspect = str(stream.get("sample_aspect_ratio"))
+            if input_pixel_aspect is not None:
```
iLLiCiTiT, line 446: It will never be `None` because of the string conversion.

openpype/hosts/nuke/plugins/publish/extract_slate_frame.py
```diff
@@ -242,7 +242,7 @@ def _render_slate_to_sequence(self, instance):
 
         # render slate as sequence frame
         nuke.execute(
-            instance.data["name"],
+            str(instance.data["name"]),
```
iLLiCiTiT, line 245: Is there a reason to convert it to string?

jrsndl: I can't reproduce now, but some combination of project settings failed without the conversion.

iLLiCiTiT: Yes please. If the name is `None` it should crash :)

64qam (approved):

openpype/hosts/nuke/plugins/publish/extract_slate_frame.py
jakubjezek001: can we revert this change to previous version, please

jakubjezek001 (approved):
after reverting of https://github.com/ynput/OpenPype/pull/5304/files#r1279339095 is good to be merged.

PR #5305, opened 2023-07-14
Nuke - Allow ExtractReviewDataMov with no timecode knob

jrsndl:
## Changelog Description
ExtractReviewDataMov allows to specify file type. Trying to write some other extension than mov fails on generate_mov assuming that mov64_write_timecode knob exists.

## Testing notes:
1. Publish render from nuke, using ExtractReviewDataMov publish plugin
2. Set the ExtractReviewDataMov extension to mxf (project_settings/nuke/publish/ExtractReviewDataMov/outputs/baking/extension) and publish again

With this PR, second step should not fail


64qam (approved):
looks good

jakubjezek001:
I was following the tasting notes and it is working. The issue is that not all ffmpeg builds are supporting mxf format so reviewables are not created and published.

Also important to mention that this is not fixing the problem with sequence extension. But let's deal with it in other PR.

Thank you @jrsndl !

jakubjezek001 (approved):
LGTM!

PR #5308, opened 2023-07-17
AYON: Fix args for workfile conversion util

iLLiCiTiT:
## Changelog Description
Workfile update conversion util function have right expected arguments.


kalisp (approved):
Helps with the problem when saving `Artist note` to Workfile.

It's a bit weird to have signature which arguments are not used inside of function.

PR #5309, opened 2023-07-17
Bug: various fixes for launch issues

antirotor:
## Changelog Description
Fixing various little issues preventing AYON desktop from launching.

## Additional info
There is inconsistency between values in `*_dir` variables - it should contain only directories, not full file paths. There is default value for `hash_type` set to sha256, but `hash_type` isn't passed and checksum is actually md5.

> **Note**
> I believe there are more places where `*_dir` variables contain wrong data.


iLLiCiTiT:
Issues will be resolved with https://github.com/ynput/OpenPype/pull/5348

PR #5310, opened 2023-07-17
Bug: Error on multiple instance rig with maya

ClementHector:
In case of multiple subset rig in the maya scene, there is an error on the plugin `rig output ids`

## Changelog Description

I change endswith method by startswith method because the set are automacaly name out_SET, out_SET1, out_SET2 ...
image

## Testing notes:
1. Open maya
2. Add multiple rig instance with different subset
3. Publish and see the error

### DIFF
--- a/openpype/hosts/maya/plugins/create/create_rig.py
+++ b/openpype/hosts/maya/plugins/create/create_rig.py
@@ -20,6 +20,6 @@ class CreateRig(plugin.MayaCreator):
         instance_node = instance.get("instance_node")
 
         self.log.info("Creating Rig instance set up ...")
-        controls = cmds.sets(name="controls_SET", empty=True)
-        pointcache = cmds.sets(name="out_SET", empty=True)
+        controls = cmds.sets(name=subset_name + "_controls_SET", empty=True)
+        pointcache = cmds.sets(name=subset_name + "_out_SET", empty=True)
         cmds.sets([controls, pointcache], forceElement=instance_node)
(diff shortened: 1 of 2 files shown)


tokejepsen:
I dont think this is the right way of fixing this. I've proposed some suggestions to fix this instead; https://github.com/quadproduction/OpenPype/pull/539

ClementHector:
> Review [quadproduction#539](https://github.com/quadproduction/OpenPype/pull/539)

Yes, it looks better, thanks.

tokejepsen (approved):
Tested in Maya 2023 successfully. Code looks good.

BigRoy:
Did you also test the publishing of the animation instances that are generated when loading the rigs that have the out_SET2 named nodes, etc?

tokejepsen:
> Did you also test the publishing and loading of the animation instances that are generated when loading the rigs that have the out_SET2 named nodes, etc?

Yup, all seems to be working. Technically there wont be any `out_SET2` but rather `{subset}_out_SET`.

Capture
cap2

BigRoy:
It seems this PR has broken rig publishing for me since it doesn't pass [this validator](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/maya/plugins/publish/validate_rig_contents.py).

It seems that the Validator is disabled by default in settings so it has likely not been tested.

antirotor:
> It seems this PR has broken rig publishing for me since it doesn't pass [this validator](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/maya/plugins/publish/validate_rig_contents.py?rgh-link-date=2023-09-07T11%3A14%3A31Z).
> 
> It seems that the Validator is disabled by default in settings so it has likely not been tested.

Also it seems it broke asset switching...

Issue #5313, opened 2023-07-17
Bug: Multiselection combobox does not show check state with PySide6 on macos


movalex:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

The multiselection combo shows empty boxes for active elements in the list.
It does not update check marks on mouse click.
It also crashes OpenPype on keyboard space press.

### Expected Behavior:

On a mouce click the checkmark should appear in the box. 
Pnce the combobox is opened, all active elements should be checkmarked

### Version

3.16.0

### What platform you are running OpenPype on?

MacOS

### Steps To Reproduce:

1. Open OP Admin settings on macos
2. Open anatomy-attributes-applications list and click on empty box, no check mark will appear
3. Select the box with a keyboard and press space, the OP will crash.

![image](https://github.com/ynput/OpenPype/assets/11698866/0d01bc82-941e-44ec-9600-7f530b740ec3)

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

[cuID:[OP-6377](https://app.clickup.com/t/861n2r90t)]

mustafa-zarkash:
I can notice that the mentioned PR is merged, Is this issue still open ?

movalex:
I will close it, thanks!

PR #5314, opened 2023-07-17
Fix/pyside6 multiselection combobox check state

movalex:
## Changelog Description
* The check states are replaced with the values for PySide6
* `ItemIsUserTristate` is used instead of `ItemIsTristate` to avoid crashes on PySide6

## Additional info
`Qt.Checked`/ and `Qt.Unchecked` are converted to values for PySide6, as suggested here:
https://forum.qt.io/topic/138298/qt-checkstaterole-not-work-in-pyside-6-3-1/6?utm_source=pocket_saves

## Related issue:
https://github.com/ynput/OpenPype/issues/5313

## Testing notes:

1. Open OP Admin settings on macos
2. Open anatomy-attributes-applications list. All active elements will be checked
3. Click on empty box, check mark will appear
4. Select the box with a keyboard and press space, check mark will appear.

photo_2023-07-17_23-10-412

### DIFF
--- a/openpype/tools/settings/settings/multiselection_combobox.py
+++ b/openpype/tools/settings/settings/multiselection_combobox.py
@@ -132,12 +132,10 @@ class MultiSelectionComboBox(QtWidgets.QComboBox):
 
         elif event.type() == QtCore.QEvent.KeyPress:
             # TODO: handle QtCore.Qt.Key_Enter, Key_Return?
-            if event.key() == QtCore.Qt.Key_Space:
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
Thank you, but I had to make few changes regarding the logic changes. We also have some functionality dealing with checkboxes in PySide6 vs. PySide2 in utils, so I've just reused them.

@movalex please verify it works as expected with these new changes.

antirotor (approved):
Tested on Windows (PySide2) and it works as expected.

Issue #4836, opened 2023-04-12
Houdini: OCIO support


antirotor:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

We need to enhance OCIO support of AYON inside Houdini. With #4700 OCIO is correctly passed to Houdini, but setting OCIO profiles (and other options) on relevant nodes needs to be done manually. Like these options in OpenGL ROP (introduced by #3839):

![image](https://user-images.githubusercontent.com/33513211/231497164-59721e27-48b9-489f-b41f-36da3a08ee4d.png)

This is somehow similar to #4833.

There might be other places to set.

### How would you imagine the implementation of the feature?

It should be done as it is already done in Nuke, where you can set color spaces like this:

![image](https://user-images.githubusercontent.com/33513211/231499570-e81aef2f-003a-4b59-b3c4-ed3c7c514912.png)

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

[cuID:[OP-5602](https://app.clickup.com/t/861mn7n0c)]

BigRoy:
What would be a decent default colorspace to set for the review output files. What setting should we obey to?
Would it the preferred rendering space? or the display space? (Houdini doesn't have any settings for that however).

mustafa-zarkash:
I was reviewing this issue [#4833](https://github.com/ynput/OpenPype/issues/4833) 
and I believe it depends on this issue. 

As far as I know, Houdini requires setting OCIO profiles (and other options) manually as Houdini nodes don't inherit color management settings like in Maya. 

I can think of two solutions 
1. Set color space dynamically so that a plugin will deduce what color space to use from default display and default view
2. Expose Parameters to Ayon/OpenPype Settings UI so that aplugin will use these exposed parameters

For the mentioned example `OpenGL` 
I used the same OCIO Color correction as `hou.Color.ocio_defaultView()` which is alsmost the same that happens with Maya playblast (_to use view transform_)

![image](https://github.com/ynput/OpenPype/assets/20871534/d44eca3e-cc57-4822-ad59-67dd38819844)

There's also another problem in houdini that not all nodes have explicit colorspace parameters, and they will use `scene_linear` implicitly (e.g. mantra, texture read). 
so, I think there will be some tweaking in all ROPs. 

what do you think ? 
@BigRoy @antirotor @moonyuet

antirotor:
[...]

Issue #4833, opened 2023-04-12
Houdini: Unmanaged colorspaces in review ROP


antirotor:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Review OpenGL ROP node introduced by #3839 is not handling colorspaces correctly in regards to selected output format. If set to for example PNG, gamma correction will stay by default as 1, resulting in dark and incorrect review video.

### How would you imagine the implementation of the feature?

We should probably define mapping between gamma and file formats if possible, something like:

```python
{
	"png": 2.2,
    "exr": 1.0,
    "jpg": 2.2
}
```

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

### Describe alternatives you've considered:

This is probably solved by using "proper" color management like OCIO as implemented here #4700, but in case studio isn't using OCIO at all, this might be issue. Maybe validator with repair action should handle this.

[cuID:[OP-5600](https://app.clickup.com/t/861mn6q1x)]

MustafaJafar:
This [PR](https://github.com/ynput/OpenPype/pull/5322) was made to solve this issue. 
as this PR was merged I believe this issue should be close. 
if the mentioned PR didn't solve the issue, feel free to reopen it.

PR #5322, opened 2023-07-18
Houdini: validate colorspace in review rop

MustafaJafar:
## Changelog Description
Adding a validator that checks if 'OCIO Colorspace' parameter on  review rop was set to a valid value.
It is a step towards managing colorspace in review rop
valid values are the ones in the dropdown menu
image

this validator also provides some helper actions 
image

This PR is related to #4836 and #4833 

## Additional info 1
This validator doesn't force a certain output color space, it only checks if it's valid or not. 
an artist may select a wrong selection from the dropdown menu and it will still be valid. 

## Additional info 2
this validation will be skipped if Houdini's colospace was set to default, _i.e. no OCIO_
like this 
image
image

This was mentioned in issue #4833 
and I think the validator introduced by this PR doesn't handle that case 

## Testing notes:
1. create a review subset instance 
2. publish


antirotor:
What puzzles me is this combination - Houdini without `OCIO` set:
```python
>>> print(os.getenv("OCIO"))
None
>>> print(hou.Color.ocio_defaultDisplay())
Display - sRGB
>>> print(hou.Color.ocio_configPath())
C:/PROGRA~1/SIDEEF~1/HOUDIN~1.435/houdini/ocio/config-aces-cg.ocio
```
That means that Houdini will default to its own ocio (without setting OCIO)....

mustafa-zarkash:
@antirotor 
So, I found out few things 
1. Houdini default ocio in H19.0 is different than H19.5 
2. I though that `PyOpenColorIO` was part of houdini however it was located in `RenderManProServer-24.4\lib`
> I believe using `PyOpenColorIO` makes it easier to get the default view space name 
3. the one came with my renderman version is old `1.1.1` 
```python
default_view_space = config.getDisplayColorSpaceName(display, view) #1.1.1 
default_view_space = config.getDisplayViewColorSpaceName(display, view) #2.2.1
```
4. this validator should be skipped if houdini set to default 
this conditions seems to work in H19.0 and H19.5 
```python
if os.getenv("OCIO") is None:
```

mustafa-zarkash:
For some reason I can't import `PyOpenColorIO`  although I found it in `project.toml` 
Maybe, I'm missing something.

mustafa-zarkash:
@antirotor  
yeah, Hound is mad at me. 
anyways I moved the function related to `PyOpenColorIO` to `openpype/pipeline/colorspace.py`

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,94 @@
...
+        display =  data.get("display")
+        view =  data.get("view")
+
+        cls.log.debug("Get default view colorspace name..")
+
+        default_view_space = get_display_view_colorspace_name(config_path,
+                                                              display, view)
+
+        rop_node.setParms({"ociocolorspace": default_view_space})
+        cls.log.debug(
```
jakubjezek001, line 90: I would kindly ask you to change the code to newer python syntax (but just to be saved not the f-string). I realize this might be a matter of preference, but this is an old python 2 syntax. This applies to all logs. Thank you.

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,94 @@
...
+        rop_node = hou.node(instance.data["instance_node"])
+
+        data =  get_color_management_preferences()
+        config_path = data.get("config")
+        display =  data.get("display")
+        view =  data.get("view")
+
+        cls.log.debug("Get default view colorspace name..")
+
+        default_view_space = get_display_view_colorspace_name(config_path,
```
jakubjezek001, line 86: I apologize for my lack of Houdini expertise, but I have a question that may seem basic. 

Is it always necessary for the ROP node to have its colorspace set to display and view? Is there no way to assign display and view names directly to the node instead of setting its parent colorspace? I ask because certain views may be associated with a specific look rather than a particular output colorspace.

mustafa-zarkash: In houdini review rop I have these options, and I thought that validating `OCIO Colorspace` leads to better color management. 
image
in this zip there's an exr that corresponds to each setting above 
[houdini_review_colorspaces.zip](https://github.com/ynput/OpenPype/files/12148508/houdini_review_colorspaces.zip)

`OCIO Colorspace` only accepts colorspace names 
image

the validator itself doesn't force any values, it only checks if  `OCIO Colorspace` was set to a valid value (i.e. "sRGB" is not valid because it is not a colorspace name unlike "Output - sRGB")

and there's an action button that suggestd to use the default color space of the default view and display if user is not sure which one to use. 
image

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
mustafa-zarkash: I agree with you.  
but, I'm a little confuse.  Is `%s` that handled by log is legacy or it's something else ?

```python
cls.log.debug( 
    "'OCIO Colorspace' parm on '%s' has been set to "
    "the default view color space '%s'",
    default_view_space, rop_node
)
```

BigRoy: As far as I know the formatting used here is actually still the recommended formatting with loggers due to it doing the formatting only when the log used used by a handler and thus is much faster for cases where no logging handler is handling the logging level. In our case publish loggers are always handled and thus the optimization does not exist as such. Just wanted to remark that this way being done here as far as I know is still the recommended way for loggers in Python. If there's a more recent replacement for that I'd love to learn about it.

moonyuet:
The validator would yell when there is no OCIO colorspace setting up.
image
Successfully publish the review after setting up the config.
image
tried with more complicated OCIO colorspace just to ensure things not broken. It is successfully published(and it is with different colorspaces for sure)
image

I am not sure about the status of this ticket. Is it still work in progress?

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,94 @@
...
+        if invalid:
+            raise PublishValidationError(
+                ("'OCIO Colorspace' parameter is not valid."),
+                title=self.label
+            )
+
+    @classmethod
+    def get_invalid(cls, instance):
+        import hou  # noqa
+        import os
```
iLLiCiTiT, line 36: Why are these imports here and not at the top?

openpype/pipeline/colorspace.py
```diff
@@ -589,3 +589,61 @@ def _get_imageio_settings(project_settings, host_name):
...
+        view color space name (str) e.g. "Output - sRGB"
+    """
+
+    if not compatibility_check():
+        # python environment is not compatible with PyOpenColorIO
+        # needs to be run in subprocess
+        return get_display_view_colorspace_subprocess(config_path,
+                                                      display, view)
+
+    from openpype.scripts.ocio_wrapper import _get_display_view_colorspace_name #noqa
```
iLLiCiTiT, line 611: Shouldn't the implementation of `_get_display_view_colorspace_name` be here and import it inside the scripts instead? Like the scripts are not meant to be source of functions.

openpype/pipeline/colorspace.py
```diff
@@ -589,3 +589,61 @@ def _get_imageio_settings(project_settings, host_name):
...
+
+    return _get_display_view_colorspace_name(config_path, display, view)
+
+def get_display_view_colorspace_subprocess(config_path, display, view):
```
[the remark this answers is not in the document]
iLLiCiTiT:
```suggestion

def get_display_view_colorspace_subprocess(config_path, display, view):
```

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,94 @@
...
+
+        import hou
+        from openpype.pipeline.colorspace import get_display_view_colorspace_name #noqa
+        from openpype.hosts.houdini.api.lib import get_color_management_preferences #noqa
+
+        rop_node = hou.node(instance.data["instance_node"])
+
+        data =  get_color_management_preferences()
+        config_path = data.get("config")
+        display =  data.get("display")
```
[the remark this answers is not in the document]
iLLiCiTiT:
```suggestion
        display = data.get("display")
```

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,94 @@
...
+        used to set colorspace on opengl node to the default view.
+        """
+
+        import hou
+        from openpype.pipeline.colorspace import get_display_view_colorspace_name #noqa
+        from openpype.hosts.houdini.api.lib import get_color_management_preferences #noqa
+
+        rop_node = hou.node(instance.data["instance_node"])
+
+        data =  get_color_management_preferences()
```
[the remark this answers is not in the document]
iLLiCiTiT:
```suggestion
        data = get_color_management_preferences()
```

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,94 @@
...
+        config_path = data.get("config")
+        display =  data.get("display")
+        view =  data.get("view")
```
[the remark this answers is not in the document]
iLLiCiTiT:
```suggestion
        view = data.get("view")
```

openpype/scripts/ocio_wrapper.py
```diff
@@ -173,6 +173,79 @@ def _get_views_data(config_path):
...
+              help="display name",
+              type=click.STRING)
+@click.option("--view", required=True,
+              help="view name",
+              type=click.STRING)
+def get_display_view_colorspace_name(in_path, out_path,
+                                     display, view):
+    """Aggregate view colorspace name to file.
+
+    Python 2 wrapped console command
```
iLLiCiTiT, line 226:
```suggestion
    Wrapper command for processes without acces to OpenColorIO
```

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
mustafa-zarkash: because I found import done in this way in many plugins 
my bad, I should have asked about the better way.

openpype/pipeline/colorspace.py
mustafa-zarkash: I thought this was the convention because it's done the same way in another two functions.

https://github.com/ynput/OpenPype/blob/0c3fa2b61079e0497bccc9ae88e381f681afe39e/openpype/pipeline/colorspace.py#L272

https://github.com/ynput/OpenPype/blob/0c3fa2b61079e0497bccc9ae88e381f681afe39e/openpype/pipeline/colorspace.py#L308

if the implementation is moved out to `colospace.py` 
then should I do this in `ocio_wrapper`

```python
def  get_display_view_colorspace_name (in_path, out_path, display, view):

    json_path = Path(out_path)
    from openpype.pipeline.colorspace import _get_display_view_colorspace_name

    out_data = _get_display_view_colorspace_name(in_path, display, view)

    with open(json_path, "w") as f:
        json.dump(out_data, f)

    print(f"Viewer data are saved to '{json_path}'")
```

mustafa-zarkash:
#### Validator
yes, the validator will only error if there's no valid value in `OCIO Colospace` 
a valid value is any value of the drop down menu. 
I made it because an artist might be lazy and write manually a wrong value like `srgb`  
this validator will only work when there's `OCIO` configured 
> BTW, validator doesn't force any particular values

image

---
#### The helper action
it sets  `OCIO Colospace`  to the default output colorspace  
it retrieves this value from `colorspace.py` using this method `get_display_view_colorspace_name` 

> if the goal of review to be viewed with any video player then default output will be sufficient
> for some reason if they want it in other colorspaces (e.g. `aces cg`) then it will be doable! they will just select it from the drop down menu! 

image

---
#### Review creation
default view color space will be set by default to  `OCIO Colospace`  when creating new review rop node

--- 
#### Future Compatibility
I believe this validator will be very handy for use cases like this one 
this is a prototype of setting default views and  display for Houdini 

This code logic is the same as my implementation in `ocio_wrapper.py` 

image

--- 
#### Conclusion 
@moonyuet  @jakubjezek001 @BigRoy 
> I am not sure about the status of this ticket. Is it still work in progress?

This PR is not a work in progress, the code itself is working as expected 
however, I'm not 100% sure that my proposed solution is what we need

from my search about "color management in Houdini" 
I found out that Houdini is linear _i.e it works in what ever is set to `scene_linear` in the ocio config file_ 
I think default settings of review opengl node will not transform aces to the default output 
So, it will look darker in a video player

these are different default settings 
image
image

image

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
fabiaserra: I always use % formatting for my log strings for this reason

mustafa-zarkash: I think we should make a community topic about this. 
tbh, agreeing on the default formatting will make our lives a little easier. 
here's the related [discussion on discord](https://discordapp.com/channels/517362899170230292/517382145552154634/1139485210833080441)

openpype/hosts/houdini/plugins/create/create_review.py
```diff
@@ -123,3 +129,46 @@ def get_pre_create_attr_defs(self):
...
+            .format(instance_node, default_view_space)
+        )
+
+        return default_view_space
+
+    def get_default_view_space(self):
+        """Get default view space for ociocolorspace parm."""
+
+        from openpype.pipeline.colorspace import get_display_view_colorspace_name  # noqa
+        from openpype.hosts.houdini.api.lib import get_color_management_preferences  # noqa
```
BigRoy, line 164: These should be top-level imports, right?

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,95 @@
...
+    @classmethod
+    def repair(cls, instance):
+        """Set Default View Space Action.
+
+        It is a helper action more than a repair action,
+        used to set colorspace on opengl node to the default view.
+        """
+
+        from openpype.pipeline.colorspace import get_display_view_colorspace_name  # noqa
+        from openpype.hosts.houdini.api.lib import get_color_management_preferences  # noqa
```
BigRoy, line 75: These should be top-level imports, right?

openpype/pipeline/colorspace.py
```diff
@@ -601,3 +601,63 @@ def _get_imageio_settings(project_settings, host_name):
     imageio_host = project_settings.get(host_name, {}).get("imageio", {})
 
     return imageio_global, imageio_host
+
+
+def get_display_view_colorspace_name(config_path, display, view):
+    """get view colorspace name for the given display and view.
```
BigRoy, line 607:
```suggestion
    """Return colorspace for the given display and view.
```

The wording _view colorspace_ somehow had me confused thinking that it wasn't a regular OCIO colorspace definition but something view-specific - which it isn't - so I'd recommend simplifying it like this.

openpype/scripts/ocio_wrapper.py
```diff
@@ -174,5 +174,80 @@ def _get_views_data(config_path):
...
+
+    json_path = Path(out_path)
+
+    out_data = _get_display_view_colorspace_name(in_path,
+                                                 display, view)
+
+    with open(json_path, "w") as f:
+        json.dump(out_data, f)
+
+    print(f"Viewer data are saved to '{json_path}'")
```
BigRoy, line 250: The printed **Viewer data** I think is copied from another function. I assume this should print something else, like "Display view colorspace saved to '{json_path}'".

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
mustafa-zarkash: in `validate_review_colorspace` they are only required to import if user clicked repair action, 
do you recommend moving them to the top ?

BigRoy: I think PEP would recommend that but I've seen the local imports a few times in OpenPype before so I'll leave it up to others to argue. Just wasn't sure whether there was any reason to turn them into local imports, that's all.

openpype/pipeline/colorspace.py
```diff
@@ -601,3 +601,63 @@ def _get_imageio_settings(project_settings, host_name):
...
+
+    Returns:
+        view color space name (str) e.g. "Output - sRGB"
+    """
+
+    if not compatibility_check():
+        # python environment is not compatible with PyOpenColorIO
+        # needs to be run in subprocess
+        return get_display_view_colorspace_subprocess(config_path,
+                                                      display, view)
```
fabiaserra, line 622: is this the standard style indentation that you guys prefer at ynput?

openpype/pipeline/colorspace.py
```diff
@@ -601,3 +601,63 @@ def _get_imageio_settings(project_settings, host_name):
...
+        ]
+        log.info("Executing: {}".format(" ".join(args)))
+
+        process_kwargs = {
+            "logger": log
+        }
+
+        run_openpype_process(*args, **process_kwargs)
+
+        # return all colorspaces
```
fabiaserra, line 661: all colorspaces or just display and view?

openpype/pipeline/colorspace.py
BigRoy: Seeing this from the sidelines I feel like the answer is "not per se". I'm seeing a lot of different contributions with a slight different style and usage of the indentation. I think at this point it's more like: If the hound doesn't complain it's fine.

```python
# A
        return get_display_view_colorspace_subprocess(config_path,
                                                      display, view)
# B
        return get_display_view_colorspace_subprocess(config_path,
                                                      display, 
                                                      view)
# C
        return get_display_view_colorspace_subprocess(
            config_path,
            display, 
            view
        )
```
Not sure if there's any preference on this really 'team-wide'.

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,95 @@
...
+
+    @classmethod
+    def get_invalid(cls, instance):
+
+        rop_node = hou.node(instance.data["instance_node"])
+        if os.getenv("OCIO") is None:
+            cls.log.warning(
+                "Default Houdini colorspace is used, "
+                " skipping check.."
+            )
```
BigRoy, line 44: Not sure if this should be a warning. It's totally fine for a production to be not color managed and if it is, it's not really a point to warn the artists about - that's more of an "admin problem" maybe? Debug log might be better suited so it doesn't show in artist facing reports.

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,95 @@
...
+        from openpype.hosts.houdini.api.lib import get_color_management_preferences  # noqa
+
+        rop_node = hou.node(instance.data["instance_node"])
+
+        data = get_color_management_preferences()
+        config_path = data.get("config")
+        display = data.get("display")
+        view = data.get("view")
+
+        cls.log.debug("Get default view colorspace name..")
```
BigRoy, line 84: Maybe a bit too versbose debug logging. Should we remove this? It's not very informational?
```suggestion
```

openpype/hosts/houdini/plugins/create/create_review.py
```diff
@@ -123,3 +129,46 @@ def get_pre_create_attr_defs(self):
...
+
+        data = get_color_management_preferences()
+        config_path = data.get("config")
+        display = data.get("display")
+        view = data.get("view")
+
+        default_view_space = get_display_view_colorspace_name(config_path,
+                                                              display, view)
```
BigRoy, line 172: I wonder whether it makes more sense to use whatever is configured in settings as view instead of what Houdini is currently using a 'view' transform. From an artist perspective I'd expect to at least get an output matching whatever my Houdini view looks like. Maybe we should be focused more at the houdini view itself being configured the way we want and to have the review instance just use whatever the houdini view is doing? Or am I using some odd reversed logic?

openpype/pipeline/colorspace.py
mustafa-zarkash: I like `B`! 

> If the hound doesn't complain it's fine.

Yeah, That's me :")

openpype/hosts/houdini/plugins/create/create_review.py
mustafa-zarkash: as far as I know, Houdini uses `scene_linear` role as render space
which is not correct when exporting reviews, as mentioned these [screenshots](https://github.com/ynput/OpenPype/pull/5322#issuecomment-1677259981)

Houdini needs three pieces of information to properly publish review in the correct output transform: 
1. View  (returned by houdini py api)
2. display (returned by houdini py api)
3. output transform (not supported, but it's displayed when clicking dropdown menu [screenshot](https://github.com/ynput/OpenPype/pull/5322#issuecomment-1672044600 ) )

That's why I added `get_display_view_colorspace_name` function  which returns the default output transform of the default view and default display

if everything kept in defaults, review output will be in `srgb`

This function call only populates review output transform with the default value 
```
 default_view_space = get_display_view_colorspace_name(config_path,
                                                              display, view)
```

openpype/pipeline/colorspace.py
mustafa-zarkash: my bad, This line was copied without editing.

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,95 @@
...
+
+    @classmethod
+    def repair(cls, instance):
+        """Set Default View Space Action.
+
+        It is a helper action more than a repair action,
+        used to set colorspace on opengl node to the default view.
+        """
+
+        from openpype.pipeline.colorspace import get_display_view_colorspace_name  # noqa
```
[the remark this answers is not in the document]
mustafa-zarkash: Cool Idea!

openpype/hosts/houdini/api/colorspace.py
```diff
@@ -54,3 +54,19 @@ def get_colorspace_data(self):
...
+
+    data = get_color_management_preferences()
+    config_path = data.get("config")
+    display = data.get("display")
+    view = data.get("view")
+
+    default_view_space = get_display_view_colorspace_name(config_path,
+                                                          display,
+                                                          view)
+    return default_view_space
```
BigRoy, line 72: I'd personally prefer it like this - but it's cosmetics mostly.
```suggestion
    prefs = get_color_management_preferences()
    return get_display_view_colorspace_name(
        config_path=prefs["config"],
        display=prefs["display"],
        view=prefs["view"]
    )
```

Note that at the very least you don't need `.get()` calls because the function always returns those three keys in the dict.

openpype/pipeline/colorspace.py
```diff
@@ -601,3 +601,63 @@ def _get_imageio_settings(project_settings, host_name):
...
+        args = [
+            "run", get_ocio_config_script_path(),
+            "config", "get_display_view_colorspace_name",
+            "--in_path", config_path,
+            "--out_path", tmp_json_path,
+            "--display", display,
+            "--view", view
+
+        ]
+        log.info("Executing: {}".format(" ".join(args)))
```
BigRoy, line 653: This should be debug log.
```suggestion
        log.debug("Executing: {}".format(" ".join(args)))
```

openpype/hosts/houdini/api/colorspace.py
fabiaserra: same here and that's also the preference of `black` formatter

openpype/pipeline/colorspace.py
fabiaserra: the alignment on `A` and `B` only work if everyone uses the same indentation configuration (spaces/tabs and same size of indent) in their IDEs, that's why `C` is generally the preferred method

openpype/hosts/houdini/api/colorspace.py
```diff
@@ -54,3 +54,19 @@ def get_colorspace_data(self):
             )
         ]
         return colorspace_data
+
+
+def get_default_display_view_colorspace():
+    """Get default display view colorspace.
+
+    It's used for 'ociocolorspace' parm in OpneGL Node."""
```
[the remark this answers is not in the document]
mustafa-zarkash: sorry I can't tell the difference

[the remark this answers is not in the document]
BigRoy: `OpneGL` -> `OpenGL`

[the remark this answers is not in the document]
mustafa-zarkash: My fingers always swap letters :"D

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
BigRoy, line 1: Just wondering - should this validator be optional?

BigRoy (approved):
Didn't test it - but the code looks good.

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
mustafa-zarkash: I thought I already made it optional but apparently I forgot!

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,92 @@
...
+
+    order = pyblish.api.ValidatorOrder + 0.1
+    families = ["review"]
+    hosts = ["houdini"]
+    label = "Validate Review Colorspace"
+    actions = [SetDefaultViewSpaceAction, SelectROPAction]
+
+    optional = True
+
+    def process(self, instance):
```
BigRoy, line 35: I'm pretty sure this is missing:
```suggestion
    def process(self, instance):
        if not self.is_active(instance.data):
            return
```

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,92 @@
...
+from openpype.pipeline.publish import RepairAction
+from openpype.hosts.houdini.api.action import SelectROPAction
+from openpype.hosts.houdini.api.colorspace import get_default_display_view_colorspace  # noqa
+
+import os
+import hou
+
+
+class SetDefaultViewSpaceAction(RepairAction):
+    label = "Set default view space"
```
BigRoy, line 16: I wonder whether this would be clearer for the artist if it mentions **colorspace** instead of **space** since it might get confusing with world-space, object-space, 3D-space / 2D-space, or a particular space the view/camera is in or whatever.

```suggestion
    label = "Set default view colorspace"
```

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,92 @@
...
+
+    optional = True
+
+    def process(self, instance):
+        invalid = self.get_invalid(instance)
+        if invalid:
+            raise PublishValidationError(
+                ("'OCIO Colorspace' parameter is not valid."),
+                title=self.label
+            )
```
BigRoy, line 41: This publish validation report is lacking a description that explains to the artist what is wrong, what it should be and how they can fix it.

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
mustafa-zarkash: Thanks for telling me! 
I thought IT was automated.

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
mustafa-zarkash: There are more description and directions in the `log.error` on line 71 and 72 

image

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,96 @@
...
+            cls.log.debug(
+                "Default Houdini colorspace is used, "
+                " skipping check.."
+            )
+            return
+
+        if rop_node.evalParm("colorcorrect") != 2:
+            # any colorspace settings other than default requires
+            # 'Color Correct' parm to be set to 'OpenColorIO'
+            rop_node.setParms({"colorcorrect": 2})
```
BigRoy, line 61: Noticed this just now - this should _not_ be in `get_invalid`  since it'd basically be changing nodes in your scene upon just validation. This should be in `repair()`

moonyuet (approved):
If I dont set anything, the validator will error out and I can set default colorspace after the validator fixes
image
It doesn't error out the validator when I set different colorspace.
image

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
BigRoy: Reading the additional log it's still not entirely clear why it's not 'valid'. What is the current value? What value were we expecting? What is wrong with the value - why is it invalid?

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,108 @@
...
+                "Invalid value: Colorspace name doesn't exist.\n"
+                "Check 'OCIO Colorspace' parameter on '{}' ROP"
+                .format(rop_node.path())
+            )
+            return rop_node, error
+
+    @classmethod
+    def get_invalid(cls, instance):
+        nodes, _ = cls.get_invalid_with_message(instance)
+        return nodes
```
BigRoy, line 80: This method is unused since you have no `SelectInvalidAction` and don't use it in `repair` either.
```suggestion
```

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,108 @@
...
+
+        if rop_node.evalParm("ociocolorspace") not in \
+                hou.Color.ocio_spaces():
+
+            error = (
+                "Invalid value: Colorspace name doesn't exist.\n"
+                "Check 'OCIO Colorspace' parameter on '{}' ROP"
+                .format(rop_node.path())
+            )
+            return rop_node, error
```
BigRoy, line 75: Since you're not using `SelectInvalidAction` nor any of the other things that need the rop_node to actually be returned I feel like we can just change this to:

```suggestion
        if os.getenv("OCIO") is None:
            self.log.debug(
                "Default Houdini colorspace is used, skipping check.."
            )
            return

        rop_node = hou.node(instance.data["instance_node"])
        if rop_node.evalParm("colorcorrect") != 2:
            # any colorspace settings other than default requires
            # 'Color Correct' parm to be set to 'OpenColorIO'
            raise PublishValidationError(
                "'Color Correction' parm on '{}' ROP must be set to"
                " 'OpenColorIO'".format(rop_node.path())
            )
        
        if rop_node.evalParm("ociocolorspace") not in hou.Color.ocio_spaces():
            raise PublishValidationError(
                "Invalid value: Colorspace name doesn't exist.\n"
                "Check 'OCIO Colorspace' parameter on '{}' ROP"
                .format(rop_node.path())
            )
```

By the way note that if you don't provide a `title` argument to `PublishValidationError` that the publisher will show the label of the plug-in, and thus `title=self.label` will be redundant.

openpype/pipeline/colorspace.py
```diff
@@ -601,3 +601,63 @@ def _get_imageio_settings(project_settings, host_name):
...
+            "--view", view
+
+        ]
+        log.debug("Executing: {}".format(" ".join(args)))
+
+        process_kwargs = {
+            "logger": log
+        }
+
+        run_openpype_process(*args, **process_kwargs)
```
BigRoy, line 659: Since we're never using the dynamic nature of `process_kwargs` in here for the arguments can't this just be:
```suggestion
        run_openpype_process(*args, logger=log)
```

openpype/pipeline/colorspace.py
```diff
@@ -601,3 +601,63 @@ def _get_imageio_settings(project_settings, host_name):
...
+    with _make_temp_json_file() as tmp_json_path:
+        # Prepare subprocess arguments
+        args = [
+            "run", get_ocio_config_script_path(),
+            "config", "get_display_view_colorspace_name",
+            "--in_path", config_path,
+            "--out_path", tmp_json_path,
+            "--display", display,
+            "--view", view
+
```
BigRoy, line 651:
```suggestion
```

openpype/scripts/ocio_wrapper.py
```diff
@@ -174,5 +174,80 @@ def _get_views_data(config_path):
...
+        --out_path=<path> --display=<display> --view=<view>
+    """
+
+    json_path = Path(out_path)
+
+    out_data = _get_display_view_colorspace_name(in_path,
+                                                 display, view)
+
+    with open(json_path, "w") as f:
+        json.dump(out_data, f)
```
BigRoy, line 248: What's the reason we need to convert `out_path` here to `Path` object to then just only pass it directly to the `open` command?

Seems like this could just be:
```suggestion
    out_data = _get_display_view_colorspace_name(in_path,
                                                 display, 
                                                 view)
    with open(out_path, "w") as f:
        json.dump(out_data, f)
```

openpype/scripts/ocio_wrapper.py
```diff
@@ -174,5 +174,80 @@ def _get_views_data(config_path):
...
+              help="display name",
+              type=click.STRING)
+@click.option("--view", required=True,
+              help="view name",
+              type=click.STRING)
+def get_display_view_colorspace_name(in_path, out_path,
+                                     display, view):
+    """Aggregate view colorspace name to file.
+
+    Wrapper command for processes without acces to OpenColorIO
```
BigRoy, line 228:
```suggestion
    Wrapper command for processes without access to OpenColorIO
```

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,91 @@
...
+            cls.log.debug(
+                "'Color Correction' parm on '{}' has been set to"
+                " 'OpenColorIO'".format(rop_node.path())
+            )
+
+        # Get default view colorspace name
+        default_view_space = get_default_display_view_colorspace()
+
+        rop_node.setParms({"ociocolorspace": default_view_space})
+        cls.log.debug(
```
BigRoy, line 86: The new publisher doesn't show logs from actions yet - but if it would I'd say this message is artist-facing and thus might be better as an info log?
```suggestion
        cls.log.info(
```

openpype/scripts/ocio_wrapper.py
```diff
@@ -174,5 +174,79 @@ def _get_views_data(config_path):
...
+    Returns:
+        view color space name (str) e.g. "Output - sRGB"
+    """
+
+    config_path = Path(config_path)
+
+    if not config_path.is_file():
+        raise IOError("Input path should be `config.ocio` file")
+
+    config = ocio.Config().CreateFromFile(str(config_path))
```
BigRoy, line 198:
```suggestion
    config = ocio.Config.CreateFromFile(str(config_path))
```
Isn't [`CreateFromFile` a static method on the config class](https://opencolorio.org/old/developers/bindings/PythonAPI.html#PyOpenColorIO.Config.CreateFromFile)? And thus you don't need an instance to run it on. In your case you're creating an OCIO instance - and then generating a new instance from the class method.

I notice the same `CreateFromFile()` call is done on `ocio.Config()` across the code above but there shouldn't be any need to initialize that empty instance.

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
mustafa-zarkash: This makes sense, I think I should modify the printed message so that the artist can know that this line is related to this action.

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,90 @@
...
+from openpype.pipeline import (
+    PublishValidationError,
+    OptionalPyblishPluginMixin
+)
+from openpype.pipeline.publish import RepairAction
+from openpype.hosts.houdini.api.action import SelectROPAction
+from openpype.hosts.houdini.api.colorspace import get_default_display_view_colorspace  # noqa
+
+import os
+import hou
```
Minkiu, line 12: Don't we normally do this kind of imports inside the `process` method? I see that the `from openpype.hosts.houdini.api.colorspace` imports it at the top, so in this instance it wont make a difference

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -0,0 +1,90 @@
...
+    optional = True
+
+    def process(self, instance):
+
+        if not self.is_active(instance.data):
+            return
+
+        if os.getenv("OCIO") is None:
+            self.log.debug(
+                "Default Houdini colorspace is used, "
```
Minkiu, line 42: I guess it would be more like: `"Using Houdini's Default Color Management"` ?

mustafa-zarkash: That's a better phrasing!

Minkiu (approved):

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
mustafa-zarkash: I'm not sure tbh. 
I can find this import for example in many places in Houdini plugins, 
sometimes it's written at the beginning of the file and sometimes it's inside methods! 
```python
import hou
```

So, should we agree that if a particular import is only used within a method, should it be imported only within that method?

Minkiu: The rationale for it being within is that then it can be loaded outside of houdini, but unsure if we are actually following this or not, so don't bother for now.

BigRoy: > The rationale for it being within is that then it can be loaded outside of houdini, but unsure if we are actually following this or not, so don't bother for now.

I'd say that use case only really applies for plugins that aren't within the hosts' addon? E.g. the one for Deadline I can see that making sense (even though there it's currently not done)

antirotor: It is also affecting performance of plugin discovery - even though it happens only one when modules are loaded and then it  is much faster. But remember pymel 😱

antirotor (approved):
It does work, but there is more to be done to make it useful. To wrap what the validator is doing currently:

If OCIO is set, it checks that review ROP has `Color Correction` param set to **OpenColorIO** and that `OCIO Colorspace` is set to one defined in the config. Repair action will set it to default output.

What it does not is that it doesn't really check if to colorspace itself make sense. For example - I am creating review PNGs but because I am a moron, I set `OCIO Colorspace` to something like `Input - Panasonic - Curve - V-Log` and the validator will happily pass because it doesn't know what it the correct one.

So for that, we need something like in Maya where you define Rendering space for the scene and Display and View:

image

and so the validator can check against that. In Houdini, perhaps we don't need to define Rendering Space because as far as I know it should be always *Scene Linear* (at least for Mantra and Karma), but maybe Redshift can be set individually? Anyway, we need to add at least the View colorspace to be able to set it correctly in MPlay for example and for Review ROP if needed. But lets do it another PR - because to fully support color management in Houdini, we'll need to handle file rules properly - for example even in Review ROP, when I set its output to EXRs instead of PNGs, the colorspace shouldn't be `Output - sRGB` or whatever. And with the textures coming in, Houdini is checking file names for token and sets color correction accordingly, but without them it only de-linearize them. So proper settings for File rules in Color Management settings for Houdini is essential.

antirotor:
@MustafaJafar if you resolve conflicts, we can merge it.

openpype/hosts/houdini/plugins/publish/validate_review_colorspace.py
MustafaJafar: I moved import statement inside their relevant functions.

MustafaJafar:
@antirotor  check it now :D

PR #5324, opened 2023-07-20
Houdini: fix wrong creator identifier in pointCache workflow

antirotor:
## Changelog Description
FIxing a bug in publishing alembics, were invalid creator identifier caused missing family association.

## Additional info
Publishing of BGEO and pointCache (alembics) should work as expected

## Testing notes:
1. Open Houdini
2. create some testing geo
3. create PointCache instance
4. publishing should work with all almebics related validations, etc.
5. create BGEO instance
6. publishing should still work as expected.


mustafa-zarkash (approved):
It works as expected 👏

PR #5297, opened 2023-07-13
Maya: Include disabled plugins

tokejepsen:
## Changelog Description
When converting `animation` family or loading a `rig` family, need to include the disabled `animation` creator.

## Testing notes:
1. Convert legacy `animation` instances.
2. Verify no errors occur.
3. Load `rig` family.
4. Verify no errors occur.

PR #5326, opened 2023-07-20
Maya: All base create plugins

iLLiCiTiT:
## Changelog Description
Prepared base classes for each creator type in Maya. Extended `MayaCreatorBase` to have default implementations of common logic with instances which is used in each type of plugin.

## Additional info
This way the default instance logic is at single place and does not need to be implemented multiple times.

Idea for this PR came from this PR https://github.com/ynput/OpenPype/pull/5297

### DIFF
--- a/openpype/hosts/maya/api/plugin.py
+++ b/openpype/hosts/maya/api/plugin.py
@@ -8,13 +8,24 @@ from maya import cmds
 from maya.app.renderSetup.model import renderSetup
 
 from openpype.lib import BoolDef, Logger
-from openpype.pipeline import AVALON_CONTAINER_ID, Anatomy, CreatedInstance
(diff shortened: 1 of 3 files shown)


tokejepsen (approved):
Tested in Maya 2023 succesfully. Code looks good.

openpype/hosts/maya/api/plugin.py
```diff
@@ -177,6 +188,36 @@ def read_instance_node(self, node):
 
         return node_data
 
+    def _default_collect_instances(self):
+        self.cache_subsets(self.collection_shared_data)
```
Minkiu, line 192: Missing docstrings

Minkiu (approved):
Haven't tested, looks good.

PR #5332, opened 2023-07-20
Nuke Render and Prerender nodes Process Order - OP-3555

tokejepsen:
## Changelog Description
This PR exposes control over the order of processing of the instances, by sorting the instances created. The sorting happens on the `render_order` and subset name. If the knob `render_order` is found on the instance, we'll sort by that first before sorting by subset name.
`render_order` instances are processed before nodes without `render_order`. This could be extended in the future by querying other knobs but I dont know of a usecase for this.

Hardcoded the creator `order` attribute of the `prerender` class to be before the `render`. Could be exposed to the user/studio but dont know of a use case for this.

## Testing notes:
1. Create `prerender` and/or `render` in Nuke.
2. Experiment with different `render_order` to control the order of the processing. NOTE: this does not reflect in the UI, which is sorted alphabetically, but will show in the processing of the instances in the details.


jakubjezek001:
Eh.. Taking it back as I have been able to read Testing notes completely and the processing is correctly sorted. 
image

jakubjezek001 (approved):

PR #5334, opened 2023-07-21
Fix colorspace compatibility check

mustafa-zarkash:
## Changelog Description
for some reason a user may have `PyOpenColorIO` installed to his machine,  _in my case it came with renderman._

it can trick the compatibility check as `import PyOpenColorIO` won't raise an error 
however it may be an old version _like my case_ 

Before
compatibility check was true and It used wrapper directly 
houdini_before

After Fix 
It will use wrapper via subprocess instead
houdini_after

### DIFF
--- a/openpype/pipeline/colorspace.py
+++ b/openpype/pipeline/colorspace.py
@@ -237,10 +237,15 @@ def get_data_subprocess(config_path, data_type):
 
 
 def compatibility_check():
-    """Making sure PyOpenColorIO is importable"""
+    """checking if user has a compatible PyOpenColorIO >= 2.
+
+    It's achieved by checking if PyOpenColorIO is importable
+    and calling any version 2 specific function
+    """
     try:
         import PyOpenColorIO
-        # old PyOpenColorIO uses `PyOpenColorIO.version` instead
+
+        # ocio versions lower than 2 will raise AttributeError
         PyOpenColorIO.GetVersion()
     except (ImportError, AttributeError):
         return False
--- a/openpype/pipeline/colorspace.py
+++ b/openpype/pipeline/colorspace.py
@@ -239,8 +239,10 @@ def get_data_subprocess(config_path, data_type):
 def compatibility_check():
     """Making sure PyOpenColorIO is importable"""
     try:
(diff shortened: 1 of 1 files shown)


openpype/pipeline/colorspace.py
```diff
@@ -239,8 +239,10 @@ def get_data_subprocess(config_path, data_type):
 def compatibility_check():
     """Making sure PyOpenColorIO is importable"""
     try:
-        import PyOpenColorIO  # noqa: F401
-    except ImportError:
+        import PyOpenColorIO
+        # old PyOpenColorIO uses `PyOpenColorIO.version` instead
+        PyOpenColorIO.GetVersion()
```
Minkiu, line 244: I don't understand this, we are checking if we can import it, but we require a specific version? Or if its too old we do not want to use it? Either way, might be worth updating the doc strings to reflect what we do, since it's no longer "is importable" only.

mustafa-zarkash: > we are checking if we can import it, but we require a specific version?

yes, we uses `2.2.1` however a user can have an older version (`1.1.1` in my case), so the older version will pass the current  compatibility check however it will crash because some methods were updated 
e.g. 
this function is `2.2.1` , [reference](https://opencolorio.readthedocs.io/en/latest/api/config.html#PyOpenColorIO.Config.getDisplayViewColorSpaceName)
https://github.com/ynput/OpenPype/blob/7ee43eef7085d6176e89b1e5da124ae7da74085d/openpype/scripts/ocio_wrapper.py#L163 

unlike in `1.1.1` , [reference](https://opencolorio.org/old/developers/bindings/PythonAPI.html#PyOpenColorIO.Config.getDisplayColorSpaceName)
```python
 colorspace = config.getDisplayColorSpaceName(display, view)
```

jakubjezek001: I agree with @Minkiu we have to write into doc-string that by requiring GetVersion call we are avoiding ocio lower then 2.

jakubjezek001 (approved):
after the doc-string change is ok to be merged

Minkiu (approved):

PR #5336, opened 2023-07-21
Maya: Validate Model Content

tokejepsen:
## Changelog Description
`assemblies` in `cmds.ls` does not seem to work;
```python
from maya import cmds


content_instance = ['|group2|pSphere1_GEO', '|group2|pSphere1_GEO|pSphere1_GEOShape', '|group1|pSphere1_GEO', '|group1|pSphere1_GEO|pSphere1_GEOShape']
assemblies = cmds.ls(content_instance, assemblies=True, long=True)
print(assemblies)
```

Fixing with string splitting instead.

## Testing notes:
1. Setup model instance with geometry from two different top level parents.
2. Publish and verify validation fails.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/validate_model_content.py
+++ b/openpype/hosts/maya/plugins/publish/validate_model_content.py
@@ -63,15 +63,10 @@ class ValidateModelContent(pyblish.api.InstancePlugin):
             return True
 
         # Top group
-        assemblies = cmds.ls(content_instance, assemblies=True, long=True)
-        if len(assemblies) != 1 and cls.validate_top_group:
+        top_parents = set([x.split("|")[1] for x in content_instance])
+        if cls.validate_top_group and len(top_parents) != 1:
             cls.log.error("Must have exactly one top group")
-            return assemblies
-        if len(assemblies) == 0:
-            cls.log.warning("No top group found. "
-                            "(Are there objects in the instance?"
-                            " Or is it parented in another group?)")
-            return assemblies or True
+            return top_parents
 
         def _is_visible(node):
             """Return whether node is visible"""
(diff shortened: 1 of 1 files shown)


moonyuet (approved):
Code looks good for me...

LiborBatek (approved):
Tested with geometries within different GRPs and when validation happens it correctly fails warning user about neccessity of single parent.
image

BigRoy:
> assemblies in cmds.ls does not seem to work;

```python
from maya import cmds

content_instance = ['|group2|pSphere1_GEO', '|group2|pSphere1_GEO|pSphere1_GEOShape', '|group1|pSphere1_GEO', '|group1|pSphere1_GEO|pSphere1_GEOShape']
assemblies = cmds.ls(content_instance, assemblies=True, long=True)
print(assemblies)
```

This for me actually works fine - it returns an empty list because `assemblies` returns root level nodes only. Maybe that's not what you were expecting but it is what the `maya.cmds.ls` flag described for `assemblies`. :)

Anyway, there's another downside to allowing adding the geometry into the `model` instance directly. Are we expecting the output publish to then EXCLUDE the parents, or INCLUDE the parents? The maya scene publish for a model will by default include them for e.g. which might be the unexpected behavior? If we were expecting to EXCLUDE them the technically they will always end up not being a unique top level node (since it'd be each individual geometry mesh at the top level that was added to the instance) and thus the validation doesn't make too much sense. To me it's a bit unclear what the expected behavior is actually.

tokejepsen:
> Anyway, there's another downside to allowing adding the geometry into the model instance directly. Are we expecting the output publish to then EXCLUDE the parents, or INCLUDE the parents? The maya scene publish for a model will by default include them for e.g. which might be the unexpected behavior? If we were expecting to EXCLUDE them the technically they will always end up not being a unique top level node (since it'd be each individual geometry mesh at the top level that was added to the instance) and thus the validation doesn't make too much sense. To me it's a bit unclear what the expected behavior is actually.

I agree its not very clear.
Would you mind making an issue for this? Dont think this PR, which is just a bugfix, is the place to evaluate and implement any of this.

PR #5338, opened 2023-07-21
Blender: Added setting to set resolution and start/end frames at startup

simonebarbieri:
## Changelog Description
This PR adds `set_resolution_startup `and `set_frames_startup` settings. They automatically set respectively the resolution and start/end frames and FPS in Blender when opening a file or creating a new one.

## Testing notes:
1. Change default settings in the Project Manager for resolution, start/end frames and FPS.
2. Disable the new settings and check when opening Blender if those settings have changed. You can check from the *Property panel*, in the *Output properties* tab.
3. Now enable the settings and try opening Blender again, and check if those settings have changed.


simonebarbieri:
@moonyuet done!
image

PR #5341, opened 2023-07-24
Maya: Extract look is not AYON compatible - OP-5375

tokejepsen:
## Changelog Description
The textures that would use hardlinking are going through texture processors. Currently all texture processors are hardcoded to copy texture instead of respecting the settings of forcing to copy.
The texture processors were last modified 4 months ago, so effectively all clients that are on any pipeline updated in the last 4 months wont be utilizing hardlinking at all, since the hardcoded texture processors will copy texture no matter the OS.

This opts for completely disabling the hardlinking feature, while we figure out what to do about it.

## Testing notes:
1. Create look in Maya and disable `Force copy`.
2. Publish and verify the textures are being copied instead of hardlinked.


iLLiCiTiT:
I think there is a mismatch in PR and branch name. This does not solve the issue with usage of `legacy_io.distinct`.

tokejepsen:
> I think there is a mismatch in PR and branch name. This does not solve the issue with usage of `legacy_io.distinct`.

Since all the texture processing is done with copying rather than hardlinking due to hardcoded copying, the `legacy_io.distinct` will never be executed. I agree its not a fix but working on the AYON `legacy_io.distinct` equivalent is not worth the effort when we are not using it anyways.

I've outlined a proposal for hardlink alternative here; https://github.com/ynput/OpenPype/issues/5342.

@iLLiCiTiT do you have something in mind instead of this PR?
We can clean up the code and completely get rid of the hardlink mentions and `legacy_io.distinct` will get removed in that process.

iLLiCiTiT:
> We can clean up the code and completely get rid of the hardlink mentions and legacy_io.distinct will get removed in that process.

I don't know how it works, so I guess it's ok if it's not used. Someone has to try out it in AYON.

tokejepsen:
> I don't know how it works, so I guess it's ok if it's not used. Someone has to try out it in AYON.

@iLLiCiTiT are you requesting to clean up the code in this PR?

iLLiCiTiT:
> > I don't know how it works, so I guess it's ok if it's not used. Someone has to try out it in AYON.
> 
> @iLLiCiTiT are you requesting to clean up the code in this PR?

I don't understand anything from the logic, so I'm not the one who can tell. I though there will have to be other way to find the versions instead of using `legacy_io.distinct`. If this is ok then just raise error in `find_paths_by_hash` if is in AYON mode?
- AYON mode can be determined by `from openpype import AYON_SERVER_ENABLED`.

BigRoy:
The logic basically finds all occurrences of a certain file hash in the database (using distinct) If there's any then it hardlinks it instead of making a new copy.

How that logic would work against the Ayon backend, but note that the query is made per texture so can easily run hundreds of times for a large asset. As such it's still crucial the query remains somewhat performant.

tokejepsen:
@iLLiCiTiT what do you think of this now?

moonyuet:
I tried with AYON. It publishes the look successfully.

LiborBatek (approved):
I did followed the testing steps and it works exactly as it suppose to work.

LiborBatek:
maybe because of inability to change the way it handles texture files and always force copy them we should hide the ```Force Copy``` option on Publish look instance.
image

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -51,6 +51,12 @@ def find_paths_by_hash(texture_hash):
         str: path to texture if found.
 
     """
+    if AYON_SERVER_ENABLED:
+        raise ValueError(
```
antirotor, line 55: What about changing this to `KnownPublishError`? Because we sort of know that it might happen here :)

```suggestion
        raise KnownPublishError(
```

openpype/hosts/maya/plugins/publish/extract_look.py
```diff
@@ -585,14 +591,12 @@ def process_resources(self, instance, staging_dir, processors):
...
 
-        # Temporary fix to NOT create hardlinks on windows machines
-        if platform.system().lower() == "windows":
-            self.log.info(
-                "Forcing copy instead of hardlink due to issues on Windows..."
-            )
-            force_copy = True
-        else:
-            force_copy = instance.data.get("forceCopy", False)
+        # Temporary disable all hardlinking, due to the feature not being used
```
BigRoy, line 594: Can we add `TODO: ` for this comment so it's easily found in the codebase as code that might need some more work?

antirotor (changes requested):
let's add the TODO as mentioned an perhaps the Exception change and we can merge it.

antirotor (approved):

PR #5343, opened 2023-07-24
Blender: Extract Camera as Alembic

simonebarbieri:
## Changelog Description
Added support to extract Alembic Cameras in Blender.

## Testing notes:
1. Check that `project_settings/blender/publish/ExtractCameraABC` is enabled.
2. Try publishing a camera from Blender (be sure that the *Extract Camera (ABC)* extractor is enabled).
3. Try loading the published camera.


moonyuet (approved):
Code looks good for me. Also tested in my machine, it works as expected.
image

PR #5354, opened 2023-07-26
Chore: Schemas inside OpenPype

iLLiCiTiT:
## Changelog Description
Moved/copied schemas from repository root inside openpype/pipeline.

## Additional info
The movement is necessary because of AYON addon conversion. Since ayon-launcher does not contain the schemas and openpype neither there is not way how it can be accessed. The easiest solution is to move schemas inside openpype codebase.

Most of the schemas are very outdated and should be probably removed, but that is not purpose of this PR.

I've kept the schema directory in the root of repository and will create different [PR in next-minor branch](https://github.com/ynput/OpenPype/pull/5355) that will remove them.

### OpenPype
There should not be any issues with schema validation.
When does schema validation happen? e.g. on host install (open of DCC) or integration.

### AYON
When used openpype as an addon via ayon-launcher should not cause any issues with schemas.

### DIFF
--- a/openpype/pipeline/schema/__init__.py
+++ b/openpype/pipeline/schema/__init__.py
@@ -122,16 +122,14 @@ def _precache():
     """Store available schemas in-memory for reduced disk access"""
     global _CACHED
 
-    schema_dir = os.path.join(CURRENT_DIR, "schema")
-
-    for schema in os.listdir(schema_dir):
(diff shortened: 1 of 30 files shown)


64qam:
OpenPype is working smoothly, Ayon returns an error on launching
image

mkolar (approved):
works correctly with AYON-launcher too

PR #5358, opened 2023-07-27
Add new action to create project structure only on ftrack

mre7a:
## Changelog Description
Add a new action to create Ftrack hierarchy and avoid folder structure creation. 

## Additional info
By now, we are only able to create folder structure and also Ftrack hierarchy if needed. The problem is for some departments, we just require the structure on Ftrack, and we don't want any folder creation. 

## Testing notes:
1. Create a folder structure with [Ftrack] identifier for the project.
2. Run the Create Ftrack Project Structure action on Ftrack.
3. Just the Ftrack part of the structure should be created and only on Ftrack.


iLLiCiTiT:
Duplicating the action to remove single line is overkill. It would be better to have interface where user could choose if wants to create folder structure, or a setting where studio can choose if wants to create folder structure when the action is called.

mre7a:
> Duplicating the action to remove single line is overkill. It would be better to have interface where user could choose if wants to create folder structure, or a setting where studio can choose if wants to create folder structure when the action is called.

Thanks for the reply. I agree.
Saying that, what do you think about adding a new identifier in addition to `[ftrack]` in the code, for example `[ftrackOnly]` ?

mkolar:

We're closing it down, but we'll he happy for a new PR to [ynput/ayon-ftrack](https://github.com/ynput/ayon-ftrack) repository.

PR #5359, opened 2023-07-27
Bugfix: Houdini Creator use selection even if it was toggled off

mustafa-zarkash:
## Changelog Description
When creating many product types (families) one after another without refreshing the creator window manually 
if you toggled `Use selection` once, all the later product types will use selection even if it was toggled off

Here's Before 
it will keep use selection even if it was toggled off, unless you refresh window manually 

https://github.com/ynput/OpenPype/assets/20871534/8b890122-5b53-4c6b-897d-6a2f3aa3388a

Here's After 
it works as expected

https://github.com/ynput/OpenPype/assets/20871534/6b1db990-de1b-428e-8828-04ab59a44e28

## Testing notes:
create publishes one after another with different `Use selection` on and off

### DIFF
--- a/openpype/hosts/houdini/api/plugin.py
+++ b/openpype/hosts/houdini/api/plugin.py
@@ -170,6 +170,8 @@ class HoudiniCreator(NewCreator, HoudiniCreatorBase):
 
     def create(self, subset_name, instance_data, pre_create_data):
         try:
+            self.selected_nodes = []
+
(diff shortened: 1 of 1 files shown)


moonyuet (approved):
tested in my side, it doesn't crash anything so far.
code looks good for me too.

mkolar (approved):

PR #5360, opened 2023-07-28
Houdini: Correct camera selection for karma renderer when using selected node

moonyuet:
## Changelog Description
When user creates the karma rop with selected camera by use selection, it will give the error message of "no render camera found in selection".
This PR is to fix the bug of creating karma rop when using selected camera node in Houdini

## Additional info
n/a

## Testing notes:
1. Select a camera in obj network
2. Create Karma ROP with "use selection" enabled.
3. It will automatically fill the camera path to the camera parameter located in "out/karma_ropMain"


openpype/hosts/houdini/plugins/create/create_karma_rop.py
```diff
@@ -66,6 +66,7 @@ def create(self, subset_name, instance_data, pre_create_data):
             # we will use as render camera
             camera = None
             for node in self.selected_nodes:
+                camera = node.path()
```
mustafa-zarkash, line 69: can this line be moved inside the if statement. 

as camera is updated even if it's not camera 
image

mustafa-zarkash (changes requested):

openpype/hosts/houdini/plugins/create/create_karma_rop.py
moonyuet: it is mainly because I haven't implemented camera = node.path() into if condition. it should be fixed now.

mustafa-zarkash (approved):
Looking Good 🚀

PR #5361, opened 2023-07-28
AYON: Environment variables and functions

iLLiCiTiT:
## Changelog Description
Prepare code for ayon-launcher compatibility. Fix ayon launcher subprocess calls, added more checks for `AYON_SERVER_ENABLED`, use ayon launcher suitable environment variables in AYON mode and changed outputs of some functions. Replaced usages of `OPENPYPE_REPOS_ROOT` environment variable with `PACKAGE_DIR` variable -> correct paths are used.

## Additional info
AYON mode with ayon-launcher is not fully compatible now at some points. It tries to launch `ayon_start.py` when running from code instead of `start.py` which is there from time when OpenPype had ayon executables.

### OpenPype
**Publishing**
1. Make sure a script with subprocess is executed (for example burnins extraction)
2. Run publishing in host of your choice
3. Publishing should not crash

**Info in tray**
1. Open info window in tray.
2. Check the information if is valid.

### AYON
1. Create openpype addon
3. Upload addon to server
4. Use the addon in a bundle -> set the bundle as production (or staging)
5. Run ayon launcher
6. Do same validations as for OpenPype

### DIFF
--- a/server_addon/create_ayon_addons.py
+++ b/server_addon/create_ayon_addons.py
@@ -203,7 +203,8 @@ def create_openpype_package(
     ignored_modules = [
         "ftrack",
         "shotgrid",
-        "sync_server",
+        # Sync server is expected as always available at multuple places
+        # "sync_server",
         "example_addons",
         "slack"
     ]
--- a/server_addon/create_ayon_addons.py
+++ b/server_addon/create_ayon_addons.py
@@ -203,7 +203,7 @@ def create_openpype_package(
     ignored_modules = [
         "ftrack",
         "shotgrid",
-        # Sync server is expected as always available at multuple places
+        # Sync server is still expected at multiple places
         # "sync_server",
         "example_addons",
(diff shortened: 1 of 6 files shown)


server_addon/create_ayon_addons.py
```diff
@@ -203,7 +203,8 @@ def create_openpype_package(
     ignored_modules = [
         "ftrack",
         "shotgrid",
-        "sync_server",
+        # Sync server is expected as always available at multuple places
```
iLLiCiTiT, line 206:
```suggestion
        # Sync server is still expected at multiple places
```

kalisp (changes requested):
`ayon_launcher` throws this
AnyDeskMSI_Cv5BnT8bH5

I cannot see applications inside of `Launcher` even when I have `applications` in my production bundle and it seems they are enabled in `Studio Settings`.

PR #5367, opened 2023-07-28
Bugfix: Houdini update defaults variable

mustafa-zarkash:
## Changelog Description
So, something was forgotten while moving out from `LegacyCreator` to `NewCreator`  
`LegacyCreator` used `defaults` to list suggested subset names 
which was changed into `default_variants` in the the `NewCreator`
and setting `defaults` to any values has no effect! 

This update affects: 
- [x] Arnold ASS
- [x] Arnold ROP
- [x] Karma ROP
- [x] Mantra ROP
- [x] Redshift ROP
- [x] VRay ROP

## Additional info1
I have removed  `default_variants` which effectively set variants to `"Main"` as default, so no need to setting it explicitly.

## Additional info2
This bug is the same as #5368

## Testing notes:
1. open creator and check `Variants` menu

### DIFF
--- a/openpype/hosts/houdini/plugins/create/create_arnold_ass.py
+++ b/openpype/hosts/houdini/plugins/create/create_arnold_ass.py
@@ -10,7 +10,6 @@ class CreateArnoldAss(plugin.HoudiniCreator):
     label = "Arnold ASS"
     family = "ass"
     icon = "magic"
-    defaults = ["Main"]
 
     # Default extension: `.ass` or `.ass.gz`
     ext = ".ass"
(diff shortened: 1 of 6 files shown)


moonyuet (approved):
Tested in the rendering instances (I have different Houdini version), Looks good for me

image

mkolar (dismissed):
can we please change all of those to `main` instead of `master` so it matches the rest of the pipeline? Other than that' it's good to go

moonyuet (approved):
image
Looks good for me

PR #5368, opened 2023-07-28
Bugfix: Maya update default_variants variable

mustafa-zarkash:
## Changelog Description
So, something was forgotten while moving out from `LegacyCreator` to `NewCreator`  
`LegacyCreator` used `defaults` to list suggested subset names 
which was changed into `default_variants` in the the `NewCreator`
and setting `defaults` to any values has no effect! 

This update affects: 
- [x] Model
- [x] Set Dress

## Additional info
This bug is the same as #5367

## Testing notes:
1. open creator and check `Variants` menu


BigRoy (approved):
Didn't know about the change, but this seems good to me if that's the new name for the variable.

Issue #5328, opened 2023-07-20
Bug: Publisher infinite growth


brycegbrazen:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

On the new publisher, when clicking between the Create and Publish tab, the window grows on the right side indefinitely.

![83f162cfd73ecc53408d83ec93693c23a18a67ef](https://github.com/ynput/OpenPype/assets/69012021/a01bb768-1977-49db-b256-0ca79c0c1585)

### Expected Behavior:

The size of the window stays the same when clicking between the tabs.

### Version

3.16.0

### What platform you are running OpenPype on?

Windows

### Steps To Reproduce:

1. Open the new publisher (I opened the new publisher in Maya).
2. Click between the Create/Publish tabs.
3. Observe the bug.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

PR #5369, opened 2023-07-28
Publisher: Fix create/publish animation

iLLiCiTiT:
## Changelog Description
Use geometry movement instead of changing min/max width.

## Additional info
When animation starts the widgets are removed from parent layout to be able position them freely. When animation is finished they're locked back to layout. The ratio of widgets widths should stay the same.

This may be safer approach how the animation should happen, but there might appear artefacts outside of window borders if Qt in a DCC does not have set window clipping properly (didn't happen to me).

## Testing notes:
1. Run host of your choice (maya, nuke, houdini, other DCCs with their Qt)
2. Open publisher
3. Swap between create, publish and other tabs

Resolves https://github.com/ynput/OpenPype/issues/5328

### DIFF
--- a/openpype/tools/publisher/widgets/overview_widget.py
+++ b/openpype/tools/publisher/widgets/overview_widget.py
@@ -161,14 +161,13 @@ class OverviewWidget(QtWidgets.QFrame):
         self._change_anim = change_anim
 
         # Start in create mode
-        self._create_widget_policy = create_widget.sizePolicy()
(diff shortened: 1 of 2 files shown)


mustafa-zarkash:
The bug is no longer exists in maya.
I tested it in Maya, Houdini and Nuke, It works well.

mustafa-zarkash (approved):
The bug is no longer exists in maya.
[as a user] I tested it in Maya, Houdini and Nuke, It works well.

PR #5373, opened 2023-07-31
Enhancement: Houdini Update pointcache labels

mustafa-zarkash:
## Changelog Description
To me it's logical to find pointcaches types listed one after another, but they were named differently 

So, I made this PR to update their labels 
image

## Testing notes:
1. open OP creator window 
2. check families names

### DIFF
--- a/openpype/hosts/houdini/plugins/create/create_bgeo.py
+++ b/openpype/hosts/houdini/plugins/create/create_bgeo.py
@@ -8,7 +8,7 @@ from openpype.lib import EnumDef
 class CreateBGEO(plugin.HoudiniCreator):
     """BGEO pointcache creator."""
     identifier = "io.openpype.creators.houdini.bgeo"
-    label = "BGEO PointCache"
+    label = "PointCache (Bgeo)"
     family = "pointcache"
     icon = "gears"
 
--- a/openpype/hosts/houdini/plugins/create/create_pointcache.py
+++ b/openpype/hosts/houdini/plugins/create/create_pointcache.py
@@ -8,7 +8,7 @@ import hou
 class CreatePointCache(plugin.HoudiniCreator):
     """Alembic ROP to pointcache"""
     identifier = "io.openpype.creators.houdini.pointcache"
-    label = "Point Cache"
+    label = "PointCache (Abc)"
     family = "pointcache"
     icon = "gears"
 


moonyuet:
I dont mind the renaming but I think it is possible to just have one point cache creator and enum options which allows user to use whether abc or bgeo in the creator. 
If the user chooses abc, the collector would add abc as part of the "families"; while it would add bgeo into the families if the bgeo is chosen.

mustafa-zarkash:
> one point cache creator and enum options
I will divide that into two parts: 
1) From UX perspective: 
I like them separated because it's much clearer for user to choose and for us to code 

2) Dynamic `EnumDef` : 
I had a quick look on `openpype\lib\attribute_definitions.py` where attribute definitions live 
I think it will be dynamic by
- having a `get_current_selection()` method
- allowing having callable attributes
- making minimal UI refresh when making a selection 

So, this code be valid 
```python
select_type = EnumDef("select_type", items={ 0: "Alembic", 1: "Bgeo"},
                      default=0, label="Select Type")
                                        
bgeo_options = EnumDef("bgeo_type", bgeo_enum, label="BGEO Options", 
                      hidden= select_type.get_current_selection()!="Bgeo",
                      disabled= select_type.get_current_selection()!="Bgeo" )  
```

I believe it may be beneficial however it will add some pain. 
Also, my analysis could be wrong, I can't tell for sure. 

@moonyuet

BigRoy:
As far as I know attribute definations are per Creator currently and not per instance - so they can't be that dynamic I believe.
I personally also like the 'cleanness' of keeping them separated but would welcome anything that "simplifies" it further, cleans things up AND gets the pointcaches closes together functionally for the artist.

Regarding this PR - I don't mind either of the label namings. The new labels do make them feel like the `pointcache` behavior will be 'quite similar' between the two which I think technically isn't the case. E.g. BGEO supports way more data than alembic, right?

mustafa-zarkash:
I agree, BGEO can include a wide range of data other than `pointcache`. 

However, the current implementation `BGEO pointcache` expects a sop selection and the companion validatior `bgeo sop path` is checking and confirming the soppath is pointing to a sop node but I can't remember if there's a validatior for checking data inside that sop node 

Maybe tomorrow, I will test it to check its limitations. 

And I really don't know what'd be more accurate  `BGEO` or `BGEO PointCache` 





FYI, I made this PR just because there are `BGEO PointCache` and `Point Cache` families and it bothered me that they share the same family but not listed near to each other.

mkolar:
I agree with mustafa that they should at least live together in the creator menu, which this PR simply solves. That being said we should alsow work towards making the whole setup better, but that's not the point of this PR.

mkolar (approved):

PR #5377, opened 2023-08-01
AYON: Fix settings conversion for ayon addons

iLLiCiTiT:
## Changelog Description
AYON addon settings are available in system settings and does not have available the same values in `"modules"` subkey.

## Additional info
The values in `"modules"` subkey is added because addons can have access only to these settings on initialization.

## Testing notes:
1. Ftrack and slack AYON addons should work with these changes.

### DIFF
--- a/openpype/settings/ayon_settings.py
+++ b/openpype/settings/ayon_settings.py
@@ -267,6 +267,7 @@ def _convert_modules_system(
     ):
         func(ayon_settings, output, addon_versions, default_settings)
 
+    modules_settings = output["modules"]
     for module_name in (
         "sync_server",
         "log_viewer",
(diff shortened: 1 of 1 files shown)


openpype/settings/ayon_settings.py
```diff
@@ -293,15 +303,17 @@ def convert_system_settings(ayon_settings, default_settings, addon_versions):
     if "core" in ayon_settings:
         _convert_general(ayon_settings, output, default_settings)
 
+
+    for key, value in default_settings.items():
```
iLLiCiTiT, line 307:
```suggestion
    for key, value in default_settings.items():
```

mkolar (approved):

PR #5378, opened 2023-08-01
Max: Adding custom modifiers back to the loaded objects 

moonyuet:
## Changelog Description
The custom parameters OpenpypeData doesn't show in the loaded container when it is being loaded through the loader.

## Additional info
<del>Loading Max Scene is a bit different from the rest.</del>
<del>It would be resolved in other ticket: https://github.com/ynput/OpenPype/pull/5406</del> 
<del>Should be merged before #5406Fixed the conflict</del>
Resolve #5406 
This ticket also affects #5434 
Related to "3dsMax-loading-an-asset-multiple-times"
## Testing notes:
1. Publish something in Max(any instance except render)
2. Load instances(e.g. model, camera etc.)
3. The OP Parameter should be embedded into the loaded container attribute 
image
4. Switch version 
5. Check the OP param and objects 
6. They should be updated


LiborBatek (changes requested):
Seems working when using ```Load``` in ```max scene format``` aka representation...
image

But it doesnt create ```OP Container``` when loaded as ```Alembic``` repre.
image

Not sure if it suppose to work also when Alembic repre used tho.

moonyuet:
@LiborBatek  that should be fixed. 

image

LiborBatek (approved):
Works also on ```Alembic``` representation now.

moonyuet:
@LiborBatek Just want you to do one more test on maxScene, thank you!

LiborBatek (changes requested):
I did re-publish of the ```model``` instance for new set of testing but getting errors when trying to ```load``` ```Alembic``` repre of the asset into empty scene:

image

also missing ```OP Parameters``` in the modifier stack when loading ```maxScene``` repre of that freshly published asset as well

image

Seems completely not working this time...neither ```maxScene``` nor ```Alembic``` properly got loaded into scene.

moonyuet:
@LiborBatek I have aligned the Openpype Data function along with #5406 
Can you please try?

moonyuet:
I just find out that there should be some adjustment after using modifiers with OP data. Will update you after the fix.

moonyuet:
The updated PR should load the OP data with modifiers in the container (while the abc loader would load as alembic container), and also set or update version correctly. @LiborBatek, can you do the test with that?
image
image

moonyuet:
> Does work for `Alembic` obviously present `OP Data` modifier now
> 
> image
> 
> however I lack any traces of `OP Data` modifier when loading the `asset` as `maxScene` tho image
> 
> Is this also correct? I guess its not properly working or?

the loading modifier for maxScene is in https://github.com/ynput/OpenPype/pull/5406

LiborBatek (approved):
Now in both scenarios all working fine.

For asset loaded as ```Alembic``` and also as ```MaxScene```  asset having ```OP Data``` modifier

image

LiborBatek (changes requested):
Did test to load multiple times the same ```asset``` model and also in 2 different repre ( ```maxScene``` and ```Alembic``` ) all seems fine.
image

but when trying to change version via ```Manage``` its screwed...

image

LiborBatek:
namespaces are generally fine but the managing of versions doesnt work ok. But maybe for other PR?

Should we split it or try to fix it within this PR?

moonyuet:
@LiborBatek I have updated the PR. It should be updating the OP parameter as expected and managing version as expected.
But one thing is the alembic object would be a bit different from the rest. I will resolve this in #5424 along with fbx and alembic objects deletion issue.
image

image

LiborBatek:
So I have tested just to load ```model``` in ```maxScene``` repre but 2 different versions at once into same scene and trying to switch versions on each...doesnt work. Pls see the video better describing what happenned...

https://github.com/ynput/OpenPype/assets/112623825/7b8a088f-c379-4d6a-912e-2092ab3f979d

moonyuet:
namespace fix:
- [x] maxScene
- [ ]  pointcloud
- [ ]  redshift proxy
- [ ]  fbx loaders
- [ ]  usd loader
- [ ]  obj loader

container position fix:
- [ ] maxScene
- [ ]  pointcloud
- [ ]  redshift proxy
- [ ]  fbx loaders
- [ ]  usd loader
- [ ]  obj loader

moonyuet:
@LiborBatek, I have updated the PR. can you test it again?

LiborBatek:
I get namespaces when loading single asset multiple times and in misc representations (maxScene and Alembic).
image

However I still experience issues when those containers changed in position/rotation and later trying ```set version``` or ```update to latest``` the transforms get messed up and also my container endup mixed up...

image
check those ```modelMain_04``` ...weird

moonyuet:
> 

For the current stages it is better to stick with the original container but still I will keep the codes there for just in case if some users also want the positions of objects/containers updating. 
For the namespace, it's better to test it after publishing several versions. It is not valid for the old versions for the update of the PR.

moonyuet:
@LiborBatek The updated PRs would update the namespace if two containers with "ModelMain" exists. The one with the OP data will have the extension of "_param"
For the current stage, the rotation of the objects wont be updated as it seems that the dictionary doesn't accept "max-type" of quat data.

LiborBatek (approved):
It correctly loads the published assets with OP Data containers, also ```Manage``` and switch versions works ok.

image

openpype/hosts/max/api/lib.py
```diff
@@ -312,3 +312,99 @@ def set_timeline(frameStart, frameEnd):
...
+    >>> unique_namespace(":hello")
+    # :hello01
+    >>> unique_namespace("bar:", suffix="_NS")
+    # bar01_NS:
+
+    """
+
+    def current_namespace():
+        current = namespace
+        # When inside a namespace Maya adds no trailing :
```
Minkiu, line 339: `Max` instead of `Maya` (?)

openpype/hosts/max/api/lib.py
```diff
@@ -312,3 +312,98 @@ def set_timeline(frameStart, frameEnd):
     """
     rt.animationRange = rt.interval(frameStart, frameEnd)
     return rt.animationRange
+
+
+def unique_namespace(namespace, format="%02d",
```
Minkiu, line 317: There seems to be a `uniqueName` method to do this?
https://help.autodesk.com/view/MAXDEV/2022/ENU/?guid=GUID-D1D7EB56-A370-4B07-99B4-BC779FB87CAF

In the "Associated Methods" section, wouldn't be better to use that?

openpype/hosts/max/plugins/load/load_camera_fbx.py
```diff
@@ -13,50 +22,76 @@ class FbxLoader(load.LoaderPlugin):
...
-        """)
-        rt.Execute(fbx_reimport_cmd)
+        node_name = container["instance_node"]
+        node = rt.getNodeByName(node_name)
+        namespace, name = get_namespace(node_name)
+        sub_node_name = f"{namespace}:{name}_{self.postfix}"
+        inst_container = rt.getNodeByName(sub_node_name)
+        rt.Select(inst_container.Children)
+        transform_data = object_transform_set(inst_container.Children)
+        for prev_fbx_obj in rt.selection:
```
Minkiu, line 69: Out of curiosity, what's the difference between `rt.selection` and `rt.GetCurrentSelection` ?

openpype/hosts/max/plugins/load/load_model_fbx.py
```diff
@@ -13,46 +21,77 @@ class FbxModelLoader(load.LoaderPlugin):
...
+                rt.Delete(prev_fbx_obj)
 
         rt.FBXImporterSetParam("Animation", False)
         rt.FBXImporterSetParam("Cameras", False)
+        rt.FBXImporterSetParam("Mode", rt.Name("merge"))
         rt.FBXImporterSetParam("AxisConversionMethod", True)
-        rt.FBXImporterSetParam("UpAxis", "Y")
         rt.FBXImporterSetParam("Preserveinstances", True)
         rt.importFile(path, rt.name("noPrompt"), using=rt.FBXIMP)
+        current_fbx_objects = rt.GetCurrentSelection()
```
Minkiu, line 74: I guess this section of the load could live in a  `load/fbx_utils.py` ina later PR, since we use it int he `load_camera_fbx` as well

openpype/hosts/max/plugins/load/load_camera_fbx.py
moonyuet: > Out of curiosity, what's the difference between `rt.selection` and `rt.GetCurrentSelection` ?

Not much differences in this case.

openpype/hosts/max/api/lib.py
moonyuet: Unqiue Name is more like re-naming two teapots into "Teapot001", "Teapot002" for example. It's like MergeMaxFile with Auto Rename Duplicate being enabled. It is doable for load function but not sure if it can be done in up-versioning and versioning setup as it will give the new name to the objects and not sure if the system in Max is wise enough to find the last merged nodes. (Some loaders like MaxScene loader would subtract the last merged nodes in previous import from the merged nodes in the latest import and remove the objects which doesn't exist in the latest import)

PR #5379, opened 2023-08-01
Nuke: OCIO env var workflow

jakubjezek001:
## Changelog Description
The OCIO environment variable needs to be consistently handled across all platforms. Nuke resolves the custom OCIO config path differently depending on the platform, so we included the ocio config path in the workfile with a partial replacement using an environment variable. Additionally, for Windows sessions, we replaced backward slashes with a TCL expression.

## Testing notes:
1. activate Colorspace management and add config path with root token
2. open Nuke in the project and see the custom ocio config path was set to some strange values  
image
In my case I was setting the path within `{root[resources]}/...`
image
If you want to make a new drive letter from some folder in case you want to test multi root - here it how to do it. 
`subs X: C:\path\to\folder`
3. submit publish deadline farm job
4. make sure your last version .zip package is created and Deadline can see it.
5. all should be rendered as expected

## Dependencies:
https://github.com/ynput/OpenPype/pull/5387
https://github.com/ynput/OpenPype/pull/5411


64qam:
Apply settings do not work for some reason. Deadline crashes with error:
```Error: FailRenderException : >>> [12:40.42] ERROR: renderTextureMain.inside_renderTextureMain: Invalid LUT selected : linear
   at Deadline.Plugins.DeadlinePlugin.FailRender(String message) (Python.Runtime.PythonException)
  File "C:\ProgramData\Thinkbox\Deadline10\workers\qamtop\plugins\64cb841b5001a6ac7cdf94ef\Nuke.py", line 264, in RenderTasks
    self.WritePython(scriptContents)
  File "C:\ProgramData\Thinkbox\Deadline10\workers\qamtop\plugins\64cb841b5001a6ac7cdf94ef\Nuke.py", line 41, in WritePython
    self.WaitForProcess()
  File "C:\ProgramData\Thinkbox\Deadline10\workers\qamtop\plugins\64cb841b5001a6ac7cdf94ef\Nuke.py", line 48, in WaitForProcess
    self.FlushMonitoredManagedProcessStdout( self.ProcessName )
  File "C:\ProgramData\Thinkbox\Deadline10\workers\qamtop\plugins\64cb841b5001a6ac7cdf94ef\Nuke.py", line 623, in HandleError
    self.deadlinePlugin.FailRender( self.GetRegexMatch( 0 ) )
   at Python.Runtime.Dispatcher.Dispatch(ArrayList args)
   at __FranticX_GenericDelegate0Dispatcher.Invoke()
   at Deadline.Plugins.DeadlinePlugin.RenderTasks()
   at Deadline.Plugins.DeadlinePlugin.DoRenderTasks()
   at Deadline.Plugins.PluginWrapper.RenderTasks(Task task, String& outMessage, AbortLevel& abortLevel)
   at Deadline.Plugins.PluginWrapper.RenderTasks(Task task, String& outMessage, AbortLevel& abortLevel)
```

openpype/hosts/nuke/api/lib.py
```diff
@@ -2094,9 +2101,29 @@ def set_root_colorspace(self, imageio_host):
...
+            log.info("OCIO config path found: `{}`".format(
+                config_data["path"]))
+
+            # check if there's a mismatch between environment and settings
+            wrong_environment = self._is_settings_different_from_environment(
+                config_data)
+
+            # if there's no mismatch between environment and settings
+            if not wrong_environment:
+                self._set_ocio_config_path_to_workfile(config_data)
```
BigRoy, line 2113: The logic here to me would be much clearer if it would just read:

```suggestion
            config_path = config_data["path"]
            log.info("OCIO config path found: `{}`".format(config_path))
            
            if config_path == os.getenv("OCIO"):
                self._set_ocio_config_path_to_workfile(config_data)
```

The double negative check of "is it different" but then `not different` seems a bit confusing to read.

Also, I suspect the OCIO path checks might need to also check with normalised `os.path` unless we know for certain the env var is set exactly (matching case + forward/backward slashes) to the value of `config_data["path"]`.

openpype/hosts/nuke/api/lib.py
```diff
@@ -2114,12 +2141,87 @@ def set_root_colorspace(self, imageio_host):
...
+            str: OCIO config path with environment variable
+        """
+        # QUESTION: should we also include other names variants
+        included_vars = [
+            "BUILTIN_OCIO_ROOT",
+            "OPENPYPE_PROJECT_ROOT"
+        ]
+        for env_var, env_path in os.environ.items():
+            # first check if variable is whitelisted
+            if all(var_ not in env_var for var_ in included_vars):
```
BigRoy, line 2201: Are we expecting the included vars to be a PART of the environment key names? Could you provide an example of what names of vars we'd be matching?

64qam (approved):
LGTM

openpype/hosts/nuke/api/lib.py
BigRoy: Ah, `OPENPYPE_PROJECT_ROOT_RESOURCES` seems to be one of them.

openpype/hosts/nuke/api/lib.py
```diff
@@ -2114,12 +2141,87 @@ def set_root_colorspace(self, imageio_host):
...
+        environment variable if possible. If not, it will set it as it is.
+        It also saves the script to apply the change, but only if it's not
+        empty Untitled script.
+
+        Args:
+            config_data (dict): OCIO config data from settings
+
+        """
+        # replace path with env var if possible
+        ocio_path = self._replace_ocio_path_with_env_var(
```
iLLiCiTiT, line 2165: I guess that most of studio will use their custom environment variables, and because `_replace_ocio_path_with_env_var` count with only default env variables, I would suggest to use "raw" ocio path from settings, instead of trying to find possible environment variable in already filled path, or maybe just use `OCIO` env variable (which is filled in global prelaunc hook).

That would break only if settings for OCIO path would change during production, I don't know if we should allow that?

NOTE: I've also seen settings for OCIO withou any env varible (full path for each platform).

openpype/hosts/nuke/api/lib.py
jakubjezek001: Thanks for the suggestion. I'll process it differently, but yeah, I agree the double negative is confusing.

openpype/hosts/nuke/api/lib.py
jakubjezek001: yes, thanks for suggestion. It is processed.

jakubjezek001:
This PR https://github.com/ynput/OpenPype/pull/5411 would be good to include so the environment variable workflow is working as it should.

openpype/hosts/nuke/api/lib.py
```diff
@@ -2120,12 +2153,119 @@ def set_root_colorspace(self, imageio_host):
...
+        ]
+
+        # include all project root related env vars
+        for env_var in os.environ:
+            if env_var.startswith("OPENPYPE_PROJECT_ROOT_"):
+                included_vars.append(env_var)
+
+        # use regex to find env var in template with format {ENV_VAR}
+        # this way we make sure only template used env vars are included
+        env_var_regex = r"\{([A-Z_]+)\}"
```
iLLiCiTiT, line 2215: Missing numbers
```suggestion
        env_var_regex = r"\{([A-Z0-9_]+)\}"
```

openpype/hosts/nuke/api/lib.py
```diff
@@ -2120,12 +2153,119 @@ def set_root_colorspace(self, imageio_host):
...
+        included_vars = self._get_included_vars(config_template)
+
+        # make sure we return original path if no env var is included
+        new_path = config_path
+
+        for env_var, env_path in os.environ.items():
+            # first check if variable is whitelisted
+            if env_var not in included_vars:
+                # included vars not found in env_var name
+                continue
```
iLLiCiTiT, line 2247:
```suggestion
        for env_var in included_vars:
            env_path = os.getenv(env_var)
```

iLLiCiTiT (approved):
Code looks ok, needs testing.

64qam (approved):

PR #5384, opened 2023-08-02
Houdini: Submit houdini patch version to deadline

mustafa-zarkash:
## Changelog Description
By default houdini  deadline submission includes `major.minor` version numbers  only 

As requested, to include patch version as follows `major.minor.patch` 
image

## Additional Info
This modifications require to update houdini deadline plugin param file. 
houdini custom plugin will override the default plugin 
I have added a mini param editing guide in deadline readme file

## Testing notes:
1. modify  `custom\plugins\Houdini\Houdini.param` as suggested 
2. copy it to deadline custom plugins 
3. submit a deadline job from OP


BigRoy:
I'm not a fan of including our own custom full Houdini plug-in for Deadline. It can easily get outdated compared to Deadline updates and I suspect it'll be hard to keep track of what changes were made for OpenPype and which were made due to being outdated - especially if currently all that's changed is the `.param` file, right? It feels quite bloated.

---

I'd say instead of this, couldn't we instead just rely on `AVALON_APP_NAME` environment value. That would actually result in a local executable to run, right - and would then actually also be using the one defined by OpenPype? Would it be possible instead to write logic that just replaces the Houdini executable that will get triggered to begin with with the one from solely that variable?

mustafa-zarkash:
> couldn't we instead just rely on `AVALON_APP_NAME` environment value.
I'm not sure how to tell deadline to resolve  `AVALON_APP_NAME` to get the right executable.
may it can call OP wrapper like `OCIO` functions. 

> replaces the Houdini executable that will get triggered to begin with with the one from solely that variable?

I think this will require modifing `plugins\Houdini\Houdini.py` 
HythonExecutable
```python
    def RenderExecutable( self ):
        version = self.GetPluginInfoEntryWithDefault( "Version", "18.5" ).replace( ".", "_" )
        return self.GetRenderExecutable( "Houdini" + version + "_Hython_Executable", "Houdini " + version)
```

SimTracker
```python
    version = self.GetPluginInfoEntryWithDefault( "Version", "18.5" ).replace( ".", "_" )
    # Retrieve houdini simtracker executable
    simTracker = self.GetRenderExecutable( "Houdini" + version + "_SimTracker", "Houdini " + version + " SimTracker")
```

Maybe we  can just send the executable path and tell `RenderExecutable` to use it if exists 
with a BIG NOTE that paths in deadline `Configure Plugins UI` will be ignored in favor of paths set in OP system settings.
```python
    def RenderExecutable( self ):
        hou_exe = self.GetPluginInfoEntryWithDefault( "Houdini_exe", "" )
        if not hou_exe:
            version = self.GetPluginInfoEntryWithDefault( "Version", "18.5" ).replace( ".", "_" )
            hou_exe = self.GetRenderExecutable( "Houdini" + version + "_Hython_Executable", "Houdini " + version)
        return hou_exe 
```
same for `SimTracker`

mkolar:
Man I don't like any of this to be honest, but it is actual scenario. I haven't tried myself, but @mustafa-zarkash can we make it so it's backwards compatible with default houdini deadline plugin? The point is that for majority of studios major.minor is probably enough, considering we've only had one request for this and I'm no too keen on maintaining the full plulgin all the time. We could possibly just make sure we're able to send the patch version and give studio the option to maintain their own deadline plugin. It might accidentally creep the scope of houdini deadline maintenance for us.

mustafa-zarkash:
@mkolar I agree. 

I think adding a toggle some where in settings so that admins can tell deadline publisher to send patch version or not with a big note that enabling this requirs modifying houdini.param with a mini guide about how to do so.

BigRoy:
So I just learned that Deadline has a "Rez" plugin apparently, [here are Deadline + Rez plugin](https://docs.thinkboxsoftware.com/products/deadline/10.2/1_User%20Manual/manual/event-rez.html). It uses the On Job Submitted event just to insert some Rez things to 'configure' things alright for the to be launched application.

I would expect to be able to do something similar where we can e.g. in GlobalJobPreLoad or whatever initialize the relevant Application environment (or make the executable accessible in a way for example add the executable's folder to start of `PATH` env var) and have the Houdini plugin in Deadline settings just refer to `houdini.exe` instead of a full executable path.

That way we can influence how Deadline launches houdini, but don't need to modify Houdini plugin itself.

mustafa-zarkash:
@mkolar @BigRoy  
So, your comments inspired me. 
I have added Houdini deadline project settings with a little toggle to switch between `major.minor` and `major.minor.patch`
and it's `False` by default so everyone will act as usual. 
whomever want enable it will be required to update only `DeadlineRepo\plugins\Houdini\Houdini.param`, I created a mini guide about how to do that but I don't know where to add it maybe to docs whenever I know how to do that!

moonyuet:
i tested with the 19.5.435 version of Houdini It doesn't work and gives the error below. (I guess this one is still in draft, right?)
```
﻿=======================================================
Error
=======================================================
Error: RenderPluginException : GetConfigEntry: Attempted to access non-existent config key: Houdini19_5_435_Hython_Executable
   at Deadline.Plugins.DeadlinePlugin.GetConfigEntry(String key) (Python.Runtime.PythonException)
  File "C:\ProgramData\Thinkbox\Deadline10\workers\desktop-1u8vp0c\plugins\64d0ceedb225af410c82de6e\Houdini.py", line 78, in RenderExecutable
    houdiniExeList = self.GetConfigEntry( "Houdini" + version + "_Hython_Executable" )
   at Python.Runtime.Dispatcher.Dispatch(ArrayList args)
   at __FranticX_GenericDelegate0`1\[\[System_String\, System_Private_CoreLib\, Version=4_0_0_0\, Culture=neutral\, PublicKeyToken=7cec85d7bea7798e\]\]Dispatcher.Invoke()
   at FranticX.Processes.ManagedProcess.RenderExecutable()
   at Deadline.Plugins.DeadlinePlugin.RenderExecutable()
   at FranticX.Processes.ManagedProcess.Execute(Boolean waitForExit)
   at Deadline.Plugins.PluginWrapper.RenderTasks(Task task, String& outMessage, AbortLevel& abortLevel)

=======================================================
Type
=======================================================
RenderPluginException

=======================================================
Stack Trace
=======================================================
   at Deadline.Plugins.SandboxedPlugin.d(DeadlineMessage bgj, CancellationToken bgk)
   at Deadline.Plugins.SandboxedPlugin.RenderTask(Task task, CancellationToken cancellationToken)
   at Deadline.Slaves.SlaveRenderThread.c(TaskLogWriter ajq, CancellationToken ajr)

=======================================================
Log
=======================================================
2023-08-07 19:01:23:  0: Render Thread - Render State transition from = 'ReceivedTask' to = 'Other'
2023-08-07 19:01:23:  0: Loading Job's Plugin timeout is Disabled
2023-08-07 19:01:23:  0: SandboxedPlugin: Render Job As User disabled, running as current user 'Kayla'
2023-08-07 19:01:25:  0: Loaded plugin Houdini
2023-08-07 19:01:25:  0: Executing plugin command of type 'Initialize Plugin'
2023-08-07 19:01:25:  0: INFO: Executing plugin script 'C:\ProgramData\Thinkbox\Deadline10\workers\desktop-1u8vp0c\plugins\64d0ceedb225af410c82de6e\Houdini.py'
2023-08-07 19:01:25:  0: INFO: Plugin execution sandbox using Python version 3
2023-08-07 19:01:25:  0: INFO: About: Houdini Plugin for Deadline
2023-08-07 19:01:25:  0: INFO: The job's environment will be merged with the current environment before rendering
2023-08-07 19:01:25:  0: Done executing plugin command of type 'Initialize Plugin'
2023-08-07 19:01:25:  0: Start Job timeout is disabled.
2023-08-07 19:01:25:  0: Task timeout is disabled.
2023-08-07 19:01:25:  0: Loaded job: new_project_ep_asset_v214.hiplc - redshift_ropMain (64d0ceedb225af410c82de6e)
2023-08-07 19:01:25:  0: Executing plugin command of type 'Start Job'
2023-08-07 19:01:25:  0: DEBUG: S3BackedCache Client is not installed.
2023-08-07 19:01:25:  0: INFO: Executing global asset transfer preload script 'C:\ProgramData\Thinkbox\Deadline10\workers\desktop-1u8vp0c\plugins\64d0ceedb225af410c82de6e\GlobalAssetTransferPreLoad.py'
2023-08-07 19:01:25:  0: INFO: Looking for legacy (pre-10.0.26) AWS Portal File Transfer...
2023-08-07 19:01:25:  0: INFO: Looking for legacy (pre-10.0.26) File Transfer controller in C:/Program Files/Thinkbox/S3BackedCache/bin/task.py...
2023-08-07 19:01:25:  0: INFO: Could not find legacy (pre-10.0.26) AWS Portal File Transfer.
2023-08-07 19:01:25:  0: INFO: Legacy (pre-10.0.26) AWS Portal File Transfer is not installed on the system.
2023-08-07 19:01:25:  0: Done executing plugin command of type 'Start Job'
2023-08-07 19:01:25:  0: Plugin rendering frame(s): 21-24
2023-08-07 19:01:25:  0: Render Thread - Render State transition from = 'Other' to = 'Rendering'
2023-08-07 19:01:25:  0: Executing plugin command of type 'Render Task'
2023-08-07 19:01:25:  0: INFO: Starting Houdini Job
2023-08-07 19:01:25:  0: INFO: Stdout Redirection Enabled: True
2023-08-07 19:01:25:  0: INFO: Stdout Handling Enabled: True
2023-08-07 19:01:25:  0: INFO: Popup Handling Enabled: True
2023-08-07 19:01:25:  0: INFO: QT Popup Handling Enabled: False
2023-08-07 19:01:25:  0: INFO: WindowsForms10.Window.8.app.* Popup Handling Enabled: False
2023-08-07 19:01:25:  0: INFO: Using Process Tree: True
2023-08-07 19:01:25:  0: INFO: Hiding DOS Window: True
2023-08-07 19:01:25:  0: INFO: Creating New Console: False
2023-08-07 19:01:25:  0: INFO: Running as user: Kayla
2023-08-07 19:01:25:  0: Done executing plugin command of type 'Render Task'

=======================================================
Details
=======================================================
Date: 08/07/2023 19:01:29
Frames: 21-24
Elapsed Time: 00:00:00:06
Job Submit Date: 08/07/2023 19:01:01
Job User: kayla
Average RAM Usage: 16114520064 (48%)
Peak RAM Usage: 16114520064 (48%)
Average CPU Usage: 1%
Peak CPU Usage: 2%
Used CPU Clocks (x10^6 cycles): 1057
Total CPU Clocks (x10^6 cycles): 105649

=======================================================
Worker Information
=======================================================
Worker Name: DESKTOP-1U8VP0C
Version: v10.1.22.5 Release (919d43564)
Operating System: Windows 11 Pro
Running As Service: No
Machine User: Kayla
IP Address: 192.168.50.249
MAC Address: 48:E7:DA:F4:8E:23
CPU Architecture: x64
CPUs: 16
CPU Usage: 6%
Memory Usage: 15.0 GB / 31.8 GB (47%)
Free Disk Space: 1.866 TB (460.553 GB on C:\, 1.416 TB on D:\)
Video Card: NVIDIA GeForce RTX 3070
```

mustafa-zarkash:
> i tested with the 19.5.435 version of Houdini It doesn't work

To check if this PR is working we need to check the job parameters.
for example

image

---

> Error: RenderPluginException : GetConfigEntry: Attempted to access non-existent config key: Houdini19_5_435_Hython_Executable

The thing is: by default deadline will complain about it can't find the houdin exe 
so that admins have to update `DeadlineRepo\plugins\Houdini\Houdini.param` 
I made a mini guide about it and I'm going to update the docs shortly (I didn't do so earlier because I didn't know where should I add this mini guide. )

In short: these blocks should added to `houdini.param` and replace `<major>`, `<minor>` and `<patch>` with the desired numbers

#### Hython Executable block
```
[Houdini<major>_<minor>_<patch>_Hython_Executable]
Label=Houdini <major>.<minor>.<patch> Hython Executable
Category=Render Executables
CategoryOrder=0
Type=multilinemultifilename
Index=14
Default=C:\Program Files\Side Effects Software\Houdini 19.0.000\bin\hython.exe;/Applications/Houdini/Houdini19.0.000/Frameworks/Houdini.framework/Versions/19.0.000/Resources/bin/hython;/opt/hfs19.0/bin/hython
Description=The path to the hython executable. It can be found in the Houdini bin folder.
```

#### Sim Tracker block
```
[Houdini<major>_<minor>_<patch>_SimTracker]
Label=Houdini <major>.<minor>.<patch> Sim Tracker File
Category=HQueue Simulation Job Options
CategoryOrder=1
Type=multilinemultifilename
Index=10
Default=C:\Program Files\Side Effects Software\Houdini 19.0.000\houdini\python3.7libs\simtracker.py;/Applications/Houdini/Houdini19.0.000/Frameworks/Houdini.framework/Versions/19.0.000/Resources/houdini/python3.7libs/simtracker.py;/opt/hfs19.0/houdini/python3.7libs/simtracker.py
Description=The path to the simtracker.py file that is used when distributing HQueue sim jobs. This file can be found in the Houdini install.
```

moonyuet:
> > i tested with the 19.5.435 version of Houdini It doesn't work
> 
> To check if this PR is working we need to check the job parameters. for example
> 
> image
> 
> > Error: RenderPluginException : GetConfigEntry: Attempted to access non-existent config key: Houdini19_5_435_Hython_Executable
> 
> The thing is: by default deadline will complain about it can't find the houdin exe so that admins have to update `DeadlineRepo\plugins\Houdini\Houdini.param` I made a mini guide about it and I'm going to update the docs shortly (I didn't do so earlier because I didn't know where should I add this mini guide. )
> 
> In short: these blocks should added to `houdini.param` and replace `<major>`, `<minor>` and `<patch>` with the desired numbers
> 
> #### Hython Executable block
> ```
> [Houdini<major>_<minor>_<patch>_Hython_Executable]
> Label=Houdini <major>.<minor>.<patch> Hython Executable
> Category=Render Executables
> CategoryOrder=0
> Type=multilinemultifilename
> Index=14
> Default=C:\Program Files\Side Effects Software\Houdini 19.0.000\bin\hython.exe;/Applications/Houdini/Houdini19.0.000/Frameworks/Houdini.framework/Versions/19.0.000/Resources/bin/hython;/opt/hfs19.0/bin/hython
> Description=The path to the hython executable. It can be found in the Houdini bin folder.
> ```
> 
> #### Sim Tracker block
> ```
> [Houdini<major>_<minor>_<patch>_SimTracker]
> Label=Houdini <major>.<minor>.<patch> Sim Tracker File
> Category=HQueue Simulation Job Options
> CategoryOrder=1
> Type=multilinemultifilename
> Index=10
> Default=C:\Program Files\Side Effects Software\Houdini 19.0.000\houdini\python3.7libs\simtracker.py;/Applications/Houdini/Houdini19.0.000/Frameworks/Houdini.framework/Versions/19.0.000/Resources/houdini/python3.7libs/simtracker.py;/opt/hfs19.0/houdini/python3.7libs/simtracker.py
> Description=The path to the simtracker.py file that is used when distributing HQueue sim jobs. This file can be found in the Houdini install.
> ```
> 
> After Updating that file, you will find new entries to add houdini exe image

Would be great if you can have a doc to tell the 19.5 users how to add the params into the Houdini (which file to edited etc.) so that it would update the plugin configuration.

mustafa-zarkash:
@moonyuet  I have updated the docs.

moonyuet:
> So I just learned that Deadline has a "Rez" plugin apparently, [here are Deadline + Rez plugin](https://docs.thinkboxsoftware.com/products/deadline/10.2/1_User%20Manual/manual/event-rez.html). It uses the On Job Submitted event just to insert some Rez things to 'configure' things alright for the to be launched application. I suspect since Rez basically also builds up an environment that could supply e.g. a Houdini version that it might've tried to solve a similar issue? Maybe it's a decent reference at least as to what we can do without touching the Houdini plugin itself.
> 
> I would expect to be able to do something similar where we can e.g. in GlobalJobPreLoad or whatever initialize the relevant Application environment (or make the executable accessible in a way for example add the executable's folder to start of `PATH` env var) and have the Houdini plugin in Deadline settings just refer to `houdini.exe` instead of a full executable path.
> 
> That way we can influence how Deadline launches houdini, but don't need to modify Houdini plugin itself.

I just saw your comment now. Adding param can be messy if the users modify Houdini plugin especially when switching from different projects with different Houdini versions(some renderers such as Arnold and Redshift can only be used in certain patch versions). I need to download two different versions of Houdini to test things and I have to make sure to download the right minor version.
After looking at the Rez link, I am thinking if we can make a customized plugin or edit the current one we have to find the executable plugin by patch version and use it for rendering. This is not only for Houdini but for other dccs.

BigRoy:
> After looking at the Rez link, I am thinking if we can make a customized plugin or edit the current one we have to find the executable plugin by patch version and use it for rendering. This is not only for Houdini but for other dccs. (Could be a later talk if it is too time-consuming)

As far as I know you don't need to alter the plug-in, just the Plugin configuration for Houdini.
If all we'd do is in the global job preload e.g. customize `PATH` so that the correct houdini version is accessible by doing `houdini.exe` then we could configure e.g. the Houdini plug-in (for all versions) to do:

image

Then it'll just take the houdini.exe that is found within the initialized environment.
We've actually used this before OpenPype exactly for Houdini in our previous pipeline - the pipeline would initialize the correct environment for the Deadline Worker (and thus provides the right accessible version of Houdini)

Doing this requires us to do NO code changes whatsoever to the Houdini Plug-in itself (just the plugin configuration itself).

---
An alternative would be to set an executable variable for e.g. the executable the application root then I suppose we could just set it to `$HOUDINI_EXECUTABLE` or `$HOUDINI_ROOT/hython.exe` but I'm not entirely sure if Houdini supports setting env vars in the executable paths. I suppose it might?

moonyuet:
> > Adding param can be messy if the users modify Houdini plugin
> 
> Do you mean that the param file list will be huge as there are many patch versions ? or will them change param file frequently ?

This is one of the scarenios I've concerned. But it's also about the possibility of the artist editing something wrong which can cause possible syntax errors and fails the deadline.

BigRoy:
Please consider [this approach](https://github.com/ynput/OpenPype/pull/5384#issuecomment-1668073944). It requires no changes to the Houdini plug-in - and actually doesn't require any change in OpenPype either if in your studio settings you set `PATH` for Houdini application environment as I described where `PATH` adds the Houdini bin folder for any launched Houdini.

If that works, then basically the behavior is already functional in OpenPype without any changes except for a minor studio configuration tweak and deadline settings tweak.

mustafa-zarkash:
> Please consider [this approach](https://github.com/ynput/OpenPype/pull/5384#issuecomment-1668073944).

I don't mind any solutions that work for us.
I wonder will that be explicit for pipeline admins or it will looks like magic ?

BigRoy:
Most admins will understand what the `PATH` environment variable is for. Any executable within the folders in `PATH` can be executed by just the executable's name instead of the full path.

Basically if no executable is found named `houdini` in the current working directory then the os will search in `PATH` for any matching executable name - if not found, only then the OS will tell you the executable is not found.

So what we're basically doing is set `PATH` to contain the specific Houdini version we want at the beginning so that it'll find that particular Houdini, then we tell Deadline to just launch `houdini` and thus do a similar search as the OS would do in `PATH`.

All OpenPype has to do is ensure that the application's environment (defined in studio settings) is set before Deadline searches for the executable's. Which I believe is what it's currently doing due to it running as a pre load script.

Does that make sense?

mustafa-zarkash:
I found out something because of #5420 
and I don't now if my PR will affect Mantra jobs or not. but I suspect it will. 
because `Mantra.py` resolves executables in the same way as `Houdini.py`

moonyuet:
> I found out something because of #5420 I don't now if my PR will affect Mantra jobs or not. but I suspect it will. because `Mantra.py` resolves executables in the same way as `Houdini.py`

I dont think the way we set the customized environment variable to the houdini in deadline would affect the default environment of Mantra in deadline. In deadline, mantra and houdini are in separated plugin configurations. 
image

If we want to also do the patch versions of the mantra, we can just do something similar to what we are gonna do with setting up houdini environment variable.

BigRoy:
> If we want to also do the patch versions of the mantra, we can just do something similar to what we are gonna do with setting up houdini environment variable.

Yes, the technique described [here](https://github.com/ynput/OpenPype/pull/5384#issuecomment-1668073944) should work for all executables. As long as the global job preload triggers to set the `PATH` environment variable accordingly.

mustafa-zarkash:
I making this comment as a checkpoint to summarize previous discussions

#### PR Objective
To submit Houdini version as `major.minor.patch` to deadline instead of the default behavior `major.minor`  

#### consequences of this request
Deadline doesn't support this by default and it requires some modifications on deadline side
the most easier solution from my perspective to update `DeadlineRepo\plugins\Houdini\Houdini.param` 
this modification can have backward compatibility with the default behavior by not deleting  the parts that expects `major.minor`

> I don't know if this PR will affect submitting Mantra jobs like in #5420

#### my suggestions to deal with these consequences
I added `Submit Patch Version` to deadline project settings which is disabled by default to
avoid applying any modifications. 

I created a mini guide about how to edit `Houdini.param` with sample code to copy and paste 
however I don't know how complicated it will be in production environment. 

#### Other solutions to consider 
This PR is not a work in progress, the code itself is working as expected *at least in my tests*
however, I'm not 100% sure that my proposed solution is what we need.

- Using Rez as BigRoy mentioned [here](https://github.com/ynput/OpenPype/pull/5384#issuecomment-1663755038)
- Making use of `PATH` env vars as BigRoy mentioned [here](https://github.com/ynput/OpenPype/pull/5384#issuecomment-1668073944)
    > I believe it won't conflict with this PR
    > By keeping `Submit Patch Version` disabled and everything will be as usual
- Submitting houdini executable path in job parameters + tweak `DeadlineRepo\plugins\Houdini\Houdini.py` to use this exe
    > This adds more work maintaing deadline houdini plugin as milan mentioned [here](https://github.com/ynput/OpenPype/pull/5384#issuecomment-1663034037)
- Making OP responsible for spawning houdini sessions instead of deadline. 
    > Deadline will spawn OP which spawns houdini (I think it will be quite annoying)
    
--- 
Regarding this suggestion: 

> - Making use of `PATH` env vars as BigRoy mentioned [here](https://github.com/ynput/OpenPype/pull/5384#issuecomment-1668073944)

OP settings only affects houdini sessions. these env vars aren't passed or set in deadline env. 
so, how would I tell deadline about `PATH` ?

> `$HOUDINI_ROOT/hython.exe` but I'm not entirely sure if Deadline supports setting env vars in the executable paths. 

This could be a way to tell deadline about houdini roots. 
Me too don't if it supported 

I found similar thing in an [article](https://jdbgraphics.nl/2021/10/28/deadline-submit-to-specific-houdini-version/) but it requires modifying `Houdini.py`

mustafa-zarkash:
> * Making use of `PATH` env vars as BigRoy mentioned [here](https://github.com/ynput/OpenPype/pull/5384#issuecomment-1668073944)

So, I gave this a shot, thanks @BigRoy for providing useful test hints.

The suggested steps above are: 
to modify OP system settings and deadline houdini config as follows
image
image

At first, deadline errored because it can't find houdini executable

--- 

Then, I made it work by adding houdini directory to my worker machine `PATH` system environment variable 

--- 

what about `GlobalJobPreLoad` ?

yes, `GlobalJobPreLoad` does inject some OP environment variables 
you can read something similar in Deadline log
```cmd
2023-08-10 19:56:31:  0: PYTHON: >>> Loading file ...
2023-08-10 19:56:31:  0: INFO: Setting Process Environment Variable AVALON_APP_NAME to houdini/19-5
2023-08-10 19:56:31:  0: INFO: Setting Process Environment Variable AVALON_ASSET to lamp
2023-08-10 19:56:31:  0: INFO: Setting Process Environment Variable AVALON_PROJECT to RnD
.
.
.
.
2023-08-10 19:56:31:  0: INFO: Setting Process Environment Variable OPENPYPE_WORKFILE_TOOL_ON_START to 0
2023-08-10 19:56:31:  0: INFO: Setting Process Environment Variable AVALON_LAST_WORKFILE to \\storage\work\pype_projects\RnD\Assets\Prop\lamp\work\Modeling\rnd_lamp_Modeling_v00000041.hip
2023-08-10 19:56:31:  0: PYTHON: >>> Removing temporary file
2023-08-10 19:56:31:  0: PYTHON: >> Injection end.
```

Strangely, I didn't find all environment variables injected!
for example, I have some test environment variables at different levels in OP settings 
- "GLOBAL_TEST_ENV" in  `general/environment` 
- "GENERIC_HOUDINI_TEST_ENV" in `applications/houdini/environment`
- "SPECIFIC_HOUDINI_TEST_ENV" and "HOUDINI_VERSION" in `applications/houdini/variants/19-5/environment` 
and I didn't find any of them injected! 
This made me wonder if the new "PATH" value was even injected! 
it can't be injected because the new value depend on "HOUDINI_VERSION" which was not injected either 

Some tests: 
This one I used the full executable path in deadline houdini config 
```cmd
2023-08-10 19:46:38:  0: INFO: Executable: "C:\Program Files\Side Effects Software\Houdini 19.5.403\bin\hython.exe"
```
This one I used `hython.exe` and `hython` in deadline houdini config however I added houdini directory to my windows system path 
```cmd
2023-08-10 19:56:31:  0: INFO: Executable: "C:\Program Files\Side Effects Software\Houdini 19.5.403\bin\hython.exe"
```
This one I only used `hython.exe` and `hython` in deadline houdini config
deadline errored and  printed that `self.GetRenderExecutable` can't find the executable
```cmd
=======================================================
Error
=======================================================
Error: FailRenderException : Deadline was unable to find Houdini 19_5 on the machine My-Machine-Name.

It tried and failed to find the render executable at these configured locations:
hython.exe
hython
```
---
#### conclusion
> I think  `GetRenderExecutable` from deadline Houdini plugin returns static full executable path by searching the worker machine (it searches in the same way as if you were using cmd)

image

My deduction, It won't be affected by any "PATH" modifications because it searches for executable before injection, 
so it worked when it was already there in my system variables.
Also note that OP didn't update path with Houdini executable that was added in OP houdini settings.

BigRoy:
I will prototype this tomorrow and report back. I'm basically 99% sure we have used this in the past without issue. Also since it's also a described use case in Deadline's own Rez plugin documentation I really feel this should work as intended.

mustafa-zarkash:
I'm wondering now 

1. Does this mean that any application environment variables (like mine above) won't be set in the deadline job ? 
2. Will your proposed solution work with other software e.g. maya or nuke ? 

> With that - I suspect it'll launch houdini's hython by just the hython name from the Houdini plugin in deadline.
3. Does the `GlobalJobPreLoad.py` run before `GetRenderExecutable` in houdini plugin ? 
I think the execution order is like this: 
    1. `GetRenderExecutable` which I think returns the full path of the executable not just `hython`  
    2. Run `GlobalJobPreLoad` which update `PATH` after few modifications 
    3. Run Executable returned by `GetRenderExecutable`

BigRoy:
> 1. Does this mean that any application environment variables (like mine above) won't be set in the deadline job ?

Hmm, not really. Sort of. Usually a process started from the plugin might inherit the environment from the Deadline plugin too - but it's not ALWAYS the case because a plugin can disable that behavior for processes it executes. That's why if you basically want Hython to start with the env vars the `DeadlinePlugin.SetProcessEnvironmentVariable` is the recommended approach because it's about explicitly passing it to the processes run from that Deadline plugin.

That's why I recommended the approach to keep the original behavior (still use `DeadlinePlugin.SetProcessEnvironmentVariable`) but only add the call to also set `PATH` for the deadline job plug-in so it can find the executables as we need by adding:

```python
        if "PATH" in contents:
            deadlinePlugin.SetEnvironmentVariable("PATH", contents["PATH"])
```
Which I believe is quite similar to:
```python
        if "PATH" in contents:
            os.environ["PATH"] = contents["PATH"]
```
Except that `deadlinePlugin.SetEnvironmentVariable` might not change the environment for the global job pre load script's environment.

> 2. Will your proposed solution work with other software e.g. maya or nuke ?

Yes, I'm quite certain it will.

> 3. Does the GlobalJobPreLoad.py run before GetRenderExecutable in houdini plugin ?

Yes, GlobalJobPreLoad runs **before** the render plugin for the job starts.

> GetRenderExecutable which I think returns the full path of the executable not just hython

It does return the full path to `hython`, but because it searches in `PATH` for that executable - and we've prepended it to `PATH` it should be the first entry it finds and thus return the one we want.

moonyuet:
@BigRoy , I did a check for the env by adding the test script(the script I used  for checking the remote_publish but it still checks the sys.path) into `openpype/scripts` and add the script into the preJobScripts as job info parameter in Houdini deadline submission script. 
It seems that it doesn't include any houdini environment variable in {PATH}
```
2023-08-11 22:18:29,584 - Deadline_Feedback_Log - DEBUG - (<class 'ModuleNotFoundError'>, ModuleNotFoundError("No module named 'remote_publish'"), <traceback object at 0x00000224C88FA308>)
2023-08-11 22:18:29,584 - Deadline_Feedback_Log - DEBUG - ['C:\\Users\\Kayla\\AppData\\Local\\Thinkbox\\Deadline10\\pythonAPIs\\2022-07-12T182241.0000000Z', 'C:\\Program Files\\Thinkbox\\Deadline10\\bin\\python3', 'C:\\Program Files\\Thinkbox\\Deadline10\\bin\\python3\\lib', 'C:\\Program Files\\Thinkbox\\Deadline10\\bin\\python3\\lib\\site-packages', 'C:\\Users\\Kayla\\Texture_Importer\\04_ui', 'C:\\Program Files\\Thinkbox\\Deadline10\\bin\\python3\\python37.zip', 'C:\\Program Files\\Thinkbox\\Deadline10\\bin', 'C:\\Program Files\\Thinkbox\\Deadline10\\bin\\python3\\Lib', 'C:\\Program Files\\Thinkbox\\Deadline10\\bin\\python3\\DLLs', 'C:\\Program Files\\Thinkbox\\Deadline10\\bin\\python3\\DLLs\\Release', 'C:\\Program Files\\Thinkbox\\Deadline10\\bin\\python3\\Lib\\plat-win32', 'C:\\Program Files\\Thinkbox\\Deadline10\\bin\\python3\\Lib\\lib-tk', 'C:\\Program Files\\Thinkbox\\Deadline10\\bin\\']
```
The script I used.

```
try:
    import remote_publish
    remote_publish.remote_publish()

except Exception:
    import sys

    import logging
    import logging.handlers

    filename = "D:/deadline.log"

    logger = logging.getLogger()
    logger.setLevel(logging.DEBUG)
    formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")

    handler = logging.handlers.RotatingFileHandler(filename, mode = 'w', backupCount = 5)
    handler.setLevel(logging.DEBUG)
    handler.setFormatter(formatter)
    logger.addHandler(handler)

    ch = logging.StreamHandler()
    ch.setLevel(logging.DEBUG)
    ch.setFormatter(formatter)
    logger.addHandler(ch)
    exc_info = sys.exc_info()

    logger = logging.getLogger('Deadline_Feedback_Log')
    logger.debug(exc_info)
    logger.debug(sys.path)

    sys.exit(1)

```
Maybe the env can be set up by adding the script which sets up the environment variable for {PATH} as PreJobScript in Houdini deadline submission. It will do the job right before all the rendering actions. 
image
If it still doesn't work, we need to adopt the original plan first and create the issue to discuss the way to enhance the workflow.

BigRoy:
> It seems that it doesn't include any houdini environment variable in {PATH}

Note that `sys.path != os.environ["PATH"]`. It's closer to `os.environ["PYTHONPATH"]` but even that isn't entirely true; it's just that on Python launch I believe it adds `PYTHONPATH` to `sys.path` but changing `os.environ["PYTHONPATH"]` during the current session doesn't change what's available on `sys.path`.

Anyway, we're not looking to adjust `sys.path` but `os.environ["PATH"]` as that defines where executables are being searched.
In short, you're logging the wrong value.

moonyuet:
> > It seems that it doesn't include any houdini environment variable in {PATH}
> 
> Note that `sys.path != os.environ["PATH"]`. It's closer to `os.environ["PYTHONPATH"]` but even that isn't entirely true; it's just that on Python launch I believe it adds `PYTHONPATH` to `sys.path` but changing `os.environ["PYTHONPATH"]` during the current session doesn't change what's available on `sys.path`.
> 
i dont mean they are the same. maybe i should explain better, what i mean is like it seems that both sys path and os.getenv("path") (see mustafa's comment) dont store the hython directory coming from the OP settings.

> Anyway, we're not looking to adjust `sys.path` but `os.environ["PATH"]` as that defines where executables are being searched. In short, you're logging the wrong value.
> 
I can check with logging but i dont really think it is there.

For prejob script, I  mean we can use the prejobscript to set up the environment for setting up the path for hython exeuctable

mustafa-zarkash:
> > 2\. Will your proposed solution work with other software e.g. maya or nuke ?
> 
> Yes, I'm quite certain it will.

As long as this proposed solution will work with other DCCs, then we should to be discussed in different place like `Enhancement Request` in issues or ideas section in forums 

This feature request is Houdini specific, and I really don't know if  it's needed to have a general solution that works with other DCCs

> If it still doesn't work, we need to adopt the original plan first and create the issue to discuss the way to enhance the workflow.

I agree with Kayla we need to find a solution that works in the first place to satisfy the current need 
then we can discuss how to enhance it. 

> I'd say that if this works - then this PR should be closed because that means you can do the exact same thing without the need of this PR. Less code and tweaks to maintain and it's a sensible approach from an admin perspective and could eventually work the same way using e.g. Rez.

I think having the two options will be nice, which adds some flexibility as admins would be able to choose from two solutions 
- one is very smart "Updating `PATH`" (_it also works for all DCCs_)
- other one is very clear (_from my perspective as a beginner_) "Updating  param file and toggling a setting in UI" (_it only works in Houdini_)

BigRoy:
> i dont mean they are the same. maybe i should explain better, what i mean is like it seems that both sys path and os.getenv("path") (see mustafa's comment) dont store the hython directory coming from the related environment set in the OP settings.
> i dont think we can use it to access the destined hython executable.

Have you tried changing the OpenPype `GlobalJobPreLoad` to use `DeadlinePlugin.SetEnvironmentVariable` as opposed to `DeadlinePlugin.SetProcessEnvironmentVariable` for the PATH environment variables supplied by OpenPype.

If that works for you it's really the simplest use case.

1. Set the application's root `PATH` in applications env settings.
2. In Deadline set the executable to just the executable name instead of full path.

> * other one is very clear (_from my perspective as a beginner_) "Updating  param file and toggling a setting in UI" (_it only works in Houdini_)

I've seen many beginners trip over the Deadline Param files (and even breaking them hehe). We'd need to clearly document how to do it, but I'd say if we need to do that anyway the other more generalized way I described is just as clear and doesn't require us to maintain a custom Houdini params/plugin, etc.

MustafaJafar:
> Have you tried changing the OpenPype GlobalJobPreLoad to use DeadlinePlugin.SetEnvironmentVariable as opposed to DeadlinePlugin.SetProcessEnvironmentVariable for the PATH environment variables supplied by OpenPype.

I'm going to try this.

MustafaJafar:
Hey, @BigRoy 
I came back. 

### Setting PATH in OP settings
Here are my Openpype settings, I printed `PATH` **inside Houdini console** and `PATH` was passed correctly.
image

```
>>> pprint(os.getenv("PATH").split(";"))
['C:\\PROGRA~1\\SIDEEF~1\\HOUDIN~1.403\\python39\\lib\\site-packages-forced\\PySide2',
 'C:/PROGRA~1/SIDEEF~1/HOUDIN~1.403/python39',
 'C:/Users/MUSTAF~1/DOCUME~1/houdini19.5/packages/HoudiniSubmitter/otls',
 '&',
 '&',
 'C:/PROGRA~1/SIDEEF~1/HOUDIN~1.403/bin',
 'C:/PROGRA~1/SIDEEF~1/HOUDIN~1.403/toolkit/bin',
 'C:/Program Files/Side Effects Software/Houdini 19.5.403/bin',
 'E:/Ynput/OpenPype/vendor/python/PySide2',
 'E:/Ynput/OpenPype/.venv/Scripts',
 'C:/Program Files (x86)/Common Files/Oracle/Java/javapath',
 'C:/ProgramData/Oracle/Java/javapath',
 'C:/Python27/Scripts',
 'C:/Python27',
 'C:/Program Files (x86)/Common Files/Intel/Shared '
 'Libraries/redist/intel64_win/compiler',
 'C:/Program Files (x86)/Common Files/Intel/Shared Files/cpp/bin/Intel64',
 'C:/WINDOWS/system32',
 'C:/WINDOWS',
 'C:/WINDOWS/System32/Wbem',
 'C:/WINDOWS/System32/WindowsPowerShell/v1.0',
 'C:/WINDOWS/System32/OpenSSH',
 'C:/Program Files/NVIDIA Corporation/NVIDIA NvDLISR',
 'C:/Program Files/Common Files/Autodesk Shared',
 'C:/Program Files/Microsoft SQL Server/120/Tools/Binn',
 'C:/WINDOWS/system32',
 'C:/WINDOWS',
 'C:/WINDOWS/System32/Wbem',
 'C:/WINDOWS/System32/WindowsPowerShell/v1.0',
 'C:/WINDOWS/System32/OpenSSH',
 'C:/Program Files/PuTTY',
 'C:/Program Files (x86)/Windows Kits/8.1/Windows Performance Toolkit',
 'C:/Program Files (x86)/IVI Foundation/VISA/WinNT/Bin',
 'C:/Program Files (x86)/NVIDIA Corporation/PhysX/Common',
 'C:/Program Files/Git/cmd',
 'C:/Program Files/CodeBlocks/MinGW/bin',
 'C:/Program Files/Graphviz/bin',
 'C:/Users/Mustafa Taher/AppData/Local/Programs/Python/Python39/Scripts',
 'C:/Users/Mustafa Taher/AppData/Local/Programs/Python/Python39',
 'C:/Users/Mustafa Taher/AppData/Local/Programs/Python/Python37/Scripts',
 'C:/Users/Mustafa Taher/AppData/Local/Programs/Python/Python37',
 'C:/Users/Mustafa Taher/AppData/Local/Microsoft/WindowsApps',
 'C:/Users/Mustafa Taher/AppData/Local/GitHubDesktop/bin',
 'C:/Users/Mustafa Taher/AppData/Local/Microsoft/WindowsApps',
 'C:/Program Files/SDCC/bin',
 'C:/Program Files/Adobe/Adobe After Effects CC 2017/Support Files',
 'C:/Users/Mustafa Taher/AppData/Local/Programs/Microsoft VS Code/bin',
 'C:/Program Files (x86)/Inno Setup 6',
 'E:/Ynput/OpenPype/.poetry/bin']
```

### Setting Deadline param file 
image

### Modifying `GlobalJobPreLoad.py`
I added these lines just after [these other line](https://github.com/ynput/OpenPype/blob/develop/openpype/modules/deadline/repository/custom/plugins/GlobalJobPreLoad.py#L385-L386)
```
if "PATH" in contents:
      PATH = contents["PATH"]
      print(f">>> Contents of 'PATH': {PATH}")
      deadlinePlugin.SetEnvironmentVariable("PATH", contents["PATH"])
```

### Job Results
```
Error: FailRenderException : Deadline was unable to find Houdini 19_5 on the machine xxxx .

It tried and failed to find the render executable at these configured locations:
hython.exe
hython
.
.
.
2023-09-12 11:05:20:  0: PYTHON: >>> Contents of 'PATH': 
C:\Program Files (x86)\OpenPype\3.16.0\lib;
C:\Program Files\Thinkbox\Deadline10\bin\python3\Library\bin;
C:\Program Files\Thinkbox\Deadline10\bin\python3\Library\bin;
C:\Program Files\Thinkbox\Deadline10\bin\;
C:\Program Files (x86)\Common Files\Oracle\Java\javapath;
C:\ProgramData\Oracle\Java\javapath;
C:\Python27\Scripts;
C:\Python27;
C:\Program Files (x86)\Common Files\Intel\Shared Libraries\redist\intel64_win\compiler;
C:\Program Files (x86)\Common Files\Intel\Shared Files\cpp\bin\Intel64;
C:\WINDOWS\system32;
C:\WINDOWS;
C:\WINDOWS\System32\Wbem;
C:\WINDOWS\System32\WindowsPowerShell\v1.0\;
C:\WINDOWS\System32\OpenSSH\;
C:\Program Files\NVIDIA Corporation\NVIDIA NvDLISR;
C:\Program Files\Common Files\Autodesk Shared\;
C:\Program Files\Microsoft SQL Server\120\Tools\Binn\;
C:\WINDOWS\system32;
C:\WINDOWS;
C:\WINDOWS\System32\Wbem;
C:\WINDOWS\System32\WindowsPowerShell\v1.0\;
C:\WINDOWS\System32\OpenSSH\;
C:\Program Files\PuTTY\;
C:\Program Files (x86)\Windows Kits\8.1\Windows Performance Toolkit\;
C:\Program Files (x86)\IVI Foundation\VISA\WinNT\Bin;
C:\Program Files (x86)\NVIDIA Corporation\PhysX\Common;
C:\Program Files\Git\cmd;
C:\Program Files\CodeBlocks\MinGW\bin;
C:\Program Files\Graphviz\bin;
;
C:\Users\Mustafa Taher\AppData\Local\Programs\Python\Python39\Scripts\;
C:\Users\Mustafa Taher\AppData\Local\Programs\Python\Python39\;
C:\Users\Mustafa Taher\AppData\Local\Programs\Python\Python37\Scripts\;
C:\Users\Mustafa Taher\AppData\Local\Programs\Python\Python37\;
C:\Users\Mustafa Taher\AppData\Local\Microsoft\WindowsApps;
C:\Users\Mustafa Taher\AppData\Local\GitHubDesktop\bin;
C:\Users\Mustafa Taher\AppData\Local\Microsoft\WindowsApps;
C:\Program Files\SDCC\bin;
C:\Program Files\Adobe\Adobe After Effects CC 2017\Support Files;
C:\Users\Mustafa Taher\AppData\Local\Programs\Microsoft VS Code\bin;
C:\Program Files (x86)\Inno Setup 6
```

### Conclusion 
1. Injection has no effect because it happens after resolving executable path by deadline.
2. Environment variables defined in OP `System Settings/Applications/Houdini` are not included in Job! 
    > Check Deadline log doesn't include PATH value that was previously defined in OP settings

For this to work I think we'd need to: 
1. Inject `PATH` before resolving executable path by deadline.
2. Include OP Applications Environment variables before submission.

BigRoy:
Your second printed log shows the issue. The environment you have there in `content` does not seem to include the app environment (or I am blind?).

I don't see the Houdini folder there?

> Include OP Applications Environment variables before submission.

No. These must be computed at runtime on the farm machine to be cross os compatible, etc.

> Inject PATH before resolving executable path by deadline.

GlobalJobPreLoad runs before the render plugin so should be on time.

MustafaJafar:
> Your second printed log shows the issue. The environment you have there in content does not seem to include the app environment

yes it does not include the app environment
any ideas about how to include app environment in `content` ? 
I'll try to dig in the code.

BigRoy:
> yes it does not include the app environment
> any ideas about how to include app environment in content ?

That'd mean the `extractenvironments` cli command doesn't return it - why it doesn't I'm not sure.
It does use `envgroup` "farm" but what that means I have no idea ;)

Probably @iLLiCiTiT knows why it might exclude the Application Settings' "environment"?

MustafaJafar:
> That'd mean the `extractenvironments` cli command doesn't return it - why it doesn't I'm not sure.

I followed that thread and found it eventually calls `get_app_environments_for_context`

so I gave it a test in houdini 
```python
import os 
from openpype.lib.applications import (
    get_app_environments_for_context,
    LaunchTypes,
)
        
project = os.getenv("AVALON_PROJECT")
asset = os.getenv("AVALON_ASSET")
task = os.getenv("AVALON_TASK")
app = os.getenv("AVALON_APP_NAME")
env_group = "farm" # "standard" 

env = get_app_environments_for_context(
    project,
    asset,
    task,
    app,
    env_group=env_group,
    launch_type=LaunchTypes.farm_render
)
print(env)
```

and I could find in the results
```
 'PATH': 'C:/Program Files/Side Effects Software/Houdini 19.5.403/bin;......
```

Also, I print `content` in deadline log and `content` hasn't the same size as the above env (although I'm calling the same function. maybe I'm missing something)
image

BigRoy:
I believe the function also returns your local environment (it includes it) and thus if you already run it within Houdini it'd include any env vars already set there - that's also why you're seeing the `_HOUDINI_...` variables.

BigRoy:
I suspect this will be a question for @iLLiCiTiT

MustafaJafar:
> I suspect this will be a question for @iLLiCiTiT

@BigRoy 
we just figured it out :) 
deadline was running another openpype version that doesn't have the path settings. 
image


and after fixing that, `PATH` was updated successfully! 

Then, I added these lines just after [these other line](https://github.com/ynput/OpenPype/blob/develop/openpype/modules/deadline/repository/custom/plugins/GlobalJobPreLoad.py#L385-L386)
```
if "PATH" in contents: # 
    PATH = contents["PATH"]
    print(f">>> Contents of 'PATH': {PATH}")
    deadlinePlugin.SetEnvironmentVariable("PATH", contents["PATH"])
```
which prints fine! 
image

but deadline still fails 
image

BigRoy:
@MustafaJafar This is perfect - so close.

I've decided to do a test run on my end.

This I how I set the Houdini plugin settings in Deadline:
image

```
hython.exe
hython
```

Then I skipped OpenPype completely and just did a 'deadline test' and submitted a Houdini scene through the Deadline Monitor for that Houdini version. Next this is what I tried in `GlobalJobPreLoad.py`:
```
import os
HOUDINI_LOCATION = r"C:\Program Files\Side Effects Software\Houdini 19.5.435\bin"
path = os.pathsep.join([HOUDINI_LOCATION, os.environ["PATH"]])

print(f"Testing deadlinePlugin.SetEnvironmentVariable: {path}")
deadlinePlugin.SetEnvironmentVariable("PATH", path)
print(f"Testing deadlinePlugin.SetProcessEnvironmentVariable: {path}")
deadlinePlugin.SetProcessEnvironmentVariable("PATH", path)
# print(f"Setting os.environ['PATH']: {path}")
#os.environ["PATH"] = path
```
Which failed.

Then I tried:
```python
print(f"Testing deadlinePlugin.SetProcessEnvironmentVariable: {path}")
deadlinePlugin.SetProcessEnvironmentVariable("PATH", path)
```
Which failed.

Then I tried:
```python

```
Which worked. ✅ 

So the solution is to just add it to `os.environ["PATH"]` in the GlobalJobPreLoad.

The logs showed:
```
2023-09-14 20:13:45:  0: PYTHON: Setting os.environ['PATH']: C:\Program Files\Side Effects Software\Houdini 19.5.435\bin;...
...
...
2023-09-14 20:13:45:  0: INFO: Executable: "C:\Program Files\Side Effects Software\Houdini 19.5.435\bin\hython.exe"
```

Could you try that @MustafaJafar

BigRoy:
It's good to note that Deadline has quite a few settings involving "environments" of the jobs. I'm not sure how any of these influence the behavior:

**Sandbox Plugins - ** it might be that this alters the behavior when disabled?
image

**Render as user** - I'm not sure if that'd similarly inherit the environment. I suspect it should and that it runs even the deadline plugin as that user from the get go.
image

MustafaJafar:
@BigRoy 
MAN! you are a genius! it worked! 
image

MustafaJafar:
This PR is closed in favor of https://github.com/ynput/OpenPype/pull/5622

PR #5386, opened 2023-08-02
Bugfix: Houdini abc validator error message

mustafa-zarkash:
## Changelog Description

When ABC path validator fails, it prints node objects not node paths or names

This bug happened because of updating `get_invalid` method to return nodes instead of node paths 

Before
image

After 
image

## Additional Info
`get_invalid` method should return a list of nodes objects because this is what `get invalid action` expects. 

## Testing notes:
1. create a geo without path attr
1. create alembic pointcache instance
2. check error message

### DIFF
--- a/openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
+++ b/openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
@@ -32,8 +32,9 @@ class ValidatePrimitiveHierarchyPaths(pyblish.api.InstancePlugin):
     def process(self, instance):
         invalid = self.get_invalid(instance)
         if invalid:
+            nodes = [n.path() for n in invalid]
             raise PublishValidationError(
-                "See log for details. " "Invalid nodes: {0}".format(invalid),
+                "See log for details. " "Invalid nodes: {0}".format(nodes),
                 title=self.label
             )
 


openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py
```diff
@@ -32,8 +32,9 @@ class ValidatePrimitiveHierarchyPaths(pyblish.api.InstancePlugin):
     def process(self, instance):
         invalid = self.get_invalid(instance)
         if invalid:
+            nodes = [n.path() for n in invalid]
             raise PublishValidationError(
-                "See log for details. " "Invalid nodes: {0}".format(invalid),
+                "See log for details. " "Invalid nodes: {0}".format(nodes),
                 title=self.label
             )
```
BigRoy, line 39: How about even `",".join(node.path() for node in invalid)` so it's `/obj/geo1, /obj/geo2`, etc. Or do we prefer the python list output formatting?

mustafa-zarkash: > do we prefer the python list output formatting?

I can't tell whether it's preferred or not.
but I can say that the convention used in all validations is to rely on `get_invalid` method which should return a list of node objects, this convention was adopted because of `SelectInvalidAction` .

https://github.com/ynput/OpenPype/blob/1b07686daf908f17f960012df0f528629bed4644/openpype/hosts/houdini/api/action.py#L27

what really `get_invalid` returns in almost all validators is a list of nodes paths (which is not compatible with `SelectInvalidAction`  and it doesn't cause an issue because most scripts never use that action)

---
So, for consistency I chose list output formatting in this PR

these scripts uses `get_invalid` and log to user 
Most of them log a list 

https://github.com/ynput/OpenPype/blob/1b07686daf908f17f960012df0f528629bed4644/openpype/hosts/houdini/plugins/publish/validate_animation_settings.py#L31-L34

https://github.com/ynput/OpenPype/blob/1b07686daf908f17f960012df0f528629bed4644/openpype/hosts/houdini/plugins/publish/validate_bypass.py#L30-L33

https://github.com/ynput/OpenPype/blob/1b07686daf908f17f960012df0f528629bed4644/openpype/hosts/houdini/plugins/publish/validate_cop_output_node.py#L29-L30

https://github.com/ynput/OpenPype/blob/1b07686daf908f17f960012df0f528629bed4644/openpype/hosts/houdini/plugins/publish/validate_file_extension.py#L36

https://github.com/ynput/OpenPype/blob/1b07686daf908f17f960012df0f528629bed4644/openpype/hosts/houdini/plugins/publish/validate_frame_token.py#L31-L34

https://github.com/ynput/OpenPype/blob/1b07686daf908f17f960012df0f528629bed4644/openpype/hosts/houdini/plugins/publish/validate_mkpaths_toggled.py#L19-L20

https://github.com/ynput/OpenPype/blob/1b07686daf908f17f960012df0f528629bed4644/openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py#L36

https://github.com/ynput/OpenPype/blob/1b07686daf908f17f960012df0f528629bed4644/openpype/hosts/houdini/plugins/publish/validate_usd_output_node.py#L26-L27

BigRoy: Consistency is good - was just wondering what an artist might prefer more. Anyway, this PR is already a massive improvement I think. ❤️

moonyuet (approved):
tested in Houdini and looks good.

PR #5387, opened 2023-08-02
Applications: Use prelaunch hooks to extract environments

iLLiCiTiT:
## Changelog Description
Environment variable preparation is based on prelaunch hooks. This should allow to pass OCIO environment variables to farm jobs.

## Additional info
To be able allow this option it was required to make `ApplicationLaunchContext` less strict about variables passed in. It was introduced new filtering option "launch type" which can be used for filtering of launch hooks. All launch hooks that should run only when application is launched locally have already set filter to be executed only locally. It is possible to create launch context and execute prelaunch hooks without launching the subprocess.

## Testing notes:
1. Applications launch should set all environments as expected (from launcher or ftrack actions)
    - last workfile
    - open workfiles tool on start
    - ocio envs
2. It should be possible to receive needed environment variables on farm for publish and render jobs with OCIO environment variables etc.
3. Remote publishing should work as expected.

### DIFF
--- a/openpype/lib/applications.py
+++ b/openpype/lib/applications.py
@@ -931,8 +931,11 @@ class ApplicationLaunchContext:
 
         self.data = dict(data)
 
+        launch_args = []
+        if executable is not None:
+            launch_args = executable.as_args()
         # subprocess.Popen launch arguments (first argument in constructor)
-        self.launch_args = executable.as_args()
+        self.launch_args = launch_args
         self.launch_args.extend(application.arguments)
         if self.data.get("app_args"):
             self.launch_args.extend(self.data.pop("app_args"))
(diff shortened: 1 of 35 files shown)


openpype/lib/applications.py
```diff
@@ -736,13 +779,15 @@ class LaunchHook:
...
-    platforms = []
+    hosts = set()
+    # Set of application groups
+    app_groups = set()
+    # Set of specific application names
+    app_names = set()
+    # Set of platform availability
+    platforms = set()
+    # Set of launch types for which is available
+    launch_types = set()
```
BigRoy, line 790: What if we default this to `{LaunchTypes.local}` for  backwards compatibility, also for hooks outside of OpenPype codebase currently. It'd also require much less changes for this PR across much less files?

jakubjezek001 (approved):
LGTM! was working

PR #5393, opened 2023-08-02
Unreal: Fix loading hero version for static and skeletal meshes

simonebarbieri:
## Changelog Description
Fixed a problem with loading hero versions for static ans skeletal meshes.

## Additional info
Static ans skeletal meshes use the version number in their name to be able to be loaded in different versions for different scenes. If trying to load a hero version for these assets, there was an error when trying to retrieve the version number.
Also added the version number to loaded static meshes from `fbx` as well.

## Testing notes:
1. Launch unreal and try to load as hero version any static or skeletal mesh from `abc` or `fbx`.


mkolar (changes requested):
Technically this fixes loading of the hero versions, but breaks loading of normal ones.
```
During load error happened on Subset: "modelHires" Representation: "abc" Version: 2

Error message: unsupported format string passed to dict.__format__

  File "C:\Users\milan\AppData\Local\Ynput\AYON\addons\openpype_3.16.3-nightly.3\openpype\tools\loader\widgets.py", line 1592, in _load_representations_by_loader
    load_with_repre_context(
  File "C:\Users\milan\AppData\Local\Ynput\AYON\addons\openpype_3.16.3-nightly.3\openpype\pipeline\load\utils.py", line 323, in load_with_repre_context
  File "C:\Users\milan\AppData\Local\Ynput\AYON\addons\openpype_3.16.3-nightly.3\openpype\hosts\unreal\plugins\load\load_staticmesh_abc.py", line 87, in load
    name_version = f"{name}_v{version:03d}"
TypeError: unsupported format string passed to dict.__format__
```

moonyuet:
I have tested some static meshes from Max, Static Mesh loaded as expected.
image
it would name the things accordingly depending on whether the "hero version" exists

LiborBatek (approved):
Both assets UE ```staticMesh``` and ```skeletalMesh``` works fine and user is able to load ```hero version``` of them.

All working fine.

mkolar (approved):

PR #5394, opened 2023-08-02
Houdini: delete redundant bgeo sop validator

mustafa-zarkash:
## Changelog Description

I found out that this `Validate BGEO SOP Path` validator is redundant, it catches two cases that are already implemented in "Validate Output Node". 
"Validate Output Node" works with `bgeo` as well as `abc` because `"pointcache"` is listed in its families 

image

## Additional Info 
"collect_output_node.py" will save output node of the bgeo rop in instance data i.e. `instance.data["output_node"]` 
and "validate_sop_output_node.py" will check the node saved in `instance.data["output_node"]` 

## Testing notes:
1. craete bgeo instance 
2. try publish with different "SOP Path" values
    - empty sop path
    - set sop path to any non sop node
3. "Validate Output Node" should error with these both cases


mkolar (approved):
good catch

PR #5396, opened 2023-08-03
Max: Use the empty modifiers in container to store AYON Parameter

moonyuet:
## Changelog Description
Instead of adding AYON/OP Parameter along with other attributes inside the container, empty modifiers would be created to store AYON/OP custom attributes

## Additional info
This PR is also affecting #5378, which loads the OP Parameters when the objects are loaded.
Old instances are definitely errored out when publishing.

## Testing notes:
1. Launch Max via AYON/OP
2. Create new instance (Camera, Model)
3. The new modifier named "OP Data" created and store the OP Parameters
image
4. Add some content to be published
5. Publish


moonyuet:
> Be carefull, you've changed unreal integration submodule.

yes saw it just now.

LiborBatek (approved):
Works as expected.
image

openpype/hosts/max/api/plugin.py
```diff
@@ -185,7 +185,10 @@ def create_instance_node(node):
             node = rt.Container(name=node)
 
         attrs = rt.Execute(MS_CUSTOM_ATTRIB)
-        rt.custAttributes.add(node.baseObject, attrs)
+        modifier = rt.EmptyModifier()
+        rt.addModifier(node, modifier)
+        node.modifiers[0].name = "OP Data"
```
iLLiCiTiT, line 190: I have question regarging this. Could be `node.modifiers[0]` replaced by `modified`? I do expect it is the same object, but accessing it through index is 1. strange, 2. potential danger if there is different modifier.

PR #5401, opened 2023-08-03
AYON: 3dsMax settings

iLLiCiTiT:
## Changelog Description
Keep `adsk_3dsmax` group in applications settings.

## Additional info
Because 3dsmax group in AYON is `adsk_3dsmax` but we're converting it to `3dsmax` it does not match and find available application in `ApplicationManage`. Because the group changed it was required to add `adsk_3dsmax` to launch hooks which had `3dsmax`.

## Testing notes:
1. 3dsMax can be launched in AYON
2. The prelaunch hooks are executed in both AYON and OpenPype modes

### DIFF
--- a/openpype/hooks/pre_add_last_workfile_arg.py
+++ b/openpype/hooks/pre_add_last_workfile_arg.py
@@ -14,7 +14,7 @@ class AddLastWorkfileToLaunchArgs(PreLaunchHook):
     # Execute after workfile template copy
     order = 10
     app_groups = {
-        "3dsmax",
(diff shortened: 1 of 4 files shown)


moonyuet:
image
It still doesn't work for some reasons

iLLiCiTiT:
You would have to create new openpype package, upload it to server, restart server and remove current openpype addon from local app data (`%LOCALAPPDATA%/Ynput/AYON/addons/openpype{version}`).

moonyuet (approved):
looks good now. thank you!
image

PR #5402, opened 2023-08-03
Maya: do not create top level group on reference

kalisp:
## Changelog Description
This PR allows to not wrapping loaded referenced assets in top level group either explicitly for artist or by configuration in Settings.
Artists can control group creation in ReferenceLoader options.
Default no group creation could be set by emptying `Group Name` in `project_settings/maya/load/reference_loader`

## Testing notes:
1. try to uncheck `Group imported assets` in Reference Loader options
AnyDeskMSI_XPqOsgRKGD
3. no group should be created
4. remove group name in `project_settings/maya/load/reference_loader` >> no need to go to Options, still no group will be created
V9nxVMBye8

AnyDeskMSI_mo3PDISCHN


openpype/hosts/maya/plugins/load/load_reference.py
```diff
@@ -205,6 +209,12 @@ def process_reference(self, context, name, namespace, options):
                 self._post_process_rig(name, namespace, context, options)
             else:
                 if "translate" in options:
+                    if not attach_to_root and new_nodes:
+                        reference_node = get_reference_node(new_nodes)
+                        for node in new_nodes:
+                            if node != reference_node:
+                                break
+                        group_name = node.lstrip("|")
```
BigRoy, line 217: This is not the way. Without the leading `|` it'll allow to match ANY node name with that name, and there can be clashes. Technically the behavior we're using is that it contains a namespaces at the very least, and we uniqify the namespace somewhere earlier in the code so it's basically a zero percent chance, but still it's odd logic doing it this way.

I feel like we're better off just getting the top level nodes from the returned nodes. I did a quick test in Maya 2024 here but it includes the group name as well if the group is generated with `groupReference=True` and `groupName="whatever"`.

Since we know that basically directly after the referencing the loaded content is NOT parented to anything else (the reference or its group is top-level maya node) we know we can just get the group name via:

```python
root_nodes = cmds.ls(new_nodes, assemblies=True, long=True)
group_name = root_nodes[0]  # we assume only a single root is ever loaded
```

As such, from the beginning `group_name` can always be retrieved like that and doesn't need altering behavior between the if else branches.
 
Like with the current logic we assume only a single root is ever loaded, be aware that this isn't necessarily true if NO grouping was enabled for the reference and the loaded content, e.g. an Alembic reference actually has multiple top level nodes. This also means that if the case happens where translating like this will happen that only ONE of the root nodes will ever be translated (as opposed to all of them.)

LiborBatek (approved):
Now both scenarios are working ok. 
When just grouping is turned off during loading using Loader Action Menu and

when Group naming is left blank

image

Resulting into just Namespaces and not Groups present when assets loaded into scene
image
I did tests on multiple asset families too.
All working fine.

openpype/hosts/maya/plugins/load/load_reference.py
kalisp: Ok, makes sense.

openpype/hosts/maya/plugins/load/load_reference.py
```diff
@@ -123,6 +124,10 @@ def process_reference(self, context, name, namespace, options):
         attach_to_root = options.get("attach_to_root", True)
         group_name = options["group_name"]
 
+        # no group shall be created
+        if not attach_to_root:
+            group_name = namespace
```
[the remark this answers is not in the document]
BigRoy: We can also filter `cmds.ls` to only ever returning a single node to begin with, to speed it up:
```python
group_name = cmds.ls(new_nodes, assemblies=True, long=True, head=1)[0]
```
So I'd likely recommend adding `head=1`.

Note that the above logic would fail if a loaded reference would have NO dag nodes whatsoever (e.g. a look).

So we might want to do:
```python
group_name = next(iter(cmds.ls(new_nodes, assemblies=True, long=True, head=1)), None)
```

Or if you prefer this readability wise:
```python
roots = cmds.ls(new_nodes, assemblies=True, long=True, head=1)
group_name = roots[0] if roots else None
```

Note that any logic that would touch `group_name` would still need to check whether it's not None.

openpype/hosts/maya/plugins/load/load_reference.py
```diff
@@ -9,7 +9,8 @@
     maintained_selection,
     get_container_members,
     parent_nodes,
-    create_rig_animation_instance
+    create_rig_animation_instance,
+    get_reference_node
```
BigRoy, line 13: Import seems unused and can be removed.

PR #5403, opened 2023-08-03
AYON: Use correct appdirs

iLLiCiTiT:
## Changelog Description
Use 'get_ayon_appdirs' in thumbnail logic instead of having own implementation.

## Additional info
If the ayon appdirs changes it is hard to track where it must be changed.


iLLiCiTiT:
It is change in `ayon-python-api`

PR #5404, opened 2023-08-03
General: Navigation to Folder from Launcher

simonebarbieri:
## Changelog Description
Adds an action in launcher to open the directory of the asset.

## Testing notes:
1. Open the launcher and navigate to an asset.
2. Select *Open in File Browser*

### DIFF
--- a/openpype/plugins/actions/open_file_explorer.py
+++ b/openpype/plugins/actions/open_file_explorer.py
@@ -1,0 +1,43 @@
+import os
+
+from openpype.client import (
+    get_asset_by_name,
+)
+from openpype.pipeline import (
+    Anatomy,
+    LauncherAction,
+)
+from openpype.pipeline.template_data import (
+    get_asset_template_data,
+)
+
+class OpenTaskPath(LauncherAction):
+    name = "open_task_path"
+    label = "Open in File Browser"
+    icon = None
+    order = 500
+
+    def is_compatible(self, session):
+        """Return whether the action is compatible with the session"""
+        return "AVALON_TASK" in session
+
+    def process(self, session, **kwargs):
+        print(session)
+        project_name = session["AVALON_PROJECT"]
+        asset_name = session["AVALON_ASSET"]
+
+        asset = get_asset_by_name(project_name, asset_name)
+
+        anatomy = Anatomy(project_name)
+        roots = anatomy.roots
+
+        # We consider the first root to be the main one
+        root_key = list(roots.keys())[0]
+        root = roots.get(root_key)
+
+        template_data = get_asset_template_data(asset, project_name)
+        hierarchy = template_data.get("hierarchy")
+
+        path = f"{root}/{project_name}/{hierarchy}/{asset_name}"
+
+        os.system(f"start {path}")
--- a/openpype/plugins/actions/open_file_explorer.py
+++ b/openpype/plugins/actions/open_file_explorer.py
@@ -1,4 +1,5 @@
 import os
+import platform
 
 from openpype.client import (
     get_asset_by_name,
@@ -41,4 +42,9 @@ class OpenTaskPath(LauncherAction):
 
         path = f"{root}/{project_name}/{hierarchy}/{asset_name}"
 
-        os.system(f"start {path}")
+        if platform.system().lower() == "windows":
+            os.system(f"start {path}")
+        elif platform.system().lower() == "darwin":
+            os.system(f"open {path}")
+        elif platform.system().lower() == "linux":
+            os.system(f"xdg-open {path}")
--- a/openpype/plugins/actions/open_file_explorer.py
+++ b/openpype/plugins/actions/open_file_explorer.py
@@ -38,6 +38,7 @@ class OpenTaskPath(LauncherAction):
         anatomy = Anatomy(project_name)
         workdir = anatomy.templates_obj["work"]["folder"].format(data)
 
+        # Remove any potential unformatted parts of the path
         valid_workdir = workdir.split("{", 1)[0]
 
         # Path is not filled
--- a/openpype/plugins/actions/open_file_explorer.py
+++ b/openpype/plugins/actions/open_file_explorer.py
@@ -42,9 +42,14 @@ class OpenTaskPath(LauncherAction):
 
         path = f"{root}/{project_name}/{hierarchy}/{asset_name}"
 
-        if platform.system().lower() == "windows":
-            os.system(f"start {path}")
-        elif platform.system().lower() == "darwin":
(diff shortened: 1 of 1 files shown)


openpype/plugins/actions/open_file_explorer.py
```diff
@@ -0,0 +1,50 @@
...
+    name = "open_task_path"
+    label = "Open in File Browser"
+    icon = None
+    order = 500
+
+    def is_compatible(self, session):
+        """Return whether the action is compatible with the session"""
+        return "AVALON_ASSET" in session
+
+    def process(self, session, **kwargs):
```
iLLiCiTiT, line 26: I don't know the background, but I do expect it should require task name, and it should use workfilte templates instead of expecting that each project has same hardcoded template.

simonebarbieri:
@BigRoy that's fantastic, thank you so much 🙏

openpype/plugins/actions/open_file_explorer.py
```diff
@@ -0,0 +1,172 @@
...
+        elif platform_name == "linux":
+            args = ["xdg-open", path]
+        else:
+            raise RuntimeError(f"Unknown platform {platform.system()}")
+        # Make sure path is converted correctly for 'os.system'
+        os.system(subprocess.list2cmdline(args))
+
+    @staticmethod
+    def copy_path_to_clipboard(path):
+        from Qt import QtWidgets
```
BigRoy, line 163: Likely needs to be `qtpy` nowadays.
```suggestion
        from qtpy import QtWidgets
```

moonyuet:
it works as expected.
image

BigRoy:
> it works as expected. image

If you have the asset's task selected - I'd expect it to actually explore into the task folder instead of the folder above it? :)

simonebarbieri:
Should have fixed all the issues discussed :) Added the folder open icon, changed the label, and now it should navigate to the right folder if you have a task selected!

moonyuet (approved):
image
image

looks good for me

openpype/plugins/actions/open_file_explorer.py
```diff
@@ -0,0 +1,167 @@
...
+        # Get template key
+        # template_key = get_workfile_template_key_from_context(
+        #    asset_name, task_name, host_name, project_name, dbcon
+        # )
+        workdir = result["work"]["folder"]
+
+        # Keep only the part of the path that was formatted by splitting up to
+        # the first stub - we can only explore up to there
+        STUB = "<<NULL>>"
+        valid_dir = workdir.split(STUB, 1)[0]
```
iLLiCiTiT, line 54: This require `get_template_data` from `openpype.pipeline.template_data` and `get_workdir_data` can be removed.
```suggestion
        data = get_template_data(project_doc, asset_doc, task_name)

        anatomy = Anatomy(project_name)
        workdir = anatomy.templates["work"]["folder"].format(data)
        
        valid_dir = workdir.split("{", 1)[0]
```

openpype/plugins/actions/open_file_explorer.py
```diff
@@ -0,0 +1,167 @@
...
+        # If the path endswith `/work` and the path does not exist but the
+        # parent does then we allow to go to the parent because it could for
+        # example be the hierarchical folder for `asset` in e.g. `asset/hero`
+        if not os.path.exists(valid_dir) and valid_dir.endswith("/work/"):
+            # /folder/ dirname is /folder so we rstrip the last / to be sure
+            parent_dir = os.path.dirname(valid_dir.rstrip("/"))
+            if os.path.exists(parent_dir):
+                valid_dir = parent_dir
+
+        path = os.path.normpath(valid_dir)
```
iLLiCiTiT, line 65:
```suggestion
        # Path is not filled
        if not valid_path:
            return

        # Normalize
        valid_workdir = os.path.normpath(valid_workdir)
        while not os.path.exists(valid_workdir):
            prev_workdir = valid_workdir
            valid_workdir = os.path.dirname(prev_workdir)
            if not valid_workdir or valid_workdir == prev_workdir:
                return None

        path = valid_workdir 
```

BigRoy: Even though I understand the logic - it's not entirely accurate to the artist. This could mean that clicking "Explore here" on a task or asset that hasn't been initialized that it'll still "think" it found a path even if the project folder itself doesn't even exist on disk. 

With this expected path:
`/projects/my_project/s01_e01/ep01/sh010/work/modeling` 

This could even just explore to `/projects` without any warning which to me personally feels misleading - it's nowhere near close to what the user would expect.

iLLiCiTiT (approved):
LGTM

openpype/plugins/actions/open_file_explorer.py
```diff
@@ -29,40 +29,30 @@ def process(self, session, **kwargs):
...
-        if not os.path.exists(valid_dir) and valid_dir.endswith("/work/"):
-            # /folder/ dirname is /folder so we rstrip the last / to be sure
-            parent_dir = os.path.dirname(valid_dir.rstrip("/"))
-            if os.path.exists(parent_dir):
-                valid_dir = parent_dir
-
-        path = os.path.normpath(valid_dir)
+        workdir = anatomy.templates_obj["work"]["folder"].format(data)
+
+        valid_workdir = workdir.split("{", 1)[0]
```
BigRoy, line 41: This should have a comment that it's removing any potential unformatted parts of the path just for clarity sake.

openpype/plugins/actions/open_file_explorer.py
BigRoy: Marking this unresolved because I feel my last comment wasn't addressed?

iLLiCiTiT: Expecting `/work/` at the end is not expected either. Especially if is used only the part of template which is formatted (this is the first part which is clunky). Both don't make sense in the same way. Using `/work/` is more matching a specific work template.

BigRoy: Can we limit it so it never passes the first `{project}` variable? 

Not entirely what we'd want. We sometimes also use the button to just quickly see if there were already ANY files whatsoever -then clicking it would say "Folder does not exist" and you'd know what's up. Now you'd be thrown into the folder regardless and need to look where you are to see whether the folder actually existed.

image

Maybe we should allow it go up only to `{asset}`? (Even though technically a template could be: `{task}/{asset}`, maniacs!) Would it at least a sensible default?

iLLiCiTiT:
Made a change, that if a task is selected but path to task folder does not exist, it will try to find only path to asset and use in that case it.

PR #5405, opened 2023-08-03
Nuke: adding BBox knob type to settings

jakubjezek001:
## Changelog Description
Nuke knob types in settings having new `Box` type for reposition nodes like Crop or Reformat.

## Additional info
Also found some typos so fixed them too

## Testing notes:
1. all should work as it was but new knob type is added
2. Open Studio Settings on any project at `project_settings/nuke/create/CreateWriteRender/prenodes/Reformat01/knobs`
3. Try to add new knob and set the type to box
4. you should see this 
image


iLLiCiTiT (approved):
Code is looking good. Needs testing.

64qam (approved):

PR #5406, opened 2023-08-03
Max: Extract Max Scene Raw should include OP metadata

moonyuet:
## Changelog Description
Extract Max Scene Raw would publish the container with OP metadata. Once the user load the published objects, the mesh objects won't be parented to the container while it would store the OP data which links to the published objects.
## Additional info
Affecting #5396 and slightly affecting #5378
<del>Should be merged after #5378 </del>
There is a room for discussion on whether the old duplicates should be deleted. If the "delete old duplicates" implemented for  switching or updating version, the old container would only includes the objects which is not duplicated in the newer version.
Please create new instance to publish something for test the loader.
## Testing notes:
1. Publish something with MaxScene Instance with more than 1 version
2. Load the instance
3. The container should include OP parameter
4. Switch the version or update the version
5. It should load the duplicated objects with the new container which also stores the data


LiborBatek (approved):
First loading ```MaxScene``` asset into scene...
image

Updating version to latest with succesfully added content into OP Container to the object list:

image

LiborBatek:
Note: as already said, there is a question how to treat objects which not present in the previous version comparing the latest. Generally speaking about all object differing in versions.

Now when switching versions (in my case going from latest version to previous one) all object list in OP Container correctly get updated containing just the valid objects for the particular version but keeping them physically in the scene even tho they arent available in that particular version.

Going from latest ver (containing Column geomtries) to previous which lacks columns still keeps columns in the scene.
image

moonyuet:
> Note: as already said, there is a question how to treat objects which not present in the previous version comparing the latest. Generally speaking about all object differing in versions.
> 
> Now when switching versions (in my case going from latest version to previous one) all object list in OP Container correctly get updated containing just the valid objects for the particular version but keeping them physically in the scene even tho they arent available in that particular version.
> 
> Going from latest ver (containing Column geomtries) to previous which lacks columns still keeps columns in the scene. image

This one can be tested with the draft switching version PR https://github.com/ynput/OpenPype/pull/5424 after this has been merged(. (Not quite sure if the deletion of the object would affect the modifier's linkage or any hierarchies when upversioning or switching version but this is definitely something we need to resolve.)

moonyuet:
@LiborBatek This one needs one more test to ensure the new modifier parameter is in the maxSceneMain regarding to the update of the instance code. Thanks!

LiborBatek (approved):
Seems ok to me.

image

openpype/hosts/max/plugins/publish/extract_max_scene_raw.py
```diff
@@ -33,6 +33,7 @@ def process(self, instance):
             instance.data["representations"] = []
 
         nodes = instance.data["members"]
+        nodes.append(container)
```
iLLiCiTiT, line 36: This will affect content of `instance.data["members"]` permanently (any future access to that key will contain the added node too).

NOTE: I'm asking if it's ok, I don't know if it's expected behavior.

moonyuet: you actually raise a good point. it affect content of the instance.data["members"] if there is multiple extractor working.

moonyuet:
I will close this one as I implement the similar logic in load MaxScene at #5378

PR #5407, opened 2023-08-03
Bugfix: update defaults to default_variants in maya and houdini OP DCC settings

mustafa-zarkash:
## Changelog Description
On moving out to new creator in Maya and Houdini 
updating settings was missed. 

## Additional Info1
I couldn't update all `defaults` variables in  `houdini.json` and `maya.json` files because they don't exist in the `schema_houdini_create.json` and  `schema_maya_create.json` respectively. 
so, I only modified what can be found in the schema create json files 

## Additional Info2: collision 
There's a collision between 
- `default_variants` values in create plugins 
- `default_variants` values in default settings (will override)

https://github.com/ynput/OpenPype/blob/c9cf6646f78223f492f21b8097f2b73ae6df6b64/openpype/hosts/maya/plugins/create/create_model.py#L15

> this code from this branch 

https://github.com/ynput/OpenPype/blob/61e45b73aed8736ff9dbc68e5a88e6e1cbbdb908/openpype/settings/defaults/project_settings/maya.json#L570-L574

>  default settings values will always override create plugins values, 

## Additional Info3: fix collision 

I think this behavior could be fixed by another PR that modifies `apply_settings` 
which would check if this attribute exists or not 
and another check that decides to override or append for certain types. 

https://github.com/ynput/OpenPype/blob/c9cf6646f78223f492f21b8097f2b73ae6df6b64/openpype/hosts/maya/api/plugin.py#L278-L279


iLLiCiTiT:
Please bump version of maya and houdini addons in `server_addon`.

iLLiCiTiT (approved):
LGTM

PR #5408, opened 2023-08-03
Applications: Attributes creation

iLLiCiTiT:
## Changelog Description
Applications addon does not cause infinite server restart loop.

## Additional info
Applications addon creates set of all applications and tools from all versions and all settings. Also changes behavior of old version to they can live next each other -> that is achieved by `pre_setup` method which is ATM available only in latest dev server.

This will keep all applications from all versions of applications addon in attributes (not easy to remove).

## Testing notes:
1. Create packages
2. Upload `applications` addon to server
3. Restart server
4. It should not cause infinite restart of server
5. Choose the addon in a bundle and set the bundle to production
6. Go to settings and add an application variant (anywhere)
7. Restart server
8. It should not cause infinite restart of server and the application should be in attributes


server_addon/applications/server/__init__.py
```diff
@@ -48,6 +103,19 @@ async def get_default_settings(self):
 
         return self.get_settings_model()(**default_values)
 
+    async def pre_setup(self):
+        """Make sure older version of addon use the new way of attributes."""
+
+        instance = AddonLibrary.getinstance()
+        app_defs = instance.data.get(self.name)
+        old_addon = app_defs.get("0.1.0")
```
iLLiCiTiT, line 111:
```suggestion
        old_addon = app_defs.versions.get("0.1.0")
```

kalisp (approved):

PR #5409, opened 2023-08-04
Nuke: publish existing frames on farm

jakubjezek001:
## Changelog Description
This PR proposes adding a fourth option in Nuke render publish called "Use Existing Frames - Farm". This would be useful when the farm is busy or when the artist lacks enough farm licenses. Additionally, some artists prefer rendering on the farm but still want to check frames before publishing.

By adding the "Use Existing Frames - Farm" option, artists will have more flexibility and control over their render publishing process. This enhancement will streamline the workflow and improve efficiency for Nuke users.

## Additional info
Validate Existing Frames on farm needed to be refactored so it is including also the logic where no dependent jobs are needed.

This PR has been originally @fabiaserra contribution https://github.com/ynput/OpenPype/pull/5196. Thank you very much!

## To do:
- [x] fix `submit_nuke_deadline` so it submit only baking part
- [x] test new `collect_write` extensively with slates

## Testing notes:
1. In Nuke test task create render instance and render locally
2. set render instance to `Use existing frames - farm`
3. publish and see that one OpenPype job is on farm
4. make sure all reviewable publishing was completed as expected


openpype/hosts/nuke/plugins/publish/collect_writes.py
```diff
@@ -142,6 +145,24 @@ def process(self, instance):
             instance.data["representations"].append(representation)
             self.log.info("Publishing rendered frames ...")
 
+            if render_target == "frames_farm":
+                # Farm rendering
+                instance.data["toBeRenderedOn"] = "deadline"
```
Minkiu, line 150: We can update the dict in one go:
```py
instance.data.update({
    "toBeRenderedOn": "deadline",
    "transfer": False,
    "farm": True, # to skip integrate
})
```

openpype/hosts/nuke/plugins/publish/collect_writes.py
```diff
@@ -142,6 +145,24 @@ def process(self, instance):
...
+                instance.data["transfer"] = False
+                instance.data["farm"] = True  # to skip integrate
+                self.log.info("Farm rendering ON ...")
+
+                self.log.info(
+                    "Adding collected files %s to expectedFiles instance.data",
+                    collected_frames
+                )
+                if "expectedFiles" not in instance.data:
+                    instance.data["expectedFiles"] = list()
```
Minkiu, line 160: This could be:
```py
instance.data["expectedFiles"] = [
    os.path.join(output_dir, source_file)
    for source_file in collected_frames
]
```

Minkiu (approved):

64qam (approved):

openpype/hosts/nuke/plugins/publish/collect_writes.py
jakubjezek001: thanks for suggestion

openpype/hosts/nuke/plugins/publish/collect_writes.py
```diff
@@ -48,113 +33,134 @@ def process(self, instance):
...
+
+            self._add_farm_instance_data(instance)
+
+        elif render_target == "farm":
+            self._add_farm_instance_data(instance)
+
+        # set additional instance data
+        self._set_additional_instance_data(instance, render_target, colorspace)
+
+    def _set_existing_files_data(self, instance, colorspace):
```
fabiaserra, line 56: should we move this function into something more generic that we can reuse elsewhere?

openpype/hosts/nuke/plugins/publish/collect_writes.py
```diff
@@ -48,113 +33,134 @@ def process(self, instance):
...
+            list: collected frames
+        """
+        collected_frames = self._get_collected_frames(instance)
+
+        representation = self._get_existing_frames_representation(
+            instance, collected_frames
+        )
+
+        # inject colorspace data
+        self.set_representation_colorspace(
```
fabiaserra, line 73: out of scope for this PR but the fact that this function exists within the `ColormanagedPyblishPluginMixin` doesn't add much benefit IMO as I have found myself requiring to call it like a staticmethod and there's nothing within the function that really requires the class attributes or anything, it could be simply a util module function.

Because of this reason, on this new `pyblish_functions` that you created in 3.16 (with the good intent to abstract some of the functions so they can be reused), you were forced to pass the mixin object as an arg https://github.com/ynput/OpenPype/blob/018677d34d8a74bbd849b129f567a11c8c3a8926/openpype/pipeline/farm/pyblish_functions.py#L446 to the function just so you can callt he method, you wouldn't need that at all and you could call directly `set_representation_colorspace` from wherever util module it lives on.

Also in a lot of these functions from `pyblish_functions` I think it would be more reusable to remove the need of having to pass the `pyblish.api.Instance` object around and breaking it down further to the parts required of the instance (i.e., `create_skeleton_instance(instance)` -> `create_skeleton_instance(instance_data, anatomy)`...). Creating the Instance object is not always obvious from outside the plugins scope and for the most parts you don't need the whole object to do whatever the function does. This might be me not knowing how to instantiate a pyblish.Instance but for the use case I had to use it for (on-demand remote publishes to Deadline from CLI) I could do exactly the same with just passing the instance data dictionary but I had to copy paste most of those functions and change its args

openpype/hosts/nuke/plugins/publish/collect_writes.py
```diff
@@ -188,15 +194,184 @@ def process(self, instance):
...
+        representation = {
+            "name": ext,
+            "ext": ext,
+            "stagingDir": output_dir,
+            "tags": []
+        }
+
+        frame_start_str = "%0{}d".format(
+            len(str(last_frame))) % first_frame
+        representation['frameStart'] = frame_start_str
```
fabiaserra, line 274: would be good to add `black` as a pre-commit hook so we don't keep mixing single quotes and double quotes on strings

openpype/hosts/nuke/plugins/publish/collect_writes.py
```diff
@@ -188,15 +194,184 @@ def process(self, instance):
...
+        # get file path knob
+        node_file_knob = write_node["file"]
+        # list file paths based on input frames
+        expected_paths = list(sorted({
+            node_file_knob.evaluate(frame)
+            for frame in range(first_frame, last_frame + 1)
+        }))
+
+        # convert only to base names
+        expected_filenames = [
```
fabiaserra, line 365: maybe you could do the same with just a single for loop?
```python
        # make sure files are existing at folder
        collected_frames = [
            filename
            for filename in os.listdir(output_dir)
            if os.path.join(output_dir, filename) in expected_paths
        ]
```

openpype/hosts/nuke/plugins/publish/collect_writes.py
```diff
@@ -188,15 +194,184 @@ def process(self, instance):
...
+                "{{:0{}d}}".format(len(str(last_frame)))
+            ).format(first_frame - 1)
+
+        slate_frame = collected_frames[0].replace(
+            frame_start_str, frame_slate_str)
+        collected_frames.insert(0, slate_frame)
+
+        return collected_frames
+
+    def _add_farm_instance_data(self, instance):
```
fabiaserra, line 320: feels like it would be useful as a generic function we could use in other plugins too

openpype/hosts/nuke/plugins/publish/collect_writes.py
jakubjezek001: I can see your points and thanks for your suggestions. Yeah you are right that we should rething some of those concepts. I would assume that best way to address those points will be to create two issues for enhancements. 

- Remove the function `set_representation_colorspace` from the `ColormanagedPyblishPluginMixin` and make it a util module function.
- Modify the **pyblish_functions** to remove the need for passing the **pyblish.api.Instance** object and break it down further into the required parts of the instance. For example, `create_skeleton_instance(instance)` should be modified to `create_skeleton_instance(instance_data, anatomy)`. 

I am resolving this comment as not related to the PR.

openpype/hosts/nuke/plugins/publish/collect_writes.py
jakubjezek001: I love the idea and would pretty much want this, but we are having PTS after we had tried to enforce some code quality via pre-commits hooks. So I am bit worried this would create big push back since we are having horrible inconsistency in whole code base.

openpype/hosts/nuke/plugins/publish/collect_writes.py
jakubjezek001: This is a the matter of preference and we all are having some. I myself prefer comprehension.

openpype/hosts/nuke/plugins/publish/collect_writes.py
```diff
@@ -188,15 +194,184 @@ def process(self, instance):
...
+        expected_paths = list(sorted({
+            node_file_knob.evaluate(frame)
+            for frame in range(first_frame, last_frame + 1)
+        }))
+
+        # convert only to base names
+        expected_filenames = [
+            os.path.basename(filepath)
+            for filepath in expected_paths
+        ]
```
BigRoy, line 368:
```suggestion
        expected_filenames = {
            os.path.basename(filepath)
            for filepath in expected_paths
        }
```
This should be a `set` looking at how this is being used below.

openpype/hosts/nuke/plugins/publish/collect_writes.py
fabiaserra: yeah I agree, for this reason `darker` (https://github.com/akaihola/darker) exists so it only reformats the lines that are changed so it's incrementally aligning to the consistency you want

openpype/hosts/nuke/plugins/publish/collect_writes.py
jakubjezek001: After further analysis of the code, I have not identified any instances where abstraction could be utilized elsewhere.

openpype/hosts/nuke/plugins/publish/collect_writes.py
jakubjezek001: After further analysis of the code, I have not identified any instances where abstraction could be utilized elsewhere. Do you have any suggestions?

fabiaserra: this same logic of setting `farm` on the instanfce to true to skip integrate step should be pretty common for any plugins that delegate the publish to the farm. For example https://github.com/search?q=repo%3Aynput%2FOpenPype+%22to+skip+integrate%22&type=code and of course, any custom plugins people write (like this PR I created that adds support to submit publish pipe to the farm https://github.com/ynput/OpenPype/pull/5195/files#diff-570c6ac4c5667daf43a0e9404fa8973fa092c74d94c8caa2ceb670c225ff77c5R118), it would be beneficial on starting to set the core API for any plugins that want to do the same

openpype/hosts/nuke/plugins/publish/collect_writes.py
fabiaserra: maybe this one is quite specific to the Nuke write node but similarly to my other comment https://github.com/ynput/OpenPype/pull/5409/files#r1286334363 there will be other plugin developers that would start implementing the same logic for setting existing files on representations. For example, the TrayPublisher farm support that I added https://github.com/ynput/OpenPype/pull/5195/files#diff-570c6ac4c5667daf43a0e9404fa8973fa092c74d94c8caa2ceb670c225ff77c5R128 I kind of hacked the framework so the `expectedFiles` attribute would turn into `files` on the `submit_publish_job.py:prepare_representations` because there was no existing common functions for doing exactly that. For example, look at how many duplicates of `generate_expected_files` we have in the codebase: https://github.com/search?q=repo%3Aynput%2FOpenPype+%22generate_expected_files%22&type=code , that's what I'd like to avoid

jakubjezek001: I agree that we need to provide an abstraction for existing frames. However, let's address this in a separate pull request.

iLLiCiTiT:
Code looks ok. Needs testing (all usecases).

Has anyone better name for `"frames_farm"`, something like `"farm_integrate_only"`?

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -474,6 +475,7 @@ def process(self, instance):
                 "FTRACK_SERVER": os.environ.get("FTRACK_SERVER"),
             }
 
+        deadline_publish_job_id = None
         if submission_type == "deadline":
```
BigRoy, line 479: Out of curiosity - when is this **not** the case whenever the Deadline plugins run?

openpype/hosts/nuke/plugins/publish/collect_writes.py
```diff
@@ -15,30 +14,16 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
     hosts = ["nuke", "nukeassist"]
     families = ["render", "prerender", "image"]
 
+    # cashing
```
BigRoy, line 17:
```suggestion
    # cache
```
or:
```suggestion
    # caching
```

fabiaserra: 💸

openpype/hosts/nuke/plugins/publish/collect_nuke_instance_data.py
```diff
@@ -40,5 +42,15 @@ def process(self, instance):
...
+
+        # add creator attributes to instance
+        creator_attributes = instance.data["creator_attributes"]
+        instance.data.update(creator_attributes)
+
+        # add review family if review activated on instance
+        if instance.data.get("review"):
+            instance.data["families"].append("review")
+
+
```
fabiaserra, line 54:
```suggestion
```

fabiaserra (approved):
looking good!! Thank you for doing this @jakubjezek001

64qam (approved):

openpype/modules/deadline/plugins/publish/submit_publish_job.py
jakubjezek001: It seems like this is only the case where `muster` would be used for farm publishing.

PR #5410, opened 2023-08-04
Max: fix the bug of handling Object deletion in OP Parameter

moonyuet:
## Changelog Description
If the object is added to the OP parameter and user delete it in the scene thereafter, it will error out the container with OP attributes. This PR resolves the bug.
This PR also fixes the bug of not adding the attribute into OP parameter correctly when the user enables "use selections" to link the object into the OP parameter.
## Additional info
Related to #5406, #5378, #5396
Please create new instance to test it
## Testing notes: (for the error)
1. Launch 3dsmax via launcher/OP
2. create instance 
3. Links some objects to the OP Parameter
4. Delete one or two objects
5. Click the container with the OP Parameter
6. Supposed it won't display the error again
## Testing notes for use selection
1. create instance by using selection
2. Click "Add to Container" and  "Delete from Container"
3. The filtering of the objects should be shown correctly


LiborBatek (approved):
Performed both tests with success.

1) Theres no errors when deleting some objects present in OP Container and showing the content of the container afterwards. All good.

2) Content of the OP Container is correctly filtered according to already present object(s) in the Container list. Again all working fine.

PR #5411, opened 2023-08-04
Colorspace: including environments from launcher process

jakubjezek001:
## Changelog Description
Fixed bug in GitHub PR where the OCIO config template was not properly formatting environment variables from System Settings `general/environment`.

## Testing notes:
1. make sure `general/environmnet` is configured for multiplatform path
image

2. use the defined environenment key in config path input
image

3. open nuke task and see it is correctly adding OCIO env var and setting config path to running process


iLLiCiTiT (approved):
Code looks ok. Didn't test.

PR #5412, opened 2023-08-04
TVPaint: Fix 'repeat' behavior

iLLiCiTiT:
## Changelog Description
Calculation of frames for repeat behavior is working correctly.

## Additional info
It did work only if layer started from frame `0`, now it should be able to handle any range.

## Testing notes:
1. Create asset with frame range 12 frames (1001 - 10013) - and task
2. Open workfile on task under the asset
3. Set mark in to 0 and mark out to 12
4. Make a layer that does start on frame 5
5. Make 4 frames in layer (with visibly different content)
6. Set pre and post behavior of the layer to repeat (loop)
7. Create -> Publish the layer
8. Check integrated frames and validate if the image content does match the content from workfile (the order of frames is correct)

### DIFF
--- a/openpype/hosts/tvpaint/api/lib.py
+++ b/openpype/hosts/tvpaint/api/lib.py
@@ -233,7 +233,7 @@ def get_layers_pre_post_behavior(layer_ids, communicator=None):
 
     Pre and Post behaviors is enumerator of possible values:
     - "none"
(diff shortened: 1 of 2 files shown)


64qam:
It is in staging at the client's end, let's wait for their feedback.

64qam (approved):

PR #5414, opened 2023-08-04
Webpublisher: Self contain test publish logic

iLLiCiTiT:
## Changelog Description
Moved test logic of publishing to webpublisher. Simplified `remote_publish` to remove webpublisher specific logic.

## Additional info
Function `remote_publish` was kept because of unknown reason of it's existence. It's used only in script `./openpype/scripts/remote_publish.py` which is used only in deadline plugin for maya remote publish. But the logic completelly skips registering of host, registering of plugins and handling of DCC process so I do expect it can be removed -> future PRs.

## Testing notes:
1. Webpublisher should work as before

### DIFF
--- a/openpype/hosts/webpublisher/addon.py
+++ b/openpype/hosts/webpublisher/addon.py
@@ -20,11 +20,10 @@ class WebpublisherAddon(OpenPypeModule, IHostAddon):
         Close Python process at the end.
         """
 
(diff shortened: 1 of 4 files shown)


kalisp (approved):
Both automatic tests and WP seems to be still working.

PR #5416, opened 2023-08-06
Fusion - Revamp OpenPype menu

EmberLightVFX:
## Changelog Description
This PR revamps the OP menu inside of Fusion. Instead of having to load an OP instance in Fusion and using a custom UI window we now communicate with OP though a socket. We send the Fusion UUID together with what menu command to execute leading to the execution happening in the correct Fusion window if the user have multiple Fusion instances open.

## Additional info
- Everything should work as expected but I haven't currently bin able to open any UI yet. When I do I get the error `openpype.pipeline.anatomy.ProjectNotSet: Implementation bug: Project name is not set. Anatomy requires to load data for specific project.` I'm sure I'm executing the command to open any window the wrong way. I have tried to look how other addons does it but I never found a solution. Would love some help with this part!
- I also left the old OP Menu UI in the Fusion menu so I could see if I broke something else while trynig to get everything working.
- Currently the socket port is a hardcoded 12345 but will be made into a setting later on.
- The path to the fusionscript dll is hardcoded. It should be fetched from the entered fusion path from the settings.

## Todo
- [ ] Create a new process for each Fusion launch
- [ ] Make socket port dynamic and passed to Fusion as a env var
- [ ] Clean up old code and remove unused files
- [ ] Make fusionscript path dynamic to the fusion location
- [ ] Add keyboard shortcuts to the menu items

## Testing notes:
1. Open an asset in Fusion
2. Use the OpenPype menu to open any window


openpype/hosts/fusion/deploy/MenuScripts/openpype_menu_launch.py
```diff
@@ -0,0 +1,23 @@
...
+
+    client_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    server_address = ("localhost", 12345)
+    client_socket.connect(server_address)
+    client_socket.send(data_json.encode())
+    client_socket.close()
+
+
+if "menu_item" in data and "uuid" in data:
+    send_message_to_OpenPype(data)
```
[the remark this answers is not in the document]
EmberLightVFX: The data is passed to this file from openpype_menu.fu
This is correct

openpype/hosts/fusion/deploy/MenuScripts/openpype_menu_launch.py
```diff
@@ -0,0 +1,23 @@
...
+    data_json = json.dumps(data_to_send)
+
+    client_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    server_address = ("localhost", 12345)
+    client_socket.connect(server_address)
+    client_socket.send(data_json.encode())
+    client_socket.close()
+
+
+if "menu_item" in data and "uuid" in data:
```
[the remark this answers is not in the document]
EmberLightVFX: The data is passed to this file from openpype_menu.fu
This is correct

openpype/hosts/fusion/api/lib.py
```diff
@@ -333,16 +334,47 @@ def get_frame_path(path):
...
     """Get bmd library"""
-    bmd = getattr(sys.modules["__main__"], "bmd", None)
-    return bmd
+    return _load_fusuionscriptlib()
+
+
+def get_fusion_module():
+    """Get current Fusion instance"""
+    try:
+        fusion_module = getattr(sys.modules["__main__"], "fusion")
```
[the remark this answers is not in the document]
EmberLightVFX: This is part of the old code and will stay until the new implementation is fully working

openpype/hosts/fusion/api/lib.py
```diff
@@ -333,16 +334,47 @@ def get_frame_path(path):
...
+    try:
+        import fusionscript as script_module
+
+        return script_module
+    except ImportError as e:
+        lib_path = ""
+        if sys.platform.startswith("darwin"):
+            lib_path = "/Applications/Blackmagic Fusion 18/Fusion.app/Contents/MacOS/fusionscript.so"
+        elif sys.platform.startswith("win"):
+            lib_path = "C:\\Program Files\\Blackmagic Design\\DaVinci Resolve\\fusionscript.dll"
```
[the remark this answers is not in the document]
EmberLightVFX: Will be shorter when I replace the hardcoded path

openpype/hosts/fusion/api/lib.py
```diff
@@ -333,16 +334,47 @@ def get_frame_path(path):
...
-    return fusion
+def _load_fusuionscriptlib():
+    try:
+        import fusionscript as script_module
+
+        return script_module
+    except ImportError as e:
+        lib_path = ""
+        if sys.platform.startswith("darwin"):
+            lib_path = "/Applications/Blackmagic Fusion 18/Fusion.app/Contents/MacOS/fusionscript.so"
```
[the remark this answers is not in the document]
EmberLightVFX: Will be shorter when I replace the hardcoded path

openpype/hosts/fusion/api/lib.py
```diff
@@ -333,16 +334,47 @@ def get_frame_path(path):
...
+
+        if not os.path.isfile(lib_path):
+            print("[Fusion] [Library Does Not Exist on Disk]", lib_path)
+
+        bmd = imp.load_dynamic("fusionscript", lib_path)
+
+        if not bmd:
+            raise ImportError(
+                "[BMD] Could not locate module dependencies"
+            ) from e
```
BigRoy, line 359: Should we cache this - so that every call to `_load_fusuionscriptlib` and thus to `get_bmd_library` doesn't need to resolve the paths and load dynamically each time and re-use it once?

openpype/hosts/fusion/deploy/MenuScripts/openpype_menu_launch.py
```diff
@@ -0,0 +1,24 @@
...
+import socket
+import json
+
+
+def send_message_to_OpenPype(data_to_send):
+    # Serialize the Python dictionary into a JSON string
+    data_json = json.dumps(data_to_send)
+
+    client_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    server_address = ("localhost", 12345)
```
BigRoy, line 12: I'm pretty sure this won't work for multiple fusion sessions open at the same time. They'd both be sending the commands to the same process - and thus, that process that should be "in a certain openpype context, like project, asset, shot, etc." now suddenly has to be in two contexts at the same time. I feel like this will not work and we'll need to have at least a separate process per fusion session.

Also, sessions can crash. I feel there'll be a need to IF no response is received that maybe we'll restart the 'menu' process for the current session.

openpype/hosts/fusion/hooks/pre_fusion_menu_communication.py
```diff
@@ -0,0 +1,28 @@
...
+    order = 3
+    launch_types = {LaunchTypes.local}
+
+    def execute(self):
+        # Connect the triggered signal to the internal function's execute slot
+
+        menu_socket_listener = MenuSocketListener()
+        menu_socket_listener.start()
+
+        self.log.info("Starting socket listener")
```
BigRoy, line 28: We could set an env var here like `OPENPYPE_FUSION_SOCKET_PORT` and make it unique per Fusion session in a way that if a socket is already in use by another fusion session then we'll need to pick another port for this particular fusion session.

openpype/hosts/fusion/deploy/Config/openpype_menu.fu
```diff
@@ -41,6 +41,176 @@
...
+        Name = "Workfiles...",
+
+        Targets =
+        {
+            Composition =
+            {
+                Execute = _Lua [=[
+                    data = {menu_item = "workfiles", uuid = bmd.getappuuid()}
+                    local scriptPath = app:MapPath("OpenPype:MenuScripts/openpype_menu_launch.py")
+                    target:RunScript(scriptPath, { data = data })
```
BigRoy, line 57: I'm wondering if at some point we can make this so that the socket emission doesn't go through Python but straight through lua. That'd require any python dependencies to be removed completely from Fusion plus I've always found that running Python scripts compared to Lua run relatively slowly (with some delay). Much less of a delay than when also importing PySide of course, but still - I wonder if it'd be a worthwhile optimization.

BigRoy (commented):
Nice prototype. I didn't test it yet. But did have some questions/comments, so I added them to the review.

[I commented on the forum about why likely the tools are not working - it's likely due to `install_host()` not having been triggered in the socket listening process](https://community.ynput.io/t/change-fusion-integration-to-socket-communication/541/9?u=bigroy) (where all the tools are actually running) **but** it seems that you are triggering `install_host(FusionHost())` - then the only thing I can still think of is that the socket listener process isn't or doesn't stay completely in sync with the Fusion session. It looks like you're running the listening process straight inside the launcher Python process. I'm quite confident that won't work. You'll at least need to spawn of a custom python process that is NOT the tray/launcher process but dedicated to the fusion process solely for the purpose of being dedicated to the right environment (os.environ tools, but also openpype context like project, asset, etc.) since I believe there currently isn't a way in OpenPype to run multiple OpenPype context including tools together within a single python process.

openpype/hosts/fusion/deploy/Config/openpype_menu.fu
EmberLightVFX: That was my hope but to use sockets in Lua you have to install an external library through LuaRocks. Maybe there are other ways to solve it but I didn't find any other way.

openpype/hosts/fusion/api/lib.py
EmberLightVFX: It is "cached" as it tries to import the script module under Try. If it can't import it then it tries to import the dll file.

BigRoy: You are absolutely right. 👍

openpype/hosts/fusion/deploy/MenuScripts/openpype_menu_launch.py
EmberLightVFX: Sounds logical. I was thinking of passing the socket to Fusion over an environment variable. That way it could be randomly picked so it won't already be in use and simply pass it like any other env var.

In short there would be a pre-hook that executes exactly like the old Menu UI object but instead of creating the UI it creates a socket listener and we control the process through socket commands.

EmberLightVFX: I think I will need to get some help on how to set up the new process correctly as I have no idea how to do that in OP.

EmberLightVFX: Best would be to have one process tho and pass what project and asset is being used. That way we could get tabs in Fusion to work correctly! But that could be done in a later PR

iLLiCiTiT:
I just quickly scrolled through the code, but if I understand the implementation correctly, it is one-way communication, which is alive only few miliseconds (send message and forget). Each tool has it's own process -> technically there can be multiple processes with the same tool (have opened workfiles tool 0-n times). I'm not sure how "current context" would be propagated to/from the fusion process, e.g. on context change with workfiles tool.

EmberLightVFX:
@iLLiCiTiT a new process would need to be started each time the user opens a comp. Fusion then communicate with that process.
Currently I'm using the traybar-process from what I understand. I would need to get a new process to start like I described above.
The python script from Fusion is only for communication.

I'm away right now but might be able to rethink some parts and see how it would be possible to implement this correct

EmberLightVFX:
So I have gotten this PR working!
When you launch Fusion a new process gets launched that Fusion will communicate with.
Some problems:
* It seems that the old publisher gets loaded instead of the new one
* When you open a menu item a new QApplication gets created. When you close the menu item and re-open it it tries to use the old QApplication but the window becomes white and freezes.

I'm starting to question if all this work is worth it or not 😅
It's nice to have the terminal outside of Fusions console as OP fills it up with error messages (even if it's not error messages) but yeah... feels like quite some coding left to make it work flawlessly

mkolar:

I'm closing it down, but we'll he happy for a new PR to [ynput/ayon-fusion](https://github.com/ynput/ayon-fusion) repository once it's up.

PR #5417, opened 2023-08-07
AYON: Add folder to template data

iLLiCiTiT:
## Changelog Description
Added `folder` to template data, so `{folder[name]}` can be used in templates.

## Additional info
The folder has only `name` at this moment for which is used asset name. Can be also used in OpenPype. As far as I'm aware there are no other places where data for templates are prepared at this moment.

### Beginning
1. Make openpype addon, upload to AYON, set in bundle
2. Make sure you're using `{folder[name]}` in anatomy templates
3. Run OpenPype

### Workfiles
- Last workfile should be found on startup
- Last workfile should be found in workfiles tool
- Save in workfiles tool should work correctly
- Save as in workfiles tool should work correctly

### Publishing
1. Publish from host of your choice
2. Publishing should went smoothly

### DIFF
--- a/openpype/client/server/conversion_utils.py
+++ b/openpype/client/server/conversion_utils.py
@@ -133,7 +133,6 @@ def _get_default_template_name(templates):
 def _template_replacements_to_v3(template):
     return (
         template
-        .replace("{folder[name]}", "{asset}")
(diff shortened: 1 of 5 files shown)


kalisp (approved):
Ayon and OP publishes work.

PR #5421, opened 2023-08-07
Houdini: Use default_variant to Houdini Node TAB Creator

fabiaserra:
## Changelog Description

Use the default variant of the creator plugins on the interactive creator from the TAB node search instead of hard-coding it to `Main`.

## Testing notes:
1. Change the default_variants for one of the Houdini creators
2. Use the Houdini node tab creation (i.e., `Create Arnold ROP`)
3. Check that the default variant on the dialog isn't `Main`


openpype/hosts/houdini/api/creator_node_shelves.py
```diff
@@ -196,7 +196,14 @@ def install():
 
             key = "openpype_create.{}".format(identifier)
             log.debug(f"Registering {key}")
-            script = CREATE_SCRIPT.format(identifier=identifier)
+            default_variant = "Main"
+            if hasattr(creator, "default_variant"):
+                default_variant = creator.default_variant
+            elif hasattr(creator, "default_variants"):
+                default_variant = creator.default_variants[0]
```
BigRoy, line 203: To me supporting both `default_variant` and `default_variants` makes no sense really. Why do we have both attributes? Can't we just only have `default_variants = ["Main"]`?

openpype/settings/defaults/project_settings/houdini.json
```diff
@@ -14,48 +14,99 @@
...
+        },
+        "CreateVDBCache": {
+            "enabled": true,
+            "default_variant": "Main",
+            "default_variants": ["Main"]
+        },
+        "CreateVrayROP": {
+            "enabled": true,
+            "default_variant": "master",
+            "default_variants": ["master"]
```
BigRoy, line 109: Shouldn't this also be `Main`?

BigRoy:
> Finally, this also now makes use of the default variant of the creator plugins on the interactive creator from the TAB node search instead of hard-coding it to Main.

❤️ Nice work!

openpype/settings/defaults/project_settings/houdini.json
fabiaserra: yeah... I basically prioritized backwards compatibiliy here before they got deleted at https://github.com/ynput/OpenPype/pull/5367/files by @mustafa-zarkash . I personally will be overriding all the default subsets in our settings but happy to change to some other more reasonable defaults

openpype/hosts/houdini/api/creator_node_shelves.py
fabiaserra: Yeah fair point, I think ideally it would be an int index instead of another text input as it definitely wouldn't make sense to have a value for `default_variant` that doesn't exist in `default_variants` but we need to have a way to specify which of all the `default_variants` items we want as the main default right? Or we could just enforce it so it's always the first one on the list given that list inputs are sortable...?
image

fabiaserra: but there's a good amount of code already expecting a single default_variant https://github.com/search?q=repo%3Aynput%2FOpenPype+%22default_variant%22&type=code so this would require a fair amount of changes

BigRoy: > I think ideally it would be an int index instead of another text input as it definitely wouldn't make sense to have a value for default_variant that doesn't exist in default_variants but we need to have a way to specify which of all the default_variants items we want as the main default right? Or we could just enforce it so it's always the first one on the list given that list inputs are sortable...?

Hmm, good point I suppose. Technically default variants might make most sense as a `set` and then there'd be no index nor "first index" so in that case `default_variant` would have to be the string variant. Anyway, it still feels duplicated haha. It's a similar issue to `family` and `families`.

openpype/settings/defaults/project_settings/houdini.json
mustafa-zarkash: the default as milan mentioned is to stick with `"Main"`
https://github.com/ynput/OpenPype/pull/5367#pullrequestreview-1557577660

openpype/hosts/houdini/api/creator_node_shelves.py
mustafa-zarkash: I think we can wire `default_variant`  and `default_variants` 
but, I'm not sure what is the best level to do that. 

> this code is an untested proof of concept

I can think of two levels : 
1. `Creator` (Openpype level): to convert this [`get_default_variant`](https://github.com/ynput/OpenPype/blob/a31b2d9d77f834441d65733c8ee91c6c3ca5b623/openpype/pipeline/create/creator_plugins.py#L609-L619) to property that dynamically gets the first value (if exists) of the  `default_variants` 
```python
@property
 def default_variant(self):
    return next(iter(self.default_variants or []), None)
```
> which eliminates the use of mentioning `default_variant` at all 

another solution (which adds backward compatibility for settings)
which may be the case when  `default_variant` found in settings and another code is trying to set `default_variant`
```python
_default_variant = None

@property
 def default_variant(self):
    return self._default_variant 

@default_variant.setter
 def default_variant(self, default_variant):
     self._default_variant  = default_variant

def get_default_variant(self):
    # get the first element of default_variants if default_variant is None
    # if default_variants is empty get a default value "Main" 
    if  not self.default_variant:
        self.default_variant = next(iter(self.default_variants or []), None) or "Main"
    
    return  self.default_variant
```


2. `HoudiniCreator` (DCC Integration level) : we can define `default_variants` as list and setting `default_variant` to the first element, but I'm not sure about how to do this may be in an `__init__` method ?

mustafa-zarkash:
So, a quick recap because I'm lost a little bit 
This PR
1. adds create subsets in TAB menu (Interactive creation)
2. adds `default_variant = "Main"` in `HoudiniCreator` 
3. updates maya and houdini project settings (add new settings)
4. updates maya and houdini project settings (set default values)

I believe there's a missing step 
5. Replicates these settings in [AYON addons settings](https://github.com/ynput/OpenPype/tree/develop/server_addon). 
and please bump version of maya and houdini addons in `server_addon`.

--- 
As you pointed out, some of these work were done in previous PRs which 
1. update `defaults` to `default_variables` #5367 and #5368
2. reflects these settings in OP and Ayon settings #5407

 > I only updated `defaults` variables in `houdini.json` and `maya.json` files that had corresponding settings in `schema_houdini_create.json` and `schema_maya_create.json`.

openpype/hosts/houdini/api/creator_node_shelves.py
BigRoy: The `next(iter(..))` approach wouldn't make much sense if `default_variants` was a `set` because it'd be unsorted and thus you'd have no idea which one it would pick. If it's a list and will stay a list then it makes sense, but then also does `default_variants[0]`.

mustafa-zarkash (commented):
#### Regarding updating creator shelves: 
it works and it shows the `default_variant` in the box 

image

image

#### Regarding updating settings: 
I will look at them in another time, I need to be more focused so that I don't miss something

also, openpype keep displaying these warnings 
image


Good Job!

openpype/hosts/houdini/api/creator_node_shelves.py
mustafa-zarkash: I didn't know that `default_variants` can be something other than a list. 

> if `default_variants` was a `set` because it'd be unsorted and thus you'd have no idea which one it would pick.

I wonder if that the reason why we have `default_variant` and not just `default_variants` 

so, what do you think about ?  we were discussing if `default_variant` can be derived from `default_variants`  
@iLLiCiTiT

openpype/hosts/houdini/api/creator_node_shelves.py
```diff
@@ -35,11 +35,11 @@
 
 CREATE_SCRIPT = """
 from openpype.hosts.houdini.api.creator_node_shelves import create_interactive
-create_interactive("{identifier}", **kwargs)
+create_interactive("{identifier}", "{variant}", **kwargs)
 """
 
 
-def create_interactive(creator_identifier, **kwargs):
+def create_interactive(creator_identifier, default_variant, **kwargs):
```
iLLiCiTiT, line 42: From the usage, this should be just variant, why `default_variant`?
```suggestion
def create_interactive(creator_identifier, variant, **kwargs):
```

BigRoy: Actually it is _just_ the default value for the prompt - the user is still free to change the value. Whether that makes the argument clearer or not I'll leave in the middle for you both to discuss.

iLLiCiTiT:
First of all, don't do any changes or PRs because I'm lost in all the conversations and we have to agree on something before any code changes.

I'll try to explain reasoning behind some of the stuff, before anything else.
1. `default_variant` and `default_variants` are class attributes to match most of attributes from other plugins (loader, publish, legacy create etc.), that's why they're not `property` and don't have `_` at the start - so they can be easily set per plugin, even without settings
2. attributes `default_variant` and `default_variants` can be completelly unused if creator overrides `get_default_variant` and `get_default_variants`, which should be used as source of truth
3. `get_default_variants` is used to fill variant hints (popup) and it's first value is used for default variant if `default_variant` is empty
4. `get_default_variant` is optional way how to use "different value from `default_variants`", so it should not show in variant hints

These are the points I would like to mention. I think the naming of the `default_variants` variable was not right, it should be `variant_hints` (NOTE: Do not change it).

Now, I do agree that `get_default_variant` 1. should always return a string and never `None` 2. should look into first value of `default_variants` or use `"main"` in default implementation, I think we should lower the default value (that why `{Variant}` is in default template...

Then `default_variant` attribute does not have to be class attribute and can be a property without setter (why setter?), that is using `get_default_variant` method, we can change `default_variant` to `_default_variant` for those creators who would like to use different values from the hints. I don't think that any create plugin should define this value in class definition and in 99.97% should be used first value from `get_default_variants`.

openpype/hosts/houdini/api/plugin.py
```diff
@@ -169,6 +169,8 @@ class HoudiniCreator(NewCreator, HoudiniCreatorBase):
     selected_nodes = []
     settings_name = None
 
+    default_variant = "Main"
```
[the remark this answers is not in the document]
fabiaserra: interesting, I got a `None` value out of the creator.default_variant on the node TAB search without this but this is no longer needed as long as all creators have a `default_variant` implemented so I will remove it

openpype/hosts/houdini/api/creator_node_shelves.py
fabiaserra: correct yeah, `default_variant` represents better what the value is used for

fabiaserra:
> EDITED: From the conversation I'm not sure if is this PR still relevant? I don't think all the settings changes should be in this PR, but it looks like there are important changes in code.

I think it's still relevant for these changes:

* Use `default_variant` from creator on Houdini node TAB search creation instead of hard-coded default value
* Expose `default_variants` and `default_variant` on Houdini Creator settings
* Add missing Houdini creators on settings (i.e., CreateArnoldROP, CreateMantraROP...)

As for how to proceed:

* Do you want me to rename all `default_variant[s]` on the settings to be lower-case `"main"` instead? Only for Houdini to no affect legacy defaults from Maya?
* Remove the fallback `"Main"` that I added on the Houdini base creator https://github.com/ynput/OpenPype/pull/5421#discussion_r1286617157

iLLiCiTiT:
> Use default_variant from creator on Houdini node TAB search creation instead of hard-coded default value
> Expose default_variants and default_variant on Houdini Creator settings

Did you read this part:
```
I don't think that any create plugin should define value, of what current default_variant is, in class definition, it's more about special usecases we had for specific workflows based on settings. It does not make sense in 99.97% cases, and first value from get_default_variants should be used instead.
```

fabiaserra:
> Did you read this part:
> I don't think that any create plugin should define value, of what current default_variant is, in class definition, it's more about special usecases we had for specific workflows based on settings. It does not make sense in 99.97% cases, and first value from get_default_variants should be used instead.

I did but I'm probably not understanding what you mean so please expand on that and answer the how to proceed questions

iLLiCiTiT:
> Do you want me to rename all default_variant[s] on the settings to be lower-case "main" instead? Only for Houdini to no affect legacy defaults from Maya?

> I did but I'm probably not understanding what you mean so please expand on that and answer the how to proceed questions

We should do this in separated PRs and first should be the code change PRs and then settings.

So if you want to fix current issues then

1.) Do a PR for houdini creators settings, in which **only** adds missing plugins and change all `"master"` to `"main"`, but does not change the template to add `default_variant`.

2.) Second PR which does fix of `defaults` in maya plugin.

Meanwhile I'll create another PR which changes behavior of `default_variant` on plugin.

With all that you won't need `default_variant` in settings, unless you hit any usecase where it make sense for production.

fabiaserra:
> > Do you want me to rename all default_variant[s] on the settings to be lower-case "main" instead? Only for Houdini to no affect legacy defaults from Maya?
> 
> > I did but I'm probably not understanding what you mean so please expand on that and answer the how to proceed questions
> 
> We should do this in separated PRs and first should be the code change PRs and then settings.
> 
> But if you want to fix the important current issues then:
> 
> 1.) Do a PR for houdini creators settings, in which **only** adds missing plugins and change all `"master"` to `"main"`, but does not change the template to add `default_variant`.
> 
> 2.) Second PR which does fix of `defaults` in maya plugin.
> 
> Meanwhile I'll create another PR which changes behavior of `default_variant` on plugin.
> 
> With all that you won't need `default_variant` in settings, unless you hit any usecase where it make sense for production.

I think it makes sense to expose the default variant for the different creators. I definitely don't want to always see `"main"` as the variant that always gets created when using the TAB node creation, we need to have the control to choose how we want to name it as default for the different families:
image

But are you saying that you don't want this code change I did on openpype/hosts/houdini/api/creator_node_shelves.py ?

iLLiCiTiT:
> I think it makes sense to expose the default variant for the different creators. I definitely don't want to always see "main" as the variant that always gets created when using the TAB node creation, we need to have the control to choose how we want to name it as default for the different families:

It's combination of changes. Check [this PR](https://github.com/ynput/OpenPype/pull/5429), that implements the changes of default variant behavior on create plugin. The reason is that if you have `default_variants = ["Something", "Else"]` and unfilled `default_variant` then `get_default_variant` will return `"Something"`. With that you probably don't need the settings for `default_variant`, or?

fabiaserra:
> It's combination of changes. Check [this PR](https://github.com/ynput/OpenPype/pull/5429), that implements the changes of default variant behavior on create plugin. The reason is that if you have `default_variants = ["Something", "Else"]` and unfilled `default_variant` then `get_default_variant` will return `"Something"`. With that you probably don't need the settings for `default_variant`, or?

Yeah that's fine for the effort of removing the need of `default_variant` on the settings but we will still need to call the function `creator.get_default_variant` on openpype/hosts/houdini/api/creator_node_shelves.py for it to populate that dialog as that's separate from the Create widget. I will do the changes and update the PR so it's more clear

iLLiCiTiT:
> Yeah that's fine for the effort of removing the need of default_variant on the settings but we will still need to call the function creator.get_default_variant on openpype/hosts/houdini/api/creator_node_shelves.py for it to populate that dialog as that's separate from the Create widget. I will do the changes and update the PR so it's more clear

Yes, but this PR contains multiple different changes. Please start again and split them. We will also have to change AYON settings in the Maya and Houdini settings PRs. Right now it has 16 commits which few of them are changing it one way and other are changing it back...

fabiaserra:
> Yes, but this PR contains multiple different changes. Please start again and split them. We will also have to change AYON settings in the Maya and Houdini settings PRs. Right now it has 16 commits and few of them are changing it one way and other are changing it back...

Well that's the nature of code reviews... isn't it worth having the history of discussion and reasoning behind the changes? I can update this PR so it's only for the use of `default_variant` for the Houdini "TAB creator" and create a separate one for the setting changes, or you want to close this one and create two new ones?

openpype/settings/entities/schemas/projects_schema/schemas/schema_maya_create_render.json
```diff
@@ -17,4 +17,4 @@
             "object_type": "text"
         }
     ]
-}
\ No newline at end of file
+}
```
fabiaserra, line 20: this must have been pushed before the end-of-file-fixer pre-commit

iLLiCiTiT:
> Well that's the nature of code reviews... isn't it worth having the history of discussion and reasoning behind the changes? I can update this PR so it's only for the use of default_variant for the Houdini "TAB creator" and create a separate one for the setting changes, or you want to close this one and create two new ones?

It's worth to keep discussion, but not 20 commits changing code one way and then back. Please use force-push with only change related to this PR or create new PR from clean branch. All commits from the PR create noise in commit history.

iLLiCiTiT:
> Why wouldn't they be on settings? They are on Maya and Nuke creators and we are making use of it, that was the point of this PR, I need to be able to set the default variants on the creators

Like I've said, we need to change it to fix the issue we have ASAP. Adding other creators is "change" that should be validated by someone who knows what the creators are and do, we don't have that one at this moment.

> RE: commit history is just a matter of merging with squash so really no argument there

Does squash merge remove blame?
image

fabiaserra:
> Like I've said, we need to change it to fix the issue we have ASAP. Adding other creators is "change" that should be validated by someone who knows what the creators are and do, we don't have that one at this moment.

Fair enough 

> Does squash merge remove blame? image

It wouldn't remove blame, it would just treat all the changes on the PR as a single commit on the git history so the touched lines that were reverted on all the commits of the PR would show the original blame

iLLiCiTiT:
> It wouldn't remove blame, it would just treat all the changes on the PR as a single commit on the git history so the touched lines that were reverted on all the commits of the PR would show the original blame

OIn that case it's ok.

openpype/hosts/houdini/api/creator_node_shelves.py
```diff
@@ -196,7 +196,9 @@ def install():
 
             key = "openpype_create.{}".format(identifier)
             log.debug(f"Registering {key}")
-            script = CREATE_SCRIPT.format(identifier=identifier)
+            script = CREATE_SCRIPT.format(
+                identifier=identifier, variant=creator.get_default_variant()
```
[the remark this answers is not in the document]
fabiaserra: yeah well that's because I removed the logic of where `default_variant` is coming from on this PR after @iLLiCiTiT addressed it on another PR (https://github.com/ynput/OpenPype/pull/5436). I would expect if we merge `develop` over now this should be fixed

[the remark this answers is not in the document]
fabiaserra: I have merged latest `develop` to this branch now

mustafa-zarkash (approved):
LGTM, It works as expected! 
Good Job 👏

openpype/hosts/houdini/api/creator_node_shelves.py
```diff
@@ -59,9 +59,9 @@ def create_interactive(creator_identifier, **kwargs):
     """
 
     # TODO Use Qt instead
```
BigRoy, line 61: Have we considered doing it like this instead?

```python

    host = registered_host()
    context = CreateContext(host)
    creator = context.manual_creators.get(creator_identifier)
    if not creator:
        raise RuntimeError("Invalid creator identifier: "
                           "{}".format(creator_identifier))

    # TODO Use Qt instead
    result, variant = hou.ui.readInput("Define variant name",
                                       buttons=("Ok", "Cancel"),
                                       initial_contents=creator.get_default_variant(),
                                       title="Define variant",
                                       help="Set the variant for the "
                                            "publish instance",
                                       close_choice=1)
    if result == 1:
        # User interrupted
        return
    variant = variant.strip()
    if not variant:
        raise RuntimeError("Empty variant value entered.")
```

That way we don't need to pass the argument plus the default variant is "live" to the creator (and thus from settings?) instead of being initialized once on Houdini launch. Not sure if better, might also be slightly slower when the node creation is actually triggered since it'll need to query the default variant from the creator at that moment in time.

fabiaserra: You are the one who wrote most of this so I would ask you if you ever consider it :)

But I don't see that many benefits on having it "live"?

fabiaserra: yeah I don't mind either TBH, I have done the changes now here https://github.com/ynput/OpenPype/pull/5421/commits/ab3d94fdb6f607b958704a06d3fa87187ac03d71 do you like it more?

antirotor (approved):
works nice, thanks!

PR #5424, opened 2023-08-08
Max: switching versions of maxScene maintain parentage/links with the loaders

moonyuet:
## Changelog Description
When using scene inventory to manage or update the version of the loading objects, the linked modifiers or parentage of the objects would be kept.

Meanwhile, loaded objects from all loaders no longer parented to the container with OP Data. 

## Additional info
#5406 affects how this PR works.
It can be ready for review after #5406 merged.
It affects #5434 

## Testing notes:
1. Launch Max via OP
2. Click Load.. via OP Menu
3. Load objects from all loaders
4. Click Manage... via OP Menu
5. Switch Version or Update to the latest by right click
6. Switch the version twice to ensure no bug has been found


LiborBatek (commented):
Now when native ```MaxScene``` asset being loaded via ```Loader``` the OP Container is not acting as Parent so there wont happen any Linking (Parent>Child)  breaking.

as seen here where the ```Dummy``` acting still as an Parent which was active during ```Publish``` of the asset. All good!
image

However there some issues happening when Managing the scene when multiple ```assets``` being loaded/managed in the workfile. 

I will post soon on this issue.

LiborBatek:
Problem being when managing loaded asset version. Whenever triggering ```Set Version``` (u can choose whichever ver you like) it re-appear new asset in the workfile.

Steps to reproduce it:

1) Load an asset (MaxSceneLoader)
2) Switch Version
3)If neccessary repeat step 2)
4) Asset got into scene and are multiplied each time

Here is an example managing several times just single asset in the workfile whichs results into multiple assets (triggered 2 times)
image

moonyuet:
@LiborBatek The switching version should be fixed....This is caused by the parameter of the OP data not being updated by the renamed max objects in namespace
https://github.com/ynput/OpenPype/assets/64118225/1cf1e657-13a2-4b41-be33-dfea2865eb90

LiborBatek (approved):
Now switching versions on multiple same assets works perfect! User being able to update/set versions freely on single or multiple assets without any issues.

Multiple version changes been tested on single / multiple assets at once and all workfing as expected...
image

PR #5425, opened 2023-08-08
Show thumbnails in Workfiles tool

BigRoy:
## Changelog Description

- Show thumbnails in Workfiles tool (if there's a `{workfile}_thumbnail.jpg` found next to the workfile).
- Note: Currently only autogenerates thumbnails on save for Maya and Houdini

## Additional info

Draft implementation for [**Thumbnails for Workfiles**](https://community.ynput.io/t/thumbnails-for-workfiles/180/10).

### Tasks

- [x] Make sure there's no overwriting of thumbnails between different file extensions/hosts, so include e.g. .hip in the thumbnail names.
- [ ] Expose settings to enable/disable the workfile thumbnails completely?
- [ ] Expose settings to adjust the thumbnail locations?
- [ ] Thumbnail widget should either hide in "Published" view mode in the workfiles OR show the published workfile thumbnail if it has any.
- [ ] The ThumbnailPainterWidget should likely be moved in the codebase to not need import from the publisher tool.

## Previews

No thumbnail:
image

With saved thumbnail:
image
image
image

## Testing notes:

1. Open Maya or Houdini
2. Save scene
3. Open workfiles tool - it should show a thumbnail for the saved scene


BigRoy:
Preferably I also make it so that the list view itself shows the icons on the left hand side - however I was unable to quickly get that to work so it did a decent alignment. I got to this:

image (21)

But since it wasn't pretty I for now skipped that.

openpype/hosts/maya/api/pipeline.py
```diff
@@ -581,6 +583,26 @@ def on_save():
...
 
 
+def generate_thumbnail():
+    if lib.IS_HEADLESS:
+        log.debug("Generating thumbnail skipped due to headless mode")
+        return
+
+    current_filepath = current_file()
+    if not current_filepath:
+        log.error("No current workfile path. Thumbnail generation skipped")
```
BigRoy, line 593:
```suggestion
        log.error("No current workfile path. Thumbnail generation skipped")
        return
```

BigRoy:
> I'm not sure if it's good idea to create the thumbnail automatically.

Could you elaborate about the why?

It works nicely - and I agree with @LiborBatek's comment on the forum, if it doesnt do it automatically then you'll basically not be seeing anyone generate thumbnails.

When I did some test runs with e.g. Prism or other pipelines I noticed the behavior was similar where screenshots got taken on saves automatically - although some expose functionality on save to customize the thumbnail I'd rarely seen that being done in practice. Anyway, that's also where I noticed I really liked seeing the workfile thumbnail.

The thumbnails aren't intended to be super-slick best preview of the shots, but just a 'here's roughly what the scene has on save' so taking just an active viewport's state to me feels like a decent approach to that.

BigRoy:
Side note: If thumbnail view gets added to workfile tool we'll also need to consider making it work with the **Published** checkbox enabled in the tool so that thumbnails also potentially show from the published workfiles.

LiborBatek:
Note: I think theres no need to keep any consistency of ```thumbnail``` visuals aka no need to force it and keep it on users decision/preference. Thumbnail being simply some additional info/value/note for artist no one else.

Its fine as it is working now.

BigRoy:
> My only concern now is wether we should keep thumbnails sitting next to the workfile or create a folder like thumbs in workfile area so they arent clutter folder structure. But this being pretty subjective from user to user.

Another concern is that we're stripping of the file extension of the source workfile. By default OpenPype is configured to save from different applications/hosts into the same work folder with version numbers being independent. So you might have a:
```
asset_v001.hip
asset_v001.ma
```
But for those there can only be one `asset_v001_thumbnail.jpg` and thus you'll conflict with the thumbnail and potentially see the one of the other host. So we might need to:
```
asset_v001.hip.thumbnail.jpg
```
Or just prefix it with a dot and don't include `_thumbnail` in the name:
```
.asset_v001.hip.jpg`
```
With a subfolder it could e.g. be:
```
.thumbnails/asset_v001.hip.jpg
```
---
Just for comparison, I believe prism pipeline just saves the thumbnails next to the workfiles.

LiborBatek:
@BigRoy  All make sense! I guess there will be OP Settings for ```thumbnail``` at least to turn it ```ON``` or ```OFF``` so potentially there could be that user preference for ```folder``` as ```Optional``` if preffered.

Are there any show stoppers then? Regarding this thumbnail feature / PR?

BigRoy:
@BigRoy  All make sense! I guess there will be OP Settings for ```thumbnail``` at least to turn it ```ON``` or ```OFF``` so potentially there could be that user preference for ```folder``` as ```Optional``` if preffered.

> I guess there will be OP Settings for ```thumbnail``` at least to turn it ```ON``` or ```OFF``` so potentially there could be that user preference for ```folder``` as ```Optional``` if preffered.

We could offer templates for it - but if you were to changet the template mid-production it'd basically mean older thumbnails before the switch would end up not being shown anymore of course.

> Are there any show stoppers then? Regarding this thumbnail feature / PR?

1. Make sure there's no overwriting of thumbnails between different file extensions/hosts, so include e.g. `.hip` in the thumbnail names.
2. Expose settings to enable/disable the workfile thumbnails completely?
3. Expose settings to adjust the thumbnail locations?
4. Thumbnail widget should either hide in "Published" view mode in the workfiles OR show the published workfile thumbnail if it has any.
5. The `ThumbnailPainterWidget` should likely be moved in the codebase to not need import from the publisher tool.
6. (Optionally) Build the automatic functionality for more hosts for which it makes sense, e.g. Blender viewport, Fusion/Nuke last active viewer state. But these could very well be PRs of their own.

BigRoy:
Here's a prototype to [save the active view from Fusion as an image](https://gist.github.com/BigRoy/28a05a2728db11cb068f4460e3f0be6d). By adding that we'd also have the save thumbnail functionality for Fusion. 👍

tokejepsen:
Great feature! Have added the tasks left to be done in the description.

tokejepsen:
@BigRoy is this still needed? If so, could you resolve the merge conflicts so we can continue testing.

BigRoy:
> If so, could you resolve the merge conflicts so we can continue testing.

@tokejepsen done.

BigRoy:
@tokejepsen 

> However, I believe AYON now uses a completely separate Loader tool and thus this logic will need porting there.

As mentioned - I suspect this will currently indeed do nothing within AYON since the workfile tool there is not updated to contain it.

tokejepsen:

@BigRoy how do you wanna continue with this PR? Implement the other tasks and update the testing notes?
Or maybe reduced the scope of the PR to just included support for AYON, then work on the other tasks separately?

mkolar:
This is really valuable. The best course of action would probably be just adding AYON compatibility and take the extra tasks as further PRs.

tokejepsen:
@BigRoy lets get this AYON compatible and make issues or PRs for the other tasks.

mkolar:

PR #5426, opened 2023-08-08
Maya: Extract active view as thumbnail when no thumbnail set

BigRoy:
## Changelog Description

This sets the Maya instance's thumbnail to the current active view if no thumbnail was set yet.

## Additional info

It's currently set to only the workfile family for which it makes most sense - but I think it could even be nice for other families as at least a visual hint?

## Preview

image

## Testing notes:
1. Make sure thumbnails are enabled for your studio
    - Ensure you have `AVALON_THUMBNAIL_ROOT` set in OpenPype settings
3. Open Maya
4. Publish a workfile
5. Published workfile subset should now have a thumbnail

### DIFF
--- a/openpype/hosts/maya/plugins/publish/extract_active_view_thumbnail.py
+++ b/openpype/hosts/maya/plugins/publish/extract_active_view_thumbnail.py
@@ -46,5 +46,6 @@ class ExtractActiveViewThumbnail(pyblish.api.InstancePlugin):
             image.writeToFile(path, "jpg")
             self.log.debug("Generated thumbnail: {}".format(path))
 
+            context.data["cleanupFullPaths"].append(path)
             context.data[cache_key] = path
         return context.data[cache_key]
--- a/openpype/hosts/maya/plugins/publish/extract_active_view_thumbnail.py
+++ b/openpype/hosts/maya/plugins/publish/extract_active_view_thumbnail.py
@@ -7,7 +7,6 @@ import tempfile
 from openpype.hosts.maya.lib import IS_HEADLESS
 
 
-
 class ExtractActiveViewThumbnail(pyblish.api.InstancePlugin):
     """Set instance thumbnail to a screengrab of current active viewport.
 
--- a/openpype/hosts/maya/plugins/publish/extract_active_view_thumbnail.py
+++ b/openpype/hosts/maya/plugins/publish/extract_active_view_thumbnail.py
@@ -48,4 +48,3 @@ class ExtractActiveViewThumbnail(pyblish.api.InstancePlugin):
 
             context.data[cache_key] = path
         return context.data[cache_key]
-
--- a/openpype/hosts/maya/plugins/publish/extract_active_view_thumbnail.py
+++ b/openpype/hosts/maya/plugins/publish/extract_active_view_thumbnail.py
@@ -1,0 +1,51 @@
+import maya.api.OpenMaya as om
+import maya.api.OpenMayaUI as omui
+
+import pyblish.api
+import tempfile
+
+
+class ExtractActiveViewThumbnail(pyblish.api.InstancePlugin):
+    """Set instance thumbnail to a screengrab of current active viewport.
+
+    This makes it so that if an instance does not have a thumbnail set yet that
+    it will get a thumbnail of the currently active view at the time of
+    publishing as a fallback.
+
+    """
+    order = pyblish.api.ExtractorOrder + 0.49
+    label = "Active View Thumbnail"
+    families = ["workfile"]
+    hosts = ["maya"]
+
+    def process(self, instance):
+        thumbnail = instance.data.get("thumbnailPath")
+        if not thumbnail:
+            view_thumbnail = self.get_view_thumbnail(instance)
+            if not view_thumbnail:
+                return
+
+            self.log.debug("Setting instance thumbnail path to: {}".format(
+                view_thumbnail
+            ))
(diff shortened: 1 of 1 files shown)


openpype/hosts/maya/plugins/publish/extract_active_view_thumbnail.py
```diff
@@ -0,0 +1,50 @@
...
+    publishing as a fallback.
+
+    """
+    order = pyblish.api.ExtractorOrder + 0.49
+    label = "Active View Thumbnail"
+    families = ["workfile"]
+    hosts = ["maya"]
+
+    def process(self, instance):
+        thumbnail = instance.data.get("thumbnailPath")
```
BigRoy, line 22: This might also need an `IS_HEADLESS` check to ignore this plugin if maya is in batch/headless mode.

moonyuet (approved):
Looks good from my end.
image

openpype/hosts/maya/plugins/publish/extract_active_view_thumbnail.py
```diff
@@ -0,0 +1,50 @@
...
+            with tempfile.NamedTemporaryFile(suffix="_thumbnail.jpg",
+                                             delete=False) as f:
+                path = f.name
+
+            view = omui.M3dView.active3dView()
+            image = om.MImage()
+            view.readColorBuffer(image, True)
+            image.writeToFile(path, "jpg")
+            self.log.debug("Generated thumbnail: {}".format(path))
+
```
iLLiCiTiT, line 48: Add to cleanup files
```suggestion

context.data["cleanupFullPaths"].append(path)
```

openpype/hosts/maya/plugins/publish/extract_active_view_thumbnail.py
```diff
@@ -0,0 +1,50 @@
...
+            if not view_thumbnail:
+                return
+
+            self.log.debug("Setting instance thumbnail path to: {}".format(
+                view_thumbnail
+            ))
+            instance.data["thumbnailPath"] = view_thumbnail
+
+    def get_view_thumbnail(self, instance):
+        cache_key = "__maya_view_thumbnail"
```
iLLiCiTiT, line 34: What is this for?

openpype/hosts/maya/plugins/publish/extract_active_view_thumbnail.py
iLLiCiTiT: If you'd add `targets =["local"]` you should be assured it's in UI mode.

BigRoy: Hmm, ... maybe. Not entirely - technically I can run a headless publish on my machine as a local target, right? I mean, I've done that just fine many times with e.g. maya.

openpype/hosts/maya/plugins/publish/extract_active_view_thumbnail.py
BigRoy: Just so the `M3DView` image is captured only once. If there's only one workfile instance ever than of course that'd currently always be the case, but I was also prototyping making this so that you can configure the families it'd operate for. We'd likely also just want the scene's thumbnail set on model publishes for example. :)

openpype/hosts/maya/plugins/publish/extract_active_view_thumbnail.py
```diff
@@ -0,0 +1,51 @@
+import maya.api.OpenMaya as om
+import maya.api.OpenMayaUI as omui
+
+import pyblish.api
+import tempfile
```
tokejepsen, line 5:
```suggestion
import tempfile

from openpype.hosts.maya.lib import IS_HEADLESS

```

openpype/hosts/maya/plugins/publish/extract_active_view_thumbnail.py
tokejepsen:
```suggestion
        if IS_HEADLESS:
            self.log.debug(
                "Skip extraction of active view thumbnail, due to being in"
                "headless mode."
            )
            return

        thumbnail = instance.data.get("thumbnailPath")
```

tokejepsen (changes requested):

Let get that headless check in there and we can merge.

openpype/hosts/maya/plugins/publish/extract_active_view_thumbnail.py
```diff
@@ -0,0 +1,61 @@
...
+
+from openpype.hosts.maya.lib import IS_HEADLESS
+
+
+
```
BigRoy, line 10:
```suggestion
```

openpype/hosts/maya/plugins/publish/extract_active_view_thumbnail.py
```diff
@@ -0,0 +1,60 @@
+import maya.api.OpenMaya as om
+import maya.api.OpenMayaUI as omui
+
+import pyblish.api
+import tempfile
+
+from openpype.hosts.maya.lib import IS_HEADLESS
```
tokejepsen, line 7:
```suggestion
from openpype.hosts.maya.api.lib import IS_HEADLESS
```

tokejepsen (changes requested):
Sorry, made a mistake on the imports. Works otherwise.

tokejepsen (approved):
Works in Maya 2023 and code looks good.

mkolar:
LGTM

PR #5427, opened 2023-08-08
Maya: allow not creation of group for Import loaders

kalisp:
## Changelog Description
This PR enhances previous one. All ReferenceLoaders could not wrap imported products into explicit group.
Also `Import` Loaders have same options. Control for this is separate in Settings, eg. Reference might wrap loaded items in group, `Import` might not.

## Testing notes:
1. Remove `_GRP` in `project_settings/maya/load/import_loader`
2. Try to `Import` anything into Maya
3. Try import `abc` representation of `animation`

### DIFF
--- a/openpype/hosts/maya/api/plugin.py
+++ b/openpype/hosts/maya/api/plugin.py
@@ -535,17 +535,20 @@ class Loader(LoaderPlugin):
             loader_key (str): key to get separate configuration from Settings
                 ('reference_loader'|'import_loader')
         """
+        options["attach_to_root"] = True
+
         asset = context['asset']
         subset = context['subset']
         settings = get_project_settings(context['project']['name'])
         custom_naming = settings['maya']['load'][loader_key]
-        options["attach_to_root"] = True
+
         if not custom_naming['namespace']:
             raise LoadError("No namespace specified in "
                             "Maya ReferenceLoader settings")
         elif not custom_naming['group_name']:
             self.log.debug("No custom group_name, no group will be created.")
             options["attach_to_root"] = False
+
         formatting_data = {
             "asset_name": asset['name'],
             "asset_type": asset['type'],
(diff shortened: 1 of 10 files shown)


openpype/hosts/maya/api/plugin.py
```diff
@@ -523,6 +523,49 @@ def remove_instances(self, instances):
...
+                    subset['data']['families'][0]
+            )
+        }
+        custom_namespace = custom_naming['namespace'].format(
+            **formatting_data
+        )
+        custom_group_name = custom_naming['group_name'].format(
+            **formatting_data
+        )
+        return custom_group_name, custom_namespace, options
```
BigRoy, line 567: To me personally the removed empty lines here in the code actually make this function much harder to read than the original. It's cosmetics and it is an opinion. Just wanted to remark that reading this version feels much more cluttered to me than the original. So if you have no preference, let's keep the original. If you prefer this, you're a maniac - haha, no, do whatever you want. 😈

BigRoy (commented):
Code wise this seems fine. I think it's massively overkill to have this amount of customization for an "import loader" since once something is imported it's unmanaged and it's trivial to remove the namespaces and group it - but I guess maybe there are cases where studios import so much stuff all the time that they need to tweak the defaults.

moonyuet:
I tried to import abc representation from animation family but it seems that it doesn't apply the namespace as expected
image

there is no "reference animation"(the animation family loader made from load_animation.py) for loading abc
image

openpype/hosts/maya/api/plugin.py
kalisp: Hmm, this is totally unwanted, probably copy&paste issue. Will fix that.

kalisp: Btw this is client requested, see `sponsored` tag on this PR which denotes that some customer requested that. Who am I to argue :D.

moonyuet:
i tried to test it in the ayon and the setting doesn't show up any host when I set the maya version to 0.1.3, it's working when I switch to 0.1.1(along with other hosts)
image
image

moonyuet (approved):
image
I tried the import options, it seems working as expected. Thanks for helping out on this.

iLLiCiTiT (approved):
Code looks ok, can't test.

PR #5429, opened 2023-08-08
Chore: Default variant in create plugin 

iLLiCiTiT:
## Changelog Description
Attribute `default_variant` on create plugins always returns string and if default variant is not filled other ways how to get one are implemented.

## Additional info
Renamed `default_variant` to `_default_variant`, and make `default_variant` as property. The property has getter, and setter for auto settings apply, the property is calling `get_default_variant` method inside. Method `get_default_variant` uses `_default_variant` if is available, otherwise uses first available value from `default_variants`, or new conststant `DEFAULT_VARIANT_VALUE` if even they are not available.

Added auto-fix of `default_variant` property if a plugin did overide it.

## Testing notes:
- Visually nothing should change.
- All plugins should work as expected, and their settings should be propagated.


openpype/pipeline/create/creator_plugins.py
```diff
@@ -606,17 +621,63 @@ def get_default_variants(self):
...
 
-        Can return `None`. In that case first element from
-        `get_default_variants` should be used.
+        Note:
+            This method does not allow to have empty string as
+                default variant.
+
+        Args:
+            only_explicit (Optional[bool]): If True, only explicit default
+                variant from '_default_variant' will be returned.
```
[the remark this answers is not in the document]
iLLiCiTiT: Debug (I've used it for testing of this PR).

openpype/pipeline/create/creator_plugins.py
```diff
@@ -606,17 +621,63 @@ def get_default_variants(self):
```
fabiaserra, line 621: Typo on docstring:
```
        Replacement of `defaults` attribute. Using method gives ability to
```

kalisp (changes requested):
I would expect to see `Default` in `Variant` field:
AnyDeskMSI_qv25KsFUmY

openpype/pipeline/create/constants.py
```diff
@@ -1,10 +1,12 @@
 SUBSET_NAME_ALLOWED_SYMBOLS = "a-zA-Z0-9_."
 DEFAULT_SUBSET_TEMPLATE = "{family}{Variant}"
 PRE_CREATE_THUMBNAIL_KEY = "thumbnail_source"
+DEFAULT_VARIANT_VALUE = "main"
```
iLLiCiTiT, line 4:
```suggestion
DEFAULT_VARIANT_VALUE = "Main"
```

PR #5434, opened 2023-08-09
Max: Bug Fix  Alembic Loaders with Ornatrix

moonyuet:
## Changelog Description
Bugfix the alembic loader with both ornatrix alembic and max alembic supports
Add the ornatrix alembic loaders for loading the alembic with Ornatrix-related modifiers.
## Additional info
Tested the loaders with model, camera, pointcache, animation families
Both alembic loaders can load ornatrix alembic.
The original alembic loaders would take longer time to load the alembics and it will cause possible 3dsmax crash. 

With the original loaders, it will convert the modifiers into the curves modifiers. (It can take several minutes to load and nurb curve conversion)

https://github.com/ynput/OpenPype/assets/64118225/b84a451a-961a-443b-b98e-a7a4fbd3b7c1

With the ornatrix loaders, It will be loaded faster with OxBaked-related modifiers(i.e. without any conversion)

https://github.com/ynput/OpenPype/assets/64118225/3dd2fdb8-7d8c-414e-bf9e-4fbb26169e12

## Testing notes:
1. Launch Max via OP
2. OP Menu -> Load...
3. Load Alembic
4. The alembic should be loaded as expected.


moonyuet:
> I did successfully tested to use normal `Alembic Load` without any issues.
> 
> However when trying to use `Load Ornatrix Alembic` it gives me error that I dont have Ornatrix plugin installed...even tho I have it running as seen from torus and hair present.
> 
> image
> 
> Any ideas whats going on?
> 
> Regarding testing the Abc Loading...I should try to load the normal alembics as published from other OP project assets, right? Because we lack publish instance for native Ornatrix right now.
> 
> Could u enlight it for me?

This is quite weird scanerio. I need to inverstigate a little bit.

LiborBatek (changes requested):
Now its possible to use standard ```Load Alembic``` and also ```Load Alembic with Ornatrix``` without missing addon error.

It correctly loads cache into scene even tho when using Ornatrix loader its empty. No errors shown tho.
image

When first loading the asset via standard ```Load Alembic``` and then trying the Ornatrix one, it fails to load it with following error:
image

Seems like the previously loaded asset (same animation asset version) interfere with it (also see outliner for broken Containers with Ornatrix outside of it)

moonyuet:
@LiborBatek I have updated the PR when the bugfix you mentioned but it will come up with namespace issue( which is fixed by https://github.com/ynput/OpenPype/pull/5378) if you already load one object with the same family.

LiborBatek (changes requested):
When performing standard ```Load Alembic``` it gives me following error now (seems valid also when doing ```Load Alembic with Ornatrix``` )
image

moonyuet:
> When performing standard `Load Alembic` it gives me following error now (seems valid also when doing `Load Alembic with Ornatrix` ) image

Should be fixed now. Can you have one more test?

LiborBatek (approved):
All seems working normally.
image

antirotor (changes requested):
looks good, I would just change the Exception.

moonyuet:
@antirotor i have updated the PR according to your comment. Thank you!

antirotor (approved):

PR #5438, opened 2023-08-10
Blender: Deadline support

simonebarbieri:
## Changelog Description
Add Deadline support for Blender.


LiborBatek (changes requested):
I have prepared lighting scene with loaded ```pointcache``` animation, created ```renderLayerMain``` instance and hit ```Publish```

I m getting following errors:
image

Also as experienced user has to select ```camera``` when trying to create ```render``` instance if not anything selected I get following  during creation error:
image

This differs from ```Maya``` host so not sure if it should be mandatory. I can imagine that there should be ```validator``` for render camera during publishing process.

LiborBatek:
As tested now it correctly upversion the ```workfile``` after successful ```Publish```.

Right now OP forces ```Image Resolution``` from ```Project``` database entry. It is preferable to let user change it to any resolution if needed.

There should be optional ```Render Resolution```  ```Validator``` similar to ```Validate Frame Range``` so user can override it if need to.

Right now this is not possible.

I have set HalfHd resolution in blender
image

but my rendered images are FullHd.

LiborBatek:
Btw there are still working render files left in work lighting folder for AOVs
image

Btw not having clue why we create those AOVs as separate files when ```multilayer EXR``` set. Would be nice to have those just when ```multilayer``` not used.

openpype/hosts/blender/plugins/create/create_render.py
```diff
@@ -0,0 +1,302 @@
...
+
+import bpy
+
+from openpype.pipeline import (
+    get_current_project_name,
+)
+from openpype.settings import (
+    get_project_settings,
+)
+from openpype.pipeline import get_current_task_name
```
iLLiCiTiT, line 12:
```suggestion
from openpype.settings import get_project_settings
from openpype.pipeline import (
    get_current_project_name,
    get_current_task_name,
)
```

openpype/hosts/blender/plugins/create/create_render.py
```diff
@@ -0,0 +1,302 @@
...
+        Args:
+            file_path (str): The path to the blender scene.
+            render_folder (str): The render folder set in settings.
+            file_name (str): The name of the blender scene.
+            instance (pyblish.api.Instance): The instance to publish.
+            ext (str): The image format to render.
+        """
+        output_file = os.path.join(output_path, name)
+
+        render_product = f"{output_file}.####"
```
Minkiu, line 81: This could be
```py
render_product = f"{os.path.join(output_path, name)}.####"
```

openpype/hosts/blender/plugins/create/create_render.py
```diff
@@ -0,0 +1,302 @@
...
+    def prepare_rendering(self, asset_group, name):
+        filepath = bpy.data.filepath
+        assert filepath, "Workfile not saved. Please save the file first."
+
+        file_path = os.path.dirname(filepath)
+        file_name = os.path.basename(filepath)
+        file_name, _ = os.path.splitext(file_name)
+
+        project = get_current_project_name()
+        settings = get_project_settings(project)
```
BigRoy, line 235: Settings should be retrieved in the `apply_settings` method. Right?

openpype/hosts/blender/plugins/create/create_render.py
```diff
@@ -0,0 +1,302 @@
...
+        subset = self.data["subset"]
+        name = plugin.asset_name(asset, subset)
+        asset_group = bpy.data.collections.new(name=name)
+
+        try:
+            instances.children.link(asset_group)
+            self.data['task'] = get_current_task_name()
+            lib.imprint(asset_group, self.data)
+
+            self.prepare_rendering(asset_group, name)
```
BigRoy, line 285: Note that the instance will be created - and imprinted even if `prepare_rendering` fails due to the file not being saved. Would it make more sense to perform the "ensure file is saved" check at the beginning of `process` to stop immediately before doing anything?

openpype/hosts/blender/plugins/create/create_render.py
```diff
@@ -0,0 +1,302 @@
...
+        # TODO: this is undesiderable, but it's the only way to be sure that
+        # the file is saved before the render starts.
+        # Blender, by design, doesn't set the file as dirty if modifications
+        # happen by script. So, when creating the instance and setting the
+        # render settings, the file is not marked as dirty. This means that
+        # there is the risk of sending to deadline a file without the right
+        # settings. Even the validator to check that the file is saved will
+        # detect the file as saved, even if it isn't. The only solution for
+        # now it is to force the file to be saved.
+        bpy.ops.wm.save_as_mainfile(filepath=bpy.data.filepath)
```
BigRoy, line 300: Would it work to just force `bpy.data.is_dirty = True`?

simonebarbieri: Yeah, I've tried everything, but I found that same post ([and others with Blender developers comments saying that it is like that by design](https://devtalk.blender.org/t/how-to-get-bpy-data-is-dirty-updating-when-changes-are-done-by-script/4852/2)), so I don't think there's any other way, for now :/

openpype/hosts/blender/plugins/create/create_render.py
simonebarbieri: Hound will complain about the second one for it being too long 😅 but the first one works, I'll change it :)

openpype/hosts/blender/plugins/create/create_render.py
simonebarbieri: The problem is that because of Blender limitations, the file is just saved, we cannot check if there are changes (unless they are made via UI). The `try` is there to remove the instance if any other error arises (I think there was a complaint about that, that if it fails for any reason, the instance must be manually deleted).

openpype/hosts/blender/plugins/create/create_render.py
simonebarbieri: I'm just finding out about this method 😅 My bad, I'll change it!

BigRoy: Wait! :) I think Blender is still using the legacy publisher and thus this might not work for Blender.

simonebarbieri: Mmmh, I see that the legacy creator has an `apply_settings` method as well, but both require the settings as parameters, so they must be retrieved before anyway! And yeah, I cannot find any old creator that uses is, so for now I'll leave it like this and maybe we can change it when we implement the new publisher in Blender :)

openpype/hosts/blender/api/colorspace.py
```diff
@@ -0,0 +1,52 @@
...
+        Returns:
+            list: List of RenderProduct
+        """
+        return [
+            RenderProduct(
+                colorspace="sRGB",
+                view="ACES 1.0",
+                productName=""
+            )
+        ]
```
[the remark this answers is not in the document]
simonebarbieri: I have to admit that I know too little about this and it is "heavily ispired" by the 3dsmax and houdini implementation 😅

LiborBatek:
There is small glitch with ```Validate Render Output for Deadline``` that ```max``` being shown even this being in ```blender``` host...need to be changed
image

LiborBatek (approved):
All working fine and without any glitches, producing desired renders and reviewables with burnins...

image

LiborBatek:
So there actually some issues regarding using other file formats besides EXR...need to be addressed first

LiborBatek (changes requested):
fully support other file formats besides EXR, like PNGs, Tiff etc.

Now not fully working.

openpype/hosts/blender/plugins/publish/validate_deadline_publish.py
```diff
@@ -0,0 +1,47 @@
...
+    def process(self, instance):
+        if not self.is_active(instance.data):
+            return
+        filepath = bpy.data.filepath
+        file = os.path.basename(filepath)
+        filename, ext = os.path.splitext(file)
+        if filename not in bpy.context.scene.render.filepath:
+            raise PublishValidationError(
+                "Render output folder "
+                "doesn't match the max scene name! "
```
BigRoy, line 37:
```suggestion
                "doesn't match the blender scene name! "
```

BigRoy: Also, shouldn't this plugin be in the Deadline module?

openpype/hosts/blender/api/colorspace.py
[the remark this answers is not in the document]
iLLiCiTiT: Use your head and don't copy paste blindly :)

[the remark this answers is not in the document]
moonyuet: > I have to admit that I know too little about this and it is "heavily ispired" by the 3dsmax and houdini implementation 😅

It's actually my bad, I implement the scripts as Max doesn't have colorspaces with 2023(but with 2024), and it needs to have some hard-coded data for both colorspace and renderproducts in order to submit the publish job. It shouldn't be in Houdini though.

LiborBatek (changes requested):
So I have tested to set ```Blender``` and its OP Render Setttings to other format than ```Exr``` to ```PNG``` and created few ```AOVs``` as seen on pic
image

...in blender all good and properly setup when creating fresh new ```Render``` instance

image

Job is successfuly rendered on DL as seen here:
image

Problem being that DL ```Publish Job``` didnt correctly produced ```review``` mp4 just rendered frames in publish folder.
image

Also question being wether we should version each ```AOVs``` separately....

In my case, Im having these:

```Main_beauty_v06```
```Main_Normal_v01```
```Main_Depth_v01```
etc.

IMHO not very practical to have different ```version``` even the rendered frames originating from the same ```workfile``` lighting scene. Should be united under same version...

Lastly, we should also have option if ```Noisy Image``` being produced when rendering too....not very practical to have it everytime even when not having accompanying ```AOVs``` turned on for later possibility to perform denoise. Most of the cases not desired to produce just ```Noisy Image``` without those additional passes needed for post denoise.

**Comment by LiborBatek** [comment]
Here is the ```Publish job``` on DL log:
[Job_2023-09-21_08-37-12_650be498530d594b3d552485.txt](https://github.com/ynput/OpenPype/files/12681528/Job_2023-09-21_08-37-12_650be498530d594b3d552485.txt)

File: openpype/modules/deadline/plugins/publish/submit_blender_deadline.py [hunk 8]
```diff
@@ -0,0 +1,181 @@
...
+
+        environment = dict({key: os.environ[key] for key in keys
+                            if key in os.environ}, **legacy_io.Session)
+
+        for key in keys:
+            value = environment.get(key)
+            if not value:
+                continue
+            job_info.EnvironmentKeyValue[key] = value
+
```
moonyuet, line 124: you may need to add `job_info.add_render_job_env_var()` as it is added in other deadline submission plugin like maya and max. 
https://github.com/ynput/OpenPype/blob/4b1c9077e6c0f109c3315ddc88dd74f09a301bfe/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py#L229-L230

LiborBatek (approved):

BigRoy:
Didn't do a review - but just wanted to say it's a massive achievement this work has been implemented. Great job! 🥳

LiborBatek:
@BigRoy I hope this one will serve for good! Let' s see how this one works in prod!

PR #5442, opened 2023-08-10
Maya: Fix wrong subset name of render family in deadline

kalisp:
## Changelog Description
New Publisher is creating different subset names than previously which resulted in duplication of `render` string in final subset name of `render` family published on Deadline.
This PR solves that, it also fixes issues with legacy instances from old publisher, it matches the subset name as was before.

This solves same issue in Max implementation.

## Additional info
`renderModelingMain_beauty` in old publisher >> `renderModelingRenderingMain_beauty`
AnyDeskMSI_56csq5vzNi

## Testing notes:
1. create new render instance, publish to deadline, check subset name (folder name)
2. use old workfile with instances created in old publisher, use `Convert legacy` to re-create instance, check subset names

### DIFF
--- a/openpype/hosts/maya/api/plugin.py
+++ b/openpype/hosts/maya/api/plugin.py
@@ -22,8 +22,6 @@ from openpype.pipeline import (
     LegacyCreator,
     LoaderPlugin,
     get_representation_path,
-
-    legacy_io,
 )
 from openpype.pipeline.load import LoadError
 from openpype.client import get_asset_by_name
--- a/openpype/hosts/maya/api/plugin.py
+++ b/openpype/hosts/maya/api/plugin.py
@@ -26,6 +26,8 @@ from openpype.pipeline import (
     legacy_io,
 )
 from openpype.pipeline.load import LoadError
+from openpype.client import get_asset_by_name
+from openpype.pipeline.create import get_subset_name
 
 from . import lib
 from .lib import imprint, read
@@ -405,14 +407,21 @@ class RenderlayerCreator(NewCreator, MayaCreatorBase):
                 # No existing scene instance node for this layer. Note that
                 # this instance will not have the `instance_node` data yet
                 # until it's been saved/persisted at least once.
-                # TODO: Correctly define the subset name using templates
-                prefix = self.layer_instance_prefix or self.family
-                subset_name = "{}{}".format(prefix, layer.name())
+                project_name = self.create_context.get_current_project_name()
                 instance_data = {
-                    "asset": legacy_io.Session["AVALON_ASSET"],
-                    "task": legacy_io.Session["AVALON_TASK"],
+                    "asset": self.create_context.get_current_asset_name,
+                    "task": self.create_context.get_current_task_name,
                     "variant": layer.name(),
                 }
+                asset_doc = get_asset_by_name(project_name,
+                                              instance_data["asset"])
+                subset_name = get_subset_name(
+                    self.layer_instance_prefix,
+                    layer.name(),
+                    instance_data["task"],
+                    asset_doc,
+                    project_name)
+
                 instance = CreatedInstance(
                     family=self.family,
                     subset_name=subset_name,
--- a/openpype/hosts/maya/api/plugin.py
+++ b/openpype/hosts/maya/api/plugin.py
@@ -413,8 +413,7 @@ class RenderlayerCreator(NewCreator, MayaCreatorBase):
                 }
                 asset_doc = get_asset_by_name(project_name,
                                               instance_data["asset"])
-                subset_name = get_subset_name(
-                    self.layer_instance_prefix,
+                subset_name = self.get_subset_name(
                     layer.name(),
                     instance_data["task"],
                     asset_doc,
@@ -525,6 +524,22 @@ class RenderlayerCreator(NewCreator, MayaCreatorBase):
                 node = instance.data.get("instance_node")
                 if node and cmds.objExists(node):
                     cmds.delete(node)
+
+    def get_subset_name(
+        self,
+        variant,
+        task_name,
+        asset_doc,
+        project_name,
+        host_name=None,
+        instance=None
+    ):
+        # creator.family != 'render' as expected
+        return get_subset_name(self.layer_instance_prefix,
+                               variant,
+                               task_name,
+                               asset_doc,
+                               project_name)
 
 
 class Loader(LoaderPlugin):
--- a/openpype/hosts/maya/api/plugin.py
+++ b/openpype/hosts/maya/api/plugin.py
@@ -406,6 +406,7 @@ class RenderlayerCreator(NewCreator, MayaCreatorBase):
                 # this instance will not have the `instance_node` data yet
                 # until it's been saved/persisted at least once.
                 project_name = self.create_context.get_current_project_name()
+
                 instance_data = {
                     "asset": self.create_context.get_current_asset_name(),
                     "task": self.create_context.get_current_task_name(),
--- a/openpype/hosts/maya/api/plugin.py
+++ b/openpype/hosts/maya/api/plugin.py
@@ -407,8 +407,8 @@ class RenderlayerCreator(NewCreator, MayaCreatorBase):
                 # until it's been saved/persisted at least once.
                 project_name = self.create_context.get_current_project_name()
                 instance_data = {
-                    "asset": self.create_context.get_current_asset_name,
-                    "task": self.create_context.get_current_task_name,
+                    "asset": self.create_context.get_current_asset_name(),
+                    "task": self.create_context.get_current_task_name(),
                     "variant": layer.name(),
                 }
                 asset_doc = get_asset_by_name(project_name,
--- a/openpype/hosts/maya/plugins/create/convert_legacy.py
+++ b/openpype/hosts/maya/plugins/create/convert_legacy.py
@@ -2,6 +2,10 @@ from openpype.pipeline.create.creator_plugins import SubsetConvertorPlugin
 from openpype.hosts.maya.api import plugin
 from openpype.hosts.maya.api.lib import read
 
+from openpype.client import get_asset_by_name
+from openpype.pipeline.create import get_subset_name
+from openpype.pipeline import get_current_project_name
+
 from maya import cmds
 from maya.app.renderSetup.model import renderSetup
 
@@ -134,6 +138,19 @@ class MayaLegacyConvertor(SubsetConvertorPlugin,
             # The family gets converted to the new family (this is due to
             # "rendering" family being converted to "renderlayer" family)
             original_data["family"] = creator.family
+
+
+            # recreate subset name as without it would be
+            # `renderingMain` vs correct `renderMain`
+            project_name = get_current_project_name()
+            asset_doc = get_asset_by_name(project_name,
+                                          original_data["asset"])
+            subset_name = get_subset_name("render",
+                                          original_data["variant"],
+                                          data["task"],
+                                          asset_doc,
+                                          project_name)
+            original_data["subset"] = subset_name
 
             # Convert to creator attributes when relevant
             creator_attributes = {}
--- a/openpype/hosts/maya/plugins/create/convert_legacy.py
+++ b/openpype/hosts/maya/plugins/create/convert_legacy.py
@@ -4,7 +4,6 @@ from openpype.hosts.maya.api.lib import read
 
 from openpype.client import get_asset_by_name
 from openpype.pipeline.create import get_subset_name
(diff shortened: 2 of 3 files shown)


openpype/hosts/maya/api/plugin.py
```diff
@@ -405,14 +405,21 @@ def collect_instances(self):
...
                 # until it's been saved/persisted at least once.
-                # TODO: Correctly define the subset name using templates
-                prefix = self.layer_instance_prefix or self.family
-                subset_name = "{}{}".format(prefix, layer.name())
+                project_name = self.create_context.get_current_project_name()
                 instance_data = {
-                    "asset": legacy_io.Session["AVALON_ASSET"],
-                    "task": legacy_io.Session["AVALON_TASK"],
+                    "asset": self.create_context.get_current_asset_name,
+                    "task": self.create_context.get_current_task_name,
```
BigRoy, line 411: I assume these are functions like `get_current_project_name()` above?
```suggestion
                    "asset": self.create_context.get_current_asset_name(),
                    "task": self.create_context.get_current_task_name(),
```

openpype/hosts/maya/api/plugin.py
```diff
@@ -405,14 +405,21 @@ def collect_instances(self):
...
                     "variant": layer.name(),
                 }
+                asset_doc = get_asset_by_name(project_name,
+                                              instance_data["asset"])
+                subset_name = get_subset_name(
```
BigRoy, line 416: Shouldn't this be using ]`self.get_subset_name()`](https://github.com/ynput/OpenPype/blob/develop/openpype/pipeline/create/creator_plugins.py)?

openpype/hosts/maya/api/plugin.py
kalisp: Fixed, wtf why it was working when I was testing it.

openpype/hosts/maya/api/plugin.py
kalisp: `self.get_subset_name` uses `self.family` which is `renderLayer` and not `render` as I would expect. 

Situation with families in Maya (`rendering`, `renderLayer`, `render`) is confusing to say the least.

iLLiCiTiT: Shouldn't be in that case overriden the method to use right value, so if the method is called out of the inner plugin logic you'll get the same result.

openpype/hosts/maya/plugins/create/convert_legacy.py
```diff
@@ -135,6 +139,18 @@ def _convert_per_renderlayer(self, instance_nodes, data, creator):
             # "rendering" family being converted to "renderlayer" family)
             original_data["family"] = creator.family
 
+            # recreate subset name as without it would be
+            # `renderingMain` vs correct `renderMain`
+            project_name = get_current_project_name()
```
iLLiCiTiT, line 144:
```suggestion
            project_name = self.create_context.get_current_project_name()
```

openpype/pipeline/farm/pyblish_functions.py
```diff
@@ -568,9 +568,15 @@ def _create_instances_for_aov(instance, skeleton, aov_filter, additional_data,
...
-        group_name = 'render{}{}{}{}'.format(
-            task[0].upper(), task[1:],
-            subset[0].upper(), subset[1:])
+        # TODO reactor/remove me
+        family = skeleton["family"]
+        if not subset.startswith(family):
+            group_name = '{}{}{}'.format(
+                family,
+                task.capitalize(),
+                task.capitalize())
```
iLLiCiTiT, line 577:
```suggestion
                subset.capitalize())
```

openpype/hosts/maya/plugins/create/convert_legacy.py
```diff
@@ -135,6 +139,18 @@ def _convert_per_renderlayer(self, instance_nodes, data, creator):
...
+            asset_doc = get_asset_by_name(project_name,
+                                          original_data["asset"])
+            subset_name = get_subset_name("render",
+                                          original_data["variant"],
+                                          data["task"],
+                                          asset_doc,
+                                          project_name)
```
iLLiCiTiT, line 151:
```suggestion
            subset_name = creator.get_subset_name(
                original_data["variant"],
                data["task"],
                asset_doc,
                project_name)
```

BigRoy:
> I got the error below when creating the render instance
> 
> ```
> Creator: Render (io.openpype.creators.maya.renderlayer)
> 
> Error: cannot encode object: <bound method CreateContext.get_current_asset_name of <openpype.pipeline.create.context.CreateContext object at 0x000002BA91F87A30>>, of type: <class 'method'>
> 
> Traceback (most recent call last):
>   File "D:\kayla\OpenPype\openpype\pipeline\create\context.py", line 2030, in _create_with_unified_error
>     result = creator.create(*args, **kwargs)
>   File "D:\kayla\OpenPype\openpype\hosts\maya\plugins\create\create_render.py", line 50, in create
>     super(CreateRenderlayer, self).create(subset_name,
>   File "D:\kayla\OpenPype\openpype\hosts\maya\api\plugin.py", line 376, in create
>     self.collect_instances()
>   File "D:\kayla\OpenPype\openpype\hosts\maya\api\plugin.py", line 414, in collect_instances
>     asset_doc = get_asset_by_name(project_name,
>   File "D:\kayla\OpenPype\openpype\client\mongo\entities.py", line 197, in get_asset_by_name
>     return conn.find_one(query_filter, _prepare_fields(fields))
>   File "D:\kayla\OpenPype\.venv\lib\site-packages\pymongo\collection.py", line 1491, in find_one
>     for result in cursor.limit(-1):
>   File "D:\kayla\OpenPype\.venv\lib\site-packages\pymongo\cursor.py", line 1280, in next
>     if len(self.__data) or self._refresh():
>   File "D:\kayla\OpenPype\.venv\lib\site-packages\pymongo\cursor.py", line 1195, in _refresh
>     self.__send_message(q)
>   File "D:\kayla\OpenPype\.venv\lib\site-packages\pymongo\cursor.py", line 1078, in __send_message
>     response = client._run_operation(
>   File "D:\kayla\OpenPype\.venv\lib\site-packages\pymongo\mongo_client.py", line 1515, in _run_operation
>     return self._retryable_read(
>   File "D:\kayla\OpenPype\.venv\lib\site-packages\pymongo\mongo_client.py", line 1617, in _retryable_read
>     return func(session, server, sock_info, secondary_ok)
>   File "D:\kayla\OpenPype\.venv\lib\site-packages\pymongo\mongo_client.py", line 1511, in _cmd
>     return server.run_operation(
>   File "D:\kayla\OpenPype\.venv\lib\site-packages\pymongo\server.py", line 99, in run_operation
>     message = operation.get_message(set_secondary_okay, sock_info, use_cmd)
>   File "D:\kayla\OpenPype\.venv\lib\site-packages\pymongo\message.py", line 398, in get_message
>     request_id, msg, size, _ = _op_msg(
>   File "D:\kayla\OpenPype\.venv\lib\site-packages\pymongo\message.py", line 833, in _op_msg
>     return _op_msg_uncompressed(flags, command, identifier, docs, check_keys, opts)
> bson.errors.InvalidDocument: cannot encode object: <bound method CreateContext.get_current_asset_name of <openpype.pipeline.create.context.CreateContext object at 0x000002BA91F87A30>>, of type: <class 'method'>
> ```

I'm pretty sure that was the code issue I commented on [here](https://github.com/ynput/OpenPype/pull/5442#discussion_r1290244129)

openpype/hosts/maya/api/plugin.py
BigRoy: Did you push the fix?

openpype/hosts/maya/api/plugin.py
BigRoy: Yup, like @iLLiCiTiT mentions. If it returns the wrong value then I suspect we override it on that creator plugin so it returns the right value instead. (There might be other things calling that method on the Creator and expecting the correct subset name!)

openpype/pipeline/farm/pyblish_functions.py
BigRoy: Pretty sure capitalize is NOT what you want. This is introducing bugs or at least very different behavior than before.

```python
"myStringVlue Hello".capitalize()
# Result: Mystringvlue hello # 
```

Instead of only uppercasing the first character this now ALSO lowercases all other characters.

openpype/pipeline/farm/pyblish_functions.py
```diff
@@ -568,9 +568,15 @@ def _create_instances_for_aov(instance, skeleton, aov_filter, additional_data,
             col = list(cols[0])
 
         # create subset name `familyTaskSubset_AOV`
-        group_name = 'render{}{}{}{}'.format(
-            task[0].upper(), task[1:],
-            subset[0].upper(), subset[1:])
+        # TODO reactor/remove me
```
BigRoy, line 571:
```suggestion
        # TODO refactor/remove me
```

openpype/hosts/maya/plugins/create/convert_legacy.py
kalisp: This won't work as `creator.family` != `render` because I don't know why.

openpype/pipeline/farm/pyblish_functions.py
kalisp: Pwned, so this is why there was that weird construct before, reverted back.

openpype/hosts/maya/api/plugin.py
kalisp: It should have been pushed with https://github.com/ynput/OpenPype/pull/5442/commits/e6a847efb6425c266d6c7800b9a64a113bac2d18

I was pulling this branch from scratch and testing it right now and it works, please try again.

openpype/hosts/maya/api/plugin.py
kalisp: Ok, modified.

openpype/hosts/maya/plugins/create/convert_legacy.py
iLLiCiTiT: Now it does

BigRoy: This suggestion needs to still be committed.

openpype/hosts/maya/plugins/create/convert_legacy.py
```diff
@@ -135,6 +138,18 @@ def _convert_per_renderlayer(self, instance_nodes, data, creator):
...
+            project_name = self.create_context.get_current_project_name()
+            asset_doc = get_asset_by_name(project_name,
+                                          original_data["asset"])
+            subset_name = get_subset_name("render",
+                                          original_data["variant"],
+                                          data["task"],
+                                          asset_doc,
+                                          project_name)
+            original_data["subset"] = subset_name
+
```
BigRoy, line 152: If [this](https://github.com/ynput/OpenPype/pull/5442/files#r1291614953) is true then this logic can all be removed.
```suggestion
```

openpype/pipeline/farm/pyblish_functions.py
```diff
@@ -568,9 +568,15 @@ def _create_instances_for_aov(instance, skeleton, aov_filter, additional_data,
...
-            subset[0].upper(), subset[1:])
+        # TODO refactor/remove me
+        family = skeleton["family"]
+        if not subset.startswith(family):
+            group_name = '{}{}{}{}{}'.format(
+                family,
+                task[0].upper(), task[1:],
+                subset[0].upper(), subset[1:])
+        else:
+            group_name = subset
```
BigRoy, line 579: Not sure about this logic change. I mean, admittedly, it's weird that originally this was hardcoded to `render` prefix. However, the new logic doesn't appear much more logical to me either. The fact that subset starts with family, isn't that a template rule in settings too?

kalisp: I think that the whole part with concatenating group_name should be removed altogether and depend only on setting subset_name in the creator. I assume that this logic expected subset_name to be `Main` only which is a bit weird.

openpype/hosts/maya/plugins/create/convert_legacy.py
kalisp: No, still needed, without it it will produce `renderingMain` subset name

BigRoy: Somehow that sounds like a different bug then - I'd expect the logic to retrieve the subset name from the creator - instead of making one up randomly?

LiborBatek (approved):
Did test both scenario and both are working ok!

All file names / folders matches the original way of naming convetions.

image

PR #5444, opened 2023-08-11
Nuke: Name of the Read Node should be updated correctly when switching versions or assets.

moonyuet:
## Changelog Description
Bug fixing of the read node's name not being updated correctly when setting version or switching asset.

## Additional info
n/a

## Testing notes:
1. Go to OP Setting `project_settings/nuke/load/LoadImage/node_name_template`
2. Set node_name template with the varying data by version/asset, such as `{class_name}_{id}_{ext}`
image

3. Launch Nuke with Launcher/OP
4. Access "Load.." via AYON/OP Menu
5. Load image with the families such as plateMain, reviewMain etc...
6. Access "Manage.." via AYON/OP Menu
7. Right click to set version or switch asset
8. The name should be changed as expected

### DIFF
--- a/openpype/hosts/nuke/plugins/load/load_image.py
+++ b/openpype/hosts/nuke/plugins/load/load_image.py
@@ -96,7 +96,8 @@ class LoadImage(load.LoaderPlugin):
 
         file = file.replace("\\", "/")
 
-        repr_cont = context["representation"]["context"]
+        representation = context["representation"]
+        repr_cont = representation["context"]
         frame = repr_cont.get("frame")
         if frame:
             padding = len(frame)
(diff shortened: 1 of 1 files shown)


openpype/hosts/nuke/plugins/load/load_image.py
```diff
@@ -250,3 +252,17 @@ def remove(self, container):
 
         with viewer_update_and_undo_stop():
             nuke.delete(node)
+
+    def _get_node_name(self, representation):
```
BigRoy, line 256: Since we're now implementing a function for this - shouldn't we also re-use it [on `load`](https://github.com/ynput/OpenPype/pull/5444/files#diff-1f8459817a04b371e8fb325f5768d8294c6754ee1cfaf893326ffdcc4cbab662R107-R116) to ensure the behavior stays the same?

moonyuet: Make sense, I have aligned this behavior in the updated PR

BigRoy:
> [roy's comment](https://github.com/ynput/OpenPype/pull/5444/commits/43796c2c1c14fee0f33e8d1e2480deb0e3c19256)

By the way, I think a commit message solely stating "roy's comment" is much less informational than what change you've actually made. I'd expect a commit message like "Reuse `_get_node_name` method in `load` logic" since it's more descriptive. Not a big issue - but it'd be my preference in the future. (Of course check with rest of Ynput crew whether they agree or not since it's just my opinion). If you'd want to highlight it's my feedback then -> "Reuse `_get_node_name` method in `load` logic (roy's comment)" would be fine of course.

jakubjezek001 (approved):
LGTM

PR #5445, opened 2023-08-11
Maya: implement matchmove publishing

kalisp:
## Changelog Description
Add possibility to export multiple cameras in single `matchmove` family instance, both in `abc` and `ma`.

Exposed flag 'Keep image planes' to control export of image planes.

## Additional info
Switch to `matchmove` from `camera` family was chosen to don't break existing workflow where only single camera is expected (Houdini, Nuke). It is important to note, that alembic and maya scenes don't support identical feature, so it is possible to have richer matchmove data in .ma, compared to .abc format. For instance Image planes will not be preserved in .abc file. 

## Testing notes:
1.create multiple cameras in Maya
2. select all of them and Create `matchmove` instance
3. publish
4. try to load both via `Reference (abc)` and `Referenca (ma)`, it should load all cameras (previously it would load only first camera even if you would select multiple for Creation step).


BigRoy (commented):
Hmm - I don't think I'll be approving this from my end. This seems to go against the whole concept of what the `camera` family is designed to be.

This just seems off. The whole intent for the camera family is to be a single camera - or maybe also a stereo camera (technically two lenses) but even then I'd say it's likely better to make that a `stereoCamera` family to be explicit about the content.

What's the reason you need multiple cameras in a single camera instance as opposed to you just publishing multiple camera instances?

Or, if you want to put any amount of cameras into an instance without rules - why not use a pointcache instance?

The Houdini (loading) logic for example is also expecting it currently to be a single camera and I anticipate other hosts to be doing the same.

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
```diff
@@ -165,7 +164,8 @@ def process(self, instance):
                                                long=True)
 
                         members = members + baked_camera_shapes
-                        members.remove(camera)
+                        for camera in cameras:
+                            members.remove(camera)
```
BigRoy, line 168: Depending on the amount of cameras you might be expecting this could be quite a slow iteration.

This could then be much faster IF cameras would be a `set`
```members = [member for member in members if member not in cameras]
```

Or:
```python
members = members - cameras
```
If both are sets.

tokejepsen:
You could have a family called "cameras" that'll publish each camera collected as "camera" family.

m-u-r-p-h-y:
> > Hmm - I don't think I'll be approving this from my end. This seems to go against the whole concept of what the `camera` family is designed to be.
> 
> I'll let @m-u-r-p-h-y or @antirotor reply to this. But thanks for valid concerns, I was expecting them.


If this family is meant for one camera only,(we should check if we have single camera only present) and we need a multicamera family to cover the use case of publishing dozens or even hundreds of cameras used for photogrammetry model reconstruction (including corresponding image planes). 


image
(I could not find a better picture quickly to demonstrate it :)

BigRoy:
Pointcache family can export any amount of cameras + include geo or locators if needed.

openpype/hosts/maya/plugins/publish/extract_camera_alembic.py
```diff
@@ -9,14 +9,17 @@
 class ExtractCameraAlembic(publish.Extractor):
     """Extract a Camera as Alembic.
 
-    The cameras gets baked to world space by default. Only when the instance's
+    The camera gets baked to world space by default. Only when the instance's
     `bakeToWorldSpace` is set to False it will include its full hierarchy.
 
+    'camera' family expects only single camera, if multiple cameras are needed,
+    'matchmove' is better choice.
```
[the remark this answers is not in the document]
kalisp: No one is reading documentation ;). And I think there should be some automatic documentation creation from docstrings, there might be discussion needed in format (artist/developer docstring).

kalisp:
Single camera family stays the same, `matchmove` should be used instead. This family allows instance to contain multiple cameras/geometries etc.

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
```diff
@@ -131,15 +134,14 @@ def process(self, instance):
...
+        members = set(cmds.ls(instance.data['setMembers'], leaf=True,
+                      shapes=True, long=True, dag=True))
+        cameras = set(cmds.ls(members, leaf=True, shapes=True, long=True,
+                      dag=True, type="camera"))
 
         # validate required settings
         assert isinstance(step, float), "Step must be a float value"
-        camera = cameras[0]
-        transform = cmds.listRelatives(camera, parent=True, fullPath=True)
+        transform = cmds.listRelatives(cameras, parent=True, fullPath=True)
```
BigRoy, line 144: Note that older versions of Maya can't deal receiving `set` as inputs. I believe it's Maya 2020 and below, but not entirely sure.

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
```diff
@@ -158,14 +160,14 @@ def process(self, instance):
...
-
-                        members = members + baked_camera_shapes
-                        members.remove(camera)
+                        baked_camera_shapes = set(cmds.ls(baked,
+                                                  type="camera",
+                                                  dag=True,
+                                                  shapes=True,
+                                                  long=True))
+
+                        members = members.union(baked_camera_shapes)
```
BigRoy, line 169: No need to return a new set if you're assigning it to the same variable anyway.
```suggestion
                        members.update(baked_camera_shapes)
```

BigRoy (commented):
I understand the thought process somewhat but I can't see yet why this currently requires to be a family of its own. What does `matchmove` do as a family that `pointcache` does not?

Maybe I'm not sure of the requirements of `matchmove` content but it doesn't seem to offer much different than the bake to world space option + that it exports both `.ma` and `.abc` - which does beg the question - for what specifically would you be needing the `.ma` for matchmove content anyway?

Can we lay out the rules of what `matchmove` family is supposed to be and why it warrants being its own family?

mkolar:
> I understand the thought process somewhat but I can't see yet why this currently requires to be a family of its own. What does `matchmove` do as a family that `pointcache` does not?

It "does" pretty much the same thing, but it portrays a very different message. Families are not only about their technical specifications, but also human understanding and `matchmove` implies cameras and associated data. `pointcache` implies...well....mostly geometry.

Same reason why we have both `mayaScene` and `setdress` same technical requirements, but very different feel within the pipeline.

moonyuet:
the publish works smoothly as far as I check.
Just wondering if there is any content validator to check the setmembers including only camera type as it seems the camera validator isn't included before publishing.

BigRoy:
> Just wondering if there is any content validator to check the setmembers including only camera type as it seems the camera validator isn't included before publishing.

I think the issue here is that the family **matchmove** is undefined. What are we expecting of that particular product type?

- Is it always only cameras? 
- Are the cameras in world-space or does it allow being grouped/parented? 
- Is there a specific naming convention for the cameras?
- Does each camera need to come with a specific image mapping (so that we know what photo belongs to it)? Like [the image here](https://github.com/ynput/OpenPype/pull/5445#issuecomment-1675308689) even though that's photogrammetry, not matchmove.
- Does it need/allow meshes? Does it require at least a mesh being present with the cameras?
- Does it need/allow curves?
- Does it need/allow points/particles? -> no good support afaik in maya alembic exporter

I feel like this PR title + description should be updated to at least describe "Implement Matchmove family" and then the actual description of what we expect that family to be and do. Now it doesn't explain the use case at all.

> > I understand the thought process somewhat but I can't see yet why this currently requires to be a family of its own. What does `matchmove` do as a family that `pointcache` does not?
> 
> It "does" pretty much the same thing, but it portrays a very different message. Families are not only about their technical specifications, but also human understanding and `matchmove` implies cameras and associated data. `pointcache` implies...well....mostly geometry.

Let me start of with _`matchmove` implies cameras and associated data. _. That's only the case if we document it like that. The biggest issue of this particular PR is that it doesn't start of explaining the exact use case and _why_ it differs.

I think we're mixing two different things here - which Ayon somewhat solves with **Product Type**. I can imagine a world where there's a **matchmove** and **pointcache** product type, but both being just an _alembic_ family extraction (or pointcache family for backwards compatibility). But from a developer's perspective `families` is exactly that, a technical specification of how the plugins process - one that we should expand upon minimally unless we have actual different technical requirements for a product type. Because the more we expand families, the more we'll need to maintain those in the future which easily will end up requiring running a test-suite of its own and bringing along a lot of upkeep/management of the functionality.

> Same reason why we have both mayaScene and setdress same technical requirements, but very different feel within the pipeline.

Wait. Those are **not** the same thing at all, right? Set dress is intended to allow very specific transformational changes and data to be loaded. It doesn't allow a user to just randomly do anything in their scene, correct? Can you do vertex changes? Can it contain yeti caches? Etc. If it is the **exact** same thing as `mayaScene` I'd say it's not a family of its own but we need some sort of tagging to tag it a setdress instead.

---

In short, I'd say let's summarize first what makes `matchmove` unique so that it's clear what this PR should end up doing. Exporting multiple cameras + geometric data without any rules is already totally doable in production - so there's no rush for this feature either - for now use pointcache. It doesn't require code changes. Which means if we want to separate `matchmove` into its own family we have some time to consider what that would encapsulate - how that'd need support across different hosts, etc. Currently this PR exports the same as `pointcache` but doesn't allow loading the data into e.g. Houdini, Nuke, Blender, etc.

mkolar:
>I can imagine a world where there's a matchmove and pointcache product type, but both being just an alembic family extraction

Family / Product Type does not imply file format....at all. Matchmove at the moment is usually actualy a python file and json I think... the family has existed for over a year, but could only be published in a specific way from tray publisher. This just expands on it. The usecase we have right is for instance not supported by alembic at all, because it needs image planes too. 

In this case the product type is more about categorization of the data, not a strict content. 

>Wait. Those are not the same thing at all, right? Set dress is intended to allow very specific transformational changes and data to be loaded. It doesn't allow a user to just randomly do anything in their scene, correct?

They are absolutely identical in a technical sense at the moment. It's just a different name so people can make sense of the data intent. 

>If it is the exact same thing as mayaScene I'd say it's not a family of its own but we need some sort of tagging to tag it a setdress instead.

I feel we're looking at what family is very differently here. It's a great discussion to have, but not in this PR. Ideal for a forum discussion. 

> Exporting multiple cameras + geometric data without any rules is already totally doable in production - so there's no rush for this feature either - for now use pointcache. 

Pointcache is not usable for this. It doesn't contain all of the required data, for instance image planes. `mayaScene` is ambiguous, catchall family, that is confusing in real production and without having robust tagging as you mentioned (and which should not be used for this I believe, but that's for another discussion) you'd just end up with tons of `mayaScene` products. 

There is a rush for this and we're merging it, because it doesn't break any existing workflows, but allow for better and more human readable categorization of scenes with a certain intent. 

I agree 100% with the PR title though. We're changing that, because it evolved from original scope.

mkolar (approved):

mkolar:
>Just wondering if there is any content validator to check the setmembers including only camera type as it seems the camera validator isn't included before publishing.

@moonyuet this allow more data than just camera, so it's been disabled purposefully.

antirotor:
@kalisp can we extend how `matchmove` is described in the documentation?

antirotor (approved):
code looks good, can't test it at the moment

BigRoy:
> Pointcache is not usable for this. It doesn't contain all of the required data, for instance image planes. 

This is great! I think this solidifies a need for this families' existence - however I'd like it still if the documentation at least describes the intent of the family (and even what is intended to receive support in the future but it doesn't currently; and have opened Issues ready for those 'to-be-added' features too). Just so we all know what the intention is of the Matchmove family.

LiborBatek (changes requested):
The core or basic functions works as expected but after ```Publish``` there are missing ```ImagePlane```s for cameras which have imagePlane attached before publish.

Here is the scene before ```Publish``` of ```matchmove``` product

image

and when loaded into fresh clean workfile via Loader:

image

LiborBatek:
@kalisp @antirotor  there is also question if the empties aka ```locators``` shouldnt be valid objects for ```matchmove``` product too...now it fails when trying to incorporate those ```locators``` inside matchmove instance.

image

openpype/hosts/maya/api/lib.py
```diff
@@ -2571,7 +2571,8 @@ def _get_attrs(node):
             new_name = "{0}_baked".format(short_name)
             new_node = cmds.duplicate(node,
                                       name=new_name,
-                                      renameChildren=True)[0]
+                                      renameChildren=True,
+                                      ic=True)[0]
```
BigRoy, line 2575: Please use the long flag name for clarity. As an example I first thought this might have been `ignoreChildren` instead of `inputConnections` without looking up maya documentation.
```suggestion
                                      inputConnections=True)[0]
```

Could you elaborate what attribute connections you were missing without this?

kalisp: Specifically Image planes were missing during default export (with `Bake Cameras to World-Space`)

kalisp:
Added new variable configurable in `project_settings/maya/publish/ExtractCameraMayaScene` which is by default set to False (to keep original behavior), studio or artist can configure if input connections (as Image Planes) should be exported also.

AnyDeskMSI_AOiJflPhdm
(I am open for better labels for variable.)

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
```diff
@@ -219,3 +229,16 @@ def process(self, instance):
...
+    @classmethod
+    def get_attribute_defs(cls):
+        defs = super(ExtractCameraMayaScene, cls).get_attribute_defs()
+
+        defs.extend([
+            BoolDef("keep_input_conn",
+                    label="Keep Input conn(Image Planes)",
+                    default=cls.keep_input_conn),
+
+        ])
```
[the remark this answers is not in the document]
kalisp: It was already exporting image plane when not baking.

openpype/hosts/maya/api/lib.py
kalisp: I think we are aiming more for regular cameras, not Perspective one..imho.

LiborBatek (changes requested):
When ```Publish``` there is new option for Keep Input Connections aka ```ImagePlane``` present however it does not work and ImagePlanes get lost after publish.
image


When loading the resulting ```MatchMove``` asset back into scene there are no ```ImagePlane``` present.

There is also need to include maya ```Groups``` and ```Locators``` in the ```MatchMove``` instance allowed for publishing.

Now it fails on  ```Camera hierarchy contains additional geometry``` and extraction fails.
image

BigRoy:
Regarding the "ancestor relationship" error. @LiborBatek can you provide a screenshot of the members in your publish instance?

LiborBatek:
heres hierarchy I have:
image

BigRoy:
Thanks @LiborBatek - the issue with the "Ancestor relationship" is [here](https://github.com/ynput/OpenPype/pull/5445/files#diff-a41ef4146dc3ba27092d7934b5cd9e9661c96da5ae2ae2581454ac7135e11f48R75-R92) since it's adding the parent group of all members of the set as `-root`. It doesn't really make any sense and I think was once written with the idea that it should take the parent of the shape members as opposed to parents of transforms.

It's basically saying, also add `|CAMP_GRP` as root in the alembic export, but it shouldn't - it should really only export the camera itself and use that as `-root` since it's used when backing to world-space.

Also, the actual logic there can be highly optimized by not actually doing the query per member but just all at once. It's likely never optimized because it was never intended to hold more than a single camera anyway.

Anway - this camera extractor should not be abused for anything other than a single camera extractor (or at least more than cameras) since it's focused on baking cameras to world-space etc. and leaves out any notion of other object types. It'd either be up for a thorough re-write or having a dedicated matchmove extractor instead.

LiborBatek:
@BigRoy thx for clarification
@antirotor @kalisp well we should decide if dedicated ```matchmove extractor``` would be the way to go (regarding what BigRoy wrote above) instead of old ```Camera extractor``` ...I can imagine that it would happen pretty quickly that user would like to include not just multiple cameras but also the usual outcome from matchmove sw like custom Grouping and empties aka locators (depends on DCC terminology) 

Thoughts?

kalisp:
I must have been wrong, I cannot make it work with Image planes and `Bake Cameras to World-Space`, 
I must have had that switched off by accident when testing (without baking it exports Image Planes), 
I think that issue is that `Bake` is duplicating camera node and publishes that new node. 
`inputConnections` adds connected Image plane to this copy (could be seen in `Node Editor`) , but it won't export it, even if Image plane is detached from original camera node. So pwned.

LiborBatek:
@antirotor could u take a look because if Im not mistake @kalisp is struggling forcing Maya to reconnect  ```ImagePlane``` back to the new baked ```Camera``` so he is a bit stuck on this issue.

Will test then when resolved this, thx.

openpype/hosts/maya/plugins/publish/extract_camera_alembic.py
```diff
@@ -35,7 +39,6 @@ def process(self, instance):
 
         # validate required settings
         assert isinstance(step, float), "Step must be a float value"
-        camera = cameras[0]
 
         # Define extract output file path
         dir_path = self.staging_dir(instance)
```
tokejepsen, line 44: BTW we need to ensure the directory exists here.
```suggestion
        dir_path = self.staging_dir(instance)
        if not os.path.exists(dir_path):
            os.makedirs(dir_path)
```

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
```diff
@@ -131,15 +140,15 @@ def process(self, instance):
...
+        members = set(cmds.ls(instance.data['setMembers'], leaf=True,
+                      shapes=True, long=True, dag=True))
+        cameras = set(cmds.ls(members, leaf=True, shapes=True, long=True,
+                      dag=True, type="camera"))
 
         # validate required settings
         assert isinstance(step, float), "Step must be a float value"
-        camera = cameras[0]
-        transform = cmds.listRelatives(camera, parent=True, fullPath=True)
+        transform = cmds.listRelatives(list(cameras),
```
tokejepsen, line 150:
```suggestion
        transforms = cmds.listRelatives(list(cameras),
```
For clarity, this variable should be plural.

openpype/hosts/maya/api/lib.py
```diff
@@ -2571,7 +2572,8 @@ def _get_attrs(node):
             new_name = "{0}_baked".format(short_name)
             new_node = cmds.duplicate(node,
                                       name=new_name,
-                                      renameChildren=True)[0]
+                                      renameChildren=True,
+                                      inputConnections=copy_input_connections)[0]  # noqa
```
tokejepsen, line 2576:
```suggestion
                                      upstreamNodes=copy_input_connections)[0]  # noqa
```
Very confusing command flags in `cmds`, but for whatever reason `upsteamNodes` is the same as using the `Duplicate input graph` through the `Duplicate Special` UI menu.

tokejepsen (changes requested):

openpype/hosts/maya/plugins/publish/extract_camera_alembic.py
BigRoy: The `staging_dir` method creates the temporary directory - so we shouldn't actually. Right?

openpype/hosts/maya/api/lib.py
BigRoy: Duplicate input graph - basically means duplicate everything that's connected to the inputs (which is the upstream nodes). Even though they use 'synonyms' it's still pretty clear right? Or was there another reason you were getting confused?

tokejepsen: When using `Duplicate Special` with `duplicate input graph` enabled, you get this as mel commands;

```
buildEditMenu("MayaWindow|mainEditMenu");
DuplicateSpecialOptions;
performDuplicateSpecial true;
duplicateSpecialCallback OptionBoxWindow|formLayout247|tabLayout4|formLayout249|tabLayout5|columnLayout32 1; hideOptionBox;
duplicate -rr -un;
```

We want to duplicate the input graph to copy the image plane to the baked camera. Unless I've missed something about this PR?

BigRoy: The input graph is easily way larger than the image plane. You'd be duplicating full rigs, with full animation curves, etc. Plus if you would just want to export inputs along then it's just as fine to duplicate just the input connections (without duplicating the actual source nodes) and allow Maya to export that along on maya scene export with `constructionHistory`,  `channels`, etc.

I suppose even when duplicating them, if you then end up exporting the camera as the selected entry then the image plane should NOT be part of the export since we're exporting without construction history in the extractor as far as I know.

The PR did discuss issues with the `matchmove` family not taking along an image plane on the export - totally fine issue. But what this does by duplicating upstream nodes should not be the way to solve that tiny issue.

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
```diff
@@ -152,22 +161,23 @@ def process(self, instance):
...
                         self.log.info(
-                            "Performing camera bakes: {}".format(transform))
+                            "Performing camera bakes: {}".format(transforms))
                         baked = lib.bake_to_world_space(
-                            transform,
+                            transforms,
                             frame_range=[start, end],
-                            step=step
+                            step=step,
+                            copy_input_connections=self.keep_input_connections
```
tokejepsen, line 169: `self.keep_input_connections` always seem to get its value from settings rather than from the instance.

tokejepsen (dismissed):

openpype/hosts/maya/api/lib.py
tokejepsen: What is your solution for duplicating the camera with its image plane?

Can only see `upstreamNodes=True` working with the `cmds.duplicate` command.

BigRoy: As far as I know we don't need to duplicate the image plane - we'd just need to ensure it gets exported along which it should if `inputConnections=True` for the duplicate (because then it is connected) and the maya export actually exports with inputs which I believe is `constructionHistory=True` flag on the file export.

I recall there being a way to duplicate just a selection of nodes and their connections in between them would be transferred to the duplicates but I'm not sure how to do that. Maybe it's just `cmds.duplicate([image_plane, camera])`. I know maya file export is capable of doing that even if you have input connections disabled.

kalisp:
Added method provided by Toke (still using `inputConnections`). Image planes had to been re-attached.

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
```diff
@@ -219,3 +235,28 @@ def process(self, instance):
...
+    def ensure_image_plane_attachment(self, camera):
+        image_planes = cmds.listConnections(camera, type="imagePlane")
+        if not image_planes:
+            return
+
+        image_planes = [x.split("->")[1] for x in image_planes]
+        for image_plane in image_planes:
+            self.log.debug("reattaching {}".format(image_plane))
+            cmds.imagePlane(image_plane, edit=True, detach=True)
+            cmds.imagePlane(image_plane, edit=True, camera=camera)
```
BigRoy, line 262: This should be a context manager so it's reverted after the export/baking because otherwise the user now is left with a scene without image plane attached after the export (and thus the next export also suddenly doesn't have an image plane). Right? Or does it attach the image plane to BOTH cameras?

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
```diff
@@ -152,22 +161,29 @@ def process(self, instance):
...
-                        members = members + baked_camera_shapes
-                        members.remove(camera)
+                        baked_camera_shapes = set(cmds.ls(baked,
+                                                  type="camera",
+                                                  dag=True,
+                                                  shapes=True,
+                                                  long=True))
+
+                        for cam in baked_camera_shapes:
+                            self.ensure_image_plane_attachment(cam)
```
BigRoy, line 181: This behavior should not happen for a `camera` instance - only for `matchmove` instance.

openpype/hosts/maya/plugins/publish/extract_camera_alembic.py
BigRoy: Yup, definitely sounds like an issue with custom staging directory then 👍

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
tokejepsen: Think we can just wrap the code in the `delete_after` context for ease.

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
tokejepsen: Yeah, the baking should only happen for `matchmove`

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
kalisp: Baked camera is deleted after export, Image Plane is kept on original camera

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
kalisp: This will trigger only if `Keep input connections` is set to True (by default is False), so studio/artist can control behavior.

BigRoy: The `camera` family should exclude image planes by design, and thus for that family it should not allow that option if that means it starts exporting image planes.

LiborBatek (changes requested):
Gives me error when extracting cameras:
image

```
  File "D:\REPO\OpenPype\.venv\lib\site-packages\pyblish\plugin.py", line 527, in __explicit_process
    runner(*args)
  File "D:\REPO\OpenPype\openpype\hosts\maya\plugins\publish\extract_camera_mayaScene.py", line 168, in process
    baked = lib.bake_to_world_space(
  File "D:\REPO\OpenPype\openpype\hosts\maya\api\lib.py", line 2629, in bake_to_world_space
    delete_bin.extend(cmds.scaleConstraint(node, new_node, mo=False))
RuntimeError: Connection not made: 'camera7_baked_scaleConstraint1.constraintScaleX' -> 'camera7_baked.scaleX'.  Destination attribute must be writable.
Could not add constraint or connections.
```
@BigRoy  I totally get your points however imagine that those cameras were comming from elswhere (DCC for image tracking etc) with connections outside of the maya project and being huge ones (file size wise). U probably wouldnt like the idea of copying those to Publish folder inside maya project to keep em versioned/organised again anyway.
It is potentionally source of issue (missing connections later on) but
Its up to the user being sensible when doing matchmove extraction (aka Publish) and the place from where the ImagePlanes are originating.

This implementation as it is now still can do good job without putting more effort into it.

LiborBatek:
@BigRoy  you are def right! So lets say that user is aware of the possible issues so he first publishes those ImagePlanes via Tray Publisher and load via Loader to the workfile scene...connecting those to the matchmove cameras. It should then produce exactly what are u after, right? Just a matter of awarness of the user. Works imho.
In any other cases then ...user fail! 

I also do agree that maybe incorporate input connections feature to just ```matchmove``` product type is a good idea.

kalisp:
The issue is that image planes **are and always were** published without Baking (as it was exporting original camera, the main difference is that Baking is creating duplicate of camera which had issues with not copying Input connections.)

So the question is if publishing IM for single camera is unwanted side-effect and it should be removed and only added for `matchmove`. And I don't know how it should be fixed. Disconnect IMs, export, reconnect?

BigRoy:
> So the question is if publishing IM for single camera is unwanted side-effect and it should be removed and only added for matchmove. And I don't know how it should be fixed. Disconnect IMs, export, reconnect?

I think the whole issue is rather that non-baked cameras are currently problematic to export to `camera` and `mayaAscii` already. At our studio we've never spent time 'fixing' that area of the camera exports because we either relied on baked camera or when we didn't want the baked camera we were relying on alembic exports - both which never included the image plane.

The issue currently with a non-baked camera export is likely much bigger than the image plane. If the camera (or its parent hierarchy) is e.g. constrained to the hand of a character rig then likely the character rig (or at least that part of driving that hand) will be include in the camera export for non-baked cameras.

I'd say that the camera family should not include the image plane, and prefably not include input connections but the logic for that can easily be more complicated when exporting a maya ascii. Since you'd still need to bake the hierarchy and above to ensure it's not driven by nodes external to that hierarchy, etc.

openpype/hosts/maya/api/lib.py
```diff
@@ -2622,10 +2624,18 @@ def _get_attrs(node):
...
+
             # Constraints
             delete_bin.extend(cmds.parentConstraint(node, new_node, mo=False))
             delete_bin.extend(cmds.scaleConstraint(node, new_node, mo=False))
 
+            # resetting scale keys on original node
+            for orig_node_attr, frame, _ in scale_keys:
+                log.debug("resetting key::{} at {}".format(orig_node_attr,
+                                                           frame))
+                cmds.setKeyframe(orig_node_attr, time=(frame,))
```
BigRoy, line 2637: Sorry, wait.. what are you doing here? :) You're breaking the scale Constraint logic by keying over the scale constraint and thus making it a blend constraint override which defaults to `1.0`.

What's going on here? What are we trying to solve but are actually breaking with this code?

openpype/hosts/maya/api/lib.py
```diff
@@ -2639,6 +2649,39 @@ def _get_attrs(node):
...
+    removes keys, they need to be set again after constraint parenting.
+    """
+    scale_attrs = set(["sx", "sy", "sz"])
+
+    saved_keys = []
+    orig_key_value = [1.0]
+    key_value_changed = False
+    for attr in scale_attrs:
+        orig_node_attr = '{0}.{1}'.format(node, attr)
+        frames = cmds.keyframe(orig_node_attr, query=True) or []
```
BigRoy, line 2665: You are likely better of doing this since you'd be doing less calls against maya's slow api:
```python
times = cmds.keyframe(orig_node_attr, query=True, timeChange=True)
if not times:
    continue
values = cmds.keyframe(orig_node_attr, query=True, valueChange=True)
time_values = {time: value for time, value in zip(times, values)}
```

openpype/hosts/maya/api/lib.py
```diff
@@ -2639,6 +2649,39 @@ def _get_attrs(node):
...
+                key_value_changed = True
+                break
+
+    if key_value_changed:
+        raise RuntimeError("{} scale keys contain non default values.".format(node) +  # noqa
+                           "These cannot be baked out. Please remove keys for " +  # noqa
+                           "scale attributes")
+
+    for orig_node_attr, frame, _ in saved_keys:
+        cmds.cutKey(orig_node_attr, time=(frame,), option="keys")
```
BigRoy, line 2680:
```suggestion
    frames = tuple(data[1] for data in saved_keys)
    cmds.cutKey(orig_node_attr, time=frames, option="keys")
```

openpype/hosts/maya/api/lib.py
kalisp: I was trying to workaround this issue https://github.com/ynput/OpenPype/pull/5445#pullrequestreview-1611062862

We were looking into that and Scale keys are breaking `cmds.scaleConstraint`. Why? IDK.
General consensus here was that keys on scale attributes are most likely just side product of keying via `S` and most likely not necessary (fingers crossed).
So the idea is to check if key values are default, eg. just generic `S`, if so, they could be safely removed (and then set back for original camera). If they are different, artist would like to set them, but this is not supported. (And we will need to figure out how to solve it, hopefully it won't happen.)
Open for ideas how to solve that `Could not add constraint or connections` without this unholy mess.

openpype/hosts/maya/api/lib.py
kalisp: Updated/cleaned up logic.

openpype/hosts/maya/api/lib.py
kalisp: Updated/cleaned up logic.

kalisp:
Workarounded Libor's issue with animated camera, see https://github.com/ynput/OpenPype/pull/5445#discussion_r1337428930

openpype/hosts/maya/api/lib.py
BigRoy: I see - this issue is **new** because you're now duplicating the original camera with input connections. As said before - you should avoid that since that's what will get you into this mess. With that disabled the input connections will not be there and thus Scale attribute will be empty. The same issue will occur on translate.

Below [this line](https://github.com/ynput/OpenPype/pull/5445/files#diff-0f94b40dc0b348d7194c61cb01a57668eb3490d22cdbff89d37dd552287fccadR2625) you could just remove any inputs whatsoever to transform attributes - as those should be propagated from the constraints instead to be 'world-space' equivalents.

Using e.g. something like this:
```python
def disconnect_inputs(plug):
    """Disconnect any input sources for the plug."""
    sources = cmds.listConnections(plug, plugs=True, source=True,
                                   destination=True, shapes=True,
                                   skipConversionNodes=False) or []
    if not sources:
        return

    for dest, source in pairwise(sources):
        if cmds.isConnected(source, dest):
            cmds.disconnectAttr(source, dest)
```

And then make it like:
```python
            # Unlock transform attributes so constraint can be created
            for attr in transform_attrs:
                plug = "{}.{}".format(new_node, attr)
                cmds.setAttr(plug, lock=False)
                disconnect_inputs(plug)
```

And voila, you should be able to make the constraints as needed. But as said before - copying input connections can definitely make the behavior much different from what you intend it to be.

I [noted on discord yesterday that there's a flag on `maya.cmds.file` for `exportSelected`](https://discord.com/channels/517362899170230292/563751989075378201/1155954318202118194) which could make this much more trivial, it's [`exportSnapshotCallback`](https://help.autodesk.com/cloudhelp/2020/ENU/Maya-Tech-Docs/CommandsPython/file.html#flagexportSnapshotCallback). Would be quite a refactor but might make it easier in the long run to maintain?



You can just add: `disconnect_inputs(

LiborBatek (changes requested):
Now ```matchmove``` being correctly produced which is ok
image

But during the publishing the active workfile being altered so ```ImagePlanes``` got disconnected and deleted from the maya scene which not desirable...

Inspected workfile after Publish
image

BigRoy:
> Now `matchmove` being correctly produced which is ok image
> 
> But during the publishing the active workfile being altered so `ImagePlanes` got disconnected and deleted from the maya scene which not desirable...
> 
> Inspected workfile after Publish image

I'm pretty sure this might become easier to solve using the `exportSnapshotCallback` for maya export that I mentioned [here](https://github.com/ynput/OpenPype/pull/5445#pullrequestreview-1646234554). Quite a refactor but maybe worth the effort if it cleans things up nicely by design.

LiborBatek:
I have also tested case when some cameraShape has animated ```Scale``` on purpose...

Gives me fail during publishing telling me to remove those keys on scale:
image

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
```diff
@@ -219,3 +238,40 @@ def process(self, instance):
...
+
+    def ensure_image_plane_attachment(self, camera, reattach=False):
+        """Reattaches image planes to baked or original cameras.
+
+        Baked cameras are duplicates of original ones, even with
+        `keep_input_connections` the connection of image plane is still on the
+        original one. This attaches it to duplicated camera properly.
+        If 'reattach' it attaches back image plane to original camera.
+        TODO figure out use of some context manager
+        (`delete_after` doesnt work)
```
BigRoy, line 264: Here's some pseudocode.
```python
import contextlib

def attached_image_planes(images_planes, camera):
    originals ={
        image_plane: cmds.imagePlane(image_plane, query=True, camera=True)
        for image_planes in image_planes
    }
    try:
        for image_plane in image_planes:
            cmds.imagePlane(image_plane, edit=True, detach=True)
            cmds.imagePlane(image_plane, edit=True, camera=camera)
        yield
    finally:
        for image_plane, original_camera in originals.items():
            cmds.imagePlane(image_plane, edit=True, detach=True)
            cmds.imagePlane(image_plane, edit=True, camera=original_camera)
```

Example usage:
```python
with attached_image_planes(image_planes, camera):
    ... export here
```

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
```diff
@@ -219,3 +238,40 @@ def process(self, instance):
...
+        """
+        image_planes = cmds.listConnections(camera, type="imagePlane")
+        if not image_planes:
+            return
+        if reattach:
+            # find original camera name
+            camera = camera[0:camera.find("_baked")]
```
BigRoy, line 271: This should not be done like this - this will fail if the original camera already happened to have `_baked` anywhere in the name.

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
```diff
@@ -219,3 +238,40 @@ def process(self, instance):
...
+
+        image_planes = [x.split("->")[1] for x in image_planes]
```
BigRoy, line 273: Be aware that it's not required for an image plane path to include `->` the image plane can be connected to the camera and NOT be parented to it as far as I know. As such, I believe what you need to do is: `x.rsplit("->", 1)[-1]` so that even if `->` is not there you get a valid result.

openpype/hosts/maya/plugins/publish/extract_camera_alembic.py
```diff
@@ -64,9 +69,10 @@ def process(self, instance):
 
                 # if baked, drop the camera hierarchy to maintain
                 # clean output and backwards compatibility
-                camera_root = cmds.listRelatives(
-                    camera, parent=True, fullPath=True)[0]
-                job_str += ' -root {0}'.format(camera_root)
+                camera_roots = cmds.listRelatives(
+                    cameras, parent=True, fullPath=True)
+                for camera_root in camera_roots:
+                    job_str += ' -root {0}'.format(camera_root)
```
BigRoy, line 75: FYI - Note that this will fail if one camera happens to be a parent of the other since they'd have ancestor relationships.

openpype/hosts/maya/api/lib.py
BigRoy: @kalisp not sure why but for me this comment chain doesn't show on the "Changes" page on github - so just commenting here once more to ensure you've seen this comment. :)

LiborBatek (changes requested):
Now the ImagePlanes stays untouched in the active workfile but they dont appear in the Published asset of ```matchmove```....

image

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
kalisp: Updated

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
kalisp: resolved by context manager

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
```diff
@@ -219,3 +238,51 @@ def process(self, instance):
...
+                    tooltip="Preserving input connections will allow Image Planes to stay connected on publish",  # noqa
+                    default=cls.keep_input_connections),
+
+        ])
+
+        return defs
+
+
+@contextlib.contextmanager
+def attached_image_planes(source_cameras, baked_cameras):
```
BigRoy, line 258: The function arguments and name to this seem odd.

```python
def _attach_image_plane(camera, image_plane):
    cmds.imagePlane(image_plane, edit=True, detach=True)
    cmds.imagePlane(image_plane, edit=True, camera=camera)

def transfer_image_planes(source_cameras, target_cameras):
    originals = {}
    try:
        for source_camera, target_camera in zip(source_cameras, target_cameras):
            if source_camera == target_camera:
                continue
            
            image_planes = cmds.listConnections(source_camera,
                                                type="imagePlane")
            if not image_planes:
                continue
                
            # Split of the parent path they are attached - we want
            # the image plane node name.
            # TODO: Does this still mean the image plane name is unique?
            image_planes = [x.split("->", 1)[1] for x in image_planes]
                
            originals[source_camera] = image_planes
            for image_plane in image_planes:
                _attach_image_plane(target_camera, image_plane)
                
        yield
    finally:
        for camera, image_planes in originals.items():
            for image_plane in image_planes:
                _attach_image_plane(camera, image_plane)
```

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
```diff
@@ -219,3 +238,51 @@ def process(self, instance):
...
+    export it reattaches it back to original to keep image plane in workfile.
+    """
+    originals = {}
+    cameras = {baked: source for baked, source in zip(baked_cameras,
+                                                      source_cameras)
+               if source != baked}
+    try:
+        for baked_camera, source_camera in cameras.items():
+            image_planes = cmds.listConnections(baked_camera,
+                                                type="imagePlane")
```
BigRoy, line 273: Shouldn't this be taking it from the `source` ?

openpype/hosts/maya/api/lib.py
```diff
@@ -2622,10 +2624,14 @@ def _get_attrs(node):
...
                 cmds.setAttr('{0}.{1}'.format(new_node, attr), lock=False)
 
+            scale_keys = _remove_scale_keys(node)
+
             # Constraints
             delete_bin.extend(cmds.parentConstraint(node, new_node, mo=False))
             delete_bin.extend(cmds.scaleConstraint(node, new_node, mo=False))
 
+            _set_scale_keys(scale_keys)
+
```
BigRoy, line 2634:
```suggestion
                plug = "{}.{}".format(new_node, attr)
                cmds.setAttr(plug, lock=False)
                disconnect_inputs(plug)

            # Constraints
            delete_bin.extend(cmds.parentConstraint(node, new_node, mo=False))
            delete_bin.extend(cmds.scaleConstraint(node, new_node, mo=False))

```

This should avoid the issues and do what originally was intended - which is to take world-space scale into account instead of transfering scale keys 1-to-1.

openpype/hosts/maya/api/lib.py
```diff
@@ -2639,6 +2645,58 @@ def _get_attrs(node):
...
+    """Adds keys back on scale attributes on source camera.
+
+    Args:
+        (dict): {camera.scaleAttribute : [(time1, time1), (time2, time2)]}
+    """
+    for orig_node_attr, times in scale_keys.items():
+        cleaned_times = [time[0] for time in times]
+        cmds.setKeyframe(orig_node_attr, time=cleaned_times)
+
+
```
BigRoy, line 2699:
```suggestion
def disconnect_inputs(plug):
    """Disconnect any input sources for the plug."""
    sources = cmds.listConnections(plug, plugs=True, source=True,
                                   destination=True, shapes=True,
                                   skipConversionNodes=False) or []
    if not sources:
        return

    for dest, source in pairwise(sources):
        if cmds.isConnected(source, dest):
            cmds.disconnectAttr(source, dest)


```

LiborBatek (changes requested):
Even tho now ```Publish``` produces correct data with all connected ```ImagePlanes``` and keyframes, there are few issues still.

1. Matchmove Instance' s ```Keep input connections``` does not have any effect when ```OFF```
2. Cameras having different names when published...this could be possible issue in some use cases ...now there is suffix present ```_baked``` in the names
3. I propose to change the terms ```Camera (Alembic)``` and ```Camera (Maya Scene)``` in the ```matchmove``` instance to 

```Extract Alembic``` and ```Extract Maya Scene```

image

File: openpype/hosts/maya/api/lib.py (hunk 23 above)
Comment (kalisp) [reply openpype/hosts/maya/api/lib.py:2699]: I don't think that `sources` contain pairs to be used in `pairwise`

Comment (BigRoy) [reply openpype/hosts/maya/api/lib.py:2699]: Ah, you're right - `destination=True` should be `connections=True` I suspect.

Comment (BigRoy) [reply openpype/hosts/maya/api/lib.py:2699]: Apparently we have used a similar implementation [here](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/maya/plugins/load/load_image_plane.py#L24-L31)

**Comment by kalisp** [comment]
Removed copying of input connections as BigRoy mentioned couple of times it might be dangerous. Explicitly exposed only flag for control of exporting Image planes.
AnyDeskMSI_RwzBUcBEp9

**Review by LiborBatek — APPROVED** [review]
Now all works as it should and ```matchMove``` asset being correctly produced with ```ImagePlanes``` just when ```Keep ImagePlanes``` being ```Enabled``` and vice versa.

There is also correctly publish Camera and its ```Scale``` if animated, so all good!

File: openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py [hunk 24]
```diff
@@ -219,3 +236,62 @@ def process(self, instance):
...
+
+        defs.extend([
+            BoolDef("keep_image_planes",
+                    label="Keep Image Planes",
+                    tooltip="Preserving connected image planes on camera",
+                    default=cls.keep_image_planes),
+
+        ])
+
+        return defs
```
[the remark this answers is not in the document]
kalisp: All combinations of regular/baked with/without image planes should work right now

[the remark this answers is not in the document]
BigRoy: I understand that it exports it - I'm saying that a camera instance should **never** export the image plane by design - so it this toggle allows it to then we're basically adding unexpected data in it? The camera product type was intended to only hold a single camera.

PR #5451, opened 2023-08-14
Farm publishing of existing frames

jakubjezek001:
## Changelog Description
- abstraction for farm publishing
- abstraction for expected files operation
- submit publish job rewrites for better support of `publish_on_farm` tagged representations
- farm targeting with global collector
- traypublisher: support of farm publishing
- nuke: implemented abstractions for farm publishing
- expected files abstraction implemented in Houdini, Max, Maya

## Additional info
I drew inspiration from [this pull request](https://github.com/ynput/OpenPype/pull/5195) and would like to express my gratitude to @fabiaserra! Please consider this pull request as a proof of concept. I understand that it is quite substantial and should be divided into smaller parts. My intention is to initiate a discussion regarding some of the modifications in farm publishing.

## Testing notes:
1. start with this step
2. follow this step


openpype/hosts/houdini/plugins/publish/collect_mantra_rop.py
```diff
@@ -31,49 +31,65 @@ class CollectMantraROPRenderProducts(pyblish.api.InstancePlugin):
...
+                continue
+
+            aov_name = "vm_filename_plane{:>1}".format(index)
+            aov_boolean = "vm_usefile_plane{:>1}".format(index)
+            aov_enabled = rop.evalParm(aov_boolean)
+            has_aov_path = rop.evalParm(aov_name)
+
+            # skip disabled AOVs
+            if not (has_aov_path and aov_enabled == 1):
+                continue
```
BigRoy, line 80: Why are you formatting using `{:>1}`? Formatting with `{}` provides the same output for this.

```suggestion
            # skip empty variables
            if not rop.evalParm("vm_variable_plane{}".format(index)):
                continue

            # skip disabled AOVs
            aov_name = "vm_filename_plane{}".format(index)
            if rop.evalParm(aov_name) != 1:
                continue
                
            # skip AOV without use file set
            has_aov_path = rop.evalParm("vm_usefile_plane{}".format(index))
            if not has_aov_path:
                continue
```

openpype/hosts/nuke/plugins/publish/collect_nuke_instance_data.py
```diff
@@ -2,11 +2,13 @@
 import pyblish.api
 
 
-class CollectInstanceData(pyblish.api.InstancePlugin):
-    """Collect all nodes with Avalon knob."""
+class CollectNukeInstanceData(pyblish.api.InstancePlugin):
```
BigRoy, line 5: Changing the class name would likely invalidate settings for `sync_workfile_version_on_families` attribute. I assume that's set from settings?

openpype/hosts/nuke/plugins/publish/collect_slate_node.py
```diff
@@ -5,18 +5,22 @@
...
-                      if not n["disable"].getValue()),
-                     None)
+        slate = next(
+            (
+                n_ for n_ in nuke.allNodes()
+                if "slate" in n_.name().lower()
+                if not n_["disable"].getValue()
+            ),
+            None
+        )
```
BigRoy, line 23: Admittedly this feels like a very unsafe approach to retrieving a global node. If anyone happens to have any additional node with `slate` (in any form of upper or lowercase) in their scene wouldn't that potentially conflict here?

Not sure what the slate node is intended to be.
Shouldn't we at least filter the `allNodes` to a specific node type classification?

openpype/hosts/nuke/plugins/publish/collect_writes.py
```diff
@@ -48,113 +27,95 @@ def process(self, instance):
...
-                self.set_representation_colorspace(
-                    representation, instance.context,
-                    colorspace=colorspace
-                )
+            self.set_farm_representation(
+                instance,
+                write_file_path,
+                frame_start,
+                frame_end,
+                only_existing=True
```
BigRoy, line 45: This looks odd - why is it generating a farm representation when the render target is not farm?

openpype/hosts/nuke/plugins/publish/collect_writes.py
```diff
@@ -48,113 +27,95 @@ def process(self, instance):
...
+            frame_start = int(write_node["first"].getValue())
+            frame_end = int(write_node["last"].getValue())
+
+        return frame_start, frame_end
+
+    def _set_additional_instance_data(
+        self, instance, write_node, frame_start, frame_end, colorspace
+    ):
+        """Set additional instance data.
+
```
BigRoy, line 90: It would be great if the docstring could elaborate on what "additional data" is. What is it setting?

openpype/hosts/nuke/plugins/publish/extract_camera.py
```diff
@@ -11,9 +11,9 @@
 
 
 class ExtractCamera(publish.Extractor):
-    """ 3D camera exctractor
+    """ 3D camera extractor
```
BigRoy, line 14:
```suggestion
    """3D camera extractor
```

openpype/hosts/nuke/plugins/publish/extract_model.py
```diff
@@ -11,9 +11,9 @@
 
 
 class ExtractModel(publish.Extractor):
-    """ 3D model exctractor
+    """ 3D model extractor
```
BigRoy, line 14:
```suggestion
    """3D model extractor
```

openpype/hosts/nuke/plugins/publish/validate_rendered_frames.py
```diff
@@ -116,15 +116,17 @@ def process(self, instance):
...
-            if ("slate" in instance.data["families"]) \
-                    and (frame_length != collected_frames_len):
-                collected_frames_len -= 1
-                f_start_h += 1
+            if (
+                    "slate" in instance.data["families"]
+                    and frame_length != collected_frames_len
+            ):
+                f_start_h -= 1
+                frame_length += 1
```
BigRoy, line 124: I know it's pre-existing code, but I'd add a comment as to what these if branches are trying to solve. It seem odd to say the least.

openpype/hosts/traypublisher/plugins/publish/collect_simple_instances.py
```diff
@@ -47,90 +52,86 @@ def process(self, instance):
...
-                " with {} representations: {}"
+                " with {} representations"
             ).format(
                 instance_label,
-                len(instance.data["representations"]),
-                ", ".join(repre_names)
+                len(instance.data["representations"])
             )
         )
+        self.log.debug(pformat(instance.data))
```
BigRoy, line 122: Please avoid overly verbose logging like this - it's slow.
```suggestion
```

openpype/modules/deadline/plugins/publish/submit_nuke_deadline.py
```diff
@@ -123,13 +121,10 @@ def process(self, instance):
         render_path = instance.data['path']
         script_path = context.data["currentFile"]
 
-        for item in context:
-            if "workfile" in item.data["families"]:
-                msg = "Workfile (scene) must be published along"
-                assert item.data["publish"] is True, msg
```
BigRoy, line 129: This not being forced anymore seems like quite a change. Who might have been relying on this?

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -294,12 +297,13 @@ def _submit_deadline_post_job(self, instance, job, instances):
...
         elif instance.data.get("bakingSubmissionJobs"):
             self.log.info("Adding baking submission jobs as dependencies...")
+
             job_index = 0
             for assembly_id in instance.data["bakingSubmissionJobs"]:
                 payload["JobInfo"]["JobDependency{}".format(
                     job_index)] = assembly_id  # noqa: E501
                 job_index += 1
-        else:
+        elif job.get("_id"):
```
BigRoy, line 306: Should we add an `else:` branch doing a debug log saying "No dependent jobs detected." just to make sure we can trace it nicely if start hitting the else branch?

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -325,9 +329,18 @@ def _submit_deadline_post_job(self, instance, job, instances):
...
+
+        r''' SUBMiT PUBLiSH JOB 2 D34DLiN3
+          ____
+        '     '            .---.  .---. .--. .---. .--..--..--..--. .---.
+        |     |   --= \   |  .  \/   _|/    \|  .  \  ||  ||   \  |/   _|
+        | JOB |   --= /   |  |  ||  __|  ..  |  |  |  |;_ ||  \   ||  __|
+        |     |           |____./ \.__|._||_.|___./|_____|||__|\__|\.___|
+        ._____.
+
+        '''
```
BigRoy, line 343:
```suggestion
```
This is way too high - it's not close to whether the actual job gets submitted whereas previously it was. This should be removed or moved back to the older position in the code.

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -521,13 +510,50 @@ def process(self, instance):
...
-            create_metadata_path(instance, anatomy)
+                create_metadata_path(instance, anatomy)
+
+        with open(metadata_path, "w") as f_:
+            json.dump(publish_job, f_, indent=4, sort_keys=True)
+
+    def _create_default_render_job(self, instance):
+        import getpass
+
+        self.log.info("Faking job data ...")
```
BigRoy, line 524: This should not be an info log. Also, I feel like it's lacking information about use cases where this might happen in the code. When do we ever fake it?

openpype/modules/deadline/plugins/publish/validate_expected_and_rendered_files.py
```diff
@@ -12,18 +11,37 @@ class ValidateExpectedFiles(pyblish.api.InstancePlugin):
...
 
     # check if actual frame range on render job wasn't different
     # case when artists wants to render only subset of frames
     allow_user_override = True
 
     def process(self, instance):
-        self.instance = instance
-        frame_list = self._get_frame_list(instance.data["render_job_id"])
+        """Process all the nodes in the instance"""
+        self.log.debug("Validate expected files for instance")
```
BigRoy, line 22: Just logging that the plugin is running is redundant - if we need to know if a plugin runs and we can't from the UX or the publishing system - then that's an issue with the publisher itself and its debugging, not with the plug-in. We shouldn't need to log a "plugin is running" statement like this unless we're actually logging information that is unique to what is going on.
```suggestion
```

openpype/pipeline/colorspace.py
```diff
@@ -13,12 +13,17 @@
     Logger
 )
 from openpype.pipeline import Anatomy
+from openpype.lib.transcoding import VIDEO_EXTENSIONS, IMAGE_EXTENSIONS
+
 
 log = Logger.get_logger(__name__)
 
 
 class CashedData:
```
BigRoy, line 22: Pretty sure this is not cashed data. 💰 💵 💲 

```suggestion
class CachedData:
```

However, will need to refactor usages of it too.

openpype/pipeline/expected_files.py
```diff
@@ -0,0 +1,297 @@
...
+    """
+    return collect_filepaths_from_sequential_path(
+        frame_start,
+        frame_end,
+        path,
+        only_existing=False,
+    )
+
+
+def collect_files_from_sequential_path(
```
BigRoy, line 87: `collect_filenames..` would be much clearer to me that this is about basenames. Also, it slightly feels like overkill to have both a full path version and a file name version if both are calling the same logic anyway. Why not just expose the filepath one, and if someone wanted filenames they can just do the `os.path.basename` mapping themselves.

Also, to me it's very confusing that `collect_files_from_sequential_path` would always return a list but `collect_filepaths_from_sequential_path` can return a list or a single filepath. That's just not intuitive.

They should either both return list or single file entry or both always return a list.

openpype/pipeline/expected_files.py
```diff
@@ -0,0 +1,297 @@
...
+        collected_files (list[str]): list of files
+
+    Returns:
+        Any[str, None]: single filepath or None if not possible
+    """
+    collections, reminders = clique.assemble(collected_files)
+    if collections:
+        return collections[0].format("{head}{padding}{tail}")
+    elif reminders:
+        return reminders[0]
```
BigRoy, line 274:
```suggestion
    elif remainders:
        return remainders[0]
```

openpype/pipeline/expected_files.py
```diff
@@ -0,0 +1,297 @@
...
+
```
BigRoy, line 275: Note that there's nothing ensuring in this logic that there aren't more than one collection or more than one remainder - so it might be obfuscating/hiding potential bugs?

openpype/pipeline/farm/pyblish_functions.py
```diff
@@ -211,9 +211,15 @@ def create_skeleton_instance(
...
-              else "prerender")
+    # pass families from original instance to skeleton
+    family = "render"
+    # try to find family with farm suffixes
+    families = instance.data["families"]
+    for family_ in families:
+        match = re.search(r"\.farm(_frames)?$", family_)
+        if match:
+            family = re.sub(r"\.farm(_frames)?$", "", family_)
+            break
```
BigRoy, line 222:
```suggestion
    # search for family with farm suffixes to
    # pass families from original instance to skeleton
    families = instance.data["families"]
    farm_suffixes = [".farm", ".farm_frames"]
    for family_ in families:
        suffix = next((s for s in farm_suffixes if family.endswith(suffix)), None)
        if suffix:
            family = family_[:-len(suffix)]
            break
    else:
        # No family found with farm suffix
        family = "render"
```
This would likely be faster (and maybe more readable?)

openpype/pipeline/frames.py
```diff
@@ -0,0 +1,62 @@
+"""Utility functions.
+
+This module contains utility functions dedicated to operations
+"""
```
BigRoy, line 4: Pretty sure this docstring provides no information whatsoever currently.

openpype/pipeline/frames.py
```diff
@@ -0,0 +1,62 @@
...
+    """Get frame start string.
+
+    Args:
+        frame_start (int): first frame
+        frame_end (int): last frame
+
+    Returns:
+        str: frame start string
+    """
+    log.debug("frame_start, frame_end: {}, {}".format(frame_start, frame_end))
```
BigRoy, line 22: I'm quite sure both these debug logs are overly verbose currently. Since all logging is handled in OpenPype I'd remove these.
```suggestion
```

openpype/pipeline/frames.py
```diff
@@ -0,0 +1,62 @@
...
+    last_frame_str = str(frame_end)
+    max_padding_length = len(last_frame_str)
+    formatting_str = "{{:0>{}}}".format(max_padding_length)
+
+    log.debug("Formatting string: {}".format(formatting_str))
```
BigRoy, line 27:
```suggestion
```

openpype/pipeline/frames.py
```diff
@@ -0,0 +1,62 @@
...
+
+from openpype.lib import Logger
+
+log = Logger.get_logger(__name__)
+
+
+def get_frame_start_str(frame_start, frame_end):
+    """Get frame start string.
+
+    Args:
```
BigRoy, line 15: I feel like it's quite unclear what this function does reading the docstring.

Basically what it does is return the start frame with padding so that it matches the length of the end frame.
Which in essence would be as simple as:
```python
def get_frame_start_str(frame_start, frame_end):
    padding = len(str(frame_end))
    return str(frame_start).zfill(padding)
```

Example:
```python
print(get_frame_start_str(10, 100))
# 010
```

openpype/pipeline/frames.py
```diff
@@ -0,0 +1,62 @@
...
+
+    handle_start = asset["data"].get("handleStart", 0)
+    handle_end = asset["data"].get("handleEnd", 0)
+
+    return {
+        "frameStart": frame_start,
+        "frameEnd": frame_end,
+        "handleStart": handle_start,
+        "handleEnd": handle_end
+    }
```
BigRoy, line 62: Wasn't there a function for this already? Anyway, I'd expect the function to at least be called `get_asset_frame_range`?

Also, I'm not sure where this is being used - but I feel like in many cases we already have an asset document to begin with and querying the database again would be overkill. This function might be better optimized if it were to take the asset document instead.

openpype/pipeline/publish/publish_plugins.py
```diff
@@ -310,43 +309,21 @@ class ColormanagedPyblishPluginMixin(object):
...
+            "'get_colorspace_settings' is deprecated, "
+            "use 'get_colorspace_settings_from_publish_context' instead"
         )
-
-        # caching settings for future instance processing
-        context.data["imageioSettings"] = (config_data, file_rules)
-
-        return config_data, file_rules
+        # TODO: for backward compatibility, remove in future
+        return get_colorspace_settings_from_publish_context(context.data)
```
BigRoy, line 326: I'm not sure about this. That'd basically mean deprecating the full mixin, right? Since just like this function the `set_representation_colorspace` method is also just calling a single function now too.

openpype/plugins/publish/collect_colorspace_data.py
```diff
@@ -24,10 +24,11 @@ class ExtractColorspaceData(publish.Extractor,
             }
         }
     """
-    label = "Extract Colorspace data"
-    order = pyblish.api.ExtractorOrder + 0.49
+    label = "Collect Colorspace data"
+    order = pyblish.api.CollectorOrder + 0.49
```
BigRoy, line 28: Pretty sure this will break quite a few cases - in the majority of cases representations are NOT set during Collecting stage yet but get defined by the Extractors. For all those cases this would now basically not trigger this logic. Seems like a huge change, no? Unless we're changing all those extractors that are expecting it?

openpype/plugins/publish/collect_farm_target.py
BigRoy, line 1: Honestly - this should just be a Collector within each of the specific modules instead of checking for the modules themselves. It might appear a bit more duplicated but it's at least clear that it's logic specific to those render managers.

openpype/plugins/publish/extract_burnin.py
```diff
@@ -78,6 +78,12 @@ class ExtractBurnin(publish.Extractor):
     options = None
 
     def process(self, instance):
+        # Instance should be integrated on a farm
+        if instance.data.get("farm"):
+            self.log.info(
+                "Instance is marked to be processed on farm. Skipping")
+            return
```
BigRoy, line 85: Note that the logs in the artist facing report do NOT show what plugin is a log is related to. So with all these skipping they'd just see e.g. five lines saying:
> Instance is marked to be processed on farm. Skipping
> Instance is marked to be processed on farm. Skipping
> Instance is marked to be processed on farm. Skipping
> Instance is marked to be processed on farm. Skipping

Having no idea of what actually got skipped.
This is either a design issue with the new publisher OR we should log what is actually being skipped, like "Skipping burnin extraction", "Skipping colorspace transcoding", etc.

BigRoy (commented):
This is a massive PR so I've likely missed A LOT in my first check - but here are a few things that stood out to me.

openpype/plugins/publish/cleanup_explicit.py
```diff
@@ -22,6 +22,12 @@ class ExplicitCleanUp(pyblish.api.ContextPlugin):
     active = True
 
     def process(self, context):
+        # Instance should be integrated on a farm
+        if context.data.get("farm"):
```
fabiaserra, line 26: I have been adding more of these same chec in my fork of OP in the plugins that I know I don't want to run locally (i.e., extract_review, extract_color_transcode...) instead of doing the families removal that we do in certain places to accomplish the same... doing this I find it way more clear than the current "families transfer" logic that we have in `submit_publish_job` to add back the families to the publish job.

Example of a case were "review" family gets removed:
```python
                if "review" in instance.families:
                    # to skip ExtractReview locally
                    instance.families.remove("review")
```
Much better to just add the same on `ExtractReview`?

openpype/plugins/publish/collect_colorspace_data.py
fabiaserra: I wonder if you had to do this for the same reason I ended up having to add this line:
```python
instance.data["colorspace"] = colorspace
```
on the existing "openpype\hosts\traypublisher\plugins\publish\collect_explicit_colorspace.py" plugin that @BigRoy created. It would be better if you could instead reuse that plugin and make it work to your needs.

I basically had to add that line because at the point where the `collect_explicit_colorspace` runs the representations that get submitted to the farm hadn't been generated but adding it to the instance data is enough for then setting the right colorspace when they do

openpype/plugins/publish/collect_farm_target.py
```diff
@@ -0,0 +1,45 @@
...
+        try:
+            deadline_module = context.data.get("openPypeModules")["deadline"]
+            if deadline_module.enabled:
+                instance.data["toBeRenderedOn"] = "deadline"
+                self.log.debug("Collected render target: deadline")
+        except AttributeError:
+            self.log.error("Cannot get OpenPype Deadline module.")
+            raise AssertionError("OpenPype Deadline module not found.")
+
+        try:
```
fabiaserra, line 27: could we return early if the prior farm scheduler was enabled already so we don't need to check on all every time?

openpype/plugins/publish/extract_review.py
```diff
@@ -77,6 +77,13 @@ class ExtractReview(pyblish.api.InstancePlugin):
     profiles = None
 
     def process(self, instance):
+
+        # Instance should be integrated on a farm
+        if instance.data.get("farm"):
```
fabiaserra, line 82: Oh never mind my prior comment haha I see you ended up doing exactly the same I did. Can we now then remove the plugins that delete the "review" family and change the logic behind the families transfer on submission to the farm?

fabiaserra (commented):
@jakubjezek001 thank you very much for starting this!

iLLiCiTiT:
This is huge and needs separated PRs. Every part that can be separated should be separated. E.g. adding `CollectFarmTarget`, adding `if instance.data.get("farm"):` into plugins, traypublisher is completelly separated part, that, to be honest, needs more discussion. There is also a lot of formatting changes that are doing noise in this PR, and it's hard to know if anything changed, or the code just changed formatting.

openpype/pipeline/frames.py
jakubjezek001: True, yeah it is implemented here 3b0840eeed373f32d09b631c4030e91286d96055

openpype/plugins/publish/collect_farm_target.py
jakubjezek001: Development has been relocated to [this location](https://github.com/ynput/OpenPype/pull/5494), which will result in greater coherence in the future. It is necessary for us to reduce the requirements for individual submitters.

openpype/hosts/traypublisher/plugins/publish/collect_simple_instances.py
```diff
@@ -47,90 +52,86 @@ def process(self, instance):
...
+            filepath_items.append(review_file_item)
+
+        representation_files = self.get_processing_file_data(
+            instance, filepath_items)
+
+        for file_path, file_data in representation_files.items():
+            # making sure file extension is publishable over farm
+            file_ext = os.path.splitext(file_path)[-1].lower().lstrip(".")
+            if file_ext not in self.pixel_ext:
+                # allow farm publishing only for pixel type data
```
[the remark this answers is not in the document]
jakubjezek001: To be honest, I wanted to restrain it for now during this POC stage, but perhaps this can be lifted up later if we see it might not break anything.

openpype/hosts/houdini/plugins/publish/collect_mantra_rop.py
```diff
@@ -31,49 +31,65 @@ class CollectMantraROPRenderProducts(pyblish.api.InstancePlugin):
...
+        render_products.append(beauty_product)
+
+        files_by_aov = {
+            "beauty": expected_files.generate_expected_filepaths(
+                frame_start, frame_end, beauty_product)
+        }
+
+        # get the number of AOVs
+        aov_numbers = rop.evalParm("vm_numaux")
+        if aov_numbers < 0:
```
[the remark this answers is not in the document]
jakubjezek001: addet here fc42095d7e6bd37fd0e65bc4f2f50446f2d96378

openpype/hosts/houdini/plugins/publish/collect_mantra_rop.py
jakubjezek001: added here fc42095d7e6bd37fd0e65bc4f2f50446f2d96378

openpype/hosts/houdini/plugins/publish/collect_mantra_rop.py
```diff
@@ -31,49 +31,65 @@ class CollectMantraROPRenderProducts(pyblish.api.InstancePlugin):
...
-            self.log.debug("Chunk Size: %s" % chunk_size)
-
-            default_prefix = evalParmNoFrame(rop, "vm_picture")
-            render_products = []
-
-            # Default beauty AOV
-            beauty_product = self.get_render_product_name(
-                prefix=default_prefix, suffix=None
+        if not chunk_size_parm:
+            return
```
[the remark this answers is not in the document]
jakubjezek001: you are right, it was wrong also in another plugin.. I was only rewriting the indention flow and haven't thought about the logic.. Here it is fixed 33214801f4f35601ef82f4be13a5f0cbbdf028dc

openpype/pipeline/frames.py
jakubjezek001: addressed here 488a2ff7a093f5c0aafe11deb09e5088dc7bfa64

openpype/pipeline/frames.py
jakubjezek001: addressed here 488a2ff7a093f5c0aafe11deb09e5088dc7bfa64

openpype/pipeline/frames.py
jakubjezek001: thanks here it is added as you suggested 859c9ac139d7393757852481ac1be7c9610b7c33

openpype/pipeline/publish/publish_plugins.py
jakubjezek001: Agree this is needs to be done differently. Look to https://github.com/ynput/OpenPype/pull/5498 the rework.

openpype/hosts/nuke/plugins/publish/validate_rendered_frames.py
jakubjezek001: variables were improved in #5500

openpype/pipeline/farm/pyblish_functions.py
jakubjezek001: > I had to remove this hard-code in my fork of OP because we are publishing plates, reference and other families in the farm

@fabiaserra proposed code is suggesting that since you set target to farm suffixe `.farm` or `.farm_frames` will be added to what ever base family [product type]. The `render` should be onlly fallback for those missing the suffix - mostly the backward compatibility.

tokejepsen:
@jakubjezek001 is this PR still needed? The ticket has been closed.

jakubjezek001:
closing this PR since it was just for research purpose only

PR #5453, opened 2023-08-15
Fix typo on Deadline OP plugin name

fabiaserra:
## Changelog Description
Surprised that no one has hit this bug yet... but it seems like there was a typo on the name of the OP Deadline plugin when submitting jobs to it.

## Testing notes:
1. Submit a publish job to Deadline


mustafa-zarkash:
I don't think it was a bug. 
I tested this PR as a user and didn't notice any difference
actually, this also works, maybe `deadline_plugin` is not case sensitive. 
```python
deadline_plugin = openpype"
```

fabiaserra:
> I don't think it was a bug. I tested this PR as a user and didn't notice any difference actually, this also works, maybe `deadline_plugin` is not case sensitive.
> 
> ```python
> deadline_plugin = openpype"
> ```

It's definitely a bug for us, "Openpype" is not recognized as the "OpenPype" plugin in deadline

mustafa-zarkash (approved):
It works, tested in windows.

kalisp (approved):

PR #5459, opened 2023-08-16
Loader: Add colorspace entry to Version Info

fabiaserra:
## Changelog Description
Small addition to add "Colorspace" entry in Version Info tab in loader if version has colorspace data

image

## Testing notes:
1. Open the loader and click on a Version entry that has colorspace data

### DIFF
--- a/openpype/tools/loader/widgets.py
+++ b/openpype/tools/loader/widgets.py
@@ -741,6 +741,16 @@ class VersionTextEdit(QtWidgets.QTextEdit):
             "source": source_label
         }
 
+        # Add colorspace info to Version Info tab
+        version_colorspace = version_doc["data"].get("colorspace")
+        colorspace_html = ""
+        if version_colorspace:
+            colorspace_html = "<b>Colorspace</b><br>{colorspace}<br><br>".format(
+                colorspace=version_colorspace
+            )
+
+        data["colorspace_html"] = colorspace_html
+
         self.setHtml((
             "<h2>{subset}</h2>"
             "<h3>{version}</h3>"
@@ -749,7 +759,7 @@ class VersionTextEdit(QtWidgets.QTextEdit):
 
             "<b>Created</b><br>"
             "{created}<br><br>"
-
+            "{colorspace_html}"
             "<b>Source</b><br>"
             "{source}"
         ).format(**data))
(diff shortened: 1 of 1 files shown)


mustafa-zarkash:
I've tested it in Houdini and Maya, and this PR doesn't seem to affect them.

openpype/tools/loader/widgets.py
```diff
@@ -741,6 +741,16 @@ def set_version(self, version_doc=None, version_id=None):
             "source": source_label
         }
 
+        # Add colorspace info to Version Info tab
+        version_colorspace = version_doc["data"].get("colorspace")
+        colorspace_html = ""
```
BigRoy, line 746: Isn't the colorspace on the representations instead of the version?

BigRoy:
> I've tested it in Houdini and Maya, and this PR doesn't seem to affect them.

It should affect them, but only for publishes that have colorspace data written into them. If you're not seeing the information right hand side even if the colorspace data exists on the version/representations than I'd say it's a bug.

openpype/tools/loader/widgets.py
fabiaserra: I checked a few entries in MongoDB and I can only see it on the version data

fabiaserra:
> > I've tested it in Houdini and Maya, and this PR doesn't seem to affect them.
> 
> It should affect them, but only for publishes that have colorspace data written into them. If you're not seeing the information right hand side even if the colorspace data exists on the version/representations than I'd say it's a bug.

Exactly

openpype/tools/loader/widgets.py
jakubjezek001: In fact @BigRoy is right @fabiaserra. The new colorspaceData distribution is kept on representation level. The version data colorspace key is still available but will be removed soon.

jakubjezek001:
I am not sure how to put my mind into this. I appreciate your attempt @fabiaserra but I am afraid this will only work for couple months only. I am suggesting to close it. But nothing will be broken if we merge it ;)

openpype/tools/loader/widgets.py
fabiaserra: yeah fair enough, no big deal, it was just a fast utility for me while testing the OpenRV integration to make sure what I was loading was getting the expected colorspace, happy to close it

PR #5464, opened 2023-08-17
Maya: Context plugin shouldn't be tied to family

kalisp:
## Changelog Description
`Maya Current File` collector was tied to `workfile` unnecessary. It should run even if `workile` instance is not being published.

## Testing notes:
1. Disable `workfile` instance and publish
2. `Collect Scene Version` shouldnt fail


BigRoy (approved):
Change makes sense to me.

moonyuet:
Code looks good to me, and i tested it by myself. The instance published successfully without workfile instance.
image

moonyuet (approved):

PR #5475, opened 2023-08-21
AYON: Fill entities during editorial

iLLiCiTiT:
## Changelog Description
Fill entities and update template data on instances during extract AYON hierarchy.

## Additional info
I think we need to change the fill of `anatomyData`, they should be prefilled in collection phase -> Future feature.

## Testing notes:
1. Run editorial publishing in AYON
2. Make sure the publishing would create new folders and tasks in project
3. The publishing should work as expected and get over integration
- Validate the hierarchy is correct, the subsets are under correct folders and files are at correct places


Minkiu (approved):
Haven't tested, but code looks good, maybe we could add docstrings.

jakubjezek001 (approved):
LGTM!

PR #5476, opened 2023-08-22
Unreal: Create project in temp location and move to final when done

antirotor:
## Changelog Description
Create Unreal project in local temporary folder and when done, move it to final destination.

## Additional info
Project creation in Unreal requires directory to be clean without any content whatsoever. But in some storage scenarios, metadata files are created and thus will project creation fail. Solution is to create it locally in temporary folder and move it to final destination only when it is done.

## Testing notes:
1. Create new Unreal project
2. It should be created as expected and copied to the final destination specified by project templates.

### DIFF
--- a/openpype/hosts/unreal/hooks/pre_workfile_preparation.py
+++ b/openpype/hosts/unreal/hooks/pre_workfile_preparation.py
@@ -233,8 +233,8 @@ class UnrealPrelaunchHook(PreLaunchHook):
                                          Path(temp_dir))
                 try:
                     self.log.info((
-                         f"Moving from {temp_dir} to "
-                         f"{project_path.as_posix()}"
+                        f"Moving from {temp_dir} to "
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
Just passing by: I think this should be optional and turned off by default, it can easily double the time of startup.

antirotor:
> Just passing by: I think this should be optional and turned off by default, it can easily double the time of startup.

Why do you think? I think letting it create locally and then copy the whole stuff to network share is actually faster. No numbers on that though. Important fact about this is that newly created project are really small - like 15Mb of data or so.

iLLiCiTiT:
>  I think letting it create locally and then copy the whole stuff to network share is actually faster. No numbers on that though. Important fact about this is that the newly created project are really small - like 15Mb of data or so.

Aaah, in that case I guess it's ok.

simonebarbieri (approved):

PR #5479, opened 2023-08-22
OCIO: windows unc path support in Nuke and Hiero

jakubjezek001:
## Changelog Description
Hiero and Nuke is not supporting windows unc path formatting in OCIO environment variable.

## Testing notes:
1. Add ocio config with unc path
2. open nuke and check that the unc path is added into custom config.

### DIFF
--- a/openpype/hooks/pre_ocio_hook.py
+++ b/openpype/hooks/pre_ocio_hook.py
@@ -45,6 +45,9 @@ class OCIOEnvHook(PreLaunchHook):
         if config_data:
             ocio_path = config_data["path"]
 
+            if self.host_name in ["nuke", "hiero"]:
+                ocio_path = ocio_path.replace("\\", "/")
+
             self.log.info(
                 f"Setting OCIO environment to config path: {ocio_path}")
 
--- a/openpype/hosts/nuke/api/lib.py
+++ b/openpype/hosts/nuke/api/lib.py
@@ -2107,8 +2107,9 @@ class WorkfileSettings(object):
 
         # set ocio config path
         if config_data:
+            config_path = config_data["path"].replace("\\", "/")
             log.info("OCIO config path found: `{}`".format(
-                config_data["path"]))
+                config_path))
 
             # check if there's a mismatch between environment and settings
             correct_settings = self._is_settings_matching_environment(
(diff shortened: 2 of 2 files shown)


Minkiu (approved):

BigRoy:
What an interesting fix. Especially because backslashes is exactly what is the Windows format - so it's interesting to see that in particular windows trips over backslashed UNC paths.

E.g. in Windows explorer `//server/share` is an invalid UNC path but `\\server\share` does work.
In Python both work however if I do `os.path.exists` so it seems it's particularly an OCIO bug or is Nuke just crazy? :D

jrsndl:
I've tested with Nuke 14 and Windows 10. Looks like Nuke protests and then fixes itself?
image
This dialog now opens for me every time, I guess artists will be worried.


image
image
image
image


Maybe the issue is not connected to UNC, but Settings just do not expect absolute paths? Checking the Nuke script:
image

BigRoy:
@jrsndl In your case this is likely the issue, `\t` is a special character used to represent TAB. Likely `\t` gets passed as a string to Nuke and thus must be converted to forward slashes. That's why it likely gets cut off at `\T`

You might be better off by making sure backslashes are escaped to begin with when applying it in Nuke, e.g. `path.replace("\\", "\\\\")` basically escaping every `\` in the string itself by making it `\\` which in a string (and being escaped) represents the single character `\`.

Minkiu:
I wonder if it's worth using forward backslashes in the OP config regardless of the workstation OS? Can you try to see if that fixes it @jrsndl

jrsndl:
> I wonder if it's worth using forward backslashes in the OP config regardless of the workstation OS?
Yeah, I normally use forward slashes (made backslashes to stress - test, sorry for not mentioning that).

The forward slashes do not fix the Nuke message about the error in OCIO config.
//dazzle.local/shares/pype_dev/projects/T029_OP3_16_3_tst/cio/aces_1.2/config.ocio


Using {root[work]} instead of absolute path doesn't silence Nuke either.
{root[work]}/T029_OP3_16_3_tst/ocio/aces_1.2/config.ocio

Minkiu:
Hey @jakubjezek001 are you able to reproduce what @jrsndl  is saying above?

jakubjezek001:
I noticed the issue that @jrsndl mentioned. It occurred only once, after I saved the script following the PR change test. When you open it again, it should work correctly. Why does this happen? Before testing, the path was saved with Windows slashes in the script. Then, when it is opened with the PR changes, Nuke recognizes an older path in the root customOCIOConfigPath
attribute and validates it, resulting in the issue being reported. However, once the script is saved, it is fixed, and the next time it is opened, it should work fine.

jakubjezek001:
@jrsndl try it now after the d3c1b84835aed4579ef7909e52631ab0732d6796 it might be improved.

Minkiu (approved):

PR #5480, opened 2023-08-22
Deadline: Added super call to init

kalisp:
## Changelog Description
DL 10.3 requires plugin inheriting from DeadlinePlugin to call super's __init__ explicitly.

## Testing notes:
1. redeploy plugins from `openpype/modules/deadline/repository/custom` to DEADLINE_REPO/custom folder


m-u-r-p-h-y (approved):
I can confirm Deadline 10.3.x is working as expected now

image

PR #5481, opened 2023-08-22
Houdini: Add Static Mesh product-type (family)

MustafaJafar:
## Changelog Description
This PR adds support to publish Unreal Static Mesh in Houdini as FBX 

Quick recap 
- [x] Add UE Static Mesh Creator
- [x] Dynamic subset name like in Maya
- [x] Collect Static Mesh Type
- [x] Update collect output node
- [x] Validate FBX output node 
- [x] Validate mesh is static
- [x] Validate Unreal Static Mesh Name
- [x] Validate Subset Name
- [x] FBX Extractor
- [x] FBX Loader 
- [x] Update OP Settings
- [x] Update AYON Settings

## Testing notes:
Create instance while selecting 
1. an Obj Level subnetwork with some node inside it. 
2. an obj node (e.g. a box)
3. a SopNode


mkolar:
@mustafa-zarkash this PR goes against most of OP concepts. FBX should can be representation of various product types, but should not be a product type in itself.

If we want to add fbx support to houdini (I'd like to hear the usecase btw), you can keep the extractor, but only add it to existing families where it makes sense and make it optional and by default disabled in the settings.

BigRoy:
Out of curiosity - what existing family exists that makes sense for `fbx` to be a part of it.

FBX supports rigs, geometry, (basic) materials, can even embed textures. It's a mad format.
Also, I believe Maya also has an fbx family, right?

mustafa-zarkash:
> FBX supports rigs, geometry, (basic) materials, can even embed textures. It's a mad format.
Yeah, that's a common misconception with all openpype/ayon new dev comers! 
there's a significant difference between a file format and what data it represents . 

 > Also, I believe Maya also has an fbx family, right?
  It's yes and no!, 
  Maya has FBX support but it uses it to publish unreal static mesh.
  which explains the existence of fbx extractor!

> If we want to add fbx support to houdini (I'd like to hear the usecase btw)
Ondraj suggested we can convert this PR to publish "Unreal - Static Mesh" from houdini just as in maya

BigRoy:
Ah yes, if this is for exporting static meshes only the FBX support is way simpler of course.

openpype/hosts/houdini/plugins/load/load_fbx.py
```diff
@@ -0,0 +1,171 @@
...
+
+        self[:] = nodes
+
+        # Call containerise function which does some
+        # automations for you
+        containerised_nodes = self.get_containerised_nodes(
+            nodes, context, node_name, namespace
+        )
+
+        return containerised_nodes
```
moonyuet, line 50: You may consider to return pipeline.containerise function here instead of having an extra get_containerised_node function to call "pipeline.containerise" function. You can take reference from other loaders.

openpype/hosts/houdini/plugins/publish/collect_fbx_type.py
```diff
@@ -0,0 +1,59 @@
...
+
+    # overrides InstancePlugin.process()
+    def process(self, instance):
+
+        if instance.data["creator_identifier"] == "io.openpype.creators.houdini.unrealstaticmesh":  # noqa: E501
+            # such a condition can be used to differentiate between
+            #  instances by identifier becuase sometimes instances
+            #  may have the same family but different identifier
+            #  e.g. bgeo and alembic
+            instance.data["families"] += ["fbx"]
```
moonyuet, line 33: Revisiting to the same staticmesh family in maya, staticmesh family extracts both abc and fbx format. If this family in Houdini acts similarly to Maya, it should be adding two families: `fbx` and `abc`. (You can find the codes of staticmesh family in Maya Hosts)
Maybe consider to rename this collector to `collect_unreal_staticmesh` and remove fbx as families in line 19.  (Otherwise, you are adding two `fbx` into families.)
```suggestion
        if instance.data["creator_identifier"] == "io.openpype.creators.houdini.unrealstaticmesh":  # noqa: E501
            # such a condition can be used to differentiate between
            #  instances by identifier becuase sometimes instances
            #  may have the same family but different identifier
            #  e.g. fbx and alembic
            instance.data["families"] += ["fbx"]
            instance.data["families"] += ["abc"]
```

openpype/hosts/houdini/plugins/load/load_fbx.py
mustafa-zarkash: You are totally right. 
I did it this way to make it easier to understand because this whole PR is part of "publish process development guide" that we also want it to work for production.

openpype/hosts/houdini/plugins/publish/collect_fbx_type.py
mustafa-zarkash: I think it might make it harder to handle collision prefixes (which basically checking path attribute values). 
anyways, I will revisit the  same staticmesh family in maya to learn more. 

I'll appreciate any recommendations.

openpype/hosts/houdini/plugins/load/load_fbx.py
moonyuet: > You are totally right. I did it this way to make it easier to understand because this whole PR is part of "publish process development guide" that we also want it to work for production.
> 
> please tell me if you think I don't have to do this and it's already easy to read

In my opinion, I think it is better to stick with using pipeline.containerise for alignment with the codes. You can find similar containerise functions in other Hosts like Maya, Max and Nuke.

openpype/hosts/houdini/plugins/publish/collect_fbx_type.py
moonyuet: > I think it might make it harder to handle collision prefixes (which basically checking path attribute values). anyways, I will revisit the same staticmesh family in maya to learn more.
> 
> I wonder if it's possible to publish same data in to representations in houdini ? In maya it can publish the same set with different commands, however in houdini we should create a rop for each representation? so, is it possible for a publish instance to have two rops ?
> 
> I'll appreciate any recommendations. @moonyuet @BigRoy

Take reference from pointcache family in Houdini. It is two creators but same families. But It's only my suggestion. You can always do the abc extractor later if there is any user wanted in staticmesh family.

mustafa-zarkash: > so, is it possible for a publish instance to have two rops ?

I don't know what will be the use case for that in production 

I can think of two options : 
1. by relying on subnets with several instances inside, each instance can have a different variant post fix ? 
2. Create multiple instances with different variants ? 

Fool Test: *Family was randomly picked*
image

openpype/hosts/houdini/plugins/load/load_fbx.py
```diff
@@ -0,0 +1,171 @@
...
+        try:
+            file_node = next(
+                n for n in node.children() if n.type().name() == "file"
+            )
+        except StopIteration:
+            self.log.error("Could not find node of type `file`")
+            return
+
+        # Update the file path
+        file_path = get_representation_path(representation)
```
BigRoy, line 64: Unlike `load()` which uses `self.get_file_path` this does not do `os.normpath` and replacing of backslashes to forward slashes. Is there any way we can do the same call here using the context?

openpype/hosts/houdini/plugins/load/load_fbx.py
```diff
@@ -0,0 +1,171 @@
...
+        file_path = self.format_path(file_path, representation)
```
BigRoy, line 65: Where is `format_path` defined?

openpype/hosts/houdini/plugins/load/load_fbx.py
```diff
@@ -0,0 +1,171 @@
...
+        null.setInput(0, attribdelete)
+
+        # Ensure display flag is on the file_node input node and not on the OUT
+        # node to optimize "debug" displaying in the viewport.
+        file_node.setDisplayFlag(True)
+
+        # Set new position for unpack node else it gets cluttered
+        nodes = [parent_node, file_node, attribdelete, null]
+        for nr, node in enumerate(nodes):
+            node.setPosition([0, (0 - nr)])
```
BigRoy, line 151: Wouldn't `parent_node.layoutChildren()` suffice?
And maybe even `parent_node.moveToGoodPosition()` as well?

openpype/hosts/houdini/plugins/load/load_fbx.py
```diff
@@ -0,0 +1,171 @@
...
+        containerised_nodes = pipeline.containerise(
+            node_name,
+            namespace,
+            nodes,
+            context,
+            self.__class__.__name__,
+            suffix="",
+        )
+
+        return containerised_nodes
```
BigRoy, line 171: Putting this in its own function to me doesn't 'help' that much because it's just a single function call. It's not really doing that much at all.

Plus, even more confusing is that `get_containerised_nodes` hints that it's just retrieving information from the scene but it is not - it is actually putting node into a container (grouping them).

openpype/hosts/houdini/plugins/publish/collect_fbx_type.py
```diff
@@ -0,0 +1,59 @@
...
+        # get output node
+        node = hou.node(instance.data["instance_node"])
+        out_node = node.parm("startnode").evalAsNode()
+
+        if not out_node:
+            self.log.warning("No output node collected.")
+            return
+
+        self.log.debug("Output node: %s" % out_node.path())
+        return out_node
```
BigRoy, line 59: To match other families you might be better off just relying on `lib.get_output_parameter` and adding the `filmboxfbx` node type to return `node.parm("startnode")`. Then you can just add the relevant fbx family to `collect_output_node`.

openpype/hosts/houdini/plugins/publish/collect_fbx_type.py
```diff
@@ -0,0 +1,59 @@
...
+
+        # Update instance.data with ouptut_node
+        out_node = self.get_output_node(instance)
+
+        if out_node:
+            instance.data["output_node"] = out_node
+
+        # Disclaimer : As a convntin we use collect_output_node.py
+        #   to Update instance.data with ouptut_node of different types
+        #   however, this collector is used for demonstration
```
BigRoy, line 43: So you're saying it's not needed, but why would you still like to "demonstrate" this here? I'm a bit unsure what you're asking here.

openpype/hosts/houdini/plugins/publish/collect_fbx_type.py
```diff
@@ -0,0 +1,59 @@
...
+
+Collectors act as a pre process for the validation stage.
+It is used mainly to update instance.data
+
+P.S.
+    There are some collectors that run by default
+    for all types.
+
+This plugin is part of publish process guide.
+"""
```
BigRoy, line 11:
```suggestion
"""Collector for filmboxfbx types."""
```

You're basically documenting here how pyblish and openpype work - that's a documentation problem, not one to be put in every collector python file I'd say.

BigRoy (commented):
I noticed the code is loaded with comments that are "explaining" how OpenPype and Pyblish works. Even though that's very nice - I'd say it's not up to the code itself explaining how the bigger system works. Even though I like how you're explaining things I'd say it's better removed from the PR and moved into documentation if a specific element is not found in developer's documentation yet.

openpype/hosts/houdini/plugins/publish/collect_fbx_type.py
mustafa-zarkash: I will save this note!

mustafa-zarkash:
> It's better removed from the PR and moved into documentation if a specific element is not found in developer's documentation yet.

That's make sense! 
I'll save this note. 
So, Let's make this PR work first, Then at the end I move these comments to the `publish process development` guide. 
I can't wait to finish it and share it with you

openpype/hosts/houdini/plugins/load/load_fbx.py
mustafa-zarkash: My bad, I forgot it :'D

openpype/hosts/houdini/plugins/load/load_fbx.py
mustafa-zarkash: `parent_node.layoutChildren()` is suffice!

openpype/hosts/houdini/plugins/publish/collect_fbx_type.py
mustafa-zarkash: I thought it would be simpler to understand if  written in  a separate plugin. 
but this doesn't help as you mentioned!

openpype/hosts/houdini/plugins/publish/collect_fbx_type.py
mustafa-zarkash: oh, it makes sense. 
but it is not used in `CollectOutputSOPPath` ! 
I will save that in a  note.

mustafa-zarkash: I found that `lib.get_output_parameter` is used to return file output path 
Anyways, I'll  just add the relevant fbx family to `collect_output_node`.

openpype/hosts/houdini/plugins/create/create_unreal_staticmesh.py
```diff
@@ -0,0 +1,213 @@
+# -*- coding: utf-8 -*-
+"""Creator plugin for creating Unreal Static Meshes.
+
+Unreal Static Meshes will be published as FBX.
+
+Filmbox by default expects an ObjNode
+however, we set the sop node explictly
+to eleminate any confusion.
```
BigRoy, line 8:
```suggestion
to eliminate any confusion.
```

openpype/hosts/houdini/plugins/create/create_unreal_staticmesh.py
```diff
@@ -0,0 +1,213 @@
...
+
+This will make Filmbox to ignore any object transformations!
```
BigRoy, line 10: If forcing a `SopNode` over `ObjNode` causes 'lack of features' in cases where one might WANT those transformations we might want to allow that?

openpype/hosts/houdini/plugins/create/create_unreal_staticmesh.py
```diff
@@ -0,0 +1,213 @@
...
+
+get_obj_output selects
+the output sop with mimimum idx
+or the node with render flag isntead.
```
BigRoy, line 14:
```suggestion
get_obj_output selects
the output sop with mimimum idx
or the node with render flag instead.
```

But it's probably best to just remove it:
```suggestion
```

mustafa-zarkash: please bare with me, I'll remove a lot of comments in upcoming commits

openpype/hosts/houdini/plugins/create/create_unreal_staticmesh.py
mustafa-zarkash: I ran out of ideas tbh. 

How can we do some validations without specifying the output node explicitly 
to be specific "path" attribute 
I keep think about how to handle collision prefixes

moonyuet: Just want to clear up some of my confusion, what kind of collision prefixes you want to collect before exporting the fbx to Unreal? Are you talking about LOD(Level of Details) or some animation data? If you talk about LODs, I am not sure about the LODs of the filmbox fbx works the same with the HDA. All I know is Sidefbx suggests to export the piece of LODs data to HDA and import the HDA to Unreal. You can find the doc here: https://www.sidefx.com/docs/houdini/unreal/collisionsetup.html#lod
If you talk about animation data, I dont think you need that for static mesh. You can always set up collision data with static mesh in Unreal. 
Or are you talking about axis issue? We need to set the Y Axis up for unreal if the axis is not right(of which Z axis goes up).

mustafa-zarkash: > what kind of collision prefixes

I mean that, meshes can't start with these name 
and I think I have two options 
A. Check Object names (I'm going to try it) 
B. Check Path attribute (can be a pain)

image

mustafa-zarkash:
### Checkpoint 
I made it less restrictive, user can select a `SopNode` or an `ObjNode` 
It checks node names for collision. 

It still lacks 
- Validate FBX output node
- Validate mesh is static
- Skip node name validation if selection is an `ObjNode` and `buildfrompath` is on 


### Note:
> Comments in files might not be updated, I'll clean them as soon as I finish the functionality

mustafa-zarkash:
## Checkpoint
@BigRoy  @moonyuet @antirotor

- UE Static mesh creator allows selecting sop node or obj node 
- Static mesh collector adds fbx family which triggers fbx extractor 
- It's possible to use other extractors but I'm not sure if that would break the validators
- It includes three validations 
    - `ValidateOutputNode` which should check if output node exists and it has geometry. 
    - `ValidateMeshIsStatic` which checks if output node or it's output children are time dependent
    - `ValidateUnrealStaticMeshName` which checks if  output node or it's output children includes collision prefix as well as checking the subset name
- it includes fbx extractor and loader

It still lacks: 
- Better error messages, I intent to use xml (to provide better guidance about dealing with errors)
- Checking geometry existence, which is not yet included `ValidateOutputNode`

openpype/hosts/houdini/api/lib.py
```diff
@@ -649,3 +649,43 @@ def get_color_management_preferences():
...
+        for child in output_node.children():
+            out_list += get_output_children(child, include_sops=include_sops)
+
+    elif include_sops and \
+            output_node.childTypeCategory() == hou.sopNodeTypeCategory():
+        out = get_obj_node_output(output_node)
+        if out:
+            out_list += [out]
+
+    return out_list
```
BigRoy, line 691: Sorry, what is this for? I feel like the naming on this is weird, `get_output_children` but it also always return itself. That is not a child?

Also, if it is similar to `allNodes()` why not use all nodes?

```python
result = {output_node}  # include self
nodes = output_node.allNodes()

obj_category = hou.objNodeTypeCategory()
sop_capegory = hou.sopNodeTypeCategory()

for node in nodes:
    if node.childTypeCategory() == sop_category:
        out = get_obj_node_output(output_node)
        result.add(out)

return list(result)
```
Only difference is that this would recurse into other types than `objNodeTypeCategory`.
Any way, could you explain what your use case is? I noticed you're using it to validate for 'animation' but do you really need to recurse further than just the main output - how does that influence the extraction?

openpype/hosts/houdini/plugins/create/create_unreal_staticmesh.py
```diff
@@ -0,0 +1,164 @@
...
+    ):
+        """
+        The default subset name templates for Unreal include {asset} and thus
+        we should pass that along as dynamic data.
+        """
+        dynamic_data = super(CreateUnrealStaticMesh, self).get_dynamic_data(
+            variant, task_name, asset_doc, project_name, host_name, instance
+        )
+        dynamic_data["asset"] = asset_doc["name"]
+        return dynamic_data
```
BigRoy, line 87: Why is this needed? The default publish templates also have `{asset}` in it and work just fine.
I'm pretty sure this is redundant. I see `asset` in the folder and file templates for `render`, `publish`, `hero`, etc.
What's so special here?

openpype/hosts/houdini/plugins/create/create_unreal_staticmesh.py
```diff
@@ -0,0 +1,164 @@
...
+        parms = {
+            "startnode": selection,
+            "sopoutput": output_path,
+            "vcformat": vcformat,
+            "convertunits": convertunits,
+            "trange": trange,
+            "createsubnetroot": createsubnetroot
+        }
+
+        return parms
```
BigRoy, line 121: The `# get` commands plus the `get_parms` method name are confusing to me. It's not getting anything from an houdini parm but is just defining the default values. To me this would be just as readable if it were:
```python
        output_path = hou.text.expandString("$HIP/pyblish/{}.fbx".format(subset_name))
        instance_node.setParms({
            "startnode": selection,
            "sopoutput": output_path,
            "vcformat": pre_create_data.get("vcformat"),
            "convertunits": pre_create_data.get("convertunits"),
            "trange": 1,
            "createsubnetroot": pre_create_data.get("createsubnetroot")
        })
```

If there are specific parm names that are confusing, like `vcformat` I would likely comment like this:
```python
        output_path = hou.text.expandString("$HIP/pyblish/{}.fbx".format(subset_name))
        instance_node.setParms({
            "startnode": selection,
            "sopoutput": output_path,
            # vertex cache format
            "vcformat": pre_create_data.get("vcformat"),
            "convertunits": pre_create_data.get("convertunits"),
            # set render range to use frame range start-end frame
            "trange": 1,
            "createsubnetroot": pre_create_data.get("createsubnetroot")
        })
```

openpype/hosts/houdini/plugins/load/load_fbx.py
```diff
@@ -0,0 +1,150 @@
...
+        try:
+            file_node = next(
+                n for n in node.children() if n.type().name() == "file"
+            )
+        except StopIteration:
+            self.log.error("Could not find node of type `file`")
+            return
+
+        # Update the file path
+        file_path = self.get_file_path(representation=representation)
```
BigRoy, line 60: Just wondering - does this work?
```suggestion
        file_path = self.get_file_path(context=representation["context"])
```

openpype/hosts/houdini/plugins/load/load_fbx.py
```diff
@@ -0,0 +1,150 @@
...
+        null.setInput(0, attribdelete)
+
+        # Ensure display flag is on the file_node input node and not on the OUT
+        # node to optimize "debug" displaying in the viewport.
+        file_node.setDisplayFlag(True)
+
+        # Set new position for children nodes
+        parent_node.layoutChildren()
+
+        # Retrun all the nodes
```
BigRoy, line 149:
```suggestion
        # Return all the nodes
```

openpype/hosts/houdini/plugins/publish/extract_fbx.py
```diff
@@ -0,0 +1,70 @@
...
+
+        # get rop node
+        ropnode = hou.node(instance.data.get("instance_node"))
+
+        # render rop
+        render_rop(ropnode)
+
+        # get required data
+        file_name, staging_dir = self.get_paths_data(ropnode)
+        representation = self.get_representation(instance,
```
BigRoy, line 30: `get_representation` is a confusing name since it doesn't get or query an existing entry. I'd rather call it `prepare_representation`

openpype/hosts/houdini/plugins/publish/extract_fbx.py
```diff
@@ -0,0 +1,70 @@
...
+    def get_paths_data(self, ropnode):
+        # Get the filename from the filename parameter
+        output = ropnode.evalParm("sopoutput")
+
+        staging_dir = os.path.normpath(os.path.dirname(output))
+
+        file_name = os.path.basename(output)
+
+        self.log.info("Writing FBX '%s' to '%s'" % (file_name,
+                                                    staging_dir))
```
BigRoy, line 51: This feels very weird as a log inside `get_paths_data`. This should really be just BEFORE the writing of the rop node.

Also note that this should be a debug log. The artist doesn't care what temp staging directory/file gets extracted (especially because the logs are not visible during publishing but only after the publisher) they just want the final publish result.

openpype/hosts/houdini/plugins/publish/extract_fbx.py
```diff
@@ -0,0 +1,70 @@
...
+
+        return file_name, staging_dir
```
BigRoy, line 53: Somehow I'd really expect a folder to come in front of the filename.

openpype/hosts/houdini/plugins/publish/validate_mesh_is_static.py
```diff
@@ -0,0 +1,61 @@
...
+from openpype.hosts.houdini.api.lib import get_output_children
+
+import hou
+
+
+class ValidateMeshIsStatic(pyblish.api.InstancePlugin,
+                           OptionalPyblishPluginMixin):
+    """Validate mesh is static.
+
+    It checks if output node is time dependant.
```
BigRoy, line 20:
```suggestion
    It checks if output node is time dependent.
```

openpype/hosts/houdini/plugins/publish/validate_mesh_is_static.py
```diff
@@ -0,0 +1,61 @@
...
+    hosts = ["houdini"]
+    label = "Validate Mesh is Static"
+    order = ValidateContentsOrder + 0.1
+    actions = [SelectInvalidAction]
+
+    def process(self, instance):
+
+        invalid = self.get_invalid(instance)
+        if invalid:
+            nodes = [n.path() for n in invalid if isinstance(n, hou.Node)]
```
BigRoy, line 33: What other types than `hou.Node` are you expecting? And if `get_invalid` is returning those are we sure we want them returned to the `SelectInvalidAction`?

openpype/hosts/houdini/plugins/publish/validate_output_node.py
```diff
@@ -0,0 +1,55 @@
...
+from openpype.pipeline import PublishValidationError
+from openpype.hosts.houdini.api.action import (
+    SelectInvalidAction,
+    SelectROPAction,
+)
+
+import hou
+
+
+class ValidateOutputNode(pyblish.api.InstancePlugin):
```
BigRoy, line 12: I feel like this class name and plug-in label is a bit too generic for an FBX specific check for validating whether the output node is either SOP or ObjNode. Should we be more specific to avoid name clashes in the future?

openpype/hosts/houdini/plugins/publish/validate_output_node.py
```diff
@@ -0,0 +1,55 @@
...
+    """Validate the instance Output Node.
+
+    This will ensure:
+        - The Output Node Path is set.
+        - The Output Node Path refers to an existing object.
```
BigRoy, line 17: The docstring should explain that also a check is done that the node is either ObjNode or SopNode.

openpype/hosts/houdini/plugins/publish/validate_output_node.py
```diff
@@ -0,0 +1,55 @@
...
+    hosts = ["houdini"]
+    label = "Validate Output Node"
+    actions = [SelectROPAction, SelectInvalidAction]
+
+    def process(self, instance):
+
+        invalid = self.get_invalid(instance)
+        if invalid:
+            raise PublishValidationError(
+                "Output node(s) are incorrect",
```
BigRoy, line 31: The output node is never supposed to be plural, right? Or can it?

openpype/hosts/houdini/plugins/publish/validate_unreal_staticmesh_naming.py
```diff
@@ -0,0 +1,108 @@
...
+
+    optional = True
+
+    @classmethod
+    def apply_settings(cls, project_settings, system_settings):
+        settings = (
+            project_settings["houdini"]["create"]["CreateUnrealStaticMesh"]
+        )
+        cls.collision_prefixes = settings["collision_prefixes"]
+        cls.static_mesh_prefix = settings["static_mesh_prefix"]
```
BigRoy, line 45: This is lacking the reading of settings for this plug-ins `optional`, `active` and `enabled` state from `houdini/publish/ValidateUnrealStaticMeshName`. Since you've overridden the function you need to apply it yourself as far as I know.

openpype/hosts/houdini/plugins/publish/validate_unreal_staticmesh_naming.py
```diff
@@ -0,0 +1,108 @@
...
+    """Validate name of Unreal Static Mesh.
+
+    This validator checks if output node name has a collision prefix:
+            - UBX
+            - UCP
+            - USP
+            - UCX
+
+    This validator also checks if subset name is correct
+            - {static mesh prefix}_{Asset-Name}{Variant}.
```
BigRoy, line 27: I feel like this is mixing two validations together that should be separate. You now can only disable both of these at the same time even though they are quite different things - right.

Also, out of curiosity - when would the subset name be incorrect?

openpype/hosts/houdini/plugins/create/create_unreal_staticmesh.py
```diff
@@ -0,0 +1,164 @@
...
+    def get_network_categories(self):
+        return [
+            hou.ropNodeTypeCategory(),
+            hou.sopNodeTypeCategory()
+        ]
+
+    def get_pre_create_attr_defs(self):
+        """Add settings for users. """
+
+        attrs = super().get_pre_create_attr_defs()
```
BigRoy, line 50: I believe `super()` call without arguments only works in Python 3+ and this doesn't work in Python 2 - which might still be relevant for older Houdini versions?

openpype/hosts/houdini/plugins/publish/validate_unreal_staticmesh_naming.py
```diff
@@ -0,0 +1,108 @@
...
+
+        rop_node = hou.node(instance.data["instance_node"])
+        output_node = instance.data.get("output_node")
+        if output_node is None:
+            cls.log.debug(
+                "No Output Node, skipping check.."
+            )
+            return
+
+        if not rop_node.evalParm('buildfrompath'):
```
BigRoy, line 73:
```suggestion
        if not rop_node.evalParm("buildfrompath"):
```

openpype/hosts/houdini/plugins/publish/validate_unreal_staticmesh_naming.py
```diff
@@ -0,0 +1,108 @@
...
+            instance.data["asset"],
+            instance.data.get("variant", "")
+        )
+
+        if instance.data.get("subset") != subset_name:
+            invalid.append(rop_node)
+            cls.log.error(
+                "Invalid subset name on rop node '%s' should be '%s'.",
+                rop_node.path(), subset_name
+            )
```
BigRoy, line 106: So what is currently ensuring that on create the subset name is like that? :) I feel like the Creator class is missing an override to its `get_subset_name` method if this is really the rule for these instances.

openpype/hosts/houdini/plugins/create/create_unreal_staticmesh.py
mustafa-zarkash: tbh, I was imitating unreal static mesh creator in maya
subset name became dynamic only after:
1. adding the `"asset"` to the dynamic data 
2. adding subset name template 
image

openpype/hosts/houdini/api/lib.py
mustafa-zarkash: `startnode` parameter on FBX ROP accepts obj node or an explicit sop node
image

fbx output depends on the start node and  its output 
1. `SopNode`, no dependencies 
2. `ObjNode` and its output 
3. obj sub network, all obj nodes inside it and their outputs 

So, the above function handle these cases as well as handling nested obj subnetworks.
it only returns the output of obj nodes unlike `allNodes()` returns all nodes.

openpype/hosts/houdini/plugins/create/create_unreal_staticmesh.py
mustafa-zarkash: man, I love your simple solutions!

openpype/hosts/houdini/plugins/publish/validate_unreal_staticmesh_naming.py
mustafa-zarkash: tbh, I was imitating the corresponding maya validator which do the two things! 
but yeah, why not splitting them into two! 

> Also, out of curiosity - when would the subset name be incorrect?

as far as I tested, it only happens when subset is changed manually. 
image

openpype/hosts/houdini/plugins/load/load_fbx.py
mustafa-zarkash: this part was confusing because I found many things when searching the code:
This is also possible I think 
```
context = get_representation_context(representation)
```

https://github.com/ynput/OpenPype/blob/c08190ccf536864e3b94b48636637fdebc114db0/openpype/pipeline/load/utils.py#L554-L556

and in this function context is re-retrieved from representation
`context=representation["context"]`

https://github.com/ynput/OpenPype/blob/c08190ccf536864e3b94b48636637fdebc114db0/openpype/pipeline/load/utils.py#L613

---
So I chose the safest way :"D

openpype/hosts/houdini/plugins/create/create_unreal_staticmesh.py
BigRoy: Still - usage of `{asset}` and `{variant}` wouldn't require the need to override `get_dynamic_data`. Right @iLLiCiTiT ?

openpype/hosts/houdini/api/lib.py
BigRoy: > obj sub network, all obj nodes inside it and their outputs

Interesting - so in that case it'd export something else than what the user would be seeing when 'displaying' that obj sub network?

openpype/hosts/houdini/plugins/publish/extract_fbx.py
mustafa-zarkash: if you are referring to the return statement, then I will no longer use it in the next commit

openpype/hosts/houdini/plugins/publish/validate_output_node.py
mustafa-zarkash: > Should we be more specific to avoid name clashes in the future?
100%  agreed! 

tbh, I don't know what to call it, any recommendations ?

openpype/hosts/houdini/plugins/publish/validate_output_node.py
mustafa-zarkash: it can, each object node child with its sop output when `startnode` is a subnetwork.

openpype/hosts/houdini/plugins/publish/validate_mesh_is_static.py
mustafa-zarkash: I think it's redundant, I will test it again.

openpype/hosts/houdini/plugins/publish/validate_unreal_staticmesh_naming.py
mustafa-zarkash: I think it's automated via dynamic data

openpype/hosts/houdini/plugins/publish/validate_output_node.py
BigRoy: We could just do `ValidateFBXOutputNode` to make it specific to the family?

openpype/hosts/houdini/plugins/publish/validate_unreal_staticmesh_naming.py
BigRoy: No, you shared a screenshot just now that the family is overridden by name due to the templating system in settings - so the default `get_subset_name` is already relying on that as it should. 👍 Anyway, I think the validator should actually just ensure it matches the template settings - not this hardcoded default?

openpype/hosts/houdini/api/lib.py
mustafa-zarkash: I'm using Maya for reference.. 

---

> so in that case it'd export something else than what the user would be seeing when 'displaying' that obj sub network?

it will work as usual, i.e. it will export what the user would be seeing when 'displaying' that obj sub network

Let show you my example: 
I have this subnetwork hierarchy: 
```
*GROUP (Subnetwork ObjNode)
  ├── **Box (ObjNode)
  │        ├── Some nodes we don't care about
  │        └──  *Output node or the node the with the display flag (SopNode)
  ├── **Sphere (ObjNode)
  │        └──  *Output node (SopNode)
  ├── **Gride (ObjNode)
  │        └──  *Output node (SopNode)
  └── **GROUP_2 (Subnetwork ObjNode)
            ├── **Platonic Plate (ObjNode)  
            │        └──  *Output node (SopNode)
            └── **Torus (ObjNode)
                      └──  *Output node (SopNode)
```
So, this function `get_output_children` should retrieve all the important nodes (that I marked with `*` and `**` above) 
I believe any node with `**` should be tested for naming collisions 
I believe any node with `*` or `**` should be tested for Time Dependency 

image

> fbx rop don't include parent node `GROUP` name in the fbx file, it's whether ignored or named `world_root`

Cube was inside a subnetwork with `Create Root for Subnet` on 
Cube was inside a subnetwork with `Create Root for Subnet` off 
image

---

I tried to mimic the same scenario in Maya, and it just ignores Groups 
while creating:
image

After publishing:
image

Let me know what do you think.

mustafa-zarkash: I hided box and sphere `ObjNode`s in houdini and this is what it looks like in maya 

image

BigRoy: So basically display flag has no influence on the export 👍 Great. Seems like the collector is handling the use cases that the FBX exporter exports.

openpype/hosts/houdini/plugins/create/create_unreal_staticmesh.py
mustafa-zarkash: I gave it a shot and it doesn't work without passing `{asset}` as dynamic data.

mustafa-zarkash:
## Checkpoint

- UE Static mesh creator allows selecting sop node or obj node 
- Static mesh collector adds fbx family which triggers fbx extractor 
- It's possible to use other extractors but I'm not sure if that would break the validators
- It includes four validations 
    - `ValidateFBXOutputNode` which should check if output node exists and it has geometry. 
    - `ValidateMeshIsStatic` which checks if output node or it's output children are time dependent
    - `ValidateUnrealStaticMeshName` which checks if  output node or it's output children includes collision prefix
    - `ValidateSubsetName` which checks if subset name is correct + a fix action
- it includes fbx extractor and loader

It still lacks: 
- Better error messages, I intent to use xml (to provide better guidance about dealing with errors)
- Checking geometry existence, which is not yet included `ValidateOutputNode`

openpype/hosts/houdini/plugins/create/create_unreal_staticmesh.py
BigRoy: Oh man - I clearly wasn't awake. `{asset}` is a default key for folder and file templates but **not** for the subset name formatting. That's why this needs to be passed along in `get_dynamic_data` explicitly.

MustafaJafar:
## Checkpoint
> Some code is refactored. 

- UE Static mesh creator allows selecting sop node or obj node 
- Static mesh collector adds fbx family which triggers fbx extractor 
- It's possible to use other extractors but I'm not sure if that would break the validators
- It includes four validations 
    - `ValidateFBXOutputNode` which checks if output node exists, has geometry and doesn't include invalid primitives. 
    - `ValidateMeshIsStatic` which checks if output node or it's output children are time dependent
    - `ValidateUnrealStaticMeshName` which checks if  output node or it's output children includes collision prefix
    - `ValidateSubsetName` which checks if subset name is correct + a fix action
- it includes fbx extractor and loader

openpype/hosts/houdini/plugins/load/load_fbx.py
MustafaJafar: @iLLiCiTiT Could you tell me what do you think ?
are there any unnecessary logic in `self.get_file_path` in `load_fbx.py` ?

openpype/hosts/houdini/plugins/load/load_fbx.py
```diff
@@ -0,0 +1,150 @@
...
+
+    order = -10
+
+    families = ["staticMesh", "fbx"]
+    representations = ["fbx"]
+
+    def load(self, context, name=None, namespace=None, data=None):
+
+        # get file path
+        file_path = self.get_file_path(context=context)
```
iLLiCiTiT, line 25:
```suggestion
        file_path = self.filepath_from_context(context)
```

antirotor (approved):
Some comments about naming, but otherwise it looks great and from what I've tested it works.

MustafaJafar:
`Create Unreal static mesh` is renamed to `Create static mesh` 
Validator `ValidateUnrealStaticMeshName` is turned off by default 

In order to enable Static mesh validation, it is required to do two steps:
1. add houdini host to staticMesh `subset name profiles`
image

2. Turn on this houdini validator  
image

PR #5483, opened 2023-08-23
Nuke: fixing thumbnail and monitor out root attributes

jakubjezek001:
## Changelog Description
Nuke Root Colorspace settings for Thumbnail and Monitor Out schema was gradually changed between version 12, 13, 14 and we needed to address those changes individually for particular version.

## Testing notes:
1. Open Settings at `project_settings/nuke/imageio/workfile/monitorLut`
2. Set the value to either `ACES/sRGB` or `sRGB (ACES)` but make sure the OCIO config is set to ACES config.
3. Open Nuke with 12, 13, 14 minor versions and use Set Colorspace menu item .
4. Look to Script properties panel to Colorspace Tab and see that the Thumbnail, Monitor Out attributes were set correctly

- [ ] Nuke 12.x 
- [ ] Nuke 13.x
- [ ] Nuke 14.x

### warning for testing Nuke 12
Once you will be testing Nuke 12 and you will open workfile which was originally created at Nuke 13/14 this will popup and it is not an isssue of this PR. This will disapeare as soon as you save the script in 12 and then reopen again.

image


openpype/hosts/nuke/api/lib.py
```diff
@@ -3320,15 +3346,35 @@ def get_viewer_config_from_string(input_string):
...
+    Returns:
+        str: viewer config string
+    """
+    if display:
+        if path_like:
+            return "{}/{}".format(display, viewer)
+        else:
+            return "{} ({})".format(viewer, display)
+    else:
+        return viewer
```
iLLiCiTiT, line 3380: No else statements :)
```suggestion
    if not display:
        return viewer

    if path_like:
        return "{}/{}".format(display, viewer)

    return "{} ({})".format(viewer, display)
```

iLLiCiTiT (approved):
No issues with code. Must be tested.

openpype/hosts/nuke/api/lib.py
```diff
@@ -2091,6 +2092,31 @@ def set_root_colorspace(self, imageio_host):
...
+
+        if nuke.NUKE_VERSION_MAJOR == 13:
+            workfile_settings["monitorOutLUT"] = create_viewer_profile_string(
+                m_viewer, m_display, path_like=False)
+            # monitorLut=thumbnails - viewerProcess makes more sense
+            workfile_settings["monitorLut"] = create_viewer_profile_string(
+                v_viewer, v_display, path_like=True)
+        if nuke.NUKE_VERSION_MAJOR <= 12:
+            workfile_settings["monitorLut"] = create_viewer_profile_string(
+                m_viewer, m_display, path_like=True)
```
BigRoy, line 2118: Would it make more sense to allow the version specific details up to do the function that is separate?

```
get_viewer_profile_settings(m_viewer, m_display)
```

Then the function returns:
```python
return {
    "monitorOutLUT": a,
    "monitorLut": b
}
```
etc. based on the nuke version itself.
It feels slightly odd being embedded directly in the `set_root_colorspace` function which is already quite a lengthy function.

jakubjezek001: good point, look here 33e18661241a59a612a06558fbc19f5de264220b

openpype/hosts/nuke/api/lib.py
```diff
@@ -2118,6 +2128,42 @@ def set_root_colorspace(self, imageio_host):
...
+            monitor_lut (str): monitor lut string
+
+        Returns:
+            dict: monitor settings
+        """
+        output_data = {}
+        m_display, m_viewer = get_viewer_config_from_string(monitor_lut)
+        v_display, v_viewer = get_viewer_config_from_string(
+            viewer_lut
+        )
```
BigRoy, line 2145:
```suggestion
        m_display, m_viewer = get_viewer_config_from_string(monitor_lut)
        v_display, v_viewer = get_viewer_config_from_string(viewer_lut)
```
Not sure why the second line was on two lines even though it's a shorter line. :)

moonyuet:
It sets the config files in Nukev14 but it doesn't set the knob value as expected after clicking (Set Colorspace from OP Menu)
image

jakubjezek001:
> It sets the config files in Nukev14 but it doesn't set the knob value as expected after clicking (Set Colorspace from OP Menu) image

image

It seems like you don't have configured correctly the nuke workfile settings. Try these, they should fit the aces_1.2 config.

PR #5484, opened 2023-08-23
Nuke: fixing missing `instance_id` error

jakubjezek001:
## Changelog Description
Workfiles with Instances created in old publisher workflow were rising error during converting method since they were missing `instance_id` key introduced in new publisher workflow. 
 
## Testing notes:
1. Open any workfile with old publisher instances
2. open Publisher and hit `Cpnvert Legacy Instances` method. 
3. All should convert as expected.


64qam:
Unfortunately, I don't have such an old work file on the machine:-(

iLLiCiTiT (approved):

Issue #4997, opened 2023-05-22
Bug: Nuke validator rendered files repair action


jakubjezek001:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

Validate rendered frames' repair action is not set chosen rendering target.

### Expected Behavior:

After particular target repair action is chosen, instance attribute enumerator will be set to the position.

### Version

3.15.8-nightly.2

### What platform you are running OpenPype on?

Windows, Linux / Centos, Linux / Ubuntu, Linux / RedHat, MacOS

### Steps To Reproduce:

1. Crated render family write instance with empty render directory
2. Enumerator for rendering target was set to `Existing frames`
3. Validator had popped up
4. Repair Local or On farm was selected. 
5. nothing had happened

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

[cuID:[OP-6012](https://app.clickup.com/t/863gtt2kz)]

PR #5486, opened 2023-08-23
Nuke: existing frames validator is repairing render target

jakubjezek001:
## Changelog Description
Nuke is now correctly repairing render target after the existing frames validator finds missing frames and repair action is used.

## Testing notes:
1. open test scene in nuke and make sure no rendered frames are available on created Render Write node. 
2. Open Publisher and set render target to existing frames publishing.
3. Validator should fail and use repair buttons on Top either to Local or Farm
4. Refresh publishing context and notice the Render Target was set according to your selection. 

Closes https://github.com/ynput/OpenPype/issues/4997

### DIFF
--- a/openpype/hosts/nuke/plugins/publish/validate_rendered_frames.py
+++ b/openpype/hosts/nuke/plugins/publish/validate_rendered_frames.py
@@ -14,19 +14,28 @@ class RepairActionBase(pyblish.api.Action):
         # Get the errored instances
         return get_errored_instances_from_context(context, plugin=plugin)
 
-    def repair_knob(self, instances, state):
+    def repair_knob(self, context, instances, state):
+        create_context = context.data["create_context"]
         for instance in instances:
-            node = instance.data["transientData"]["node"]
-            files_remove = [os.path.join(instance.data["outputDir"], f)
-                            for r in instance.data.get("representations", [])
-                            for f in r.get("files", [])
-                            ]
+            files_remove = [
+                os.path.join(instance.data["outputDir"], f_)
+                for r_ in instance.data.get("representations", [])
+                for f_ in r_.get("files", [])
+            ]
             self.log.info("Files to be removed: {}".format(files_remove))
-            for f in files_remove:
-                os.remove(f)
-                self.log.debug("removing file: {}".format(f))
-            node["render"].setValue(state)
+            for f_ in files_remove:
+                os.remove(f_)
+                self.log.debug("removing file: {}".format(f_))
+
+            # Reset the render knob
+            instance_id = instance.data["instance_id"]
+            created_instance = create_context.get_instance_by_id(
+                instance_id
+            )
+            created_instance.creator_attributes["render_target"] = state
+
             self.log.info("Rendering toggled to `{}`".format(state))
+        create_context.save_changes()
 
 
 class RepairCollectionActionToLocal(RepairActionBase):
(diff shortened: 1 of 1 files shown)


openpype/hosts/nuke/plugins/publish/validate_rendered_frames.py
```diff
@@ -14,35 +14,44 @@ def get_instance(context, plugin):
...
         for instance in instances:
-            node = instance.data["transientData"]["node"]
-            files_remove = [os.path.join(instance.data["outputDir"], f)
-                            for r in instance.data.get("representations", [])
-                            for f in r.get("files", [])
-                            ]
+            files_remove = [
+                os.path.join(instance.data["outputDir"], f_)
+                for r_ in instance.data.get("representations", [])
+                for f_ in r_.get("files", [])
```
iLLiCiTiT, line 23: What if `"files"` is single file?

jakubjezek001: In the future we should support single video file output, but at the moment we don't, so this should not happen.

openpype/hosts/nuke/plugins/publish/validate_rendered_frames.py
```diff
@@ -14,35 +14,44 @@ def get_instance(context, plugin):
...
+
             self.log.info("Rendering toggled to `{}`".format(state))
+        create_context.save_changes()
 
 
 class RepairCollectionActionToLocal(RepairActionBase):
     label = "Repair - rerender with \"Local\""
 
     def process(self, context, plugin):
         instances = self.get_instance(context, plugin)
```
BigRoy, line 50: It's so weird that this method is called `get_instance` even though it returns a list.

openpype/hosts/nuke/plugins/publish/validate_rendered_frames.py
```diff
@@ -14,35 +14,44 @@ def get_instance(context, plugin):
...
+                for f_ in r_.get("files", [])
+            ]
             self.log.info("Files to be removed: {}".format(files_remove))
-            for f in files_remove:
-                os.remove(f)
-                self.log.debug("removing file: {}".format(f))
-            node["render"].setValue(state)
+            for f_ in files_remove:
+                os.remove(f_)
+                self.log.debug("removing file: {}".format(f_))
```
BigRoy: I wouldn't want it to delete any files on disk just when clicking 'repair' to set the render mode. Isn't Nuke capable of just overwriting existing files?  Or does publishing in Nuke work in such a way that it only ever publishes existing frames from disk?

Minkiu (approved):

openpype/hosts/nuke/plugins/publish/validate_rendered_frames.py
jakubjezek001: I'm not sure what exactly seems strange. I can only explain why I am using the `create_context.save_changes()` method since it is highlighted. The reason it needs to be saved is because the repair tool is modifying `creator_attributes` and those changes need to be saved back into the instance data knob at the instance node. This ensures that they are correctly invoked after a refresh action in the publisher.

openpype/hosts/nuke/plugins/publish/validate_rendered_frames.py
jakubjezek001: I agree. The leftover code seems unrelated to the validator's purpose. Thanks for the fresh perspective. I hadn't considered it before.

PR #5490, opened 2023-08-23
Use host's node uniqueness for instance id in new publisher

### LINKED ISSUE #4820: enhancement: Houdini: Unable to duplicate ROP node to duplicate a publish instance

### Please state which of feature you have in mind and describe what are its shortcomings?

It's currently hard to duplicate an existing publish instance (ROP node) to allow quickly duplicating the behavior and create a new instance natively to houdini. 

This is due to the node storing the `instance_node` value and the `instance_id` which means that the duplicated instance would not show in the Publisher UI (due to same `instance_id`), once you change the `instance_id` the `instance_node` fails due to the duplicated node not having the same path.

**Warning:** The `instance_node` value being stored on the node also means it breaks publishing whenever an artist would change the node name but not the attribute.

### How would you imagine the enhancement of the feature?

1. Upon a conflict of `instance_id` it should be able to differentiate the two and maybe generate a new id for one of the two?
2. `instance_node` should **never** be stored on the node since that value is the node's path in the workfile. That should **always** be based on the node's actual path (thus parent + name, `node.path()`).

### Is there any more labels you wish to add?

- [X] I have searched labels and added any

### Describe alternatives you've considered:

_No response_

### Additional context:

_No response_

[cuID:[OP-5573](https://app.clickup.com/t/861mmp5j1)]

BigRoy:
## Changelog Description

Instead of writing `instance_id` as parm or attributes on the publish instances we can, for some hosts, just rely on a unique name or path within the scene to refer to that particular instance. By doing so we fix #4820 because upon duplicating such a publish instance using the host's (DCC) functionality the uniqueness for the duplicate is then already ensured instead of attributes remaining exact same value as where to were duplicated from, making `instance_id` a non-unique value. 

## Additional info

I've now tested the Houdini, Maya and Fusion change and for me these seem to work.

I'm not entirely sure whether there are any other potential issues due to `instance_id` now not being a uuid generated by the new publisher for these instances.

## Testing notes:

1. Create a publish instance in Maya, Fusion, Houdini
2. Duplicate the publish instance node inside the host (e.g. ROP node in Fusion)
3. Refresh the publisher - both the old and duplicated entry should show up and publish fine.
4. Upon changing attributes e.g. on the instance in the Publisher UI and saving that the changed values should still persist on the node in the workfile (even when altering on duplicated instances)


BigRoy:
Had a brief chat with @iLLiCiTiT about the Publisher's API/backend and whether this PR performs changes that would start introducing subtle bugs with the new publisher - the conclusion was that **_it's fine._**

There was an inconclusive remark that having a unique id to persist unrelated to the DCC could help with 'information loss' within the DCC if a user e.g. renamed the node (or whatever represented the unique of the node in the DCC). However, the conclusion there was that it wasn't necessarily a huge problem. Plus even if it'd occur, you'd only lose the changes of your publish session instead of the original full node, due to the fact that the renamed node would still exist but just not get the updated instance data applied. Again, this is an example case that only occurs if you were to rename the node whilst the publisher is open.

As such, the approach of this PR seems like an ok solution to solve the issue at hand.
It'd be good to have some testers do a test in the different DCCs to see if any (other) issues or dangers might be found or whether this is fully functional as is in Maya, Houdini and Fusion.

mustafa-zarkash (approved):
It works in Houdini..
Makes me wonder
- what are the basic minimum parameters that make a node discoverable?
- Are there any redundant parameters that can be safely removed ?
- What parameters can be dynamic using expressions in Houdini ?

BigRoy:
@mustafa-zarkash thanks for providing screenshots. It's good to note that if you were to Create a new instance using this PR you should actually get a ROP node without `instance_node` and `instance_id` parms. (Old instances will preserve them however, they'd just be unused because the Creator doesn't clean up redundant attributes on existing instances).

moonyuet:
Looks good. The OP parameter doesn't include any instance_node or instance_id 
image

image
Also double check in python shell (some clinche check)
Both instance_node and instance_id dont' "exist" as OP parameters anymore.
image

moonyuet (approved):
Approved for Houdini, Heven't tested Maya and Fusion.

mustafa-zarkash:
> Technically other attributes like 'task', etc. could be made "live" to your current context

Revisiting this comment https://github.com/ynput/OpenPype/issues/4820#issuecomment-1679444263 
I think having a dynamic `subset` name will let us duplicate instances freely without doing any changes.

image

tokejepsen:
Tested in Maya 2023 successfully, but I agree with @mustafa-zarkash that the workflow for the subset is slightly confusing cause just duplicating the instance (objectset) will produce instances that go to the same subset.

Not sure what logic we can introduce in Maya to get the subset from the objectset's name.

BigRoy:
> Tested in Maya 2023 successfully, but I agree with @mustafa-zarkash that the workflow for the subset is slightly confusing cause just duplicating the instance (objectset) will produce instances that go to the same subset.

That's true - but at the very least this PR brings the capabilities back to the level to how it was possible in the old publisher with duplicating. The subset+asset can always be adjusted in the publisher afterwards, but without this PR the duplicated nodes would start confusing the new publisher since only one of the two nodes would show up - you wouldn't really know which one, etc. So I think this is a massive improvement and we can solve other enhancements separately.

tokejepsen (approved):

tokejepsen:
@LiborBatek you able to test fusion? Else assign someone who can.

LiborBatek:
> @LiborBatek you able to test fusion? Else assign someone who can.

@jakubjezek001 could u test in fusion? I guess it would be the quickest?

BigRoy:
> Maybe we can solve this later by some visualization inside Publisher UI on instances - that it could colorize items with the same subset just as a warning or reminder.

Note that the clash would only be important if they match in `asset + subset` since publishing a same named subset to different assets is totally fine.

Also note that [this validator](https://github.com/ynput/OpenPype/blob/develop/openpype/plugins/publish/validate_unique_subsets.py) currently exists to validate this exact use case where there might be clashes on asset+subset.

BigRoy:
Or @EmberLightVFX for a Fusion check perhaps?

BigRoy:
Would it be ok if I'd personally do a test run in Fusion - maybe screen capturing the process?

tokejepsen:
> Would it be ok if I'd personally do a test run in Fusion - maybe screen capturing the process?

Yeah, I think that would be great!

BigRoy:
Quick screen recording in Fusion creating a saver, duplicating it, changing the values of the second saver in the publisher and then validating.

https://github.com/ynput/OpenPype/assets/2439881/51b3c8e5-2467-464f-bcc0-86ac70ba67d3

Does this suffice?

EmberLightVFX (approved):

antirotor (approved):
Looks great!

PR #5494, opened 2023-08-24
Farm: adding target collector

jakubjezek001:
## Changelog Description
Enhancing farm publishing workflow.

## Additional info
- removing `toBeRenderedOn` key from individual host submitters
- introducing new collector plugin which will define target by activated farm modules.

## Testing notes:
1. All hosts should work the same with farm publishing

Primary hosts to be tested:
- [x] Maya
- [x] Nuke
- [x] Houdini 
- [x] Max
- [x] AfterEffects

## Dependencies
Changes from this PR were initially implemented in [this pull request](https://github.com/ynput/OpenPype/pull/5451). However, due to its significant size, we made the decision to divide it into smaller portions. This PR is the first one in the sequence, and subsequent PRs will indicate the interdependencies and the order in which the merges should be executed.


openpype/plugins/publish/collect_farm_target.py
```diff
@@ -0,0 +1,45 @@
...
+    order = pyblish.api.CollectorOrder + 0.499
+    label = "Collect Farm Target"
+    targets = ["local"]
+
+    def process(self, instance):
+        if not instance.data.get("farm"):
+            return
+
+        context = instance.context
+        try:
```
Minkiu, line 18: So for someone to render on muster, they have to have both `deadline` and `royalrenderer` installed as modules?

Wouldn't an approach like this be better?
```py
farm_name = ""

op_modules = context.data.get("openPypeModules")

for farm_renderer in ["deadline","royalrender", "muster"]:
	op_module  = op_modules.get(farm_renderer, False)
	
	if op_module and op_module.enabled:
		farm_name = farm_renderer
	elif not op_module:
		self.log.error("Cannot get OpenPype {0} module.".format(farm_renderer))


if farm_name:
	self.log.debug("Collected render target: {0}".format(farm_name))
	instance.data["toBeRenderedOn"] = farm_name
```

jakubjezek001: Yeah, definitely improvement. But also we are planning to remove `muster` from codebase.

Minkiu: I used "muster" as an example, but the same applies to "royalrender", my point was that unless you have "deadline" module loaded, you won't be able to render in any of the other two

kalisp: I was thinking about exposing renderer as a dropdown in Publisher and filling it only with enabled modules. Not sure if it not a overkill if there is no use case for studio having multiple renderers active. (Maybe for different projects?)

moonyuet:
Tested with Max.
It successfully submit to deadline 

image

image

LiborBatek:
Did tested for ```Max``` host and all good.

@jakubjezek001 how about testing also ```blender``` host which was recently added in OP (with support of submit to DL). should I run tests for this too??

LiborBatek (approved):
Did test ```maya``` and ```max``` hosts apps and all good and working.

openpype/plugins/publish/collect_farm_target.py
BigRoy: > I was thinking about exposing renderer as a dropdown in Publisher and filling it only with enabled modules. Not sure if it not a overkill if there is no use case for studio having multiple renderers active. (Maybe for different projects?)

It could be `hidden=True` if there's only one active?

moonyuet:
Works for Houdini....
image

kalisp (approved):
Tested in Nuke and AE

openpype/plugins/publish/collect_farm_target.py
jakubjezek001: > I used "muster" as an example, but the same applies to "royalrender", my point was that unless you have "deadline" module loaded, you won't be able to render in any of the other two

Definitely! I haven't seen the raise after each try and it would be blocking. I had implemented our way and all works ;) Thanks

PR #5496, opened 2023-08-24
General, Houdini: abstraction of file collections

jakubjezek001:
## Changelog Description
Unified way of collecting expected files for farm publishing.

## Additional info
- added into host plugins
- enhanced some code style in houdini collectors

## Testing notes:
1. all should work as it was before

## Dependencies
Changes from this PR were initially implemented in [this pull request](https://github.com/ynput/OpenPype/pull/5451). However, due to its significant size, we made the decision to divide it into smaller portions. 

## Edit
This needed to be retargeted to develop so some changes in the original code and commits needed to be done.

### DIFF
--- a/openpype/hosts/houdini/plugins/publish/collect_mantra_rop.py
+++ b/openpype/hosts/houdini/plugins/publish/collect_mantra_rop.py
@@ -1,9 +1,9 @@
-import re
 import os
 
 import hou
 import pyblish.api
 
+from openpype.pipeline import file_collection
 from openpype.hosts.houdini.api.lib import (
     evalParmNoFrame,
     get_color_management_preferences
@@ -31,49 +31,63 @@ class CollectMantraROPRenderProducts(pyblish.api.InstancePlugin):
     def process(self, instance):
 
         rop = hou.node(instance.data.get("instance_node"))
+        frame_start = instance.data["frameStart"]
+        frame_end = instance.data["frameEnd"]
 
         # Collect chunkSize
         chunk_size_parm = rop.parm("chunkSize")
         if chunk_size_parm:
             chunk_size = int(chunk_size_parm.eval())
             instance.data["chunkSize"] = chunk_size
-            self.log.debug("Chunk Size: %s" % chunk_size)
+            self.log.debug("Chunk Size: {}".format(chunk_size))
 
-            default_prefix = evalParmNoFrame(rop, "vm_picture")
-            render_products = []
+        default_prefix = evalParmNoFrame(rop, "vm_picture")
+        render_products = []
 
-            # Default beauty AOV
-            beauty_product = self.get_render_product_name(
-                prefix=default_prefix, suffix=None
+        # Default beauty AOV
+        beauty_product = self.get_render_product_name(
+            prefix=default_prefix, suffix=None
+        )
+        render_products.append(beauty_product)
+
+        files_by_aov = {
+            "beauty": file_collection.generate_expected_filepaths(
+                frame_start, frame_end, beauty_product)
+        }
+
+        # get the number of AOVs
+        aov_numbers = rop.evalParm("vm_numaux")
+        if aov_numbers == 0:
+            return
+
+        # get the filenames of the AOVs
+        for index in range(1, aov_numbers + 1):
+            var_ = rop.evalParm("vm_variable_plane{}".format(index))
+
+            # skip empty variables
+            if not var_:
+                continue
+
+            aov_name = "vm_filename_plane{}".format(index)
+            aov_boolean = "vm_usefile_plane{}".format(index)
+            aov_enabled = rop.evalParm(aov_boolean)
+            has_aov_path = rop.evalParm(aov_name)
+
+            # skip disabled AOVs
+            if not (has_aov_path and aov_enabled == 1):
+                continue
+
+            aov_prefix = evalParmNoFrame(rop, aov_name)
+            aov_product = self.get_render_product_name(
+                prefix=aov_prefix, suffix=None
             )
-            render_products.append(beauty_product)
+            render_products.append(aov_product)
 
-            files_by_aov = {
-                "beauty": self.generate_expected_files(instance,
-                                                       beauty_product)
-            }
-
-            aov_numbers = rop.evalParm("vm_numaux")
-            if aov_numbers > 0:
-                # get the filenames of the AOVs
-                for i in range(1, aov_numbers + 1):
-                    var = rop.evalParm("vm_variable_plane%d" % i)
-                    if var:
-                        aov_name = "vm_filename_plane%d" % i
-                        aov_boolean = "vm_usefile_plane%d" % i
-                        aov_enabled = rop.evalParm(aov_boolean)
-                        has_aov_path = rop.evalParm(aov_name)
-                        if has_aov_path and aov_enabled == 1:
-                            aov_prefix = evalParmNoFrame(rop, aov_name)
-                            aov_product = self.get_render_product_name(
-                                prefix=aov_prefix, suffix=None
-                            )
-                            render_products.append(aov_product)
-
-                            files_by_aov[var] = self.generate_expected_files(instance, aov_product)     # noqa
+            files_by_aov[var_] = file_collection.generate_expected_filepaths(
+                frame_start, frame_end, aov_product)
 
         for product in render_products:
-            self.log.debug("Found render product: %s" % product)
+            self.log.debug("Found render product: {}".format(product))
 
         filenames = list(render_products)
         instance.data["files"] = filenames
@@ -101,27 +115,3 @@ class CollectMantraROPRenderProducts(pyblish.api.InstancePlugin):
             product_name = prefix_base + "." + suffix + ext
 
         return product_name
-
-    def generate_expected_files(self, instance, path):
-        """Create expected files in instance data"""
-
-        dir = os.path.dirname(path)
-        file = os.path.basename(path)
-
-        if "#" in file:
-            def replace(match):
-                return "%0{}d".format(len(match.group()))
-
-            file = re.sub("#+", replace, file)
-
-        if "%" not in file:
-            return path
-
-        expected_files = []
-        start = instance.data["frameStart"]
-        end = instance.data["frameEnd"]
-        for i in range(int(start), (int(end) + 1)):
-            expected_files.append(
-                os.path.join(dir, (file % i)).replace("\\", "/"))
-
-        return expected_files
--- a/openpype/pipeline/file_collection.py
+++ b/openpype/pipeline/file_collection.py
@@ -1,0 +1,279 @@
+"""Utility functions for file sequences and collections
+"""
+import os
+import re
+import clique
+from openpype.lib import Logger
+from openpype.pipeline import frame_utils
+
+
+def collect_filepaths_from_sequential_path(
+    file_path,
+    frame_start=None,
+    frame_end=None,
+    only_existing=False,
+):
+    """Generate expected files from path
+
+    Args:
+        file_path (str): Absolute path with any sequential pattern (##, %02d)
+        frame_start (Optional[int]): Start frame of the sequence
+        frame_end (Optional[int]): End frame of the sequence
+        only_existing (Optional[bool]): Ensure that files exists.
+
+    Returns:
+        Any[list[str], str]: List of absolute paths to files
+            (frames of a sequence) or path if single file
+    """
+
+    dirpath = os.path.normpath(
+        os.path.dirname(file_path))
+
+    filename = os.path.basename(file_path)
+
+    formattable_string = convert_filename_to_formattable_string(
+        filename)
+
+    if (
+        not formattable_string
+        or frame_start is None
+        or frame_end is None
+    ):
+        return file_path
+
+    expected_files = []
+    for frame in range(int(frame_start), (int(frame_end) + 1)):
+        frame_file_path = os.path.join(
+            dirpath, formattable_string.format(frame)
+        )
+
+        # make sure file exists if ensure_exists is enabled
+        if only_existing and not os.path.exists(frame_file_path):
+            continue
+
+        # add to expected files
+        expected_files.append(
+            # add normalized path
+            frame_file_path
+        )
+
+    return expected_files
+
+
+def generate_expected_filepaths(
+    frame_start,
+    frame_end,
+    file_path,
+):
+    """Generate expected files from path
+
+    Args:
+        frame_start (int): Start frame of the sequence
+        frame_end (int): End frame of the sequence
+        file_path (str): Absolute path with any sequential pattern (##, %02d)
+
+    Returns:
+        Any[list[str], str]: List of expected absolute paths to files
+            (frames of a sequence)
+    """
+    files = collect_filepaths_from_sequential_path(
+        file_path,
+        frame_start,
+        frame_end,
+        only_existing=False,
+    )
+    return [files] if isinstance(files, str) else files
+
+
+def collect_basenames_from_sequential_path(
+    file_path,
+    frame_start=None,
+    frame_end=None,
+    only_existing=False,
+):
+    """Collect files from sequential path
+
+    Args:
+        file_path (str): Absolute path with any sequential pattern (##, %02d)
+        frame_start (Optional[int]): Start frame of the sequence
+        frame_end (Optional[int]): End frame of the sequence
+        only_existing (Optional[bool]): Ensure that files exists.
+
+    Returns:
+        list[str]: List of collected file names
+            (frames of a sequence) or path if single file
+    """
+
+    collected_abs_filepaths = collect_filepaths_from_sequential_path(
+        file_path,
+        frame_start,
+        frame_end,
+        only_existing,
+    )
+
+    if isinstance(collected_abs_filepaths, list):
+        return [os.path.basename(f_) for f_ in collected_abs_filepaths]
+
+    # return single file in list
+    return [os.path.basename(collected_abs_filepaths)]
+
+
+def convert_filename_to_formattable_string(filename):
+    """Convert filename to formattable string
+
+    Args:
+        filename (str): Filename to convert
+
+    Returns:
+        Any[str, None]: Formattable string or None if not possible
+                        to convert
+    """
+    new_filename = None
+
+    if "#" in filename:
+        # use regex to convert #### to {:0>4}
+        def replace(match):
+            return "{{:0>{}}}".format(len(match.group()))
+        new_filename = re.sub("#+", replace, filename)
+
+    elif "%" in filename:
+        # use regex to convert %04d to {:0>4}
+        padding = re.search("%(\\d)+d", filename)
+        padding = padding.group(1) if padding else 1
+        new_filename = re.sub(
+            "%.*d",
+            "{{:0>{}}}".format(padding),
+            filename
+        )
+
+    return new_filename
+
+
+def get_publishing_representation(
+    instance,
+    file_path,
+    frame_start=None,
+    frame_end=None,
+    log=None,
+    only_existing=False,
+    reviewable=False,
+):
+    """Get representation with expected files.
+
+    Args:
+        instance (pyblish.Instance): instance
+        file_path (str): file path
+        frame_start (Optional[int]): first frame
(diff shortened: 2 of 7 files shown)


openpype/pipeline/file_collection.py
```diff
@@ -0,0 +1,274 @@
...
+from openpype.pipeline import frames
+
+
+def collect_filepaths_from_sequential_path(
+    file_path,
+    frame_start=None,
+    frame_end=None,
+    only_existing=False,
+):
+    """Generate expected files from path
```
BigRoy, line 16: Can we add some example `path` to the docstring?

```python
collect_filepaths_from_sequential_path("/path/to/fname.####.exr")
collect_filepaths_from_sequential_path("/path/to/fname.%d.exr")
collect_filepaths_from_sequential_path("name.%04d.exr")
collect_filepaths_from_sequential_path("name.<frame>.exr")
collect_filepaths_from_sequential_path("name.[1001-1050].exr")
```

It's not that clear from the docstring that:
- The paths must be absolute paths, not filenames
- That only #### and %04d are supported

openpype/pipeline/file_collection.py
```diff
@@ -0,0 +1,274 @@
...
+    ):
+        return file_path
+
+    expected_files = []
+    for frame in range(int(frame_start), (int(frame_end) + 1)):
+        frame_file_path = os.path.join(
+            dirpath, formattable_string.format(frame)
+        )
+        # normalize path
+        frame_file_path = os.path.normpath(frame_file_path)
```
BigRoy, line 48: We don't need to do this per file right, we can do outside the loop
```python
dirpath = os.path.normpath(dirpath)
```
And `os.path.join` will always join in the normpath way, right?

openpype/pipeline/file_collection.py
```diff
@@ -0,0 +1,274 @@
...
+        # use regex to convert #### to {:0>4}
+        def replace(match):
+            return "{{:0>{}}}".format(len(match.group()))
+        new_filename = re.sub("#+", replace, filename)
+
+    elif "%" in filename:
+        # use regex to convert %04d to {:0>4}
+        def replace(match):
+            return "{{:0>{}}}".format(match.group()[1:])
+        new_filename = re.sub("%\\d+d", replace, filename)
```
BigRoy, line 142: Are we sure that this catches all we want?
```python
import re

def replace(match):
    return "{{:0>{}}}".format(match.group()[1:])
    
    
for filename in [
    "hello.%04d",
    "hello.%4d",
    "hello.%d",
    "hello.%1d"
]:
    new_filename = re.sub("%\\d+d", replace, filename)
    print(filename, "  -->  ", new_filename)
```

```
hello.%04d   -->   hello.{:0>04d}
hello.%4d   -->   hello.{:0>4d}
hello.%d   -->   hello.%d
hello.%1d   -->   hello.{:0>1d}
```

openpype/pipeline/file_collection.py
```diff
@@ -0,0 +1,274 @@
...
+        collected_files (list[str]): list of files
+
+    Returns:
+        Any[str, None]: single filepath or None if not possible
+    """
+    collections, reminders = clique.assemble(collected_files)
+    if collections:
+        return collections[0].format("{head}{padding}{tail}")
+    elif reminders:
+        return reminders[0]
```
BigRoy, line 274:
```suggestion
    elif remainders:
        return remainders[0]
```

openpype/pipeline/file_collection.py
```diff
@@ -0,0 +1,274 @@
...
+        frame_slate_str = frames.get_frame_start_str(
+            frame_start - 1,
+            frame_end
+        )
+
+        slate_frame = collected_file_frames[0].replace(
+            frame_start_str, frame_slate_str)
+        collected_file_frames.insert(0, slate_frame)
+
+        return True
```
BigRoy, line 258: I feel like this is such an odd function haha - it changes a list in place if it the amount of file frames matches the frame start and end. Such an odd requirement.

openpype/pipeline/file_collection.py
```diff
@@ -0,0 +1,274 @@
...
+    frame_start_str = frames.get_frame_start_str(
+        frame_start, frame_end)
+    frame_length = int(frame_end - frame_start + 1)
+
+    # add slate frame only if it is not already in collected frames
+    if frame_length == len(collected_file_frames):
+        frame_slate_str = frames.get_frame_start_str(
+            frame_start - 1,
+            frame_end
+        )
```
BigRoy, line 252: The logic here is wrong. You want to insert a file into an existing sequence that already has 'some sort of potential padding' and want to insert a frame before the first.

The `get_frame_start_str` call returns the first frame padded with the length of the last frame.
If the current collected file frames range are
```
frame.999.exr
frame.1000.exr
frame.1001.exr
```
You are now suddenly prepending `frame.0998.exr`
You should actually be matching the padding from the actual padding in the existing sequence for this function to be correct.

openpype/pipeline/frames.py
```diff
@@ -0,0 +1,79 @@
...
+
+def get_asset_frame_range(asset_name=None, asset_id=None, fields=None):
+    """Get the current assets frame range and handles.
+
+    Args:
+        asset_name Optional[str]: Name of asset used for filter.
+        asset_id Optional[Union[str, ObjectId]]: Asset document id.
+            If entered then is used as only filter.
+        fields Optional[Union[List[str], None]]: Limit returned data
+            of asset documents to specific keys.
```
BigRoy, line 34: Arguments are unused.

I feel like this makes more sense as an abstraction like ```get_asset_frame_range(asset_doc)``` because for moments when you already have the asset document that'd be much faster.

Currently also feels way too much like a function that should be in `context_tools`

openpype/pipeline/frames.py
```diff
@@ -0,0 +1,79 @@
...
+    Returns:
+        dict: with frame start, frame end, handle start, handle end.
+    """
+    from openpype.pipeline.context_tools import get_current_project_asset
+
+    # Set frame start/end
+    asset = get_current_project_asset()
```
BigRoy, line 41: We only need a few fields, right?
```suggestion
    asset = get_current_project_asset(fields=["data.frameStart", "data.frameEnd", "data.handleStart", "data.handleEnd"])
```

openpype/pipeline/frames.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+
+    collections, remainder = clique.assemble(collected_files)
+    if not collections:
+        # No sequences detected and we can't retrieve
+        # frame range from single file
+        return None, None
+
+    collection = collections[0]
+    repres_frames = list(collection.indexes)
```
BigRoy, line 77: I think this hides away bugs/issues where the collected files might actually be detected as separate collections - so we should at least be asserting that there isn't more than one.

openpype/pipeline/file_collection.py
```diff
@@ -0,0 +1,274 @@
...
+
+    Args:
+        collected_files (list[str]): list of files
+
+    Returns:
+        Any[str, None]: single filepath or None if not possible
+    """
+    collections, reminders = clique.assemble(collected_files)
+    if collections:
+        return collections[0].format("{head}{padding}{tail}")
```
BigRoy, line 272: It's good to be aware that this doesn't return a valid filepath - it'd be a 'human readable representation of the sequence' right? The docstring doesn't mention at all that this is just for 'nice labeling' for the user.

Also, is this function unused? I don't see it being used. Remove?

openpype/pipeline/file_collection.py
```diff
@@ -0,0 +1,274 @@
...
+    Returns:
+        Any[list[str], str]: List of expected absolute paths to files
+            (frames of a sequence) or path if single file
+    """
+    return collect_filepaths_from_sequential_path(
+        frame_start,
+        frame_end,
+        path,
+        only_existing=False,
+    )
```
BigRoy, line 84: It's funny that this function calls `collect_filepaths_from_sequential_path` even though I'd have expected it to be the opposite:
```python
def collect_filepaths_from_sequential_path(*args, **kwargs):
    expected = generate_expected_files(*args, **kwargs)
    return [path for path in expected if os.path.exists(path)]
```

To me it makes no sense by the way that this will return either a single filepath or a list of files. I'd expect it to always return a list of files.

BigRoy (commented):
To me this adds a lot of code that isn't necessarily a super helpful pipeline-wide abstraction. I'd say at most that it helps a bit with logic on the houdini side.

It being in pipeline makes it feel like it'll work for much more cases than it does. There are unfortunately more ways of writing a sequential path out there than just `####` and `%04d`. E.g. the ridiculous `<frames>`

openpype/pipeline/file_collection.py
jakubjezek001: Great catch! But my head still bit hurt with some of your examples.. why whould you use either `%4d` or `%d` - have never seen it on any dccs.. Here is the fix.. 

```
import re

for filename in [
    "hello.%04d",
    "hello.%4d",
    "hello.%d",
    "hello.%1d"
]:
    number = 20
    padding = re.search("%(\\d)+d", filename)
    padding = padding.group(1) if padding else 1
    print("padding", padding)
    new_filename = re.sub(
        "%.*d",
        "{{:0>{}}}".format(padding),
        filename
    )
    print(filename, "  -->  ", new_filename)
    print(filename % number, "  -->  ", new_filename.format(number))
    ```

File: openpype/pipeline/file_collection.py (hunk 5 above)
Comment (jakubjezek001) [reply openpype/pipeline/file_collection.py]: issue is that there might be cases where slate frame is already part of collected files

File: openpype/pipeline/file_collection.py (hunk 10 above)
Comment (jakubjezek001) [reply openpype/pipeline/file_collection.py:277]: it is implemented here https://github.com/ynput/OpenPype/pull/5498

File: openpype/pipeline/frames.py [hunk 12]
```diff
@@ -0,0 +1,83 @@
...
+        frame_end (int): last frame
+
+    Returns:
+        str: frame start string
+    """
+    padding = len(str(frame_end))
+    return str(frame_start).zfill(padding)
+
+
+def get_asset_frame_range(asset_name=None, asset_id=None, fields=None):
```
iLLiCiTiT, line 26: Pipeline functions should require what is minimum required to run properly and never use current context.
```suggestion
def get_asset_frame_range(project_name, asset_name, asset_doc=None):
```

openpype/pipeline/frames.py
```diff
@@ -0,0 +1,83 @@
+"""Utility functions.
```
iLLiCiTiT, line 1: I would renamed the file to `frame_utils`.

openpype/pipeline/file_collection.py
```diff
@@ -0,0 +1,279 @@
...
+        frame_end (Optional[int]): last frame
+        log (Optional[Logger]): logger
+        only_existing (Optional[bool]): Ensure that files exists.
+        reviewable (Optional[bool]): reviewable
+
+
+    Returns:
+        dict[str, Any]: representation
+    """
+    log = log or Logger().get_logger(__name__)
```
iLLiCiTiT, line 176:
```suggestion
    log = log or Logger.get_logger(__name__)
```

openpype/pipeline/frames.py
```diff
@@ -0,0 +1,83 @@
...
+            If entered then is used as only filter.
+        fields Optional[Union[List[str], None]]: Limit returned data
+            of asset documents to specific keys.
+    Returns:
+        dict: with frame start, frame end, handle start, handle end.
+    """
+    from openpype.pipeline.context_tools import get_current_project_asset
+
+    # Set frame start/end
+    asset = get_current_project_asset(fields=["data.frameStart", "data.frameEnd", "data.handleStart", "data.handleEnd"])
```
iLLiCiTiT, line 41:
```suggestion
    if not asset_doc:
        asset_doc = get_asset_by_name(
            project_name, 
            asset_name,
            fields=[
                "data.frameStart", "data.frameEnd", "data.handleStart", "data.handleEnd"
            ]
        )
```

openpype/hosts/houdini/plugins/publish/collect_mantra_rop.py
```diff
@@ -31,49 +31,63 @@ class CollectMantraROPRenderProducts(pyblish.api.InstancePlugin):
     def process(self, instance):
 
         rop = hou.node(instance.data.get("instance_node"))
+        frame_start = instance.data["frameStart"]
+        frame_end = instance.data["frameEnd"]
```
moonyuet, line 35:
```suggestion
        frame_start = str(instance.data["frameStart"])
        frame_end = str(instance.data["frameEnd"])
```
If it is not a string it will error out in the collectors. See the log below.
```
DEBUG: Chunk Size: 10
  File "D:\kayla\OpenPype\.venv\lib\site-packages\pyblish\plugin.py", line 527, in __explicit_process
    runner(*args)
  File "D:\kayla\OpenPype\openpype\hosts\houdini\plugins\publish\collect_mantra_rop.py", line 54, in process
    "beauty": file_collection.generate_expected_filepaths(
  File "D:\kayla\OpenPype\openpype\pipeline\file_collection.py", line 79, in generate_expected_filepaths
    files = collect_filepaths_from_sequential_path(
  File "D:\kayla\OpenPype\openpype\pipeline\file_collection.py", line 30, in collect_filepaths_from_sequential_path
    os.path.dirname(file_path))
  File "C:\PROGRA~1/SIDEEF~1/HOUDIN~1.435/python39\lib\ntpath.py", line 223, in dirname
    return split(p)[0]
  File "C:\PROGRA~1/SIDEEF~1/HOUDIN~1.435/python39\lib\ntpath.py", line 185, in split
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not float
```

jakubjezek001: It doesn't look the frame start and end values are causing this issue. Also, those values are usually integer numbers. Please look carefully into the trackback.

jakubjezek001: it seems like the issue is in the order of input attributes. Fixing it

moonyuet: The problem resolved after the change of the order of input attributes

openpype/pipeline/file_collection.py
jakubjezek001: more info here 953e1d8fb48ee0ce21ca4c6bb6c8ea35a52a8846

moonyuet:
For the file collections, looks good for me.
But I can't really test out as  it errors out the submit_houdini_deadline, but this is not related to this issue as it also errors out when I tested with the latest develop branch
```
  File "D:\kayla\OpenPype\.venv\lib\site-packages\pyblish\plugin.py", line 527, in __explicit_process
    runner(*args)
  File "D:\kayla\OpenPype\openpype\modules\deadline\plugins\publish\submit_houdini_render_deadline.py", line 138, in process
    super(HoudiniSubmitDeadline, self).process(instance)
  File "D:\kayla\OpenPype\openpype\modules\deadline\abstract_submit_deadline.py", line 446, in process
    file_path = self.from_published_scene()
  File "D:\kayla\OpenPype\openpype\modules\deadline\abstract_submit_deadline.py", line 541, in from_published_scene
    return replace_with_published_scene_path(
  File "D:\kayla\OpenPype\openpype\pipeline\publish\lib.py", line 924, in replace_with_published_scene_path
    if isinstance(expected_files[0], dict):
TypeError: 'NoneType' object is not subscriptable

```

openpype/pipeline/file_collection.py
```diff
@@ -0,0 +1,292 @@
...
+):
+    """Generate expected files from path
+
+    Args:
+        frame_start (int): Start frame of the sequence
+        frame_end (int): End frame of the sequence
+        file_path (str): Absolute path with any sequential pattern (##, %02d)
+
+    Returns:
+        Any[list[str], str]: List of expected absolute paths to files
```
iLLiCiTiT, line 76:
```suggestion
        Union[list[str], str]: List of expected absolute paths to files
```

openpype/pipeline/file_collection.py
```diff
@@ -0,0 +1,292 @@
...
+        new_filename = re.sub(
+            "%.*d",
+            "{{:0>{}}}".format(padding),
+            filename
+        )
+
+    return new_filename
+
+
+def get_publishing_representation(
```
iLLiCiTiT, line 165:
```suggestion
def prepare_publish_representation(
```

openpype/pipeline/file_collection.py
```diff
@@ -0,0 +1,292 @@
...
+
+    Args:
+        instance (pyblish.Instance): instance
+        file_path (str): file path
+        frame_start (Optional[int]): first frame
+        frame_end (Optional[int]): last frame
+        log (Optional[Logger]): logger
+        only_existing (Optional[bool]): Ensure that files exists.
+        reviewable (Optional[bool]): reviewable
+
```
iLLiCiTiT, line 184:
```suggestion
```

openpype/pipeline/file_collection.py
BigRoy: Sorry, could you elaborate what happens in the other PR? Yes, that PR uses it - but that doesn't resolve my other note about this not returning actual valid paths in some cases.

openpype/pipeline/file_collection.py
```diff
@@ -0,0 +1,292 @@
...
+    if len(collected_file_frames) == 1:
+        representation["files"] = collected_file_frames.pop()
+    else:
+        representation["files"] = collected_file_frames
+
+    return representation
+
+
+def _add_slate_frame_to_collected_frames(
```
BigRoy, line 240: Would this be more understandable if the wording was `_prepend_slate_frame_...` so that you know where it adds it?

openpype/pipeline/file_collection.py
```diff
@@ -0,0 +1,292 @@
...
+        Any[list[str], str]: List of absolute paths to files
+            (frames of a sequence) or path if single file
+    """
+
+    dirpath = os.path.normpath(
+        os.path.dirname(file_path))
+
+    filename = os.path.basename(file_path)
+
+    formattable_string = convert_filename_to_formattable_string(
```
Minkiu, line 34: Wouldn't we be able to achieve something like what this function is doing by using `clique.assemble` ?

openpype/hosts/houdini/plugins/publish/collect_mantra_rop.py
```diff
@@ -31,49 +31,63 @@ class CollectMantraROPRenderProducts(pyblish.api.InstancePlugin):
...
+
+        files_by_aov = {
+            "beauty": file_collection.generate_expected_filepaths(
+                frame_start, frame_end, beauty_product)
+        }
+
+        # get the number of AOVs
+        aov_numbers = rop.evalParm("vm_numaux")
+        if aov_numbers == 0:
+            return
```
moonyuet, line 61: it will return everything and the collector wont get any expected files.

moonyuet (changes requested):
I tested with karma, redshift, vray rop. They all look good. But mantra fails to pass render deadline submission plugin as instance.data["files"] is returning to empty. When the houdini render deadline submission plugin tries to get files, it errors out.
I suggest some pesudo codes for the possible fix in the comment above which can help to get the data.
```
WARNING: Falling back to workfile
INFO: Using D:/test_project/new_project/ep/work/asset/new_project_ep_asset_v232.hiplc for render/export.
  File "D:\ayon_desktop\OpenPype\.venv\lib\site-packages\pyblish\plugin.py", line 527, in __explicit_process
    runner(*args)
  File "D:\ayon_desktop\OpenPype\openpype\modules\deadline\plugins\publish\submit_houdini_render_deadline.py", line 138, in process
    super(HoudiniSubmitDeadline, self).process(instance)
  File "D:\ayon_desktop\OpenPype\openpype\modules\deadline\abstract_submit_deadline.py", line 459, in process
    self.job_info = self.get_job_info()
  File "D:\ayon_desktop\OpenPype\openpype\modules\deadline\plugins\publish\submit_houdini_render_deadline.py", line 111, in get_job_info
    for i, filepath in enumerate(instance.data["files"]):
KeyError: 'files'

```

jakubjezek001:
closing this PR since it was just for research purpose only

PR #5498, opened 2023-08-24
Abstraction for publishing with prepared representation

jakubjezek001:
## Changelog Description
Unified way of preparing representation data for local and farm publishing.

## Additional info
- added into traypublisher simple collector
- added into nuke collect write

## Testing notes:
1. try publishing and nuke should work as it was before

## Dependencies
Changes from this PR were initially implemented in [this pull request](https://github.com/ynput/OpenPype/pull/5451). However, due to its significant size, we made the decision to divide it into smaller portions. 

**!!! This PR should be merged after #5496. But before the merge the target branch should be changed to develop !!!**

### DIFF
--- a/openpype/hosts/nuke/plugins/publish/collect_writes.py
+++ b/openpype/hosts/nuke/plugins/publish/collect_writes.py
@@ -6,6 +6,7 @@ from openpype.pipeline import publish
 
 
 class CollectNukeWrites(pyblish.api.InstancePlugin,
+                        publish.PrepRepresentationPluginMixin,
                         publish.ColormanagedPyblishPluginMixin):
     """Collect all write nodes."""
 
@@ -14,17 +15,11 @@ class CollectNukeWrites(pyblish.api.InstancePlugin,
     hosts = ["nuke", "nukeassist"]
     families = ["render", "prerender", "image"]
 
-    # cache
-    _write_nodes = {}
-    _frame_ranges = {}
-
     def process(self, instance):
 
         group_node = instance.data["transientData"]["node"]
-        render_target = instance.data["render_target"]
 
         write_node = self._write_node_helper(instance)
-
         if write_node is None:
             self.log.warning(
                 "Created node '{}' is missing write node!".format(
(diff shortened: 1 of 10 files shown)


openpype/hosts/traypublisher/plugins/publish/collect_simple_instances.py
```diff
@@ -47,90 +53,93 @@ def process(self, instance):
...
         ).format(instance_label, tmp_folder))
 
+        creator_attributes = instance.data["creator_attributes"]
+        publish_target = creator_attributes["publish_target"]
+        review_file_item = creator_attributes["reviewable"]
+        review_filenames = review_file_item.get("filenames")
+
+        # make instance reviewable if reviewable attribute is set
+        if review_filenames:
+            instance.data["review"] = True
```
iLLiCiTiT, line 63: Why this? We want to get rid of this key for few years.

openpype/pipeline/publish/publish_plugins.py
```diff
@@ -380,64 +349,149 @@ def set_representation_colorspace(
...
 
-        self.log.debug("__ filename: `{}`".format(filename))
+        # make sure rendered sequence on farm will
+        # be used for extract review
+        instance.data["useSequenceForReview"] = bool(
+            instance.data.get("review"))
+
+        # Farm rendering
+        instance.data.update({
+            "transfer": False,
```
iLLiCiTiT, line 406: `"transfer"` is not used anywhere.

openpype/hosts/traypublisher/plugins/publish/collect_simple_instances.py
```diff
@@ -47,90 +53,93 @@ def process(self, instance):
...
-            instance.data["originalBasename"] = origin_basename
+        representation_files = self.get_processing_file_data(
+            instance, filepath_items)
+
+        for file_path, file_data in representation_files.items():
+            # making sure file extension is publishable over farm
+            file_ext = os.path.splitext(file_path)[-1].lower().lstrip(".")
+            if file_ext not in self.pixel_ext:
+                # allow farm publishing only for pixel type data
+                publish_target = "local"
```
iLLiCiTiT, line 85: Publish targets should not be part of this PR.

openpype/pipeline/publish/publish_plugins.py
```diff
@@ -380,64 +349,149 @@ def set_representation_colorspace(
...
-            filename = filename[0]
+        Args:
+            instance (pyblish.api.Instance): pyblish instance
+        """
 
-        self.log.debug("__ filename: `{}`".format(filename))
+        # make sure rendered sequence on farm will
+        # be used for extract review
+        instance.data["useSequenceForReview"] = bool(
+            instance.data.get("review"))
```
iLLiCiTiT, line 402: Review of instance is define by `"review"` family. As metioned we wanted to get rid of the key `"review"` for a very long time, this is adding new, same, key that we actually don't want.

openpype/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -192,8 +191,7 @@ def _submit_deadline_post_job(self, instance, job, instances):
 
         # Transfer the environment from the original job to this dependent
         # job so they use the same environment
-        metadata_path, rootless_metadata_path = \
-            create_metadata_path(instance, anatomy)
+        _, rootless_metadata_path = create_metadata_path(instance, anatomy)
```
fabiaserra, line 194: in our fork of OP I have changed the `_submit_deadline_post_job` function so it not only returns `deadline_publish_job_id` but also the `metadata_path` (the one you are now ignoring with `_`) to avoid the second call of this same function later on the code path here https://github.com/ynput/OpenPype/blob/eec21037a3c79b000afea44edd1494af397f3c14/openpype/modules/deadline/plugins/publish/submit_publish_job.py#L508

So it ends looking like this:
```python
deadline_publish_job_id, metadata_path = self._submit_deadline_post_job(instance, render_job, instances)
```

The reason we had to do this is because in our Hiero integration we are submitting farm jobs for different assets and subsets from the same parent "asset" (so we can ingest multiple shots from just the episode context) and so I changed the metadata.json path to be unique for each subset and it also helps to add a timestamp for debugging purposes and enforcing even more uniqueness:
```
    # Prefixing metadata file with timestamp and asset so the .json files are
    # unique and not overwrite each other. This is necessary because in Hiero
    # we use the same working directory to publish multiple subsets at once
    # and when the subset was called the same, it was overwriting the same file
    # over and over
    metadata_filename = "{}_{}_{}_metadata.json".format(
        datetime.datetime.now().strftime("%d%m%Y%H%M%S"),
        ins_data["asset"],
        ins_data["subset"]
    )
```
And because of the current logic of calling the `create_metadata_path` twice... that timestamp would return a second difference between both executions,

Making the json files unique might not be something you are interested in but I think it makes the `submit_publish_job` code more understandable too if you only "create" that metadata path once (it was confusing for me to read the first time why there were two "creations" of the path)

** Bonus points: I think `get_metadata_path` defines semantically a bit better what the function does as the file isn't being created in the function

Minkiu: @jakubjezek001 have you seen this?

jakubjezek001: @fabiaserra I totally agree this should be redesigned with your suggestion. The plan is to include the editorial publishing at the farm too, so this would need to be dealt with anyway. What do  you thing @iLLiCiTiT - should this be part of this PR?

openpype/pipeline/file_collection.py
```diff
@@ -252,22 +286,30 @@ def _add_slate_frame_to_collected_frames(
     Returns:
         Bool: True if slate frame was added
     """
+    print(type(frame_start), type(frame_end))
```
iLLiCiTiT, line 289:
```suggestion
```

kalisp:
I think there are some conflicts with `develop` in `submit_publish_job.py` and `openpype/pipeline/publish/publish_plugins.py`

openpype/hosts/nuke/plugins/publish/collect_writes.py
```diff
@@ -33,112 +28,93 @@ def process(self, instance):
...
+            # QUESTION: should we set colorspace at this moment or downstream?
+            self.set_representation_colorspace(
+                representation, instance.context,
+                colorspace=colorspace
+            )
+            instance.data["representations"].append(representation)
 
-        write_file_path = nuke.filename(write_node)
-        output_dir = os.path.dirname(write_file_path)
+            self.add_farm_instance_data(instance)
```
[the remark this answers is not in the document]
iLLiCiTiT: Agree on this it could be just
```python
if render_target in {"frame", "frames_farm"}:
    file_paths = self.prepare_collection_of_file_paths(
        write_file_path, frame_start, frame_end, only_existing=True
    )

    representation = self.prepare_representation(
        instance, file_paths, frame_start, frame_end
    )

    # QUESTION: should we set colorspace at this moment or downstream?
    self.set_representation_colorspace(
        representation, instance.context,
        colorspace=colorspace
    )

if render_target in {"farm", "frames_farm"}:
    self.add_farm_instance_data(instance)
```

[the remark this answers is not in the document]
BigRoy: I'd personally like it actually if we just have a basic mixin we can use everywhere - similar to the extractor for `staging_dir` method that would expose e.g. the `add_representation` method and have all extractors use that instead. That way, in some time, if the data storage on the instances need to change we can just change whatever the method does.

Plus, having a dedicated function can also make sure the actual representation gets validated at that point in time, etc.

Anyway, having dedicated methods for that has come up quite a bit as a request actually. I believe there's a forum discussion.

openpype/hosts/traypublisher/plugins/publish/collect_simple_instances.py
```diff
@@ -140,128 +123,68 @@ def _create_main_representations(
...
-                " Creation of review representation was skipped."
-            ))
-            return
+            file_path = self.get_single_filepath_from_list_of_files(
+                filepaths)
+            frame_start, frame_end = self.get_frame_range_from_list_of_files(
+                filepaths)
+            processing_file_data = {
+                file_path: {
+                    "framerange": (frame_start, frame_end)
```
Minkiu, line 137: As far as I can tell, the method returns a `tuple` and we are not checking if they are non, so we can skip line 133 and call it here directly:
```py

processing_file_data = {
	file_path: {
		"framerange": self.get_frame_range_from_list_of_files(
			filepaths
		)
	}
}
```

tokejepsen:
@jakubjezek001 can we resolved the comments on this before further testing?

mkolar:

PR #5499, opened 2023-08-25
Max: Implementation of OCIO configuration 

### LINKED ISSUE #5473: Bug: 3dsmax is not using predefined OCIO config file in OP Settings when Enabled CMS

### Current Behavior:

When turning ```On``` global Color Management in OP Project Settings
image

3dsmax host CM also turned ```On``` too

image

It does not start up with that particular OCIO config file defined but with different Color Management settings.

image

There have to be taken extra step manually switching to OCIO based CMS:

image

### Expected Behavior:

_No response_

### Version

3.16.4-nightly.3

Windows

### Steps To Reproduce:

1.Turn Global CMS on in OP Settings
2. Turn CMS for Max host app in OP Settings
3. Launch Max via OP Launcher
4. Check File>Preferences>Color Management Settings for OCIO config file based management
5. Missing OCIO based Color Management mode present (need to be set manually by user which is wrong)

### Relevant log output:

_No response_

### Additional context:

All CMS settings with OCIO config file should be set automatically on launch via OP.

[cuID:[OP-6732](https://app.clickup.com/t/861najumt)]

moonyuet:
## Changelog Description
Resolve #5473 
Implementation of OCIO configuration for Max 2024 regarding to the update of Max 2024

## Additional info
Not applicable for Max 2023 or lower.
(Will update the test note later)

## Testing notes 1:
1. Launch 3dsMax via launcher
2. Start a new scene for 3dsMax
3. Create Render Instance
4. Publish

## Testing notes 2:
1. Launch 3dsMax via launcher
2. Go to Openpype -> Set Colorspace
3. Create Review Instance
4. Publish


LiborBatek:
Just a quick question, is this corelating to our present way of doing CMS accross the all host?

I mean CMS should be set withing global OP Settings so is this still the case or?
image

moonyuet:
> Just a quick question, is this corelating to our present way of doing CMS accross the all host?
> 
> I mean CMS should be set withing global OP Settings so is this still the case or?
> 
> Global CMS and OCIO config defined in OP Settings
> 
> image
> 
> and for Max host specifically:
> 
> image

The color configuration should be taking the last OCIO config file(aces 1.2) from the global CMS after set colorspace in OP menu or start a new scene.
The configuration is the same with the one in Houdini.
image

image

LiborBatek:
Ok cool then...I was more worried it doesnt work like that :) ...so all good

Will test asap. thx!

openpype/hosts/max/api/lib.py
```diff
@@ -312,3 +316,30 @@ def set_timeline(frameStart, frameEnd):
...
+        ocio_config = global_imageio["ocio_config"]
+        ocio_config_path = ocio_config["filepath"][-1]
+
+    max_imageio = get_project_settings(
+        project_name)["max"]["imageio"]
+    if max_imageio["activate_host_color_management"]:
+        ocio_config = max_imageio["ocio_config"]
+        if ocio_config["override_global_config"]:
+            ocio_config_path = ocio_config["filepath"][0]
+
```
Minkiu, line 344: Should we check here that `ocio_config_path` is not empty?

Minkiu (approved):

LiborBatek:
Meanwhile when testing (will post comments on that soon) ...we should also treat case of older workfiles in max set to ```Gamma Workflow``` which being ```3dsmax``` default one (at least in max ```2024``` )
image

We should somehow force it...
 or at least user should get sort of Warning that OCIO color management not set and need to be switched to our custom OCIO config.

This mainly happening when re opening older max workfiles before this PR.

moonyuet:
> Meanwhile when testing (will post comments on that soon) ...we should also treat case of older workfiles in max set to `Gamma Workflow` which being `3dsmax` default one (at least in max `2024` ) image
> 
> We should somehow force our OCIO config based CMS... or at least user should get sort of Warning that OCIO color management not set and need to be switched to our custom OCIO config.
> 
> This mainly happening when re opening older max workfiles before this PR.
Maybe we can do an action similar to the one in maya which fix fps when the old 3dsmax workfile scene is being opened....

LiborBatek:
@moonyuet  yeah, is as simple as going into Max Preferences and choose custom OCIO config based CMS (our config already being in the list present just not set as working color management setting)

...basically just not being switched on

LiborBatek (commented):
Now OCIO correctly been visible when creating ```Render``` instance
image

I propose to keep it exposed on ```Render``` instance when re opening ```Publish``` now its not being visible after going into it again...so user is not aware/able to change color profiles anymore easily
image

moonyuet:
I have updated the scripts with collect review can set colorspace and the colorspace options move to the publish tab.
Also for the old workfile it would have the popup menu but for some reason the color of the push button is not applicable for the main_window set for max. 
image
image
image

LiborBatek (approved):
All seems working fine.  Max being correctly set to OP OCIO config file and if not,
then the pop up menu shows with option to fix that. Also works fine.

image

There is now visible the option box on ```render``` instance for color profiles too, so user has opportunity to tweak the color profiles later on if needed.

image

openpype/hosts/max/api/lib.py
```diff
@@ -314,6 +345,54 @@ def set_timeline(frameStart, frameEnd):
...
+                 "because Max main window can't be found.")
+    if int(get_max_version()) >= 2024:
+        color_mgr = rt.ColorPipelineMgr
+        if color_mgr.Mode != rt.Name("OCIO_Custom"):
+            from openpype.widgets import popup
+            dialog = popup.Popup(parent=parent)
+            dialog.setWindowTitle("Warning: Wrong OCIO Mode")
+            dialog.setMessage("This scene has wrong OCIO "
+                              "Mode setting.")
+            dialog.widgets["button"].setText("Fix")
```
iLLiCiTiT, line 391:
```suggestion
            dialog.setButtonText("Fix")
```

openpype/hosts/max/api/lib.py
```diff
@@ -314,6 +345,54 @@ def set_timeline(frameStart, frameEnd):
...
+    if int(get_max_version()) >= 2024:
+        color_mgr = rt.ColorPipelineMgr
+        if color_mgr.Mode != rt.Name("OCIO_Custom"):
+            from openpype.widgets import popup
+            dialog = popup.Popup(parent=parent)
+            dialog.setWindowTitle("Warning: Wrong OCIO Mode")
+            dialog.setMessage("This scene has wrong OCIO "
+                              "Mode setting.")
+            dialog.setButtonText("Fix")
+            dialog.on_clicked.connect(reset_colorspace)
```
iLLiCiTiT, line 392:
```suggestion
            dialog.setStyleSheet(load_stylesheet())
            dialog.on_clicked.connect(reset_colorspace)
```

Requires to add import at the top of the file
```
from openpype.style import load_stylesheet
```

openpype/hosts/max/api/lib.py
```diff
@@ -314,6 +345,54 @@ def set_timeline(frameStart, frameEnd):
...
+def reset_colorspace():
+    """OCIO Configuration
+    Supports in 3dsMax 2024+
+
+    """
+    if int(get_max_version()) < 2024:
+        return
+    project_name = get_current_project_name()
+    colorspace_mgr = rt.ColorPipelineMgr
+    colorspace_mgr.Mode = rt.Name("OCIO_Custom")
```
BigRoy, line 359: This should only really be applied if color management is enabled, right?

openpype/hosts/max/api/lib.py
```diff
@@ -314,6 +345,54 @@ def set_timeline(frameStart, frameEnd):
...
+
+    max_imageio = get_project_settings(
+        project_name)["max"]["imageio"]
+    if max_imageio["activate_host_color_management"]:
+        ocio_config = max_imageio["ocio_config"]
+        if ocio_config["override_global_config"]:
+            ocio_config_path = ocio_config["filepath"][0]
+            if not ocio_config_path:
+                # use the default ocio config path instead
+                ocio_config_path = ocio_config["filepath"][-1]
```
BigRoy, line 375: You are retrieving the full project settings twice - re-use it please.

Also, shouldn't `colorspace.get_imageio_config` be doing this to get the config you need to apply instead? @jakubjezek001 


```python

def reset_colorspace():
    """OCIO Configuration
    Supports in 3dsMax 2024+
    """
    if int(get_max_version()) < 2024:
        return

    config_data = colorspace.get_imageio_config(project_name, 
                                                "max", 
                                                project_settings)
    if config_data:
        path = config_data["path"]    
        colorspace_mgr = rt.ColorPipelineMgr
        colorspace_mgr.Mode = rt.Name("OCIO_Custom")
        colorspace_mgr.OCIOConfigPath = ocio_config_path
```

openpype/hosts/max/api/lib.py
```diff
@@ -314,6 +345,54 @@ def set_timeline(frameStart, frameEnd):
...
+
+
+def check_colorspace():
+    parent = get_main_window()
+    if parent is None:
+        log.info("Skipping outdated pop-up "
+                 "because Max main window can't be found.")
+    if int(get_max_version()) >= 2024:
+        color_mgr = rt.ColorPipelineMgr
+        if color_mgr.Mode != rt.Name("OCIO_Custom"):
```
BigRoy, line 387: The OCIO custom mode should only be enforced if color management is enabled, right?

openpype/hosts/max/api/lib.py
BigRoy: Note that this won't work in headless/batch mode so you might want to consider a fallback for that.

openpype/hosts/max/plugins/create/create_render.py
```diff
@@ -30,3 +30,28 @@ def create(self, subset_name, instance_data, pre_create_data):
...
+            for display in sorted(displays):
+                views = colorspace_mgr.GetViewList(display)
+                for view in sorted(views):
+                    ocio_display_view_transform_list.append({
+                        "value": "||".join((display, view))
+                    })
+                    if display == "ACES" and view == "sRGB":
+                        display_view_default = "{0}||{1}".format(
+                            display, view
+                        )
```
BigRoy, line 50: Note that `get_instance_attr_defs` is called PER instance and quite often, more often that `apply_settings`. You could potentially optimize this by caching the result in apply settings to a class attribute.

openpype/hosts/max/plugins/publish/collect_render.py
```diff
@@ -35,6 +35,8 @@ def process(self, instance):
         files_by_aov.update(aovs)
 
         camera = rt.viewport.GetCamera()
+        if instance.data.get("members"):
+            camera = instance.data["members"][-1]
```
BigRoy, line 39: Is there something currently ensuring that the contained member is actually a camera? If not, we should probably prepare for that.

openpype/modules/deadline/plugins/publish/submit_max_deadline.py
```diff
@@ -238,7 +238,7 @@ def _use_published_name(self, data, project_settings):
             plugin_data["redshift_SeparateAovFiles"] = instance.data.get(
                 "separateAovFiles")
         if instance.data["cameras"]:
-            plugin_info["Camera0"] = None
+            plugin_info["Camera0"] = instance.data["cameras"][0]
             plugin_info["Camera"] = instance.data["cameras"][0]
             plugin_info["Camera1"] = instance.data["cameras"][0]
```
BigRoy, line 243:
```suggestion
            camera = instance.data["cameras"][0]
            plugin_info["Camera0"] = camera
            plugin_info["Camera"] = camera
            plugin_info["Camera1"] = camera
```

openpype/hosts/max/plugins/publish/collect_review.py
```diff
@@ -54,8 +65,28 @@ def process(self, instance):
...
+            for display in sorted(displays):
+                views = colorspace_mgr.GetViewList(display)
+                for view in sorted(views):
+                    ocio_display_view_transform_list.append({
+                        "value": "||".join((display, view))
+                    })
+                    if display == "ACES" and view == "sRGB":
+                        display_view_default = "{0}||{1}".format(
+                            display, view
+                        )
```
BigRoy, line 84: Can't we just label them like this:

```python
display_views = {}
if int(get_max_version()) >= 2024:
    colorspace_mgr = rt.ColorPipelineMgr
    for display in sorted(colorspace_mgr.GetDisplayList()):
        for view in sorted(colorspace_mgr.GetViewList(display)):
            display_views[(display, view)] = "{} || {}".format(display, view)
else:
    display = "sRGB"
    view = "ACES 1.0 SDR-Video"
    display_views[(display, view)] = "{} || {}".format(display, view)

# if the default value does not exist it is ignored by EnumDef
# and the first entry is taken instead
default_value = ("ACES", "sRGB")  
```

Where `items` for the EnumDef then is the `display_views` dict?
The resulting value from the enumdef would then also be that tuple.
So in the code you can then do: `display, view = creator_attributes["ocio_display_view_transform"]` if I'm not mistaken.

openpype/hosts/max/api/lib.py
```diff
@@ -1,15 +1,45 @@
...
+log = logging.getLogger("openpype.hosts.max")
+
+
+class Context:
+    main_window = None
+    context_label = None
+    project_name = os.getenv("AVALON_PROJECT")
+    # Workfile related code
+    workfiles_launched = False
+    workfiles_tool_timer = None
```
BigRoy, line 28: These seem unused in this PR and I feel it's also not up to the max lib to keep these 'global singleton' attributes available. What's the need for this?
```suggestion
```

openpype/hosts/max/api/pipeline.py
```diff
@@ -58,6 +58,9 @@ def context_setting():
         rt.callbacks.addScript(rt.Name('systemPostNew'),
                                context_setting)
 
+        rt.callbacks.addScript(rt.Name('filePostOpen'),
+                               lib.check_colorspace)
```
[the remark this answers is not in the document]
moonyuet: i agree that this one needs to be resolved somehow. I will create issue soon...and find some time to fix it..

openpype/hosts/max/plugins/publish/collect_render.py
moonyuet: we can do a check to get the list of object in instance.data["members"] which is camera. 
Some code similar to this : 
```
camera_list = [member for member in instance.data["members"] if rt.ClassOf(member) == rt.Camera.Classes]
```

openpype/hosts/max/plugins/publish/collect_review.py
moonyuet: I agree this method is more concise than with the list being in enumDef for this case (less lines of code),  but I have tried to pass the tuple as default_value it seems it seems the enum settings doesn't get it (let say the default value is ("ACES, "sRGB", but the setting doesn't show ACES || sRGB as default_value (the first value of the dict as the default instead)). 
So I stick with the list as the enum setting.

openpype/hosts/max/api/lib.py
moonyuet: I have tested it from my side, this code works.

openpype/hosts/max/api/pipeline.py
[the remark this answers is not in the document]
moonyuet: The related issue is https://github.com/ynput/OpenPype/issues/5577

openpype/hosts/max/api/lib.py
moonyuet: I have added the condition with the new headless function check in max.

openpype/hosts/max/api/lib.py
```diff
@@ -384,12 +365,20 @@ def check_colorspace():
...
         color_mgr = rt.ColorPipelineMgr
-        if color_mgr.Mode != rt.Name("OCIO_Custom"):
+        project_name = get_current_project_name()
+        project_settings = get_project_settings(
+        project_name)
+        global_imageio = project_settings["global"]["imageio"]
+        max_config_data = colorspace.get_imageio_config(
+        project_name, "max", project_settings)
+        config_enabled = global_imageio["activate_global_color_management"] or (
+            max_config_data)
```
BigRoy: I believe the config enabled check isn't needed explicitly because I believe `get_imageio_config` actually returns an empty dict if it is disabled, right?

PR #5505, opened 2023-08-25
Publisher: Fix multiselection value

iLLiCiTiT:
## Changelog Description
Selection of multiple instances in Publisher does not cause that all instances change all publish attributes to the same value.

## Additional info
It seems that text and boolean attributes are only one which are actually changing value of widget, which cause callback that change the value back to the instances. Changed the logic to not trigger any signals of a multiselection value is set.

## Testing notes:
1. Choose a creator which creates instances that have in Publish Attributes Boolean and Text fields
2. Create 2 instances
3. Select first instance and change both values
4. Select second instance and change values to be different from the first
5. Select both
6. Select only first, the values should not change from state before
7. Select only second, the values should not change from state before

### DIFF
--- a/openpype/tools/attribute_defs/widgets.py
+++ b/openpype/tools/attribute_defs/widgets.py
@@ -343,6 +343,7 @@ class TextAttrWidget(_BaseAttrDefWidget):
         return self._input_widget.text()
 
     def set_value(self, value, multivalue=False):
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
This fixes the issue I was facing on my end.

PR #5508, opened 2023-08-25
Maya: fix error when exporting alembic cameras

friquette:
## Changelog Description
When exporting cameras in alembic, there's an ancestor relationship error in the root arguments. So we put the second root argument in a condition.
What do you think of this solution ?

## Additional info
Here's the error we got:
image

### DIFF
--- a/openpype/hosts/maya/api/lib.py
+++ b/openpype/hosts/maya/api/lib.py
@@ -2260,7 +2260,7 @@ def get_frame_range(include_animation_range=False):
     return frame_range
 
 
-def reset_frame_range(playback=True, render=True, fps=True):
+def reset_frame_range(playback=True, render=True, fps=True, instances=True):
     """Set frame range to current asset
 
     Args:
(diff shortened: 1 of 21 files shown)


BigRoy:
Could it be that your publish instance (the object set) contains both cameras and parent transforms, or camera shapes and their transforms?

Could you show your set members in the outliner. I suspect the issue is there because I can't reproduce it on my end.

This is what I'd expect:
image

BigRoy:
Also note that your screenshot is using the old style publisher. Current `develop` uses the new style publisher. Would be good to confirm the issue there too.

friquette:
> Could it be that your publish instance (the object set) contains both cameras and parent transforms, or camera shapes and their transforms?
> 
> Could you show your set members in the outliner. I suspect the issue is there because I can't reproduce it on my end.
> 
> This is what I'd expect: image

This is what our set looks like:
image
Our cameras are in the grp_cam group

tokejepsen:
Can confirm that having a camera parented to a group and publishing the group works in latest `develop`, so cant reproduce the issue of this PR.

@friquette when you have a reproducible scene or steps, let us know and mark the PR ready for review.

tokejepsen:
@friquette is this still needed?
If so, we'll need some testing notes to replicate the issue.

mkolar:
@friquette we're closing this PR as it is now stale and can't be reproduced.

PR #5514, opened 2023-08-25
Chore: Queued event system

iLLiCiTiT:
## Changelog Description
Implemented event system with more expected behavior of event system. If an event is triggered during other event callback, it is not processed immediately but waits until all callbacks of previous events are done. The event system also allows to not trigger events directly once `emit_event` is called which gives option to process events in custom loops.

## Additional info
The event system is not used anywhere, I need it for refactor of workfiles tool, where order of events matter. The changes can be added as part of that PR, but I thought it would be easier to do a review in smaller PR.

### DIFF
--- a/openpype/lib/events.py
+++ b/openpype/lib/events.py
@@ -3,6 +3,7 @@ import os
 import re
 import copy
 import inspect
+import collections
 import logging
 import weakref
 from uuid import uuid4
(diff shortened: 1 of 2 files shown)


BigRoy:
I like that this is implemented in isolation for review. 

However, with it unused currently - right? - I think it's the perfect moment to implement tests now to ensure the behavior is correct. It'll also help to ensure it will remain to function the expected way over time.

kalisp (approved):

PR #5520, opened 2023-08-28
Max: Fixing the bug of falling back to use workfile for Arnold or any renderers except Redshift

moonyuet:
## Changelog Description
Fix the bug of falling back to use workfile for Arnold 
## Additional info
n/a
## Testing notes:
1. Launch Max via launcher
2. Create Render Instance with Arnold or some renderers
3. Publish

### DIFF
--- a/openpype/modules/deadline/plugins/publish/submit_max_deadline.py
+++ b/openpype/modules/deadline/plugins/publish/submit_max_deadline.py
@@ -249,12 +249,8 @@ class MaxSubmitDeadline(abstract_submit_deadline.AbstractSubmitDeadline,
         if instance.data["renderer"] == "Redshift_Renderer":
             self.log.debug("Using Redshift...published scene wont be used..")
             replace_in_path = False
-            return replace_with_published_scene_path(
-                instance, replace_in_path)
-        else:
-            return replace_with_published_scene_path(
-                instance, replace_in_path)
-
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
Now OP correctly uses published workfile for rendering via ```Arnold``` render on DL.

image

BigRoy:
Thanks for [the screenshot](https://github.com/ynput/OpenPype/pull/5520#pullrequestreview-1598114714) @LiborBatek 

Can we get a separate PR to solve the issue that some of the details in that report log aren't artist-facing logs?

> data: {'asset':... etc}

This should at most be a debug log, but I'd even argue that this should not be logged at all since it looks like it's just logging the full instance's data?

> deadline_url:blablabla

This should be a debug log. It's not information required by the artist.

> pools: ['none', 'pool_a', 'pool_b', 'pool_tile']

This should be a debug log.

> No representations at instance : `maxrenderMain`

If important for the artist, please describe what they should be doing with that information in the log. Otherwise, it's a debug log.

> Submitting Deadline Job ...

I feel like this is a debug log at most. If we want to actually provide some useful artist facing information about the submission I'd expect a log like: "Submitted Deadline Job: {deadline_job_name}" so that it actually provides some information they might want to use (e.g. to look it up in Deadline). For now I'd just say, make it a debug log.

PR #5523, opened 2023-08-28
Houdini: Improve VDB Selection

mustafa-zarkash:
## Changelog Description
Improves VDB selection 
if selection is `SopNode`: 
  return the selected sop node
if selection is `ObjNode`: 
  get the output node with the minimum 'outputidx' or the node with display flag

## Additional info
It's the same as pointcache (ABC) selection

## Testing notes:
Try create an instance while selecting:
1. empty `ObjNode` 
2. `ObjNode` with many sop nodes inside it
3. `SopNode`

### DIFF
--- a/openpype/hosts/houdini/plugins/create/create_vbd_cache.py
+++ b/openpype/hosts/houdini/plugins/create/create_vbd_cache.py
@@ -33,18 +33,7 @@ class CreateVDBCache(plugin.HoudiniCreator):
         }
 
         if self.selected_nodes:
-            sop_path = self.get_sop_node_path(self.selected_nodes[0])
-
-            if sop_path:
-                parms["soppath"] = sop_path
-            else:
-                self.log.debug(
-                    "Selection isn't valid. 'SOP Path' in ROP will be empty."
-                )
-        else:
-            self.log.debug(
-                "No Selection. 'SOP Path' in ROP will be empty."
-            )
+            parms["soppath"] = self.get_sop_node_path(self.selected_nodes[0])
 
         instance_node.setParms(parms)
 
@@ -72,10 +61,9 @@ class CreateVDBCache(plugin.HoudiniCreator):
         # Allow object level paths to Geometry nodes (e.g. /obj/geo1)
         # but do not allow other object level nodes types like cameras, etc.
         elif isinstance(selected_node, hou.ObjNode) and \
-                selected_node.type().name() in ["geo"]:
+                selected_node.type().name() == "geo":
 
-            # get the output node with the minimum
-            # 'outputidx' or the node with display flag
+            # Try to find output node.
             sop_node = self.get_obj_output(selected_node)
             if sop_node:
                 self.log.debug(
@@ -84,8 +72,19 @@ class CreateVDBCache(plugin.HoudiniCreator):
                 )
                 return sop_node.path()
 
+        self.log.debug(
+            "Selection isn't valid. 'SOP Path' in ROP will be empty."
+        )
+        return ""
+
     def get_obj_output(self, obj_node):
-        """Find output node with the smallest 'outputidx'."""
+        """Try to find output node.
+
+        If any output nodes are present, return the output node with
+          the minimum 'outputidx'
+        If no output nodes are present, return the node with display flag
+        If no nodes are present at all, return None
+        """
 
         outputs = obj_node.subnetOutputs()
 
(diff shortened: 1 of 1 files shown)


openpype/hosts/houdini/plugins/create/create_vbd_cache.py
```diff
@@ -33,7 +33,18 @@ def create(self, subset_name, instance_data, pre_create_data):
...
+            if sop_path:
+                parms["soppath"] = sop_path
+            else:
+                self.log.debug(
+                    "Selection isn't valid. 'SOP Path' in ROP will be empty."
+                )
+        else:
+            self.log.debug(
+                "No Selection. 'SOP Path' in ROP will be empty."
+            )
```
BigRoy, line 47: Note that this will ALSO be logged when the user had `Use selection` disabled on the Creator. Right? Which might seem a tad confusing?

openpype/hosts/houdini/plugins/create/create_vbd_cache.py
```diff
@@ -42,3 +53,53 @@ def get_network_categories(self):
...
+            self.log.debug(
+                "Valid SopNode selection, 'SOP Path' in ROP will"
+                " be set to '%s'.", selected_node.path()
+            )
+            return selected_node.path()
+
+        # Allow object level paths to Geometry nodes (e.g. /obj/geo1)
+        # but do not allow other object level nodes types like cameras, etc.
+        elif isinstance(selected_node, hou.ObjNode) and \
+                selected_node.type().name() in ["geo"]:
```
BigRoy, line 75: If we only allow geo and no other types then this is the more sensible check:
```suggestion
        elif isinstance(selected_node, hou.ObjNode) and \
                selected_node.type().name() == "geo":
```
No need for the list.

openpype/hosts/houdini/plugins/create/create_vbd_cache.py
```diff
@@ -42,3 +53,53 @@ def get_network_categories(self):
...
+
+            # get the output node with the minimum
+            # 'outputidx' or the node with display flag
+            sop_node = self.get_obj_output(selected_node)
+            if sop_node:
+                self.log.debug(
+                    "Valid ObjNode selection, 'SOP Path' in ROP will "
+                    "be set to the child path '%s'.", sop_node.path()
+                )
+                return sop_node.path()
```
BigRoy, line 85: There's no message in the case there's no valid `sop_node` here?

openpype/hosts/houdini/plugins/create/create_vbd_cache.py
```diff
@@ -42,3 +53,53 @@ def get_network_categories(self):
...
+
+    def get_obj_output(self, obj_node):
+        """Find output node with the smallest 'outputidx'."""
```
BigRoy, line 88: The docstring isn't entirely true - because it doesn't necessarily find an output node.

If there is no output node, then it takes the node with a `Display flag` set as per `SopNode.subnetOutputs()` documentation:

> Certain networks, such as SOPs, have special Output nodes to override the canonical Display or Render nodes and allow multiple outputs. If any output nodes are present, this returns a list of those nodes. If no output nodes are present, either the display or render node is returned depending on if the target of cooking is the display or an output driver. If the node has no children, an empty list is returned.

I think it's a good thing that node is considered - but just wanted to remark that the docstring isn't entirely accurate.

openpype/hosts/houdini/plugins/create/create_vbd_cache.py
```diff
@@ -42,3 +53,53 @@ def get_network_categories(self):
...
+        # if obj_node is empty
+        if not outputs:
+            return
+
+        # if obj_node has one output child whether its
+        # sop output node or a node with the render flag
+        elif len(outputs) == 1:
+            return outputs[0]
+
+        # if there are more than one, then it have multiple ouput nodes
```
BigRoy, line 101:
```suggestion
        # if there are more than one, then it has multiple output nodes
```

openpype/hosts/houdini/plugins/create/create_vbd_cache.py
mustafa-zarkash: my bad, I wrote that message in `process`, I will move it to the right place.

moonyuet (approved):
It looks good in all three scarnerios.
image
image
image

BigRoy (approved):
Didn't test the code but the code looks good.

PR #5525, opened 2023-08-28
Maya: Remove Validate Instance Attributes plug-in

BigRoy:
## Changelog Description

Remove Validate Instance Attributes plug-in.

## Additional info

Maya has adopted the new publisher. Since the new publisher always updates and imprints the publish instances with the latest Attribute Definitions of the Creator it makes this legacy validator redundant. Currently legacy instances do not work through the new publisher (they need to be converted to work correctly) and thus even for legacy workfiles this validator is redundant.

For new style instances the `objset` key isn't set in `instance.data` and thus each instance would always redundantly log:
```python
"Skipping {} because no objectset found.".format(instance)
```

## Testing notes:

1. As far as I know there isn't any use case currently to still test this validator.


tokejepsen (approved):
Nice, lets get rid of it.

antirotor (approved):

PR #5526, opened 2023-08-28
Maya: Refactor/tweak Validate Instance In same Context plug-in

BigRoy:
## Changelog Description

- Chore/Refactor: Re-use existing select invalid and repair actions
- Enhancement: provide more elaborate PublishValidationError report
- Bugfix: fix "optional" support by using `OptionalPyblishPluginMixin` base class.

## Testing notes:

1. Create an instance that publishes into a different asset than your current asset.
2. Publish (make sure the validator catches it, the report is nice and clear.
3. Make sure the "optional" checkbox works for the validator.


tokejepsen (approved):
Tested in Maya 2023 successfully and code looks good.

Issue #5516, opened 2023-08-28
Houdini: Reset FPS


mustafa-zarkash:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

_Similar to Maya_, Make `Set Frame Range` resets FPS 

![image](https://github.com/ynput/OpenPype/assets/20871534/561d2779-40f8-40c5-ada1-22946e32b2e7)

### How would you imagine the implementation of the feature?

add `hou.setFps` to this function:

https://github.com/ynput/OpenPype/blob/437f483bc48c8f8cdb38b181d9a1c6ec305388aa/openpype/hosts/houdini/api/lib.py#L474

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

BigRoy:
Sounds good to me. Just make sure to also update [these lines of code](https://github.com/ynput/OpenPype/blob/437f483bc48c8f8cdb38b181d9a1c6ec305388aa/openpype/hosts/houdini/api/pipeline.py#L388-L393) to avoid it doing so twice.

mustafa-zarkash:
I've found out that there's a `before_save` callback that validates FPS! 
So, we may not need `reset_framerange()` to deal with that! 
What do you think ? 

https://github.com/ynput/OpenPype/blob/fd64e261df0371b16395aa5238fee1eb6f0fd705/openpype/hosts/houdini/api/pipeline.py#L294-L295

PR #5528, opened 2023-08-29
Houdini: Reset FPS when clicking Set Frame Range

mustafa-zarkash:
## Changelog Description
*Similar to Maya,* Make `Set Frame Range` resets FPS, issue https://github.com/ynput/OpenPype/issues/5516

## Additional Info
There are couple of callbacks that validates FPS (e.g.  `before_save`)
> If you think This PR is not necessary, I can close it! 

image

## Testing notes:
1. open new houdini session 
2. try setting fps to different values
3. click `Set Frame Range`

### DIFF
--- a/openpype/hosts/houdini/api/lib.py
+++ b/openpype/hosts/houdini/api/lib.py
@@ -474,6 +474,11 @@ def maintained_selection():
 def reset_framerange():
     """Set frame range to current asset"""
 
+    # Set new scene fps
+    fps = get_asset_fps()
+    print("Setting scene FPS to {}".format(int(fps)))
+    set_scene_fps(fps)
+
     project_name = get_current_project_name()
     asset_name = get_current_asset_name()
     # Get the asset ID from the database for the asset of current context
--- a/openpype/hosts/houdini/api/lib.py
+++ b/openpype/hosts/houdini/api/lib.py
@@ -27,7 +27,7 @@ def get_asset_fps(asset_doc=None):
 
     if asset_doc is None:
         asset_doc = get_current_project_asset(fields=["data.fps"])
-    return asset_doc["data"].get("fps")
+    return asset_doc["data"]["fps"]
 
 
 def set_id(node, unique_id, overwrite=False):
--- a/openpype/hosts/houdini/api/lib.py
+++ b/openpype/hosts/houdini/api/lib.py
@@ -22,9 +22,12 @@ log = logging.getLogger(__name__)
 JSON_PREFIX = "JSON:::"
 
 
-def get_asset_fps():
+def get_asset_fps(asset_doc=None):
     """Return current asset fps."""
-    return get_current_project_asset()["data"].get("fps")
+
+    if asset_doc is None:
+        asset_doc = get_current_project_asset(fields=["data.fps"])
+    return asset_doc["data"].get("fps")
 
 
 def set_id(node, unique_id, overwrite=False):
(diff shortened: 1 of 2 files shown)


openpype/hosts/houdini/api/lib.py
```diff
@@ -474,6 +474,11 @@ def maintained_selection():
 def reset_framerange():
     """Set frame range to current asset"""
 
+    # Set new scene fps
+    fps = get_asset_fps()
+    print("Setting scene FPS to {}".format(int(fps)))
+    set_scene_fps(fps)
```
BigRoy, line 480: Since the asset document is retrieved below you might be best of allowing `get_asset_fps` to also take the asset document as argument and use it slightly down below to make sure you don't redundantly query the database twice for the asset's document.

While you're at it - it might be worth to change the `get_asset_fps` function to also filter the database query to only `data.fps` field. Something like this:

```python
def get_asset_fps(asset_doc=None):
    if asset_doc is None:
        asset_doc = get_current_project_asset(fields=["data.fps"])
    return asset_doc["data"]["fps"]
```

openpype/hosts/houdini/api/lib.py
```diff
@@ -22,9 +22,12 @@
...
 
 
-def get_asset_fps():
+def get_asset_fps(asset_doc=None):
     """Return current asset fps."""
-    return get_current_project_asset()["data"].get("fps")
+
+    if asset_doc is None:
+        asset_doc = get_current_project_asset(fields=["data.fps"])
+    return asset_doc["data"].get("fps")
```
BigRoy, line 30: I know this is the way "it was before" but the `.get("fps")` doesn't really make much sense looking at how `get_asset_fps` is used. It's used to set the FPS in `reset_frame_range` and that logic NEVER expects `None` to be the fps value.

As such, should we maybe just start 'believing' `fps` to be available on the asset so we can at least rely on this `get_asset_fps` function to return some data from the asset and error if otherwise?

```suggestion
    return asset_doc["data"]["fps"]
```

mustafa-zarkash:
I want to point out to a weird result. 
`hou.setFrame` is ignored when setting these values as follows, then running set frame script. 
current frame keeps at `1043` 

is that a problem in houdini's API ? 

image

openpype/hosts/houdini/api/lib.py
```diff
@@ -493,8 +501,12 @@ def reset_framerange():
     frame_start -= int(handle_start)
     frame_end += int(handle_end)
 
+    # Set frame range and FPS
+    print("Setting scene FPS to {}".format(int(fps)))
+    set_scene_fps(fps)
     hou.playbar.setFrameRange(frame_start, frame_end)
     hou.playbar.setPlaybackRange(frame_start, frame_end)
+    print("Setting current frame to {}".format(frame_start))
```
BigRoy, line 509: I'd prefer printing less than more unless it's something that's extremely unreliable or just needs often debugging whether it actually did a thing - I feel like "current frame" isn't either of those things? Is there a reason you added this particular print statement?

mustafa-zarkash: `hou.setFrame(frame_start)` was somehow ignored. I thought adding some delay may solve it but it didn't work, and I forgot to delete it.

moonyuet (approved):
Looks good. 
Tested in Houdini. 
image

Issue #5507, opened 2023-08-25
Houdini: Bgeo extractor crashes when rendering current frame


MustafaJafar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

`instance.data` will not have `"frameStart"` and `"frameEnd"` 
![image](https://github.com/ynput/OpenPype/assets/20871534/024c6a9a-807a-4843-b09a-af74696c8e98)

### Expected Behavior:

A quick fix to modify extractor as such
```python
# A single frame may also be rendered without start/end frame.
if "frameStart" in instance.data and "frameEnd" in instance.data:
    representation["frameStart"] = instance.data["frameStart"]
    representation["frameEnd"] = instance.data["frameEnd"]
```

A more in depth fix to 
- add a validator that add or remove `$F4` in `sopoutput` according to the value of `trange`

### Version

3.16.4

### What platform you are running OpenPype on?

Windows

### Steps To Reproduce:

1. Create a bgeo instance
2. Set  `Valid Frame Range` to `Render Current Frame` 
![image](https://github.com/ynput/OpenPype/assets/20871534/9b78cc55-687d-4e46-a8d9-dec9471c8a3e)

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

[cuID:[OP-6724](https://app.clickup.com/t/861najun9)]

BigRoy:
I've mentioned this elsewhere - but from a pipeline perspective the "Render Current Frame" isn't a very informative and safe option. Opening the workfile makes it quite unclear what was originally the intended frame be published?
Not saying it's unimaginable to support it - but even if we would from a safety perspective I'd still recommend refraining from using it.

mustafa-zarkash:
> I'd still recommend refraining from using it.
So, it's better to add a validator that make sure trange != 0 instead of adding support for single frame at the moment?

BigRoy:
> So, it's better to add a validator that makes sure trange != 0 instead of adding support for single frame at the moment?

Hard to say what the community wants. For our production use cases I'd say not support `trange == 0` would be perfectly fine. A validator could solve that yes, which could offer a repair which can set the toggle to "Render frame range" but with start and end frame on the current frame to help the artist.

mustafa-zarkash:
This comment relates to this issue. 
https://github.com/ynput/OpenPype/pull/5529#issuecomment-1697480805

Issue #5488, opened 2023-08-23
Houdini: Hard coded "start_frame" raises error in validations


MustafaJafar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

Hard coded "frameStart" value in several houdini plugins 
```python
frame = instance.data.get("frameStart", 0)
```
It causes an error when imported asset doesn't have a 0000 frame. 

> [!NOTE]
> This only happens during validation (alembic, fbx static mesh, vdb)
> when the associated ROP trange is set to render current frame 
> one solution is to make a validator that 
> - changes `render current frame` to `render frame range`
> - set `f1` and `f2` of the ROP to the current frame value.

![image](https://github.com/ynput/OpenPype/assets/20871534/e6926425-24cb-4f43-b186-afd8913f8f70)

### Expected Behavior:

Validators should validate what the user see in the view port whether the current frame or the start frame of the current frame range. 
Depending on the situation: 

### Version

3.16.4

### What platform you are running OpenPype on?

Windows

### Steps To Reproduce:

1. import an asset that doesn't exists at frame 0 (e.g. a BEGO cache that doesn't have 0000 frame) 
2. publish

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

### Additional context:
Suggested Solution
```python
        # Get frame
        frame = hou.intFrame()
        trange = rop_node.parm("trange").eval()
        if trange:
            frame = int(hou.playbar.frameRange()[0])

        frame = instance.data.get("frameStart", frame)

        # Get Geo at that frame
        geo = output_node.geometryAtFrame(frame)
        if not geo:
            cls.log.warning("No geometry found,"
                            " skipping check..")
            return
```
Validators associated with this issue

https://github.com/ynput/OpenPype/blob/4b00d2a9b96f0e877baae03371e4d8fd785ded05/openpype/hosts/houdini/plugins/publish/validate_abc_primitive_to_detail.py#L76-L86

https://github.com/ynput/OpenPype/blob/4b00d2a9b96f0e877baae03371e4d8fd785ded05/openpype/hosts/houdini/plugins/publish/validate_alembic_input_node.py#L53

https://github.com/ynput/OpenPype/blob/4b00d2a9b96f0e877baae03371e4d8fd785ded05/openpype/hosts/houdini/plugins/publish/validate_primitive_hierarchy_paths.py#L83

[...]

PR #5529, opened 2023-08-29
Houdini: make validations use currentFrame if frameStart doesn't exist

mustafa-zarkash:
## Changelog Description
Hard coded "frameStart" value can cause a bug when `"frameStart"` is missing when user render current frame (`trange == 0`)

## Additional Info 1
Hard coded "frameStart" leaded to this  issue https://github.com/ynput/OpenPype/issues/5488
when validator was checking geo at frame 0 and imported file hadn't any geo at frame 0 but it workd fine in the current frame

## Additional Info 2
Some Product-types crash when rendering the current frame range like BGEO. 
which is another issue https://github.com/ynput/OpenPype/issues/5507
image

## Testing notes:
1. create and instance  (e.g abc)
2. publish

### DIFF
--- a/openpype/hosts/houdini/plugins/publish/validate_abc_primitive_to_detail.py
+++ b/openpype/hosts/houdini/plugins/publish/validate_abc_primitive_to_detail.py
@@ -82,7 +82,7 @@ class ValidateAbcPrimitiveToDetail(pyblish.api.InstancePlugin):
             return
 
         # Check if the primitive attribute exists
-        frame = instance.data.get("frameStart", 0)
+        frame = instance.data.get("frameStart", hou.intFrame())
         geo = output_node.geometryAtFrame(frame)
 
         # If there are no primitives on the start frame then it might be
--- a/openpype/hosts/houdini/plugins/publish/validate_alembic_input_node.py
+++ b/openpype/hosts/houdini/plugins/publish/validate_alembic_input_node.py
@@ -50,7 +50,7 @@ class ValidateAlembicInputNode(pyblish.api.InstancePlugin):
             cls.log.warning("No geometry output node found, skipping check..")
             return
 
-        frame = instance.data.get("frameStart", 0)
+        frame = instance.data.get("frameStart", hou.intFrame())
         geo = output_node.geometryAtFrame(frame)
 
         invalid = False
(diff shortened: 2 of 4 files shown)


BigRoy:
Code seems fine. 

Note that it's always worrying if the user is on a different frame and the "validation" suddenly doesn't work or does work so I'm still hesitant making it rely on the 'current frame' of the user instead of something static to ensure whatever it does is completely reproducable from the instance alone. We'd like to avoid situations where we need to tell the user "go to another current frame and then publish; then it'll work" for whatever reason.

mustafa-zarkash:
>  We'd like to avoid situations where we need to tell the user "go to another current frame and then publish; then it'll work" for whatever reason.

It makes sense! 
I recall your previous explanations and I'm more aware now! Thanks! 

Do you have any recommendations other than falling to 0 ? if not I will close this PR.

BigRoy:
Only thing I can come up with is:

1. "start of current timeline" or 
2. "start frame of current asset"

Not sure if any of those is much better.

To me it sounds more reasonable somehow to enforce NOT using a 'Render current frame' that just happens to start publishing a different frame if the user happened to have gone to the next frame in the timeline and that whenever that is used a validator catches it and says: _"Your publish range is set to "Current frame". It's required to be more explicit about what frame you'd like to extract. bla-bla-bla."_ - which then offers a repair option to just change the ROP node to Render Range with start and end set to whatever the current frame was. So if I was on frame 1004, it'd just set 1004-1004 explicitly.

From that moment onwards it'd also mean that future publishes actually end up using that frame explicitly and we can then still rely more accurately on "frameStart" since it'd be explicitly set on the instance from that point in time. If that's enforced - then I suppose for any case where frameStart would still **not** be defined we can indeed rely on `hou.intFrame()` making the rest of this PR fine from that perspective.

However, I welcome any opinions from others since I'm solely coming at it from a pipeline-perspective and having a 'reliable reproducable scene' instead of what is maybe considered artist friendly.

@fabiaserra @mkolar @antirotor @krishna8008 any thoughts?

mustafa-zarkash:
> a validator catches it and says: "Your publish range is set to "Current frame". It's required to be more explicit about what frame you'd like to extract. bla-bla-bla."

This idea sounds reasonable, and it can be disabled from settings

PR #5530, opened 2023-08-29
TrayPublisher: set default frame values to sequential data

jakubjezek001:
## Changelog Description
We are inheriting  default frame handles and fps data either from project or setting them to 0. This is just for case a production will decide not to injest the sequential representations with asset based metadata.

## Testing notes:
1. Open TrayPublisher on a context
2. drop some imagesequence data to Plate family.
3. Disable instance attribute for `Collect Sequence Frame Data`
4. Publish and see that all had been published as expected.
5. look into Loader and find the published product and see that the frame-range defined on the version is the same as the original frame-range and is not the same as it is defined on asset parent.

### DIFF
--- a/openpype/plugins/publish/collect_sequence_frame_data.py
+++ b/openpype/plugins/publish/collect_sequence_frame_data.py
@@ -50,4 +50,7 @@ class CollectSequenceFrameData(pyblish.api.InstancePlugin):
             return {
                 "frameStart": repres_frames[0],
                 "frameEnd": repres_frames[-1],
+                "handleStart": 0,
+                "handleEnd": 0,
+                "fps": instance.context.data["projectEntity"]["data"]["fps"]
             }
--- a/openpype/plugins/publish/collect_sequence_frame_data.py
+++ b/openpype/plugins/publish/collect_sequence_frame_data.py
@@ -52,5 +52,5 @@ class CollectSequenceFrameData(pyblish.api.InstancePlugin):
                 "frameEnd": repres_frames[-1],
                 "handleStart": 0,
                 "handleEnd": 0,
-                "fps": instance.context.data["projectEntity"]["data"]["fps"]
+                "fps": instance.context.data["assetEntity"]["data"]["fps"]
             }
--- a/openpype/hosts/traypublisher/plugins/publish/collect_missing_frame_range_asset_entity.py
+++ b/openpype/hosts/traypublisher/plugins/publish/collect_missing_frame_range_asset_entity.py
@@ -2,16 +2,18 @@ import pyblish.api
 from openpype.pipeline import OptionalPyblishPluginMixin
 
 
-class CollectFrameDataFromAssetEntity(pyblish.api.InstancePlugin,
-                                      OptionalPyblishPluginMixin):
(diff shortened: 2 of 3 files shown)


openpype/plugins/publish/collect_sequence_frame_data.py
```diff
@@ -50,4 +50,7 @@ def get_frame_data_from_repre_sequence(self, instance):
             return {
                 "frameStart": repres_frames[0],
                 "frameEnd": repres_frames[-1],
+                "handleStart": 0,
+                "handleEnd": 0,
+                "fps": instance.context.data["projectEntity"]["data"]["fps"]
```
iLLiCiTiT, line 55: Shouldn't that be taken from `"assetEntity"`?

jakubjezek001: Not in this plugin. We are having dedicated plugin which is optional `CollectFrameDataFromAssetEntity`.

BigRoy: Note that due to this PR that particular plug-in will not do anything since it'll only handle missing keys in `instance.data` and it runs after this collector, right?

iLLiCiTiT: > Not in this plugin. We are having dedicated plugin which is optional CollectFrameDataFromAssetEntity.

As BigRoy mentioned, if `"fps"` is filled here, `CollectFrameDataFromAssetEntity` won't do anything. At the same time `CollectFrameDataFromAssetEntity` can't override the value if is enabled, because the values could be filled from dedicated host collector. So it should be filled from `"assetEntity"` or not filled.

jakubjezek001: If it is not added here and the optional plugin `CollectFrameDataFromAssetEntity` is disabled then it is crashing on ExctractReview, since it is missing `fps` at the instance. 

>  values could be filled from dedicated host collector
We are solving only TrayPublisher here since it is only host added into hosts filter.

BigRoy: > If it is not added here and the optional plugin CollectFrameDataFromAssetEntity is disabled then it is crashing on ExctractReview, since it is missing fps at the instance.

I understand it might fail for that scenario and thus there's a bug we want to fix but this PR will still be breaking `CollectFrameDataFromAssetEntity` even if it were enabled. Yes, it'll allow to publish but it'll be publishing the wrong data. Right?

jakubjezek001: Yeah, I see the issue here finally. I had changed it it

BigRoy:
I personally don't understand why [this conversation](https://github.com/ynput/OpenPype/pull/5530#discussion_r1308528677) was marked resolved multiple times with no code changes. This PR has completely broken the behavior for `CollectFrameDataFromAssetEntity`. Sure, you relabeled it "Collect missing data" but it's basically rendered useless as now it should **never** be the case that it's missing. If the decision is that _that_ collecting from asset entity is thus redundant the collector could just be removed. It kinds of feels like 'patching' something whilst allowing other stuff to break.

PR #5533, opened 2023-08-29
Tests: fix unit tests

kalisp:
## Changelog Description
Fixed failing tests.
Updated Unreal's validator to match removed general one which had a couple of issues fixed.

### DIFF
--- a/openpype/hosts/unreal/plugins/publish/validate_sequence_frames.py
+++ b/openpype/hosts/unreal/plugins/publish/validate_sequence_frames.py
@@ -1,4 +1,6 @@
 import clique
+import os
+import re
 
 import pyblish.api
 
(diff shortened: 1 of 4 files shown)


openpype/hosts/unreal/plugins/publish/validate_sequence_frames.py
```diff
@@ -21,7 +23,19 @@ def process(self, instance):
...
+                continue
+
+            ext = repr.get("ext")
+            if not ext:
+                _, ext = os.path.splitext(repr_files[0])
+            elif not ext.startswith("."):
+                ext = ".{}".format(ext)
+            pattern = r"\D?(?P<index>(?P<padding>0*)\d+){}$".format(
+                re.escape(ext))
+            patterns = [pattern]
```
BigRoy, line 37: Oh boy - this looks more complicated than I'd appreciate in this function. :) Should we separate this out into a `get_collection(files)` function which maybe also comments and explains why certain workarounds are needed. (Actually didn't we actually build that some time ago AND have tests implemented for that?)

Note that `clique.assemble` is used in many more places - just patching this particular one with this relatively unique way of detecting the collection seems off since we can't actually ensure that e.g. the integrator will detect the collection the same way?

Isn't it just time to have an actual "collect" or "assemble" call in OpenPype that we re-use throughout that actually contains all these 'fixes'?

simonebarbieri (approved):
LGTM

openpype/hosts/unreal/plugins/publish/validate_sequence_frames.py
kalisp: This is from that general plugin and that we have tests for.

BigRoy: Doesn't the integrator does its own `clique.assemble` calls too? Wouldn't they potentially differ from this logic?

64qam (approved):

PR #5534, opened 2023-08-29
General: Fix Validate Publish Dir Validator

kalisp:
## Changelog Description
Nonsensical "family" key was used instead of real value (as 'render' etc.) which would result in wrong translation of intermediate family names.

Updated docstring.

### DIFF
--- a/openpype/plugins/publish/validate_publish_dir.py
+++ b/openpype/plugins/publish/validate_publish_dir.py
@@ -60,7 +60,7 @@ class ValidatePublishDir(pyblish.api.InstancePlugin):
         host_name = instance.context.data["hostName"]
         anatomy_data = instance.data["anatomyData"]
         family = anatomy_data["family"]
-        family = self.family_mapping.get("family") or family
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
Docstring seems clearer to me, so good improvement.

Please make sure the PR title + description is more descriptive of the change - they are used to generate changelogs for releases of OpenPype. For example **Fix Validate Publish Dir Validator**. The changelog description is also used as an expandable section in the changelogs containing more information, so also those should be worded in a way that it's understandable what the change _does_ instead of just mentioning "x was broken". 

Off-topic question: How does a 'source' template publish work with Site Sync? Does it just happen to sync to the other user relative to the project root even though the file could live 'anywhere in the project'?

kalisp:
Not sure I follow "sync to the other user".

`source` is used when studio has already source data in expected data structure (likely via some external script) and they would like to push it into OP database without regular copy into `publish` folder (use case is massive folder from camera capture system) as this copy might be expensive.

PR #5535, opened 2023-08-29
Chore: PowerShell script for docker build

iLLiCiTiT:
## Changelog Description
Added PowerShell script to run docker build.

## Testing notes:
1. Run `./tools/docker_build` (with possible variant `centos7` or `debian`)
2. Build should be in `./build` once finished

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,8 @@ If all is successful, you'll find built OpenPype in `./build/` folder.
 
 Docker build can be also started from Windows machine, just use `./tools/docker_build.ps1` instead of shell script.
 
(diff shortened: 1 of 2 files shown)


kalisp:
Shouldn't this be added to readme? What is actually a use case for this in OP?

iLLiCiTiT:
> Shouldn't this be added to readme

Added, if you think there should be more, don't be shy and add more :)

> What is actually a use case for this in OP?

To be able to create linux builds on windows...

kalisp (approved):
Works even for `centos7` argument.

PR #5541, opened 2023-08-30
Colorspace: ociolook file product type workflow

jakubjezek001:
## Changelog Description
Traypublisher support for publishing of colorspace look files (ociolook) which are json files holding any LUT files. This new product is available for loading in Nuke host at the moment. 

Added colorspace selector to publisher attribute with better labeling. We are supporting also Roles and Alias (only v2 configs).
image
image

## TODO:
- [x] Enhanced colorspace data from config
- [x] fix families in Collector for explicit colorspace
- [x] enhance Collector explicit colorspace so new labeled items can be show. Aliases from ocio v2 also included.
- [x] Creator for colorspace look
- [x] Collector for colorspace look
- [x] Validator for colorspace look
- [x] Extractor for colorspace look
- [x] Nuke loader for colorspace look

### Traypubisher Publishing
0. make sure your project Colormanagement is activated in global `project_settings/global/imageio/activate_global_color_management`
1. open traypublisher at some context and set product type to `Colorspace Look`
2. Drop any of defined lut formats. Here is some example packed in zip.
[Canon-Log3_to_linear.zip](https://github.com/ynput/OpenPype/files/12882264/Canon-Log3_to_linear.zip)
3. set working/Input/output space at created instance. Publish
4. look there is a Validator failing which is navigating you to set some missing attributes. Set those to anything. Publish again
5. all should be published as expected

### Nuke Loading
0. make sure colorspace management is activated on nuke `project_settings/nuke/imageio/activate_host_color_management`
1. also make sure colorspace on workfile and nodes is correctly defined to the globaly selected aces 1.2 
2. open workfile and open loader 
3. load `ocioLook` file


openpype/hosts/traypublisher/plugins/publish/collect_explicit_colorspace.py
```diff
@@ -46,12 +50,31 @@ def apply_settings(cls, project_settings):
...
+                    continue
+                if not value_.get("aliases"):
+                    continue
+                for alias in value_.get("aliases"):
+                    aliases.add(alias)
+
+            colorspaces = {
+                name for name, data_ in config_items.items()
+                if data_.get("type") == "colorspace"
+            }
```
BigRoy, line 68: Out of curiosity - why are you suffixing with `_` underscores after `value_` and `data_`. Feels a bit redundant?

openpype/hosts/traypublisher/plugins/publish/validate_colorspace_look.py
```diff
@@ -0,0 +1,70 @@
...
+        not_set_keys = []
+        for key in check_keys:
+            if ociolook_item[key]:
+                # key is set and it is correct
+                continue
+
+            def_label = next(
+                (d_.label for d_ in creator_defs if key == d_.key),
+                None
+            )
```
BigRoy, line 51: If you would do this once outside the for loop: 
`creator_defs_by_key = {def.key: def.label for def in creator_defs}`

Then inside the for loop you can just do:
```python
def_label = creator_defs_by_key.get(key)
```
You can then also just log `def_attrs` below by just logging `creator_defs_by_key` instead.


Although I'm not entirely sure what you're trying to achieve here.

openpype/scripts/ocio_wrapper.py
```diff
@@ -96,11 +96,41 @@ def _get_colorspace_data(config_path):
...
+    if roles:
+        colorspace_data.update({
+            role[0]: {
+                "type": "role",
+                "colorspace": role[1]
+            }
+            for role in roles
+        })
+
+    return colorspace_data
```
BigRoy, line 132: Isn't it technically possible for an OCIO config to define names for looks, roles and colorspaces that clash. E.g. the OCIO config can technically hold a role `srgb` and a look `srgb` and a colorspace `srgb`. By flattening them into a dict here you're essentially losing data in that case. (Plus you're making it harder later to just query, 'give me all roles' from that data).

Why note return:
```{
  "colorspaces": {},
  "roles": {},
  "looks": {}
}
```
Which can always be expanded later with even more keys if somehow OCIO starts adding features.

openpype/hosts/traypublisher/plugins/publish/extract_colorspace_look.py
```diff
@@ -0,0 +1,43 @@
...
+
+        # create lut representation data
+        ociolook_repre = {
+            "name": "ocioLookFile",
+            "ext": "json",
+            "files": ociolook_file_name,
+            "stagingDir": staging_dir,
+            "tags": []
+        }
+        instance.data["representations"].append(ociolook_repre)
```
BigRoy, line 43: What is ensuring the `representations` data key already exists?

openpype/scripts/ocio_wrapper.py
```diff
@@ -96,11 +96,41 @@ def _get_colorspace_data(config_path):
...
+
+    # add roles
+    roles = config.getRoles()
+    if roles:
+        colorspace_data.update({
+            role[0]: {
+                "type": "role",
+                "colorspace": role[1]
+            }
+            for role in roles
```
BigRoy, line 129: Does `roles` provide more than just index zero and one or is it a two-tuple? If a two-tuple it might be more readable as:
```suggestion
            role: {
                "type": "role",
                "colorspace": colorspace
            }
            for (role, colorspace) in roles
```

openpype/hosts/traypublisher/plugins/publish/validate_colorspace_look.py
```diff
@@ -0,0 +1,70 @@
...
+                # raise since key is not recognized by creator defs
+                raise KeyError(
+                    f"Colorspace look attribute '{key}' is not "
+                    f"recognized by creator attributes: {def_attrs}"
+                )
+            not_set_keys.append(def_label)
+
+        if not_set_keys:
+            message = (
+                f"Colorspace look attributes are not set: "
```
BigRoy, line 63:
```suggestion
                "Colorspace look attributes are not set: "
```

openpype/hosts/traypublisher/plugins/publish/collect_explicit_colorspace.py
jakubjezek001: issue is the Flake8 is flagging it as not recommended name for variable. Underscore is at right so it should not confuse you with non-public and this is more  matter of preference. For example if one letter variable is awful  to use because of search and replace could pick anything, so I am also adding underscore for those `f_` .

BigRoy: All these code styling tools just forcing us into evil 👿 - anyway, thanks for the explanation. I just love code as brief and simple as possible and all these `_` are just not my taste I suppose. Anyway, I'm fine with it if nobody else cares.

openpype/hosts/traypublisher/plugins/create/create_colorspace_look.py
```diff
@@ -0,0 +1,185 @@
...
+
+        files = repr_file.get("filenames")
+        if not files:
+            # this should never happen
+            raise CreatorError("Missing files from representation")
+
+        asset_doc = get_asset_by_name(
+            self.project_name, instance_data["asset"])
+
+        subset_name = self._get_subset(
```
iLLiCiTiT, line 62: You should always use `self.get_subset_name`.

openpype/hosts/traypublisher/plugins/publish/validate_colorspace_look.py
jakubjezek001: done

openpype/hosts/traypublisher/plugins/publish/extract_colorspace_look.py
jakubjezek001: here https://github.com/ynput/OpenPype/pull/5541/files#diff-6d33590917e9654fbee62d3961bd6d38d81869b19b060a3d81450201548793bcR33

openpype/scripts/ocio_wrapper.py
jakubjezek001: Yes lets do it this way.

openpype/scripts/ocio_wrapper.py
```diff
@@ -106,11 +106,36 @@ def _get_colorspace_data(config_path):
...
+            for look in looks
+        }
+
+    # add roles
+    roles = config.getRoles()
+    if roles:
+        colorspace_data["role"] = {
+            role: {"colorspace": colorspace}
+            for (role, colorspace) in roles
+        }
```
BigRoy, line 135:
```suggestion
    looks = config.getLooks()
    if looks:
        colorspace_data["looks"] = {
            look.getName(): {"process_space": look.getProcessSpace()}
            for look in looks
        }

    # add roles
    roles = config.getRoles()
    if roles:
        colorspace_data["roles"] = {
            role: {"colorspace": colorspace}
            for (role, colorspace) in roles
        }
```
Seems weird that these keys are singular even though they contain multiple entries?

BigRoy: Also, what's holding us from **always** adding the `looks` and `roles` key in the returned data - even when e.g. empty? At least then our ouput dict structure is more consistent.

openpype/pipeline/colorspace.py
```diff
@@ -568,46 +617,61 @@ def get_labeled_colorspaces(
...
     if roles and include_roles:
+        roles = sorted(roles, key=lambda x: x[0])
         labeled_colorspaces.extend(roles)
 
-    # add colorspace after roles so it is first in menu
-    labeled_colorspaces.extend((
-        (name, f"[colorspace] {name}") for name in colorspaces
-    ))
+    # add colorspaces as second so it is not first in menu
+    colorspaces = sorted(colorspaces, key=lambda x: x[0])
```
BigRoy, line 662: Since the values are tuples can't we just sort on both items. It'd only check the second item if the first one matched anyway? Which means we can just do `sorted(colorspaces)` etc. Right?

openpype/pipeline/colorspace.py
```diff
@@ -533,29 +533,78 @@ def get_ocio_config_colorspaces(config_path):
...
-
 ):
     """Get all colorspace data with labels
 
     Wrapper function for aggregating all names and its families.
     Families can be used for building menu and submenus in gui.
 
     Args:
-        config_path (str): path leading to config.ocio file
+        config_items (dict[str,dict]): colorspace data
```
BigRoy, line 600: It's a bit unclear how I generate the `config_items` entry I need here from an OCIO file. Which function does that? Is it `get_ocio_config_colorspaces(filepath)`?

openpype/pipeline/colorspace.py
```diff
@@ -533,29 +533,78 @@ def get_ocio_config_colorspaces(config_path):
...
+
+    Args:
+        colorspace_item (str): colorspace and family in couple
+        config_items (dict[str,dict]): colorspace data
+
+    Returns:
+        dict: colorspace data
+    """
+    # split string with `::` separator and set first as key and second as value
+    item_type, item_name = colorspace_enum_item.split("::")
```
BigRoy, line 550:
```suggestion
    item_type, item_name = colorspace_enum_item.split("::", 1)
```

openpype/pipeline/colorspace.py
```diff
@@ -568,46 +617,61 @@ def get_labeled_colorspaces(
...
 
         elif items_type == "roles":
             roles.update([
-                (name, "[role] {} ({})".format(name, role_data["colorspace"]))
+                (
+                    "{}::{}".format(items_type, name),
+                    "[role] {} ({})".format(name, role_data["colorspace"])
+                )
                 for name, role_data in colorspace_items.items()
             ])
```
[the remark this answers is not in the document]
BigRoy: Also, doesn't it make sense to have the prefix be `role::` if the header is `[role]` instead of `roles::` just for consistency.

So:

- `role::` and `[role]`
- `look::` and `[look]`
- `alias::` and `[alias]`
- `colorspace::` and `[colorspace]`

The only outlier is `[view (display)]`  but also there it's likely even better to then also make that singular:

- `display_view::` and `[view (display)]

openpype/pipeline/colorspace.py
BigRoy: Also note that the sorting is Case sensitive - which I'm not sure whether a user would want that visually?

openpype/pipeline/colorspace.py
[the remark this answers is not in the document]
jakubjezek001: Since the prefix is referring to the plural key in data but label makes sense to be singular, it would be better to keep it this way.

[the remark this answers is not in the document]
BigRoy: To me it's clearer that that single entry refers to a `role::` type than it being part of the `roles::` group - but, whatever rocks your boat. I just got tripped up on them not matching from label to the data.

openpype/pipeline/colorspace.py
```diff
@@ -568,46 +617,61 @@ def get_labeled_colorspaces(
...
+        aliases = sorted(aliases, key=lambda x: x[0])
         labeled_colorspaces.extend(aliases)
 
     if looks and include_looks:
+        looks = sorted(looks, key=lambda x: x[0])
         labeled_colorspaces.extend(looks)
 
     if display_views and include_display_views:
+        display_views = sorted(display_views, key=lambda x: x[0])
         labeled_colorspaces.extend(display_views)
```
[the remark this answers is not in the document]
jakubjezek001: That is why it is optional and anyone could include them on demand.

moonyuet (approved):
Looks good.
Traypublisher can successfully publish the ocio color data and sucessfully loaded into nuke too
image

PR #5544, opened 2023-08-30
AYON: Update settings

iLLiCiTiT:
## Changelog Description
Updated settings in AYON addons and conversion of AYON settings in OpenPype.

## Additional information
Bumped only version of core and aftereffects addons. Both had model changes.


Minkiu (approved):

PR #5545, opened 2023-08-30
Deadline: files on representation cannot be single item list

kalisp:
## Changelog Description
Further logic expects that single item files will be only 'string' not 'list' (eg. repre["files"] = "abc.exr" not repre["files"] = ["abc.exr"].

This would cause an issue in ExtractReview later.

This could happen if DL rendered single frame file with different frame value.


Minkiu (approved):

iLLiCiTiT (approved):

PR #5546, opened 2023-08-30
Webpublisher: better encode list values for click

kalisp:
## Changelog Description
Targets could be a list, original implementation pushed it as a separate items, it must be added as `--targets webpulish --targets filepublish`.

`wepublish_routes` handles triggering from UI, changes in `publish_functions` handle triggering from cmd (for tests, api access).

### DIFF
--- a/openpype/hosts/webpublisher/publish_functions.py
+++ b/openpype/hosts/webpublisher/publish_functions.py
@@ -65,8 +65,7 @@ def cli_publish(project_name, batch_path, user_email, targets):
         if isinstance(targets, str):
             targets = [targets]
         for target in targets:
-            for target_item in target.split(" "):
-                pyblish.api.register_target(target_item)
+            pyblish.api.register_target(target)
 
     install_host(webpublisher_host)
 
--- a/openpype/hosts/webpublisher/publish_functions.py
+++ b/openpype/hosts/webpublisher/publish_functions.py
@@ -65,7 +65,8 @@ def cli_publish(project_name, batch_path, user_email, targets):
         if isinstance(targets, str):
             targets = [targets]
         for target in targets:
-            pyblish.api.register_target(target)
+            for target_item in target.split(" "):
+                pyblish.api.register_target(target_item)
 
     install_host(webpublisher_host)
 
--- a/openpype/hosts/webpublisher/webserver_service/webpublish_routes.py
+++ b/openpype/hosts/webpublisher/webserver_service/webpublish_routes.py
@@ -284,7 +284,7 @@ class BatchPublishEndpoint(WebpublishApiEndpoint):
                 args.append("--{}".format(key))
                 # Extend list into arguments (targets can be a list)
                 if isinstance(value, (tuple, list)):
-                    args.extend(value)
+                    args.append(" ".join(value))
                 else:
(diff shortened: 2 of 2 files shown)


64qam (approved):

64qam (approved):

64qam (approved):

openpype/hosts/webpublisher/webserver_service/webpublish_routes.py
```diff
@@ -284,7 +284,7 @@ async def post(self, request) -> Response:
                 args.append("--{}".format(key))
                 # Extend list into arguments (targets can be a list)
                 if isinstance(value, (tuple, list)):
-                    args.extend(value)
+                    args.append(" ".join(value))
```
BigRoy, line 287: Does this work for `subprocess.Popen` calls? That's not how you should be passing lists of arguments to Popen. Seems odd that this is a fix unless maybe the script it's triggering is not correctly parsing input arguments.

iLLiCiTiT: I do agree, for what usecase this is?

iLLiCiTiT:
> Targets could be a list, original implementation pushed it as a comma separated, it must be separated by space for click to understand.

It did not pass it as coma separated values, it passed it as space separated. Where did you get the comma separation?

openpype/hosts/webpublisher/webserver_service/webpublish_routes.py
iLLiCiTiT: I think the problem is that `click` is expecting each value as new argument ["target1", "target2"] should be `-t target1 -t target2` instead of `-t target1 target2`.

This should work:
```python
        for key, value in add_args.items():
            # Skip key values where value is None
            if value is None:
                continue
            arg_key = "--{}".format(key)
            if not isinstance(value, (tuple, list)):
                args.extend([arg_key, value])
                continue
                
            for item in value:
                args.extend([arg_key, item])
```

openpype/hosts/webpublisher/publish_functions.py
```diff
@@ -65,7 +65,8 @@ def cli_publish(project_name, batch_path, user_email, targets):
         if isinstance(targets, str):
             targets = [targets]
         for target in targets:
-            pyblish.api.register_target(target)
+            for target_item in target.split(" "):
+                pyblish.api.register_target(target_item)
```
iLLiCiTiT, line 69:
```suggestion
           pyblish.api.register_target(target_item)
```

kalisp: This is necessary for triggering this from command line, as it is much more probable that one will add there `--targets "filepublish webpublish"` (which will be shown when one google it) than `--targets filepublish --targets webpublish`

64qam (approved):

iLLiCiTiT (approved):

openpype/hosts/webpublisher/publish_functions.py
kalisp: Ok, this was unnecessary as correct way (which is also in click documentation is only `--targets filepublish --targets webpublish`)

PR #5547, opened 2023-08-30
Attribute Definitions: Multiselection enum def

iLLiCiTiT:
## Changelog Description
Added `multiselection` option to `EnumDef`.

## Additional info
I've just tried it and it looks like it works, need some testing.

## Testing notes:
1. Add `EnumDef` to creator and pass `multiselection=True` to init arguments.
2. Use the creator.
3. Try few possible usecases with the attribute.

### DIFF
--- a/openpype/lib/attribute_definitions.py
+++ b/openpype/lib/attribute_definitions.py
@@ -424,7 +424,10 @@ class TextDef(AbstractAttrDef):
 
 
 class EnumDef(AbstractAttrDef):
-    """Enumeration of single item from items.
+    """Enumeration of items.
+
+    Enumeration of single item from items. Or list of items if multiselection
+    is enabled.
 
     Args:
         items (Union[list[str], list[dict[str, Any]]): Items definition that


openpype/lib/attribute_definitions.py
```diff
@@ -443,7 +445,10 @@ def __init__(self, key, items, default=None, **kwargs):
...
         items = self.prepare_enum_items(items)
         item_values = [item["value"] for item in items]
-        if default not in item_values:
+        if multiselection and default is None:
+            default = []
+
+        if not multiselection and default not in item_values:
             for value in item_values:
                 default = value
                 break
```
BigRoy, line 454: Seems like this was intended to just get the first value?
```suggestion
            default = next(iter(item_values), None)
```

BigRoy, line 454: I'm pretty sure looking at this logic that there's currently no `default value` check for `defaults` if `multiselection` were enabled.
```suggestion
        if multiselection:
            if not isinstance(default, (list, tuple, set)):
                # allow single value as default
                default = [default]
            default = list(item_values.intersection(default))
        
        elif default not in item_values:
            # single selection; use first value if no valid default
            default = next(iter(item_values), None)
```
Note that this pseudocode logic assumes `item_values` is a `set` already

openpype/lib/attribute_definitions.py
```diff
@@ -452,21 +457,34 @@ def __init__(self, key, items, default=None, **kwargs):
 
         self.items = items
         self._item_values = set(item_values)
```
BigRoy, line 459:
```suggestion
        self._item_values = item_values
```
Assuming the `item_values` is changed to set already above - see other review suggestion.

openpype/lib/attribute_definitions.py
```diff
@@ -452,21 +457,34 @@ def __init__(self, key, items, default=None, **kwargs):
...
+        if not self.multiselection:
+            if value in self._item_values:
+                return value
+            return self.default
+
+        if value is None:
+            return copy.deepcopy(self.default)
+        new_value = set(value)
+        rem = new_value - self._item_values
+        return list(new_value - rem)
```
BigRoy, line 481:
```suggestion
    return list(self._item_values.intersection(value))
```

openpype/tools/attribute_defs/widgets.py
```diff
@@ -442,17 +454,40 @@ def _on_value_change(self):
...
+                    if _value is None:
+                        _value = _v
+                    elif not are_same or _value != _v:
+                        _value |= _v
+                        are_same = False
+                if _value is None:
+                    _value = []
+                else:
+                    value = list(_value)
+                multivalue = not are_same
```
BigRoy, line 478: What godly thing is this doing?

It seems like, if all values are `None` then you want: `[]` (so basically an empty set, but as a list)
If any value is not None then you want the intersections (but as a list)

So you always basically the intersection of all sets combined.
Which is basically this: (assuming each `v` in `value` is an iterable and never e.g. `None`)
```python
value = set.intersection(*list(map(set, value)))
```
Or maybe more readable:
```python
value = set.intersection(*[set(v) for v in value])
```
Or:
```python
sets = [set(v) for v in value]
value = set.intersection(*sets)
```
---
Example:
```python
a = [1, 2, 3]
b = [2, 5]
c = [1, 2, 4]
values = [a, b, c]

def intersection(*iterables):
    sets = [set(it) for it in iterables]
    return set.intersection(*sets)


print(intersection(a, b, c))
# {2}
print(intersection(*values))
# {2}
```

openpype/lib/attribute_definitions.py
```diff
@@ -443,7 +445,10 @@ def __init__(self, key, items, default=None, **kwargs):
 
         items = self.prepare_enum_items(items)
         item_values = [item["value"] for item in items]
```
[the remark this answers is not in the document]
iLLiCiTiT: It's kept as list, to use first item as `default`, `set` re-orders it.

openpype/tools/attribute_defs/widgets.py
iLLiCiTiT: How you find out if the values are same using this?

openpype/lib/attribute_definitions.py
iLLiCiTiT: If you pass `multiselection=True`, you should be able pass `default=[value]`...

openpype/tools/attribute_defs/widgets.py
BigRoy: How about:
```python
                value_sets = [set(v) for v in value]
                value = list(set.intersection(*value_sets))
                are_same = all(v == value_sets[0] for v in value_sets)
                multivalue = not are_same
```

openpype/tools/attribute_defs/widgets.py
```diff
@@ -459,23 +459,23 @@ def current_value(self):
...
         return self._input_widget.itemData(idx)
 
+    def _multiselection_multivalue_prep(self, values):
+        final = None
+        are_same = True
+        for value in values:
+            value = set(value)
+            if final is None:
+                final = value
+            elif not are_same or final != value:
```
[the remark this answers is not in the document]
iLLiCiTiT: The first check is boolean check (is instant) whereare comparison of values is slower, and because if `are_same` is already set to `False` it does not make sense to check values, it is actually faster this way.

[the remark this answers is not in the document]
BigRoy: You are completely right!

If we're optimizing anyway. Note that if:
- final is not None
- not final
- not are same

You can actually opt out early since there's no more need for doing intersections and you already know values are not the same.

Which means you could do:
```python
if final is None:
    final = value
elif not are_same and not final:
    return final, not are_same
elif not are_same or final != value:
    ...
```

openpype/tools/attribute_defs/widgets.py
```diff
@@ -442,17 +454,40 @@ def _on_value_change(self):
...
+    def _multiselection_multivalue_prep(self, values):
+        final = None
+        multivalue = False
+        for value in values:
+            value = set(value)
+            if final is None:
+                final = value
+            elif multivalue or final != value:
+                final |= value
+                multivalue = True
```
BigRoy, line 471:
```suggestion
            if final is None:
                final = value
            elif multivalue and not final:
                break
            elif multivalue or final != value:
                final |= value
                multivalue = True
```

iLLiCiTiT: I'm doing Union, not intersection.

BigRoy: Damn those operator overloads - thanks. I clearly need to just go to bed tonight.

BigRoy:
Ok, this appears to be 99% perfect. There's just one issue with multiselect.

Say you have two instances, one that has selected **a, b, c** and the other has **a, b**.
Select both instances, it shows `<Multiselect>` ✅ , but when you now deselect the "c" option it doesn't update the label/display and will actually show "a", "b", "c" instead of "a", "b". Note that if you click the dropdown menu again, you can still see "c" deactivated. Also if you now select just one of the two instances, it correctly has "a", "b" both and that is also what it saves on the instances.

As such, it seems when going from Multiselection and then toggling some things the actual display bar doesn't update as you'd expect.

If after the "multiselection" label had disappeared becaues you changed some things in the popup menu - then if you do so again, it'll THEN update the label accordingly. So the issue is that the label just doesn't update whenever "Multiselection" disappears.

BigRoy:
Note: We should [update the docstring](https://github.com/ynput/OpenPype/blob/5b4bdeb6259dc5bd69ec773ba717a1efad17a695/openpype/lib/attribute_definitions.py#L427) since now in multiselection mode it's not about single items. :)

openpype/lib/attribute_definitions.py
```diff
@@ -443,30 +448,45 @@ def __init__(self, key, items, default=None, **kwargs):
...
-        return self.default
+        if not self.multiselection:
+            if value in self._item_values:
+                return value
+            return self.default
+
+        if value is None:
+            return copy.deepcopy(self.default)
+        return list(self._item_values.intersection(value))
+
```
BigRoy, line 484:
```suggestion
```

iLLiCiTiT:
> As such, it seems when going from Multiselection and then toggling some things the actual display bar doesn't update as you'd expect.

Fixed

BigRoy (approved):
Great work - works as expected.

jakubjezek001 (approved):
LGTM

PR #5553, opened 2023-09-01
Chore: Create plugin does not expect system settings

iLLiCiTiT:
## Changelog Description
System settings are not passed to initialization of create plugin initialization (and `apply_settings`).

## Additional info
Reasons: The system settings are not used anywhere, and since AYON there won't be 2 types of settings.

We need validation of all Create plugins: They should not use `system_settings` in `apply_settings` method and they should not override `__init__` method.

To make this PR safer the system settings can still be passed to creator initialization, but eventually it will be removed in future PR anyway, so it would be just temporary backwards compatibility.

This PR must be updated with latest develop and revalidated changes before merge.

## Testing notes:
1. Technically run all available creators in Publisher

- [x] Traypublisher
- [x] Maya
- [x] Nuke
- [x] AfterEffects
- [x] Photoshop
- [x] 3dsMax
- [x] Houdini
- [ ] TVPaint

### DIFF
--- a/openpype/pipeline/create/creator_plugins.py
+++ b/openpype/pipeline/create/creator_plugins.py
@@ -193,7 +193,9 @@ class BaseCreator:
     # QUESTION make this required?
     host_name = None
 
-    def __init__(self, project_settings, create_context, headless=False):
+    def __init__(
+        self, project_settings, system_settings, create_context, headless=False
+    ):
         # Reference to CreateContext
         self.create_context = create_context
         self.project_settings = project_settings
@@ -202,13 +204,26 @@ class BaseCreator:
         # - we may use UI inside processing this attribute should be checked
         self.headless = headless
 
+        expect_system_settings = False
         if is_func_signature_supported(
             self.apply_settings, project_settings
         ):
             self.apply_settings(project_settings)
         else:
+            expect_system_settings = True
             # Backwards compatibility for system settings
-            self.apply_settings(project_settings, {})
+            self.apply_settings(project_settings, system_settings)
+
+        init_overriden = self.__class__.__init__ is not BaseCreator.__init__
+        if init_overriden or expect_system_settings:
+            self.log.warning((
+                "WARNING: Source - Create plugin {}."
+                " System settings argument will not be passed to"
+                " '__init__' and 'apply_settings' methods in future versions"
+                " of OpenPype. Planned version to drop the support"
+                " is 3.15.6 or 3.16.0. Please contact Ynput core team if you"
+                " need to keep system settings."
+            ).format(self.__class__.__name__))
 
     def apply_settings(self, project_settings):
         """Method called on initialization of plugin to apply settings.
--- a/openpype/pipeline/create/creator_plugins.py
+++ b/openpype/pipeline/create/creator_plugins.py
@@ -221,7 +221,7 @@ class BaseCreator:
                 " System settings argument will not be passed to"
                 " '__init__' and 'apply_settings' methods in future versions"
                 " of OpenPype. Planned version to drop the support"
-                " is 3.15.6 or 3.16.0. Please contact Ynput core team if you"
+                " is 3.16.6 or 3.17.0. Please contact Ynput core team if you"
                 " need to keep system settings."
             ).format(self.__class__.__name__))
 
--- a/openpype/pipeline/create/creator_plugins.py
+++ b/openpype/pipeline/create/creator_plugins.py
@@ -1,11 +1,7 @@
 import copy
 import collections
 
-from abc import (
-    ABCMeta,
-    abstractmethod,
-    abstractproperty
-)
+from abc import ABCMeta, abstractmethod
 
 import six
 
@@ -84,7 +80,8 @@ class SubsetConvertorPlugin(object):
     def host(self):
         return self._create_context.host
 
-    @abstractproperty
+    @property
+    @abstractmethod
     def identifier(self):
         """Converted identifier.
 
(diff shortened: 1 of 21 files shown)


BigRoy:
Unfortunately I did end up using `system_settings` on our end for a Deadline creator plug-in.

See [here](https://github.com/BigRoy/OpenPype/blob/colorbleed_develop3/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py#L164-L172) to use it [here](https://github.com/BigRoy/OpenPype/blob/colorbleed_develop3/openpype/modules/deadline/plugins/publish/submit_maya_deadline.py#L884-L889)

See [here](https://github.com/BigRoy/OpenPype/blob/colorbleed_develop3/openpype/modules/deadline/plugins/publish/collect_pools.py#L34-L39) to use it [here](https://github.com/BigRoy/OpenPype/blob/colorbleed_develop3/openpype/modules/deadline/plugins/publish/collect_pools.py#L61-L75).

Any good ideas on how I could still optimally retrieve our globally define deadline URL - since the project settings URL just picks one by name from the Module Settings? The only thing I can think of is that the `apply_settings`? If I'd initialize a `ModuleManager` I'd still need to do that per plug-in to get the instantiated modules at the time of `apply_settings` so I couldn't get it from a module instance either.

---

Also, since this is a backwards incompatible change for external code bases (e.g. anyone having a custom new style creator using `apply_settings`) it might be good to merge this into a next minor release maybe with a clear note in the changelog?

iLLiCiTiT:
> Does this PR affect those too?

No.

> Looking at the changed files it doesn't but I'm assuming a PR for that might be incoming too?

Publish plugins already have changed code to allow expect only project settings, but it is "less radical" there, because system settings are still used, which won't change until full AYON release.
I've chose more radical approach in create plugins, because I didn't find any usage of them anywhere and couldn't think of the usage...

> Also, since this is a backwards incompatible change for external code bases (e.g. anyone having a custom new style creator using apply_settings) it might be good to merge this into a next minor release maybe with a clear note in the changelog?

I would rather change it to be more backwards compatible than wait for next-minor, last next-minor took more than 6 months.

BigRoy:
For Create plug-ins I'm personally fine with removing it - I'd just need to sync up potential custom creators but if it's clear WHEN I should be doing that then I'm totally fine with it. It's a trivial change.

iLLiCiTiT:
Added backwards compatibility, System settings are still passed to Create plugin. A warning message is logged if `__init__` is overriden or `apply_settings` still expect `system_settings`.

openpype/pipeline/create/creator_plugins.py
```diff
@@ -224,7 +243,8 @@ def identifier(self):
 
         return self.family
 
-    @abstractproperty
+    @property
+    @abstractmethod
```
BigRoy, line 247: Isn't this Python 3.3+ only? [src](https://stackoverflow.com/a/48710068)

openpype/pipeline/create/creator_plugins.py
```diff
@@ -208,10 +204,33 @@ def __init__(
...
+            self.apply_settings(project_settings, system_settings)
+
+        init_overriden = self.__class__.__init__ is not BaseCreator.__init__
+        if init_overriden or expect_system_settings:
+            self.log.warning((
+                "WARNING: Source - Create plugin {}."
+                " System settings argument will not be passed to"
+                " '__init__' and 'apply_settings' methods in future versions"
+                " of OpenPype. Planned version to drop the support"
+                " is 3.15.6 or 3.16.0. Please contact Ynput core team if you"
```
BigRoy, line 224: The version numbers mentioned here seem low. `develop` is currently on: `3.16.5-nightly.3`?

openpype/pipeline/create/creator_plugins.py
iLLiCiTiT: Yes, but this is not Python 2 breaking. I would not bend to Python 2 at this moment.

openpype/pipeline/create/creator_plugins.py
iLLiCiTiT: > The version numbers mentioned here seem low. develop is currently on: 3.16.5-nightly.3?

It's 2 patch releases or 1 minor release.

BigRoy: So then it should be `3.16.7 or 3.17.0` right?

iLLiCiTiT: `3.16.5` is not out yet. `3.16.5-nightly.3` is preparation for `3.16.5`.

iLLiCiTiT:
```suggestion
                " is 3.16.6 or 3.17.0. Please contact Ynput core team if you"
```

jakubjezek001 (approved):
Nuke and TrayPublisher tested and worked normally

antirotor (approved):
Creators in Maya, Houdini, 3dsmax and Blender works

kalisp (approved):
Tested in PS, AE, works.

openpype/pipeline/create/creator_plugins.py
```diff
@@ -208,10 +204,33 @@ def __init__(
...
+            self.apply_settings, project_settings
+        ):
+            self.apply_settings(project_settings)
+        else:
+            expect_system_settings = True
+            # Backwards compatibility for system settings
+            self.apply_settings(project_settings, system_settings)
+
+        init_overriden = self.__class__.__init__ is not BaseCreator.__init__
+        if init_overriden or expect_system_settings:
```
BigRoy, line 218: Ok, so this warning is now always logged for all classes inheriting from the new `Creator` due to the `Creator` class already overriding `__init__`. :)

PR #5559, opened 2023-09-04
Enhancement: Maya fixes, optimizations and logging tweaks

BigRoy:
## Changelog Description

- Tweak logging levels
- Fix a lot of validators to now actually report a `PublishValidationError` for better UX
- Fix Collect Inputs for Renderlayers (I suspect this was broken since new publisher)
- Optimize `get_id_required_nodes` (this is much faster e.g. on scene save for many nodes)
- Fix Global AOV Mode support for Arnold (to disable/enable all AOVs in a renderlayer)
- Add Global AOV Mode support for Redshift (to disable/enable all AOVs in a renderlayer)
- Remove some pre-Maya 2017 support
- Some cosmetics as well 
- Only collect file input dependencies if the setting is enabled in project settings

## Additional info

Sorry for the massive PR - it was just for much for me to break this into smaller parts. Most commits are 'isolated' so if you have the spare time feel free to pick the commits you need to break this into smaller PRs.

## Testing notes:

1. Go crazy in Maya
2. Publish a variety of families, models, looks, renders, pointcaches


BigRoy:
> I've discovered one more issue - some of the validators are not inheriting from the mixin, therefor not respecting "optional" states and so on. This needs to be fixed but from the scope of this PR and possible conflicts, I am wondering if I shouldn't push those fixes here or create separate PR and solve conflicts later.

I had noticed that a lot of things that can be made 'optional' in settings indeed do not inherit from the optional mixin class. Whether it's best kept separate or fixed in this PR I'll leave to others to decide - this one is already quite a hefty one but since we need to test the different cases anyway we might as well put it in here.

@iLLiCiTiT thoughts?

antirotor:
> I am wondering if I shouldn't push those fixes here or create separate PR and solve conflicts later.

I did it here #5575

tokejepsen:
@BigRoy I think this PR might be too big. Could you maybe split the changes up into smaller PRs, maybe 1 per changelog?
I think that'll be faster than getting through this PR.

BigRoy:
> I think this PR might be too big. Could you maybe split the changes up into smaller PRs, maybe 1 per changelog?
I think that'll be faster than getting through this PR.


Unfortunately I don't have time to do so, but I agree that there are definitely a few features in here that are 'big changes' (like the `cbId` optimizations). Will close this PR and maybe look into generating tiny PRs from our `colorbleed` branch over time?

This is just a part of the `diff` between the branches even though most features should be able to work fine in OP and aren't too specific to our studio.

PR #5560, opened 2023-09-04
Chore: Removed Ass export script

iLLiCiTiT:
## Changelog Description
Removed Arnold render script, which was obsolete and unused.

## Additional info
The logic of the script looks like it was written for early Pype 2, the whole logic would not work as is at all in current state of OpenPype, thus there's not reason to keep it in codebase.


BigRoy (approved):
Sounds fine with me.

Maybe @tokejepsen would like to check this too?

antirotor (approved):
yes. This isn't working anymore (in fact it wasn't working properly even then, but 🤫 )

PR #5561, opened 2023-09-04
Maya: Yeti Cache Include viewport preview settings from source

BigRoy:
## Changelog Description

When publishing and loading yeti caches persist the display output and preview colors + settings to ensure consistency in the view

## Testing notes:

1. Publish yeti cache in Maya with some tweaked preview settings (e.g. viewport preview color, etc)
2. Load in yeti cache, settings should be preserved.

### DIFF
--- a/openpype/hosts/maya/plugins/load/load_yeti_cache.py
+++ b/openpype/hosts/maya/plugins/load/load_yeti_cache.py
@@ -15,6 +15,16 @@ from openpype.hosts.maya.api import lib
 from openpype.hosts.maya.api.pipeline import containerise
 
 
+# Do not reset these values on update but only apply on first load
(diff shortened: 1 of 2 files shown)


LiborBatek (approved):
My Yeti settings display overrides during Publish the ```YetiRig``` being as
image

And when loaded into some scne:

image

All works ok!

Just a note: would be lovely to not loose shader assignment on the ```groom mesh``` tho

or at least having the ```defaultShading```

BigRoy:
> Just a note: would be lovely to not loose shader assignment on the groom mesh tho
> or at least having the defaultShading

Valid point - can we create a separate issue to track this?

antirotor (approved):

PR #5563, opened 2023-09-04
Max: OP Data updates in Loaders

moonyuet:
## Changelog Description
Fix the bug on the loaders not being able to load the objects when iterating key and values with the dict.
Max prefers list over the list in dict.

## Additional info
n/a

## Testing notes:
1. Launch Max via Launcher
2. Openpype -> Load...
3. Load any families.


LiborBatek (approved):
All seems normal and functional. Did try to load multiple product types and switch theirs versions.

All good!

PR #5565, opened 2023-09-04
Blender: Set --python-use-system-env from PreLaunchHook

BigRoy:
## Changelog Description

Automatically add `--python-use-system-env` when running Blender when it's not pre-configured in the application settings for Blender (In Studio Settings > System Settings > Applications > Blender > {version} > arguments.

## Additional info

It often happens that admins add a new Blender version in the applications but don't also apply the `--python-use-system-env`  arguments meaning OpenPype will not be loaded for Blender plus won't be available in Blender on `PYTHONPATH`.

This avoids that issue by always enforcing the argument is added on launch plus updates the defaults so that `--python-use-system-env` isn't there by default either.

@iLLiCiTiT I've set it to run on `launch_types = set()` so it always runs. I basically want it to run on ANY Blender launch where prior to this it'd have also been applied due to being in the arguments settings for the application. (I feel like that's likely more than just 'local', right?)

---

Before:
image

After:
image

---

See [this reference discussion on Discord for somebody who was hitting this issue](https://discord.com/channels/517362899170230292/517382145552154634/1148250026099875990). Although there have also been [reports that it should be working without the flag](https://discord.com/channels/517362899170230292/563751989075378201/943193907842404362).

**Note:** For those who were already succsesfully launching blender without `--python-use-system-env` this PR might be giving unexpected results if - for whatever reason - they don't actually use the system env in Blender.

## Testing notes:

1. Run blender without `--python-use-system-env` argument in settings
2. Run blender with `--python-use-system-env` argument in settings (this is just to ensure it's backwards compatible with any settings a studio might currently have for custom versions.)
3. Everything should load and run fine for both


iLLiCiTiT:
> Automatically add --python-use-system-env when running Blender when it's not pre-configured in the application settings for Blender (In Studio Settings > System Settings > Applications > Blender > {version} > arguments.

We didn't add it automatically because someone wanted to launch it using rez or script (.bat/.sh) where it caused issues... (not saying I'm against).

> I've set it to run on launch_types = set() so it always runs. I basically want it to run on ANY Blender launch where prior to this it'd have also been applied due to being in the arguments settings for the application. (I feel like that's likely more than just 'local', right?)

It make sense only for `{LaunchTypes.local, LaunchTypes.remote, LaunchTypes.automated}`, the others are farm specific, where blender is not launched. But because `remote` and `automated` are not implemented for blender it could be just `{LaunchTypes.local}`.

BigRoy:
> We didn't add it automatically because someone wanted to launch it using rez or script (.bat/.sh) where it caused issues... (not saying I'm against).
> 
> EDITED: It would nice to have option to turn the prelaunch hook off.

Thanks - yes, then I suspect instead having a toggle somewhere in settings might be the decent other solution? But I was expecting this might've been the case. Any good place you think this could work? Should it somehow be a part of the Blender-specific application versions? or more globally?

> It make sense only for {LaunchTypes.local, LaunchTypes.remote, LaunchTypes.automated}, the others are farm specific, where blender is not launched. But because remote and automated are not implemented for blender it could be just {LaunchTypes.local}.

Thanks - I'll just make it the first three just in preparation of what might come; will save some debugging for those cases maybe?

iLLiCiTiT:
> will save some debugging for those cases maybe?

Not needed, that's unrelated to the launch hook itself.

BigRoy:
> It would nice to have option to turn the prelaunch hook off.

Any pointers on where I could best put that?

Here's an example option:
image

LiborBatek:
@iLLiCiTiT any insight for aformentioned things? thx

Tilix4 (commented):
Code LGTM!

About the UI, I'd move it right before the custom arguments list, before the executables, it is confusing.

BigRoy:
Added a commit that added the toggle in settings:
image

BigRoy:
> About the UI, I'd move it right before the custom arguments list, before the executables, it is confusing.

@Tilix4 This unfortunately wasn't as trivial because it builds that section using a template which is the full environment section - so I've opted to just add it at the end.

Tilix4:
@BigRoy This is fine for me!

openpype/hosts/blender/hooks/pre_add_python_use_system_env.py
```diff
@@ -0,0 +1,34 @@
...
+                           f"this prelaunch hook.")
+
+    def get_application_setting(self, key, default=None):
+        app = self.launch_context.application
+        group_name = app.group.name
+        app_name = app.name
+        return (
+            self.data["system_settings"]["applications"][group_name]
+                     ["variants"][app_name].get(key, default)
+        )
```
BigRoy, line 34: This likely moves in AYON? Is there a better approach to this @iLLiCiTiT ?

LiborBatek (approved):
Works for me in both scenarios...

when still having that ```Argument``` present
image

and also when removed but using the new Python path knob ```Enabled```
image

openpype/hosts/blender/hooks/pre_add_python_use_system_env.py
mkolar: Yep this will need to be made forward compatible with AYON. Settings and their defaults are here ;) https://github.com/ynput/OpenPype/tree/develop/server_addon/applications/server

mkolar (changes requested):

mkolar:
Because we're splitting OpenPype into ayon-core and individual host addons, this PR would have to be re-created to target one of those. It will need to be AYON compatible for it to get merged. 

We're closing it down, but we'll he happy for a new PRs to [ynput/ayon-blender](https://github.com/ynput/ayon-blender) repository once it's up.

PR #5567, opened 2023-09-04
Houdini: Fix outdated containers pop-up on opening last workfile on launch

BigRoy:
## Changelog Description

Fix Houdini not showing outdated containers pop-up on scene open when launching with last workfile argument

## Testing notes:

1. Launch Houdini into an asset that has an existing workfile with outdated containers
2. On launch (with launch last workfile) this should now popup the dialog that scene contains outdated containers


moonyuet (approved):
Show the popup dialog
image
and it will allow you to access the scene inventory to change version
image

antirotor (approved):

PR #5569, opened 2023-09-05
Fusion: Code updates

iLLiCiTiT:
## Changelog Description
Update fusion code which contains obsolete code. Removed `switch_ui.py` script from fusion with related script in scripts.

## Additional info
This is all I know should be removed, if other scripts should be removed too, just let me know, or do it.

## Testing notes:
Fusion integration should still work.


BigRoy (approved):
These were just utility scripts that aren't crucial to OpenPype's functionality and were originally just intended as quick scripts for our studio at the time so I feel like these can safely be removed from OpenPype's codebase.

PR #5570, opened 2023-09-05
Chore: Maya / Nuke remove publish gui filters from settings

BigRoy:
## Changelog Description

- Remove Publish GUI Filters from Nuke settings
- Remove Publish GUI Filters from Maya settings

## Additional info

Note that it seems like both TVPaint and Hiero also have the Publish GUI Filter settings currently - I am not sure if those use the new publisher already and whether it's worth also removing those in this PR or not.

I'm not sure how AYON settings work with its versioning - so let me know if the changes/commits seem like the right way.

## Testing notes:

1. Ensure Nuke + Maya settings + publishing still work (in both AYON and OpenPype)


BigRoy:
I'm not sure either - as far as I can tell these were publisher presets for the old publisher that'd set the Optional states of certain plugins according to the preset. The new publisher doesn't have this concept since it stores and remembers tweaked settings and thus there's much less need for these presets.

@iLLiCiTiT @antirotor any idea whether it makes sense to remove these settings?

Also note the comment I made about TV Paint in the PR description under **Additional Info**

iLLiCiTiT:
> I'm not sure either - as far as I can tell these were publisher presets for the old publisher that'd set the Optional states of certain plugins according to the preset. The new publisher doesn't have this concept since it stores and remembers tweaked settings and thus there's much less need for these presets.

Exactly, this cannot be tested, because settings that were removed, were not used.

BigRoy:
> Yes, we should remove the settings for them too. I'll leave it up to you if you want to add it to this PR.

@iLLiCiTiT only for those that do **not** use the old publisher anymore, correct? (I believe Hiero still does)

iLLiCiTiT:
> only for those that do not use the old publisher anymore, correct? (I believe Hiero still does)

Yes, from the remaining it is only TVPaint.

openpype/settings/ayon_settings.py
```diff
@@ -694,16 +694,6 @@ def _convert_nuke_project_settings(ayon_settings, output):
...
 
-    # --- Filters ---
-    new_gui_filters = {}
-    for item in ayon_nuke.pop("filters"):
-        subvalue = {}
-        key = item["name"]
-        for subitem in item["value"]:
-            subvalue[subitem["name"]] = subitem["value"]
-        new_gui_filters[key] = subvalue
-    ayon_nuke["filters"] = new_gui_filters
```
BigRoy, line 705: Not entirely sure why Maya/TVPaint were lacking this conversion even though Nuke/Hiero settings did have it.

BigRoy:
Ready to check again @iLLiCiTiT

openpype/settings/ayon_settings.py
antirotor: I think this was only mistake during the conversion

BigRoy: So the removal in this PR is correct?

iLLiCiTiT (approved):

PR #5571, opened 2023-09-05
Create Plugins: Better check of overriden '__init__' method

iLLiCiTiT:
## Changelog Description
Create plugins do not log warning messages about each create plugin because of wrong `__init__` method check.

## Additional info
Use `__init__` check for each base class of create plugin. Issue caused by https://github.com/ynput/OpenPype/pull/5553 .

## Testing notes:
1. Open publisher.
2. There should not be any logged warnings if create plugin does not override `__init__` or expect system settings in `apply_settings`.

### DIFF
--- a/openpype/pipeline/create/creator_plugins.py
+++ b/openpype/pipeline/create/creator_plugins.py
@@ -214,8 +214,16 @@ class BaseCreator:
             # Backwards compatibility for system settings
             self.apply_settings(project_settings, system_settings)
 
-        init_overriden = self._method_is_overriden("__init__")
-        if init_overriden or expect_system_settings:
+        init_use_base = any(
+            self.__class__.__init__ is cls.__init__
+            for cls in {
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
Yes, this works for me! 

I did still see this logged:
```
# *** WRN: >>> { CreateYetiCache }: [ WARNING: Source - Create plugin CreateYetiCache. System settings argument will not be passed to '__init__' and 'apply_settings' methods in future versions of OpenPype. Planned version to drop the support is 3.16.6 or 3.17.0. Please contact Ynput core team if you need to keep system settings. ] 
```

Which shows [that that creator has something wrong](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/maya/plugins/create/create_yeti_cache.py#L16) and indeed - that's not how you should define attribute definitions! Haha. Will make a PR to fix that too.

PR #5573, opened 2023-09-05
Maya: Implement USD publish and load using native `mayaUsdPlugin`

BigRoy:
## Changelog Description

Implement Creator and Loaders for extraction and loading of USD files using Maya's own `mayaUsdPlugin`.

Also adds support to load a `usd` file into an Arnold Standin (`aiStandin`).

## Additional info

This now lives alongside the Multiverse implementation. You can decide to either create a MayaUSD instance or a Multiverse instance. Both will publish a `usd` family result (product-type) but during publishing it will also have the family specific to that creator type to choose the right extractor (and potentially validations) which are `mayaUsd` and `mvUsd`.

## Testing notes:

1. Launch Maya
2. Create USD instance
3. Publish USD
4. Load USD in Maya

### DIFF
--- a/openpype/hosts/maya/plugins/load/load_arnold_standin.py
+++ b/openpype/hosts/maya/plugins/load/load_arnold_standin.py
@@ -17,6 +17,7 @@ from openpype.hosts.maya.api.lib import (
 )
 from openpype.hosts.maya.api.pipeline import containerise
 
+
 def is_sequence(files):
     sequence = False
     collections, remainder = clique.assemble(files, minimum_items=1)
@@ -28,6 +29,7 @@ def is_sequence(files):
 def get_current_session_fps():
     session_fps = float(legacy_io.Session.get('AVALON_FPS', 25))
     return convert_to_maya_fps(session_fps)
+
 
 class ArnoldStandinLoader(load.LoaderPlugin):
     """Load as Arnold standin"""
--- a/openpype/hosts/maya/plugins/load/load_arnold_standin.py
+++ b/openpype/hosts/maya/plugins/load/load_arnold_standin.py
@@ -32,8 +32,8 @@ def get_current_session_fps():
 class ArnoldStandinLoader(load.LoaderPlugin):
     """Load as Arnold standin"""
 
-    families = ["ass", "animation", "model", "proxyAbc", "pointcache"]
-    representations = ["ass", "abc"]
+    families = ["ass", "animation", "model", "proxyAbc", "pointcache", "usd"]
+    representations = ["ass", "abc", "usda", "usdc", "usd"]
 
     label = "Load as Arnold standin"
     order = -5
--- a/openpype/hosts/maya/api/plugin.py
+++ b/openpype/hosts/maya/api/plugin.py
@@ -129,11 +129,33 @@ class MayaCreatorBase(object):
             shared_data["maya_cached_legacy_subsets"] = cache_legacy
         return shared_data
 
+    def get_publish_families(self):
+        """Return families for the instances of this creator.
+
+        Allow a Creator to define multiple families so that a creator can
+        e.g. specify `usd` and `usdMaya` and another USD creator can also
+        specify `usd` but apply different extractors like `usdMultiverse`.
+
+        There is no need to override this method if you only have the
+        primary family defined by the `family` property as that will always
+        be set.
+
+        Returns:
+            list: families for instances of this creator
+
+        """
+        return []
+
     def imprint_instance_node(self, node, data):
 
         # We never store the instance_node as value on the node since
         # it's the node name itself
         data.pop("instance_node", None)
+
+        # Don't store `families` since it's up to the creator itself
+        # to define the initial publish families - not a stored attribute of
+        # `families`
+        data.pop("families", None)
 
         # We store creator attributes at the root level and assume they
         # will not clash in names with `subset`, `task`, etc. and other
@@ -186,6 +208,11 @@ class MayaCreatorBase(object):
         # Explicitly re-parse the node name
         node_data["instance_node"] = node
 
+        # If the creator plug-in specifies
+        families = self.get_publish_families()
+        if families:
+            node_data["families"] = families
+
         return node_data
 
     def _default_collect_instances(self):
@@ -229,6 +256,14 @@ class MayaCreator(NewCreator, MayaCreatorBase):
         members = list()
         if pre_create_data.get("use_selection"):
             members = cmds.ls(selection=True)
+
+        # Allow a Creator to define multiple families
+        publish_families = self.get_publish_families()
+        if publish_families:
+            families = instance_data.setdefault("families", [])
+            for family in self.get_publish_families():
+                if family not in families:
+                    families.append(family)
 
         with lib.undo_chunk():
             instance_node = cmds.sets(members, name=subset_name)
--- a/openpype/hosts/maya/plugins/create/create_maya_usd.py
+++ b/openpype/hosts/maya/plugins/create/create_maya_usd.py
@@ -1,0 +1,143 @@
+from openpype.hosts.maya.api import plugin, lib
+from openpype.lib import (
+    BoolDef,
+    NumberDef,
+    TextDef,
+    EnumDef
+)
+
+
+class CreateMayaUsd(plugin.MayaCreator):
+    """Create Maya USD Export"""
+
+    identifier = "io.openpype.creators.maya.mayausd"
+    label = "Maya USD"
+    family = "usd"
+    icon = "cubes"
+    description = "Create Maya USD Export"
+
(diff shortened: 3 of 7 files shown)


antirotor:
I've used similar code for my internal native maya usd publishing tests and I like the trick with the publishing families.

openpype/hosts/maya/plugins/load/load_arnold_standin.py
```diff
@@ -29,11 +30,12 @@ def get_current_session_fps():
...
     return convert_to_maya_fps(session_fps)
 
+
 class ArnoldStandinLoader(load.LoaderPlugin):
     """Load as Arnold standin"""
 
-    families = ["ass", "animation", "model", "proxyAbc", "pointcache"]
-    representations = ["ass", "abc"]
+    families = ["ass", "animation", "model", "proxyAbc", "pointcache", "usd"]
+    representations = ["ass", "abc", "usda", "usdc", "usd"]
```
BigRoy, line 38: I wonder if all of these extensions `usda` and `usdc` should actually be representation `usd`?
```suggestion
    representations = ["ass", "abc", "usd"]
```

BigRoy:
I'm marking this ready for review just so I can get some early input on the current state to make sure I'm moving the correct direction forward. It's not ready to merge I'd say but would love to get some input.

**Maya USD Proxy Shape -> Shareable**
The Maya USD system has a lot of features to 'edit' the data of loaded content in the maya usd proxy shapes - which by default I think even saves into the loaded USD file. For loaded published content this is of course NOT what we want. We can set the proxys to `shareable=False` (attribute on the node) so that any local changes to the USD in that proxy will happen in an anonymous layer just for that node and will get saved along with the current maya file instead. I assume that's likely what we'd want, yes?

Also see:
- [Save USD Data in Maya](https://help.autodesk.com/view/MAYAUL/2024/ENU/?guid=GUID-EEBC1C65-0805-4488-80C7-3C061B316294)

LiborBatek:
@BigRoy sounds awesome! Will test asap!

openpype/hosts/maya/plugins/load/load_arnold_standin.py
tokejepsen: What is `usda` and `usdc`?

BigRoy: > USDA files are plain ascii, while USDC are binary files

Also see [here](https://openusd.org/release/tut_converting_between_layer_formats.html#converting-between-usda-usdc-and-usd)

tokejepsen:
> I'm marking this ready for review just so I can get some early input on the current state to make sure I'm moving the correct direction forward. It's not ready to merge I'd say but would love to get some input.

Should we maybe make this a draft if its not ready to be merged yet?

BigRoy:
> Should we maybe make this a draft if its not ready to be merged yet?

I'm looking to get reviews in for this - so I wasn't sure. Usually when PRs are marked "Draft" not a lot of PR reviewers around here take a look, but that taking a look is exactly what I need to steer this direction forward.

For what it's worth, what's there in the PR currently is functional so could even be merged, but I'd rather work out the USD workflow further than just what it is now.

LiborBatek:
Even being it silly...when no USDmaya plugin running it will crash Maya during Publish...

I guess there should be sort of check if mayaUSD plugin is running before Publish.
image

LiborBatek:
So when only standard ```maya``` enviroment used (no Vray present) I am able to ```Publish``` ```usd``` asset (model in my case) and also being able to ```load``` it via ```Loader``` into scene.

I was trying to use ```usd``` asset for ```look``` publishing (not sure if it should be possible) but it will fail on validation of the content for ```Look``` set.

image

I guess I should be able to use the ```Look``` from existing ```model``` asset (ma, abc) but maybe its worth noticing this limitation...

LiborBatek:
Ok so its not even listed in ```Look Manager``` when loaded as ```usdProxy``` but just as ```arnold ass```...

Again not sure if this should be supported when speaking of USD assets tho...
image

LiborBatek (approved):
I have tested the core functionality in ```maya 2023``` following the testing steps and all working fine.

- [x] Publish ```usd``` asset
- [ ] Load ```usd``` asset into scene via ```Loader```
- [ ] Assign existing ```Look``` on ```usd``` loaded as ```Arnold Standin```

image


As a side note I was not able to produce any ```Look``` publishes / loads when using ```usdProxy``` loader which I guess is not suppose to work. So not a deal breaker at all.

LiborBatek:
So I did successfully ```Publish``` also full ```layout``` into ```usd``` asset
image

From user pov there could be some friendlier ```error report``` during ```Publish``` validations if ```Strip Namespaces``` being ```On``` and using loaded (aka Referenecd files) assets so user have to turn it ```Off```. Mine scene was kind of "edge case" tho...having multiple of the same assets present so non unique names present after stripping.

Here is the error report if using Namespace stripping during Publish:

```
DEBUG: Performing extraction ...
DEBUG: Collected objects: ['|street_SET']
DEBUG: Exporting USD: C:/TEMP/pyblish_tmp_kjkn6_t9/usdMain.usd / ['|street_SET']
  File "D:\REPO\OpenPype\.venv\lib\site-packages\pyblish\plugin.py", line 527, in __explicit_process
    runner(*args)
  File "D:\REPO\OpenPype\openpype\hosts\maya\plugins\publish\extract_maya_usd.py", line 254, in process
    cmds.mayaUSDExport(file=file_path,
  File "D:\REPO\OpenPype\openpype\hosts\maya\plugins\publish\extract_maya_usd.py", line 2, in mayaUSDExport
RuntimeError: The xformOp 'xformOp:translate' already exists in xformOpOrder [[xformOp:translate]].
Could not add xform op
Multiple dag nodes map to the same prim path after stripping namespaces: |street_SET|buildings_GRP|buildingC_modelMain_01_:_GRP - |street_SET|buildings_GRP|buildingA_modelMain_01_:_GRP
```

BigRoy:
- [Crash with V-ray](https://github.com/ynput/OpenPype/pull/5573#issuecomment-1720867225) - interesting. Still, even with V-Ray it should not crash. Would be good if we could figure out _why_ it crashes and on what? Any case you have more info?
- [Publishing look on mayaUsdProxyShape](https://github.com/ynput/OpenPype/pull/5573#issuecomment-1720913932) is indeed not supported. Looks for USD proxy shapes should actually be done within the USD data itself and doesn't allow assigning maya native shaders per mesh, etc. as far as I know. It'd be interesting however to define a workflow on how we'd want to approach that look assignment workflow.
- [Look assignment not possible on mayaUsdProxyShape](https://github.com/ynput/OpenPype/pull/5573#issuecomment-1720919071) again, correct behavior, it's not possible to assign maya native shaders like that. They'd really need to be "USD edits" in the first place. Note that the Lookdev workflow has much improved with latest maya usd plugin version and Maya 2024 - so definitely play with it there to get familiar; not that this PR supports it much but hopefully it might help you brainstorm on how you'd like the lookdev workflow to be. I believe [this video](https://www.youtube.com/watch?v=O9HXeSQd2Fk) covers it.
- **TODO** [that strip namespaces error](https://github.com/ynput/OpenPype/pull/5573#issuecomment-1721118328) should definitely get a validator - thanks.

So what works for you is indeed the basics of what this PR offers now - extracting just maya geometry (and arnold lookdevs) directly into a single USD file. It doesn't work with actual custom layer overrides or building full on USD assets with variants, etc.
For that it's likely best to first get a "workflow description" going of what we'd want to support and how so we know what code is needed.

LiborBatek:
@antirotor maybe you could put your thoughts here
> So what works for you is indeed the basics of what this PR offers now - extracting just maya geometry (and arnold lookdevs) directly into a single USD file. It doesn't work with actual custom layer overrides or building full on USD assets with variants, etc. For that it's likely best to first get a "workflow description" going of what we'd want to support and how so we know what code is needed.

antirotor (approved):
I think this is great basis for additional work. There are few caveats as mentioned, but nothing breaking. Thanks

BigRoy:
Some follow up prototyping to this PR:

https://github.com/ynput/OpenPype/assets/2439881/04f5c923-d796-4735-b409-4e43c4b7fdc9

_Mostly posting it here so that it's hosted somewhere_.

BigRoy:
Some more prototyping:

https://github.com/ynput/OpenPype/assets/2439881/eaad169a-d592-42ff-b9f2-9182d6d3d333


https://github.com/ynput/OpenPype/assets/2439881/88709671-2f50-4765-8056-bf4e3300cac4

BigRoy:
And a Blender test:

https://github.com/ynput/OpenPype/assets/2439881/488dfe73-cb40-4180-b439-a7aed49142db

Issue #5468, opened 2023-08-18
Bug: Maya - Extract Yeti Cache Error


LiborBatek:
Using OP 3.16.x with new Publisher in ```maya 2023``` tested to publish ```YetiCache``` instance but getting error during extraction.

![image](https://github.com/ynput/OpenPype/assets/112623825/e80079e2-3cb6-48fa-a3e4-52fddcea01e9)

```
Traceback (most recent call last):
  File "D:\REPO\OpenPype\.venv\lib\site-packages\pyblish\plugin.py", line 527, in __explicit_process
    runner(*args)
  File "D:\REPO\OpenPype\openpype\hosts\maya\plugins\publish\extract_yeti_cache.py", line 29, in process
    start_frame = instance.data["frameStartHandle"]
KeyError: 'frameStartHandle'
```

Also not sure if the ```YetiCache``` instance doesnt miss ```frameStart``` and ```frameEnd```.

Note: this was done in ```Asset``` with ```frameStart``` ```1001``` and ```frameEnd``` also ```1001```.  But also happens if different ```frameRange``` set

Happens even if not ```Validate FrameRange``` is turned on.

![image](https://github.com/ynput/OpenPype/assets/112623825/a0495759-b0d5-45a7-96ff-74949b9237f4)

### Expected Behavior:

Should not be failing on ```frameStartHandle``` attribute.

### Version

3.16.4-nightly.2

### What platform you are running OpenPype on?

Windows

### Steps To Reproduce:

1. Setup Yeti
2. create ```YetiCache``` instance
3. Hit Publish

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

[cuID:[OP-6711](https://app.clickup.com/t/861najcby)]

LiborBatek:
https://app.clickup.com/t/6658547/OP-6616

LiborBatek:
@kalisp @iLLiCiTiT thats the issue u were quickly fixin before me doing the Yeti demo for client...

Also valid issue not just speaking of ```handle start``` ```handle end``` but also for ```preroll``` values/entries

Will be happy to review it then. Thx

PR #5574, opened 2023-09-05
Maya: Fix attribute definitions for `CreateYetiCache`

BigRoy:
## Changelog Description

Fix attribute definitions for `CreateYetiCache`

## Additional info

`__init__` was overridden instead of using `get_instance_attr_defs`. Previously the attribute definitions should not have worked.

Note that even with this PR [the issue here](https://github.com/ynput/OpenPype/issues/5468#issuecomment-1706070618) might still be present.

## Testing notes:

1. Create yeti cache instance
2. Publish it


LiborBatek (approved):
All good!

YetiCache published without any issues.

PR #5572, opened 2023-09-05
Chore: add default isort config

antirotor:
## Changelog Description
Add default configuration for [isort tool](https://pycqa.github.io/isort/)

## Testing notes:
This shouldn't affect any functionality of OpenPype/AYON

PR #5575, opened 2023-09-05
Maya: Add `OptionalPyblishPluginMixin` to all validators

antirotor:
## Changelog Description
This PR is adding `OptionalPyblishPluginMixin` inheritance to all Maya validators. This fixes issues where validators are set as *Optional* in the Settings but that state is ignored on them.

## Additional info
Rationale to add `OptionalPyblishPluginMixin` to all validators is that it should be decided in the Settings if the plugin is optional or not - or at least on one place and currently there is no way to make Settings dynamic enough so that they will get that from the plugins itself. If this mixin is present, but the setting *Optional* in Settings is not, it won't allow anyone to disable even the important **"must-run"** validators. And at the same time, adding it to the Settings will make the validator settable to *Optional* state.

This is also re-sorted imports with [isort tools](https://pycqa.github.io/isort/) - for that I am truly sorry as it might complicate code review, but since I had to change import on most of these plugins anyway, I've chose to do it for the sake of consisten and more readable code. I've added default isort configuration in #5572.

## Testing notes:
1. Open Maya
2. Publish - validators that are set to be optional should behave so

### DIFF
--- a/openpype/hosts/maya/plugins/publish/validate_animation_out_set_related_node_ids.py
+++ b/openpype/hosts/maya/plugins/publish/validate_animation_out_set_related_node_ids.py
@@ -1,16 +1,18 @@
 import maya.cmds as cmds
-
 import pyblish.api
+
 import openpype.hosts.maya.api.action
 from openpype.hosts.maya.api import lib
 from openpype.pipeline.publish import (
+    OptionalPyblishPluginMixin,
+    PublishValidationError,
     RepairAction,
     ValidateContentsOrder,
-    PublishValidationError
 )
 
 
-class ValidateOutRelatedNodeIds(pyblish.api.InstancePlugin):
+class ValidateOutRelatedNodeIds(pyblish.api.InstancePlugin,
+                                OptionalPyblishPluginMixin):
     """Validate if deformed shapes have related IDs to the original shapes
 
     When a deformer is applied in the scene on a referenced mesh that already
--- a/setup.cfg
+++ b/setup.cfg
@@ -29,3 +29,10 @@ directory = ./coverage
 
 [tool:pytest]
 norecursedirs = repos/* openpype/modules/ftrack/*
+
+[isort]
+line_length = 79
+multi_line_output = 3
+include_trailing_comma = True
+force_grid_wrap = 0
+combine_as_imports = True
--- a/openpype/hosts/maya/plugins/publish/validate_animation_content.py
+++ b/openpype/hosts/maya/plugins/publish/validate_animation_content.py
@@ -1,8 +1,9 @@
 import pyblish.api
+
 import openpype.hosts.maya.api.action
 from openpype.pipeline.publish import (
     PublishValidationError,
-    ValidateContentsOrder
+    ValidateContentsOrder,
 )
 
 
--- a/openpype/hosts/maya/plugins/publish/validate_arnold_scene_source.py
+++ b/openpype/hosts/maya/plugins/publish/validate_arnold_scene_source.py
@@ -1,10 +1,14 @@
 import pyblish.api
+
 from openpype.pipeline.publish import (
-    ValidateContentsOrder, PublishValidationError
+    OptionalPyblishPluginMixin,
+    PublishValidationError,
+    ValidateContentsOrder,
 )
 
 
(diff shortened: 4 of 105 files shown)


openpype/hosts/maya/plugins/publish/validate_animation_out_set_related_node_ids.py
```diff
@@ -1,16 +1,18 @@
...
+    PublishValidationError,
     RepairAction,
     ValidateContentsOrder,
-    PublishValidationError
 )
 
 
-class ValidateOutRelatedNodeIds(pyblish.api.InstancePlugin):
+class ValidateOutRelatedNodeIds(pyblish.api.InstancePlugin,
+                                OptionalPyblishPluginMixin):
```
BigRoy, line 15: Unfortunately the changes made of only adding the mixin does not suffice. The `process` method should also be doing a `self.is_active(instance.data)` check to make sure it's actually skipped if the instance is toggled off. (Note that for ContextPlugin it is `self.is_active(context.data)`.

BigRoy (changes requested):
1. See comment - all of the mixin inherited classes should do the `is_active` check to actually work as intended.
2. Also validators that ensure the correct behavior of the family for publishing should NEVER be optional. I noticed you've added the mixin to some of the classes that are actually validators that ensure the publishing works as expected like `ValidateLookContent`, `ValidateModelContent`, `ValidateStepSize`. I think for now you're best off ONLY changing those plugins that actually are exposed in settings currently?
    - You seem to have skipped `ValidateAnimationContent` which is ok because it should not be optional but I think you just 'forgot' that one? :)

setup.cfg
BigRoy, line 1: Let's make sure the PR that sets the isort settings is merged first so it doesn't historically find its way into the codebase through a Maya PR _or_ remove this change from this PR.

antirotor: shouldn't be there but then again, it really does nothing except adding configuration for tool, it doesn't even install it.

openpype/hosts/maya/plugins/publish/validate_animation_out_set_related_node_ids.py
antirotor: > Also validators that ensure the correct behavior of the family for publishing should NEVER be optional.

True, but to really make them optional, there is another step and that is defining that in the Settings. My point is that even if we add the mixin, it doesn't really make the plugin optional, only the existance of the switch in the Settings does. I would even go that far later on as to unify inheritance of all validator plugin to one class that will wrap it together. Because then it wouldn't be *"confusing"* in the code as you wouldn't see `OptionalPyblishPluginMixin` in the signature of the validator class on non-optional plugins.

LiborBatek (approved):
Now any ```validator``` can be set as ```optional``` and visible then in the ```publisher```
image
Works fine.

BigRoy:
> Works fine.

This is not true - the plug-ins are still lacking the `is_active` check and thus when toggled off in the UI they still process as usual. The checkboxes in the publisher thus do nothing.

LiborBatek:
> > Works fine.
> 
> This is not true - the plug-ins are still lacking the `is_active` check and thus when toggled off in the UI they still process as usual. The checkboxes in the publisher thus do nothing.

Will try it out again to confirm. Thx

LiborBatek:
@BigRoy well to be honest I did not tested every single validator but obviously the one I tested does take into account wether ```ON``` or ```OFF```


https://github.com/ynput/OpenPype/assets/112623825/d5bec4ed-91b7-48a2-8717-ae139322e34f

I can of course do more of them one by one but seems fine...at first glance. Could u give me some validator exactly not working for you so I can test that special one, pls?

LiborBatek:
@BigRoy  so I finally found first not working

```Transform Zero (Freeze)``` which does not respect state of the ENABLED / DISABLED knob

@antirotor  not sure why some validators work and some doesnt

LiborBatek (changes requested):
Seems like some ```validators``` doesnt respect the ```Enabled / Disabled``` knob state.

Need to be addressed.

BigRoy:
The reason they don't work is described [here](https://github.com/ynput/OpenPype/pull/5575#discussion_r1316209476)

mkolar:

PR #5580, opened 2023-09-06
Nuke: removing redundant workfile colorspace attributes

jakubjezek001:
## Changelog Description
Nuke root workfile colorspace data type knobs are long time configured automatically via config roles or the default values are also working well. Therefore there is no need for pipeline managed knobs. 

## Additional info
- removed in OpenPype
- removed in server addon

## Testing notes:
1. Open nuke with color-managed enabled and see all works as expected

### DIFF
--- a/openpype/settings/defaults/project_settings/nuke.json
+++ b/openpype/settings/defaults/project_settings/nuke.json
@@ -28,11 +28,7 @@
             "colorManagement": "Nuke",
             "OCIO_config": "nuke-default",
             "workingSpaceLUT": "linear",
-            "monitorLut": "sRGB",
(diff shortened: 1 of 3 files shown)


jrsndl:
Having root workfile colorspace data type knobs pipeline managed is maybe not strictly necessary, but quite convenient.
Does it need to be removed?

jakubjezek001:
> Having root workfile colorspace data type knobs pipeline managed is maybe not strictly necessary, but quite convenient. Does it need to be removed?

We are monitoring client feedback, and most clients are happy to use the config roles instead of defining them here.

64qam (approved):

PR #5591, opened 2023-09-07
AYON: Fix representation context conversion

iLLiCiTiT:
## Changelog Description
Do not fix `"folder"` key in representation context until it is needed.

## Additional info
The folder is already filled during publishing so it is not necessary to convert template and change value.

## Testing notes:
1. Published representations could be loaded

### DIFF
--- a/openpype/client/server/conversion_utils.py
+++ b/openpype/client/server/conversion_utils.py
@@ -667,6 +667,9 @@ def convert_v4_representation_to_v3(representation):
             _c_folder = context["folder"]
             context["asset"] = _c_folder["name"]
 
+        elif "asset" in context and "folder" not in context:
+            context["folder"] = {"name": context["asset"]}
+
         if "product" in context:
             _c_product = context.pop("product")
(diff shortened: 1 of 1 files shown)


openpype/client/server/conversion_utils.py
```diff
@@ -1285,7 +1287,7 @@ def convert_update_representation_to_v4(
 
     if "context" in update_data:
         context = update_data["context"]
-        if "asset" in context:
+        if "folder" not in context and "asset" in context:
```
BigRoy, line 1290: Shouldn't we still always pop `asset` from context even if folder already exists in `context`?

iLLiCiTiT: We made few mistakes, and now there are representations which have `{folder[name]}` in template bude context data have `"asset"`, then we have few representations which have `{asset}` in template but `"folder"` in data, so nope.

jakubjezek001 (approved):
LGTM! works after testing

PR #5593, opened 2023-09-07
AYON: Avoid creation of duplicated links

iLLiCiTiT:
## Changelog Description
Handle cases when an existing link should be recreated and do not create the same link multitple times during single publishing.

## Additional info
AYON fails to create link if already exists. We didn't handle that case at all which caused alot crashed publishes. It crashed if single representation was loaded multiple times or if a scene was republished under the same version.

### DIFF
--- a/openpype/plugins/publish/integrate_inputlinks_ayon.py
+++ b/openpype/plugins/publish/integrate_inputlinks_ayon.py
@@ -1,7 +1,11 @@
 import collections
 
 import pyblish.api
-from ayon_api import create_link, make_sure_link_type_exists
+from ayon_api import (
+    create_link,
+    make_sure_link_type_exists,
+    get_versions_links,
+)
 
 from openpype import AYON_SERVER_ENABLED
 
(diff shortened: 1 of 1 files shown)


BigRoy:
This discussion has come up before, the 'duplicated' links previously were **intentionally** duplicated. At least, that's the case if this is about input links to e.g. used containers in the scene. This is because when e.g. set dressing it can be good to know **how many times** a certain asset was loaded in the shot. Without that information stored you won't be able to retrieve that information.

It's not crucial to retain that but just know that with this change you're thus losing certain information about the input links that were available before.

openpype/plugins/publish/integrate_inputlinks_ayon.py
```diff
@@ -144,16 +175,30 @@ def create_links_on_server(self, context, new_links):
...
+                    if output_id in existing_links:
+                        continue
+                    create_link(
+                        project_name,
+                        link_type,
+                        input_id,
+                        "version",
+                        output_id,
+                        "version"
+                    )
```
BigRoy, line 201: Question!
I have a feeling that creating each link individually on the backend as a separate creation request would be massively slow, no? You can easily have 10s or 100s of links per instance. With say 10 instances that's doing quite a few calls to the backend - or is the ayon api caching these and is a `session.commit()` or whatever needed to finish the submission?

iLLiCiTiT: > Also, another question - the link is like this right output_id -> input_id? Somehow my OCD trips over the fact that you need to specify the destination first and the source last as arguments to create_link

I don't understand this comment (nothing of it), but I guess that's because I don't understand the links flow much...

iLLiCiTiT:
> This discussion has come up before - I just can't find it now, the 'duplicated' links previously were intentionally duplicated. At least, that's the case if this is about input links to e.g. used containers in the scene. This is because when e.g. set dressing it can be good to know how many times a certain asset was loaded in the shot. Without that information stored you won't be able to retrieve that information.

I think it should be done in a different way then. Combination of entity type, in id and out id is unique identifier on a server. Don't know if that changed, or it was like that all the time. If we would want to support that we should be carefull about re-publishing.

BigRoy:
> I think it should be done in a different way then. Combination of entity type, in id and out id is unique identifier on a server. Don't know if that changed, or it was like that all the time. If we would want to support that we should be carefull about re-publishing.

Previously each version just had `inputs` or something as `data` published along (within the document) and the links themselves weren't separate entities (and there was nothing ensuring their uniqueness) as far as I know.

PR #5594, opened 2023-09-07
ayon-nuke: default factory to lists 

jakubjezek001:
## Changelog Description
Default factory were missing in settings schemas for complicated objects like lists and it was causing settings to be failing saving.

## Additional info
- Also clearing some already processed TODOs.
- improving of some docstrings

## Testing notes:
Just a code review should be ok.


iLLiCiTiT (approved):
LGTM

PR #5595, opened 2023-09-07
Maya: Fix rig validators for new out_SET and controls_SET names 

BigRoy:
## Changelog Description

Fix usage of `out_SET` and `controls_SET` since #5310 because they can now be prefixed by the subset name.

## Additional info

- Collect the rig sets only once (I've ordered it before Collect History so that the instance still contains less node, as an optimization)
- Also fixes a hard error when `out_SET` is not found, instead now only relevant `PublishValidationError` are raised to generate a nice report

## Testing notes:

_Note that quite a few validators for the rig can be disabled and are disabled by default - so you might need to enable them to validate they work correctly._

1. Enable all the rig validators
2. Create a new rig instance
3. Validate the rig publish

### DIFF
--- a/openpype/hosts/maya/plugins/publish/collect_rig_sets.py
+++ b/openpype/hosts/maya/plugins/publish/collect_rig_sets.py
@@ -1,0 +1,39 @@
+import pyblish.api
+from maya import cmds
+
+
+class CollectRigSets(pyblish.api.InstancePlugin):
+    """Ensure rig contains pipeline-critical content
+
+    Every rig must contain at least two object sets:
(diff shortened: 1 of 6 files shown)


LiborBatek (approved):
I have kept all ```rig validators```  ```ON``` in OP Settings  for ```rig``` publishing first...
image


When keeping the ```out_set``` empty user get correct validation error 
image this happens also correctly when ```control_set``` being empty
image

Now its possible to publish multiple rig instances from current workfile...no validation errors present.

image

antirotor (approved):
great! I think this was lurking there for really long time...

PR #5603, opened 2023-09-11
Blender: Remove Hardcoded Subset Name for Reviews

simonebarbieri:
## Changelog Description
Fixes hardcoded subset name for Reviews in Blender.

## Testing notes:
1. Create a review instance and publish it.
2. Check if it is published with the right subset name.


LiborBatek (approved):
Now when ```Review``` made all matching the rest of the other host integrations...

maya publish on the left / blender one on the right side
image

image

Issue #5602, opened 2023-09-11
Bug: 3dsmax usd plugin validator should run only when needed


antirotor:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

Currently USD plugin validator is bound to `model` family and will run always, no matter if "Extract Geometry (USD)" is enabled on the Instance or not. When you are missing `usdexport.dle` in your 3ds max, this validator will always fail even when not needed.

### Expected Behavior:

Model creator should have "Extract Geometry (USD)" defined on itself. Additional collector should add it to the model instance if this is enabled and validator and extractor should only have `usd` family on them.

### Version

3.16.6-nightly.1

### What platform you are running OpenPype on?

Windows

### Steps To Reproduce:

1. Run 3ds max without `usdexport.dle` available
2. Try to publish model without `Extract Geometry (USD)` enabled.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

antirotor:
I think the validator is OK, imagine additional plugin in the future needed just for USD part - some validator that will validate only USD specific stuff, not just presence of the 3ds max plugin. It is similar to BGEO in Houdini. There are two ways to handle this, all with their caveats:

**Way if the plugin attributes**
That is the current way. When adding new plugin, it would have to have `model` family, but additionaly some check to determine if the `Extract USD` is enabled or not. Or you would have that on the model creator itself and those additional plugins would have to check it manually against the instance.

**The Way of the Families**
Attribute `Extract USD` will move to the creator. Creator (or new collector) will add family USD to instance and the validator will run only on `usd` family.

I vote for the latter.

PR #5605, opened 2023-09-11
3ds max: small tweaks to obj extractor and model publishing flow

antirotor:
## Changelog Description
There migh be situation where OBJ Extractor passes without failure, but no obj file is produced. This is adding simple check directly into the extractor to catch it earlier then in the integration phase. Also switched `Validate USD Plugin` to optional, because it was always run no matter if the Extract USD was enabled or not, hindering testing (and publishing).

## Additional info
Apart of simple check for the file, this is modifying seemingly unrelated Validate USD Plugin. This was really needed without actually installing it into the max because of the above mentioned reason. I think we should look into #5602 to amend that properly.

## Testing notes:
Publishing model from Max should work as before when `Extract obj` is switched on. The validator for USD Plugin should be optional too.

### DIFF
--- a/openpype/hosts/max/plugins/publish/extract_model_obj.py
+++ b/openpype/hosts/max/plugins/publish/extract_model_obj.py
@@ -3,6 +3,7 @@ import pyblish.api
 from openpype.pipeline import publish, OptionalPyblishPluginMixin
 from pymxs import runtime as rt
 from openpype.hosts.max.api import maintained_selection
+from openpype.pipeline.publish import KnownPublishError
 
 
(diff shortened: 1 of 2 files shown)


moonyuet:
image
I need to disable the usd extractor and validator usd plugin for the valid publish.
Otherwises looks good. Maybe we can also implement what we come up in #5602 at here too?

LiborBatek (approved):
Publish producing ```OBJ``` model repre.

image

Also been tested via ```Loader``` and all seems normal and working ok.
image

PR #5606, opened 2023-09-12
Maya: problem ppublishing camera from workfile template builder

friquette:
## Changelog Description
When importing camera from template workfile builder, there's an error at the publishing because the maya commands take the placeholder as a camera and publish it instead of the actual camera.

## Additional info
The maya command cmds.ls(camera, type="camera") doen't work properly if any other flag is set in this command. With any other flag, the command ignore completely the "type" flag.

## Testing notes:
1. Import a camera with the template build workfile
2. Animate the camera (or camera rig)
3. Publish the camera
4. Load the camera (you will see that the loaded camera lost its animation)

### DIFF
--- a/openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
+++ b/openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
@@ -133,8 +133,7 @@ class ExtractCameraMayaScene(publish.Extractor):
         # get cameras
         members = cmds.ls(instance.data['setMembers'], leaf=True, shapes=True,
                           long=True, dag=True)
-        cameras = cmds.ls(members, leaf=True, shapes=True, long=True,
-                          dag=True, type="camera")
+        cameras = cmds.ls(members, type="camera")
 
         # validate required settings
         assert isinstance(step, float), "Step must be a float value"
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -35,6 +35,7 @@ body:
       label: Version
       description: What version are you running? Look to OpenPype Tray
       options:
+        - 3.16.5
         - 3.16.5-nightly.5
         - 3.16.5-nightly.4
         - 3.16.5-nightly.3
@@ -134,7 +135,6 @@ body:
         - 3.14.9-nightly.3
         - 3.14.9-nightly.2
         - 3.14.9-nightly.1
-        - 3.14.8
     validations:
       required: true
   - type: dropdown


openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
```diff
@@ -133,8 +133,7 @@ def process(self, instance):
         # get cameras
         members = cmds.ls(instance.data['setMembers'], leaf=True, shapes=True,
                           long=True, dag=True)
-        cameras = cmds.ls(members, leaf=True, shapes=True, long=True,
-                          dag=True, type="camera")
+        cameras = cmds.ls(members, type="camera")
```
BigRoy, line 136: Please keep `long=True`

LiborBatek:
@friquette could you more elaborate on how you setup your camera in ```template file1``` because this doesnt happen when just using ```camera``` object created in it.

Works normally and animation get published later on...

thx

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
friquette: If I keep `long=True` it will be the same issue as before. If there's an other flag than the `type="camera"` the command will not work properly.

LiborBatek:
@friquette ok and when speaking of Publishing the camera from current workfile...how u approach that? Just selecting that CameraShape (I mean transform but just the object acting as Camera) ?

openpype/hosts/maya/plugins/publish/extract_camera_mayaScene.py
BigRoy: `long=True` should work fine. It's the combination of `dag`, `leaf` or `shapes` causing the actual issue. I believe it's `dag`.

LiborBatek:
@friquette any info on how you output the camera from your anim scene? which node u use for ```Camera``` publishing...

OlivierOsotimehin:
Hello i can reply : since we create in advance in the template scene the instance camera to be publish , we add to the maya set the parent group in which the camera is. 
So in the screenshot above we add the grp_cam to the set. 

Otherwise if we are force to put just the camera transform, the template build is useless : for each layout / anim shot the animators have to create themselves the set to publish the camera. It's not that much for sure but it's something less for them to learn, to do wrong. And it's not a useful skill.

BigRoy:
@OlivierOsotimehin your use case makes sense - but I'd still say the issue then is in the template build system. The locator should not be in the camera instance set but the camera should. I'd say it's not an issue of the camera family - it's restrictive on purpose.

Is there a reason you have not enabled the settings to "delete the workfile template placeholders on a build" since I suppose it'd make a cleaner scene and also avoid the issue you're facing? Are you doing anything with the placeholders afterwards?

OlivierOsotimehin:
i would like to activate delete template placeholder, but when i do that, the system doesn't work afterwards (asset stop beeing loaded ,etc )  in 3.15 , so for now we are forced to keep it but i have no real need to keep the placeholder.

If the camera family is so strict that it can't accept locator that mean that  it would failed also if there was locator driving the rig of the cam.
In fact, in our camera rig we have curve to preview the different format that we are now forced to produce : 16/9 9/16 4/5 etc. 
The camera publish failed if the camera is the parent of curves. I did'nt expected that 

When we want to publish a camera family that mean for me : take the camera rig and publish a clean camera . 
Otherwise we would be using cameraRig family

BigRoy:
> When we want to publish a camera family that mean for me : take the camera rig and publish a clean camera .

Yes, you provide the camera to the camera instance - so the `camera` should be in the camera instance (not the locator). From there on out it should be able to solely extract the camera, excluding children. Whether it does that I'm not sure; the extractor also exports a mayaAscii where this logic to exclude children is way more involved since maya scene exports would ALWAYS include children unless explicilty unparented in the source scene (something that's not done easily).

The idea is indeed that the camera instance object set holds a camera - and then it'd only extract that camera ignoring other data. If there are cases where it fails where you've explicitly added only the camera in the publish instance then **that** is an issue of its own that we should fix.

> If the camera family is so strict that it can't accept locator that mean that it would failed also if there was locator driving the rig of the cam.

Not necessarily - the camera should be in the publish instance so that _what we export_ is really only the camera and not other shapes. If the locator isn't in the instance then it shouldn't export that either and if it then still fails on that I suppose it's a bug like the curve issue you mentioned.

OlivierOsotimehin:
To do that cleanly i guess we need tools to do actions on thing after they have been add to the scene by the scene builder. In the means time i have hopping that the pipeline code will magically find my camera in the hierarchy i give him : ) 

I will manage the camera failing publish validation by rearranging our camera rig and removing the curve parented to it . For the place holder locator for now i cannot do anything

BigRoy:
> I will manage the camera failing publish validation by rearranging our camera rig and removing the curve parented to it . For the place holder locator for now i cannot do anything

Great. I understand the production use case and need for your production fixes. I'd however like to keep this PR to `develop` open (or have a dedicated issue) to solve the root of the issue instead of this hotfix which _can_ cause issues elsewhere?

LiborBatek:
@OlivierOsotimehin thanks for your indepth info...I do agree that  ```Camera``` instance being somewhat limiting but as @BigRoy noted its on purpose by design choices made.

We are implementing new instance family ```matchMove``` in [this PR](https://github.com/ynput/OpenPype/pull/5445) which should allow user to include more than just solely ```Camera``` also allowing to have input/output connections present. There is ongoing discussion what kind of data should be included tho and if any of those changes should be incorporated also into present ```Camera``` family too or not.

tokejepsen:
@friquette @OlivierOsotimehin what is the status of this PR? Still needed?

LiborBatek:
@tokejepsen I think its not relevant because the way it was used (production practice described by @friquette ) is not corelating to the designed workflow as we having in OP now, so irrelevant ATM

mkolar:
I agree, this should be closed. @friquette we're closing this PR as it is now stale and it doesn't really align with the rest of the workflow.

PR #5607, opened 2023-09-12
Houdini: Make increment current file optional

MustafaJafar:
## Changelog Description
This PR makes `HoudiniIncrementCurrentFile` publish plugin optional and it's `on` by default 

image

## Testing notes:
1. toggle `Increment Current File` setting on and off


BigRoy:
Should we not make this optional by default to preserve existing behavior? Or what's the reason we want it to be optional?

MustafaJafar:
> Should we not make this optional by default to preserve existing behavior? 

This PR makes it optional and I think it preserves the existing behavior as it's on by default

> what's the reason we want it to be optional?

I don't know the best way to describe it but incrementing Houdini file messed up all my testing files and it became so confusing to find a particular file even while using artist notes. 

I think incrementing file is a mandatory in production but it kills me when testing :(

BigRoy:
Ah, yes - I see the `optional = True` in the old code too - ignore me. :)

> I don't know the best way to describe it but incrementing Houdini file messed up all my testing files and it became so confusing to find a particular file even while using artist notes.

Mess up as in - hard to remember which file was which? or did it actually break something for your scenes itself?

MustafaJafar:
> Mess up as in - hard to remember which file was which? or did it actually break something for your scenes itself?

It doesn't break anything. 
It just makes it hard to remember, 
I like building things gradually and 
Saving new versions can be quite confusing

moonyuet:
Maybe itis also good idea to include the increment current file in publisher?
I am open to have it set optional. But by my experience, I personally do increment and save for version by versions from time to time to avoid the possible crash in the project, especially when I work for relatively big scene such as filp fluid or rigid body, There is no guarantee that crashing the houdini scene when I am working on the fracturing scene in progress and I click something wrong with the loop nodes. I would save versions from time to time, and I can always go back to the previous version to continue my work. 
The files with incrementing versions are also good for communication (and life-savers too) when somebody needs to work on your files. If the files you sent are not working or not the updated version or that somebody prefers the previous version, you can always send them the correct files.

MustafaJafar:
I believe it's crucial for production to always increment current file! 
I know there's no reason to make this optional other than it bothers me a lot (I'm working solely in my home studio)

> I would save versions from time to time, and I can always go back to the previous version to continue my work.
This is a good idea! I will do that if we agreed to close this PR XD 

> Maybe it is also good idea to include the increment current file in publisher?
sorry, I did get this.

moonyuet:
> I believe it's crucial for production to always increment current file! I know there's no reason to make this optional other than it bothers me a lot (I'm working solely in my home studio)
> 
> > I would save versions from time to time, and I can always go back to the previous version to continue my work.
> 
> This is a good idea! I will do that if we agreed to close this PR XD
> 
> > Maybe it is also good idea to include the increment current file in publisher?
> 
> sorry, I did get this.

Sorry, I saw it just now. You did implement that in the code....Got a friday's brain

MustafaJafar:
Turning off publishing the workfile effectively stops the file from incrementing. 

image

MustafaJafar:
Turning off publishing the workfile effectively stops the file from incrementing. 

image

PR #5613, opened 2023-09-13
Extract Review: Multilayer specification for ffmpeg

iLLiCiTiT:
## Changelog Description
Extract review is specifying layer name when exr is multilayer.

## Additional info
This issue was discovered with renders from blender where output exr had channels `['Main.AOV.R', 'Main.AOV.G', 'Main.AOV.B', 'Main.AOV.A', 'Main.Combined.R', 'Main.Combined.G', 'Main.Combined.B', 'Main.Combined.A', 'Main.Normal.X', 'Main.Normal.Y', 'Main.Normal.Z', 'Main.Test.X']` which failed in ffmpeg.

## Testing notes:
1. First of all, any other current exr processing should still work (e.g. multipart exrs from maya)
2. Render from blender (testing for @simonebarbieri )

### DIFF
--- a/openpype/lib/transcoding.py
+++ b/openpype/lib/transcoding.py
@@ -315,6 +315,92 @@ def parse_oiio_xml_output(xml_string, logger=None):
     return output
 
 
+def get_channels_info_by_layer_name(channel_names):
+    """Get channels info grouped by layer name.
+
+    Finds all layers in channel names and returns list of dictionaries with
+    information about channels in layer.
+    Example output (not real world example):
+        [
+            {
+                "name": "Main",
+                "review_channels": {
+                    "R": "Main.red",
+                    "G": "Main.green",
+                    "B": "Main.blue",
+                    "A": None,
+                },
+                "channel_names": {
+                    "Main.red": "red",
+                    "Main.green": "green",
+                    "Main.blue": "blue",
+                    "Main.z": "z"
+                }
+            },
+            {
+                "name": "Composed",
+                "review_channels": {
+                    "R": "Composed.R",
+                    "G": "Composed.G",
+                    "B": "Composed.B",
+                    "A": "Composed.A",
+                },
+                "channel_names": {
+                    "Composed.R": "R",
+                    "Composed.G": "G",
+                    "Composed.B": "B",
+                    "Composed.A": "A",
+                }
+            },
+            ...
+        ]
+
+    Args:
+        channel_names (list[str]): List of channel names.
+
+    Returns:
+        list[dict]: List of channels information.
+    """
+
+    layer_names_order = []
+    rgba_by_layer_name = collections.defaultdict(dict)
+    channels_by_layer_name = collections.defaultdict(dict)
+    for channel_name in channel_names:
+        name_parts = channel_name.split(".")
+        last_part = name_parts.pop(-1)
+        layer_name = ".".join(name_parts)
+        channels_by_layer_name[layer_name][channel_name] = last_part
+        rgb_part = last_part.lower()
+        if layer_name not in layer_names_order:
+            layer_names_order.append(layer_name)
+        if rgb_part in ("r", "red"):
+            rgba_by_layer_name[layer_name]["R"] = channel_name
+        elif rgb_part in ("g", "green"):
+            rgba_by_layer_name[layer_name]["G"] = channel_name
+        elif rgb_part in ("b", "blue"):
+            rgba_by_layer_name[layer_name]["B"] = channel_name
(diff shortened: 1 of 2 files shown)


openpype/lib/transcoding.py
```diff
@@ -315,6 +315,92 @@ def parse_oiio_xml_output(xml_string, logger=None):
...
+        if layer_name not in layer_names_order:
+            layer_names_order.append(layer_name)
+        if rgb_part in ("r", "red"):
+            rgba_by_layer_name[layer_name]["R"] = channel_name
+        elif rgb_part in ("g", "green"):
+            rgba_by_layer_name[layer_name]["G"] = channel_name
+        elif rgb_part in ("b", "blue"):
+            rgba_by_layer_name[layer_name]["B"] = channel_name
+        elif rgb_part in ("a", "alpha"):
+            rgba_by_layer_name[layer_name]["A"] = channel_name
```
BigRoy, line 383: I wonder if something like this would make more sense:
```suggestion
        match = re.match(r"^(.*\).(r|red|g|green|b|blue|a|alpha)$", flags=re.IGNORECASE)
        if match:
            # ends with rgba identifier, e.g. main.red or main.A
            layer_name = match.group(0)
            rgb_part = match.group(1)
            rgb_channel = rgb_part[0].upper()  # R, G, B or A
            rgba_by_layer_name[layer_name][rgb_channel] = channel_name
        else:
            # do whatever if it does not end with rgba identifier
```

BigRoy: Or just:
```suggestion
        if "." in channel_name:
            layer_name, last_part = channel_name.rsplit(".", 1)
        else:
            layer_name = ""
            last_part = channel_name

        channels_by_layer_name[layer_name][channel_name] = last_part
        if layer_name not in layer_names_order:
            layer_names_order.append(layer_name)

        if last_part.lower() in {"r", "red", "g", "green", "b", "blue", "a", "alpha"}:
            channel = last_part[0].upper() # R, G, B or A
            rgba_by_layer_name[layer_name][channel] = channel_name
```

openpype/lib/transcoding.py
```diff
@@ -315,6 +315,92 @@ def parse_oiio_xml_output(xml_string, logger=None):
...
+
+    if "" in layer_names_order:
+        layer_names_order.remove("")
+        layer_names_order.insert(0, "")
```
BigRoy, line 387: So what's this? We want to prioritize channels that do not have one of the RGBA suffixes? Note that there could be muliple channels with "" layer name here if inputs are:
```python
channel_names = ["red", "green", "blue", "z", "custom_aov"]
```

openpype/plugins/publish/extract_review.py
```diff
@@ -599,6 +619,10 @@ def _ffmpeg_arguments(
 
         duration_seconds = float(output_frames_len / temp_data["fps"])
 
+        # Define which layer should be used
+        if layer_name:
+            ffmpeg_input_args.extend(["-layer", layer_name])
```
BigRoy, line 624: Am I correct in understanding that if the input has: `main.r`, `main.g` and `main.b` that here we should be passing it `main`?

openpype/lib/transcoding.py
```diff
@@ -315,6 +315,92 @@ def parse_oiio_xml_output(xml_string, logger=None):
...
+        output.append({
+            "name": layer_name,
+            "review_channels": {
+                "R": rgba_layer_info.get("R"),
+                "G": rgba_layer_info.get("G"),
+                "B": rgba_layer_info.get("B"),
+                "A": rgba_layer_info.get("A"),
+            },
+            "channel_names": channels_by_layer_name[layer_name],
+        })
```
BigRoy, line 400: It seems the `channel_names` returned data is unused. I feel like we might get an easier method if we just call this function: `iter_review_channels` and it'd just yield the review channels with the name?

openpype/plugins/publish/extract_review.py
iLLiCiTiT: yes

iLLiCiTiT: This PR is because of blender, so it is not much surprise I've got this information from blender forum https://blenderartists.org/t/converting-blenders-exr-to-video-with-ffmpeg/1290550/6 .

kalisp (approved):
This doesn't seem to break anything in automatic tests, but I don't think that tests would be actually testing this.

iLLiCiTiT:
We also want to add option to be able define preferred layer names to be used for review, but that's for another PR.

Issue #4641, opened 2023-03-16
Implement Rez for Ayon


BigRoy:
**Is your feature request related to a problem? Please describe.**

Ayon and even OpenPype had to handle dependencies, versions of packages and distributing vendored binaries. With Ayon resolving versions and its dependencies becomes more dynamic due to being able to mix and match addons which in itself can require specific versions of other addons.

As a part of this some of the work for Ayon has been involving resolving these dependencies and building an environment for launching a specific version of Ayon, etc. and has with OpenPype (v3) involved managing environments for Applications and Tools when launching even specific assets in a project.

This addon (package) management and dependency resolving with version ranges is exactly what [Rez](https://github.com/AcademySoftwareFoundation/rez) is built to solve and I feel like before we set of implementing complex resolving algorithms for the addons and their 'dependencies' we should maybe avoid time spent there and instead spent time on relying on Rez as much as we can.

By adopting Rez we can also get other benefits like:
- Setting up a central repository of package versions (e.g. ffmpeg, ocio, otio, python packages) without requiring each build of Ayon to embed all those dependencies.
- It can support local caching of having the dependencies on a network share but copy certian packages (or all) locally to improve runtime performance. This would then also automatically become possible for things like "tools" where e.g. redshift could be loaded through ayon configuration and its resolving could use that Redshift locally to improve the startup running time for the renderer.
- Since Rez is built to resolve dependencies like this we can use its feature set to also help debug dependency problems, etc. without again having to re-write and manage our own.
- It can even help the "version resolving" we are using for OpenPype itself currently by turning OpenPype ZIP packages into rez packages, etc. (e.g. at deployment time or alike) to avoid maintaining our own version resolving logic and be able to rely on Rez's mechanisms which have a much stronger backbone in production already.

**Describe the solution you'd like**

Have Ayon use `rez` to resolve the environment to launch an application in would be one first step.

I'd say the final goal would be to be able to run even Ayon itself through rez so from there you could e.g. pick a particular Ayon version to start with too - but that doesn't have to be a first priority I'd say.

**Describe alternatives you've considered**

We've had Applications and Tools in OPv3 but with separating all addons for Ayon and them versioning independently with dependencies (requirements) to other addons will start complicating the version resolving logic that needs to be written.

I'd say we are getting to a point where Rez might be best way forward into development time and what we gain from it in the end.

**Additional context**

### Rez feature requests have come up before

[...]

PR #5614, opened 2023-09-13
Fussion: added support for Fusion 17

kalisp:
## Changelog Description
Fusion 17 still uses Python 3.6 which causes issues with some our delivered libraries. Vendorized necessary set for Python 3.6

### Additional Information
It should be tested in Fusion 18 too, but it is explicitly adding older libraries for versions lower than 18.
 
## Testing notes:
1. start Fusion 17
2. open Openpype Menu - should show up
3. (you could check console output in `View > Console`


openpype/hosts/fusion/addon.py
```diff
@@ -60,8 +60,18 @@ def get_launch_hook_paths(self, app):
             return []
         return [os.path.join(FUSION_HOST_DIR, "hooks")]
 
-    def add_implementation_envs(self, env, _app):
+    def add_implementation_envs(self, env, app):
         # Set default values if are not already set via settings
+        app_version = app.name
+        if int(app_version) < 18:
```
BigRoy, line 66: Be aware that app names can be customized (they can literally be any string). That discussion also came up when implementing [this](https://github.com/ynput/OpenPype/blob/00b83e58792db4e33d5d218ce942a8667d724265/openpype/hosts/fusion/addon.py#L20-L47) in this PR: https://github.com/ynput/OpenPype/pull/4572

If anyone had a `18.5.beta` version registered this would currently just error on it.

Additionally, Fusion 18 can also still use Python 3.6. With Fusion 18 I believe they only added the possibility to a Python between 3.3 and 3.10 as Python 3 interpreter. If so, this issue could be just as relevant for Fusion 18+

kalisp: Now used `get_fusion_version`.

What would be the safer option? Trying to import `requests` in `openpype/hosts/fusion/deploy/MenuScripts/openpype_menu.py` and modify environment only if it fails there? That doesn't look particularly elegant.

BigRoy: The best solution would be packaging the right tools for the right python interpreters - e.g. using Rez #4641 

A hacky solution could indeed be hot-patching it within the interpreter, but it'd only fix the Python session if we were to run OpenPype tools - but other Py3 scripts Fusion would then run that might use a similar dependency would then still break. So it's not as nice as making sure we're not adding the wrong packages to Fusion's Python.

openpype/hosts/fusion/addon.py
```diff
@@ -60,8 +60,18 @@ def get_launch_hook_paths(self, app):
             return []
         return [os.path.join(FUSION_HOST_DIR, "hooks")]
 
-    def add_implementation_envs(self, env, _app):
+    def add_implementation_envs(self, env, app):
         # Set default values if are not already set via settings
+        app_version = get_fusion_version(app.name)
+        if int(app_version) < 18:
```
BigRoy, line 66:
```suggestion
        if app_version and app_version < 18:
```

openpype/hosts/fusion/addon.py
```diff
@@ -60,8 +60,18 @@ def get_launch_hook_paths(self, app):
...
+            python_path = env["PYTHONPATH"]
```
BigRoy, line 67: Let's also add a comment explaining what we're trying to do here - plus maybe add a `# todo` to clean it up later whenever we can or have better packaging tooling available for OpenPype/AYON's launching architecture.

m-u-r-p-h-y (approved):
tested with all combinations and with app strings as Fusion17/Fusion18

- [x] Fusion17 with Python 3.6 
- [x] Fusion 18 with Python 3.6
- [x] Fusion 18 with Python 3.9

openpype/hosts/fusion/deploy/MenuScripts/openpype_menu.py
```diff
@@ -1,6 +1,17 @@
...
+if sys.version_info < (3, 7):
+    from openpype import PACKAGE_DIR
+    FUSION_HOST_DIR = os.path.join(PACKAGE_DIR, "hosts", "fusion")
+
+    vendor_path = os.path.join(FUSION_HOST_DIR, "vendor")
+    print("openpype_menu vendor_path::{}".format(vendor_path))
+    if vendor_path not in sys.path:
+        sys.path.insert(0, vendor_path)
+
+    print(f"Added vendorized libraries from {vendor_path}")
```
BigRoy, line 13: Should we add a comment as to what this is doing and why?

BigRoy:
@m-u-r-p-h-y would actually be good to _also_ test it with a Python 3.9 version (or even Python 3.10) in Fusion 18 so that we can see if issues exist there too.

PR #5615, opened 2023-09-13
Publisher: Fix screenshot widget

iLLiCiTiT:
## Changelog Description
Use correct super method name.

EDITED:
Removed fade animation which is not triggered at some cases, e.g. in Nuke the animation does not start. I do expect that is caused by `exec_` on the dialog, which blocks event processing to the animation, even when I've added the window as parent it still didn't trigger registered callback.

Modified how the "empty" space is not filled by using paths instead of clear mode on painter. Added render hints to add antialiasing.

## Testing notes:
1. Use take a screenshot feature in publisher
2. Press any keyboard key
3. It should not crash


BigRoy (approved):
Code change makes sense :D

64qam (approved):
schmeckt gut

PR #5618, opened 2023-09-14
Maya: Missing "data" field and enabling of audio

tokejepsen:
## Changelog Description
When updating audio containers, the field "data" was missing and the audio node was not enabled on the timeline.

## Testing notes:
1. Load audio in Maya.
2. Publish another version of the audio.
3. Update audio in Maya.


openpype/hosts/maya/plugins/load/load_audio.py
```diff
@@ -61,6 +61,14 @@ def update(self, container, representation):
 
         path = get_representation_path(representation)
         cmds.setAttr("{}.filename".format(audio_node), path, type="string")
+
+        cmds.timeControl(
+            mel.eval("$tmpVar=$gPlayBackSlider"),
+            edit=True,
+            sound=audio_node,
+            displaySound=True
+        )
```
BigRoy, line 70: Not too sure about doing this on update - I feel like it'd just be interfering with the scene state without a real reason. In the past I had cases where I had multiple audio files loaded using the **Use Trax Sounds** option so it'd play multiple audio files instead of just the one.

image

Having this code here would mean it'd just be changing that setting everytime you update the audio file?

openpype/hosts/maya/plugins/load/load_audio.py
```diff
@@ -76,7 +84,7 @@ def update(self, container, representation):
             project_name, version["parent"], fields=["parent"]
         )
         asset = get_asset_by_id(
-            project_name, subset["parent"], fields=["parent"]
+            project_name, subset["parent"], fields=["parent", "data"]
```
BigRoy, line 87: This would also be a possibility, right? Or doesn't AYON support that?
```suggestion
            project_name, subset["parent"], fields=["parent", "data.frameStart", "data.frameEnd"]
```

openpype/hosts/maya/plugins/load/load_audio.py
tokejepsen: Thought I could be lazy about this, but you always catch me. Yeah, we'll need to check the audio node is used before updating.

BigRoy: We could just remove this right? If the user had already set it to the audio node (e.g. from the `load` itself and not changing it) then updating shouldn't remove it right since the audio node remains the same node?

tokejepsen: Changing the file path makes the audio node not active on the timeline.

openpype/hosts/maya/plugins/load/load_audio.py
```diff
@@ -76,7 +84,7 @@ def update(self, container, representation):
             project_name, version["parent"], fields=["parent"]
         )
         asset = get_asset_by_id(
-            project_name, subset["parent"], fields=["parent"]
+            project_name, subset["parent"], fields=["parent", "data.frameStart", "data.frameEnd"]
```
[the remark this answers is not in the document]
tokejepsen:
```suggestion
            project_name, subset["parent"],
            fields=["parent", "data.frameStart", "data.frameEnd"]
```

openpype/hosts/maya/plugins/load/load_audio.py
```diff
@@ -76,7 +84,8 @@ def update(self, container, representation):
             project_name, version["parent"], fields=["parent"]
         )
         asset = get_asset_by_id(
-            project_name, subset["parent"], fields=["parent"]
+            project_name, subset["parent"],
+            fields=["parent", "data.frameStart", "data.frameEnd"]
```
tokejepsen, line 88:
```suggestion
            project_name,
            subset["parent"],
            fields=["parent", "data.frameStart", "data.frameEnd"]
```

openpype/hosts/maya/plugins/load/load_audio.py
```diff
@@ -59,8 +59,24 @@ def update(self, container, representation):
         assert audio_nodes is not None, "Audio node not found."
         audio_node = audio_nodes[0]
 
+        current_sound = cmds.timeControl(
+            mel.eval("$tmpVar=$gPlayBackSlider"),
```
BigRoy, line 63: By the way - this `mel.eval` query should really be: `$gPlayBackSlider=$gPlayBackSlider`

The reason for that is that if you use `$tmpVar` more than once you might end up getting MEL "variable redeclaration errors" since if you also use `$tmpVar` for retrieving other global variables that if those other variables have a different type (say integer or float instead of string) then MEL will complain because it's statically typed and thus warns on the variable declaration.

tokejepsen:
```suggestion
            mel.eval("$gPlayBackSlider=$gPlayBackSlider"),
```

openpype/hosts/maya/plugins/load/load_audio.py
```diff
@@ -59,8 +59,24 @@ def update(self, container, representation):
...
+            sound=True
+        )
+        activate_sound = current_sound == audio_node
+
         path = get_representation_path(representation)
         cmds.setAttr("{}.filename".format(audio_node), path, type="string")
+
+        if activate_sound:
+            cmds.timeControl(
+                mel.eval("$tmpVar=$gPlayBackSlider"),
```
tokejepsen, line 74:
```suggestion
                mel.eval("$gPlayBackSlider=$gPlayBackSlider"),
```

openpype/hosts/maya/plugins/load/load_audio.py
```diff
@@ -76,7 +92,9 @@ def update(self, container, representation):
             project_name, version["parent"], fields=["parent"]
         )
         asset = get_asset_by_id(
-            project_name, subset["parent"], fields=["parent"]
+            project_name,
+            subset["parent"],
+            fields=["parent", "data.frameStart", "data.frameEnd"]
         )
 
         source_start = 1 - asset["data"]["frameStart"]
```
BigRoy, line 100: I wonder, instead of setting the attributes manually. Would it also work to do this?

```python
cmds.sound(audio_node, edit=True, offset=start_frame)
```

Also, is it technically weird that we set `sourceStart` and `sourceEnd` on the sound node from the asset's data here but we don't do that on `load`? Doesn't that mean that technically the sound node behaves differently after `update` than on `load`?

I have a feeling that updating to the SAME version might even start showing different offsets?

openpype/hosts/maya/plugins/load/load_audio.py
```diff
@@ -30,7 +30,7 @@ def load(self, context, name, namespace, data):
             file=context["representation"]["data"]["path"], offset=start_frame
```
BigRoy, line 30: Interesting approach - I have a feeling that this should actually be `self.filepath_from_context(context)` instead of just taking it from the repsesentation's data directly. I have a feeling that might provide issues since (OR nobody knew it's there?) since I haven't seen that data entry used before.

Does that still work with AYON too?

openpype/hosts/maya/plugins/load/load_audio.py
```diff
@@ -59,8 +59,24 @@ def update(self, container, representation):
...
+            query=True,
+            sound=True
+        )
+        activate_sound = current_sound == audio_node
+
         path = get_representation_path(representation)
         cmds.setAttr("{}.filename".format(audio_node), path, type="string")
+
+        if activate_sound:
+            cmds.timeControl(
```
BigRoy, line 73: Should we add a comment why we're doing this?
```suggestion
            # maya by default deactivates it from timeline on file change
            cmds.timeControl(
```

BigRoy (approved):
Commented some other notes just now that aren't necessarily related to this PR but if they are issues might as well get fixed here - other than that the code changes you've made make sense. Nice work.

openpype/hosts/maya/plugins/load/load_audio.py
tokejepsen: Think it might just old code maybe?
Will let @iLLiCiTiT weigh in on this.

iLLiCiTiT: It is as @BigRoy wrote. It should use `self.filepath_from_context(context)`. This would work only on the platform where it was published (and does not use root overrides).

BigRoy (approved):
Didn't test it but changes look good to me.

openpype/hosts/maya/plugins/load/load_audio.py
BigRoy: Since it's consistent with how  it was I feel like we can push it to a separate PR but we should definitely pursue consistent behavior between load and  updating the audio node.

tokejepsen: Would rather just work it into this PR cause I doubt I'll be doing another PR for this.

We can also just leave it to an issue (that wont be resolved :) )

LiborBatek (changes requested):
Did tests in ```maya 2023``` and it fails when triggering ```Update``` via Scene Inventory.


https://github.com/ynput/OpenPype/assets/112623825/9bed9654-a3b7-4467-8d70-638f9cf11c0a



Also tried simple ```Load``` audio asset and different versions and that works normally so its just issue when ```Update``` outdated assets action.

tokejepsen:
> Did tests in maya 2023 and it fails when triggering Update via Scene Inventory.

@LiborBatek I'm guessing no error messages?
I suspect it might be something to do with the source frame ranges. Could you check on the audio node, what the source range is before and after update?

LiborBatek:
> > Did tests in maya 2023 and it fails when triggering Update via Scene Inventory.
> 
> @LiborBatek I'm guessing no error messages? I suspect it might be something to do with the source frame ranges. Could you check on the audio node, what the source range is before and after update?

This is what I have:
image

I guess its missing completely or set to these weird values:

Source Start ```0```
Source End ```298.5```

Side note:  I have produced this audio asset via ```Tray Publisher``` within this particular ```shot```  before using it as an asset.

LiborBatek:
And here is scrngrab of attributes for that audio asset after performing ```Update to Latest```

image

Minkiu (approved):
Haven't tested, code looks good.

BigRoy:
@LiborBatek I've committed some tweaks. Could you try again?

Minkiu (approved):

LiborBatek (approved):
Now ```Update``` works fine. And audio asset get loaded and set to timeline correctly.

image

LiborBatek:
@BigRoy there is some unresolved conversation...still valid or should be merged evnt.?

BigRoy:
> @BigRoy there is some unresolved conversation...still valid or should be merged evnt.?

Should've been resolved with latest changes.

PR #5619, opened 2023-09-14
AYON: Ignore separated modules

iLLiCiTiT:
## Changelog Description
Do not load already separated modules from default directory.

## Additional info
The modules are not in client code when created using `create_ayon_addon.py` script, but for testing of PRs is usually copied whole openpype into local app data which causes troubles when these modules are loaded.

## Testing notes:
1. All ignored addons should be loaded in OpenPype (ftrack, shotgrid, slack, sync_server)
2. The same addons should not be loaded in AYON mode from OpenPype

### DIFF
--- a/openpype/modules/base.py
+++ b/openpype/modules/base.py
@@ -401,7 +401,7 @@ def _load_ayon_addons(openpype_modules, modules_key, log):
         addon_dir = os.path.join(addons_dir, folder_name)
         if not os.path.exists(addon_dir):
             log.debug((
-                "Addon {} {} does not have available code."
+                "No localized client code found for addon {} {}."
             ).format(addon_name, addon_version))
             continue
 
--- a/openpype/modules/base.py
+++ b/openpype/modules/base.py
@@ -400,9 +400,9 @@ def _load_ayon_addons(openpype_modules, modules_key, log):
         folder_name = "{}_{}".format(addon_name, addon_version)
         addon_dir = os.path.join(addons_dir, folder_name)
         if not os.path.exists(addon_dir):
-            log.warning((
-                "Directory for addon {} {} does not exists. Path \"{}\""
-            ).format(addon_name, addon_version, addon_dir))
+            log.debug((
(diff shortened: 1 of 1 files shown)


openpype/modules/base.py
```diff
@@ -59,6 +59,14 @@
     "example_addons",
     "default_modules",
 )
+# Modules that won't be loaded in AYON mode from "./openpype/modules"
+# - the same modules are ignored in "./server_addon/create_ayon_addons.py"
+IGNORED_FILENAMES_IN_AYON = {
```
antirotor, line 64: not sure if this shouldn't be
```suggestion
IGNORED_MODULE_NAMES_IN_AYON = {
```

openpype/modules/base.py
```diff
@@ -392,9 +400,9 @@ def _load_ayon_addons(openpype_modules, modules_key, log):
         folder_name = "{}_{}".format(addon_name, addon_version)
         addon_dir = os.path.join(addons_dir, folder_name)
         if not os.path.exists(addon_dir):
-            log.warning((
-                "Directory for addon {} {} does not exists. Path \"{}\""
-            ).format(addon_name, addon_version, addon_dir))
+            log.debug((
+                "Addon {} {} does not have available code."
```
antirotor, line 404: I think previous error was more clear for me. I am not sure what the addon without available code is.

antirotor (changes requested):
Otherwise it looks fine and works

openpype/modules/base.py
iLLiCiTiT:
```suggestion
                "No localized client code found for addon {} {}."
```

openpype/modules/base.py
iLLiCiTiT: Those are actually filenames.

antirotor: ...ok

antirotor (approved):

PR #5621, opened 2023-09-14
Houdini: Add self publish button

MustafaJafar:
## Changelog Description
This PR allow publishing single instance from publish rop nodes in Houdini 
image
image

## Additional info1
This PR was made to prototype Fabia's idea about publishing one instance in houdini, [discord discussion](https://discord.com/channels/517362899170230292/517382145552154634/1151589724914454679)

## Additional info2 
I made this PR as a prototype. 
and I have no idea if it will be beneficial to the production or cause confusion for the artists. 

## Testing notes:
1. Create multiple publish instances in houdini 
2. try publishing from the button

### DIFF
--- a/openpype/hosts/houdini/api/lib.py
+++ b/openpype/hosts/houdini/api/lib.py
@@ -649,3 +649,65 @@ def get_color_management_preferences():
         "display": hou.Color.ocio_defaultDisplay(),
         "view": hou.Color.ocio_defaultView()
     }
+
+
+def publisher_show_and_publish():
+    """Open publisher window and trigger publishing action. """
+
+    from openpype.tools.utils.host_tools import get_tool_by_name
+
+    main_window = get_main_window()
+    publisher_window = get_tool_by_name(
+        tool_name="publisher",
+        parent=main_window
+    )
+
+    publisher_window.set_current_tab("publish")
+    publisher_window.make_sure_is_visible()
+    publisher_window._reset_on_show = False
+
+    publisher_window._controller.reset()
+    publisher_window._controller.publish()
+
+
+def self_publish():
+    """Self publish from ROP nodes. """
+    from openpype.pipeline import registered_host
+    from openpype.pipeline.create import CreateContext
+
+    current_node = hou.node(".").path()
+
+    host = registered_host()
+    context = CreateContext(host, reset=True)
+
+    for instance in context.instances:
+        node_path = instance.data.get("instance_node")
+        if not node_path:
+            continue
+        print(node_path)
+
+        if current_node == node_path:
+            instance["active"] = True
+        else:
+            instance["active"] = False
+
+    context.save_changes()
+    publisher_show_and_publish()
+
+def add_self_publish_button(node):
+    """Adds a self publish button in the rop node. """
+    label = os.environ.get("AVALON_LABEL") or "OpenPype"
+
+    button_parm = hou.ButtonParmTemplate(
+        "{}_publish".format(label.lower()),
+        "{} Publish".format(label),
+        script_callback="from openpype.hosts.houdini.api.lib import "
+                        "self_publish; self_publish()",
+        script_callback_language=hou.scriptLanguage.Python,
+        join_with_next=True
+    )
+
+    template = node.parmTemplateGroup()
+    template.insertBefore((0,), button_parm)
+    # parm_group.append(button_parm)
+    node.setParmTemplateGroup(template)
--- a/openpype/hosts/houdini/api/plugin.py
+++ b/openpype/hosts/houdini/api/plugin.py
@@ -13,7 +13,7 @@ from openpype.pipeline import (
     CreatedInstance
 )
 from openpype.lib import BoolDef
-from .lib import imprint, read, lsattr
+from .lib import imprint, read, lsattr, add_self_publish_button
 
 
 class OpenPypeCreatorError(CreatorError):
@@ -194,6 +194,7 @@ class HoudiniCreator(NewCreator, HoudiniCreatorBase):
                 self)
             self._add_instance_to_context(instance)
             imprint(instance_node, instance.data_to_store())
+            add_self_publish_button(instance_node)
             return instance
 
         except hou.Error as er:


openpype/hosts/houdini/api/lib.py
```diff
@@ -649,3 +649,66 @@ def get_color_management_preferences():
...
+    current_node = hou.node(".").path()
+
+    host = registered_host()
+    context = CreateContext(host, reset=True)
+
+    for instance in context.instances:
+        node_path = instance.data.get("instance_node")
+        if not node_path:
+            continue
+        print(node_path)
```
BigRoy, line 687:
```suggestion
```

openpype/hosts/houdini/api/lib.py
```diff
@@ -649,3 +649,66 @@ def get_color_management_preferences():
...
+
+        if current_node == node_path:
+            instance["active"] = True
+        else:
+            instance["active"] = False
```
BigRoy, line 692: Can be simplified to:
```suggestion
        instance["active"] = current_node == node_path
```

Or if you find that unreadable/unclear:
```suggestion
        active = current_node == node_path
        instance["active"] = active
```

openpype/hosts/houdini/api/lib.py
```diff
@@ -649,3 +649,66 @@ def get_color_management_preferences():
         "display": hou.Color.ocio_defaultDisplay(),
         "view": hou.Color.ocio_defaultView()
     }
+
+
+def publisher_show_and_publish():
+    """Open publisher window and trigger publishing action. """
```
BigRoy, line 655:
```suggestion
    """Open publisher window and trigger publishing action."""
```

openpype/hosts/houdini/api/lib.py
```diff
@@ -649,3 +649,66 @@ def get_color_management_preferences():
...
+
+    from openpype.tools.utils.host_tools import get_tool_by_name
```
BigRoy, line 657: Why is this import here and not at the top?

openpype/hosts/houdini/api/lib.py
```diff
@@ -649,3 +649,66 @@ def get_color_management_preferences():
...
+    publisher_window._reset_on_show = False
+
+    publisher_window._controller.reset()
+    publisher_window._controller.publish()
+
+
+def self_publish():
+    """Self publish from ROP nodes. """
+    from openpype.pipeline import registered_host
+    from openpype.pipeline.create import CreateContext
```
BigRoy, line 676: Why are these imports here and not at the top?

openpype/hosts/houdini/api/lib.py
```diff
@@ -649,3 +649,66 @@ def get_color_management_preferences():
...
+        "{} Publish".format(label),
+        script_callback="from openpype.hosts.houdini.api.lib import "
+                        "self_publish; self_publish()",
+        script_callback_language=hou.scriptLanguage.Python,
+        join_with_next=True
+    )
+
+    template = node.parmTemplateGroup()
+    template.insertBefore((0,), button_parm)
+    # parm_group.append(button_parm)
```
BigRoy, line 713:
```suggestion
```

openpype/hosts/houdini/api/plugin.py
```diff
@@ -194,6 +194,7 @@ def create(self, subset_name, instance_data, pre_create_data):
                 self)
             self._add_instance_to_context(instance)
             imprint(instance_node, instance.data_to_store())
+            add_self_publish_button(instance_node)
```
BigRoy, line 197: If this were to be merged at some point I feel like this might be something that studios want to disable / hide away and thus potentially expose this in studio settings. In the meantime however I think let's first see what everyone thinks of the PR / implementation in general.

openpype/hosts/houdini/api/lib.py
MustafaJafar: No reason actually.

openpype/hosts/houdini/api/plugin.py
MustafaJafar: I will keep this one unresolved as a reminder!

fabiaserra:
as I said on Discord, could we also expose a string parm called "Comment" (or a pop-up dialog after clicking Publish) where they can  type the comment corresponding to that publish?

fabiaserra:
I like it! Thanks for doing this! I will test it soon

openpype/hosts/houdini/api/lib.py
```diff
@@ -649,3 +655,71 @@ def get_color_management_preferences():
...
+
+    main_window = get_main_window()
+    publisher_window = get_tool_by_name(
+        tool_name="publisher",
+        parent=main_window
+    )
+
+    publisher_window.set_current_tab("publish")
+    publisher_window.make_sure_is_visible()
+    publisher_window._reset_on_show = False
```
iLLiCiTiT, line 671: I'm strongly against any usage of private functions or attributes. They're private, so they can be changed in future without checking if are used.

openpype/hosts/houdini/api/lib.py
```diff
@@ -649,3 +655,71 @@ def get_color_management_preferences():
...
+
+def self_publish():
+    """Self publish from ROP nodes."""
+
+    current_node = hou.node(".").path()
+
+    host = registered_host()
+    context = CreateContext(host, reset=True)
+
+    for instance in context.instances:
```
iLLiCiTiT, line 685: This loop won't disable other instances, so if will publish more than just selected instance, or?

openpype/hosts/houdini/api/lib.py
```diff
@@ -649,3 +655,71 @@ def get_color_management_preferences():
...
+        instance["active"] = active
+
+        result, comment = hou.ui.readInput(
+            "Add Publish Note",
+            buttons=("Ok", "Cancel"),
+            title="Publish Note",
+            close_choice=1
+        )
+        if not result:
+            instance.data["comment"] = comment
```
iLLiCiTiT, line 703: With this you will set the comment on instance, but the comment value will stay there forever and user won't have any option to remove it.

MustafaJafar: are there any alternatives ?

openpype/hosts/houdini/api/lib.py
MustafaJafar: Fair enough! 
I will try finding another way.

iLLiCiTiT: I can add the option if you tell me what you want to do...

openpype/hosts/houdini/api/lib.py
iLLiCiTiT: It's marked as resolved, but nothing changed?

openpype/hosts/houdini/api/lib.py
iLLiCiTiT: Don't use per-instance comment.

openpype/hosts/houdini/api/lib.py
MustafaJafar: omg, I changed it locally but I didn't push my commit sorry

openpype/hosts/houdini/api/lib.py
MustafaJafar: I was coping from this [code snippet](https://discord.com/channels/517362899170230292/517382145552154634/1151589724914454679) on discord. 

how should we trigger the publish action ?

openpype/hosts/houdini/api/lib.py
MustafaJafar: how can I set this comment from my python code? 
image

openpype/hosts/houdini/api/lib.py
iLLiCiTiT: what about these?
```
if not node_path:
    continue
```

openpype/hosts/houdini/api/lib.py
BigRoy: Without that variable set @fabiaserra said the publish would still be interrupted by the UI because it resets on show ;)

openpype/hosts/houdini/api/lib.py
MustafaJafar: There's an instance that hasn't a `node_path`, this `instance.data.get("instance_node")`  equals `None`
so I skipped it as I deactivate instances using `node_path`

BigRoy: What @iLLiCiTiT is trying to say is those nodes without `instance_node`, like the workfile instance will not be deactivated even though the user expects likely only that single instance they chose to be publishes.

I agree with @iLLiCiTiT that we're better off always setting active = False unless it's the instance we want.

openpype/hosts/houdini/api/lib.py
MustafaJafar: I found that input text. I will push commit shortly.

openpype/hosts/houdini/api/lib.py
MustafaJafar: it didn't work until I made this 
```python
active = current_node == node_path
instance["active"] = active
hou.node(node_path).parm("active").set(active)
```

openpype/hosts/houdini/api/lib.py
```diff
@@ -657,21 +657,22 @@ def get_color_management_preferences():
     }
 
 
-def publisher_show_and_publish():
+def publisher_show_and_publish(comment = ""):
```
BigRoy, line 660:
```suggestion
def publisher_show_and_publish(comment=""):
```

openpype/hosts/houdini/api/lib.py
```diff
@@ -689,19 +690,18 @@ def self_publish():
...
-        if active:
-            result, comment = hou.ui.readInput(
-                "Add Publish Note",
-                buttons=("Ok", "Cancel"),
-                title="Publish Note",
-                close_choice=1
-            )
-            if not result:
-                instance.data["comment"] = comment
+        hou.node(node_path).parm("active").set(active)
```
BigRoy, line 693: This can still be removed, right?
```suggestion
```

openpype/tools/publisher/window.py
```diff
@@ -388,6 +388,20 @@ def __init__(self, parent=None, controller=None, reset_on_show=None):
...
+    @property
+    def reset_on_show(self):
+        return self._reset_on_show
+
+    @reset_on_show.setter
+    def reset_on_show(self, value):
+        self._reset_on_show = value
+
+    def set_comment_input_text(self, text=""):
+        self._comment_input.setText(text)
```
BigRoy, line 400: I'm pretty sure this should go through the controller - not just the UI element. @iLLiCiTiT what do you think?

openpype/hosts/houdini/api/lib.py
MustafaJafar: I'm afraid not. 
I tried this many times in Houdini console, here's the snippet. 
```python

current_node = hou.node('/out/pointcacheTestComment').path()

host = registered_host()
context = CreateContext(host, reset=True)

for instance in context.instances:
    node_path = instance.data.get("instance_node")
    if not node_path:
        continue

    active = current_node == node_path
    instance["active"] = active
    hou.node(node_path).parm("active").set(active)

context.save_changes()
```

openpype/tools/publisher/window.py
MustafaJafar: I think they are not alternatives 
image

Also, most code in publisher depends on `self._comment_input` when updating the comment field in the UI

openpype/hosts/houdini/api/lib.py
fabiaserra: yeah I'm also against accessing private variables, I just did that as a quick hack

openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +753,69 @@ def get_camera_from_container(container):
...
+    publisher_window.click_publish()
+
+
+def self_publish():
+    """Self publish from ROP nodes."""
+
+    current_node = hou.node(".").path()
+
+    host = registered_host()
+    context = CreateContext(host, reset=True)
```
BigRoy, line 782: Since we already have a CreateContext it would be nice if we could actually pass that on directly to the Publisher UI itself @iLLiCiTiTas we open it. Or is there good reason to not allow that? That could potentially even allow to define custom CreateContext classes (that discover plugins differently?) but use the same UI.

openpype/hosts/houdini/api/lib.py
BigRoy: Interesting. Will investigate, if it doesn't work without changing the actual hou parm ourselves I feel like that's a bug of its own. Let me unresolved this as I investigate it (mostly as a reminder to myself)

openpype/hosts/houdini/api/plugin.py
```diff
@@ -168,6 +168,7 @@ class HoudiniCreator(NewCreator, HoudiniCreatorBase):
     """Base class for most of the Houdini creator plugins."""
     selected_nodes = []
     settings_name = None
+    _add_self_publish_button = False
```
BigRoy, line 172: Is there a reason we prefix this with underscore but usually do not do so for similar attributes that get set from studio settings?

BigRoy: Also if you wouldn't prefix it I believe you don't need the changes in apply settings just for this attribute since then it matches by name already

openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +753,69 @@ def get_camera_from_container(container):
...
+        hou.node(node_path).parm("active").set(active)
+
+    context.save_changes()
+
+    result, comment = hou.ui.readInput(
+        "Add Publish Note",
+        buttons=("Ok", "Cancel"),
+        title="Publish Note",
+        close_choice=1
+    )
```
BigRoy, line 800: Having both Ok and Cancel button seems a bit off - because when 'cancelled' the publish still starts. I would actually move this `readInput` earlier in the code to BEFORE changing the active instances. Like, first ask the user directly for a comment and whether they want to initiate the publish - and if they cancel, just stop completely. If they want no comment they just leave it empty and press enter.

Also, you labeled it "Publish note" but the Publisher itself calls it a _comment_ and e.g. has placeholder text: _Add a comment to your publish_. For consistency I'd refactor this to also read "Add publish comment" and title "Publish comment".

moonyuet (approved):
Looks good in my side. 
image

I agree with @BigRoy that add comment section shouldn't have two buttons included. One button named like "Add Publish Comment" should be okay. 

I actually have another suggestion but it could be somehow not quite useful or good idea for this. I actually wondered it is possible to have new clean null node and includes a button inside to do multiple publishes by depending how many inputs of this node. I am not sure if it would mess up the things.

BigRoy:
> I actually have another suggestion but it could be somehow not quite useful or good idea for this. I actually wondered it is possible to have new clean null node and includes a button inside to do multiple publishes by depending how many inputs of this node. I am not sure if it would mess up the things.

I feel like then we're a bit getting into the real of why not use the publisher UI - haha. But there's definitely merit in reconceptualizing the OpenPype x Houdini workflows to better support ROP node inputs/stacking and through that maybe get what you're asking.

openpype/hosts/houdini/api/plugin.py
MustafaJafar: I added that `_` prefix because I found out that I picked the the same name for the attribute 
and I wasn't comfortable with something like this 

```
if self.add_self_publish_button:
    add_self_publish_button(instance_node)
```

I feel a little dumb, why didn't I just pick another name for the attribute :) ?!

MustafaJafar:
@iLLiCiTiT  
could you review my edits ?, I made few tweaks in `window.py` and `host_tools.py`

moonyuet:
> > I actually have another suggestion but it could be somehow not quite useful or good idea for this. I actually wondered it is possible to have new clean null node and includes a button inside to do multiple publishes by depending how many inputs of this node. I am not sure if it would mess up the things.
> 
> I feel like then we're a bit getting into the real of why not use the publisher UI - haha. But there's definitely merit in reconceptualizing the OpenPype x Houdini workflows to better support ROP node inputs/stacking and through that maybe get what you're asking. I believe @fabiaserra brought it up elsewhere as well to get that logic into the OP system.

I find this openpype_publish is way more convenient than clicking create, and validate than publish, and the custom node for mutliple publish would be only benefitical if there are too much instances created(users would prefer to publish some instances). 
They can always choose what to publish from the output network and they can make the well use of both merge and swift nodes to choose what to render first and last.
But It is not quite helpful if users want to have some specific settings by instances.

MustafaJafar:
Speaking of mulitpublishing, what about this idea. 
A multi input node (e.g. merge node) 
its publish button should publish input instances on the same order as they are connected ? 
which do nothing but set active and trigger publish action (I don't know how to force a certain order)
image

BigRoy:
> its publish button should publish input instances on the same order as they are connected ?
> which do nothing but set active and trigger publish action (I don't know how to force a certain order)

As far as I know the publish system doesn't allow to set custom order for the instances but this shows that it definitely should allow it?

I think supporting these ROP node hierarchies is exactly what @fabiaserra described as well. However, I'd even expect you don't need to use `Merge` nodes but can also just stack ROP nodes after each other which if we were to JUST support the "Merge" node case would still fail.

I feel like we should just do:
1. Find a way to define the "Order of instances" for the publisher
2. Then using the ROP node ordering sort the instances in Houdini by that order.
3. Then on each ROP node the multi publish node would just have a "Self publish"  and "Self publish + upstream" entry I suppose to allow to be specific what you'd want?

I suspect 2) would be trivial by going over `inputDependencies` of the ROP node and similarly 3) would be easy as well. As soon as we can define the processing order of instances (in particular the Extractors I suppose?)

It's good to note however that due to how publishing works even if we were to order instances in a single publish that the instances that finished "before" the current one would still not be integrated plus it'd only really work if the Extractors or at the same `InstancePlugin.order` - otherwise the instance that is earlier would still process it's plugin later than the other extractor.

moonyuet:
> I think supporting these ROP node hierarchies is exactly what @fabiaserra described as well. However, I'd even expect you don't need to use `Merge` nodes but can also just stack ROP nodes after each other which if we were to JUST support the "Merge" node case would still fail.

I am thinking if the screenshot below possible in the multiple publish button.
image

MustafaJafar:
> I believe in native pyblish you could just sort the context[:] 

if this is going to work, then it should be handled in the publisher itself, because I think `context[:]` is generated when the publisher triggered. the only thing I could find is `self._controller.instances.values()` which is very similar to `context.instances` which is just `context._instances_by_id.values()` 

> 3. Then on each ROP node the multi publish node would just have a "Self publish" and "Self publish + upstream" entry I suppose to allow to be specific what you'd want?

> I am thinking if the screenshot below possible in the multiple publish button.
> I admit that it can be complicated

I like these ideas!

MustafaJafar:
> I am thinking if the screenshot below possible in the multiple publish button.
> I admit that it can be complicated

I think this should work, it can be complicated but this `inputDependencies` does the hard work as  BigRoy pointed, 
and why using `merge` as long as I can connect ROPs like this! 
I suggested using `merge` because I forgot that we can really connect these as such! 

image


```python
import hou 
from openpype.hosts.houdini.api.lib import publisher_show_and_publish

def find_rops(input_tuple):
    out_list = []
    if isinstance(input_tuple[0], hou.RopNode):
        return input_tuple[0].path()
    
    if isinstance(input_tuple[0], tuple):
        for item in input_tuple:
            out_list.append(find_rops(item))
    
    return out_list

def multi_publish():
    """Self publish from ROP nodes."""
    
    result, comment = hou.ui.readInput(
        "Add Publish Comment",
        buttons=("Publish", "Cancel"),
        title="Publish comment",
        close_choice=1
    )
    
    if result:
        return
        
    node = hou.node('/out/pointcacheBgeoSelfPublish') # hou.node(".")
    inputs_paths = find_rops(node.inputDependencies())
    inputs_paths.append(node.path())
    
    host = registered_host()
    context = CreateContext(host, reset=True)
    
    for instance in context.instances:
        node_path = instance.data.get("instance_node")
        if not node_path:
            continue
    
        active = node_path in inputs_paths
        instance["active"] = active
        hou.node(node_path).parm("active").set(active)
    
    context.save_changes()
    publisher_show_and_publish(comment)
    
multi_publish()
```

fabiaserra:
First of all, thank you very much for doing this work, however I'm not sure about the multi-input dependency publish triggers and whether that's the right approach. First let me forward some messages I sent on Discord this last week for extra context:

> check me on this but one limitation with that approach from a Houdini POV is that the OP publish triggers of Houdini are currently not that smart to create complex job dependency networks in the farm (i.e., multiple cache ROP nodes connected to a render ROP and you want all of the cache nodes to cache independently as Deadline jobs and be dependencies of the render ROP). To do those kind of dependency networks, we would need to use the Deadline ROP submitter which parse the ROP node network and maps it to the Deadline job graph
maybe we should provide an OP publish ROP/TOP  instead so we can leverage the existing Deadline submitter (or PDG) to create complex networks that automate most of those processes, so we aren't limited by a single node creating the Deadline graph
instead of duplicating efforts that Deadline already does out of the box, I think we should redesign the approach of the Houdini OP integration and leverage that instead by providing some basic building blocks that play together with the vanilla Houdini ROP/TOP submissions as that opens up a lot of much more powerful workflows
if the OP publish was just another Houdini node that when it evaluates it runs the publish, then that would play well with the existing submitter logic and you could create very complex Deadline graphs that sets up all the caching, rendering and publishing as independent tasks

I have to say that I still haven't used the Houdini OP implementation that much to have a good grasp of all the things its capable to do so let me know if I'm understanding some things wrong.

@MustafaJafar what you wrote will simply collect all the existing OpenPype instances (and ONLY the OpenPype ones, not other vanilla nodes like a vanilla ROP geo cache) in the ROP graph and publish them one by one without any particular order, is that correct? Could you submit that into the farm?

While I could see some uses of that, most Houdini TDs will set up quite complex ROP dependency graphs with multiple vanilla nodes or custom HDAs that when they submit them to the farm, they expect those to be submitted in a smart way so all the nodes that can run independently of each other (i.e., different wedges) or frame by frame they will get different chunk sizes, farm priorities and farm groups (some machines might be reserved for rendering, others for expensive simulations while others for simple processes like I/O or publishing). 

For example, you could have a graph like the following:
image

The different Deadline ROP nodes allow you to tweak all the upstream nodes so you can control whether those nodes evaluate on a single machine or you split in multiple machines if they aren't time dependant... and see that in my example I also made one of the nodes bypass (does your `find_rops` take into account bypassed nodes?)

And you could also make use other dependency ROP nodes for having further control of the dependency evaluation:
image

This becomes even more complex and powerful with the introduction of PDG in Houdini (https://www.sidefx.com/docs/houdini/tops/intro.html) and the already existent Deadline Scheduler for it which should handle most of that logic already (https://www.sidefx.com/docs/houdini/nodes/top/deadlinescheduler.html), where you can define very granular control on which thinks depend on what and have multiple different types of tasks.

Now, having that in mind... try to think how does that play together with the OpenPype integration in Houdini and whether we are duplicating work in the plugins evaluation. Would users combine the use of Deadline submissions so they get all the power of submitting their dependency graphs to the farm while also submitting some other things to publish through OpenPype? That becomes a nightmare to manage and we are duplicating logic that Deadline and other schedulers already have implemented in a better way. 

Instead, if OpenPype had its own ROP node (or TOP node for PDG uses) that provided the simple building block of running a publish of its upstream node, then we could start wrapping it on other util HDAs (i.e., a camera publish node that runs the extract and the publish task and also has a Deadline node inside so it creates a separate job in the farm for each) I think it would fit much better with the existing vanilla way of working in Houdini and open up a lot more procedural workflows of using OpenPype and encourage Houdini TDs to build their own toolsets.

Does it make sense that we are writing an extract plugin node for each ROP node in Houdini and taking over the control of the evaluation graph?

MustafaJafar:
## How this PR works
> what you wrote will simply collect all the existing OpenPype instances (and ONLY the OpenPype ones, not other vanilla nodes like a vanilla ROP geo cache) in the ROP graph and publish them one by one without any particular order, is that correct?

This PR is doing nothing other than Activating/Deactivating publish instances then triggering publish action, which quite similar to your Nuke snippet. 
So, Both screenshots shows how to publish particular two nodes. 
<table>
<tr>
<th>Activate nodes from publisher UI, then trigger publish action</th>
<th>Connect desired nodes as such, then click the `OpenPype Publish` to activate the whole tree and deactivate all other publish instances then trigger publish action</th>
</tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

vanilla ROPs will have no effect because publisher just ignores them. 

> Could you submit that into the farm? 

As long as the publish instances is configured to publish to farm, then the publisher will publish it to farm. 
I don't interfere in how how publisher publish these instances.

---
## Publish Order
I think in the current state of the Houdini plugin, we just depend on the publisher which sorts them in ascii order. 
I don't know the publisher's inner component responsible for publishing instances which could be hacked to publish single instances without using the publisher. 
So, is publishing order important in other DCCs like Maya ?

image

---
## Houdini Future
> if OpenPype had its own ROP node (or TOP node for PDG uses) that provided the simple building block of running a publish of its upstream node

Our discussion makes me feel that Houdini can change how the system is built. or have an alternate publisher system. 
Some other DCCs don't allow having task processors (like TOP nodes in houdini) and there are only these top level API calls to (export data - import data - save working file) 

<table>
<tr>
<th>In Maya publishing is straight forward, you only need to call a dedicated export command</th>
<th>The Houdini equivalent is to trigger the render action</th>
</tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

If  publisher is broken into smaller components so we can replace some components with Houdini's components 
I really find it hard to imagine how it would look like, maybe something like this where we have some Houdini nodes that imitates the publisher functionality ?? 

it should work in a similar way to the deadline submissions (render task + publish task) ? 
and then we can set publishing order explicitly ? 

image

fabiaserra:
> As long as the publish instances is configured to publish to farm, then the publisher will publish it to farm.
I don't interfere in how how publisher publish these instances.

What's the current support for caching and publishing geometry in the farm for simulations?

> it should work in a similar way to the deadline submissions (render task + publish task) ?
and then we can set publishing order explicitly ?

Yeah that's how I imagine it, the publish task would be just a simple task that given some expected files it just runs a publish of them.

RIght now I feel like the publish integration is very limited because the publish plugins are trying to also run the export/render and that's way behind the capabilities that you can do with the existing vanilla Deadline integrations where you can set much more complex exporters so I'm still struggling a bit to understand how they would be mixing both workflows in a friendly way. 

I think that as a first small improvement we could make the publish plugins of Houdini support the same workflow Nuke supports where you can run an OP publish with "Use existing frames" (in the farm as well) so that a Houdini TD would for the most part just submit their ROP/TOP graph in Deadline as they would normally and when they are happy with the results they would just run a publish of those nodes validating that the output files exist. This however would require them to use the OP ROP nodes always instead of vanilla ones and from the feedback I got from my artists... they will probably forget often to use those. Has there been any work on providing a method to convert vanilla ROP nodes into OP wrapped ones? We could probably hide the vanilla ones or default the creation to always be OP ROP nodes but that's another friction of workflow in the current OP Houdini workflow.

Because of these limitations, we are actually considering providing DCC agnostic publishers (like the TrayPublisher but easier to use) so they can publish renders directly from RV or Deadline by just running it off the written files on disk and the tool automatically inferring the context where it needs to be published.

MustafaJafar:
> What's the current support for caching and publishing geometry in the farm for simulations?

There's this WIP implementation https://github.com/ynput/OpenPype/pull/4903 which works similarly to the implemented farm render https://github.com/ynput/OpenPype/pull/4825 

> Right now I feel like the publish integration is very limited

Recalling how Maya integration works, I think the limitation came from there because adopting Maya solutions in Houdini can limit Houdini quite a bit. 
maybe @BigRoy can share us more thoughts 

> that's another friction of workflow in the current OP Houdini workflow.

Thanks for helping me thinking in a different way, I think we can say that this PR is not a cure it's just a pain relief

fabiaserra:
> Thanks for helping me thinking in a different way, I think we can say that this PR is not a cure it's just a pain relief

Correct yeah, I think it's still useful to add the functionality this PR provides but we need to rethink the publish workflow so it fits better with the existing Houdini Deadline integration without duplicating all that logic in OP

BigRoy:
> so it fits better with the existing Houdini Deadline integration without duplicating all that logic in OP

I just want to note that Deadline isn't the only renderfarm to be considered so I just wanted to remark that we should design it for Deadline, but anything that can keep our supported system more native to how Houdini works will likely have the system deal better with other renderfarms as well - e.g. openCue, royal render or someone's custom system.

fabiaserra:
> > so it fits better with the existing Houdini Deadline integration without duplicating all that logic in OP
> 
> I just want to note that Deadline isn't the only renderfarm to be considered so I just wanted to remark that we should design it for Deadline, but anything that can keep our supported system more native to how Houdini works will likely have the system deal better with other renderfarms as well - e.g. openCue, royal render or someone's custom system.

Yeah I agree and that adds even more arguments in favor of moving away from recreating Deadline specific logic in OpenPype. Those other farm schedulers would provide native ways of submitting the Houdini scenes without OpenPype so the integration of OP should just be a thin layer on top that allows you to integrate/publish the results of it

MustafaJafar:
> I think it's still useful to add the functionality this PR provides but we need to rethink the publish workflow

So, Let's finalize this PR and we can open the discussion about rethinking the publish workflow so it fits better with the existing Houdini features 

---

> ## How this PR works
> > what you wrote will simply collect all the existing OpenPype instances (and ONLY the OpenPype ones, not other vanilla nodes like a vanilla ROP geo cache) in the ROP graph and publish them one by one without any particular order, is that correct?
> 
> This PR is doing nothing other than Activating/Deactivating publish instances then triggering publish action, which quite similar to your Nuke snippet. So, Both screenshots shows how to publish particular two nodes.
> 
> Activate nodes from publisher UI, then trigger publish action	Connect desired nodes as such, then click the `OpenPype Publish` to activate the whole tree and deactivate all other publish instances then trigger publish action
> 	
> vanilla ROPs will have no effect because publisher just ignores them.
> 
> > Could you submit that into the farm?
> 
> As long as the publish instances is configured to publish to farm, then the publisher will publish it to farm. I don't interfere in how how publisher publish these instances.

Do you guys have any further suggestions other than this implementation ?

krishna8008:
An idea here - Can we create a otl that publishes the geometry in sop network itself ?

fabiaserra:
> An idea here - Can we create a otl that publishes the geometry in sop network itself ?

Yeah I agree and this also falls on that republish refactor discussion. If we added the ROP node that simply handles the publish, creating and wrapping new util OTLs would be very easy to do. We would just create a SOP otl that wraps a ropnet with the geo rop export linked to the publish node as a util node so they can just call that from the SOP network

fabiaserra:
> > An idea here - Can we create a otl that publishes the geometry in sop network itself ?
> 
> Yeah I agree and this also falls on that republish refactor discussion. If we added the ROP node that simply handles the publish, creating and wrapping new util OTLs would be very easy to do. We would just create a SOP otl that wraps a ropnet with the geo rop export linked to the publish node as a util node so they can just call that from the SOP network

one of the most used nodes in my prior studio was the "<studio code> cache" SOP OTL that handled all the export, publish and read back of SOP geo. If I had to write that again I would probably just wrap the existing Labs File Cache node (or the non labs one but the labs one adds some cool new utilities) and add the publish functionality on top of it so they can publish the output directly from a button in the node

fabiaserra:
> > I think it's still useful to add the functionality this PR provides but we need to rethink the publish workflow
> 
> So, Let's finalize this PR and we can open the discussion about rethinking the publish workflow so it fits better with the existing Houdini features
> 
> > ## How this PR works
> > > what you wrote will simply collect all the existing OpenPype instances (and ONLY the OpenPype ones, not other vanilla nodes like a vanilla ROP geo cache) in the ROP graph and publish them one by one without any particular order, is that correct?
> > 
> > 
> > This PR is doing nothing other than Activating/Deactivating publish instances then triggering publish action, which quite similar to your Nuke snippet. So, Both screenshots shows how to publish particular two nodes.
> > Activate nodes from publisher UI, then trigger publish action	Connect desired nodes as such, then click the `OpenPype Publish` to activate the whole tree and deactivate all other publish instances then trigger publish action
> > 	
> > vanilla ROPs will have no effect because publisher just ignores them.
> > > Could you submit that into the farm?
> > 
> > 
> > As long as the publish instances is configured to publish to farm, then the publisher will publish it to farm. I don't interfere in how how publisher publish these instances.
> 
> Do you guys have any further suggestions other than this implementation ?

I'm happy merging this! I have it on my TODO list to test it ASAP

krishna8008:
> > > An idea here - Can we create a otl that publishes the geometry in sop network itself ?
> 
> > 
> 
> > Yeah I agree and this also falls on that republish refactor discussion. If we added the ROP node that simply handles the publish, creating and wrapping new util OTLs would be very easy to do. We would just create a SOP otl that wraps a ropnet with the geo rop export linked to the publish node as a util node so they can just call that from the SOP network
> 
> 
> 
> one of the most used nodes in my prior studio was the "\<studio code\> cache" SOP OTL that handled all the export, publish and read back of SOP geo. If I had to write that again I would probably just wrap the existing Labs File Cache node (or the non labs one but the labs one adds some cool new utilities) and add the publish functionality on top of it so they can publish the output directly from a button in the node


Exactly! We can recreate to our in-house file cache node by copying parameters and code... I would request this🙏🏻

MustafaJafar:
> An idea here - Can we create a otl that publishes the geometry in sop network itself ?

I'm not I got that right, 
so, 
are you referring to an `ayon cache node` ? 
if yes let's discuss it on forums, [Link here] 
or 
are you referring to support hda from non object level ? 
if yes, we can discuss it in the relevant GH issue

MustafaJafar:
After the recent discussions about Houdini, I think this PR may be misleading. 
because, we don't make use of the dependency graph other than to know which nodes to activate/deactivate!
I believe the feature introduced by this PR is better converted to a shelf button that publish selection (like the snippet provided by fabia)

So, the shelf button should be exactly like this: 
```python
from openpype.hosts.houdini.api import lib
lib.publish_nodes() 
```
also, you'll be able to pass some nodes to it (it should be a list)
```python
from openpype.hosts.houdini.api import lib
nodes = []
lib.publish_nodes(nodes) 
```

Could we vote ? 
@fabiaserra @BigRoy @moonyuet 

I'm suggesting : 
1. removing `publish` button from ROPs and its settings
2. change `self_publish` to `publish_nodes`

BigRoy:
I think the PR is a better _intermediate_ workflow than a shelf button for 'selected nodes'. If you want to publish multiple, just run the publisher UI.

Having said that, after this temporary solution to improve the workflow a tiny bit we should look into the other solutions. Especially with making sure the publishing system itself (how it runs Houdini ROPs etc.) mimics closer how Houdini itself works.

MustafaJafar:
> Having said that, after this temporary solution to improve the workflow a tiny bit we should look into the other solutions. Especially with making sure the publishing system itself (how it runs Houdini ROPs etc.) mimics closer how Houdini itself works.

Then, we will just wait for this PR #5651 to be merged, then I'll deal with few conflicts and bump the addon version then it will be good to go for a final review before merging.

fabiaserra:
Yeah I agree with @BigRoy, the button is much better than a shelve tool. I actually already cherry picked the changes of this PR on my fork of OP and I have been using it for quick submission of render -> publish for simple scenes with OP render ROP nodes with good success. However, as I mentioned on Discord and the forums, we are likely going to stop using this whole workflow thanks to the OP Publisher node that I wrote. We will be running a much more vanilla Houdini/Deadline workflow and then just using the OP publisher node to push the generated outputs to the OP database. Unfortunately there's still too many things we would need to do in the OP publish workflow to be comparable with the things we can do in vanilla.

pcoip_client_vbHHUxQCFV

BigRoy:
> However, as I mentioned on Discord and the forums, we are likely going to stop using this whole workflow thanks to the OP Publisher node that I wrote. We will be running a much more vanilla Houdini/Deadline workflow and then just using the OP publisher node to push the generated outputs to the OP database. Unfortunately there's still too many things we would need to do in the OP publish workflow to be comparable with the things we can do in vanilla.

Just wanted to say - digging the mini-demo. Looking forward to a prototype / draft PR just showing off the potential so it can be iterated upon for a potential future merge of this into OpenPype/AYON itself. So if you are able to, or if you have had one or two iterations on your end already it'd be great to see this shared.

_Sidenote:_ It looks like you're literally naming the subset `cool_sphere` which would actually be considered incorrect since it doesn't adhere to the project setting's publish template system. At best you might want to use it to name `variant` but that _could_ get complicated due to nodes requiring unique names inside Houdini but you might want to publish the same variant into a different product type / family. I wrote some more about that 'complication' [here](https://github.com/ynput/OpenPype/pull/5666#issuecomment-1741710742) and [here](https://github.com/ynput/OpenPype/pull/5666#issuecomment-1750671747)

fabiaserra:
> > However, as I mentioned on Discord and the forums, we are likely going to stop using this whole workflow thanks to the OP Publisher node that I wrote. We will be running a much more vanilla Houdini/Deadline workflow and then just using the OP publisher node to push the generated outputs to the OP database. Unfortunately there's still too many things we would need to do in the OP publish workflow to be comparable with the things we can do in vanilla.
> 
> Just wanted to say - digging the mini-demo. Looking forward to a prototype / draft PR just showing off the potential so it can be iterated upon for a potential future merge of this into OpenPype/AYON itself. So if you are able to, or if you have had one or two iterations on your end already it'd be great to see this shared.
> 
> _Sidenote:_ It looks like you're literally naming the subset `cool_sphere` which would actually be considered incorrect since it doesn't adhere to the project setting's publish template system - unless your studio templates are not setup like `{family}{Variant}` and `{family}{Task}{Variant}`. At best you might want to use it to name `variant` but that _could_ get complicated due to nodes requiring unique names inside Houdini but you might want to publish the same variant into a different product type / family. I wrote some more about that 'complication' [here](https://github.com/ynput/OpenPype/pull/5666#issuecomment-1741710742) and [here](https://github.com/ynput/OpenPype/pull/5666#issuecomment-1750671747)
> 
> Technically it is valid for subset name to be whatever (and not even include the variant I suppose?) it's just that it gets way more likely that e.g. a render called `Main`, a pointcache called `Main` and a vdb cache called `Main` start publishing into the same thing because artists will then at some point forget to prefix with e.g. `vdb` or `fx` or whatever to ensure uniqueness. So prefixes default to at least include the family - even though it's not an **absolute** requirement by the backend.
> 
> To adhere to the subset naming templates you'll need to use [`get_subset_name`](https://github.com/ynput/OpenPype/blob/d41ac2a3fd5bac2492c6d97f9e1003564489f7fa/openpype/pipeline/create/subset_name.py#L64-L157)

Yeah our subset name template is basically just `{variant}` so it's directly the name that you set (my artists didn't like all the prefixes we were getting on all publishes) and I made that default to `opinput(".")` for now just as a first default of the subset name but they can break that channel reference and change it to whatever name. You are right on the name uniqueness and I might change it to something else if this becomes a problem but from my experience I have found that most artists prefer naming their publishes with a unique name on their scene to avoid confusion. For multi-shot workflows this could become a problem but then again, the node isn't really enforcing it to be `opinput`, I could change the logic and even remove integer suffixes easily

Happy to demo once we get more traction on this, I literally just deployed it for my artists to test yesterday

MustafaJafar:
>  we are likely going to stop using this whole workflow thanks to the OP Publisher node that I wrote.

Good to hear man, Nice Demo! 

Anyways, I updated this PR and resolved conflicts. 
Now what to do ? 
mrbean

moonyuet:
@antirotor I think this PR is ready for review and merge

MustafaJafar:
> @antirotor I think this PR is ready for review and merge
I think we would need @iLLiCiTiT review too because I modified the publisher a little bit. 

Quick recap
This PR provides a quick way to specify what nodes to publish in a "Houdini way" where we connect ROPs to each other and click `publish` button. 

It automates the current workflow and it doesn't enable procedural workflows. 
Houdini does provide me with the execution order of nodes but I cannot use this information in any way.

antirotor (approved):
I've tested it and the code looks good. This is one step on a path to make OP/AYON workflow in Houdini more native a procedural but even in this form, it is useful. Thanks @MustafaJafar

openpype/hosts/houdini/api/lib.py
MustafaJafar: I figured it out. 
So, `imprint` creates/updates `ParmTemplates` as well as default values but never updates the values. 

In my test, It didn't work until I added `instance_node.setParms(new_values)`  at the end of `update_instances` in `HoudiniCreator`

fabiaserra (approved):

openpype/hosts/houdini/api/lib.py
MustafaJafar: it should work now.

openpype/hosts/houdini/api/lib.py
```diff
@@ -847,3 +854,101 @@ def update_houdini_vars_context_dialog():
...
+    inputs_paths.append(current_node.path())
+
+    host = registered_host()
+    context = CreateContext(host, reset=True)
+
+    for instance in context.instances:
+        node_path = instance.data.get("instance_node")
+        if not node_path:
+            continue
```
iLLiCiTiT, line 928: We should disable all other instances, or?
```suggestion
        if not node_path:
            instance["active"] = False
            continue
```

openpype/hosts/houdini/api/plugin.py
```diff
@@ -316,6 +324,11 @@ def get_network_categories(self):
     def apply_settings(self, project_settings):
         """Method called on initialization of plugin to apply settings."""
 
+        # Apply General Settings
+        self.add_publish_button = \
+            project_settings["houdini"]["general"]["add_self_publish_button"]
```
iLLiCiTiT, line 329: Be carefull about new keys in settings...
```suggestion
        houdini_general_settings = project_settings["houdini"]["general"]
        self.add_publish_button = houdini_general_settings.get(
            "add_self_publish_button", False)
```

iLLiCiTiT:
Moved the show and publish logic to Publisher window, which reduced public functions of the window. Also removed `reset_on_show` argument from host tools. The proper implementation of the functionality would be more complicated. Please re-test if still works as before.

openpype/hosts/houdini/api/lib.py
iLLiCiTiT: > Or this is maybe more readable?

Much better.

antirotor:
First time publish will (probably) publish stuff ok, but then it will reset the UI and the publishing report/state will be lost. Subsequent publishes are ok.

antirotor (approved):
With the reset fix, it works!

PR #5623, opened 2023-09-15
Nuke: Multiple format supports for ExtractReviewDataMov

moonyuet:
## Changelog Description
This PR would fix the bug of  the plugin `ExtractReviewDataMov` not being able to support extensions other than `mov`. The plugin is also renamed to `ExtractReviewDataBakingStreams` as i provides multiple format supoort.
## Additional info
n/a

## Testing notes:
1. add `png` to the settings `project_settings/nuke/publish/ExtractReviewDataBakingStreams/outputs/baking/extension`
2. publish review from nuke


CHANGELOG.md
jakubjezek001, line 1: Please remove changes from changelog.md. Those changes are not related to the enhancement.

openpype/hosts/nuke/api/plugin.py
```diff
@@ -799,8 +800,11 @@ def __init__(self,
         self.get_file_info()
 
         self.log.info("File info was set...")
-
         self.file = self.fhead + self.name + ".{}".format(self.ext)
+        if self.ext !="mov":
```
jakubjezek001, line 804: I would rather unlock this from only `mov` to all video extensions. Here is usable variable in transcoding.py

```Python
from openpype.lib.transcoding import (
    VIDEO_EXTENSIONS
)
```

openpype/hosts/nuke/api/lib.py
```diff
@@ -1167,6 +1167,24 @@ def format_anatomy(data):
     return anatomy.format(data)
 
 
+def get_frame():
```
jakubjezek001, line 1170: The name could be improved to `get_hashed_frame_padding_from_anatomy()`

openpype/hosts/nuke/api/plugin.py
```diff
@@ -799,8 +800,11 @@ def __init__(self,
...
+            frame = get_frame()
```
jakubjezek001, line 805: It would be good to use this only if a padding is not already used in `self.in_path` since the variable is inheriting value from publishing context of `self.instance.data.get("path", None)` and the hash can already be part for the path, so we can use it. Read the above comment with `get_hashed_frame_padding_from_anatomy`.

In case of video file is used for rendering path then we can get the value from Anatomy. And the anatomy would be better to pass from publishing context too. It means we could have optional input arguments for `get_hashed_frame_padding_from_anatomy` to accept Anatomy.

openpype/settings/entities/schemas/projects_schema/schemas/schema_nuke_publish.json
jakubjezek001, line 1: We should be maintaining backward compatibility for projects which are already having configured multiple presets which are not delivered by code defaults. So we have to acknowledge users that the settings will be deprecated and they should move settings to the new category.  For inspiration look here: https://github.com/ynput/OpenPype/blob/36a8976c580d822c816992c752181b48a884202d/openpype/settings/entities/schemas/projects_schema/schema_project_nuke.json#L287-L290

openpype/settings/defaults/project_settings/nuke.json
```diff
@@ -447,7 +447,7 @@
         "ExtractReviewDataLut": {
             "enabled": false
         },
-        "ExtractReviewDataMov": {
```
jakubjezek001, line 450: Need to maintain backward compatibility. Look below for more information.

openpype/settings/ayon_settings.py
jakubjezek001, line 1: Need to maintain backward compatibility. Look below for more information.

jakubjezek001 (changes requested):

server_addon/nuke/server/settings/publish_plugins.py
```diff
@@ -165,7 +165,7 @@ class BakingStreamModel(BaseSettingsModel):
         title="Custom tags", default_factory=list)
 
 
-class ExtractReviewDataMovModel(BaseSettingsModel):
```
jakubjezek001, line 168: This change will also require an increase of patch version in `server_addon\nuke\server\version.py`

jakubjezek001: We will also need to preserve the original plugin settings in place, but will need to add some annotation for users to understand the future depreciation. For more inspiration about this look to `server_addon\nuke\server\settings\workfile_builder.py:61` Here it is added as docstrings and since those are shown in the settings it could be good way to inform user about the changes.

openpype/hosts/nuke/plugins/publish/extract_review_data_mov.py
```diff
@@ -8,7 +8,7 @@
 from openpype.hosts.nuke.api.lib import maintained_selection
 
 
-class ExtractReviewDataMov(publish.Extractor):
```
jakubjezek001, line 11: Label and file name should be also changed to `Extract Review Baking Streams` and `extract_review_baking_streams.py`

jakubjezek001: Also we will need to figure out the way to do the backward compatibility for the old plugin settings inside of the plugin. You could use classmethod `self.apply_settings(project_settings)` to use similar approach as it is used here https://github.com/ynput/OpenPype/pull/5623/files#diff-79438371778038156c3abd11150aff9b31a510034f4891c40e16e6f55482885bR752-R762

jakubjezek001: here is also some more information about apply_settings during plugin discovery in publishing `openpype\pipeline\publish\lib.py:500`

openpype/hosts/nuke/api/plugin.py
```diff
@@ -801,6 +804,8 @@ def __init__(self,
         self.log.info("File info was set...")
 
         self.file = self.fhead + self.name + ".{}".format(self.ext)
+        if self.ext != VIDEO_EXTENSIONS:
```
jakubjezek001, line 807:
```suggestion
        if ".{}".format(self.ext) != VIDEO_EXTENSIONS:
```

jakubjezek001: extensions are with dot and `self.ext` is without.

jakubjezek001 (changes requested):
```
  File "C:\CODE\__PYPE\OpenPype_3\.venv\lib\site-packages\pyblish\plugin.py", line 527, in __explicit_process
    runner(*args)
  File "C:\CODE\__PYPE\OpenPype_3\openpype\hosts\nuke\plugins\publish\extract_review_data_mov.py", line 126, in process
    data = exporter.generate_mov(**o_data)
  File "C:\CODE\__PYPE\OpenPype_3\openpype\hosts\nuke\api\plugin.py", line 968, in generate_mov
    self.render(write_node.name())
  File "C:\CODE\__PYPE\OpenPype_3\openpype\hosts\nuke\api\plugin.py", line 824, in render
    nuke.execute(
RuntimeError: Write1: could not deduce output format from file extension```

File: openpype/hosts/nuke/api/plugin.py [hunk 8]
```diff
@@ -804,8 +804,13 @@ def __init__(self,
         self.log.info("File info was set...")
 
         self.file = self.fhead + self.name + ".{}".format(self.ext)
-        if self.ext != VIDEO_EXTENSIONS:
-            self.file = os.path.basename(self.path_in)
+        if ".{}".format(self.ext) not in VIDEO_EXTENSIONS:
+            filename = os.path.basename(self.path_in)
+            self.file = filename
+            if ".{}".format(self.ext) not in self.file:
+                wrg_ext = filename.split(".")[-1]
```
jakubjezek001, line 811: `wrg_ext` is a wrong extension or what exactly? If it is `wrong` than it is suggesting there is something wrong with it and It doesn't seem to be correct in this case. I would suggest to use `original_ext` also it is easier to read if it is full name rather than the abbreviation.

openpype/hosts/nuke/plugins/publish/extract_review_baking_streams.py
```diff
@@ -25,6 +25,34 @@ class ExtractReviewDataMov(publish.Extractor):
     viewer_lut_raw = None
     outputs = {}
 
+    @classmethod
+    def apply_settings(cls, project_settings):
+        """just in case there are some old presets
+        in deprecrated ExtractReviewDataMov Plugins
```
jakubjezek001, line 31: there is a typo in the plugin `deprecrated` should be `deprecated`

openpype/hosts/nuke/plugins/publish/extract_review_baking_streams.py
```diff
@@ -25,6 +25,34 @@ class ExtractReviewDataMov(publish.Extractor):
...
+            cls.viewer_lut_raw = (
+                deprecrated_review_settings["viewer_lut_raw"]
+            )
+
+        if deprecrated_review_settings["outputs"] == (
+            current_review_settings["outputs"]
+        ):
+            cls.outputs = current_review_settings["outputs"]
+        else:
+            cls.outputs = deprecrated_review_settings["outputs"]
```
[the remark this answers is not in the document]
jakubjezek001: there is a attribute `enabled` we could use `project_settings/nuke/publish/ExtractReviewDataMov/enabled` So if it is enabled then it is having priority on `ExtractReviewDataMov` and once you disable it it would prioritize `ExtractReviewDataBakingStreams`

jakubjezek001 (changes requested):
At the moment there is no way to use new plugins settings since it is always preferred old deprecated plugin settings.

jakubjezek001 (changes requested):
We also need to add baking stream profile name as suffix to write filepath. Since the name will remain the same it will be colliding with the original sequence.. For example, I had set baking stream to be EXR and original was EXR.

```
{'name': 'bakingNew', 'ext': 'exr', 'files': 'renderCompositingMain.####.exr', 'stagingDir': 'C:\\projects\\jeza_test_project_01\\shots\\shot00010\\work\\compositing\\renders\\nuke\\renderCompositingMain', 'tags': ['bakingNew', 'review', 'delete'], 'frameStart': 996, 'frameEnd': 1013}]
```

It the example above it would be desirable to have files `renderCompositingMain_bakingNew.####.exr`

openpype/hosts/nuke/api/lib.py
```diff
@@ -3397,3 +3397,24 @@ def create_viewer_profile_string(viewer, display=None, path_like=False):
     if path_like:
         return "{}/{}".format(display, viewer)
     return "{} ({})".format(viewer, display)
+
+
+def get_head_filename_without_hashes(original_path, name):
+    """Function to get the ranmed head filename without frame hashes
```
Minkiu, line 3403: `ranmed` -> `renamed`

openpype/settings/ayon_settings.py
```diff
@@ -748,7 +748,20 @@ def _convert_nuke_project_settings(ayon_settings, output):
...
 
     new_review_data_outputs = {}
-    for item in ayon_publish["ExtractReviewDataMov"]["outputs"]:
+    outputs_settings = None
+    # just in case that the users having old presets in outputs setting
+    deprecrated_review_settings = ayon_publish["ExtractReviewDataMov"]
+    current_review_settings = (
+        ayon_publish["ExtractReviewDataBakingStreams"]
+    )
+    if deprecrated_review_settings["outputs"] == (
```
Minkiu, line 757: We only care about `current_review_settings["outputs"]` if they are the same as `deprecrated_review_settings["outputs"]`, so why not just do ` outputs_settings = deprecrated_review_settings["outputs"]` directly without the `if...else` ?

Minkiu (approved):
Haven't tested but code looks good, aside from the comment I made above.

openpype/settings/ayon_settings.py
moonyuet: i would do the same with `apply_setting` function in the extractor.

openpype/hosts/nuke/api/lib.py
```diff
@@ -3397,3 +3397,24 @@ def create_viewer_profile_string(viewer, display=None, path_like=False):
...
+    To avoid the system being confused on finding the filename with
+    frame hashes if the head of the filename has the hashed symbol
+
+    Args:
+        original_path (str): the filename with frame hashes
+        name (str): the name of the tags
+
+    Returns:
+        filename: the renamed filename with the tag
+    """
```
jakubjezek001, line 3413: Can we add an example of incoming values and expected output value?

openpype/hosts/nuke/api/lib.py
```diff
@@ -3397,3 +3397,24 @@ def create_viewer_profile_string(viewer, display=None, path_like=False):
...
+    filename = os.path.basename(original_path)
+    fhead = filename.split(".")[0]
+    if "#" in fhead:
+        fhead = fhead.replace("#", "")[:-1]
```
jakubjezek001, line 3417: There might be also case where instead of `###` in name can be `%03d`

jakubjezek001, line 3417:
```suggestion
    if "#" in fhead:
        fhead = fhead.replace("#", "").rstrinp(".")
```
this might be more safe. Since the dot might or might not be there.

openpype/hosts/nuke/api/plugin.py
```diff
@@ -808,6 +812,14 @@ def __init__(self,
         self.log.info("File info was set...")
 
         self.file = self.fhead + self.name + ".{}".format(self.ext)
+        if ".{}".format(self.ext) not in VIDEO_EXTENSIONS:
+            filename = get_head_filename_without_hashes(
+                self.path_in, self.name)
+            self.file = filename
+            if ".{}".format(self.ext) not in self.file:
+                original_ext = filename.split(".")[-1]
```
jakubjezek001, line 820:
```suggestion
                original_ext = os.path.splitext(filename)[-1].strip(".")
```
why not to use os path function dedicated extension split?

BigRoy:
The error reported [here](https://github.com/ynput/OpenPype/pull/5623#issuecomment-1733568153) I suspect is due to this PR https://github.com/ynput/OpenPype/pull/5273 which just got merged since that removed the attribute [here](https://github.com/ynput/OpenPype/pull/5273/files#diff-c4b595d393443040b1d7c9551351bfffe69f9c00ad87bd9b6243da47a643f337L24)

jakubjezek001:
Issue is fixed here https://github.com/ynput/OpenPype/pull/5634. Thanks @BigRoy for pointing at it!

openpype/hosts/nuke/api/lib.py
moonyuet: I have updated the commit to check on both "#" and "%03d", which is inspired by your "abstraction of file collection in Houdini" ticket

moonyuet:
@jakubjezek001 I get some inspirations from your PR #5496 and make some updates on the files from representation for png publishing besides searching the frame hashes ("#" and "%03d")

jakubjezek001 (approved):
LGTM!

openpype/hosts/nuke/api/lib.py
```diff
@@ -3423,3 +3423,60 @@ def create_viewer_profile_string(viewer, display=None, path_like=False):
...
+        fhead = re.sub("#+", "", fhead).rstrip(".")
+    elif "%" in fhead:
+        # use regex to convert %04d to {:0>4}
+        padding = re.search("%(\\d)+d", fhead)
+        padding = padding.group(1) if padding else 1
+        fhead = re.sub(
+            "%.*d",
+            "{{:0>{}}}".format(padding),
+            fhead
+        ).rstip(".")
```
BigRoy, line 3455:
```suggestion
        ).rstrip(".")
```
Typo?

moonyuet: nice catch

BigRoy: Even with this fixed it returns:
```python
print(get_head_filename_without_hashes("renderCompositingMain.%04d.exr", "baking"))
# renderCompositingMain.%04d.exr
```
Shouldn't it include `.baking`?

BigRoy: It kind of feels like what you're trying to do is this:
```python
import os
import re


def get_head_filename_without_hashes(original_path, name):
    """Function to get the renamed head filename without frame hashes
    To avoid the system being confused on finding the filename with
    frame hashes if the head of the filename has the hashed symbol

    Args:
        original_path (str): the filename with frame hashes
        e.g. "renderCompositingMain.####.exr"
        name (str): the name of the tags
        e.g. "baking"

    Returns:
        filename: the renamed filename with the tag
        e.g. "renderCompositingMain.baking.####.exr"
    """
    filename = os.path.basename(original_path)
    fhead = os.path.splitext(filename)[0].strip(".")
    
    
    def insert_name(matchobj):
        return "{}.{}".format(name, matchobj.group(0))
    
    return re.sub(r"(%\d*d)|#+", insert_name, filename)
    
    
print(get_head_filename_without_hashes("renderCompositingMain.%4d.exr", "baking"))
print(get_head_filename_without_hashes("renderCompositingMain.####.exr", "baking"))
print(get_head_filename_without_hashes("renderCompositingMain.%d.exr", "baking"))
```

Note that this would put the baking naming in front of EVERY occurence of `%4d` or `####`
```python
print(get_head_filename_without_hashes("renderCompositingMain.###.%d.exr", "baking"))
renderCompositingMain.baking.###.baking.%d.exr
```

moonyuet: > Even with this fixed it returns:
> 
> ```python
> print(get_head_filename_without_hashes("renderCompositingMain.%04d.exr", "baking"))
> # renderCompositingMain.%04d.exr
> ```
> 
> Shouldn't it include `.baking`?

have you tried in Nuke?

moonyuet: > Even with this fixed it returns:
> 
> ```python
> print(get_head_filename_without_hashes("renderCompositingMain.%04d.exr", "baking"))
> # renderCompositingMain.%04d.exr
> ```
> 
> Shouldn't it include `.baking`?

It should include the baking as I checked.
image

BigRoy: It doesn't work for the case with `%d` or `%04d` as per my example code.

moonyuet: For` print(get_head_filename_without_hashes("renderCompositingMain.%4d.exr", "baking"))`, it is supposed it would be `renderCompositingMain.0004.baking.exr`, if I dont predict wrong.

BigRoy: I also wonder whether it's worth to add pytest-able docstrings here. They are quite easy to read, make sense to a developer and can actually be run as part of your test suite.

It would look e.g. like this:

```
import os
import re


def get_head_filename_without_hashes(original_path, name):
    """Function to get the renamed head filename without frame hashes
    To avoid the system being confused on finding the filename with
    frame hashes if the head of the filename has the hashed symbol

    Examples:
        >>> get_head_filename_without_hashes("render.####.exr", "baking")
        render.baking.####.exr
        >>> get_head_filename_without_hashes("render.%d.exr", "tag")
        render.tag.%d.exr
        >>> get_head_filename_without_hashes("exr.####.exr", "foo")
        exr.foo.%04d.exr

    Args:
        original_path (str): the filename with frame hashes
        name (str): the name of the tags

    Returns:
        str: the renamed filename with the tag

    """
    filename = os.path.basename(original_path)
        
    def insert_name(matchobj):
        return "{}.{}".format(name, matchobj.group(0))
    
    return re.sub(r"(%\d*d)|#+", insert_name, filename)
```

openpype/hosts/nuke/api/lib.py
```diff
@@ -3423,3 +3423,55 @@ def create_viewer_profile_string(viewer, display=None, path_like=False):
...
+    """Get filenames without frame hash
+        i.e. "renderCompositingMain.baking.0001.exr"
+
+    Args:
+        filename (str): filename with frame hash
+        frame_start (str): start of the frame
+        frame_end (str): end of the frame
+
+    Returns:
+        filenames(list): list of filename
```
BigRoy, line 3466:
```suggestion
        list: filename per frame of the sequence
```

openpype/hosts/nuke/api/lib.py
moonyuet: > > For` print(get_head_filename_without_hashes("renderCompositingMain.%4d.exr", "baking"))`, it is supposed it would be `renderCompositingMain.0004.baking.exr`, if I dont predict wrong.
> 
> Why would it be `.0004`?
> 
> Anyway - the refactored code still has the issue that the function name says it returns the "head" of the filename even though it returns the whole filename. Also, it still returns it with hashes.
> 
> All it really does is insert a particular string in front of the frame number. Still not sure why and what this is for.

it is just for example if you have frame 4 as that case....

openpype/hosts/nuke/plugins/publish/extract_review_baking_streams.py
```diff
@@ -25,6 +25,22 @@ class ExtractReviewDataMov(publish.Extractor):
...
+        nuke_publish = project_settings["nuke"]["publish"]
+        deprecated_setting = nuke_publish["ExtractReviewDataMov"]
+        current_setting = nuke_publish["ExtractReviewDataBakingStreams"]
+        if not deprecated_setting["enabled"]:
+            if current_setting["enabled"]:
+                cls.viewer_lut_raw = current_setting["viewer_lut_raw"]
+                cls.outputs = current_setting["outputs"]
+        else:
+            cls.viewer_lut_raw = deprecated_setting["viewer_lut_raw"]
+            cls.outputs = deprecated_setting["outputs"]
```
BigRoy, line 42: Would this be more readable?
```suggestion
        if deprecated_setting["enabled"]:
            # Use deprecated settings if they are still enabled
            cls.viewer_lut_raw = deprecated_setting["viewer_lut_raw"]
            cls.outputs = deprecated_setting["outputs"]
        elif current_setting["enabled"]:
            cls.viewer_lut_raw = current_setting["viewer_lut_raw"]
            cls.outputs = current_setting["outputs"]
```

openpype/settings/ayon_settings.py
```diff
@@ -748,7 +748,19 @@ def _convert_nuke_project_settings(ayon_settings, output):
...
+    # just in case that the users having old presets in outputs setting
+    deprecrated_review_settings = ayon_publish["ExtractReviewDataMov"]
+    current_review_settings = (
+        ayon_publish["ExtractReviewDataBakingStreams"]
+    )
+    if not deprecrated_review_settings["enabled"]:
+        if current_review_settings["enabled"]:
+            outputs_settings = current_review_settings["outputs"]
+    else:
+        outputs_settings = deprecrated_review_settings["outputs"]
```
BigRoy, line 761: I suspect this might also be more readable with the logic changed like [here](https://github.com/ynput/OpenPype/pull/5623/files#r1338759890).

openpype/hosts/nuke/api/plugin.py
```diff
@@ -808,6 +817,18 @@ def __init__(self,
...
+            # filename would be with frame hashes if
+            # the file extension is not in video format
+            filename = get_head_filename_without_hashes(
+                self.path_in, self.name)
+            self.file = filename
+            # make sure the filename are in
+            # correct image output format
+            if ".{}".format(self.ext) not in self.file:
+                original_ext = os.path.splitext(filename)[-1].strip(".")    # noqa
+                self.file = filename.replace(original_ext, ext)
```
BigRoy, line 830: Looking at this logic we're much better off just always setting the output extension we want instead of replacing `original_ext` to `ext` in the filename, etc. even though what we really want is to ever just swap around the extension - so let's just do that.
```suggestion
            # ensure output has expected extension
            filename_no_ext, _ = os.path.splitext(filename)
            self.file = "{}.{}".format(filename_no_ext, self.ext)
```
No need for the if statement and much clearer that we're always enforcing the extension `self.ext`

openpype/settings/entities/schemas/projects_schema/schemas/schema_nuke_publish.json
```diff
@@ -371,6 +371,151 @@
 
             ]
         },
+        {
+            "type": "label",
+            "label": "^ Settings and for <span style=\"color:#FF0000\";><b>ExtractReviewDataMov</b></span> is deprecated and will be soon removed.  <br> Please use <b>ExtractReviewDataBakingStreams</b> instead."
```
BigRoy, line 376: Should we mention when **soon** is? 

Can we e.g. mention a version number when it should be removed? And I've seen "deprecated by version x.y.z" messages here and there - do we have a good "todo" tracking list for those removals so we actually remove those when intended? @iLLiCiTiT do you know? If not, any ideas on how we could best set that up?

openpype/hosts/nuke/api/lib.py
```diff
@@ -3423,3 +3423,55 @@ def create_viewer_profile_string(viewer, display=None, path_like=False):
...
+    To avoid the system being confused on finding the filename with
+    frame hashes if the head of the filename has the hashed symbol
+
+    Examples:
+        >>> get_head_filename_without_hashes("render.####.exr", "baking")
+        render.baking.####.exr
+        >>> get_head_filename_without_hashes("render.%d.exr", "tag")
+        render.tag.%d.exr
+        >>> get_head_filename_without_hashes("exr.####.exr", "foo")
+        exr.foo.%04d.exr
```
BigRoy, line 3439: This example is wrong.
```suggestion
        >>> get_head_filename_without_hashes("exr.%04d.exr", "foo")
        exr.foo.%04d.exr
```

openpype/hosts/nuke/plugins/publish/extract_review_baking_streams.py
```diff
@@ -8,15 +8,15 @@
...
+class ExtractReviewDataBakingStreams(publish.Extractor):
     """Extracts movie and thumbnail with baked in luts
 
     must be run after extract_render_local.py
 
     """
 
     order = pyblish.api.ExtractorOrder + 0.01
-    label = "Extract Review Data Mov"
+    label = "Extract Review Data Baking Streams"
```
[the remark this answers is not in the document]
BigRoy: Just wanted to say I got confused by 'streams' as well.

openpype/hosts/nuke/plugins/publish/extract_review_baking_streams.py
```diff
@@ -8,15 +8,15 @@
 from openpype.hosts.nuke.api.lib import maintained_selection
 
 
-class ExtractReviewDataMov(publish.Extractor):
+class ExtractReviewDataBakingStreams(publish.Extractor):
     """Extracts movie and thumbnail with baked in luts
```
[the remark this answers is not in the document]
moonyuet: I agree that the docstring needs to be changed.

openpype/hosts/nuke/plugins/publish/extract_review_baking_streams.py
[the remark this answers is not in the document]
moonyuet: Maybe we can have a name like `Extract Review Data Sequences`?

openpype/hosts/nuke/api/lib.py
```diff
@@ -3423,3 +3423,55 @@ def create_viewer_profile_string(viewer, display=None, path_like=False):
     if path_like:
         return "{}/{}".format(display, viewer)
     return "{} ({})".format(viewer, display)
+
```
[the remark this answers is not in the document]
moonyuet: This is not quite related to this PR but @jakubjezek001 probably can answer you this

openpype/hosts/nuke/plugins/publish/extract_review_baking_streams.py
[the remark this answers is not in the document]
BigRoy: Not sure what the 'data' refers to in this case. Also, it works for both mov and sequences, right? Or does it only generates moves? (The `plugin.ExporterReviewMov` plugin it runs in the code does seem to hint it generates movs?)


- `ExtractReviewConversions`? 
- `ExtractConvertReviews`
- `ExtractReviewIntermediates` - if they are never intended to be finals but just preprocessed sequences for further processing?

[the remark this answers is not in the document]
jakubjezek001: I agree that the current name is not ideal. Initially, it was called "Bake Streams" because it referred to data created in Upstream and related to the Baking of colorspace, which can involve multiple streams.

However, I really like the name "ExtractReviewIntermediates". The only problem I have with it is that the word "Intermediates" can be complex, especially considering that we will be mentioning it frequently every week since we are supporting this feature. This could potentially cause difficulties for non-native speakers. 😄

openpype/hosts/nuke/api/lib.py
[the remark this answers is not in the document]
jakubjezek001: Not yet, but yes, this can be a global abstraction for future - If you want the Issue is yours @fabiaserra   ;)

openpype/hosts/nuke/api/plugin.py
```diff
@@ -21,6 +21,9 @@
     CreatedInstance,
     get_current_task_name
 )
+from openpype.lib.transcoding import (
+    VIDEO_EXTENSIONS
```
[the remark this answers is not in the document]
jakubjezek001: The VIDEO_EXTENSIONS constant was created to store all possible video extensions in the pipeline, eliminating the need to do it repeatedly in each plugin or host. Reverting this change would remove this advantage. Can you explain the potential benefits of your suggestion?

openpype/settings/entities/schemas/projects_schema/schemas/schema_nuke_publish.json
jakubjezek001: I would suggest avoiding the restriction for now since we don't currently have a proper workflow for tracking TODOs. This theme is quite complex. I recommend opening a discussion topic in the community as it has the potential to be very lively.

openpype/hosts/nuke/api/lib.py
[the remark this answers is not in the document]
BigRoy: Admittedly the abstraction is more of an issue with the host DCC (especially with nuke apparently throwing this around) and Maya also doing it like that in the past. With OCIO v2 becoming more mainstream I feel like almost all APIs are updating with it and start supporting actual separate `display`  and `view` arguments/settings for things instead of requiring to combine them into a string. Hopefully this 'workaround' is just temporary and future versions of apps actually just supporting display separate from view.

Anyway, looking at this now. Are we sure this is correct?
```python
        return "{}/{}".format(display, viewer)
    return "{} ({})".format(viewer, display)
```
Did they reverse / swap around the arguments? 😢 That must make for quite some confusion.

openpype/settings/entities/schemas/projects_schema/schemas/schema_nuke_publish.json
BigRoy: Should we still name the version we're expecting to deprecate it like we've done for other deprecations recently?

openpype/hosts/nuke/plugins/publish/extract_review_baking_streams.py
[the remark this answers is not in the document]
BigRoy: ChatGPT gives me:

> Certainly, here are some simpler words you can use to describe an intermediate file that is meant for processing and not a final output:
> 
> 1. **Working File**: This term suggests that the file is actively being used in the process.
> 
> 2. **Temporary File**: This implies that the file is only temporary and will be discarded after processing.
> 
> 3. **Process File**: This term emphasizes that the file is part of the processing steps.
> 
> 4. **In-Progress File**: This indicates that the file is still being worked on and isn't a final result.
> 
> 5. **Transitory File**: This term implies that the file is transitional and not permanent.
> 
> 6. **Draft File**: This suggests that the file is a preliminary version.
> 
> 7. **Processing File**: This term makes it clear that the file is involved in processing tasks.
> 
> 8. **Intermediate Data**: While it includes "data," it's a clear term for files used during processing.
> 
> 9. **Placeholder File**: This indicates that the file holds a place in the process but isn't final.
> 
> 10. **Provisional File**: This term implies that the file is provisional or temporary in nature.
> 
> Choose the term that best fits your context and your audience's understanding of English.

Still feel like intermediate is the best wording if the files coming out of that extractor aren't intended to be the final output files for exports.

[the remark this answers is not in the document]
BigRoy: @moonyuet I noticed with https://github.com/ynput/OpenPype/pull/5623/commits/6c1e066b3b7d58eba28d38900a886a0c89c556b5 you just changed the name to `ExtractReviewIntermediate`. If the extractor is able to output multiple intermediate files I'd still think the plural for Intermediates would make that clearer? That'd be `ExtractReviewIntermediate` -> `ExtractReviewIntermediates`

server_addon/nuke/server/settings/publish_plugins.py
```diff
@@ -166,6 +166,18 @@ class BakingStreamModel(BaseSettingsModel):
...
+    outputs: list[BakingStreamModel] = Field(
+        default_factory=list,
+        title="Baking streams"
+    )
+
+
+class ExtractReviewIntermediatesModel(BaseSettingsModel):
     enabled: bool = Field(title="Enabled")
     viewer_lut_raw: bool = Field(title="Viewer lut raw")
     outputs: list[BakingStreamModel] = Field(
```
jakubjezek001, line 183: I believe since we had renamed it to **Intermediates** any trace of **BakingStream** can be renamed too, right? In this case `IntermediateOutputModel`.

openpype/hosts/nuke/plugins/publish/extract_review_intermediates.py
```diff
@@ -8,15 +8,15 @@
 from openpype.hosts.nuke.api.lib import maintained_selection
 
 
-class ExtractReviewDataMov(publish.Extractor):
-    """Extracts movie and thumbnail with baked in luts
+class ExtractReviewIntermediates(publish.Extractor):
+    """Extracts Sequences and thumbnail with baked in luts
```
jakubjezek001, line 12: This plugin is not only for generating Intermediate Sequences. This is misleading and I would rather recommend 

```suggestion
    """Extracting intermediate files for transcoding. 
```

BigRoy: If it only does videos or sequences I think mentioning that is clearer than "files" by the way - I'd not expect it to process word documents or audio files? :) Does it do single images?

moonyuet: I edited the docsting to """Extracting intermediate videos or sequences with thumbnail for transcoding"""

openpype/hosts/nuke/api/lib.py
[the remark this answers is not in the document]
fabiaserra: Oh sorry I might have confused you by the placing where I put the comment, I meant the functions actually added on this PR that simply remove frame hashes on filename `get_head_filename_without_hashes` and `get_filenames_without_hash`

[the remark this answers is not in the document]
fabiaserra: there's a lot of opportunities of creating better generic python util modules that we could reuse throughout hosts but the argument is generally "that needs a bigger refactor and it's outside of this PR"... so we just keep growing the codebase and duplicating the same utility functions with slight modifications over and over. I understand it's not always easy to do that but just wanted to raise it

openpype/hosts/nuke/api/plugin.py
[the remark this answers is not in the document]
fabiaserra: yep, exactly what @BigRoy said

openpype/hosts/nuke/api/lib.py
```diff
@@ -3423,3 +3423,55 @@ def create_viewer_profile_string(viewer, display=None, path_like=False):
...
+    Returns:
+        str: the renamed filename with the tag
+    """
+    filename = os.path.basename(original_path)
+
+    def insert_name(matchobj):
+        return "{}.{}".format(name, matchobj.group(0))
+
+    return re.sub(r"(%\d*d)|#+", insert_name, filename)
```
BigRoy, line 3453: I saw this get merged just now but some of the discussions didn't seem entirely resolved. For example, the function name here `get_head_filename_without_hashes` AND the top description of the docstring are entirely incorrect.

It does **not** return the head of a filename, it returns a file filename.
It does **not** remove the hashes from the filename.
It does **not** avoid issues with finding files on the file system - right? How does it do that and why?

The only thing this function does is prepend **{name}.** before any frame hash (where frame hash is %d, %04d `####` or alike.

Can we get a new PR to still clean this up? :)

openpype/hosts/nuke/api/plugin.py
```diff
@@ -808,6 +817,18 @@ def __init__(self,
...
         self.file = self.fhead + self.name + ".{}".format(self.ext)
+        if ".{}".format(self.ext) not in VIDEO_EXTENSIONS:
+            # filename would be with frame hashes if
+            # the file extension is not in video format
+            filename = get_head_filename_without_hashes(
+                self.path_in, self.name)
+            self.file = filename
+            # make sure the filename are in
+            # correct image output format
+            if ".{}".format(self.ext) not in self.file:
```
BigRoy, line 828: I suspect that we're always expecting the file should END with the extension - right? So we should still change the extension if the extension `.exr` for the filename was originally `hello.exr.0001.png`.

This should've been:
```python
if not self.file.endswith(".{}".format(ext)):
    # Swap the extension
    ...
```

openpype/hosts/nuke/api/lib.py
jakubjezek001: @moonyuet shell we do the additional changes in extra (linked to this one please). Also include the other comment https://github.com/ynput/OpenPype/pull/5623#discussion_r1341521543

PR #5624, opened 2023-09-15
Max: tycache family support

moonyuet:
## Changelog Description
Tycache family supports for Tyflow Plugin in Max

## Additional info
Users should have tyFlow Pro Plugins to use this instance
As the tyCache format requires strict naming rules for storing the particle data, the published tyCache would adopt the originalBasename instead of using default publish naming convention.
Both tycache and tyspline export are working in the same export mode.
## Testing notes:
OpenPype
1. Launch Max via launcher
2. Create -> Create tyCache
3. Toggle the settings you preferred in Publish tab
4. Publish.

Ayon
1. Go to Studio Setting -> Anatomy Presets -> add tyCache name according to template.json located in Openpype setting.
image
2. Launch Max via launcher
3. Create -> Create tyCache
4. Toggle the settings you preferred in Publish tab
5. Publish.


sebasti1a:
I just notice that opening a project when the Ayon plugin is active takes much longer to open. I had already noticed this on projects with Unreal 5.2.
I think it retries a cache of assets when opening, is there perhaps a way to remove this procedure?
Capture_ayon_01
image
image

openpype/hosts/max/plugins/publish/extract_tycache.py
```diff
@@ -0,0 +1,169 @@
...
+            'name': 'tyc',
+            'ext': 'tyc',
+            'files': filenames if len(filenames) > 1 else filenames[0],
+            "stagingDir": stagingdir
+        }
+        instance.data["representations"].append(representation)
+        self.log.info(f"Extracted instance '{instance.name}' to: {filenames}")
+
+    def get_file(self, filepath, start_frame, end_frame):
+        filenames = []
```
Minkiu, line 70: Missing docstrings.

openpype/hosts/max/plugins/publish/extract_tycache.py
```diff
@@ -0,0 +1,169 @@
...
+            added by user
+
+        Returns:
+            additional_args(list): a list of additional args for MAX script
+        """
+        additional_args = []
+        for key, value in attrs.items():
+            tyc_attribute = None
+            if isinstance(value, bool):
+                tyc_attribute = f"{operator}.{key}=True"
```
Minkiu, line 165: Regardless of the `bool` value, it will always be `True` ?

moonyuet: it's only happened to be True when the user add the tycache parameter for publish in the publish tab.

openpype/hosts/max/plugins/publish/validate_tyflow_data.py
```diff
@@ -0,0 +1,74 @@
...
+            if rt.ClassOf(sel) in [rt.tyFlow,
+                                   rt.Editable_Mesh]:
+                if "tyFlow" not in sel_tmp:
+                    invalid.append(sel)
+            else:
+                invalid.append(sel)
+
+        return invalid
+
+    def get_tyflow_operator(self, instance):
```
Minkiu, line 53: missing docstrings

Minkiu (approved):

LiborBatek (changes requested):
So I have finally manage to produce some data via tyFlow Pro version...

Been tested in ```3dsmax 2022``` and having issues when trying to ```Publish``` tyCache product.

getting this error:
image

also note the ```outliner``` there happens to appear new tyCache node each time trying to ```Publish```...not desirable

Checking that failure log I have also checked that Global/3dsmax Color Management System being off (maybe unrelated tho) and still not working

BigRoy:
@LiborBatek could you try again? :)

LiborBatek (changes requested):
Ok with latest develop merged in it doesnt fail on color space anymore but during
```Integrate Asset``` publishing...

here is the log:
[tyFlow_Log.txt](https://github.com/ynput/OpenPype/files/12750215/tyFlow_Log.txt)

**Comment by BigRoy** [comment]
> Ok with latest develop merged in it doesnt fail on color space anymore but during `Integrate Asset` publishing...
> 
> here is the log: [tyFlow_Log.txt](https://github.com/ynput/OpenPype/files/12750215/tyFlow_Log.txt)

Just wanted to remark:

What stands out to me here is that you're actually sharing logs from "details" instead of the full publishing log which is intended to be used for 'sharing your reports' by design so they can be loaded into the report viewer by devs. I have a feeling you're also 'misguided' by the visual cues of thew new publisher's design and hitting the [Publisher design not actionable enough thread brought up here](https://community.ynput.io/t/publishing-better-error-reporting/55/30?u=bigroy).

**Comment by LiborBatek** [comment]
@BigRoy you probably mean sharing full JSON file log. I am aware of it but I have just simply copy/paste the failing one to make it more compact (less lines of text).

If preferable, I can always attach the full JSON file. Is that it? Thanks for your thoughts!

**Review by LiborBatek — CHANGES_REQUESTED** [review]
Did another test go but again failing on me... I will provide full JSON too this time!

image

[publish-report-230929-17-16.json.txt](https://github.com/ynput/OpenPype/files/12768219/publish-report-230929-17-16.json.txt)

Just edit the ext back to json (because of GIThub limit for attachements) thx.

File: openpype/hosts/max/plugins/publish/extract_tycache.py [hunk 4]
```diff
@@ -0,0 +1,197 @@
...
+
+        stagingdir = self.staging_dir(instance)
+        filename = "{name}.tyc".format(**instance.data)
+        path = os.path.join(stagingdir, filename)
+        filenames = self.get_file(instance, start, end)
+        additional_attributes = instance.data.get("tyc_attrs", {})
+
+        with maintained_selection():
+            job_args = None
+            if instance.data["tycache_type"] == "tycache":
```
Minkiu, line 45: Wouldn't this be a check we do earlier and just return `if instance.data["tycache_type"] not in ["tycache", "tycachespline"]` ?

Then here we could do:
```py
job_args = self.export_particle(
                    instance.data["members"],
                    start, end, path,
                    additional_attributes,
                    tycache_spline_enabled=True if instance.data["tycache_type"] == "tycachespline" else False
)
```

openpype/hosts/max/plugins/publish/validate_tyflow_data.py
Minkiu: Is `_summary_` supposed to be there?

Minkiu (approved):

openpype/hosts/max/plugins/publish/validate_tyflow_data.py
moonyuet: ah yes. I will add the docstring to this one.

openpype/hosts/max/plugins/publish/validate_pointcloud.py
```diff
@@ -52,46 +39,6 @@ def process(self, instance):
         if report:
             raise PublishValidationError(f"{report}")
```
BigRoy, line 40:
```suggestion
            raise PublishValidationError("\n".join(report))
```

openpype/hosts/max/plugins/publish/validate_tyflow_data.py
```diff
@@ -0,0 +1,94 @@
...
+        if invalid_object:
+            report.append(f"Non tyFlow object found: {invalid_object}")
+
+        invalid_operator = self.get_tyflow_operator(instance)
+        if invalid_operator:
+            report.append((
+                "tyFlow ExportParticle operator not "
+                f"found: {invalid_operator}"))
+        if report:
+            raise PublishValidationError(f"{report}")
```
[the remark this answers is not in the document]
moonyuet: I would use raise PublishValidationError for invalid object and invalid operator respectively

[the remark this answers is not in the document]
BigRoy: You are best of doing:

psuedocode:

```python
log.error("issue-1")
log.error("issue-2")
log.error("issue-3")
log.error("issue-4")
raise PublishValidationError("issues occurred", description="describe the validator and how the artist can solve any issues it might find")
```

By **logging directly** instead of building the report into a list first it also means that **if any unexpected error occurs** that the first few logs had been generated at least and it's much easier to pinpoint how far the code actually got. Plus the code gets more readable as well. :)

moonyuet:
There would be some default channel data for tycache export preseted similar to PRT export in the updated commiot.
image

LiborBatek (approved):
All works now!

I was able to produce the ```tyCache``` asset and succesfully load it into scene as asset.

image

moonyuet:
@Minkiu and @antirotor would be great if you can take a look at the code before the merge:).

openpype/hosts/max/plugins/publish/extract_tycache.py
```diff
@@ -0,0 +1,197 @@
...
+        filenames = self.get_files(instance, start, end)
+        additional_attributes = instance.data.get("tyc_attrs", {})
+
+        with maintained_selection():
+            job_args = None
+            has_tyc_spline = (
+                True
+                if instance.data["tycache_type"] == "tycachespline"
+                else False
+            )
```
antirotor, line 49: This is a little bit simpler:

```suggestion
            has_tyc_spline = instance.data["tycache_type"] == "tycachespline"
```

openpype/hosts/max/plugins/publish/extract_tycache.py
```diff
@@ -0,0 +1,197 @@
...
+            'name': 'tyc',
+            'ext': 'tyc',
+            'files': filenames if len(filenames) > 1 else filenames[0],
+            "stagingDir": stagingdir,
+        }
+        representations.append(representation)
+        self.log.info(f"Extracted instance '{instance.name}' to: {filenames}")
+
+        # Get the tyMesh filename for extraction
+        mesh_filename = "{}__tyMesh.tyc".format(instance.name)
```
antirotor, line 69: we can use f-string here too:

```suggestion
        mesh_filename = f"{instance.name}__tyMesh.tyc"
```

openpype/hosts/max/plugins/publish/extract_tycache.py
```diff
@@ -0,0 +1,197 @@
...
+        Set the filenames accordingly to the tyCache file
+        naming extension(.tyc) for the publishing purpose
+
+        Actual File Output from tyFlow in tyCache format:
+        <InstanceName>__tyPart_<frame>.tyc
+
+        e.g. tycacheMain__tyPart_00000.tyc
+
+        Args:
+            instance (str): instance.
```
antirotor, line 93: Instance isn't `str` - my linter complains a lot :)

```suggestion
            instance (pyblish.api.Instance): instance.
```

openpype/hosts/max/plugins/publish/extract_tycache.py
```diff
@@ -0,0 +1,197 @@
...
+            end_frame (int): End frame.
+
+        Returns:
+            filenames(list): list of filenames
+
+        """
+        filenames = []
+        # should we include frame 0 ?
+        for frame in range(int(start_frame), int(end_frame) + 1):
+            filename = "{}__tyPart_{:05}.tyc".format(instance.name, frame)
```
antirotor, line 104:
```suggestion
            filename = "{instance.name}__tyPart_{frame:05}.tyc"
```

openpype/hosts/max/plugins/publish/extract_tycache.py
```diff
@@ -0,0 +1,197 @@
...
+                        tycache_spline_enabled=False):
+        """Sets up all job arguments for attributes.
+
+        Those attributes are to be exported in MAX Script.
+
+        Args:
+            members (list): Member nodes of the instance.
+            start (int): Start frame.
+            end (int): End frame.
+            filepath (str): Output path of the TyCache file.
```
antirotor, line 119: this is missing documentation for the rest of the arguments -- `additional_attributes` and `tycache_spline_enabled`

openpype/hosts/max/plugins/publish/extract_tycache.py
```diff
@@ -0,0 +1,197 @@
...
+        """
+        job_args = []
+        opt_list = self.get_operators(members)
+        for operator in opt_list:
+            if tycache_spline_enabled:
+                export_mode = f'{operator}.exportMode=3'
+                job_args.append(export_mode)
+            else:
+                export_mode = f'{operator}.exportMode=2'
+                job_args.append(export_mode)
```
antirotor, line 133:
```suggestion
            if tycache_spline_enabled:
                export_mode = f'{operator}.exportMode=3'
            else:
                export_mode = f'{operator}.exportMode=2'
            job_args.append(export_mode)
```

openpype/hosts/max/plugins/publish/validate_tyflow_data.py
```diff
@@ -0,0 +1,96 @@
...
+                description="Container should only include tyFlow object\n "
+                "and tyflow operator 'Export Particle' should be in \n"
+                "the tyFlow editor")
+
+    def get_tyflow_object(self, instance):
+        """Get the nodes which are not tyFlow object(s)
+        and editable mesh(es)
+
+        Args:
+            instance (str): instance node
```
antirotor, line 43:
```suggestion
            instance (pyblish.api.Instance): instance node
```

openpype/hosts/max/plugins/publish/validate_tyflow_data.py
```diff
@@ -0,0 +1,96 @@
...
+                invalid.append(sel)
+
+        return invalid
+
+    def get_tyflow_operator(self, instance):
+        """Check if the Export Particle Operators in the node
+        connections.
+
+        Args:
+            instance (str): instance node
```
antirotor, line 70:
```suggestion
            instance (pyblish.api.Instance): instance
```

antirotor (changes requested):
just a few code-wise nodes - I haven't been able to test it in Max

openpype/hosts/max/plugins/publish/extract_tycache.py
BigRoy: Or even:
```python
export_mode = 3 if tycache_spline_enabled else 2
job_args.append(f"{operator}.exportMode={export_mode}")
```

openpype/hosts/max/plugins/publish/collect_tycache_attributes.py
```diff
@@ -0,0 +1,79 @@
...
+
+    order = pyblish.api.CollectorOrder + 0.02
+    label = "Collect tyCache attribute Data"
+    hosts = ['max']
+    families = ["tycache"]
+
+    def process(self, instance):
+        all_tyc_attributes_dict = {}
+        attr_values = self.get_attr_values_from_data(instance.data)
+        tycache_boolean_attributes = attr_values.get("all_tyc_attrs")
```
Minkiu, line 19: Would it work if we:
```py
tycache_boolean_attributes = attr_values.get("all_tyc_attrs", [])
for attrs in tycache_boolean_attributes:
    all_tyc_attributes_dict[attrs] = True
```

So we skip checking `if tycache_boolean_attributes:` ?

openpype/hosts/max/plugins/publish/extract_tycache.py
```diff
@@ -0,0 +1,197 @@
...
+from openpype.pipeline import publish
+
+
+class ExtractTyCache(publish.Extractor):
+    """
+    Extract tycache format with tyFlow operators.
+    Notes:
+        - TyCache only works for TyFlow Pro Plugin.
+
+    Args:
```
Minkiu, line 16: These are not args, but methods, right?

openpype/hosts/max/plugins/publish/extract_tycache.py
```diff
@@ -0,0 +1,197 @@
...
+    """
+
+    order = pyblish.api.ExtractorOrder - 0.2
+    label = "Extract TyCache"
+    hosts = ["max"]
+    families = ["tycache"]
+
+    def process(self, instance):
+        # TODO: let user decide the param
+        start = int(instance.context.data.get("frameStart"))
```
Minkiu, line 33: Unsure if "frameStart" would be ever `None` in which instance it would fail to `int()`

BigRoy: With what @Minkiu says here the case of `frameStart` not being defined would error - so we're just as well of assuming it exists:
```suggestion
        start = instance.context.data["frameStart"])
```

Aside of that - why are we using the **context** start frame and not a start frame of the instance? 😮

Minkiu (approved):

openpype/hosts/max/plugins/publish/collect_tycache_attributes.py
```diff
@@ -0,0 +1,79 @@
...
+        tyc_layer_attr = attr_values.get("tycache_layer")
+        if tyc_layer_attr:
+            all_tyc_attributes_dict["tycacheLayer"] = (
+                tyc_layer_attr)
+        tyc_objname_attr = attr_values.get("tycache_objname")
+        if tyc_objname_attr:
+            all_tyc_attributes_dict["tycache_objname"] = (
+                tyc_objname_attr)
+        self.log.debug(
+            f"Found tycache attributes: {all_tyc_attributes_dict}")
```
BigRoy, line 32: Other than reporting the found attributes does this do ANYTHING with that data? It doesn't store it on the instance or context?

Also, `all_tyc_attributes_dict` is way verbose making this code hard to read:

```suggestion
    def process(self, instance):
        attr_values = self.get_attr_values_from_data(instance.data)

        attributes = {}
        for attr_key in attr_values.get("all_tyc_attrs", []):
            attributes[attr_key] = True
            
        tyc_layer_attr = attr_values.get("tycache_layer")
        if tyc_layer_attr:
            attributes["tycacheLayer"] = tyc_layer_attr
            
        tyc_objname_attr = attr_values.get("tycache_objname")
        if tyc_objname_attr:
            attributes["tycache_objname"] = tyc_objname_attr
        
        # TODO: Do we need this data somewhere?
        instance.data["storeWhereverYouWant"] = attributes
        self.log.debug(
            f"Found tycache attributes: {attributes}"
        )
```

also, what's the reason we're converting attribute `tycache_layer` to `tycacheLayer` from attribute definition to the attributes dict but `tycache_objname` remains `tycache_objname`? I feel this is a bit of a code style issue. If the attribute def keys would match the names you want to store them in the dict with you could just do:

```suggestion
    def process(self, instance):
        attr_values = self.get_attr_values_from_data(instance.data)

        attributes = {}
        for attr_key in attr_values.get("all_tyc_attrs", []):
            attributes[attr_key] = True
        
        for key in ["tycacheLayer", "tycacheObjName"]:
            if key in attr_values:
                attributes[key] = attr_values[key]
        
        # TODO: Do we need this data somewhere?
        instance.data["storeWhereverYouWant"] = attributes
        self.log.debug(
            f"Found tycache attributes: {attributes}"
        )
```

However, you might be better off always SETTING the 'default' value for an option even if `""` for consistency of the `instance.data`
```suggestion
    def process(self, instance):
        attr_values = self.get_attr_values_from_data(instance.data)

        attributes = {}
        for attr_key in attr_values.get("all_tyc_attrs", []):
            attributes[attr_key] = True
        
        for key in ["tycacheLayer", "tycacheObjName"]:
            attributes[key] = attr_values.get(key, "")
        
        # TODO: Do we need this data somewhere?
        instance.data["storeWhereverYouWant"] = attributes
        self.log.debug(
            f"Found tycache attributes: {attributes}"
        )
```

moonyuet: I would try to see if it pass the validation tmr

moonyuet:
I have updated the code, and I find out that there is only one export mode for tycache when I tested with the code. When I tested with the export mode =3, it turns out to be alembic mesh export. So both tycachespline and tycache works in the same export mode.

openpype/hosts/max/plugins/publish/extract_tycache.py
```diff
@@ -0,0 +1,186 @@
...
+            # TODO: add the additional job args for tycache attributes
+            if additional_attributes:
+                additional_args = self.get_additional_attribute_args(
+                    operator, additional_attributes
+                )
+                job_args.extend(additional_args)
+            tycache_export = f"{operator}.exportTyCache()"
+            job_args.append(tycache_export)
+
+        return job_args
```
BigRoy, line 138: I wonder if this is more readable.
```suggestion
        settings = {
            "exportMode": 2,
            "frameStart": start,
            "frameEnd": end,
            "tyCacheFilename": filepath.replace("\\", "/")
        }
        settings.update(additional_attributes)
        
        job_args = []
        for operator in self.get_operators(members):
            for key, value in settings.items():
                if isinstance(value, str):
                    # embed in quotes
                    value = f'"{value}"'
            
                job_args.append(f"{operator}.{key}=[value}")

            job_args.append(f"{operator}.exportTyCache()")

        return job_args
```

Note that I've removed the `bool` conversion to always be `True` . I assumed it'd be weird that if for whatever reason there's a bool value in additional attributes like: `key=False` that you'd still be forcing it `key=True` in the job args in the `get_additional_attribute_args()` function. Seemed like a bug.

openpype/hosts/max/plugins/publish/validate_tyflow_data.py
```diff
@@ -0,0 +1,96 @@
...
+        invalid = []
+        container = instance.data["instance_node"]
+        self.log.debug(f"Validating tyFlow container for {container}")
+
+        selection_list = instance.data["members"]
+        for sel in selection_list:
+            sel_tmp = str(sel)
+            if rt.ClassOf(sel) in [rt.tyFlow,
+                                   rt.Editable_Mesh]:
+                if "tyFlow" not in sel_tmp:
```
BigRoy, line 58: Why does it need `tyFlow` in the name if you're already checking the class type?

openpype/hosts/max/plugins/publish/validate_tyflow_data.py
```diff
@@ -0,0 +1,96 @@
...
+            for anim_name in anim_names:
+                # get all the names of the related tyFlow nodes
+                sub_anim = rt.GetSubAnim(obj, anim_name)
+                # check if there is export particle operator
+                boolean = rt.IsProperty(sub_anim, "Export_Particles")
+                bool_list.append(str(boolean))
+            # if the export_particles property is not there
+            # it means there is not a "Export Particle" operator
+            if "True" not in bool_list:
+                invalid.append(sel)
```
BigRoy, line 94: This is weird - this would consider `sel` invalid until it has found at least one other object in selection that has `bool_list` set to True.

Something is off here - that must be a bug. Did you mean to put the `bool_list` variable INSIDE the for loop so it triggers per `sel`?


Did you mean something like this?
```suggestion
        members = instance.data["members"]
        for member in members:
            obj = member.baseobject
            
            # There must be at least one animation with export 
            # particles enabled
            has_export_particles = False
            anim_names = rt.GetSubAnimNames(obj)
            for anim_name in anim_names:
                # get name of the related tyFlow node
                sub_anim = rt.GetSubAnim(obj, anim_name)
                # check if there is export particle operator
                if rt.IsProperty(sub_anim, "Export_Particles"):
                    has_export_particles = True
                    break
                
            if not has_export_particles:
                invalid.append(member)
```

openpype/hosts/max/plugins/publish/collect_tycache_attributes.py
```diff
@@ -0,0 +1,76 @@
+import pyblish.api
+
+from openpype.lib import EnumDef, TextDef
+from openpype.pipeline.publish import OpenPypePyblishPluginMixin
+
+
+class CollectTyCacheData(pyblish.api.InstancePlugin,
+                         OpenPypePyblishPluginMixin):
+    """Collect Review Data for Preview Animation"""
```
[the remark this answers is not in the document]
moonyuet: No it's not

openpype/hosts/max/plugins/publish/extract_tycache.py
moonyuet: Why you are removing it? The tyflow channel data needs to be set to True in order to get the data export.

moonyuet: It's not a bug, it's just hardcoded.

BigRoy: Could you provide an example of the input data where that is not a bug? :)

Say `additional_attributes` is:
```python
additional_attributes = {
    "key1": True,
    "key2": False,
    "key3": True
}
```

I don't ever see a reason to set key2 to True? :p

openpype/hosts/max/plugins/publish/extract_tycache.py
```diff
@@ -0,0 +1,183 @@
...
+        additional_args = []
+        for key, value in attrs.items():
+            tyc_attribute = None
+            if isinstance(value, bool):
+                tyc_attribute = f"{operator}.{key}=True"
+            elif isinstance(value, str):
+                tyc_attribute = f'{operator}.{key}="{value}"'
+            additional_args.append(tyc_attribute)
+        self.log.debug(additional_args)
+        return additional_args
```
BigRoy, line 183:
```suggestion

```

openpype/hosts/max/plugins/publish/extract_tycache.py
```diff
@@ -0,0 +1,183 @@
...
+        mesh_repres = {
+            'name': 'tyMesh',
+            'ext': 'tyc',
+            'files': mesh_filename,
+            "stagingDir": stagingdir,
+            "outputName": '__tyMesh'
+        }
+        representations.append(mesh_repres)
+        self.log.info(
+            f"Extracted instance '{instance.name}' to: {mesh_filename}")
```
BigRoy, line 71:
```suggestion
        self.log.debug(f"Extracted instance '{instance.name}' to: {filenames}")

        # Get the tyMesh filename for extraction
        mesh_filename = f"{instance.name}__tyMesh.tyc"
        mesh_repres = {
            'name': 'tyMesh',
            'ext': 'tyc',
            'files': mesh_filename,
            "stagingDir": stagingdir,
            "outputName": '__tyMesh'
        }
        representations.append(mesh_repres)
        self.log.debug(
            f"Extracted instance '{instance.name}' to: {mesh_filename}")
```
Also, are you sure you want to report both with the same `Extracted instance ` message? Maybe a bit confusing to see twice?

openpype/hosts/max/plugins/publish/extract_tycache.py
```diff
@@ -0,0 +1,183 @@
...
+    order = pyblish.api.ExtractorOrder - 0.2
+    label = "Extract TyCache"
+    hosts = ["max"]
+    families = ["tycache"]
+
+    def process(self, instance):
+        # TODO: let user decide the param
+        start = int(instance.context.data["frameStart"])
+        end = int(instance.context.data.get("frameEnd"))
+        self.log.info("Extracting Tycache...")
```
BigRoy, line 35:
```suggestion
        self.log.debug("Extracting Tycache...")
```

openpype/hosts/max/plugins/publish/extract_tycache.py
moonyuet: Maybe just do it once for tyCache part as we usually load tyCache instead of tyMesh

openpype/hosts/max/plugins/load/load_tycache.py
```diff
@@ -0,0 +1,65 @@
...
+    families = ["tycache"]
+    representations = ["tyc"]
+    order = -8
+    icon = "code-fork"
+    color = "green"
+
+    def load(self, context, name=None, namespace=None, data=None):
+        """Load tyCache"""
+        from pymxs import runtime as rt
+        filepath = os.path.normpath(self.filepath_from_context(context))
```
BigRoy, line 28: Looking at the trace for `self.filepath_from_context(context)` it ends up calling `get_representation_path` which seems to already return a normalized path (if the file exists that is).
```suggestion
        filepath = self.filepath_from_context(context)
```

openpype/hosts/max/plugins/publish/extract_tycache.py
```diff
@@ -0,0 +1,159 @@
...
+from pymxs import runtime as rt
+
+from openpype.hosts.max.api import maintained_selection
+from openpype.pipeline import publish
+
+
+class ExtractTyCache(publish.Extractor):
+    """
+    Extract tycache format with tyFlow operators.
+    Notes:
```
BigRoy, line 13:
```suggestion
    """Extract tycache format with tyFlow operators.
    
    Notes:
```
I've mostly seen this code style.

openpype/hosts/max/plugins/publish/extract_tycache.py
```diff
@@ -0,0 +1,159 @@
...
+            instance (pyblish.api.Instance): instance.
+            start_frame (int): Start frame.
+            end_frame (int): End frame.
+
+        Returns:
+            filenames(list): list of filenames
+
+        """
+        filenames = []
+        # should we include frame 0 ?
```
BigRoy, line 92: Is this comment still relevant? Should it be a `TODO`? Also not entirely sure where that frame 0 comes from?

openpype/hosts/max/plugins/publish/extract_tycache.py
```diff
@@ -0,0 +1,159 @@
...
+            members (list): Instance members.
+
+        Returns:
+            list of particle operators
+
+        """
+        opt_list = []
+        for member in members:
+            obj = member.baseobject
+        # TODO: to see if it can be used maxscript instead
```
BigRoy, line 149:
```suggestion
            # TODO: see if it can use maxscript instead
```

openpype/hosts/max/plugins/publish/validate_tyflow_data.py
```diff
@@ -0,0 +1,91 @@
+import pyblish.api
+from openpype.pipeline import PublishValidationError
+from pymxs import runtime as rt
+
+
+class ValidateTyFlowData(pyblish.api.InstancePlugin):
+    """Validate that TyFlow plugins or
+    relevant operators being set correctly."""
```
BigRoy, line 8:
```suggestion
    """Validate TyFlow plugins or relevant operators are set correctly."""
```

openpype/hosts/max/plugins/publish/validate_tyflow_data.py
```diff
@@ -0,0 +1,91 @@
...
+        invalid_operator = self.get_tyflow_operator(instance)
+        if invalid_operator:
+            self.log.error(
+                "Operator 'Export Particles' not found in tyFlow editor.")
+        if invalid_object or invalid_operator:
+            raise PublishValidationError(
+                "issues occurred",
+                description="Container should only include tyFlow object\n "
+                "and tyflow operator 'Export Particle' should be in \n"
+                "the tyFlow editor")
```
BigRoy, line 36: The new lines `\n` seem to be in odd locations here mid sentences.

```suggestion
                description="Container should only include tyFlow object "
                "and tyflow operator 'Export Particle' should be in "
                "the tyFlow editor.")
```

openpype/hosts/max/plugins/publish/validate_tyflow_data.py
```diff
@@ -0,0 +1,91 @@
...
+    def get_tyflow_object(self, instance):
+        """Get the nodes which are not tyFlow object(s)
+        and editable mesh(es)
+
+        Args:
+            instance (pyblish.api.Instance): instance
+
+        Returns:
+            invalid(list): list of invalid nodes which are not
+            tyFlow object(s) and editable mesh(es).
```
BigRoy, line 47:
```suggestion
        Returns:
            list: invalid nodes which are not tyFlow 
                object(s) and editable mesh(es).
```

openpype/hosts/max/plugins/publish/validate_tyflow_data.py
```diff
@@ -0,0 +1,91 @@
...
+        invalid = []
+        container = instance.data["instance_node"]
+        self.log.debug(f"Validating tyFlow container for {container}")
+
+        selection_list = instance.data["members"]
+        for sel in selection_list:
+            if rt.ClassOf(sel) not in [rt.tyFlow, rt.Editable_Mesh]:
+                invalid.append(sel)
+
+        return invalid
```
BigRoy, line 58:
```suggestion
        allowed_classes = [rt.tyFlow, rt.Editable_Mesh]
        return [
            member for member in instance.data["members"]
            if rt.ClassOf(member) not in allowed_classes
        ]
```

openpype/hosts/max/plugins/publish/extract_tycache.py
```diff
@@ -0,0 +1,157 @@
...
+            members (list): Instance members.
+
+        Returns:
+            list of particle operators
+
+        """
+        opt_list = []
+        for member in members:
+            obj = member.baseobject
+        # TODO: see if it can use maxscript instead
```
BigRoy, line 147: Fix indentation
```suggestion
            # TODO: see if it can use maxscript instead
```

BigRoy (approved):
Can't test - but code looks good to me.

LiborBatek (changes requested):
TyCache being produced ok, but during Load via loader it fails to set the path to the ```tyCache``` produced.
image

When trying manually put the path in, the ```tyCache``` correctly loads and all fine.
image

Need to be adressed.

moonyuet:
> TyCache being produced ok, but during Load via loader it fails to set the path to the `tyCache` produced. image
> 
> When trying manually put the path in, the `tyCache` correctly loads and all fine. image
> 
> Need to be adressed.

@LiborBatek I restored the original code for the loader plugins for the test if it is being loaded correctly. Can you test again? Thanks.

LiborBatek:
Works like charm...Loads the ```tyCache``` immedeately...
image

openpype/hosts/max/plugins/load/load_tycache.py
```diff
@@ -0,0 +1,64 @@
...
+
+        return containerise(
+            name, [obj], context,
+            namespace, loader=self.__class__.__name__)
+
+    def update(self, container, representation):
+        """update the container"""
+        from pymxs import runtime as rt
+
+        path = get_representation_path(representation)
```
BigRoy, line 45: If the `os.path.normpath` fixed the issue on `load` shouldn't we also do it on `update`?

@LiborBatek can you confirm updating also works fine?

LiborBatek:
This is what I get when trying to set version via ```Manage```

image

BigRoy:
> This is what I get when trying to set version via `Manage`
> 
> image
> 
> image
> 
> Seems like when `Frame Range` in the active scene doesnt match the asset one it fails on it...

@LiborBatek this might be something with your specific environment where the previous publish does not also have a `tyc` representation - likely because when you published that one a longer time ago this PR was still generating different publishes or there was a bug at the time that made it fail, etc.. You might be better of doing another v10 and v11 publish and then see if you can switch to the newer ones and between them.

LiborBatek:
@BigRoy hmm did try to produce brand new publishes and switching between those only...still fails. but it cant find the 
```tycacheMain__tyPart_00000_00000.tyc``` file

BigRoy:
> did try to produce brand new publishes and switching between those only...still fails. but it cant find the
> `tycacheMain__tyPart_00000_00000.tyc` file

Just to be clear - you are **not** getting the same error, but now it just fails to find the right path to set on the node? Does it give any errors?

LiborBatek:
This also happens when brand new asset created and frame range set to more ussual ```1001-1100```

Does the same error:
image

obivously we are looking for wrong file name, here are the actual ones:
image

moonyuet:
> This also happens when brand new asset created and frame range set to more ussual `1001-1100`
> 
> Does the same error: image
> 
> obivously we are looking for wrong file name, here are the actual ones: image

can you test again to see if the issue still exists?

LiborBatek:
Well that commit didnt helped. at least here is more of a log (sorry for being it scrngrab but its faster when in remote on site)

image

moonyuet:
> Well that commit didnt helped. at least here is more of a log (sorry for being it scrngrab but its faster when in remote on site)
> 
> image

can you go to the error path {publish_folder}/{version} to see if there is anything?

LiborBatek (approved):
So after adjusting the ```Template``` in Projects Settings for ```tycache``` all working nicely!

I was able to produce publish of ```tyCache``` (several asset versions for switching tests) successfuly.

I was able to ```Load``` the published ```tyCache``` in particular version selected.

And lastly I was able to successfuly ```Manage``` versions of it by ```Set version``` and ```Update to Latest ver```.

I also tried to load multiple ```tyCache``` assets and ```Manage``` those successfuly too...

Im enclosing the video for all of that.

https://github.com/ynput/OpenPype/assets/112623825/ce56f6dd-123e-497d-9b5e-0b4212dc2f15

moonyuet:
@antirotor can you do the final check before the merge? Thanks!

antirotor (approved):

PR #5629, opened 2023-09-18
Chore: Lower click dependency version

iLLiCiTiT:
## Changelog Description
Lower click version to support older versions of python.

## Additional info
The newest version is only for python >=3.7 which cause issues with addons that have defined their cli methods.

## Testing notes:
1. Run create env
2. Run older DCC via OpenPype (ideally with Python 2.x)
3. Test publishing with addons that use click for cli commands (e.g. ftrack)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -560,17 +560,14 @@ files = [
 
 [[package]]
 name = "click"
-version = "8.1.3"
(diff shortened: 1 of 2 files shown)


64qam (approved):
lgbt

PR #5630, opened 2023-09-18
Maya: Support new publisher for colorsets validation.

tokejepsen:
## Changelog Description
Fix `validate_color_sets` for the new publisher.

In current `develop` the repair option does not appear due to wrong error raising.

## Testing notes:
1. Setup multiple colorsets on a mesh in Maya with `Modeling > Mesh Display > Color Sets Editor`
2. Publish and repair through publisher.

### DIFF
--- a/openpype/hosts/maya/plugins/publish/validate_color_sets.py
+++ b/openpype/hosts/maya/plugins/publish/validate_color_sets.py
@@ -3,9 +3,10 @@ from maya import cmds
 import pyblish.api
 import openpype.hosts.maya.api.action
 from openpype.pipeline.publish import (
-    RepairAction,
     ValidateMeshOrder,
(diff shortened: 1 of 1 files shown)


BigRoy:
Note that this fix is also included in this PR https://github.com/ynput/OpenPype/pull/5559/files#diff-681791944245f024737c7408ed4e704c98c53d355e989b7b505a1bc08c698c47 

Aside of that - I noticed just now that the first line of the docstring for this plug-in is wrong; it's likely copied from another plugin. Cosmetics maybe but might be nice to fix.

LiborBatek (approved):
The ```validator``` correctly kicks in when some ```color set``` present on the mesh.

image


 Repair function works too and Publish then happens without any issues.
image

Minkiu (approved):

PR #5632, opened 2023-09-18
Maya: Remove context prompt.

tokejepsen:
## Changelog Description
More of a plea than a PR, but could we please remove the context prompt in Maya when switching tasks?

## Testing notes:
1. Switch task in Maya.
2. Validate no prompt appears about which context you are entering.


BigRoy:
Haha, yes please - we've done that originally too. We've now replaced it however with a pop-up about what settings the user wants to reset on the context change but ONLY if the user is saving somewhere else - not when opening.

BigRoy (approved):

antirotor (approved):

PR #5635, opened 2023-09-19
nuke: extract review data mov read node with expression

jakubjezek001:
## Changelog Description
Some productions might have set default values for read nodes, those settings are not colliding anymore now.

## Testing notes:
1. all should work as it was before


jakubjezek001:
> Publishing fails with error:

This doesnt look like the issue of this PR,  but it rather seems to be problem with your testing data. Please try to recreate them.

64qam (approved):
working now

PR #5637, opened 2023-09-19
TVPaint: Fix review family extraction

iLLiCiTiT:
## Changelog Description
Extractor marks representation of review instance with review tag.

## Testing notes:
1. Open TVPaint
2. Review auto-creator should create an instance
3. Enable the instance
4. Publish
5. Extract review should catch the representation and create review (based on settings)

### DIFF
--- a/openpype/hosts/tvpaint/plugins/publish/extract_sequence.py
+++ b/openpype/hosts/tvpaint/plugins/publish/extract_sequence.py
@@ -108,11 +108,16 @@ class ExtractSequence(pyblish.api.Extractor):
             "Files will be rendered to folder: {}".format(output_dir)
         )
 
+        # Fill tags and new families from project settings
+        tags = []
         if instance.data["family"] == "review":
+            tags.append("review")
             result = self.render_review(
                 output_dir, mark_in, mark_out, scene_bg_color
             )
         else:
+            if "review" in instance.data["families"]:
+                tags.append("review")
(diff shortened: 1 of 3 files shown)


64qam (approved):

openpype/hosts/tvpaint/plugins/publish/extract_sequence.py
```diff
@@ -108,11 +108,16 @@ def process(self, instance):
...
+        # Fill tags and new families from project settings
+        tags = []
         if instance.data["family"] == "review":
+            tags.append("review")
             result = self.render_review(
                 output_dir, mark_in, mark_out, scene_bg_color
             )
         else:
+            if "review" in instance.data["families"]:
+                tags.append("review")
```
BigRoy, line 120: So basically if there's any review family you always want to append the review tag?

The oddity that `family` isn't always present in `families` has come up before as confusing but and the workaround of making sure to merge the two is usually done.

```python
families = [instance.data["family"]] + instance.data.get("families", [])
if "review" in families:
    tags.append("review")
```
Would this make it easier to understand that the tag appears always if "review" is one of its families? To me that's much clearer than basically doing the check across the if else branches?

```python
families = get_instance_families(instance)
if "review" in families:
    tags.append("review")
```
I recall a function like this also being present somewhere in lib or somewhere.

PR #5639, opened 2023-09-21
AYON settings: Extract OIIO transcode settings

iLLiCiTiT:
## Changelog Description
Output definitions of Extract OIIO transcode have name to match OpenPype settings, and the settings are converted to dictionary in settings conversion.

## Additional info
Added partial fix of older AYON settings.

## Testing notes:
1. Extract OIIO transcode plugin should work with filled profiles


LiborBatek (approved):
Checking ```Studio Settings``` and ```OIIO Transcode``` options present on AYON.
image

PR #5643, opened 2023-09-22
Unreal: Yeti support

simonebarbieri:
## Changelog Description
Implemented Yeti support for Unreal.

## Additional info
A new family has been implemented, `ue_yeticache`, to differentiate from the existing `yeticache`. Yeti has different exporter to be compatible with the Unreal built-in Groom system, so it needs a different extractor.

## Testing notes:
1. In Maya, import or create a Yeti Node
2. Create a new instance of family Unreal Yeti Cache and publish.
3. In Unreal, load the published asset.

### DIFF
--- a/openpype/hosts/maya/plugins/create/create_unreal_yeticache.py
+++ b/openpype/hosts/maya/plugins/create/create_unreal_yeticache.py
@@ -10,7 +10,7 @@ class CreateYetiCache(plugin.MayaCreator):
 
     identifier = "io.openpype.creators.maya.unrealyeticache"
     label = "Unreal Yeti Cache"
-    family = "ue_yeticache"
+    family = "yeticacheUE"
     icon = "pagelines"
 
     def get_instance_attr_defs(self):
--- a/openpype/hosts/maya/plugins/create/create_unreal_yeticache.py
+++ b/openpype/hosts/maya/plugins/create/create_unreal_yeticache.py
@@ -9,7 +9,7 @@ class CreateYetiCache(plugin.MayaCreator):
     """Output for procedural plugin nodes of Yeti """
 
     identifier = "io.openpype.creators.maya.unrealyeticache"
-    label = "Unreal Yeti Cache"
+    label = "Unreal - Yeti Cache"
     family = "yeticacheUE"
     icon = "pagelines"
 
--- a/openpype/hosts/maya/plugins/create/create_unreal_yeticache.py
+++ b/openpype/hosts/maya/plugins/create/create_unreal_yeticache.py
@@ -1,0 +1,39 @@
+from openpype.hosts.maya.api import (
+    lib,
+    plugin
+)
+from openpype.lib import NumberDef
+
+
+class CreateYetiCache(plugin.MayaCreator):
+    """Output for procedural plugin nodes of Yeti """
+
+    identifier = "io.openpype.creators.maya.unrealyeticache"
+    label = "Unreal Yeti Cache"
+    family = "ue_yeticache"
+    icon = "pagelines"
+
+    def get_instance_attr_defs(self):
+
+        defs = [
+            NumberDef("preroll",
+                      label="Preroll",
+                      minimum=0,
+                      default=0,
+                      decimals=0)
+        ]
+
+        # Add animation data without step and handles
+        defs.extend(lib.collect_animation_defs())
+        remove = {"step", "handleStart", "handleEnd"}
+        defs = [attr_def for attr_def in defs if attr_def.key not in remove]
+
+        # Add samples after frame range
+        defs.append(
+            NumberDef("samples",
+                      label="Samples",
+                      default=3,
+                      decimals=0)
+        )
+
+        return defs
--- a/openpype/hosts/maya/plugins/publish/extract_unreal_yeticache.py
+++ b/openpype/hosts/maya/plugins/publish/extract_unreal_yeticache.py
@@ -1,0 +1,62 @@
+import os
+import json
+
+from maya import cmds
+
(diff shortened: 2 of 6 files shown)


LiborBatek (changes requested):
1. Publish from maya is working fine

image

2. Loading into UE works however if not ```Groom``` addons activated only ```Main_CON``` present and user do not get any sort of warning basically ending up with empty ```asset``` without any actual data for ```yeti```
3. When trying to use ```Manage``` and ```Update version to latest``` it will happen but overwriting the existent version so no new version being produced.  We need to treat it as versioned the same way we do for ```uasset``` we are producing via Ayon tools. So new folder version should be produced.

image


4. The product name imho not being ideal now...

```ue_yeticache```

because its not sitting alphabetically next to other ```yeti``` products Ayon is capable of as seen here
image

it should be something like ```yetiCacheUE``` or similar

**Comment by simonebarbieri** [comment]
I should have fixed all your points @LiborBatek :)

The only note is about this:
> 2. Loading into UE works however if not `Groom` addons activated only `Main_CON` present and user do not get any sort of warning basically ending up with empty `asset` without any actual data for `yeti`

Unreal Python API don't allow to check if a plugin is active, so I implemented a sort of workaround to check if the plugin is in the Unreal Project file!

File: openpype/hosts/maya/plugins/create/create_unreal_yeticache.py [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+        defs = [
+            NumberDef("preroll",
+                      label="Preroll",
+                      minimum=0,
+                      default=0,
+                      decimals=0)
+        ]
+
+        # Add animation data without step and handles
+        defs.extend(lib.collect_animation_defs())
```
Minkiu, line 27: Couldn't this be added when we initialize `defs` ?:
```py
defs = [
	NumberDef("preroll",
		label="Preroll",
		minimum=0,
		default=0,
		decimals=0),
	lib.collect_animation_defs()
]

File: openpype/hosts/maya/plugins/publish/extract_unreal_yeticache.py [hunk 2]
```diff
@@ -0,0 +1,61 @@
...
+
+        kwargs = {}
+        samples = instance.data.get("samples", 0)
+        if samples == 0:
+            kwargs.update({"sampleTimes": "0.0 1.0"})
+        else:
+            kwargs.update({"samples": samples})
+
+        self.log.debug(f"Writing out cache {start_frame} - {end_frame}")
+        filename = "{0}.abc".format(instance.name)
```
Minkiu, line 42: If we are using `f-trsings` might as well do it here too?
```py
filename = f"{instance.name}.abc"

File: openpype/hosts/maya/plugins/create/create_unreal_yeticache.py (hunk 1 above)
Comment (simonebarbieri) [reply openpype/hosts/maya/plugins/create/create_unreal_yeticache.py:27]: Technically yes, but in the lines below we're removing a few keys from that, so I think it makes the code more readable this way. What do you think?

File: openpype/hosts/maya/plugins/publish/extract_unreal_yeticache.py [hunk 3]
```diff
@@ -0,0 +1,61 @@
...
+        path = os.path.join(dirname, filename)
+        cmds.pgYetiCommand(yeti_nodes,
+                           writeAlembic=path,
+                           range=(start_frame, end_frame),
+                           asUnrealAbc=True,
+                           **kwargs)
+
+        if "representations" not in instance.data:
```
Minkiu, line 50: This could be:
```py
representations = instance.data.setdefault("representations", [])
representations.append({
	'name': 'abc',
	'ext': 'abc',
	'files': filename,
	'stagingDir': dirname
})
```

or:
````py
instance.data.setdefault("representations", []).append({
	'name': 'abc',
	'ext': 'abc',
	'files': filename,
	'stagingDir': dirname
})

File: openpype/hosts/unreal/plugins/load/load_yeticache.py [hunk 4]
```diff
@@ -0,0 +1,179 @@
...
+
+        tools = unreal.AssetToolsHelpers().get_asset_tools()
+        asset_dir, container_name = tools.create_unique_asset_name(
+            f"{root}/{asset}/{name}", suffix="")
+
+        unique_number = 1
+        while unreal.EditorAssetLibrary.does_directory_exist(
+            f"{asset_dir}_{unique_number:02}"
+        ):
+            unique_number += 1
```
Minkiu, line 102: I don't get what this `while` is trying to do? Would it raise if the dir doesnt exist? We check again in line `107`?

LiborBatek (approved):
Now all works as expected.

User gets warning aka failure message if ```Groom addons``` arent loaded in ```UE```.

image


Each ```yetiCacheUE``` having own ```version``` folder when Imported new version to UE.

image

openpype/hosts/maya/plugins/create/create_unreal_yeticache.py
Minkiu: If it's for readability it's fine then; but by the time we do the removal they are already merged, it's just happens in two steps instead of one, or I got it wrong?

openpype/hosts/unreal/plugins/load/load_yeticache.py
simonebarbieri: We load the asset in a unique folder, so if the asset already exists (`asset_01`), it would be loaded as `asset_02`. The while exists so that if we have deleted one asset and i.e. we have `asset_01` and `asset_03`, `02` would free again, so we would load it as `asset_02`.

PR #5648, opened 2023-09-25
traypublisher: missing `assetEntity` in context data

jakubjezek001:
## Changelog Description
Issue with missing `assetEnity` key in context data is not problem anymore. 

## Testing notes:
1. Try to publish plate imagesequence in traypublisher
2. enable `Collect Original Sequence Frame Data` toggle on instance and publish
3. Open Tray Loader and check the framerange is inherited form original sequence and handles are set to 0
4. reload the tray publishing context
5. desable`Collect Original Sequence Frame Data` toggle on instance and publish
6. publish again 
7. Open Tray Loader and check the framerange is inherited form asset frame range including handles

### DIFF
--- a/openpype/hosts/traypublisher/plugins/publish/collect_frame_data_from_asset_entity.py
+++ b/openpype/hosts/traypublisher/plugins/publish/collect_frame_data_from_asset_entity.py
@@ -2,18 +2,18 @@ import pyblish.api
 from openpype.pipeline import OptionalPyblishPluginMixin
 
 
-class CollectMissingFrameDataFromAssetEntity(
+class CollectFrameDataFromAssetEntity(
     pyblish.api.InstancePlugin,
     OptionalPyblishPluginMixin
 ):
-    """Collect Missing Frame Range data From Asset Entity
+    """Collect Frame Data From AssetEntity found in context
 
     Frame range data will only be collected if the keys
     are not yet collected for the instance.
     """
 
     order = pyblish.api.CollectorOrder + 0.491
-    label = "Collect Missing Frame Data From Asset Entity"
+    label = "Collect Frame Data From Asset"
     families = ["plate", "pointcache",
                 "vdbcache", "online",
                 "render"]
@@ -23,7 +23,9 @@ class CollectMissingFrameDataFromAssetEntity(
     def process(self, instance):
         if not self.is_active(instance.data):
             return
-        missing_keys = []
+
+        asset_data = instance.data["assetEntity"]["data"]
+
         for key in (
             "fps",
             "frameStart",
@@ -31,14 +33,5 @@ class CollectMissingFrameDataFromAssetEntity(
             "handleStart",
             "handleEnd"
         ):
-            if key not in instance.data:
-                missing_keys.append(key)
-        keys_set = []
-        for key in missing_keys:
-            asset_data = instance.data["assetEntity"]["data"]
-            if key in asset_data:
-                instance.data[key] = asset_data[key]
-                keys_set.append(key)
-        if keys_set:
-            self.log.debug(f"Frame range data {keys_set} "
-                           "has been collected from asset entity.")
+            instance.data[key] = asset_data[key]
+            self.log.debug(f"Collected Frame range data '{key}':{asset_data[key]} ")
--- a/openpype/hosts/traypublisher/plugins/publish/collect_frame_data_from_asset_entity.py
+++ b/openpype/hosts/traypublisher/plugins/publish/collect_frame_data_from_asset_entity.py
@@ -1,11 +1,7 @@
 import pyblish.api
-from openpype.pipeline import OptionalPyblishPluginMixin
 
 
-class CollectFrameDataFromAssetEntity(
-    pyblish.api.InstancePlugin,
-    OptionalPyblishPluginMixin
-):
+class CollectFrameDataFromAssetEntity(pyblish.api.InstancePlugin):
     """Collect Frame Data From AssetEntity found in context
 
     Frame range data will only be collected if the keys
@@ -18,14 +14,9 @@ class CollectFrameDataFromAssetEntity(
                 "vdbcache", "online",
                 "render"]
     hosts = ["traypublisher"]
-    optional = True
 
     def process(self, instance):
-        if not self.is_active(instance.data):
-            return
-
-        asset_data = instance.data["assetEntity"]["data"]
-
+        missing_keys = []
         for key in (
             "fps",
             "frameStart",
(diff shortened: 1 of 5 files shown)


openpype/plugins/publish/collect_sequence_frame_data.py
```diff
@@ -28,6 +28,10 @@ def process(self, instance):
 
     def get_frame_data_from_repre_sequence(self, instance):
         repres = instance.data.get("representations")
+        parent_entity = (
```
iLLiCiTiT, line 31: To repeat myself from previous PR. This must be only `instance.data["assetEntity"]` or not filled at all.

iLLiCiTiT: BTW You've also renamed `CollectFrameDataFromAssetEntity` to `CollectMissingFrameDataFromAssetEntity`, but didn't change settings name.

iLLiCiTiT: Well you've changed it to use `"assetEntity"` from `instance.data`, but that is filled only after `CollectAnatomyInstanceData` which is always after this plugin (+0.49). So I guess we have to create new collector plugin which adds those data from the asset entity.

iLLiCiTiT: I just realized. Why this does not happen in `CollectMissingFrameDataFromAssetEntity`? It has same family filtering and the plugin is in order which already have set `assetEntity` on instance data?

openpype/plugins/publish/collect_sequence_frame_data.py
```diff
@@ -28,10 +28,12 @@ def process(self, instance):
...
         repres = instance.data.get("representations")
-        parent_entity = (
-            instance.data.get("assetEntity")
-            or instance.context.data["projectEntity"]
-        )
+        parent_entity = instance.data.get("assetEntity")
+
+        if not parent_entity:
+            self.log.warning("Cannot find parent entity data")
+            return
```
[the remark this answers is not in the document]
jakubjezek001: yes agree, but it was rewritten so resolving this

jakubjezek001:
Just to summarize the current state of the feature rework, the original pull request (PR) that addressed the need to preserve sequential data found on representation, rather than always publishing asset entity collected frame data, can be found here: [OpenPype PR #5530](https://github.com/ynput/OpenPype/pull/5530). This PR also resolved an issue where fps and handles were missing by adding them. And because the "CollectFrameDataFromAssetEntity" plugin only added the data if it was missing, I have decided to rename it to "CollectFrameDataFromAssetEntity".

This update takes a different approach to the issue. Instead of only filling missing data, it now adds asset entity data when needed. As a result, I am renaming the plugin back to CollectFrameDataFromAssetEntity, as it aligns better with its purpose and also it fixes the issue with unliked settings metioned [here](https://github.com/ynput/OpenPype/pull/5648#discussion_r1336090708). By default, it adds sequential data from the representation in CollectSequenceFrameData.

I have observed that the asset entity data is already added when the sequential data is collected. Therefore, if no sequential representation data is found, it will remain in its original state and nothing is missing.

openpype/hosts/traypublisher/plugins/publish/collect_frame_data_from_asset_entity.py
```diff
@@ -23,22 +23,16 @@ class CollectMissingFrameDataFromAssetEntity(
...
-        keys_set = []
-        for key in missing_keys:
-            asset_data = instance.data["assetEntity"]["data"]
-            if key in asset_data:
-                instance.data[key] = asset_data[key]
-                keys_set.append(key)
-        if keys_set:
-            self.log.debug(f"Frame range data {keys_set} "
-                           "has been collected from asset entity.")
+            instance.data[key] = asset_data[key]
```
iLLiCiTiT, line 36: You should not override the values if are already filled, and you should not expect thay are filled on asset data. As far as I remember, during editorial are all of these values prefilled from completelly different source and should not be change by asset entity, which may not exist yet.

openpype/hosts/traypublisher/plugins/publish/collect_frame_data_from_asset_entity.py
```diff
@@ -23,22 +23,16 @@ class CollectMissingFrameDataFromAssetEntity(
...
+            self.log.debug(
+                f"Collected Frame range data '{key}':{asset_data[key]} ")
```
BigRoy, line 38: This logic will now ALWAYS overwrite the collected frame range data which is NOT what I'd expect as a user if my input footage has actual frames in its filename - unless maybe this particular plugin is disabled by default?

Should maybe the default `active` state for this particular plugin be "false" in settings so that it requires the user to ENABLE it to actually say 'no matter what - just use the frame range of the asset as data'?

jakubjezek001:
Ok so after call with @iLLiCiTiT we had decided to do it differently:

- sequencial original frame data should be optional plugin
- sequential data are added if activated
- asset data frame data are not optional anymore and are added only if missing

64qam (approved):
lgbt

openpype/settings/defaults/project_settings/traypublisher.json
```diff
@@ -346,10 +346,10 @@
         }
     },
     "publish": {
-        "CollectFrameDataFromAssetEntity": {
+        "CollectSequenceFrameData": {
             "enabled": true,
             "optional": true,
-            "active": true
+            "active": false
```
BigRoy, line 352: I think this is a weird default - for 99% of the files you'd drop into the publisher why wouldn't you want the frame range you're dropping in to be the frame data?

If I drop in files:
```
seq.001.exr, seq.002.exr, seq.003.exr, seq.004.exr
```
I'd expect seq to be frame 1-4 - not an arbitrary asset frame range which (might just as well be completely different, e.g. asset might be 1001-1150. Now suddenly the frame data shows 150 frames instead of the four frames the sequence actually exists of?

I feel like that'd be confusing for the majority of users as default behavior, no?
Pretty sure the default here should be `active: true` and maybe even `optional: false`?

BigRoy: > This would be an issue with editorial data. This is rather safe to have it as optional.

I mean - isn't that an issue then with the actual editorial data system? :) Shouldn't that itself already provide frame ranges and if set - this plugin should not override it?

BigRoy: Hope this doesn't come off as a rant but do want to highlight this:

I'm a bit worried about PRs getting merged with conversations getting marked resolved that didn't reach a valid conclusion. I want to highlight that this is already, what?, the third PR in succession covering this particular 'issue' where the fix got implemented the wrong way. It's weird that these then, especially after how much has already gone wrong with it, get merged within just a few hours span of the comments that were made on it.

It might be me being completely off here - but I really see no reason why uploading a file sequence in the tray publisher should **not** default to the frame range of the actual sequence I'm throwing into it. If it causes bugs for Editorial publishing then what we'd need to solve is those bugs and not set some odd default behavior because it's "safer". What's holding anyone from enabling this checkbox on Editorial publishing? It's just as broken if this feature is really what was causing issues. As such, it feels like this PR solved nothing except for making more odd design choices.

PR #5649, opened 2023-09-25
Extended error message when getting subset name

kalisp:
## Changelog Description
Each Creator is using `get_subset_name` functions which collects context data and fills configured template with placeholders.
If any key is missing in the template, non descriptive error is thrown.
Discord_vbHC59nQ1t

This should provide more verbose message:
AnyDeskMSI_HEl3Q0LbPq

## Testing notes:
1. configure `project_settings/global/tools/creator/subset_name_profiles` for particular host and add there non existent placeholder value
2. open Publisher and create instance that will use that template

### DIFF
--- a/openpype/pipeline/create/subset_name.py
+++ b/openpype/pipeline/create/subset_name.py
@@ -112,6 +112,10 @@ def get_subset_name(
             for project. Settings are queried if not passed.
         family_filter (Optional[str]): Use different family for subset template
             filtering. Value of 'family' is used when not passed.
+
+    Raises:
+        KeyError if filled template contains placeholder key which is not
+            collected.
     """
 
     if not family:
@@ -154,4 +158,10 @@ def get_subset_name(
         for key, value in dynamic_data.items():
             fill_pairs[key] = value
 
-    return template.format(**prepare_template_data(fill_pairs))
+    try:
+        return template.format(**prepare_template_data(fill_pairs))
+    except KeyError as exp:
+        error = "Value for '{}' key is missing in template '{}'.".format(
+            str(exp), template)
+        error += "Collected values are {}".format(fill_pairs)
+        raise KeyError(error)


openpype/pipeline/create/subset_name.py
```diff
@@ -154,4 +158,10 @@ def get_subset_name(
         for key, value in dynamic_data.items():
             fill_pairs[key] = value
 
-    return template.format(**prepare_template_data(fill_pairs))
+    try:
+        return template.format(**prepare_template_data(fill_pairs))
+    except KeyError as exp:
+        error = "Value for '{}' key is missing in template '{}'.".format(
```
BigRoy, line 164:
```suggestion
        error = "Value for {} key is missing in template '{}'.".format(
```
Looking at your screenshot the reported value has double single quotes. Likely due to the KeyError's message value:
```python
a = {}
a["foo"]
# Error: 'foo'
# #   File "<maya console>", line 2, in <module>
# # KeyError: 'foo'
```
The string already includes single quotes.

openpype/pipeline/create/subset_name.py
```diff
@@ -154,4 +158,10 @@ def get_subset_name(
...
+            str(exp), template)
+        error += "Collected values are {}".format(fill_pairs)
```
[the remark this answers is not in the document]
BigRoy: Also note that string concatenation is relatively slow. You might be better of writing this as:
```
    raise KeyError(
        "Value for {} key is missing in template '{}'."
        " Collected values are {}".format(str(exp), template, fill_pairs)
    )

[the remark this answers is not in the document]
Comment (kalisp) [reply openpype/pipeline/create/subset_name.py]: Ok, fixed accordingly.

File: openpype/pipeline/create/subset_name.py [hunk 3]
```diff
@@ -154,4 +158,10 @@ def get_subset_name(
         for key, value in dynamic_data.items():
             fill_pairs[key] = value
 
-    return template.format(**prepare_template_data(fill_pairs))
+    try:
+        return template.format(**prepare_template_data(fill_pairs))
+    except KeyError as exp:
+        raise KeyError(
```
iLLiCiTiT, line 164: I still think we should not raise `KeyError`, but custom exception.

kalisp: Changed to new `MissingKeyError` exception

openpype/pipeline/create/subset_name.py
```diff
@@ -14,6 +14,13 @@ def __init__(self, msg=None):
         super(TaskNotSetError, self).__init__(msg)
 
 
+class MissingKeyError(KeyError):
```
iLLiCiTiT, line 17:
```suggestion
class TemplateFillError(Exception):
```

openpype/pipeline/create/subset_name.py
```diff
@@ -112,6 +119,10 @@ def get_subset_name(
             for project. Settings are queried if not passed.
         family_filter (Optional[str]): Use different family for subset template
             filtering. Value of 'family' is used when not passed.
+
+    Raises:
+        KeyError if filled template contains placeholder key which is not
```
iLLiCiTiT, line 124:
```suggestion
        TemplateFillError: If filled template contains placeholder key which is not
```

PR #5651, opened 2023-09-25
Bugfix: houdini switching context doesnt update variables

MustafaJafar:
## Changelog Description
Allows admins to have a list of vars (e.g. JOB) with (dynamic) values that will be updated on context changes, e.g. when switching to another asset or task.

Using template keys is supported but formatting keys capitalization variants is not, e.g. {Asset} and {ASSET} won't work

> [!NOTE]  
> If is Dir Path toggle is activated, Openpype will consider the given value as a path to a folder.
>
> If the folder does not exist on the context change it will be created by this feature so that the path will always try to point to an existing folder.

Disabling Update Houdini vars on context change feature will leave all Houdini vars unmanaged and thus no context update changes will occur.

> If $JOB is present in the Houdini var list and has an empty value, OpenPype will set its value to $HIP

image

image

Also, this PR adds a new button in menu to update vars on demand. 
image

## Testing notes:
1. try different job path combinations 
2. open houdini files and switch contexts


openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +752,32 @@ def get_camera_from_container(container):
 
     assert len(cameras) == 1, "Camera instance must have only one camera"
     return cameras[0]
+
+
+def validate_job_path():
+    """Validate job path to ensure it matches the settings."""
```
BigRoy, line 758: I'd say that `validate_` is the wrong function name for this since it's actually _changing_ the values. How about `update_context_for_job_var`? or `update_job_var_context`? or something along those lines?

openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +752,32 @@ def get_camera_from_container(container):
...
+            # and has empty value.
+            job_path = os.environ["HIP"]
+
+        current_job = hou.hscript("echo -n `$JOB`")[0]
+        if current_job != job_path:
+            hou.hscript("set JOB=" + job_path)
+            os.environ["JOB"] = job_path
+            print("  - set $JOB to " + job_path)
+    else:
+        print("  - JOB Path is disabled, Skipping Check...")
```
BigRoy, line 783: Shouldn't this be `logging` calls instead?

openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +752,32 @@ def get_camera_from_container(container):
 
     assert len(cameras) == 1, "Camera instance must have only one camera"
     return cameras[0]
+
+
+def validate_job_path():
```
BigRoy, line 757: We  might want to allow to pass settings document directly for optimization purposes if you have the settings `dict` available already?

openpype/hosts/houdini/api/lib.py
BigRoy, line 783: Pretty sure this is a very low level debug message at most - doesn't make sense to log this all the time I'd say. Remove the message when the feature is disabled?

openpype/settings/entities/schemas/projects_schema/schemas/schema_houdini_general.json
```diff
@@ -0,0 +1,28 @@
...
+    "is_group": true,
+    "children": [
+        {
+            "type": "dict",
+            "collapsible": true,
+            "checkbox_key": "enabled",
+            "key": "job_path",
+            "label": "JOB Path",
+            "children": [
+                {
```
BigRoy, line 15: Admittedly it's pretty unclear what the checkbox means really for `JOB Path` for an admin. Can imagine someone not fully grasping what enabling/disabling it would change. Should we add a more descriptive label inside the collapsible box?

openpype/hosts/houdini/api/lib.py
MustafaJafar: > if you have the settings dict available already?

I'm afraid it's not, but why not making it available!

openpype/hosts/houdini/api/lib.py
MustafaJafar: yes, but for some reason logging doesn't work, so I chose print. 
same problem found here 
https://github.com/ynput/OpenPype/blob/1bd07bd15b1bf238e186b23b7d112e9fe4637737/openpype/hosts/houdini/api/lib.py#L505

moonyuet: it is supposed that you can simply type `log.debug("Setting scene FPS to {}".format(int(fps)))` to do the logging calls as the logging system is implemented in https://github.com/ynput/OpenPype/blob/fdea715fe0d7aafd7f3000b8aca7780d432aeacb/openpype/hosts/houdini/api/lib.py#L26

MustafaJafar:
I made a little cool update in `openpype.hosts.houdini.api` 
there are two different ways to get project settings 

```python

# it's possible to access houdini settings any where inside HoudiniHost class
houdini_host = registered_host()
print(houdini_host.houdini_project_settings.general)

# by default it gets the current project settings
from openpype.hosts.houdini.api import HoudiniProjectSettings
print(HoudiniProjectSettings().general)

# you can pass a project_name argument to get project settings regards that project
from openpype.hosts.houdini.api import HoudiniProjectSettings
print(HoudiniProjectSettings(project_name="a project").general)
```

openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +751,27 @@ def get_camera_from_container(container):
...
+    """Validate job path to ensure it matches the settings."""
+
+    if job_var_settings["enabled"]:
+
+        # get and resolve job path template
+        job_path_template = job_var_settings["job_path"]
+        job_path = StringTemplate.format_template(
+            job_path_template, get_current_context_template_data()
+        )
+        job_path = job_path.replace("\\", "/")
```
moonyuet, line 766:
```suggestion
        job_path = job_path.replace("\\", "/")
        os.makedirs(job_path, exist_ok=True)
```
what if the `job_path `is successfully generated from anatomy data but it doesn't exist as path yet? For example, you have set JOB_PATH to {root[work]}/{project[name]}/{hierarchy}/{asset}/work/{task[name]}/{asset}

openpype/hosts/houdini/api/pipeline.py
```diff
@@ -48,6 +52,10 @@ def __init__(self):
         self._op_events = {}
         self._has_been_setup = False
 
+    @property
+    def houdini_project_settings(self):
+        return HoudiniProjectSettings()
```
MustafaJafar, line 57: @BigRoy 
Would it be still clear that it's `houdini_project_settings` if I removed the `houdini` prefix ? 
```suggestion
    # Current project houdini settings
    @property
    def project_settings(self):
        return HoudiniProjectSettings()
```

openpype/hosts/houdini/api/lib.py
MustafaJafar: wow good catch. I totally forgot that!

openpype/hosts/houdini/api/pipeline.py
```diff
@@ -450,3 +466,33 @@ def main_take(no_update=True):
             instance_node.bypass(not new_value)
     except hou.PermissionError as exc:
         log.warning("%s - %s", instance_node.path(), exc)
+
+
+class HoudiniProjectSettings:
+    def __init__(self, project_name=""):
```
moonyuet, line 472:
```suggestion
    def __init__(self, project_name=get_current_project_name()):
```
Is it going to be used in somewhere else? If not, implement` from openpype.pipeline import get_current_project_name` and use `get_current_project_name()` . Otherwises the system would give the KeyError below
```
Installing OpenPype ...
Setting scene FPS to 25
>>> [ *** Discovered 1 abstract plugins
- AutoCreator ] 
  - { CreateArnoldRop }: [  No settings found for CreateArnoldRop  ] 
  - { CreateBGEO }: [  No settings found for CreateBGEO  ] 
  - { CreateHDA }: [  No settings found for CreateHDA  ] 
  - { CreateKarmaROP }: [  No settings found for CreateKarmaROP  ] 
  - { CreateMantraROP }: [  No settings found for CreateMantraROP  ] 
  - { CreateRedshiftProxy }: [  No settings found for CreateRedshiftProxy  ] 
  - { CreateReview }: [  No settings found for CreateReview  ] 
  - { CreateVrayROP }: [  No settings found for CreateVrayROP  ] 
Using existing QApplication..
Registered font families: Noto Sans, Noto Sans, Noto Sans, Noto Sans, Noto Sans Mono, Noto Sans Mono, Noto Sans Mono, Noto Sans Mono, Noto Sans Mono, Noto Sans Mono, Noto Sans Mono, Noto Sans Mono, Noto Sans Mono
Failed to execute event callback < EventCallback - on_open > D:\kayla\OpenPype\openpype\hosts\houdini\api\pipeline.py
  File "D:\kayla\OpenPype\openpype\lib\events.py", line 163, in process_event
    callback()
  File "D:\kayla\OpenPype\openpype\hosts\houdini\api\pipeline.py", line 352, in on_open
    HoudiniProjectSettings().general["update_job_var_context"])
KeyError: 'update_job_var_context'
  File "D:\kayla\OpenPype\openpype\tools\workfiles\files_widget.py", line 610, in _on_workarea_open_pressed
    self.open_file(path)
  File "D:\kayla\OpenPype\openpype\tools\workfiles\files_widget.py", line 517, in open_file
    host.open_workfile(filepath)
  File "D:\kayla\OpenPype\openpype\hosts\houdini\api\pipeline.py", line 123, in open_workfile
    hou.hipFile.load(filepath,
  File "C:\PROGRA~1/SIDEEF~1/HOUDIN~1.435/houdini/python3.9libs\houpythonportion\ui.py", line 1119, in decorator
    return func(*args, **kwargs)
  File "C:\PROGRA~1/SIDEEF~1/HOUDIN~1.435/houdini/python3.9libs\hou.py", line 52104, in load
    return _hou.hipFile_load(self, file_name, suppress_save_prompt, ignore_load_warnings)
hou.LoadWarning: Warnings were generated during load.
Error loading: D:/test_project/new_project/ep/work/asset/new_project_ep_asset_v232.hiplc
Warning:     Problem while synchronizing child node:
             Warning:     Skipping unrecognized parameter "sepparm11".
             Skipping unrecognized parameter "tonemap".
             Skipping unrecognized parameter "tonemap_aovs".
             Skipping unrecognized parameter "tonemap_toe".
             Skipping unrecognized parameter "tonemap_shoulder".
             Skipping unrecognized parameter "tonemap_utoe".
             Skipping unrecognized parameter "tonemap_ushoulder".
             Skipping unrecognized parameter "tonemap_slope".
             Skipping unrecognized parameter "tonemap_linear".
             Skipping unrecognized parameter "tonemap_linearAngle".
             Skipping unrecognized parameter "tonemap_toeLength".
             Skipping unrecognized parameter "tonemap_shoulderLength".
             Skipping unrecognized parameter "tonemap_shoulderAngle".
             Skipping unrecognized parameter "sepparm13".
	[/out/karma_ropMain/lopnet/karma]

Setting scene FPS to 25
Using existing QApplication..
Failed to execute event callback < EventCallback - on_open > D:\kayla\OpenPype\openpype\hosts\houdini\api\pipeline.py
  File "D:\kayla\OpenPype\openpype\lib\events.py", line 163, in process_event
    callback()
  File "D:\kayla\OpenPype\openpype\hosts\houdini\api\pipeline.py", line 352, in on_open
    HoudiniProjectSettings().general["update_job_var_context"])
KeyError: 'update_job_var_context'
  File "D:\kayla\OpenPype\openpype\tools\workfiles\files_widget.py", line 610, in _on_workarea_open_pressed
    self.open_file(path)
  File "D:\kayla\OpenPype\openpype\tools\workfiles\files_widget.py", line 517, in open_file
    host.open_workfile(filepath)
  File "D:\kayla\OpenPype\openpype\hosts\houdini\api\pipeline.py", line 123, in open_workfile
    hou.hipFile.load(filepath,
  File "C:\PROGRA~1/SIDEEF~1/HOUDIN~1.435/houdini/python3.9libs\houpythonportion\ui.py", line 1119, in decorator
    return func(*args, **kwargs)
  File "C:\PROGRA~1/SIDEEF~1/HOUDIN~1.435/houdini/python3.9libs\hou.py", line 52104, in load
    return _hou.hipFile_load(self, file_name, suppress_save_prompt, ignore_load_warnings)
hou.LoadWarning: Warnings were generated during load.
Error loading: D:/test_project/new_project/ep/work/asset/new_project_ep_asset_v232.hiplc
Warning:     Problem while synchronizing child node:
             Warning:     Skipping unrecognized parameter "sepparm11".
             Skipping unrecognized parameter "tonemap".
             Skipping unrecognized parameter "tonemap_aovs".
             Skipping unrecognized parameter "tonemap_toe".
             Skipping unrecognized parameter "tonemap_shoulder".
             Skipping unrecognized parameter "tonemap_utoe".
             Skipping unrecognized parameter "tonemap_ushoulder".
             Skipping unrecognized parameter "tonemap_slope".
             Skipping unrecognized parameter "tonemap_linear".
             Skipping unrecognized parameter "tonemap_linearAngle".
             Skipping unrecognized parameter "tonemap_toeLength".
             Skipping unrecognized parameter "tonemap_shoulderLength".
             Skipping unrecognized parameter "tonemap_shoulderAngle".
             Skipping unrecognized parameter "sepparm13".
	[/out/karma_ropMain/lopnet/karma]

Using existing QApplication..
Failed to execute event callback < EventCallback - on_open > D:\kayla\OpenPype\openpype\hosts\houdini\api\pipeline.py
  File "D:\kayla\OpenPype\openpype\lib\events.py", line 163, in process_event
    callback()
  File "D:\kayla\OpenPype\openpype\hosts\houdini\api\pipeline.py", line 352, in on_open
    HoudiniProjectSettings().general["update_job_var_context"])
KeyError: 'update_job_var_context'
```

openpype/hosts/houdini/api/pipeline.py
BigRoy: No it wouldn't but I would also want to say, don't do this. It doesn't cache it .. it being a property makes it feel like would be cached and doesn't touch the database each time. This doesn't solve anything of it potentially being slow but just made it more complex code without any benefit.

If we were to cache the local project settings, then I think that could be nice if e.g. available on a dedicated host mixing instead of implementing it only in the Houdini side. Plus that'd be a separate PR.

I'd revert this change.

openpype/hosts/houdini/api/lib.py
BigRoy: Actually, do we really care if the folder does not exist? Nothing breaks if it doesn't exist. Right?

I don't feel this is the time and place to create the folder. If the folder is needed to exist for an asset they should be dedicated work folders and not depend on the job var logic.

We are just now touching the file system more for no reason.

Or does Houdini error if we set a JOB path that does not exist?

BigRoy: Also, IF we absolutely need this we should move this into the if path doesn't equal current path logic so it only ever triggers when changing the value. But, again, I'd argue it shouldn't create the folder.

moonyuet: > Actually, do we really care if the folder does not exist? Nothing breaks if it doesn't exist. Right?
> 
> I don't feel this is the time and place to create the folder. If the folder is needed to exist for an asset they should be dedicated work folders and not depend on the job var logic.
> 
> We are just now touching the file system more for no reason.
> 
> Or does Houdini error if we set a JOB path that does not exist?

The $JOB would turns to be in red in the dialog.
image
Dont think we want this happened.

BigRoy: Hmm. I see. In that case we might want to only do it if the path changed to ensure less disk access.

Also, we could add a settings toggle "create folder if not exists" or alike. I'd still argue that creating the folder should not be up to this functionality but I can see how others might think it should.

moonyuet: > we could add a settings toggle "create folder if not exists" or alike. 

Good idea. If it isn't enabled, we can just use $HIP to be $JOB (like when Job Path setting is empty) and inform the user that the Job Path  they set doesn't exist, and the system would use $HIP path instead.
And if it is enabled, just inform the user using the job path :{the_path} for $JOB.

BigRoy: > and inform the user that the Job Path they set doesn't exist, and the system would use $HIP path instead.

I feel like then we don't need to inform the user at all. If this feature is disabled we should just do nothing.
Be aware that ANY console messages, even if "X is skipped" will have users start asking questions about, why is this skipped? Should I be worried? Did I break something? etc.

MustafaJafar: > we could add a settings toggle "create folder if not exists" or alike.

So cool! 

> If this feature is disabled we should just do nothing.

I'll vote for this and it's the current implementation

openpype/hosts/houdini/api/pipeline.py
MustafaJafar: I reverted this change as BigRoy suggested here https://github.com/ynput/OpenPype/pull/5651#discussion_r1336874211

openpype/hosts/houdini/api/lib.py
MustafaJafar: I start to give up the idea of `os.makedirs(job_path, exist_ok=True)` 

Both of you are correct 
This code relates to houdini and it shouldn't touch file system 
also, users may have not enough privileges to create folders

however, some admins may assume that Ayon/openpype will create JOB folder for them  if not exits

how do you think we can solve this ?

moonyuet: Maybe we can just use $HIP to be $JOB (like when Job Path setting is empty) if the setting  "create folder if not exists"  isn't enabled,

moonyuet:
i have tested the code again. 
If "$JOB path update on context change " is enabled and the JOB env does not exist, the dialog would be in red .

image

But if the toggle is disable, the $JOB env is the same with the $HIP env.

image

moonyuet (approved):

BigRoy:
Should we add admin docs for this feature?

moonyuet:
> Should we add admin docs for this feature?

Yes, we should. At least let the user know $JOB would be not showed or errored out if they enabled the toggled button with the non-existent path.

openpype/hosts/houdini/api/lib.py
MustafaJafar: let's just use `print` and we can change it when log works

openpype/hosts/houdini/api/lib.py
MustafaJafar: I discussed this with iLLiCiTiT and `os.makedirs(job_path, exist_ok=True)` is just fine.

openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +752,41 @@ def get_camera_from_container(container):
...
+        # sync both environment variables.
+        # because when opening new file $JOB is overridden with
+        # the value saved in the HIP file but os.environ["JOB"] is not!
+        os.environ["JOB"] = current_job
+
+        if current_job != job_path:
+            hou.hscript("set JOB=" + job_path)
+            os.environ["JOB"] = job_path
+
+            os.makedirs(job_path, exist_ok=True)
```
BigRoy, line 789: Do we still support Houdini versions with Py2? If that's the case `exist_ok` argument won't exist in Py2.

MustafaJafar: I think Houdini versions with Py2 still work, and I think there are no yet critical code that has to be py3 specific.
but I didn't try testing with a houdini version with py2. 
so, is this better: 
```python
if not os.path.exists(job_path):
    os.makedirs(job_path)
```

MustafaJafar:
> Should we add admin docs for this feature?

I will update Openpype docs as it can be done in this PR, AYON requires a different PR to another repo

openpype/hosts/houdini/api/lib.py
BigRoy:
```python
if not os.path.exists(job_path):
    os.makedirs(job_path)
```

This would work in Py2 yes. Whether it's better? Purists would say it's worse because it introduces a race condition where just after the `exists` call the folder might have been created already (or removed). See devs having fun discussing the race condition [here](https://stackoverflow.com/a/273227).

That's where they recommend doing e.g.:
```python
import os, errno

try:
    os.makedirs(directory)
except OSError as e:
    if e.errno != errno.EEXIST:
        raise
```

openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +752,41 @@ def get_camera_from_container(container):
 
     assert len(cameras) == 1, "Camera instance must have only one camera"
     return cameras[0]
+
+
+def update_job_var_context():
+    """Validate job path to ensure it matches the settings."""
+
+    project_settings = get_current_project_settings()
+    project_settings = \
```
Minkiu, line 762: This is no longer `project_settings` so maybe store it as `update_job_var_context_settings` ? Something on the lines of that

website/docs/admin_hosts_houdini.md
```diff
@@ -3,9 +3,26 @@ id: admin_hosts_houdini
 title: Houdini
 sidebar_label: Houdini
 ---
+## General Settings
+### JOB Path
+you can add your studios preffered JOB Path, JOB value will be checked and updated on file save and open.
+Disableing this option will effectivly turn off this feature.
```
BigRoy, line 9:
```suggestion
The Houdini `$JOB` path can be customized through project settings with a dynamic path that will be updated on context changes, e.g. when switching to another asset or task.

Disabling this feature will leave `$JOB` var unmanaged and thus no context update changes will occur.
```

Might it be good to also link to the Houdini documentation as to what the JOB variable is intended to be for Houdini itself?

openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +752,41 @@ def get_camera_from_container(container):
...
+        project_settings["houdini"]["general"]["update_job_var_context"]
+
+    if project_settings["enabled"]:
+
+        # get and resolve job path template
+        job_path_template = project_settings["job_path"]
+        job_path = StringTemplate.format_template(
+            job_path_template, get_current_context_template_data()
```
Minkiu, line 770: We don't seem to be using `job_path_template` anywhere else so maybe it could be:
```py
job_path = StringTemplate.format_template(
	project_settings["job_path"],
	get_current_context_template_data()
)
```

website/docs/admin_hosts_houdini.md
```diff
@@ -3,9 +3,26 @@ id: admin_hosts_houdini
...
+
+JOB Path can be:
+- Arbitrary hardcoded path
+- Openpype template path
+    > This allows dynamic values for assets or shots.<br>
+    > Using template keys is supported but formatting keys capitalization variants is not,
+    >   e.g. {Asset} and {ASSET} won't work
+- empty
```
BigRoy, line 17:
```suggestion
- Empty
```

website/docs/admin_hosts_houdini.md
```diff
@@ -3,9 +3,26 @@ id: admin_hosts_houdini
...
+    > In this case, JOB will be synced to HIP
+
+![update job on context change](assets/houdini/update-job-context-change.png)
+
+
```
BigRoy, line 22: Should we add a note that it creates the folder as well?

```suggestion
If the folder does not create on the context change it will be created by this feature so that `$JOB` will always try to point to an existing folder.
```

```

File: website/docs/admin_hosts_houdini.md [hunk 14]
```diff
@@ -3,9 +3,26 @@ id: admin_hosts_houdini
 title: Houdini
 sidebar_label: Houdini
 ---
+## General Settings
+### JOB Path
+you can add your studios preffered JOB Path, JOB value will be checked and updated on file save and open.
```
Minkiu, line 8: I'd probably phrase more ont he lines of:
```md
Specify a studio-wide `JOB` path ...
``` 

There's a typos:
 `Disableing` -> `Disabling`; 
`effectivly` -> `effectively`

website/docs/admin_hosts_houdini.md
```diff
@@ -3,9 +3,26 @@ id: admin_hosts_houdini
...
+Disableing this option will effectivly turn off this feature.
+
+JOB Path can be:
+- Arbitrary hardcoded path
```
Minkiu, line 12: I think `hardcoded` is not necessary here, since it's implied we are setting it to a path

Minkiu (approved):
Looking good code wise.

website/docs/admin_hosts_houdini.md
```diff
@@ -5,16 +5,18 @@ sidebar_label: Houdini
 ---
 ## General Settings
 ### JOB Path
-you can add your studios preffered JOB Path, JOB value will be checked and updated on file save and open.
-Disableing this option will effectivly turn off this feature.
+Specify a studio-wide `JOB` path.<br>
```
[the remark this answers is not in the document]
BigRoy: A bit of semantics - but it's not necessarily studio-wide. It's a project setting, so project specific. Plus it can be dynamic per asset or task - so whether 'studio-wide' sounds correct for that I'm not sure. It kind of sounds like a studio specific static path. But maybe it's just my interpretation of the sentence.

website/docs/admin_hosts_houdini.md
MustafaJafar: okay, no more spell typos (I hope), I'm going to install vs code spell checker.

website/docs/admin_hosts_houdini.md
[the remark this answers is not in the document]
MustafaJafar: `<br>` doesn't generate an empty line, if it's not a big deal no problem. 
image

without `<br>`
image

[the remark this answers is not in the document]
BigRoy: It does add the empty line if you add an extra enter like in my suggestion, right?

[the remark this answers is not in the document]
BigRoy: I see - I personally prefer the second style anyway.

openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +753,48 @@ def get_camera_from_container(container):
 
     assert len(cameras) == 1, "Camera instance must have only one camera"
     return cameras[0]
+
+
+def update_job_var_context():
+    """Validate job path to ensure it matches the settings."""
```
BigRoy, line 759:
```suggestion
    """Update $JOB to match current context.

    This will only do something if the setting is enabled in project settings.
    """
```

openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +753,48 @@ def get_camera_from_container(container):
...
+            os.environ["JOB"] = job_path
+
+            try:
+                os.makedirs(job_path)
+            except OSError as e:
+                if e.errno != errno.EEXIST:
+                    print(
+                        "  - Failed to create JOB dir. Maybe due to "
+                        "insufficient permissions."
+                    )
```
BigRoy, line 797:
```suggestion
                    raise
```
If the dirs couldn't be created that's the perfect time to re-raise the error?

openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +753,48 @@ def get_camera_from_container(container):
...
+
+            print("  - Context changed, update $JOB respectively to "
```
BigRoy, line 799:
```suggestion
            print("Updated $JOB to {}".format(job_path))
```
Anyone is free to call this function - thus context techhnically doesn't have to be changed for it to occur.

website/docs/admin_hosts_houdini.md
```diff
@@ -3,9 +3,30 @@ id: admin_hosts_houdini
 title: Houdini
 sidebar_label: Houdini
 ---
+## General Settings
+### JOB Path
+
+The Houdini `$JOB` path can be customized through project settings with a (dynamic) path that will be updated on context changes, e.g. when switching to another asset or task.
+
+> If the folder does not exist on the context change it will be created by this feature so that `$JOB` will always try to point to an existing folder.
```
BigRoy, line 11: Would this be better as as note in the docs?
```suggestion
:::note
If the folder does not exist on the context change it will be created by this feature so that `$JOB` will always try to point to an existing folder.
:::
```

openpype/hosts/houdini/api/lib.py
moonyuet: i am actually wondering if it is possible to use logging here.

MustafaJafar: > i am actually wondering if it is possible to use logging here. 

I tried using `log.debug` and `log.info` but `log` doesn't seem to log anything

openpype/hosts/houdini/api/lib.py
MustafaJafar: Recalling this, will `raise` be helpful ? because Houdini is fine when setting JOB to a path that doesn't exists (I mean Houdini API won't raise an error when doing so)

> If "$JOB path update on context change " is enabled and the JOB env does not exist, the dialog would be in red .
> 
> image

website/docs/admin_hosts_houdini.md
MustafaJafar: That's even better! 
this is not markdown right ? it only works on the website ? 

--- 

I just found that this is supported in markdown  and it's very cool! 
[[Markdown] An option to highlight a "Note" and "Warning" using blockquote (Beta)](https://github.com/orgs/community/discussions/16925)

BigRoy (commented):
I was on the bike home and thought about this feature some more. I feel like we could actually tweak this feature slightly and allow much more studio customization through it.

The current feature is focused on the `$JOB` var only - but what if the studio also wants to maintain other vars per context, e.g. `$TASK` or `$SEQUENCE_LIB` or whatever.

Could we potentially just make this a dict in settings so that one could specify multiple entries, with the default just being:
```python
{"JOB": "{root[work]}/{project[name]}/{hierarchy}/{asset}/work/{task[name]}"}
```

It'd make the feature just a generalized "update (houdini) vars on context change" feature. Would that make sense?
The documentation could then still name `$JOB` updating as a concrete example but the use case would then be much wider than just maintaining `$JOB`.

I've heard of @krishna8008 requesting quite a similar feature where he could have vars like `$ASSETSTART`, `$ASSETEND` or render resolutions `$ASSET_RESOLUTIONX`,  `$ASSET_RESOLUTIONY` or whatever available to always have the rop nodes set to the current asset's frame start and frame end or rendering set to the render resolution based on the database. Could 90% be based off of this feature too in a way. If these would update the same way that could automatically get supported. We'll just somehow need to support formatting those vars too?

openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +753,60 @@ def get_camera_from_container(container):
...
+                try:
+                    os.makedirs(path)
+                except OSError as e:
+                    if e.errno != errno.EEXIST:
+                        print(
+                            "  - Failed to create {} dir. Maybe due to "
+                            "insufficient permissions.".format(var)
+                        )
+
+                print("  - Updated ${} to {}".format(var, path))
```
BigRoy, line 812: Since as per my examples provided when explaining why we could benefit from expanding this functionality for user requests to set vars for asset resolution, frame ranges, etc. the vars set through this function now aren't always actual paths on disk - but could be other arbitrary data, like a frame start value. So likely we need an extra toggle per var in settings that says "Treat as path and create folder when it does not exist" so that you can enable it for vars that are supposed to be absolute paths.

And likely we'd need to refactor the function to something like this:
```suggestion


def update_houdini_vars_context():
    """Update Houdini vars to match current context based.

    This will only do something if the setting is enabled in project settings.

    """

    project_settings = get_current_project_settings()
    houdini_vars_settings = \
        project_settings["houdini"]["general"]["update_houdini_var_context"]

    if not houdini_vars_settings["enabled"]:
        return

    houdini_vars = houdini_vars_settings["houdini_vars"]

    # Remap AYON settings structure to OpenPype settings structure
    # It allows me to use the same logic for both AYON and OpenPype
    if isinstance(houdini_vars, list):
        items = {}
        for item in houdini_vars:
            items.update({item["var"]: item["path"]})

        houdini_vars = items

    if not houdini_vars:
        # No vars specified - nothing to do
        return

    # For consistency reasons we always force all vars to be uppercase
    houdini_vars = {key.upper(): value for key, value in houdini_vars.items()}

    # TODO: I've now NOT used the separate function you made for this since
    #   now we're accessing the asset document itself as well.
    # Retrieve any context data for formatting
    # pre-prepare get_template_data args
    current_context = get_current_context()
    project_name = current_context["project_name"]
    asset_name = current_context["asset_name"]
    anatomy = Anatomy(project_name)

    # prepare get_template_data args
    project_doc = get_project(project_name)
    asset_doc = get_asset_by_name(project_name, asset_name)
    task_name = current_context["task_name"]
    host_name = get_current_host_name()

    # get template data
    template_data = get_template_data(
        project_doc, asset_doc, task_name, host_name
    )
    template_data["root"] = anatomy.roots

    # TODO: We add to documentation asset data examples that an admin might
    #    want to use.
    # To allow setting vars to context specific frame ranges, resolution
    # we need to provide more than the regular template data
    # By doing this you can access data like:
    # - {assetData[frameStart]}
    # - {assetData[frameEnd]}
    # - {assetData[handleStart]}
    # - {assetData[handleEnd]}
    # - {assetData[frameStartHandle]}
    # - {assetData[frameEndHandle]}
    # - {assetData[resolutionHeight]}
    # - {assetData[resolutionWidth]}
    # This would also expose e.g. (if ftrack module is enabled/synced):
    # - {assetData[ftrackId]}
    asset_data = asset_doc["data"].copy()
    template_data["asset_data"] = asset_data
    # The `frameStartHandle` and `frameEndHandle` doesn't exist in the
    # document's data thus we always need to compute it so it can be used
    # as formattable data
    if "frameStart" in asset_data and "handleStart" in asset_data:
        asset_data["frameStartHandle"] = (
                asset_data["frameStart"] - asset_data["frameStartHandle"]
        )
    if "frameEnd" in asset_data and "handleEnd" in asset_data:
        asset_data["frameEndHandle"] = (
                asset_data["frameEnd"] + asset_data["handleEnd"]
        )
    template_data["assetData"] = asset_data

    for var, template_value in houdini_vars.items():
        # get template and format with template data
        value = StringTemplate.format_template(template_value, template_data)

        if var == "JOB" and path == "":
            # sync $JOB to $HIP if $JOB is empty
            path = os.environ["HIP"]

        # TODO: Allow each entry in settings to specify whether it should
        #       be treated as a path (and with that also force the folder
        #       creation.
        if var_is_path:
            value = value.replace("\\", "/")

            # Force create folder when treating it as a path
            try:
                os.makedirs(value)
            except OSError as e:
                if e.errno != errno.EEXIST:
                    raise

        # sync both environment variables and houdini vars
        # because houdini doesn't do sync os value when opening new files
        # we force it to the current value if current value already matches
        # the new value on the hscript var
        current_value = hou.hscript("echo -n `${}`".format(var))[0]
        os.environ[var] = current_value

        if current_value != value:
            hou.hscript("set {}={}".format(var, value))
            os.environ[var] = value

            print("  - Updated ${} to {}".format(var, value))
```

Note the **TODO** blocks that I added which might still be relevant for you. also, I haven't used this before so please do check that if we expose this whether e.g. setting var `$ASSET_FRAMESTART` to `{assetData[frameStart]}` actually makes it so that you can use it as an int where we want to? If not, we might even need to set the correct type, etc. Anyway, have fun puzzling! :)

openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +753,64 @@ def get_camera_from_container(container):
...
+            # sync both environment variables.
+            # because houdini doesn't do that by default
+            # on opening new files
+            os.environ[item["var"]] = current_value
+
+            if current_value != item_value:
+                hou.hscript("set {}={}".format(item["var"], item_value))
+                os.environ[item["var"]] = item_value
+
+                print("  - Updated ${} to {}".format(item["var"], item_value))
```
BigRoy:
```suggestion
    if not houdini_vars_settings["enabled"]:
        return
    
    houdini_vars = houdini_vars_settings["houdini_vars"]

    # No vars specified - nothing to do
    if not houdini_vars:
        return

    # Get Template data
    template_data = get_current_context_template_data()

    # Set Houdini Vars
    for item in houdini_vars:

        # For consistency reasons we always force all vars to be uppercase
        item["var"] = item["var"].upper()

        # get and resolve job path template
        item_value = StringTemplate.format_template(
            item["value"],
            template_data
        )

        if item["is_path"]:
            item_value = item_value.replace("\\", "/")
            try:
                os.makedirs(item_value)
            except OSError as e:
                if e.errno != errno.EEXIST:
                    print(
                        "  - Failed to create ${} dir. Maybe due to "
                        "insufficient permissions.".format(item["var"])
                    )

        if item["var"] == "JOB" and item_value == "":
            # sync $JOB to $HIP if $JOB is empty
            item_value = os.environ["HIP"]

        current_value = hou.hscript("echo -n `${}`".format(item["var"]))[0]

        # sync both environment variables.
        # because houdini doesn't do that by default
        # on opening new files
        os.environ[item["var"]] = current_value

        if current_value != item_value:
            hou.hscript("set {}={}".format(item["var"], item_value))
            os.environ[item["var"]] = item_value

            print("  - Updated ${} to {}".format(item["var"], item_value))
```

BigRoy: Seems like this is now outdated - anyway the change would be:

Avoid the nested if branch
```python
if not enabled:
    return
```

Change comment:
```
# get and resolve job path template
```
to:
```
# get and resolve template in value
```

server_addon/houdini/server/settings/general.py
```diff
@@ -0,0 +1,39 @@
+from pydantic import Field
+from ayon_server.settings import BaseSettingsModel
+
+
+class HoudiniVarModel(BaseSettingsModel):
+    _layout = "expanded"
+    var: str = Field("", title="Var")
+    value: str = Field("", title="Value")
+    is_dir_path: bool = Field(False, title="is Dir Path")
```
BigRoy, line 9: Title seems a bit off with the `Dir Path` capitalized but `is` being lowercase? Should this be `Is dir path` or if the title is the descriptive label probably "Value should be treated as a path" would be nicer for the setting?

Also, why do you feel `is_dir_path` is clearer than `is_path`? Technically one _could_ point it to a file right? `{asset}_thumbnail.png` or whatever.

MustafaJafar: >  Technically one could point it to a file right? {asset}_thumbnail.png or whatever.

yes, That's why I though `is_dir_path` is clearer 

> Title seems a bit off with the Dir Path capitalized but is being lowercase? 

I was going to write it as `isDirPath`

BigRoy: Just to rephrase what I said. The value can point to a file just fine - it isn't necessarily a directory - technically someone COULD point it to a `{asset}_thumbnail.png` **but** thinking about it now that'd definitely give odd behavior if it were to CREATE that as a  folder if it didn't exist ha so I see what you're getting at. Thanks.

Anyway, I still think "Treat value as a director (create if it does not exist)" is the most clear. "Is folder" or "Is directory" doesn't tell me enough about what the checkbox would be doing. Right?

MustafaJafar: "Treat value as a directory, create directory if it does not exist" is very long! 
is this clear? I wish if we could make it shorter, 
image

or `Is value a directory` 
image

BigRoy: **Treat as directory** is short and gets the point across. Would that work for you?


I just wish there was a label, tooltip or description the user could rely on. Maybe at the top of the section there could be a label that says: "If a value is treated as a directory on update it will be ensured the folder exists" or alike.

BigRoy:
@krishna8008 Is there any chance you would be able to test this feature whether it allows to do what you wanted?

server_addon/houdini/server/settings/general.py
MustafaJafar: > Treat as directory 

That's fair enough 

> I just wish there was a label, tooltip or description the user could rely on.

I think this is supported in Ayon, but I'm not sure how to use it (yet). 

in openpype we can just insert a label as tip.
image

BigRoy: I personally quite like this!

MustafaJafar: omg! you just add a doc string!  and click `?` button to show it! 

image

image

BigRoy: But how do you set a tooltip for a single field?
Still, this is already nice for this particular use case.

MustafaJafar:
This is a reminder for me to include these (resolution/pixel aspect )
and I can ~steal~ get inspiration from here: https://github.com/ynput/OpenPype/blob/a66edaf1d051be4c6be2cfe189fe7a5912296968/openpype/hosts/nuke/api/lib.py#L2554

BigRoy:
> This is a reminder for me to include these (resolution/pixel aspect ) and I can ~steal~ get inspiration from here:
> 
> https://github.com/ynput/OpenPype/blob/a66edaf1d051be4c6be2cfe189fe7a5912296968/openpype/hosts/nuke/api/lib.py#L2554
As far as I know these are in asset data and thus already present now in the `assetData` key.

MustafaJafar:
> As far as I know these are in asset data and thus already present now in the assetData key.

yes, They are! 
I can found these keys in`assetData` in my setup 
```
['zou', 'tasks', 'frameStart', 'frameEnd', 'handleStart', 'handleEnd', 'frameStartHandle', 'frameEndHandle',
'fps', 'resolutionWidth', 'resolutionHeight', 'pixelAspect',  'clipIn', 'clipOut',
'visualParent', 'parents', 'tools_env']
```

@fabiaserra , Could you tell if there are missing keys ?

BigRoy:
> I think env vars should be all in capitals and ideally some short code name. For resolutionWidth and resolutionHeight we used to use `RESX` and `RESY`, and for pixel aspect I guess it could be something like `PIX_AR` or something along those lines

I think you're misinterpreting. These are not environment variables. How those are called are totally up to you via project settings.

Basically you would configure through the settings:
```
RESX = {assetData[resolutionWidt']}
RESY = {assetData[resolutionHeight]}
PIX_AR = {assetData[pixelAspect]}
```
The asset data is a `dict` that is just the plain asset's entity data you can then use in that templating system. It will now always sync the left hand side (your names for hscript vars and environment variables)

MustafaJafar:
> The asset data is a dict that is just the plain asset's entity data you can then use in that templating system. It will now always sync the left hand side (your names for hscript vars and environment variables)

There's a catch, values will have type of `str` 
Settings
image

For Houdini Vars I don't think it will not make a difference 
image

but for python os.environ vars (I sync houdini vars to python os.environ in code) 
I think we could leave it as it's for now and see if it will make a problem ? 
image

BigRoy:
> There's a catch, values will have type of str

FYI, Environment Variable values must always be `str` anyway so I'd this is to be the expected behavior.

BigRoy:
@fabiaserra Any chance you could contribute [that update vars pop-up logic](https://discord.com/channels/517362899170230292/563751989075378201/1158438058996289577) to this branch? I feel like the feature would align one-to-one.

I assume it would be something like this?

**pseudocode**

```python
# First we need to collect what vars will change and what there current values are.
# This will need to have a function available for that @MustafaJafar so we need to
# update the `update_houdini_vars_context()` function to use that to.
# E.g.

def get_context_var_changes():
    """Return asset context variables that need to be updated"""
    vars_to_update = {}
    vars_to_update["RESX"] = (old, new)
    return vars_to_update

def update_houdini_vars_context()
    """Update asset context variables"""
    vars_to_update = get_context_var_changes()
    for var, (old, new) in get_context_var_changes().items():
        # perform the update
        ...

def update_houdini_vars_context_dialog():
    """Show pop-up to update asset context variables"""
    update_vars = get_context_var_changes()
    if not update_vars:
        # Nothing to change
        return
    
    message = "\n".join(
        "${}: {} -> {}".format(var, old, new)
        for var, (old, new) in update_vars.items()
    )
    
    dialog = popup.PopupUpdateKeys(parent=parent)
    dialog.setModal(True)
    dialog.setWindowTitle("Houdini scene has outdated asset variables")
    dialog.setMessage(message)
    dialog.setButtonText("Fix")

    # on_show is the Fix button clicked callback
    dialog.on_clicked_state.connect(lambda: update_houdini_vars_context())

    dialog.show()
```

openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +753,66 @@ def get_camera_from_container(container):
...
+
+    # No vars specified - nothing to do
+    if not houdini_vars:
+        return
+
+    # Get Template data
+    template_data = get_current_context_template_data()
+
+    # Set Houdini Vars
+    for item in houdini_vars:
```
Minkiu, line 781: Is `houdini_vars` a dictionary? If so we could do:
`for var_name, var_value in houdini_vars.items():`

openpype/pipeline/context_tools.py
```diff
@@ -658,3 +661,70 @@ def get_process_id():
...
+
+    # get regular template data
+    template_data = get_template_data(
+        project_doc, asset_doc, task_name, host_name
+    )
+
+    template_data["root"] = anatomy.roots
+
+    # get context specific vars
+    asset_data = asset_doc["data"].copy()
```
Minkiu, line 716: By doing this copy I think you are effectively changing the original one, unsure if you are okay with that or not.

Minkiu (approved):
Haven't tested but looks good.

openpype/pipeline/context_tools.py
BigRoy: Sorry, what do you mean. By copying you're basically copying the top level `dict` (not the nested `mutable` entries) so adding a key to that `dict` does not alter the original `dict`.

Minkiu: True that, i got it wrong, ignor me.

openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +751,104 @@ def get_camera_from_container(container):
...
+    )
+    parent = hou.ui.mainQtWindow()
+    dialog = popup.PopupUpdateKeys(parent=parent)
+    dialog.setModal(True)
+    dialog.setWindowTitle("Houdini scene has outdated asset variables")
+    dialog.setMessage(message)
+    dialog.setButtonText("Fix")
+
+    # on_show is the Fix button clicked callback
+    dialog.on_clicked_state.connect(lambda: update_houdini_vars_context())
```
BigRoy, line 853: Sorry, this could've just been like this:
```suggestion
    dialog.on_clicked_state.connect(update_houdini_vars_context)
```
Should've written better pseudocode. :)

openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +751,104 @@ def get_camera_from_container(container):
...
+        os.environ[item["var"]] = current_value
+
+        if current_value != item_value:
+            houdini_vars_to_update.update(
+                {
+                    item["var"]: (
+                        current_value, item_value, item["is_directory"]
+                    )
+                }
+            )
```
BigRoy, line 809:
```suggestion
        # For consistency reasons we always force all vars to be uppercase
        var = item["var"].upper()

        # get and resolve template in value
        item_value = StringTemplate.format_template(
            item["value"],
            template_data
        )

        if var == "JOB" and item_value == "":
            # sync $JOB to $HIP if $JOB is empty
            item_value = os.environ["HIP"]

        if item["is_directory"]:
            item_value = item_value.replace("\\", "/")

        current_value = hou.hscript("echo -n `${}`".format(var))[0]

        # sync both environment variables.
        # because houdini doesn't do that by default
        # on opening new files
        os.environ[var] = current_value

        if current_value != item_value:
            houdini_vars_to_update[var] = (
                current_value, item_value, item["is_directory"]
            )
```

BigRoy: I did notice another issue aside of my above suggestion. We're changing `os.environ[var]` in a function that should only **get** the changes we might want to make. I know that's because in this case you'd also need to return those values that have non-changed values.

How about changing the `get_context_var_changes` to have an extra argument? For example:
```python
def get_context_var_changes(differences_only=True):
```

Then `update_houdini_vars_context` would call it: `get_context_var_changes(differences_only=False)` and would set the `os.environ[]` value even for non-changed values but continues to do the rest for changed values.

Thoughts?

openpype/hosts/houdini/api/lib.py
```diff
@@ -835,6 +835,7 @@ def update_houdini_vars_context_dialog():
     update_vars = get_context_var_changes()
     if not update_vars:
         # Nothing to change
+        print("  - Nothing to change, Houdini Vars are up to date.")
```
BigRoy, line 838:
```suggestion
        log.debug("Nothing to change, Houdini vars are already up to date.")
```
_Assumes there's a `log` in this `lib.py` file already_

What's up with all the indentations and prefixed `  - ` in prints? :D Why is that needed? I see that a lot and never understood why.

BigRoy: You're a mad man. Use `logging`.

MustafaJafar: I can't see any printed messages whenever I use `log` in lib file :/ 
I can tell log is defined but it doesn't work

https://github.com/ynput/OpenPype/blob/e255c20c440211d3578fc7bcc7b350b6756dd859/openpype/hosts/houdini/api/lib.py#L21

openpype/hosts/houdini/api/lib.py
MustafaJafar: I have two ideas 
1. we can just move it and update function will eventually update and sync them with the new value 
2. sync old values before the popup

MustafaJafar: I will just remove it.

openpype/hosts/houdini/api/lib.py
MustafaJafar: Let's stick to `print` (without these annoying indentations) till we know why logging not working in this scope.

openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +751,96 @@ def get_camera_from_container(container):
...
+def update_houdini_vars_context_dialog():
+    """Show pop-up to update asset context variables"""
+    update_vars = get_context_var_changes()
+    if not update_vars:
+        # Nothing to change
+        print("Nothing to change, Houdini vars are already up to date.")
+        return
+
+    message = "\n".join(
+        "${}: {} -> {}".format(var, old or "None", new)
```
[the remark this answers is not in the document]
MustafaJafar: I checked it many times, and it's an empty string

BigRoy:
Can you share a screenshot of the current dialog you get when context changes?

openpype/hosts/houdini/api/lib.py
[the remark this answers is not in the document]
BigRoy: That confirms my suspicion, thanks.

MustafaJafar:
> Can you share a screenshot of the current dialog you get when context changes?

image

fabiaserra:
> @fabiaserra Any chance you could contribute [that update vars pop-up logic](https://discord.com/channels/517362899170230292/563751989075378201/1158438058996289577) to this branch? I feel like the feature would align one-to-one.
> 
> First we need to collect what vars will change and what their current values are. This will need to have a function available for that @MustafaJafar so we need to update the `update_houdini_vars_context()` function to use that too.
> 
> I assume it would be something like this?
> 
> **pseudocode**
> 
> ```python
> def get_context_var_changes():
>     """Return asset context variables that need to be updated"""
>     vars_to_update = {}
>     vars_to_update["RESX"] = (old, new)
>     return vars_to_update
> 
> def update_houdini_vars_context():
>     """Update asset context variables"""
>     vars_to_update = get_context_var_changes()
>     for var, (old, new) in get_context_var_changes().items():
>         # perform the update
>         # ...
> 
> def update_houdini_vars_context_dialog():
>     """Show pop-up to update asset context variables"""
>     update_vars = get_context_var_changes()
>     if not update_vars:
>         # Nothing to change
>         return
>     
>     message = "\n".join(
>         "${}: {} -> {}".format(var, old, new)
>         for var, (old, new) in update_vars.items()
>     )
>     
>     dialog = popup.PopupUpdateKeys(parent=parent)
>     dialog.setModal(True)
>     dialog.setWindowTitle("Houdini scene has outdated asset variables")
>     dialog.setMessage(message)
>     dialog.setButtonText("Fix")
> 
>     # on_show is the Fix button clicked callback
>     dialog.on_clicked_state.connect(lambda: update_houdini_vars_context())
> 
>     dialog.show()
> ```
> 
> But maybe a slightly better dialog could also be nice to implement since the "Update keys []" in your example seems weird (due to re-using a different pop-up of course)

Hey yeah I'm sorry I've been hammered in my studio firefighting production needs and haven't had time to do what I said I would. Let me share here the quick and dirty implementation I did of this:

I added this on the `on_open` function of `houdini.api.pipeline`:
```python
def on_open():

    if not hou.isUIAvailable():
        log.debug("Batch mode detected, ignoring `on_open` callbacks..")
        return

    log.info("Running callback on open..")

    outdated_asset_variables = lib.get_outdated_asset_variables()

    if outdated_asset_variables:
        parent = lib.get_main_window()
        if parent is None:
            # When opening Houdini with last workfile on launch the UI hasn't
            # initialized yet completely when the `on_open` callback triggers.
            # We defer the dialog popup to wait for the UI to become available.
            # We assume it will open because `hou.isUIAvailable()` returns True
            import hdefereval
            hdefereval.executeDeferred(
                lib.show_outdated_asset_variables_popup,
                outdated_asset_variables
            )
        else:
            lib.show_outdated_asset_variables_popup(outdated_asset_variables)

        log.warning("Scene has outdated context variables.")
````

And these are the functions that Ia dded on the `houdini.lib`:

```python
def set_scene_resolution(width, height, pix_aspect):
    if width:
        hou.hscript(f"set -g RESX={width}")
        hou.hscript("varchange -V RESX")
    if height:
        hou.hscript(f"set -g RESY={height}")
        hou.hscript("varchange -V RESY")
    if pix_aspect:
        hou.hscript(f"set -g PIX_AR={pix_aspect}")
        hou.hscript("varchange -V PIX_AR")


def get_outdated_asset_variables():
    outdated_variables = {}

    # Validate FPS
    fps = get_asset_fps()
    current_fps = hou.fps()  # returns float

    if current_fps != fps:
        outdated_variables["FPS"] = (current_fps, fps)

    # Validate resolution
    width, height, pix_aspect = get_resolution_from_doc(asset_doc)
    cur_width = hou.getenv("RESX")
    cur_height = hou.getenv("RESY")
    cur_pix_aspect = hou.getenv("PIX_AR")

    if width != cur_width:
        outdated_variables["RESX"] = (cur_width, width)

    if height != cur_height:
        outdated_variables["RESY"] = (cur_height, height)

    if pix_aspect != cur_pix_aspect:
        outdated_variables["PIX_AR"] = (cur_pix_aspect, pix_aspect)

    return outdated_variables


def show_outdated_asset_variables_popup(outdated_variables):
    # Get main window
    parent = get_main_window()
    if parent is None:
        log.info("Skipping outdated content pop-up "
                 "because Houdini window can't be found.")
    else:
        from openpype.widgets import popup
        dialog = popup.PopupUpdateKeys(parent=parent)
        dialog.setModal(True)
        dialog.setWindowTitle("Houdini scene has outdated asset variables")
        outaded_vars_list = []
        for variable_name, outdated_value in outdated_variables.items():
            cur_value, new_value = outdated_value
            outaded_vars_list.append(
                f"${variable_name}: {cur_value} -> {new_value}"
            )

        dialog.setMessage("\n".join(outaded_vars_list))
        dialog.setButtonText("Fix")

        new_width = None
        outdated_width = outdated_variables.get("RESX")
        if outdated_width:
            new_width = outdated_width[1]

        new_height = None
        outdated_height = outdated_variables.get("RESY")
        if outdated_height:
            new_height = outdated_height[1]

        new_pix_aspect = None
        outdated_pix_aspect = outdated_variables.get("PIX_AR")
        if outdated_pix_aspect:
            new_pix_aspect = outdated_pix_aspect[1]

        new_fps = None
        outdated_fps = outdated_variables.get("FPS")
        if outdated_fps:
            new_fps = outdated_fps[1]

        # on_show is the Fix button clicked callback
        dialog.on_clicked_state.connect(
            lambda: (
                set_scene_resolution(new_width, new_height, new_pix_aspect),
                set_scene_fps(new_fps)
            )
        )

        dialog.show()
```

Notice that I changed the existing `get_resolution_from_doc` to also return pixel aspect:
```python
def get_resolution_from_doc(doc):
    """Get resolution from the given asset document. """

    if not doc or "data" not in doc:
        print("Entered document is not valid. \"{}\"".format(str(doc)))
        return None

    resolution_width = doc["data"].get("resolutionWidth")
    resolution_height = doc["data"].get("resolutionHeight")
    pixel_aspect = doc["data"].get("pixelAspect")

    # Make sure both width and height are set
    if resolution_width is None or resolution_height is None \
            or pixel_aspect is None:
        print("No resolution information found for \"{}\"".format(doc["name"]))
        return None

    return int(resolution_width), int(resolution_height), float(pixel_aspect)
```

This is very dirty and we should make a proper Qt dialog that's more user friendly and ideally something we can reuse in all other OP DCCs for startup dialog

fabiaserra:
This startup dialog should show up automatically every time an artist opens (not saves) a scene and has outdated variables. Doing it as an on-demand button it's helpful for the cases you want to validate or trigger an update because you know some have been updated but the most common case should be when an artist starts to work on a scene, they want to know if there's been an update on any of those and choose if he wants to pull them or not

fabiaserra:
> @fabiaserra Never mind man!
> 
> > you could add FPS to houdini vars, but that's redundant because `lib.validate_fps()` do that and it's registered in `on_open` and `on_save` callbacks.
> 
> I made a quick test with this PR
> 
> 1. I added these vars in settings
>    image
> 2. I made these changes in project manager
>    image
> 3. here's what I found when I opened my file
>    image
> 4. also, it's possible to update on demand from this menu button
>    image

Nice! that's looking good. Could we spend some cycles on this PR to make a better Qt dialog for this use case?

MustafaJafar:
> Nice! that's looking good. Could we spend some cycles on this PR to make a better Qt dialog for this use case? We should be able to update each of the variables independently (although resolution I would probably expect to be a single entry)

It's a great idea, however I prefer to add it as a `# TODO` in code because the main concern of this PR to update the Houdini vars when context changes.

BigRoy:
> It's a great idea, however I prefer to add it as a # TODO in code because the main concern of this PR to update the Houdini vars when context changes.

Can you create a Github issue to track it once this PR is merged?

The UX on path changes (since they are long) felt way off in [this screenshot](https://github.com/ynput/OpenPype/pull/5651#issuecomment-1749269611).

openpype/hosts/houdini/api/lib.py
```diff
@@ -747,3 +751,97 @@ def get_camera_from_container(container):
...
+        for var, (old, new, _is_directory) in update_vars.items()
+    )
+    parent = hou.ui.mainQtWindow()
+    dialog = popup.PopupUpdateKeys(parent=parent)
```
BigRoy, line 838:
```suggestion
    dialog = popup.Popup(parent=parent)
```
Can you test whether this would work too?
It should show the same pop-up but without the Update Keys checkbox :)

moonyuet:
> Nice! that's looking good. Could we spend some cycles on this PR to make a better Qt dialog for this use case? We should be able to update each of the variables independently (although resolution I would probably expect to be a single entry)

@fabiaserrathere there is popup dialog which can be imported by using `from openpype.widgets import popup`.
That one is used for checking outdated container and user can click fix button for updating container. Not sure it is a good choice and this Qtdialog looks so similar to that popup dialog(I used it for setting ocio config in 3dsmax in regards to the DCC's launching ocio config in 2024 while it doesn't have any feature before that)

openpype/hosts/houdini/api/lib.py
MustafaJafar: image

BigRoy: Ah I see, change `on_clicked_state` to `on_clicked`

BigRoy:
> @fabiaserra there there is popup dialog which can be imported by using `from openpype.widgets import popup`.
> That one is used for checking outdated container and user can click fix button for updating container. Not sure it is a good choice and this Qtdialog looks so similar to that popup dialog(I used it for setting ocio config in 3dsmax in regards to the DCC's launching ocio config in 2024 while it doesn't have any feature before that)

That's the one currently being used, also see [this comment](https://github.com/ynput/OpenPype/pull/5651#discussion_r1348410492).

openpype/hosts/houdini/api/lib.py
BigRoy: https://github.com/ynput/OpenPype/pull/5651#discussion_r1348444804

That just looks like your "new" values are empty strings? :)

moonyuet:
> > @fabiaserra there there is popup dialog which can be imported by using `from openpype.widgets import popup`.
> > That one is used for checking outdated container and user can click fix button for updating container. Not sure it is a good choice and this Qtdialog looks so similar to that popup dialog(I used it for setting ocio config in 3dsmax in regards to the DCC's launching ocio config in 2024 while it doesn't have any feature before that)
> 
> That's the one currently being used, also see [this comment](https://github.com/ynput/OpenPype/pull/5651#discussion_r1348410492).

Yes I noticed that just now too :). So ignore my previous comment.

openpype/hosts/houdini/api/lib.py
MustafaJafar: image
image

BigRoy: Seems better at least without the checkbox we we weren't actually using - right?
I mean it's far from beautiful, but fine prototype.

BigRoy (approved):
Code looks good for a first prototype. Solid work.

openpype/hosts/houdini/api/lib.py
MustafaJafar: okay, Then I will just add a TODO : Use better popup UI

fabiaserra:
> > Nice! that's looking good. Could we spend some cycles on this PR to make a better Qt dialog for this use case? We should be able to update each of the variables independently (although resolution I would probably expect to be a single entry)
> 
> @fabiaserra there there is popup dialog which can be imported by using `from openpype.widgets import popup`. That one is used for checking outdated container and user can click fix button for updating container. Not sure it is a good choice and this Qtdialog looks so similar to that popup dialog(I used it for setting ocio config in 3dsmax in regards to the DCC's launching ocio config in 2024 while it doesn't have any feature before that)

Yeah that's the same one I used for my dirty implementation of this... but it's just so Windows 98 haha I wish I could show you a screenshot of the startup dialog we had in my prior company, it was so pretty and clear for the artists. I think a simple vertical layout with horizontal layouts of three QLabels (variable, current and new value) and an "Update" button next to it, inheriting the stylesheet of the OP widgets with a title of "Outdated scene variables" and the OP logo would have already gone a long way and shouldn't take more than an hour to write

openpype/pipeline/context_tools.py
```diff
@@ -658,3 +661,70 @@ def get_process_id():
...
+
+    # compute `frameStartHandle` and `frameEndHandle`
+    if "frameStart" in asset_data and "handleStart" in asset_data:
```
iLLiCiTiT, line 719: Are these data used somewhere? We need to remove the `"assetData"` from here.

iLLiCiTiT: The point of all template data functions is that they return the same values. This is unique and does not return the same. If those data are needed on other places (other usecase than houdini environments), we can add it to base function `get_template_data`. Otherwise it's maintainance hell...

fabiaserra: I think this is used for the case where we want to set environment variables pointing to `{frameStartHandle}`/`{frameEndHandle}`

MustafaJafar: Hello, 
Is this sufficient ?  https://github.com/ynput/OpenPype/pull/5846

BigRoy: @MustafaJafar I believe @iLLiCiTiT already made a PR actually here https://github.com/ynput/OpenPype/pull/5845

PR #5652, opened 2023-09-26
General: Avoid fallback if value is 0 for handle start/end

fabiaserra:
## Changelog Description
There's a bug on the `pyblish_functions.get_time_data_from_instance_or_context` where if `handleStart` or `handleEnd` on the instance are set to value 0 it's falling back to grabbing the handles from the instance context. Instead, the logic should be that it only falls back to the `instance.context` if the key doesn't exist.

This change was only affecting me on the `handleStart`/`handleEnd` and it's unlikely it could cause issues on `frameStart`, `frameEnd` or `fps` but regardless, the `get` logic is wrong.

## Testing notes:
1. Assign handleStart/handleEnd value to 0 on the instance and some other value on the context data
2. Run a publish to the farm and check that the handleStart/handleEnd are set to 0

### DIFF
--- a/openpype/pipeline/farm/pyblish_functions.py
+++ b/openpype/pipeline/farm/pyblish_functions.py
@@ -112,6 +112,7 @@ def get_time_data_from_instance_or_context(instance):
         start=instance.data.get("frameStart", context.data.get("frameStart")),
         end=instance.data.get("frameEnd", context.data.get("frameEnd")),
         fps=instance.data.get("fps", context.data.get("fps")),
+        step=instance.data.get("byFrameStep", instance.data.get("step", 1)),
         handle_start=instance.data.get(
             "handleStart", context.data.get("handleStart")
         ),
--- a/openpype/pipeline/farm/pyblish_functions.py
+++ b/openpype/pipeline/farm/pyblish_functions.py
@@ -107,21 +107,16 @@ def get_time_data_from_instance_or_context(instance):
         TimeData: dataclass holding time information.
 
     """
+    context = instance.context
     return TimeData(
-        start=instance.data.get(
-            "frameStart", instance.context.data.get("frameStart")
-        ),
-        end=instance.data.get(
-            "frameEnd", instance.context.data.get("frameEnd")
-        ),
-        fps=instance.data.get(
-            "fps", instance.context.data.get("fps")
-        ),
+        start=instance.data.get("frameStart", context.data.get("frameStart")),
+        end=instance.data.get("frameEnd", context.data.get("frameEnd")),
+        fps=instance.data.get("fps", context.data.get("fps")),
         handle_start=instance.data.get(
-            "handleStart", instance.context.data.get("handleStart")
+            "handleStart", context.data.get("handleStart")
         ),
         handle_end=instance.data.get(
-            "handleEnd", instance.context.data.get("handleEnd")
+            "handleEnd", context.data.get("handleEnd")
         )
     )
 
--- a/openpype/pipeline/farm/pyblish_functions.py
+++ b/openpype/pipeline/farm/pyblish_functions.py
@@ -114,10 +114,12 @@ def get_time_data_from_instance_or_context(instance):
              instance.context.data.get("frameEnd")),
         fps=(instance.data.get("fps") or
              instance.context.data.get("fps")),
-        handle_start=(instance.data.get("handleStart") or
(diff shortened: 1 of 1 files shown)


fabiaserra:
@kalisp you are the last one that touched these lines so adding you as a reviewer

BigRoy (approved):
Looks good to me.
Would this be more readable if at the beginning of the function we do `context = instance.context` os that the lines below get clearer:
```python
instance.data.get("frameStart", context.data.get("frameStart"))
```

fabiaserra:
> Looks good to me. Would this be more readable if at the beginning of the function we do `context = instance.context` os that the lines below get clearer:
> 
> ```python
> instance.data.get("frameStart", context.data.get("frameStart"))
> ```

sounds good to me yeah, let me push that change

fabiaserra:
> > Looks good to me. Would this be more readable if at the beginning of the function we do `context = instance.context` os that the lines below get clearer:
> > ```python
> > instance.data.get("frameStart", context.data.get("frameStart"))
> > ```
> 
> sounds good to me yeah, let me push that change

Done here https://github.com/ynput/OpenPype/pull/5652/commits/6e9257c31be430c92f82d0a385634966c6a7fca3

BigRoy:
Shouldn't the `step` data also get set from the instance's data to the `TimeData` instance?

Basically adding:
```python
        step=instance.data.get("step", 1)
```

I believe however the OpenPype codebase uses a mixture of `step` or `byFrameStep` for this attribute. So we might need to do:
```python
        step=instance.data.get("byFrameStep", instance.data.get("step", 1))
```

fabiaserra:
> ### What about `step`?
> Shouldn't the `step` data also get set from the instance's data to the `TimeData` instance?
> 
> Basically adding:
> 
> ```python
>         step=instance.data.get("step", 1)
> ```
> 
> I believe however the OpenPype codebase uses a mixture of `step` or `byFrameStep` for this attribute. So we might need to do:
> 
> ```python
>         step=instance.data.get("byFrameStep", instance.data.get("step", 1))
> ```
> 
> I did notice that `TimeData` forces `step` to be `int` even though the `step` is used in e.g. alembic exports or yeti exports to indicate subframe steps, so usually are `0.1` or alike. Is this another issue with the TimeData class?
> 
> Might need to be an issue/PR of its own.

Fair enough, I will let @jakubjezek001 comment on that one

> ### What about `step`?
> Shouldn't the `step` data also get set from the instance's data to the `TimeData` instance?
> 
> Basically adding:
> 
> ```python
>         step=instance.data.get("step", 1)
> ```
> 
> I believe however the OpenPype codebase uses a mixture of `step` or `byFrameStep` for this attribute. So we might need to do:
> 
> ```python
>         step=instance.data.get("byFrameStep", instance.data.get("step", 1))
> ```
> 
> I did notice that `TimeData` forces `step` to be `int` even though the `step` is used in e.g. alembic exports or yeti exports to indicate subframe steps, so usually are `0.1` or alike. Is this another issue with the TimeData class?
> 
> Might need to be an issue/PR of its own.

Fair enough, I have now added `step` here https://github.com/ynput/OpenPype/pull/5652/commits/86bbadf7ce10930bc4a25517235cf70d788a3c93.

Should we also check under `context.data`?

I will let @jakubjezek001 comment on the int vs float and confirm what we should do here

iLLiCiTiT:
NOTE: Is there a chance that the key is filled with `None`? In that case `.get("...key...", other.get("...key..."))` won't help.

BigRoy:
> NOTE: Is there a chance that the key is filled with `None`? In that case `.get("...key...", other.get("...key..."))` won't help.

Good question - I'd say _if_ that's the case then **that** is a bug. :)

fabiaserra:
> > NOTE: Is there a chance that the key is filled with `None`? In that case `.get("...key...", other.get("...key..."))` won't help.
> 
> Good question - I'd say _if_ that's the case then **that** is a bug. :)

Yeah... and if values are "explicitly" set on the instance, those should take preference regardless if it's None IMO (although probably it should never be None...)

kalisp (approved):

BigRoy:
> I did notice that TimeData forces step to be int even though the step is used in e.g. alembic exports or yeti exports to indicate subframe steps, so usually are 0.1 or alike. Is this another issue with the TimeData class?
> 
> Might need to be an issue/PR of its own.

Does anyone know more about this? I suspect there's a hidden bug there?

PR #5654, opened 2023-09-26
Increase timout for deadline test

kalisp:
## Changelog Description
DL picks up jobs quite slow, so bump up delay.


iLLiCiTiT (approved):

64qam (approved):

BigRoy:
I believe it's the "Worker intervals" you're looking for?

image

Since publish jobs also need to release their "Dependent" state you might need to tweak the pending job scan interval too:

image

Does that help?

kalisp:
I tweaked DL similarly, test suite was a bit quicker, but I'll keep monitoring it for better average numbers.

I'll merge it nevertheless, it shouldn't hurt and it will make suite a more stable previous timeout wasn't hit all the time.

PR #5656, opened 2023-09-27
Unreal: Parent Qt UI to Window main Unreal

sebasti1a:
## Changelog Description
Improved Qt window parenting for Unreal window. I disabled the setWindowFlags : WindowStaysOnTopHint and I use this function : unreal.parent_external_window_to_slate

## Additional info
Test with Unreal 5.3

## Testing notes:
1. Open Unreal
2. Click on Ayon Menu
3. The popup UI is now correctly parent with unreal

### DIFF
--- a/openpype/hosts/unreal/api/tools_ui.py
+++ b/openpype/hosts/unreal/api/tools_ui.py
@@ -62,6 +62,7 @@ class ToolsBtnsWidget(QtWidgets.QWidget):
 
 class ToolsDialog(QtWidgets.QDialog):
     """Dialog with tool buttons that will stay opened until user close it."""
+
     def __init__(self, *args, **kwargs):
         super(ToolsDialog, self).__init__(*args, **kwargs)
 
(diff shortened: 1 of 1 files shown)


simonebarbieri:
Hello. I'm not entirely sure about the purpose of this PR. The current window is made to behave like a dropdown menu, but we have a dialog accessible from the Tools menu that already behaves like this PR!
image

sebasti1a:
Actually, It's the Qt window that is not correctly linked to the Unreal application, for example when I minimize Unreal, the Ayon tools window still remains visible in WindowStaysOnTop. My PR fixes that.

image

simonebarbieri:
Ah, in that case, that works, but I think that the menu should still behave as a dropdown menu and disappear when out of focus :)
Now the two options (popup and dialog) behaves the same, at least with Unreal 5.2! Haven't tested 5.3.

sebasti1a:
I tested with Unreal 5.3 and "AyonTools" is still a drop-down menu (the window is not persistent and disappear when out of focus ). Like expected behavior.

antirotor (changes requested):
Strange thing is that my  UE 5.3 is actually showing this:

op_ue_ayontools_qt_01 gif

Could it be because of the Qt version? Mine is:

image

mkolar:

We're closing it down, but we'll he happy for a new PR to [ynput/ayon-unreal](https://github.com/ynput/ayon-unreal) repository once it's up.

PR #5660, opened 2023-09-28
Feature/force download last wf

Sharkitty:
## Changelog Description
Adding a `Force download last workfile` feature, which forces openpype to download the last workfile (using the pre copy last workfile hook) even if there already are local workfiles. This doesn't check if the last local workfile is the same as the one downloaded, as a design choice. The point of this feature is to allow users to easily ensure they get a brand new download of the last published workfile, for instance in case their last local workfile is borked, or if they already know their workfile is not the latest and they will need to download a new one anyway.

Newly downloaded workfile version number will either be:
- Last local workfile + 1 if the last local workfile has a higher version number than the last published workfiles.
- Last published workfile + 1 otherwise.

This feature available in the launcher by right clicking on the host app icon and ticking `Force download last workfile`. A small download icon then appears to the lower right of the host app icon to show the user that this feature is active for the selected project, asset, task and host, along with `(Force download last workfile)` added to the app tooltip.

This is compatible with the existing `Skip opening last workfile` feature in case not only you need a force download, but you also need to open the workfile directly from the DCC.

This PR also includes some refactor the `Skip opening last workfile` feature, in order to make some of the code more reusable and accommodate required changes for the `Force download last workfile` feature. It would now be easier to add more options to the context menu, but it would still require quite a bit of work and understanding of the implementation compared to the system used for plugins.

Finally the `Skip opening last workfile` marker visible on the host app button has been moved to be closer to the icon.

## Known limitations
When activating both `Skip opening last workfile` and `Force download last workfile` at the same time:
- Sometimes `Force download last workfile` gets unticked automatically (I don't understand why).
- The order in which these feature appear in the host app tooltip depend on the order in which these features have been activated.

## Testing notes:
- Open the launcher and select any asset and task.
- Right click on the host app and tick `Force download last workfile`, a download icon should appear and the host app tool tip should contain `(Force download last workfile)`.
- Click on the host app, the last published workfile should be downloaded and opened. Its version number should be `last local workfile + 1` (If there were local workfiles already) or `last published workfile + 1`. No workfile should ever be overwritten when doing this.
- Close the DCC.
- Right click on the host app and untick `Force download last workfile`, the download icon should disappear, the tooltip should no longer contain `(Force download last workfile)` (may require to hover out and hover back in).
- Click on the host app, it should launch without downloading a new workfile.

### DIFF
--- a/openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
+++ b/openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
@@ -10,6 +10,8 @@ from openpype.modules.sync_server.sync_server import (
 from openpype.pipeline.template_data import get_template_data
 from openpype.pipeline.workfile.path_resolving import (
     get_workfile_template_key,
+    get_last_workfile_with_version,
+    get_workdir,
 )
 from openpype.settings.lib import get_project_settings
 
@@ -49,7 +51,9 @@ class CopyLastPublishedWorkfile(PreLaunchHook):
 
         # Check there is no workfile available
         last_workfile = self.data.get("last_workfile_path")
-        if os.path.exists(last_workfile):
+        if os.path.exists(last_workfile) and not self.data.get(
+            "force_download_last_workfile"
+        ):
             self.log.debug(
                 "Last workfile exists. Skipping {} process.".format(
                     self.__class__.__name__
@@ -179,7 +183,22 @@ class CopyLastPublishedWorkfile(PreLaunchHook):
 
         extension = last_published_workfile_path.split(".")[-1]
         workfile_data["version"] = (
-                workfile_representation["context"]["version"] + 1)
+            get_last_workfile_with_version(
+                get_workdir(
+                    project_doc,
+                    asset_doc,
+                    task_name,
+                    host_name,
+                    anatomy=anatomy,
+                    template_key=template_key,
+                    project_settings=project_settings,
+                ),
+                anatomy.templates[template_key]["file"],
+                workfile_data,
+                [extension],
+            )[1]
+            or workfile_representation["context"]["version"]
+        ) + 1
         workfile_data["ext"] = extension
 
         anatomy_result = anatomy.format(workfile_data)
--- a/openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
+++ b/openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
@@ -184,24 +184,36 @@ class CopyLastPublishedWorkfile(PreLaunchHook):
 
         extension = last_published_workfile_path.split(".")[-1]
 
-        # Use last local workfile version + 1
-        # or last published workfile version + 1
+        # Get last published workfile version
+        last_published_workfile_version = workfile_representation["context"][
+            "version"
+        ]
+
+        # Get last local workfile version
+        last_local_workfile_version = get_last_workfile_with_version(
+            get_workdir(
+                project_doc,
+                asset_doc,
+                task_name,
+                host_name,
+                anatomy=anatomy,
+                template_key=template_key,
+                project_settings=project_settings,
+            ),
+            anatomy.templates[template_key]["file"],
+            workfile_data,
+            [extension],
+        )[1]
+
+        # Use last published workfile version + 1
+        # or last local workfile version + 1 if it's higher
         workfile_data["version"] = (
-            get_last_workfile_with_version(
-                get_workdir(
-                    project_doc,
-                    asset_doc,
-                    task_name,
-                    host_name,
-                    anatomy=anatomy,
-                    template_key=template_key,
-                    project_settings=project_settings,
-                ),
-                anatomy.templates[template_key]["file"],
-                workfile_data,
-                [extension],
-            )[1]
-            or workfile_representation["context"]["version"]
+            last_local_workfile_version if (
+                last_local_workfile_version
+                and last_local_workfile_version
+                > last_published_workfile_version
+            )
+            else last_published_workfile_version
         ) + 1
         workfile_data["ext"] = extension
 
--- a/openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
+++ b/openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
@@ -49,7 +49,8 @@ class CopyLastPublishedWorkfile(PreLaunchHook):
             self.log.debug("Sync server module is not enabled or available")
             return
 
-        # Check there is no workfile available
+        # Skip if there are local workfiles
+        # and force download last workfile is active
         last_workfile = self.data.get("last_workfile_path")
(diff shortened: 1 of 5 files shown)


BigRoy:
> Newly downloaded workfile version number will either be:
> 
> - Last local workfile + 1 if any local workfile is found.
> - Last published workfile + 1 if no local workfile is found (i.e. what the pre copy last workfile hook already does even without this feature).

Few notes or questions about this.

### Always increment publish version + 1 unless a newer workfile version already exists

Wouldn't you expect maybe that if the publish version is v010 and you do have a workfile v001 that the new workfile still ends up being v011 or did you find that more confusing?

### Allow settings to specify a subversion/comment string for downloaded files.

Should it download to a version with a subversion string like `download` or allow in settings to define if a subversion/comment string should be appended to the filename when it's a downloaded file?

It'd basically be what Work Files tool allows to set via _subversion_ here:

image

### Set Artist Note for downloaded files

Additionally - it'd be epic if we can make it so that the downloaded file also gets an "Artist Note" (like you can set in the Work Files tool) and the Artist Note would e.g. be: "Download of published file X Y Z" to really clarify the situation?

E.g. like:
image

Sharkitty:
> ### Always increment publish version + 1 unless a newer workfile version already exists
> 
> Wouldn't you expect maybe that if the publish version is v010 and you do have a workfile v001 that the new workfile still ends up being v011 or did you find that more confusing?

Yes now that I think about it, what you describe would make more sense. I'll see what I can do about it.

> ### Allow settings to specify a subversion/comment string for downloaded files.
> 
> Should it download to a version with a subversion string like `download` or allow in settings to define if a subversion/comment string should be appended to the filename when it's a downloaded file?
> 
> It'd basically be what Work Files tool allows to set via _subversion_ here:
> 
> image
> ### Set Artist Note for downloaded files

That's a good idea, I'll make that change.

> Additionally - it'd be epic if we can make it so that the downloaded file also gets an "Artist Note" (like you can set in the Work Files tool) and the Artist Note would e.g. be: "Download of published file X Y Z" to really clarify the situation?
> 
> E.g. like: image

Might as well try that too.

openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
```diff
@@ -178,8 +183,38 @@ def execute(self):
...
+        # or last local workfile version + 1 if it's higher
         workfile_data["version"] = (
-                workfile_representation["context"]["version"] + 1)
+            last_local_workfile_version if (
+                last_local_workfile_version
+                and last_local_workfile_version
+                > last_published_workfile_version
+            )
+            else last_published_workfile_version
+        ) + 1
```
BigRoy, line 217: This is quite an unreadable statement.

A bit verbose but maybe this is more readable?
```python
versions = []
if last_local_workfile_version is not None:
    versions.append(last_local_workfile_version)
if last_published_workfile_version is not None:
    versions.append(last_published_workfile_version)
new_version = max(versions, default=0) + 1
```
Or:
```python
new_version = max((
    version for version in [
        last_local_workfile_version,
        last_published_workfile_version
    ]
    if version is not None
), default=0) + 1
```
But admittedly that looks like crazy mumbo jumbo again.

iLLiCiTiT:
```
new_version = max([
    last_local_workfile_version or 0,
    last_published_workfile_version or 0
]) + 1
```

Sharkitty: I went with @iLLiCiTiT implementation, but with a generator. I don't think it's worth using a list in this case.

Sharkitty:
I add `downloaded` ass the subversion but let me know if you think of a better name.

openpype/modules/sync_server/launch_hooks/pre_copy_last_published_workfile.py
BigRoy: FYI - looking at the code I don't think that's a generator but it produces a `tuple`.

Sharkitty:
Since artist notes are pretty much implemented within the UI, I think it would be tricky to use them cleanly in this context. So I think it might be best not to use artist notes for this feature.

Besides that, versioning has been fixed, and subversions are used as requested.

kalisp:
I am not sold on this. It seems as a lot of code to solve an issue that could be solved by single hit on Delete key.

Sharkitty:
> I am not sold on this. It seems as a lot of code to solve an issue that could be solved by single hit on Delete key.

There's no delete option within the launcher as far as I know. If you suggest going to the workfile directory to delete workfiles. the whole point is to avoid it. It's much simpler for users to have a button that does what they need than to fidget around in files, and what I'm proposing is non destructive. I also consider risky to expect user to manually do these file operations. This would be prone to human error, in fact I already had to fix human errors due to the lack of a feature such as the one I'm proposing.

BigRoy:
> There's no delete option within the launcher as far as I know. If you suggest going to the workfile directory to delete workfiles. the whole point is to avoid it. It's much simpler for users to have a button that does what they need than to fidget around in files, and what I'm proposing is non destructive. I also consider risky to expect user to manually do these file operations. This would be prone to human error, in fact I already had to fix human errors due to the lack of a feature such as the one I'm proposing.

Having said that - would you be ok if this were an individual launcher action? So that in the launcher there's an extra entry "download last workfile" instead of adding the 'force download toggle' and everything? I feel like you'd rarely want to always have force download enabled the whole time and it's a on-off thing anyway?

Sharkitty:
> > There's no delete option within the launcher as far as I know. If you suggest going to the workfile directory to delete workfiles. the whole point is to avoid it. It's much simpler for users to have a button that does what they need than to fidget around in files, and what I'm proposing is non destructive. I also consider risky to expect user to manually do these file operations. This would be prone to human error, in fact I already had to fix human errors due to the lack of a feature such as the one I'm proposing.
> 
> Having said that - would you be ok if this were an individual launcher action? So that in the launcher there's an extra entry "download last workfile" instead of adding the 'force download toggle' and everything? I feel like you'd rarely want to always have force download enabled the whole time and it's a on-off thing anyway?
> 
> Like you said - that'd make it just a simple button they could press - which could even then offer them options like, which workfile and show some little info - what site did it copy from? what will it save it as? maybe even prompt to the artist note field and file suffix etc so they can customize if they want.

Yes I agree. I chose a tick box to keep the current design of the skip opening last workfile feature. But what you're describing sounds even better. However it would take more development time, and the feature freeze is in just two weeks. I guess for now a download button that just downloads the last published workfile might be more realistic (I don't know if I'll have time to do it though), and the more feature rich menu you're proposing could be done directly in Ayon. What do you think?

mkolar:
Because we're splitting OpenPype into ayon-core and individual host addons, this PR would have to be re-created to target one of those. Also, similarly to @kalisp I'm not convinced that this logic should even make it to the main codebase. We should find a more robust way of handling the core problem here in AYON

We're closing it down, but we'll he happy for a new PRs to [ynput/ayon-core](https://github.com/ynput/ayon-core) repository once it's up.

PR #5665, opened 2023-09-29
Resolve: load clip to timeline at set time

jakubjezek001:
## Changelog Description
It is possible to load clip to correct place on timeline.

# Todo
- [x] add option to loader plugin to add sequencial loading as an opose to in place loading
- [ ] ~seek the way to add it always to new layer as it is done in Hiero. layer name inherited from product name~

## Testing notes:
1. Open testing timeline in project
2. load clip from loader and see it is added at correct place on timeline which had been captured during original editorial ingest. 
2. follow this step


openpype/hosts/resolve/api/lib.py
```diff
@@ -278,6 +279,12 @@ def create_timeline_item(media_pool_item: object,
             clip_data.update({"endFrame": source_end})
 
         print(clip_data)
+
+        if timeline_in:
```
BigRoy, line 283: Should this be?
```suggestion
        if timeline_in is not None:
```
Or does default load always start at timeline at 0 anyway?

openpype/hosts/resolve/api/plugin.py
```diff
@@ -402,6 +402,9 @@ def load(self):
         if handle_end is None:
             handle_end = int(self.data["assetData"]["handleEnd"])
 
+        self.timeline_in = int(self.data["assetData"]["clipIn"])
```
BigRoy, line 405: Interesting - I though the PR description meant it would place the loaded item at the current playbar time (like at current position) and not use the clip in from the asset data.

Also, there's no reason this is stored on `self` - please just make it `timeline_in =` etc. and use it as the var like that. Right?

jakubjezek001:
I have been searching for a solution to consistently add new layers in the same way as it is done in Hiero. Specifically, I want the layer name to be inherited from the product name and representation. Unfortunately, the latest Resolve API does not support setting any timeline track as active. As a result, any new objects loaded into the timeline will not automatically be added to the activated track.

BigRoy:
Could you make a screen recording of how it behaves in Hiero and how it currently behaves in Resolve to show the differences?

openpype/hosts/resolve/api/plugin.py
```diff
@@ -410,13 +423,18 @@ def load(self):
             source_out -= handle_end
 
         # include handles
-        if self.with_handles:
-            source_in -= handle_start
-            source_out += handle_end
+        if not self.with_handles:
```
Minkiu, line 426: Could it be a situation where line 421 and this one to be true, so we add /substract twice the handles?

openpype/hosts/resolve/api/lib.py
```diff
@@ -243,10 +246,13 @@ def get_media_pool_item(fpath, root: object = None) -> object:
...
 
-def create_timeline_item(media_pool_item: object,
-                         timeline: object = None,
-                         source_start: int = None,
-                         source_end: int = None) -> object:
+def create_timeline_item(
+        media_pool_item: object,
+        source_start: int,
+        source_end: int,
+        timeline_in: int,
```
Minkiu, line 253: We are missing these new ones on the docstring.

m-u-r-p-h-y (approved):
sequence loaded according the Clip In data

image

openpype/hosts/resolve/api/lib.py
```diff
@@ -266,28 +272,41 @@ def create_timeline_item(media_pool_item: object,
...
+        clip_data = {
+            "mediaPoolItem": media_pool_item,
+            "startFrame": source_start,
+            "endFrame": source_end,
+            "recordFrame": timeline_in,
+        }
 
+        print("clip_data", "_" * 50)
+        print(media_pool_item.GetName())
         print(clip_data)
```
BigRoy: Let's remove the debug prints?
```suggestion
```

openpype/hosts/resolve/api/lib.py
```diff
@@ -243,10 +246,13 @@ def get_media_pool_item(fpath, root: object = None) -> object:
...
-                         timeline: object = None,
-                         source_start: int = None,
-                         source_end: int = None) -> object:
+def create_timeline_item(
+        media_pool_item: object,
+        source_start: int,
+        source_end: int,
+        timeline_in: int,
+        timeline: object = None
+) -> object:
```
BigRoy, line 258: Why do we now require these arguments? I'd expect the lib function to still just allow to the relevant entry if no source start is specificied, etc.

I'd also expect the arguments ordered:
```suggestion
def create_timeline_item(
        media_pool_item: object,
        timeline: object = None,
        timeline_in: int = None,
        source_start: int = None,
        source_end: int = None,
) -> object:
```

So that to just add a clip to a timeline would be: `create_timeline_item(item, timeline)`

openpype/hosts/resolve/api/lib.py
```diff
@@ -266,28 +272,41 @@ def create_timeline_item(media_pool_item: object,
...
         output_timeline_item = get_timeline_item(
             media_pool_item, timeline)
 
-    assert output_timeline_item, AssertionError(
-        "Track Item with name `{}` doesn't exist on the timeline: `{}`".format(
-            clip_name, timeline.GetName()
-        ))
+    assert output_timeline_item, AssertionError((
+        "Clip name '{}' was't created on the timeline: '{}' \n\n"
+        "Please check if the clip is in the media pool or if the timeline \n"
```
BigRoy, line 307: How would the `media_pool_item` object that is an input to this function NOT be in the media pool? :)

openpype/hosts/resolve/api/lib.py
```diff
@@ -266,28 +272,41 @@ def create_timeline_item(media_pool_item: object,
...
+        "is having activated correct track name, or if it is not already \n"
+        "having any clip add place on the timeline in: '{}' out: '{}'. \n\n"
+        "Clip data: {}"
+    ).format(
+        clip_name, timeline.GetName(), timecode_in, timecode_out, clip_data
+    ))
```
BigRoy, line 313:
```suggestion
    assert output_timeline_item, AssertionError((
        "Clip name '{}' wasn't created on the timeline: '{}' \n\n"
        "Please check if the clip is in the media pool or if the timeline \n"
        "has correct track name activated, or if it does not already \n"
        "have any clip add place on the timeline in: '{}' out: '{}'. \n\n"
        "Clip data: {}"
    ).format(
        clip_name, timeline.GetName(), timecode_in, timecode_out, clip_data
    ))
```

Even with these grammar tweaks suggested I'm not sure what the last OR statement means. If it doesn't have any clip on the timeline yet, what should I do? And what did I do wrong if I hit that particular case?

openpype/hosts/resolve/api/plugin.py
jakubjezek001: good point, better to merge both those conditions

openpype/hosts/resolve/api/lib.py
```diff
@@ -266,28 +272,41 @@ def create_timeline_item(media_pool_item: object,
     clip_name = _clip_property("File Name")
     timeline = timeline or get_current_timeline()
 
+    # timing variables
+    fps = project.GetSetting("timelineFrameRate")
```
BigRoy, line 280: A resole timeline can deviate from the project FPS.
E.g. if Create New Timeline had Use Project Settings enabled bottom left:

image

This should instead be getting the setting from the timeline:
```suggestion
    fps = timeline.GetSetting("timelineFrameRate")
```

openpype/hosts/resolve/api/plugin.py
```diff
@@ -418,12 +418,11 @@ def load(self):
...
 
-        if _clip_property("Type") == "Video":
-            source_in += handle_start
-            source_out -= handle_end
-
         # include handles
-        if not self.with_handles:
+        if (
+            not self.with_handles
+            or _clip_property("Type") == "Video"
```
BigRoy, line 429: Why would a video type **require** handles to be applied?

jakubjezek001: Yeah, that is a very good question. I am not sure myself to be honest. It was there historically and I have already forgotten the reason. In fact, I would also remove the condition. Video or sequence, either it is with handles or not, there should not be a difference in how we are treating it.

openpype/hosts/resolve/api/lib.py
jakubjezek001: Yeah, you are right, this lost the backward compatibility..

jakubjezek001: and other related changes here 435408dc142cb934424333257f8092db37974b01

openpype/hosts/resolve/api/lib.py
jakubjezek001: done

openpype/hosts/resolve/api/lib.py
jakubjezek001: Yeah, it doesn't make sense. Fixing it.

openpype/hosts/resolve/api/lib.py
jakubjezek001: Yeah, I have almost broken my brain too.. here is fix a810dfe78a97eb1467d3745ee786f2743f6faac2

openpype/hosts/resolve/api/lib.py
jakubjezek001: fixed here a11467883b5d1cec302f43361e7d939ff502ac07

openpype/hosts/resolve/api/plugin.py
jakubjezek001: check this commit if you agree with the solution. fbc24f565415cc6cbb88547c9b333bf41a52a3e0

Now the db frame duration is compared with representation frames availability (the same on sequence or video). This is for case that representation was published with tag `no_handles`. Function is first checking versionData frame start/end and of keys are not available then it looks to the asset data.

BigRoy: I disagree with that actually - the asset range is not a replacement for whatever the frame range a published representation would be. I'd say, if no frame range is known that we either get the actual clip's length somehow (does Resolve API allow us to get the length of a clip? I mean the Media Pool can show the length of a clip - I'd assume we could query that somehow?

For example:
```python
media_pool_item.GetClipProperty("Duration")
# 00:00:02:02
media_pool_item.GetClipProperty("End TC")
# 00:00:02:02
media_pool_item.GetClipProperty("Start TC")
# 00:00:00:00
media_pool_item.GetClipProperty("Type")
# Video + Audio
media_pool_item.GetClipProperty("Frames")
# 52
media_pool_item.GetClipProperty("FPS")
# 25
``` 

Why couldn't we use any of that clip information to know the **actual** clip length?

openpype/hosts/resolve/api/plugin.py
```diff
@@ -312,6 +312,9 @@ def __init__(self, cls, context, path, **options):
         # try to get value from options or evaluate key value for `load_to`
         self.new_timeline = options.get("newTimeline") or bool(
             "New timeline" in options.get("load_to", ""))
+        # try to get value from options or evaluate key value for `load_how`
+        self.sequential_load = options.get("sequentially") or bool(
+            "Sequentially in order" in options.get("load_how", ""))
```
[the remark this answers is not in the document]
jakubjezek001: good point ed2756aa52372b296b5e0307e763339dd5890250

openpype/hosts/resolve/api/plugin.py
BigRoy: I'd say that on load we'd only ever apply the handles of the version or representation then, yes.
If someone wanted to instead apply the handles of the `asset` that should then be an explicit toggle somewhere - I wouldn't expect that to be the default behavior on load.

jakubjezek001: We are trying to obtain frame data from version first and then as fallback we are getting them from the asset. The only reason is that version data are not always having them (although if version is not having them it would be really strange).

jakubjezek001: And what about the case where handles are defined on version data, but representation files are not integrated  handles? 

Anyway, I agree that we should remove the asset data fallback.

BigRoy: > And what about the case where handles are defined on version data, but representation files are not integrated handles?

In that case - again, it's an error on the ingestion. If the representation differs from the version then the representation data should contain the differed data. That's an error on the publish itself, not something we need to 'fix' on load.

The frame range should thus be:

- frame range + handles of representation
- if not found, frame range + handles of version

Minkiu (approved):
Haven't tested, but code looks good

openpype/hosts/resolve/api/plugin.py
mkolar: >In that case - again, it's an error on the ingestion. If the representation differs from the version then the representation data should contain the differed data. That's an error on the publish itself, not something we need to 'fix' on load.

probably just a difference in opinions, but I don't think it's necessarily an error on ingest. We have plenty of production where .exr representation would have handles, but .mov would not. That way it was easy to quickly assemble the reviewables and see shot hookups in ftrack for instance, by allowing this `no-handles` representation. For episodic animation, it was an absolute must.

jakubjezek001:
Task linked: [OP-6991 Resolve: loader failing load](https://app.clickup.com/t/6658547/OP-6991)

openpype/hosts/resolve/api/plugin.py
jakubjezek001: hey @BigRoy are we ok to merge this PR.

openpype/hosts/resolve/api/plugin.py
```diff
@@ -391,30 +398,70 @@ def load(self, files):
...
+        source_with_handles = True
+        if source_duration < db_frame_duration:
+            source_with_handles = False
+
+        # only exclude handles if source has no handles or
+        # if user wants to load without handles
+        if (
+            not self.with_handles
+            or not source_with_handles
+        ):
```
jakubjezek001, line 453: I believe this condition should be working as substitution for the previous case where only **Video** property type was allowed `if _clip_property("Type") == "Video":`. Here we are not excluding handles to all, but only to those resources, which are either missing handles or it is chosed by user not to load with handles.

BigRoy: I personally don't understand the logic.

```
        source_with_handles = True
        if source_duration < db_frame_duration:
            source_with_handles = False
```
This does not tell me whether the source has no handles, right? Could you explain your thought process?

To me it's this:
```python
start_handles = version_data.get("handleStart", 0)
end_handles = version_data.get("handleEnd", 0)
```
And **never** the asset handles on file load - the version in the database, is the ground truth of the file I'd argue.

My argument is that: If a publish has no handles data defined, then it has no handles. Again, it's an issue at the time of ingesting not something we need to solve here. As such, we don't need to somehow magically "solve" what the handles for a file are - the database is the ground truth of what we're expecting to load. If it doesn't tell me it has handles, then it has no handles.

openpype/hosts/resolve/api/plugin.py
BigRoy: > probably just a difference in opinions, but I don't think it's necessarily an error on ingest. We have plenty of production where .exr representation would have handles, but .mov would not.

Doesn't that just mean that we're lacking `frameStart`, `frameEnd`, `handleStart` and `handleEnd` data per representation for those cases?

> That way it was easy to quickly assemble the reviewables and see shot hookups in ftrack for instance, by allowing this no-handles representation. For episodic animation, it was an absolute must.

Totally agree. That's definitely useful. But still, I'd argue that the database would be capable of what file it's providing me instead of somehow saying "just use whatever the asset is set to".

jakubjezek001: This is not guessing of mis-published data so this conversation is misleading. Just to wrap this up before I merge it. Data are published correctly without handles since Extract review output preset is supporting tag `no handles` - this is then supporting to define two representations to be created under one version which is having handles - one representation with hadles and other without handles.

This loading solution is supporting loading both of those representation under one version.

openpype/hosts/resolve/api/plugin.py
jakubjezek001:
```
source_with_handles = True
if source_duration < db_frame_duration:
            source_with_handles = False
```
By defautl we asume source data in representation are having handles. 
Comparing length of frams in database on version data with representation amount of frames and if those are smaller then source is with no handles. 

```
start_handles = version_data.get("handleStart", 0)
end_handles = version_data.get("handleEnd", 0)
```
This solution would not work in context of this comment https://github.com/ynput/OpenPype/pull/5665#discussion_r1381523374

I am closing the conversation since we really need to mege this PR. 

Any other suggestions can be created in form of issue referencing this PR.

BigRoy: Ah I see - I finally understand.

In that case your logic is wrong because now you're checking whether the duration of the file is shorter than the database frame length WITHOUT handles. You're checking it against the frame range without handles.

Instead you **should** be doing this:
```python
frame_start = version_data.get("frameStart")
frame_end = version_data.get("frameEnd")
if frame_start is not None and frame_end is not None:
    # The version has defined frame range data
    handle_start = version_data.get("handleStart", 0)
    handle_end = version_data.get("handleEnd", 0)
else:
    # Unfortunately the publish we are loading has no knowledge about its frame range
    # so all we can really do is rely on the asset's frame range as a fallback.
    # TODO: If the source does not have any frame range information we should actually
    #   be performing no custom fallbacks whatsoever
    frame_start = asset_data["frameStart"]
    frame_end = asset_data["frameEnd"]
    handle_start = asset_data["handleStart"]
    frame_end = asset_data["handleEnd"]

frame_start_handle = frame_start - handle_start
frame_end_handle = frame_end + handle_end

# The representation we are loading *can be* a representation with less frames than the version
# if it were e.g. a preview output without handles of the version. Best case scenario this data should
# be stored with the `representation` as its range explicitly, but for now we will check whether the
# loaded media is actually shorter than the database range or not - if so, we will assume the file
# is without handles
db_duration = frame_end_handle - frame_start_handle + 1
source_with_handles = source_duration >= db_duration 
```

PR #5666, opened 2023-09-29
Enhancement : Houdini dynamic publish parameters

MustafaJafar:
## Changelog Description
I was exploring Houdini ideas/issues/PRs and came across allow houdini publish instances duplication. 
I ~stole some code from~ was inspired by these two PRs https://github.com/ynput/OpenPype/pull/5490 and https://github.com/ynput/OpenPype/pull/5616 

I was abstracting their ideas and trying imitating their functionalities by Houdini expressions. 
one interesting thing that when houdini evaluates the value of a parameter it also evaluate any expressions used inside it. 
so, this is actually works! 

This is my first test, where I edited manually these parameters 
> one limitation is that family must be included in the node's name because you can have the same variant for multiple families
> and that's why I can't use node's name for variant value

- subset name = `opname(".")`
- variant = `strreplace(chs("subset"), chs("family"), "")`
- instance_node = `opfullpath(".")`
- instance_id = `opfullpath(".")`
- use `chs('subset')`

> and that's a thing that you can try it yourself without pulling this branch! 
 
image

Here are some tests from this branch 

https://github.com/ynput/OpenPype/assets/20871534/e9d01f86-53e9-40cc-b29b-61d9d357198d

https://github.com/ynput/OpenPype/assets/20871534/be5e2758-0e6a-427f-8018-b5233847be78

## Testing notes:
1. Create publish instances
2. Duplicate them as many as you want
3. publish

### DIFF
--- a/openpype/hosts/houdini/api/plugin.py
+++ b/openpype/hosts/houdini/api/plugin.py
@@ -182,18 +182,19 @@ class HoudiniCreator(NewCreator, HoudiniCreatorBase):
                 node_type = "geometry"
 
             instance_node = self.create_instance_node(
-                subset_name, "/out", node_type)
+                instance_data["variant"], "/out", node_type)
 
             self.customize_node_look(instance_node)
 
             instance_data["instance_node"] = instance_node.path()
+            instance_data["instance_id"] = instance_node.path()
             instance = CreatedInstance(
                 self.family,
                 subset_name,
                 instance_data,
                 self)
             self._add_instance_to_context(instance)
-            imprint(instance_node, instance.data_to_store())
+            self.imprint(instance_node, instance.data_to_store())
             return instance
 
         except hou.Error as er:
@@ -222,24 +223,43 @@ class HoudiniCreator(NewCreator, HoudiniCreatorBase):
         self.cache_subsets(self.collection_shared_data)
         for instance in self.collection_shared_data[
                 "houdini_cached_subsets"].get(self.identifier, []):
+
+            node_data = read(instance)
+
+            # Node paths are always the full node path since that is unique
+            # Because it's the node's path it's not written into attributes
+            # but explicitly collected
+            node_path = instance.path()
+            node_data["instance_id"] = node_path
+            node_data["instance_node"] = node_path
+
             created_instance = CreatedInstance.from_existing(
-                read(instance), self
+                node_data, self
             )
             self._add_instance_to_context(created_instance)
 
     def update_instances(self, update_list):
         for created_inst, changes in update_list:
             instance_node = hou.node(created_inst.get("instance_node"))
-
             new_values = {
                 key: changes[key].new_value
                 for key in changes.changed_keys
             }
-            imprint(
+            self.imprint(
                 instance_node,
                 new_values,
                 update=True
             )
+
+    def imprint(self, node, values, update=False):
+        # Update values to be dynamic
+
+        values["subset"] = '`chs("family")+chs("variant")`'
+        values["variant"] = '`opname(".")`'
+        values["instance_node"] = '`opfullpath(".")`'
+        values["instance_id"]  = '`opfullpath(".")`'
+
+        imprint(node, values, update=update)
 
     def remove_instances(self, instances):
         """Remove specified instance from the scene.
--- a/openpype/hosts/houdini/api/plugin.py
+++ b/openpype/hosts/houdini/api/plugin.py
@@ -182,7 +182,7 @@ class HoudiniCreator(NewCreator, HoudiniCreatorBase):
                 node_type = "geometry"
 
             instance_node = self.create_instance_node(
-                instance_data["variant"], "/out", node_type)
+                subset_name, "/out", node_type)
 
             self.customize_node_look(instance_node)
 
@@ -254,8 +254,8 @@ class HoudiniCreator(NewCreator, HoudiniCreatorBase):
     def imprint(self, node, values, update=False):
         # Update values to be dynamic
 
-        values["subset"] = '`chs("family")+chs("variant")`'
-        values["variant"] = '`opname(".")`'
+        values["subset"] = '`opname(".")`'
+        values["variant"] = '`strreplace(chs("subset"), chs("family"), "")`'
         values["instance_node"] = '`opfullpath(".")`'
         values["instance_id"]  = '`opfullpath(".")`'
 
--- a/openpype/hosts/houdini/api/plugin.py
+++ b/openpype/hosts/houdini/api/plugin.py
@@ -257,7 +257,7 @@ class HoudiniCreator(NewCreator, HoudiniCreatorBase):
         values["subset"] = '`opname(".")`'
         values["variant"] = '`strreplace(chs("subset"), chs("family"), "")`'
         values["instance_node"] = '`opfullpath(".")`'
-        values["instance_id"]  = '`opfullpath(".")`'
+        values["instance_id"] = '`opfullpath(".")`'
 
         imprint(node, values, update=update)
 
--- a/openpype/hosts/houdini/plugins/create/create_composite.py
+++ b/openpype/hosts/houdini/plugins/create/create_composite.py
@@ -28,13 +28,13 @@ class CreateCompositeSequence(plugin.HoudiniCreator):
             pre_create_data)  # type: CreatedInstance
 
         instance_node = hou.node(instance.get("instance_node"))
-        filepath = "{}{}".format(
-            hou.text.expandString("$HIP/pyblish/"),
-            "{}.$F4{}".format(subset_name, self.ext)
+        file_path = "{}/pyblish/`chs('subset')`.$F4.{}".format(
+            hou.text.expandString("$HIP"),
+            self.ext
         )
         parms = {
             "trange": 1,
-            "copoutput": filepath
+            "copoutput": file_path
         }
 
         if self.selected_nodes:
--- a/openpype/hosts/houdini/plugins/create/create_alembic_camera.py
+++ b/openpype/hosts/houdini/plugins/create/create_alembic_camera.py
@@ -26,9 +26,11 @@ class CreateAlembicCamera(plugin.HoudiniCreator):
             pre_create_data)  # type: CreatedInstance
 
         instance_node = hou.node(instance.get("instance_node"))
+        file_path = "{}/pyblish/`chs('subset')`.abc".format(
+            hou.text.expandString("$HIP")
+        )
         parms = {
-            "filename": hou.text.expandString(
-                "$HIP/pyblish/{}.abc".format(subset_name)),
+            "filename": file_path,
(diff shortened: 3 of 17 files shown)


BigRoy:
> * subset name = `opname(".")`
> 
> * variant = `strreplace(chs("subset"), chs("family"), "")`
> 
> * instance_node = `opfullpath(".")`
> 
> * instance_id = `opfullpath(".")`
> 
> * use `chs('subset')`

Let me just reply to these choices. 

Skip `instance_node` and `instance_id` - these should not be 'imprinted' as attributes on the nodes as they represent the node itself. Which this PR solves: https://github.com/ynput/OpenPype/pull/5490 by removing them from as parms on the node.

Then `subset_name` and `variant`. It's up to the template system of OpenPype and the actual creator to define the subset name based on the variant - by default it includes the family, but **it may not!** which means we cannot rely on the variant always being the stripped subset name, etc. Because it's not always just `subset = {family}{Variant}` but can be different, e.g. a render usually includes the Task as well `{family}{Task}{Variant}` - based on templates.

Additionally, using the node name as the 'subset name' is dangerous. In practice the artist should **not** have full control over how to name the subset - it's based on the variant and the creator (which the creator + templates default to "{family}{Variant}`). However if the node name is to represent the subset name, what happens if an artist renames the node "hello" - you're suddenly working with a subset name `hello` that doesn't adhere to these rules.

I also want to remark that storing the **family** on the node itself to me personally makes no sense whatsoever. It's the Creator that defines what family it produces, so if the node already has the creator identifier technically we can deduce what family it should be. Having this **always** defined by the creator and not saved into the node will also allow us to maintain the code easier since we don't need to ensure whatever data is on the node matches what the Creator has as family (or families; which has been an occurrence with bgeo + alembic in Houdini; but also with USD families in Maya which I solved exactly like that; the **Creator** always defines the families and is not embedded onto the container, see this [PR](https://github.com/ynput/OpenPype/pull/5490))

Then the output paths for the intermediate/staged paths, using `chs("subset")` - go for it! That makes total sense.

In short, the first step is removing ALL data from the node that isn't actually defined by what is stored in attributes:

- Instance ID: Remove, it's just the node's path -> solved by https://github.com/ynput/OpenPype/pull/5490
- Instance node: Remove, it's just the node's path -> solved by https://github.com/ynput/OpenPype/pull/5490
- Family: Remove, the creator identifier links it to a creator which is what defines the family - can be solved with similar approach as https://github.com/ynput/OpenPype/pull/5490

- Subset: Technically we can remove this, since it's always computed from creator + variant **but** in this case having subset defined is somewhat of a safety mechanism. It's basically saying: this was the last computation of the subset which we want to keep no matter what; unless you change the variant again. Just so that if someone _were_ to change the template for subset names that older publishes still update into the legacy names and don't start creating new subset names midway in production.

---

More importantly I think this definitly tries to solve some concerns or remarks that people have had on the Houdini integration but as far as I understand are not solving it the way people like @fabiaserra requested. All of these would already be solved by a custom output HDA which doesn't even show these 'raw text fields' but e.g. has a button for "Set asset" which shows a tiny asset selector to update the asset, or a text field "Variant" when edited that will automatically compute the new "Subset" name on changes. Similarly - yes we could use `chs("subset")` in the path name but since the HDA would take full control over what attributes the user changes the output name can be arbitrarily updated as well if needed. Nonetheless, I would even move the output paths for the saver to the template system as well so that settings just define a default like:

```
rop_output_template: $HIP/pyblish/`chs("subset")`{ext}
```
And using the template give the rest of the control to the creator to manage the path. The custom HDA would just on attribute changes ask the Creator: "variant has changed, what should I update"? Which in practice is just updating the instance data as far as I know and triggering a save changes on the CreateContext. And there we have it, a trivial implementation of embedding the Creator/publish logic into an HDA.

openpype/hosts/houdini/api/plugin.py
```diff
@@ -222,25 +223,44 @@ def collect_instances(self):
...
                 update=True
             )
 
+    def imprint(self, node, values, update=False):
+        # Update values to be dynamic
+
+        values["subset"] = '`opname(".")`'
+        values["variant"] = '`strreplace(chs("subset"), chs("family"), "")`'
+        values["instance_node"] = '`opfullpath(".")`'
+        values["instance_id"] = '`opfullpath(".")`'
```
BigRoy, line 260:
```suggestion
        values.pop("instance_node")
        values.pop("instance_id")
        values.pop("family")
```

openpype/hosts/houdini/api/plugin.py
```diff
@@ -187,13 +187,14 @@ def create(self, subset_name, instance_data, pre_create_data):
             self.customize_node_look(instance_node)
 
             instance_data["instance_node"] = instance_node.path()
+            instance_data["instance_id"] = instance_node.path()
```
BigRoy, line 190:
```suggestion
            instance_data["instance_id"] = instance_node.path()
            instance_data["family"] = self.family
```

I'd even argue that the Creator should also always be capable of defining `families` to like in the Maya USD PR I linked where the creator could provide families `["usd", "mayaUsd"]`  or alike to create a product type **usd**, also pass through any USD validators, but have a dedicated **mayaUsd** extractor for that Creator.

By allowing the Creator to define these instead of a post collector adding in `usd` for example also allows Attribute Definitions on plugin on the different families to show for that Creator because before publishing it then knows it has those families.

This could e.g. then also allow a Render Creator itself, to aside of just `render` family add `render.{renderer}` family like `render.arnold` to on the publish tab e.g. show Arnold specific attributes only if the render instance is set to render with arnold ❤️ 

However as far as I know (@iLLiCiTiT can probably confirm) the new publisher isn't capable yet of producing different attribute definitions per instance but does so PER creator currently (even though it does call the `get_attribute_defs()` method per instance - I recall hearing before that's still a bug to be fixed.

openpype/hosts/houdini/api/plugin.py
```diff
@@ -222,25 +223,44 @@ def collect_instances(self):
...
                 "houdini_cached_subsets"].get(self.identifier, []):
+
+            node_data = read(instance)
+
+            # Node paths are always the full node path since that is unique
+            # Because it's the node's path it's not written into attributes
+            # but explicitly collected
+            node_path = instance.path()
+            node_data["instance_id"] = node_path
+            node_data["instance_node"] = node_path
```
BigRoy, line 234:
```suggestion
            node_data["instance_node"] = node_path
            node_data["family"] = self.family
```

openpype/hosts/houdini/api/plugin.py
```diff
@@ -222,25 +223,44 @@ def collect_instances(self):
...
                 instance_node,
                 new_values,
                 update=True
             )
 
+    def imprint(self, node, values, update=False):
+        # Update values to be dynamic
+
+        values["subset"] = '`opname(".")`'
+        values["variant"] = '`strreplace(chs("subset"), chs("family"), "")`'
```
BigRoy, line 258: As explained [here](https://github.com/ynput/OpenPype/pull/5666#issuecomment-1741710742) this isn't a correct approach unfortunately.

openpype/hosts/houdini/plugins/create/create_composite.py
```diff
@@ -28,13 +28,13 @@ def create(self, subset_name, instance_data, pre_create_data):
             pre_create_data)  # type: CreatedInstance
 
         instance_node = hou.node(instance.get("instance_node"))
-        filepath = "{}{}".format(
-            hou.text.expandString("$HIP/pyblish/"),
-            "{}.$F4{}".format(subset_name, self.ext)
+        file_path = "{}/pyblish/`chs('subset')`.$F4.{}".format(
+            hou.text.expandString("$HIP"),
+            self.ext
```
BigRoy, line 33: As explained [here](https://github.com/ynput/OpenPype/pull/5666#issuecomment-1741710742) we're likely better of leaving these to the template system to decide where files should go.

And better yet is that a custom HDA could also then allow to add 'versioning' to the staging outputs as @fabiaserra and others had requested. The HDA could have a "Staging version" integer attribute which could be used, etc.

BigRoy (commented):
Just commenting code changes that'd somewhat align with what I explained [here](https://github.com/ynput/OpenPype/pull/5666#issuecomment-1741710742). Still an approach with a dedicated HDA as requested by many might be the better long term refactor we'd be looking for. Nonetheless some of things I presented could be intermediate changes that help to move there or simplify what HDA we might need to make.

openpype/hosts/houdini/plugins/create/create_composite.py
MustafaJafar: I think we would need to add Houdini vars to template data, 
so that we can have something like this ? 

image

MustafaJafar:
I needed some time to digest your comments. 
Here's a quick recap: 
Let me know if I got you right. 
1. There's no dependency between `node name` and  (`subset name`, `variant name`)
2. we should add an issue about deducing families from creator identifier in Houdini
    - Should that happen while creation like your PR #5616 ? 
3. we should add an issue about using templates for file output 
    - I imagine these templates should exist in `Houdini > General` settings 
    Here's a screenshot for visualization
    image

BigRoy:
> There's no dependency between node name and (subset name, variant name)

There could be - but it makes no sense since:

1. The user can't fully control the subset name, they can choose the creator + variant name. Then the pipeline defines the subset name based on the template system.
2. Using the node name has issue with "uniqueness" it's totally fine to publish a variant `Main` and `Main` for two different nodes, e.g. the `pointcacheMain` and `vdbcacheMain`. The same issue exists for subsets as well actually, it's totally fine for 2x `pointcacheMain` and `pointcacheMain` subsets to be published from a single scene, for example when those are publishing into different assets. We do that a lot - single workfile, publish into multiple assets.

> we should add an issue about deducing families from creator identifier in Houdini 

This should happen on create AND on collect_instances, since that's when the Creator creates the instances for the CreateContext. Additionally this data should **not** be stored on the instance node in the scene (it should not be stored in parms on the node, it should not be serialized because the Creator defines the families).

One could argue it's actually `transient_data` but the publish system needs it in `instance.data["families"]` due to how pyblish works.

> we should add an issue about using templates for file output
> 
>    **I imagine these templates should exist in Houdini > General settings**
>    Here's a screenshot for visualization

Potentially - I initially imagined each Creator just had its own entry so studios could go crazy configuring it per creator if they wanted.

So you'd have:
`CreateVDB/output_path` and `CreatePointcache/output_path` etc.

MustafaJafar:
Thanks a lot, I have added two issues. 
https://github.com/ynput/OpenPype/issues/5712
https://github.com/ynput/OpenPype/issues/5714

Regarding the subset name, This validator I made for static mesh FBX, from my testing I found out it works well for computing the subset name and it doesn't depend on the instance name. 
https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/houdini/plugins/publish/validate_subset_name.py

PR #5668, opened 2023-10-01
Publisher: Refactor Report Maker plugin data storage to be a dict by plugin.id

BigRoy:
## Changelog Description

Refactor Report Maker plugin data storage to be a dict by `plugin.id`

Also fixes `_current_plugin_data` type on `__init__`

## Additional info

Avoids the need for _stored plugins_, _plugin data_ and _plugin data with plugin to be stored separately.

Should be faster because `_get_plugin_data_item`  can now be replaced with O1 lookup instead of requiring to iterate all plugin data items.

## Testing notes:

1. Use publisher, reset, validate, publish, view the different reports, use actions.

**Preferably test in different DCCs with different Python versions.**

### DIFF
--- a/openpype/tools/publisher/control.py
+++ b/openpype/tools/publisher/control.py
@@ -176,11 +176,10 @@ class PublishReportMaker:
         self._create_discover_result = None
         self._convert_discover_result = None
         self._publish_discover_result = None
-        self._plugin_data = []
-        self._plugin_data_with_plugin = []
 
-        self._stored_plugins = set()
-        self._current_plugin_data = []
+        self._plugin_data_by_plugin_id = {}
+        self._current_plugin = None
+        self._current_plugin_data = {}
         self._all_instances_by_id = {}
         self._current_context = None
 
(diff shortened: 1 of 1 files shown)


openpype/tools/publisher/control.py
```diff
@@ -176,11 +176,10 @@ def __init__(self, controller):
...
         self._convert_discover_result = None
         self._publish_discover_result = None
-        self._plugin_data = []
-        self._plugin_data_with_plugin = []
 
-        self._stored_plugins = set()
-        self._current_plugin_data = []
+        self._plugin_data_by_plugin_id = {}
+        self._current_plugin = None
+        self._current_plugin_data = {}
```
BigRoy, line 182: This is still there solely to be 'faster' lookup?

Note that basically whenever `self._current_plugin is not None` then `self._current_plugin_data` should always be the same as: `self._plugin_data_by_plugin_id[self._current_plugin.id]`.

openpype/tools/publisher/control.py
```diff
@@ -176,11 +176,10 @@ def __init__(self, controller):
         self._create_discover_result = None
         self._convert_discover_result = None
         self._publish_discover_result = None
-        self._plugin_data = []
-        self._plugin_data_with_plugin = []
 
-        self._stored_plugins = set()
-        self._current_plugin_data = []
+        self._plugin_data_by_plugin_id = {}
```
BigRoy, line 180: This could also be shortened to this if that's preferred - let me know.
```suggestion
        self._plugin_data_by_id = {}
```

iLLiCiTiT (approved):
Tested, looks ok. I would use (your) suggestion to rename `_plugin_data_by_plugin_id` to `_plugin_data_by_id`.

PR #5669, opened 2023-10-01
Chore: Maya reduce get project settings calls

BigRoy:
## Changelog Description

Re-use system settings / project settings where we can instead of requerying.

### Additional Info

Note: Moved the Muster-specific submitter for maya into the muster module

## Testing notes:

1. Maya should launch ok
2. Maya menu should install ok
3. Muster submissions should still work from maya
4. Custom naming for import and reference loaders should this be configurable in `project_settings/maya/load/import_loader` and `project_settings/maya/load/reference_loader` and should still work as intended.

### DIFF
--- a/openpype/modules/muster/plugins/publish/submit_maya_muster.py
+++ b/openpype/modules/muster/plugins/publish/submit_maya_muster.py
@@ -25,6 +25,7 @@ def _get_template_id(renderer):
     :rtype: int
     """
 
+    # TODO: Use setings from context?
     templates = get_system_settings()["modules"]["muster"]["templates_mapping"]
     if not templates:
         raise RuntimeError(("Muster template mapping missing in "
--- a/openpype/modules/muster/plugins/publish/submit_maya_muster.py
+++ b/openpype/modules/muster/plugins/publish/submit_maya_muster.py
@@ -25,7 +25,7 @@ def _get_template_id(renderer):
     :rtype: int
     """
 
-    # TODO: Use setings from context?
+    # TODO: Use settings from context?
     templates = get_system_settings()["modules"]["muster"]["templates_mapping"]
     if not templates:
         raise RuntimeError(("Muster template mapping missing in "
--- a/openpype/hosts/maya/api/menu.py
+++ b/openpype/hosts/maya/api/menu.py
@@ -1,14 +1,13 @@
 import os
 import logging
+from functools import partial
 
 from qtpy import QtWidgets, QtGui
 
 import maya.utils
 import maya.cmds as cmds
 
-from openpype.settings import get_project_settings
 from openpype.pipeline import (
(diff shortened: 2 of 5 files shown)


openpype/modules/muster/plugins/publish/submit_maya_muster.py
```diff
@@ -25,6 +25,7 @@ def _get_template_id(renderer):
     :rtype: int
     """
 
+    # TODO: Use setings from context?
```
BigRoy, line 28:
```suggestion
    # TODO: Use settings from context?
```

openpype/modules/muster/plugins/publish/submit_maya_muster.py
```diff
@@ -25,6 +25,7 @@ def _get_template_id(renderer):
     :rtype: int
     """
 
+    # TODO: Use settings from context?
     templates = get_system_settings()["modules"]["muster"]["templates_mapping"]
```
BigRoy, line 29: From the `InstancePlugin` what would be the best way to access these settings?

However, aside of that - it even seems the function is unused? It's only used in the [plugin's `_resolve_template()` method](https://github.com/ynput/OpenPype/blob/61f7a2039b60567416d80005c046aab7a5e28de2/openpype/modules/muster/plugins/publish/submit_maya_muster.py#L209C9-L209C26) which doesn't get used within the file.

tokejepsen:
Tested in Maya 2023 successfully for 1, 2, and 4.

I dont have Muster so that is left to test. Maybe @LiborBatek can handle that?

tokejepsen (approved):

openpype/modules/muster/plugins/publish/submit_maya_muster.py
iLLiCiTiT: > From the InstancePlugin what would be the best way to access these settings?

```
instance.context.data["system_settings"]
```

> However, aside of that - it even seems the function is unused? It's only used in the [plugin's _resolve_template() method](https://github.com/ynput/OpenPype/blob/61f7a2039b60567416d80005c046aab7a5e28de2/openpype/modules/muster/plugins/publish/submit_maya_muster.py#L209C9-L209C26) which doesn't get used within the file.

Maybe remove it in that case? I would be surprised if Muster would work anyway.

BigRoy: Remove the unused functions, remove the todo or remove the change?

LiborBatek:
So it seems that Muster is not at the target....no need to test it then (according to the devs).

LiborBatek (approved):
Muster is not in the scope of Ayon. So doesnt matter now.

antirotor (approved):

PR #5670, opened 2023-10-02
Unreal: Changed behaviour for updating assets

simonebarbieri:
## Changelog Description
Changed how assets are updated in Unreal.

## Additional info
When updating assets in Unreal, instead of completely replacing the old version, the new version is loaded in a different folder (it is checked beforehand if it was already loaded), and replaces the asset in the current level. Then, we check if the old asset is used in any other level. If it isn't we remove the previous version.

Updated:
- Static Meshes (FBX and ABC).
- Skeletal Meshes (FBX and ABC).
- Geometry Caches.

## Testing notes:
1. Load an asset.
2. Add it to a level and save it.
3. Try changing the version of the asset.

### DIFF
--- a/openpype/hosts/unreal/plugins/load/load_geometrycache_abc.py
+++ b/openpype/hosts/unreal/plugins/load/load_geometrycache_abc.py
@@ -147,7 +147,6 @@ class PointCacheAlembicLoader(plugin.Loader):
         if not unreal.EditorAssetLibrary.does_directory_exist(asset_dir):
             path = self.filepath_from_context(context)
 
-
             self.import_and_containerize(
                 path, asset_dir, asset_name, container_name,
                 frame_start, frame_end)
--- a/openpype/hosts/unreal/plugins/load/load_geometrycache_abc.py
+++ b/openpype/hosts/unreal/plugins/load/load_geometrycache_abc.py
@@ -7,7 +7,12 @@ from openpype.pipeline import (
     AYON_CONTAINER_ID
 )
 from openpype.hosts.unreal.api import plugin
-from openpype.hosts.unreal.api import pipeline as unreal_pipeline
+from openpype.hosts.unreal.api.pipeline import (
+    create_container,
+    imprint,
+    replace_geometry_cache_actors,
+    delete_previous_asset_if_unused,
+)
 
 import unreal  # noqa
 
@@ -21,8 +26,11 @@ class PointCacheAlembicLoader(plugin.Loader):
     icon = "cube"
     color = "orange"
 
+    root = "/Game/Ayon/Assets"
+
+    @staticmethod
     def get_task(
-        self, filename, asset_dir, asset_name, replace,
+        filename, asset_dir, asset_name, replace,
         frame_start=None, frame_end=None
     ):
         task = unreal.AssetImportTask()
@@ -38,8 +46,6 @@ class PointCacheAlembicLoader(plugin.Loader):
         task.set_editor_property('automated', True)
         task.set_editor_property('save', True)
 
-        # set import options here
-        # Unreal 4.24 ignores the settings. It works with Unreal 4.26
         options.set_editor_property(
             'import_type', unreal.AlembicImportType.GEOMETRY_CACHE)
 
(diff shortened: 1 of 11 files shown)


LiborBatek (approved):
Did tests in ```UE 5.1``` and also in ```UE 5.3``` and all works as supposed to.

Did not find any issues.

openpype/hosts/unreal/plugins/load/load_geometrycache_abc.py
```diff
@@ -21,8 +26,11 @@ class PointCacheAlembicLoader(plugin.Loader):
     icon = "cube"
     color = "orange"
 
+    root = "/Game/Ayon/Assets"
```
antirotor, line 29: I would yank this out to some constant living in the unreal host code.

antirotor (changes requested):
I didn't go through the code in detail, but I've tested it in UE5.3 and it worked. Just to recap (and please correct me if I am wrong):

- when updating version that is not used as an actor in any level, the version is replaced
- when updating verison of something that is used,  existing version is kept and new one is loaded

If this is so, I'd like to propose one addition to this PR:

Lets create Inventory `Update in Current Level` plugin that will run on selected item and replace actors in current level too.

openpype/hosts/unreal/plugins/load/load_geometrycache_abc.py
simonebarbieri: Maybe it would be even better to make a setting to make the studio choose where to import the assets. What do you think?

simonebarbieri:
@antirotor it already updates the assets in the current level! And it checks if it is still used after this update. Do you think it would be better to add the plugin anyway to give the choice?

antirotor:
> @antirotor it already updates the assets in the current level! And it checks if it is still used after this update. Do you think it would be better to add the plugin anyway to give the choice?

I think it must be obvious what it is updating - we shouldn't mix things together. Artist should know that when he's updating, it is updating assets loaded in Unreal but not used as actors. And only when requested, it should update actors in levels. But that's just my opinion. Any thoughts about it @LiborBatek ?

antirotor:
This is probably related to the new Scene Manager for UE that needs to be done anyway, but we need some reasonable functionality until then.

LiborBatek:
@antirotor yes I agree, the asset updating and actor updating should be split and be separate action so user is in control of all his levels and Ayon not forcing anything without users will.

I think right now it behaves like that...only updating asset in the content browser and not touching any actors present in the UE level. @simonebarbieri 

At least that was happening when I have tested this PR if I remember correctly.

simonebarbieri:
> I think it must be obvious what it is updating - we shouldn't mix things together. Artist should know that when he's updating, it is updating assets loaded in Unreal but not used as actors. And only when requested, it should update actors in levels.

@antirotor Yeah I agree! I'll change it and make a new inventory plugin.

> I think right now it behaves like that...only updating asset in the content browser and not touching any actors present in the UE level.

@LiborBatek no, it also updates the actors in the opened level, if there is any!

simonebarbieri:
@antirotor now updating an asset doesn't update any actor! And I've added an Inventory plugin to update all the actors of the same subset to the selected version!

LiborBatek (commented):
The ```Manage``` and its ```Update to latest version``` brings new uasset version into UE project and do not touch any already present ```actors``` in the level(s) which is right and works fine.

Also ```Action``` in the Manage actions menu ```Update actors in level to this version``` works fine. 

Things to consider:

Question being if it also should remove preceding asset version from content browser aka UE Project or keep it intact. Imho safer would be to keep any ```uasset``` older versions still in the project. 
There could be possibly Manage Action dedicated for removing unused uasset in the current project as separate tool? Something like:
Manage>Actions> ```Remove unused assets from the project```

One more thing to think about is, if there shouldnt be way to trigger ```Update actors in level to this version``` on only selected actors...now this happening globally accros the active UE level for all actors matching the selected asset for replacing, so user have no full control which actors get replaced by selected asset version which is not always desirable and need users decision.

LiborBatek:
I also propose to change name of the Action from ```Update actors in level to this version``` to something like
```Replace actors by current version```
@antirotor whats your opinion on that?
image

simonebarbieri:
> Question being if it also should remove preceding asset version from content browser aka UE Project or keep it intact.

Yeah that's a good question. For now it deletes them if they're unused, but with big assets it could be bothersome if you want to use them in the future and you need to reimport them...
 
> One more thing to think about is, if there shouldnt be way to trigger `Update actors in level to this version` on only selected actors...now this happening globally accros the active UE level for all actors matching the selected asset for replacing, so user have no full control which actors get replaced by selected asset version which is not always desirable and need users decision.

This could be doable, I just need to change what assets I'm checking. Maybe split it in two options? "Update all" and "Update selected?"

antirotor:
> This could be doable, I just need to change what assets I'm checking. Maybe split it in two options? "Update all" and "Update selected?"

Yep, two options are the way! :)

simonebarbieri:
@antirotor @LiborBatek 
Now there are two options:
- "Replace all Actors in level to this version"
- "Replace selected Actors in level to this version".

Also, now unused assets are not deleted automatically, but there is a new action to do that (with a dialog to confirm, just to avoid any mistake)

LiborBatek (approved):
Works like charm! 

I did tested various scenarios like ```Replace Selection by current version``` and ```Replace All actors...``` on single and multiple selected ```actors``` and all working perfectly!

image

Loving this feature...big improvement for UE integration!

openpype/hosts/unreal/plugins/load/load_geometrycache_abc.py
simonebarbieri: Changed to a constant for now!

Issue #4361, opened 2023-01-21
Implement a format representation sequence path function for Loaders


BigRoy:
**Is your feature request related to a problem? Please describe.**

Many loader plug-ins involved quite some code on finding the right file or file pattern for a sequence. They are trying to construct patterns like:

```
/path/to/publish/v001/filename.####.exr
/path/to/publish/v001/filename.$F4.exr
/path/to/publish/v001/filename.%04d.exr
/path/to/publish/v001/filename.<f>.exr
/path/to/publish/v001/filename.<udim>.exr
/path/to/publish/v001/filename.<udim>.<f>.exr
```

These currently all appear to be implemented over and over again, and also appear to parse this data from the filename itself even though most of these frames or udims or alike is information that was available at the time of publish.

Some examples
- [Replacing part of the filename in the Nuke image loader](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/nuke/plugins/load/load_image.py#L191-L196) - this would already fail if the subset was called imageMain20 and we're loading frame 20, it would also replace the `20` of the subset name to 
- [Houdini image loader, listing files in the publish folder and going from there](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/houdini/plugins/load/load_image.py#L121-L127)
- [Houdini bgeo loader loading a sequence and always expecting the `dot` in front of the frame number even though I believe that's customizable to be another frame separator in OpenPype](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/houdini/plugins/load/load_bgeo.py#L65-L83) and [something similar for Houdini's VDB Loader](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/houdini/plugins/load/load_vdb.py#L61-L79)
- [Fusion image sequence loader listing folder's contents](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/fusion/plugins/load/load_sequence.py#L252-L255)
- [Maya ass loader listing publish contents using `os.listdir` and then `clique.assemble`](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/maya/plugins/load/load_ass.py#L255-L259)
- [Maya VDB loader to V-Ray trying to parse path from listing folder](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/maya/plugins/load/load_vdb_to_vray.py#L170-L203)
[...]

PR #5673, opened 2023-10-02
Resolve: inventory management fix

jakubjezek001:
## Changelog Description
Loaded Timeline item containers are now updating correctly and version management is working as it suppose to.

- [x] updating loaded timeline items
- [x] Removing of loaded timeline items

> Warning! For versions under 18.5 Beta 3
> The removal of loaded containers has reached the limits of Resolve API (v18.0.4). We can only remove the linked mediapool item, not the timeline item. The removed containers will leave the offline timeline item intact, requiring manual removal.

## Testing notes:
1. Load multiple products from couple different folder contexts. Make sure multiple versions of each are available and load some older versions.
2. Open `Manager ...` from OpenPype menu and update all versions to latest
image


openpype/hosts/resolve/api/lib.py
```diff
@@ -125,15 +125,16 @@ def get_any_timeline():
         return project.GetTimelineByIndex(1)
 
 
-def get_new_timeline():
+def get_new_timeline(timeline_name: str = None):
     """Get new timeline object.
```
BigRoy, line 130:
```suggestion

Arguments:
    timeline_name (str): New timeline name.

```

openpype/hosts/resolve/api/plugin.py
```diff
@@ -291,17 +291,17 @@ class ClipLoader:
...
         Arguments:
-            cls (openpype.pipeline.load.LoaderPlugin): plugin object
+            loader_obj (openpype.pipeline.load.LoaderPlugin): plugin object
             context (dict): loader plugin context
             options (dict)[optional]: possible keys:
                 projectBinPath: "path/to/binItem"
 
         """
-        self.__dict__.update(cls.__dict__)
+        self.__dict__.update(loader_obj.__dict__)
```
BigRoy, line 304: What's the reason we need to do this? Instead of dedicated passing the actual arguments that we expect to exist? It's quite hard to grasp what data we're now expecting on the loader obj?

openpype/hosts/resolve/api/plugin.py
```diff
@@ -349,10 +355,14 @@ def _populate_data(self):
...
-        representation = str(repr_cntx["representation"])
-        self.data["clip_name"] = "_".join([asset, subset, representation])
+        self.asset = str(repr_cntx["asset"])
+        self.subset = str(repr_cntx["subset"])
+        self.representation = str(repr_cntx["representation"])
+        self.data["clip_name"] = "_".join([
+            self.asset,
+            self.subset,
+            self.representation
+        ])
```
BigRoy, line 365: Is there a reason these need to be stored on `self`? It's making it look like it's used more elsewhere but it's hard to pinpoint exactly where.

openpype/hosts/resolve/api/plugin.py
```diff
@@ -319,23 +319,29 @@ def __init__(self, cls, context, path, **options):
...
+                self.active_timeline = lib.get_new_timeline(
+                    "{}_{}_{}".format(
+                        self.subset,
+                        self.representation,
+                        str(uuid.uuid4())[:8]
+                    )
                 )
+                loader_cls.timeline = self.active_timeline
+
+            print(self.active_timeline.GetName())
```
BigRoy, line 340: Debugging?
```suggestion
            print(self.active_timeline.GetName())
```

BigRoy:
Updating and setting versions works for me in Resolve Studio 18.5 with Python 3.9.13

However I did hit these issues:

#### Load just the one representation instead of all

When loading just the EXR it seems to load all the representations into the project - is that necessary?

image

#### Changing versions seems to work, but now has both old + new loaded into the project?

When changing versions it now has both loaded into the project's media pool?

image

#### Broken: Manager > Remove does nothing

I can't seem to remove loaded content through the manager - nothing happens?

#### Would it be possible to tag the media pool item instead of the timeline entry?

Probably outside of scope of this PR

I was wondering whether it'd be possible to tag the media pool item instead of the timeline entries so that if a user were to drag content from the media pool that was loaded with the loader that it'd also still detect that clip as an managed entry?

openpype/hosts/resolve/api/plugin.py
jakubjezek001: Agree that this needs some refactor. This is old code of mine created 3 years ago with the knowledge of those days 🥲 

But since it is not the themem of this PR lets leave it as it is and move on. This refactor should be done separatelly.

openpype/hosts/resolve/api/plugin.py
jakubjezek001: yeah, this is quite dirty ;)

jakubjezek001:
> Load just the one representation instead of all

Those are limitations of Resolve API. Perhaps we could add the clips to mediapool items, then use them in timeline and if the rest is not used then remove them. But again this is not the theme of this PR. And this conversation will be lost in history of this PR.

jakubjezek001:
> Broken: Manager > Remove does nothing
> I can't seem to remove loaded content through the manager - nothing happens?

Will fix this in this PR. Thanks a lot for the testing!

BigRoy:
> > Load just the one representation instead of all
> 
> Those are limitations of Resolve API. Perhaps we could add the clips to mediapool items, then use them in timeline and if the rest is not used then remove them. But again this is not the theme of this PR. And this conversation will be lost in history of this PR.

It is possible actually  - I believe you're still using what is actually a deprecated API call. I believe you are still using `MediaStorage.AddItemsToMediaPool` but should be using `media_pool.ImportMedia` like [here](https://github.com/ynput/OpenPype/pull/4976/files#diff-bf9f8c0751e9055db60409c3670716a6027c6951c596757fd86bb340e89f26c0R65-R71) - there however you need to explicitly pass it the full list of files, not the filepath.

jakubjezek001:
closes https://github.com/ynput/OpenPype/issues/5683 since it was implemented here [`908e980` (#5673)](https://github.com/ynput/OpenPype/pull/5673/commits/908e980a404bf33dd7657414658cbd801ceb86d0) and 19840862426e74c6558803864ec212014ada186f

openpype/hosts/resolve/api/lib.py
```diff
@@ -202,22 +212,39 @@ def create_media_pool_item(fpath: str,
...
+
+    # get all files in folder
+    for file in glob.glob(os.path.join(dir_path, base_name)):
+        files.append(file)
+
+    # iterate all files and check if they exists
+    # if not then remove them from list
+    for file in files[:]:
+        if not os.path.exists(file):
+            files.remove(file)
```
BigRoy, line 242:
```suggestion
    files = [f for f in files if os.path.exists(f)]
```

BigRoy, line 242: This is a perfect example of a function that we should be abstracting out of here.

A `get_representation_files(representation)` or whatever so that we can accurately get the sequence we need instead of implementing the file logic into this loader itself.

jakubjezek001: Are you talking about pipeline wide abstraction? This is abstracted to `resolve.api.lib`

BigRoy: I did mean that, yes. ❤️ 

But didn't see this was in resolve lib already. In that case - can you just still separate out the function that retrieves the paths themselves, so that `create_media_pool_item()`  just calls that function.

I can imagine there are reasons to retrieve the expected filepaths for a representation (which is what this is doing!) without directly calling `ImportMedia` ;)

openpype/hosts/resolve/plugins/load/load_clip.py
```diff
@@ -168,5 +168,9 @@ def remove(self, container):
         namespace = container['namespace']
         timeline_item = lib.get_pype_timeline_item_by_name(namespace)
         take_mp_item = timeline_item.GetMediaPoolItem()
+        timeline = lib.get_current_timeline()
+
+        if timeline.DeleteClips is not None:
```
BigRoy, line 173:
```suggestion
        # DeleteClips function was added in Resolve 18.5+
        # by checking None we can detect whether the
        # function exists in Resolve
        if timeline.DeleteClips is not None:
```

BigRoy:
Did a quick test-run.

- ✅ Loading works fine.
- 🔴 Updating doesn't work as expected, when updating an **image sequence** the updated version has only the single frame loaded not the full sequence. - however video files (single files) work fine on updating.
- 🔴 When you load the same file twice - and from the manager "remove" only **one** of the two then it will still remove the media from the Media Pool leaving the other loaded clip invalid (no media)
- ⚠️ When updating versions it still leaves the 'old' version dangling in your media pool as well - so now you basically have both there. I'd expect that if after updating the MediaPoolItem has no remaining `Usage` that it'd remove it to. Not super critical.

E.g. in this case I'd have expected the top one to be removed:
image

openpype/hosts/resolve/api/lib.py
BigRoy: Just wanted to cross-reference this issue since it's _somewhat_ related: https://github.com/ynput/OpenPype/issues/4361

BigRoy:
Issues mentioned [here](https://github.com/ynput/OpenPype/pull/5673#issuecomment-1750883828) should be resolved with https://github.com/ynput/OpenPype/pull/5717

openpype/pipeline/load/utils.py
```diff
@@ -286,6 +288,54 @@ def get_representation_context(representation):
...
+    glob_pattern = re.sub(r"\d", "?", str_first_frame)
+
+    # in filename replace number with glob pattern
+    # example: `filename.0001.exr` -> `filename.????.exr`
+    base_name = re.sub(str_first_frame, glob_pattern, base_name)
+
+    files = []
+    # get all files in folder
+    for file in glob.glob(os.path.join(dir_path, base_name)):
+        files.append(file)
```
BigRoy, line 328: Can we add a `todo`here that we should improve the logic over time for safety reasons - it's not super reliable to just assume we can match on any digits?

openpype/pipeline/load/utils.py
```diff
@@ -286,6 +288,54 @@ def get_representation_context(representation):
...
+        filepath (str): Filepath of the representation.
+
+    Returns:
+        list[str]: List of files for representation.
+    """
+    version = context["version"]
+    frame_start = version["data"]["frameStart"]
+    frame_end = version["data"]["frameEnd"]
+    handle_start = version["data"]["handleStart"]
+    handle_end = version["data"]["handleEnd"]
```
BigRoy, line 305: Note: Not all versions have frame range data.
Also, representation can also store frame range data that supersedes the version data, so that should be taken into account.

openpype/pipeline/load/utils.py
```diff
@@ -286,6 +288,54 @@ def get_representation_context(representation):
     return context
 
 
+def get_representation_files(context, filepath):
+    """Return list of files for representation.
+
+    Args:
+        representation (dict): Representation document.
```
BigRoy, line 295: This argument does not exist :)

openpype/pipeline/load/utils.py
```diff
@@ -286,6 +288,54 @@ def get_representation_context(representation):
     return context
 
 
+def get_representation_files(context, filepath):
```
BigRoy, line 291: I feel the logic here is still off. Can't we rely on the data in `representation["files"]` or something alike?

```python
def get_representation_files(representation, anatomy):
    
    context = representation["context"]
    files = []
    for file_data in representation["files"]:
        path = file_data["path"].format(root=anatomy.roots)
        files.append(path)
    return files
```

This would return all files **including** resources (that are not necessarily part of the sequence, e.g. for maya looks I believe this should also include the textures aisde of the maya ascii file). Filtering those might be done by matching the paths against `representation["data"]["template"]` to see if it's a match or not or reversing it and computing all files based on that:
```python
import copy
from openpype.lib import StringTemplate

def get_representation_frame_files(representation, anatomy, frames):
    
    template = representation["data"]["template"]
    template = StringTemplate(template)
    
    context = representation["context"]
    context = copy.deepcopy(context)
    context["root"] = anatomy.roots
    
    files = []
    for frame in frames:
        context["frame"] = frame
        path = template.format(**context)
        files.append(path)
        
    return files
```

BigRoy: Not sure how well the last formatting example works with publishes that use `source` as template because it doesn't include `{frame}` I believe.

We might be able to capture those case to by doing:
```python
if "frame" not in context:
    raise ValueError("Can't resolve frame files for published contexts without frame data")

if "{frame" not in template:
    raise ValueError("Can't resolve frame files for publishes without {frame} key in template")
```

openpype/hosts/resolve/api/lib.py
jakubjezek001: implemented via [`0a71b89` (#5673)](https://github.com/ynput/OpenPype/pull/5673/commits/0a71b89ddd857676d1561c3cdbfeb690ebae6103)

jakubjezek001:
> *🔴 Updating doesn't work as expected, when updating an **image sequence** the updated version has only the single frame loaded not the full sequence. - however video files (single files) work fine on updating.
> * 🔴 When you load the same file twice - and from the manager "remove" only **one** of the two then it will still remove the media from the Media Pool leaving the other loaded clip invalid (no media)

This should be fixed now.

openpype/pipeline/load/utils.py
BigRoy: For those edge cases we might be ok with falling back to a straight up parsing the files as a collection logic? Maybe even using `clique.assemble`?

iLLiCiTiT: I do agree with the comments, but for sake of this PR please keep this function only inside resolve for now. This function definetelly should not be in public functions, at least not with this implementation.

BigRoy: Thanks, good point - unless we can define an obvious 99% always correct function to retrieve this let's keep it Resolve lib only - so that we have some wiggle room to have it 'unstable' to some extent :P

openpype/hosts/resolve/api/lib.py
```diff
@@ -192,48 +203,35 @@ def create_media_pool_item(fpath: str,
         object: resolve.MediaPoolItem
     """
     # get all variables
-    media_storage = get_media_storage()
     media_pool = get_current_project().GetMediaPool()
     root_bin = root or media_pool.GetRootFolder()
 
     # try to search in bin if the clip does not exist
-    existing_mpi = get_media_pool_item(fpath, root_bin)
+    existing_mpi = get_media_pool_item(files[0], root_bin)
```
Minkiu, line 210: Would it be worth checking that it's not an empty list?
Also docstring for this arg is wrong.

openpype/hosts/resolve/api/plugin.py
```diff
@@ -383,25 +381,23 @@ def _get_asset_data(self):
...
+
+        Arguments:
+            files (list): list of files to load into timeline
+        """
         # create project bin for the media to be imported into
         self.active_bin = lib.create_bin(self.data["binPath"])
-
-        # create mediaItem in active project bin
-        # create clip media
+        handle_start = self.data["versionData"].get("handleStart", 0)
```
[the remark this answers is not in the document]
BigRoy: No idea but if that's the case then:
```suggestion
        handle_start = self.data["versionData"].get("handleStart") or 0
```

openpype/hosts/resolve/api/lib.py
jakubjezek001: implemented via 61f381cb5cee14f9f2c85b6db04c9144f9818ac5

openpype/hosts/resolve/api/plugin.py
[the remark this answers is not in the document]
jakubjezek001: implemented via f03be42e9d62882a501303abfbee37b83463c946

BigRoy (approved):

PR #5674, opened 2023-10-02
Maya: Add Maya 2024 and remove pre 2022.

tokejepsen:
## Changelog Description
Adding Maya 2024 as default application variant.

Removing Maya 2020 and older, as these are not supported anymore.

## Testing notes:
1. Launch settings with a new database and validate only Maya 2024-2022 is present.

Split from https://github.com/ynput/OpenPype/pull/5644

### DIFF
--- a/openpype/settings/defaults/system_settings/applications.json
+++ b/openpype/settings/defaults/system_settings/applications.json
@@ -12,6 +12,26 @@
             "LC_ALL": "C"
         },
         "variants": {
+            "2024": {
+                "use_python_2": false,
+                "executables": {
+                    "windows": [
+                        "C:\\Program Files\\Autodesk\\Maya2024\\bin\\maya.exe"
+                    ],
+                    "darwin": [],
+                    "linux": [
+                        "/usr/autodesk/maya2024/bin/maya"
+                    ]
+                },
+                "arguments": {
+                    "windows": [],
+                    "darwin": [],
+                    "linux": []
+                },
+                "environment": {
+                    "MAYA_VERSION": "2024"
+                }
+            },
             "2023": {
                 "use_python_2": false,
                 "executables": {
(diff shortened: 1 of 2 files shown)


BigRoy (approved):
Didn't test but looks good to me ❤️

BigRoy:
Actually - is this missing an update to AYON settings for Maya?

iLLiCiTiT:
Please add AYON settings changes too.

BigRoy (changes requested):
As stated - this should also update AYON defaults.
_Sorry for commenting the same thing - I had to place a comment to change my approval status_

BigRoy (approved):
Didn't test it - but code looks good. Probably best to check against new deployment of Ayon and OpenPype I suppose? Probably for @LiborBatek?

Minkiu (approved):

iLLiCiTiT (approved):
Code changes look ok. Someone has to validate if we should limit the maya versions to this set. @m-u-r-p-h-y ?

antirotor (approved):

antirotor:
> Code changes look ok. Someone has to validate if we should limit the maya versions to this set. @m-u-r-p-h-y ?

I don't think we support Maya 2018 and with Maya 2020 I would be surprised if there are no issues. Anyway, there's no problem to  define them locally if needed

m-u-r-p-h-y:
I agree to have only M2021 and up

if we are touching the defaults I would like to unify the paths for other platforms. We have linux paths defined somewhere but no Mac (not sure what is the default path on MAc though)

something like `/Applications/Autodesk/maya2021/Maya.app/Contents/bin/maya` ?

tokejepsen:
> if we are touching the defaults I would like to unify the paths for other platforms. We have linux paths defined somewhere but no Mac (not sure what is the default path on MAc though)

Could this be a different PR for another time?

mkolar:
I agree we can tweak the paths in another PR. let's merge this, so it's not festering.

PR #5676, opened 2023-10-02
Add openpype_mongo command flag for testing.

tokejepsen:
## Changelog Description
Instead of changing the environment, this command flag allows for changing the database.

## Testing notes:
1. Run test suite with command flag `--openpype_mongo [URI]`.

### DIFF
--- a/openpype/cli.py
+++ b/openpype/cli.py
@@ -290,11 +290,15 @@ def run(script):
               "--setup_only",
               help="Only create dbs, do not run tests",
               default=None)
+@click.option("--openpype_mongo",
+              help="MongoDB for testing.",
+              default=None)
 def runtests(folder, mark, pyargs, test_data_folder, persist, app_variant,
-             timeout, setup_only):
+             timeout, setup_only, openpype_mongo):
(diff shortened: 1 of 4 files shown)


openpype/cli.py
```diff
@@ -290,11 +290,15 @@ def run(script):
               "--setup_only",
               help="Only create dbs, do not run tests",
               default=None)
+@click.option("--openpype_mongo",
```
iLLiCiTiT, line 293: I would say that `mongo_url` would be better name for cli.

kalisp (approved):
Had to revert TEST_OPENPYPE_MONGO as it is used in existing .zip files.
Fixed `openpype_mongo` fixture to be actually used.
Renamed `openpype_mongo` to `mongo_url` according to Kuba's comment.

tokejepsen:
> Had to revert TEST_OPENPYPE_MONGO as it is used in existing .zip files. Fixed `openpype_mongo` fixture to be actually used. Renamed `openpype_mongo` to `mongo_url` according to Kuba's comment.
> 
> Please revisit/test on your side and I could merge this.

Looks good to me. Merge at will!

PR #5678, opened 2023-10-03
Fusion: implement toggle to use Deadline plugin FusionCmd

kalisp:
## Changelog Description
Fusion 17 doesn't work in DL 10.3, but FusionCmd does. It might be probably better option as headless variant.
Fusion plugin seems to be closing and reopening application when worker is running on artist machine, not so with FusionCmd

Added configuration to Project Settings for admin to select appropriate Deadline plugin:
AnyDeskMSI_2Nk3v96H3t

## Testing notes:
1. Publish with any version of Fusion to Deadline


kalisp:
Question is if we should switch to `FusionCmd` or provide option on `submit_fusion_deadline` for artist to choose from (in that case what should be default value?)

server_addon/deadline/server/settings/publish_plugins.py
BigRoy, line 1: I believe the AYON settings are lacking the default value?

openpype/settings/entities/schemas/projects_schema/schema_project_deadline.json
```diff
@@ -289,6 +289,15 @@
                             "type": "text",
                             "key": "group",
                             "label": "Group Name"
+                        },
+                        {
+                            "type": "enum",
+                            "key": "plugin",
+                            "label": "Deadline plugin",
```
BigRoy, line 296: Labels differed between AYON and OpenPype - AYON has uppercase P on Plugin?
```suggestion
                            "label": "Deadline Plugin",
```

BigRoy (approved):
I haven't tested the submissions or the differences between the two Fusion deadline plugins. 

It's interesting that apparently you can just swap out one for the other (both use same parameters?) but if it works it works. But I guess looking at [this documentation page](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/app-fusion.html) it's designed to be used interchangeably.

> - **Command Line Mode:** Render using separate command line calls instead of keeping the scene loaded in memory between tasks. Using this feature disables the High Quality, Proxy, and Check Saver Output options. This uses the **FusionCmd** plug-in, instead of the **Fusion** one.

So I suppose that might explain the difference somewhat. And the [FAQ](https://docs.thinkboxsoftware.com/products/deadline/10.1/1_User%20Manual/manual/app-fusion.html#faq) says more:

> **What’s the difference between the Fusion and FusionCmd plugins?**
> 
> The **Fusion** plugin starts the Fusion Render Node in server mode and uses eyeonscript to communicate with the Fusion renderer. Fusion and the comp remain loaded in memory between tasks to reduce overhead. This is usually the preferred way of rendering with Fusion.
> 
> The **FusionCmd** plugin renders with Fusion by executing command lines, and can be used by selecting the Command Line mode option in the Fusion submitter. Because Fusion needs to be launched for each task, there is some additional overhead when using this plugin. In addition, the Proxy, High Quality, and Saver Output Checking features are not supported in this mode. However, this mode tends to print out better debugging information when there are problems (especially when the Fusion complains that it can’t load the comp), so we recommend using it to help figure out problems that may be occurring when using the Fusion plugin.

BigRoy:
> Fusion plugin seems to be closing and reopening application when worker is running on artist machine,

These tend to hint that the comp might be failing to load a certain plug-in or the comp uses nodes/plugins not available to the machine. The RenderSlave has always had the tendency to just error out whilst saying nothing and bye bye. Might be this issue exactly where Fusion isn't reporting why it's closing down.

m-u-r-p-h-y:
Shouldn't we put `FusionCMD `as the default value so it will work by default?

kalisp:
I think that `FusionCmd` plugin is broken in DL 10.2, because Deadline :/

m-u-r-p-h-y (approved):
works as expected

image

PR #5679, opened 2023-10-03
Blender: Fix remove/update in new layout instance

simonebarbieri:
## Changelog Description
Fixes an error that occurs when removing or updating an asset in a new layout instance.

## Additional info
When updating or removing the assets, it looks for `members` in the parent container, which doesn't exist (it exists in loaded layouts, not in instances).

## Testing notes:
1. Create a new layout instance in Blender.
2. Try updating or removing one of the assets in the layout.


LiborBatek (approved):
When removing asset from existing ```Layout``` Instance it correctly publishes without raising any validation erros.

Also tested when loaded into scene as out of date version and later updating to the latest version and all working normally.

Issue #5681, opened 2023-10-03
Bug: Ayon Launcher Refresh btn doesnt work


LiborBatek:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

When using Launcher and in need of refreshing Projects assets etc when hitting ```Refresh``` btn icon on the top right nothing happens and no newly added assets/tasks appear.

### Expected Behavior:

When using ```Refresh``` btn all assets in the active Project should get refreshed and showing latest state of the active Project on the Ayon server.

### Version

3.17.2-nightly.1

### What platform you are running OpenPype on?

Windows

### Steps To Reproduce:

1. Start Ayon Launcher
2. Pick some Project
3. Add some asset on the Ayon Server and Save the changes
4. Hit Refresh btn in the Launcher app
5. Nothing changes
6. Step out from the active selected Project in the Launcher UI
7. Get back and check newly added assets appearing

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

PR #5685, opened 2023-10-03
AYON Launcher tool: Fix refresh btn

iLLiCiTiT:
## Changelog Description
Refresh button does propagate refreshed content properly. Folders and tasks are cached for 60 seconds instead of 10 seconds. Auto-refresh in launcher will refresh only actions and related data which is project and project settings.

## Additional info
Refresh button did trigger refresh in controller but if project/asset/task did not change, the other widgets did not update their data.

## Testing notes:
1. Open launcher tools
2. Go to a project in launcher tool
3. Open the same project in AYON web UI
4. Go to Edit tab and add few folders/tasks
5. Go back to launcher tool
6. It should update new folders/tasks when you click on refresh
NOTE: There is auto-refresh in launcher each 10 seconds so you may not need to click on refresh button at all.

Resolves https://github.com/ynput/OpenPype/issues/5681

### DIFF
--- a/openpype/tools/ayon_launcher/abstract.py
+++ b/openpype/tools/ayon_launcher/abstract.py
@@ -295,3 +295,13 @@ class AbstractLauncherFrontEnd(AbstractLauncherCommon):
         """
 
         pass
+
+    @abstractmethod
+    def refresh_actions(self):
+        """Refresh actions and all related data.
+
+        Triggers 'controller.refresh.actions.started' event at the beginning and
+        'controller.refresh.actions.finished' at the end.
+        """
+
+        pass
--- a/openpype/tools/ayon_launcher/abstract.py
+++ b/openpype/tools/ayon_launcher/abstract.py
@@ -300,8 +300,8 @@ class AbstractLauncherFrontEnd(AbstractLauncherCommon):
     def refresh_actions(self):
         """Refresh actions and all related data.
 
-        Triggers 'controller.refresh.actions.started' event at the beginning and
-        'controller.refresh.actions.finished' at the end.
+        Triggers 'controller.refresh.actions.started' event at the beginning
+        and 'controller.refresh.actions.finished' at the end.
         """
 
         pass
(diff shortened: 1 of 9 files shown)


LiborBatek (commented):
Works as declared in testing steps.

However the Refresh is so instant theres actually no need to trigger ```Refresh``` btn at all (also as mentioned in testing steps).

However it crashes ```Launcher``` a lot. It just dissapear quite often without any trace (shut down).

I have also tried to run it with console but no chance to log anything....it is just gone.

I have only uploaded package for OP and set it as ```Production``` no any other addons.

There is also weird thing in the Launcher that when switching ```assets``` with preselected ```task``` that it looses its focus on newly created assets/tasks. It seems this happening only on newly created items and recent time (with time it gets fixed?) Check my video show that behavior...

https://github.com/ynput/OpenPype/assets/112623825/e938958b-f50f-4651-b288-0dec4b8c13ba

iLLiCiTiT:
> However it crashes Launcher a lot. It just dissapear quite often without any trace (shut down).

I've tried to fix it, please validate if still happens.

I'll do few more changes regarding auto-refresh, so the assets and tasks are not refreshed automatically. It causes troubles with bigger projects.

> There is also weird thing in the Launcher that when switching assets with preselected task that it looses its focus on newly created assets/tasks

I will skip this issue in this PR.

iLLiCiTiT:
> Seems still not working...refresh has no effect. But its not crashing this time!

Please make sure you've replaced the code in localappdata. I don't have the issue on my side.

LiborBatek:
Did another test round adding ```sh040``` into ```shots``` folder but still no luck when triggering ```Refresh``` action...only when stepping out from the project and going back shows the newly created asset aka shot.

https://github.com/ynput/OpenPype/assets/112623825/c689ae26-eefe-4aee-86e9-e4b6c54c993a

LiborBatek (approved):
After setting the right OP package on Ayon, all works fine!

PR #5687, opened 2023-10-03
Maya: Bug in validate Plug-in Path Attribute

antirotor:
## Changelog Description
Overwriting list with string is causing `TypeError: string indices must be integers` in subsequent iterations, crashing the validator plugin.

## Testing notes:
1. Plugin should work as expected


BigRoy (approved):
Yes, 100% correct.

PR #5689, opened 2023-10-03
Delete Old Versions Enhancements

tokejepsen:
## Changelog Description
- Account for using the plugin without any UI.
- Return size for further processing outside of plugin.

## Testing notes:
1. Load the plugin through a python script like below.

```python
os.environ["AVALON_PROJECT"] = project_name
io.install()
project = io.find_one({"type": "project"})

session = ftrack_api.Session()
query = 'select parent.name from Task where type.name is "Compositing" '
query += 'and status.name is "Delivered" '
query += 'and project.name is "{}"'.format(project_name)
shot_names = []
for task in session.query(query):
    shot_name = task["parent"]["name"]
    if shot_name not in shot_names:
        shot_names.append(shot_name)

contexts = []
assets = []
for shot_name in shot_names:
    asset = io.find_one({"type": "asset", "name": shot_name})

    if asset is None:
        print("Could not find asset for shot name: {}".format(shot_name))
        continue

    assets.append(asset)

    for subset in io.find({"type": "subset", "parent": asset["_id"]}):
        # Ignore workfiles.
        if "workfile" in subset["name"].lower():
            continue

        contexts.append(
            {
                "subset": subset,
                "asset": asset,
                "project": {"name": project_name}
            }
        )

register_loader_plugin_path(
    os.path.join(os.path.dirname(openpype.__file__), "plugins", "load")
)
plugins = get_loaders_by_name()
plugin = plugins["CalculateOldVersions"]()
if delete_files:
    plugin = plugins["DeleteOldVersions"]()

size = plugin.load(
    contexts,
    options={"versions_to_keep": 1, "remove_publish_folder": True}
)
```

### DIFF
--- a/openpype/plugins/load/delete_old_versions.py
+++ b/openpype/plugins/load/delete_old_versions.py
@@ -468,6 +468,8 @@ class DeleteOldVersions(load.SubsetLoaderPlugin):
         except Exception:
             self.log.error("Failed to delete versions.", exc_info=True)
 
+        return size
+
 
 class CalculateOldVersions(DeleteOldVersions):
     """Calculate file size of old versions"""
--- a/openpype/plugins/load/delete_old_versions.py
+++ b/openpype/plugins/load/delete_old_versions.py
@@ -5,7 +5,6 @@ import uuid
 import clique
 from pymongo import UpdateOne
 import qargparse
-from qtpy import QtWidgets, QtCore
 
 from openpype import style
 from openpype.client import get_versions, get_representations
@@ -185,6 +184,14 @@ class DeleteOldVersions(load.SubsetLoaderPlugin):
         return size
 
     def message(self, text):
+        from qtpy import QtWidgets, QtCore
+
+        if not QtWidgets.QApplication.instance():
+            self.log.debug(
+                "No QApplication found to show text with:\n{}" + text
+            )
+            return
+
         msgBox = QtWidgets.QMessageBox()
         msgBox.setText(text)
         msgBox.setStyleSheet(style.load_stylesheet())


openpype/plugins/load/delete_old_versions.py
```diff
@@ -185,6 +184,14 @@ def delete_only_repre_files(self, dir_paths, file_paths, delete=True):
         return size
 
     def message(self, text):
+        from qtpy import QtWidgets, QtCore
+
+        if not QtWidgets.QApplication.instance():
+            self.log.debug(
+                "No QApplication found to show text with:\n{}" + text
+            )
+            return
```
iLLiCiTiT, line 193: Considering you've moved the imports here, you wanted to avoid import error.
```suggestion
        qt_app_running = False
        try:
            from qtpy import QtWidgets, QtCore

            qt_app_running = QtWidgets.QApplication.instance()
        except Exception:
                pass

        if not qt_app_running:
            self.log.debug(
                "No QApplication found to show text with:\n{}" + text
            )
            return
```

iLLiCiTiT:
I do understand the reasons, and I'm ok with merging it, but when the functionality is needed to be headless, it would make more sense to separate the logic from the plugin, so it can be used out of the loader plugin.

tokejepsen:
> I do understand the reasons, and I'm ok with merging it, but when the functionality is needed to be headless, it would make more sense to separate the logic from the plugin, so it can be used out of the loader plugin.

Hmm, good point. Actually just ran into an issue where running it through Deadine will show the Qt message cause the workers have the QApplication available.

Will have a look at refactor to some library code.

BigRoy:
> I do understand the reasons, and I'm ok with merging it, but when the functionality is needed to be headless, it would make more sense to separate the logic from the plugin, so it can be used out of the loader plugin.

Agree with this - this is the more sensible approach. Turn the main logic in a simple API - and have the loader just use that with a nice UI to the user.

mkolar:
Because we're splitting OpenPype into ayon-core and individual host addons, this PR would have to be re-created to target one of those. The PR has merit, the solution, not so much. Let's try to revisit in AYON

We're closing it down, but we'll he happy for a new PR to [ynput/ayon-core](https://github.com/ynput/ayon-core) repository once it's up.

PR #5691, opened 2023-10-04
Resolve: Add save current file button + "Save" shortcut when menu is active

### LINKED ISSUE #5684: Enhancement: Resolve allow to easily 'save current file' without incrementing

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

A downside of Resolve is that it works with a database of projects locally instead of just a project file. The OpenPype integration's workfiles works around that by working with Project exports to `.drp` files upon saving/loading with the workfiles tool.

Unfortunately, due to that design, whenever pressing CTRL+S inside the host you aren't actually saving into the project's Workfile `.drp` but into Resolve's database local to your machine.

The Work Files tool only allows to save into a new version. As such there's no OpenPype way to save into the existing version in the project (unless a user would manually do "Export Project" on top of that file using `File > Export Project (CTRL+E)`.

### How would you imagine the implementation of the feature?

I am not sure what the _best_ way would be that would make this feel super native to Resolve itself - preferably clicking CTRL+S would just saving into the workfile's `.drp` as fluently as possible.

I've thought of:

- Add a "Save current file" to the pop-up OpenPype menu which does not increment but just re-exports into the current project name. This is likely the easiest to implement.
- See if there's any way we can listen to "Save callbacks/events" in Resolve so that we can trigger the re-export on CTRL+S. For this we'd need to be able to connect to Resolve callbacks/events which hasn't been done yet and likely isn't trivial (looking at e.g. Fusion it might involve quite a bit of code.

- [X] I have added the relevant labels to the enhancement request.

### Describe alternatives you've considered:

n/a

### Additional context:

_No response_

[cuID:[OP-7101](https://app.clickup.com/t/86bw4gg4g)]

BigRoy:
## Changelog Description

Adds a "Save current file" to the OpenPype menu.

Also adds a "Save" shortcut key sequence (CTRL+S on Windows) to the button, so that clicking CTRL+S when the menu is active will save the current workfile. However this of course does not work if the menu does not receive the key press event (e.g. when Resolve UI is active instead)

Resolves #5684 

## Additional info

image

> A downside of Resolve is that it works with a database of projects locally instead of just a project file. The OpenPype integration's workfiles works around that by working with Project exports to .drp files upon saving/loading with the workfiles tool.
> 
> Unfortunately, due to that design, whenever pressing CTRL+S inside the host you aren't actually saving into the project's Workfile .drp but into Resolve's database local to your machine.

As such, this PR solves that by adding a Save current file button.

## Testing notes:

1. Use OpenPype menu in Resolve
2. Trigger Save Current File


Minkiu (approved):

openpype/hosts/resolve/api/menu.py
```diff
@@ -111,6 +119,15 @@ def __init__(self, *args, **kwargs):
         # reset_resolution_btn.clicked.connect(self.on_set_resolution_clicked)
         experimental_btn.clicked.connect(self.on_experimental_clicked)
 
+    def on_save_current_clicked(self):
+        host = registered_host()
+        current_file = host.current_file()
+        if not current_file:
```
Minkiu, line 125: Maybe this check could be done in the UI building and disable the button altogether

BigRoy: It's tricky that there's no callbacks coming from Resolve currently and the UI is intended to remain open over time - so it'd be hard to detect when to disable/enable the button unfortunately. It would definitely have been nicer :)

But if we have _save callbacks_ we could always then start sneaking in the "save current file" logic there so it also always exports to the project. Making the button redundant. But alas, we're not there yet.

jakubjezek001 (approved):
It works well, but UX wise it might be confusing for some and it would be good to implement some warning for cases described here  https://github.com/ynput/OpenPype/pull/5691/files#r1373359414

openpype/hosts/resolve/api/menu.py
BigRoy: @jakubjezek001 could you try again to see if it's better now?

jakubjezek001 (approved):
LGTM! thanks @BigRoy

PR #5693, opened 2023-10-04
Maya: Optional Resolution Validator for Render

moonyuet:
## Changelog Description
Adding optional resolution validator for maya in render family, similar to the one in Max.
It checks if the resolution in render setting aligns with that in setting from the db.

## Additional info
n/a

## Testing notes:
1. Launch Maya via launcher
2. Create Render Instance
3. Choose whether enable the resolution validator or not
image

4. Publish


openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,54 @@
+import pyblish.api
+from openpype.pipeline import (
+    PublishValidationError,
+    OptionalPyblishPluginMixin
+)
+from maya import cmds
+from openpype.hosts.maya.api.lib import reset_scene_resolution
+
+
+class ValidateResolution(pyblish.api.InstancePlugin,
```
moonyuet, line 10: I am thinking if we can possibly have an abstract class for this validator as the code looks super similar to the Max Hosts. It can be used across different hosts eventually.

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,54 @@
...
+    hosts = ["maya"]
+    label = "Validate Resolution"
+    optional = True
+
+    def process(self, instance):
+        if not self.is_active(instance.data):
+            return
+        width, height = self.get_db_resolution(instance)
+        current_width = cmds.getAttr("defaultResolution.width")
+        current_height = cmds.getAttr("defaultResolution.height")
```
BigRoy, line 25: This is wrong - it should actually check the value **in** the renderlayer because there might be renderlayer overrides. Look at `lib.get_attr_in_layer`.

I believe the layer would be `layer = instance.data["renderlayer"]`
So this would be:
```python
layer = instance.data["renderlayer"]
current_width = lib.get_attr_in_layer("defaultResolution.width", layer=layer)
current_height = lib.get_attr_in_layer("defaultResolution.height", layer=layer)
```

Then in comes the next complication, because V-Ray doesn't use the width/height from `defaultResolution` but from its own `VRaySettingsNode`. Have a look at `lib_rendersettings.RenderSettings` class. Here's [the V-Ray part](https://github.com/ynput/OpenPype/blob/develop/openpype/hosts/maya/api/lib_rendersettings.py#L341-L342) setting the values.

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,54 @@
...
+                                         "set on asset or shot.")
+        if current_width != width:
+            raise PublishValidationError("Width in Resolution Setting "
+                                         "not matching resolution set "
+                                         "on asset or shot.")
+
+        if current_height != height:
+            raise PublishValidationError("Height in Resolution Setting "
+                                         "not matching resolution set "
+                                         "on asset or shot.")
```
BigRoy, line 38: Each of these reports should be showing the current value **and** the expected value so that the user knows what the differences are!

I also don't see a reason to really create a fully different report for which attribute mismatches. I feel like we can just log:

> Render resolution is 1920x1080 does not match asset resolution is 2100x1900

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,54 @@
...
+                width = data["resolutionWidth"]
+                height = data["resolutionHeight"]
+                return int(width), int(height)
+
+        # Defaults if not found in asset document or project document
+        return 1920, 1080
+
+    @classmethod
+    def repair(cls, instance):
+        return reset_scene_resolution()
```
BigRoy, line 54: This being renderlayers complicates this a bit, because technically there _can_ be renderlayer overrides on a single layer. I suspect what repair should be doing is triggering the repair onto the layer to the best of its capabilities, so that _if_ there's a renderlayer override that it sets it there.

Also here not that render resolution _can_ be a different attribute than just `defaultResolution.width` e.g. like the V-Ray case I commented above.

BigRoy: By the way, **note** that `reset_scene_resolution` **also** sets the Pixel Aspect Ratio which this validator does **not** validate/report about which might be confusing. As such, we might want to add that as a validation in here too so it validates resolution + pixel aspect?

LiborBatek (approved):
The Resoluton validator what it should and correctly validate the resolution when enabled and vice versa.
image

Note: it would be nice to have ```Repair``` btn to set the Resolution directly in ```Publisher``` UI without need to go to OP>Set Resolution menu.

openpype/hosts/maya/plugins/publish/validate_resolution.py
moonyuet: We can have two validators for the resolution, one is for scene resolution and the other is for render resolution.

BigRoy: Not sure whether that's necessary and it'd be a bit confusing because for most renderers scene resolution == render resolution.

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,81 @@
...
+    label = "Validate Resolution"
+    actions = [RepairAction]
+    optional = True
+
+    def process(self, instance):
+        if not self.is_active(instance.data):
+            return
+        width, height, pixelAspect = self.get_db_resolution(instance)
+        current_renderer = cmds.getAttr(
+            "defaultRenderGlobals.currentRenderer")
```
BigRoy, line 28: This one should also be acquired from the renderlayer, but you're in luck - it already is collected. I believe it's `instance.data["renderer"]`.

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,81 @@
...
+            vray_node = "vraySettings"
+            if cmds.objExists(vray_node):
+                control_node = vray_node
+                current_width = lib.get_attr_in_layer(
+                    "{}.width".format(control_node), layer=layer)
+                current_height = lib.get_attr_in_layer(
+                    "{}.height".format(control_node), layer=layer)
+                current_pixelAspect = lib.get_attr_in_layer(
+                    "{}.pixelAspect".format(control_node), layer=layer
+                )
```
BigRoy, line 40:
```suggestion
                current_width = lib.get_attr_in_layer(
                    "{}.width".format(vray_node), layer=layer)
                current_height = lib.get_attr_in_layer(
                    "{}.height".format(vray_node), layer=layer)
                current_pixelAspect = lib.get_attr_in_layer(
                    "{}.pixelAspect".format(vray_node), layer=layer)
```

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,81 @@
...
+            else:
+                raise PublishValidationError(
+                    "Can't set VRay resolution because there is no node "
+                    "named: `%s`" % vray_node)
```
BigRoy, line 44:
```suggestion
                raise PublishValidationError(
                    "Can't detect VRay resolution because there is no node "
                    "named: `{}`".format(vray_node))
```

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,81 @@
...
+        asset_doc = instance.data["assetEntity"]
+        project_doc = instance.context.data["projectEntity"]
+        for data in [asset_doc["data"], project_doc["data"]]:
+            if "resolutionWidth" in data and (
+                "resolutionHeight" in data and "pixelAspect" in data
+            ):
+                width = data["resolutionWidth"]
+                height = data["resolutionHeight"]
+                pixelAspect = data["pixelAspect"]
+                return int(width), int(height), int(pixelAspect)
```
BigRoy, line 74: Pixel aspect is not an integer, right? It's a float I presume?
```suggestion
                return int(width), int(height), float(pixelAspect)
```

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,81 @@
...
+
+        # Defaults if not found in asset document or project document
+        return 1920, 1080, 1
```
BigRoy, line 77:
```suggestion
        return 1920, 1080, 1.0
```

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @classmethod
+    def repair(cls, instance):
+        return reset_scene_resolution()
```
BigRoy, line 81: Issue still exists that this does not repair the width/height if the renderlayer were using a renderlayer override on resolution.

However, we _can_ create an issue for that so we can track this 'bug' for later and get this first pass of the feature merged in for community testing.

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,81 @@
...
+    PublishValidationError,
+    OptionalPyblishPluginMixin
+)
+from maya import cmds
+from openpype.pipeline.publish import RepairAction
+from openpype.hosts.maya.api import lib
+from openpype.hosts.maya.api.lib import reset_scene_resolution
+
+
+class ValidateSceneResolution(pyblish.api.InstancePlugin,
```
BigRoy, line 12: Can we add project settings for this plug-in so studios can tweak the enabled, optional, active state from settings?

m-u-r-p-h-y:
> The Resoluton validator does what it should and correctly validate the resolution when enabled and vice versa.
> Note: it would be nice to have `Repair` btn to set the Resolution directly in `Publisher` UI without need to go to OP>Set Resolution menu.

Is it true for all major renderers we support? They tend to hijack the render globals and use their own custom attributes . . . 

Arnold
Redshift
RfM
Vray

(3Delight)

BigRoy:
> > The Resoluton validator does what it should and correctly validate the resolution when enabled and vice versa.
> > Note: it would be nice to have `Repair` btn to set the Resolution directly in `Publisher` UI without need to go to OP>Set Resolution menu.
> 
> Is it true for all major renderers we support? They tend to hijack the render globals and use their own custom attributes . . .
> 
> Arnold Redshift RfM Vray
> 
> (3Delight)

If not supported, then it's a missing abstraction for that renderer in [`lib_rendersettings`](https://github.com/ynput/OpenPype/blob/3daa0749d1a40eb0c22214fb69cc5ef76965b65d/openpype/hosts/maya/api/lib_rendersettings.py) and should be implemented in a separate PR - it'd be unrelated to this particular validator.

Looking at that file it's at least supporting:
image

openpype/hosts/maya/plugins/publish/validate_resolution.py
moonyuet: maybe we can build some new function fix it by referencing` lib.get_attr_in_layer`

BigRoy: Well, the logic to reset scene resolution can grow complicated across multiple renderers since each has their own settings, etc. - That's why we have abstracted that into the `lib_rendersettings` so there's a hopefully somewhat ok API to reset all different renders, etc.

Anyway, I believe however that `reset_scene_resolution` just sets the current value. If you're **inside a renderlayer with a renderlayer override** I believe setting the value equals to updating the layer override value.

We could potentially use [this function `get_attr_overrides`](https://github.com/ynput/OpenPype/blob/a93b63561814ed2d6bed0b97503b3f51acd2c757/openpype/hosts/maya/api/lib_rendersetup.py#L186) to detect if a layer has overrides and if so switch to the renderlayer.

**pseudocode**

```python
attr = "attr"
current_layer = "x"
layer = "y"

# We are in the renderlayer - we can just set the value directly
# even if it has a layer override
if current_layer == layer:
    # Just reset since we're in the layer
    reset_scene_resolution()
    return

# If the layer has an override or the current layer has override
# then we must switch to the layer to ensure we change the
# correct value that affects that layer
if (
    get_attr_overrides(attr, layer=layer) or
    get_attr_overrides(attr, layer=current_layer)
):
    with lib.renderlayer(layer):
        reset_scene_resolution()
        return

# We are not in the same layer but both don't have layer overrides
# so whatever we set in our current layer will affect that layer too
reset_scene_resolution()
```

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,99 @@
...
+                "defaultResolution.height", layer=layer)
+            current_pixelAspect = lib.get_attr_in_layer(
+                "defaultResolution.pixelAspect", layer=layer
+            )
+        if current_width != width or current_height != height:
+            invalid = self.log.error(
+                "Render resolution {}x{} does not match asset resolution {}x{}".format(         # noqa:E501
+                    current_width, current_height,
+                    width, height
+                ))
```
BigRoy, line 70: I don't think a log function returns the value - right? 
```python
import logging
log = logging.getLogger(__name__)
print(log.info("hello"))
# __main__ : hello
None
```

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,99 @@
...
+                    "{}.height".format(vray_node), layer=layer)
+                current_pixelAspect = lib.get_attr_in_layer(
+                    "{}.pixelAspect".format(vray_node), layer=layer
+                )
+            else:
+                invalid = self.log.error(
+                    "Can't detect VRay resolution because there is no node "
+                    "named: `{}`".format(vray_node)
+                )
+                invalids.append(invalid)
```
BigRoy, line 56:
```suggestion
                self.log.error(
                    "Can't detect VRay resolution because there is no node "
                    "named: `{}`".format(vray_node)
                )
                return True
```

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,99 @@
...
+                    width, height
+                ))
+            invalids.append("{0}\n".format(invalid))
+        if current_pixelAspect != pixelAspect:
+            invalid = self.log.error(
+                "Render pixel aspect {} does not match asset pixel aspect {}".format(            # noqa:E501
+                    current_pixelAspect, pixelAspect
+                ))
+            invalids.append("{0}\n".format(invalid))
+        return invalids
```
BigRoy, line 78:
```suggestion
        invalid = False
        if current_width != width or current_height != height:
            self.log.error(
                "Render resolution {}x{} does not match asset resolution {}x{}".format(         # noqa:E501
                    current_width, current_height,
                    width, height
                ))
            invalid = True
        if current_pixelAspect != pixelAspect:
            self.log.error(
                "Render pixel aspect {} does not match asset pixel aspect {}".format(            # noqa:E501
                    current_pixelAspect, pixelAspect
                ))
            invalid = True
        return invalid
```

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,100 @@
...
+                return int(width), int(height), float(pixelAspect)
+
+        # Defaults if not found in asset document or project document
+        return 1920, 1080, 1.0
+
+    @classmethod
+    def repair(cls, instance):
+        layer = instance.data["renderlayer"]
+        with lib.renderlayer(layer):
+            reset_scene_resolution()
```
BigRoy, line 100:
```suggestion
        # Usually without renderlayer overrides the renderlayers
        # all share the same resolution value - so fixing the first
        # will have fixed all the others too. It's much faster to
        # check whether it's invalid first instead of switching
        # into all layers individually
        if not self.get_invalid_resolution(instance):
            self.log.debug(
                "Nothing to repair on instance: {}".format(instance)
            )
            return
        layer = instance.data["renderlayer"]
        with lib.renderlayer(layer):
            reset_scene_resolution()
```

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,100 @@
...
+            invalid = True
+        return invalid
+
+    def get_db_resolution(self, instance):
+        asset_doc = instance.data["assetEntity"]
+        project_doc = instance.context.data["projectEntity"]
+        for data in [asset_doc["data"], project_doc["data"]]:
+            if "resolutionWidth" in data and (
+                "resolutionHeight" in data and "pixelAspect" in data
+            ):
```
BigRoy, line 87:
```suggestion
            if (
                "resolutionWidth" in data and
                "resolutionHeight" in data and 
                "pixelAspect" in data
            ):
```

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,100 @@
...
+                    "{}.height".format(vray_node), layer=layer)
+                current_pixelAspect = lib.get_attr_in_layer(
+                    "{}.pixelAspect".format(vray_node), layer=layer
+                )
+            else:
+                self.log.error(
+                    "Can't detect VRay resolution because there is no node "
+                    "named: `{}`".format(vray_node)
+                )
+                invalid = True
```
BigRoy, line 55: If not returning directly here then `current_width` and `current_height`, etc. are still undefined variables down below
```suggestion
                return True
```
Since we know it's invalid and we can't check resolution then anyway, let's return "invalid" directly.

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,100 @@
...
+        if not self.is_active(instance.data):
+            return
+        invalid = self.get_invalid_resolution(instance)
+        if invalid:
+            raise PublishValidationError(
+                "issues occurred", description=(
+                    "Wrong render resolution setting. "
+                    "Please use repair button to fix it.\n"
+                    "If current renderer is V-Ray, "
+                    "make sure vraySettings node has been created"))
```
BigRoy, line 33: Haha, I know this was my pseudocode example - but we can be slightly more descriptive in the message itself.
```suggestion
            raise PublishValidationError(
                "Render resolution is invalid. See log for details.", 
                description=(
                    "Wrong render resolution setting. "
                    "Please use repair button to fix it.\n"
                    "If current renderer is V-Ray, "
                    "make sure vraySettings node has been created"
                )
            )
```

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,100 @@
...
+from openpype.pipeline.publish import RepairAction
+from openpype.hosts.maya.api import lib
+from openpype.hosts.maya.api.lib import reset_scene_resolution
+
+
+class ValidateResolution(pyblish.api.InstancePlugin,
+                         OptionalPyblishPluginMixin):
+    """Validate the render resolution setting aligned with DB"""
+
+    order = pyblish.api.ValidatorOrder - 0.01
```
BigRoy, line 16: Out of curiosity, is there are reason we check this a validator order - 0.01? Why not just `ValidatorOrder`?

BigRoy: Great - let's remove it.
```suggestion
    order = pyblish.api.ValidatorOrder
```

BigRoy (approved):
Code looks good to me.

LiborBatek (approved):
Works ok now.
image

moonyuet:
@antirotor, this can be merged if you think it's okay.

openpype/hosts/maya/plugins/publish/validate_resolution.py
```diff
@@ -0,0 +1,117 @@
...
+    def process(self, instance):
+        if not self.is_active(instance.data):
+            return
+        invalid = self.get_invalid_resolution(instance)
+        if invalid:
+            raise PublishValidationError(
+                "Render resolution is invalid. See log for details.",
+                description=(
+                    "Wrong render resolution setting. "
+                    "Please use repair button to fix it.\n"
```
BigRoy, line 32:
```suggestion
                    "Please use repair button to fix it.\n\n"
```

moonyuet:
Should be resolved after the adjustment @BigRoy 
image

antirotor (approved):

PR #5695, opened 2023-10-04
Nuke: minor docstring and code tweaks for ExtractReviewMov

moonyuet:
## Changelog Description
Code and docstring tweaks on https://github.com/ynput/OpenPype/pull/5623

## Additional info
n/a

## Testing notes:
1. add `png` to the settings `project_settings/nuke/publish/ExtractReviewDataBakingStreams/outputs/baking/extension`
2. publish review from nuke

### DIFF
--- a/openpype/hosts/nuke/api/plugin.py
+++ b/openpype/hosts/nuke/api/plugin.py
@@ -819,6 +819,11 @@ class ExporterReviewMov(ExporterReview):
             self.file = "{}{}.{}".format(
                 self.fhead, self.name, self.ext)
         else:
+            # Output is image (or image sequence)
+            # When the file is an image it's possible it
+            # has extra information after the `fhead` that
+            # we want to preserve, e.g. like frame numbers
+            # or frames hashes like `####`
             filename_no_ext = os.path.splitext(
                 os.path.basename(self.path_in))[0]
             after_head = filename_no_ext[len(self.fhead):]
--- a/openpype/hosts/nuke/api/plugin.py
+++ b/openpype/hosts/nuke/api/plugin.py
@@ -39,7 +39,7 @@ from .lib import (
     get_view_process_node,
     get_viewer_config_from_string,
     deprecated,
-    get_head_filename_without_hashes,
+    prepend_name_before_hashed_frame,
     get_filenames_without_hash
 )
 from .pipeline import (
@@ -820,12 +820,12 @@ class ExporterReviewMov(ExporterReview):
         if ".{}".format(self.ext) not in VIDEO_EXTENSIONS:
             # filename would be with frame hashes if
             # the file extension is not in video format
-            filename = get_head_filename_without_hashes(
+            filename = prepend_name_before_hashed_frame(
                 self.path_in, self.name)
             self.file = filename
             # make sure the filename are in
             # correct image output format
-            if ".{}".format(self.ext) not in self.file:
+            if not self.file.endswith(".{}".format(ext)):
                 filename_no_ext, _ = os.path.splitext(filename)
                 self.file = "{}.{}".format(filename_no_ext, self.ext)
 
--- a/openpype/hosts/nuke/api/plugin.py
+++ b/openpype/hosts/nuke/api/plugin.py
@@ -39,7 +39,6 @@ from .lib import (
     get_view_process_node,
     get_viewer_config_from_string,
     deprecated,
-    prepend_name_before_hashed_frame,
     get_filenames_without_hash
 )
 from .pipeline import (
@@ -818,15 +817,14 @@ class ExporterReviewMov(ExporterReview):
 
         self.file = self.fhead + self.name + ".{}".format(self.ext)
         if ".{}".format(self.ext) not in VIDEO_EXTENSIONS:
-            # filename would be with frame hashes if
-            # the file extension is not in video format
-            filename = prepend_name_before_hashed_frame(
-                self.path_in, self.name)
-            self.file = filename
+            filename = os.path.basename(self.path_in)
+            self.file = self.fhead + self.name + ".{}".format(
+                filename.split(".", 1)[-1]
+            )
             # make sure the filename are in
             # correct image output format
             if not self.file.endswith(".{}".format(ext)):
-                filename_no_ext, _ = os.path.splitext(filename)
+                filename_no_ext, _ = os.path.splitext(self.file)
                 self.file = "{}.{}".format(filename_no_ext, self.ext)
 
         self.path = os.path.join(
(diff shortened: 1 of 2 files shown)


openpype/hosts/nuke/api/plugin.py
```diff
@@ -820,12 +820,12 @@ def __init__(self,
         if ".{}".format(self.ext) not in VIDEO_EXTENSIONS:
             # filename would be with frame hashes if
             # the file extension is not in video format
-            filename = get_head_filename_without_hashes(
+            filename = prepend_name_before_hashed_frame(
```
BigRoy, line 823: I still think the code is weird.

The class inherits from [`ExporterReview`](https://github.com/ynput/OpenPype/blob/4a2417d2ca741c6b5a8db81b042e0bfff7a4d12a/openpype/hosts/nuke/api/plugin.py#L559) which already computes `self.fhead` like is done outside of this logic to define `self.fhead + self.name + ...` etc.

Doesn't that just mean we should somehow be fixing the logic [here](https://github.com/ynput/OpenPype/blob/4a2417d2ca741c6b5a8db81b042e0bfff7a4d12a/openpype/hosts/nuke/api/plugin.py#L585-L609)?

What's the reason we can't rely on that `self.fhead`? Doesn't this just mean that the design of `ExporterReview` class is buggy to begin with and doesn't support sequences (or at least not paths with hashes - and if it's just about the hashes, _why_ are we passing the filepath with hashes to begin with).

jakubjezek001 (approved):
LGTM!

openpype/hosts/nuke/api/plugin.py
```diff
@@ -818,15 +817,14 @@ def __init__(self,
...
         if ".{}".format(self.ext) not in VIDEO_EXTENSIONS:
-            # filename would be with frame hashes if
-            # the file extension is not in video format
-            filename = get_head_filename_without_hashes(
-                self.path_in, self.name)
-            self.file = filename
+            filename = os.path.basename(self.path_in)
+            self.file = self.fhead + self.name + ".{}".format(
+                filename.split(".", 1)[-1]
+            )
```
BigRoy, line 823: This would still fail with files that contain a `dot` before the hashes or frame numbers.

```
hello.world.####.exr
hello.world.0001.exr
```
Would now both make:
```
hello.world.{name}.world
```

Also, with this logic we now know for a fact that `self.file` will not have the extension since we creating the filename ourselves without the extension

so the logic below would actually be splitting of `.world` and then replace that with the extension so we end up with `hello.world.{name}.{ext}` which does not have the frame number or hashes which would not work for image based outputs?

moonyuet: maybe we can try to use `replace `here for the files(although it is not likeable)
`self.file =self.fname.replace(self.fhead, self.fhead+self.name)`

openpype/hosts/nuke/api/plugin.py
```diff
@@ -818,15 +817,13 @@ def __init__(self,
...
         self.file = self.fhead + self.name + ".{}".format(self.ext)
         if ".{}".format(self.ext) not in VIDEO_EXTENSIONS:
-            # filename would be with frame hashes if
-            # the file extension is not in video format
-            filename = get_head_filename_without_hashes(
-                self.path_in, self.name)
-            self.file = filename
+            filename = os.path.basename(self.path_in)
+            self.file = re.sub(
+                self.fhead, self.fhead + self.name + ".", filename)
```
BigRoy, line 822: `self.fhead` does currently not include the frame number or frame hashing as far as I could tell from the logic in [`get_file_info`](https://github.com/ynput/OpenPype/blob/13b46070fe9fef0b6474a50e9cebb7d7b73eac43/openpype/hosts/nuke/api/plugin.py#L584-L608) - is that what you intended to produce here?

Also, if you use regex on a name/pattern you want to exactly match use `re.escape()` on the input because otherwise e.g. `hello.world` would also match `hello_world` or `hellooworld` since the `.` means any character. 

Could you explain what, in your eyes the different variables should be?
```
self.path_in = "/path/to/hello.####.exr"
self.fhead = "hello."
self.name = "baking"
filename = `hello.####.exr`
```
Does this seem about right?

If so, this is likely your best shot to 'keeping whatever is in between' head and extension.
```python
# Strip off the head but preserve as much of the remainder of the filename
filename = filename[len(self.fhead):]
```
Better:
```python
filename_no_ext = os.path.splitext(os.path.basename(self.path_in))[0]
after_head = filename_no_ext[len(self.fhead):]
self.file = self.fhead + self.name + after_head + self.ext
```

And this logic I suspect might even hold true for both video files and image files, for video files `after_head` would just be an empty string.

This is just quickly out of the top of my head.

openpype/hosts/nuke/api/plugin.py
```diff
@@ -818,17 +817,11 @@ def __init__(self,
...
-            # correct image output format
-            if ".{}".format(self.ext) not in self.file:
-                filename_no_ext, _ = os.path.splitext(filename)
-                self.file = "{}.{}".format(filename_no_ext, self.ext)
-
+            filename_no_ext = os.path.splitext(
+                os.path.basename(self.path_in))[0]
+            after_head = filename_no_ext[len(self.fhead):]
+            self.file = "{}{}.{}.{}".format(
+                self.fhead, self.name, after_head, self.ext)
```
BigRoy, line 824:
```suggestion
            # When the file is an image it's possible it
            # has extra information after the `fhead` that
            # we want to preserve, e.g. like frame numbers
            # or frames hashes like `####`
            filename_no_ext = os.path.splitext(
                os.path.basename(self.path_in))[0]
            after_head = filename_no_ext[len(self.fhead):]
            self.file = "{}{}.{}.{}".format(
                self.fhead, self.name, after_head, self.ext)
```

openpype/hosts/nuke/api/plugin.py
```diff
@@ -818,17 +817,11 @@ def __init__(self,
 
         self.file = self.fhead + self.name + ".{}".format(self.ext)
```
BigRoy, line 818: Let's match the code style here with what we're doing in the if statement.
```suggestion
        self.file = "{}{}.{}".format(self.fhead, self.name, self.ext)
```

BigRoy: Also, what's the reason we're not doing `if ... else`?

```python
        if ".{}".format(self.ext) in VIDEO_EXTENSIONS:
            # Output is video
            self.file = "{}{}.{}".format(self.fhead, self.name, self.ext)
        else:
            # Output is expected to be image (or image sequence)
            # When the file is an image it's possible it
            # has extra information after the `fhead` that
            # we want to preserve, e.g. like frame numbers
            # or frames hashes like `####`
            filename_no_ext = os.path.splitext(
                os.path.basename(self.path_in))[0]
            after_head = filename_no_ext[len(self.fhead):]
            self.file = "{}{}.{}.{}".format(
                self.fhead, self.name, after_head, self.ext)
```

BigRoy (approved):
Haven't tested the code (don't use Nuke) but the logic seems more sound than before.

jakubjezek001 (approved):
LGTM!

PR #5696, opened 2023-10-04
Tests: Remove checks for env var

kalisp:
## Changelog Description
Env var will be filled in `env_var` fixture, here it is too early to check


iLLiCiTiT:
Is the check sometimes later? It looks like the assertion was there for a long time? https://github.com/ynput/OpenPype/pull/5676 did not change it.

iLLiCiTiT (approved):

